PR #1333, opened 2020-04-09
Set viewed threshold to 18px for remote epic

nicl:
## What does this change?

Set viewed threshold to 18px into the component (previously 10px).

## Why?

I misread the Frontend value here. It's actually 18:

https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/commercial/contributions-utilities.js#L276

n/a


nicl:
thanks @gtrufitt !

tjmw:
Retrospective :+1: on this - good catch @nicl!

PR #1335, opened 2020-04-09
Display double MPU for comment ads slot

shtukas:
## What does this change?

Modify the available sizes for the comment ad slot to only display double MPUs


shtukas:
Screenshot 2020-04-09 at 11 46 34

oliverlloyd:
Be good to show how this looks when there are no or very few comments and the advert creates whitespace. Be great to have something to fill that space with

shtukas:
Yeah, let's see how it displays up when that happens :)

PR #1336, opened 2020-04-09
Use property attribute for opengraph meta tags

jfsoul:
## What does this change?
Uses the `property` attribute in meta tags for opengraph specifically. We believe that is why echobox is using the page title instead of the `og:title` when rendering previews. The facebook opengraph debugger specifically flags that we are not using the property attribute:

image

For extra confidence, this will also match the behaviour of frontend - which has never had these problems.

## Why?
Fixes social shares, particularly ones that rely on opengraph - like hangout chat (see trello card), facebook (see screenshot above) or echobox.

https://trello.com/c/lGkJLv2E/730-implement-open-graph-in-dotcom-rendering-correctly

### DIFF
--- a/src/web/server/document.test.tsx
+++ b/src/web/server/document.test.tsx
@@ -23,6 +23,33 @@ test('that all the required meta SEO fields exist', () => {
     );
 });
 
+test('that all some of the most important opengraph meta tags exist', () => {
+    const names = [
+        'og:url',
+        'og:description',
+        'og:title',
+        'og:type',
+        'og:image',
+        'article:author',
+    ];
+
+    names.map(name =>
+        expect(result).toEqual(
+            expect.stringContaining(`<meta property="${name}"`),
+        ),
+    );
+});
+
+test('that all some of the most important twitter meta tags exist', () => {
+    const names = ['card', 'image', 'site'];
+
+    names.map(name =>
+        expect(result).toEqual(
+            expect.stringContaining(`<meta name="twitter:${name}"`),
+        ),
+    );
+});
+
 test('that all the required links exist', () => {
     const names = ['amphtml'];
 
--- a/src/web/server/document.test.tsx
+++ b/src/web/server/document.test.tsx
@@ -40,7 +40,7 @@ test('that the most important opengraph meta tags exist', () => {
     );
 });
 
-test('that all some of the most important twitter meta tags exist', () => {
+test('that the most important twitter meta tags exist', () => {
     const names = ['card', 'image', 'site'];
 
     names.map(name =>
--- a/src/web/server/document.test.tsx
+++ b/src/web/server/document.test.tsx
@@ -23,7 +23,7 @@ test('that all the required meta SEO fields exist', () => {
     );
 });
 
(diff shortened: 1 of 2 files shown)


jfsoul:
I guess something to flag is that I haven't tested this beyond checking the page source looks good myself, because all the debuggers need to be able to scrape the page AFAIK - so I'd need to do it on CODE.

src/web/server/htmlTemplate.ts
```diff
@@ -84,18 +84,24 @@ export const htmlTemplate = ({
             )}" as="font" crossorigin>`,
     );
 
-    const generateMetaTags = (dataObject: { [key: string]: string }) => {
+    const generateMetaTags = (
+        dataObject: { [key: string]: string },
+        attributeName: 'name' | 'property',
```
jfsoul, line 89: not sure whether this is the idiomatic way to do it...

src/web/server/document.test.tsx
```diff
@@ -23,6 +23,33 @@ test('that all the required meta SEO fields exist', () => {
     );
 });
 
+test('that all some of the most important opengraph meta tags exist', () => {
```
oliverlloyd, line 26:
```suggestion
test('that the most important opengraph meta tags exist', () => {
```

oliverlloyd: Sorry, a pedantic grammar change!

src/web/server/document.test.tsx
```diff
@@ -23,6 +23,33 @@ test('that all the required meta SEO fields exist', () => {
...
+    ];
+
+    names.map(name =>
+        expect(result).toEqual(
+            expect.stringContaining(`<meta property="${name}"`),
+        ),
+    );
+});
+
+test('that all some of the most important twitter meta tags exist', () => {
```
oliverlloyd, line 43:
```suggestion
test('that the most important twitter meta tags exist', () => {
```

src/web/server/htmlTemplate.ts
oliverlloyd: Sometimes I upgrade to:

```
const generateMetaTags = ({
        dataObject,
        attributeName,
}: {
        dataObject: { [key: string]: string },
        attributeName: 'name' | 'property',
}) => ...
```

To guard against property order confusion but it's more verbose. I think you can go both ways here.

jfsoul:
> Just a small thought What do you think about splitting generateMetaTags into two functions, generateOGTags & generateTwitterTags? The downside is repetition but it could make things easier to maintain if there are more use cases like this in future where the requirement branch.

Neither here nor there for me. I've tried your suggestion and I'm leaning towards the original for brevity and to hint at the namespace of meta attributes i.e. `name | property` and not necessarily tie that to twitter or og.

src/web/server/document.test.tsx
```diff
@@ -23,6 +23,33 @@ test('that all the required meta SEO fields exist', () => {
     );
 });
 
+test('that the most important opengraph meta tags exist', () => {
```
gtrufitt, line 26: 🎊

PR #1341, opened 2020-04-12
ArticleTitle Direction

oliverlloyd:
## What does this change?
This adds the logic to inline the article titles below `leftCol`

Note. Inlining always occurs on frontend at this breakpoint but here we add the design decision to _not_ inline the items when a badge is showing on DCR because with the height of the badge present, there's no need and we may as well keep the text stacked

@HarryFischer Just a heads up that we're doing this ☝️ shout if we missed something or you'd rather we didn't, happy to change this to keep parity if that's the design call.

### Before (with badge)
Screenshot 2020-04-12 at 12 50 56

### After (with badge)
Screenshot 2020-04-12 at 12 50 23

### Before (without badge)
Screenshot 2020-04-12 at 12 51 08

### After (without badge)
Screenshot 2020-04-12 at 12 49 59

## Why?
Because at smaller screen widths we want to reduce the amount that we push content down

https://trello.com/c/TKNUhPZP/1113-section-headers-are-inline-below-wide

### DIFF
--- a/src/web/components/ArticleTitle.tsx
+++ b/src/web/components/ArticleTitle.tsx
@@ -64,6 +64,7 @@ export const ArticleTitle = ({
                 guardianBaseURL={guardianBaseURL}
                 pillar={pillar}
                 fallbackToSection={true}
+                badge={badge}
             />
         </div>
     </div>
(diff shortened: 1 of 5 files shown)


HarryFischer:
looks good to me

PR #1342, opened 2020-04-13
Adjust spacing for immersive headlines

oliverlloyd:
## What does this change?
Fixes spacing for an immersive headline

### Before
Screenshot 2020-04-13 at 10 35 58

### After
Screenshot 2020-04-13 at 10 34 31

## Why?
Because we want to handle shorter headlines better

https://trello.com/c/KZ7TdHwq/1439-immersive-headline-styles

### DIFF
--- a/src/web/components/ArticleHeadline.stories.tsx
+++ b/src/web/components/ArticleHeadline.stories.tsx
@@ -385,7 +385,7 @@ export const Live = () => (
 );
 Live.story = { name: 'Live' };
 
-export const Immersive = () => (
+export const LongImmersive = () => (
     <>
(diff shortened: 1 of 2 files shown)


liywjl:
I noticed some of the diffs in chromatic are due to dates in the fixtures, It might be worth generating dates at runtime to avoid legacy styling

oliverlloyd:
> I noticed some of the diffs in chromatic are due to dates in the fixtures, It might be worth generating dates at runtime to avoid legacy styling

Yeah, I think we need to fix the date to a specific point in time. We do the same in discussion. One for another PR

PR #1344, opened 2020-04-13
Immersive images

oliverlloyd:
## What does this change?
Adds the css for images with the `immersive` weighting

Immersive images are stretched to the left and right to fill the main display column. They do not completely fill the viewport the way immersive main media does

Screenshot 2020-04-13 at 21 36 41

### Gutters
On Frontend the margins are pushed an extra 20 pixels on the left and right filling the article gutters, this works for photo essays because the gutters aren't shown but looks broken for other article types. In order to allow support for immersive images over all articles the margins are cropped by 20 pixels here, preserving the gutter.

If we wanted to restore that extra 20 pixels and lose the gutters then we need to solve the caption problem. This is addressed on photo essays by deleting the normal captions and showing special elements but this solution is extremely limited and inflexible.

## Why?
So we support immersive articles and other artilce types that show immersive images

https://trello.com/c/EPmYufDg/1082-immersive-in-body-images

### DIFF
--- a/src/web/components/Caption.tsx
+++ b/src/web/components/Caption.tsx
@@ -81,7 +81,8 @@ export const Caption: React.FC<{
     };
 
     const shouldLimitWidth =
-        !isMainMedia && (role === 'showcase' || role === 'supporting');
+        !isMainMedia &&
+        (role === 'showcase' || role === 'supporting' || role === 'immersive');
 
     return (
         <figure className={figureStyle}>
--- a/src/web/components/elements/ImageBlockComponent.tsx
+++ b/src/web/components/elements/ImageBlockComponent.tsx
@@ -30,25 +30,21 @@ const imageCss = {
     `,
 
     immersive: css`
-        ${from.mobileMedium} {
-            margin-left: -20px;
-            margin-right: -20px;
-        }
         ${from.tablet} {
-            margin-left: -20px;
-            margin-right: -120px;
+            margin-left: 0px;
+            margin-right: -100px;
         }
         ${from.desktop} {
-            margin-left: -20px;
-            margin-right: -340px;
+            margin-left: 0px;
+            margin-right: -320px;
         }
         ${from.leftCol} {
-            margin-left: -180px;
-            margin-right: -340px;
+            margin-left: -160px;
+            margin-right: -320px;
         }
         ${from.wide} {
-            margin-left: -260px;
-            margin-right: -420px;
+            margin-left: -240px;
+            margin-right: -400px;
         }
     `,
 
--- a/src/web/components/elements/ImageBlockComponent.tsx
+++ b/src/web/components/elements/ImageBlockComponent.tsx
@@ -29,8 +29,28 @@ const imageCss = {
         }
     `,
 
-    // TODO: immersive is pending review of different article types
-    immersive: css``,
+    immersive: css`
+        ${from.mobileMedium} {
+            margin-left: -20px;
+            margin-right: -20px;
+        }
+        ${from.tablet} {
+            margin-left: -20px;
+            margin-right: -120px;
+        }
+        ${from.desktop} {
+            margin-left: -20px;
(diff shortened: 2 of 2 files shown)


liywjl:
We dont have a storybook for this, maybe worth adding one?

paperboyo:
Just some random thoughts/archeology:
1. Historically, immersive weighting was introduced for Immersive display hint (when this was the only one). That’s why current frontend ignores immersive weighting on `No display hint`. I think there is strong case for dcr to be better and respect it regardless of display hint.
2. Captions on Photo essays are an ugly hack as the whole essay is (we have at least 3 failed PRs trying to bring styling to post-Garnett). In Photo essays, frontend ignores correct, structured captions from Composer (no idea what hacks apps have here, but at least Daily is completely broken) and instead forces editors to use a bulletted body paragraph to provide a caption. Apart from all frontends having to support this hack (which, again, they do not), this means that sorting this out properly will mean we will have legacy content with stupid body paras. Also, sorting it out properly, probably requires Composer to allow for captioning eg half-width-weighted images with a single caption (initial reason for a hack). I still, naively, think we should sort it out properly, but this will require coordination between all frontends and Tools (if not CAPI too). Oh, boy. There is a [card for that](https://trello.com/c/D4xjei2L/295-with-a-half-width-image-would-it-be-possible-to-make-the-caption-run-at-the-side-of-the-image-as-a-block-rather-than-underneath), by Yors Truly.
3. Not sure what you mean by
   > On Frontend the margins are pushed an extra 20 pixels on the left and right filling the article gutters, this works for photo essays because the gutters aren't shown but looks broken for other article types. In order to allow support for immersive images over all articles the margins are cropped by 20 pixels here, preserving the gutter.

   I can’t see a difference in how immersive image looks at Immersive vs. Photo essay hint (apart from captions, obviously)

4. If I would have more time, I would write a shorter comment, apologies (lemme know if you can’t see trello link too!).

Exciting!

oliverlloyd:
@paperboyo 💯 agree

>     3. Not sure what you mean by

By gutters I mean the 20px padding at the side of the article before the side borders. On photo essays there are no side borders so the fact the image touches the sides doesn't look strange. See below. 

Screenshot 2020-04-13 at 21 22 44

If I make sure to respect that 20px gutter then it means the `immersive` weighting will work in all articles, regardless of display hint

### Captions
I have an idea for how to approach this. It's multi staged.

**Stage 1**
We (DCR) build a new element type, say `MultiImageElementBlock`, and we define the props that it accepts something like:

```typescript
type Props {
    images: ImageElementType[];
    caption: string;
    captionStyle: 'normal' | 'fancy' | 'bold' | 'etc';
}
```

DCR can then use this data to render the 'side by side', or 'grid of four' or even '1 above 2' views that we get now, with a single caption showing for them all.

**Stage 2**
Before we send json to DCR, the server needs to take the existing hack data and convert it to this new format. The same process happens on frontend now - it deletes the captions from iamges and other things - so we just need to change this to a DCR version

**Stage 3**
We get the other rendering tiers to support our new `MultiImageElementBlock`

**Stage 4**
We build a new image insert tool in Composer that supports multi images with a single caption and outputs data in the format expected by `MultiImageElementBlock`

paperboyo:
> On photo essays there are no side borders

I wouldn’t care much for how Photo essay looks, as it’s still pre-Garnett design, because [our](https://github.com/guardian/frontend/pull/19209) [three](https://github.com/guardian/frontend/pull/20975) (sic!) [attempts](https://github.com/guardian/frontend/pull/21008) at bringing the styling up to date (and borders with it), bombed. Let’s go with consistency with other design hints.

I very much like your plan! I hope for some discussion with Tools on how to sort it out properly one day, although, obviously, priorities…

oliverlloyd:
> We dont have a storybook for this, maybe worth adding one?

I'm working on building out the immersive stories now. I'll be sure to make sure we include an immersive image in there 👍

PR #1346, opened 2020-04-13
Only show an avatar when there's one contributor

oliverlloyd:
## What does this change?
Prevents the author avatar from showing when there is more then one contributor to a comment piece

### Before
Screenshot 2020-04-13 at 21 50 47

### After
Screenshot 2020-04-13 at 21 57 40

## Why?
To keep everyone happy

https://trello.com/c/V6tV7HCn/1333-multiple-comment-contributors-hide-image

### DIFF
--- a/src/web/layouts/CommentLayout.tsx
+++ b/src/web/layouts/CommentLayout.tsx
@@ -209,6 +209,10 @@ export const CommentLayout = ({ CAPI, NAV }: Props) => {
 
     const contributorTag = CAPI.tags.find(tag => tag.type === 'Contributor');
     const avatarUrl = contributorTag && contributorTag.bylineImageUrl;
+    const onlyOneContributor: boolean =
+        CAPI.tags.filter(tag => tag.type === 'Contributor').length === 1;
+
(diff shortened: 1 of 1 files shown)


src/web/layouts/CommentLayout.tsx
```diff
@@ -209,6 +209,10 @@ export const CommentLayout = ({ CAPI, NAV }: Props) => {
 
     const contributorTag = CAPI.tags.find(tag => tag.type === 'Contributor');
     const avatarUrl = contributorTag && contributorTag.bylineImageUrl;
+    const onlyOneContributor: boolean =
+        CAPI.tags.filter(tag => tag.type === 'Contributor').length === 1;
```
liywjl, line 213:
```suggestion
        CAPI.tags.filter(tag => tag.type === 'Contributor').length >= 1;
```
in case there is more than one of the same tag (just a precaution)

oliverlloyd: Actually, we only want to return true when the count is exactly 1

PR #1351, opened 2020-04-14
Dynamic imports CMP example

liywjl:
## What does this change?
Implement dynamic imports using `loadable-components`  on the frontend (`react.js`) bundle.

## Why?
A proof of concept of how we can break react components out of the core `react.js` bundle.

Using `@loadable/component` to import a component will automatically split this component out of the `react.js` bundle. `loadable-components` will then fetch that component remotely if the component is trying to be rendered.

Main benefit, fetch the component if it is being used, reducing the bundle size for components that are not often used.

## Note
Q: Why are we not implementing SSR using `loadable-components`?
A: `loadable-components` SSR requires you to render the app server side and allows you to extract script tags for each component/library, giving you the opportunity to load each chunk in parallel on page load.

Our app has 2 sections, SSR-ed html and a bundled `react.js` payload. It makes no sense to do dynamic imports to our server code, as all of it gets used to generate HTML and is never used by the client. Regarding the `react.js` bundle, we would need to render it sever side and use the ChunkExtractor in order to extract script tags (refer to [step 3](https://loadable-components.com/docs/server-side-rendering/) ). However rendering the `react.js` server side would fail as we depend on the window object in several sections of the code and of course window does not exist in node.

This is something we could explore if we remove the window dependency from the bundle.

Q: why was `babel-plugin-dynamic-import-node` removed?
A: it replaced instances of `import` to `require`, thus causing issues to get dynamic imports to work

Q: why has the `CMP` component been changed into a default export?
A: Currently `loadable-components` does not support non default exports of components, there is an active PR to allow non default exports https://github.com/gregberge/loadable-components/pull/483

Q: Can this be used for other components?
A: Yes, here is a breakdown of other components that are conditionally rendered in `App.tsx`
Screenshot 2020-04-15 at 11 04 00

## How to test
You can use local dev environment to test. `CMP` is rendered conditionally using `CAPI.config.cmpUi` in `src/web/components/App.tsx`. As `CAPI.config.cmpUi` is determined on BE, you will need to replace it with a boolean `true` or `false` to see it in action.

When `CMP` is being rendered, you should be able to see an import in the network tab. When it is not, you should not see any `CMP` call in the network tab.

https://trello.com/c/0qCpwbbO/1285-dynamic-imports


src/web/components/CMP.test.tsx
```diff
@@ -35,23 +35,23 @@ describe('CMP', () => {
...
+        const { container } = render(<CMP />);
 
         expect(container.firstChild).toBeNull();
     });
 
     it('It should render null if shouldShow ireturns true but cmpUi switch is false', () => {
         shouldShow.mockImplementation(() => true);
 
-        const { container } = render(<CMP cmpUi={false} />);
+        const { container } = render(<>{false && <CMP />}</>);
```
liywjl, line 46: This test can likely be removed

src/web/components/App.tsx
```diff
@@ -316,9 +318,13 @@ export const App = ({ CAPI, NAV }: Props) => {
                     />
                 </Lazy>
             </Portal>
-            <Portal root="cmp">
-                <CMP cmpUi={CAPI.config.cmpUi} />
-            </Portal>
+            {!CAPI.config.cmpUi && (
+                <Portal root="cmp">
+                    <Suspense fallback={<div>Loading...</div>}>
```
liywjl, line 323: Unsure if we should use `Suspense` here as I assume we don't need a fallback loading state

babel.config.js
```diff
@@ -1,8 +1,8 @@
 module.exports = {
     plugins: [
         '@babel/plugin-transform-react-jsx',
+        'babel-plugin-syntax-dynamic-import',
```
liywjl, line 4: to support webpack dynamic imports

babel.config.js
```diff
@@ -46,6 +47,9 @@ module.exports = {
                 ],
             ],
         },
+        test: {
+            plugins: ['babel-plugin-dynamic-import-node'],
```
liywjl, line 51: in order to allow unit tests to still work when we dynamically load

package.json
```diff
@@ -93,12 +98,15 @@
...
         "@babel/preset-typescript": "^7.3.3",
+        "@loadable/babel-plugin": "^5.12.0",
+        "@loadable/webpack-plugin": "^5.12.0",
         "@storybook/addon-actions": "^5.2.5",
         "@storybook/addon-links": "^5.2.5",
         "@storybook/addon-storysource": "^5.2.5",
         "@storybook/addon-viewport": "^5.2.5",
         "@storybook/addons": "^5.2.5",
         "@storybook/react": "^5.2.5",
+        "@testing-library/dom": "^7.2.1",
```
liywjl, line 109: allows us to use `waitFor` in unit tests

scripts/webpack/browser.js
```diff
@@ -53,46 +50,54 @@ module.exports = ({
...
-        PROD &&
-        new AssetsManifest({
-            writeToDisk: true,
-            assets: isLegacyJS ? legacyManifestData : manifestData,
-            // Need to explicitly define output file names to avoid overwrites
-            output: isLegacyJS ? 'manifest.legacy.json' : 'manifest.json',
+        new LoadablePlugin({
+            filename: isLegacyJS
+                ? 'loadable-stats.browser.legacy.json'
+                : 'loadable-stats.browser.json',
```
liywjl, line 56: this is the only change made to the file, the rest are styling changes from prettier

src/web/components/App.tsx
```diff
@@ -317,7 +317,10 @@ export const App = ({ CAPI, NAV }: Props) => {
                 </Lazy>
             </Portal>
             <Portal root="cmp">
-                <CMP cmpUi={CAPI.config.cmpUi} />
+                <>
+                    {/* need to assure cmpUi is false and not undefined for storybook */}
+                    {CAPI.config.cmpUi === false && <CMP />}
```
oliverlloyd, line 322: This means the CMP component will only render if `cmpUI` is explicitly `false` - is that correct? The naming kind of suggests the opposite

src/web/components/CMP.test.tsx
```diff
@@ -35,24 +36,23 @@ describe('CMP', () => {
     it('It should render null if shouldShow returns false', () => {
         shouldShow.mockImplementation(() => false);
 
-        const { container } = render(<CMP cmpUi={true} />);
+        const { container } = render(<CMP />);
 
         expect(container.firstChild).toBeNull();
     });
 
     it('It should render null if shouldShow ireturns true but cmpUi switch is false', () => {
```
oliverlloyd, line 44: This test implies the component should not render if `cmpUI` is false

src/web/components/CMP.tsx
```diff
@@ -1,35 +1,43 @@
...
-};
+import loadable from '@loadable/component';
+
+const ConsentManagementComponent = loadable.lib(() =>
+    import(
+        '@guardian/consent-management-platform/lib/ConsentManagementPlatform'
+    ),
+);
 
-export const CMP = ({ cmpUi }: Props) => {
```
oliverlloyd, line 12: I feel like we shouls be using `cmpUI` here to block the import

src/web/components/CMP.tsx
```diff
@@ -1,35 +1,43 @@
...
         return null;
     }
 
-    return <ConsentManagementPlatform source="dcr" onClose={onClose} />;
+    return (
+        <ConsentManagementComponent>
+            {({ ConsentManagementPlatform }) => (
+                <ConsentManagementPlatform source="dcr" onClose={onClose} />
+            )}
+        </ConsentManagementComponent>
```
oliverlloyd, line 41: Render props pattern?

oliverlloyd:
> # New build sizes
> ## Before
> 
> Screenshot 2020-04-15 at 14 49 30
> ## After
> 
> Screenshot 2020-04-15 at 14 54 10
> 
> `react bunde 264.43kb => 198.2kb` 🎊

Do you know why there are 3 separate bundles for CMP in the build images above?

src/web/components/CMP.tsx
```diff
@@ -1,35 +1,43 @@
...
-            setShow(true);
-
-            // setErrorHandler takes function to be called on errors in the CMP UI
-            setErrorHandler((errMsg: string): void => {
-                const err = new Error(errMsg);
-
-                window.guardian.modules.sentry.reportError(err, 'cmp');
-            });
-        }
+        import(
```
gtrufitt, line 14: Just for my understanding (and prosperity) - this is a plain [dynamic import](https://webpack.js.org/guides/code-splitting/#dynamic-imports) causing the code split on 'vendors' for consent platform (answering @oliverlloyd's question on why it's split into three).

This is what we can do to dynamically import and split libraries vs using `loadable` which is basically some nice sugar and API around dynamic imports (and thus Webpack treats both the same RE code splitting?)

src/web/components/CMP.tsx
gtrufitt, line 41: You're using the [`lib splitting`](https://loadable-components.com/docs/library-splitting/) here, can you not just use the standard [`component splitting`](https://loadable-components.com/docs/component-splitting/) in this case for the same goal?

My understanding from the docs is that library splitting is useful for things like utils, but in this case you're importing a component?

src/web/components/CMP.tsx
gtrufitt: Although there are methods in loadable to import `lib` files - when would we use plain dynamic imports vs loadable?

src/web/components/App.tsx
GHaberis: Yes, the value of `CAPI.config.cmpUi` should be `true` to show the `<CMP />`, it is derived from the `cmpUi` feature switch value, which is exposed at `window.guardian.config.switches.cmpUi` on `frontend`.

liywjl: yes, this is only rendered when `CAPI.config.cmpUi` is false, idk why this was decided, you can look in the old `CMP` and it returns null when `! cmpUi`

src/web/components/CMP.test.tsx
liywjl: yes I should remove that

src/web/components/CMP.tsx
liywjl: as in use `cmpUi` to dynamically import the component? This is what I originally did but did not see a substantial decrease in the bundle size. Thus why I ended up lazy loading the lib

src/web/components/CMP.tsx
liywjl: yeah this is how you can achieve lazy lib loading https://loadable-components.com/docs/library-splitting/

src/web/components/App.tsx
GHaberis: Just took a look and it was previously returning `null` not the component if `!cmpUi`

src/web/components/CMP.tsx
liywjl: yes this is the classic webpack dynamic import, I was planning on using exclusively `loadable-components` but as you can see I am importing functions and not react components.

In `loadable-components` you can split out components using the classic way:
https://loadable-components.com/docs/component-splitting/

or split out libraries using HOCs:
https://loadable-components.com/docs/library-splitting/

I cannot use either within `useEffect` so I decided to instead implement webpack's dynamic import. I think both are ok but personally if it is not a react component I think webpack's dynamic imports are preferable

src/web/components/CMP.tsx
liywjl: so its a little more complex, there are 2 main reasons:

- This component originates from a lib https://github.com/guardian/dotcom-rendering/pull/1351/files#diff-90ded79e2aa2d6deb1270cc58818484aR4

- `ConsentManagementPlatform` is not a default export

When I just used a dynamic import fro CMP, I noticed the bundle size was not as small as I would have [predicted](https://github.com/guardian/dotcom-rendering/pull/1351#issuecomment-613511842). I was only able to achive the much greater reduction in [bundle size](https://github.com/guardian/dotcom-rendering/pull/1351#issuecomment-614057994) when splitting out the lib. I therefore assume that the lib does not get implicitly split out on dynamic imports. I would need to further test this assumption however.

@oliverlloyd this should answer your question https://github.com/guardian/dotcom-rendering/pull/1351/files#r409735054

src/web/components/App.tsx
liywjl: @GHaberis so this is a bug? We should change this to:
```suggestion
                    {CAPI.config.cmpUi && <CMP />}
```

src/web/components/CMP.tsx
oliverlloyd: As far as I understand it, `cmpUI` could be renamed `cmpIsEnabled` (site wide) and then `shouldShow` is a local flag for that user/browser.

Based on that, we only need to import the lib if `cmpIsEnabled` and then within the lib/component there's local state to show hide based on `shouldShow`. But it's fair to assume it will always be enabled so the advantage to dynamically loading on this switch is, I think, that the loading is removed from the initial bundle and moved to App - but still immediate?

The next logical step from here would be to extract the `shouldShow` logic so that we can dynamically import based on if the user has seen the banner or not - that might need a refactor of CMP though

src/web/components/App.tsx
GHaberis: @liywjl yes your `{CAPI.config.cmpUi && <CMP />}` suggestion is correct 👍

oliverlloyd:
I don't yet have this fully uploaded into my head so this might not work but is there a way to abstract the dynamic import such that it can be composed? Similar to `<Lazy>`?

src/web/components/CMP.tsx
liywjl: I hope to get som clarification on what `cmpUi` means in this thread https://github.com/guardian/dotcom-rendering/pull/1351#discussion_r410088412. In the meantime I am just focused on not changing any of the condutional logic.

I fully agree, we should only  fetch the CMP and related libs if `cmpUi` is false. Thats is why I lifted the `cmpUi` condition logic to `App.tsx`, meaning none of the library is dynamically loaded if `cmpUi` is true. Also the component `ConsentManagementPlatform` only gets fetched if `show` is true. We therefore have dynamically loaded both the functions `shouldShow` and `setErrorHandler` (which are  ~50kb) and the component `ConsentManagementPlatform` (around ~20kb).

ripecosta: We opted for not having the `shouldShow` within the component because the CMP UI can still be surfaced even if this function returns false (by the user clicking the privacy settings link at the bottom of the page). Having a separate shouldShow function also helps when determining if other banners can be shown or not.

liywjl: @ripecosta in other words `shouldShow` should always be fired irregardless of `cmpUi`?

ripecosta: In this case, yes. You always need to check both to surface the UI.

src/web/components/CMP.test.tsx
oliverlloyd: I'd prefer we keep the test to make sure we're doing things right and also as a form of documentation

src/web/components/CMP.tsx
oliverlloyd: > I fully agree, we should only fetch the CMP and related libs if `cmpUi` is false. Thats is why I lifted the `cmpUi` condition logic to `App.tsx`, meaning none of the library is dynamically loaded if `cmpUi` is true.

I think the logic here is the opposite. We should only render `<CMP />` if cmpUI is `true`

ripecosta: Correct. We should only render `<CMP />` when `cmpUI` is true and `shouldShow` returns true.

liywjl: thanks for catching this @ripecosta I now see i did in fact change the logic here. I will revert to a previous version.

Also I wanted to double cheack, accorinding to @GHaberis https://github.com/guardian/dotcom-rendering/pull/1351#discussion_r410088412 the `cmpUi` logic was wrongly implemented before, we want to render the `CMP` component if `cmpUi` is true?

liywjl: @ripecosta so just to confirm: `if(cmpUI && shouldShow() === true)`  then display `<CMP />` ? 

so its not an `||` condition but a `&&`?

ripecosta: This was what we had initially: `<CMP cmpUi={CAPI.config.cmpUi} />` which was then being correctly checked inside the component:
```
if (!show || !cmpUi) {
    return null;
}
```

One of the versions on this PR changed the logic to `{CAPI.config.cmpUi === false && <CMP />}` which is incorrect, but has since been fixed to `{CAPI.config.cmpUi && <CMP />}`.

Extra points if you do something like `{CAPI.config.cmpUi && shouldShow() && <CMP />}` because in most pageviews `shouldShould()` will return false, despite the switch being enabled, because the UI won't need to be show.

liywjl: to clarify, negations make things hard to discuss ^^ here is some recap:
`{CAPI.config.cmpUi && <CMP />}` => we life the `CAPI.config.cmpUi` logic to `App.tsx`
`show` should be used to render the component, and not to return null (for clarity)

liywjl:
@oliverlloyd I decided to change the name of the webpack dynamic import to better give you an idea of what each file means in the build:

Screenshot 2020-04-17 at 11 56 01


`CMP-api-calls.2f4084ce137073d551e1.js` => my guess is this is some webpack glue used to pull the compnent down?

`vendors~CMP-api-calls~guardian-consent-management-platform-lib-ConsentManagementPlatform.dd5e88666211bdffd775.js` => this is the webpack tree shaken our library (you can notice the size is much larger)

`vendors~guardian-consent-management-platform-lib-ConsentManagementPlatform.22981d6cf9f3dcc2e776.js` => this is the component that is broken out using `loadable-components`

liywjl:
Closing in favor of:
React Lazy: https://github.com/guardian/dotcom-rendering/pull/1376
Dynamic Imports: https://github.com/guardian/dotcom-rendering/pull/1479

PR #1352, opened 2020-04-15
Use display

oliverlloyd:
## What does this change?
Propagates the `display` property down into thos sub components that branch based on if the context is an immersive article or not

Previously this components applied different styles based on when `designType` was `Immersive` but now we do that using `display === 'immersive'

### Why so many changes?
There are a lot of changes here but they're repetative and well distributed by commit so I suggest reviewing on a commit by commit bases for increased reviewing pleasure

### Why are my stories broken?
The immersive layout stories are not broken but this is a good thing. We don't yet support immersive layout but we're starting to and these visual changes in the stories show this movement

### Just why?
This prepares the way for things like having immersive Feature articles and also means we'll be able to remove the `Immersive` design type completely soon.

https://trello.com/c/cIDwM4es/1436-format

### DIFF
--- a/src/web/components/ArticleHeadline.tsx
+++ b/src/web/components/ArticleHeadline.tsx
@@ -239,6 +239,7 @@ const renderHeadline = ({
                     <h1 className={lightFont}>{curly(headlineString)}</h1>
                     {byline && (
                         <HeadlineByline
+                            display={display}
                             designType={designType}
                             pillar={pillar}
                             byline={byline}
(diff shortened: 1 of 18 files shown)


src/web/components/ArticleHeadline.tsx
```diff
@@ -183,7 +186,28 @@ const renderHeadline = ({
         colour?: string;
     };
 }) => {
+    if (display === 'immersive') {
```
liywjl, line 189: so `designType` has capitalisation ex: Immersive
and display lowercase

oliverlloyd: Indeed, `pillar` is lower so I followed that. Either way we're inconsistent. We can look to refactor capitalisation or even move over to typescript's const strings later

PR #1356, opened 2020-04-16
Refactor article title

oliverlloyd:
## What does this change?
Adds the `display` prop to `ArticleTitle` using it to create a new inverted style

Screenshot 2020-04-16 at 08 32 05

## Why?
So we can use this component, and its associated link logic, with immersive articles

https://trello.com/c/HcWLFBfG/1438-review-articletitle


oliverlloyd:
> are we ok to loose the text under the pillar name?

Good spot. But yes, this is desired - immersive articles only show the main title

PR #1367, opened 2020-04-17
Simplify CMP logic

liywjl:
## What does this change?
Simplify CMP logic

## Why?
CMP will be used in a dynamic import example and we need to make sure it is clear to understand

https://trello.com/c/ecKE6YoC/1466-simplify-cmp-logic

### DIFF
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -317,7 +317,7 @@ export const App = ({ CAPI, NAV }: Props) => {
                 </Lazy>
             </Portal>
             <Portal root="cmp">
-                <CMP cmpUi={CAPI.config.cmpUi} />
+                <>{CAPI.config.cmpUi && <CMP />}</>
             </Portal>
         </>
     );
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -316,9 +316,11 @@ export const App = ({ CAPI, NAV }: Props) => {
                     />
                 </Lazy>
             </Portal>
-            <Portal root="cmp">
-                <>{CAPI.config.cmpUi && <CMP />}</>
-            </Portal>
+            {CAPI.config.cmpUi && (
+                <Portal root="cmp">
+                    <CMP />
+                </Portal>
+            )}
         </>
     );
 };
--- a/src/web/components/CMP.tsx
+++ b/src/web/components/CMP.tsx
@@ -5,11 +5,7 @@ import {
 } from '@guardian/consent-management-platform';
 import { ConsentManagementPlatform } from '@guardian/consent-management-platform/lib/ConsentManagementPlatform';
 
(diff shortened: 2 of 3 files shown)


src/web/components/CMP.tsx
```diff
@@ -25,11 +21,10 @@ export const CMP = ({ cmpUi }: Props) => {
...
-        return null;
-    }
-
-    return <ConsentManagementPlatform source="dcr" onClose={onClose} />;
+    return show ? (
+        <ConsentManagementPlatform
+            source="dcr"
+            onClose={() => setShow(false)}
+        />
+    ) : null;
```
oliverlloyd, line 29:
```suggestion
    if (!show) {
        return null;
    }
   
    return <ConsentManagementPlatform source="dcr" onClose={onClose} />;
```

oliverlloyd: This seems clearer and removes the need for the ternary

src/web/components/App.tsx
```diff
@@ -317,7 +317,7 @@ export const App = ({ CAPI, NAV }: Props) => {
                 </Lazy>
             </Portal>
             <Portal root="cmp">
-                <CMP cmpUi={CAPI.config.cmpUi} />
+                <>{CAPI.config.cmpUi && <CMP />}</>
```
oliverlloyd, line 320: We could also wrap the `<Portal>` in a check for `cmpUI`

src/web/components/CMP.tsx
liywjl: I didn't like it because of the negation, it became quite confusing when we were discussing the logic in dynamic loading, because you need to make the constant distinction between the value and the if statement.

regarding the ternary, how about this instead? 

```suggestion
    return (
    <>
        {show && 
            <ConsentManagementPlatform
                source="dcr"
                onClose={() => setShow(false)}
            />
        ) }
    </>
```

src/web/components/App.tsx
liywjl: good point 👍

src/web/components/CMP.tsx
liywjl: I feel this is more personal preference, so I am happy with whichever. I just had a lot of confusing conversations with people when the code was:
```
if (!show || !cmpUi) {
    return null;
}
```
the double negation made it difficult to discuss. But if its just one variable then it is much more clearer

oliverlloyd: Yes, I agree that double negation with an OR combiner is hard to reason through.

But here we just have one element so it's not complicated. I actually find it a lot clearer.

gtrufitt: I think the `show && ...` is the one more prevalent in the codebase

PR #1368, opened 2020-04-17
Add exclusion of discussion rendering from webpack exclude

gtrufitt:
## What does this change?

What if I told you... we didn't have to double build our modules.

This removes the exclusion of `@guardian/discussion-rendering` from the babel-loader, so that that module will get compiled. This is because in [discussion rendering](https://github.com/guardian/discussion-rendering/pull/222/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L9) I've set the `module` field to the **source entry** unbuilt code.

## Why?

Delegating the build to platforms is risky but reasonable for internal modules. We can take advantage of the platform's requirements (eg dotcom-rendering builds against modern (type="module") browsers and older browsers but apps-rendering might build differently) and build the source as we would if it was a part of our codebase. 

Really, this is almost as if the module was a part of the codebase in general, just another module - but with the benefits of versioning and deployment to help with pushing out to platforms in the future.

It means we can much more clearly see what is going on in our bundles (though this was mostly my mis-configuration of the build previously I think)
Screen Shot 2020-04-17 at 17 23 48

And it comes with modest bundle size benefits (due to being able to optimize for the platform, while supporting older browsers as expected)

|Before| After |
|--|--|
|Screen Shot 2020-04-17 at 17 24 49 | Screen Shot 2020-04-17 at 17 18 29 |

## More info

- This thread https://chat.google.com/room/AAAAWwBdSMs/1S3h7fd4CHk
- This article on JasonFormat https://jasonformat.com/enabling-modern-js-on-npm/
- Sindre Sorhus :godmode: thoughts https://github.com/sindresorhus/ama/issues/446

Need to write up this into more of a recommendation if it proves to work... (cc @SiAdcock @sndrs)

### DIFF
--- a/scripts/webpack/browser.js
+++ b/scripts/webpack/browser.js
@@ -69,7 +69,7 @@ module.exports = ({
     module: {
         rules: [{
                 test: /(\.tsx)|(\.js)|(\.ts)$/,
-                exclude: /node_modules/,
+                exclude: /node_modules\/(?!(@guardian\/discussion-rendering)\/).*/,
                 use: [{
                     loader: 'babel-loader',
                     options: {


gtrufitt:
Note: I don't know if this is a really stupid idea and I'm missing something.

oliverlloyd:
I'm trying to get my head around this. Is this a bit like we're publishing discussion in 'raw' format and leaving it up to the consumer to 'process' it?

There's this mental model in my head which very much sees 'node_modules' as this greyed out place that should be excluded from everything because 1. It's so big and 2. It's out of my control but that means there needs to be some kind of agreed standard for how the javascript gets turned into something useful cross browser and I'm guessing that's a mess.

But what you're saying is to put that mental model to one side and that we're making DCR the boss and  discussion is delegating responsibility for processing javascript?

gtrufitt:
> I'm trying to get my head around this. Is this a bit like we're publishing discussion in 'raw' format and leaving it up to the consumer to 'process' it?

Yup, that’s right.

> There's this mental model in my head which very much sees 'node_modules' as this greyed out place that should be excluded from everything because 1. It's so big 

Indeed, we’re only removing discussion rendering from the exclude here.

> and 2. It's out of my control but that means there needs to be some kind of agreed standard for how the javascript gets turned into something useful cross browser and I'm guessing that's a mess.

Exactly, you can’t rely, as a library author where you’re less in control of the consumers that those consumers will be transpiling from whatever fancy modern js you write to whatever they need to support. They might write ES5 and not transpile.

However, within the Guardian, we can probably rely on a minimum modern version and typescript being transpiled or at least communicate that somehow.

> But what you're saying is to put that mental model to one side and that we're making DCR the boss and  discussion is delegating responsibility for processing javascript?

Yup. The mess of not knowing your consumers transpilation methods will always be there, at least until package.json has a way to say - this code is *this* modern and is written in typescript so you need to deal with that please. 

But if we know that already, do we need to transpile in the Guardian libraries? I don’t reckon so? Or at least I think giving the option based on a set of transpilation requirements.

oliverlloyd:
I think that this is a good solution to the problem of trying to manage sharing code via NPM.

I also think if we're getting down to this level of efficiencies and complexity then we should just use a monorepo

Because:
- We're only sharing code internally
- We don't really care about versioning, as far as I can see it's only adding hassle
- We (or at least I) are not enjoying the linked PR dependency dance
- Solving the same tooling setup issues seems to be taking up a lot of everyone's time

gtrufitt:
Indeed, I don't know if it will or won't work in the long run, but I believe it's the right direction for module bundling. It's likely that we'll still need *some* bundling in discussion rendering - but maybe that's *very modern* JS, as opposed to source TS. Lets see how we get on with this.

sndrs:
i think i'm really coming round to this idea. but i wonder if what we should expose is not so much the source, but an untranspiled public version i.e. `.js` or `.mjs` etc files, with any type defs available as separate files? 

so that the package is still very much a JS package, it just makes no accommodation for where it will be run – that's the installer's job.

in this way, guardian npm modules for internal use become more like git submodules, but with the versioning benefits that @gtrufitt outlined

sndrs:
not sure that was clear: so you might still write your package in TS, but you publish js. e.g. the target in the TS config is `ESNext` and `lib` is empty

gtrufitt:
Indeed @sndrs - that's my thinking too. I've, for now, published src but I think you're right we'll need to bundle some kind of high level JS not TS. 

Iteration!

PR #1371, opened 2020-04-17
Immersive Spacing

oliverlloyd:
## What does this change?
Refacotrs the spacing around the standfirst and headline components

## Why?
So that we can support immersive articles later

https://trello.com/c/f2GSOAzC/1468-roundel-colours


oliverlloyd:
> I am unsure of the size variation here, it seems quite large

It's intentional, the standfirst for immersive articles is smaller

oliverlloyd:
@liywjl Good spot on the size difference but this is correct. The byline is smaller for immersive articles. You can check one [here](https://www.theguardian.com/film/2019/oct/17/ken-loach-all-his-films-ranked?dcr=false)

PR #1372, opened 2020-04-19
Refactor Nav to support immersives

oliverlloyd:
## What does this change?
Adds support for an immersive version of `Nav`

The nav bar is slightly higher when showing on an immersive article and shows a Roundel. This PR adds the styling to support these things and passes in the `display` prop to decide when to do this. There's a lot of prop passing in this PR as well as a small refactor to `App` so we can pass in `display` when hydrating (as well as SSR)

### Standard
Screenshot 2020-04-19 at 21 21 13

### Immersive
Screenshot 2020-04-19 at 21 21 20

## Why?
To support immersive articles

https://trello.com/c/MnvVu3Bn/1470-immersive-nav

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -277,6 +277,8 @@ interface CAPIType {
 }
 
 type CAPIBrowserType = {
+    designType: DesignType;
+    pillar: Pillar;
     config: {
         isDev: boolean;
         ajaxUrl: string;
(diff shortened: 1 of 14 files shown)


oliverlloyd:
> It looks like the icon, increases the height of the Nav bar, so it might be worth either increasing the default hight of the Nav or reducing the height of the icon. But bar that all looks good

The increase in the Nav height is intentional. It's actually a big part of the reason for this PR!

liywjl:
> > It looks like the icon, increases the height of the Nav bar, so it might be worth either increasing the default hight of the Nav or reducing the height of the icon. But bar that all looks good
> 
> The increase in the Nav height is intentional. It's actually a big part of the reason for this PR!

cool, well then all looks good

PR #1375, opened 2020-04-20
Update CMP for copy change test

sndrs:
## What does this change?

updates version of CMP

## Why?

There's a test running

https://trello.com/c/hRLkjCAn

### DIFF
--- a/src/web/components/CMP.tsx
+++ b/src/web/components/CMP.tsx
@@ -7,15 +7,23 @@ import { ConsentManagementPlatform } from '@guardian/consent-management-platform
 
 let variant: string | undefined;
 
-const participations = localStorage.getItem('gu.ab.participations');
+if (
+    window &&
+    window.guardian &&
+    window.guardian.config &&
+    window.guardian.config.switches &&
+    window.guardian.config.switches.abCommercialCmpCopy
+) {
+    const participations = localStorage.getItem('gu.ab.participations');
 
-if (participations) {
-    try {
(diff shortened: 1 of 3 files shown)


sndrs:
sorry @oliverlloyd @liywjl – @ripecosta made the good point that this didn't actually pass in the param the CMP needed!

I've updated the PR.

i've hardcoded the check of LocalStorage in this PR because we need to get the test out, but is there a better way to do it?

src/web/components/CMP.tsx
```diff
@@ -5,6 +5,20 @@ import {
 } from '@guardian/consent-management-platform';
 import { ConsentManagementPlatform } from '@guardian/consent-management-platform/lib/ConsentManagementPlatform';
 
+let variant: string | undefined;
+
+const participations = localStorage.getItem('gu.ab.participations');
```
oliverlloyd: Calls to local storage often fail (Safari private mode, quota exceeded, etc.) causing errors to throw so a try catch here would be good. Maybe it could be dropped down into the one below?

I guess you could argue this should be abstracted but I'm not sure we're there yet, abstractions come at a price and I find being able to access things like localStorage directly a big dev ex plus.

sndrs:
thank you all for your input – we've made a call not to run this on DCR for now, since it's only to monitor impact and not base a decision on it anyway.

it def felt like the wrong thing to be doing, all those `&&` etc 

sorry to waste your time!

PR #1376, opened 2020-04-20
React Lazy for CMP

liywjl:
## What does this change?
Dynamically imports `ConsentManagementPlatform` from the ConsentManagementPlatform lib.

## Why?
`ConsentManagementPlatform` is not always used and we gains a major bundle reduction by dynamically loading it.

## How to test?
To avoid having to guess when CMP gets run, you can make sure the condition in `App.tsx` and `CMP.tsx` both return true and then you should be able to see the component being rendered. Look at the network tab in devtools to see that componet being loaded dynamically.

## Why React.lazy rather than `loadable-components`? I though we did SSR?
`loadable-components` can be used for SSR dynamic imports, but we don't gain any bundle reduction by doing this. We aim to reduce the frontend bundle (mainly `react.js`) so using React's native lazy API fits out need.

## doesn't `loadable-components` support code splitting libraries? 
Yes it does, but so does webpack, and unlike `loadable-components` we are not forced to wrap libraries in render-props.

## You are using Suspense, isn't that experimental?
Yes, this is a concern, the API could likely change. However the benefit right now is quite substantial, and the code changes needed to revert are quite minimal. 

https://trello.com/c/0qCpwbbO/1285-dynamic-imports


src/web/components/CMP.test.tsx
```diff
@@ -32,19 +33,19 @@ describe('CMP', () => {
...
     });
 
-    it('It should render null if shouldShow returns false', () => {
+    it('It should render null if shouldShow returns false', async () => {
         shouldShow.mockImplementation(() => false);
 
         const { container } = render(<CMP />);
 
-        expect(container.firstChild).toBeNull();
+        await waitFor(() => expect(container.firstChild).toBeNull());
```
liywjl, line 41: component rendering is now dynamic so we need to have a `waitFor`

src/web/components/CMP.tsx
```diff
@@ -1,9 +1,14 @@
...
+import React, { useState, useEffect, Suspense } from 'react';
 import {
     shouldShow,
     setErrorHandler,
 } from '@guardian/consent-management-platform';
-import { ConsentManagementPlatform } from '@guardian/consent-management-platform/lib/ConsentManagementPlatform';
+
+const ConsentManagementPlatform = React.lazy(() =>
+    import(
+        /* webpackChunkName: "ConsentManagementPlatform" */ '@guardian/consent-management-platform/lib/ConsentManagementPlatform'
```
liywjl, line 9: `webpackChunkName` allows us to specify the name of the component, else webpack would default it to some random key, making hard to read

src/web/components/CMP.tsx
```diff
@@ -24,10 +29,12 @@ export const CMP = () => {
     return (
         <>
             {show && (
-                <ConsentManagementPlatform
-                    source="dcr"
-                    onClose={() => setShow(false)}
-                />
+                <Suspense fallback={<></>}>
```
liywjl, line 32: we fallback to a react fragment instead of a div for the tests

babel.config.js
```diff
@@ -1,8 +1,8 @@
 module.exports = {
     plugins: [
+        '@babel/plugin-syntax-dynamic-import',
         '@babel/plugin-transform-react-jsx',
         'babel-plugin-preval',
-        'babel-plugin-dynamic-import-node',
```
liywjl, line 5: `babel-plugin-dynamic-import-node` makes dynamic imports impossible, it overrides `import` to `require`. It was intended to be used for server code (because Node needs require), but typescript handles this conversion from what I gather

src/web/components/CMP.tsx
gtrufitt: Why do we need suspense? Just for the tests?

gtrufitt:
I *really* like the though of "using the platform" here in terms of support for this stuff, the complexity vs that loadable plugin is much reduced too.

This doesn't seem to have as aggressively split the CMP stuff - the reduced bundle size is `-9kb` vs the `-18kb` in your loadable branch. What's not being split? The lib? I think maybe then, we need to look at webpack split chunks for that stuff (the lib) as opposed to the Components (if we're moving away from the use of loadable.lib) - lets look at that in a future PR!

I have some misgivings about Supense and Lazy due to this in the Preact docs...

image

I think it's probably ok, there's [active work](https://github.com/preactjs/preact/releases/tag/10.2.0) so I'd be keen to get this out and keep an eye on issues. 

In that respect, have you thought about how errors or issues might manifest themselves and how we guard against them or catch/log them in production / build time?

liywjl:
> This doesn't seem to have as aggressively split the CMP stuff 

Yes this is because the other functions (which are large) were done using webpack dynamic imports. Here I only wanted to focus on React Lazy. I wanted to do it in another PR to make a clear distinction on how we would implement.

> I have some misgivings about Supense and Lazy due to this in the Preact docs...

Great catch! yes this is a little concerning. Would you prefer `loadable-components` implementations if that is the case? I searched open issues on preact's github and have only found one issue regarding Suspense.

> In that respect, have you thought about how errors or issues might manifest themselves and how we guard against them or catch/log them in production / build time?

There are 2 main issues I see arrising:

- build time: build issues are likely caused because the breaking off of chunks to not work, we would get an error the build failed, which would prevent a release. We could maybe add a step in the CI to generate a build and check that everything works.

- run time: when the browser fails to fetch or render a component. This is a little more difficult to guard against mainly because of all the different browsers and versions. The best we can do is make sure we thoroughly test this on all browsers, and use sentry to log any issues.

gtrufitt:
> Here I only wanted to focus on React Lazy

Yup cool makes sense. 

> Would you prefer loadable-components implementations if that is the case?

I'm not sure, part of me thinks that it's worth sticking with this because a) it's not that complex and b) the migration to loadable is not that difficult if we discover issues c) I reckon the Preact suspense and lazy is likely more robust than loadable anyway (despite being 'experimental') purely because it's a part of the core and not a module supported by one person. 

Lets stick with this one.

> We could maybe add a step in the CI to generate a build and check that everything works.

Do the layouts + chromatic capture this? Can we get them to? Or Cypress?

liywjl:
> I'm not sure, part of me thinks that it's worth sticking with this 

Yeah I agree with your points, we should still be wary when lazy loading errors appear

> Do the layouts + chromatic capture this? Can we get them to? Or Cypress?

I suspect `continuous-integration/teamcity` would pick up a build issue. For a client side loading issue its a little more complex, we would need to test this in all browsers before launch (just to check it can fetch the component and load it correctly)

Storybook wont be a good testing solution, as it runs its own webpack config to build the stories, we could explore using DCR's webpack config but i think the benefit remains limited.

Cypress would make more sense, as if the component is rendered we can assume that it loaded correctly. CMP might be a bad example though as we do not control the logic of rendering.

src/web/components/CMP.tsx
liywjl: I thought this was for tests, but looks we should always use React.lazy with Suspense

gtrufitt:
[Dynamic imports for React](https://trello.com/c/0qCpwbbO/1285-dynamic-imports-for-react)

liywjl:
Tested on CODE

oliverlloyd:
I had a play with using this branch to code split some of the other components that load at the bottom of the page and the bundle breaks down like this

Screenshot 2020-05-09 at 21 56 04

Which looks pretty good. Loving that new React bundle size. I wasn't really able to translate it into any huge perf score differences though 😐  but I'm keen to spend more time looking at this.

I also tested the changes I made using a production build and found a problem. The issue is [documented here](https://github.com/webpack/webpack/issues/8996) and resolved using [a suggestion](https://github.com/webpack/webpack/issues/8996#issuecomment-494777363) in the same ticket, adding 

```
    optimization: {
        namedModules: true,
        namedChunks: true,
        splitChunks: { cacheGroups: { default: false } },
    },
```

to the webpack config fixed the issue.

Finally, along the way I learnt that we need to have a seaprate `Suspense` context for each `Portal`, not a global one at the top of the App. That might save you some time!

gtrufitt:
Really keen to get this out on Monday! 

Side note:
The section code splitting in https://engineering.fb.com/web/facebook-redesign/ is super interesting!

We should think about what our ‘tiers’ are!

liywjl:
This PR not only applies dynamic imports to CMP component

liywjl:
closed in favor of https://github.com/guardian/dotcom-rendering/pull/1492

PR #1377, opened 2020-04-20
Increase minimum PROD capacity to 6

gtrufitt:
## What does this change?

Increase ASG minimum PROD capacity to 6

## Why?

Struggles occasionally on 4 instances, so doubling gives us room to investigate a longer term fix (different instance types?)

https://trello.com/c/hKPOpMqY


gtrufitt:
> Could do 6 and 24?

Yeah, lets try

PR #1386, opened 2020-04-21
Fix NoJS support for Nav

liywjl:
## What does this change?
We fix the support for NoJS on the nav by using checkbox input and CSS.

In doing so we removed a lot of JS logic that was implemented before

## Why?
Nav intended to support NoJS, I also will allow us to SSR the Nav and to potentially remove a big chunk of the `window.guardian` object

## How?
We used `~` in CSS to allow you to apply styles to sibling DOM elements. The technique was to use an ID for the input checkbox and then use the CSS `:checked` descriptor combined with `~` to then apply the new style to sibling elements. 

ex:
```
<input type="checkbox" id="someCheckbox" />
<h1>Hello World</h1>

#someCheckbox ~ h1 {
  color: red;
}
```

If the target element is nested you then need to explicitly state the path
ex:
```
<input type="checkbox" id="someCheckbox" />
<div>
  <h1>Hello World</h1>
</div>

#someCheckbox ~ div h1 {
  color: red;
}
```

It is also important to note that this only works for DOM elements that `preceded`  the input
ex:
```
// NOTE: This will NOT work
<div>
  <h1>Hello World</h1>
</div>
<input type="checkbox" id="someCheckbox" />

#someCheckbox ~ div h1 {
  color: red;
}
```

https://trello.com/c/cpO2xeKH/1290-fix-nav-nojs-components

### DIFF
--- a/src/web/components/Nav/Nav.test.tsx
+++ b/src/web/components/Nav/Nav.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, fireEvent } from '@testing-library/react';
+import { render, within } from '@testing-library/react';
 import { Nav } from './Nav';
 import { nav } from './Nav.mock';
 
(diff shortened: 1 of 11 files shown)


src/web/components/Nav/Nav.test.tsx
```diff
@@ -1,44 +1,45 @@
...
-
-        expect(queryAllByRole('menu')).toEqual([]);
-        fireEvent.click(getByText('More'));
-        expect(expandedMenu).toHaveStyle('display: block');
-        expect(queryAllByRole('menu').length).toBeGreaterThan(1);
-        fireEvent.click(getByText('More'));
-        expect(queryAllByRole('menu')).toEqual([]);
-    });
+    // TODO:
+    // it('should open and close the expanded menu by clicking More', () => {
```
liywjl, line 34: due to the way we render things this test will not longer work. Trying to update it and I found some complications with how to render an element using `@testing-library/react`. 

We render the Nav differently depending on the width of the window. However rendering it using `@testing-library/react` means we do not have control of this. When we render this component on smaller screens the only CSS effect we apply is transfrom, which I struggle to extract using `toHaveStyle`.

liywjl:
closed in favor of https://github.com/guardian/dotcom-rendering/pull/1399

PR #1387, opened 2020-04-21
ImmersiveHeadline

oliverlloyd:
## What does this change?
Adds `ImmersiveHeadline` as a component

```typescript
type Props = {
    display: Display;
    designType: DesignType;
    tags: TagType[];
    author: AuthorType;
    headline: string;
    sectionLabel: string;
    sectionUrl: string;
    guardianBaseURL: string;
    pillar: Pillar;
    badge?: BadgeType;
};
```

### Why the propocalypse?
Because we need to prop drill the props needed for `ArticleTitle` and `ArticleHeadline`

## Why?
Immersive headlines are complicated. In particular, they use css pseudo selectors to make their black background stretch to the right edge of the page. To achieve this within the existing frameword of `ArticleHeadline` would have added needless complexity to that component, and doing it directly in `ImmersiveLayout` would bloat that file. In addition, we also need some of the layout logic from `Section` to position the headline but not all of it.

Rather than confuse our existing components with new requirements, a separate component encapsulates this similar but special logic in one place, hopefully with clear dividing lines making it easier to reason about

https://trello.com/c/andzBqpC/1494-immersive-headline


src/web/components/ArticleHeadline.stories.tsx
```diff
@@ -387,93 +387,3 @@ export const Live = () => (
...
-                            pillar="culture"
-                            tags={[]}
-                        />
-                    </div>
-                </ArticleContainer>
-            </Flex>
-        </Section>
-    </>
-);
-ShortImmersive.story = { name: 'Immersive with a short headline' };
```
oliverlloyd, line 479: These stories would still work but keeping them is missleading

PR #1389, opened 2020-04-22
Asset Manifest Localhost

liywjl:
## What does this change?
We remove  the PROD flag from the Asset Manifest generation in webpack

## Why?
We should load both legacy and non legacy file locally to be able to test older browsers like IE.

https://trello.com/c/WLdlGWct/1499-webpack-support-legacy-build-localhost

### DIFF
--- a/scripts/webpack/browser.js
+++ b/scripts/webpack/browser.js
@@ -2,9 +2,7 @@ const webpack = require('webpack');
 const AssetsManifest = require('webpack-assets-manifest');
 const FriendlyErrorsWebpackPlugin = require('friendly-errors-webpack-plugin');
 const chalk = require('chalk');
-const {
-    siteName
-} = require('../frontend/config');
+const { siteName } = require('../frontend/config');
 
 const friendlyErrorsWebpackPlugin = () =>
     new FriendlyErrorsWebpackPlugin({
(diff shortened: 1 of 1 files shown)


scripts/webpack/browser.js
```diff
@@ -53,7 +50,6 @@ module.exports = ({
         chunkFilename: generateName(isLegacyJS),
     },
     plugins: [
-        PROD &&
```
liywjl, line 56: this is the only change, the rest are formatting

gtrufitt:
Would you be able to explain a bit more about what was happening previously and why this fixes it? Is it purely by virtue of a having an assets manifest locally, it uses the two versions? Seems... magic.

liywjl:
> Would you be able to explain a bit more about what was happening previously and why this fixes it? Is it purely by virtue of a having an assets manifest locally, it uses the two versions? Seems... magic.

of course, in `/src/web/server/document.tsx` we use `getDist` to get the path to each file. That function is defined `/src/lib/assets.ts` and what it tried to do is to find the `assetHash` and `assetHashLegacy` by looking for `manifest.json` and `manifest.legacy.json`.

The `manifest.json` contains a list of all files plus their hash key (the hash key is not defined on local dev). Before this PR we did not generate `manifest.json` when running on local so `getDist` would fall back to the name of the file requested:

```
export const getDist = ({
    path,
    legacy,
}: {
    path: string;
    legacy: boolean;
}): string => {
    const selectedAssetHash = legacy ? assetHashLegacy : assetHash;
    return `${CDN}assets/${selectedAssetHash[path] || path}`;
};
```

This allowing the build to generate the manifest would allow us to support legecy and non-legacy bundles locally

PR #1390, opened 2020-04-22
Add amp-experiment component

buck06191:
## What does this change?
This commit adds this component and places it in the Article component,
test document and `render.tsx`.

It does not update the analytics to send the variant information as
this will require updating the ophan config information.

## Why?
In order to perform proper A/B testing in AMP we need to use the
`amp-experiment` tag to create variant groups.

https://trello.com/c/EfkqCZue/744-create-test-implementation-of-amp-experiment-tag


buck06191:
We need to add the analytics tags to the ophan config before merging this.

buck06191:
image

Example of using the `amp-experiment` tag to change the header colour

src/amp/abTests.ts
```diff
@@ -0,0 +1,16 @@
+import {
+    ExperimentFullConfig
```
jeteve, line 2: Importing a type from an AMP component is confusing.

src/amp/components/Experiment.tsx
```diff
@@ -0,0 +1,89 @@
+import React from 'react';
+
+export interface ExperimentModel {
```
jeteve, line 3: I think the types and functions should be moved to src/amp/lib, in order to keep amp/components/ more strictly related to the rendering.

src/amp/components/Experiment.tsx
```diff
@@ -0,0 +1,89 @@
...
+const extractStyleFromVariants = (variants: StyledVariants): [{[key: string]: number}, {[key: string]: string}] =>
+    Object.entries(variants).reduce(
+    (acc, [variant, config])=>{
+        const {proportion, style} = config;
+        acc = [{...acc[0], ...{[variant]: proportion}}, {...acc[1], ...{[variant]: style}}];
+        return acc
+    },
+    [{}, {}]
+);
+export const extractModelAndStyle = (fullConfig: ExperimentFullConfig): [ExperimentModel, ExperimentStyle] => {
```
jeteve, line 37: Can you elaborate on/document this function?

src/amp/server/document.test.tsx
```diff
@@ -56,12 +59,17 @@ test('produces valid AMP doc', async () => {
...
+        getActiveTests(abTestsFullConfig, {"ab-zero-test-experiment": true})
+    )[0];
+
     const body = (
         <Article
             nav={nav}
             articleData={CAPI}
             config={config}
             analytics={analytics}
+            experiment={activeTests}
```
jeteve, line 72: assigning something plural to something singular confuses me :)

src/amp/server/document.test.tsx
```diff
@@ -56,12 +59,17 @@ test('produces valid AMP doc', async () => {
         },
     };
 
+    const activeTests = extractModelAndStyle(
+        getActiveTests(abTestsFullConfig, {"ab-zero-test-experiment": true})
+    )[0];
```
jeteve, line 64: why just the first one?

src/amp/server/render.tsx
```diff
@@ -58,6 +64,7 @@ export const render = ({ body }: express.Request, res: express.Response) => {
                     articleData={CAPI}
                     nav={extractNAV(CAPI.nav)}
                     analytics={analytics}
+                    experiment={activeTests}
```
jeteve, line 67: same :)

src/amp/server/document.test.tsx
buck06191: The first element of the array is the model, second element is the style. I originally had this as an object but it was adding complexity to the `extractModelAndStyle` function.

src/amp/components/Experiment.tsx
buck06191: Happy to do this, but worth noting that I took this format from the existing Analytics component and module. If we move this around for the `Experiment` component, we might want to do the same for the `Analytics` component and any others that match this.

src/amp/components/Experiment.tsx
buck06191: Where should this be done? In the module or in external documentation?

src/amp/components/Experiment.tsx
nicl: Would be good to get more thoughts on this. I personally think that types should be as far as possible declared at the *point of use*. So here would make sense in this case.

src/amp/components/Experiment.tsx
```diff
@@ -0,0 +1,89 @@
...
+        ''
+    )
+};
+
+export const Experiment: React.FC<{
+    experiment: ExperimentModel;
+}> = experiment => {
+    const script: string = `<amp-experiment>
+        <script type="application/json">
+        ${JSON.stringify(experiment.experiment)}
```
jeteve, line 84: Where's experiment.experiment? When I look at the example data  ( ab-zero-test-experiment ), or the ExperimentModel type (well, interface)  I can't see an 'experiment' key, so I don't understand where this comes from.

src/amp/abTests.ts
```diff
@@ -0,0 +1,16 @@
...
+} from "@root/src/amp/components/Experiment";
+
+// Variant proportions must be >0, so we set extremely low to simulate a 0% test
+export const abTestsFullConfig: ExperimentFullConfig = {
```
jeteve, line 6: We'd probably better off avoiding introducing a terminology (ABTest) that's more restrictive than the AMP terminology (experiment)

src/amp/components/Experiment.tsx
buck06191: For some reason the object passed through to experiment was being put inside another object where the key was the prop (here that's `experiment`). I'm certain there's a way of accessing it that's better than what I've used but I couldn't figure it out myself.

src/amp/components/Experiment.tsx
oliverlloyd: I agree with Nic, if a type is only used in one place then it makes sense to declare it right there. It's easier for others to reason about / maintain the code when it's not distributed

oliverlloyd: I'd also +1 Jerome's other point around keeping the component related to rendering. We don't have any strict rules around this but I find things easier to manage when files do one thing and have one export. Sometimes it feels wrong to break something out into a new file though and keeping it local is the way to go. If there had to be a dogmatic rule then it would be to 'always consider maintainability'

In DCR (web) we have a lib folder with a series of files in it, one file per function. It sounds weird but it makes finding what you need easier and prevents the `utils.js` explosion that normally happens on most projects. Not sure if that helps 🤷

buck06191:
Changes made regarding the previously given comments.
Code now has unit tests which were done prior to refactor. These unit tests still pass and visual test shows the code still works:
image

src/amp/components/Experiment.tsx
buck06191: I've moved non-component code into the `src/lib` directory to keep the code responsible for rendering separate and cleaner.

src/amp/server/document.test.tsx
buck06191: Changed naming conventions in refactor

src/amp/abTests.ts
buck06191: Moved Types out of this file.

src/amp/components/Experiment.tsx
buck06191: Refactored this function into two functions which should hopefully be more understandable.

src/amp/components/Experiment.tsx
buck06191: Improved this to use idiomatic React code with destructuring of the props argument.

src/amp/server/render.tsx
buck06191: See above- refactored to improve naming convention

src/amp/server/document.test.tsx
buck06191: `extractModelAndStyle` has been refactored into two functions - one for AmpModel, one for Style - to improve clarity

src/amp/abTests.ts
buck06191: Improved naming in recent refactor

src/amp/lib/experiment.ts
```diff
@@ -0,0 +1,150 @@
...
+                },
+                '',
+            );
+            return `${experimentStyleString} ${experimentStyle}`.trim();
+        },
+        '',
+    );
+};
+
+// export const extractProportionFromVariants = (variants: {
```
liywjl, line 117: why is this commented out?

src/amp/pages/Article.tsx
```diff
@@ -9,6 +9,8 @@ import { AdConsent } from '@root/src/amp/components/AdConsent';
 import { css } from 'emotion';
 import { Sidebar } from '@root/src/amp/components/Sidebar';
 import { Analytics, AnalyticsModel } from '@root/src/amp/components/Analytics';
+import { Experiment, } from '@root/src/amp/components/Experiment';
```
liywjl, line 12: extra `,`, check if prettier is working correctly on your IDE, this should be auto removed

src/amp/lib/experiment.ts
buck06191: There were two approaches to this - one using a single `extractFromVariants` with type predicates and one with two almost identical `extract...FromVariants` functions. I wanted to check which would be preferable at this stage and then remove the one that isn't in use.

Which approach do you think is better here?

liywjl: I think the commented out section seems more readable, it uses reduce more sparingly and is easier to follow, so I would favor that if possible

src/amp/pages/Article.tsx
buck06191: This is after using Intellij. I'll move back to VS code and make sure to run the Prettier formatter on it. Thanks for catching this.

src/amp/experimentConfigs.ts
```diff
@@ -0,0 +1,14 @@
+import { StyledModelCollection } from '@root/src/amp/lib/experiment';
+
+// Variant proportions must be >0, so we set extremely low to simulate a 0% test
+export const experimentFullConfig: StyledModelCollection = {
+    'ab-zero-test-experiment': {
+        sticky: false,
+        variants: {
+            treatment1: {
+                proportion: 0.0000001,
+                style: `.test-header { background-color: yellow;}`,
```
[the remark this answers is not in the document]
buck06191: Is there an example of a more subtle CSS change that we can still check for?

src/amp/lib/experiment.ts
```diff
@@ -0,0 +1,150 @@
...
+type StylesCollection = ExperimentCollection<{ [variantName: string]: string }>;
+
+const extractFromVariants = (
+    variants: { [variantName: string]: VariantConfig },
+    fieldName: keyof VariantConfig,
+): { [variantName: string]: string | number } => {
+    return Object.entries(variants).reduce((acc, [variantName, config]) => {
+        acc = { ...acc, ...{ [variantName]: config[fieldName] } };
+        return acc;
+    }, {});
```
twrichards, line 35: - probs not the best idea to reassign the accumulator variable
- no need to build a new inner object with one key&value just to spread it
- needn't have explicit return, can just be an expression

```suggestion
    return Object.entries(variants).reduce((acc, [variantName, config]) => ({
        ...acc, 
        [variantName]: config[fieldName]
    }), {});
```

src/amp/lib/experiment.ts
```diff
@@ -0,0 +1,150 @@
...
+};
+
+const isValidAmpModelVariant = (variant: {
+    [variantName: string]: string | number;
+}): variant is { [variantName: string]: number } => {
+    return Object.values(variant).every(x => typeof x === 'number');
+};
```
twrichards, line 42: can just be an expression 
```suggestion
}): variant is { [variantName: string]: number } => 
   Object.values(variant).every(x => typeof x === 'number');
```

src/amp/lib/experiment.ts
```diff
@@ -0,0 +1,150 @@
...
+
+const isValidStyleVariant = (variant: {
+    [variantName: string]: string | number;
+}): variant is { [variantName: string]: string } => {
+    return Object.values(variant).every(x => typeof x === 'string');
+};
```
twrichards, line 48:
```suggestion
}): variant is { [variantName: string]: string } => 
   Object.values(variant).every(x => typeof x === 'string');
```

buck06191:
> Some areas in `src/amp/lib/experiment.ts` rely heavily on nested reduce function and a lot of
> `Object.entries` calls. I think this makes the code quite hard to follow. I think `buildExperimentStyle` is a great candidate for a function that could be broken down into smaller chunks.
> 
> Also it is worth checking that your IDE has prettier enabled, else it will cause future diffs for devs

Thanks for this. With regards to nested reduce functions, is extracting some of those nested `reduce` functions out into separate functions that are then called by the larger scope okay e.g.
```typescript
export const extractProportionFromVariants = (variants: {
    [variantName: string]: VariantConfig;
}): { [variantName: string]: number } =>
    Object.entries(variants).reduce((acc, [variantName, config]) => {
        acc = { ...acc, ...{ [variantName]: config.proportion } };
        return acc;
    }, {});

export const extractExperimentModels = (
    fullConfig: StyledModelCollection,
): AmpModelCollection => {
    return Object.entries(fullConfig).reduce(
        (extractedModels, [experimentName, styledModel]) => {
            const { variants, ...configWithoutVariants } = styledModel;
            const proportions = extractProportionFromVariants(variants);
            extractedModels[experimentName] = {
                ...configWithoutVariants,
                ...{
                    variants: proportions,
                },
            };
            return extractedModels;
        },
        {} as AmpModelCollection,
    );
};
```
 or do you mean to try and decrease the number of uses of `Object.entries` and `reduce` in general?

src/amp/lib/experiment.ts
```diff
@@ -0,0 +1,150 @@
...
+                    ...{
+                        variants: proportions,
+                    },
+                };
+            }
+            return extractedModels;
+        },
+        {} as AmpModelCollection,
+    );
+};
```
twrichards, line 69: - outer can be expression rather than return
- no need to create inner object with one key&value just to spread it
- rather than mutating the accumulator you could return a new object
```suggestion
): AmpModelCollection => Object.entries(fullConfig).reduce(
      (extractedModels, [experimentName, styledModel]) => {
          const { variants, ...configWithoutVariants } = styledModel;
          const proportions = extractFromVariants(variants, 'proportion');
          if (isValidAmpModelVariant(proportions)) {
              return {
                ...extractedModels,
                [experimentName]: {
                    ...configWithoutVariants,
                    variants: proportions,
                },
              };
          }
          return extractedModels;
      },
      {} as AmpModelCollection,
  );
```

src/amp/lib/experiment.ts
```diff
@@ -0,0 +1,150 @@
...
+            const { variants } = styledModel;
+            const styles = extractFromVariants(variants, 'style');
+            if (isValidStyleVariant(styles)) {
+                extractedModels[experimentName] = { ...styles };
+            }
+            return extractedModels;
+        },
+        {} as StylesCollection,
+    );
+};
```
twrichards, line 85:
```suggestion
export const extractExperimentStyles = (
    fullConfig: StyledModelCollection,
): StylesCollection => Object.entries(fullConfig).reduce(
      (extractedModels, [experimentName, styledModel]) => {
          const styles = extractFromVariants(styledModel.variants, 'style');
          if (isValidStyleVariant(styles)) {
              return {
                ...extractedModels,
                [experimentName]: styles,
              };
          }
          return extractedModels;
      },
      {} as StylesCollection,
  );
```

src/amp/lib/experiment.ts
```diff
@@ -0,0 +1,150 @@
...
+    return Object.entries(abTestObject).reduce(
+        (acc, [testName, testConfig]) => {
+            if (switches[testName]) {
+                acc[testName] = testConfig;
+            }
+            return acc;
+        },
+        {} as StyledModelCollection,
+    );
+};
```
twrichards, line 100:
```suggestion
): StyledModelCollection => Object.entries(abTestObject).reduce(
      (acc, [testName, testConfig]) => {
          if (switches[testName]) {
              return {
                ...acc,
                [testName]: testConfig,
              };
          }
          return acc;
      },
      {} as StyledModelCollection,
  );
```

twrichards: this could also be written as a ternary expression which would be shorter but not sure if it's more or less readable...
```suggestion
): StyledModelCollection => Object.entries(abTestObject).reduce(
      (acc, [testName, testConfig]) => switches[testName] ? {
                ...acc,
                [testName]: testConfig,
              } : acc,
      {} as StyledModelCollection,
  );
```

liywjl:
> ... or do you mean to try and decrease the number of uses of Object.entries and reduce in general?

I think using `reduce` should always be done sparingly,  the code here in the comment is much more readable as it only had 1 `reduce` function. I am much happier with it :)

buck06191:
> > ... or do you mean to try and decrease the number of uses of Object.entries and reduce in general?
> 
> I think using `reduce` should always be done sparingly, the code here in the comment is much more readable as it only had 1 `reduce` function. I am much happier with it :)

Fantastic. I've made those changes.
One thing I couldn't figure out how to do in a satisfactory way was to have a  general function for looping over an `ExperimentCollection` that would then apply specific methods. For example, loop over a StyledModelCollection and extract the Styles from it. That way there wouldn't need to be so many case of `Object.entries(experiments).reduce(...)` at the start of so many functions.

If this was possible to do, would this be a preferable approach to how it works currently?

oliverlloyd:
> One thing I couldn't figure out how to do in a satisfactory way was to have a general function for looping over an `ExperimentCollection` that would then apply specific methods. For example, loop over a StyledModelCollection and extract the Styles from it. That way there wouldn't need to be so many case of `Object.entries(experiments).reduce(...)` at the start of so many functions.
> 
> If this was possible to do, would this be a preferable approach to how it works currently?

This is just my subjective feeling on this so feel free to take this or leave it, but I've found it's often best to hold off from abstracting common code into functions until you get to the point where 'the code screams for it'. The idea is that it's simpler to read the code if there are less abstractions (new devs fresh to the code don't need to bounce around learning what different functions do) and it's also easier to make changes to the code in the future when it's duplicated because you don't have to worry about your changes affecting other use cases.

In general though, this is a fantastic example of collaboration on a PR - great work!

src/amp/lib/experiment.ts
buck06191: Thanks. I took a lot of what you suggested and applied it to the recommendations from the other reviewers. Particularly glad to know that I don't have to build an inner object to spread and combine into a new one.

src/amp/server/document.tsx
```diff
@@ -30,6 +37,12 @@ export const document = ({
     scripts: string[];
     metadata: Metadata;
 }) => {
+    const testCss = buildExperimentStyle(
```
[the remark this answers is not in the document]
buck06191: I've moved this logic inside `lib` as suggested. Same with the `activeExperiments` code.

src/amp/experimentConfigs.ts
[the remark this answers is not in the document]
buck06191: I've changed this to use a slightly different shade of blue...
If we need to check, we can just check  the hexcode and see which it matches.

src/amp/components/Experiment.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import { AmpModelCollection } from '@root/src/amp/lib/experiment';
```
sndrs, line 3: what's in `AmpModelCollection`? i'm not quite sure from the name?

what if it was just `AmpExperiment`?

src/amp/experimentConfigs.ts
```diff
@@ -0,0 +1,14 @@
+import { StyledModelCollection } from '@root/src/amp/lib/experiment';
+
+// Variant proportions must be >0, so we set extremely low to simulate a 0% test
+export const experimentFullConfig: StyledModelCollection = {
```
sndrs, line 4: again, not too sure what this type is?

src/amp/lib/experiment.ts
```diff
@@ -0,0 +1,137 @@
+interface ExperimentModel<VariantData> {
```
sndrs, line 1: sometimes more natural language for names can make them easier to read when you first see the code (if it doesn't remove too much context). e.g.
```suggestion
interface Experiment<Variant> {
```

src/amp/lib/experiment.ts
```diff
@@ -0,0 +1,137 @@
...
+    sticky: boolean;
+    consentNotificationId?: string;
+    variants: {
+        [variantName: string]: VariantData;
+    };
+}
+
+type AmpModel = ExperimentModel<number>;
```
sndrs, line 9:
```suggestion
type AmpExperiment = Experiment<number>;
```

etc

src/amp/components/Experiment.tsx
buck06191:
```typescript
interface ExperimentCollection<ExperimentObject> {
    [experimentName: string]: ExperimentObject;
}

export type AmpModelCollection = ExperimentCollection<AmpModel>;

export type StyledModelCollection = ExperimentCollection<StyledExperimentModel>;
```
`AmpModelCollection` and `StyledModelCollection` are objects containing data for each AB Test we'd be running. Google requires that we add the data to the `amp-experiment` tag in the format I've called `AmpModelCollection`, which is basically `{ExperimentName: ExperimentConfig}`
In a bid to try and keep the styles associated with a given experiment and variant close to the definition of the variant, `StyledModelCollection` puts the CSS string for the variant style as a property of the `variant` field in the `StyledModel`.

The data model I've ended up with here is arguably way too complex given that it started as an attempt to keep things simple. Ideally we'd have something like the following.
## Data Model ##
### Experiment Definition ###
```json
{
      "button-color-experiment": {
"sticky": false,
        "variants": {
          "0": 30,
          "1": 30,
          "2": 30
        }
      }
    }
```
### Style definition ###
```json
{
      "button-color-experiment": {
          "0": "style1",
          "1": "style2",
          "2": "style3"
        }
    }
```
### Overall definition of an experiment ###
 Combination of the above in something tidy and not confusing...

I'm not sure the best way to define that data model without it getting inevitably complex.

src/amp/lib/experiment.ts
buck06191: This is a great spot and suggestion. Thanks

PR #1391, opened 2020-04-22
Documentation: Update detailed set up guide

shtukas:
## What does this change?

Make the detailed set up guide slightly more useful.


SiAdcock:
@shtukas it looks like this PR checks a package-lock.json into the repo. I think this might be by mistake!

liywjl:
good catch @SiAdcock, this should not have a lock

shtukas:
Oh. Thanks Simon. Give me one second...

PR #1392, opened 2020-04-22
Reactor Nav components

liywjl:
## What does this change?
Remove NoJS conditional rendering from Nav components and refactor the components to not use React Class components

## Why?
We plan on fixing NoJS support for Nav, but in order to make the changes smaller and easier to review we are starting with a min refactor of the components.

We are removing the conditional rendering that was used to enable NoJS as it doesn't seem to work, this will make the components smaller and easier to reason with once we make changes to support NoJS

https://trello.com/c/m6Bej36R/1511-refactor-nav-components

### DIFF
--- a/src/web/components/Nav/ExpandedMenu/Column.tsx
+++ b/src/web/components/Nav/ExpandedMenu/Column.tsx
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React, { useState } from 'react';
 import { css, cx } from 'emotion';
 import { brand, brandText, brandAlt } from '@guardian/src-foundations/palette';
 import { textSans } from '@guardian/src-foundations/typography';
@@ -9,7 +9,6 @@ import { CollapseColumnButton } from './CollapseColumnButton';
 const pillarHeight = 42;
 
 // CSS
-
 export const hideDesktop = css`
     ${from.desktop} {
         display: none;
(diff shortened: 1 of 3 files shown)


src/web/components/Nav/ExpandedMenu/Column.tsx
```diff
@@ -138,7 +137,7 @@ const pillarColumnLinks = css`
     }
 `;
 
-const hide = css`
+const hideStyles = css`
```
oliverlloyd, line 140: 👍

src/web/components/Nav/ExpandedMenu/Column.tsx
```diff
@@ -155,7 +154,7 @@ const ColumnLinks: React.FC<{
                 { [firstColumnLinks]: index === 0 },
                 { [pillarColumnLinks]: !!column.pillar },
                 {
-                    [hide]: !showColumnLinks,
+                    [hideStyles]: !showColumnLinks,
```
oliverlloyd, line 157: We don't need to change this here but just to say we have the convention of using `cx` to conditionally combine classes

liywjl:
> have this been tested for accessibility

I tested this for accessibility, and it preforms no better than master. There is some work to do to increase accessibility, though I prefer to do this in another PR. The most glaringly obvious issue is that the veggie burger is not highlighted on focus, so arnt pillars on mobile view when highlighted.

One thing that may be complex to implement if we are planning on doing NoJS is encapsulating menu selection within the dialog.

Some other issues are regarding the z-index of the top Nav advert, veggie burger z-index, and mobile background colour not being greyed out. Will look at fixing these in a future PRs

PR #1394, opened 2020-04-22
Introduce the ContentAtom Element

shtukas:
## What does this change?

This introduce a currently empty but important "catch all" new element for any possible ContentAtom that would be released to AMP due to an apparent blanket support of `ContentAtomBlockElement` here: https://github.com/guardian/frontend/pull/22517/files#diff-5775f8f47d521441655ea7d7f0ce47d8L151

### DIFF
--- a/src/amp/components/elements/ContentAtom.tsx
+++ b/src/amp/components/elements/ContentAtom.tsx
@@ -1,0 +1,7 @@
+import React from 'react';
+
+export const ContentAtom: React.FC<{
+    element: ContentAtomElement;
+}> = ({}) => {
+    return <div />;
+};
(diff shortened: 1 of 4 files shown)


src/amp/components/elements/ContentAtom.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+export const ContentAtom: React.FC<{
+    element: ContentAtomElement;
+}> = ({}) => {
+    return <div />;
```
oliverlloyd, line 6: Could we just return `null` to prevent anything rendering?

shtukas: Sure!

PR #1396, opened 2020-04-22
Support onPermalinkClick prop

oliverlloyd:
## What does this change?
This supports using an `onPermalinkClick` function to give to discussion for it to call with the comment id clicked upon

## Why?
This allows us to prevent the default of refreshing the page in favour of manually passing in the `commentIdToScrollTo` prop

https://trello.com/c/o95XCIxg/1509-jump-to-comment-causes-page-refresh

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
         "@emotion/core": "^10.0.28",
         "@guardian/automat-client": "^0.2.16",
         "@guardian/consent-management-platform": "^3.0.0",
-        "@guardian/discussion-rendering": "^2.2.4",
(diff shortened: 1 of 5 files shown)


oliverlloyd:
> github found a type issue with `src/web/components/CommentsLayout.tsx`, I suspect its been there for some time, and is beyond the scope of this PR

No, that was indeed caused by thiis PR. It was waiting for the corresponding code in discussion to be published. This is now done and I've updated this PR.

Sorry, I should have made a comment in the description about this!

PR #1397, opened 2020-04-23
Debounce useHasBeenSeen callback

andre1050:
## What does this change?
Debounces the Intersection Observer callback in the `useHasBeenSeen` hook to ensure the target element is visible/intersecting for at least 200ms before the intersection is registered.

## Why?
Because in Frontend we track the intersection of elements differently, using scroll event callbacks, but also debouncing them with a 200ms delay. Replicating the delay in DCR should help narrow the gap we currently have between views registered on Frontend and DCR.

## Notes
This hooks is also used by the Lazy component. It would be good if one of the team could confirm this change won't cause unwanted side effects there.

Co-authored with @tjmw


oliverlloyd:
By the way, I don't have a strong preference (or really any preference) for those options and I'm happy to let you choose. Just so long as we have a way to manage both use cases and don't force  delay on rendering for the Lazy content

src/web/lib/useHasBeenSeen.ts
```diff
@@ -11,11 +12,16 @@ const useHasBeenSeen = (options: IntersectionObserverInit) => {
             observer.current.disconnect();
         }
 
-        observer.current = new window.IntersectionObserver(([entry]) => {
-            if (entry.isIntersecting) {
-                setHasBeenSeen(true);
-            }
-        }, options);
+        // Debounce to ensure the node is seen for at least 200ms before the
+        // boolean is set to true
```
tjmw, line 16: :+1: I think this comment is justified and useful here. I think it also highlights the difference between what we're doing here vs what we're doing on frontend. Here the element has to intersect for at least 200ms before we fire the callback (you could still be scrolling, the callback will still fire is my understanding) whereas on frontend I think you have to have _stopped_ scrolling for 200ms and the element be intersecting for the callback to fire.

oliverlloyd:
> we could (optionally) supply a wrapper function for the callback when you invoke `useHasBeenSeen`, which in our case would be a wrapper which debounces

This is actually an interesting idea. If you could solve this with composition, leaving the original `useHasBeenSeen` untouched that'd be a great solution in terms of separation of concerns and would potentially make the code easier to manage going forward

tjmw:
@oliverlloyd yeah agreed that would be really nice. I can't think of a way of implementing a solution that keeps the `hasBeenSeen` behaviour completely separate from the debouncing. I guess because the interface to `hasBeenSeen` doesn't give us back something we can debounce - the event callback is an internal thing. You _could_ pass an optional wrapper function for the callback to the hook though, which would keep the concept of debouncing out of the hook implementation, but would require a small change to `useHasBeenSeen`. E.g. something like

```js
// In the caller
const debouncer = (fn) => debounce(fn, 200);

const [hasBeenSeen, setNode] = useHasBeenSeen(
        {
            rootMargin: '-18px',
            threshold: 0,
        },
        debouncer, // optional
    ) as HasBeenSeen;

// In useHasBeenSeen
const useHasBeenSeen = (
    options: IntersectionObserverInit,
    callbackWrapper = null, // or maybe possible to default to a no-op wrapper?
) => {
...
        const callback: IntersectionObserverCallback = ([entry]) => {
            if (entry.isIntersecting) {
                setHasBeenSeen(true);
            }
        };

        const wrappedCallback = callbackWrapper ? callbackWrapper(callback) : callback;

        observer.current = new window.IntersectionObserver(
            wrappedCallback,
            options,
        );

...
```

oliverlloyd:
@tjmw I'm guessing there won't be many, if any, other cases where a the intersection would need to be wrapped so maybe just a plain, honest prop `debounce?: boolean` is the best?

tjmw:
> @tjmw I'm guessing there won't be many, if any, other cases where a the intersection would need to be wrapped so maybe just a plain, honest prop debounce?: boolean is the best?

Yep that's completely valid, thanks for steering me back from an over-engineered tangent :-)

src/web/lib/useHasBeenSeen.ts
andre1050: That's my understanding too. This fix should take us some way there but not completely. For parity with Frontend, we'd need to actually replicate their behaviour and implement something new using scroll callbacks. But this approach is so much better it's worth trying a couple of fixes first to see how impactful they are.

andre1050:
Many thanks to @oliverlloyd for the detailed feedback and suggestions.
@tjmw I ended up going for the simpler approach of adding a new boolean parameter to the `useHasBeenSeen` hook, which we can use to wrap the callback in a debounce function. So to recap, the hook supports debouncing, but we're only using it when observing the Slot intersection, not the Lazy component intersection.

PR #1398, opened 2020-04-23
AtomEmbedMarkup and AtomEmbedUrl migration 1/3

shtukas:
## What does this change?

First part of the renaming of 

```
InteractiveMarkupBlockElement
InteractiveUrlBlockElement
```

Into

``` 
AtomEmbedMarkupBlockElement
AtomEmbedUrlBlockElement
```

## Why?

Both the Scala classes and the DCR Elements they are referring to were born as part of the general Interactive Atom work, but are now used for both Interactives and Charts and are in fact (and will be more and more) used for general embeds.


shtukas:
Screenshot 2020-04-23 at 17 12 03

shtukas:
Also tested on CODE:

Screenshot 2020-04-24 at 08 52 01

PR #1399, opened 2020-04-23
Desktop nav NoJS support

liywjl:
## What does this change?
We are enabling NoJS support for the desktop Nav (mobile will need to be done in another PR). We achieve this by using checkbox input and CSS.

## Why?
Nav intended to support NoJS, it also will allow us to SSR the Nav and to potentially remove a big chunk of the `window.guardian` object (estimated ~20kb ziped  reduction)

## How do we achieve this with NoJS?
We used `~` in CSS to allow you to apply styles to sibling DOM elements. The technique was to use an ID for the input checkbox and then use the CSS `:checked` descriptor combined with `~` to then apply the new style to sibling elements. 

ex:
```
<input type="checkbox" id="someCheckbox" />
<h1>Hello World</h1>

#someCheckbox ~ h1 {
  color: red;
}
```

If the target element is nested you then need to explicitly state the path
ex:
```
<input type="checkbox" id="someCheckbox" />
<div>
  <h1>Hello World</h1>
</div>

#someCheckbox ~ div h1 {
  color: red;
}
```

It is also important to note that this only works for DOM elements that `preceded`  the input
ex:
```
// NOTE: This will NOT work
<div>
  <h1>Hello World</h1>
</div>
<input type="checkbox" id="someCheckbox" />

#someCheckbox ~ div h1 {
  color: red;
}
```

## Why do we need a new `LabelButton` component?

All the styles depend on an input checkbox, we therefore can only have 1 checkbox for the nav to toggle as each component doesnt have a local state.

We use `label` HTML elements and use `htmlFor={CHECKBOX_ID}` to link that label to the input checkbox. This means if a user clicks on the label it will trigger a checkbox toggle. But doing so mean we lose the semantic accessibility settings that would come from a button. Therefore we needed to override a lot of the properties of the HTML `label` element causing typescript and eslint to complain. Therefore it made sense to abstract this logic out of the `VeggieBurger` and `ExpandedMenuToggle`.

https://trello.com/c/cpO2xeKH/1290-fix-nav-nojs-components


src/web/components/Footer.tsx
```diff
@@ -213,7 +213,6 @@ export const Footer: React.FC<{
         <div className={pillarWrap}>
             <Pillars
                 display="standard"
-                mainMenuOpen={false}
```
liywjl, line 216: if `CHECKBOX_ID` is not set, `Pillars` assumes this is sub nav

src/web/components/Nav/Nav.tsx
```diff
@@ -41,38 +42,56 @@ type Props = {
     nav: NavType;
     display: Display;
 };
-export const Nav = ({ display, pillar, nav }: Props) => {
-    const [showExpandedMenu, toggleExpandedMenu] = useState<boolean>(false);
```
liywjl, line 45: we are removing the local state of `Nav` in favor of a input checkbox

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/LabelButton.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+
+type Props = {
+    CHECKBOX_ID: string;
+    className: string;
+    ariaControls: string;
+    children: JSX.Element | JSX.Element[];
+};
+
+export const LabelButton = ({
```
oliverlloyd, line 10: Is there a better name than 'LabelButton'? It sounds so innocent, which is risky because we probably don't want to invite future developers to use this elsewhere, given the fact it is built for a very specific purpose and contains so many overrides.

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/LabelButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+    CHECKBOX_ID,
+    className,
```
oliverlloyd, line 12: Do we need to pass a `className` prop? I'm assuming we know what the styles are so could they just be set here in this component rather than trying to make it generic?

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/VeggieBurger.tsx
```diff
@@ -20,12 +16,10 @@ const veggieBurger = ({
...
     outline: none;
-    ${until.tablet} {
-        z-index: 1;
-    }
-    ${from.tablet} {
-        z-index: ${showExpandedMenu ? 1071 : 0};
-    }
+
+    /* TODO: we should not use such a hight z-index number  */
+    z-index: 1071;
```
oliverlloyd, line 21: Shall we try changing this as part of this PR? Boy scouting and all that

src/web/components/Nav/Nav.test.tsx
```diff
@@ -1,44 +1,52 @@
...
-        const expandedMenu = getByTestId('expanded-menu');
-
-        expect(queryAllByRole('menu')).toEqual([]);
-        fireEvent.click(getByText('More'));
-        expect(expandedMenu).toHaveStyle('display: block');
-        expect(queryAllByRole('menu').length).toBeGreaterThan(1);
-        fireEvent.click(getByText('More'));
-        expect(queryAllByRole('menu')).toEqual([]);
-    });
+    // TODO: fix test
```
oliverlloyd, line 32: I would say we probably want a fairly high level of test confidence for this change. Because the solution is complex it will be brittle and a good test suite will help guard againt things breaking badly in the future.

I'd be keen to see both Jest and Cypress tests that prove as many of the key requirements as possible

src/web/components/Nav/Nav.tsx
```diff
@@ -1,6 +1,7 @@
-import React, { useState } from 'react';
+import React from 'react';
 import { css, cx } from 'emotion';
 
+import { visuallyHidden } from '@guardian/src-foundations/accessibility';
```
oliverlloyd, line 4: 👍 👍 👍

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/ExpandedMenuToggle.tsx
```diff
@@ -44,52 +48,61 @@ const text = ({ showExpandedMenu }: { showExpandedMenu: boolean }) => css`
...
             ariaControls={ariaControls}
             key="VeggieBurger"
         />
-        <button
+        <LabelButton
             className={openExpandedMenu(display)}
-            onClick={() => toggleExpandedMenu(!showExpandedMenu)}
-            aria-controls={ariaControls}
-            key="OpenExpandedMenuButton"
-            data-link-name={`nav2 : veggie-burger : ${
```
oliverlloyd, line 87: We need to preserve this logic

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/ExpandedMenuToggle.tsx
```diff
@@ -44,52 +48,61 @@ const text = ({ showExpandedMenu }: { showExpandedMenu: boolean }) => css`
...
-                showExpandedMenu ? 'show' : 'hide'
-            }`}
+            ariaControls={ariaControls}
```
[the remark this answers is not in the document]
SiAdcock: Not sure if it helps, but Heydon Pickering has long advocates [ignoring aria-controls](https://heydonworks.com/article/aria-controls-is-poop/). Would removing aria-controls make the component easier to work with? If so it would be worth checking whether the advice in this article is still accurate.

src/web/components/Nav/ExpandedMenu/ExpandedMenu.tsx
```diff
@@ -72,26 +78,21 @@ export const ExpandedMenu: React.FC<{
     display: Display;
     id: string;
     nav: NavType;
-    showExpandedMenu: boolean;
-    toggleExpandedMenu: (value: boolean) => void;
-}> = ({ display, id, nav, showExpandedMenu, toggleExpandedMenu }) => {
+    CHECKBOX_ID: string;
```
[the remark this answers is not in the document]
liywjl: how does `displayMenuCheckboxId` sound like as a more meaningful name?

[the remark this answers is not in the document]
oliverlloyd: How about `navCheckboxId` or `menuCheckboxId`? 'display' is a verb and confuses my brain when I read it

src/web/components/Nav/ExpandedMenu/ExpandedMenu.tsx
```diff
@@ -72,26 +78,21 @@ export const ExpandedMenu: React.FC<{
...
+}> = ({ display, id, nav, CHECKBOX_ID }) => {
     return (
         <>
             <ExpandedMenuToggle
                 display={display}
-                showExpandedMenu={showExpandedMenu}
-                toggleExpandedMenu={toggleExpandedMenu}
+                CHECKBOX_ID={CHECKBOX_ID}
                 ariaControls={id}
```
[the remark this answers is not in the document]
liywjl: it appears to refer to what that element is being controlled. But as mentioned [here](https://heydonworks.com/article/aria-controls-is-poop/) we should probably not use them.

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/ExpandedMenuToggle.tsx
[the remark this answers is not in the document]
liywjl: yeah I just read that, I think I will as a result remove it

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/LabelButton.tsx
liywjl: `NavLabelButton` a better name?

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/LabelButton.tsx
liywjl: No the styles are very different, I could move the styles to this component and change this to a `isVeggieBurger` boolean to conditionally render the styling?

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/VeggieBurger.tsx
liywjl: this may open up a can of worms later, I prefer to do this outside of this PR as it is already quite large, there are several areas where we have a hard coded `z-index`.

Screenshot 2020-04-27 at 11 15 35

Also I plan on doing some fixes to veggie burger and Top Nav Ad z-index, I can expand the scope of these tickets?

https://trello.com/c/y8CyFKJm/1524-top-nav-ad-and-nav-z-index-issue
https://trello.com/c/cgzbjP6u/1525-veggie-burger-z-index

src/web/components/Nav/Nav.test.tsx
liywjl: I have spent quite some time trying to get this test operational, and had no luck at all. I am happy to move it to a Cypress test instead and make sure we test both for desktop and mobile view.

I would be curious to know if we can add unit tests for accessibility or if that is a little complex

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/VeggieBurger.tsx
gtrufitt: Should maybe be handled by https://github.com/guardian/dotcom-rendering/blob/master/src/web/lib/getZIndex.tsx

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/LabelButton.tsx
oliverlloyd: `NoJSButton`?

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/VeggieBurger.tsx
oliverlloyd: Fair enough 👍

src/web/components/Nav/ExpandedMenu/ExpandedMenu.tsx
```diff
@@ -8,17 +8,23 @@ import { from, until } from '@guardian/src-foundations/mq';
...
-    }
-    ${until.desktop} {
-        transform: translateX(0%);
-    }
-`;
+const mainMenuStyles = (CHECKBOX_ID: string) => css`
+        /*
+        IMPORTANT NOTE:
+        we need to specify the adjacent path to the a (current) tag
+        to apply styles to the nested tabs due to the face we use ~
```
gtrufitt, line 15:
```suggestion
        to apply styles to the nested tabs due to the fact we use ~
```

src/web/components/Nav/ExpandedMenu/ExpandedMenu.tsx
```diff
@@ -8,17 +8,23 @@ import { from, until } from '@guardian/src-foundations/mq';
...
+        to support NoJS
+    */
+   ${`#${CHECKBOX_ID}:checked ~ & {
+        ${from.desktop} {
+            display: block;
+            overflow: visible;
+        }
+        ${until.desktop} {
+            transform: translateX(0%);
```
gtrufitt, line 24: Comment why?

src/web/components/Nav/ExpandedMenu/ExpandedMenu.tsx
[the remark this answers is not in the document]
gtrufitt: +1 menuCheckboxId (as we're in menu component, even though I think this should all be Nav not Menu 🤷‍♂️)

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/ExpandedMenuToggle.tsx
```diff
@@ -44,52 +48,49 @@ const text = ({ showExpandedMenu }: { showExpandedMenu: boolean }) => css`
...
+        /*
+            IMPORTANT NOTE:
+            we need to specify the adjacent path to the a (current) tag
+            to apply styles to the nested tabs due to the face we use ~
+            to support NoJS
+        */
+        transform: translateY(-3px) rotate(45deg);
+        ${`#${CHECKBOX_ID}:checked ~ & {
+            transform: translateY(1px) rotate(-135deg);
+        }`}
```
gtrufitt: Why doesn't this work, why the double string interpolation?

```suggestion
        ${CHECKBOX_ID}:checked ~ & {
            transform: translateY(1px) rotate(-135deg);
        }
```

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/ExpandedMenuToggle.tsx
liywjl: It is unclear to me what this does, and how to preserve this?

oliverlloyd: They're data attributes used for tracking

gtrufitt:
It seems we lose the ability to tab into the sublinks here which is an accessibility blocker (sorry if I missed this in the comments elsewhere!).

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/VeggieBurger.tsx
```diff
@@ -60,51 +50,67 @@ const veggieBurgerIcon = ({
...
+                IMPORTANT NOTE:
+                we need to specify the adjacent path to the a (current) tag
+                to apply styles to the nested tabs due to the face we use ~
+                to support NoJS
+            */
+            bottom: -6px;
+            ${`#${navCheckboxId}:checked ~ label & {
+                bottom: 0;
+                transform: rotate(45deg);
+            }`}
```
gtrufitt, line 99: These are pretty gnarly, is there a way to write them longer-hand, but with greater clarity on what's going on?

src/web/components/Nav/Nav.test.tsx
gtrufitt: I would say on the navigation component we should have good test coverage for a11y

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/VeggieBurger.tsx
```diff
@@ -60,51 +50,67 @@ const veggieBurgerIcon = ({
...
-            data-link-name={`nav2 : veggie-burger : ${
-                showExpandedMenu ? 'hide' : 'show'
-            }`}
+        <NoJSButton
+            className={veggieBurgerStyles(display)}
+            navCheckboxId={navCheckboxId}
         >
-            <span className={veggieBurgerIcon({ showExpandedMenu })} />
-        </button>
+            <span className={veggieBurgerIconStyles(navCheckboxId)} />
```
gtrufitt, line 113: Is this an empty Label? It's likely we want to have some kind of text content in here for screenreaders, unless it's defined elsewhere?

gtrufitt:
Missing this bit of flair (ok you can't really see past the gifox logo, but the arrow spinning).

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/ExpandedMenuToggle.tsx
gtrufitt: data-link-name and data-component attributes are automatically hooked into by the tracker.js script from Ophan, as Oliver mentions they're required and they need to be the exact same as Frontend.

oliverlloyd:
One of the reasons I find it hard to reason about this Nav stuff is for some reason it's developed it own entirely different and pretty wild folder structure. Why is Nav special?

Could we, as a first pass to refactoring it into line with everything else, change it to a flat structure under `/Nav`? Rather than doubling down on this nested structure it would be a lot simpler to remove it completely.

From there, some changes start to become obvious. For example: `ExpandedMenuToggle.tsx` should be called `More.tsx`. This brings it in line with it's partner, `VeggieBurger` and 'CollapseColumnButton'; is this related to the mobile view? If so, it could be better named.

liywjl:
> It seems we lose the ability to tab into the sublinks here which is an accessibility blocker (sorry if I missed this in the comments elsewhere!).
> 
> 

Great catch @gtrufitt, with the agreements that accessibility can be supported with JS, I will make sure to focus on the open menu when open.

src/web/components/Nav/ExpandedMenu/ExpandedMenu.tsx
liywjl: below there is a style `transform: translateX(-110%);` that pushes the Nav off the screen. If the checkbox is `:checked` then it is set to 0% which makes it reappear. We could look at changing this styling logic, but I kept it here as it was not the main focus. Changing this could potentially help us fix the unit test

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/ExpandedMenuToggle.tsx
liywjl: I am going at add a prop to `NoJSButton` called `dataLinkName` because this component is NoJS `showExpandedMenu ? 'show' : 'hide'` wont be possible

liywjl: also, this is not a veggie burger? this is just a Nav. Should I remove `veggie-burger` from `data-link-name`? and keep it for the label in the `<VeggieBurger />` ?

gtrufitt: I don't think so as to match up with Frontend it needs to be the same linkname when clicked and it's the same element for both desktop and mobile - 
image

liywjl: https://github.com/guardian/dotcom-rendering/pull/1399/files#diff-6b05eaaad73e58efb4df4b1cbe70e9d3R109

` data-link-name="nav2 : veggie-burger"` has been added for both buttons

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/NoJSButton.tsx
```diff
@@ -0,0 +1,127 @@
...
+                // keyCode: 32 => Space key
+                if (e.keyCode === 13 || e.keyCode === 32) {
+                    // @ts-ignore
+                    document && document.getElementById(menuCheckboxId).click();
+                }
+            }}
+            // @ts-ignore
+            tabindex={0}
+            role="button"
+            data-cy={isVeggieBurger ? 'veggie-burger' : 'show-more-button'}
```
oliverlloyd, line 121: This couples the application javascript with the test but this is an anti-pattern that can lead to brittle tests.

https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/NoJSButton.tsx
```diff
@@ -0,0 +1,127 @@
+import React from 'react';
```
oliverlloyd, line 1: This component is getting more and more complex as we try to make it support two different use cases, the veggie and non veggie buttons. I think we might have got to the point where it makes sense to split it out into two, what do you think? There would be some code duplication but it would create two much simpler components that are easier to test and easier to maintain and reason about.

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/NoJSButton.tsx
```diff
@@ -0,0 +1,127 @@
...
+    }
+    ${from.tablet} {
+        bottom: 3px;
+    }
+    ${from.desktop} {
+        display: none;
+    }
+`;
+
+const openExpandedMenu = (display: Display) => css`
```
oliverlloyd, line 50: `openExpandedMenu` is an action, a doing thing, but this variable refers to some styles. Maybe `moreStyles`?

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/NoJSButton.tsx
liywjl: I am happy to move the CSS back to their respective sections, I still think the `NoJSButton` should be its own component however.

liywjl: I removed this component in the end, moving the logic to VeggieBurger and ExpandedMenueToggle

cypress/integration/e2e/article.interactivity.spec.js
```diff
@@ -17,10 +17,20 @@ describe('Interactivity', function() {
...
             cy.visit(`/Article?url=${articleUrl}`);
-            cy.contains('Crosswords').should('not.exist');
-            cy.contains('More').click();
+            cy.contains('Crosswords').should('not.be.visible');
+            cy.get('#veggie-burger').click();
+            cy.get('#main-menu').should('be.visible'); // TODO: should select menu item when AD z-index fixed
+        });
+
+        // Mobile
+        it('should display the expanded pillar menu when More is clicked', function() {
```
oliverlloyd, line 30: Can we change these test names to be more meaningful?

cypress/integration/e2e/article.interactivity.spec.js
```diff
@@ -17,10 +17,20 @@ describe('Interactivity', function() {
...
+            cy.visit(`/Article?url=${articleUrl}`);
+            cy.contains('Crosswords').should('not.be.visible');
+            cy.get('#nav-show-more-button').click();
```
oliverlloyd, line 33: In Cypress, selecting by id is a bit of a no, no. We should be aiming to select by text or using data-cy attributes

cypress/integration/e2e/article.interactivity.spec.js
```diff
@@ -17,10 +17,20 @@ describe('Interactivity', function() {
...
         });
 
+        // Desktop
         it('should display the expanded pillar menu when More is clicked', function() {
+            cy.viewport('iphone-x');
             cy.visit(`/Article?url=${articleUrl}`);
-            cy.contains('Crosswords').should('not.exist');
-            cy.contains('More').click();
+            cy.contains('Crosswords').should('not.be.visible');
+            cy.get('#veggie-burger').click();
```
oliverlloyd, line 25: No select by id

src/model/window-guardian.ts
```diff
@@ -175,6 +175,10 @@ export const makeWindowGuardian = (
             cssIDs,
             data: {
                 ...dcrDocumentData,
+                NAV: {
+                    subNavSections: dcrDocumentData.NAV.subNavSections,
+                    currentNavLink: dcrDocumentData.NAV.currentNavLink,
+                },
```
oliverlloyd, line 181: I'm probably missing something, but if we no longer hydrate this why do we need this data on `window`?

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/ExpandedMenuToggle.tsx
```diff
@@ -33,63 +76,43 @@ const openExpandedMenu = (display: Display) => css`
...
-`;
-
 interface Props {
     display: Display;
-    showExpandedMenu: boolean;
-    toggleExpandedMenu: (value: boolean) => void;
-    ariaControls: string;
+    menuCheckboxId: string;
+    veggieBurgerId: string;
+    showMoreNavId: string;
```
oliverlloyd, line 83: I presume these values are fixed, right? So do we need to pass them as props?

src/web/components/Nav/Nav.test.tsx
```diff
@@ -34,29 +35,36 @@ describe('Nav', () => {
...
-        const expandedMenu = getByTestId('expanded-menu');
-
-        expect(queryAllByRole('menu')).toEqual([]);
-        fireEvent.click(getByText('More'));
-        expect(expandedMenu).toHaveStyle('display: block');
-        expect(queryAllByRole('menu').length).toBeGreaterThan(1);
-        fireEvent.click(getByText('More'));
-        expect(queryAllByRole('menu')).toEqual([]);
-    });
+    // it('should open and close the expanded menu by clicking More', () => {
```
oliverlloyd, line 53: Best to delete dead code now or it will live forever

src/web/components/Nav/Nav.tsx
```diff
@@ -65,12 +66,50 @@ const PositionButton = ({ children }: { children: React.ReactNode }) => (
     </div>
 );
 
-export const Nav = ({ display, pillar, nav, subscribeUrl, edition }: Props) => {
-    const [showExpandedMenu, toggleExpandedMenu] = useState<boolean>(false);
-    const mainMenuId = 'main-menu';
+// The checkbox ID is used as a CSS selector to enable NoJS support
+const menuCheckboxId = 'main-menu-toggle';
```
oliverlloyd, line 70: I think we've evolved past the point where these variables are needed. We can just inline the text

src/web/components/Pillars.tsx
```diff
@@ -215,31 +224,34 @@ const isNotLastPillar = (i: number, noOfPillars: number): boolean =>
 
 export const Pillars: React.FC<{
     display: Display;
-    mainMenuOpen: boolean;
+    menuCheckboxId?: string;
```
oliverlloyd, line 227: Might be better to have a more descriptive prop here like, `showFoobar: boolean` rather than use the id (which is fixed) as a flag

src/model/window-guardian.ts
liywjl: The only probs here are used for the SubNav which is still hydrated

src/web/components/Pillars.tsx
liywjl: we need to know the `menuCheckboxId` as it is what is used in the CSS to get the ID of the checkbox. The boolean was removed as we are supporting NoJS and we do not want components to have internal states

liywjl:
2 issues that remain:

- `:onfocus` veggie burger does not display any highlighting
- on mobile views, the Nav menu items are visible, as therefore the browser assumes they are able to focus on them. This breaks usability on mobile devices. Is this something important @gtrufitt ?

src/web/components/Pillars.tsx
oliverlloyd: But isn't `menuCheckboxId ` a static string. Why does it need to be a prop if it never changes?

oliverlloyd:
So my point about static strings is this:

```
// The checkbox ID is used as a CSS selector to enable NoJS support
const navInputCheckboxId = 'top-nav-input-checkbox';

const showMoreButtonId = 'show-more-button';
const veggieBurgerId = 'veggie-burger';
```
These variables are constants in the true form, they never change. In React props are meant to be things that can change. If we know they will never change then by not having them as props we win both on performance (no need for pointless diffs) and on readability, no need to trace the variables to see what they are and understand them.

liywjl:
> So my point about static strings is this:
> 
> ```
> // The checkbox ID is used as a CSS selector to enable NoJS support
> const navInputCheckboxId = 'top-nav-input-checkbox';
> 
> const showMoreButtonId = 'show-more-button';
> const veggieBurgerId = 'veggie-burger';
> ```
> 
> These variables are constants in the true form, they never change. In React props are meant to be things that can change. If we know they will never change then by not having them as props we win both on performance (no need for pointless diffs) and on readability, no need to trace the variables to see what they are and understand them.
> 
> The counter argument to this is: but what if I want to change them?
> I say, you never will but if you do then do a find and replace

The reason for these constants was to create some coupling between these components, as they do depend on one another, and breaking that contract should be flagged in some way.

I see your point, these wont change, I still think this way is a little safer, but I don't really mind.

oliverlloyd:
> he reason for these constants was to create some coupling between these components

I see your point. The idea is that if someone changes the id in one file it will break the code in the other one so by putting it in a variable you prevent that. This is true.

If you want to achieve this then maybe you could think about exporting the name as a constant from a config file? This does what you are trying to achieve but without abusing the prop system.

Treating these strings as config params, not props, will help future developers understand your intentions better, making their lives easier.

src/web/components/App.tsx
```diff
@@ -198,15 +190,6 @@ export const App = ({ CAPI, NAV }: Props) => {
                     dataLinkName="nav2 : topbar : edition-picker: toggle"
                 />
             </Hydrate>
-            <Hydrate root="nav-root">
```
oliverlloyd, line 201: Can we also clean up `nav-root` from the type declaration too?

index.d.ts
```diff
@@ -141,6 +141,11 @@ interface NavType {
     readerRevenueLinks: ReaderRevenuePositions;
 }
 
+interface NavBrowserType {
```
oliverlloyd, line 144: Would it make more sense to call this `SubNavBrowserType` - calling it `NavBrowserType` confused me as I thought it was meant for `Nav`

index.d.ts
```diff
@@ -557,7 +562,7 @@ interface DCRBrowserDocumentData {
     page: string;
     site: string;
     CAPI: CAPIBrowserType;
-    NAV: NavType;
+    NAV: NavBrowserType;
```
oliverlloyd, line 565: Having this as one `NAV` object doesn't really make sense anymore if the object only has 2 properties and isn't the same as `NAV` on the server

index.d.ts
oliverlloyd: But thinking about it, it makes a lot more sense to not have a NavType anymore because we're no longer sending Nav done. We're just sending `currentNavLink` and `subNavSections` which could live by themselves?

src/model/window-guardian.ts
oliverlloyd, line 181: As above, we've removed the need to pass the 'NAV' object so these properties can just be put on CAPI?

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/ExpandedMenuToggle.tsx
```diff
@@ -33,63 +77,27 @@ const openExpandedMenu = (display: Display) => css`
...
+            className={openExpandedMenuStyles(display)}
+            aria-label="Toggle main menu"
             key="OpenExpandedMenuButton"
-            data-link-name={`nav2 : veggie-burger : ${
-                showExpandedMenu ? 'show' : 'hide'
-            }`}
+            htmlFor={navInputCheckboxId}
+            data-link-name="nav2 : veggie-burger: show"
+            // @ts-ignore
+            tabindex={0}
```
gtrufitt, line 94: tabIndex (with the I)

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/VeggieBurger.tsx
```diff
@@ -60,51 +24,109 @@ const veggieBurgerIcon = ({
...
 
     return css`
+        background-color: currentColor;
+        /*
+            IMPORTANT NOTE:
+            we need to specify the adjacent path to the a (current) tag
+            to apply styles to the nested tabs due to the fact we use ~
+            to support NoJS
+        */
+        /* stylelint-disable */
```
gtrufitt, line 34:
```suggestion
		/* stylelint-disable-next-line selector-type-no-unknown */
```

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/VeggieBurger.tsx
```diff
@@ -60,51 +24,109 @@ const veggieBurgerIcon = ({
...
+        ${`#${navInputCheckboxId}`}:checked ~ label & {
+            background-color: transparent;
+        }
+        /* stylelint-enable */
```
gtrufitt, line 38:
```suggestion
```

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/VeggieBurger.tsx
```diff
@@ -60,51 +24,109 @@ const veggieBurgerIcon = ({
...
-            `
-                : 'top: -6px;'};
+            top: -6px;
+            /* refer to comment above */
+            /* stylelint-disable */
+            ${`#${navInputCheckboxId}`}:checked ~ label & {
+                top: 0;
+                transform: rotate(-45deg);
+            }
+            /* stylelint-enable */
```
gtrufitt, line 57:
```suggestion
			/* stylelint-disable-next-line selector-type-no-unknown */
            ${`#${navInputCheckboxId}`}:checked ~ label & {
                top: 0;
                transform: rotate(-45deg);
            }
```

src/web/components/Nav/ExpandedMenu/ExpandedMenuToggle/VeggieBurger.tsx
```diff
@@ -60,51 +24,109 @@ const veggieBurgerIcon = ({
...
-            `
-                : 'bottom: -6px;'};
+            bottom: -6px;
+            /* refer to comment above */
+            /* stylelint-disable */
+            ${`#${navInputCheckboxId}`}:checked ~ label & {
+                bottom: 0;
+                transform: rotate(45deg);
+            }
+            /* stylelint-enable */
```
gtrufitt, line 69:
```suggestion
            /* stylelint-disable-next-line selector-type-no-unknown */
            ${`#${navInputCheckboxId}`}:checked ~ label & {
                bottom: 0;
                transform: rotate(45deg);
            }
```

liywjl:
> Is this in hand?
> 
> 

Yes I wanted to do that in another PR https://trello.com/c/70zV5gbi/1523-support-nojs-mobile-view

essentially we will be doing that same input trick to get that effect to work in NoJS

gtrufitt:
> Yes I wanted to do that in another PR

Doesn't that mean we will push a broken nav on mobile to users though?

liywjl:
> Doesn't that mean we will push a broken nav on mobile to users though?

Oh I though you meant this was not working on NoJS, but you are right it is not working all-together. Good catch!

I will therefore implement the NoJS implementation of this menu

oliverlloyd:
> > Is this in hand?
> > 
> 
> Yes I wanted to do that in another PR https://trello.com/c/70zV5gbi/1523-support-nojs-mobile-view
> 
> essentially we will be doing that same input trick to get that effect to work in NoJS

I added a failing test for expanding a section in the mobile view

src/model/window-guardian.ts
liywjl: I think using the same approach we used to support NoJS we could apply this to the Subnav, I do not think it belongs in CAPI

oliverlloyd:
I'm not able to tab through the mobile menu. It seems as though these elements can't be focussed at all.

Also the VeggieBurger is no longer turning into a cross on click

liywjl:
## Launch Checklist

We need to add cypress tests to test accessibility. Most of these tests need to be done on mobile view and pertain to how the user would step through the page.

[] - change `z-index` top Nav ad overlaying on menu
[] - view, focus veggie burger, keydown tab should not jump to menu item
[] - when focusing a menu item, keydown esc should focus menu button
[] - focus should be locked on open menu
[] - can only focus on sub menu items that are visible

src/web/components/Nav/ExpandedMenu/Column.tsx
```diff
@@ -213,83 +163,102 @@ const columnStyle = css`
...
+                as we use ~ to apply certain styles on checkbox checked and ~ can only
+                apply to styles with elements that are preceded
+            */}
+            <input
+                type="checkbox"
+                className={css`
+                    ${visuallyHidden};
+                `}
+                id={columnInputId}
+                // aria-controls={ariaControls}
```
oliverlloyd, line 213: Is this needed?

tsconfig.json
```diff
@@ -66,5 +66,6 @@
         /* Experimental Options */
         // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
         // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
-    }
+    },
+    "exclude": ["cypress"]
```
oliverlloyd, line 70: What's the reason for adding this here?

liywjl: good question, in order to get the `tab`action to work in cypress we had to add this experimental packages: `cypress-plugin-tab`. However after adding it I started getting type issues on all the unit tests in the repo. After trying out several solutions, I found out that getting TS to ignore the cypress folder fixed the problem. As cypress does not support TS, i think it should not affect anything anyway.

src/web/components/Nav/ExpandedMenu/Column.tsx
liywjl: good catch, I should delete this

paperboyo:
Wow! This is so impressive. I cannot test, but just looking at (pretty) pics, I don’t think the text here (“Opinion”) should be selectable? Although this is mobile, so not such a problem, I guess?

liywjl:
> Wow! This is so impressive. I cannot test, but just looking at (pretty) pics, I don’t think the text here (“Opinion”) should be selectable? Although this is mobile, so not such a problem, I guess?
> 
> 

Options should be selectable on mobile view. I am a little concerned as when you click the veggie burger you should have focus on the news tab 1st, and in your gif you can see that you are not focused on that element. Could you indicate what browser you are running this on? also are you running the page without JS?

paperboyo:
> Options should be selectable on mobile view.
&
> Could you indicate what browser you are running this on? also are you running the page without JS?

I **just** meant that the text should not be selectable (yellow highlight on Opinion). I am not running the branch, I just repasted Oliver’s GIF :-). Also, text being selectable is not such a problem on mobile as it’s not possible to drag a mouse-downed cursor through them. To see what I mean, try to select text on current dotcom nav on desktop (they aren’t selectable, they are draggable, which is better, but maybe they should just stay put).
Also, this is an extremely minor point!

liywjl:
> I **just** meant that the text should not be selectable...
This brings up a good point. Yes It appears that the pillars on mobile ('Opinion") do not change colour when highlighted. This however means we loose accessibility, as the user who would tab though the page would not know if they have hit that item. 

For this reason I feel we need to have some indication of what is focused. Should we instead try use Pillar colour when highlighting each element?

I am happy to remove the onHover styles though.

paperboyo:
> This however means we loose accessibility, as the user who would tab though the page would not know if they have hit that item.

Accessibility should trump whatever I say, obviously (and I’m no expert here). But on current gen desktop, hover highlight seems to indicate tab focus and the text of the options is correctly non-selectable. On mobile… can we even tab (when connecting physical keyboard, I guess)?

I will shut up now. My whole point was only that buttons should not have selectable text as this is confusing when one mistakenly drags over them with a mouse.

oliverlloyd:
> > I **just** meant that the text should not be selectable...
> > This brings up a good point. Yes It appears that the pillars on mobile ('Opinion") do not change colour when highlighted. This however means we loose accessibility, as the user who would tab though the page would not know if they have hit that item.
> 
> For this reason I feel we need to have some indication of what is focused. Should we instead try use Pillar colour when highlighting each element?
> 
> I am happy to remove the onHover styles though.

I think there's some confusion here between _selecting_ test, eg. dragging my mouse over it as though I wanted to copy it, and _focussing_ text. Mateusz is talking about selection, which we don't want and can prevent with a simple css property (from memory, `test-select: none`, but if not that, then similar)

liywjl:
> I think there's some confusion here between selecting test, eg. dragging my mouse over it as though I wanted to copy it, and focussing text. Mateusz is talking about selection, which we don't want and can prevent with a simple css property (from memory, test-select: none, but if not that, then similar)

I misunderstood that, I have made the change 🙂

PR #1401, opened 2020-04-24
AtomEmbedMarkup and AtomEmbedUrl migration 3/3

shtukas:
## What does this change?

Third part of the renaming of 

```
InteractiveMarkupBlockElement
InteractiveUrlBlockElement
```

Into

``` 
AtomEmbedMarkupBlockElement
AtomEmbedUrlBlockElement
```

Follow up of: 
1. https://github.com/guardian/dotcom-rendering/pull/1398, and 
2. https://github.com/guardian/frontend/pull/22526


shtukas:
@oliverlloyd that would be great !

shtukas:
Tested on CODE:

Screenshot 2020-04-24 at 11 32 30

oliverlloyd:
> Should be fairly simple to do this

Why do I make statements like this?

Anyway, so adding tests for elements rendered inside iframes was not easy but is now working and is part of this PR. I created a new 'elements' spec file so we have a place to put all in body element focussed tests so we can now extend the same test method for web iframes once we have support there.

gtrufitt:
> Anyway, so adding tests for elements rendered inside iframes was not easy but is now working and is part of this PR. I created a new 'elements' spec file so we have a place to put all in body element focussed tests so we can now extend the same test method for web iframes once we have support there.

+1 This is great to have!

cypress/integration/e2e/article.elements.spec.js
```diff
@@ -0,0 +1,40 @@
...
+                    // chaining more Cypress commands, like ".find(...)"
+                    // https://on.cypress.io/wrap
+                    .then(cy.wrap)
+            );
+        };
+
+        it('should render the corona embed', function() {
+            // https://api.nextgen.guardianapps.co.uk/embed/atom/interactive/interactives%2F2020%2F03%2Fcovid-19-uk%2Fdefault#amp=1
+            cy.server();
+            // Listen for share count call
```
gtrufitt, line 27: Is this the wrong comment?

cypress/integration/e2e/article.elements.spec.js
```diff
@@ -0,0 +1,40 @@
...
+            cy.route({
+                method: 'GET',
+                url: '**/interactives/**/covid-19-uk/**',
+            }).as('getCoronaEmbed');
```
gtrufitt, line 31: Don't think you need the alias?

gtrufitt: What's this route listener actually for?

oliverlloyd: > What's this route listener actually for?

Thank you. I completely missed that I wasn't using this anymore!

PR #1403, opened 2020-04-24
Dynamically set CDN path using webpack

liywjl:
## What does this change?
specify the entire publicPath of the asset in webpack

## Why?
when it comes to dynamic import, webpack determines the path of each chuck so we need to specify the full path or it will fall back to the URL

https://webpack.js.org/configuration/output/#outputpublicpath

https://trello.com/c/v8XUWxzT/1528-dynamic-cdn-webpack


liywjl:
this should be incorporated into the dynamic import PR

PR #1405, opened 2020-04-24
Comments should fill the available space

oliverlloyd:
## What does this change?
Removes the width limit that was being set on the comments container when the reader was logged in

## Why?
So the comments can fill the available space

https://trello.com/c/xTwk3XYQ/1493-skinny-commenting-container


oliverlloyd:
> such a simple fix, I got lost working on this issues

It helps that I made the error in the first place

PR #1321, opened 2020-04-06
Implement Picnic ad slot recommendations

buck06191:
##  What does this change?
Add ad slot format details as per Picnic's recommendations. The result of these changes is that we will add the `data-multi-size` property to ad tags in AMP and increase the primary ad size to be 320x480.

This will have the effect of making the primary ad slot larger than the current size of 300x250. If the new larger ad slot cannot be filled with an ad of that size there are two resulting situations:

1. If the ad renders *below the viewport*, the runtime automatically resizes the ad below the line, with not visual reflow for the user.

2. If the ad renders *inside the viewport*, the runtime will centre the ad in the middle of the reserved ad space.

Given that picnic is intended to serve ads mid-article, they should appear below the fold meaning that situation 1) is the most likely.

Information about `multi-data-size` can be found [here](https://medium.com/ampfuel/increase-ad-competition-with-multi-size-fluid-ads-in-amp-9cff90066c16)

## Why?
As per the [Trello card](https://trello.com/c/CoEW5KzV/535-picnic-update-amp-placements-to-run-320-x-480-size), there is significant revenue opportunity available if we can introduce these changes successfully.

## Link to supporting Trello card
https://trello.com/c/CoEW5KzV/535-picnic-update-amp-placements-to-run-320-x-480-size

PR #1418, opened 2020-04-28
Implement picnic ad slot on AMP

buck06191:
_Note: This is a clean version of https://github.com/guardian/dotcom-rendering/pull/1321, which broke after a bad rebase._
## What does this change?
Add ad slot format details as per Picnic's recommendations. The result of these changes is that we will add the data-multi-size property to ad tags in AMP and increase the primary ad size to be 320x480.

This will have the effect of making the primary ad slot larger than the current size of 300x250. If the new larger ad slot cannot be filled with an ad of that size there are two resulting situations:

If the ad renders below the viewport, the runtime automatically resizes the ad below the line, with not visual reflow for the user.

If the ad renders inside the viewport, the runtime will centre the ad in the middle of the reserved ad space.

Given that picnic is intended to serve ads mid-article, they should appear below the fold meaning that situation 1) is the most likely.

Information about multi-data-size can be found here

## Why?
As per the Trello card, there is significant revenue opportunity available if we can introduce these changes successfully.

https://trello.com/c/CoEW5KzV/535-picnic-update-amp-placements-to-run-320-x-480-size

### DIFF
--- a/src/amp/components/Ad.tsx
+++ b/src/amp/components/Ad.tsx
@@ -6,7 +6,40 @@ import { palette } from '@guardian/src-foundations';
 import { textSans } from '@guardian/src-foundations/typography';
 import { adJson, stringify } from '@root/src/amp/lib/ad-json';
 
-const adStyle = css`
+const picnicAdContainerStyle = css`
+    background: ${palette.neutral[93]};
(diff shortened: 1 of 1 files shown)


buck06191:
This needs integrating with the work on `amp-experiment` https://github.com/guardian/dotcom-rendering/pull/1390 in order to separate out the two ad styles.

buck06191:
Closing this until we're able to AB test AMP pages

PR #1426, opened 2020-04-29
Bugfix for IE11 flexbox

jlieb10:
## What does this change?
Defaults the `Flex` container to have a positive grow property in the case of IE10 or IE11

## Why?
There's an IE bug causing the `Flex` container to collapse on initialisation, particularly when it is inside another container which has the property `display: flex` on it.

https://trello.com/c/IaAtcy0x/1484-layout-issues-in-onwards-in-ie11

IE Bug:
ie 11 broken

IE "Fix":
ie11 fix

### DIFF
--- a/src/web/components/Flex.tsx
+++ b/src/web/components/Flex.tsx
@@ -17,6 +17,8 @@ export const Flex = ({
             display: flex;
             flex-direction: ${direction};
             justify-content: ${justify};
+            /* Fixes IE 10/11 bug that collapses this container by default: */
+            -ms-flex-positive: 1;
         `}
     >
         {children}
--- a/src/web/components/Flex.tsx
+++ b/src/web/components/Flex.tsx
@@ -12,6 +12,7 @@ export const Flex = ({
     direction = 'row',
     justify = 'space-between',
 }: Props) => (
+    /* eslint-disable property-no-vendor-prefix */
     <div
         className={css`
             display: flex;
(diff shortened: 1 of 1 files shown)


src/web/components/Flex.tsx
```diff
@@ -17,6 +17,9 @@ export const Flex = ({
             display: flex;
             flex-direction: ${direction};
             justify-content: ${justify};
+            /* Fixes IE 10/11 bug that collapses this container by default: */
+            /* stylelint-disable-next-line property-no-vendor-prefix */
+            -ms-flex-positive: 1;
```
jfsoul, line 22: Not sure it makes any difference here, but is it generally good to have the browser specific prefixed properties first so that standards ones override it in browsers that understand both?

liywjl: Emotion compiles down CSS to support all browsers, I am unsure how it handles flexbox but I do know that IE does not understand `flex: 1` but instead needs `flex: 1 1 auto`

jlieb10: @jfsoul I don't think the order matters in this case, because we are only using the prefix to instruct specific behaviour for IE rather than to make a general rule that should apply across all browsers, if that makes sense

@liywjl I don't think that's true in this case. The compiled CSS only has "-ms-flex-positive" in it, not "flex-grow" for other browsers, so it is ignored in the case of other browsers, which is the desired behaviour.

PR #1427, opened 2020-04-30
Create lighthouse ci workflow action

jlieb10:
## What does this change?
This will allow us to see the lighthouse scores of our builds to be aware if we are regressing before pushing to master

## Why?
We want to keep dotcom-rendering ultra performant and accessible

https://trello.com/c/Rgtlap7f/1582-implement-lighthouse-regression-testing-github-integration

### DIFF
--- a/makefile
+++ b/makefile
@@ -60,7 +60,7 @@ monitor:
 logs:
 	@pm2 logs
 
-run: build stop start
+run: build stop start-without-logs
 
 # dev #########################################
 
--- a/makefile
+++ b/makefile
@@ -30,7 +30,14 @@ deploy:
 
 # prod #########################################
 
-build: clean-dist install
+#  Setting "TERM" is required for lighthouse-ci github action because it relies on 'clear'
+#  From stackoverflow: https://stackoverflow.com/questions/19425727/how-to-remove-term-environment-variable-not-set
+#  Running a program that demands a terminal via cron can lead to problems; it won't have a terminal when it is run by cron.
+build:
+	@export TERM=${TERM:-dumb}
+	@clear
+	@make clean-dist
+	@make install
 	$(call log, "building production bundles")
 	@NODE_ENV=production webpack --config scripts/webpack/frontend
 
--- a/makefile
+++ b/makefile
@@ -60,7 +60,7 @@ monitor:
 logs:
 	@pm2 logs
 
-run: build stop start-without-logs
+run: build stop start
 
 # dev #########################################
 
--- a/makefile
+++ b/makefile
@@ -52,7 +52,7 @@ monitor:
 logs:
 	@pm2 logs
 
-run: stop build start
+run: build stop start
 
 # dev #########################################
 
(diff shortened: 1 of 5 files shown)


jfsoul:
@Jpsi you might be interested in this too

jlieb10:
I need to figure out how to configure this to use the appropriate URL

oliverlloyd:
> I need to figure out how to configure this to use the appropriate URL

This could be tricky. In dev mode you can make GET requests against DCR so it works as a webserver in the classic sense. But in production mode, the one you want to test here, it only accepts POSTS and depends on the existence of `frontend` to serve the page.

We get around this when running Cypress tests by testing in dev mode like this:

```
@NODE_ENV=development start-server-and-test 'node scripts/frontend/dev-server' 3030 'cypress run --spec "cypress/integration/e2e/**/*"'
```

This isn't great but works.

I wonder if we need to think about adding an endpoint to DCR to act as a webserver. It would be somewhat odd though.

1. User (CI test process) hits DCR GET endpoint with article path
2. DCR calls out to `frontend` to ask for the article
3. Frontend would call CAPI, make some changes, and then POST the result to DCR
4. DCR returns the rendered html to Frontend
5. Frontend returns the same back to the original request from DCR
6. DCR returns html to user

Hmmm, that's not confusing at all!

jlieb10:
@oliverlloyd l'm a bit lost on that, to be honest. Is it something you'd be able to work on with me? Thanks a lot.

oliverlloyd:
> @oliverlloyd l'm a bit lost on that, to be honest. Is it something you'd be able to work on with me? Thanks a lot.

Absolutely.

To be honest, it still confuses me sometimes but lets have a chat directly to lay it all out. @gtrufitt maybe you can join as well as I have a couple of questions myself.

liywjl:
I think this is a common patter to have when you have a review apps that auto build on each PR. Maybe this is something we could explore? its a nice way of testing the PR without having to run it locally

gtrufitt:
> @gtrufitt maybe you can join as well as I have a couple of questions myself.

Lets chat after standup!

makefile
```diff
@@ -30,7 +30,7 @@ deploy:
 
 # prod #########################################
 
-build: clear clean-dist install
```
liywjl, line 33: why remove `clear`? this would cause us to bloat the `/dist` folder

liywjl:
🎊 looks like lighthouse works on the CI!

liywjl:
where is the output of lighthouse found?

jlieb10:
@liywjl I am hoping to get it working first then worry about where the results go or setting a pass/fail threshold

makefile
jlieb10: @liywjl it relies on the commands being run with a $TERM set, which is not the case for Github actions. Another option is set up a custom 'lighthouse build' command

oliverlloyd:
Closing this for now. We can reopen if this is still needed

PR #1430, opened 2020-05-01
Tablet right col position

oliverlloyd:
## What does this change?
Adjusts the position of the right colum below `leftCol`

### Before
Screenshot 2020-05-01 at 11 20 44

### After
Screenshot 2020-05-01 at 11 18 04

## Why?
For this breakpoint, below `leftCol` but above the point where the column is hidden, we had a problem where the content was being pushed to the left. This was caused by the fact the grid still had hidden items on the page and they has gaps 10px left and right of them causing a 20px space to be filled on the right of the page.

We still want to continue using grid gaps so the solution was to manually adjust the margins in this situation.

### DIFF
--- a/src/web/components/RightColumn.tsx
+++ b/src/web/components/RightColumn.tsx
@@ -15,6 +15,14 @@ const hideBelowDesktop = css`
         flex-basis: 300px;
         flex-grow: 0;
         flex-shrink: 0;
+        margin-left: 20px;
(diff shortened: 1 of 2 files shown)


oliverlloyd:
[Right Column in wrong place](https://trello.com/c/DCXj2opS/1586-right-column-in-wrong-place)

PR #1431, opened 2020-05-01
Hide Border

oliverlloyd:
## What does this change?
Hides the `Border` when it's not in use

### Before
Screenshot 2020-05-01 at 11 07 23

### After
Screenshot 2020-05-01 at 11 06 52

## Why?
We were still displaying the `Border` on the page even when it wasn't included in the `grid`

### DIFF
--- a/src/web/components/Border.tsx
+++ b/src/web/components/Border.tsx
@@ -2,17 +2,15 @@ import React from 'react';
 import { css } from 'emotion';
 
 import { border } from '@guardian/src-foundations/palette';
-import { until } from '@guardian/src-foundations/mq';
+import { from } from '@guardian/src-foundations/mq';
 
 export const Border = () => (
     <div
         className={css`
-            ${until.leftCol} {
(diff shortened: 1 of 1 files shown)


oliverlloyd:
[Hide Border](https://trello.com/c/v1N5auXU/1587-hide-border)

liywjl:
the screenshot shows the border has turned into dots, so its still there? 

Screenshot 2020-05-01 at 11 51 56

liywjl:
The diffs seem to be unrelated to this change, I am surprised they did not pick up this change

oliverlloyd:
> the screenshot shows the border has turned into dots, so its still there?
> 
> Screenshot 2020-05-01 at 11 51 56

Impressive 👀 

Fixed

oliverlloyd:
> The diffs seem to be unrelated to this change, I am surprised they did not pick up this change

I think this is because we don't take snapshots at this breakpoint?

PR #1432, opened 2020-05-01
Adjust use of negative margins below mobileLandscape

oliverlloyd:
## What does this change?
Prevents content expanding outside the viewport

## Why?
To prevent a horizontal scrollbar appearing


oliverlloyd:
[Width on Mobile](https://trello.com/c/bL1UAUHG/1572-width-on-mobile)

PR #1433, opened 2020-05-01
prefetch/preconnect performance

jlieb10:
## What does this change?
Duplicating performance enhancing behaviour from DCP onto DCR

## Why?
To increase the performance of DCR

https://trello.com/c/2PGEM8c9/1581-use-prefetch-and-preconnect-in-dcr-the-same-way-it-is-being-used-on-dcp

### DIFF
--- a/src/web/server/htmlTemplate.ts
+++ b/src/web/server/htmlTemplate.ts
@@ -103,22 +103,33 @@ export const htmlTemplate = ({
 
     const twitterMetaTags = generateMetaTags(twitterData, 'name');
 
-    // Duplicated performance tags from DCP:
-    const staticPrefetchPreconnectTags = `
-        <link rel="dns-prefetch" href="https://assets.guim.co.uk/">
-        <link rel="dns-prefetch" href="https://i.guim.co.uk">
-        <link rel="dns-prefetch" href="https://api.nextgen.guardianapps.co.uk">
-        <link rel="dns-prefetch" href="https://hits-secure.theguardian.com">
-        <link rel="dns-prefetch" href="//j.ophan.co.uk">
-        <link rel="dns-prefetch" href="//ophan.theguardian.com">
-        <link rel="dns-prefetch" href="//phar.gu-web.net">
-        <link rel="dns-prefetch" href="//www.google-analytics.com">
-        <link rel="preconnect" href="https://assets.guim.co.uk/">
-        <link rel="preconnect" href="https://i.guim.co.uk">
-        <link rel="preconnect" href="//interactive.guim.co.uk">
-        <link rel="preconnect" href="//www.google-analytics.com">
-        <link rel="dns-prefetch" href="//sb.scorecardresearch.com">
-    `;
+    // Duplicated prefetch and preconnect tags from DCP:
+    const staticPrefetchUrls = [
+        `https://assets.guim.co.uk/`,
+        `https://i.guim.co.uk`,
+        `https://api.nextgen.guardianapps.co.uk`,
+        `https://hits-secure.theguardian.com`,
+        `//j.ophan.co.uk`,
+        `//ophan.theguardian.com`,
+        `//phar.gu-web.net`,
+        `//www.google-analytics.com`,
+        '//sb.scorecardresearch.com',
+    ];
+
+    const staticPreconnectUrls = [
+        `https://assets.guim.co.uk/`,
+        `https://i.guim.co.uk`,
+        `//interactive.guim.co.uk`,
+        `//www.google-analytics.com`,
+    ];
+
+    const prefetchTags = staticPrefetchUrls.map(
+        src => `<link rel="dns-prefetch" href="${src}">`,
+    );
+
+    const preconnectTags = staticPreconnectUrls.map(
+        src => `<link rel="preconnect" href="${src}">`,
+    );
 
     return `<!doctype html>
         <html lang="en">
@@ -131,7 +142,8 @@ export const htmlTemplate = ({
                 <meta name="theme-color" content="${brandBackground.primary}" />
                 <link rel="icon" href="https://static.guim.co.uk/images/${favicon}">
 
-                ${staticPrefetchPreconnectTags}
+                ${prefetchTags}
+                ${preconnectTags}
 
                 <script type="application/ld+json">
                     ${JSON.stringify(linkedData)}
--- a/src/web/server/htmlTemplate.ts
+++ b/src/web/server/htmlTemplate.ts
@@ -103,6 +103,23 @@ export const htmlTemplate = ({
 
     const twitterMetaTags = generateMetaTags(twitterData, 'name');
 
+    // Duplicated performance tags from DCP:
+    const staticPrefetchPreconnectTags = `
+        <link rel="dns-prefetch" href="https://assets.guim.co.uk/">
+        <link rel="dns-prefetch" href="https://i.guim.co.uk">
+        <link rel="dns-prefetch" href="https://api.nextgen.guardianapps.co.uk">
+        <link rel="dns-prefetch" href="https://hits-secure.theguardian.com">
+        <link rel="dns-prefetch" href="//j.ophan.co.uk">
+        <link rel="dns-prefetch" href="//ophan.theguardian.com">
+        <link rel="dns-prefetch" href="//phar.gu-web.net">
+        <link rel="dns-prefetch" href="//www.google-analytics.com">
+        <link rel="preconnect" href="https://assets.guim.co.uk/">
+        <link rel="preconnect" href="https://i.guim.co.uk">
+        <link rel="preconnect" href="//interactive.guim.co.uk">
+        <link rel="preconnect" href="//www.google-analytics.com">
+        <link rel="dns-prefetch" href="//sb.scorecardresearch.com">
+    `;
+
     return `<!doctype html>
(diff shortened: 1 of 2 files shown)


oliverlloyd:
[Use prefetch and preconnect in DCR the same way it is being used on DCP](https://trello.com/c/2PGEM8c9/1581-use-prefetch-and-preconnect-in-dcr-the-same-way-it-is-being-used-on-dcp)

src/web/server/htmlTemplate.ts
```diff
@@ -103,6 +103,34 @@ export const htmlTemplate = ({
...
+        `//www.google-analytics.com`,
+    ];
+
+    const prefetchTags = staticPrefetchUrls.map(
+        src => `<link rel="dns-prefetch" href="${src}">`,
+    );
+
+    const preconnectTags = staticPreconnectUrls.map(
+        src => `<link rel="preconnect" href="${src}">`,
+    );
```
oliverlloyd, line 132: Does order matter here?

Reading [the docs](https://developer.mozilla.org/en-US/docs/Web/Performance/dns-prefetch), it seems to indicate the `preconnect` should happen first. It's not explicit though but maybe just to be safe?

src/web/server/htmlTemplate.ts
```diff
@@ -103,6 +103,34 @@ export const htmlTemplate = ({
...
+        `//phar.gu-web.net`,
+        `//www.google-analytics.com`,
+        '//sb.scorecardresearch.com',
+    ];
+
+    const staticPreconnectUrls = [
+        `https://assets.guim.co.uk/`,
+        `https://i.guim.co.uk`,
+        `//interactive.guim.co.uk`,
+        `//www.google-analytics.com`,
```
oliverlloyd, line 123: Can we add a comment around why these are commented out? Or maybe just remove them?

src/web/server/htmlTemplate.ts
```diff
@@ -103,6 +103,34 @@ export const htmlTemplate = ({
...
+    const staticPrefetchUrls = [
+        `https://assets.guim.co.uk/`,
+        `https://i.guim.co.uk`,
+        `https://api.nextgen.guardianapps.co.uk`,
+        `https://hits-secure.theguardian.com`,
+        `//j.ophan.co.uk`,
+        `//ophan.theguardian.com`,
+        `//phar.gu-web.net`,
+        `//www.google-analytics.com`,
+        '//sb.scorecardresearch.com',
```
oliverlloyd, line 116: Can we add a comment around why these are commented out? Or maybe just remove them?

src/web/server/htmlTemplate.ts
```diff
@@ -103,6 +103,34 @@ export const htmlTemplate = ({
 
     const twitterMetaTags = generateMetaTags(twitterData, 'name');
 
+    // Duplicated prefetch and preconnect tags from DCP:
```
oliverlloyd, line 106: Is it worth adding a link to the DCP code? Or maybe we can just add a link to the mdn docs so this stands by itself?

https://developer.mozilla.org/en-US/docs/Web/Performance/dns-prefetch

Or both! 🤷

src/web/server/htmlTemplate.ts
```diff
@@ -103,6 +103,34 @@ export const htmlTemplate = ({
...
+
+    const staticPreconnectUrls = [
+        `https://assets.guim.co.uk/`,
+        `https://i.guim.co.uk`,
+        `//interactive.guim.co.uk`,
+        `//www.google-analytics.com`,
+    ];
+
+    const prefetchTags = staticPrefetchUrls.map(
+        src => `<link rel="dns-prefetch" href="${src}">`,
```
liywjl, line 127: would be great to link to an explanation of the benefits of `dns-prefetch` https://developer.mozilla.org/en-US/docs/Web/Performance/dns-prefetch

src/web/server/htmlTemplate.ts
```diff
@@ -103,6 +103,34 @@ export const htmlTemplate = ({
...
+    );
+
+    const preconnectTags = staticPreconnectUrls.map(
+        src => `<link rel="preconnect" href="${src}">`,
```
liywjl, line 131: same here, this is a great link https://css-tricks.com/using-relpreconnect-to-establish-network-connections-early-and-increase-performance/

src/web/server/htmlTemplate.ts
```diff
@@ -103,6 +103,34 @@ export const htmlTemplate = ({
...
+        `https://hits-secure.theguardian.com`,
+        `//j.ophan.co.uk`,
+        `//ophan.theguardian.com`,
+        `//phar.gu-web.net`,
+        `//www.google-analytics.com`,
+        '//sb.scorecardresearch.com',
+    ];
+
+    const staticPreconnectUrls = [
+        `https://assets.guim.co.uk/`,
```
liywjl, line 120: if the deploy is happening to CODE this will be different, refer to `dotcom-rendering/src/lib/assets.ts` and how CDN is defined

src/web/server/htmlTemplate.ts
jlieb10: they're not commented out - `//` means use `https://` if it's available or `http://` otherwise. I think probably everything is https at this point, and this is just legacy from when it was implemented on DCP. I will check that https works for all of those and update accordingly.

src/web/server/htmlTemplate.ts
jlieb10: they're not commented out - // means use https:// if it's available or http:// otherwise. I think probably everything is https at this point, and this is just legacy from when it was implemented on DCP. I will check that https works for all of those and update accordingly.

src/web/server/htmlTemplate.ts
jlieb10: oh cool - i can do that. it just did it in the same order as on DCP

src/web/server/htmlTemplate.ts
liywjl: they arent commented out, but they just do not have the `https:` defined. I would be curious as to why? are these http and not https requests?

jlieb10:
@liywjl I copied from DCP

gtrufitt:
https://github.com/guardian/frontend/pull/12935

src/web/server/htmlTemplate.ts
jlieb10: 👍

src/web/server/htmlTemplate.ts
jlieb10: 👍

src/web/server/htmlTemplate.ts
jlieb10: 👍

src/web/server/htmlTemplate.ts
liywjl, line 120: this URL can change depending on whether this is being deployed to `CODE` or to `PROD`

jlieb10: @liywjl can you confirm i've updated it correctly?

liywjl: Great change :) looks good

src/web/server/htmlTemplate.ts
oliverlloyd: Oh! Fair enough, that makes a lot more sense

src/web/server/htmlTemplate.ts
jlieb10: They probably were at some point but i've updated

oliverlloyd:
> [guardian/frontend#12935](https://github.com/guardian/frontend/pull/12935)

Thanks for the link - this helps.

Also: remember Omniture? Such a great name

src/web/server/htmlTemplate.ts
```diff
@@ -103,6 +103,40 @@ export const htmlTemplate = ({
...
+    const staticPrefetchUrls = [
+        `${CDN}`,
+        `https://i.guim.co.uk`,
+        `https://api.nextgen.guardianapps.co.uk`,
+        `https://hits-secure.theguardian.com`,
+        `https://j.ophan.co.uk`,
+        `https://ophan.theguardian.com`,
+        `https://phar.gu-web.net`,
+        `https://www.google-analytics.com`,
+        'https://sb.scorecardresearch.com',
```
gtrufitt, line 129:
```suggestion
        `https://sb.scorecardresearch.com`,
```

just because all the others are

oliverlloyd: That's odd, prettier should have picked that up

gtrufitt:
Brill! Could you run a WPT on a couple of DCR pages before merge and then when it's out to get see the difference it makes on PROD?

jlieb10:
Good idea @gtrufitt

jlieb10:
Before:
Link 1: https://www.theguardian.com/football/2020/may/01/amiens-vow-to-fight-ligue-1-relegation-after-injustice-of-season-resolution?dcr
Ran 3x, roughly the same each time:
Performance (only relevant metric): 54

Link 2:
https://www.theguardian.com/education/2020/may/01/indian-students-trapped-in-uk-by-coronavirus-actually-starving?dcr
Performance (only relevant metric): 61

Link 3:
https://www.theguardian.com/environment/2020/may/01/alarm-over-deaths-of-bees-from-rapidly-spreading-viral-disease?dcr
Performance: 59

PR #1438, opened 2020-05-03
Enhance CAPI

oliverlloyd:
## What does this change?
Adds a new step to enhance the CAPI response 

Screenshot 2020-05-03 at 11 51 44

## What types of enhancements is this for?
- Drop caps (* * *) [[fontend code](https://github.com/guardian/frontend/blob/08b3732d113c28c5165ab2ce8e9b7b52909da336/common/app/views/support/HtmlCleaner.scala#L519)]
- Divider lines (also * * * but immersive only) [[fontend code](https://github.com/guardian/frontend/blob/08b3732d113c28c5165ab2ce8e9b7b52909da336/common/app/views/support/HtmlCleaner.scala#L539)]
- Auto linking [[fontend code](https://github.com/guardian/frontend/blob/08b3732d113c28c5165ab2ce8e9b7b52909da336/common/app/views/support/HtmlCleaner.scala#L849)?]
- Unicode bullets [[fontend code](https://github.com/guardian/frontend/blob/08b3732d113c28c5165ab2ce8e9b7b52909da336/common/app/views/support/HtmlCleaner.scala#L207)]
- Data attributes [[fontend code](https://github.com/guardian/frontend/blob/08b3732d113c28c5165ab2ce8e9b7b52909da336/common/app/views/support/HtmlCleaner.scala#L227)?]
- Photo essays [[frontend code](https://github.com/guardian/frontend/blob/08b3732d113c28c5165ab2ce8e9b7b52909da336/common/app/views/support/HtmlCleaner.scala#L453)]

## Why?
This initially helps us by giving us a way to replace some of the work done in [HtmlCleaner](https://github.com/guardian/frontend/blob/08b3732d113c28c5165ab2ce8e9b7b52909da336/common/app/views/support/HtmlCleaner.scala) on frontend. Having the code here, in DCR, speeds up development.

It also helps us identify and isolate these cleaning requirements so that we can work towards the goal of moving this logic further up the stack, away from the rendering tier.

### DIFF
--- a/src/model/add-dropcaps.ts
+++ b/src/model/add-dropcaps.ts
@@ -47,7 +47,7 @@ const enhanceElements = (elements: CAPIElement[]): CAPIElement[] => {
     return enhanced;
 };
 
-export const enhanceCAPI = (data: any): CAPIType => {
+export const addDropCaps = (data: CAPIType): CAPIType => {
     const enhancedBlocks = data.blocks.map((block: Block) => {
         return {
             ...block,
--- a/src/model/add-dropcaps.ts
+++ b/src/model/add-dropcaps.ts
@@ -3,45 +3,51 @@ import { JSDOM } from 'jsdom';
 const TEXT = 'model.dotcomrendering.pageElements.TextBlockElement';
 const SUBHEADING = 'model.dotcomrendering.pageElements.SubheadingBlockElement';
 
-const isDropCapFlag = (element: SubheadingBlockElement): boolean => {
+const isDropCapFlag = (element: CAPIElement): boolean => {
     // A drop cap flag: <h2><strong>* * *</strong></h2>
+    if (element._type !== SUBHEADING) return false;
     const frag = JSDOM.fragment(element.html);
     if (!frag || !frag.firstChild) return false;
     return frag.firstChild.nodeName === 'H2' && frag.textContent === '* * *';
 };
 
-const enhanceElements = (elements: CAPIElement[]): CAPIElement[] => {
-    // enhanceElements loops the array of article elements making 'enhancements' (read hacks)
-    // to allow us to support the data model coming out of CAPI. Similar logic exists in frontend
-    // here: https://github.com/guardian/frontend/blob/08b3732d113c28c5165ab2ce8e9b7b52909da336/common/app/views/support/HtmlCleaner.scala
-    // TODO: Take these changes and push them up the stack so that we don't need to make logic
-    // overrides in the rendering tier.
-    //
-    // The current list of enhancements are:
-    //
-    // 1. Drop Cap
-    //    If a h2 tag containing * * * is found, then the proceeding element should be givin drop
-    //    cap styling (so long as it's a text element).
-    //
-    let dropCapFlagSeen = false;
+const isSubheading = (element: CAPIElement): boolean => {
+    return element._type === SUBHEADING;
+};
 
+const prevElementWasDropCapFlag = (
+    elements: CAPIElement[],
+    i: number,
+): boolean => {
+    return isSubheading(elements[i - 1]) && isDropCapFlag(elements[i - 1]);
+};
+
+const checkForDropCaps = (elements: CAPIElement[]): CAPIElement[] => {
+    // checkForDropCaps loops the array of article elements looking for drop cap flags and
+    // enhancing the data accordingly. In short, if a h2 tag is equal to * * * then the
+    // text element immediately aftwards should have dropCap set to true
     const enhanced: CAPIElement[] = [];
-    elements.forEach(element => {
-        if (element._type === TEXT && dropCapFlagSeen) {
-            // The previous element was the drop cap * * * flag so set the
-            // dropCap property to true on this, the next, element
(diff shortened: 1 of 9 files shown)


oliverlloyd:
[CAPI enhancement layer to enable html cleaning](https://trello.com/c/ngKRs7Bc/1585-capi-enhancement-layer-to-enable-html-cleaning)

shtukas:
Very nice! Also, the tests completely document what exactly any future upstream implementation must do. +1.

oliverlloyd:
@gtrufitt I refactored the enhance logic to match [your gist](https://gist.github.com/gtrufitt/1dc327066dbfbfa34a00ed61bd610549) but didn't quite make it all the way! Using a switch statement was a big improvement but I couldn't get the `(prev, current)` model to work, mainly because, you know, typescript. But I managed to get rid of the 'state' and things read a lot better now. Let me know what you think.

Oh, and I also renamed everything to be drop cap specific so we keep each 'enhancement' separate, with CAPIType contracts inbetween, so removing these later should be as simple as deleting one line and the file.

JamieB-gu:
This got me thinking... here's an immutable alternative that I _think_ should pass the tests:

```ts
const isText = (element: CAPIElement): boolean =>
    element !== undefined && element._type === TEXT;

const checkForDropCaps = (elements: CAPIElement[]): CAPIElement[] => {
    const [ element, ...tail ] = elements;

    if (element === undefined) {
        return [];
    }

    if (isDropCapFlag(element)) {
        const [ nextElement, ...tail2 ] = tail;

        if (isText(nextElement)) {
            return [ { ...nextElement, dropCap: true }, ...checkForDropcaps(tail2) ];
        }

        return checkForDropcaps(tail);
    }

    return [ element, ...checkForDropcaps(tail) ];
}
```

This may or may not be useful to you, I just thought it was an interesting problem 🙂

src/web/server/render.ts
```diff
@@ -3,12 +3,14 @@ import { extract as extractNAV } from '@root/src/model/extract-nav';
...
 import { document } from '@root/src/web/server/document';
 import { validateAsCAPIType } from '@root/src/model/validate';
+import { addDropCaps } from '@root/src/model/add-dropcaps';
 import { extract as extractGA } from '@root/src/model/extract-ga';
 import { bodyJSON } from '@root/src/model/exampleBodyJSON';
 
 export const render = ({ body }: express.Request, res: express.Response) => {
     try {
-        const CAPI = validateAsCAPIType(body);
+        const withDropCaps: CAPIType = addDropCaps(body);
```
gtrufitt, line 12: I think maybe having a hop into a `enhanceCAPI` file, then then collates all the hacks might make it easier to reason with as render.ts could get large. But yeah, not urgent for this PR.

src/model/add-dropcaps.ts
```diff
@@ -0,0 +1,68 @@
+import { JSDOM } from 'jsdom';
```
gtrufitt, line 1: Yeah, think it's much cleaner to have each 'mod' in a separate file. Does 'model' describe the folder well? Maybe `enhanceCAPI` or `hacks` - model feels a little too generic and I think it's worth have the folder name descriptive.

src/model/add-dropcaps.ts
```diff
@@ -0,0 +1,68 @@
...
+
+const TEXT = 'model.dotcomrendering.pageElements.TextBlockElement';
```
gtrufitt, line 3: Maybe just inline these?

src/web/server/render.ts
oliverlloyd: I had the same thought 👍

src/model/add-dropcaps.ts
oliverlloyd: Maybe if we call it `hacks` it's a clear message that we want to remove this code at a later stage

oliverlloyd:
> This got me thinking... here's an immutable alternative that I _think_ should pass the tests:
> 
> ```ts
> const isText = (element: CAPIElement): boolean =>
>     element !== undefined && element._type === TEXT;
> 
> const checkForDropCaps = (elements: CAPIElement[]): CAPIElement[] => {
>     const [ element, ...tail ] = elements;
> 
>     // End of the list
>     if (element === undefined) {
>         return [];
>     }
> 
>     if (isDropCapFlag(element)) {
>         const [ nextElement, ...tail2 ] = tail;
> 
>         // If TEXT, add the `dropCap` flag
>         if (isText(nextElement)) {
>             return [ { ...nextElement, dropCap: true }, ...checkForDropcaps(tail2) ];
>         }
> 
>         // Otherwise remove the dropcap and continue
>         return checkForDropcaps(tail);
>     }
> 
>     // No dropcaps here, so append the element and continue
>     return [ element, ...checkForDropcaps(tail) ];
> }
> ```
> 
> This may or may not be useful to you, I just thought it was an interesting problem 🙂

Using the `rest` trick when spreading is a clever way to go about this!

My instinct is to keep things a bit more basic (albeit more verbose) though - a lot of people don't know how to do this and it can be hard to reason about when you're not used to it

PR #1439, opened 2020-05-03
Dropcap

oliverlloyd:
Depends on https://github.com/guardian/dotcom-rendering/pull/1438

## What
Adds support for the first letter of a paragraph being given the fairytale drop cap styling

### Before
Screenshot 2020-05-03 at 12 19 03

### After
Screenshot 2020-05-03 at 12 19 29


oliverlloyd:
[We don't support the big leading letters](https://trello.com/c/P0LbnQYZ/432-we-dont-support-the-big-leading-letters)

src/model/enhance.ts
```diff
@@ -0,0 +1,62 @@
...
+    // The current list of enhancements are:
+    //
+    // 1. Drop Cap
+    //    If a h2 tag containing * * * is found, then the proceeding element should be givin drop
+    //    cap styling (so long as it's a text element).
+    //
+    let dropCapFlagSeen = false;
+
+    const enhanced: CAPIElement[] = [];
+    elements.forEach(element => {
```
gtrufitt, line 29: Hey, so I quickly tried to write a spike of what I imagined of enhance CAPI (it's not working, so assume it pseudo code). 

The main difference I had in mind was that rather than tracking through state, each 'block' has an enhance method that take the previous block and current and self contains the updating. Within that I imagined self contained methods to do the rewriting.

It's very quickly written and rough but wondered on your thoughts as an alternative?

https://gist.github.com/gtrufitt/1dc327066dbfbfa34a00ed61bd610549

gtrufitt: I actually think I still like the 'sanitize-html' approach as a way to better define an API for handling this and avoiding dom soup 

https://github.com/guardian/dotcom-rendering/blob/2616d4d30bfaed89cbdc3acd8d119c357aa6e824/src/web/components/elements/TextBlockComponent.tsx#L41

But obviously everything can't be handled with it

oliverlloyd: This is really nice:

```
let shouldDropCap = prev.type === 'subheadingBlock' && prev.content.includes('***');
```

I worry it might look less nice if the same logic needs to be aware of more than just the previous element though, like with photo essays

But if we take the one enhancer file per use case approach, then we can implement each one as we like. Only having to write code to deal with one problem is going to keep things a lot easier to reason about and also make removing it later easier too

oliverlloyd: Also, again, assuming we keep processing separate, then perhaps we could use the `sanitzer` on a case by case bases? It seems like it would work well for edits in place without the  need for context

paperboyo:
🎉 
Just re:appearance. We *should not* follow current frontend where drop caps are broken since new fonts were introduced with new vertical metrics, but the sizing was kept based on old ones. Baseline of the drop cap should align to the baseline of the body text. Ditto for the top. Latter is correct in the screenshot above, while former is wrong.

src/web/layouts/Standard.stories.tsx
```diff
@@ -25,7 +25,7 @@ mockRESTCalls();
 export default {
     title: 'Layouts/Standard',
     parameters: {
-        chromatic: { viewports: [1300], delay: 800 },
+        chromatic: { viewports: [1300], delay: 800, diffThreshold: 0.2 },
```
jfsoul, line 28: Is this related to this PR?

oliverlloyd: Not directly. But I needed to use _a_ PR to see if this change fixed this problem which has been occuring recently

PR #1441, opened 2020-05-03
Allow zero comments or shares to be displayed

oliverlloyd:
## What does this change?
Previously we didn't render anything if the comment or share count was zero, now we do

## Why?
Zero is useful information


oliverlloyd:
[0 comments removes the comment link and it shouldn't](https://trello.com/c/WSL5rkuz/1570-0-comments-removes-the-comment-link-and-it-shouldnt)

PR #1444, opened 2020-05-04
Adds support for embed blocks

rcrphillips:
## What does this change?
Screen Shot 2020-05-04 at 13 40 54
Screen Shot 2020-05-04 at 13 41 18

## Why?
Parity with frontend

### DIFF
--- a/cypress/integration/e2e/article.elements.spec.js
+++ b/cypress/integration/e2e/article.elements.spec.js
@@ -31,35 +31,30 @@ describe('Elements', function() {
     });
 
     describe('WEB', function() {
-        // https://www.cypress.io/blog/2020/02/12/working-with-iframes-in-cypress/
-        const getIframeBody = () => {
-            return cy
-                .get('div[data-cy="instagram-embed"] > iframe')
-                .its('0.contentDocument.body')
-                .should('not.be.empty')
-                .then(cy.wrap);
-        };
-
         it('should render the instagram embed', function() {
+            // https://www.cypress.io/blog/2020/02/12/working-with-iframes-in-cypress/
+            const getIframeBody = () => {
+                return cy
+                    .get('div[data-cy="instagram-embed"] > iframe')
+                    .its('0.contentDocument.body')
+                    .should('not.be.empty')
+                    .then(cy.wrap);
+            };
             cy.visit(
                 'Article?url=https://www.theguardian.com/media/2018/aug/29/flat-tummy-instagram-women-appetite-suppressant-lollipops',
             );
 
             getIframeBody().contains('View More on Instagram');
         });
-    });
-
-    describe('WEB', function() {
-        // https://www.cypress.io/blog/2020/02/12/working-with-iframes-in-cypress/
-        const getIframeBody = () => {
-            return cy
-                .get('div[data-cy="embed-block"] > div > iframe')
-                .its('0.contentDocument.body')
-                .should('not.be.empty')
-                .then(cy.wrap);
-        };
 
         it('should render the embed', function() {
+            const getIframeBody = () => {
+                return cy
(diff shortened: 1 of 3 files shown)


cypress/integration/e2e/article.elements.spec.js
```diff
@@ -48,4 +48,23 @@ describe('Elements', function() {
             getIframeBody().contains('View More on Instagram');
         });
     });
+
+    describe('WEB', function() {
```
oliverlloyd, line 52: If we keep the other 'describe WEB as the parent and nest two new ones for, say, describe 'InstagramBlockElement' and 'describe EmbedBlockComponent' then these tests will print [really pretty output](https://jestjs.io/docs/en/api#describename-fn) which makes me feel good inside.

Plus it will help make it clearer what the scope of the tests are.

Maybe like:

```
describe('WEB', function() {
    describe('for InstagramBlockElement', function() {
        // ... etc.
    })
    describe('for EmbedBlockComponent', function() {
        // ... etc.
    })
})
```

src/web/components/elements/EmbedBlockComponent.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { text } from '@guardian/src-foundations/palette';
+
+type Props = {
+    html: string;
+    alt?: string;
+};
+
+const emailCaptionStyle = css`
+    ${textSans.xsmall()};
+    word-wrap: break-word;
```
oliverlloyd, line 15: Does this degrade okay on IE11? I don't think that browser support `break-word`

https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

src/web/components/elements/EmbedBlockComponent.tsx
```diff
@@ -0,0 +1,38 @@
...
+    color: ${text.supporting};
```
oliverlloyd, line 16: Nice use of Source 👍

src/web/components/elements/EmbedBlockComponent.tsx
```diff
@@ -0,0 +1,38 @@
...
+const embedContainer = css`
+    iframe {
+        /* Some embeds can hijack the iframe and calculate an incorrect width, which pushes the body out */
+        width: 100% !important;
+    }
+    margin-bottom: 16px;
+`;
+
+export const EmbedBlockComponent = ({ html, alt }: Props) => {
+    // Email embeds are being turned into atoms, so we can remove this hack when that happens
```
oliverlloyd, line 28:
```suggestion
    // TODO: Email embeds are being turned into atoms, so we can remove this hack when that happens
```

oliverlloyd: The TODO: prefix is useful mainly because it gets highlighted by some IDEs

src/web/components/elements/EmbedBlockComponent.tsx
```diff
@@ -0,0 +1,38 @@
...
+    const isEmailEmbed = html.includes('email/form');
+    return (
+        <div data-cy="embed-block" className={embedContainer}>
+            <div dangerouslySetInnerHTML={{ __html: unescapeData(html) }} />
+            {isEmailEmbed && alt && (
+                <div className={emailCaptionStyle}>{alt}</div>
```
oliverlloyd, line 34: +1 for the meaningful name

src/web/components/elements/EmbedBlockComponent.tsx
rcrphillips: Have updated to a more supported property.

rcrphillips:
[Embed block types](https://trello.com/c/BGH5dpLV/1584-embed-block-types)

PR #1315, opened 2020-04-03
Dynamically set Epic tracking params

tjmw:
## What does this change?

Now that the contributions service provides us all the tracking params we need in the `meta` part of the response, we can use those instead of hardcoding.

## Why?

For the default epic which we're currently serving this should all be identical to the hardcoded values, but going forward once we're selecting tests/variants dynamically this is required.

## Link to supporting Trello card

https://trello.com/c/CNlYq1rK/93-update-platforms-to-use-tracking-metadata-returned-by-epic

PR #1447, opened 2020-05-04
Supporting all Epic variants outside of A/B test

andre1050:
## What does this change?
Makes the necessary changes to serve all Epic variants in `SlotBodyEnd` outside of an A/B test. The switch used is `slotBodyEnd` but it's also possible to render the slot using the URL parameter `?slot-machine-flags=showBodyEnd`.

### Before
- Slot was behind the `FrontendDotcomRenderingEpic` test switch.
- Ophan events were logging data for the `FrontendDotcomRenderingEpic` test.

### After
- Slot is behind a dedicated switch (`slotBodyEnd`) with option to override via URL params.
- Ophan events will log data specific to the Epic served by the service.

## Why?
Because we want to support all variants in DCR (behind a switch) and ensure the correct data is sent to Ophan.

## Notes
This includes all the work by @tjmw in https://github.com/guardian/dotcom-rendering/pull/1315 with some additions to include the switch handling.

Merging this PR will effectively finish the `FrontendDotcomRenderingEpic` test on DCR, but the environment variable `ENABLE_DYNAMIC_TESTS` will need to be set `true` for the Contributions service to ensure all variants get served.


andre1050:
> I think this looks good 👍
> 
> I think i did this at the time with #1315 but would be good to verify in the browser that we're sending the right tracking event data to Ophan, just to verify everything is wired up right.
> 
> I noticed we've still got this condition:
> 
> ```js
> // FIXME - used temporarily to exclude US from the DCR 5% test.
>     // Please check with Slot Machine team if this condition can be removed.
>     if (countryCode === 'US') {
>         return null;
>     }
> ```
> 
> Can we remove this now do you think? Tests can have their own geo config with the `locations` param so seems like it's probably not needed?

Thanks @tjmw I totally forgot the US exclusion but I agree we shouldn't need it anymore so I got rid of it. It makes sense to rely on the test's geo rules.

andre1050:
> > I think this looks good 👍
> > I think i did this at the time with #1315 but would be good to verify in the browser that we're sending the right tracking event data to Ophan, just to verify everything is wired up right.
> > I noticed we've still got this condition:
> > ```js
> > // FIXME - used temporarily to exclude US from the DCR 5% test.
> >     // Please check with Slot Machine team if this condition can be removed.
> >     if (countryCode === 'US') {
> >         return null;
> >     }
> > ```
> > 
> > 
> > Can we remove this now do you think? Tests can have their own geo config with the `locations` param so seems like it's probably not needed?
> 
> Thanks @tjmw I totally forgot the US exclusion but I agree we shouldn't need it anymore so I got rid of it. It makes sense to rely on the test's geo rules.

Also, forgot to mention before but I did log the `componentEvent` object we sent to Ophan and the values all seem to reflect the test being served (not the default epic hardcoded stuff).

src/web/components/SlotBodyEnd.tsx
```diff
@@ -189,11 +198,22 @@ const MemoisedInner = ({
...
         }
-    }, [hasBeenSeen]);
+        // At the moment, given the information we currently store in data, it's
+        // fine to add it as a dependency because it won't ever change after
+        // viewing, so we shouldn't ever run the hook multiple times. If we _do_
+        // ever have an implementation where information in data could change
+        // after viewing, we'll need to revisit this. Perhaps it's clearer to
+        // remove data from the dependencies and disable the exhaustive-deps
+        // rule, as above.
+    }, [hasBeenSeen, data]);
```
nicl, line 216: Yeah would much rather disable exhaustive deps for this part and improve the code clarity (and avoid the massive comment too!)

src/web/layouts/CommentLayout.tsx
```diff
@@ -216,12 +216,10 @@ export const CommentLayout = ({
 
     const adTargeting: AdTargeting = buildAdTargeting(CAPI.config);
 
-    // Render the slot if one is true:
-    // 1) The flag for this slot exists in the URL (i.e. ?slot-machine-flags=showBodyEnd)
-    // 2) The global switch for the Frontend/DCR Epic test is true
+    // Render the slot if switch is set to true or flag set in URL (i.e. ?slot-machine-flags=showBodyEnd)
```
nicl, line 219: This is a case where the code is clear enough that a comment doesn't add much.

andre1050: Fair enough, I've removed those

src/web/components/SlotBodyEnd.tsx
andre1050: I've removed the `data` object from the deps array and updated the comment.

PR #1448, opened 2020-05-04
Type additional elements

oliverlloyd:
## What does this change?
Adds types for `DocumentBlockElement` and `TableBlockElement`

### Before
Without these we were throwing validation errors causing 500 errors

### After
We still don't support these elements fully but we correctly return null preventing errors

### DIFF
--- a/src/lib/content.d.ts
+++ b/src/lib/content.d.ts
@@ -277,6 +277,11 @@ interface DocumentBlockElement {
     isMandatory: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


oliverlloyd:
[Update types for extra elements](https://trello.com/c/0SkgyDA1/1607-update-types-for-extra-elements)

gtrufitt:
Think we need to add json schema generation to a hook!

PR #1454, opened 2020-05-05
Update index.html with missing element types

shtukas:
## What does this change?
Update index.html with missing element types

## Why?
Useful


shtukas:
Screenshot 2020-05-05 at 13 49 36

PR #1457, opened 2020-05-05
Remove caption if displayCredit false

liywjl:
## What does this change?
Remove caption from `ImageBlockElement` if `displayCredit` is false

### Before
Screenshot 2020-05-05 at 17 25 05

### After
Screenshot 2020-05-05 at 17 24 44

## Why?
parity with frontend


paperboyo:
Wot Oliver said. We hide credit if `displayCredit` is `false`. We hide caption if it’s empty. We hide triangle when both aren’t there (frontend has a [bug there](https://trello.com/c/wdMLGGZu/414-triangle-visible-when-caption-empty-on-immersive-main-media)).

src/web/components/Caption.tsx
```diff
@@ -114,6 +114,7 @@ export const Caption = ({
                             <TriangleIcon />
                         </span>
                         {getCaptionHtml()} {displayCredit && credit}
+                        {displayCredit && credit}
```
oliverlloyd, line 117: This line should be removed

PR #1458, opened 2020-05-05
Caption Toggle

oliverlloyd:
## What does this change?
Adds the caption toggle to show the image caption as an overlay below `tablet`

## Why?
We don't want to push content too far down on smaller screens but it's still important that the caption information is available

## How was this implemented?
This PR uses a [css only toggle](https://www.outofscope.com/css-only-menu-toggle-no-javascript-required/). We don't use react to control this state, instead a hidden checkbox is used to hold when the caption should show or not with a `label` acting as the button to capture the click event on the checkbox

In simple terms:

```
<label for="the-checkbox">Click me to toggle menu</label>
<input type="checkbox" id="the-checkbox" />
<div id="the-caption">
  <Caption />
</div>
```

```
label {
  cursor: pointer;
  // etc.
}
#the-checkbox {
  display: none; /* hide */
}
#the-caption {
  display: none;
}
#the-checkbox:checked + #the-caption {
  display: block;
}
```

## We still show the triangle
Screenshot 2020-05-05 at 23 34 15

This triangle icon shouldn't appear. There are actually several use cases for not having it so we need an additional PR to make it an optional property for `Caption`


oliverlloyd:
[Image caption on mobile](https://trello.com/c/rw7mr1DC/979-image-caption-on-mobile)

src/web/components/Caption.tsx
```diff
@@ -93,30 +118,29 @@ export const Caption = ({
...
+        >
+            <span
+                className={cx(
+                    iconStyle,
+                    display === 'immersive' && hideIconBelowLeftCol,
+                )}
+            >
+                <TriangleIcon />
+            </span>
+            {getCaptionHtml()} {displayCredit && credit}
```
liywjl, line 143: I think `getCaptionHtml()` should be move into the rendering, having it as its own function doesn't make much sense IMO

oliverlloyd: Agreed. It's not that big and only used once. I'll do that refactor now 👍

PR #1461, opened 2020-05-06
Fix NoJS for Dropdown

oliverlloyd:
## What does this change?
This fixes the problem where the `Dropdown` component was no longer working without javascript

## Why?
To support users that don't have JS turned on

## What about just not using JS completely?
We don't want to do this because then we would lose support for the closing the menu using the escape key and also when clicking away


oliverlloyd:
[NoJS Dropdown Toggle](https://trello.com/c/au1f3myX/1614-nojs-dropdown-toggle)

PR #1469, opened 2020-05-07
AMP Component Renaming

shtukas:
## What does this change?

He we apply to amp the conventions we decided to adopt for web. See description here ( https://github.com/guardian/dotcom-rendering/pull/1463 ). The idea is to ensure that the name of the element as given by the backend, drives the canonical name of the associated component and the associated type. 

###


shtukas:
Oh yeah. Sorry. It came because moving from `YouTubeBlockComponent.tsx` to `YoutubeBlockComponent.tsx` was only a case change and the macos' file system doesn't see any difference

PR #1472, opened 2020-05-10
MultiImageBlockElement

oliverlloyd:
## What does this change?
Adds a new element type `MultiImageBlockElement` for use when multiple images should be arranged together with a single caption (optional)

### Props
```typescript
type Props = {
    designType: DesignType;
    images: ImageBlockElement[];
    caption?: string;
    pillar: Pillar;
};
```

### Element type
```typescript
interface MultiImageBlockElement {
    _type: 'model.dotcomrendering.pageElements.MultiImageBlockElement';
    images: ImageBlockElement[];
    caption?: string;
}
```

## Why?
In the short term, this prepares support for Photo Essays, in the long term it offers a more flexible api for how we insert images into an article, not just photo essays.

### DIFF
--- a/src/web/components/elements/MultiImageBlockComponent.tsx
+++ b/src/web/components/elements/MultiImageBlockComponent.tsx
@@ -1,0 +1,287 @@
+import React from 'react';
+import { css } from 'emotion';
+
+import { space } from '@guardian/src-foundations';
+import { from, until } from '@guardian/src-foundations/mq';
+
+import { ImageComponent } from '@root/src/web/components/elements/ImageComponent';
+import { Caption } from '@frontend/web/components/Caption';
+import { GridItem } from '@root/src/web/components/GridItem';
+
+type Props = {
+    designType: DesignType;
+    images: ImageBlockElement[];
+    caption?: string;
+    pillar: Pillar;
+};
+
+const ieFallback = css`
+    display: flex;
+    flex-direction: column;
+    ${until.leftCol} {
+        margin-left: 0px;
+    }
+    ${from.leftCol} {
+        margin-left: 151px;
+    }
+    ${from.wide} {
+        margin-left: 230px;
+    }
+`;
+
+const SideBySideGrid = ({
+    children,
+}: {
+    children: JSX.Element | JSX.Element[];
+}) => (
+    <div
+        className={css`
+            ${ieFallback}
+            @supports (display: grid) {
+                margin-left: 0;
+                margin-right: ${space[3]}px;
+                display: grid;
+                grid-gap: ${space[3]}px;
+                grid-template-columns:
+                    50% /* Left column */
+                    50%; /* Right column */
+                grid-template-areas: 'first second';
+            }
+        `}
+    >
+        {children}
+    </div>
+);
+
+const OneAboveTwoGrid = ({
+    children,
+}: {
+    children: JSX.Element | JSX.Element[];
+}) => (
+    <div
+        className={css`
+            ${ieFallback}
+            @supports (display: grid) {
+                margin-left: 0;
+                margin-right: ${space[3]}px;
+                display: grid;
+                grid-gap: ${space[3]}px;
+                grid-template-columns:
+                    50% /* Left column */
+                    50%; /* Right column */
+                grid-template-areas:
+                    'first first'
+                    'second third';
+            }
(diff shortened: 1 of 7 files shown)


oliverlloyd:
[Create `MultiImageBlockElement`](https://trello.com/c/s9WmQXdH/1633-create-multiimageblockelement)

src/web/components/elements/MultiImageBlockComponent.tsx
```diff
@@ -0,0 +1,287 @@
...
+                            captionText={caption}
+                            pillar={pillar}
+                            shouldLimitWidth={false}
+                        />
+                    )}
+                </div>
+            );
+        case 3:
+            return (
+                <div className={css``}>
```
liywjl, line 186: you can remove

src/web/components/elements/MultiImageBlockComponent.tsx
```diff
@@ -0,0 +1,294 @@
...
+            return (
+                <div
+                    className={css`
+                        img {
+                            object-fit: cover;
+                            width: 100%;
+                        }
+                    `}
+                >
+                    <OneAboveTwoGrid>
```
JamieB-gu, line 194: What about scenarios where it's "two above one"? Example near the bottom of [this article](https://www.theguardian.com/travel/2019/jan/25/mountain-rescue-why-bothies-need-a-helping-hand-a-photo-essay).

oliverlloyd: Dang. That example really messes up my api!

Either:
1. You just accept that when there are 3 images they will always appear 1 above 2
2. You refactor the api here to have a way to communicate the fact you want 2 above 1 or vice versa
3. The journo constructs the look they want using a combination of 1 image + 2

Actually, 1 & 3 are complementary.

JamieB-gu: Perhaps it's possible to parse this using the image `role`? In the "two above one" example the first two are `role: halfWidth`, and the third has no `role` which means full width.

> The journo constructs the look they want using a combination of 1 image + 2

Would this mean losing the shared caption?

oliverlloyd: > Would this mean losing the shared caption?

It would mean you'd lose some data integrity because you'd have to deliberately not set a caption for the set of 2 and then have the caption for the 1 apply to all of them. Which is basically where we're at right now.

I suppose you could enhance the api to have a `twoAtTop` boolean. It would make cleaning the data harder but is probably the best (most flexible) way to present the new element in Composer

JamieB-gu: Is there a way to make the container layout-agnostic?

```tsx
type Width = Half | Full;

interface Image {
  width: Width;
  sources: ...;
  data: ...;
}

interface Props {
  images: Image[];
  caption: string | null;
}

const halfStyles = css`
  flex-basis: 50%;
`;

const fullStyles = css`
  flex-basis: 100%;
`;

const styles = css`
  display: flex;
  flex-wrap: wrap;
`;

const imgStyles = (width: Width) => {
  switch (width) {
    case Half:
      return halfStyles;
    case Full:
    default:
      return fullStyles;
  }
}

const MultiImage = (props: Props) =>
  <figure css={styles}>
    {images.map(image => <img css={imgStyles(image.width)} {...image.data} />}
    {props.caption ? <figcaption>{props.caption}</figcaption> : null}
  </figure>

```

I have a vague memory of one of these having _five_ images in it...

gtrufitt: Basically, we're going to have to support any combination of images in half width or non-half-width combinations...

https://github.com/guardian/frontend/blob/3095a28e951b63ae993aa3d3fcc59a3cf83afc4b/static/src/stylesheets/inline/article-photo-essay-garnett.scss#L222

https://github.com/guardian/frontend/blob/08b3732d113c28c5165ab2ce8e9b7b52909da336/common/app/views/support/HtmlCleaner.scala#L467

If they wanted 4 half width image, a big image in the middle then 4 more half width they could right now. So we either descope this or we have to adjust the API as @JamieB-gu mentions to handle the combo...

Probably latter as we've got existing content to support and we don't wanna have to split the components.

PR #1473, opened 2020-05-10
Missing immersive captions

oliverlloyd:
## What does this change?
Ensures that we display a caption, even if that text is only the credit

### Before
Screenshot 2020-05-10 at 14 45 15

### After
Screenshot 2020-05-10 at 14 45 29

## Why?
We previously thought that captions should not show at all if there was now caption text, even if there was a credit. But it's now clear that sometimes the text can be wholy made up of just the credit

### DIFF
--- a/src/web/components/Caption.tsx
+++ b/src/web/components/Caption.tsx
@@ -104,7 +104,10 @@ export const Caption = ({
         }
     `;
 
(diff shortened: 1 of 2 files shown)


oliverlloyd:
[Missing Captions](https://trello.com/c/Z8jmq7Sw/1624-missing-captions)

PR #1474, opened 2020-05-10
PhotoEssay Type

oliverlloyd:
## What does this change?
Adds the `photoEssay` design type

## Why a design type and not a display?
Articles become photo essays in Composer using a display hint so it would be natural to carry that over into DCR. However, it works better to represent photo essays as a DesignType, for several reasons:

1. By using a design type, we could potentially have a showcase or a standard photo essay, we're not forced to make every photo essay immersive.

2. The relationship between `immersive` and photo essays is confusing and mainly so because we need to use lots of OR statements when trying to handle the logic. Creating this new design type removes the need for that logic and makes things more declarative.

3. As part of the support for photo essays we are going to have to have new versions of several components. Caption, Standfirst, PullQuotes, etc. all have special styles for photo essays. It's a lot more logical to make this branch based on designType because we already use the same value for all out other design branches.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -28,7 +28,8 @@ type DesignType =
     | 'GuardianView'
     | 'GuardianLabs'
     | 'Quiz'
-    | 'AdvertisementFeature';
+    | 'AdvertisementFeature'
+    | 'PhotoEssay';
 
 // This is an object that allows you Type defaults of the designTypes.
 // The return type looks like: { Feature: any, Live: any, ...}
@@ -334,6 +335,7 @@ type CAPIBrowserType = {
     };
     contributionsServiceUrl: string;
     isImmersive: boolean;
+    isPhotoEssay: boolean;
 };
 
 interface TagType {
@@ -527,6 +529,7 @@ interface ConfigType extends CommercialConfigType {
     discussionApiUrl: string;
     discussionD2Uid: string;
     discussionApiClientHeader: string;
+    isPhotoEssay: boolean;
 }
 
 interface GADataType {
--- a/fixtures/CAPI/comment.ts
+++ b/fixtures/CAPI/comment.ts
@@ -2605,6 +2605,7 @@ export const comment: CAPIType = {
         discussionApiClientHeader: 'testClientHeader',
         ampIframeUrl:
             'https://assets.guim.co.uk/data/vendor/a994b749adae30cd58f0e84c8fa28013/amp-iframe.html',
+        isPhotoEssay: false,
     },
     sectionLabel: 'Conservatives',
 };
--- a/fixtures/CAPI/review/showcaseReview.ts
+++ b/fixtures/CAPI/review/showcaseReview.ts
@@ -2711,6 +2711,7 @@ export const showcaseReviewCAPI: CAPIType = {
             'https://assets.guim.co.uk/data/vendor/a994b749adae30cd58f0e84c8fa28013/amp-iframe.html',
         discussionD2Uid: 'testD2Header',
         discussionApiClientHeader: 'testClientHeader',
+        isPhotoEssay: false,
     },
     sectionLabel: 'TV comedy',
 };
--- a/fixtures/CAPI/review/standardReview.ts
+++ b/fixtures/CAPI/review/standardReview.ts
@@ -2388,6 +2388,7 @@ export const standardReviewCAPI: CAPIType = {
             'https://assets.guim.co.uk/data/vendor/a994b749adae30cd58f0e84c8fa28013/amp-iframe.html',
         discussionD2Uid: 'testD2Header',
         discussionApiClientHeader: 'testClientHeader',
+        isPhotoEssay: false,
     },
     sectionLabel: 'Theatre',
 };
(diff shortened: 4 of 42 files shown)


oliverlloyd:
[Add `PhotoEssay` DesignType](https://trello.com/c/rdLuaTlC/1639-add-photoessay-designtype)

index.d.ts
```diff
@@ -28,7 +28,8 @@ type DesignType =
     | 'GuardianView'
     | 'GuardianLabs'
     | 'Quiz'
-    | 'AdvertisementFeature';
+    | 'AdvertisementFeature'
+    | 'PhotoEssay';
```
gtrufitt, line 32: Hmm... `designTypes` are one-to-one on articles, and It's definitely the case that you can have  `photoEssays` that are many of the other `DesignTypes` - how would we handle styling of these articles?

The assumption here is that there can never be `Feature` photoEssay etc which I don't think plays out?

What is the relationship between a `photoEssay` designType and `isPhotoEssay` for browser and TagType? Why do we need both?

gtrufitt: Discussed elsewhere but for prosperity:

> designTypes are one-to-one on articles, and It's definitely the case that you can have photoEssays that are many of the other DesignTypes - how would we handle styling of these articles?

PhotoEssays are a separate enough entity, particularly in Frontend where designType is overridden, for this not to be a problem - photoEssay designType may in most cases default to the same as the `immersive` designType component styling but can be nicely separated for the cases it does not match up.

This issue is more with the special report and it's relationship with everything else but @JamieB-gu's colour scheme suggestion might help us there instead (Special Report can't be a designType really as it is definitely possible to have specialReports that are other designTypes)

JamieB-gu: > the same as the immersive designType component styling

Isn't `Immersive` a `Display` type?

> Special Report can't be a designType really as it is definitely possible to have specialReports that are other designTypes

Does this mean that we could have `Feature`s, `Liveblog`s, `Analysis` pieces etc. that are also special reports?

This could be problematic 🤔. Given the following type:

```ts
type Format = {
  design: Design;
  display: Display;
  pillar: Pillar;
}
```

How do you derive whether something is a special report?

oliverlloyd: > Isn't Immersive a Display type?

Yes. But frontend spits out designType = 'Immersive'

> How do you derive whether something is a special report?
This is the million dollar question. From a purely technical point of view I think you check the tags and see if it has one that appears in the 'Special' list. But how do you then represent this state using the model?

I guess the options are:
1. As a `display` type
2. As a `design` type
3. As a `pillar`
or 4. As an additional `isSpecialReport` flag

The one that we choose depends on what combinations we want to support. If a special report is a display type then it can't be combined with another design type, etc. Using an additional flag is most flexible but adds a lot of complexity

JamieB-gu: I think (3) is problematic because special reports also have pillars (e.g. [this piece](https://www.theguardian.com/world/2016/sep/27/panama-papers-inquiry-opens-at-european-parliament) is `pillar/news`)

(2) also suffers from the same problem, special reports can be different `Design`s (such as [analysis](https://www.theguardian.com/politics/2016/apr/06/the-cameron-network-inherited-wealth-and-family-companies))

But (1) is _also_ an issue because they can also be different `Display`s (like [immersive](https://www.theguardian.com/news/2016/apr/08/mossack-fonseca-law-firm-hide-money-panama-papers))

Which seems to leave use with (4), which would mean updating the `Format` type 🤔

oliverlloyd: Or we stamp our feet a little bit and draw some lines in the sand? I think there's a good case to start pushing back on some of these things.

Do editorial really _need_ the ability to have a different pillar or an Analysis version of Special Report or have these things just happened over time because they could?

JamieB-gu: I'd guess the answer to whether special reports need to be different `Design`s (such as `Analysis` or `Interactive` or `Feature`) will be yes.

As to whether they need a separate `Pillar`, I don't know. Maybe @gtrufitt does? If not, we _could_ formally define `SpecialReport` as our sixth pillar. But if we do that we'd have to accept that special reports can't fall under any of the other pillars because that would no longer make sense.

oliverlloyd: What about `SpecialAnalysis`, `SpecialFeature`?

It's pushing the envelope of the Design type but I like how it formalises what's possible. You can have a 'special' Feature, Interview and Analysis. But never a special Recipe, Quiz, etc.

Using Pillar just _feels wrong_. We've been careful to keep pillar all about colour but here we'd have to start using it to decide on other things as well, at that point it's really not that different from a Design type.

JamieB-gu: I'd agree that I'm not thrilled about using `Pillar` for these, although possibly for different reasons 🤔. I _don't_ see `Pillar` as necessarily purely about colour; I'm looking at it as some abstract property of a piece of content that we often to use to _derive_ colour in combination with other properties, like `Design`.

> What about `SpecialAnalysis`, `SpecialFeature`?
> You can have a 'special' Feature, Interview and Analysis. But never a special Recipe, Quiz, etc.

This is a fair point, it's unlikely a special report for `Recipe` makes sense... If we make it a `Pillar` or a `Display` then it implies that it works with _any_ `Design`, which is invalid.

PR #1475, opened 2020-05-10
Fix horizontal scrolling

oliverlloyd:
## What does this change?
Only push the guardian lines out 10px below mobileLandscape

## Why?
Using 20px before was pushing the lines outside the viewport causing overflow problems


oliverlloyd:
[Line Margins Causing Scrolling](https://trello.com/c/kZq88vTr/1627-line-margins-causing-scrolling)

PR #1478, opened 2020-05-11
Remove Outbrain code from DCR

shtukas:
## What does this change?

Remove Outbrain code from DCR. Note that `plistaForOutbrainAu` is being renamed in Frontend ( https://github.com/guardian/frontend/pull/22568  ), but the order in which both PRs go is not important.


src/web/layouts/ImmersiveLayout.tsx
```diff
@@ -166,7 +165,7 @@ export const ImmersiveLayout = ({
 }: Props) => {
     const {
         config: { isPaidContent },
-        pageType: { isSensitive },
+        pageType: {},
```
oliverlloyd, line 168:
```suggestion
```

oliverlloyd: That line isn't needed

oliverlloyd:
I noticed on the snapshots that we have double borders in some places but this seems to be a larger problem so I've raised a [separate ticket](https://trello.com/c/m1oR7C7g/1651-review-section-borders) for it

PR #1479, opened 2020-05-11
Dynamic import consent management platform utilities

liywjl:
## What does this change?
Dynamic import for utilities used in '@guardian/consent-management-platform'

### Before
`shouldShow` and `setErrorHandler` where bundled as part of the `react.js` bundle

### After
`shouldShow` and `setErrorHandler` removed from `react.js` and are loaded dynamically when `CAPI.config.cmpUi` is true.

## Why?
`shouldShow` and `setErrorHandler` are some of the larger parts of '@guardian/consent-management-platform'. We can use webpack's dynamic import functionality to break them out of the bundle.

## Note
`shouldShow` and `setErrorHandler` are needed when `CAPI.config.cmpUi` is true. This is known on the server, so it may be worth splitting out and importing server side?

### DIFF
--- a/src/web/components/CMP.tsx
+++ b/src/web/components/CMP.tsx
@@ -1,24 +1,31 @@
 import React, { useState, useEffect } from 'react';
-import {
-    shouldShow,
-    setErrorHandler,
-} from '@guardian/consent-management-platform';
+
 import { ConsentManagementPlatform } from '@guardian/consent-management-platform/lib/ConsentManagementPlatform';
+import { initPerf } from '@root/src/web/browser/initPerf';
 
 export const CMP = () => {
     const [show, setShow] = useState(false);
 
     useEffect(() => {
-        if (shouldShow()) {
-            setShow(true);
+        const { start, end } = initPerf(
+            'consent-management-platform-utilities',
+        );
+        start();
+        import(
+            /* webpackChunkName: "consent-management-platform-utilities" */ '@guardian/consent-management-platform'
+        ).then(({ shouldShow, setErrorHandler }) => {
+            end();
+            if (shouldShow()) {
+                setShow(true);
 
-            // setErrorHandler takes function to be called on errors in the CMP UI
-            setErrorHandler((errMsg: string): void => {
-                const err = new Error(errMsg);
+                // setErrorHandler takes function to be called on errors in the CMP UI
+                setErrorHandler((errMsg: string): void => {
+                    const err = new Error(errMsg);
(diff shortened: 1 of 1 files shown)


liywjl:
Used https://github.com/guardian/dotcom-rendering/pull/1376 as base, waiting for that PR to be merged

src/web/components/CMP.tsx
```diff
@@ -1,33 +1,52 @@
...
     }, []);
 
     return (
         <>
             {show && (
-                <ConsentManagementPlatform
-                    source="dcr"
-                    onClose={() => setShow(false)}
-                />
+                <Suspense fallback={<></>}>
```
jlieb10, line 44: _bites nails_

liywjl: fallback is required, does not support `null`, but we should not render anything while CMP is loading, I could write it as `React.Fragment`.

btw this PR is more about dynamic imports of, this change is from the base PR https://github.com/guardian/dotcom-rendering/pull/1376

ripecosta: Because these changes are based on changes made in another branch  may I suggest editing the branch you're merging into on this PR to be that base branch? Only temporarily until it gets merged. That way you'll only see the relevant changes in this PR, which makes it easier to read.
Examples: https://github.com/guardian/frontend/pull/22269 or https://github.com/guardian/frontend/pull/22268

src/web/components/CMP.tsx
```diff
@@ -21,16 +17,25 @@ export const CMP = () => {
...
+                setErrorHandler((errMsg: string): void => {
+                    const err = new Error(errMsg);
 
-                window.guardian.modules.sentry.reportError(err, 'cmp');
-            });
-        }
+                    window.guardian.modules.sentry.reportError(err, 'cmp');
+                });
+            }
+        });
```
oliverlloyd, line 38: Should this be inside a `useEffect` like this? I'm a little confused by what's gong on here. I though you can set the dynamic import at the top of the file and then it only gets pulled down when it's used. I'm not 100% following this though so I could be off

liywjl: The examples on webpack move the import logic within the function https://webpack.js.org/guides/code-splitting/#dynamic-imports

I think the behavior you are referring to is only accomplished thanks to `React.lazy`.

liywjl:
closed in favor of https://github.com/guardian/dotcom-rendering/pull/1492

PR #1481, opened 2020-05-12
Remove 'clear' from the build and validate steps

shtukas:
## What does this change?

Remove 'clear' from the build and validate steps. Known to break things, in contexts without the `TERM` env variable set.

### DIFF
--- a/makefile
+++ b/makefile
@@ -30,7 +30,7 @@ deploy:
 
 # prod #########################################
 
-build: clear clean-dist install
+build: clean-dist install
 	$(call log, "building production bundles")
 	@NODE_ENV=production webpack --config scripts/webpack/frontend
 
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -101,10 +101,10 @@ test-ci: clear clean-dist install
 bundlesize: clear clean-dist install build
 	@bundlesize
 
-validate: clear clean-dist install tsc lint stylelint test validate-build
+validate: clean-dist install tsc lint stylelint test validate-build
```
oliverlloyd, line 104: Do we also need to take this command off of the child commands here too? `tsc` and `test`?

liywjl: why remove `tsc` and `test` ? are they run independently

shtukas: @liywjl 
I think @oliverlloyd was only asking to remove `clear` from the sub commands not to remove the sub commands themselves :)

shtukas: @oliverlloyd 
Removed from `test`, but `tsc` didn't have one as far as I can see.

liywjl: I see 👍

PR #1482, opened 2020-05-12
Fix comment width pre-expand

liywjl:
## What does this change?
Makes sure that comment layout is full width before expand

### Before
Screenshot 2020-05-12 at 15 48 52

### After
Screenshot 2020-05-12 at 15 49 05

## Why?
Width was incorrect, causing it to jump when the user click on the expand button

## Ticket
https://trello.com/c/luHoc4q1/1655-comment-width-pre-expand

### DIFF
--- a/src/web/components/CommentsLayout.tsx
+++ b/src/web/components/CommentsLayout.tsx
@@ -6,7 +6,6 @@ import { App as Comments } from '@guardian/discussion-rendering';
 import { LeftColumn } from '@frontend/web/components/LeftColumn';
 import { SignedInAs } from '@frontend/web/components/SignedInAs';
 import { Hide } from '@frontend/web/components/Hide';
-import { Flex } from '@frontend/web/components/Flex';
 
 type Props = {
     user?: UserProfile;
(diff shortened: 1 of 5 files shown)


src/web/layouts/CommentLayout.tsx
```diff
@@ -465,7 +465,12 @@ export const CommentLayout = ({
                     {showComments && (
                         <Section sectionId="comments">
                             <Flex>
-                                <div id="comments-root" />
+                                <div
+                                    id="comments-root"
+                                    className={css`
+                                        width: 100%;
+                                    `}
+                                />
```
oliverlloyd, line 473: I'm not sure we want to start applying styles to these root marker divs - I worry that could confuse thing

src/web/components/CommentsLayout.tsx
```diff
@@ -40,6 +39,16 @@ const bottomPadding = css`
...
 `;
 
+const flexRowWrapper = css`
+    width: 100%;
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    /* Fixes IE 10/11 bug that collapses this container by default: */
+    /* stylelint-disable-next-line property-no-vendor-prefix */
+    -ms-flex-positive: 1;
```
oliverlloyd, line 49: This is `flex-grow: 1`. Do we want to be consistent and also use that for non IE browsers too? It might make for a better solution rather than width: 100%

liywjl:
a better fix would be to use `flex-grow` on the `<Lazy />` wrapper used to wrap `<CommentContainer />`, which is already being worked on elsewhere

PR #1485, opened 2020-05-13
Immersive sans main media

oliverlloyd:
## What does this change?
Adds support for Immersive articles without main media

https://www.theguardian.com/artanddesign/2020/may/12/buy-a-classic-sport-photograph-euphoria-at-the-1966-world-cup

These are special articles that:

- Have `isImmersive: true`
- Have `DesignType: Immersive`
- Empty main media
- Don't show the top right ad slot

### Before
Screenshot 2020-05-13 at 10 45 18

### After
Screenshot 2020-05-13 at 10 44 25

### DIFF
--- a/src/web/components/ArticleHeadline.tsx
+++ b/src/web/components/ArticleHeadline.tsx
@@ -17,6 +17,8 @@ type Props = {
     byline?: string;
     tags: TagType[];
     isShowcase?: boolean; // Used for Interviews to change headline position
+    noMainMedia?: boolean; // Used for Immersives where the headline styles
+    // change when there is no main media
 };
 
 const curly = (x: any) => x;
@@ -160,14 +162,11 @@ export const ArticleHeadline = ({
     pillar,
     tags,
     byline,
+    noMainMedia,
 }: Props) => {
-    const isPrintShop = tags.find(
-        tag => tag.id === 'artanddesign/series/guardian-print-shop',
-    );
-
-    if (isPrintShop) {
+    if (display === 'immersive' && noMainMedia) {
         return (
-            // Immersive headlines are larger than normal
+            // Immersive headlines have two versions, with main media, and (this one) without
             <h1
                 className={cx(
                     jumboFont,
@@ -183,8 +182,8 @@ export const ArticleHeadline = ({
 
     if (display === 'immersive') {
         return (
-            // Immersive headlines are large and inverted and have their black background
-            // extended to the right
+            // Immersive headlines with main media present, are large and inverted with
+            // a black background
             <h1 className={cx(invertedWrapper, blackBackground)}>
                 <span
                     className={cx(
--- a/src/web/components/ArticleHeadline.tsx
+++ b/src/web/components/ArticleHeadline.tsx
@@ -161,6 +161,26 @@ export const ArticleHeadline = ({
     tags,
     byline,
 }: Props) => {
+    const isPrintShop = tags.find(
+        tag => tag.id === 'artanddesign/series/guardian-print-shop',
+    );
+
+    if (isPrintShop) {
+        return (
+            // Immersive headlines are larger than normal
+            <h1
+                className={cx(
+                    jumboFont,
+                    maxWidth,
+                    immersiveStyles,
+                    displayBlock,
+                )}
+            >
+                {curly(headlineString)}
+            </h1>
+        );
+    }
+
     if (display === 'immersive') {
         return (
             // Immersive headlines are large and inverted and have their black background
--- a/src/web/components/ArticleHeadline.tsx
+++ b/src/web/components/ArticleHeadline.tsx
@@ -182,18 +182,6 @@ export const ArticleHeadline = ({
     }
 
     switch (designType) {
-        case 'Immersive':
-        case 'Article':
-        case 'Media':
-        case 'PhotoEssay':
-        case 'Live':
-        case 'SpecialReport':
-        case 'MatchReport':
-        case 'GuardianLabs':
-        case 'Quiz':
-        case 'AdvertisementFeature':
-            return <h1 className={standardFont}>{curly(headlineString)}</h1>;
-
         case 'Review':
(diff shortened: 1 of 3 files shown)


src/web/components/ArticleHeadline.tsx
```diff
@@ -161,6 +161,26 @@ export const ArticleHeadline = ({
     tags,
     byline,
 }: Props) => {
+    const isPrintShop = tags.find(
+        tag => tag.id === 'artanddesign/series/guardian-print-shop',
```
liywjl, line 165: is `artanddesign/series` important?

src/web/layouts/ImmersiveLayout.tsx
```diff
@@ -280,6 +299,48 @@ export const ImmersiveLayout = ({
                     <GridItem area="border">
                         <Border />
                     </GridItem>
+                    <GridItem area="title">
```
liywjl, line 302: do you still want `GridItem` if the child is being conditionally rendered?

oliverlloyd:
> I don't think we should be modifying the immersive headline styles for a specific series like this, and the design should be descoped back to standard white on black immersive headline

That would greatly simplify this PR. It feels really wrong to have a version of the headline based on a series tag. That seems like a line we don't want to cross.

oliverlloyd:
> should the advert avoid it [the immersive image]?

That seems like a problem for spacefinder. Maybe that is what's failing here?

I completely agree that not rendering the slot for this series tag is pretty awful

gtrufitt:
> That would greatly simplify this PR. It feels really wrong to have a version of the headline based on a series tag. That seems like a line we don't want to cross.

I think now and in the future there would need to be significant editorial or commercial advantages to applying those differences as the maintenance and complexity is not free.

src/web/components/ArticleHeadline.tsx
oliverlloyd: The string `artanddesign/series/guardian-print-shop` is the literal id for the tag so I think it's correct to use it all.

gtrufitt:
> That seems like a problem for spacefinder. Maybe that is what's failing here?

> I completely agree that not rendering the slot for this series tag is pretty awful


I was just reading this PR https://github.com/guardian/frontend/pull/20933/files#diff-9d70892042a7201f76b5bdad674cfd67R34 and it seems to, based on classes of immersive elements, calculate the offset required and adjust the allowed slots. 

We could add a class to the immersive elements and main column so that the commercial scripts can hook into that maybe...

oliverlloyd:
> > That seems like a problem for spacefinder. Maybe that is what's failing here?
> 
> > I completely agree that not rendering the slot for this series tag is pretty awful
> 
> I was just reading this PR https://github.com/guardian/frontend/pull/20933/files#diff-9d70892042a7201f76b5bdad674cfd67R34 and it seems to, based on classes of immersive elements, calculate the offset required and adjust the allowed slots.
> 
> We could add a class to the immersive elements and main column so that the commercial scripts can hook into that maybe...

Oh man, these problems are really coming out of the woodwork today! So this is another example of DCR being a bit distant from commercial. We can add the classes, but it feels like going backwards... But the only other alternative is to improve the skill palette of DCR so it can do more advanced as tasks, like dynamically setting slots based on height

paperboyo:
In the spirit of Gareth’s comment, I feel (not actually know!) this shouldn’t be specific. [There](https://www.theguardian.com/lifeandstyle/2015/may/01/last-bites-david-everitt-mathias-final-meal) are more [articles](https://www.theguardian.com/science/2020/may/03/starwatch-grains-from-halleys-comet-make-a-meteor-show) without [main media](https://www.theguardian.com/world/2019/apr/16/notre-dame-cathedral-fire-a-visual-guide-and-timeline), because main media is not mandatory. They are rare, but still.

paperboyo:
>     * We already have the option to turn adverts off on articles, I think it's strange that we're creating an override for a specific series

We have the tech to block ads at a tag level:

image

Here, it would be for the whole series. CP can help.

gtrufitt:
> In the spirit of Gareth’s comment, I feel (not actually know!) this shouldn’t be specific. [There](https://www.theguardian.com/lifeandstyle/2015/may/01/last-bites-david-everitt-mathias-final-meal) are more [articles](https://www.theguardian.com/science/2020/may/03/starwatch-grains-from-halleys-comet-make-a-meteor-show) without [main media](https://www.theguardian.com/world/2019/apr/16/notre-dame-cathedral-fire-a-visual-guide-and-timeline), because main media is not mandatory. They are rare, but still.

Yeah, we’ve realised we need to support this for all immersives without main media rather than the specific series, I think a Oliver has updated the PR to that effect 👍🏻👍🏻

paperboyo:
> for all immersives

Some examples aren’t immersives…

Anyhow, thanks and sorry, this happens when one comments late, doesn’t know the context, cannot read the code, draws quick conclusions without much thought. Basically: normal me.

oliverlloyd:
> We could add a class to the immersive elements and main column so that the commercial scripts can hook into that maybe...

I don't think we can maintain parity using classnames. Similar to the comment ad slot, it's not realistic to try and make all this commercial logic which is so frontend specific also work on DCR. Hopefully, the work AS is starting can be extended to fix these kinds of issues?

Shall we put this out as is (hiding the ad slot in this case), and then raise a card to fix it properly (hopefully based on the module approach)?

src/web/layouts/ImmersiveLayout.tsx
```diff
@@ -362,13 +424,21 @@ export const ImmersiveLayout = ({
...
                         <RightColumn>
-                            <div
-                                className={css`
-                                    margin-top: ${space[4]}px;
-                                `}
-                            >
-                                <AdSlot asps={namedAdSlotParameters('right')} />
-                            </div>
+                            <>
+                                {mainMedia && (
```
gtrufitt, line 428: So are we just removing the right ad in the case of  no main media?

PR #1486, opened 2020-05-13
Auto rich links

liywjl:
## What does this change?
Extract `rich-link` from references object and pass down, `referencesRichLink` as a new object to CAPI.

## Why?
Some rich links can be found in `CAPI.config.references`, if the object contains the key `rich-link` it means the value should be a URL path to the json of that rich link. This PR extracts this data out of the the CAPI sent to the server. 

## Note
Unlike rich links found in the elements array, which use their position index to calculate, there is not index position for auto rich links. Normally we would use spaceFinder to determine their position, however we do not currently have access to that function

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -312,6 +312,7 @@ type CAPIBrowserType = {
         discussionApiClientHeader: string;
     };
     richLinks: RichLinkBlockElementBrowser[];
+    referencesRichLink: ReferenceRichLinkBlockElementBrowser[];
     editionId: Edition;
     contentType: string;
     sectionName?: string;
--- a/index.d.ts
+++ b/index.d.ts
@@ -502,7 +502,7 @@ interface ConfigType extends CommercialConfigType {
     ajaxUrl: string;
     sentryPublicApiKey: string;
     sentryHost: string;
-    references: { [key: string]: string }[];
+    references?: { [key: string]: string }[];
     dcrSentryDsn: string;
     switches: { [key: string]: boolean };
     abTests: { [key: string]: string };
--- a/index.d.ts
+++ b/index.d.ts
@@ -311,7 +311,7 @@ type CAPIBrowserType = {
         discussionD2Uid: string;
         discussionApiClientHeader: string;
     };
-    richLinks: RichLinkBlockElement[];
+    richLinks: RichLinkBlockElementBrowser[];
     editionId: Edition;
     contentType: string;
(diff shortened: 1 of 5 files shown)


src/model/window-guardian.ts
```diff
@@ -67,23 +67,46 @@ const makeWindowGuardianConfig = (
...
+                } as RichLinkBlockElementBrowser);
             }
             return acc;
         },
-        [] as RichLinkBlockElement[],
+        [] as RichLinkBlockElementBrowser[],
     );
 
+    // references with the key `rich-link` work similar to RichLinkBlockElement
+    // however we do care about thier index, as we will use spaceFinder to determin thier location
```
gtrufitt, line 89:
```suggestion
    // however we do care about their index, as we will use spaceFinder to determine their location
```

src/model/window-guardian.ts
```diff
@@ -67,23 +67,46 @@ const makeWindowGuardianConfig = (
...
+    const referencesRichLink: ReferenceRichLinkBlockElementBrowser[] = CAPI
```
gtrufitt, line 90: I think as this is array, could be plural?

index.d.ts
```diff
@@ -502,6 +503,7 @@ interface ConfigType extends CommercialConfigType {
     ajaxUrl: string;
     sentryPublicApiKey: string;
     sentryHost: string;
+    references?: { [key: string]: string }[];
```
gtrufitt, line 506: capiReferences might be clearer

src/model/window-guardian.ts
gtrufitt: is this supposed to be `don't care about their index` rather than `do`? Otherwise the code doesn't match the comment :)

liywjl:
In order to complete this PR, we need to have a way to use spaceFinder in DCR

gtrufitt:
> In order to complete this PR, we need to have a way to use spaceFinder in DCR

I think we can work around this.

liywjl:
@gtrufitt awesome, I can have a look at this monday 😄

gtrufitt:
> @gtrufitt awesome, I can have a look at this monday 😄

👋

oliverlloyd:
I'm closing this PR because it's been inactive for a while. This is part of a general clean up of our PRs and nothing specific about this piece of work.

If you're still working on it please reopen

PR #1492, opened 2020-05-14
Dynamic imports

liywjl:
## What does this change?

Implements Dynamic imports for all child elements of the `Lazy` and `CMP`

### Before
Screenshot 2020-05-14 at 11 52 17

### After
Screenshot 2020-05-14 at 17 46 00

## Why?
Reduce the bundle size, and only fetch components if they are being rendered.

### Over use of dynamic imports
We need to be careful to not over use dynamic imports as this could increase the number of round trips, that could increase the overall page load time.

However we can use [`prefetching` and `preloading`](https://webpack.js.org/guides/code-splitting/#prefetchingpreloading-modules) to help speed up the fetching of some components. We can also include components that are commonly used server side in the list of scripts loaded and leverage parallel loading of smaller scripts than one lager one that could still lead to notable performance increases. 

### Notes
- `CMP`
CMP is a heavy component that is not Lazy loaded, therefore we still do apply dynamic imports to it

- `MostViewedFooter`
`Lazy` component is found within the component due to some complexities with rendering the Ad, so we moved the `Lazy` logic to exist within the component

- `RichLinkComponent`
`RichLinkComponent` is loaded 2x, once with `Lazy` wrapper, the other without, therefore we should not apply dynamic imports to it


oliverlloyd:
Note. The exception to the only in App rule is `MostViewedFooter`. We can't (yet) lazy load this section because we _must_ server render the ad slot. Instead, the data component is lazy loaded inside and this can be dynamically imported (`MostViewedFooterData`)

liywjl:
> A cleaner approach for dynamic loading might be to change the imports in `App`? That way we
> 
> 1. Keep this logic all in one place and
> 2. Ensure it is only ever run on the client (not on the server)
> 
> I really like how `App` is this one place that describes how we manage client side loading, hydration, Portals, etc. Let's try to maintain that!

The drawback is that many of these components still have a lot of hidden conditional logic nested within them. Doing it in `App.tsx` without lifting this logic would mean we would fetch the entire component even if it is not necessarily needed.

oliverlloyd:
> The drawback is that many of these components still have a lot of hidden conditional logic nested within them. Doing it in `App.tsx` without lifting this logic would mean we would fetch the entire component even if it is not necessarily needed.

That's a good point. Sometimes an article might not have onwards content.

But that sounds like a secondary enhancement. The first thing we want to do is _lazy load_ the components below the fold. This should be step #1.

After that, it's true that _some_ sections have secondary logic that might mean the component won't need to be loaded. Things like MostViewedFooter, MVRight though will always be loaded and others such as OnwardsUpper will mostly be loaded (95% of the time?)

Shall we implement lazy loading in it's classic sense (the code is only loaded when the user scrolls down) and then look to maybe further enhance individual components as required? I think it's probably just `OnwardsLower` that needs this*.

*I'm excluding CMP and Automat stuff from this thinking

oliverlloyd:
Also, it looks like React is not hydrating the stories. Checking the console I can see some errors:

```
Warning: render(...): Replacing React-rendered children with a new root component. If you intended to update the children of this node, you should instead have the existing children update their state and render the new components instead of calling ReactDOM.render.
    in Hydrate (created by App)

Warning: Expected server HTML to contain a matching <div> in <div>.
    in div (created by Dropdown)

Error: A React component suspended while rendering, but no fallback UI was specified.

Add a <Suspense fallback=...> component higher in the tree to provide a loading indicator or placeholder to display.
    in Unknown (created by MostViewedRightWrapper)
```

gtrufitt:
> However we can use prefetching and preloading to help speed up the fetching of some components. We can also include components that are commonly used server side in the list of scripts loaded and leverage parallel loading of smaller scripts than one lager one that could still lead to notable performance increases.

👍 

The next key perf enhancement (and another cornerstone of the original DCR perf vision) is loading only scripts we need. We can break out serverside the components, and load based on the elements array, similar to what we do with AMP.

src/web/components/App.tsx
```diff
@@ -174,12 +215,14 @@ export const App = ({ CAPI, NAV }: Props) => {
...
-                    inHeader={true}
-                />
+                <Suspense fallback={<></>}>
+                    <ReaderRevenueLinks
+                        urls={CAPI.nav.readerRevenueLinks.footer}
+                        edition={CAPI.editionId}
+                        dataLinkNamePrefix="footer : "
+                        inHeader={true}
+                    />
+                </Suspense>
```
oliverlloyd, line 225: This component is in the header so there's no benefit to lazy loading it?

liywjl: if the user scrolls down the page fast, or the page is small and both are trying to be rendered, the fact this component is not wrapped in a `Suspense`, will cause the JS to crash. 

The component will on be fetched once anyway, `Suspense` is just a wrapper to handle the state while the component is loading

src/web/components/Onwards/OnwardsUpper.tsx
```diff
@@ -1,7 +1,18 @@
...
+const OnwardsData = React.lazy(() => {
+    const { start, end } = initPerf('OnwardsData');
+    start();
+    return import(/* webpackChunkName: "OnwardsData" */ './OnwardsData').then(
+        module => {
+            end();
+            return { default: module.OnwardsData };
+        },
+    );
+});
```
oliverlloyd: Did you mean to keep this in here?

I think it's right that we lazy load the OnwardsUpper component only when the reader scrolls the content into view. It's a lot less clear if this secondary split adds benefit.

We know without doubt we only want to load based on scroll position but beyond that it could be argued that most articles have onwards content so splitting here might possibly slow things down for the majority of readers? Tbh, I'm not sure. It'd be great to keep this change separate where we can discuss and test it in isolation?

src/web/components/App.tsx
oliverlloyd: I think though that, if the component isn't being lazy loaded, then we shouldn't dynamically load it, no?

src/web/components/Onwards/OnwardsUpper.tsx
liywjl: ^^ I was removing it

src/web/components/App.tsx
liywjl: oh I see what you mean, yeah good point, ill remove the dynamic import

src/web/components/App.tsx
```diff
@@ -174,12 +206,14 @@ export const App = ({ CAPI, NAV }: Props) => {
...
```
oliverlloyd, line 216: We can lose this suspense?

liywjl: good catch, removed

PR #1494, opened 2020-05-14
removes condition limiting amp epic to single test article

michaelbjacobson:
AMP epic tested live on PROD, on various devices, operating systems, and browsers. Currency symbol variable confirmed to be working.

This rolls out AMP epic implementation on all AMP articles (unless reader revenue asks are suppressed).


michaelbjacobson:
This is ready to go, but we're waiting on feedback from Aus team ahead of merging into master.

PR #1495, opened 2020-05-15
Sticky comment ad

oliverlloyd:
## What does this change?
Sticks the comment double MPU to the top of the viewport as thereader scrolls the list of comments

## Implementation Notes
To achieve this we now server side render comments. This only results in a loading indicator being rendered but it means we are able to render the ad slot on the server which is a key requirement and also later hydrate the whole comments container, including the ad slot. This lets us pass properties between the ad slot container and `Comments`.

## Why?
💵


oliverlloyd:
[Comment AdSlot is sticky](https://trello.com/c/x4aOLTLx/1595-comment-adslot-is-sticky)

src/web/components/CommentsLayout.tsx
```diff
@@ -56,49 +61,123 @@ export const CommentsLayout = ({
...
+    }, [heightChangeCount]);
+
+    useEffect(() => {
+        // Whoah, why on earth are you doing this? Are you mad?
+        // Yup, this is a hackity, hack, hack. It's bad because we have no way to be sure that the dom
+        // has finished rendering so maybe the check will fail (return the wrong height), and it's also
+        // bad because we could be running the check too late (delaying the dom update). So why?
+        // Because we are already working on a better way to render this ad slot so we're happy
+        // to leave this code sitting here for now as a workaround
+        // TODO: **** Remove this ****
```
shtukas, line 97: Understood! :)
Also my philosophy is that there are no hack, just badly documented solutions. So thanks for the comments :)

src/web/layouts/CommentLayout.tsx
```diff
@@ -438,7 +438,7 @@ export const CommentLayout = ({
                     </GridItem>
                     <GridItem area="right-column">
                         <RightColumn>
-                            <StickyAd />
+                            <StickyAd name="right" height={1059} />
```
shtukas, line 441: Just curious, where does the value 1059 comes from ? :)

src/web/components/CommentsLayout.tsx
```diff
@@ -56,49 +61,123 @@ export const CommentsLayout = ({
...
+
+    const handleHeightChange = () => {
+        setheightChangeCount(heightChangeCount + 1);
+    };
+
+    const checkHeight = (ref: RefObject<HTMLDivElement>) => {
+        if (ref.current) {
+            const availableHeight = ref.current.clientHeight;
+            // Don't run the sticky container all the way to the bottom
+            let heightWereUsing = availableHeight - 600;
```
gtrufitt, line 77: Assign 600 to a var above so that it's clear what `600` is?

src/web/components/App.tsx
```diff
@@ -330,55 +330,27 @@ export const App = ({ CAPI, NAV }: Props) => {
...
+                    commentPage={commentPage}
+                    commentPageSize={commentPageSize}
+                    commentOrderBy={commentOrderBy}
+                    isClosedForComments={isClosedForComments}
+                    discussionD2Uid={CAPI.config.discussionD2Uid}
+                    discussionApiClientHeader={
+                        CAPI.config.discussionApiClientHeader
+                    }
+                    enableDiscussionSwitch={CAPI.config.enableDiscussionSwitch}
+                    expanded={!!openComments}
```
gtrufitt, line 349: `openComments` isn't very clear what it is from var name so the `!!` makes it even harder to reason with. Is it 'comments being open on the page' or 'comment box is open' or something else.

src/web/components/CommentsLayout.tsx
gtrufitt, line 77: Nitpick: `heightWereUsing` vs `heightWeAreUsing`, because you can't have apostrophes in var names.

src/web/components/CommentsLayout.tsx
```diff
@@ -56,49 +61,123 @@ export const CommentsLayout = ({
...
+        // Because we are already working on a better way to render this ad slot so we're happy
+        // to leave this code sitting here for now as a workaround
+        // TODO: **** Remove this ****
+        if (expanded) {
+            setTimeout(() => {
+                checkHeight(heightRef);
+            }, 500);
+            setTimeout(() => {
+                checkHeight(heightRef);
+            }, 3000);
```
gtrufitt, line 104: As it's a hack, maybe not worth it but `setInterval` might respond quicker?
```suggestion
			let intervalCount = 0;
            let checkHeightInterval = setInterval(() => {
	            intervalCount++;
                checkHeight(heightRef);
    			if (intervalCount === 8) {
    				clearInterval(checkHeightInterval)
    			}
            }, 500);
```

src/web/layouts/CommentLayout.tsx
gtrufitt: +1 named var here would be useful context

src/web/components/CommentsLayout.tsx
oliverlloyd: Thanks. This bugged me too but I never got around to fixing it

src/web/components/App.tsx
oliverlloyd: Agreed 👍

oliverlloyd:
> LGTM, I would still like a change to the component name to avoid having duplicate `CommentLayout`

I added [a ticket](https://trello.com/c/cHj1R5Mx/1671-commentlayout-opinionlayout) for this

PR #1496, opened 2020-05-16
Refactor stickiness

oliverlloyd:
## What does this change?
Makes the top above nav slot on Comment pieces sticky

Also refactors how we make the header sticky by abstracting the logic into a lib.

## Why?
We repeat this code in three places now and it's deemed unlikely that we would want to stick the advert in a different manner in different layouts so an abstraction makes sense in this case.


oliverlloyd:
[Stick top-above-nav for Opinion](https://trello.com/c/pb1uHSJs/1669-stick-top-above-nav-for-opinion)

PR #1499, opened 2020-05-17
Refactor RichlinkComponent / Fix quote size

oliverlloyd:
## What does this change?
Refactors `RichLinkComponent` to:

- Break out the presentation code into a separate component (for testing and reabability). `RichLinkComponent` now handles data fetching and types, `RichLink` manages how rish links look
- Adds stories for the various states a rich link can appear in

Once refactored, we replaced the use of a direct quote svg with the `QuoteIcon` component. The breakpoints that the quote icons change size at for rich links are different to other places where quote icons are used so we're show / hiding two versions of the `QuoteIcon` component to cover this. In doing this I noticed that the `Hide` component added `display: block` by default (my text wrapped) so I refacotred the `Hide` component to use a `span`


oliverlloyd:
[Quote icon the wrong size in comment rich link](https://trello.com/c/Sm6eJUVZ/1302-quote-icon-the-wrong-size-in-comment-rich-link)

PR #1500, opened 2020-05-18
Added specific widths to some layout stories

oliverlloyd:
## What does this change?
Causes 2 stories to be rendred at each specific breakpoint

## Why?
So that we capture more coverage at these layouts


oliverlloyd:
[Add additional widths to some layout stories](https://trello.com/c/oOEq06tD/1594-add-additional-widths-to-some-layout-stories)

PR #1503, opened 2020-05-18
Increase test coverage for different image roles

oliverlloyd:
## What does this change?
Inserts additional image elements into the fixtures used for Storybook and also adds a new article url to the list used for Cypress

## Why?
To increase test coverage of the different image types. We've previously seen how changes to the image logic can cause issues in DCR when an article has a speific image type so we wanted to be sure to have all types covered


oliverlloyd:
[Add additional images weights to layout stories/cypress urls](https://trello.com/c/uZgQ7EUm/1596-add-additional-images-weights-to-layout-stories-cypress-urls)

PR #1507, opened 2020-05-19
Top above ad not showing on mobile

liywjl:
## What does this change?
We hydrate the rendering of the header as slot. We also conditionally render the Ad based on the page width.

## Why?
The header ad slot is rendered in body on widths lower than tablet. Previously we used CSS display to hide this element, however the commercial script would still identify the DOM element and assume this element is already being rendered.

Therefore we need to avoid rendering this DOM element at lower widths. The commercial script will then auto generate the new header ad

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -594,7 +594,8 @@ type IslandType =
     | 'onwards-lower-whensignedout'
     | 'rich-link'
     | 'links-root'
-    | 'comments';
+    | 'comments'
+    | 'header-ad-slot-root';
 
 interface TrailType {
     designType: DesignType;
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -10,6 +10,7 @@ import { SlotBodyEnd } from '@frontend/web/components/SlotBodyEnd';
 import { Links } from '@frontend/web/components/Links';
 import { SubNav } from '@frontend/web/components/SubNav/SubNav';
 import { CommentsLayout } from '@frontend/web/components/CommentsLayout';
+import { HeaderAdSlot } from '@root/src/web/components/HeaderAdSlot';
 import { incrementWeeklyArticleCount } from '@guardian/automat-client';
 
 import { Portal } from '@frontend/web/components/Portal';
@@ -215,6 +216,12 @@ export const App = ({ CAPI, NAV }: Props) => {
             </Portal>
             <Hydrate root="links-root">
                 <Links userId={user ? user.userId : undefined} />
+            </Hydrate>
+            <Hydrate root="header-ad-slot-root">
+                <HeaderAdSlot
+                    isAdFreeUser={CAPI.isAdFreeUser}
+                    shouldHideAds={CAPI.shouldHideAds}
+                />
             </Hydrate>
             <Hydrate root="edition-root">
                 <EditionDropdown
--- a/src/web/components/HeaderAdSlot.tsx
+++ b/src/web/components/HeaderAdSlot.tsx
@@ -35,19 +35,31 @@ export const HeaderAdSlot: React.FC<{
     isAdFreeUser: boolean;
     shouldHideAds: boolean;
 }> = ({ isAdFreeUser, shouldHideAds }) => (
-    <div className={headerWrapper}>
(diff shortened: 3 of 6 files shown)


liywjl:
Tested on CODE

gtrufitt:
I need to look at this properly but there are a couple of things off the top of my head 1) is this going to push back ad rendering if we have to wait until hydration of that component before starting to ad the top ad? That might push down ad ratio quite a bit for the above nav ad. 2) The 'react' bundle doing hydration is after the commercial bundle load-wise - does the commercial bundle not expect a slot, and therefore you might get a race condition with this?

liywjl:
> I need to look at this properly but there are a couple of things off the top of my head 1) is this going to push back ad rendering if we have to wait until hydration of that component before starting to ad the top ad? That might push down ad ratio quite a bit for the above nav ad. 2) The 'react' bundle doing hydration is after the commercial bundle load-wise - does the commercial bundle not expect a slot, and therefore you might get a race condition with this?

1) Good point, we can work on then solving this in frontend rather than here. 

2) There could be a race condition, but I havent noticed anything on my end. 

So commercial doesn't expect a slot for smaller screen sizes. Someone added a work around to delete that node if it exists [here](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/commercial/modules/dfp/create-slots.js#L157), that code seems to only be running on smaller widths. 

An ideal solution would be to keep this element on the page, not hide it at widths below tablet, remove the code in frontend commercial/spacefinder that renders this DOM element in DCR (using a similar flag to the workaround mentioned above) and then let commercial just render the element.

oliverlloyd:
The code in frontend is deleting the `top-above-nav` element from the top of the page (where DCR inserts it) so that the commercial code can then insert another element _with the same id_ in the article body.  It's quirky, but that's what happens on mobile. The top above nav ad slot becomes inline1 effectively.

If this code is required, it suggests the frontend fix is no longer working or there are other use cases that need understanding. I'd be keen to focus on that before making commercial changes in DCR.

Side note. Based on [that frontend code](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/commercial/modules/dfp/create-slots.js#L157), we should never need to hide this ad slot at all, using React or css - the commercial code should delete it

Side, side note. This is another good use case for Commercial Modules

liywjl:
@oliverlloyd in order to reduce DOM modifications (which are pretty costly to the browser) we could simply move the conditional wrapper for DCR [here](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/commercial/modules/article-body-adverts.js#L178)

This bit of code is specifically related to inline ads in mobile and `top-above-nav` is determined by the index of the elements. I have been able to get frontend and DCR to work locally so I can test this. This could be a good enough compromise?

gtrufitt:
The concern here is that there's a very undefined contract between DCR and the commercial code, and if we modify DCR then we're breaking that contract and thus open ourselves up to bugs further down the road - while this fixes this particular thing, it might have knock on effects down the code path. 

The separate implementations in the platform also adds to the cognitive load of making changes - rather than trusting making those in the commercial code and the platforms being close enough to work, we have to be aware that the platforms have diverged.

This is an issue with the commercial code - it's too far reaching to assume it's behaviour safely and so we need to mirror Frontend as close as possible.

oliverlloyd:
> This is an issue with the commercial code - it's too far reaching to assume it's behaviour safely and so we need to mirror Frontend as close as possible.

Is this even an issue we should be fixing in DCR? Could this be a problem that Commercial fixes?

liywjl:
I tried running frontend locally and testing out some ideas, but unfortunately running frontend and DCR locally is clunky at best, so it was hard to get any takeaways.

I believe this issue come from a race condition that comes from [here](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/commercial/modules/article-body-adverts.js#L46), as `forEach` may not have been finished executing until the `.then` get called. But as mentioned before, unable to conclude anything on that hypothesis.

liywjl:
closing this in favor of a fix on frontend: https://github.com/guardian/frontend/pull/22609

PR #1508, opened 2020-05-19
Fix Sentry by adding dcrSentryDsn and editionLongForm to the browser window

gtrufitt:
## What does this change?

As title, but adds typing to the value being used so that we can more clearly see what's missing.

Screen Shot 2020-05-19 at 17 11 33
Screen Shot 2020-05-19 at 17 29 29

## Why?
Sentry errors are useful.

## Note

This is just *fixing* sentry, not setting up the lazy loading (wanna make sure this works first)

### DIFF
--- a/src/web/browser/sentry/sentry.tsx
+++ b/src/web/browser/sentry/sentry.tsx
@@ -26,12 +26,12 @@ const ignoreErrors = [
 ];
 
 export const initialiseSentry = (adBlockInUse: boolean) => {
-    const CAPIbrowser: CAPIBrowserType = window.guardian.app.data.CAPI;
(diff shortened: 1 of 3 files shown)


src/web/browser/sentry/sentry.tsx
```diff
@@ -26,11 +26,12 @@ const ignoreErrors = [
 ];
 
 export const initialiseSentry = (adBlockInUse: boolean) => {
+    const CAPIBrowser: CAPIBrowserType = window.guardian.app.data.CAPI;
```
oliverlloyd, line 29: 👍

PR #1509, opened 2020-05-20
Use AtomEmbedHtmlDocumentBlockElement in [web]

shtukas:
## What does this change?

This is an experimental enabling of `AtomEmbedUrlBlockElement` in [web]. Currently the AudioAtom is sent to DCR as `AtomEmbedUrlBlockElement`.


shtukas:
@liywjl Good question. It's just a visual reminder for myself that they are in the pipeline, but you are right, not necessary at all for this particular PR itself :)

shtukas:
Yeah, the ones that I moved are meant to be given a component in the next few days.

PR #1510, opened 2020-05-20
Do not notify Sentry when there’s no epic to show

tjmw:
## What does this change?

It is expected that the data property of the contributions service response can be `null`. Because our accessing of downstream properties (html, css etc.) wasn’t wrapped in a condition we were causing an exception which was then being passed to Sentry unnecessarily.

### Before

Sentry exceptions when the contributions service returns `null` for the data property (no epic to show, based on business rules).

### After

These Sentry exceptions go away! 👋 

## Why?

Unnecessary exceptions clutter up Sentry and potentially push us towards the limit of our plan.

### DIFF
--- a/src/web/components/SlotBodyEnd.tsx
+++ b/src/web/components/SlotBodyEnd.tsx
@@ -172,16 +172,18 @@ const MemoisedInner = ({
             .then(checkForErrors)
             .then(response => response.json())
             .then(json => {
-                setData({
-                    slot: {
-                        html: json.data.html,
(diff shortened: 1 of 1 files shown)


tjmw:
@gtrufitt thanks for the review! I'm going to wait for Nic or Andre to give a sanity check :+1: too just in case I'm missing something here.

> Do you want to track when there's no epic in some way?

Good question - we do track that but it's on the contributions service side.

PR #1511, opened 2020-05-20
Move CMP stub to JS bundle

sndrs:
## What does this change?

moves the CMP stub code from page head to a new `commercial.js` entry point

## Why?

- creates a place to add commercial JS that will be extracted from frontend
- prepares the ground for testing putting sourcepoint branch of new CMP behind 0% test


src/web/server/document.tsx
```diff
@@ -61,6 +61,7 @@ export const document = ({ data }: Props) => {
      */
     const priorityScripts = [
         polyfillIO,
+        getDist({ path: 'commercial.js', legacy: false }),
```
sndrs, line 64: this needs to run before the commercial frontend>dcr bundle. 

but that bundle doesn't have module/no-module builds, so it can't be moved to the `priorityLegacyScripts`/`priorityNonLegacyScripts` arrays, otherwise modern browsers try to load it a module and break.

hence loading it like this. which feels icky – any suggestions?

gtrufitt: Thinking building the scripts needs a slight refactor along the lines of:

```js
const priorityScriptTags = [{
  'path': 'one',
  'legacy': true
},
{
  'path': 'two'
},
{
  'path': 'three',
  'legacy': false
}];



const mapToScriptTags = (pathObj) => {
  if (pathObj.legacy === undefined) {
    return `Normal script tag: ${pathObj.path}`
  } else if (pathObj.legacy === true) {
    return `Script tag with momodule: ${pathObj.path}`
  } else {
    return `Script tag with type="module": ${pathObj.path}`
  }
}

// Insert these into the page instead
console.log(priorityScriptTags.map(mapToScriptTags).join('\n'))
```

So the ordering is maintained through one array but we maintain the module / nomodule and normal script tag mechanism:

```js
`Script tag with momodule: one
Normal script tag: two
Script tag with type="module": three`
```

Does that make sense?

gtrufitt: Maybe even more simplified:

```js
const priorityScriptTags = [{
  'path': 'one',
  'hasLegacy': true
},
{
  'path': 'two',
  'hasLegacy': false
},
{
  'path': 'three',
  'hasLegacy': true
}];



const mapToScriptTags = (pathObj) => {
  if (pathObj.hasLegacy === false) {
    return `Normal script tag: ${pathObj.path}`
  } else if (pathObj.hasLegacy === true) {
    return `
Script tag with momodule: ${pathObj.path}
Script tag with type="module": ${pathObj.path}
`
  }
}

console.log(priorityScriptTags.map(mapToScriptTags).join('\n'))
```

liywjl: what is wrong with calling it as a `module/no-module`? if we compile the code here and it gets build in our webpack config, then it should be fine?

liywjl: we use `module/no-module` purely as a way of checking if the browser supports those features, its an indication that the browser is modern enough to support a build that used `@babel/preset-modules` rather than `@babel/preset-env`

gtrufitt: Hey @liywjl I think @sndrs is referring to the Commercial JS bundle that we get passed a url to from Frontend (via CAPI json). So it's not a bundled file.

src/web/browser/commercial/init.ts
```diff
@@ -0,0 +1,9 @@
+import '@root/src/web/browser/prepareCmp';
```
liywjl, line 1: you may want to add `import '../webpackPublicPath';` to specify the path in case there is the need to do any dynamic imports in the future.

sndrs:
closing this for now - will come back to it again soon and open a new one taking comments into account, thanks @liywjl and @gtrufitt

gtrufitt:
👍🏻

PR #1512, opened 2020-05-21
Show a fallback message when error.statusText isn't present

oliverlloyd:
## What does this change?
Often when getting a 500 for an api call we don't see any `statusText` value. In these cases we're currently logging an undefined error message. Here we use a more useful fallback message

## Why?
This update helps with Sentry error messages like this
Screenshot 2020-05-21 at 11 10 10

The fallback error message doesn't add any additional conext, it's present in the error page, but it does surface it sooner, which is useful

### DIFF
--- a/src/web/lib/api.tsx
+++ b/src/web/lib/api.tsx
@@ -25,7 +25,11 @@ interface FetchOptions {
 
 function checkForErrors(response: any) {
     if (!response.ok) {
-        throw Error(response.statusText);
+        console.log(response);
+        throw Error(
+            response.statusText ||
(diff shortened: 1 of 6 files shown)


src/web/lib/api.tsx
```diff
@@ -25,7 +25,11 @@ interface FetchOptions {
 
 function checkForErrors(response: any) {
     if (!response.ok) {
-        throw Error(response.statusText);
+        console.log(response);
```
shtukas, line 28: That ^

shtukas: Actually, let me make a commit :)

PR #1517, opened 2020-05-22
MatchNav

oliverlloyd:
## What does this change?
Adds the `MatchNav` component

### Mobile
Screenshot 2020-05-22 at 10 08 34

### Desktop
Screenshot 2020-05-22 at 10 08 52

## Why?
This component forms part of the work towards supporting match reports

## What's remaining?
This PR includes no solution for extending the yellow background colour all the way to the left

Screenshot 2020-05-22 at 10 20 11

This is currently achieved on frontend using negative margins so if we took the same approach then that code would live here or close to it. But we could also take a more structured approach with a separate `Section` or grid for this component and I wanted to do that investigation work before committing on one approach or another.


oliverlloyd:
[MatchNav component](https://trello.com/c/MnEXjCN0/1687-matchnav-component)

gtrufitt:
Amazing! Ccing @JamieB-gu @webb04 as think it's interesting for them given this could be a component used in their project based on the new JSON endpoint

JamieB-gu:
> based on the new JSON endpoint

What's the new JSON endpoint?

oliverlloyd:
> What's the new JSON endpoint? cc @alexduf

This one: https://github.com/guardian/frontend/pull/22601

alexduf:
I've missed something, is that new endpoint a complete re-implementation that goes to PA and fetches the data?

oliverlloyd:
> I've missed something, is that new endpoint a complete re-implementation that goes to PA and fetches the data?

No, it's the same code up until the point where it's returning the response. At that point it branches based on if the `?dcr` param is set

PR #1523, opened 2020-05-27
Remote es modules

nicl:
**UPDATE: THIS IS NOW READY FOR REVIEW.**

## What does this change?

Load epic as an ES module from contributions.theguardian.com 

### Before

Epic looks the same, but done with nasty custom JS.

### After

Loaded as a module.

## Why?

To include React/dynamic components into DCR in way that is decoupled, so that DCR itself can stay simple, and product teams can develop and update their components faster and more easily.

For more context and related code, see https://github.com/guardian/contributions-service/pull/138

## Next steps

This PR is to run this on PROD and see if there are any errors and also what performance is like. We can check this using latencies from the contributions service.


src/web/components/SlotBodyEnd.tsx
```diff
@@ -1,5 +1,8 @@
 import React, { useEffect, useState, useRef } from 'react';
+import * as emotion from 'emotion';
 import { css } from 'emotion';
+import * as emotionCore from '@emotion/core';
+import * as emotionTheming from 'emotion-theming';
```
andre1050, line 5: Important to note here that the `sendOphanReminderEvent` function below needs to be assigned to the `onReminderOpen` prop in the Epic.

src/web/components/SlotBodyEnd.tsx
```diff
@@ -168,96 +163,60 @@ const MemoisedInner = ({
...
-            }
-        }
-    // The 'data' object used in the hook never changes after 'hasBeenSeen'
-    // is set to true, so we're intentionally leaving it out of the deps array.
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [hasBeenSeen]);
+                // TODO, we need to reconsider event handling, as this is all
+                // highly coupled at the moment (better would be data
+                // attributes).
+                const props = {
```
liywjl, line 191: I do not see the point in splitting this out, nor in having the variable `epic` defined , suggestion:

```
import(/* webpackIgnore: true */ module.url).then(() => {
    setEpicMeta(meta);
    setEpicProps({
        ...module.props,
        onReminderOpen: sendOphanReminderOpenEvent,
    });
    setEpic(() => epicModule.ContributionsEpic); // useState requires functions to be wrapped
    sendOphanEpicEvent('INSERT', meta);
}).catch(error => console.log(`epic - error is: ${error}`));

```

src/web/components/SlotBodyEnd.tsx
```diff
@@ -168,96 +163,60 @@ const MemoisedInner = ({
             contributionsServiceUrl,
             isSensitive,
         });
-        getBodyEnd(contributionsPayload, `${contributionsServiceUrl}/epic`)
+
+        window.automat = {
```
liywjl, line 167: Is there no way to pass this directly to the component? If not maybe worth having this as `window.guardian.automat`? Ideally having this defined in `DCRBrowserDocumentData` type object in `index.d.ts` would be great, but I am not sure how best to achieve this is

src/web/components/SlotBodyEnd.tsx
liywjl: it is also worth adding some performance analysis here, refer to https://github.com/guardian/dotcom-rendering/blob/master/src/web/components/CMP.tsx#L20 for how to implement it in imports

src/web/components/SlotBodyEnd.tsx
nicl: Initially (for the simple prototype), it worked as you suggest - passing React as an explicit dependency. But It gets tricky when you have nested components as you want to develop them in the usual way (import react, etc.) so that it's easy to use them with other tooling, such as storybook and for familiarity - basically to keep them as regular React functions. If you force react/emotion etc as direct arguments, you have to depart from this.

I agree exporting on the window feels sad. Nesting under `window.guardian`makes sense though.

window.guardian.d.ts
```diff
@@ -34,7 +34,14 @@ declare global {
         };
         GoogleAnalyticsObject: string;
         ga: UniversalAnalytics.ga;
-        initAutomatJs?: Function;
+
+        // TODO expose as type from Automat client lib
+        automat: {
+            react: any;
```
andre1050, line 40: Is there any way to cast these to something more specific, ideally including definitions from the actual libraries? I can't think of a quick way of doing that but there might be one. It would stop anybody assigning the wrong library to any of these vars (it could even go further and add some guard against versioning).

src/web/components/SlotBodyEnd.tsx
nicl: Ah great, both really useful points. I was wondering about how to track performance.

nicl: @liywjl does the perf tracking get sent somewhere?

liywjl: it gets picked up by [speed curve](https://speedcurve.com/) its hidden deep in one of the reports, best to ask @gtrufitt for access

window.guardian.d.ts
nicl: Yeah agree this would be nice. I think these could be good things to add to the Automat client lib if/when we have fully validated this approach.

src/web/components/SlotBodyEnd.tsx
gtrufitt: Yeah, if you add user timing metrics, they'll be captured on runs and we can add a graph to speedcurve. I am going to create a new 'user timings' dashboard so we can get an overview of our module perf

nicl: awesome sounds good. Thanks both.

PR #1527, opened 2020-05-29
Donut

oliverlloyd:
## What does this change?
Adds a `Donut` component.

## Limits
Does not handle a single segment. You must provide at least 2 segments where each segment has a non zero value.

## Why?
For use in the MatchStats section where we want to visually represent possession between the teams during a football match


src/web/components/Donut.tsx
```diff
@@ -0,0 +1,189 @@
...
+    let segmentAngle;
+    let endRadius;
+    let arc;
+    let outer;
+    let inner;
+    let r;
+    let a;
+    let startRadius = -halfPI;
+
+    withoutZeroSections(sections).forEach(section => {
```
liywjl, line 86: i would personally use `map` and defined the `const segments` here rather than `forEach` as `map` expects something to be returned (so you do not need to do `segments.push`)

src/web/components/Donut.tsx
```diff
@@ -0,0 +1,189 @@
...
+        // We handle these at the moment by filtering them out using withoutZeroSections()
+        // TODO: Support displaying 0% for sections where value is zero
+    }
+
+    const radius = Math.min(height / 2, width / 2);
+    const cutoutRadius = radius * (percentCutout / 100);
+
+    const totalValue = sections
+        .map(section => section.value)
+        .reduce((prev, next) => prev + next);
```
liywjl, line 65:
```suggestion
    const totalValue = sections
        .reduce((acc, cur) => cur.value + acc, 0);
```

src/web/components/Donut.tsx
```diff
@@ -0,0 +1,189 @@
...
+    height = 300,
+}: Props) => {
+    if (withoutZeroSections(sections).length === 1) {
+        console.warn('The Donut component requires at least 2 sections');
+        // TODO: Support showing 100% for a single section
+        return null;
+    }
+
+    if (sections.find(section => section.value === 0)) {
+        console.warn(
```
liywjl, line 53: github doesnt seem happy with `console.warn`, maybe you can add a eslint ignore?

src/web/components/Donut.tsx
oliverlloyd: I actually find the more verbose version easier to read and understand. But I agree `prev` and `next` is misleading - updated

src/web/components/Donut.tsx
oliverlloyd: 👍 Removed

src/web/components/Donut.tsx
oliverlloyd: You're right, it's cleaner to use map but when I tried this I ended up returning the object but then needing to still run code inside the loop (updating `v`) after the return statement, which felt wrong

oliverlloyd:
> why was `src/web/components/Dropdown.test.tsx` deleted?

That's was not intentional. Restored

src/web/components/Donut.tsx
liywjl: if you move `startRadius += (section.value / totalValue) * doublePI;` above `segments.push` and then convert `segments.push` to `return { ... }` the rest should be fine?

oliverlloyd: I think that having the start position reset at the end of the loop is useful. The code does everything using startRadius at one point and _then_ it gets reset at the end. Preserving this order might not be explicitly necessary but in terms of readability I think it helps communicate what's happening.

PR #1529, opened 2020-06-01
Update pullquote element to be full width at low breakpoint

rcrphillips:
## What does this change?
Makes the pullquote element full width on lowest breakpoint

### Before
Screen Shot 2020-06-01 at 10 37 56

### After
Screen Shot 2020-06-01 at 10 37 28

## Why?
For parity with frontend, and because it looked a  bit borked
https://trello.com/c/BL5qIg8d/1575-pullquotes-should-be-full-width-on-mobile


rcrphillips:
[PullQuotes should be full width on mobile](https://trello.com/c/BL5qIg8d/1575-pullquotes-should-be-full-width-on-mobile)

PR #1530, opened 2020-06-01
isLight

oliverlloyd:
## What does this change?
Adds the `isLight()` utility function

## Why?
So we can decide when to show white or black text above dynamic background colours. Eg. team colours for football in `MatchStats`


oliverlloyd:
[isLight function](https://trello.com/c/GRNdjydx/1723-islight-function)

PR #1532, opened 2020-06-02
Add support for supported by badging

rcrphillips:
## What does this change?
Adds the rendering of 'commercial' badges to Web. 'Commercial' happens to handle all the badging for articles, regardless of whether or not it's actually a commercial sponsorship.

### Frontend
Screen Shot 2020-06-02 at 14 01 25
Screen Shot 2020-06-02 at 14 02 18

### DCR
Screen Shot 2020-06-02 at 14 01 56
Screen Shot 2020-06-02 at 14 02 27

## Why?
https://trello.com/c/znrv4cCk/1674-add-support-for-supported-by-badging

### DIFF
--- a/src/web/components/Branding.tsx
+++ b/src/web/components/Branding.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 import { LinkStyle } from '@root/src/amp/components/elements/TextBlockComponent';
 import { textSans } from '@guardian/src-foundations/typography';
 import { css } from 'emotion';
-import { ImageWrapper } from './Card/components/ImageWrapper';
 
 const brandingStyle = (pillar: Pillar) => css`
     padding: 10px 0;
--- a/src/web/components/Branding.tsx
+++ b/src/web/components/Branding.tsx
@@ -1,10 +1,12 @@
 import React from 'react';
 import { LinkStyle } from '@root/src/amp/components/elements/TextBlockComponent';
 import { textSans } from '@guardian/src-foundations/typography';
+
 import { css } from 'emotion';
+import { palette } from '@guardian/src-foundations';
 
 const brandingStyle = (pillar: Pillar) => css`
-    padding: 10px 0;
+    padding-bottom: 10px;
     ${LinkStyle(pillar)}
 
     a, a:hover {
@@ -16,6 +18,8 @@ const brandingStyle = (pillar: Pillar) => css`
 
 const brandingLabelStyle = css`
     ${textSans.xsmall()};
+    color: ${palette.neutral[46]};
+    font-weight: bold;
 `;
 
 const brandingLogoStyle = css`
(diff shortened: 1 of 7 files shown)


src/web/components/Branding.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import { LinkStyle } from '@root/src/amp/components/elements/TextBlockComponent';
```
oliverlloyd, line 2: I could be wrong but I think we have a bit of an invisible rule where we don't use amp components in DCR?

oliverlloyd: Maybe we could use `Link` from `src-link`? If you look in discussion you'll see how we combine `Link` with a wrapper to get pillar styles. Not sure if @SiAdcock  has introduced a `PillarLink version since then?

rcrphillips:
[Add support for 'supported by' badging](https://trello.com/c/znrv4cCk/1674-add-support-for-supported-by-badging)

src/web/components/Branding.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { textSans } from '@guardian/src-foundations/typography';
+
+import { css } from 'emotion';
+import { palette } from '@guardian/src-foundations';
```
oliverlloyd, line 6: We recently moved away from importing the whole `palette` object in favour of just the thing you need. Eg:

`import { neutral} from '@guardian/src-foundations/palette';`

src/web/components/Branding.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    a, a:hover {
+        display: block;
+        border-bottom: none;
+        ${textSans.xsmall()}
+    }
+`;
+
+const brandingLabelStyle = css`
+    ${textSans.xsmall()};
```
oliverlloyd, line 20: You can: `${textSans.xsmall({ fontWeight: "bold" })};`

src/web/components/Branding.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const brandingStyle = (pillar: Pillar) => css`
+    padding-bottom: 10px;
+    ${LinkStyle(pillar)}
+
+    a, a:hover {
+        display: block;
+        border-bottom: none;
+        ${textSans.xsmall()}
+    }
```
oliverlloyd, line 16: This will probably go away if we use `Link` but in general, with css-in-js it's a good rule to never use css selectors. They create coupling which can slowly grow in bad ways...

src/web/components/Branding.tsx
gtrufitt: Yeh, we maybe want to lint against that

rcrphillips: Ah yes. Good spot. Was from the initial copy I took of the AMP branding component.

paperboyo:
Sorry that a) I’m late to this, b) I can’t answer my own question by looking at the code, but will it support different [logo resolutions](https://github.com/guardian/tagmanager/pull/355) (scale them to the same size)?

PR #1533, opened 2020-06-02
Use ts-loader Over Babel's preset-typescript

JamieB-gu:
### Before

Used [`babel-preset-typescript`](https://babeljs.io/docs/en/babel-preset-typescript) to transform TypeScript.

### After

Uses [`ts-loader`](https://github.com/TypeStrong/ts-loader) to transform TypeScript.

## Why?

This uses `ts-loader` to handle TypeScript within Webpack, as described in the [Webpack docs](https://webpack.js.org/guides/typescript/) and the [TypeScript docs](https://www.typescriptlang.org/docs/handbook/react-&-webpack.html). Making this change avoids some of [the caveats](https://babeljs.io/docs/en/babel-plugin-transform-typescript#caveats) that come with Babel's transformation of TypeScript. One benefit of this is that DCR can start to make use of `const enum`s, such as those used in the [`guardian/types`](https://github.com/guardian/types) project and [apps-rendering](https://github.com/guardian/apps-rendering), which may help with code sharing. Another is consolidation of the validation step, editor language servers, and build step on the `tsconfig.json` file, as `ts-loader` draws from this whereas the Babel preset does not.

One potential downside of performing full type-checking in development is that it takes longer and requires more memory. I've made use of the `transpileOnly` option to avoid this, as described in [the docs](https://github.com/TypeStrong/ts-loader#faster-builds), so the build is actually slightly faster than before. That said, I'm not completely clear whether this would introduce the same issue with `const enum`s, or if it's handled in a different way to Babel's transformation, thus avoiding that problem. If it becomes an issue we can turn this flag off. We could also turn this flag off in production builds to allow a full type-check as part of the webpack build, which might mean we could drop `tsc` from the validation step if desired.

**Note:** As I'm not too familiar with this project I'd appreciate any advice on what to test to make sure everything behaves as expected.


JamieB-gu:
The TC build is failing because the `stdout maxBuffer` has been exceeded 🤔, not sure why there'd be such a big change in content going to `stdout` here...

PR #1536, opened 2020-06-03
White Headline

oliverlloyd:
## What does this change?
Show a white headline for immersive comment pieces

### Before
Screenshot 2020-06-03 at 19 48 43

### After
Screenshot 2020-06-03 at 19 47 26

## Why?
In preparation for the new immersive comment layout
Screenshot 2020-06-03 at 19 49 13

### Why does it look broken?
Because this PR only adds partial support for the new layout. Further work is pending but because these styles aren't surfaced (there are no immersive comment pieces) we can release the code as it is

### DIFF
--- a/src/web/components/ArticleHeadline.stories.tsx
+++ b/src/web/components/ArticleHeadline.stories.tsx
@@ -369,6 +369,30 @@ export const ImmersiveNoMainMedia = () => (
 );
 ImmersiveNoMainMedia.story = { name: 'Immersive (with no main media)' };
 
+export const ImmersiveComment = () => (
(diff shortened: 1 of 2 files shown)


oliverlloyd:
[White Headline](https://trello.com/c/m7vUoIxt/1741-white-headline)

jfsoul:
Actually - does this need a new story?

oliverlloyd:
> Actually - does this need a new story?

Thanks, I missed this. This is a much better way to see how the changes look.

Screenshot 2020-06-04 at 09 50 04

PR #1537, opened 2020-06-03
White Byline

oliverlloyd:
## What does this change?
Show white byline text for immersive comment pieces

### Before
Screenshot 2020-06-03 at 20 08 53

### After
Screenshot 2020-06-03 at 20 08 04

## Why?
In preparation for the new immersive comment layout
Screenshot 2020-06-03 at 19 49 13

### Why does it look broken?
Because this PR only adds partial support for the new layout. Further work is pending but because these styles aren't surfaced (there are no immersive comment pieces) we can release the code as it is

### DIFF
--- a/src/web/components/HeadlineByline.stories.tsx
+++ b/src/web/components/HeadlineByline.stories.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-
+import { css } from 'emotion';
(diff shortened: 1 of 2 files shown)


oliverlloyd:
[White Byline](https://trello.com/c/edFi9ZUe/1742-white-byline)

PR #1538, opened 2020-06-03
Remote reader revenue banner pattern

jfsoul:
## What does this change?
Adds a skeleton implementation of reader revenue banner(s) served from the contributions service. It uses the ES module pattern introduced by @nicl for serving the epic. It extracts shouldShow logic into an exported method of the CMP so a decision can be made outside the component of which banner to show (we might need to make that smarter later if there are more banners added).

This cuts corners in several ways which will need to be addressed later:
- Avoids creating any types liable to change (i.e. the ones that rely on knowing what targeting or tracking data the contributions service will need)
- No tests
- Avoids changing the automat client for now to save a bit of time and to avoid publishing a version unlikely to be used properly in production. I'll add another PR for this.

This goes with https://github.com/guardian/contributions-service/pull/141

### Before
No banner

### After
image

## Why?
Gives a platform for a proper reader revenue banner to built on.


jfsoul:
[Support delegation to the contributions service for banner rendering on DCR](https://trello.com/c/1e4EiVCH/1702-support-delegation-to-the-contributions-service-for-banner-rendering-on-dcr)

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -0,0 +1,59 @@
...
+    useEffect(() => {
+        const callShouldShow = () => setShowCMP(shouldShowCMP());
+
+        if (CAPI.config.cmpUi) {
+            callShouldShow();
+        }
+    }, [CAPI.config.cmpUi]);
+
+    // Don't render anything until we know whether we can show the CMP
+    if (showCMP === null) {
```
nicl, line 30: I found this a bit confusing but I don't know of a better pattern here. Could expess intent even clearer by adding an `isUnset` method or something which just does the `=== null` check, but that is probably overkill.

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -0,0 +1,158 @@
...
+            sectionName,
+            shouldHideReaderRevenue,
+            isMinuteArticle,
+            isPaidContent,
+            tags,
+            contributionsServiceUrl,
+            isSensitive,
+        });
+
+        window.guardian.automat = {
```
nicl, line 75: Note more for myself, but we should centralise this somehow to avoid every component having to do it

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -0,0 +1,158 @@
...
+        throw Error(
+            response.statusText ||
+                `SlotBanner | An api call returned HTTP status ${response.status}`,
+        );
+    }
+    return response;
+};
+
+const sendOphanEvent = (): void => {
+    // TODO stub
```
nicl, line 17: Good reminder that we should build a better solution for this!

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
jfsoul: Yeah, and maybe this becomes a bit more obvious/rigorous if we introduce some "banner picking" functionality to support extra banners OR preferably we defer all decision making to automat.

PR #1539, opened 2020-06-03
Refactored Article Title

oliverlloyd:
## What does this change?
Refactors article title / section links

## Why?
The code here needs to handle several use cases and, because we needed to add one more, needed cleaning up.

I added the `designType` prop to the component so we can act differently for comment pieces (so we can support white test for immersive opinion pieces) and then replaced the series of if statements we had before with a nested switch statement based on `display` and `designType`.

From here there is still some branching based on what type of tag is available but it is hopefully more readable.


oliverlloyd:
[White Title](https://trello.com/c/IzlCmC6H/1743-white-title)

PR #1541, opened 2020-06-04
Don't absolute position main media for immersive opinion

oliverlloyd:
## What does this change?
To ensure the main media image is correctly positioned full width and matches the viewport height we use absolute positioning when `display` is immersive. But now that we have a new layout for immersive comment pieces we need to adjust this logic, based on `designType`

Most of the changes here are around prop drilling `designType`. The actual logic change is in `ImageComponent` and fairly straight forward.

### DIFF
--- a/src/web/components/ArticleHeadline.stories.tsx
+++ b/src/web/components/ArticleHeadline.stories.tsx
@@ -106,6 +106,7 @@ export const ShowcaseInterview = () => (
                 </div>
                 <MainMedia
                     display="standard"
+                    designType="Article"
                     hideCaption={true}
                     elements={mainMediaElements}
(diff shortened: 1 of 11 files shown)


gtrufitt:
Do we have a story for this immersive opinion or will that come later?

oliverlloyd:
> Do we have a story for this immersive opinion or will that come later?

We do have a story, it's [one of the layouts](https://www.chromatic.com/component?appId=5dfcbf3012392c0020e7140b&name=Layouts%2FImmersive&buildNumber=1594&specName=Comment%20Story). It currently looks a bit off while we drip in the required changes but should all come together soon

PR #1548, opened 2020-06-06
Refactor getIndex()

oliverlloyd:
After using `getZIndex()` in another PR I suddenly realised how great this function is! I really like how it solves the z index ducks in a row problem. But one thing that confused me was the order of the `indices` array. My strong instinct was that the stack of items should visually represent the actual z index stack and that the items at the top of the array should be at the top of the stack. Even after working out it was reversed I still made the mistake of inserting my item in the wrong place.

So here I refactor the function to work the way my brain insists that it does. This actually adds more complexity to the implementation but by paying this price of extra code, we offer future developers a more intuitive experience.

### DIFF
--- a/src/web/components/Links.tsx
+++ b/src/web/components/Links.tsx
@@ -14,28 +14,6 @@ type Props = {
     userId?: string;
 };
 
-const search = css`
-    :after {
-        content: '';
-        display: inline-block;
-        width: 5px;
-        height: 5px;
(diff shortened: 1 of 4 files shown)


gtrufitt:
This is great, makes a lot more sense to my brain looking at that array as a visual stack!

oliverlloyd:
[Refactor getIndex()](https://trello.com/c/13ptuz3m/1751-refactor-getindex)

PR #1551, opened 2020-06-08
Add guardian.functions.import as an all-browser dynamic import

nicl:
*Note, relates to https://github.com/guardian/dotcom-rendering/pull/1523 but it's unclear the best way to support dynamic import so have separated it into this PR.*

## What does this change?

    dynamicImport.b4e492c2d22f7902cc40.js             4.44 kB → 1.93 kB (gzip)
    shimport.0a5fc5187c536a619484.js                  7.81 kB → 3.18 kB (gzip)

(shimport is only loaded for IE11/non-module supporting browsers.)

This PR polyfills dynamic import for non-supporting browsers. To do this it adds a < 2kb (gz) high priority script to all requests. In addition, for browsers that don't support ES modules (IE 11), a larger polyfill is dynamically loaded (3kb gzipped).

As browser support improves we can simplify this code.

It's not possible to directly polyfill `import` as it is a keyword in Javascript.

## Why?

Dynamic import is useful for loading remote ES modules, but isn't supported in IE 11 and also older versions of the main browsers.

At the moment ~90% of browsers support it out of the box.

## Alternatives

We could remove the high priority script and do this within the main react script. This would be a bit less visible, but would perform better. (Aside from the http side, there is a fixed size overhead to each init script as they are self-contained so include polyfills.

We could just use shimport for all non-supporting browsers. This would be simpler, but also increase loaded JS for module-supporting browsers - the Google polyfill used for these is much smaller and reliable too.


src/web/browser/dynamicImport/init.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import dynamicImportPolyfill from 'dynamic-import-polyfill';
+import { startup } from '@root/src/web/browser/startup';
+
+// Provides an __import__ function to use for dynamic imports. **Only works on
+// browsers that cut the mustard (support modules).**
+export const initialiseDynamicImport = () => {
+    try {
+        new Function('url', `return import(url)`);
+        self.__import__ = (url: string) =>
```
gtrufitt, line 13: What's the benefit of `self` over `window`? It seems like it might be not be required in this context given we're unlikely to run this in a worker?

src/web/server/document.tsx
```diff
@@ -64,10 +64,12 @@ export const document = ({ data }: Props) => {
         CAPI.config && CAPI.config.commercialBundleUrl,
     ];
     const priorityLegacyScripts = [
+        getDist({ path: 'dynamicImport.js', legacy: true }),
         getDist({ path: 'sentry.js', legacy: true }),
         getDist({ path: 'react.js', legacy: true }),
     ];
     const priorityNonLegacyScripts = [
+        getDist({ path: 'dynamicImport.js', legacy: true }),
```
gtrufitt, line 72: Is `legacy: true` correct? (might need a comment to explain why if so)

window.guardian.d.ts
```diff
@@ -35,6 +35,7 @@ declare global {
         GoogleAnalyticsObject: string;
         ga: UniversalAnalytics.ga;
         initAutomatJs?: Function;
+        __import__: (url: string) => Promise<any>;
```
gtrufitt, line 38: Can we namespace the import if that's the case to something like `window.guardian.functions.import`? 

Probably needs thought if external modules are going to rely on it being available on the window in the future, so refactoring out at a later stage will be difficult.

src/web/browser/dynamicImport/init.ts
```diff
@@ -0,0 +1,45 @@
...
+            self.__import__ = shimport.load;
+        },
+    );
+};
+
+const init = (): Promise<void> => {
+    let p;
+
+    if (window.guardian.mustardCut) {
+        p = Promise.resolve(initialiseDynamicImport());
```
gtrufitt, line 37: Can you not just return within the `if` and `else` rather than reassign `p`?

window.guardian.d.ts
gtrufitt: Note: This the reason for `request changes` just because this is not easily back-out-able in the long term.

nicl:
thanks @gtrufitt - very helpful feedback.

nicl:
@gtrufitt Let me know if looks good now. In particular, I've updated the import function location as suggested to live under `guardian.functions` now.

package.json
```diff
@@ -31,6 +31,7 @@
         "@guardian/automat-client": "^0.2.16",
         "@guardian/consent-management-platform": "^3.0.4",
         "@guardian/discussion-rendering": "^2.2.22",
+        "@guardian/shimport": "^1.0.2",
```
SiAdcock, line 34: Can I just give a big 👏  to whoever came up with this name!

nicl: Ha, it's actually a fork of https://github.com/Rich-Harris/shimport I made to make it easier to consume. So credit has to go to the original author!

PR #1553, opened 2020-06-08
Callout campaign type and fixture

liywjl:
## What does this change?
Adds TS type for campaign callouts and relevant fixtures

## Why?
Needed for testing and type checking future callout campaign component


index.d.ts
```diff
@@ -133,6 +133,38 @@ type ReaderRevenuePosition =
     | 'ampHeader'
     | 'ampFooter';
 
+interface CampaignsType {
```
oliverlloyd, line 136: This name is pluralised but it isn't an array

index.d.ts
```diff
@@ -133,6 +133,38 @@ type ReaderRevenuePosition =
...
+    id: string;
+    name: string;
+    rules: [];
```
oliverlloyd, line 139: What is rule? Should this be `rules:RuleType[];`?

index.d.ts
```diff
@@ -133,6 +133,38 @@ type ReaderRevenuePosition =
...
+    priority: number;
```
oliverlloyd, line 140: Is there a more defined list of possible numbers? Like DefCon 1-5?

index.d.ts
```diff
@@ -133,6 +133,38 @@ type ReaderRevenuePosition =
...
+    activeFrom: number;
+    displayOnSensitive: boolean;
+    fields: {
+        formId: number;
+        callout: string;
+        _type: string;
+        description: string;
+        tagName: string;
+        formFields: CampaignsFeildType[];
+    };
```
oliverlloyd, line 150: Is there a point to having this object  property? Could you just not have it?

index.d.ts
```diff
@@ -133,6 +133,38 @@ type ReaderRevenuePosition =
...
+    id: string;
+    name: string;
+    rules: [];
+    priority: number;
+    activeFrom: number;
+    displayOnSensitive: boolean;
+    fields: {
+        formId: number;
+        callout: string;
+        _type: string;
```
oliverlloyd, line 146: Is there a limited list of _types? I think it would be better to create strict types to start with and fix any errors by adding more types as they surface, rather than allowing everything

index.d.ts
```diff
@@ -133,6 +133,38 @@ type ReaderRevenuePosition =
...
+        description: string;
+        tagName: string;
+        formFields: CampaignsFeildType[];
+    };
+}
+
+interface CampaignsFeildType {
```
oliverlloyd, line 153: Maybe `CampaignFeildType` if it only represents a single object?

index.d.ts
```diff
@@ -133,6 +133,38 @@ type ReaderRevenuePosition =
...
+    text_size?: number | string; // TODO: need to fix at source
```
oliverlloyd, line 154: Is it better to make this fix first rather than move past it and potentially (lets face it, probably) never fix it?

index.d.ts
```diff
@@ -133,6 +133,38 @@ type ReaderRevenuePosition =
...
+    name?: string;
+    description?: string;
+    hide_label?: string;
```
oliverlloyd, line 157: `hideLabel`?

index.d.ts
```diff
@@ -133,6 +133,38 @@ type ReaderRevenuePosition =
...
+    label?: string;
+    id?: string;
+    type: string;
+    required: string;
+    options?: {
+        label: string;
+        value: string;
+    }[];
```
oliverlloyd, line 165: It might be clearer if this were defined separately as `CampaignOptionType`.

That said, the name `options` makes no sense in this context. They _look_ more like tags of some sort. Perhaps a comment here describing what this is would help?

oliverlloyd:
Finally, typo on Field in `CampaignsFeildType`

index.d.ts
liywjl: Good point. Honestly I am not sure, the only example I have found has had an empty array. I struggle to find this value being used in frontend. Will investigate further

index.d.ts
liywjl: again not sure,  will investigate further

index.d.ts
liywjl: I agree, the structure is strange, `fields` seems to always be the same it only `formFields` that changes dynamically, so i think it makes sense to keep it as is

index.d.ts
oliverlloyd: I think it's better to only type and include the data we use. There's no reason the DCR types need to match those from frontend or even the data sent from frontend. Our types should document the data needed by the rendering tier.

index.d.ts
oliverlloyd: Lets remove any properties that we aren't using

index.d.ts
oliverlloyd: I think it's a mistake to build our types based on the data we receive. We should do the opposite, and build the types based on what we use

index.d.ts
liywjl: yeah its either a '0' or a '1' ...

oliverlloyd: It's a boolean

liywjl:
@oliverlloyd you raise some good points regarding the lack of clarity and unnecessary fields present in of some of the type defs.

One of my the future PRs, base off the work here, was to filter and move campaign callout data out of CAPI config key, and instead move it into the related block element in the block elements array in CAPI (something that could be quite costly to do with such large JSON size).

Fundamentally the core issue is that the returned data from CAPI is not always clear, consistent, nor correctly placed and would require a refactor at the source. 

However I am reluctant to make changes to the original CAPI object, as it is not always clear who depends on what, and any change made would need to be applied to the FE codebase as well. We would likely need these changes to be done first before continuing with this task.

A current workaround would be to define the types returned by CAPI (as they are) and then parsing them into a clearer format (but at the cost of making changes to a large JSON file).

oliverlloyd:
> Fundamentally the core issue is that the returned data from CAPI is not always clear, consistent, nor correctly placed and would require a refactor at the source.

I agree. However, I think that _now_ is the time to fix this problem. In my experience, the first step in these types of migrations is to look at the data coming out of frontend, learn it, understand it's boundaries and quirks, and then build the DCR model that we need.

It can be tempting to try and work quickly towards the goal of just getting something working but DCR has wider goals beyond migration and we should not neglect our other objectives such as code quality and maintainability.

src/lib/content.d.ts
```diff
@@ -249,6 +249,18 @@ interface YoutubeBlockElement {
     width?: string;
 }
 
+interface CalloutBlockElement {
+    _type: 'model.dotcomrendering.pageElements.CalloutBlockElement';
+    id: string;
+    activeFrom: number;
+    displayOnSensitive: boolean;
+    formId: number;
+    calloutTitle: string;
```
oliverlloyd, line 258: We know this is a callout so shall we just call it 'title'?

PR #1554, opened 2020-06-08
MatchStats

oliverlloyd:
## What does this change?
Adds the `MatchStats` component

```typescript
type Props = {
    home: TeamType;
    away: TeamType;
};
```

Screenshot 2020-06-08 at 13 27 44
Screenshot 2020-06-08 at 13 28 10

## Why?
So that we have the presentational component ready to insert in the page for `MatchReport` style articles


fixtures/game.ts
```diff
@@ -0,0 +1,439 @@
+export const game = {
```
liywjl, line 1: worth adding a type

PR #1557, opened 2020-06-09
Placeholder

oliverlloyd:
## What does this change?
Adds the `Placeholder` component

```typescript
type Props = {
    height: number;
    rootId?: string;
    width?: number;
    spaceBelow?: 1 | 2 | 3 | 4 | 5 | 6 | 9;
    spaceLeft?: 1 | 2 | 3 | 4 | 5 | 6 | 9;
    shouldShimmer?: boolean;
};

```

## Why?
So we have a generic grey box with a shimmer effect that can be used to build loading skeletons with the aim of reducing our [Cumulative Layout Shift](https://web.dev/cls/)

### DIFF
--- a/src/web/components/Placeholder.stories.tsx
+++ b/src/web/components/Placeholder.stories.tsx
@@ -93,3 +93,12 @@ export const Root = () => {
     );
 };
 Root.story = { name: 'with rootId set' };
+
+export const NoShimmer = () => {
+    return (
+        <Container>
+            <Placeholder height={200} shouldShimmer={false} />
+        </Container>
+    );
+};
+NoShimmer.story = { name: 'without shimmer' };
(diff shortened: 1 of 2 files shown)


gtrufitt:
I don't a portal would be the right direction to go for something like the main media. We can just reserve the space by setting the height and width of the image and a background colour without getting React on the client involved.

What scenario do you imagine this used? At the moment, I can't see a use case.

oliverlloyd:
> I don't a portal would be the right direction to go for something like the main media

Agreed. This PR is nothing to do with Portals, it simply, and only, renders a grey box. All this component does is offer some helper props around height, width and margins + a bonus optional shimmer effect.

These kind of helpers are useful for building skeletons for loading states for things like Onwards, Most Viewed, etc. I think for images it's a different use case and you can solve the problem using existing patterns such as, like you say, just setting the background colour.

oliverlloyd:
I confused things here by adding the `rootId` prop. That was added to solve a special use case where sometimes we will need to server side render the grey box to reserve space and then only later, on the client, does the box get replaced by dynamic content. To do this we needed a way to couple the server component and the client side code, and typically we do this using root ids.

PR #1558, opened 2020-06-09
Remove placeholder from root before inserting Portals

oliverlloyd:
## What does this change?
Adds an additional step when inserting a Portal to remove the root div's children

## Why?
Because we sometimes server side render Placeholders in the root divs where the Portal is being inserted and we want to remove then when we load in the client side content to prevent them continuing to exist in the background

Ref: https://stackoverflow.com/questions/3955229/remove-all-child-elements-of-a-dom-node-in-javascript


oliverlloyd:
> I feel like this would be better to fix on the server end

You can't remove the children on the server because you want them to exist on the client. It's only when the client is ready to replace the content that we want to make sure we remove any existing children prior to inserting new ones

PR #1563, opened 2020-06-09
Normalize callout CAPI data

liywjl:
## What does this change?
We convert `EmbedBlockElement` to `CalloutBlockElement` if callout campaign tag name is found matches contents of `EmbedBlockElement`

## Why?
Currently there is no way to distinguish between `EmbedBlockElement` that are callouts. Also all the data relevant to rendering the callout campaign form is found in the CAPI `config.campaigns` key.

This PR will make it easier to handle these block elements, and enable us to SSR render them.

### DIFF
--- a/src/lib/content.d.ts
+++ b/src/lib/content.d.ts
@@ -295,7 +295,8 @@ type CAPIElement =
     | VideoFacebookBlockElement
     | VideoVimeoBlockElement
     | VideoYoutubeBlockElement
-    | YoutubeBlockElement;
+    | YoutubeBlockElement
+    | CalloutBlockElement;
 
 // -------------------------------------
 // Misc
--- a/src/lib/content.d.ts
+++ b/src/lib/content.d.ts
@@ -255,7 +255,7 @@ interface CalloutBlockElement {
     activeFrom: number;
     displayOnSensitive: boolean;
     formId: number;
-    calloutTitle: string;
+    title: string;
     description: string;
     tagName: string;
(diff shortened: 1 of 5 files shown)


src/lib/content.d.ts
```diff
@@ -443,3 +445,103 @@ interface CampaignFieldSelect {
...
+
+type preNormalizedCampaignFieldType =
+    | preNormalizedCampaignFieldText
+    | preNormalizedCampaignFieldTextArea
+    | preNormalizedCampaignFieldFile
+    | preNormalizedCampaignFieldRadio
+    | preNormalizedCampaignFieldCheckbox
+    | preNormalizedCampaignFieldSelect;
+
+type preNormalizedCampaignFieldText = {
```
gtrufitt, line 478: You can create an `formInput` interface and `extends` it for each input for the common properties

gtrufitt:
This might have been answered elsewhere but why are we not doing this in the Frontend dotcomponents model? 

I don't think we should see model normalisation to this level within DCR.

src/lib/content.d.ts
liywjl: good point! I think I will break this out in another PR bec we have quite a bit of type logic changed here

liywjl:
@gtrufitt yes We discussed this here https://github.com/guardian/dotcom-rendering/pull/1553#issuecomment-640676953

But I agree this is not idea. I think making the change in the `Frontend dotcomponents model` would be way better, but my knowledge of that area is limited.

PR #1564, opened 2020-06-09
Update MatchNav

oliverlloyd:
## What does this change?
Updates the TeamType to include additional properties and adds support for a minimum height

## Why?
We want to expand `TeamType` to support responses from both the match-nav and match-stats endpoints.

And having a minimum height will help reduce jank when using a grey placeholder box while the component it loading

### DIFF
--- a/src/web/components/MatchNav.tsx
+++ b/src/web/components/MatchNav.tsx
@@ -54,17 +54,11 @@ const StretchBackground = ({ children }: { children: React.ReactNode }) => (
                 position: absolute;
                 top: 0;
                 bottom: 0;
-                /* Always stretch left */
+                width: 100vw;
                 left: -100vw;
-                /* Only stretch right below desktop */
-                right: 0;
                 ${until.desktop} {
-                    right: -20px;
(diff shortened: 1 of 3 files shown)


oliverlloyd:
[Update MatchNav](https://trello.com/c/4EsmyS6g/1764-update-matchnav)

src/web/components/MatchNav.tsx
```diff
@@ -31,11 +25,43 @@ const Row = ({ children }: { children: React.ReactNode }) => (
...
+            min-height: 230px;
+            background-color: ${brandYellow.main};
+
+            :before {
+                content: '';
+                position: absolute;
+                top: 0;
+                bottom: 0;
+                width: 100vw;
+                left: -100vw;
```
liywjl, line 58: I am curious of the use of `vw`? Is this done for parity?

oliverlloyd: By using viewport width we can expand the background to cover the full viewport. This is very similar to how frontend does the same thing

PR #1565, opened 2020-06-09
Show black title text for MatchReports

oliverlloyd:
## What does this change?
If `designType === 'MatchReport' set the title text colour to black

## Why?
Because MatchReports have a yellow background behind the title and the pilar colour looks off in this situation


oliverlloyd:
[Black title for match reports](https://trello.com/c/T5rMK8G4/1765-black-title-for-match-reports)

PR #1578, opened 2020-06-11
Wrap dropcap with p tag

liywjl:
## What does this change?

Make sure that `DropCap` component is rendered in `p` tag.

## Before
Screenshot 2020-06-11 at 16 00 32

## After
Screenshot 2020-06-11 at 21 22 00

## Why?

Google crawlers see a `span` that is outsite of `p` tag as its own content. This caused an issue when showing a preview of the article, it only includes the drop cap character and not the rest of the sentence.

example:
image_(5)

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,6 @@
         "glob": "^7.1.3",
         "he": "^1.2.0",
         "html-minifier": "^3.5.14",
-        "html-react-parser": "^0.13.0",
         "inquirer": "^5.2.0",
         "jsdom": "^16.2.2",
         "lodash.debounce": "^4.0.6",
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
         "glob": "^7.1.3",
         "he": "^1.2.0",
         "html-minifier": "^3.5.14",
+        "html-react-parser": "^0.13.0",
         "inquirer": "^5.2.0",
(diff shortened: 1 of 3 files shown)


oliverlloyd:
Do you think we can simplify this by replacing this

```typescript
        return (
            <>
                <DropCap
                    letter={firstLetter}
                    pillar={pillar}
                    designType={designType}
                />
                <RewrappedComponent
                    isUnwrapped={isUnwrapped}
                    html={sanitise(remainingLetters, sanitiserOptions)}
                    elCss={paraStyles}
                    tagName="p"
                />
            </>
        );
```

with

```
        return (
            <p>
                <DropCap
                    letter={firstLetter}
                    pillar={pillar}
                    designType={designType}
                />
                <RewrappedComponent
                    isUnwrapped={isUnwrapped}
                    html={sanitise(remainingLetters, sanitiserOptions)}
                    elCss={paraStyles}
                    tagName="p=span"
                />
            </p>
        );
```

This results in
Screenshot 2020-06-11 at 18 31 00

liywjl:
> Do you think we can simplify this by replacing this

I was unsure about adding an extra `span` to the `p` tag, but on second thought, you are right there shouldn't be downside to that approach, plus its simpler and removes the need for an HTML parser.

PR #1579, opened 2020-06-12
Callout form

liywjl:
## What does this change?
Supports a callout from that is dynamically generated based on an array of `formFields`

### After
Screenshot 2020-06-12 at 17 09 16

## Why?
Parity with FE


src/web/components/Callout/FieldLabel.tsx
```diff
@@ -0,0 +1,37 @@
...
+    ${textSans.small({ fontStyle: 'italic' })}
+    color: ${neutral[46]};
+    padding-left: 5px;
+`;
+
+export const FieldLabel = ({ formField }: { formField: CampaignField }) => (
+    <label className={fieldLabelStyles} htmlFor={formField.name}>
+        <div>
+            {formField.label}
+            {formField.required && (
```
oliverlloyd, line 25: Is this logic a bit reversed? I would expect the optional text to only show if the field is _not_ required

src/web/components/Callout/Checkboxes.tsx
```diff
@@ -0,0 +1,65 @@
...
+type Props = {
+    formField: CampaignFieldCheckbox;
+    formData: { [key in string]: any };
+    setFormData: React.Dispatch<React.SetStateAction<{ [x: string]: any }>>;
+};
+
+// We encapsulate the Component's state here instead of directly sourcing it from
+// `formData` to avoid unnessesary rerenders, as this was causing selection animation
+// to be applied whenever formData was changing
+export const Checkboxes = ({ formField, formData, setFormData }: Props) => {
```
oliverlloyd, line 15: `Checkboxes` implies multiple checkboxes will be returned but it seems this is a single item

src/web/components/Callout/Checkboxes.tsx
```diff
@@ -0,0 +1,65 @@
...
+    return (
+        <>
+            <FieldLabel formField={formField} />
+            {formField.options && (
+                <CheckboxGroup name={formField.name || ''}>
+                    {formField.options.map((option, index) => {
+                        const checkboxSelection =
+                            formField.id && formField.id in formData
+                                ? formData[formField.id]
+                                : [];
```
oliverlloyd, line 34: This logic isn't very clear. Maybe we could use some named vars to make it more obvious what's happening here?

src/web/components/Callout/Checkboxes.tsx
```diff
@@ -0,0 +1,65 @@
...
+                                    updateState(
+                                        isCheckboxChecked
+                                            ? checkboxSelection.filter(
+                                                  (ele: string) =>
+                                                      ele !== option.value,
+                                              )
+                                            : [
+                                                  ...checkboxSelection,
+                                                  option.value,
+                                              ],
```
oliverlloyd, line 55: Ditto on the unclear ternary

src/web/components/Callout/Checkboxes.tsx
```diff
@@ -0,0 +1,65 @@
...
+                                : [];
+                        const isCheckboxChecked = checkboxSelection.find(
+                            (ele: string) => ele === option.value,
+                        );
+                        return (
+                            <Checkbox
+                                key={index}
+                                label={option.value}
+                                value={option.value}
+                                name={`${formField.id}`}
```
oliverlloyd, line 43: Should this be name={formField.name}?

Minus the string literal selectors and using `name`

src/web/components/Callout/Checkboxes.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <>
+            <FieldLabel formField={formField} />
+            {formField.options && (
+                <CheckboxGroup name={formField.name || ''}>
+                    {formField.options.map((option, index) => {
+                        const checkboxSelection =
+                            formField.id && formField.id in formData
+                                ? formData[formField.id]
+                                : [];
+                        const isCheckboxChecked = checkboxSelection.find(
```
oliverlloyd, line 35: `isCheckboxChecked` suggests it's a boolean but it's actually the `checkedCheckbox`

src/web/components/Callout/FieldLabel.tsx
```diff
@@ -0,0 +1,37 @@
...
+`;
+
+const optionalTextStyles = css`
+    ${textSans.small({ fontStyle: 'italic' })}
+    color: ${neutral[46]};
+    padding-left: 5px;
+`;
+
+export const FieldLabel = ({ formField }: { formField: CampaignField }) => (
+    <label className={fieldLabelStyles} htmlFor={formField.name}>
```
oliverlloyd, line 22: I'm a bit confused with the way `name` and `id` seem to be used interchangeably - I think we should be using id here

src/web/components/Callout/FileUpload.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { FieldLabel } from './FieldLabel';
+
+const fileUploadInputStyles = css`
+    padding-top: 10px;
+    padding-bottom: 10px;
+`;
+
+type Props = {
+    formField: CampaignFieldFile;
+    formData: { [key in string]: any };
```
oliverlloyd, line 13: Is it possible to type this data better? I think we can predict what the data will be so it'd be great to build the type for it

src/web/components/Callout/FileUpload.tsx
```diff
@@ -0,0 +1,36 @@
...
+            type="file"
+            accept="image/*, .pdf"
+            required={formField.required}
+            onChange={e =>
+                setFormData({
+                    ...formData,
+                    [formField.id || '']: e.target.files && e.target.files[0],
+                })}
+        />
+        <p className="form-info-text">
```
oliverlloyd, line 32: Eh? Is there a reason for this classname?

src/web/components/Callout/Form.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+        >
+            <Form
+                formFields={calloutCampaign.formFields}
+                submitForm={() => {}}
+                formId={1234}
+            />
+        </div>
+    );
+};
+Default.story = { name: 'default' };
```
oliverlloyd, line 28: It's not clear what this test is doing. It's a nice pattern when the test data is either named to describe what we're testing or we use spreading to amend it in a way which describes it. When the data is hidden in an abstraction then there's no way to know the purpose of the test.

src/web/components/Callout/Form.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+
```
oliverlloyd, line 29: It'd be great to see more stories with variations on the form data being passed in

src/web/components/Callout/Form.tsx
```diff
@@ -0,0 +1,158 @@
...
+const footerPaddingStyles = css`
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+`;
+
+type addFormFieldProp = {
+    formField: CampaignFieldType;
+    formData: { [key in string]: any };
+    setFormData: React.Dispatch<React.SetStateAction<{ [x: string]: any }>>;
```
oliverlloyd, line 31: 'x'?

src/web/components/Callout/Form.tsx
```diff
@@ -0,0 +1,158 @@
...
+            className={formStyles}
+            onSubmit={e => {
+                e.preventDefault();
+                submitForm();
+            }}
+        >
+            <input name="formId" type="hidden" value={formId} />
+            {formFields.map((formField, index) => (
+                <div key={index} custom-guardian="callout-form-field">
+                    {addFormField({
```
oliverlloyd, line 116: You don't need to use a function here. A better method would be to compose using a component. Eg:
```
<AddFormField prop={blah} />
```

And in this case the verb 'add' probably makes less sense and you could just name it `Field` or something

src/web/components/Callout/Form.tsx
```diff
@@ -0,0 +1,158 @@
...
+            method="post"
+            className={formStyles}
+            onSubmit={e => {
+                e.preventDefault();
+                submitForm();
+            }}
+        >
+            <input name="formId" type="hidden" value={formId} />
+            {formFields.map((formField, index) => (
+                <div key={index} custom-guardian="callout-form-field">
```
oliverlloyd, line 115: What's 'custom-guardian'? I've never seen that before

src/web/components/Callout/Form.tsx
```diff
@@ -0,0 +1,158 @@
...
+                    name="twitter-handle"
+                    type="text"
+                    id="twitter-handle"
+                    tabIndex={-1}
+                    placeholder="@mytwitterhandle"
+                    value={twitterHandle}
+                    onChange={e => setTwitterHandle(e.target.value)}
+                />
+            </div>
+            {error && <div>{error}</div>}
```
oliverlloyd, line 142: Maybe need some styling here?

src/web/components/Callout/Radios.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+type fieldProp = {
+    formField: CampaignFieldRadio;
+    formData: { [key in string]: any };
+    setFormData: React.Dispatch<React.SetStateAction<{ [x: string]: any }>>;
+};
+
+// We encapsulate the Component's state here instead of directly sourcing it from
+// `formData` to avoid unnessesary rerenders, as this was causing selection animation
+// to be applied whenever formData was changin
```
oliverlloyd, line 14: I wonder if you couldn't fix the rerender problem with use of composition and useEffect?

src/web/components/Callout/TextArea.tsx
```diff
@@ -0,0 +1,35 @@
...
+            required={formField.required}
+            value={
+                formField.id && formField.id in formData
+                    ? formData[formField.id]
+                    : ''
+            }
+            onChange={e =>
+                setFormData({
+                    ...formData,
+                    [formField.id || '']: e.target.value,
```
oliverlloyd, line 31: Does it work using an empty string as key name?

src/web/components/Callout/FieldLabel.tsx
liywjl: good catch! yes previously this was comparing a string '0' and I forgot to inverse the bool

src/web/components/Callout/Checkboxes.tsx
liywjl: agreed, I was not sure what would be a better name, would `CheckboxList` be a better idea? or  `CheckboxGroup`?

src/web/components/Callout/FileUpload.tsx
liywjl: `formData` is constructed on data that is fetched at runtime, there is no way to know what those keys will be beforehand as a result

src/web/components/Callout/Form.stories.tsx
liywjl: I am not sure I follow, this is just rendering the component using fixtures, there isnt really any test (unless you are referring to visual regression?) 

Note: I am working on adding `Form.test.tsx` that will test this component

src/web/components/Callout/Form.stories.tsx
liywjl: That reminds me, I am unsure how to support a form with no fields? (its an edge case that should not arise in theory)

The `calloutCampaign` fixture contain all the types of fields we support. We can split out each field into its own story but I do not see this being particularly beneficial bar have diffs for an exact Field rather than the form

src/web/components/Callout/Checkboxes.tsx
oliverlloyd: I thought this component was returning one checkbox, no?

liywjl: no this combines all the checkboxes for that field, it also adds a `FieldLabel` which is on the roadmap to be added to `CheckboxGroup` in `src-*`.

knowing this returns a list of checkboxes, do you think I should still change the name?

oliverlloyd: Oh, these are _options_! This isn't a checkbox at all, it's a radio control

I'd lose the connotation with checkboxes and stick to 'Radio' and 'Option'. To me, a checkbox is the think you check when agreeing to terms and conditions. Not when choosing from a list of options

src/web/components/Callout/FieldLabel.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <div>
```
oliverlloyd, line 23: Is this div needed?

src/web/components/Callout/FileUpload.tsx
oliverlloyd: How about if we check the code that builds the form data to see what will be sent back and use that to construct a contract? Or, we just work out the contract that we need to prevent breaking the form and demand that this be met.

If we don't set any contract at all then we have the worst of both worlds. We are open to errors and have no contract to learn from. At least if we set a contract there could still be errors (if the backend doesn't meet it) but we're on the way to being properly typed.

src/web/components/Callout/Form.tsx
```diff
@@ -0,0 +1,180 @@
...
+                    />
+                    <hr />
+                </>
+            );
+        }
+    }
+};
+
+type FormProps = {
+    submitForm: () => void;
```
oliverlloyd, line 117: Should this be `onSubmit()`? It looks like this function is a callback

src/web/components/Callout/Radios.tsx
liywjl: I am curious what you have in mind?
I spent some time trying to get a simpler solution working but to not much luck.

Do you think it may be preferable to instead of passing down `formData` to pass down the just the field value and lift the state to the parent to avoid a re-render?

src/web/components/Callout/Form.tsx
```diff
@@ -0,0 +1,180 @@
...
+            return (
+                <>
+                    <CheckboxList
+                        formField={formField}
+                        formData={formData}
+                        setFormData={setFormData}
+                    />
+                    <hr />
+                </>
+            );
```
oliverlloyd, line 100: This is still confusing. If we're returning an array of things then why are we passing it `formField`? My brain jars when it sees this. This might seem like nitpicking but incorrect use of plurals is one of the biggest sources of confusion and can cause hard to find bugs.

I'm actually confused as to what this _should_ be because I actually can't tell from the code what type of data this is, which is a flag that there are possible improvements to be had!

oliverlloyd: In the past, when I've built forms with radios and checkboxes I've reserved the words 'radio' and 'checkbox' to only describe the actual circles or squares that a user can click on. Instead I create a 'MultiSelect' or 'Choices' or 'OptionsGroup' or something, which is the grouping of multiple choices. That component can then have a prop `multiple: boolean` which decides if radios or checkboxes are used.

liywjl: Finding the right level of abstraction is not always clear.

I see your point, I dnt like the names `CheckboxList` and `RadioList`.

From what I gather you would prefer to lift the logic of looping through options here? If so we would then run into the same issue of having unnecessary re-renders, because we loose the local state of `CheckboxList` and `RadioList`.

liywjl: > `MultiSelect`
that sounds like a much better approach! I will implement that. Thanks

src/web/components/Callout/FileUpload.tsx
liywjl: The only reason we pass down `formData` is so we can update the parent state object. Technically there is minimal relation between `formData` and `FileUpload`.
We could remove it as a prop an instead pass down a function that would only update the `formField.id` relevant in `formData` ?

liywjl:
this PR need and update to `src-*` before merge

shtukas:
Awesome job!

PR #1586, opened 2020-06-14
Adding support for football table embeds

rcrphillips:
## What does this change?
Adds the support for rendering football table embeds. E.g. https://www.theguardian.com/football/2020/jun/10/premier-league-restart-preview-no-5-burnley.

Source does not have the same font-size available, so there is a slight visual discrepancy

### Frontend
Screen Shot 2020-06-14 at 21 49 01

### DCR
Screen Shot 2020-06-14 at 21 48 39

### Why?
https://trello.com/c/5eyG4hWH/1258-tableblockelement

### DIFF
--- a/cypress/integration/e2e/article.elements.spec.js
+++ b/cypress/integration/e2e/article.elements.spec.js
@@ -75,5 +75,19 @@ describe('Elements', function() {
 
             getIframeBody().contains('Cookie policy');
         });
+
+        it('should render the football embed', function() {
+            const getBody = () => {
+                return cy
+                    .get('div[data-cy="football-table-embed"]')
+                    .should('not.be.empty')
+                    .then(cy.wrap);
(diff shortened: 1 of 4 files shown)


rcrphillips:
[TableBlockElement](https://trello.com/c/5eyG4hWH/1258-tableblockelement)

oliverlloyd:
So, retrospectively, I just wanted to place my opposition to this approach.

I don't think we should be creating this kind of coupling between our css selectors and uncontrolled strings of html from frontend. If either of these 2 things change or if there are other use cases we've not seen then there's no contract here to control the impact it will have and could lead to the kind of bugs that can only be fixed by layering on more css, and then more, and...

I would much prefer that we look at how this data is sent through on frontend and send a json object down to DCR so we can control the styling ourselves.

PR #1593, opened 2020-06-16
Create dependabot.yml

gtrufitt:
## What does this change?

Adds dependabot to watch our dependencies.


gtrufitt:
> looks good, what would the next stages be?
Dunno, merge it and see what happens? 😆

PR #1600, opened 2020-06-16
Match Reports

oliverlloyd:
## What does this change?
Adds support for Match Reports

Match report articles are special because they contain 2 components that get loaded client side. MatchNav and MatchStats. These components are complete, this PR integrates them into the page by adding them to the layout and passing in the data they need to load.

The backend currently doesn't provide the url to use to make the api call so we're using an enhancement file to construct it. This file can be removed once [this ticket](https://trello.com/c/7ODqlmN7/1758-add-match-match-url-to-dcr-model) is complete.

### Loading
To prevent content moving aorund the page while these components are loading we're using a placeholder

There's an assumption built into these placeholders that the content when it arrives will not cause the height of the final component to be greater than a fixed value. If it is then there will be some content movement. This is unavoidable without knowing in advance how much content to expect.

### Standard layout only
We're only introducing these comnponents into the standard layout. Not immersive, not showcase. By convention, Match Reports are not given the showcase or immersive display hints.


oliverlloyd:
[Integrate MatchNav and MatchStats components into the page](https://trello.com/c/2UY8eJq8/1689-integrate-matchnav-and-matchstats-components-into-the-page)

PR #1601, opened 2020-06-16
Vimeo block element support

rcrphillips:
## What does this change?
Adds DCR support for the vimeo block elements. Does include a bit of a tricksy css workaround that @oliverlloyd managed to solve. 

### DCR
Screen Shot 2020-06-16 at 21 07 38

### Frontend
Screen Shot 2020-06-16 at 22 13 31

### DIFF
--- a/src/lib/content.d.ts
+++ b/src/lib/content.d.ts
@@ -227,7 +227,6 @@ interface VideoVimeoBlockElement {
     height: number;
     width: number;
     caption: string;
-    html: string;
     credit: string;
     title: string;
     description: string;
--- a/src/lib/content.d.ts
+++ b/src/lib/content.d.ts
@@ -226,7 +226,7 @@ interface VideoVimeoBlockElement {
     url: string;
     height: number;
     width: number;
-    caption: string;
+    caption?: string;
     credit: string;
     title: string;
     description: string;
--- a/src/lib/content.d.ts
+++ b/src/lib/content.d.ts
@@ -228,6 +228,9 @@ interface VideoVimeoBlockElement {
     width: number;
     caption: string;
     html: string;
+    credit: string;
+    title: string;
+    description: string;
 }
 
 interface VideoYoutubeBlockElement {
--- a/src/lib/content.d.ts
+++ b/src/lib/content.d.ts
@@ -227,6 +227,7 @@ interface VideoVimeoBlockElement {
     height: number;
     width: number;
     caption: string;
+    html: string;
 }
 
 interface VideoYoutubeBlockElement {
--- a/src/web/components/elements/VimeoBlockComponent.tsx
+++ b/src/web/components/elements/VimeoBlockComponent.tsx
@@ -1,22 +1,17 @@
 import React from 'react';
 import { css } from 'emotion';
-// import { border } from '@guardian/src-foundations/palette';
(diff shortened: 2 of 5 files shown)


src/web/components/elements/VimeoBlockComponent.tsx
```diff
@@ -0,0 +1,56 @@
...
+    height: number;
+    width: number;
+    caption: string;
+    credit: string;
+    title: string;
+}> = ({ url, caption, title, pillar, width, height }) => {
+    // 812 is the full height on an iphone X
+    const maxHeight = 812;
+    const aspectRatio = width / height;
+    const maxWidth = maxHeight * aspectRatio;
```
gtrufitt, line 29: Could you add a comment for what this this max width is avoiding happening (and why it needs iphone x height)?

src/lib/content.d.ts
```diff
@@ -227,6 +227,10 @@ interface VideoVimeoBlockElement {
     height: number;
     width: number;
     caption: string;
+    html: string;
```
oliverlloyd, line 230: I think `html` is rendundant here and can be removed

src/web/components/elements/VimeoBlockComponent.tsx
rcrphillips: @gtrufitt It's taken from here, https://github.com/guardian/frontend/pull/17902
We can blame @RichieAHB

gtrufitt, line 29:
```suggestion
	// Constrain iframe embeds with a width to their natural width 
	// rather than stretch them to the container using
	// a max that would prevent portrait videos from being taller than an iphone X (baseline)
	// More context: https://github.com/guardian/frontend/pull/17902
    const maxWidth = maxHeight * aspectRatio;
```

gtrufitt:
```suggestion
	// Constrain iframe embeds with a width to their natural width 
	// rather than stretch them to the container using
	// a max that would prevent portrait videos from being taller than an iphone X (baseline)
	// More context: https://github.com/guardian/frontend/pull/17902
    const maxWidth = maxHeight * aspectRatio;
```

src/lib/content.d.ts
rcrphillips: good catch

RichieAHB:
Github seems to think I was mentioned on this PR but I can't see where, maybe they just wanted me to lean in and say hi 👋

shtukas:
Hi @RichieAHB :)

gtrufitt:
> Github seems to think I was mentioned on this PR but I can't see where, maybe they just wanted me to lean in and say hi 👋

Rob was blaming you for something, standard

gtrufitt:
@rcrphillips Do you need to `make gen-schema`? I can't remember if content.d.ts changes it, worth seeing

rcrphillips:
@RichieAHB https://github.com/guardian/dotcom-rendering/pull/1601#discussion_r441395430Your work on responsive embed!

RichieAHB:
I hope I left some good tech debt ... some tech debt that was soooo hard to remove that it will be there forever 😿

Oh, it was a missing comment ... that does sound like something I would do ...

PR #1612, opened 2020-06-18
Ophan tracking on the RR banner

tomrf1:
## What does this change?
Implements ophan tracking events for the reader revenue banner.
We use the INSERT and VIEW componentEvents. For the click it's enough to track the user's arrival on support.theguardian.com.

Note - there's some mostly duplicated code taken from the epic (`SlotBodyEnd.tsx`), I'm not sure where the best place to put this is.

Screen Shot 2020-06-18 at 11 05 50
Screen Shot 2020-06-18 at 11 05 55

### DIFF
--- a/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
+++ b/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
@@ -23,7 +23,6 @@ type TestMeta = {
     abTestName: string;
     abTestVariant: string;
     campaignCode: string;
-    campaignId: string;
 };
 
 const sendOphanBannerEvent = (action: OphanAction, testMeta: TestMeta): void => {
@@ -32,7 +31,7 @@ const sendOphanBannerEvent = (action: OphanAction, testMeta: TestMeta): void => 
             componentType: 'ACQUISITIONS_BANNER',
             products: ['CONTRIBUTION', 'MEMBERSHIP_SUPPORTER'],
             campaignCode: testMeta.campaignCode,
-            id: testMeta.campaignId,
+            id: testMeta.campaignCode,
         },
         abTest: {
             name: testMeta.abTestName,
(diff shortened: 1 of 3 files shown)


src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -13,8 +15,32 @@ const checkForErrors = (response: any) => {
     return response;
 };
 
-const sendOphanEvent = (): void => {
-    // TODO stub
+type HasBeenSeen = [boolean, (el: HTMLDivElement) => void];
+
+type OphanAction = 'INSERT' | 'VIEW';
```
JoemG, line 20: can this be shared with the one in slotBodyEnd?

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -13,8 +15,32 @@ const checkForErrors = (response: any) => {
...
+
+type TestMeta = {
```
JoemG, line 22: could be shared with slotBodyEnd too although would require `campaignId` to be optional

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -97,26 +130,39 @@ const MemoisedInner = ({
...
 
                 import(/* webpackIgnore: true */ module.url)
                     .then(bannerModule => {
                         setBannerProps({
                             ...module.props,
                         });
                         setBanner(() => bannerModule.Banner); // useState requires functions to be wrapped
-                        sendOphanEvent();
+                        setBannerMeta(meta);
+                        sendOphanBannerEvent('INSERT', meta);
```
JoemG, line 142: `INSERT` could be a const or enum in a shared file as it's currently used here and in slotBody as a string. Same for `VIEW`.

PR #1614, opened 2020-06-18
Use embedly embed card for embeds with url

gtrufitt:
Note: requires the model to update https://github.com/guardian/frontend/blob/master/common/app/model/dotcomrendering/pageElements/PageElement.scala#L69 to support the 'source url' passed through with embeds. 

Also, need to check that this it is ok to show the 'powered by embedly' link. 

## What does this change?
Adds [embedly cards](https://amp.dev/documentation/components/amp-embedly-card/) for embeds that we don't support natively through the amp components (we support twitter, instagram etc for example). Tik tok is an example of one we don't support natively, nor does it have an amp component to actually support it.

### Before
Either doesn't show or errors if mandatory.

### After

Screen Shot 2020-06-18 at 12 04 26

## Why?
This will vastly increase the amount of embeds we can support and the UX of the pages where the embeds are not mandatory.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -742,6 +742,7 @@ declare namespace JSX {
         'amp-live-list': any;
         'amp-audio': any;
         'amp-embed': any;
+        'amp-embedly-card': any;
     }
     /* eslint-enable @typescript-eslint/no-explicit-any */
 }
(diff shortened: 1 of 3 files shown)


mchv:
There is currently some work on reducing number of embed we will allow regarding our legal obligations for ensuring privacy of our readers. Probably worth for you to be aware of the changes (@davidfurey @juanvaccari) planned before merging this PR.

frj:
It turns out that all of the elements in CAPI with an element type of 'embed' have been created by manually pasting fragments of html embed code into the tools.

In these cases there is no 'url' field stored in CAPI for those elements.

To use the embedly component we would need to extract a 'url' from the embed html its self, its not clear if we could do this reliably in all cases.

PR #1615, opened 2020-06-18
Add missing Props for Caption within Vimeo block

rcrphillips:
## What does this change?
Adds the required props for the caption component imported into the Vimeo block.

### DIFF
--- a/src/lib/content.d.ts
+++ b/src/lib/content.d.ts
@@ -234,9 +234,8 @@ interface VideoVimeoBlockElement {
     height: number;
     width: number;
     caption?: string;
-    credit: string;
-    title: string;
-    description: string;
+    credit?: string;
+    title?: string;
 }
 
 interface VideoYoutubeBlockElement {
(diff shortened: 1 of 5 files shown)


src/lib/content.d.ts
```diff
@@ -234,9 +234,8 @@ interface VideoVimeoBlockElement {
     height: number;
     width: number;
     caption?: string;
-    credit: string;
-    title: string;
-    description: string;
```
oliverlloyd, line 239: We're losing `description` here, is that intentional?

src/web/lib/ArticleRenderer.tsx
```diff
@@ -133,12 +133,14 @@ export const ArticleRenderer: React.FC<{
                     return (
                         <VimeoBlockComponent
                             pillar={pillar}
-                            url={element.url}
+                            url="https://player.vimeo.com/video/327310297?app_id=122963"
```
gtrufitt, line 136: Guessing you don't want every vimeo to be this one 😉

rcrphillips: damnit. Testing!

oliverlloyd, line 136: I'm guessing this slipped in by mistake?

rcrphillips: It is a sweet video though.....

src/lib/content.d.ts
rcrphillips: yep

PR #1618, opened 2020-06-18
Return 404 for pages with mandatory content we don't support

gtrufitt:
Note: needs testing, not sure if this will work as expected through Fastly.

## What does this change?

Returns 404 for pages with mandatory content that isn't supported - mainly embeds.

### Before
Returned a 500 error.

### After
Returns a 404.

image

## Why?
We were caching the error in fastly which meant that the page was not purged.

Detailed info here: https://chat.google.com/room/AAAAPTsU18A/xGbM-gEDr5o

### DIFF
--- a/src/amp/server/render.tsx
+++ b/src/amp/server/render.tsx
@@ -65,8 +65,11 @@ export const render = ({ body }: express.Request, res: express.Response) => {
 
         res.status(200).send(resp);
     } catch (e) {
-        // a validation error
-        if (e instanceof TypeError) {
+        if (e.message.startsWith('404')) {
+            // Message starts with 404 (mandatory content)
+            res.status(404).send(`<pre>${e.message}</pre>`);
+        } else if (e instanceof TypeError) {
+            // a validation error
             res.status(400).send(`<pre>${e.message}</pre>`);
         } else {
             res.status(500).send(`<pre>${e.message}</pre>`);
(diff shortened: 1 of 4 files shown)


gtrufitt:
Needs article/app/renderers/RemoteRenderer.scala in Frontend updating

src/amp/server/render.tsx
```diff
@@ -65,8 +65,11 @@ export const render = ({ body }: express.Request, res: express.Response) => {
 
         res.status(200).send(resp);
     } catch (e) {
-        // a validation error
-        if (e instanceof TypeError) {
+        if (e.message.startsWith('404')) {
```
jfsoul, line 68: Could we avoid this by creating a custom error type or using the `name` property? The latter would still be a string comparison, but at least it's more obvious that it shouldn't be changed without checking elsewhere.

shtukas (approved):

shtukas:
Comes together with https://github.com/guardian/frontend/pull/22728

src/amp/server/render.tsx
shtukas: Agree with Jon, but we can card it for later :)

gtrufitt:
I think while this is probably still a problem, we've had exactly 0 reports so is probably not really a requirement

PR #1620, opened 2020-06-18
Adding DCR support for youtube embeds

rcrphillips:
## What does this change?
Enables us to support non-Guardian youtube embeds. 

One thing to note is that the responsive embeds trick may not be required as youtube seems to serve them all in a 16:9 player, but I wanted to keep it in there just in case....

### Frontend
Screen Shot 2020-06-18 at 21 43 25

### DCR
Screen Shot 2020-06-18 at 21 36 33

## Why?
https://trello.com/c/z1H0hB23/1803-youtube-video-embed


rcrphillips:
[Youtube video embed](https://trello.com/c/z1H0hB23/1803-youtube-video-embed)

rcrphillips:
@oliverlloyd I can't workout where `YoutubeEmbed` or `YoutubeBlockElement` comes in. I think this may be an issue with the elements we have in frontend?

Agree that we should standardise it though as the  existing ones clearly aren't rendering youtube  embeds on DCR - https://www.theguardian.com/film/2018/oct/17/ray-liz-review-brutal-study-of-a-family-coming-to-pieces or https://www.theguardian.com/music/2020/jun/17/madonna-where-to-start-in-her-back-catalogue?dcr

oliverlloyd:
> Agree that we should standardise it though as the existing ones clearly aren't rendering the expected youtube embeds on DCR

👍 

Seems like we have 2 distinct ways to insert you tube videos into a page. Fun.

Shall we try aiming for having just one react YouTube component and trying to merge the two frontend types into one model? I suspect the ideal YouTubeComponent will have a bit of logic from both sides. Looks like the existing one has logic we want to keep around `adTargetting` and also supports an overlay, but the new version deals with variable height which we want as well.

Maybe it's worth doing a review the backend model to see if we can bring these different embeds together in frontend? If that works then we can move forward with the work to build the ideal component

gtrufitt:
FYI https://github.com/guardian/dotcom-rendering/blob/master/docs/youtube-variations.md

PR #1630, opened 2020-06-19
Fix banner z-index, use module name

tomrf1:
1. Brings the banner to the front
2. Uses the name of the module after fetching it (this is configured per-test, server-side)

### DIFF
--- a/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
+++ b/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
@@ -129,7 +129,7 @@ const MemoisedInner = ({
                         setBannerProps({
                             ...module.props,
                         });
-                        setBanner(() => bannerModule.Banner); // useState requires functions to be wrapped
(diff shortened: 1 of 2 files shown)


src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -150,8 +150,7 @@ const MemoisedInner = ({
     if (Banner) {
         return (
             // The css here is necessary to put the container div in view, so that we can track the view
-            <div ref={setNode} className={emotion.css`position: fixed; bottom: -1px;`}>
-                hello
+            <div ref={setNode} className={emotion.css`position: fixed; bottom: -1px; width: 100%; z-index: 999`}>
```
gtrufitt, line 153: We'll probably lint use of z-index, you want https://github.com/guardian/dotcom-rendering/blob/master/src/web/lib/getZIndex.tsx

tomrf1: 👍

PR #1634, opened 2020-06-22
Image margins

oliverlloyd:
## What does this change?
Adjusts the margins used above and below images to always be 12px

As part of this change, I removed the article body level css on `figure` to cross component css

## Why?
Before we had spaces of 12 and 16 pixels but this caused inconsistent spacing when you had multiple images in close proximity, such as in photo essays


oliverlloyd:
[Refactor Image Margins](https://trello.com/c/xys2JC9y/1815-refactor-image-margins)

PR #1637, opened 2020-06-22
Less article level css

oliverlloyd:
## What does this change?
Deletes the 'global' `ul` and `li` css set at the article body level and drops it down to inside the `TextBlockComponent`

## Why?
Because in photo essays the `ul` and `li`tags are used to denote captions and we want to style them for that but this global pattern creates a css war. Rather than get in a fight I changed the scope of the original overrides to only apply to text blocks.


oliverlloyd:
[Reduce Global CSS](https://trello.com/c/jSwX13gR/1816-reduce-global-css)

src/web/components/ArticleBody.tsx
```diff
@@ -43,42 +43,6 @@ const bodyStyle = (display: Display) => css`
         width: 100%;
         height: auto;
     }
-
-    figure {
```
gtrufitt, line 47: Is this completely removed because we don't have a `figure` used without styling being applied directly to the figure element?

PR #1638, opened 2020-06-22
Added css for halfwidth images

oliverlloyd:
## What does this change?
Adds specific css for sinlge images with the role `halfWidth`

This css sets the width to 50% and moves the image to the left wrapping any adjacent content around it.

Screenshot 2020-06-22 at 09 43 33

## Why?
Normally, `halfWidth` images appear as a pair in photo essays and are 'enhanced' to be passed to the `MultiImageBlockComponent`. But sometimes they are placed by themselves and need to be styled.

### How is this different to frontend?
This is a deviation from frontend where the image is placed on the right and adjacent content given it's own 50% width styling to appear alongside. In DCR we can't do this for 2 reasons:

1. We don't want to try and clean the html to apply special 50% width css to other peices of content so it sits alongside the image. This level of coupling creates hard to maintain code.

2. Without this special css ^, we can't position the image on the right without breaking ad placement. Ads use negative right margins to be positioned in the right column but if the image is on the right then the ad will appear half inside the article body, overlapping the text.

By taking this approach of moving away from a dependency on leaners, we maintain the integrity of the code and also offer up a new image type for all article types, not just photo essays.

### DIFF
--- a/src/web/components/elements/ImageBlockComponent.stories.tsx
+++ b/src/web/components/elements/ImageBlockComponent.stories.tsx
@@ -226,3 +226,171 @@ ShowcaseTitle.story = {
         chromatic: { viewports: [980] },
     },
 };
+
+export const HalfWidth = () => {
+    return (
+        <Container>
+            <p>
+                Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+                eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut
+                enim ad minim veniam, quis nostrud exercitation ullamco laboris
+                nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor
+                in reprehenderit in voluptate velit esse cillum dolore eu fugiat
+                nulla pariatur. Excepteur sint occaecat cupidatat non proident,
+                sunt in culpa qui officia deserunt mollit anim id est laborum.
+                Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+                eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut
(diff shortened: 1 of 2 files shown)


oliverlloyd:
[halfWidth Images](https://trello.com/c/zSTHQwGR/1813-halfwidth-images)

oliverlloyd:
@HarryFischer @paperboyo How does this look to you?

https://www.theguardian.com/sport/2020/may/24/knocking-down-barriers-through-boxing-photo-essay

### Before
Screenshot 2020-06-22 at 10 20 59

### After (with other photo essay work)
Screenshot 2020-06-22 at 10 27 33

paperboyo:
I welcome the decision to open up this weighting to more (all) content*, although this is obviously Design competence. Just to check: if two subsequent images/embeds would be set to this, they would appear next to each other? And if yes: do we have styling for their (non-hacked**) captions?

*although Composer needs to be updated to allow this
** because we **definitely** do not want to allow Photo essay-style hacked captions anywhere else

oliverlloyd:
> if two subsequent images/embeds would be set to this, they would appear next to each other?

Yes, by happenstance this is true. For photoessays we do something difference when we see 2 halfWdith images. We put them into a `MultiImageBlockElement` but even outside of that this styling should create a very similar effect.

Captions might look strange though, you'd get one caption per image. As you say, to fix this we'd really need to look at an update to composer rather than more cleaning.

paperboyo:
> Captions might look strange though, you'd get one caption per image.

This is fine as long as they render correctly below each image. Makes no sense to replicate craziness of multi-image captions hack beyond photo essays without doing it properly (in Composer).

PR #1639, opened 2020-06-22
Maintain gu.alreadyVisited count

tomrf1:
## What does this change?
In frontend we increment the `gu.alreadyVisited` local storage value on each page view, and it's used to decide when to show banners.

This value is passed to the new remotely-imported banner component, see https://github.com/guardian/dotcom-rendering/blob/master/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx#L50

### DIFF
--- a/src/web/lib/alreadyVisited.ts
+++ b/src/web/lib/alreadyVisited.ts
@@ -1,0 +1,11 @@
+const AlreadyVisitedKey = 'gu.alreadyVisited';
+
+export const getAlreadyVisitedCount = (): number => {
+    const alreadyVisited = parseInt(localStorage.getItem(AlreadyVisitedKey) || "", 10);
+    return !Number.isNaN(alreadyVisited) ? alreadyVisited : 0;
+};
+
+export const incrementAlreadyVisited = () => {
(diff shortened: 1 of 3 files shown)


src/web/lib/alreadyVisited.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const getAlreadyVisitedCount = (): number => {
+    const alreadyVisited = parseInt(localStorage.getItem(AlreadyVisitedKey) || "", 10);
+    return !Number.isNaN(alreadyVisited) ? alreadyVisited : 0;
+};
+
+export const incrementAlreadyVisited = () => {
+    const alreadyVisited = getAlreadyVisitedCount();
+    localStorage.setItem(AlreadyVisitedKey, `${alreadyVisited + 1}`);
+};
```
oliverlloyd, line 11: Perhaps in a follow up PR, it would be good to wrap these attempts to access `localStorage` with a try catch. In certain situations (Safari in private mode, when out of space, and others) the call will fail and without a try catch it throws an unhandled error.

PR #1642, opened 2020-06-23
Photo Essays

oliverlloyd:
# Photo Essays
This PR adds support for photo essays to DCR. It builds on previous PRs adding new essay sspecific styles but mainly it adds the file `enhance-photoessays.ts` which acts like a cleaner, transforming the structure of the `block` array to make essays work.

Photo essays are also often wrong or broken on frontend. So this PR contains changes from frontend to try and fix some of these things.

## How do we describe them
Photo essays are a display hint in composer and appear in `CAPI` as `CAPI.config.isPhotoEssay:  boolean;` but in DCR they are a `designType: 'PhotoEssay';`.

## Quirkiness
This is a non exhaustive list of the things that make photo essays fun.

### Side by side images
In essays, images are often placed side by side, sometimes you even get a grid. This is acheived using the `halfWidth` role. In DCR, rather then treat images as separate elements, with coupled css, we use the `MultiImageBlockElement` component to solve this problem.

Screenshot 2020-06-22 at 19 44 19

### Captions
In order to display a single caption for multiple images side by side (and other groups of images) a convention or rule exists in composer that says: if an image is followed by a `<ul><li>` tag, then this tag should become the caption for the 1 or more images before it.

Screenshot 2020-06-22 at 19 57 16

### Image Titles
Similar to captions, a convention exists where if a `<h2>` tag follows an image then the text should be considered the title for the image and overlayed upon it.

Image titles are, by convention, only used in photo essays but in DCR they are an optional prop for all images. If in the future composer were updated to offer this ability to add a title, DCR will just work. DCR versions of image titles take their colour from pillar.

Images can have both captions and titles

Screenshot 2020-06-22 at 20 12 55

## Things that are different in DCR
Some things have been done differently in the DCR versions of photo essays

### Colours
Before, colours were sometimes based on pillar and sometimes blue. This created an inconsistent feel so here the pillar has been used throughput

#### Before
Screenshot 2020-06-22 at 19 49 45

#### After
Screenshot 2020-06-22 at 19 49 27

### Alignment
Consistent vertical alignment is just nice. Now we have more of it.

#### Before
Screenshot 2020-06-22 at 19 53 05

#### After
Screenshot 2020-06-22 at 19 52 47


gtrufitt:
cc/ @JamieB-gu @webb04 @paperboyo

paperboyo:
This is great! (as much as putting a 💄 on a 🐷 can ever be ;-)

PR #1645, opened 2020-06-23
Update browser-support.md

gtrufitt:
Deprecated docs message. Need to update, but been saying that a while so at least note that it needs updating!


paperboyo:
Badger.

gtrufitt:
> Badger.

Lol, not even merged but yeah 😆

PR #1647, opened 2020-06-23
Rename dynamic import polyfill to be directly on the window

jfsoul:
## What does this change?

Make the dynamic import polyfill available directly on the window, not nested under `window.guardian.functions`. This is because the "lighter" Google polyfill for import doesn't support making it available under a nested path (passing in a path with dot notation didn't have the effect of nesting). The light polyfill is used by browsers which support native imports but not dynamic imports (e.g. Chrome 62 and Edge 17).

### Before

Banner and epic not rendered on affected browsers.

### After

Chrome 62:

Screenshot 2020-06-24 at 09 55 00

Edge 17:

Screenshot 2020-06-24 at 10 04 43

## Why?

The new banner and epic both use dynamic imports to render the relevant component. This adds support for some older browsers which don't support dynamic imports out the box.

### DIFF
--- a/window.guardian.d.ts
+++ b/window.guardian.d.ts
@@ -31,9 +31,6 @@ declare global {
                     reportError: (error: Error, feature: string) => void;
                 };
             };
-            functions: {
-                import: (url: string) => Promise<any>;
-            };
             // TODO expose as type from Automat client lib
             automat: {
                 react: any;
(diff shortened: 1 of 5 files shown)


window.guardian.d.ts
```diff
@@ -44,6 +41,7 @@ declare global {
         };
         GoogleAnalyticsObject: string;
         ga: UniversalAnalytics.ga;
+        guardianPolyfilledImport: (url: string) => Promise<any>; // can't be nested beyond top level
```
gtrufitt, line 44: It'd be ideal to pop a detailed comment for what this for above this as it'll then appear in the type intellisense.

tjmw: Hey yeah that's a great shout, thanks. I've added something, does it match what you were thinking?

PR #1651, opened 2020-06-24
HighlightBlockElement

oliverlloyd:
## What does this change?
Adds a new element `HighlightBlockComponent` to  support unquoted blockquote text

```typescript
type Props = {
    html: string;
};
```

Screenshot 2020-06-24 at 07 47 59

We've branched from the style in frontend slightly here to align better with Source

Example: https://www.theguardian.com/uk-news/2020/apr/22/eddies-last-farewell-a-funeral-in-the-time-of-coronavirus-photo-essay?dcr=false

## Why?
Because we sometimes want to highlight a piece of text but it might not be a quote

### DIFF
--- a/scripts/frontend/landing/index.html
+++ b/scripts/frontend/landing/index.html
@@ -152,6 +152,11 @@
                     article: '/film/2013/feb/14/keanu-reeves-future-of-cinema',
                 },
                 {
+                    name: 'HighlightBlockElement',
+                    article:
+                        '/uk-news/2020/apr/22/eddies-last-farewell-a-funeral-in-the-time-of-coronavirus-photo-essay',
+                },
+                {
                     name: 'ImageBlockElement',
                     article:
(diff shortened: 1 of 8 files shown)


gtrufitt:
This looks semantically incorrect to me? As in, it's not used for quotes. 

I'm assuming there's some logic for how this works in photo essays, but my preference would be we re-write the element in `enhanceCapi` (or Frontend?) and send through as a different blockElement type. Something like 'highlight' maybe.

oliverlloyd:
> This looks semantically incorrect to me? As in, it's not used for quotes.
> 
> I'm assuming there's some logic for how this works in photo essays, but my preference would be we re-write the element in `enhanceCapi` (or Frontend?) and send through as a different blockElement type. Something like 'highlight' maybe.

You're right, a `blockquote` should contain a _quote_. To have a `quoted: boolean` prop or class breaks this. I've brokwn this out into a new element

oliverlloyd:
[HighlightBlockElement](https://trello.com/c/lJ8A0GzI/1823-highlightblockelement)

PR #1652, opened 2020-06-24
CaptionBlockComponent

oliverlloyd:
## What does this change?
Adds a separate `CaptionBlockComponent` for standalone captions

## Why?
Because sometimes captions can appear out of context of a specific image. This is currently only happening in photo essays but by having a separate element we allow this pattern to be used anywhere, adding flexibility.

If Composer were to be updated to allow stand alone captions at any point, this is the element and the props that would support that.


oliverlloyd:
[CaptionBlockElement](https://trello.com/c/Wu4BMzpp/1824-captionblockelement)

PR #1653, opened 2020-06-24
Speed up pre-push hook

gtrufitt:
## What does this change?

- Jest only runs on changed files
- Eslint only runs on staged files
- We don't check bundling

TSC still runs on everything, maybe opportunities there.

Before and after is with *no changes to files* so it will be slower, but still much quicker.

### Before

image

### After

image

### Update After

Ok, so added parallelisation with https://github.com/mysticatea/npm-run-all/blob/HEAD/docs/run-p.md

image

## Why?
Pre push is valuable, but 2 minutes is a long time.

### DIFF
--- a/.huskyrc
+++ b/.huskyrc
@@ -1,6 +1,6 @@
 {
     "hooks": {
-        "pre-push": "make validate && yarn run createtoc && yarn run addandcommittoc",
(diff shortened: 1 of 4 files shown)


gtrufitt:
I consider this @oliverlloyds's going away present

gtrufitt:
Merging now, will tidy up terminal later

PR #1663, opened 2020-06-25
Use Highlights for non quoted blockquotes

oliverlloyd:
## What does this change?
Replaces blockquotes with highlights when not quoted

https://www.theguardian.com/world/2020/feb/21/how-italy-blends-culture-with-cuisine-a-photo-essay

## Why?
Because there's a `quoted` class on the `blockquote` element as passed down from CAPI which denotes that the text is in fact a quote and we should show quote marks. When this class is missing the text should be styled differently, and for better semantics, placed into a new element, the `HighlightBlockElement`

### DIFF
--- a/src/model/add-highlights.test.ts
+++ b/src/model/add-highlights.test.ts
@@ -1,0 +1,157 @@
+import { addHighlights } from './add-highlights';
+import { bodyJSON } from './exampleBodyJSON';
+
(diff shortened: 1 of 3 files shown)


oliverlloyd:
[Use Caption and Highlight block element in photo essays](https://trello.com/c/kHPgcSor/1825-use-caption-and-highlight-block-element-in-photo-essays)

PR #1668, opened 2020-06-26
Hide revenue links when logged in

nitro-marky:
## What does this change?

The reader revenue links are now hidden when the user is logged in and a:

 - Paying Member
 - Digital Subscriber
 - Recent contributor
 - Hide support message is true

This will need optimisation as the links appear for a few milliseconds and then are removed

### Before

image

### After

image

### DIFF
--- a/src/web/components/ReaderRevenueLinks.tsx
+++ b/src/web/components/ReaderRevenueLinks.tsx
@@ -132,7 +132,15 @@ export const ReaderRevenueLinks: React.FC<Props> = ({
     dataLinkNamePrefix,
     inHeader,
 }) => {
+    const [isDigitalSubscriber, setIsDigitalSubscriber] = useState<boolean>(
+        false,
+    );
+    const [hideSupportMessage, setShouldHideSupportMessage] = useState<boolean>(
+        false,
+    );
+
     const [isPayingMember, setIsPayingMember] = useState<boolean>(false);
+
     const [isRecentContributor, setIsRecentContributor] = useState<boolean>(
         false,
     );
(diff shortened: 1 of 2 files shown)


nitro-marky:
[Revenue links when logged in](https://trello.com/c/FAFxNWRH/1622-revenue-links-when-logged-in)

nitro-marky:
[Contribute showing when signed in on mobile](https://trello.com/c/NDXaBMe3/1682-contribute-showing-when-signed-in-on-mobile)

src/web/components/ReaderRevenueLinks.tsx
```diff
@@ -132,74 +132,98 @@ export const ReaderRevenueLinks: React.FC<Props> = ({
...
-                    className={cx({
-                        [hiddenUntilTablet]: inHeader,
-                    })}
+    /*
+        Changed to OR statement as it's likely that at least one will will be false.
+        Default response is to show the banners
+    */
+    if (
+        isDigitalSubscriber ||
+        isPayingMember ||
```
twrichards, line 170: you don't really need these as the `gu_hide_support_messaging` intentionally aggregates all the different products into a single flag (see https://github.com/guardian/members-data-api#dotcom) so the control over this moves into the Reader Revenue sphere (makes it easy for us to add/recognise new products etc)
```suggestion
```

twrichards: note though that `isRecentContributor` is still required though as this helps to recognise contributors who bought using 'guest checkout' (not signed-in, all other products require sign-in to purchase)

twrichards:
furthermore I suspect this needs a more holistic solution, as if you look at the old implementation ([user-features.js](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/commercial/user-features.js) and [applyRenderConditions.scala.js](https://github.com/guardian/frontend/blob/master/common/app/templates/inlineJS/blocking/applyRenderConditions.scala.js) the `shouldHideSupportMessaging` function is used in various places to hide the plethora of 'support asks' across the site (EPICs, header, footer, contributions banner, subs banner, thrashers etc.)

mchv:
@nicl @amyhughes seems like another good use case to have this behind `automat`. Quite sure it was even one of your initial presentation.

jfsoul:
Definitely @mchv - it was one we considered looking at in Q1 https://docs.google.com/document/d/1_ADSACK6MQQWw5fxgIq__1E1a2f9TbZ7Jo6jUq4FFmE/edit#heading=h.nelcjvb8ar0w

nitro-marky:
[Head over byline](https://trello.com/c/dL8xiGpN/1680-head-over-byline)

PR #1673, opened 2020-06-29
Using contributions library for revenue link logic

nitro-marky:
## What does this change?

The reader revenue links are now hidden when the user is logged in.

**Now uses the Contributions library to provide the credentials**

before banner

after banner

### DIFF
--- a/src/web/components/ReaderRevenueLinks.test.tsx
+++ b/src/web/components/ReaderRevenueLinks.test.tsx
@@ -17,8 +17,6 @@ jest.mock('@frontend/web/browser/cookie', () => ({
 describe('ReaderRevenueLinks', () => {
     const contributionsCookie = 'gu.contributions.contrib-timestamp';
     const payingMemberCookie = 'gu_paying_member';
-    const digitalSubscriberCookie = 'gu_digital_subscriber';
-    const recentContributorCookie = 'gu.contributions.contrib-timestamp';
 
     const urls = {
         contribute: 'https://www.theguardian.com/contribute',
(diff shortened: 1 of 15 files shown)


nitro-marky:
[Revenue links when logged in](https://trello.com/c/FAFxNWRH/1622-revenue-links-when-logged-in)

nitro-marky:
[Contribute showing when signed in on mobile](https://trello.com/c/NDXaBMe3/1682-contribute-showing-when-signed-in-on-mobile)

src/web/components/ReaderRevenueLinks.test.tsx
```diff
@@ -17,8 +17,6 @@ jest.mock('@frontend/web/browser/cookie', () => ({
 describe('ReaderRevenueLinks', () => {
```
jfsoul, line 17: Within this test suite we had 3 tests, corresponding to the logic in ReaderRevenueLinks source, but we've dramatically changed that logic now. It no longer checks several conditions, the logic for which is within that file. 

Instead we're just making one call out to `shouldHideSupportMessaging` and rendering based off that, so I think the tests should be updated to match.  My suggestion would be 2 tests, one that checks nothing is rendered when `true` is returned, and the other that checks _something_ is rendered when `false` is returned.

To control the return value of `shouldHideSupportMessaging` you can mock it out, as has been done for `getCookie`/`addCookie` previously. You can also remove the mocking for those things, because they're no longer needed.

Last thing to check (if being completely thorough) would be to see if the tests we're removing here have already been implemented in the contributions utility we're now importing. If not, then we ought to add some, or add a card to the backlog to do that later (or more likely ask another team to own this).

nitro-marky: Ok, I'll update the test cases. 

I guess it would be best to ask the reader revenue team to update the contributions, it needs to have digital subscriber checks added to it as well.

jfsoul: I think digital subscriber would be included in the `gu_hide_support_messaging` cookie that already gets used.  (In fact all relevant products are). I suspect the extra checks for contributions are because they had an extra requirement to immediately remove all support messaging after a contribution - I don't think that necessarily makes sense for subscriptions.

nitro-marky:
Rebased to pass TeamCity CI

jfsoul:
Your rebase is probably good, but github isn't smart enough to update the comparison with master. You can fix this like so: https://stackoverflow.com/a/46782679/2584664

PR #1674, opened 2020-06-29
Add support for CCPA

sndrs:
## What does this change?

will use the CCPA for consent is the user is in the US, otherwise use the current CMP

## Why?

it will be the law


src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -13,53 +14,52 @@ type Props = {
...
+        );
     }, [CAPI.config.cmpUi]);
 
-    // Don't render anything until we know whether we can show the CMP
-    if (showCMP === null) {
-        return null;
-    }
+    if (newCMPWillShow) return null;
+
+    if (showOldCMP) return <CMP />;
```
gtrufitt, line 42: Do we need two pieces of state here, as you wait for the promise.all to resolve with both values so you can have a single pieces of state with an if and if else statement?

sndrs: i had to rejig the logic, sorry

sndrs: not sure i get you - `newCMPWillShow` could be false and `showOldCMP`, what would the single piece of state look like?

src/web/components/StickyBottomBanner/CMP.tsx
```diff
@@ -1,11 +1,30 @@
...
     setErrorHandler,
+    checkWillShowUi as checkWillShowCCPAUi,
 } from '@guardian/consent-management-platform';
+import { ccpaApplies } from '@root/src/web/lib/ccpaApplies';
 
-export const shouldShow = shouldShow_;
+// this is the wrong place to be doing this, but it's temporary
+// until we remove the old react CMP component and go 100% sourcepoint
+export const willShowNewCMP = async () => {
+    const useCCPA = await ccpaApplies();
```
liywjl, line 14: `ccpaApplies` gets called 2x here, idk how costly it is, but would it not make sense to make sure that it is only ever called once?

sndrs: it's defined in a file further down this PR – it's memoised so the check is only ever performed once

sndrs: in src/web/lib/ccpaApplies.ts

PR #1680, opened 2020-06-30
Supporter count in Aus moment header

JoemG:
Adds the Aus Moment header for DCR - we already have this live on frontend: https://github.com/guardian/frontend/pull/22692

A custom header for the upcoming Australia moment.
A special message is displayed if edition is 'au'. It includes the supporter count, which is fetched client-side and stored in a cookie for future page loads.
Also, if the user is a supporter they'll see 'Welcome back' with no ctas.

### AU Non-supporter:
Screen Shot 2020-07-01 at 10 14 41
Screen Shot 2020-07-01 at 13 25 11

### AU supporter:
Screen Shot 2020-07-01 at 10 47 10
Screen Shot 2020-07-01 at 13 25 22

### UK Non-supporter (no change)
Screen Shot 2020-07-01 at 11 50 07

### UK supporter (no change)
Screen Shot 2020-07-01 at 11 50 18


src/lib/fetchTickerData.ts
```diff
@@ -0,0 +1,47 @@
+import fetch, { Response } from 'node-fetch';
```
tomrf1, line 1: this file is very similar to https://github.com/guardian/contributions-service/blob/master/src/lib/fetchTickerData.ts

src/web/components/ReaderRevenueLinks.tsx
```diff
@@ -21,6 +23,7 @@ type Props = {
     };
     dataLinkNamePrefix: string;
     inHeader: boolean;
+    enableAusMoment2020Header: boolean,
```
paulbrown1982, line 26: Just noticed this ends in a comma not a semicolon like the others. I’d’ve expected it to be a comma tbh.

src/web/components/ReaderRevenueLinks.tsx
```diff
@@ -11,6 +11,8 @@ import { textSans, headline } from '@guardian/src-foundations/typography';
 import { from, until } from '@guardian/src-foundations/mq';
 
 import { shouldHideSupportMessaging } from '@root/src/web/lib/contributions';
+import {fetchTickerData, TickerCountType, TickerData} from "@root/src/lib/fetchTickerData";
```
liywjl, line 14: looks like prettier is not working on `pre-push`, worth running it and fixing the styling diffs

src/lib/fetchTickerData.ts
liywjl: I do not think using `node-fetch` here is a good idea, as this lib is called in the frontend part of the app. I would favor using the native `fetch` API instead.

Note: `reader-revenue-links` is hydrated client side so you can use JS without having to worry about SSR

gtrufitt: William is right here, we don't need node-fetch, can we please refactor this out?

This should be using the lib/api functionality I think

tomrf1: See https://github.com/guardian/dotcom-rendering/pull/1690

gtrufitt: 🤟

PR #1681, opened 2020-06-30
Add sourcepoint privacy link

sndrs:
## What does this change?

injects a link into the footer once CCPA CMP is initialised

#### nb
- a subsequent PR will add an equivalent TCFv1 link (it's more complex)
- in a few weeks we will remove this react component entirely, at which point `src/web/components/StickyBottomBanner/CMPPrivacyLink.ts` will also get a new home. it seems like the wrong place for this code but it keeps it together. happy to move it if @guardian/dotcom-platform prefer it elsewhere

## Why?

this link is a legal requirement

### DIFF
--- a/src/web/components/StickyBottomBanner/CMPPrivacyLink.ts
+++ b/src/web/components/StickyBottomBanner/CMPPrivacyLink.ts
@@ -1,0 +1,60 @@
+import { ccpaApplies } from '@root/src/web/lib/ccpaApplies';
+import { showPrivacyManager } from '@guardian/consent-management-platform';
+
+const show = (forceModal?: boolean) => {
+    ccpaApplies().then((useCCPA) => {
+        if (useCCPA) {
+            if (forceModal) {
+                showPrivacyManager();
+            }
+        }
+    });
+};
+
+export const addPrivacySettingsLink = (): void => {
+    if (
+        'guardian' in window &&
+        'config' in window.guardian &&
+        'switches' in window.guardian.config &&
+        'cmpUi' in window.guardian.config.switches &&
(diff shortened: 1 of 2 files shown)


src/web/components/StickyBottomBanner/CMPPrivacyLink.ts
```diff
@@ -0,0 +1,60 @@
+import { ccpaApplies } from '@root/src/web/lib/ccpaApplies';
+import { showPrivacyManager } from '@guardian/consent-management-platform';
+
+const show = (forceModal?: boolean) => {
+    ccpaApplies().then((useCCPA) => {
+        if (useCCPA) {
+            if (forceModal) {
+                showPrivacyManager();
+            }
```
gtrufitt, line 9: nit...
```suggestion
        if (useCCPA && forceModal) {
			showPrivacyManager();
```

sndrs:
@liywjl i've merged this because it needs to go live, but am totally open to updating it to work better with DCR. currently the footer is not hydrated, and the link should only show/be active when JS is enabled, so i'm not sure how we could manage it with react in a way that doesn't need _something_ vanilla?

liywjl:
> @liywjl i've merged this because it needs to go live, but am totally open to updating it to work better with DCR. currently the footer is not hydrated, and the link should only show/be active when JS is enabled, so i'm not sure how we could manage it with react in a way that doesn't need _something_ vanilla?

I was under the impression that DCR's footer was hydrated, but you are right. It may be worth Hydrating it in the future to dynamically set this Nav item. But this works for the moment 👍

PR #1685, opened 2020-07-01
integrate InteractiveAtomBlockElement

liywjl:
## What does this change?
Update `atoms-rendering` and support `InteractiveAtomBlockElement`

### Before
Screenshot 2020-07-01 at 15 08 15

### After
Screenshot 2020-07-01 at 15 07 09

## Why?

Because InteractiveAtomBlockElement is now supported thanks to the new release


gtrufitt:
cc/ @JamieB-gu @webb04

PR #1686, opened 2020-07-01
Ad slots relative positioning

mxdvl:
## What does this change?

`AdSlot.tx` now have `position: relative;`

### Before

image

### After

image

## Why?

Ad slots should have relative positioning, as some native add will have absolutely positioned elements inside. This sets the AdSlot as a [CSS Containing Block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_Block), therefore stopping any absolutely position ad to grow beyond its remit.


src/web/components/AdSlot.tsx
```diff
@@ -5,6 +5,9 @@ import { border, neutral, text } from '@guardian/src-foundations/palette';
 import { textSans } from '@guardian/src-foundations/typography';
 import { from } from '@guardian/src-foundations/mq';
 
+const adSlotStyles = css`
+    position: relative;
```
mxdvl, line 9: Adding `overflow: hidden;` seems unnecessary as only native ads will be able to interact with the commercial bundle, and we have control of the templates to prevent spilling.

PR #1702, opened 2020-07-03
Dropcaps colour fixed for comment types

nitro-marky:
## What does this change?

It stops comment Dropcaps being forced to the opinion[400] colour, it now defaults to the dark variant of the pillar type. Additional stories have been added with different comment types for verification.

There was some confusion as to why opinion comments had to be [400] and not it's dark variant but it's likely down to maintaining look and feel:

Screenshot 2020-07-03 at 09 24 16
`opinion[400]`

Screenshot 2020-07-03 at 09 25 03
`opinion dark`

**But the main changes are:** 

### Before

Screenshot 2020-07-03 at 09 44 12

### After

Screenshot 2020-07-03 at 09 44 24

### DIFF
--- a/src/web/components/DropCap.tsx
+++ b/src/web/components/DropCap.tsx
@@ -21,13 +21,14 @@ const outerStyles = (pillar: Pillar, designType: DesignType) => {
         margin-right: 4px;
     `;
 
+    /*
+        The reason pillar type 'opinion' is forced to opinion[400] is that
+        opinion.dark is much darker so it is forced to keep with similar colour
+        tones (that's my understanding anyway!)
+    */
     switch (designType) {
         case 'GuardianView':
         case 'Comment':
-            return css`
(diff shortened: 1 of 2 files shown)


nitro-marky:
[Dropcap colour](https://trello.com/c/IbOTqOzw/1836-dropcap-colour)

paperboyo:
Just a note that current dotcom has dropcaps all wrong-looking because they weren’t adjusted when fonts have changed (so, vertical metrics too). Pending an OK from eg @HarryFischer, I think the baseline of a dropcap should align to the baseline of adjacent body text (pretty sure this is how it was before font change, and in any case, it’s better). cc @blongden73

nitro-marky:
[Change Dropcap position](https://trello.com/c/Hpp7Lg25/1854-change-dropcap-position)

PR #1703, opened 2020-07-03
SVG optimisation

mxdvl:
## What does this change?

Keep SVG raw, rather than URI-encoded in source files.

### Screenshots

image
https://www.chromatic.com/snapshot?appId=5dfcbf3012392c0020e7140b&id=5eff0a1dca65740022b1d5b4

Lineups are so identical they don't get picked up by Chromatic as different:
image

## Why?

Reduce file size, keep source readable and easier to update. Every byte saved counts!


mxdvl:
Following a chat with @gtrufitt on SVG optimisation, I think I found ways to squeeze out over 200B ! 😃 

Further reading on [CSS-Tricks: Probably Don’t Base64 SVG](https://css-tricks.com/probably-dont-base64-svg/).

src/web/components/Lineup.tsx
```diff
@@ -19,6 +19,10 @@ const Row = ({ children }: { children: React.ReactNode }) => (
     </div>
 );
 
+const BackgroundRed = '#cc2b12';
```
gtrufitt, line 22: I think just keep these inline

mxdvl: I was wondering if they should move to palette, actually. Having them inline fails the 'color-no-hex' test in TeamCity:

https://teamcity.gutools.co.uk/viewLog.html?tab=buildLog&buildTypeId=dotcom_rendering&buildId=544694&_focus=149
> Please use the src-foundations palette variables instead of hex values  color-no-hex

gtrufitt: Looking at a few match reports, it actually looks like these are dynamic and need to be passed down with the match report data https://www.theguardian.com/tone/matchreports

gtrufitt: @shtukas are these colours on the data model?

mxdvl: These are the yellow and red card icons and as far as I am aware their colour [is compiled once in frontend](https://github.com/guardian/frontend/blob/9ae6637/tools/__tasks__/compile/images/icons.js#L56).

https://www.theguardian.com/football/2020/jul/04/derby-nottingham-forest-championship-match-report

I could not find any information in the JSON file that could suggest these are in fact dynamic.

gtrufitt: Ah ok, yup, leave as is then, I don't think they need to be in palette in this case unless @SiAdcock thinks so

SiAdcock: Yeah these colours look pretty unique to me, I'd say keep them hardcoded in DCR.

PR #1714, opened 2020-07-08
Callout terms and conditions link styles override

liywjl:
## What does this change?
Remove a tag styling, also added callout to article story for demo purposes 

### Before
Screenshot 2020-07-08 at 17 22 07
Screenshot 2020-07-08 at 17 22 03

### After
Screenshot 2020-07-08 at 17 17 30
Screenshot 2020-07-08 at 17 17 34

## Why?
underline should only apply on hover


amyhughes (approved):
👍 looks good

PR #1718, opened 2020-07-09
CSS Typo: Height =/= Width

mxdvl:
## What does this change?

Fixes a typo, as there is an overriding rule for: `max-width: 80vw` further down.

## Why?

The original `80vh` seemed like a typo, and this removes the horizontal scroll bar showing on AMP.


mxdvl:
I've noticed the horizontal scroll bar shown is actually due to the `.expander` elements.

PR #1722, opened 2020-07-13
reversed sections array of donut

mxdvl:
## What does this change?

Reverses array of possession donut, so `home` and `away` are aligned with the rest of the report.
[DCR now](https://www.theguardian.com/football/2020/jul/12/bournemouths-dominic-solanke-spells-double-trouble-for-10-man-leicester?dcr) matches the [frontend version](https://www.theguardian.com/football/2020/jul/12/bournemouths-dominic-solanke-spells-double-trouble-for-10-man-leicester).

### Before

image

### After

image

## Why?

The `Donut` element is filled clockwise, making the first item (`home`)of the array visually on the right, whereas it is otherwise on the left. Maybe the fill direction could be a prop?

### DIFF
--- a/src/web/components/MatchStats.tsx
+++ b/src/web/components/MatchStats.tsx
@@ -203,7 +203,7 @@ export const MatchStats = ({ home, away }: Props) => (
                                     label: away.name.slice(0, 3).toUpperCase(),
                                     color: away.colours,
                                 },
-                            ]}
+                            ].reverse()}
                         />
                     </Center>
                     <br />


mxdvl:
Maybe @SiAdcock and the design team would have an opinion on which direction—clockwise or counterclockwise—a donut should fill in by default, and set this as the default value for the prop?

benwuersching:
In the above Match stats example, I also find it's better if the donut fills counterclockwise.

PR #1724, opened 2020-07-14
Implement the AB Test Framework & Cypress Integration Tests

gtrufitt:
## What does this change?

- Adds the AB Test Framework core and react libraries (core is a peer dependency of react library)
- Uses the provider `<ABProvider>` to initialise the AB test library in the root of the project to allow use of `useAB` hook
- Implements a simple 'test' AB test that makes invisible (data attribute) modifications to the page so that we can use Cypress to run integration tests

It implements [AB Tests library Version 1 release candidate](https://github.com/guardian/ab-testing/pull/22)

*Other small changes:*

- Update Cypress
- Add switches to the CAPIBrowserType

*Notes:*

- Implements a `GU_mvt_id_local` with `isDev` check so that we're able to test against localhost cookies

## Why?

Ab tests are a requirement on DCR for incoming work, this implementation ensures that we're supporting as expected and testing the integration.

### DIFF
--- a/cypress/integration/e2e/article.e2e.spec.js
+++ b/cypress/integration/e2e/article.e2e.spec.js
@@ -64,7 +64,13 @@ describe('E2E Page rendering', function () {
 
     describe('AB Tests - Can modify page', function () {
         it('should set the correct AB Test Variant', function () {
-            // Variant by cookie
+            // The A/B test has an audience of 0.001 and test offset of 0
+            // Therefore the test will run from MVTIds 0 - 100
+            // As there are two variants and therefore each variant falls into odd or even numbers
+            // The 'control' will be even numbers, and the 'variant' will be odd
+            // We test 99 here for the MVT cookie (set by Fastly usually) as expecting it to return
+            // the 'variant' of the A/B test
+            // See https://ab-tests.netlify.app/ for help caluclating buckets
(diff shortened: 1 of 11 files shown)


cypress/integration/e2e/article.e2e.spec.js
```diff
@@ -62,6 +62,51 @@ describe('E2E Page rendering', function () {
...
+            cy.get('[data-cy-ab-runnable-test=variant]').should('be.visible');
+
+            cy.get('[data-cy-ab-user-in-variant=ab-test-not-in-test]').should(
+                'not.be.visible',
+            );
+        });
+
+        it('should not edit the page if not in an AB test', function () {
+            // Not in test
+            cy.setCookie('GU_mvt_id_local', '500', {
```
nicl, line 90: I think this is fine for now, but it might be nice in future to have helper methods to get the right numbers/values here and also guide the reader in what they mean.

gtrufitt: Yeah, for sure, I've add a more details comment for now!

gtrufitt:
> Looks good! I need to read up on the actual AB library/implementation to get a better sense of things.

I'm part way though the docs, will share them when finished! The actual API & underlying code is very similar to Frontend

PR #1728, opened 2020-07-16
Minimum 8 instances in ASG

jfsoul:
Adds a couple more t3.small instances to the ASG as we are anticipating needing more capacity when we do an audience size increase of ~67%.

I think this should be considered a short term change - we should review more thoroughly the best configuration for running node/express efficiently (i.e. experiment with instance type, number and pm2 cluster mode).

### DIFF
--- a/cloudformation.yml
+++ b/cloudformation.yml
@@ -60,8 +60,8 @@ Mappings:
       Value: rendering
   StageMap:
     PROD:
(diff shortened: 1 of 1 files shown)


jfsoul:
On 8 instances we do accrue some CPU credits overnight, on 6 we were not:

image

PR #1732, opened 2020-07-17
Use article count opt out flag

TomPretty:
## What does this change?
We now send contributions service a flag indicating if the user has opted out of article counts. This allows contributions service to send the correct epic. Additionally, only increment the article count when this flag isn't set.

### DIFF
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -27,11 +27,12 @@ import { getUser } from '@root/src/web/lib/getUser';
 import { getCommentContext } from '@root/src/web/lib/getCommentContext';
 import { FocusStyleManager } from '@guardian/src-foundations/utils';
 import { incrementAlreadyVisited } from '@root/src/web/lib/alreadyVisited';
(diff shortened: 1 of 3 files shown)


src/web/components/App.tsx
```diff
@@ -166,7 +166,9 @@ export const App = ({ CAPI, NAV }: Props) => {
     // We should monitor this function call to ensure it only happens within an
     // article pages when other pages are supported by DCR.
     useEffect(() => {
-        incrementWeeklyArticleCount();
+        if (getCookie('gu_article_count_opt_out') == null) {
```
tomrf1, line 169: let's add a const for `"gu_article_count_opt_out"` to here https://github.com/guardian/dotcom-rendering/blob/master/src/web/lib/contributions.tsx#L15

TomPretty: Yeah great shout - I've added the const + added a hasOptedOutOfArticleCount function to clean up the logic a little

PR #1734, opened 2020-07-20
Revert #1733 and fix dailyArticleCount unshift issue

coldlink:
## What does this change?
- Reverts #1733 
- Fixes issue introduced in #1730 

In #1730 I introduced the `dailyArticleCount` to the `localStorage` based on code from `frontend`. However the format I added the JSON to localStorage was different to that in frontend.

In frontend the `gu.history.dailyArticleCount` key was in the following JSON format:
```json
{
   "value": [
      {
         "day":18459,
         "count":1
      },
      {
         "day":18457,
         "count":1
      },
      {
         "day":18446,
         "count":1
      }
   ]
}
```
However in my previous PR I had stored it into local storage as an array instead:
```json
[
      {
         "day":18459,
         "count":1
      },
      {
         "day":18457,
         "count":1
      },
      {
         "day":18446,
         "count":1
      }
 ]
```
This caused an issue when merged into production where if a browser already had the dailyArticleCount in the `{ value: ... }` format, it would parse this into an object okay, but when trying to increment (using `unshift` see #1733 ) it was unable to do so, as it was parsed as an object.

I've now changed this so that in DCR to it also adds the JSON to the localStorage in the same format as frontend so that it's compatible.

### DIFF
--- a/src/web/lib/dailyArticleCount.test.ts
+++ b/src/web/lib/dailyArticleCount.test.ts
@@ -1,0 +1,135 @@
+import {
+    DailyArticleCount,
+    DailyArticleCountKey,
+    getDailyArticleCount,
+    incrementDailyArticleCount,
+} from './dailyArticleCount';
+
+const today = Math.floor(Date.now() / 86400000);
+
+const validDailyArticleCount: DailyArticleCount = [
+    {
+        day: today,
+        count: 3,
+    },
+    {
+        day: today - 1,
+        count: 2,
+    },
+    {
+        day: today - 2,
+        count: 1,
+    },
+];
+
+describe('dailyArticleCount', () => {
+    beforeEach(() => {
+        localStorage.clear();
+    });
+
+    it('gets an empty array of daily article counts by default', () => {
+        const output = getDailyArticleCount();
+
+        expect(output).toEqual([]);
+    });
+
+    it('returns array of valid daily article counts if they exist', () => {
+        localStorage.setItem(
+            DailyArticleCountKey,
+            JSON.stringify(validDailyArticleCount),
+        );
+
+        const output = getDailyArticleCount();
+
+        expect(output).toEqual(validDailyArticleCount);
+    });
+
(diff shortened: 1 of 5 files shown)


src/web/lib/dailyArticleCount.test.ts
```diff
@@ -0,0 +1,153 @@
...
+    it('returns empty array of daily article counts if failed to parse local storage, and remove the key from localStorage', () => {
+        localStorage.setItem(DailyArticleCountKey, 'not a valid json string');
+
+        const output = getDailyArticleCount();
+
+        expect(output).toEqual([]);
+        expect(localStorage.getItem(DailyArticleCountKey)).toBeNull();
+    });
+
+    it('returns empty array of daily article counts if invalid json format, and removes the key from localStorage', () => {
```
coldlink, line 56: this test now checks that the localStorage string is in the correct format when parsed

src/web/lib/dailyArticleCount.ts
```diff
@@ -0,0 +1,76 @@
+interface DailyArticle {
+    day: number;
+    count: number;
+}
+
+export type DailyArticleCount = Array<DailyArticle>;
+
+// in localStorage, has format {"value":[{"day":18459,"count":1},{"day":18457,"count":1},{"day":18446,"count":1}]} to match frontend
+interface DailyArticleCountLocalStorage {
```
coldlink, line 9: interface which determines the format that dailyArticleCount should be stored in localStorage as

src/web/lib/dailyArticleCount.ts
```diff
@@ -0,0 +1,76 @@
...
+    if (!dailyCount) {
+        return [];
+    }
+
+    try {
+        const { value }: DailyArticleCountLocalStorage = JSON.parse(dailyCount);
+
+        // check if value parsed correctly
+        if (!value || !value.length) {
+            throw new Error('Invalid gu.history.dailyArticleCount value');
```
coldlink, line 27: throw an error if the parsed JSON was not in correct format

PR #1738, opened 2020-07-22
Update uwraphtml to support multiple html elements

gtrufitt:
## What does this change?

In order to support different element types in `textBlockComponent` - ie: `<ul>` elements, and not have to wrap them in a span, then we need to support multiple checks in the `unwrapHtml` utility.

The underlying issue being fixed here is lack of styling for UL elements because they were nested within a span. Ensuring that we don't do this is more semantic markup. 

### Before
image

image

### After

image

image

## Why?
ul - Styling and semantic markup.

There's likely other elements where this is useful, so extending out to allow multiple 'fixes' ensures that we can re-use the utility effectively.

### DIFF
--- a/src/web/components/elements/TextBlockComponent.tsx
+++ b/src/web/components/elements/TextBlockComponent.tsx
@@ -108,9 +108,23 @@ export const TextBlockComponent: React.FC<Props> = ({
     forceDropCap,
     isFirstParagraph,
 }: Props) => {
-    const { willUnwrap: isUnwrapped, unwrappedHtml } = unwrapHtml({
-        prefix: '<p>',
-        suffix: '</p>',
+    const {
+        willUnwrap: isUnwrapped,
+        unwrappedHtml,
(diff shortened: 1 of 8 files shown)


src/web/components/elements/TextBlockComponent.tsx
```diff
@@ -108,9 +108,23 @@ export const TextBlockComponent: React.FC<Props> = ({
...
+            {
+                unwrappedElement: 'p',
+                prefix: '<p>',
+                suffix: '</p>',
+            },
+            {
+                unwrappedElement: 'ul',
+                prefix: '<ul>',
+                suffix: '</ul>',
+            },
```
gtrufitt, line 126: This is the main part that fixes the ul issue - because we can now check multiple elements, we can ensure that we're unwrapping `ul`s too.

shtukas: +1

PR #1739, opened 2020-07-22
Add padding to slot body end between tablet and desktop

TomPretty:
## What does this change?
Add some padding to the slot body end component between tablet and desktop. 

### Before
Screenshot 2020-07-22 at 11 57 42

### After
Screenshot 2020-07-22 at 11 58 03

## Why?
This aligns the article body with the slot body end at this breakpoint.

### DIFF
--- a/src/web/components/SlotBodyEnd.tsx
+++ b/src/web/components/SlotBodyEnd.tsx
@@ -8,6 +8,7 @@ import {
     logView,
     getWeeklyArticleHistory,
 } from '@guardian/automat-client';
+import { between } from '@guardian/src-foundations/mq';
 import {
     isRecurringContributor,
     getLastOneOffContributionDate,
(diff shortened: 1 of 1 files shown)


oliverlloyd (commented):
This creates a bit of a coupling between here and the same css in `ArticleBody`. I wonder if it's worth dropping the root tag for this element inside `ArticleBody` so it gets wrapped by the same container (with this padding already set). A bit like:

```
        <div className={cx(bodyStyle(display), linkColour[pillar])}>
            <ArticleRenderer
                display={display}
                elements={blocks[0] ? blocks[0].elements : []}
                pillar={pillar}
                designType={designType}
                adTargeting={adTargeting}
                host={host}
            />
            {showBodyEndSlot && <div id="slot-body-end" />}
        </div>
```

You'd need to prop drill `showBodyEndSlot` but this feel natural.

The downside I guess is we're no longer rendering all of these root tags in one place but I think that was a nice to have but not essential.

oliverlloyd:
@TomPretty what's the best thing to do with this PR? It looks like we probably still want this fix

PR #1742, opened 2020-07-23
Default the element to paragraph for textblockcomponent

gtrufitt:
## What does this change?

I had previously made the default element for unwrapped components a `span` but they should be paragraphs.

### Before

Screen Shot 2020-07-23 at 14 31 00

### After

Screen Shot 2020-07-23 at 14 30 53
## Why?

In this case, the markup we get through is super odd, so we need to fix that root issue but this more gracefully handles badmarkup

Screen Shot 2020-07-23 at 14 33 20

### DIFF
--- a/src/model/unwrapHtml.ts
+++ b/src/model/unwrapHtml.ts
@@ -38,7 +38,7 @@ export const unwrapHtml = ({
 
     // span is the default unwrappedElement
     const unwrappedElement =
-        (willUnwrap && matchingFix && matchingFix.unwrappedElement) || 'span';
+        (willUnwrap && matchingFix && matchingFix.unwrappedElement) || 'p';
 
     return {
         unwrappedHtml,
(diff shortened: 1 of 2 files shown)


paperboyo:
> the markup we get through is super odd, so we need to fix that root issue

That Composer’s fault?

gtrufitt:
> > the markup we get through is super odd, so we need to fix that root issue
> 
> That Composer’s fault?

I don't *think* so, it comes through CAPI fairly clean:

image

It looks remarkably similar to the issue @jfsoul fixed here: https://github.com/guardian/frontend/pull/22675/files#diff-a78830a0e574fa7224f8e492f8ac2c9aL601

Likely need `.body` here in the cleaner: https://github.com/guardian/frontend/blob/976ff4a1b99c030b2526b0384072bd7bf39bc0c1/common/app/views/support/HtmlCleaner.scala#L844

Will have to test locally!

PR #1747, opened 2020-07-24
QandA Atom Integration

nitro-marky:
## What does this change?

It adds the QandA Atom definition to `types` and returns the component in `ArticleRenderer` and `App` for hydration. 

Ophan tracking actions have also been added to record 'EXPAND', 'LIKE' and 'DISLIKE'

To see the Atom structure check out this link to the PR in atoms-rendering:

[https://github.com/guardian/atoms-rendering/pull/28](url)


nitro-marky:
[Atom: QandaAtom](https://trello.com/c/He2VyjwF/1731-atom-qandaatom)

gtrufitt:
> One thing to check is if this componentEvent is valid in Ophan, current production sends:

It looks like they're an Option, so no https://github.com/guardian/ophan/blob/078d62eb9ea12018055c3cea6cbcc41d0f30883a/the-slab/app/extractors/ComponentEventExtractor.scala#L17 however, if you make it generic as per https://github.com/guardian/ophan/blob/078d62eb9ea12018055c3cea6cbcc41d0f30883a/the-slab/app/extractors/ComponentEventExtractor.scala#L17 you probably want to send the default empty array

src/model/window-guardian.ts
```diff
@@ -99,6 +99,22 @@ export const makeGuardianBrowserCAPI = (CAPI: CAPIType): CAPIBrowserType => {
         [] as CalloutBlockElement[],
     );
 
+    const qandaAtomsWithIndex: QABlockElement[] = CAPI.blocks[0].elements.reduce(
```
gtrufitt, line 102: As per the [Rule of Three](https://blog.codinghorror.com/rule-of-three/) it may be worth abstracting this into a `blockElementWithIndex` function now, probably in lib so it's testable.

```
const blockElementWithIndex = (blocks, blockType, indexName) => {
	// same body
}
```

?

src/web/components/App.tsx
```diff
@@ -299,6 +302,38 @@ export const App = ({ CAPI, NAV }: Props) => {
...
+                        }}
+                        expandHandler={() => {
+                            recordComponentEvent(
+                                'EXPAND',
+                                qandaAtom.id,
+                                'QANDA_ATOM',
+                            );
+                        }}
+                    />
+                </Hydrate>
```
gtrufitt, line 335: As per https://github.com/guardian/atoms-rendering/pull/28/files#r460810610 you can probably pass an `ophanHandler` as single prop, and then deal with the function names inside the atom

PR #1749, opened 2020-07-24
Identity | Sign In Gate | Documentation and Unit Tests

coldlink:
## What does this change?
- Adds comprehensive documentation on setting up, developing, and testing Sign In Gate things.
- Adds unit tests for the exported methods in `displayRule` and `dismissGate`
- Update `localStorageLookupKey` in `dismissGate` to remove optional values since both `name` and `variant` are required.

## Why?
- Engineering knowledge sharing on working with the sign in gate
- Although Cypress tests cover display rules and dismiss gate, the unit tests check the individual method functionality


src/web/components/SignInGate/README.md
```diff
@@ -0,0 +1,389 @@
...
+7. Add it to Storybook by modifying `SignInGate.stories.tsx`
+8. Update Cypress tests in the `cypress/integration/e2e/sign-in-gate.spec.js`, and any unit tests e.g. `displayRule.test.ts`
+
+## Full Guide
+
+### General
+
+For all possible sign in gates, the following rules should apply:
+
+-   All variant names should be unique across all tests
```
dominickendrick, line 20: Do you think this is worth enforcing at a type level? Using an `enum` or `union` type ?

coldlink: That would only be useful to stop spelling mistakes, but not enforce the uniqueness.

It might still be useful though, as you could do something like:

```ts
// in one file
enum SignInGateVariants {
  MainVariant = 'main-variant-1',
  MainControl = 'main-control-1',
  PatientiaVariant = 'patientia-variant-1',
  PatientiaControl = 'patientia-control-1',
}

// in another file
const variant = SignInGateVariants.MainVariant;
```

And then you'll only need to change the variant name in one place to make a change. 🤔

gtrufitt:
Great docs!

PR #1750, opened 2020-07-24
Add lighthouse ci github action

jorgeazevedo:
## What does this change?

Adds lighthouse-ci as a github action

jorgeazevedo_dotcom-rendering_at_ja-gh-test

## Why?

We want to be able to measure performance at the PR level!

I had to patch the prod server to make this work, and would appreciate feedback on at least 2 fronts:

* I had to fix a few routing problems that were causing lighthouse to fail. Since this is the prod server, I'm looking to understand if my fixes actually didn't break anything else! Also, if I can get rid of one of the routes
* I need to make some functionality optional, but it feels a bit clunky. I'm looking for feedback on how best to make it optional!

### Fixes

There's two main fixes

First, adding `next` back to `app.use((err: any`. This was removed during a rollout of linting since the argument is unused., but that fourth argument is not optional. Without it, the `res` argument will change and the server will throw the following error

```
TypeError: res.status is not a function
```

Second, removing ` app.get('*'`. AFAIK This was added to facilitate lighthouse ci, but it actually makes the server throw a bunch of errors that ultimately break lighthouse. Joshua report the issue to google [here](https://github.com/GoogleChrome/lighthouse-ci/issues/313), but it's actually due to our server misconfiguration. 

Basically, that generic route will swallow up calls to `assets` and other paths. This will make a cascade of errors like

```
{ FetchError: invalid json response body at https://www.theguardian.com/assets/dynamicImport.d6662415116a9da9a585.js.json?dcr=true reason: Unexpected token < in JSON at position 0
    at /Users/jorge_azevedo/code/dotcom-rendering/node_modules/node-fetch/lib/index.js:272:32
    at process._tickCallback (internal/process/next_tick.js:68:7)
  message:
   'invalid json response body at https://www.theguardian.com/assets/dynamicImport.d6662415116a9da9a585.js.json?dcr=true reason: Unexpected token < in JSON at position 0',
  type: 'invalid-json' }
```

And others paths such as `https://www.theguardian.com/asset-manifest.json` and `https://www.theguardian.com/robots.txt`. This breaks lighthouse, like literally when you run it in the chrome UI, not just the CI tool.

### Optionals

For our production server to run in the github action virtual environment, we need to
- Not log to `/var/log/dotcom-rendering/dotcom-rendering.log` since we won't have the permissions
- Not request aws credentials
- Not push metrics to cloudwatch

I've ended going for a somewhat verbose env variable DISABLE_LOGGING_AND_METRICS to turn both metric collection and log4js off, instead of keeping log4js logging to a file that the CI environment can access.

### DIFF
--- a/src/app/logging.ts
+++ b/src/app/logging.ts
@@ -3,7 +3,7 @@ import { configure, getLogger, addLayout, shutdown } from 'log4js';
 
 const logLocation =
     process.env.NODE_ENV === 'production'
-        ? '/var/log/dotcom-rendering/dotcom-rendering.log'
+        ? `${path.resolve('logs')}/dotcom-rendering.log`
         : `${path.resolve('logs')}/dotcom-rendering.log`;
 
 const logFields = (logEvent: any): any => {
--- a/src/app/logging.ts
+++ b/src/app/logging.ts
@@ -2,9 +2,6 @@ import path from 'path';
 import { configure, getLogger, addLayout, shutdown } from 'log4js';
 
 const logLocation =
-    process.env.DISABLE_LOGGING_AND_METRICS === 'true'
-        ? `${path.resolve('logs')}/dotcom-rendering.log`
-        :
     process.env.NODE_ENV === 'production'
         ? '/var/log/dotcom-rendering/dotcom-rendering.log'
         : `${path.resolve('logs')}/dotcom-rendering.log`;
@@ -28,7 +25,17 @@ addLayout('json', () => {
     };
 });
 
-configure({
+const log4js_disabled = {
+    appenders: {
+        console: { type: 'console' },
+    },
+    categories: {
+        default: { appenders: ['console'], level: 'off' },
+        off: { appenders: ['console'], level: 'off' },
+    },
+};
+
+const log4js_enabled = {
     appenders: {
         console: { type: 'console' },
         fileAppender: {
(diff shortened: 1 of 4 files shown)


jorgeazevedo:
@shtukas I ended up following your suggestion in standup and actually getting the checks to pass. They leave annotations on on the diff, which makes it quite hard to read!

src/app/logging.ts
```diff
@@ -56,8 +60,9 @@ if (process.env.NODE_ENV === 'development') {
 }
 
 export const logger =
+    process.env.DISABLE_LOGGING_AND_METRICS === 'true'
+        ? getLogger('off')
+        :
     process.env.NODE_ENV === 'development'
         ? getLogger('development')
         : getLogger();
```
gtrufitt, line 68: Linter not happy, maybe splitting it out something like this makes sense?

```suggestion
const loggerOff = process.env.DISABLE_LOGGING_AND_METRICS === 'true' && 'off';
const loggerDev = process.env.NODE_ENV === 'development' && 'development';

// If Disabled, turn off logger. If in Dev, set logger to development.
export const logger = getLogger(loggerOff || loggerDev || undefined);
```

jorgeazevedo: Thanks for the suggestion! I've seen a similar construct to set default values for variables, but it's the first time I'm seeing this one. Learned something!

There was also another nested tertiary operator that I had to untangle at the top of the file. In doing that I ended up splitting the initialisation of log4js, which then changed this part of the code. Have a look, see what you think!

src/app/server.ts
```diff
@@ -144,32 +140,19 @@ if (process.env.NODE_ENV === 'production') {
...
+    if (process.env.DISABLE_LOGGING_AND_METRICS !== "true") {
+        setInterval(() => {
+            recordBaselineCloudWatchMetrics();
+        }, 10 * 1000);
+    }
 
     app.listen(port);
     // eslint-disable-next-line no-console
-    console.log(`Started production server on port ${port}`);
+    console.log(`Started production server on port ${port}\nready`);
```
gtrufitt, line 157: Did you mean to keep the ready here?

jorgeazevedo: yeah, lighthouse ci is waiting for that to show up in stdout to know when to kick off the tests!

jorgeazevedo:
Thanks! We need to decide what we want to test (we can have multiple URLs) and maybe if we want to add a threshold for the check to go red.

PR #1753, opened 2020-07-27
Pass submitComponentEvent prop to remote banner

GHaberis:
## What does this change?

This PR passes the `submitComponentEvent` function as prop to remote Banner components as a prop. This means we can track Ophan "actions" from our Banners, such as click events.

Related PRs:

`frontend`: https://github.com/guardian/frontend/pull/22832
`contributions-service`: https://github.com/guardian/contributions-service/pull/196

NOTE: The original PR https://github.com/guardian/dotcom-rendering/pull/1743 was reverted due to changes going into master after it had been raised which conflicted with it's changes causing the validation Github actions to fail.

### DIFF
--- a/src/web/browser/ophan/ophan.ts
+++ b/src/web/browser/ophan/ophan.ts
@@ -87,7 +87,15 @@ export type TestMeta = {
 export const submitComponentEvent = (
     componentEvent: OphanComponentEvent,
 ): void => {
-    window.guardian.ophan.record({ componentEvent });
+    if (
+        window.guardian &&
+        window.guardian.ophan &&
+        window.guardian.ophan.record
+    ) {
+        window.guardian.ophan.record({ componentEvent });
+    } else {
+        throw new Error("window.guardian.ophan.record doesn't exist");
+    }
 };
 
 export const sendOphanComponentEvent = (
--- a/src/web/browser/ophan/ophan.ts
+++ b/src/web/browser/ophan/ophan.ts
@@ -84,6 +84,12 @@ export type TestMeta = {
     products?: OphanProduct[];
 };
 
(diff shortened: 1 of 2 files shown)


src/web/browser/ophan/ophan.ts
```diff
@@ -84,6 +84,12 @@ export type TestMeta = {
     products?: OphanProduct[];
 };
 
+export const submitComponentEvent = (
+    componentEvent: OphanComponentEvent,
+): void => {
+    window.guardian.ophan.record({ componentEvent });
+};
```
mxdvl, line 91: I think you need to make sure that `window.guardian.ophan.record`, as is done on [line 137][L137].

```suggestion
): void => {
  if (
        window.guardian &&
        window.guardian.ophan &&
        window.guardian.ophan.record
    ) {
      window.guardian.ophan.record({ componentEvent });
    } else {
        throw new Error("window.guardian.ophan.record doesn't exist");
    }
};
```

[L137]: https://github.com/guardian/dotcom-rendering/pull/1753/files#diff-6bc07cd001d12de643e0551db421254eR137

GHaberis: Thanks @mxdvl, we weren't making this check previously when calling`window.guardian.ophan.record` from `sendOphanComponentEvent` so didn't deem it necessary, however I appreciate the extra security of doing so.

gtrufitt: Looks like the third usage, so abstracting it out into a `record` function and using that might be beneficial!

GHaberis:
@mxdvl @gtrufitt I've created a `record` function in Ophan as discussed.

PR #1759, opened 2020-07-27
Track (remote) subscriptions banner view event in GA

GHaberis:
## What does this change?

This PR adds a tracking call to Google Analytics (GA) that's fired when a subscriptions banner served from the remote `contributions-service` has been seen. This event is currently recorded in GA for subscriptions banners that are NOT served from the remote `contributions-service`, so we're just making sure we retain this data, as requested by the stakeholder.

To achieve this I've introduced the `trackNonClickInteraction` function into the `ga` module, the same implementation is used in `frontend`.

**frontend equivalent PR:** https://github.com/guardian/frontend/pull/22842

### DIFF
--- a/src/web/browser/ga/ga.ts
+++ b/src/web/browser/ga/ga.ts
@@ -182,10 +182,17 @@ export const trackNonClickInteraction = (actionName: string): void => {
         const send = `${tracker.name}.send`;
 
         ga(send, 'event', 'Interaction', actionName, {
-            nonInteraction: true, // to avoid affecting bounce rate
+            /**
+             * set nonInteraction to avoid affecting bounce rate
+             * https://support.google.com/analytics/answer/1033068#NonInteractionEvents
+             */
+            nonInteraction: true,
         });
     } else {
         const error = new Error("window.ga doesn't exist");
-        window.guardian.modules.sentry.reportError(error);
+        window.guardian.modules.sentry.reportError(
+            error,
+            'trackNonClickInteraction',
+        );
     }
 };
--- a/src/web/browser/ga/ga.ts
+++ b/src/web/browser/ga/ga.ts
@@ -176,9 +176,13 @@ export const sendPageView = (): void => {
 };
 
 export const trackNonClickInteraction = (actionName: string): void => {
-    const send = `${tracker.name}.send`;
+    const { ga } = window;
 
-    window.ga(send, 'event', 'Interaction', actionName, {
-        nonInteraction: true, // to avoid affecting bounce rate
-    });
+    if (ga) {
+        const send = `${tracker.name}.send`;
+
+        ga(send, 'event', 'Interaction', actionName, {
+            nonInteraction: true, // to avoid affecting bounce rate
+        });
+    }
 };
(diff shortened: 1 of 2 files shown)


src/web/browser/ga/ga.ts
```diff
@@ -174,3 +174,11 @@ export const sendPageView = (): void => {
 
     trackLCP(send);
 };
+
+export const trackNonClickInteraction = (actionName: string): void => {
+    const send = `${tracker.name}.send`;
+
+    window.ga(send, 'event', 'Interaction', actionName, {
```
paulbrown1982, line 181: Do you need a safety check that the `ga` function exists on `window`?

GHaberis: I took a look at the the other functions in this module such as `sendPageView` and they're not checking, but I appreciate a check will provide an extra layer of security so will add one.

src/web/browser/ga/ga.ts
```diff
@@ -174,3 +174,15 @@ export const sendPageView = (): void => {
...
 };
+
+export const trackNonClickInteraction = (actionName: string): void => {
+    const { ga } = window;
+
+    if (ga) {
+        const send = `${tracker.name}.send`;
+
+        ga(send, 'event', 'Interaction', actionName, {
+            nonInteraction: true, // to avoid affecting bounce rate
```
johnduffell, line 185: > send event `Interaction`, non interaction = true
doesn't sound right

src/web/browser/ga/ga.ts
johnduffell: depends on your idea of safety, I know we differ on this Paul, but I'd rather fail loudly than silently lose tracking.  The more bad data we accumulate, the harder it is to get out of.

src/web/browser/ga/ga.ts
i-hardy: Could be worth adding in a bit more detail about what exactly we're doing here, eg a link to https://support.google.com/analytics/answer/1033068#NonInteractionEvents or a similar explanation

GHaberis: Good points. Honestly, I copied the long-standing (4+ years) implementation of `trackNonClickInteraction`, including comments that we've been using on `frontend`: https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/analytics/google.js#L13, so unfortunately I don't have the detail behind why we've done this in the past.

src/web/browser/ga/ga.ts
GHaberis: We could throw an error in the event it's not present, as we do here: https://github.com/guardian/dotcom-rendering/blob/master/src/web/browser/ophan/ophan.ts#L95, should then be captured and reported in `sentry`.

GHaberis: In fact there's a `reportError` function exported by the `sentry` module that can be used rather than throwing a `new Error`: https://github.com/guardian/dotcom-rendering/blob/e9741e59c9120198fc1fc76ebefeeb3d5427c0de/src/web/browser/sentry/sentry.tsx#L58

src/web/browser/ga/ga.ts
johnduffell: wasn't clear it was just a copy! good old github.  But that link is useful Imogen!

GHaberis: I've updated the comment, adding the link @i-hardy provided.

src/web/browser/ga/ga.ts
GHaberis: I'm now reporting any instances where `window.ga` is undefined with `reportError`.

PR #1760, opened 2020-07-28
Branding image should not be enlarged

jfsoul:
## What does this change?
Removes the usage of MaintainAspectRatio for branding images, because images do that natively (if I understand what that's for). Instead we control the max width of the image by setting to 100% and let the image's dimensions dictate the width below that threshold. This means we're not enlarging low resolution images that get used for branding.

### Before
image

### After
image

## Why?
Spotted by editorial that images were blurry and too large.

### DIFF
--- a/src/web/components/Branding.tsx
+++ b/src/web/components/Branding.tsx
@@ -52,7 +52,12 @@ export const Branding: React.FC<{
                     rel="nofollow"
                     aria-label={`Visit the ${branding.sponsorName} website`}
                 >
-                    <img src={branding.logo.src} alt={branding.sponsorName} />
+                    <img
+                        src={branding.logo.src}
+                        alt={branding.sponsorName}
+                        height={branding.logo.dimensions.height}
+                        width={branding.logo.dimensions.width}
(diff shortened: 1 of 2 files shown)


src/web/components/Branding.tsx
```diff
@@ -44,22 +46,14 @@ export const Branding: React.FC<{
...
-                        rel="nofollow"
-                        aria-label={`Visit the ${branding.sponsorName} website`}
-                    >
-                        <img
-                            src={branding.logo.src}
-                            alt={branding.sponsorName}
-                        />
-                    </a>
-                </MaintainAspectRatio>
+                    <img src={branding.logo.src} alt={branding.sponsorName} />
```
gtrufitt, line 55: I think we want height and width here again?

jfsoul: thanks, will add those

jfsoul: Turns out that we lose the fixed aspect ratio by adding these, I'll have to remove this in another PR I think.

PR #1764, opened 2020-07-28
Remove "AU" restriction on use of contributions-service

GHaberis:
## What does this change?

The `contributions-service` remote banners were previously only relevant to Australia so we restricted requests to the `countryCode` matching "AU". Now subscriptions banners are served from the service we should remove this restriction.

### DIFF
--- a/src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
+++ b/src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
@@ -44,7 +44,7 @@ export const StickyBottomBanner = ({
 
     if (showOldCMP) return <CMP />;
 
-    const showRRBanner = CAPI.config.remoteBanner && countryCode === 'AU';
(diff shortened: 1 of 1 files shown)


tomrf1:
should this wait for the ec2 migration?

GHaberis:
> should this wait for the ec2 migration?

I'm going to merge this, we still have the `remoteBanner` switch to turn off requests to the service if this is still an issue.

GHaberis:
> > should this wait for the ec2 migration?
> 
> I'm going to merge this, as we still have the `remoteBanner` switch to turn off requests to the service if this is still an issue.

OK, I've checked with @nicl and the migration to ec2 is due to be completed this week, we discussed the impact of removing this restriction before the migration and it was deemed low so the change should be OK. If it's a problem we can turn the `remoteBanner` switch off until migration is complete.

PR #1771, opened 2020-07-29
Add disclaimer block for articles containing affiliate links

rcrphillips:
## What does this change?
Adds the disclaimer block for when we have an article that contains affiliate links. The font-size is unfortunately magic number, but there's no existing size within the design system to enable it otherwise.

### Frontend
Screen Shot 2020-07-29 at 16 39 41

### DCR
Screen Shot 2020-07-29 at 16 39 23

## Why?
https://trello.com/c/yI6FAg1H/1679-support-affiliate-links

### DIFF
--- a/src/web/components/elements/DisclaimerBlockComponent.tsx
+++ b/src/web/components/elements/DisclaimerBlockComponent.tsx
@@ -11,7 +11,7 @@ const style = (pillar: Pillar) => css`
     }
 
     sup {
-        font-size: 12.75px;
+        font-size: 85%;
     }
 `;
 
@@ -19,8 +19,9 @@ export const DisclaimerBlockComponent: React.FC<{
     html: string;
     pillar: Pillar;
 }> = ({ html, pillar }) => (
-    <div
+    <span
         className={style(pillar)}
+        data-cy="affiliate-disclaimer"
         dangerouslySetInnerHTML={{
             __html: html,
         }}
--- a/src/web/components/elements/DisclaimerBlockComponent.tsx
+++ b/src/web/components/elements/DisclaimerBlockComponent.tsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import { body } from '@guardian/src-foundations/typography';
+import { pillarPalette } from '@root/src/lib/pillars';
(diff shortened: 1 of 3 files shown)


src/web/components/elements/DisclaimerBlockComponent.tsx
```diff
@@ -0,0 +1,28 @@
...
+    sup {
+        font-size: 12.75px;
+    }
+`;
+
+export const DisclaimerBlockComponent: React.FC<{
+    html: string;
+    pillar: Pillar;
+}> = ({ html, pillar }) => (
+    <div
```
jfsoul, line 22: I'm wondering whether there's a better choice for semantic markup rather than div...

rcrphillips: Uses <span> in frontend, but it felt easy to use <div> to ensure padding is properly set.

src/web/components/elements/DisclaimerBlockComponent.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const style = (pillar: Pillar) => css`
+    ${body.small()};
+
+    a {
+        color: ${pillarPalette[pillar].dark};
+    }
+
+    sup {
+        font-size: 12.75px;
```
jfsoul, line 14: Frontend is `font-size: 75%` which might get you the same but a little less magic.

rcrphillips: Have resolved.

src/web/components/elements/DisclaimerBlockComponent.tsx
rcrphillips: Have altered back to span.

jfsoul: I couldn't see any wrapping element on frontend, but I think we have to have one here to style the `html`. In which case maybe a `footer` is a reasonable choice?

>The HTML footer element represents a footer for its nearest sectioning content or sectioning root element. A footer typically contains information about the author of the section, copyright data or links to related documents.

PR #1773, opened 2020-07-30
Conditionally add MoDI analytics tag to AMP

jfsoul:
## What does this change?
Adds an extra snowplow amp-analytics tag to the business section on AMP for a trial that we are going to be part of. This has been signed off as a proof-of-concept by the necessary groups and an engineering review has been conducted and recorded by @mchv . This should be revisited once taken further than the PoC.

It is also worth noting that the tag will sit behind consent, which will be more appropriate once the TCFv2 implementation is complete in #1830 . Currently it sits behind "personalised ads" consent banner.

cc: @guardian/commercial-dev because I think this would probably be a tag that's owned by yourselves in the long run (maybe), but we're expediting this for the trial.


src/amp/components/Analytics.tsx
```diff
@@ -102,5 +102,50 @@ export const Analytics: React.FC<{
         </amp-analytics>`,
     ];
 
+    // Trial implementation of MoDI tracker tag. For now only to appear on the business section on AMP
+    if (section && section === 'business') {
+        scripts.push(
+            `<amp-analytics type="snowplow_v2">
+                <script type="application/json">
```
mxdvl, line 109: Should we be using the `<JsonScript o={{a: b}} />` syntax?

jfsoul: Not sure, I'm not familiar with that - but this is following the implementation guide given to us.

docs/architecture/3rd party technical review/000-technical-review-records.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Decision
+
+For all 3rd party technical reviews in this project, a record will be created with a filename of `XXX-technical-review.md` where XXX is the monotonically increasing number described in the linked document.
+
+Pleae note that as records will be public, specific attention should be paid about information being disclosed.
+
+## Status
+
+Approved
```
mchv, line 43: @jfsoul @nicl @gtrufitt I am documenting here some business process change regarding adding third-party scripts to our website. I think using `architecture` doc and formalism is a great way to document further technical reviews but let me know if you disagree.

docs/architecture/3rd party technical review/001-modi-poc.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### Recommended mitigations
+
+Outside of the proof of concept there are 2 options than can be investigated to ensure our performances
+are less degraded:
+ * Use [idle-until-urgent pattern](https://philipwalton.com/articles/idle-until-urgent/) using a library like [idlize](https://github.com/GoogleChromeLabs/idlize)
+ * Do not send the events client-side but server-side. This approach have several benefits:
+    - Do not require sending custom events in both apps and web platforms
+    - Ensure better privacy by design by not sharing any PII information such as IP addresses. 
+
```
mchv, line 114: @jfsoul @nicl @gtrufitt I have shared my analysis for this third-party showing the kind of details expected in such a technical review. As the scope of this proof-of-concept is limited I have not really made a performance analysis but I would expect it to be the case should it not be one.

docs/architecture/3rd party technical review/000-technical-review-records.md
jfsoul: This sounds good to me although we may want to also think about how we document (or carry out even) these reviews when they cover multiple platforms outside DCR (e.g. apps)

docs/architecture/3rd party technical review/001-modi-poc.md
jfsoul: Thanks this is very useful for us and other teams as a reference example.

docs/architecture/3rd party technical review/000-technical-review-records.md
```diff
@@ -0,0 +1,40 @@
+# 3rd-Party Technical Review Format
+
+## Context
+
+We are using succefully lightweight architecture decision records to document decisions being made about architecture.
+Last year legislation changes around data privacy (GDPR, CCPA, PECR) and their impact on our obligations means we need to increase our technical measures to ensure the safety of our readers data through:
+  * Reducing number of third-party
```
mxdvl, line 7: A little inconsistent with the main title.

There's no entry in the [style guide], but **third party** can be a noun (no hyphen, can be plural) or adjective (with hyphen, no plural)

```suggestion
  * Reducing the number of third parties
  * Reducing the number of third-party scripts
```

mxdvl (changes requested):
With the work we're doing in #1830, most analytics tags are behind consent, so I would think snowplow should be, too.

src/amp/components/Analytics.tsx
```diff
@@ -102,5 +102,50 @@ export const Analytics: React.FC<{
         </amp-analytics>`,
     ];
 
+    // Trial implementation of MoDI tracker tag. For now only to appear on the business section on AMP
+    if (section && section === 'business') {
+        scripts.push(
+            `<amp-analytics type="snowplow_v2">
```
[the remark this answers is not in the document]
mchv: Yes we were discussing this with @jfsoul, this should be under consent.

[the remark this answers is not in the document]
jfsoul: I've added this in and it works as expected. Thanks for pointing this out.

mxdvl (approved):

src/amp/components/Analytics.tsx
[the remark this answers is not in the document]
mxdvl: For reference, the [official docs on AMP](https://amp.dev/documentation/components/amp-consent/#basic-blocking-behaviors)

PR #1775, opened 2020-07-31
Run server in github action

jorgeazevedo:
## What does this change?

This is a basic attempt at us getting public urls for our open PRs. It does two things.

First, it leaves an ngrok github actions running continuously (actually times out after 6 hours)

Comparing_master___ja-ngrok_·_guardian_dotcom-rendering

When the ngrok public URL is ready, the action posts back to github a new status check 

Comparing_master___ja-ngrok_·_guardian_dotcom-rendering

The details link is set to the ngrok URL.

I think this is actually in a usable state to merge and have a play with. The big risk would be if github limits the number of parallel actions a repository can have running, for example.

Also:

When the ngrok action times out after 6 hours, it will look like this. You can just click details and re-run it

Commits_·_jorgeazevedo_dotcom-rendering

### DIFF
--- a/.github/workflows/ngrok.yml
+++ b/.github/workflows/ngrok.yml
@@ -1,0 +1,22 @@
+name: Ngrok
+on: [push]
+jobs:
+    lhci:
+        name: run
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v1
+            - name: Use Node.js 10.15.3
+              uses: actions/setup-node@v1
+              with:
+                  node-version: 10.15.3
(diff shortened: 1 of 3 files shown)


gtrufitt:
Had a thought (for future improvements) - ngrok being a bit gnarly here, we could run a local proxy server using something like https://github.com/localtunnel/server with our own domain etc. Generic enough for a dept wide solution too!

gtrufitt:
Cypress tests are now passing in main branch, please rebase, thanks! https://github.com/guardian/dotcom-rendering/pull/1829

PR #1790, opened 2020-08-04
Identity | Sign In Gate | Fix issue with Component ID not reporting correctly

coldlink:
## What does this change?
- Maps the component id in the `testIdToComponentId` object using the AB test ID now, rather than the name, since the name may not be the AB test ID as it could be the AB test `dataLinkName` instead
  - This is because `dataLinkName` is used to combine multiple tests under one reporting name, but the component id for each of those tests may be different
- Renamed `component` to `ophanComponentId` in the `SignInGateProps` type, as that better reflects the name

## Why?
- In the data lake the component id for that particular component in the data lake was sometimes reporting as `undefined` due to this issue, which may cause issues with reporting if segmenting by the component id
  - It's still possible to segment by the AB test variant, which is unique, so using the variant is actually a better way of doing it, but this was still a bug which may cause confusion

### DIFF
--- a/src/web/components/SignInGate/SignInGate.stories.tsx
+++ b/src/web/components/SignInGate/SignInGate.stories.tsx
@@ -19,7 +19,7 @@ export const mainStandalone = () => {
                 guUrl="https://theguardian.com"
                 signInUrl="https://profile.theguardian.com/"
                 dismissGate={() => {}}
-                component="test"
+                ophanComponentId="test"
             />
         </Section>
     );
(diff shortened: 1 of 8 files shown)


coldlink:
> Would some tests have caught this ? Is it worth adding tests for the tracking ?

Hmm.. it's difficult, not sure how i'd be able to test this to catch the error, since it was a problem with the implementation when looking up from the `testIdToComponentId` object.

PR #1791, opened 2020-08-04
[Experiment] Load Braze Web SDK

tjmw:
## What does this change?

Load the Braze Web SDK lazily for logged in users and show a message. This is not intended to be merged as is, in reality it looks likely we'll show the user the message in the form of a banner.

## Why?

Verifying that the mechanism we're using to load the Braze Web SDK is acceptable to teams involved.

This is the DCR equivalent of guardian/frontend#22850 (and depends on that PR for config/switch changes).


src/web/lib/getBrazeUuid.ts
```diff
@@ -0,0 +1,22 @@
+import { joinUrl } from '@root/src/web/lib/joinUrl';
+
+export const getBrazeUuid = async (ajaxUrl: string): Promise<string> => {
+    const url = joinUrl([ajaxUrl, 'user/me']);
```
jfsoul, line 4: incidentally I think the identity team were looking into whether they could remove calls to `user/me` on every page view (or at least it was on their backlog). Better let them know there's a new use case for this if you're going to merge something similar to this.

tjmw: Ah cool thanks, thats good to know :+1: We noticed that we were making a call to `/user/me` on every page for non-dcr rendered pages and it's on my radar to figure out if we could persist the data we need (the braze UUID) somewhere so we don't need to make the request every time.

tjmw:
Closing in favour of #1859

PR #1793, opened 2020-08-04
Track Core Vitals to Google Analytics

gtrufitt:
## What does this change?

Implements the official [GoogleChrome web-vitals](https://github.com/GoogleChrome/web-vitals) metrics module reporting the results to Google Analytics.

The library is small (with easy to read code) and well tested, it makes more sense than implementing these ourselves. Reporting to GA is simple and matches how we were previously reporting LCP.

We are reporting:

- *Largest Contentful Paint* when page lifecycle becomes 'hidden'
- *Cumulative Layout Shift* when page lifecycle becomes 'hidden'
- *First Input Delay* on a user interaction

I am limiting the data to these three for now to avoid sending many more events to GA. We may want to expand on this in the future, this library in particular also tracks FCP and TTFB which are not automatically tracked by GA.

## Why?

Tracking core vitals at RUM is required to allow us to measure pages that are damaging to user experience and web performance. We can see Core Vitals at a RUM level using the CrUX report which feeds into the Google Search Console but capturing the data ourselves here (and later, maybe, Ophan) means we can better segment the data to pages, sections and platforms.

Be aware: the data is only available for Chromium browsers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,8 @@
         "prettier-eslint": "^11.0.0",
         "pretty-bytes": "^4.0.2",
         "reset-css": "^5.0.1",
-        "sanitize-html": "^1.27.0"
+        "sanitize-html": "^1.27.0",
+        "web-vitals": "^0.2.4"
     },
     "devDependencies": {
         "@babel/core": "^7.5.5",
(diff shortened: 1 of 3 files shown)


gtrufitt:
> 	🚨

drama

PR #1796, opened 2020-08-05
Support reader revenue bookmarklets

tomrf1:
## What does this change?
Re-implements the logic from frontend here: https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/commercial/reader-revenue-dev-utils.js

These functions support the [bookmarklets](http://reader-revenue-bookmarklets.s3-website-eu-west-1.amazonaws.com/), which are used heavily by lots of people for viewing epics/banners.

### DIFF
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -238,12 +238,14 @@ export const App = ({ CAPI, NAV }: Props) => {
                 .catch(error => console.log('Error loading readerRevenueDevUtils', error));
                 /* eslint-enable no-console */
 
-        window.guardian.readerRevenue = {
-            changeGeolocation: loadAndRun('changeGeolocation'),
-            showMeTheEpic: loadAndRun('showMeTheEpic'),
-            showMeTheBanner: loadAndRun('showMeTheBanner'),
-            showNextVariant: loadAndRun('showNextVariant'),
-            showPreviousVariant: loadAndRun('showPreviousVariant'),
+        if (window && window.guardian) {
+            window.guardian.readerRevenue = {
+                changeGeolocation: loadAndRun('changeGeolocation'),
+                showMeTheEpic: loadAndRun('showMeTheEpic'),
+                showMeTheBanner: loadAndRun('showMeTheBanner'),
+                showNextVariant: loadAndRun('showNextVariant'),
+                showPreviousVariant: loadAndRun('showPreviousVariant'),
+            }
         }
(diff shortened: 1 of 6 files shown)


src/web/lib/getCountryCode.ts
```diff
@@ -3,14 +3,37 @@ type LocalCountryCodeType = {
...
+const TEN_DAYS = 60 * 60 * 24 * 10;
+
 function hasExpired(whenItExpires: number) {
     return new Date().getTime() > whenItExpires;
 }
 
-export const getCountryCode = async () => {
-    const TEN_DAYS = 60 * 60 * 24 * 10;
-    const COUNTRY_CODE_KEY = 'gu.geolocation';
+export const setCountryCode = (countryCode: string): void => {
```
tomrf1, line 13: this is just copied from below

src/web/components/App.tsx
```diff
@@ -228,6 +229,10 @@ export const App = ({ CAPI, NAV }: Props) => {
         FocusStyleManager.onlyShowFocusOnTabs();
     }, []);
 
+    useEffect(() => {
```
gtrufitt, line 232: I would prefer we dynamically load this, it's not going to be used by our users so having it in the bundle doesn't make sense

tomrf1: hmm, I cannot think of a way to do that without changing the bookmarklets in everyone's browsers.
Any ideas?
The bookmarklet just calls e.g. `javascript:window.guardian.readerRevenue.changeGeolocation()`

tomrf1: The best I can think of is to update the bookmarklet scripts to e.g.:
```
window.guardian.initialiseReaderRevenue()
  .then(() => window.guardian.readerRevenue.changeGeolocation())
```

gtrufitt: To save updating bookmarks could you change them all in this code to roughly
```
window.guardian.readerRevenue.changeGeolocation = () => import(/* webpackChunkName: "readerRevenueHelpers" */ 'readerRevenueHelpers').then(({ default: _ }) => {
	changeGeoLocation()	
   }).catch(error => 'An error occurred while loading the reader revenue helpers');
```

tomrf1: thanks for the tip @gtrufitt !
I've pushed a change which uses that strategy

PR #1729, opened 2020-07-16
Enable TCFv2

mxdvl:
## What does this change?

TCFv2 implemented in RotW, following the CCPA implementation from #1674.
The new [guardian CMP](https://github.com/guardian/consent-management-platform/) is bumped to version 4, and all methods have now different signatures.

- [x] Requires frontend 1% functional test to be completed
- [ ] Implement testing before it rolls out to 100% of DCR

### Before

image

### After

image

## Why?

It's the law.

PR #1797, opened 2020-08-07
Implement TCFv2

sndrs:
## What does this change?

- implements TCFv2 outside the US, behind a switch (https://github.com/guardian/frontend/pull/22870) – supersedes #1729
- deletes a test file that didn't contain meaningful tests (better to have no tests than tests that do nothing)
- deletes the story – it's 100% 3rd party now, no react so it's not really a meaningful test

### Before

image

### After

image

## Why?

> It's the law.


src/web/components/SignInGate/gateDesigns/SignInGateMain.tsx
```diff
@@ -250,7 +250,7 @@ export const SignInGateMain = ({
             </div>
             {showCpmUi && (
                 <Suspense fallback={<></>}>
-                    <ConsentManagementPlatform
+                    <oldCmp.ConsentManagementPlatform
```
sndrs, line 253: @guardian/dotcom-platform I'm not sure what to do about these – does this sign in gate component appear in place of `StickyBottomBanner.tsx`?

src/web/lib/getPrivacyFramework.ts
```diff
@@ -0,0 +1,23 @@
+import { getCountryCode } from '@frontend/web/lib/getCountryCode';
+
+const switches =
+    'guardian' in window &&
+    'config' in window.guardian &&
+    'switches' in window.guardian.config
+        ? window.guardian.config.switches
+        : {};
```
mxdvl, line 8: I've noticed that changes made the the `window.guardian.config.switches` object seem persistent across page views, when testing locally. Might be worth double-checking how the config is set/read in DCR?

src/web/components/SignInGate/gateDesigns/SignInGateMain.tsx
sndrs: so we do need to add support for v4 in here, confirmed

PR #1800, opened 2020-08-10
Disable rendering of AudioAtomBlockElement momentarily

shtukas:
## What does this change?

Disable rendering of AudioAtomBlockElement momentarily, while I update the backend model (and then the DCR model will have to be updated as well). Unfortunately we need to do this in several stages because no monorepo. (This should not be a problem because it was broken for months and I repaired it few days ago and I am just disabling it for a couple of hours)


shtukas:
Before:

Before

shtukas:
After:

After

PR #1805, opened 2020-08-11
Use tcfv2 in the signin gate

sndrs:
## What does this change?

supports TCFv2 via CMP@4 in the sign in gate:

- removes the check for whether the privacy message is showing
- shows TCFv2 privacy manager if TCFv2 is enabled

also:
- removes privacy manager deps from `SignInGatePatientia.tsx` since it's not being used

## Why?

the law

### DIFF
--- a/cypress/integration/e2e/sign-in-gate.spec.js
+++ b/cypress/integration/e2e/sign-in-gate.spec.js
@@ -13,20 +13,6 @@ describe('Sign In Gate Tests', () => {
         })
     }
 
-    const setCookieConsent = () => {
-        // the value for the consent cookies is an actual value added to the cookie when the cmp is dismissed as setting it to arbitrary values
-        // caused the following error "ConsentString - Unsupported version 45 in the string to decode"
-        cy.setCookie('euconsent', 'BO08eEOO08eEOBwABAENDOiAAAAwJ7_______9______9uz_Ov_v_f__33e8__9v_l_7_-___u_-23d4u_1vf99yfm1-7etr3tp_47ues2_Xur__71__3z3_9pxP78k89r7335Ew_v-_v-b7BCPN9Y3v-8K96g', {
-            log: true
-        });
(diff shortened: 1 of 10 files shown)


sndrs:
@coldlink the new CMP v4 API doesn't have a `showShow` method that returns a boolean, but instead has `willShowPrivacyMessage` which return a promise of a boolean. 

this affects your gate designs where they create their own `canShow` methods, which i've also updated but i'm not sure i've covered the use of them properly, because now they're also functions that return a promise of a boolean, not the boolean itself.

i think i need to do more but i'm not sure what?

coldlink:
> @coldlink the new CMP v4 API doesn't have a `showShow` method that returns a boolean, but instead has `willShowPrivacyMessage` which return a promise of a boolean.
> 
> this affects your gate designs where they create their own `canShow` methods, which i've also updated but i'm not sure i've covered the use of them properly, because now they're also functions that return a promise of a boolean, not the boolean itself.
> 
> i think i need to do more but i'm not sure what?

We'll need to refactor the place where the `canShow` method is called to deal with the promise now: https://github.com/guardian/dotcom-rendering/blob/4726845e286164c3d1f7cc8c8e890de414b5edce/src/web/components/SignInGate/SignInGateSelector.tsx#L195

I don't have time this afternoon, but will be able to look at this tomorrow if that's okay? Unless you can see an easy solution? 🤔

coldlink:
Running the Sign In Gate Cypress tests would the best/easiest way to check things are working properly with the gate.

sndrs:
I've updated the `SignInGateSelector` to hopefully handle the newly async functions. but there's an issue in the e2e test that i think is related to expecting TCFv1 cookies, here: https://github.com/guardian/dotcom-rendering/blob/8bc9bb76e92fb4d4ff285fe116ca7521ca75f2a5/cypress/integration/e2e/sign-in-gate.spec.js#L16-L28

it _appears_ to be working as expected running it locally and forcing TCFv2, so this looks to be a test config thing 🙏

PR #1806, opened 2020-08-11
Guide Atom

nitro-marky:
## What does this change?
This adds the guide atom to DCR via the latest version of atoms-rendering. As per the Q and A atom, it uses the Ophan event tracking functions.

Screenshot 2020-08-10 at 12 10 55

Screenshot 2020-08-05 at 16 24 21

Screenshot 2020-08-05 at 16 24 07

Screenshot 2020-08-05 at 16 25 36

### DIFF
--- a/cypress/integration/e2e/atom.interactivity.spec.js
+++ b/cypress/integration/e2e/atom.interactivity.spec.js
@@ -4,6 +4,9 @@
 const qandaUrl =
     'https://www.theguardian.com/technology/2018/sep/19/time-to-regulate-bitcoin-says-treasury-committee-report';
 
+const guideUrl =
+    'https://www.theguardian.com/environment/2020/aug/01/plan-to-curb-englands-most-polluted-spot-divides-residents';
+
 describe('Atom Interactivity', function () {
     describe('Q&A Atom', function () {
         it('should render', function () {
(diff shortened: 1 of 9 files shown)


cypress/integration/e2e/atom.interactivity.spec.js
```diff
@@ -61,4 +64,37 @@ describe('Atom Interactivity', function () {
...
+            cy.get('[data-testid="like"]').click();
+            cy.get('[data-testid="feedback"]').should('be.visible');
+        });
+
+        it('should show feedback message when dislike is clicked', function () {
+            cy.visit(`/Article?url=${guideUrl}`);
+            cy.get('[data-snippet-type="guide"]').click();
+            cy.get('[data-testid="dislike"]').click();
+            cy.get('[data-testid="feedback"]').should('be.visible');
+        });
```
gtrufitt, line 98: I think as this is generally copied, you could abstract into a function and pass in the `type` so you'd have a function:

```
const atomTests = (type, url) = {
... all your tests
it('should show feedback message when dislike is clicked', function () {
            cy.visit(`/Article?url=${url}`);
            cy.get(`[data-snippet-type="${type}"]`).click(); // I think this will work!?
            cy.get('[data-testid="dislike"]').click();
            cy.get('[data-testid="feedback"]').should('be.visible');
        });
...
}
```

and call them like 

```
atomTests('qanda', qandaUrl);
atomTests('guide', guideUrl);
```

That way, adding and amending these similar types should be quick (and you can do the other atom type easily)

src/web/components/App.tsx
```diff
@@ -351,6 +351,54 @@ export const App = ({ CAPI, NAV }: Props) => {
...
+                                component: {
+                                    componentType: 'GUIDE_ATOM',
+                                    id: guideAtom.id,
+                                    labels: [],
+                                    products: [],
+                                },
+                                action: 'LIKE',
+                            };
+                            submitComponentEvent(componentEvent);
+                        }}
```
gtrufitt, line 375: Maybe this is another good opportunity to make this clearer and easier to reuse?

```
// Function that returns a function
const componentEventHandler = (componentType, id, action) => () => {
                            const componentEvent: OphanComponentEvent = {
                                component: {
                                    componentType,
                                    id,
                                    labels: [],
                                    products: [],
                                },
                                action,
                            };
                            submitComponentEvent(componentEvent);
                        }

```
```suggestion
	likeHandler={componentEventHandler('GUIDE_ATOM', guideAtom.id, 'LIKE')}
```

and you can reuse for this and other atoms:

`dislikeHandle={componentEventHandler('GUIDE_ATOM', guideAtom.id, 'DISLIKE')}`

What do you reckon?

nitro-marky: Yeah, that looks like a plan

PR #1710, opened 2020-07-07
AMP⚡TCFv2 CMP w/ Sourcepoint

mxdvl:
## What does this change?

AMP TCFv2 consent is now handled by SourcePoint. CCPA follows in https://github.com/guardian/dotcom-rendering/pull/1809

### Before

image

### After

image

## Why?

TCFv2 is a requirement and it is handled by SourcePoint on the web platform (non-AMP).

PR #1809, opened 2020-08-12
AMP⚡CCPA CMP w/ Sourcepoint

mxdvl:
## What does this change?

Follow-up on #1710 to implement CCPA

### Before

image

### After

image

## Why?

We must comply with the CCPA framework.


sndrs (approved):
image 👍

PR #1810, opened 2020-08-12
Correct SpotifyBlockElement's type

shtukas:
## What does this change?

Correct SpotifyBlockElement's type. To align it with the backend type: https://github.com/guardian/frontend/blob/0e6b2252b80d555961315278f47307edc15789ba/common/app/model/dotcomrendering/pageElements/PageElement.scala#L111

### DIFF
--- a/src/lib/content.d.ts
+++ b/src/lib/content.d.ts
@@ -230,10 +230,11 @@ interface SoundcloudBlockElement {
 
 interface SpotifyBlockElement {
     _type: 'model.dotcomrendering.pageElements.SpotifyBlockElement';
-    embedUrl: string;
-    title: string;
-    height: number;
-    width: number;
+    embedUrl?: string;
+    title?: string;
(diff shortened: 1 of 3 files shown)


gtrufitt:
Do you not need to handle the optional-ness in the components now then?

src/lib/content.d.ts
```diff
@@ -230,10 +230,11 @@ interface SoundcloudBlockElement {
...
     _type: 'model.dotcomrendering.pageElements.SpotifyBlockElement';
-    embedUrl: string;
-    title: string;
-    height: number;
-    width: number;
+    embedUrl?: string;
+    title?: string;
+    height?: number;
+    width?: number;
+    caption?: string;
```
rcrphillips, line 237: The component doesn't need the caption field

shtukas:
@gtrufitt Yes.

PR #1817, opened 2020-08-12
Send browserId in pubData to SourcePoint via CMP

janua:
## What does this change?

This adds adding the `browserId` to the `pubData` object that can be passed to the `init` of the CMP.

## Why?

We would like to report on the consent each pageview has; in order to do this we would like to associate the current `browserId` with a consent signal from the CMP.

@guardian/commercial-dev

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
         "@guardian/ab-react": "1.0.0-next.0",
         "@guardian/atoms-rendering": "^1.5.0",
         "@guardian/automat-client": "^0.2.16",
-        "@guardian/consent-management-platform": "4.0.0-12",
+        "@guardian/consent-management-platform": "4.0.0-13",
         "@guardian/discussion-rendering": "^2.6.0",
         "@guardian/shimport": "^1.0.2",
         "@guardian/src-button": "^1.5.0",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
         "@guardian/ab-react": "1.0.0-next.0",
         "@guardian/atoms-rendering": "^1.5.0",
         "@guardian/automat-client": "^0.2.16",
-        "@guardian/consent-management-platform": "4.0.0-13",
+        "@guardian/consent-management-platform": "4.0.0-14",
         "@guardian/discussion-rendering": "^2.6.0",
         "@guardian/shimport": "^1.0.2",
         "@guardian/src-button": "^1.5.0",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
         "@guardian/ab-react": "1.0.0-next.0",
         "@guardian/atoms-rendering": "^1.5.0",
         "@guardian/automat-client": "^0.2.16",
-        "@guardian/consent-management-platform": "4.0.0-10",
+        "@guardian/consent-management-platform": "4.0.0-11",
         "@guardian/discussion-rendering": "^2.6.0",
         "@guardian/shimport": "^1.0.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
         "@guardian/ab-react": "1.0.0-next.0",
         "@guardian/atoms-rendering": "^1.5.0",
         "@guardian/automat-client": "^0.2.16",
-        "@guardian/consent-management-platform": "4.0.0-9",
+        "@guardian/consent-management-platform": "4.0.0-13",
```
janua, line 40: This is currently a pre-release; we will switch to a major release when that it put out.

src/web/components/StickyBottomBanner/CMP.tsx
```diff
@@ -13,7 +14,12 @@ export const willShowCMP = async () => {
     }
 
     const isInUsa = (await getCountryCode()) === 'US';
+    const browserId: string | undefined = getCookie('bwid') || undefined;
```
janua, line 17: Not sure if this is the best method to retreive the `browserId`; could someone from DCR advise?

jfsoul: I think this should be available at `window.guardian.config.ophan.browserId`?

gtrufitt: It's currently not set in DCR, it looks like the sign-in-gate uses the same mechanism https://github.com/guardian/dotcom-rendering/blob/35c9a86b4e6967815512a56f56b7d10085c5abe3/src/web/components/SignInGate/SignInGateSelector.tsx#L103 but looks like we do run `window.guardian.config.ophan.browserId = getCookieValue("bwid");` in `htmlTemplate.ts` so could use that?

janua: @jfsoul It probably is available at the location you give. However, I took it being acceptable from here;

https://github.com/guardian/dotcom-rendering/blob/master/src/web/components/SignInGate/SignInGateSelector.tsx#L103

The line before is uses `window.guardian.ophan.viewId`, but for some reason does not do the same for browserId.

gtrufitt: `window.guardian.config.ophan.browserId` is from the same `bwid` cookie (note, *not* viewId)

janua: Actually, it doesn't even seem to be defined here;

https://github.com/guardian/dotcom-rendering/blob/master/window.guardian.d.ts#L19-L28

jfsoul: I suspect the `SignInGateSelector` could do with updating to use the same window property.

janua: @gtrufitt @jfsoul What about the types? It's not represented in either `window.guardian.ophan` or `window.guardian.config.ophan`.

https://github.com/guardian/dotcom-rendering/blob/d2f957fcb3eea6d3df7bb8f9c4fc2373e5e4dc9d/src/model/window-guardian.ts#L28

gtrufitt: They're all a bit of a mess at the moment and need fixing, but I'd suggest adding it to config.ophan

gtrufitt: We'll look at this on Dotcom as the types need correcting

PR #1819, opened 2020-08-13
Add Sign In Gate Dismiss Window test

vlbee:
## What does this change?

- Adds a new AB(C) test for different dismiss windows on the Sign In Gate before reshowing gate to users. 
- Updates the README
- Adds the option to add a dismissal window (in hours) to the `hadUserDismissedGate` function
- Adds new `unsetUserDismissedGate` function
- Fixes how the sign in gate placeholder div is inserted into the Article html, as before it would only work if the second article element was a text block. 

Uses same gate UI as the current main variant

### TODO

Cypress integration tests

## Why?

Because

### DIFF
--- a/cypress/integration/e2e/sign-in-gate.spec.js
+++ b/cypress/integration/e2e/sign-in-gate.spec.js
@@ -1,56 +1,79 @@
 /* eslint-disable no-undef */
 /* eslint-disable func-names */
 
-describe('Sign In Gate Tests', function() {
+describe('Sign In Gate Tests', function () {
     const setArticleCount = (n) => {
         // set article count for today to be n
-        localStorage.setItem('gu.history.dailyArticleCount', JSON.stringify({ value: [{ day: Math.floor(Date.now() / 86400000), count: n }] }));
-    }
+        localStorage.setItem(
+            'gu.history.dailyArticleCount',
+            JSON.stringify({
+                value: [
+                    {
+                        day: Math.floor(Date.now() / 86400000),
+                        count: n,
+                    },
+                ],
+            }),
+        );
+    };
 
     const setMvtCookie = (str) => {
         cy.setCookie('GU_mvt_id_local', str, {
-            log: true
-        })
-    }
+            log: true,
+        });
+    };
 
     // helper method over the cypress visit method to avoid having to repeat the same url by setting a default
     // can override the parameter if required
-    const visitArticle = (url = 'https://www.theguardian.com/games/2018/aug/23/nier-automata-yoko-taro-interview') => {
-        cy.visit(
-            `Article?url=${url}`,
-        );
-    }
+    const visitArticle = (
+        url = 'https://www.theguardian.com/games/2018/aug/23/nier-automata-yoko-taro-interview',
+    ) => {
+        cy.visit(`Article?url=${url}`);
+    };
 
     // as the sign in gate is lazy loaded, we need to scroll to the rough position where it
     // will be inserted to make it visible
     // can override position if required
     const scrollToGateForLazyLoading = (roughPosition = 1000) => {
         cy.scrollTo(0, roughPosition, { duration: 500 });
-    }
+    };
 
     // we call visit and scroll for most test, so this wrapper combines the two
     // while preserving the ability to set the parameters if required
-    const visitArticleAndScrollToGateForLazyLoad = ({ url, roughPosition } = {}) => {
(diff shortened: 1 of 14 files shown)


cypress/integration/e2e/sign-in-gate.spec.js
```diff
@@ -1,109 +1,156 @@
...
+    } = {}) => {
         visitArticle(url);
         scrollToGateForLazyLoading(roughPosition);
-    }
+    };
+
+    // prevents the CMP banner from showing which creates a race condition that prevents tests from passing
+    // use before each test
+    const setConsentCookieToBlockCMPBanner = () => {
+        cy.setCookie('consentUUID', '9c1d1bdd-3322-416a-ad9f-33773423e17c', {
```
gtrufitt, line 54: This is now set before every test so you probably don't need it here. See https://github.com/guardian/dotcom-rendering/pull/1829

cypress/integration/e2e/sign-in-gate.spec.js
```diff
@@ -113,52 +160,333 @@ describe('Sign In Gate Tests', function() {
...
         });
 
-        it('should show cmp ui when privacy settings link is clicked', function() {
+        it('should show cmp ui when privacy settings link is clicked', function () {
             visitArticleAndScrollToGateForLazyLoad();
 
             cy.get('[data-cy=sign-in-gate-main]').should('be.visible');
 
             cy.get('[data-cy=sign-in-gate-main_privacy]').click();
+            getCMPiFrame().contains('privacy settings').should('be.visible');
```
gtrufitt, line 189: I think I may have fixed this in https://github.com/guardian/dotcom-rendering/pull/1829/files#diff-ae97a2c8603811d3bc137eddacb4a4f1L142 but I didn't seem to have to do the getCMPIframe bit so not sure why!

cypress/integration/e2e/sign-in-gate.spec.js
vlbee: ok - thanks!

src/web/components/SignInGate/gateDesigns/SignInGateDismissWindow.tsx
```diff
@@ -0,0 +1,268 @@
+import React, { useState, Suspense } from 'react';
```
coldlink, line 1: Do we need a whole separate gate design if it's the same as the previous design, and just re use that?

coldlink: in cypress we can reuse the same `data-cy` selectors that main uses, and in the files we set up the variants we can just the main design too

coldlink: as that's all i can see thats different?

vlbee: Yeah, I did this originally then decided to just dupe the design but can switch back to using main if you prefer.

coldlink: Yeah I think it's better we reuse as much as possible without duping.
I'm thinking instead of `SignInGateMainDesign` we should call is `SignInGateViiDesign` to match the name in the sheet?

vlbee: done. you mean rename the design file back to vii-variant?

coldlink: Yeah that makes sense. Thanks.

vlbee:
Bug fix: the span for the gate placeholder will now be added to any second element type. Unfortuantley this can have unpredictable results, eg, the first two paragraphs are created with <br> tags in this article, but at lease the gate renders:


image

src/web/lib/withSignInGateSlot.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+// This span is used to insert the sign in gate into the appropirate location within body of an article,
+// if the SignInGateSector determines a gate should be rendered.
+
+const SignInGateSlot = <span id="sign-in-gate" />;
+
+export const withSignInGateSlot = (
```
gtrufitt, line 8: Looks ok to me, just pinging @nicl for awareness of the method

vlbee: thanks!

PR #1820, opened 2020-08-13
Match tabs

rcrphillips:
## What does this change?
Adds the match tabs component to football match report pages, when a minute by minute is present. These are also some subtle design changes. The fake 'tab' look needed updating. I think we still have an accessibility issue with using #0084C6 @HarryFischer do you think you could take a look?

Also, removing the 'stats' tab on mobile. The stats are displayed at the bottom of the article anyway. 

Ordering of components is addressed in this PR - https://github.com/guardian/dotcom-rendering/pull/1836

## DCR - Desktop
Screen Shot 2020-08-21 at 11 38 04

## Frontend - desktop
Screen Shot 2020-08-21 at 11 38 47

## DCR - Mobile
Screen Shot 2020-08-21 at 11 39 41

## Frontend - Mobile
Screen Shot 2020-08-21 at 11 39 58


shtukas:
Backend work completed: https://github.com/guardian/frontend/pull/22922

shtukas:
👍

shtukas:
Screenshot 2020-08-20 at 14 40 55

src/web/components/MatchNav.tsx
```diff
@@ -204,23 +211,101 @@ const Border = () => (
...
+            font-weight: 600;
+            min-height: 36px;
+            display: block;
+            width: 100%;
+        }
+        a span:hover {
+            background-color: ${palette.neutral[93]};
+        }
+
+        a span {
```
gtrufitt, line 258: This is a little brittle as you are overriding spans based on the parent without context of the previous styling.

It would be more idomatic of DCR styling to add `className`'s for 'tab', `tabLabel` and `tabAnchor` and style off of these. You can then also style within `tabLabel` like 

```
// tabLabel is the `span`
tabLabel = css`
& a {
	// styles that only apply inside an anchor
}

& a:hover {
	// styles that only apply inside a hovered anchor
}
`
```

this will make it easier to reason with the code and contextualise the styling to the parent

rcrphillips: @gtrufitt could you take another look? Happy to go through it on a call if needs be.

gtrufitt (approved):
Looks great!

One query, I'm not seeing them in the Chromatic screenshot?

image

rcrphillips:
[Match tabs](https://trello.com/c/WH7pzo9f/1828-match-tabs)

PR #1821, opened 2020-08-13
Fix byline image when viewing full size screen

nitro-marky:
## What does this change?
If there is more than one contributor it defaults the byline width to 100%, if there is only 1 it limits it to 68% to accommodate an image. 

This is a temporary fix.

### Before

Screenshot 2020-08-13 at 12 10 20

### After

Screenshot 2020-08-13 at 12 05 44

Screenshot 2020-08-13 at 12 05 53

### DIFF
--- a/src/web/components/HeadlineByline.tsx
+++ b/src/web/components/HeadlineByline.tsx
@@ -78,10 +78,8 @@ const immersiveLinkStyles = (pillar: Pillar) => css`
 `;
 
 // If there is an image reduce the width of the author div
-const mobileCommentAuthor = css`
-    @media (max-width: 400px) {
-        width: 65%;
-    }
+const authorBylineWithImage = css`
+    width: 68%;
 `;
 
 type Props = {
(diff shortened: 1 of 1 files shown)


src/web/components/HeadlineByline.tsx
```diff
@@ -152,8 +150,9 @@ export const HeadlineByline = ({
                     return (
                         <div
                             className={`${opinionStyles(pillar)} ${
-                                tags.find((tag) => tag.bylineImageUrl)
-                                    ? mobileCommentAuthor
+                                tags.filter((tag) => tag.type === 'Contributor')
+                                    .length === 1
+                                    ? authorBylineWithImage
                                     : ''
```
gtrufitt, line 156: I know this is merged but when we look at logic wider, it would be better to assign this to a meaningful class and use the cx helper:

`const hasSingleContributor = tags.filter((tag) => tag.type === 'Contributor').length === 1`

`cx({ [authorBylineWithImage]: hasSingleContributor }, opinionStyles(pillar))`

and then extend that with the other checks like: 

`cx({ [authorBylineWithImage]: hasSingleContributor && authorNameMatchesBlah && thisOtherThing}, opinionStyles(pillar))`

gtrufitt:
Would be great to add a story for this

PR #1826, opened 2020-08-13
Introduce Banner Picker to DCR

tjmw:
## What does this change?

This introduces banner picker logic similar to frontend on DCR. We provide a list of banners in priority order and concurrently ask each whether they can show in the current context. The highest priority banner which answers "yes" gets displayed.

It's also possible to give each banner a timeout, which they must respond within otherwise they lose their opportunity. This can be configured per banner, if we want to give some more leeway than others (I noticed that right now we'll wait indefinitely for the CMP to decide, before moving on to the next priority, and didn't want to break that).

Screenshot 2020-08-13 at 16 18 19

Screenshot 2020-08-13 at 16 18 29

### Before

The existing setup of either showing the CMP, or otherwise falling back on the Reader Revenue banner service (or nothing) is fairly baked in, which makes it harder to introduce new banners.

### After

A clearer path for adding new banners with their own `canShow` and rendering functionality.

## Why?

We're planning on using Braze to display messages to users on the web, using the banner as the format of the first one. So we need to be able to extend the current banner selection logic to integrate the new banner.

## Notes

* ~~I've not yet split apart the true canShow logic for the RR banner and the show part. At the moment the canShow just considers the switch value, but really it should take into account the first (data) call to the contributions service, which is the real canShow logic. Right now that happens downstream of show. This only works because the RR banner is bottom of the priority list, it wouldn't if there were other candidates after it.~~

* ~~A more fundamental issue that I ran into was how to handle dependencies for the canShow logic. Two props, `countryCode` and `isSignedIn` are needed for the canShow logic of the banner. These aren't guaranteed to be set by the time we first render `StickyBottomBanner`. For now I've made them both dependencies for the effect in which we kick off the banner selection logic. But that doesn't seem reasonable since we're potentially holding up the CMP (and any other future banners) until those props are set, even though it doesn't care about them.~~ We worked around this by creating a `Promise` which wraps the country code so we pass that down to the `ReaderRevenueBanner` and it no longer blocks the banner picking code. I left `isSignedIn` as is. Technically that does still block the banner picking code but it seems so fast that it's not a problem, and not worth introducing extra complexity to work around. I measured like this, and in consistently reports 0 on my laptop:

```js
    useEffect(() => {
        const startTime = new Date().getTime();
        const guCookie = getCookie('GU_U');
        const endTime = new Date().getTime();
        console.log('Reading GU_U took', endTime - startTime);
        setIsSignedIn(!!guCookie);
    }, []);
```

That'll be slower on on older devices, but presumably not hugely?

* ~~It was suggested by Gareth that we might want to have timeouts be opt-out rather than opt-in which sounds like a good suggestion. So for the CMP if that shouldn't have a timeout we explicitly configure that, but others get a default (presumably of 2 seconds, like frontend).~~

* Previously we always waited to get a response from the CMP before asking the contributions service about the reader revenue banner. That's now changed, and we'll kick off all `canShow` requests concurrently (like frontend). This means that there are scenarios where we send the request to the contributions service but don't end up using it because the CMP says yes. I need to check this with the RR team (since it will result in additional requests to the service).

### DIFF
--- a/src/web/browser/ophan/ophan.ts
+++ b/src/web/browser/ophan/ophan.ts
@@ -84,7 +84,7 @@ export type TestMeta = {
     products?: OphanProduct[];
 };
 
-const record = (event: {}): void => {
+export const record = (event: {}): void => {
     if (
         window.guardian &&
         window.guardian.ophan &&
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -134,6 +134,13 @@ export const App = ({ CAPI, NAV }: Props) => {
     const [isSignedIn, setIsSignedIn] = useState<boolean>();
     const [user, setUser] = useState<UserProfile>();
     const [countryCode, setCountryCode] = useState<string>();
+    // This is an async version of the countryCode state value defined above.
+    // This can be used where you've got logic which depends on countryCode but
+    // don't want to block on it becoming available, as you would with the
+    // non-async version (this is the case in the banner picker where some
+    // banners need countryCode but we don't want to block all banners from
+    // executing their canShow logic until countryCode is available):
+    const [asyncCountryCode, setAsyncCountryCode] = useState<Promise<string>>();
     const [commentCount, setCommentCount] = useState<number>(0);
     const [isClosedForComments, setIsClosedForComments] = useState<boolean>(
         true,
@@ -176,8 +183,11 @@ export const App = ({ CAPI, NAV }: Props) => {
     }, [isSignedIn, CAPI.config.discussionApiUrl]);
 
     useEffect(() => {
-        const callFetch = async () =>
-            setCountryCode((await getCountryCode()) || '');
+        const callFetch = () => {
+            const countryCodePromise = getCountryCode();
+            setAsyncCountryCode(countryCodePromise);
+            countryCodePromise.then((cc) => setCountryCode(cc || ''));
+        };
         callFetch();
     }, []);
 
@@ -605,7 +615,7 @@ export const App = ({ CAPI, NAV }: Props) => {
             <Portal root="bottom-banner">
                 <StickyBottomBanner
                     isSignedIn={isSignedIn}
-                    countryCode={countryCode}
+                    asyncCountryCode={asyncCountryCode}
                     CAPI={CAPI}
                 />
             </Portal>
--- a/src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
+++ b/src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
@@ -4,12 +4,16 @@ import {
     shouldShowOldCMP,
     CMP,
 } from '@root/src/web/components/StickyBottomBanner/CMP';
-import { ReaderRevenueBanner } from '@root/src/web/components/StickyBottomBanner/ReaderRevenueBanner';
+import {
+    canShow as canShowRRBanner,
+    ReaderRevenueBanner,
+} from '@root/src/web/components/StickyBottomBanner/ReaderRevenueBanner';
 import { getAlreadyVisitedCount } from '@root/src/web/lib/alreadyVisited';
+import { pickBanner, BannerConfig, MaybeFC, Banner } from './bannerPicker';
 
 type Props = {
     isSignedIn?: boolean;
-    countryCode?: string;
+    asyncCountryCode?: Promise<string>;
     CAPI: CAPIBrowserType;
 };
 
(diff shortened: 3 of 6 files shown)


gtrufitt:
Cypress tests are now passing in main branch, please rebase, thanks! https://github.com/guardian/dotcom-rendering/pull/1829

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -20,61 +24,75 @@ const getBannerLastClosedAt = (key: string): string | undefined => {
...
+                    subscriptionBannerLastClosedAt: getBannerLastClosedAt(
+                        'subscriptionBannerLastClosedAt',
+                    ),
+                    switches: {
+                        remoteSubscriptionsBanner: !!CAPI.config
+                            .remoteSubscriptionsBanner,
+                    },
+                }),
+            /* eslint-disable-next-line react/jsx-props-no-spreading */
+            show: (meta: any) => () => <ReaderRevenueBanner {...meta} />,
```
tjmw, line 83: Note that by omitting the optional `timeoutMillis` parameter here we're automatically opting in the reader revenue banner to the 2 second timeout rule, which may not be what we want.

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -18,63 +22,96 @@ const getBannerLastClosedAt = (key: string): string | undefined => {
...
+                subscriptionBannerLastClosedAt: getBannerLastClosedAt(
+                    'subscriptionBannerLastClosedAt',
+                ),
+                switches: {
+                    remoteSubscriptionsBanner: !!CAPI.config
+                        .remoteSubscriptionsBanner,
+                },
+            }),
+        /* eslint-disable-next-line react/jsx-props-no-spreading */
+        show: (meta: any) => () => <ReaderRevenueBanner {...meta} />,
```
tjmw, line 81: Note that by omitting the optional `timeoutMillis` parameter here we're automatically opting in the reader revenue banner to the 2 second timeout rule, which may not be what we want. We can opt out of this by specifying `timeoutMillis: null`.

src/web/browser/ophan/ophan.ts
```diff
@@ -84,7 +84,7 @@ export type TestMeta = {
     products?: OphanProduct[];
 };
 
-const record = (event: {}): void => {
+export const record = (event: {}): void => {
```
tjmw, line 87: I had to export this as I wanted to call `record` directly instead of one of the abstractions built on top of it. This matches [what we do on frontend](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/common/modules/ui/bannerPicker.js#L80).

tjmw:
Hi @tomrf1 this is the PR for the banner picker changes I mentioned on Chat. I'd be interested in any feedback you have if you've got time, especially since this does touch the reader revenue banner code.

tjmw:
Hi @sndrs I've been working on this PR to make DCR's banner picking logic closer to frontend (with the aim of adding another banner in future). I'd be interested in your feedback if you've got a moment, since this touches code around the CMP.

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
tomrf1: You're sending timeout events to ophan so I'm ok with this (it behaves the same as in frontend). We can monitor timeouts and reassess later

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -18,63 +22,96 @@ const getBannerLastClosedAt = (key: string): string | undefined => {
     return (item && JSON.parse(item).value) || undefined;
 };
 
+const buildNewCmpBannerConfig = () => {
```
tomrf1, line 25: might be helpful to specify the type here
`(): Banner => ...`

src/web/components/StickyBottomBanner/bannerPicker.ts
```diff
@@ -0,0 +1,102 @@
...
+        value: bannerId,
+    });
+
+const timeoutify = (banner: Banner): BannerWithTimeout => {
+    let timer: number | undefined;
+
+    const canShow = (): Promise<CanShowResult> =>
+        new Promise((resolve) => {
+            const timeout =
+                typeof banner.timeoutMillis === 'undefined'
```
tomrf1, line 38: Having such different meanings for null and undefined worries me a bit, is this a common pattern?
I think I'd prefer making `timeoutMillis` a mandatory field:
`timeoutMillis: number | null`

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
tjmw: Great idea. Resolved in 6a6c7fd0fac3daec8b0184899dcb843585a28165.

src/web/components/StickyBottomBanner/bannerPicker.ts
tjmw: Yeah that's a really good point, I can see that this is a little confusing now. My intention was around the difference between omitting the property (which gives you the default 2s timeout):

```js
{
  id: 'some-banner',
  // etc
}
```

and including it but specifying `null` (which gives no timeout):

```js
{
  id: 'some-banner',
  // etc
  timeoutMillis: null,
}
```

But your suggestion of making it a mandatory field sounds sensible. I think the reasoning behind having a default was that when adding a new banner you wouldn't have to think about the timeout too much. But realistically new banners aren't added that often, and you probable _do_ need to think about the timeout anyway. Perhaps making it mandatory but defining a const `DEFAULT_TIMEOUT = 2000` would keep the intent clear.

tomrf1: Makes sense - the developer has to consider whether or not they want a timeout

tjmw: I've made `timeoutMillis` mandatory in this commit: d4b8e269aa729d633d9bc43177ccc5b9182cf21a.

tomrf1 (approved):
great!

tjmw:
👋 Hey @gtrufitt @jfsoul this is the banner picker work we talked about before. Just wanted to give a little bump in case you get time for a review. Cheers!

src/web/components/StickyBottomBanner/bannerPicker.test.tsx
```diff
@@ -0,0 +1,195 @@
...
+                id: 'banner-3',
+                canShow: () => Promise.resolve({ result: true }),
+                show: () => MockComponent,
+                timeoutMillis: null,
+            },
+        ];
+
+        const got = await pickBanner(config);
+
+        expect(got()).toEqual(ChosenMockComponent);
```
jfsoul, line 48: What does toEqual do with a component? Presumably checks by reference rather than by value!

jfsoul (approved):
👍  apart from comments

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -18,63 +22,99 @@ const getBannerLastClosedAt = (key: string): string | undefined => {
...
+        id: 'cmpUi',
+        canShow: () => willShowCMP().then((result) => ({ result: !!result })),
+        // New CMP is not a react component and is shown outside of react's world
+        // so render nothing if it will show
+        show: () => null,
+        timeoutMillis: null,
+    };
+};
+
+const buildOldCmpBannerConfig = (CAPI: CAPIBrowserType): Banner => {
```
sndrs, line 38: the old cmp will be going away as soon as we can get it working properly (working on it now) – if it helps you might want to hang on till then. no worries if not, we can remove it from instead once merged if that;s better for you

sndrs: but will be removed from this file entirely (I think)

src/web/components/StickyBottomBanner/bannerPicker.test.tsx
tjmw: Hmm good question, yes I think that's correct. I guess the thing to note is that it's the component definition (e.g. `const ChosenMockComponent = () => { ... }` in the real world) that we're comparing here not a rendering of the component.

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
tjmw: Hey thanks @sndrs :+1: yeah I figured this would probably go away at some point. As long as you're happy with it I'd vote to keep it around in this new form until we're ready to remove it (i.e. to not hold off on merging this PR).

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -18,63 +22,99 @@ const getBannerLastClosedAt = (key: string): string | undefined => {
...
-            setShowOldCMP(shouldShowOld && CAPI.config.cmpUi),
-        );
-        willShowCMP().then(setCMPWillShow);
-    }, [CAPI.config.cmpUi]);
-
-    // Don't render anything until we know whether we can show the CMP
-    if (showOldCMP === null || CMPWillShow === null) {
-        return null;
-    }
+        if (isSignedIn === undefined) {
```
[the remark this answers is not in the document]
tjmw: That's a good suggestion, I've made this change in 0d16ec1.

src/web/components/App.tsx
```diff
@@ -176,8 +177,11 @@ export const App = ({ CAPI, NAV }: Props) => {
     }, [isSignedIn, CAPI.config.discussionApiUrl]);
 
     useEffect(() => {
-        const callFetch = async () =>
-            setCountryCode((await getCountryCode()) || '');
```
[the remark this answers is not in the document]
tjmw: Good idea, comment added in 2a97a73.

PR #1837, opened 2020-08-19
Meta Timestamp

nitro-marky:
## What does this change?
Adds the published timestamp.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -172,6 +172,7 @@ interface Block {
     title?: string;
     firstPublished?: number;
     firstPublishedDisplay?: string;
+    secondaryDateLine?: string;
 }
 
 interface Pagination {
--- a/src/web/components/ArticleMeta.tsx
+++ b/src/web/components/ArticleMeta.tsx
@@ -20,6 +20,7 @@ type Props = {
     author: AuthorType;
     tags: TagType[];
     webPublicationDateDisplay: string;
+    secondaryDateline?: string;
     branding?: Branding;
 };
 
@@ -262,6 +263,7 @@ export const ArticleMeta = ({
     author,
     tags,
     webPublicationDateDisplay,
+    secondaryDateline,
 }: Props) => {
     const sharingUrls = getSharingUrls(pageId, webTitle);
     const bylineImageUrl = getBylineImageUrl(tags);
@@ -297,8 +299,9 @@ export const ArticleMeta = ({
                                 />
                             )}
                             <Dateline
-                                dateDisplay={webPublicationDateDisplay}
+                                primaryDateline={webPublicationDateDisplay}
                                 descriptionText="Published on"
+                                secondaryDateline={secondaryDateline}
                             />
                         </div>
                     </>
--- a/src/web/components/Dateline.tsx
+++ b/src/web/components/Dateline.tsx
@@ -34,6 +34,21 @@ const toggleClass = css`
         opacity: 0;
         display: none;
     }
+
+    input[type='checkbox']:checked ~ span ~ p {
+        max-height: 80px;
+    }
+`;
+
+const pCheckboxStyle = css`
+    p {
+        max-height: 0;
+        overflow: hidden;
+        transition: max-height 0.4s ease;
+    }
+`;
+
+const labelStyles = css`
     label {
         cursor: pointer;
     }
@@ -52,15 +67,6 @@ const toggleClass = css`
         top: 50%;
         left: 10px;
     }
-
-    p {
-        max-height: 0;
-        overflow: hidden;
-        transition: max-height 0.4s ease;
-    }
-    input[type='checkbox']:checked ~ span ~ p {
-        max-height: 80px;
-    }
 `;
 
 // At the moment there is no way to tell from primaryDateline and secondaryDatline
@@ -73,7 +79,14 @@ export const Dateline: React.FC<{
     return (
         <div className={dateline}>
             {primaryDateline !== secondaryDateline ? (
-                <div className={cx(toggleClass, dateline)}>
+                <div
+                    className={cx(
+                        toggleClass,
+                        pCheckboxStyle,
+                        labelStyles,
+                        dateline,
+                    )}
+                >
                     <input type="checkbox" id="dateToggle" />
                     <span>
                         <label htmlFor="dateToggle">{primaryDateline}</label>
--- a/src/web/components/Dateline.tsx
+++ b/src/web/components/Dateline.tsx
@@ -1,9 +1,9 @@
 import React from 'react';
 
-import { css } from 'emotion';
+import { css, cx } from 'emotion';
 import { text } from '@guardian/src-foundations/palette';
 import { textSans } from '@guardian/src-foundations/typography';
-import { visuallyHidden } from '@guardian/src-foundations/accessibility';
+import { from } from '@guardian/src-foundations/mq';
 
 const captionFont = css`
     ${textSans.xsmall()};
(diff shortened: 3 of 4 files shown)


shtukas (approved):

src/web/components/Dateline.tsx
```diff
@@ -17,18 +17,89 @@ const dateline = css`
...
+// At the moment there is no way to tell from primaryDateline and secondaryDatline
+// what the prepend should be (either 'Last modified on' or 'First published on ')
 export const Dateline: React.FC<{
-    dateDisplay: string;
+    primaryDateline: string;
     descriptionText: string;
-}> = ({ dateDisplay, descriptionText }) => (
-    <>
-        <div className={dateline} role="textbox">
-            <span className={description}>{descriptionText} </span>
```
gtrufitt, line 30: What happened to the descriptionText here?

src/web/components/Dateline.tsx
```diff
@@ -17,18 +17,89 @@ const dateline = css`
...
+                        pCheckboxStyle,
+                        labelStyles,
+                        dateline,
+                    )}
+                >
+                    <input type="checkbox" id="dateToggle" />
+                    <span>
+                        <label htmlFor="dateToggle">{primaryDateline}</label>
+                    </span>
+                    {secondaryDateline && (
```
gtrufitt, line 94: I think we would want this conditional check higher up, as we don't want the checkbox etc when we don't have a secondaryDateLine to show

gtrufitt: I think as it's currently *not* optional, you can remove the optionality in the props though and rely on the `!==` check for primary and secondary date lines.

src/web/components/Dateline.tsx
```diff
@@ -17,18 +17,89 @@ const dateline = css`
...
+`;
+
+const labelStyles = css`
+    label {
+        cursor: pointer;
+    }
+    label {
+        position: relative;
+        display: block;
+    }
```
gtrufitt, line 58:
```suggestion
    label {
	    cursor: pointer;
        position: relative;
        display: block;
    }
```

src/web/components/Dateline.tsx
```diff
@@ -17,18 +17,89 @@ const dateline = css`
...
+    label::before {
+        content: '';
+        position: absolute;
+        width: 0;
+        height: 0;
+        top: 50%;
+        left: 10px;
+    }
+`;
```
gtrufitt, line 70: Not sure what this is for, could you add a comment?

src/web/components/Dateline.tsx
```diff
@@ -17,18 +17,89 @@ const dateline = css`
...
+}> = ({ primaryDateline, secondaryDateline }) => {
+    return (
+        <div className={dateline}>
+            {primaryDateline !== secondaryDateline ? (
+                <div
+                    className={cx(
+                        toggleClass,
+                        pCheckboxStyle,
+                        labelStyles,
+                        dateline,
```
gtrufitt, line 87: Could you apply these classes directly the the elements that use them?

src/web/components/Dateline.tsx
```diff
@@ -17,18 +17,89 @@ const dateline = css`
...
+const toggleClass = css`
+    input[type='checkbox'] {
+        position: absolute;
+        top: 0;
+        left: 0;
+        width: 100%;
+        height: 100%;
+        opacity: 0;
+        display: none;
+    }
```
gtrufitt, line 36: Does this not just need the `display: none`?

src/web/components/Dateline.tsx
gtrufitt: In storybook, I am seeing a hover state when there's no secondaryDateLine in the markup though https://5dfcbf3012392c0020e7140b-pfylxkkawk.chromatic.com/?path=/story/components-articlemeta--article-story because it's not in the fixtures but primary !== secondary then becomes true

src/web/components/Dateline.tsx
```diff
@@ -17,18 +17,89 @@ const dateline = css`
...
+
+    input[type='checkbox']:checked ~ span ~ p {
```
gtrufitt, line 38: Are you able to move the input below the span to avoid having to do the double `~`?

src/web/components/Dateline.tsx
```diff
@@ -17,18 +17,89 @@ const dateline = css`
...
+        max-height: 80px;
+    }
+`;
+
+const pCheckboxStyle = css`
+    p {
+        max-height: 0;
+        overflow: hidden;
+        transition: max-height 0.4s ease;
```
gtrufitt, line 47: Nice little bit of UX!

src/web/components/Dateline.tsx
```diff
@@ -17,18 +17,89 @@ const dateline = css`
...
+    }
+`;
+
+const labelStyles = css`
```
gtrufitt, line 51: Yeah, I think it would great to move the `labelStyles` directly to a className attr on label

src/web/components/Dateline.tsx
```diff
@@ -17,18 +17,89 @@ const dateline = css`
...
+                        toggleClass,
+                        pCheckboxStyle,
+                        labelStyles,
+                        dateline,
+                    )}
+                >
+                    <input type="checkbox" id="dateToggle" />
+                    <span>
+                        <label htmlFor="dateToggle">{primaryDateline}</label>
+                    </span>
```
gtrufitt, line 93: Is this wrapping span needed here?

src/web/layouts/StandardLayout.tsx
```diff
@@ -425,6 +424,9 @@ export const StandardLayout = ({
                                 webPublicationDateDisplay={
                                     CAPI.webPublicationDateDisplay
                                 }
+                                secondaryDateline={
```
gtrufitt, line 427: This will need adding to all layouts

index.d.ts
```diff
@@ -172,6 +172,7 @@ interface Block {
     title?: string;
     firstPublished?: number;
     firstPublishedDisplay?: string;
+    secondaryDateLine?: string;
```
gtrufitt, line 175: I think you need to run `make gen-schema` too

gtrufitt, line 175: I don't think secondaryDateLine is optional from DCR CAPI currently (@shtukas).

src/web/components/Dateline.tsx
```diff
@@ -17,18 +17,89 @@ const dateline = css`
...
 
-const description = css`
-    ${visuallyHidden}
+const secondaryDateWidth = css`
+    ${from.leftCol} {
+        width: 90%;
+    }
 `;
 
+// Checkbox functionality
```
gtrufitt, line 26:
```suggestion
// We use the 'Checkbox Hack' for the show-hide functionality of the secondary date line.
// https://css-tricks.com/the-checkbox-hack/
```

gtrufitt (changes requested):
Nice one! Nothing major here, just a few bits and bobs around fixtures and the way we use emotion styling in DCR.

It would be great to see a cypress test too for clicking on the dateline, though you will have to do it with a 'fixture' to ensure stable data.

src/web/components/Dateline.tsx
nitro-marky: I think the idea for descriptionText was to use it to show "Last modified on" and "First published on", but with the data coming through as strings there's no simple way of determining which to use. I'll remove descriptionText for now.

src/web/components/ArticleMeta.tsx
```diff
@@ -297,8 +299,9 @@ export const ArticleMeta = ({
                                 />
                             )}
                             <Dateline
-                                dateDisplay={webPublicationDateDisplay}
+                                primaryDateline={webPublicationDateDisplay}
```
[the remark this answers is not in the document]
nitro-marky: Yeah, I'll change this

src/web/components/Dateline.tsx
gtrufitt: Right, I see what you mean, so I guess we need to know that from the data model..?

nitro-marky: On some articles the publication date is back-dated so the primary date is actually after the secondary date (https://www.theguardian.com/uk-news/2020/aug/12/reports-of-serious-injuries-after-train-derails-scotland-stonehaven). It may be easier to pass these dates as timestamps as they can be checked against each other mathematically to determine if the date is before or after.

gtrufitt: I think this is why we originally wanted to pass the full dateline with the 'last modified' 'first published' etc text intact 🤦

src/web/components/Dateline.tsx
nitro-marky: Didn't need it! I've removed it.

src/web/components/Dateline.tsx
nitro-marky: Yeah, I had a play with frontend yesterday and it's pretty straightforward to get the full string in secondaryDateLine if needed.

PR #1839, opened 2020-08-20
Remove old CMP

sndrs:
## What does this change?

- removes all usage of the old CMP
- adds a new `commercial` entry in webpack
  - will be filled in more as we extract commercial JS from frontend
- updates the way entry scripts are defined and loaded so that the order can be maintained regardless of the module type

## Why?

- old CMP not legal any more
- nor is it react, so initialising it from a react component made no sense.


src/web/server/document.tsx
```diff
@@ -81,19 +100,12 @@ export const document = ({ data }: Props) => {
...
-        getDist({ path: 'ophan.js', legacy: true }),
-        getDist({ path: 'lotame.js', legacy: true }),
-    ];
-    const lowPriorityNonLegacyScripts = [
-        getDist({ path: 'ga.js', legacy: false }),
-        getDist({ path: 'ophan.js', legacy: false }),
-        getDist({ path: 'lotame.js', legacy: false }),
-    ];
+    const lowPriorityScriptTags = generateScriptTags([
+        { src: 'https://www.google-analytics.com/analytics.js' },
```
mxdvl, line 104: Why is this added?

src/web/server/document.tsx
```diff
@@ -59,20 +82,16 @@ export const document = ({ data }: Props) => {
...
-    ];
-    const priorityNonLegacyScripts = [
-        getDist({ path: 'sentry.js', legacy: false }),
-        getDist({ path: 'dynamicImport.js', legacy: false }),
-        getDist({ path: 'react.js', legacy: false }),
-    ];
+    const priorityScriptTags = generateScriptTags(
+        [
+            { src: polyfillIO },
+            { src: 'commercial.js', module: true },
```
mxdvl, line 88: What's `commercial.js`?

PR #1845, opened 2020-08-24
Add retries to the cypress config

gtrufitt:
# What does this change?

Adds 2 retries for failing cypress tests. We sometimes get test failures in CI due to network or test runner oddness - these optional retries should prevent that flakiness from failing the build.

image

## Bump to Cypress 5.0

https://docs.cypress.io/guides/references/changelog.html#5-0-0

Retries only exists in Cypress 5 so I also bumped the Cypress version. There were no code changes required to migrate.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -3,5 +3,6 @@
     "viewportHeight": 860,
     "baseUrl": "http://localhost:3030/",
     "video": false,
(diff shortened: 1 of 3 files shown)


shtukas (approved):

shtukas:
+1

PR #1847, opened 2020-08-24
Remove old cmp, take 2

sndrs:
## What does this change?

- removes all usage of the old CMP
- adds a new `commercial` entry in webpack
  - will be filled in more as we extract commercial JS from frontend
- updates the way entry scripts are defined and loaded so that the order can be maintained regardless of the module type
- tries to handle uncaught errors caused by 3rd party vendors making incorrect calls to the `__tcfapi` not being handled by sourcepoint and so breaking our tests

## Why?

- old CMP not legal any more
- nor is it react, so initialising it from a react component made no sense.

rebuild of #1511 and take 2 of #1839


cypress/support/index.js
```diff
@@ -36,3 +36,11 @@ beforeEach(function () {
 
 // Alternatively you can use CommonJS syntax:
 // require('./commands')
+
+Cypress.on('uncaught:exception', (err, runnable) => {
```
sndrs, line 40: @gtrufitt this PR is the same as before, except for this bit

sndrs:
need a strategy, closing for now

PR #1851, opened 2020-08-25
Refactor bundle loading

sndrs:
# What does this change?

- refactors bundle loading so that guardian bundles and 3rd party scripts can be easily interleaved while priority is maintained

## Why?

should make it easier to add new bundles of different types to the page while maintaining a global order

_was part of #1511 but then became unneeded – is it still useful?_

### DIFF
--- a/scripts/webpack/browser.js
+++ b/scripts/webpack/browser.js
@@ -45,6 +45,7 @@ module.exports = ({ isLegacyJS }) => ({
         react: scriptPath('react'),
         lotame: scriptPath('lotame'),
         dynamicImport: scriptPath('dynamicImport'),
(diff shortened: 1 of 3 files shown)


sndrs:
@gtrufitt is this PR useful for you still? 

at some point we'll have a commercial entry point which will probably need this, but CMP doesn't need it. so there's no use for this from commdev side just yet...

gtrufitt:
> @gtrufitt is this PR useful for you still? 

Yeah, I would be keen to see this in as its defo an architectural improvement and a step to making it easier to breaking out scripts

gtrufitt (approved):

PR #1854, opened 2020-08-25
Use Trail Text for the meta description

gtrufitt:
# What does this change?

Uses trailText as per Frontend for the meta description.

I checked `og:description` which is correct. I checked AMP description meta which is correct.

Json schema doesn't exist, that requires an update on Frontend to the linkedData passed from DCR CAPI.

### Before

Screen Shot 2020-08-25 at 13 58 25

### After

Screen Shot 2020-08-25 at 13 57 52

## Why?

S EE OOOOH.


shtukas (approved):

nicl (approved):
Looks controversial to me :)

PR #1859, opened 2020-08-26
Braze banner integration

tjmw:
## What does this change?

Replaces #1791 as a more realistic POC for showing a message from Braze as a banner. We lazily load the Braze Web SDK only for users who we know may see the banner (logged in, digital subscriber). The banner component itself comes from the @guardian/braze-components NPM package (also lazily loaded).

### Before

No messages from Braze on the web.

### After

Messages from Braze on the web!

## Why?

Marketing already use Braze across multiple other platforms (email, apps) and they would like to extend this to the web.

## TODO

- [x] Once #1826 has been merged, integrate this work with the banner picker
- [ ] Analytics events
- [ ] Implement finalized braze banner designs
- [x] Move away from our custom TypeScript type definitions, since the braze sdk now ships with its own

### DIFF
--- a/fixtures/CAPI/review/showcaseReview.ts
+++ b/fixtures/CAPI/review/showcaseReview.ts
@@ -2660,6 +2660,7 @@ export const showcaseReviewCAPI: CAPIType = {
     config: {
         ajaxUrl: 'https://api.nextgen.guardianapps.co.uk',
         discussionApiUrl: 'https://discussion.theguardian.com/discussion-api',
+        idApiUrl: 'https://idapi.theguardian.com',
         sentryPublicApiKey: '344003a8d11c41d8800fbad8383fdc50',
         sentryHost: 'app.getsentry.com/35463',
         dcrSentryDsn:
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -1,0 +1,134 @@
+import React, { useEffect, useState } from 'react';
+import * as emotion from 'emotion';
+import * as emotionCore from '@emotion/core';
+import * as emotionTheming from 'emotion-theming';
+import { onConsentChange } from '@guardian/consent-management-platform';
+import { getZIndex } from '@root/src/web/lib/getZIndex';
+import { Props as BrazeBannerProps } from '@guardian/braze-components';
+import { CanShowResult } from './bannerPicker';
+
+export const brazeVendorId = '5ed8c49c4b8ce4571c7ad801';
+
+type Props = {
+    meta: any;
+};
+
+const containerStyles = emotion.css`
+    position: fixed;
+    bottom: -1px;
+    width: 100%;
+    ${getZIndex('banner')}
+`;
+
+export const hasRequiredConsents = (): Promise<boolean> =>
+    new Promise((resolve) => {
+        onConsentChange(({ tcfv2, ccpa }) => {
+            const consentGivenUnderCcpa = ccpa && !ccpa.doNotSell;
+            const consentGivenUnderTcfv2 =
+                tcfv2 && tcfv2.vendorConsents[brazeVendorId];
+
+            resolve(!!(consentGivenUnderCcpa || consentGivenUnderTcfv2));
+        });
+    });
+
+export const canShow = async (
+    asyncBrazeUuid: Promise<null | string>,
+    isDigitalSubscriber: undefined | boolean,
+): Promise<CanShowResult> => {
+    const { brazeSwitch } = window.guardian.config.switches;
+    const apiKey = window.guardian.config.page.brazeApiKey;
+
+    if (!(brazeSwitch && apiKey)) {
+        return { result: false };
+    }
+
+    if (!isDigitalSubscriber) {
+        return { result: false };
+    }
+
+    const [brazeUuid, hasGivenConsent] = await Promise.all([
+        asyncBrazeUuid,
+        hasRequiredConsents(),
+    ]);
+
+    if (!(brazeUuid && hasGivenConsent)) {
+        return { result: false };
+    }
+
+    try {
+        const { default: appboy } = await import(
+            /* webpackChunkName: "braze-web-sdk-core" */ '@braze/web-sdk-core'
+        );
+
+        appboy.initialize(apiKey, {
+            noCookies: true,
+            baseUrl: 'https://sdk.fra-01.braze.eu/api/v3',
+            sessionTimeoutInSeconds: 1,
+            minimumIntervalBetweenTriggerActionsInSeconds: 0,
+        });
+
+        return new Promise((resolve) => {
+            appboy.subscribeToInAppMessage((message) => {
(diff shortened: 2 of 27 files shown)


src/web/layouts/StandardLayout.tsx
```diff
@@ -475,6 +475,7 @@ export const StandardLayout = ({
                                 {showMatchStats && <div id="match-stats" />}
 
                                 {showBodyEndSlot && <div id="slot-body-end" />}
+                                <div id="braze" />
```
paulbrown1982, line 478: Can we give a more unique id to this element than just `"braze"` e.g. `"braze-message-container"`?

tjmw: Ah good spot @paulbrown1982. This is actually left over from early work where we were rendering a braze message outside of the banner picker flow (which didn't exist at that point). So actually I think this element just needs to be removed.

tjmw:
Hi @guardian/guardian-frontend-team 👋 Bumping this in case anyone is available to give it a review! We won't be turning the Braze banner integration on in this current state (we're ironing out some details with the banner design and configuration) but it'd be great to get this foundation merged and follow up with smaller more focused PRs for the remainder of the work. Cheers!

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -0,0 +1,135 @@
...
+        onConsentChange(({ tcfv2, ccpa }) => {
+            const consentGivenUnderCcpa = ccpa && !ccpa.doNotSell;
+            const consentGivenUnderTcfv2 =
+                tcfv2 && tcfv2.vendorConsents[brazeVendorId];
+
+            resolve(!!(consentGivenUnderCcpa || consentGivenUnderTcfv2));
+        });
+    });
+
+export const canShow = async (
```
nicl, line 34: This function gets a bit complex and hard to read, partly because of the nested promises going on. It might be worth lifting out some of the promises to make things easier (or defining an inner function to do this). Something like:

```
export const canShow = async (
    hasGivenConsent: boolean,
    isDigitalSubscriber: undefined | boolean,
    brazeUUID: string | null,
    const { brazeSwitch } = window.guardian.config.switches;
    const apiKey = window.guardian.config.page.brazeApiKey;

    if (!(brazeSwitch && apiKey)) {
        return Promise.resolve({ result: false });
    }

    if (isDigitalSubscriber) {
        return Promise.resolve({ result: false });
    }

    // then if hasconsent (promise)
    if (!brazeUUID || !hasGivenConsent) {
        return Promise.resolve({ result: false });
    }

    return import(
        /* webpackChunkName: "braze-web-sdk-core" */ '@braze/web-sdk-core'
    )
        .then(({ default: appboy }) => {
            appboy.initialize(apiKey, {
                noCookies: true,
                baseUrl: 'https://sdk.fra-01.braze.eu/api/v3',
                sessionTimeoutInSeconds: 1,
                minimumIntervalBetweenTriggerActionsInSeconds: 0,
            });

            return appboy.subscribeToInAppMessage((message: any) => {
                const meta = message.extras;

                appboy.changeUser(brazeUUID);
                appboy.openSession();

                if (meta) {
                    return { result: true, meta };
                }

                return { result: false };
            });
        })
        .catch(() => ({ result: false }));
};
```

Even if you don't want to go this far, you can avoid the outer Promise wrapping, by using `Promise.resolve` for the initial early returns.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -0,0 +1,135 @@
...
+    const [ExampleComponent, setExampleComponent] = useState<
+        React.FC<BrazeBannerProps>
+    >();
+
+    useEffect(() => {
+        if (meta && meta['test-key']) {
+            // TODO: unify the way we handle sharing these deps (this is
+            // duplicated in SlotBodyEnd). Probably via the automat client
+            // library.
+            window.guardian.automat = {
```
nicl, line 99: Let's use `automat-modules` for this, but can be a separate PR (which I'm happy to raise for here and slot body end too).

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -0,0 +1,135 @@
...
+                .catch((error) =>
+                    window.guardian.modules.sentry.reportError(
+                        error,
+                        'braze-banner',
+                    ),
+                );
+        }
+    }, [meta]);
+
+    if (ExampleComponent && meta && meta['test-key']) {
```
nicl, line 122: You might be able to use the funky `meta?.['test-key']` now instead of `meta && meta['test-key']`.

nicl (approved):
The only thing I'd really suggest changing is the `canShow` function, which is hard to read atm.

fixtures/CAPI/review/showcaseReview.ts
```diff
@@ -2660,6 +2660,7 @@ export const showcaseReviewCAPI: CAPIType = {
     config: {
         ajaxUrl: 'https://api.nextgen.guardianapps.co.uk',
         discussionApiUrl: 'https://discussion.theguardian.com/discussion-api',
+        idApiUrl: 'https://idapi.theguardian.com',
```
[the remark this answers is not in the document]
tjmw: That's an interesting point, but yep agreed probably for a separate PR.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: Yeah that's a great point, this could definitely be made more readable by flattening. You're right that the outer Promise wrapping can go. I guess that does make a small difference in that the initial checks go from being async to synchronous but I'm sure that's not a big deal.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: Yep good point. Let's add that in a separate PR.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: Ah yeah for some reason I thought that the linter or something wasn't happy with optional chaining in DCR. Definitely better if I can use it.

tjmw: It turns out there is an issue with optional chaining currently - for some reason Jest isn't happy with it (I get an `unexpected token` error).

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: I ended up re-writing this with `async/await` and i think it's an improvement (ed8da98).

src/web/components/StickyBottomBanner/BrazeBanner.tsx
nicl: Ah okay, yeah not worth fighting that really.

PR #1861, opened 2020-08-26
Remove old CMP

sndrs:
## What does this change?

- removes the old CMP and switches
- moves initialisation of current CMP to `App.tsx`
- disables CMP in storybook
- updates some fixtures to use the main `switches` fixture

## Why?

- new CMP is not a react component, and we don't need to trigger its display, just initialise it


src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -24,28 +20,17 @@ const getBannerLastClosedAt = (key: string): string | undefined => {
 
 const DEFAULT_BANNER_TIMEOUT_MILLIS = 2000;
 
-const buildNewCmpBannerConfig = (): Banner => {
-    return {
-        id: 'cmpUi',
-        canShow: () => willShowCMP().then((result) => ({ result: !!result })),
+const buildCmpBannerConfig = (): Banner => ({
```
sndrs, line 23: @tjmw does this look ok to you? locally i'm not seeing any banners after accepting the privacy one, is this wrong or ok?

src/web/components/App.tsx
```diff
@@ -292,6 +295,18 @@ export const App = ({ CAPI, NAV }: Props) => {
         }
     }, [CAPI.shouldHideReaderRevenue]);
 
+    useEffect(() => {
+        if (CAPI.config.switches.consentManagement && countryCode) {
```
gtrufitt, line 299: Pass in CAPI.config.switches.consent directly?

src/web/components/App.tsx
```diff
@@ -292,6 +295,18 @@ export const App = ({ CAPI, NAV }: Props) => {
         }
     }, [CAPI.shouldHideReaderRevenue]);
 
+    useEffect(() => {
```
gtrufitt, line 298: Could add a explainer comment?

gtrufitt (approved):

src/web/components/App.tsx
sndrs: how do i do that? in the `useEffect` dependency list?

jorgeazevedo:
@sndrs does this PR enable us to remove `old-cmp` from DCR and commercial's JS bundles? 
I was curious from this PR's title if it removed it, but I can see in the lighthouse report it's still being included

PageSpeed_Insights

sndrs:
@jorgeazevedo not yet...but nearly! this is part of the process. once nothing on dotcom is using it anymore, we'll release v5 of the CMP which will remove it entirely. we tried earlier but did too much at once. this is part of breaking the changes up into chunks

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
sndrs: talked offline

jorgeazevedo:
@sndrs big fan of splitting work into chunks, so it sounds like a good plan to me! thanks for clarifying 👍

src/web/components/App.tsx
gtrufitt: Sorry, missed this comment. 

I meant, rather than re-run the effect on `CAPI`, in the array` pass in `CAPI.config.switches.consentManagement`? I think that's doable?

`}, [countryCode, CAPI.config.switches.consentManagement]);`

gtrufitt (approved):

mxdvl (approved):

PR #1865, opened 2020-08-26
Add comment about WithSignInGateSlot

vlbee:
Adds some comments for clarity.

splicing the array in place does not really add much performance benefit as article elements are not very large arrays, so leaving as is. If the Long Reads become really really long, it might be worth changing.


nicl (approved):

gtrufitt:
oh... I didn't meant to merge this PR! 😆

PR #1869, opened 2020-08-27
Send article count to backend for banners

tomrf1:
This enables the backend to use article count restrictions in banner tests, as well as display the user's article count


TomPretty (approved):
💯

PR #1870, opened 2020-08-27
Disable cmp in lighthouse ci

jorgeazevedo:
## What does this change?

This disables the consent management platform from loading during lighthouse testing by setting the following cookie `gu-cmp-disabled=true`

Unfortunately lighthouse ci only supports setting cookies in the header of the original GET request, not actually in the browser state, which is what the cmp js code will look for in runtime.

This means that the only implementation available is based on puppeteer (a chrome instrumentation tool) which will basically open the URL under test and set a cookie *before* lighthouse runs its tests.

I'm also using this PR to disable the PWA category since we don't use it!

### Before

Banners_and_Alerts_and_Lighthouse_Report_Viewer

### After

Lighthouse_Report_Viewer

## Why?

The purpose is to make the results of the lighthouse tests more deterministic, but also to keep the report focused on issues introduced in this codebase

### DIFF
--- a/scripts/lighthouse/puppeteer-script.js
+++ b/scripts/lighthouse/puppeteer-script.js
@@ -3,13 +3,12 @@
  * @param {{url: string, options: LHCI.CollectCommand.Options}} context
  */
 module.exports = async (browser, context) => {
-  const url = 'http://localhost:9000/Article?url=https://www.theguardian.com/commentisfree/2020/feb/08/hungary-now-for-the-new-right-what-venezuela-once-was-for-the-left#noads';
   const page = await browser.newPage();
-  await page.goto(url);
+  await page.goto(context.url);
   const cookie = {
           name: 'gu-cmp-disabled',
           value: 'true',
-          url: url,
+          url: context.url,
           'Max-Age': '31536000', //year
         };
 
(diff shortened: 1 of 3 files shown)


scripts/lighthouse/puppeteer-script.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * @param {puppeteer.Browser} browser
+ * @param {{url: string, options: LHCI.CollectCommand.Options}} context
+ */
+module.exports = async (browser, context) => {
+  const url = 'http://localhost:9000/Article?url=https://www.theguardian.com/commentisfree/2020/feb/08/hungary-now-for-the-new-right-what-venezuela-once-was-for-the-left#noads';
```
gtrufitt, line 6: Think you can get this off of context?

`{{url: string, options: LHCI.CollectCommand.Options}} context`

Which will be useful if URLs is multiple in the array in lighthouserc!

gtrufitt (approved):
This actually looks really useful!

scripts/lighthouse/puppeteer-script.js
jorgeazevedo: oh!! great point, I totally missed that. I'll give it a try

PR #1872, opened 2020-09-01
Banner - do not overlap footer and override ad-takeover styles

tomrf1:
Fixes 2 issues which both involve styling the parent `bottom-banner` element:
### 1. The banner currently overlaps the footer:

Current behaviour:
Screen Shot 2020-08-28 at 14 28 04

Desired behaviour:
Screen Shot 2020-08-28 at 14 28 10

### 2. Ad-takeovers (which inject css into the page) break the layout:
(This is similar to the frontend fix here: https://github.com/guardian/frontend/blob/main/static/src/stylesheets/module/site-messages/_engagement-banner.scss#L223)

Current behaviour:
Screen Shot 2020-10-02 at 09 08 06

Desired behaviour:
Screen Shot 2020-10-02 at 09 09 03

And proof that the view event still works:
Screen Shot 2020-10-02 at 09 13 03

And proof that the consent banner still works:
Screen Shot 2020-10-02 at 09 18 06

### DIFF
--- a/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
+++ b/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
@@ -212,7 +212,7 @@ export const ReaderRevenueBanner = ({ meta, module }: Props) => {
             // The css here is necessary to put the container div in view, so that we can track the view
             <div
                 ref={setNode}
-                className={emotion.css`position: fixed; bottom: -1px; width: 100%; ${getZIndex(
+                className={emotion.css`width: 100%; ${getZIndex(
                     'banner',
                 )}`}
             >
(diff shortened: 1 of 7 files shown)


src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -208,7 +208,7 @@ export const ReaderRevenueBanner = ({ meta, module }: Props) => {
             // The css here is necessary to put the container div in view, so that we can track the view
```
jfsoul, line 208: Are we tracking the view still, given this comment?

jfsoul (approved):
Can't really see what fixes the issue here, is it `bottom: -1px` -> `bottom: 0`?

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
tomrf1: good question - I've checked this still works and added a screenshot of the ophan request

tomrf1:
> Can't really see what fixes the issue here, is it `bottom: -1px` -> `bottom: 0`?

@jfsoul the fix to make the banner not overlap the footer is moving the css up to the parent element

PR #1873, opened 2020-09-01
Use the Polyfill IO URL builder

gtrufitt:
## What does this change?

Adds a simple script to read the built JS and runs through [Polyfill Service URL Builder](https://github.com/Financial-Times/polyfill-service-url-builder).

## Size difference

Testing the two download sizes in IE11 (the most polyfilled supported browser) with a test page in WPT https://www.webpagetest.org/result/200901_5G_84980ed82d8c1a3cd68fb224a2d2af55/1/details/

### Before

Content Download: 363 ms
Bytes In (downloaded): 35.4 KB
Uncompressed Size: 127.6 KB

`https://assets.guim.co.uk/polyfill.io/v3/polyfill.min.js?rum=0&features=es6,es7,es2017,es2018,default-3.6,HTMLPictureElement,IntersectionObserver,IntersectionObserverEntry,fetch,NodeList.prototype.forEach&flags=gated&callback=guardianPolyfilled&unknown=polyfill&cacheClear=1`

### After

Content Download: 196 ms
Bytes In (downloaded): 23.8 KB
Uncompressed Size: 87.2 KB

`https://polyfill.io/v3/polyfill.min.js?features=Array.from,Array.isArray,Array.prototype.entries,Array.prototype.every,Array.prototype.fill,Array.prototype.filter,Array.prototype.find,Array.prototype.findIndex,Array.prototype.forEach,Array.prototype.includes,Array.prototype.indexOf,Array.prototype.keys,Array.prototype.lastIndexOf,Array.prototype.map,Array.prototype.reduce,Array.prototype.some,Array.prototype.sort,Array.prototype.values,ArrayBuffer,Blob,console,CustomEvent,DataView,Date.now,Date.prototype.toISOString,document,Element,Event,fetch,Function.prototype.bind,globalThis,JSON,localStorage,Map,modernizr:es5object,Number.isNaN,Object.assign,Object.entries,Object.freeze,Object.getOwnPropertyDescriptors,Object.getOwnPropertySymbols,Object.setPrototypeOf,Promise,Promise.prototype.finally,Reflect,Reflect.construct,requestAnimationFrame,Set,String.prototype.endsWith,String.prototype.includes,String.prototype.startsWith,String.prototype.trim,String.raw,Symbol,Symbol.for,Symbol.iterator,Symbol.prototype.description,Symbol.toStringTag,Uint16Array,Uint8Array,URL,URLSearchParams,WeakMap,WeakSet,XMLHttpRequest`

## Pros

- Ensures that polyfill.io only downloads polyfills features that are required from our JS bundles. I.E smaller JS download for older browsers
- If kept up to date, means that we don't miss any polyfills

## Cons

- Not a part of the current build process so may miss polyfills going forward (how might we add it? how might we add it without holding up push or commit or builds?)
- Doesn't capture the requirements of things like interactives, epics, banners that are loaded from elsewhere (breaks the contract of polyfill support?)

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
         "storybook": "node --max-old-space-size=4096 $(yarn bin)/start-storybook -s ./src/static -p 6006",
         "build-storybook": "node --max-old-space-size=4096 $(yarn bin)/build-storybook",
         "chromatic": "chromatic --build-script-name=build-storybook --exit-zero-on-changes",
-        "makeBuild": "NODE_ENV=production CI_ENV=github webpack --config scripts/webpack/frontend"
+        "makeBuild": "NODE_ENV=production CI_ENV=github webpack --config scripts/webpack/frontend",
+        "getPfUrl": "make build && node ./scripts/buildPolyfilllIoUrl.js"
     },
     "bundlesize": [
         {
@@ -64,6 +65,7 @@
         "compose-function": "^3.0.3",
         "compression": "^1.7.3",
         "cpy": "^8.1.0",
+        "create-polyfill-service-url": "^2.2.2",
         "curlyquotes": "^1.4.3",
         "dateformat": "^3.0.3",
         "diskusage": "^0.2.5",
--- a/makefile
+++ b/makefile
@@ -97,7 +97,7 @@ test: clean-dist install
 
 test-ci: clear clean-dist install
 	$(call log, "running tests")
-	@yarn test --verbose  --runInBand --collectCoverage --coverageReporters=lcov
+	@yarn 	 test --verbose  --runInBand --collectCoverage --coverageReporters=lcov
 
 bundlesize: clear clean-dist install build
 	@bundlesize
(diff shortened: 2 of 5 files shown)


shtukas (approved):

package.json
```diff
@@ -64,6 +65,7 @@
         "compose-function": "^3.0.3",
         "compression": "^1.7.3",
         "cpy": "^8.1.0",
+        "create-polyfill-service-url": "^2.2.2",
```
jfsoul, line 68: Is it not a devDependency?

jfsoul (commented):
Don't we have some uncontrolled page elements that will utilise the polyfill we provide?

gtrufitt:
> Don't we have some uncontrolled page elements that will utilise the polyfill we provide?

Ah. Yeah. Good point... 🤔

gtrufitt:
This is not something we can currently achieve with the architecture

PR #1874, opened 2020-09-01
Add default rich link

nicl:
## What does this change?

Show (a default) rich link even if:

* no JS
* ajax call to enrich fails

Note, we just show a 'news' default here, which is equivalent to Frontend. See:

https://github.com/guardian/frontend/blob/master/common/app/views/fragments/richLinkDefault.scala.html

### Before

(first rich link doesn't even appear when ajax call to enrich fails)

Screenshot 2020-09-01 at 16 49 40

### After

Screenshot 2020-09-01 at 16 43 02

## Why?

We want to show the rich link even if missing some of the info.

### DIFF
--- a/src/web/lib/ArticleRenderer.tsx
+++ b/src/web/lib/ArticleRenderer.tsx
@@ -35,6 +35,7 @@ import {
 import { Display } from '@root/src/lib/display';
 import { withSignInGateSlot } from '@root/src/web/lib/withSignInGateSlot';
 import { GuVideoBlockComponent } from '@root/src/web/components/elements/GuVideoBlockComponent';
+import { DefaultRichLink } from '../components/RichLink';
 
 // This is required for spacefinder to work!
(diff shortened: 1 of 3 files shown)


shtukas (approved):

src/web/lib/ArticleRenderer.tsx
```diff
@@ -235,7 +236,18 @@ export const ArticleRenderer: React.FC<{
...
+                            <DefaultRichLink
+                                index={i}
+                                headlineText={element.text}
+                                url={element.url}
+                                isPlaceholder={true}
+                            />
+                        </div>
+                    );
+
+                // return <div key={i} id={`rich-link-${i}`} />;
```
jfsoul, line 250: Don't forget to remove!

nicl: doh, will raise a quick PR to do this. Thanks.

PR #1876, opened 2020-09-02
Add switches to article fixtures

sndrs:
## What does this change?

- creates a `switches` fixture
- uses it in all article fixtures

## Why?

switch settings are not picked up in story book, meaning you can't test regressions in things that require a switch to be on. 

this is causing failing tests in #1861. this will enable us to add its switch.


jfsoul (approved):
nice 👍

PR #1877, opened 2020-09-02
Skip SignInGate Cypress e2e test suites if Frontend AB test switch is "off" or missing

vlbee:
## Why?

AB test switches are defined in Frontend and required to be "ON" for Cypress E2E Sign In Gate Test to successfully run and pass. However, in practice it is useful to be able to switch an AB test on and off in Frontend PROD Admin UI without having to make any code changes to a different repo. 
 
In order not to break CI builds in DCR if someone switches off a Sign In Gate AB test in the Frontend Admin tool, we now run a setup test suite to load active switches from Frontend into a Cypress Environment Variable available to this spec only.

 With the help of a utility function `onlyRunIf` that wraps each AB test suite ("describe"), we can use this env var to check if a switch is ON to conditionally decide which AB test suites to run. .

For local development and CODE, you will still need to ensure the switches are ON to run your cypress tests.

### DIFF
--- a/cypress/integration/e2e/sign-in-gate.spec.js
+++ b/cypress/integration/e2e/sign-in-gate.spec.js
@@ -60,6 +60,16 @@ describe('Sign In Gate Tests', function () {
             .then(cy.wrap);
     };
 
+    const switchIsOn = (switchName) => {
+        const switches = Cypress.env('SWITCHES');
+        const isOn = !!switches[switchName];
+        if (!isOn)
+            console.log(
+                `${switchName} switch in Frontend does not exist or is turned off - skipping ${switchName} test suite`,
+            );
+        return isOn;
+    };
+
     /**
      * AB test switches are defined in Frontend and required to be "ON" for Cypress E2E Sign In Gate Test to successfully pass.
      * In order not to break CI builds in DCR if someone switches off a Sign In Gate AB test in the Frontend Admin tool, we run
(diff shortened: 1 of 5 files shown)


cypress/integration/e2e/sign-in-gate.spec.js
```diff
@@ -58,380 +60,424 @@ describe('Sign In Gate Tests', function () {
...
+     * but you will need to make sure switches are ON when developing locally or in CODE.
+     *
+     * Ensure you wrap all your AB test suites with "onlyRunIf(switchIsOn('abNameOfTest', () => {})" for this to work.
+     */
+    describe('Get live Switches from Frontend', function () {
+        it('Load article and get switches', function () {
+            visitArticle();
+            cy.window().then((win) => {
+                const switches = win.guardian.config.switches;
+                Cypress.env('SWITCHES', switches);
```
vlbee, line 79: This is  a global var scoped to this spec only

cypress/integration/e2e/sign-in-gate.spec.js
```diff
@@ -58,380 +60,424 @@ describe('Sign In Gate Tests', function () {
...
+     *
+     * Warning - this means we can turn switches on/off in the Frontend Admin PROD UI without needing to make code changes to DCR,
+     * but you will need to make sure switches are ON when developing locally or in CODE.
+     *
+     * Ensure you wrap all your AB test suites with "onlyRunIf(switchIsOn('abNameOfTest', () => {})" for this to work.
+     */
+    describe('Get live Switches from Frontend', function () {
+        it('Load article and get switches', function () {
+            visitArticle();
+            cy.window().then((win) => {
```
vlbee, line 77: I have to run "cy" commands within a test, which is why there is a suite to load the live switches from Frontend.  Open to better suggestions!

cypress/integration/e2e/sign-in-gate.spec.js
```diff
@@ -1,5 +1,7 @@
 /* eslint-disable no-undef */
 /* eslint-disable func-names */
+/* eslint-disable mocha/no-setup-in-describe */
+import { onlyOn as onlyRunIf } from '@cypress/skip-test';
```
vlbee, line 4: Neither Jest nor Cypress support skipping tests conditionally out of the box, so this helper package seemed the best way to go. 
https://github.com/facebook/jest/issues/7245
https://github.com/cypress-io/cypress-skip-test

gtrufitt:
Checked this out locally and had a play and seems like a good solution to me! You could hit an endpoint 	like https://www.theguardian.com/film/2020/sep/03/tenet-dialogue-christopher-nolan-sound-technology.json?dcr and get the switches from there rather than load a page and get from the window but I don't think it's the end of the world to do it that way, and actually is maybe more realistic to what a *user* will see (switches defining whether a test runs from the window of a browser) than the JSON endpoint anyway, which is what Cypress tests should more closely align with.

vlbee:
@gtrufitt  Thanks!  That would make it faster but because Cypress runs asynchronously, I still don't think it would guarantee the switch information would be available when we needed them in Cypress. 

@coldlink had look and came round independently to the original solution to just override the switches in HydratedApp if the app is running in the Cypress environment. He will raise a PR for this and tag  you. I think overall it's a cleaner solution, more reliable, and adds less "bespoke" code to what devs have to write when adding e2e tests that rely on Switches.

PR #1879, opened 2020-09-03
Add Height and Width to Images

gtrufitt:
## What does this change?

This adds a height and width to the image element, so the pre-parser in the browser can reserve space and prevent re-rendering on image load.

## Why?

Improves CLS and LCP times on DCR.

### DIFF
--- a/src/web/components/Picture.tsx
+++ b/src/web/components/Picture.tsx
@@ -22,7 +22,9 @@ export const Picture: React.FC<{
     sources: PictureSource[];
     alt: string;
     src: string;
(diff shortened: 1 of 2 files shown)


gtrufitt:
Have some fixes to add to storybook that are dealt with in the templates.

shtukas (approved):

SiAdcock:
I just enjoyed watching [Jenn Simmons' explanation](https://css-tricks.com/do-this-to-improve-image-loading-on-your-website/) of how the browser figures out the aspect ratio and reserves the correct amount of space responsively 😍

PR #1880, opened 2020-09-03
Updating standfirst to correct li element styling

rcrphillips:
## What does this change?
Updates the styling of the <li> item within a standfirst to use a lighter font style in order to be more aligned with frontend.

### Before
Screen Shot 2020-09-03 at 14 32 12

### After
Screen Shot 2020-09-03 at 14 35 15

## Why?
Parity

### DIFF
--- a/src/web/components/Standfirst.tsx
+++ b/src/web/components/Standfirst.tsx
@@ -3,7 +3,7 @@ import { css, cx } from 'emotion';
 import { neutral } from '@guardian/src-foundations/palette';
 import { space } from '@guardian/src-foundations';
 import { from } from '@guardian/src-foundations/mq';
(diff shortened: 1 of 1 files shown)


gtrufitt:
I think the original style was correct in some cases: eg: https://www.theguardian.com/world/2020/feb/10/irish-general-election-everything-you-need-to-know?dcr=false

gtrufitt:
> I think the original style was correct in some cases: eg: [theguardian.com/world/2020/feb/10/irish-general-election-everything-you-need-to-know?dcr=false](https://www.theguardian.com/world/2020/feb/10/irish-general-election-everything-you-need-to-know?dcr=false)

Actually, you're right, this lgtm I think

gtrufitt (approved):

PR #1881, opened 2020-09-03
Fix Yarn link bug with Typescript

gtrufitt:
## What does this change?

Typescript gets confused with `yarn link` because it follows the symlink, this adds this fix https://github.com/microsoft/TypeScript/issues/11916#issuecomment-257130001

### Before

image

### After

Type checks work.


amyhughes:
Thank you, I've lost a lot of time to this problem this week, it's great to know others won't have to suffer this...

shtukas (approved):

PR #1883, opened 2020-09-03
Deactivate SignInGateDismissWindow test

vlbee:
# What does this change?

Deletes Cypress e2e test for the DismissWindow sign in gate

Deletes the DismissWindow sign in gate test from the code base except the canShow variant  files in case we want to roll out either of these.


coldlink (approved):
☕

PR #1885, opened 2020-09-03
Refactoring captions in DCR

rcrphillips:
## What does this change?
Refactors some of the captions within components in order to add conditional padding. Also turns out that Spotify was using the AMP Caption component.

### Before
Screen Shot 2020-09-03 at 17 52 27

### After
Screen Shot 2020-09-03 at 17 50 42

## Why?
Parity


shtukas (approved):

src/web/components/elements/GuVideoBlockComponent.tsx
```diff
@@ -13,18 +13,18 @@ export const GuVideoBlockComponent: React.FC<{
     credit: string;
     caption?: string;
 }> = ({ html, pillar, designType, display, credit, caption }) => {
+    const hasCaption = caption != null;
```
gtrufitt, line 16: Not sure it'll be `null` here?

gtrufitt: I would maybe just do a truthy check on `caption` because in all cases (undefined, null, empty string), we don't want to add the caption, so though the descriptive var is useful, it's probably a bit of indirection that isn't required

src/web/components/elements/GuVideoBlockComponent.tsx
```diff
@@ -13,18 +13,18 @@ export const GuVideoBlockComponent: React.FC<{
...
+
+    const embedContainer = css`
+        width: 100%;
```
gtrufitt, line 19: You lost `video { width: 100% }` here, checking that's correct?

rcrphillips: Nice catch

src/web/components/elements/GuVideoBlockComponent.tsx
rcrphillips: Updated to adopt that

shtukas:
+1

PR #1887, opened 2020-09-04
Ability to force AB Switch state for Cypress Tests

coldlink:
AB test switches are defined in Frontend and required to be "ON" for Cypress E2E Sign In Gate Test to successfully run and pass. However, in practice it is useful to be able to switch an AB test on and off in Frontend PROD Admin UI without having to make any code changes to a different repo.

In order not to break CI builds in DCR if someone switches off a Sign In Gate AB test in the Frontend Admin tool, we've added the ability to define a state that a switch should be in DCR, and will only run within a Cypress environment.

In #1877 @vlbee tried to do a similar thing by skipping tests if the switch was set to off in Frontend PROD, but this caused non-deterministic behaviour due to tests running asynchronously.

This presents a simpler solution, as we can set the state of the test switch in `cypressSwitches` which will only be picked up by Cypress, meaning that tests will still run regardless of the state in PROD.

@gtrufitt it would be useful to get your thoughts in this too?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,21 +6,20 @@ Frontend rendering framework for theguardian.com. It uses [React](https://reactj
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 <!-- Automatically created with yarn run createtoc and on push hook -->
 
-- [Dotcom Rendering](#dotcom-rendering)
-  - [Where can I see Dotcom Rendering in Production?](#where-can-i-see-dotcom-rendering-in-production)
-  - [Quick start](#quick-start)
-    - [Install Node.js](#install-nodejs)
-    - [Running instructions](#running-instructions)
-    - [Detailed Setup](#detailed-setup)
-    - [Technologies](#technologies)
-    - [Architecture Diagram](#architecture-diagram)
-    - [Concepts](#concepts)
-    - [Feedback](#feedback)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -6,21 +6,20 @@ Frontend rendering framework for theguardian.com. It uses [React](https://reactj
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 <!-- Automatically created with yarn run createtoc and on push hook -->
 
-- [Dotcom Rendering](#dotcom-rendering)
```
coldlink, line 9: not sure why this file updated automatically?

coldlink: since i didn't make changes to it

src/web/experiments/cypress-switches.ts
```diff
@@ -0,0 +1,25 @@
+// Forced AB Test Switches for Cypress testing
+
+// AB test switches are defined in Frontend and required to be "ON" for Cypress E2E Tests to successfully run and pass.
```
vlbee, line 3: Thanks - inline comments more helpful than docs to quickly orient devs

vlbee (approved):
I prefer this solution so giving it a ✅   but see what @gtrufitt thinks!

gtrufitt (approved):
LGTM!

PR #1899, opened 2020-09-08
Remove lazy loading from main media images

gtrufitt:
## What does this change?

Makes main media non-lazy. This was a mistake https://web.dev/native-lazy-loading/#avoid-lazy-loading-images-that-are-in-the-first-visible-viewport

> images which are loaded lazily the browser currently needs to wait until it knows where the image is positioned on the page,


nicl (approved):
Makes sense!

shtukas (approved):

PR #1903, opened 2020-09-11
Curated container test

nicl:
## What does this change?

Introduce a basic curated container component. 

**Note, we still need to do some work in Frontend to return the current headlines component - currently it's including the 'Brexitu' container.** There are also issues here - for example, the `sss` in the first container item, which we'll need to consider.

### Before

Only this:

Screenshot 2020-09-11 at 11 43 17

### After

Screenshot 2020-09-11 at 11 36 26

## Why?

See the test description etc., but the basic hypothesis is that this additional curated container will prove valuable to users and increase CTR on article pages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ Frontend rendering framework for theguardian.com. It uses [React](https://reactj
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 <!-- Automatically created with yarn run createtoc and on push hook -->
 
+- [Moving to main as default branch](#moving-to-main-as-default-branch)
 - [Where can I see Dotcom Rendering in Production?](#where-can-i-see-dotcom-rendering-in-production)
 - [Quick start](#quick-start)
   - [Install Node.js](#install-nodejs)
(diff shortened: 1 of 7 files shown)


paperboyo:
When the journey follows from a container whose title is clickable, eg https://www.theguardian.com/uk#opinion, will its title also clickable on the article page?

nicl:
@paperboyo initially this is going to be fixed as the headlines component - so not the version where it's the container you came from. The title could definitely be clickable, but for this test I think the intention is the keep behaviour/styling the same as the 'More on this story' style ones, to control for design and only test the content and fact that the new container exists.

But I think cc @rcrphillips that we are likely to run further tests like the one you are describing!

paperboyo:
> fixed as the headlines component

May bad, sorry (was in the PR title!). Headlines aren’t clickable, so a non-issue. Of course, Headlines should be clickable as we should be able to render historical content on curated fronts just like we do with tag pages, but that’s for when we will consider Onward Journeys as a backend problem as well ;-)

gtrufitt (approved):
Nice to see the A/B test framework in the wild, fairly nice ergonomics. Anything that'd make it easier (I am glad I changed isUserInVariant to take string IDs!)

liywjl (approved):
it looks good 🙂  

I do not fully understand what is meant by `isUserInVariant` (does it mean the users is logged in during the test?). 

I am still brushing up on the use of `useAB`, looks really cool, but not sure how all the bits and bobs work.

+1 on changing: `if (!url)`

PR #1904, opened 2020-09-11
Revert "Braze banner integration"

tomrf1:
Reverts guardian/dotcom-rendering#1859
because DCR is no longer requesting banners from support-dotcom-components

PR #1905, opened 2020-09-11
Braze Banner Integration (Attempt 2)

tjmw:
## What does this change?

This PR reverts #1904 which itself reverted #1859. This re-adds the (initally turned off) Braze banner integration but fixes an issue which was preventing reader revenue banners from showing. See #1859 for details of that functionality. The second commit here (23cc963) is the interesting part with the fix.

### DIFF
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -35,6 +35,7 @@ import { getCookie } from '@root/src/web/browser/cookie';
 import { getCountryCode } from '@frontend/web/lib/getCountryCode';
 import { getDiscussion } from '@root/src/web/lib/getDiscussion';
 import { getUser } from '@root/src/web/lib/getUser';
+import { getBrazeUuid } from '@root/src/web/lib/getBrazeUuid';
 import { getCommentContext } from '@root/src/web/lib/getCommentContext';
 import { FocusStyleManager } from '@guardian/src-foundations/utils';
 import { incrementAlreadyVisited } from '@root/src/web/lib/alreadyVisited';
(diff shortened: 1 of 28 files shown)


src/web/components/App.tsx
```diff
@@ -187,14 +187,19 @@ export const App = ({ CAPI, NAV }: Props) => {
...
-            callGetBrazeUuid();
         }
-    }, [isSignedIn, CAPI.config.discussionApiUrl, CAPI.config.idApiUrl]);
+    }, [isSignedIn, CAPI.config.discussionApiUrl]);
+
+    useEffect(() => {
+        if (isSignedIn) {
+            setAsyncBrazeUuid(getBrazeUuid(CAPI.config.idApiUrl));
+        } else {
+            setAsyncBrazeUuid(Promise.resolve(null));
```
tomrf1, line 200: could you instead pass `Promise.resolve(null)` as the initial value when you call `useState`? Would the behaviour be the same?

tomrf1 (approved):
looks good

src/web/components/App.tsx
tjmw: Annoyingly I don't think that quite works in the case where the user is signed in and there _is_ a braze UUID. Since we'll use the first promise and the banner logic will only run once (by design, since we only want to show one banner per page view).

Now we'll wait until the `asyncBrazeUuid` is defined (which should always happen now) before running the banner picker.

tjmw:
Thanks for the review @tomrf1 :+1: I pushed this to code and verified that I'm seeing banners as expected there.

PR #1906, opened 2020-09-11
Outstream ad css fix

ioanna0:
## What does this change?
Fixes the stying of outsteam ads css to support a new ad patner called Connatix

### Before
Screenshot 2020-09-11 at 15 26 49

### After
Screenshot 2020-09-11 at 15 26 31

## Why?
Connatix ads will render OK

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,6 @@ Frontend rendering framework for theguardian.com. It uses [React](https://reactj
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 <!-- Automatically created with yarn run createtoc and on push hook -->
 
-- [Moving to main as default branch](#moving-to-main-as-default-branch)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -6,6 +6,7 @@ Frontend rendering framework for theguardian.com. It uses [React](https://reactj
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 <!-- Automatically created with yarn run createtoc and on push hook -->
 
+- [Moving to main as default branch](#moving-to-main-as-default-branch)
```
mxdvl, line 9: Worth rebasing?

mxdvl (approved):
💯

README.md
ioanna0: I've tried that but seems like that change is still there. I think it is auto-generated via `yarn run` to provide a quick link of the new section `Moving to main as default branch`

ioanna0: looks like I cannot even revert it. I committed a revert but an autogenerated commit called `Add TOC update` followed straight away and put it back 😅

PR #1908, opened 2020-09-11
Add Braze Banner Integration

tjmw:
## What does this change?

This PR follows on from #1859, which was previously merged and then reverted as it caused an issue with reader revenue banners not showing. That issue should now be resolved in this PR. There are two commits:

* the revert of the revert of the original commit
* the fix for the issue

See #1859 for details on the Braze integration.

Regarding the fix, here's a preview of the banner picking logic working correctly on CODE:

### DIFF
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -31,6 +31,14 @@ export const hasRequiredConsents = (): Promise<boolean> =>
         });
     });
 
+// We can show a Braze banner if:
+// - The Braze switch is on
+// - We have a Braze API key
+// - The user is a digital subscriber
+// - We've got a Braze UUID from the API, given a user's ID Creds
+// - The user has given Consent via CCPA or TCFV2
+// - The Braze websdk appboy initialisation does not throw an error
+// - The Braze app Boy subscription to in app message returns meta info
 export const canShow = async (
     asyncBrazeUuid: Promise<null | string>,
     isDigitalSubscriber: undefined | boolean,
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -1,0 +1,134 @@
+import React, { useEffect, useState } from 'react';
+import * as emotion from 'emotion';
+import * as emotionCore from '@emotion/core';
+import * as emotionTheming from 'emotion-theming';
+import { onConsentChange } from '@guardian/consent-management-platform';
+import { getZIndex } from '@root/src/web/lib/getZIndex';
+import { Props as BrazeBannerProps } from '@guardian/braze-components';
+import { CanShowResult } from './bannerPicker';
+
+export const brazeVendorId = '5ed8c49c4b8ce4571c7ad801';
+
+type Props = {
+    meta: any;
+};
+
+const containerStyles = emotion.css`
+    position: fixed;
+    bottom: -1px;
+    width: 100%;
+    ${getZIndex('banner')}
+`;
+
+export const hasRequiredConsents = (): Promise<boolean> =>
+    new Promise((resolve) => {
+        onConsentChange(({ tcfv2, ccpa }) => {
+            const consentGivenUnderCcpa = ccpa && !ccpa.doNotSell;
+            const consentGivenUnderTcfv2 =
+                tcfv2 && tcfv2.vendorConsents[brazeVendorId];
+
+            resolve(!!(consentGivenUnderCcpa || consentGivenUnderTcfv2));
+        });
+    });
+
+export const canShow = async (
+    asyncBrazeUuid: Promise<null | string>,
+    isDigitalSubscriber: undefined | boolean,
+): Promise<CanShowResult> => {
+    const { brazeSwitch } = window.guardian.config.switches;
+    const apiKey = window.guardian.config.page.brazeApiKey;
+
+    if (!(brazeSwitch && apiKey)) {
+        return { result: false };
+    }
+
+    if (!isDigitalSubscriber) {
+        return { result: false };
+    }
+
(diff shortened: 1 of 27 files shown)


src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -117,11 +117,41 @@ export const StickyBottomBanner = ({
...
+// until all of our dependencies are defined. Then when they are all defined we
+// only want to run the banner picker once.
+export const StickyBottomBanner = ({
+    isSignedIn,
+    asyncCountryCode,
+    CAPI,
+    asyncBrazeUuid,
+    isDigitalSubscriber,
+}: Props) => {
+    if (
```
nicl, line 139: Is this just defensive in purpose?

nicl (approved):
(Though might be worth getting another pair of eyes that are more client-side experty.)

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
tjmw: Yeah exactly. All of those props will be set to _something_ but not necessarily on the first render cycle (because they're set via effects in the parent component). So we're making sure that the banner picker doesn't run until we've got all the inputs we need as we only want it to run once per page.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -0,0 +1,134 @@
...
+                tcfv2 && tcfv2.vendorConsents[brazeVendorId];
+
+            resolve(!!(consentGivenUnderCcpa || consentGivenUnderTcfv2));
+        });
+    });
+
+export const canShow = async (
+    asyncBrazeUuid: Promise<null | string>,
+    isDigitalSubscriber: undefined | boolean,
+): Promise<CanShowResult> => {
```
gtrufitt, line 37: I don't know if this is useful or not, but for me a comment that encompasses the following steps in code is useful for understanding and sense-checking everything is captured. 

```suggestion
): Promise<CanShowResult> => {
// We can show a Braze banner if:
// - The Braze switch is on
// - We have a braze API key
// - The user is not a digital subscriber
// - We've received a Braze UUID from the API, given a user's ID Creds
// - The user has given Consent via CCPA or TCFV2
// - The Braze websdk appboy intialisation does not throw an error
// - The Braze app Boy subscription to in app message returns meta info
```

gtrufitt (approved):
👍  Nice

liywjl (approved):
LGTM 👍

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -0,0 +1,134 @@
...
+};
+
+export const BrazeBanner = ({ meta }: Props) => {
+    const [ExampleComponent, setExampleComponent] = useState<
+        React.FC<BrazeBannerProps>
+    >();
+
+    useEffect(() => {
+        if (meta && meta['test-key']) {
+            // TODO: unify the way we handle sharing these deps (this is
```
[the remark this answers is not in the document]
tjmw: :+1: I agreed with Nic that we'd follow up on this in a separate PR (and make the same change in `SlotBodyEnd` too.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: Ah that's an interesting point. My approach usually is to name things in a way that makes the logic clear from the code (because comments get out of date). Perhaps adding this to the commit message would be useful 🤔

src/web/lib/getBrazeUuid.ts
```diff
@@ -0,0 +1,22 @@
+import { joinUrl } from '@root/src/web/lib/joinUrl';
+
+export const getBrazeUuid = async (ajaxUrl: string): Promise<string> => {
+    const url = joinUrl([ajaxUrl, 'user/me']);
+    return fetch(url, {
```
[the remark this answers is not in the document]
tjmw: Oh ha I "borrowed" my code from src/web/lib/getUser.ts. If it sounds okay I'll follow up and update both to use `callApi` in a separate PR? (I think both would benefit).

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -70,35 +81,77 @@ const buildReaderRevenueBannerConfig = (
...
-export const StickyBottomBanner = ({
+const buildBrazeBanner = (
+    asyncBrazeUuid: Promise<null | string>,
+    isDigitalSubscriber: undefined | boolean,
+): Banner => ({
+    id: 'braze-banner',
+    canShow: () => canShowBrazeBanner(asyncBrazeUuid, isDigitalSubscriber),
+    show: (meta: any) => () => <BrazeBanner meta={meta} />,
+    timeoutMillis: DEFAULT_BANNER_TIMEOUT_MILLIS,
+});
```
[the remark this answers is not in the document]
tjmw: Ah yeah it was more about giving something a meaningful name than re-usability, so I think I'll leave this as-is (also welcome back! 👋 )

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: Actually, having thought this over, because this list of conditions is quite long (and most are unlikely to change) I think that the comment is worth having :+1:

gtrufitt: > Actually, having thought this over, because this list of conditions is quite long (and most are unlikely to change) I think that the comment is worth having 👍

Yeah, I think that was my thinking, I had to go through quite a few steps to get to that point (it's not in one big if statement because of the early returns), so found it useful. But defo could go the other way if it gets out of date! Useful until it's not I guess 😆

PR #1910, opened 2020-09-14
Braze Banner -> 0.0.7 - support for DigitalSubscriberAppBanner and Braze click analytics (DCR edition)

rhystmills:
### Before
The ExampleComponent was used in DCR. There was no click tracking or impression tracking reaching Braze.

### After
DCR will now use the DigitalSubscriberAppBanner. Clicks and impressions are tracked in Braze. There is also a fix for some unintended space under the banner.

There is an equivalent PR in frontend [here](https://github.com/guardian/frontend/pull/22997).

## Why?
The new banner is an early version of a banner required by marketing, which will be sent to digital subscribers who have never signed in to the premium app, letting them know that they are eligible for this perk.

## Testing
The changes have been tested in a local instance of the website.

### DIFF
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -69,6 +69,7 @@ export const canShow = async (
         );
 
         appboy.initialize(apiKey, {
+            enableLogging: false,
             noCookies: true,
             baseUrl: 'https://sdk.fra-01.braze.eu/api/v3',
             sessionTimeoutInSeconds: 1,
@@ -76,12 +77,36 @@ export const canShow = async (
         });
 
         return new Promise((resolve) => {
-            appboy.subscribeToInAppMessage((message) => {
-                const meta = (message as any).extras;
+            appboy.subscribeToInAppMessage((message: any) => {
+                const meta = message.extras;
+
+                const buttonHandler = (buttonId: number) => {
+                    const thisButton = new appboy.InAppMessageButton(
+                        `Button: ID ${buttonId}`,
+                        undefined,
+                        undefined,
+                        undefined,
+                        undefined,
+                        undefined,
+                        buttonId,
+                    );
+                    appboy.logInAppMessageButtonClick(thisButton, message);
+                };
+
+                const logImpression = () => {
+                    // Log the impression with Braze
+                    appboy.logInAppMessageImpression(message);
+                };
 
                 if (meta) {
-                    resolve({ result: true, meta });
+                    const newMeta = {
+                        ...meta,
+                        logImpression,
+                        buttonHandler,
+                    };
+                    resolve({ result: true, meta: newMeta });
                 }
+
                 resolve({ result: false });
             });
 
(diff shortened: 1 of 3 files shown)


src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -61,15 +63,34 @@ export const canShow = async (
...
+                        undefined,
+                        buttonId,
+                    );
+                    appboy.logInAppMessageButtonClick(thisButton, message);
+                };
+
+                if (appboy) {
+                    // Log the impression with Braze
+                    appboy.logInAppMessageImpression(message);
+                }
```
tjmw, line 93: I think the right way to expose the click handler and the impression log functionality to the component is via the `meta` object which is threaded through to the component as a prop. So `meta` could be extended to included two additional fields:

```js
{
  logImpression: () => { appboy.logInAppMessageImpression(message) },
  buttonHandler: (buttonId: number) => {...},
}
```

In general having React components rely on dependencies which aren't passed as props (which is what we're doing with `buttonHandler` here) can cause confusing situations as they exist outside of React's lifecycle. It's definitely confusing though because this is how we thread things together on frontend (where we're not in React world).

Really happy to pair on this if it'd be useful!

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -61,15 +63,34 @@ export const canShow = async (
...
+                        undefined,
+                        undefined,
+                        buttonId,
+                    );
+                    appboy.logInAppMessageButtonClick(thisButton, message);
+                };
+
+                if (appboy) {
+                    // Log the impression with Braze
+                    appboy.logInAppMessageImpression(message);
```
tjmw, line 92: It's subtle but this will log the impression when the callback is fired, not when we actually show the banner component to the user, which is what we want. I think using the approach where this is threaded through using the `meta` object is the way to go.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -107,7 +128,7 @@ export const BrazeBanner = ({ meta }: Props) => {
                 /* webpackChunkName: "guardian-braze-components" */ '@guardian/braze-components'
             )
                 .then((module) => {
-                    setExampleComponent(() => module.ExampleComponent);
+                    setBrazeComponent(() => module.DigitalSubscriberAppBanner);
```
tjmw, line 131: :+1: Much better naming!

tjmw (commented):
This is looking good. I left a couple of suggestions around how we thread tracking callbacks to the component itself. Happy to pair if anything isn't clear!

src/web/components/StickyBottomBanner/BrazeBanner.tsx
rhystmills: Cheers Tom! I've pushed the fix we made for logImpression earlier today.

rhystmills: Also have added the `buttonHandler` to the meta prop.

tjmw (approved):
Looks good!

shtukas (approved):

PR #1911, opened 2020-09-14
Showcase opinion

liywjl:
## What does this change?
Support Showcase weighting of main media images

example showcase: https://www.theguardian.com/commentisfree/2020/aug/09/work-britain-industries-jobs-tory-manufacturing-investment

example non-showcase opinion: https://www.theguardian.com/commentisfree/2020/sep/14/covid-19-global-education-crisis-girls-impacted

### Before
Screenshot 2020-09-14 at 17 36 39

### After
Screenshot 2020-09-14 at 17 35 01

## Why?
Parity

## Note:
this does not apply if  `hasShowcaseMainElement` is false, to avoid having full width for showcase opinions (not sure this is the best approach)


src/web/layouts/CommentLayout.tsx
```diff
@@ -78,13 +90,17 @@ const StandardGrid = ({
...
                     grid-template-areas:
                         'title      border  headline    right-column'
                         'lines      border  headline    right-column'
                         'meta       border  standfirst  right-column'
                         'meta       border  media       right-column'
                         '.          border  body        right-column'
                         '.          border  .           right-column';
+
+                    /* Override grid-template-areas if showcase */
+                    ${display === Display.Showcase ? showcaseGrid : ''}
```
rcrphillips, line 103: I like this instead of the other conditional rendering we have in StandardLayout.tsx 

Just need to make sure we're covering all breakpoints?

liywjl: thanks, yeah I can throw up a PR to make those changes. Will test other breakpoints to make sure I havent missed anything

gtrufitt, line 103: Worth seperating into a standard grid and a showcaseGrid rather than rely on an override?

src/web/layouts/DecideLayout.tsx
```diff
@@ -73,6 +73,15 @@ export const DecideLayout = ({ CAPI, NAV }: Props) => {
         case Display.Showcase: {
             switch (designType) {
                 case 'Comment':
+                    return (
+                        <CommentLayout
```
gtrufitt, line 77: I don't think this is comment only, I believe any article layout can have showcase images. I.E: https://www.theguardian.com/science/grrlscientist/2012/aug/07/3

gtrufitt:
Looking at https://www.chromatic.com/snapshot?appId=5dfcbf3012392c0020e7140b&id=5f5f995982be0d00225eeb5d I think we want to adjust the most viewed positioning in the column based on whether show case main media or not - ideally we want the most viewed to be in a similar position from the top of the article (for a 5:3 image) as without a showcase image. At the moment most viewed has been shifted quite far down the article.

src/web/layouts/DecideLayout.tsx
gtrufitt: Not sure whether this was opinion specific in this instance?

src/web/layouts/CommentLayout.tsx
liywjl: agreed 👍

src/web/components/StickyAd.tsx
```diff
@@ -6,11 +6,12 @@ import { namedAdSlotParameters } from '@root/src/model/advertisement';
...
 type Props = {
     name: AdSlotType;
-    height: number;
+    height?: number;
 };
-const adSlotWrapper = (height: number) => css`
+const adSlotWrapper = ({ height }: { height?: number }) => css`
     position: static;
-    height: ${height}px;
+    min-height: 274px;
```
liywjl, line 13: to mimic frontend, I set a min-width instead of a width. This avoids having spacing between the right column ad and the most read column

liywjl: Note: local testing shows that this doesn't prevent with larger ad blocks from rendering

shtukas (approved):

PR #1914, opened 2020-09-15
Always add the index to the root, even when the index is 0 (falsy)

amyhughes:
## What does this change?
Here we fix a bug where we don't identify and hydrate elements where their index is 0, since 0 evaluates to false and is not appended to the `root` to give the expected `rootWithIndex`.

### Before
This QuizAtom element with an index of 0 was given the wrong `rootWithIndex` (missing the 0), meaning the element was not hydrated:

quiz-atom-not-hydrated

### After

With this change, the `rootWithIndex` was as expected, and the element was hydrated:

quiz-atom-hydrated

### DIFF
--- a/src/web/components/Hydrate.tsx
+++ b/src/web/components/Hydrate.tsx
@@ -7,7 +7,7 @@ type Props = {
 };
 
 export const Hydrate = ({ root, index, children }: Props) => {
-    const rootWithIndex = index ? `${root}-${index}` : root;
+    const rootWithIndex = index===0 || index ? `${root}-${index}` : root;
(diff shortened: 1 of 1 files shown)


liywjl (approved):
👍

gtrufitt (approved):
Sigh... 😆

nitro-marky (approved):
🎉

liywjl:
a cool alternative work around https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

PR #1915, opened 2020-09-15
Add end-to-end banner tests

tjmw:
## What does this change?

Adds two very basic e2e banner tests:

* ensure that when consent cookies aren't set we show the CMP
* ensure that we make a single request to the reader revenue /banner endpoint

## Why?

Additional confidence in the banner picking logic.

### DIFF
--- a/cypress/integration/e2e/article.bannerPicker.e2e.spec.js
+++ b/cypress/integration/e2e/article.bannerPicker.e2e.spec.js
@@ -1,0 +1,45 @@
+/* eslint-disable no-undef */
+/* eslint-disable func-names */
+
+describe('Banner Picker Integration', function () {
+    const visitArticle = (
+        url = 'https://www.theguardian.com/games/2018/aug/23/nier-automata-yoko-taro-interview',
+    ) => {
+        cy.visit(`Article?url=${url}`);
+    };
+
+    const cmpIframe = () => {
+        return cy
+            .get('iframe#sp_message_iframe_208529')
+            .its('0.contentDocument.body')
+            .should('not.be.empty')
+            .then(cy.wrap);
+    };
+
+    describe('When consent cookies are not set', function () {
+        it('shows the CMP', function () {
+            cy.clearCookie('consentUUID');
+            cy.clearCookie('euconsent-v2');
+
+            visitArticle();
+
+            cmpIframe().contains("It's your choice");
+        });
+    });
+
+    it('makes a single request to the banner service', function () {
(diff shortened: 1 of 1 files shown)


cypress/integration/e2e/article.bannerPicker.e2e.spec.js
```diff
@@ -0,0 +1,44 @@
...
+describe('Banner Picker Integration', function () {
+    const visitArticle = (
+        url = 'https://www.theguardian.com/games/2018/aug/23/nier-automata-yoko-taro-interview',
+    ) => {
+        cy.visit(`Article?url=${url}`);
+    };
+
+    const cmpIframe = () => {
+        return cy
+            .get('iframe#sp_message_iframe_208529')
```
tjmw, line 13: It seemed to me that the presence of this iframe was a good indicator of the CMP being shown. Does someone more familiar with the CMP know if that's the case? I tried matching on content in the CMP, and while I could see it in the Cypress browser instance, the match was failing (it seems like iframe support is a little tricky in Cypress).

liywjl: Iframes and testing is a nightmare, so I think this is a good work around althought I am not sure the ID `sp_message_iframe_208529` will stay they same?

Maybe add a comment to explain why you used this work around?

tjmw: Yeah great question about the ID. I'll see if I can get someone from commercial dev to take a look.

gtrufitt: You may have seen it but there's some digging into iframes here for Cypress https://github.com/guardian/dotcom-rendering/blob/main/cypress/integration/e2e/article.elements.spec.js#L5

tjmw: Ah nice. That looks pretty similar to what I'm doing afaict (I was looking at the same blog post reference in the comment). Interesting that chaining onto the returned cypress wrapped element did seem to work when querying for content in that case.

tjmw: @gtrufitt I revisited this and making a small change to match the example you linked to has got things working, so thanks for that! I'm now able to test for content inside the CMP iframe, which I think is a better test (since it's closer to what a user actually sees).

gtrufitt (approved):

PR #1917, opened 2020-09-16
Refactor getUser and getBrazeUuid to use callApi

tjmw:
### What does this change?

Refactor `getUser` and `getBrazeUuid` to use `callApi` instead of using `fetch` directly.

### Why?

Using `callApi` has the benefit of being able to remove some boilerplate (error handling, json decoding) from the two call sites.


liywjl (approved):
👍

tjmw:
Thanks @liywjl!

PR #1918, opened 2020-09-16
Add index entry for YoutubeBlockElement

shtukas:
## What does this change?

Add index entry for `YoutubeBlockElement`. (This particular video comes with an expired flag, but it's ok, for the moment, for index purposes).


liywjl (approved):
Nice! 👍

PR #1925, opened 2020-09-18
Rich link text color

liywjl:
# What does this change? 
Set rich link text colours

### Before
Screenshot 2020-09-18 at 13 00 21

### After
Screenshot 2020-09-18 at 12 55 53

## Why?
We should not have pallet colours applied to the text on rich links


rcrphillips (approved):
As long as @HarryFischer approves!

PR #1929, opened 2020-09-22
Fix minimum height problem

buck06191:
## What does this change?
Fixes email sign up being cut off following changes to the iframe on Frontend

### Before
image

### After
image

## Why?

The iframe was updated on Frontend and this change is needed to reflect that.


liywjl (approved):
LGTM

PR #1930, opened 2020-09-22
Carousel / onwards test

nicl:
Questions:

* do we want to add some scroll behaviour on non-mobile? I.e. show scrollbar, or translate vertical scroll to horizontal, or support drag+drop (these are doable but require more effort for the last in particular)
* do we want to make the nav arrows larger for the larger breakpoints?
* text vertical position - ideal is for first two lines to sit over image and then expand downwards after that. I'll try and add that with some CSS wizardry!

## What does this change?

Adds a Carousel component and updates the (not yet live) curated content test to include this as a C variant.

### Before

n/a

### After

Kapture 2020-09-22 at 14 58 41
Kapture 2020-09-22 at 15 00 56

## Why?

As part on our curated content test.

### DIFF
--- a/src/web/components/Onwards/Onwards.test.tsx
+++ b/src/web/components/Onwards/Onwards.test.tsx
@@ -19,7 +19,8 @@ describe('OnwardsLayout', () => {
         useApi.mockReturnValue(response);
 
         const { asFragment } = render(
-            <OnwardsLayout onwardSections={[linkAndDescription]} />,
+            // eslint-disable-next-line react/jsx-props-no-spreading
+            <OnwardsLayout {...linkAndDescription} />,
         );
 
         // Renders data-component
--- a/src/web/components/Onwards/OnwardsUpper.tsx
+++ b/src/web/components/Onwards/OnwardsUpper.tsx
@@ -70,6 +70,19 @@ const onwardsWrapper = css`
     width: 100%;
 `;
 
+const headlinesContainer = (edition: Edition): string => {
+    switch (edition) {
+        case 'UK':
+            return 'uk-alpha/news/regular-stories';
+        case 'US':
+            return 'c5cad9ee-584d-4e85-85cd-bf8ee481b026';
+        case 'AU':
+            return 'au-alpha/news/regular-stories';
+        case 'INT':
+            return '10f21d96-18f6-426f-821b-19df55dfb831';
+    }
+};
+
 type Props = {
     ajaxUrl: string;
     hasRelated: boolean;
@@ -81,6 +94,7 @@ type Props = {
     keywordIds: string | string[];
     contentType: string;
     tags: TagType[];
+    edition: Edition;
 };
 
 export const OnwardsUpper = ({
@@ -94,6 +108,7 @@ export const OnwardsUpper = ({
     keywordIds,
     contentType,
     tags,
+    edition,
 }: Props) => {
     const dontShowRelatedContent = !showRelatedContent || !hasRelated;
 
@@ -167,7 +182,8 @@ export const OnwardsUpper = ({
     const ABTestAPI = useAB();
     const headlinesDataUrl = joinUrl([
         ajaxUrl,
-        '/container/data/uk-alpha/news/regular-stories.json',
+        'container/data',
+        `${headlinesContainer(edition)}.json`,
     ]);
 
     const inCuratedContainerTest = ABTestAPI.isUserInVariant(
@@ -175,11 +191,11 @@ export const OnwardsUpper = ({
         'fixed',
     );
 
-    const inCuratedCarouselTest = true; // ABTestAPI.isUserInVariant(
-    /*         'CuratedContainerTest',
+    const inCuratedCarouselTest = ABTestAPI.isUserInVariant(
+        'CuratedContainerTest',
         'carousel',
     );
- */
+
     return (
         <div className={onwardsWrapper}>
             {inCuratedCarouselTest && (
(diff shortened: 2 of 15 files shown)


HarryFischer:
Thanks Nic - looking good, couple small bits:

- I think the container title should be either just 'Headlines' or 'From Headlines'
- On mobile can we align the 'Headlines' title and the pagination dots with the lefthand side of the first card, would also make the arrows a little bigger and some padding between them so easy to tap
- Re desktop navigation, I think a scroll bar is a bit ugly and the scroll thing might confuse people. It would be amazing if we can support drag&drop but if not then I would just make the arrows larger and maybe add some space between them so they're easy to click

Let me know if any of that doesn't make sense

src/web/components/Onwards/Onwards.test.tsx
```diff
@@ -19,7 +19,8 @@ describe('OnwardsLayout', () => {
         useApi.mockReturnValue(response);
 
         const { asFragment } = render(
-            <OnwardsLayout onwardSections={[linkAndDescription]} />,
+            // eslint-disable-next-line react/jsx-props-no-spreading
+            <OnwardsLayout {...linkAndDescription} />,
```
liywjl, line 23: more of a preference, but I like to avoid using spread operator when possible (same for the stories)

liywjl (approved):
The code looks great 👍 

I was wondering if having some sort of logic to `scrollTo` the next element might be nicer for when clicking the next/previous buttons? Maybe thats something to get @HarryFischer 's advise on

src/web/components/Onwards/OnwardsUpper.tsx
```diff
@@ -165,23 +172,38 @@ export const OnwardsUpper = ({
...
         'CuratedContainerTest',
-        'variant',
+        'fixed',
     );
 
+    const inCuratedCarouselTest = true; // ABTestAPI.isUserInVariant(
+    /*         'CuratedContainerTest',
+        'carousel',
+    );
+ */
```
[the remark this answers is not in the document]
HarryFischer: Yeah scrollTo would be nice!

paperboyo:
Cool stuff!

> do we want to add some scroll behaviour on non-mobile? I.e. show scrollbar, or translate vertical scroll to horizontal, or support drag+drop (these are doable but require more effort for the last in particular)

These are excellent questions, but they apply to live (for years) main carousel too (https://www.theguardian.com/uk#video) and we should be consistent, so… (my humble opinion would be the last one in particular on every carousel we got ;-)

> I think the container title should be either just 'Headlines' or 'From Headlines'
IMHO _from_ is useful for context. Live apps support swiping between stories from containers for years, but it wasn’t ever very clear what’s the context (where do the stories come from)

For consistency, I would normally suggest arrows in circle, but I don’t like them looking like that there, so I won’t ;-)

In your screenshots, images do not look 5:3? Are they 5:3 (should be!)? Maybe a touch of transparency on those headline overlays, because we will cover something important no way we won’t (video editors jump through all sorts of hoops to make sure main video carousel doesn’t cover nothing, resorting even to flipping a picture!). In general: covering images is always dangerous, just like blind cropping/changing ratio. But re:overlays, I will obviously defer to @HarryFischer’s opinion!

nicl:
@paperboyo thanks for the feedback. I agree on not adding mouse scroll to the carousel, partly for consistency as you say and also for ease of implementation!

Image formats are currently consistent with our existing below-article onwards - which is 5:3 yep!

Re image covering: am going to move the overlap down slightly so as to not cover so much of the image, and also ensure it expands downwards if there is more text (currently it expands upwards). These should help a bit, but agree overlay always can potentially obscure useful stuff.

Re arrows in a circle, I will defer to @HarryFischer !

nicl:
@HarryFischer thanks. 

* title is easy to change
* as is alignment - though, the heading for our other onwards is indented at this break point so worth noting the two will be inconsistent if we change this (perhaps okay though)
* re: arrows agree, will make them larger and update screenshots.

Thanks!

nicl:
Updated gif is here:

Kapture 2020-09-23 at 15 52 41

cc @paperboyo @HarryFischer and @rcrphillips .

rcrphillips:
Lovely!

paperboyo:
Looks cool! Eager to have a play. Some things come to mind (would this get in proper): rubber-banding, being able to click on the half-visible one to move to it etc…

HarryFischer:
Looks great! thanks for looking at those tweaks

src/web/components/Onwards/Onwards.test.tsx
nicl: Yeah, I generally agree. I wanted `Carousel` and `OnwardsLayout` to have the same interface so that the calling code could treat them generically. The simplest interface was just to accept an `OnwardsType`, which has a lot of fields. Is the motivation here that avoiding spread is more explicit about dependencies?

nicl:
@liywjl meant to say, thanks for the feedback! It's not too hard to add basic scroll-to behaviour with `scroll-behavior: smooth` at least. But it felt a bit sluggish when I tried it - you have to wait for the animation to complete before being able to click next/prev again. And I was reluctant to add some JS scroll animation. Maybe we'll realise in the testing that we do want this and it's worth the effort.

src/web/components/Onwards/Onwards.test.tsx
liywjl: yeah, its more that I want to discourage the use of spread operator in general for react components, as its can make it a little more challenging to follow the props when debugging. Also future developers may look at `test/storybook` code to see how to implement things; and this could invite them to use this approaches. Thats why I would personally not have used spread. But again this is just preference.

PR #1931, opened 2020-09-23
Use image sources from frontend directly

jfsoul:


shtukas (approved):

jfsoul:
This wasn't the right approach

PR #1935, opened 2020-09-24
Bump @guardian/braze-components to 0.0.10

tjmw:
## What does this change?

This PR bumps @guardian/braze-components to 0.0.10 which gives a few changes:

* We can now use the new `BrazeMessage` component which takes a `componentName` prop provided by Braze
* 0.0.10 is published as cjs and esm, which seems to fix a linting issue we were seeing before
* The buttons in the underlying banner component are now wired correctly

Here it is working locally:

Screenshot 2020-09-24 at 13 02 15

This is the DCR equivalent of guardian/frontend#23021 and guardian/frontend#23028 rolled into one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
         "@guardian/atoms-rendering": "^1.11.2",
         "@guardian/automat-client": "^0.2.16",
         "@guardian/consent-management-platform": "6.0.0",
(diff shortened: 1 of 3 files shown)


liywjl (approved):
👍

PR #1942, opened 2020-09-25
Use minimum that's a multiple of 3

jfsoul:
## What does this change?
Changes our ASG min size to 9 from 8 so that we always have balance between the 3 AZs.

## Why?
There have been occasional issues when deploys happen very close together that we have attributed to automatic rebalancing done by AWS (that they have confirmed can't be controlled by us).

### DIFF
--- a/cloudformation.yml
+++ b/cloudformation.yml
@@ -60,8 +60,8 @@ Mappings:
       Value: rendering
   StageMap:
     PROD:
-      MinCapacity: 6
-      MaxCapacity: 24
+      MinCapacity: 9
+      MaxCapacity: 36
     CODE:
       MinCapacity: 1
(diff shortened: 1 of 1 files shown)


liywjl (approved):
👍

shtukas (approved):

shtukas:
Interesting 🤔

nicl:
Worth checking the instance type and CPU credits monitoring as for some of the burstable instances even a lowish CPU can indicate exceeding the burst credits and end up costing more than you want.

jfsoul:
Thanks @nicl that's a great point. Every time I come back to these instances I forget that 🤦, but it is absolutely the case here. Based on previous experience I am now thinking we should go to 9 instead of 6.

PR #1944, opened 2020-09-25
Send VIEW event to Ophan for Braze banner

tjmw:
## What does this change?

Start sending Braze banner view events to Ophan.

## Why?

To get an accurate idea of how often we're showing the banner, which is useful feedback to understand how the campaign is performing (and whether it's configured correctly).

### DIFF
--- a/src/web/browser/ophan/ophan.ts
+++ b/src/web/browser/ophan/ophan.ts
@@ -54,7 +54,8 @@ export type OphanComponentType =
     | 'ACQUISITIONS_EDITORIAL_LINK'
     | 'ACQUISITIONS_SUBSCRIPTIONS_BANNER'
     | 'ACQUISITIONS_OTHER'
-    | 'SIGN_IN_GATE';
+    | 'SIGN_IN_GATE'
+    | 'RETENTION_ENGAGEMENT_BANNER';
 
 export type OphanComponent = {
     componentType: OphanComponentType;
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -5,6 +5,7 @@ import * as emotionTheming from 'emotion-theming';
 import { onConsentChange } from '@guardian/consent-management-platform';
 import { getZIndex } from '@root/src/web/lib/getZIndex';
 import { Props as BrazeBannerProps } from '@guardian/braze-components';
+import { submitComponentEvent } from '@root/src/web/browser/ophan/ophan';
 import { CanShowResult } from './bannerPicker';
 
 export const brazeVendorId = '5ed8c49c4b8ce4571c7ad801';
@@ -152,7 +153,17 @@ const BrazeBannerWithSatisfiedDependencies = ({
     meta,
 }: InnerProps) => {
     useEffect(() => {
+        // Log the impression with Braze
         meta.logImpression();
+
+        // Log VIEW event with Ophan
+        submitComponentEvent({
+            component: {
+                componentType: 'RETENTION_ENGAGEMENT_BANNER',
+                id: meta.dataFromBraze.componentName,
+            },
+            action: 'VIEW',
+        });
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 


src/web/browser/ophan/ophan.ts
```diff
@@ -54,7 +54,8 @@ export type OphanComponentType =
     | 'ACQUISITIONS_EDITORIAL_LINK'
     | 'ACQUISITIONS_SUBSCRIPTIONS_BANNER'
     | 'ACQUISITIONS_OTHER'
-    | 'SIGN_IN_GATE';
+    | 'SIGN_IN_GATE'
+    | 'BRAZE_BANNER';
```
tjmw, line 58: Is creating a new component type as simple as defining it here? Does it need to be added somewhere else first?

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -152,7 +153,17 @@ const BrazeBannerWithSatisfiedDependencies = ({
...
         meta.logImpression();
+
+        // Log VIEW event with Ophan
+        submitComponentEvent({
+            component: {
+                componentType: 'BRAZE_BANNER',
+                id: meta.dataFromBraze.componentName,
+            },
+            action: 'VIEW',
+        });
```
tjmw, line 166: According to the types, this is the minimum data we need to provide to `submitComponentOphanEvent`. Is there other data we should be sending?

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -152,7 +153,17 @@ const BrazeBannerWithSatisfiedDependencies = ({
...
 }: InnerProps) => {
     useEffect(() => {
+        // Log the impression with Braze
         meta.logImpression();
+
+        // Log VIEW event with Ophan
+        submitComponentEvent({
+            component: {
+                componentType: 'BRAZE_BANNER',
+                id: meta.dataFromBraze.componentName,
```
tjmw, line 163: For now this will only ever be one value: `DigitalSubscriberAppBanner` since that's the only component we currently support.

src/web/browser/ophan/ophan.ts
paulbrown1982: You might need to add this to the Thrift model, if you do you then have to release it to 5+ other apps so I'd like to add another one in there for Contributions if you do. cc @tomrf1

src/web/components/StickyBottomBanner/BrazeBanner.tsx
rtyley: > According to the types, this is the minimum data we need to provide to `submitComponentOphanEvent`. Is there other data we should be sending?

Just to be clear, you're mentioning `submitComponentOphanEvent` here but using `submitComponentEvent` above:

https://github.com/guardian/dotcom-rendering/blob/adf3a321e24a369ed496e6bfd83da212a0816694/src/web/components/StickyBottomBanner/BrazeBanner.tsx#L160

..and they are two different functions (tho `submitComponentOphanEvent` calls `submitComponentEvent`):

https://github.com/guardian/dotcom-rendering/blob/ab86bc6f16d28ad703465e635af1dedb242f124a/src/web/browser/ophan/ophan.ts#L99-L132

Anyway - I _think_ what you have is indeed enough information - you can check out [`extractors.ComponentEventExtractor.ComponentRepr`](https://github.com/guardian/ophan/blob/de959dbe4b506d97e56181c0688d0e6d94446744/the-slab/app/extractors/ComponentEventExtractor.scala#L14-L20) in Ophan for the definitive word on what you can send, with some examples in the [`ComponentEventExtractorTest`](https://github.com/guardian/ophan/blob/de959dbe4b506d97e56181c0688d0e6d94446744/the-slab/test/extractors/ComponentEventExtractorTest.scala).

src/web/browser/ophan/ophan.ts
tomrf1: The enum in the ophan model will need to be updated before it can receive these events, but they probably do not need to release the thrift model to maven (unless something else depends on that)

rtyley: Yep - and note that within [the Ophan stack, the Slab and everything downstream from it](https://github.com/guardian/ophan/blob/master/docs/how-ophan-works/ophan-stack.md) will need to be deployed with the update before Ophan will recognise the new value.

tjmw: This is good to know, thanks all :+1:

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: Thanks @rtyley that's really useful, I'll check out the links. Ah sorry yeah that's a mistake on my end (the names are v similar!), I meant `submitComponentEvent`. Like you said, `submitOphanComponentEvent` calls `submitComponentEvent` under the hood. It looks like the former is there to make the call more convenient (presumably it's common to send the params expected by `sendOphanComponentEvent`?)

adamnfish (approved):

liywjl (approved):

PR #1945, opened 2020-09-28
Hide trailing carousel dots on wider breakpoints

nicl:
## What does this change?

Hide trailing dots on wider breakpoints.

### Before
(Even with scrolling to the end, the right two dots will never become active.)
Screenshot 2020-09-28 at 11 31 13

### After

(Fewer dots, but all can be made active with scrolling.)
Screenshot 2020-09-28 at 11 23 50

## Why?

The purpose here is to not show dots that will never be made active through scrolling - as 'active' is defined by whatever is at the midpoint of the container.

This 'solution' is quite fragile; we'll need to think of a better approach to the UX here in time.

### DIFF
--- a/src/web/components/Onwards/Carousel/Carousel.tsx
+++ b/src/web/components/Onwards/Carousel/Carousel.tsx
@@ -161,17 +161,27 @@ const dotsStyle = css`
     }
 `;
 
-const dotStyle = css`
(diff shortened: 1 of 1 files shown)


rcrphillips (approved):
LGTM

PR #1946, opened 2020-09-28
Unsafe embed block elements

liywjl:
## What does this change?
Add support for Unsafe embed block elements

### After
example https://www.theguardian.com/us-news/2016/sep/10/cia-senate-investigation-constitutional-crisis-daniel-jones
NOTE: example has issue also on frontend, but is used as a baseline for compatibility

Screenshot 2020-09-28 at 11 46 27

 
## Why?
Parity


src/model/filterUnsafeBlockElements.ts
```diff
@@ -0,0 +1,32 @@
+export const filterUnsafeBlockElements = (data: CAPIType): CAPIType => {
+    const elements: CAPIElement[] = data.blocks[0].elements.reduce(
+        (acc: CAPIElement[], cur: CAPIElement) => {
+            if (
+                cur._type ===
+                    'model.dotcomrendering.pageElements.EmbedBlockElement' &&
+                !cur.safe
+            ) {
```
liywjl, line 8: we could explore `frontend` doing this for us, rather us having to filter out these elements ourself

shtukas: +1

shtukas (approved):

src/web/components/elements/UnsafeEmbedBlockComponent.tsx
```diff
@@ -0,0 +1,51 @@
...
+        if (numberOfIntervals >= 12) return;
+
+        const timer = setInterval(() => {
+            const doc =
+                iFrameRef.current &&
+                iFrameRef.current.contentWindow &&
+                iFrameRef.current.contentWindow.document;
+
+            numberOfIntervals += 1;
+
```
gtrufitt, line 24: I wonder if we could align the DCR resizing contract here. We already have resizing for atoms using iframeMessenger and postMessage with a listener on DCR. We could remove this different way of doing it and align on one approach for iframes as you're building the iframe yourself?

Then we could use this https://github.com/guardian/dotcom-rendering/blob/55ccdbe592a5895133b67234ea18fb81f1467e0c/src/web/browser/atomIframe/init.ts#L10 and extend it out from just `atom-iframe`, to fire `set-height` from the iframe.

You'd have to add it in the inner iframe using the `<gu-script>` mechanism as there's three steps to it:

- Page loads, all iframes on the page are rendered
- DCR starts listening for postmessage set-height from recognised iframes (Currently atom-iframe but I think we should extend that)
- DCR runs over all recognised iframes to convert `<gu-script>` to `script` tags, those contain postMessage `set-height` messages

You would need to implement something that `iframeMessenger` within the iframe https://github.com/guardian/interactive-us-campaign-finance-charts/blob/f746ae26e1de87c6c1f2f2bfa58df5e37018a34d/rich-zips-table/src/js/iframeMessenger.js

I guess best thing to do would be to inline that script in the iframe?

You could then avoid the hydration I think because you're using the existing mechanism and avoiding multiple ways for setting heights on iframes

gtrufitt (dismissed):
This is great, I think we can diverge from the parity in this case though by aligning on one approach for iframe resizing (based on what we use for atoms already), what do you think?

src/model/filterUnsafeBlockElements.ts
jfsoul: I think I'm missing something - why create a new BlockElement type rather than just branching `EmbedBlockElement`  react component into `SafeEmbedBlockElement` and `UnsafeEmbedBlockElement`?

liywjl: good point, this was because the workaround I used to set the high was done using react hooks, and in order to get hooks to work we needed to hydrate that component on the client. Therefore I thought i would be simpler to split out `SafeEmbedBlockElement`. Though the changes @gtrufitt recommended means we wont need to do this

src/web/components/elements/UnsafeEmbedBlockComponent.tsx
```diff
@@ -0,0 +1,21 @@
...
+const fullWidthStyles = css`
+    width: 100%;
+`;
+export const UnsafeEmbedBlockComponent = ({ html, alt }: Props) => (
+    <iframe
+        className={`${fullWidthStyles} atom__iframe`}
+        title={alt}
+        data-cy="embed-block"
+        srcDoc={`<script src="https://interactive.guim.co.uk/libs/iframe-messenger/iframeMessenger.js"></script>
```
gtrufitt, line 17: This won't be initalised in the platform yet, see my comment around gu-script here: https://github.com/guardian/dotcom-rendering/pull/1946#discussion_r495889625

src/web/components/elements/UnsafeEmbedBlockComponent.tsx
liywjl: > You would need to implement something that iframeMessenger within the iframe
I assumed that the HTML code came from frontend, so would we need to add `iframeMessenger.js` within the frontend returned HTML?

src/web/components/elements/UnsafeEmbedBlockComponent.tsx
```diff
@@ -0,0 +1,23 @@
...
+const fullWidthStyles = css`
+    width: 100%;
+`;
+export const UnsafeEmbedBlockComponent = ({ html, alt, index }: Props) => (
+    <iframe
+        className={`${fullWidthStyles} atom__iframe`}
+        title={alt}
+        name={`unsafe-embed-${index}`}
+        data-cy="embed-block"
+        srcDoc={`<script src="https://interactive.guim.co.uk/libs/iframe-messenger/iframeMessenger.js"></script>
```
gtrufitt, line 19: I have a feeling you want this to be after the html, to avoid blocking?

src/web/lib/ArticleRenderer.tsx
```diff
@@ -46,6 +49,20 @@ const commercialPosition = css`
     position: relative;
 `;
 
+const unsafeEmbedWrapperStyles = css`
```
gtrufitt, line 52: Is this an odd place for these styles? Seems very high in the render, can the figure not be inside the `UnsafeEmbeBlockComponent`?

src/web/lib/ArticleRenderer.tsx
```diff
@@ -46,6 +49,20 @@ const commercialPosition = css`
...
+    position: relative;
+    float: left;
+    margin-top: 6px;
```
gtrufitt, line 55: Probably want to use the space variables from the design system?

src/web/lib/ArticleRenderer.tsx
```diff
@@ -46,6 +49,20 @@ const commercialPosition = css`
...
+    margin-bottom: 12px;
+    margin-right: 20px;
+    ${from.leftCol} {
+        margin-left: -100px;
```
gtrufitt, line 59: Are we sure we want to do this for every one? Also with the width below?

src/web/components/elements/UnsafeEmbedBlockComponent.tsx
```diff
@@ -0,0 +1,23 @@
...
+};
+
+const fullWidthStyles = css`
+    width: 100%;
+`;
+export const UnsafeEmbedBlockComponent = ({ html, alt, index }: Props) => (
+    <iframe
+        className={`${fullWidthStyles} atom__iframe`}
+        title={alt}
+        name={`unsafe-embed-${index}`}
```
gtrufitt, line 17: Can you add a comment for what this is used for? Tripped me up for a sec, but it's the 'contract' required for the iframe script, right? Along with the `atom__iframe` class - if you can add a comment so that that's clear for the future, that'd be great!

src/web/browser/atomIframe/atomIframe.ts
```diff
@@ -0,0 +1,37 @@
+export const atomIframe = (): Promise<void> => {
+    const iframes: HTMLIFrameElement[] = [].slice.call(
+        document.querySelectorAll('.atom__iframe'),
```
gtrufitt, line 3: I wonder if we should add a separate class for the embeds, just to make sure we keep 'atoms' and 'embeds' distinct? It'd be great for that class to be 'js-' prefixed too, so we know it's *just* for JS behaviour (it should be that for atom__iframe but possible that ship sailed!)

gtrufitt (changes requested):

src/web/components/elements/UnsafeEmbedBlockComponent.tsx
jfsoul: Have been thinking we should have a docs directory called `contracts` with all the basic info on these types of things.

liywjl: you need a unique identifying name for the iFrame, it is used in the script here: https://github.com/guardian/dotcom-rendering/pull/1946/files#diff-e500d36de993a83661c36277af8605152cc33bd2aa0d03b7f5b1563655c4d805R10

src/web/browser/atomIframe/atomIframe.ts
liywjl: I agree with the `js-` prefix but I am scared it might be too late, that this may have knock on effects elsewhere.

I am not sure I see the benefit between the separation between `atoms` and `embeds`. This scrip just cares about resizing iFrames so I think `atom__iframe` is ok (although granted it would be better to highlight the action of resizing/JS execution in the name)

src/web/components/elements/UnsafeEmbedBlockComponent.tsx
gtrufitt: I know yeah, I meant a comment in the code for future people :D

src/web/browser/atomIframe/atomIframe.ts
gtrufitt: It isn't an `atom__iframe` though. We may decide in the future that we change the mechanism of embeds or atoms, and therefore by separating them now, we're making it easier to make that change. I think we should add a new class `js-embed__iframe` and use that

src/web/lib/ArticleRenderer.tsx
liywjl: I have had a look at the SCSS in frontend, it uses a bunch of calculations to determine width/margin etc...

in this case particularly : `margin-left: -1 * ($left-column + $gs-gutter);`


**Global variables**
`$gs-gutter: 20px;`
`$gs-column-width: 60px;`

**$left-column**: `gs-span(2)` = `$n-columns * $gs-column-width + $gs-gutter * ($n-columns - 1)` = `2 * 60px + 20px * (2 - 1)` = `140px`

**$gs-gutter**: `20px`

After having stepped through the SCSS logic, I think it is a lot simpler to just have the exact values defined rather than using the methods defined as the calculations will always resolve to the same numbers.

src/web/browser/atomIframe/atomIframe.ts
gtrufitt: Could we do this then as a part of this PR, and add the class @buck06191 needs too as part of this, so we can close https://github.com/guardian/dotcom-rendering/pull/1980 ?

liywjl: done

gtrufitt (approved):

PR #1947, opened 2020-09-28
Reduce carousel text overal to single line

nicl:
(Previously we had two lines of text overap the image for each card.)

Screenshot 2020-09-28 at 12 24 28
Screenshot 2020-09-28 at 12 24 46


rcrphillips (approved):
👍

PR #1949, opened 2020-09-28
Identity | Sign In Gate | Fire customevent to re-init commercial code when gate dismissed

coldlink:
## What does this change?
- Adds `useEffect` hook that fires once the sign in gate has been dismissed by the user (`showGate` set to `false`) that emits an `dcr:page:article:redisplayed` custom event to reinitialse the inline article body ads in case they failed to initalise before the gate had been dismissed

## Why?
- Inline article body ads may not have displayed previously due to a race condition as the ad initialisation competes with the sign in gate


gtrufitt (approved):
Great!

PR #1950, opened 2020-09-28
Enable curated container test

nicl:
...initially at 1% to verify data is coming through correctly and no major issues. We'll then increase.


rcrphillips (approved):
:shipit:

PR #1951, opened 2020-09-28
Added Ophan click tracking for Braze Banner

rhystmills:
## What does this change?

Adds Ophan click tracking for Braze Banner to DCR.

Equivalent to https://github.com/guardian/frontend/pull/23043

Depends on guardian/braze-components#17 (already reviewed/merged).

## Why?

More visibility and flexibility than Braze tracking - providing more confidence in our analytics


tjmw:
I've pushed some additional commits to this repo, which move towards an approach where more work is done in the component. The platform now provides an implementation of `submitComponentEvent` and the component does the rest. This reduces duplication and makes things easier to test. This PR relies on guardian/braze-components#17, which in turn relies on a changes in `@guardian/types`. The build failures are expected (locally I've been `yarn link`ing to the updated `braze-components`, but we need to do a release and bump the version used here).

shtukas (approved):

PR #1953, opened 2020-09-30
Use shared Ophan types from @guardian/types

tjmw:
## What does this change?

Use shared Ophan types from `@guardian/types` (see guardian/types#24).

~~(Note that currently this branch is getting `@guardian/types` from a GH branch. If we're happy with the changes I'll release a new version of the types package and bump the version in this project).~~

### Before

These types are defined in DCR directly (and duplicated across a few projects).

### After

Import the types from `@guardian/types`.

## Why?

Reduced duplication and increase ease of sharing the types across TypeScript projects.


yarn.lock
```diff
@@ -2436,10 +2436,9 @@
     "@guardian/src-helpers" "^1.5.0"
     "@guardian/src-icons" "^1.5.0"
 
-"@guardian/types@^0.4.4":
+"@guardian/types@guardian/types#tw-add-ophan-types":
   version "0.4.4"
-  resolved "https://registry.yarnpkg.com/@guardian/types/-/types-0.4.4.tgz#dba5c1ef0e95cfe9b8411c21354a528cba7b338c"
-  integrity sha512-kBnfpMouvhheuEXaWLFEDQRv2Lyx1wbrz/4IidSK04RqALAMdrsZiXRLHIApUDcgohgEGYp9Jwquml0Vvp7o5Q==
+  resolved "https://codeload.github.com/guardian/types/tar.gz/3d9e1c37c5acbb572dbf7561fea5653340b5c120"
```
JamieB-gu, line 2441: Oh, this is new. How does this `tar.gz` URL work? 🤔

Apps-rendering also loads `guardian/types` [from a GH URL](https://github.com/guardian/apps-rendering/blob/main/package-lock.json), but it's of the form `github:guardian/types#<commit>`. This is achieved using `semver` together with a GH URL, as seen [here](https://github.com/guardian/apps-rendering/blob/51a78cbb2e1ae3440df7d56c0902a6f9179d3d01/package.json#L50).

tjmw: Good question 😄 I'd not used it before. The format used in `package.json` looks similar to the example you've linked to (maybe omitting `github:` works if `npm` is able to infer it from the format). That URL is a link to a tar.gz for the repo at the commit specified. I guess both approaches work because there aren't any build steps for `@guardian/types`?

Just to be clear though, I'm not planning on leaving it like this. If the changes look okay then I'd plan on releasing a new version of `@guardian/types` and switching back to npm/bumping the version.

JamieB-gu: > That URL is a link to a tar.gz for the repo at the commit specified

Ah that makes sense. Yeah perhaps it varies a bit depending on exactly what command you use to add the dependency - I assume you used `yarn` here?

> because there aren't any build steps for @guardian/types

There aren't, no, although I'd love to figure out a straightforward way to include GH dependencies that _do_ have build steps.

tjmw: > I assume you used yarn here?

Yeah exactly. Specifically I used `yarn add -D guardian/types#tw-add-ophan-types` (so `<org>/<repo>#<ref e.g. branch or commit sha>`).

> I'd love to figure out a straightforward way to include GH dependencies that do have build steps.

Yep that'd be great, given that we're standardising on targeting ES2020 for @guardian packages. Related- do you know if anyone at the guardian has played with [using GitHub as a package registry](https://docs.github.com/en/free-pro-team@latest/packages/using-github-packages-with-your-projects-ecosystem/configuring-npm-for-use-with-github-packages)?

JamieB-gu: Yeah I believe Oliver tested it out. I think he ran into a couple of difficulties with using it in parallel with NPM:

1. Namespace clash - `guardian` on NPM vs `guardian` on GitHub
2. Having to change your package registry to the GitHub one

Don't know if things have changed since then.

gtrufitt: > I think he ran into a couple of difficulties with using it in parallel with NPM:

Indeed, I think the conclusion we came to at the time was that if we wanted to use Github registry, the DevEx would be nasty unless we moved *all* the modules over.

Again, maybe this has changed, but I think it's unlikely to have done given the mechanism for package registries.

tjmw: Resolving this as we're using the version from NPM now.

shtukas (approved):

PR #1954, opened 2020-09-30
Show signin gate 5 times after the first dismissal if user does not signin

QuarpT:
## What does this change?

Show signin gate 5 times after the first dismissal if user does not signin or register.
See https://trello.com/c/Xp9mwQGy/2471-goal-update-dismiss-rules-to-display-the-sign-in-gate-on-every-eligible-article-after-dismissing-t

Equivalent PR for dotcom
https://github.com/guardian/frontend/pull/23058

### Before

Signin gate was dismissed once and did not come back

### After

Show signin gate 5 times after the first dismissal if user does not signin

## Why?

Increase number of users registering / signing in. Data shows that a large percentage of users signin after 5 gate viewings

### DIFF
--- a/src/web/components/SignInGate/SignInGateSelector.tsx
+++ b/src/web/components/SignInGate/SignInGateSelector.tsx
@@ -3,7 +3,7 @@ import { useAB } from '@guardian/ab-react';
 import { ABTest, Runnable } from '@guardian/ab-core';
 import { constructQuery } from '@root/src/lib/querystring';
 
-import { setUserDismissedGate } from '@frontend/web/components/SignInGate/dismissGate';
+import { incrementUserDismissedGateCount, setUserDismissedGate } from '@frontend/web/components/SignInGate/dismissGate';
 import {
     SignInGateComponent,
     CurrentABTest,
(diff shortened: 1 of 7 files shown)


src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -74,13 +75,13 @@ const tests: ReadonlyArray<ABTest> = [
 const testVariantToGateMapping: GateTestMap = {
     'patientia-control-1': gatePatientiaControl,
     'patientia-variant-1': gatePatientiaVariant,
-    'main-control-1': gateMainControl,
-    'main-variant-1': gateMainVariant,
+    'main-control-2': gateMainControl,
```
QuarpT, line 78: @coldlink is it correct to bump the control/variant Ids for this variation?

coldlink: yeah thats right

coldlink (approved):
Looks fine to me. 👍 

Have you tested on `CODE`?

dominickendrick (approved):

PR #1958, opened 2020-10-01
Update source packages

liywjl:
## What does this change?
Update guardian source packages

## Why?
Some PRs are blocked because they need more recent icons only available in the most recent source icon package.
As source packages depend on one another it is preferable to update them together.


gtrufitt:
Hey, you'll need to update the dependencies first as they have a peer-dependency on source. discussion-rendering and CMP are the ones that I can think of to check, but any @guardian dependencies that are now moaning about peer-dependencies when you to a yarn install

There's also a migration guide here in case you hadn't seen it https://docs.google.com/document/d/1y2WNKiVifbPRuyTVYYwHCJIgEuFhd4_c8CnKkblCouI/edit

CC/ @SiAdcock

liywjl:
@gtrufitt Good point 👍 

On another note, I assumed that its preferable to update all that packages at once rather than just updating them individually (I feel like I remember once having an issue being caused by difference guardian source packages being installed)

liywjl:
Note:
One of the visual regression tests show a snapshot diff that is wrong:
https://www.chromatic.com/snapshot?appId=5dfcbf3012392c0020e7140b&id=5f918983ab599d00222336a0

This is likely due to the fact that the Chrome they are using to take the snapshot is not the latest and therefore still has that previous underlying issue that is discussed here https://github.com/guardian/source/issues/580 . The most recent Chrome update has resolved this issue.

shtukas (approved):

gtrufitt (approved):

PR #1962, opened 2020-10-06
Add isCountry display rule for sign in gate

vlbee:
# What does this change?

Adds a new sign in gate display rule `isCountry(countryCode)` based on the available local storage value of `gu.geolocation` which is set asynchronously on page load.  

## Why?

This display rule will be used on upcoming tests.

### DIFF
--- a/src/web/lib/getCountryCode.ts
+++ b/src/web/lib/getCountryCode.ts
@@ -33,6 +33,30 @@ export const setCountryCode = (countryCode: string): void => {
     });
 };
 
+export const setCountryCodeSynchronous = (countryCode: string): void => {
+    if (countryCode) {
+        localStorage.setItem(
+            COUNTRY_CODE_KEY,
+            JSON.stringify({
+                value: countryCode,
+                expires: new Date().getTime() + TEN_DAYS,
+            }),
+        );
+    }
+};
+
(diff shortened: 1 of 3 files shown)


src/web/lib/getCountryCode.ts
```diff
@@ -33,6 +33,30 @@ export const setCountryCode = (countryCode: string): void => {
...
+            COUNTRY_CODE_KEY,
+            JSON.stringify({
+                value: countryCode,
+                expires: new Date().getTime() + TEN_DAYS,
+            }),
+        );
+    }
+};
+
+export const getCountryCodeFromLocalStorage = (): string | null => {
```
dominickendrick, line 48: Could you just `return` from the `try...catch` block ?

```
    try {
        const item = localStorage.getItem(COUNTRY_CODE_KEY);
        localCountryCode = item ? JSON.parse(item) : null; 
        return localCountryCode.value || null
    } catch (error) {
        return null
    }
```
   

```

dominickendrick (approved):
Looks good !

src/web/lib/getCountryCode.ts
vlbee: thanks. will fix before merging

PR #1965, opened 2020-10-07
Tell Chromatic to ignore YouTube player

tjmw:
## What does this change?

We get quite a lot of differences reported by Chromatic related to the YouTube player which in general aren't related to the change at hand. Reduce these false positives by [telling Chromatic to ignore](https://www.chromatic.com/docs/ignoring-elements#ignore-dom-elements) the YouTube player.


gtrufitt (approved):
Just repeating for travellers:

I think in the long run it makes sense to make sure the video stays static as we're not testing for visual regression on the vide player but defo better to do this for now!

liywjl (approved):

PR #1967, opened 2020-10-08
Send pageViewId into the initalisation of the CMP

janua:
## What does this change?

This adds a the current `pageViewId` to the initialisation of the CMP in the App component.

The frontend version of this PR is [here](https://github.com/guardian/frontend/pull/23084).

## Why?

We require the pageview to be associated with the consent given for reporting purposes; this will allow us to report on how different consent affects our revenue and user behaviour.

@guardian/commercial-dev


src/web/components/App.tsx
```diff
@@ -327,7 +327,10 @@ export const App = ({ CAPI, NAV }: Props) => {
             injectPrivacySettingsLink(); // manually updates the footer DOM because it's not hydrated
             cmp.init({
                 isInUsa: countryCode === 'US',
-                pubData: { browserId: getCookie('bwid') || undefined },
+                pubData: {
+                    browserId: getCookie('bwid') || undefined,
+                    pageViewId: window.guardian.config.ophan.pageViewId,
```
janua, line 332: I'm not sure if this is safe or if there is a better way to access the `pageViewId`; would someone be able to advise?

sndrs: i think you might want the following:
```suggestion
                    pageViewId: window.guardian?.config?.ophan?.pageViewId,
```
at least that way you'll just get `undefined` if it fails

janua: @sndrs Thanks, I didn't realise chaining was available in TypeScript!

src/web/components/App.tsx
```diff
@@ -323,14 +323,17 @@ export const App = ({ CAPI, NAV }: Props) => {
...
                 isInUsa: countryCode === 'US',
-                pubData: { browserId: getCookie('bwid') || undefined },
+                pubData: {
+                    browserId: getCookie('bwid') || undefined,
+                    pageViewId: window.guardian?.config?.ophan?.pageViewId,
+                },
             });
         }
-    }, [countryCode, CAPI.config.switches.consentManagement]);
+    }, [countryCode, CAPI.config.switches.consentManagement, window.guardian?.config?.ophan?.pageViewId]);
```
janua, line 336: This `pageViewId` is defined outside of the React system, will this suffice as a dependency or does it need to be defined through `useState`?

sndrs: i htink you just dont need it in that array. you won't rerender if it changes right?

janua: Is it guaranteed to be available? What if it's unavailable at the start of the render then becomes available?

I have tested this and it seems to work fine though.

janua: @tjmw I have just tested this dependency by changing `pageViewId` myself within the browser; nothing has been update. I think this probably needs to hook into React somehow for it to work. React internally uses all callbacks so this is not surprising that it doesn't work; maybe it's not a fair test?

janua: I am going to remove this dependency for now; we can add it back in if we think we need it.

tjmw: Cool I think that sounds like a good decision :+1: Like Alex says if it changes you don't need to re-run the effect. I'm not 100% on this but changing it wouldn't trigger a re-render for sure and if there isn't a re-render the effect won't be re-run _but_ if something else triggered a re-render and `pageViewId` had changed then I think maybe it would re-run the effect (just speculating here for interest's sake, it sounds too flaky to rely on even if it this is right!) So the only scenario I'd want to make sure of is will the `pageViewId` definitely be available by the time we run the effect when the inner condition passes (i.e. when `countryId` and `CAPI.config.switches.consentManagement` are both truthy).

janua:
Tested on CODE; working as expected.

src/web/components/App.tsx
janua: @tjmw On the client side; we run the `pageViewId` generation code at the top of the DOM in a script tag; [see here](https://github.com/guardian/dotcom-rendering/blob/main/src/web/server/htmlTemplate.ts#L152-L175). 

This is happening well before the the application is initialised; so I would say that yes it is always available in the client side at least.

It's relationship with backend rendering through is what I am unsure about; does it even apply in this case?

tjmw: Ah great, thanks for pointing that out :+1: Sorry if I derailed things talking about the dependencies of the `useEffect`! Yeah I guess this all isn't relevant when it comes to server side rendering? (Not sure but I guess `useEffect` doesn't run server side).

gtrufitt: > This is happening well before the the application is initialised; so I would say that yes it is always available in the client side at least.

+1, you're right, this looks like it should be fine. As you've discovered, because the window value is not 'state' in a React sense, it wouldn't re-run a useEffect. You don't need it in this case. I think the Github Action is warning you of this.

It is available as on configuration that is passed through to the hydration step, but I would leave it as this for now, we need to review that as a whole anyway.

gtrufitt (approved):

PR #1968, opened 2020-10-08
Use placeholderUrl and set atom height based on comment in html

gtrufitt:
## What does this change?

- Adds a default height of 150px to Interactive Atoms on AMP
- Creates a mechanism for adding a height hint for interactive atoms on AMP through the use of a comment within the `html` field of the interactive atom
- Uses the `placeholderUrl` of an atom as a an image placeholder for interactive atoms on AMP, ensuring that we can show atoms at the top of articles
- Does not render the interactive atom if the `html` prop is an empty string (as this is used for modifying pages on Dotcom, not possible on AMP)

## Why?

Allows the showing of interactive atoms in AMP that exist at the top of an article - which includes full page interactive articles.


shtukas:
Nice one :)

shtukas (approved):

liywjl (approved):
👍 great work around

PR #1969, opened 2020-10-09
Fix broken CMP banner picker test

tjmw:
## What does this change?

This makes the id matching for the CMP iframe fuzzier in the e2e test. I'm not sure what the final number is, but it's recently changed and has caused this test to fail.

### Before

Test: 🔴 

### After

Test: 💚


shtukas (approved):

mxdvl:
@liywjl’s hunch was right, the id can change: https://github.com/guardian/dotcom-rendering/pull/1915#discussion_r488763736

Worth noting that the iframe ID changing _might be due_ to the Vendor List being updated.

PR #1971, opened 2020-10-12
Replace immersive black block with headline width

liywjl:
#  What does this change?
Uses the headline to determine the width rather than block

## Why?
Instead of guessing the width and the height of the end block element to match the headline, we should use the headline to cover the entire width


liywjl:
does not resolve issue

PR #1973, opened 2020-10-13
send hasOptedOutOfArticleCount for banners

tomrf1:
to indicate if the article count feature should be displayed or not


TomPretty (approved):
💯

PR #1975, opened 2020-10-13
Opt out of article count when the user removes consent

TomPretty:
## What does this change?
Opt users out of article count epics/banners if we don't have TCFv2 consent, as per [this card](https://trello.com/c/5QeO8WKw/2313-do-not-show-article-count-to-users-who-have-not-consented-to-the-cmp)

## Why?
We need to be compliant!


src/web/lib/contributions.ts
```diff
@@ -112,5 +117,35 @@ export const shouldHideSupportMessaging = (
     isRecurringContributor(isSignedIn) ||
     isRecentOneOffContributor();
 
+const REQUIRED_CONSENTS_FOR_ARTICLE_COUNT = [1, 3, 7];
```
tomrf1, line 120: not the fault of this PR, but this is quite cryptic!

tomrf1 (approved):
great work!

src/web/lib/contributions.ts
```diff
@@ -112,5 +117,35 @@ export const shouldHideSupportMessaging = (
...
+        onConsentChange(({ ccpa, tcfv2 }) => {
+            if (ccpa) {
+                resolve(true);
+            } else if (tcfv2) {
+                const hasRequiredConsents = REQUIRED_CONSENTS_FOR_ARTICLE_COUNT.every(
+                    (consent) => tcfv2.consents[consent],
+                );
+
+                if (!hasRequiredConsents) {
+                    removeArticleCountsFromLocalStorage();
```
tomrf1, line 144: 👍

src/web/lib/contributions.ts
TomPretty: it really is!

PR #1976, opened 2020-10-13
Correct spelling in index

shtukas:
## What does this change?
Correct spelling in index.


shtukas:
@gtrufitt Thanks for spotting that :)

gtrufitt (approved):
😁

PR #1977, opened 2020-10-13
Identity / Add AB test SignInGatePersonalisedAdCopy

vlbee:
## What does this change?

Adds an AB test to compare sign in gate copy about 'personalised advertising'

- Frontend version - https://github.com/guardian/frontend/pull/23103

### DIFF
--- a/src/web/components/SignInGate/gateDesigns/SignInGatePersonalisedAdCopyVariant2.tsx
+++ b/src/web/components/SignInGate/gateDesigns/SignInGatePersonalisedAdCopyVariant2.tsx
@@ -1,0 +1,251 @@
+import React from 'react';
+import { css, cx } from 'emotion';
+
+import { headline, textSans } from '@guardian/src-foundations/typography';
+import { from } from '@guardian/src-foundations/mq';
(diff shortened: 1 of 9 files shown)


src/web/components/SignInGate/gateDesigns/SignInGatePersonalisedAdCopyVariant2.tsx
```diff
@@ -0,0 +1,251 @@
...
+`;
+
+const actionButtons = css`
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-start;
+    margin-bottom: 42px;
+
+    > a {
+        margin-right: ${space[9]}px !important;
```
QuarpT, line 72: is it possible to avoid `!important` using emotion's class composition?

vlbee: These `!importants` are used on all the designs for a few months now. Not going to fix this now so we can get this test out, but the next set of tests are all design/UI changes so this can be looked at then.

QuarpT (approved):

PR #1978, opened 2020-10-14
Make banner container position fixed

tomrf1:
This fixes a bug in firefox on android.
The only disadvantage of using `fixed` over `sticky` is that the banner overlaps the footer at the bottom of the page.

The bug is caused by firefox's (new) search bar at the bottom of the screen. E.g. if I highlight a word (double tap) while the search bar is visible then it works correctly:
android_1

If I scroll down then firefox hides the search bar. If I try to highlight the same word then firefox selects the wrong part of the screen:
android_2

This is particularly problematic for buttons, because if a user scrolls down a bit then wants to click the close button on the banner then they have to click above the button.


shtukas (approved):

TomPretty (approved):
👍

PR #1980, opened 2020-10-14
Add resize listener for newsletter embeds

buck06191:
## What does this change?
This uses a similar approach to [`atomIframe/init.ts`](https://github.com/guardian/dotcom-rendering/blob/main/src/web/browser/atomIframe/init.ts), but relies on the iframe `src` instead of the `name`. This is necessary as we have to support legacy iframes that will have been created without a `name` field.

_Note: I haven't managed to get this to work locally as the iframe `src` and the `event.source.location.href` never match up when using the local Dotcom architecture. It might need CODE for testing._

### Before
image

### After
image

## Why?
The new designs require iframe resizing.

### DIFF
--- a/scripts/webpack/browser.js
+++ b/scripts/webpack/browser.js
@@ -47,6 +47,7 @@ module.exports = ({ isLegacyJS }) => ({
         dynamicImport: scriptPath('dynamicImport'),
         atomIframe: scriptPath('atomIframe'),
         embedIframe: scriptPath('embedIframe'),
+        newsletterEmbedIframe: scriptPath('newsletterEmbedIframe'),
     },
     output: {
         filename: generateName(isLegacyJS),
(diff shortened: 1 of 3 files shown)


liywjl (approved):
👍 

I will be splitting out the atom iframe logic into another file, maybe we can look at making the querySelector string a param in a later PR.

buck06191:
> 👍
> 
> I will be splitting out the atom iframe logic into another file, maybe we can look at making the querySelector string a param in a later PR.

That's good to know. The only difference we have here is that there is the small risk that we have the same iframe appear in multiple places on a page. It _shouldn't_ happen, but it might, so there has to be slightly different logic to deal with that edge case of resizing multiple at once.

buck06191:
Not for any future code archaeologists: You need to have the same domain for the iframe code to work. This means amending the `src` to be on the same domain. This probably means you need CODE in order to test any changes like this

PR #1981, opened 2020-10-15
Braze banner 'canShow' timeout metrics to Ophan

rhystmills:
## What does this change?
Measures the time taken for the braze-banner eligibility logic to take place, and sends that data to Ophan, so that we can measure how long it tends to take, and crucially how often it exceeds the 2 second limit specified for this logic to take place.

The equivalent Frontend PR can be found [here](https://github.com/guardian/frontend/pull/23127).

### DIFF
--- a/src/web/browser/initPerf.ts
+++ b/src/web/browser/initPerf.ts
@@ -9,13 +9,13 @@ export const initPerf = (name: string) => {
         perf.mark(startKey);
     };
 
-    const end = (shouldLog: boolean = true): TimeTakenInMilliseconds => {
+    const end = (): TimeTakenInMilliseconds => {
         perf.mark(endKey);
         perf.measure(name, startKey, endKey);
-        if (shouldLog) {
-            // eslint-disable-next-line no-console
-            console.log(JSON.stringify(perf.getEntriesByName(name)));
-        }
+
+        // eslint-disable-next-line no-console
+        console.log(JSON.stringify(perf.getEntriesByName(name)));
+
         const measureEntries = perf.getEntriesByName(name, "measure");
         const timeTakenFloat = measureEntries[0].duration;
         const timeTakenInt = Math.round(timeTakenFloat);
--- a/src/web/browser/initPerf.ts
+++ b/src/web/browser/initPerf.ts
@@ -1,4 +1,6 @@
 export const initPerf = (name: string) => {
+    type TimeTakenInMilliseconds = number;
+
     const perf = window.performance;
     const startKey = `${name}-start`;
     const endKey = `${name}-end`;
@@ -6,11 +8,20 @@ export const initPerf = (name: string) => {
     const start = () => {
         perf.mark(startKey);
     };
-    const end = () => {
+
+    const end = (dontLog: boolean = false): TimeTakenInMilliseconds => {
         perf.mark(endKey);
         perf.measure(name, startKey, endKey);
-        // eslint-disable-next-line no-console
-        console.log(JSON.stringify(perf.getEntriesByName(name)));
+
+        if (!dontLog) {
+            // eslint-disable-next-line no-console
+            console.log(JSON.stringify(perf.getEntriesByName(name)));
+        }
+        const measureEntries = perf.getEntriesByName(name, "measure");
+        const timeTakenFloat = measureEntries[0].duration;
+        const timeTakenInt = Math.round(timeTakenFloat);
+
+        return timeTakenInt;
     };
 
     return {
--- a/src/web/browser/initPerf.ts
+++ b/src/web/browser/initPerf.ts
@@ -9,11 +9,10 @@ export const initPerf = (name: string) => {
         perf.mark(startKey);
     };
 
-    const end = (dontLog: boolean = false): TimeTakenInMilliseconds => {
+    const end = (shouldLog: boolean = true): TimeTakenInMilliseconds => {
         perf.mark(endKey);
         perf.measure(name, startKey, endKey);
-
-        if (!dontLog) {
+        if (shouldLog) {
             // eslint-disable-next-line no-console
             console.log(JSON.stringify(perf.getEntriesByName(name)));
         }
--- a/src/web/browser/startup.ts
+++ b/src/web/browser/startup.ts
@@ -4,7 +4,7 @@ export interface Reporter {
     report: (err: Error, tags: { [key: string]: string }) => void;
 }
 
-const measure = (name: string, task: () => Promise<any>): void => {
+const measure = (name: string, task: () => Promise<void>): void => {
     const { start, end } = initPerf(name);
 
     start();
--- a/src/web/browser/startup.ts
+++ b/src/web/browser/startup.ts
@@ -14,8 +14,12 @@ const measure = (name: string, task: () => Promise<any>): void => {
         // in Chrome 59 which is used by Cypress in headless mode so if you do it will
         // break the Cypress tests on CI
         // See: https://github.com/cypress-io/cypress/issues/2651#issuecomment-432698837
-        .then(end)
-        .catch(end);
+        .then(() => {
+            end()
+        })
+        .catch(()=> {
+            end()
+        });
 };
 
 export const startup = <A>(
--- a/src/web/browser/startup.ts
+++ b/src/web/browser/startup.ts
@@ -4,7 +4,7 @@ export interface Reporter {
     report: (err: Error, tags: { [key: string]: string }) => void;
 }
 
-const measure = (name: string, task: () => Promise<void>): void => {
+const measure = (name: string, task: () => Promise<any>): void => {
     const { start, end } = initPerf(name);
 
     start();
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -155,8 +155,8 @@ export const canShow = async (
     asyncBrazeUuid: Promise<null | string>,
     isDigitalSubscriber: undefined | boolean,
 ): Promise<CanShowResult> => {
-    const canShowTiming = initPerf('braze-banner');
-    canShowTiming.start();
+    const timing = initPerf('braze-banner');
+    timing.start();
 
     const forcedBrazeMeta = getBrazeMetaFromQueryString();
     if (forcedBrazeMeta) {
@@ -192,7 +192,7 @@ export const canShow = async (
     try {
         const result = await getMessageFromBraze(apiKey as string, brazeUuid)
 
-        const timeTaken = canShowTiming.end(true);
+        const timeTaken = timing.end();
         record({
             component: 'braze-banner-timing',
             value: timeTaken,
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -5,9 +5,8 @@ import * as emotionTheming from 'emotion-theming';
 import { onConsentChange } from '@guardian/consent-management-platform';
 import { getZIndex } from '@root/src/web/lib/getZIndex';
 import { Props as BrazeBannerProps } from '@guardian/braze-components';
-import { submitComponentEvent } from '@root/src/web/browser/ophan/ophan';
+import { submitComponentEvent, record } from '@root/src/web/browser/ophan/ophan';
 import { CanShowResult } from './bannerPicker';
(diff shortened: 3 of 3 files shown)


src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -29,6 +30,31 @@ const containerStyles = emotion.css`
     ${getZIndex('banner')}
 `;
 
+const measureTiming = (name: string) => {
```
tjmw, line 33: What do you think about splitting this out into it's own file? I think that would make sense since this isn't core logic tied to the Braze banner, and would make it reusable. Even better perhaps we could extend [`initPerf`](https://github.com/guardian/dotcom-rendering/blob/main/src/web/browser/initPerf.ts) to fit our use case? What if we updated the `end` function returned by `initPerf` to return the performance data? Existing callsites wouldn't have to care about that (they just won't use the returned data). And then in this component we'd just need to extract the data we need. If it's a concern that the version of `end` from `initPerf` logs when we might now want it to, we could add an (optional) arg to `end` which prevents the logging. So the method signature would become:

```js
const end = (dontLog: boolean = false): PerformanceEntryList => { ... }
```

What do you reckon?

(I found the type definition for `PerformanceEntryList` [here](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.performance.html#getentriesbyname)).

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -29,6 +30,31 @@ const containerStyles = emotion.css`
...
+    const start = () => {
+        perf.mark(startKey);
+    };
+
+    const endAndReturn = () => {
+        perf.mark(endKey);
+        perf.measure(name, startKey, endKey);
+        const measureEntries = perf.getEntriesByName(name, "measure");
+        const timeTakenFloat = measureEntries[0].duration;
+        const timeTakenInt = Math.trunc(timeTakenFloat);
```
tjmw, line 47: It probably doesn't really matter, but would [Math.round](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round) make more sense here?

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -186,7 +216,14 @@ export const canShow = async (
     }
 
     try {
-        return getMessageFromBraze(apiKey as string, brazeUuid);
```
tjmw, line 189: Since we're in an `async` function here we could also use `await` if we felt it would make things more readable. This would look like:

```js
const result = await getMessageFromBraze(apiKey as string, brazeUuid);

const timeTaken = canShowTiming.endAndReturn();
record({
  component: 'braze-banner-timing',
  value: timeTaken,
});

return result;

```

I think I have a _slight_ preference for this, as it's also more consistent with the rest of this file. But I definitely don't feel strongly either way!

tjmw (commented):
I think this is looking good :+1: I left a few comments and suggestions - see what you think. Happy to chat if anything isn't clear!

src/web/components/StickyBottomBanner/BrazeBanner.tsx
rhystmills: Yep, definitely a better suggestion.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
rhystmills: Great idea.

Would you have a preference for returning the PerformanceEntryList rather than returning the time taken?

rhystmills: (I.e. processing the PerformanceEntryList to product the millisecond value integer)

src/web/components/StickyBottomBanner/BrazeBanner.tsx
rhystmills: I like it! Will work well in Frontend too.

src/web/browser/startup.ts
```diff
@@ -14,8 +14,12 @@ const measure = (name: string, task: () => Promise<void>): void => {
...
         // break the Cypress tests on CI
         // See: https://github.com/cypress-io/cypress/issues/2651#issuecomment-432698837
-        .then(end)
-        .catch(end);
+        .then(() => {
+            end()
+        })
+        .catch(()=> {
+            end()
+        });
```
gtrufitt, line 22: Isn't think equivalent to what it was originally?

src/web/browser/initPerf.ts
```diff
@@ -1,16 +1,27 @@
...
     };
-    const end = () => {
+
+    const end = (dontLog: boolean = false): TimeTakenInMilliseconds => {
         perf.mark(endKey);
         perf.measure(name, startKey, endKey);
-        // eslint-disable-next-line no-console
-        console.log(JSON.stringify(perf.getEntriesByName(name)));
+
+        if (!dontLog) {
```
gtrufitt, line 16: I think this might be clearer without the double negative, so make it `shouldLog` with a default of `true` instead?

rhystmills: I reckon that would be better, though it would change the interface of the function, which is used in quite a few places across DCR, and I'd have to update those to reflect the new interface. If you think this is a good idea, I'd be happy to do it - though it might be one for a follow up PR as it will touch quite a few files across DCR and is a fairly distinct piece of work.

Incidentally I think initPerf is responsible for quite a lot of console logging on production at the moment, it might not be a bad thing for the logging to be 'opt in' as it would make it a bit clearer that it's going to do that rather than tucking away the logging as a (potentially unintended) side effect.

src/web/browser/startup.ts
rhystmills: We were seeing some Typescript errors and @tjmw came up with this solution - it solved the problem, but it slightly went over my head - we're going to run over it once he's back.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -154,6 +155,9 @@ export const canShow = async (
     asyncBrazeUuid: Promise<null | string>,
     isDigitalSubscriber: undefined | boolean,
 ): Promise<CanShowResult> => {
+    const canShowTiming = initPerf('braze-banner');
```
adamnfish, line 158: This is fine, but if we find it useful, I wonder if it would be nice to wrap it around the can show logic for all banners? Out of scope for this PR and let's see how we find it.

src/web/browser/initPerf.ts
adamnfish: Agreed. It might be nicer for the start and end calls to return things so that the caller can decide whether they want to log, but I guess that's more a question for the platform team.

It comes down to what and who this initPerf module is really for. If it's going to have all these side-effects mixed in then it isn't very re-usable and I think we should consider calling the `window.performance` methods ourselves directly instead of adding abstraction on top of it to cater to our use-case.
As far as I can tell the module's value is pretty minimal right now, but maybe I'm missing some details or a future strategy reason for it? If the typical use-case is to call `record` and send the data to Ophan then should that be part of this module, for example?

gtrufitt: Interesting, what's your concern with logging to the console?

src/web/components/StickyBottomBanner/BrazeBanner.tsx
adamnfish: I'm not crazy about how the performance logic is all intertwined with the business logic of the Braze banner. If we wanted to tidy things now would we be able to wrap the business logic promise in timer functionality?

Something like:

    time(config params e.g. logging / ophan record data, <promise to time>)

If such a `time` function accepted and returned a promise then it can be kept separate from business logic by wrapping. Such a function would probably be useful in a separate (re-usable) module, subject to the same questions above about purpose and audience (and thus, side effects).

```typescript
const canShowLogic = async (
    asyncBrazeUuid: Promise<null | string>,
    isDigitalSubscriber: undefined | boolean,
  // Braze logic here as before
}

export const canShow = time(logging = false, canShowLogic)
```

I don't feel strongly here though, very happy for this to go out as-is for us to investigate. We can look at how future use-cases appear and decide whether there's a good abstraction to address those in the future when we have a better view of how it should work.

src/web/browser/startup.ts
```diff
@@ -4,7 +4,7 @@ export interface Reporter {
     report: (err: Error, tags: { [key: string]: string }) => void;
 }
 
-const measure = (name: string, task: () => Promise<void>): void => {
+const measure = (name: string, task: () => Promise<any>): void => {
```
tjmw, line 7: I think this can go back to `Promise<void>` since we now shouldn't have changed the interface `measure` has?

src/web/browser/startup.ts
tjmw: Yeah it's because we've changed `end` to return the time taken, but didn't want the knock on effect of changing the interface/return value here. So we're maintaining `measure`'s return type of `Promise<void>`.

src/web/browser/initPerf.ts
rhystmills: I can see what you mean Adam - I had implemented my performance measurement separately at first but Tom had suggested I could reduce duplication by modifying initPerf instead as I was basically replicating all of it short of returning the value.

On logging - I can't speak for Adam but my thoughts were:

- I've found more 'noise' in the console can make it harder to use for development/debugging. You end up having to sort through irrelevant information to find what you've deliberately logged during the development process.
- It feels a bit of a shotgun approach - logging to every user's console every time they load a page, when probably what was needed was to log to a specific developer's device at a particular moment in time in order for them to observe the performance measurement. 
- If it's being used for performance monitoring I think the quality of analysis would be better if we were gathering performance information from real user devices and storing it somewhere to subsequently analyse it, rather than just logging it to the console and presumably reading specific instances on our own devices.

In the future we could read from somewhere like localStorage for a developer 'performance logging opt-in' before logging to the console.

gtrufitt: The marks are used within speedcurve already, via WPT which automatically recognises them. We can graph module load times over time with those metric. This is still synthetic and, you're right, it would be ideal to automatically capture those marks and measures and send them to ophan to get better RUM data.

Logging to user's consoles is not a huge problem, just because it will be thrown away if dev tools aren't open.

We don't need the console logging, but it's sometimes useful in production - that said, I've never used it in anger in production so I am happy to remove the console logging of these entirely. Dev tools will give us the same thing anyway.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
rhystmills: That sounds sensible - I know with the current structure it's much simpler to put the timing within the BrazeBanner itself but it is probably worth revisiting in a future PR if we want to monitor their performance in the future.

src/web/browser/initPerf.ts
rhystmills: Thanks Gareth, useful to know! On your original suggestion - I've switched over to 'shouldLog' from the double negative now.

src/web/browser/startup.ts
rhystmills: Good catch, thanks! Sorted now.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: Hey sorry I somehow missed this comment before. I think the way you've got it right now where you do the processing in `measureTiming` and return the ms sounds good 👍  If we have a use case in future where we want something different that requires the whole `PerformanceEntryList` we can revisit then.

src/web/browser/initPerf.ts
tjmw: Hey this feels like the one possibly unresolved thread on this PR. It feels like the three options are:

1) Remove the opt in/out of logging flag and continue to always log and not worry about it - we're already logging some performance metrics to the console and it seems like we're not worried about logging one additional metric.
2) Remove the console logging completely
3) Do nothing and merge as is

I think maybe I'd lean towards option 1 (followed closely by 2!) and have the conversation about what's useful to log in production separately.

rhystmills: Thanks for taking another look Tom! I've followed number 1 and removed it - if we do feel the logging is a problem in the future we can sort it globally for DCR.

shtukas (approved):

PR #1983, opened 2020-10-16
Identity | Sign In Gate and Spacefinder interactions

coldlink:
## What does this change?
- Changes the element where the sign in gate is inserted to a `p` tag instead of `span`, to solve issues with space finder being unable to find a slot to place an ad.
- Changes to the CSS while the gate is visible, to hide any other inline article elements while the gate is visible, e.g. asides/ads.  - 
  - When the gate is dismissed, remove this CSS to make content reappear.

### DIFF
--- a/src/web/components/SignInGate/gateDesigns/SignInGateMain.tsx
+++ b/src/web/components/SignInGate/gateDesigns/SignInGateMain.tsx
@@ -126,8 +126,11 @@ const firstParagraphOverlay = (isComment: boolean) => css`
     );
 `;
 
-// This css hides all the elements in the article after the #sign-in-gate
-// using the General sibling combinator https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator
+// This css does 3 things
+// 1. first hide all article conent using display: none;
+// 2. make the sign in gate, and the first 2 paragraphs visible
+// 3. hide any siblings after the sign in gate incase because of the css in 2
+//    a paragraph is still visible
 const hideElementsCss = `
     .article-body-commercial-selector > * {
         display: none;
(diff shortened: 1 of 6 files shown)


src/web/components/SignInGate/gateDesigns/SignInGateMain.tsx
```diff
@@ -129,10 +129,18 @@ const firstParagraphOverlay = (isComment: boolean) => css`
 // This css hides all the elements in the article after the #sign-in-gate
 // using the General sibling combinator https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator
 const hideElementsCss = `
+    .article-body-commercial-selector > * {
+        display: none;
+    }
+
+    #sign-in-gate, .article-body-commercial-selector p:nth-of-type(-n+3) {
```
gtrufitt, line 136: Would you be able to comment this as not immediately clear :)

gtrufitt (commented):
I'm happy if you're confident with this short-term but I think it's probably a little brittle to be a long term fix?

src/web/components/SignInGate/gateDesigns/SignInGateMain.tsx
coldlink: i've added in comments above the variable

coldlink:
> I'm happy if you're confident with this short-term but I think it's probably a little brittle to be a long term fix?

yeah, we're just testing it to make sure it's okay for the short term

But it is brittle for the long term, I think we will need to rethink the way that we're doing the sign in gate on DCR.

gtrufitt (changes requested):
Next steps being:
1. Confirm changes on PROD
2. Check on Monday ad ratio improvements
3. Work with spacefinder to see if we can use that instead
4. Review whether this solution for sign-in gate makes sense long term

Nice work on this, thanks!

gtrufitt (approved):

liywjl (approved):
I would of changed https://github.com/guardian/frontend/blob/main/static/src/javascripts/projects/commercial/modules/article-body-adverts.js#L237 to `' > :not(p):not(h2):not(.ad-slot):not(span#sign-in-gate)': {` 

But given the timeframe I am happy with changing `SignInGateSlot` from `span` to `p`.

Great work guys

PR #1988, opened 2020-10-19
Limit banner height and allow scroll

tomrf1:
We have had reports of the close button being out of view on our banners if the user has set a very high font size in their browser.
This is because the banner container is `position: fixed`.

We can ensure the button is not too high with `80vh`.
This doesn't work for the width, which still overflows on mobile if the font is too large. So we allow scrolling in this case.

Now scrollable and with a limited height:
Screen Shot 2020-10-20 at 08 05 42
Screen Shot 2020-10-20 at 08 06 42

### DIFF
--- a/src/web/layouts/lib/stickiness.tsx
+++ b/src/web/layouts/lib/stickiness.tsx
@@ -32,6 +32,8 @@ const bannerWrapper = css`
     position: fixed !important;
     bottom: 0;
     ${getZIndexImportant('banner')}
+    max-height: 80vh;
+    overflow: auto;
     
     width: 100% !important;
     background: none !important;


TomPretty (approved):
👍

liywjl (commented):
This feels like a work around for a more fundamental issue of how we should display text at that screen width. Idk if it maybe worth bringing this up with @HarryFischer ? @gtrufitt do you think this is an ok workaround in the meantime?

Side note: the background article seems to have lots of format issues, is that something that should be raised?

tomrf1:
Yes if we just used px instead of rem then it wouldn't be an issue, but would that make it less accessible?

liywjl:
@tomrf1 could you add a couple more details to replicate this locally? 

I spoke to @HarryFischer who suggests it may be best to avoid having a left right scroll for the text as button, and instead make sure that the user can only scroll up and down.

This may require the banner to be reworked at certain breakpoints

liywjl (changes requested):
@tomrf1 could you add a couple more details to replicate this locally?

I spoke to @HarryFischer who suggests it may be best to avoid having a left right scroll for the text as button, and instead make sure that the user can only scroll up and down.

This may require the banner to be reworked at certain breakpoints

tomrf1:
@liywjl to reproduce this, set the browser's font size to e.g. 40. To get the banner to display again you can use a private browser window.
Because we define our font sizes in rem, it overflows. So I don't think this is a case of needing to change the banner layout at different breakpoints. We either set font size in px (making it less accessible?), or allow scrolling.

liywjl (approved):
Good workaround for the current issue.

Will need a better solution requiring some redesigns for the banner on these widths. Notable things to consider:

- needing to scroll left and right on mobile is uncomfortable (always favor up-down scroll)
- we should be able to see the close button on load (ideally no need to scroll to see it)
- text could be smaller, but should no break accessibility principles

tomrf1:
Thanks @liywjl !
As a quick fix we will go ahead with this PR, and the contributions team will look into improving the design of the banner on mobile

PR #1994, opened 2020-10-21
Add story for Braze banner

tjmw:
## What does this change?

@liywjl and I encountered a scenario where a visual change only cropped up when rendering the banner in the context of the DCR project (it was related to peer dependencies). We manually spotted the issue but we'd like to pick up visual regressions like that in a more automated way in future, so add a basic story here.

We have to lazily load @guardian/braze-components in the same way we do in prod because we have to configure the shared deps (via the window) _before_ the component is loaded.


liywjl (approved):
Great work! 👍

PR #1995, opened 2020-10-21
Match ad slots breakpoint label values with frontend

ioanna0:
###   What does this change?
@gtrufitt  found some inconsistencies on ad slots `data` labels for each breakpoint. In particular `most-pop` on `tablet` had a `null` value.
This PR makes sure we get 100% parity between frontend and dcr on those values

### Before

(no `data-tablet` attribute present)

Screenshot 2020-10-21 at 17 56 12

### After

Screenshot 2020-10-21 at 17 45 18

## Why?
This will bring some more revenue on tablets on most-pop ad slots

### DIFF
--- a/src/model/advertisement.ts
+++ b/src/model/advertisement.ts
@@ -106,7 +106,7 @@ export const namedAdSlotParameters = (name: AdSlotType): AdSlotParameters => {
             name: 'right',
             adTypes: ['mpu-banner-ad', 'rendered'],
             sizeMapping: {
-                mobile: ['1,1|2,2|300,250|300,274|300,600|fluid'],
+                mobile: ['1,1|2,2|300,250|300,274|fluid'],
                 // Pascal: It's not totally obvious where this sequence should come from the frontend code,
                 // but it should be correct.
             },
--- a/src/model/advertisement.ts
+++ b/src/model/advertisement.ts
@@ -145,11 +145,9 @@ export const namedAdSlotParameters = (name: AdSlotType): AdSlotParameters => {
                             'outOfPage',
                             'empty',
                             'fabric',
-                            'outstreamMobile',
-                            'mpu',
                             'fluid',
+                            'leaderboard',
                         ]),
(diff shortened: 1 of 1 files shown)


gtrufitt (approved):
LGTM, though feel free to get a second (third) opinion! 👍

src/model/advertisement.ts
```diff
@@ -106,7 +106,7 @@ export const namedAdSlotParameters = (name: AdSlotType): AdSlotParameters => {
             name: 'right',
             adTypes: ['mpu-banner-ad', 'rendered'],
             sizeMapping: {
-                mobile: ['1,1|2,2|300,250|300,274|300,600|fluid'],
+                mobile: ['1,1|2,2|300,250|300,274|fluid'],
                 // Pascal: It's not totally obvious where this sequence should come from the frontend code,
                 // but it should be correct.
```
mxdvl, line 111: I think the use of [file marks][uuid] as below can be quite useful to navigate between `DCR` and `frontend`.

[uuid]: https://www.uuidgenerator.net/version4

mxdvl (commented):
Great!

Could it be captured by a Cypress test perhaps?

If a change is made on one platform, how to we ensure it gets updated on the other?

src/model/advertisement.ts
ioanna0: Ha, i'm not familiar with them but will have a look. It's coming from this file:

https://github.com/guardian/frontend/blob/beb2d69bfcb901403498d942291d59f62d2064dc/static/src/javascripts/projects/commercial/modules/article-aside-adverts.js#L70

ioanna0:
> Could it be captured by a Cypress test perhaps?
Very good question, don't think we have any cypress tests for the commercial stack but would be great if we start looking into them as it will help finding all the regressions. Let's discuss with team and card up.

> If a change is made on one platform, how to we ensure it gets updated on the other?
Another good question! Server side ad slots are indeed shared but dynamic ones had to leave inside each platform due to the differences they have in nature (it didn't work with the shared commercial stack either). I suppose the commercial-repo sounds promising for moving them out into a shared place?

PR #1996, opened 2020-10-21
Hf rp comment block

HarryFischer:
## What does this change?
Re-designed comment block

### example article
https://www.theguardian.com/tv-and-radio/2017/jan/31/peter-capaldi-doctor-who-bbc-your-verdict

### Before
image

### After
image

## Why?
Old design didn't fit with the style of website


shtukas (approved):

src/web/components/elements/CommentBlockComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+    border-bottom: none !important;
+    img {
+        border-radius: 50%;
+    }
+`;
+
+const userName = css`
+    a {
+        ${textSans.medium({ fontWeight: 'bold' })};
+        /* color: ${palette.neutral[86]}; */
```
liywjl, line 49: remove

src/web/components/elements/CommentBlockComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+    padding: 4px 0 0 10px;
+    border-left: 1px ${palette.neutral[86]} solid;
+    border-top: 1px ${palette.neutral[86]} solid;
+    margin-top: 12px;
+`;
+
+const bodyContent = css`
+    p {
+        word-break: break-word;
+        margin-bottom: 8px;
```
liywjl, line 26: use `space` from source

src/web/components/elements/CommentBlockComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+    profileName: string;
+    dateTime: string;
+};
+
+const commentWrapper = css`
+    ${textSans.medium()};
+    padding: 4px 0 0 10px;
+    border-left: 1px ${palette.neutral[86]} solid;
+    border-top: 1px ${palette.neutral[86]} solid;
+    margin-top: 12px;
```
liywjl, line 20: use `space` from source

src/web/components/elements/CommentBlockComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+    display: flex;
+    margin-top: ${space[4]}px;
+    margin-bottom: ${space[12]}px;
+`;
+
+const avatar = css`
+    width: ${space[12]}px;
+    height: ${space[12]}px;
+    margin-right: ${space[2]}px;
+    border-bottom: none !important;
```
liywjl, line 40: is this needed?

liywjl (changes requested):
looks good, just some minor changes

src/web/components/elements/CommentBlockComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+    return (
+        <div className={commentWrapper}>
+            <div
+                className={bodyContent}
+                dangerouslySetInnerHTML={{ __html: unescapeData(body) }}
+            />
+            <div className={profileWrapper}>
+                <a className={avatar} href={profileURL}>
+                    <img src={avatarURL} alt="" />
+                </a>
```
gtrufitt, line 73: This will appear to screen reader users as an empty link. Because we don't have descriptive text for this image other than `avatar of profileName`, then maybe it's worth removing this anchor with `aria-hidden="true"` @SiAdcock what do you think?

src/web/components/elements/CommentBlockComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+    }
+`;
+
+const profileWrapper = css`
+    display: flex;
+    margin-top: ${space[4]}px;
+    margin-bottom: ${space[12]}px;
+`;
+
+const avatar = css`
```
liywjl, line 36: as a conversion we try to call these variables with the suffix `-Styles` so here we could call it `avatarStyles`

gtrufitt (commented):
It looks great! I just have some suggestions for semantics and accessibility

src/web/components/elements/CommentBlockComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+const userName = css`
+    a {
+        ${textSans.medium({ fontWeight: 'bold' })};
+        /* color: ${palette.neutral[86]}; */
+        border-bottom: none;
+    }
+    p {
+        ${textSans.small()};
+    }
```
liywjl, line 54: applying styles this way is not ideal especially if you control the rendering of each element. It would be better to create a separate const for `a` and `p` tag.

src/web/components/elements/CommentBlockComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+    width: ${space[12]}px;
+    height: ${space[12]}px;
+    margin-right: ${space[2]}px;
+    border-bottom: none !important;
+    img {
+        border-radius: 50%;
+    }
+`;
+
+const userName = css`
```
liywjl, line 46: name change needed

src/web/components/elements/CommentBlockComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+`;
+
+const avatar = css`
+    width: ${space[12]}px;
+    height: ${space[12]}px;
+    margin-right: ${space[2]}px;
+    border-bottom: none !important;
+    img {
+        border-radius: 50%;
+    }
```
liywjl, line 43: apply this style directly to the image, not through the avatar const

liywjl (changes requested):
If you apply the changes and directly apply the css to the `tags` rather than referencing them in parent CSS styles, your styles will take precedence over the styles causing the issue. This means you will no longer need to use `!important`

src/web/components/elements/CommentBlockComponent.tsx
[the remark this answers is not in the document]
gtrufitt: Actually this won't work.

I think the real fix here is to remove the linkColour styles from esentially 'global' https://github.com/guardian/dotcom-rendering/blob/a2e3c7f46bb1fd37eea79ddd5ba374f1e9083742/src/web/components/ArticleBody.tsx#L49 and add them as needed in components...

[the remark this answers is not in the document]
HarryFischer: Yeah I couldn't find a way to get it to work – will need to do as you suggested I think – can chat tomorrow about it

rcrphillips:
[CommentBlockElement](https://trello.com/c/d9qL6w2o/1768-commentblockelement)

liywjl (approved):
LGTM

Some nice to haves however: 
- It would be good to however add an `a` tag around the `time` tag to link to `permalink` prop. 
- It might be worth adding `target="_blank"` in the `a` tags here to open new tabs on click

liywjl (approved):

PR #2001, opened 2020-10-23
Atoms-rendering docs update

nitro-marky:
## What does this change
@amyhughes and I have made additional documentation to the existing which will hopefully make creating and adding atoms to DCR a bit simpler (hopefully!).

Feedback very welcome 👍


nitro-marky:
[Atoms-rendering doc](https://trello.com/c/IHl15IhS/2056-atoms-rendering-doc)

gtrufitt (approved):

PR #2004, opened 2020-10-27
Parity DCR DOM tags with frontend

liywjl:
## What does this change?
Changes some DOM tags to better match frontend

## Why?
for both google indexing and spaceFinder (which uses DOM tags to determine where to place elements in a page)

### DIFF
--- a/src/web/components/elements/BlockquoteBlockComponent.tsx
+++ b/src/web/components/elements/BlockquoteBlockComponent.tsx
@@ -13,17 +13,6 @@ type Props = {
     pillar: Pillar;
 };
 
-const Row = ({ children }: { children: React.ReactNode }) => (
-    <div
-        className={css`
-            display: flex;
-            flex-direction: row;
-        `}
-    >
-        {children}
-    </div>
-);
-
 export const BlockquoteBlockComponent: React.FC<Props> = ({
     html,
     pillar,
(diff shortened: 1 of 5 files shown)


src/web/components/elements/BlockquoteBlockComponent.tsx
```diff
@@ -47,6 +41,6 @@ export const BlockquoteBlockComponent: React.FC<Props> = ({
                 `}
                 tagName="blockquote"
```
gtrufitt, line 42: You're gonna double blockquote here.

src/web/lib/ArticleRenderer.tsx
```diff
@@ -278,14 +278,14 @@ export const ArticleRenderer: React.FC<{
                     );
                 case 'model.dotcomrendering.pageElements.RichLinkBlockElement':
                     return (
-                        <div key={i} id={`rich-link-${i}`}>
+                        <aside key={i} id={`rich-link-${i}`}>
```
gtrufitt, line 281: Might it make sense to move this `aside` along with the key and id generation into the component itself?

gtrufitt (approved):
Yay to semantics! and parity! Couple of suggestions but +1 when you've looked at them 👍

oliverlloyd:
@liywjl do we still want to move forward with this PR?

liywjl:
I thought this had an impact on Ads and space finder. It seemed not to in the end, and I had some challenges with lifting dangerously insterted DOM elements to fit the syntax. It might be worth revisiting one day, but is fairly low impact

PR #2007, opened 2020-10-28
Youtube overlay

liywjl:
## What does this change?
Adds overlay image to youtube

### Example article 
https://www.theguardian.com/world/2020/apr/01/what-is-coronavirus-and-what-is-the-mortality-rate-covid-19?dcr=false

### Before
Screen_Shot_2020-03-12_at_08 53 46

### After
Screenshot 2020-10-30 at 16 55 39

## Why?
Parity with frontend


paperboyo:
Old frontend never made them overlays HiDPI aware (coz harder to do within `background-image`, I suppose, although possible?). Will this support them?

nitro-marky (approved):

liywjl:
> Old frontend never made them overlays HiDPI-aware (coz harder to do within `background-image`, I suppose, although possible?). Will this support them?

An interesting point, I have never really thought much about HiDPI. After reading a bit about it, I suspect the scope of supporting Hight res screens is a whole challenge in itself.

The overlay logic implemented here is heavily based on how frontend implements it (you should be able to see 
```
background-image: url(https://i.guim.co.uk/img/media/4b3808707ec341629932a9d443ff5a812cf4df14/0_309_1800_1081/master/1800.jpg?width=700&quality=85&auto=format&fit=max&s=afebec1662ceabf6a65889ba386d8b7d
``` in the overlay `div` on frontend https://www.theguardian.com/world/2020/apr/01/what-is-coronavirus-and-what-is-the-mortality-rate-covid-19?dcr=false )

I think it would be interesting to explore how we can adapt the page for different screen sizes, if you could link me some things to read more on the topic I would love to find out more. But for now, I think this is beyond the scope of this ticket

PR #2011, opened 2020-10-29
Delay re-fetching RR banner if closed

tomrf1:
This is an optimisation to reduce the number of requests made by the client to support-dotcom-components.

Currently we make a request to find out if the user should see a RR banner even if the user has previously closed the banner.
This is because we need to check if the banner has been redeployed since the user last closed it.
(Note - there are actually 2 banner 'channels', each with its own 'last closed at' timestamp).
This means a lot of unnecessary requests because banners are redeployed very infrequently, and most page views should not see a banner.

E.g. in the last 24 hours:
- 25m requests were sent to /banner
- 19m of these returned no banner

The solution in this PR is to avoid making any banner requests for 20 minutes after a request returns no banner, if the user has closed the banner.
So, if the user has already closed the banner (both channels) and the response has no banner, we add a local storage item with a timestamp 20mins into the future.

This is a bit like caching the /banner response when it returns no banner, but it's a POST so that would not be reliable.

Edit:
We also discussed having the client fetch the banner deploy timestamps directly (via fastly).
This would certainly avoid even more requests to /banner, but at the cost of more complexity (we'd have to make that data public, and behind fastly).
We'll try this solution for now.

### DIFF
--- a/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
+++ b/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
@@ -10,6 +10,8 @@ import {
 import {
     shouldHideSupportMessaging,
     getArticleCountConsent,
+    noBannerUntilLater,
+    setNoBannerUntilLaterKey,
 } from '@root/src/web/lib/contributions';
 import { getCookie } from '@root/src/web/browser/cookie';
 import {
@@ -118,6 +120,11 @@ export const canShow = async ({
         return Promise.resolve({ result: false });
     }
 
+    if (engagementBannerLastClosedAt && subscriptionBannerLastClosedAt && noBannerUntilLater()) {
+        // Both banners have been dismissed *and* we have recently been told not to display a banner
+        return Promise.resolve({ result: false });
+    }
+
     const countryCode = await asyncCountryCode;
     const hasConsentedToArticleCounts = await getArticleCountConsent();
     const bannerPayload = buildPayload({
@@ -147,6 +154,10 @@ export const canShow = async ({
         .then((response) => response.json())
         .then((json) => {
             if (!json.data) {
+                if (engagementBannerLastClosedAt && subscriptionBannerLastClosedAt) {
+                    // Both banners have been dismissed *and* we have been told not too display a banner - do not request banner for a while
+                    setNoBannerUntilLaterKey();
+                }
                 return { result: false };
             }
 
--- a/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
+++ b/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
@@ -10,8 +10,8 @@ import {
 import {
     shouldHideSupportMessaging,
     getArticleCountConsent,
-    noBannerUntilLater,
-    setNoBannerUntilLater,
+    withinLocalNoBannerCachePeriod,
+    setLocalNoBannerCachePeriod,
 } from '@root/src/web/lib/contributions';
 import { getCookie } from '@root/src/web/browser/cookie';
 import {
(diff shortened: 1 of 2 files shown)


src/web/lib/contributions.ts
```diff
@@ -149,3 +150,20 @@ export const getArticleCountConsent = (): Promise<boolean> => {
         });
     });
 };
+
+export const noBannerUntilLater = (): boolean => {
```
rupertbates, line 154: I think the approach is a good one. I think I'd call this function something more specific like `withinLocalCachePeriod` to make it really clear what it relates to.

rupertbates (approved):
I like the approach, just one suggestion for naming 👍

src/web/lib/contributions.ts
tomrf1: I've renamed to:
`withinLocalNoBannerCachePeriod`
`setLocalNoBannerCachePeriod`

src/web/lib/contributions.ts
```diff
@@ -149,3 +150,20 @@ export const getArticleCountConsent = (): Promise<boolean> => {
...
+            window.localStorage.removeItem(NO_BANNER_UNTIL_LATER_KEY);
+        }
+        return noBanner;
+    }
+    return false;
+};
+
+const twentyMins = 20*60000;
+export const setLocalNoBannerCachePeriod = (): void =>
+    window.localStorage.setItem(NO_BANNER_UNTIL_LATER_KEY, `${Date.now() + twentyMins}`);
```
johnduffell, line 169: minor point given 20 minutes, but it might be better to make this the last banner refreshed time, and then add the 20 minutes when checking (on line 157).  This just means if you want to change the period it takes effect as soon as you deploy, rather than after the caching period.

src/web/lib/contributions.ts
```diff
@@ -149,3 +150,20 @@ export const getArticleCountConsent = (): Promise<boolean> => {
         });
     });
 };
+
+export const withinLocalNoBannerCachePeriod = (): boolean => {
+    const item = window.localStorage.getItem(NO_BANNER_UNTIL_LATER_KEY);
+    if (item && !Number.isNaN(parseInt(item, 10))) {
+        const noBanner = parseInt(item, 10) > Date.now();
+        if (!noBanner) {
```
johnduffell, line 158: can we flip the logic so this isn't a double negative? or just change NoBanner to IsCached

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -147,6 +153,9 @@ export const canShow = async ({
         .then((response) => response.json())
         .then((json) => {
             if (!json.data) {
```
johnduffell, line 155: `data` meaning "should show banner"?

src/web/lib/contributions.ts
tomrf1: I like this

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
tomrf1: yes, if `data` is missing in the response then there is no banner to show

johnduffell (approved):
looks like a great and simple idea overall to see how it affects things, just a few comments about style really.

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
johnduffell: I know that's not even your change, but we could  const that up to make it clearer

src/web/lib/contributions.ts
tomrf1: done!

PR #2017, opened 2020-11-02
Global Ad slot css for ad imbed

liywjl:
# What does this change?
Applies global CSS rules to inline Ad slots

## NOTE
We need https://github.com/guardian/frontend/pull/23204 to remove inline styles, but we should first merge this. To test this PR you will need to manually delete the styles attr from the div

### Example article
https://www.theguardian.com/us-news/2020/oct/29/lordstown-ohio-trump-gm-plant-election

### Before
Screenshot 2020-10-30 at 10 06 30

### After
!!!! IF YOU REMOVE STYLES ATTR FROM DIV

Screenshot 2020-11-02 at 15 53 00

## Why?
Some CSS styles are applied globally from frontend to the ads based on classNames

### DIFF
--- a/src/web/components/ArticleBody.tsx
+++ b/src/web/components/ArticleBody.tsx
@@ -46,30 +46,6 @@ const bodyStyle = (display: Display) => css`
     }
 `;
 
-const imbedAdSlotStyles = css`
-    div > div.ad-slot--im {
-        float: left;
-        width: 130px;
-
-        ${from.mobileLandscape} {
-            width: 220px;
-        }
-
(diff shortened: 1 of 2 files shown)


shtukas (approved):

gtrufitt (changes requested):
I think we concluded yesterday that we want this CSS to live alongside the definition of the test within Frontend, in the commerical JS, rather than here in DCR. @mxdvl has implemented this previously in this PR https://github.com/guardian/frontend/pull/22751 I believe the same thing?

PR #2018, opened 2020-11-02
Double PROD ASG Min and Max for Election

gtrufitt:
## What does this change?

As per https://github.com/guardian/platform/pull/1420 - here we double the (much smaller) DCR server min and max.

## Why?

Election traffic baseline is higher and spiky.

### DIFF
--- a/cloudformation.yml
+++ b/cloudformation.yml
@@ -60,8 +60,8 @@ Mappings:
       Value: rendering
   StageMap:
     PROD:
-      MinCapacity: 9
-      MaxCapacity: 36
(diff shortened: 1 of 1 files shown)


oliverlloyd:
> much smaller

Love the subtle dig there

Just a thought, if this is only a short term change, maybe increase the min size so we're running steady at, say, 36? I know the issues with starting new boxes have been addressed but its extra risk nonetheless and for a few extra ££s might be worth it?

liywjl (approved):

jfsoul (approved):

PR #2019, opened 2020-11-02
Youtube Atom overlay

liywjl:
## What does this change?
Add support for overlay image on youtube atom ~embeds~

### Before/After
TODO: need to find example

## Why?
Overlay images can exist for embed youtube blocks, therefore we need to support it similar to how main media implements it


shtukas (approved):

liywjl:
changing branch name https://github.com/guardian/dotcom-rendering/pull/2020

PR #2022, opened 2020-11-03
A longer shortUrl

oliverlloyd:
Replaces `gu.com` with `theguardian.com`

Short urls are no longer as important as they were and, because we want to reduce redirects, the `gu.com` domain used in Guardian short urls is being replaced with `theguardian.com`

See: https://trello.com/c/kBHVNERc/2116-check-capi-short-url-change-will-not-break-anything-on-dotcom

### What is actually being changed here?
Two things:

1. All the fixture data is being updated to use `shortUrl` values that have the new domain string. There's no immediate requirement to do this but it's good practice to keep our test data up to date.

2. The string manipulation logic used around fetching comment counts has been updated to handle both domains. This will ensure if and when these shortUrls used bu the commenting api change, DCR will continue to work as expected.


nitro-marky (approved):

shtukas (approved):

liywjl (approved):

mxdvl:
Worth noting that the recommendation is now to keep the `www` in the short URL, to reduce redirections: https://github.com/guardian/content-api/pull/2547

PR #2024, opened 2020-11-04
Youtube atom origin prop

liywjl:
## What does this change?
Support optional origin prop to be set in youtube iframe URL

## Why?
When origin is set, if the URL doesn't match then the video will be prevented from playing. As we have `host` passed down in the JSON, we should use that to set the origin

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
         "@emotion/core": "^10.0.28",
         "@guardian/ab-core": "^2.0.0",
         "@guardian/ab-react": "^2.0.1",
-        "@guardian/atoms-rendering": "^2.0.5",
+        "@guardian/atoms-rendering": "^2.0.7",
         "@guardian/automat-client": "^0.2.16",
         "@guardian/braze-components": "^0.0.13",
         "@guardian/consent-management-platform": "^6.6.0",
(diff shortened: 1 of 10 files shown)


liywjl:
> `host` and `origin` are subtly different things. See: https://stackoverflow.com/questions/13851946/header-origin-vs-host
> 
> But we seem to have ended up with both names here. Not sure if this is:
> 
> a) worth it
> or b) semantically possible
> 
> But can we rationalise this down to one name?

Thats a really good point, but I think in this case it makes sense, because we are telling youtube what the origin is, so our `host` is youtube's `origin`?

Granted idk what other option might be better thought to set origin.

oliverlloyd:
> > `host` and `origin` are subtly different things. See: https://stackoverflow.com/questions/13851946/header-origin-vs-host
> > But we seem to have ended up with both names here. Not sure if this is:
> > a) worth it
> > or b) semantically possible
> > But can we rationalise this down to one name?
> 
> Thats a really good point, but I think in this case it makes sense, because we are telling youtube what the origin is, so our `host` is youtube's `origin`?
> 
> Granted idk what other option might be better thought to set origin.

Fair point. It sounds like each use fo origin and host is valid

👍

PR #2026, opened 2020-11-06
Use acast enabled version of AudioAtom

oliverlloyd:
### What
This PR adds `@guardian/atoms-rendering` at `v2.0.9` which encludes support for Acast via the `should UseAcast` prop.

We decide how to set the `shouldUseAcast` prop within a `useEffect` during hydration on the client (in `App.jsx`). The logic is:

```typescript
            if (
                aCastisEnabled &&
                consentGiven &&
                readerCanBeShownAds && // Eg. Not a subscriber
                contentIsNotSensitive
            ) {
                setShouldUseAcast(true);
            }
```

### Why make the decision about whether to display Acast ads here, and not in the atom itself?
Because:

1, it keeps the atom simple and
2, becuase different platforms might use different methods to decide if a user has consented to ads or not

### DIFF
--- a/fixtures/CAPI/CAPI.ts
+++ b/fixtures/CAPI/CAPI.ts
@@ -2792,6 +2792,7 @@ export const CAPI: CAPIType = modifyDCRJsonResponse({
             videojs: true,
             surveys: true,
             inizio: true,
+            acast: true,
         },
         hasYouTubeAtom: false,
         inBodyInternalLinkCount: 15,
--- a/fixtures/switches.ts
+++ b/fixtures/switches.ts
@@ -90,4 +90,5 @@ export const switches = {
     surveys: true,
     abCommercialAdVerification: false,
     inizio: true,
+    acast: true,
 };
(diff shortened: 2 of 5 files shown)


oliverlloyd:
@sndrs @mxdvl , I've put the check to see if consent was given into a separate `useEffect / onConsentChange` section. But I see that [there is an existing](https://github.com/guardian/dotcom-rendering/blob/main/src/web/components/App.tsx#L335) `useEffect` which also contains an `onConsentChange` function. Do you think there is benefit to merging this code? I quite like keeping things separate to help with maintainability but I'm not sure if there's a performance hit to having multiple `onConsentChange` listeners on the page.

liywjl (approved):
LGTM

oliverlloyd:
[Serve ad-free podcasts when user has not consented to targeted advertising (on DCR)](https://trello.com/c/HGaimQ0W/2031-serve-ad-free-podcasts-when-user-has-not-consented-to-targeted-advertising-on-dcr)

mxdvl (approved):
👍 from a CMP perspective. Adding `onConsent` change in several instances should have minimal performance impact. This is how we do it in the commercial bundle now.

PR #2029, opened 2020-11-10
Mj implement ticker for amp epics

michaelbjacobson:
### What does this PR do?
Implements a ticker component for use with AMP epics. The epic data is requested from `support-dotcom-components`, and if the returned data includes a ticker, it's rendered on the page. Both money and people ticker types are supported, see screenshots below.

### Screenshots
_Contributions (money) ticker_
money

_Supporters (people) ticker_
people

_No ticker_
no-ticker


michaelbjacobson:
[Explore the possibility to have a ticker on AMP and Apple news](https://trello.com/c/k7SNJTol/2377-explore-the-possibility-to-have-a-ticker-on-amp-and-apple-news)

src/amp/components/Epic.tsx
```diff
@@ -9,6 +9,7 @@ import {
 } from '@root/src/amp/components/moustache';
 import { palette } from '@guardian/src-foundations';
 import { headline, body, textSans } from '@guardian/src-foundations/typography';
+import {Ticker} from "@root/src/amp/components/Ticker";
```
liywjl, line 12: I think that prettier doesn't seem to be working, it might be worth configuring it to run on save.

src/amp/components/Ticker.tsx
```diff
@@ -0,0 +1,94 @@
...
+    justify-content: flex-end;
+`;
+const right = css`
+    text-align: right;
+    display: flex;
+    flex-wrap: nowrap;
+    flex-direction: column;
+    justify-content: flex-end;
+`;
+const tickerInfo = css`
```
liywjl, line 23: as a convention we are trying to add the suffix `Styles` to const related to styling. Thus we could name this variable `tickerInfoStyle`, this should also apply to other style consts.

src/amp/components/Ticker.tsx
```diff
@@ -0,0 +1,94 @@
...
+    flex-wrap: nowrap;
+    flex-direction: row;
+    justify-content: space-between;
+`;
+const tickerProgress = css`
+    position: absolute;
+    margin: 0;
+    padding: 0;
+    left: 0;
+    background-color: ${brandAlt[400]} !important;
```
liywjl, line 35: do we need `!important` here?

src/amp/components/Ticker.tsx
```diff
@@ -0,0 +1,94 @@
...
+    margin: 5px 0;
+    height: 10px;
+    width: 100%;
+    background-color: ${neutral[86]};
+    border: none;
+`;
+const currentAmount = css`
+    font-style: normal;
+    font-weight: 700;
+    font-size: 24px !important;
```
liywjl, line 49: do we need `!important` here?

src/amp/components/Ticker.tsx
```diff
@@ -0,0 +1,94 @@
...
+                        <p className={currentAmount}>{currentAmountFigure}</p>
+                        <p className={amountCaption}>{currentAmountCaption}</p>
+                    </div>
+                    <div className={right}>
+                        <p className={goalAmount}>{goalAmountFigure}</p>
+                        <p className={amountCaption}>{goalAmountCaption}</p>
+                    </div>
+                </div>
+                <div className={tickerBackground}>
+                    <div className={tickerProgress} style={{width: `${percentage}%`}} />
```
liywjl, line 89: convert `tickerProgress` into a function to dynamically set `width` instead of using styles

src/amp/components/Ticker.tsx
```diff
@@ -0,0 +1,94 @@
+import {css} from 'emotion';
+import {headline} from '@guardian/src-foundations/typography';
+import {neutral, brandAlt} from '@guardian/src-foundations/palette';
+import React from 'react';
+
```
liywjl, line 5: As a convention its nice to have `React` at the top, and try to separate external libs to internal libs

liywjl (changes requested):
Overall this PR looks great, I think there are just some minor stylistic changes to be made, and just want to double check that all options have been exhausted before using `!important`. I also noticed a lot of font style css being applied, would it be better to use some of the source lib instead of defining it on our own here?

src/amp/components/Epic.tsx
michaelbjacobson: ✅

src/amp/components/Ticker.tsx
michaelbjacobson: ✅

src/amp/components/Ticker.tsx
michaelbjacobson: ✅

src/amp/components/Ticker.tsx
michaelbjacobson: ✅

src/amp/components/Ticker.tsx
michaelbjacobson: ✅

michaelbjacobson:
@liywjl thanks for the feedback! I've made most of the changes you've requested, but the suggestion to move the progress bar progression into a function is unfortunately not practical/doable, because of limitations with `amp-list` rendering.

michaelbjacobson:
Can anyone advise how to tackle the build issue with Cypress?

liywjl:
> @liywjl thanks for the feedback! I've made most of the changes you've requested, but the suggestion to move the progress bar progression into a function is unfortunately not practical/doable, because of limitations with `amp-list` rendering.

Awesome! I am not super familiar with AMP so I do not fully understand its limitations, but in that case all looks good 👍 

> Can anyone advise how to tackle the build issue with Cypress?

I have noticed another PR having a similar issue, I will investigate. Thanks for the heads up

On am final though, we have been pushing to use `'@guardian/src-foundations/typography'` for text styling, idk if you would be able to use it in `currentAmountStyle`, `goalAmountStyle` and `amountCaptionStyle` instead of custom defining font styling css?

src/amp/components/Ticker.tsx
michaelbjacobson: As discussed, this isn't doable with our current use of AMP. We'll consider it for future improvements/updates though.

michaelbjacobson:
@liywjl I'd prefer to keep the explicit CSS font sizes for the time being, just to ensure that the AMP ticker's styling is consistent with our other tickers. Is that ok?

liywjl:
> @liywjl I'd prefer to keep the explicit CSS font sizes for the time being, just to ensure that the AMP ticker's styling is consistent with our other tickers. Is that ok?

Whenever setting font sizes in DCR we always try to use `Source` even if this means slight differences with frontend. If there is something you think we need something new in `Source` it is best to reach out to @SiAdcock .

oliverlloyd:
> > @liywjl I'd prefer to keep the explicit CSS font sizes for the time being, just to ensure that the AMP ticker's styling is consistent with our other tickers. Is that ok?
> 
> Whenever setting font sizes in DCR we always try to use `Source` even if this means slight differences with frontend. If there is something you think we need something new in `Source` it is best to reach out to @SiAdcock .

Fun fact: If you look on DCR, the Sub Nav text is 1px different to Frontend. It's moderately noticeable but the call was made to favour using Source over parity. There's actually a PR open to change the font size on frontend to align with Source instead.

michaelbjacobson:
Per feedback from @liywjl, this now makes use of Source typography. See screenshot below for latest styling. 🙂

Screenshot 2020-11-11 at 12 44 23

liywjl (approved):
🎊  Thank you so much for all those changes! I have been working on a fix for the cypress tests that should be out soon https://github.com/guardian/dotcom-rendering/pulls

tomrf1 (approved):
great work! 👍 
Adjusting the post-goal behaviour can be done in a follow up PR

michaelbjacobson:
> 🎊 Thank you so much for all those changes! I have been working on a fix for the cypress tests that should be out soon https://github.com/guardian/dotcom-rendering/pulls

Awesome, I'll keep an eye on that PR 🙂

PR #2030, opened 2020-11-10
Add first commercial cypress test

ioanna0:
## What does this change?
Adds first commercial cypress test
Picks a long read read and expects the right amount of ad slots

Still to check more cases and that the right slots are contained in both dcr and frontend.

Note: Had to wait for 4.5 seconds for all the ads to load

## Why?
Will help find bugs


cypress/integration/e2e/commercial.spec.js
```diff
@@ -0,0 +1,27 @@
...
+
+    beforeEach(function () {
+        // Disable CMP
+        cy.setCookie('gu-cmp-disabled', 'true', {
+            log: true,
+        });
+
+    });
+
+    describe('for WEB', function () {
```
oliverlloyd, line 15: This describe is optional. It made sense before when we different sets of tests but you can delete it here in this context (if you like 🤷 )

cypress/integration/e2e/commercial.spec.js
```diff
@@ -0,0 +1,27 @@
+import { getPolyfill } from '../../lib/polyfill';
+import { fetchPolyfill } from '../../lib/config';
+
+describe('Commercial E2E tests', function () {
+    before(getPolyfill);
+
+    beforeEach(function () {
+        // Disable CMP
+        cy.setCookie('gu-cmp-disabled', 'true', {
```
oliverlloyd, line 9: I suspect you already know this but @liywjl has a function for this now which does the same - not merged yet though...

cypress/integration/e2e/commercial.spec.js
```diff
@@ -0,0 +1,27 @@
...
+            log: true,
+        });
+
+    });
+
+    describe('for WEB', function () {
+        // eslint-disable-next-line mocha/no-setup-in-describe
+        it(`It should check slots for a long article`, function () {
+                const longReadURL = "https://www.theguardian.com/environment/2020/oct/13/maverick-rewilders-endangered-species-extinction-conservation-uk-wildlife";
+                cy.log(`Opening A long read`);
```
oliverlloyd, line 19: `cy.log()`! I never knew you could do this, neat

cypress/integration/e2e/commercial.spec.js
```diff
@@ -0,0 +1,27 @@
...
+                cy.visit(`Article?url=${longReadURL}`, fetchPolyfill);
+
+                cy.scrollTo('bottom', { duration: 4500 });
+
+                cy.get('.js-ad-slot').should('have.length', 15)
```
oliverlloyd, line 24: I really like this method for verifiying the ads loaded as expected. Simple but effective.

cypress/integration/e2e/commercial.spec.js
```diff
@@ -0,0 +1,27 @@
...
+    });
+
+    describe('for WEB', function () {
+        // eslint-disable-next-line mocha/no-setup-in-describe
+        it(`It should check slots for a long article`, function () {
+                const longReadURL = "https://www.theguardian.com/environment/2020/oct/13/maverick-rewilders-endangered-species-extinction-conservation-uk-wildlife";
+                cy.log(`Opening A long read`);
+                cy.visit(`Article?url=${longReadURL}`, fetchPolyfill);
+
+                cy.scrollTo('bottom', { duration: 4500 });
```
oliverlloyd, line 22: Hopefully this duration won't need to be extended but I've seen it in other places (I think the BBC also had similar issues) where this was flakey. But I don't see a better way as things stand so 👍 

Note. Hopefully, in some bright future, we will have callbacks like `onAdLoad` that we can use to count things like this 🤞

oliverlloyd (approved):
👏  for doing this. Adding this test is probably the single biggest improvement in our test coverage in a long time!

gtrufitt:
This is ace! Do you think it might be inconsistent occasionally? Not sure whether you expect the number of ads to be exactly the same every time.

I guess you could test by wrapping the subscribe in a for loop to repeat the test x times and then running that particular spec to see if it passes every time?

liywjl (approved):
🎊  This is a great start to testing the commercial stack in DCR. I am just concerned with the use of `scrollTo` that could make the test somewhat flaky, but I guess that is something we will keep an eye on.

Great work! 👍

cypress/integration/e2e/commercial.spec.js
ioanna0: I've now added a second frontend test and have renamed the describe description

cypress/integration/e2e/commercial.spec.js
ioanna0: I've switched now to 500ms after I realised the long duration was only needed while testing with --headed flag 🙂

ioanna0:
>  Do you think it might be inconsistent occasionally? / > I am just concerned with the use of scrollTo that could make the test somewhat flaky, but I guess that is something we will keep an eye on.

Yes good points @gtrufitt and @liywjl those test could be flaky but from what I saw so far by running them multiple times, theres already a 3 times retry rule which is  enough to make them pass. But definitely should keep an eye on them and monitor their performance  

> Not sure whether you expect the number of ads to be exactly the same every time.
I would expect the same number every time unless we introduce a change in placing ad slots on the page, in which case we should make a change on these tests as well

liywjl:
> > Do you think it might be inconsistent occasionally? /  I am just concerned with the use of scrollTo that could make the test somewhat flaky, but I guess that is something we will keep an eye on.
> 
> Yes good points @gtrufitt and @liywjl those tests could be flaky but from what I saw so far by running them multiple times, there is already a 3 times retry rule which is enough to make them pass. But definitely should keep an eye on them and monitor their performance
> 
> > Not sure whether you expect the number of ads to be exactly the same every time.
> 
> I would expect the same number every time unless we introduce a change in placing ad slots on the page, in which case we should make a change on these tests as well

I agree, it may be flaky, but the benefits outweighs the risk IMO. Also I was thinking, this might be a great opportunity to test the number of Ads on different screen sizes?

mxdvl:
I’ve approved the Chromatic test as it was a case of “8 months” &rarr; “9 month”

mxdvl (approved):
This is an excellent addition to testing the Commercial stack!

I’ve approved the Chromatic test as it was a case of “8 months” &rarr; “9 month”.

oliverlloyd (approved):
Hearty approval

I merged in main to fix the failing tests - LGTM!

cypress/integration/e2e/commercial.spec.js
```diff
@@ -36,9 +36,12 @@ describe('Commercial E2E tests', function () {
...
-            Cypress.config('baseUrl', '');
-            runLongReadTestFor(`${longReadURL}?dcr=false`);
-        });
+        // Uncomment to check Locally.
+        // It seems overriding the baseURL works fine locally but hangs on CI.
+
+        // it(`It should check slots for a long article in Frontend`, function () {
+        //     Cypress.config('baseUrl', '');
+        //     runLongReadTestFor(`${longReadURL}?dcr=false`);
+        // });
```
mxdvl, line 45: You can use Jest.skipOn &rarr; https://github.com/guardian/consent-management-platform/blob/95d97ecb13c45158bf9b1fae0cc284eb588653e1/cypress/integration/sourcepoint-tcfv2.spec.js#L41

ioanna0: Ooos I saw that after I merged! Let me do that on a subsequent PR

PR #2034, opened 2020-11-11
Set Radio selection vertically for callout on condition

liywjl:
## What does this change?
Change the orientation of Radio buttons in Callout component depending on the number of radio inputs

### Before
Screenshot 2020-11-11 at 12 40 18

### After
2 Radio
Screenshot 2020-11-11 at 13 24 20

More than 2 Radio
Screenshot 2020-11-11 at 12 39 51

## Why?
As mentioned by @SiAdcock https://github.com/guardian/source/issues/580#issuecomment-718728816, it is recommended  to use horizontal radio buttons if there are only 2 answers.

### DIFF
--- a/src/web/components/Callout/RadioSelect.tsx
+++ b/src/web/components/Callout/RadioSelect.tsx
@@ -25,7 +25,7 @@ export const RadioSelect = ({
         `}
     >
         <FieldLabel formField={formField} />
-        <RadioGroup name={formField.name} orientation="horizontal">
+        <RadioGroup name={formField.name} orientation="vertical">
             {formField.options.map((option, index) => {
                 const isRadioChecked =
                     formField.id in formData &&
--- a/src/web/components/Callout/RadioSelect.tsx
+++ b/src/web/components/Callout/RadioSelect.tsx
@@ -25,7 +25,12 @@ export const RadioSelect = ({
         `}
     >
         <FieldLabel formField={formField} />
-        <RadioGroup name={formField.name} orientation="vertical">
+        <RadioGroup
+            name={formField.name}
+            orientation={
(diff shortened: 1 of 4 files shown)


src/web/components/Callout/RadioSelect.tsx
```diff
@@ -25,7 +25,12 @@ export const RadioSelect = ({
         `}
     >
         <FieldLabel formField={formField} />
-        <RadioGroup name={formField.name} orientation="horizontal">
+        <RadioGroup
+            name={formField.name}
+            orientation={
+                formField.options.length > 2 ? 'vertical' : 'horizontal'
```
oliverlloyd, line 31: 👍

oliverlloyd (approved):

oliverlloyd:
In the Chromatic snapshot it looks like the radio field with 2 options is not showing horizontally

Screenshot 2020-11-11 at 13 39 50

Is that desired?

liywjl:
> In the Chromatic snapshot it looks like the radio field with 2 options is not showing horizontally
> 
> Screenshot 2020-11-11 at 13 39 50
> 
> Is that desired?

Yes so that is an interesting point, this is because Chromatic uses an older version of Chrome that has a bug in `flex` https://github.com/guardian/source/issues/580 , the bug has since been fixed

oliverlloyd:
> > In the Chromatic snapshot it looks like the radio field with 2 options is not showing horizontally
> > Screenshot 2020-11-11 at 13 39 50
> > Is that desired?
> 
> Yes so that is an interesting point, this is because Chromatic uses an older version of Chrome that has a bug in `flex` [guardian/source#580](https://github.com/guardian/source/issues/580) , the bug has since been fixed

Wild. But out of our control so 👍

PR #2035, opened 2020-11-11
Update ophan tracker js to 1.3.23

mchv:
## What does this change?

We may receive garbage in iframe message that will result in client side errors which we could safely ignore.
We currently have about [10k errors per day in Sentry with this problem](https://sentry.io/organizations/the-guardian/issues/1813121539/?project=35463&query=is%3Aunresolved) 

See https://github.com/guardian/ophan/pull/3964


oliverlloyd (approved):
Looks good 👍

liywjl (approved):

PR #2036, opened 2020-11-11
Added timings for the Braze SDK load and Braze 'appboy'-related calls

rhystmills:
Equivalent PR for Frontend [here](**url**).

## What does this change?
This adds additional timing measurements. We (TX) want to collect RUM data, which we will pull from Ophan in order to see which calls are taking up the most time. Currently our banner exceeds its time limit roughly 1 in 3 page views where it would otherwise be valid to show.

## Why?
We want to figure out what takes the most time, in order to identify where improvements are made. We are especially interested in Braze's API calls and SDK load time.

### DIFF
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -92,7 +92,7 @@ const getMessageFromBraze = async (
         minimumIntervalBetweenTriggerActionsInSeconds: 0,
     });
 
-    return new Promise((resolve) => {
+    const canShowPromise: Promise<CanShowResult> = new Promise((resolve) => {
 
         appboy.subscribeToInAppMessage((message: any) => {
             const { extras } = message;
@@ -116,12 +116,6 @@ const getMessageFromBraze = async (
             };
 
             if (extras) {
-                const appboyTimeTaken = appboyTiming.end()
-                record({
-                    component: 'braze-appboy-timing',
-                    value: appboyTimeTaken,
-                });
-
                 const meta = {
                     dataFromBraze: extras,
                     logImpressionWithBraze,
@@ -137,6 +131,20 @@ const getMessageFromBraze = async (
         appboy.changeUser(brazeUuid);
         appboy.openSession();
     });
+
+    canShowPromise.then(() => {
+        const appboyTimeTaken = appboyTiming.end();
+
+        record({
+            component: 'braze-appboy-timing',
+            value: appboyTimeTaken,
+        });
+    }).catch(() => {
+        // eslint-disable-next-line no-console
+        console.log("Appboy Timing failed.");
+    });
+
+    return canShowPromise
 };
 
 const getBrazeMetaFromQueryString = (): Meta | null => {
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -140,8 +140,8 @@ const getMessageFromBraze = async (
             value: appboyTimeTaken,
         });
     }).catch(() => {
-        // eslint-disable-next-line no-console
         appboyTiming.clear();
+        // eslint-disable-next-line no-console
         console.log("Appboy Timing failed.");
     });
 
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -141,6 +141,7 @@ const getMessageFromBraze = async (
         });
     }).catch(() => {
         // eslint-disable-next-line no-console
+        appboyTiming.clear();
         console.log("Appboy Timing failed.");
     });
 
@@ -232,6 +233,7 @@ export const canShow = async (
 
         return result;
     } catch (e) {
+        bannerTiming.clear();
         return { result: false };
     }
 };
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -61,13 +61,25 @@ export const canShowPreChecks = ({
             !pageConfig.isPaidContent,
     );
 
+const appboyTiming = initPerf('braze-appboy');
+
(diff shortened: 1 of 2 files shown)


src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -78,6 +90,8 @@ const getMessageFromBraze = async (
     });
 
     return new Promise((resolve) => {
+
+        appboyTiming.start();
```
tjmw, line 94: It feels to me like the `appboyTiming` should start when we initialize the SDK and stop/record when (if) we get a message back. What do you think? That would allow us to scope the `const appboyTiming = initPerf('braze-appboy');` more closely to the timing calls, instead of having it be global at the top. So init and start just before `appboy.initialize` and then stop/record when the promise we're in here resolves. What do you think?

tjmw (approved):
This looks good to me. I left one comment about the scope of the appboy timing, see what you think.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
rhystmills: Hey Tom, thanks for looking over this! Sounds good to me - I've updated the PR now, let me know what you think.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -100,11 +116,18 @@ const getMessageFromBraze = async (
             };
 
             if (extras) {
+                const appboyTimeTaken = appboyTiming.end()
```
tjmw, line 119: Nice! Yeah I think this will more accurately measure the time taken from braze initialising to getting a message. One small improvement we could make is that in _theory_ this callback could fire multiple times, though we only care about the first one. The way we handle that in terms of what `canShow` returns is through the promise we return in that a promise is guaranteed to only resolve once. So if this callback did fire multiple times, only the first call would have any effect outwardly (i.e. the first time we call `resolve` in here). So it feels to me like the most correct thing would be to handle the timing end when the promise resolves e.g.:

So instead of:

```js
return new Promise((resolve) => {
  // ...
})
```

We could do:
```js
const canShowPromise = new Promise((resolve) => {
  // ...
});

canShowPromise.then((result) => {

  record({
    value: appboyTimeTaken,
  });
}).catch(() => {
  console.log("Something went wrong recording timing");
});

return canShowPromise;
```

What do you reckon?

rhystmills: Ah yes, good point - I hadn't thought about multiple calls. This makes perfect sense to me.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -114,6 +131,20 @@ const getMessageFromBraze = async (
...
+        const appboyTimeTaken = appboyTiming.end();
+
+        record({
+            component: 'braze-appboy-timing',
+            value: appboyTimeTaken,
+        });
+    }).catch(() => {
+        // eslint-disable-next-line no-console
+        console.log("Appboy Timing failed.");
+    });
```
liywjl, line 145: this is more of a question, but would there be any benefit to running `const appboyTimeTaken = appboyTiming.end();` here?

liywjl (approved):
LGTM 👍 , just curious about adding `const appboyTimeTaken = appboyTiming.end();` in the `catch` block.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
rhystmills: Thanks for looking over this! My feeling is we would not want to log a timing value if the promise results in an error - though we could handle the error more elegantly than we do here.

What do you reckon @tjmw ?

liywjl: I thought the logging was done by `record`? I was snooping around on MDN and found this https://developer.mozilla.org/en-US/docs/Web/API/Performance/clearMarks   Might be worth creating another export in `initPerf` like:

```
const close = () => {
        perf. clearMarks(startKey);
    };
```

we could then close it by running `appboyTiming .close()`

rhystmills: You are right, I had assumed you were suggesting we `.end()` and `record` since there would not be much value in running `.end()` by itself given that it will add a second performance mark and measure the difference between the two without using the value to do anything.

We could clear the mark - but I don't know if the performance mark sitting in memory somewhere will have a greater impact on performance than any attempt we could make to remove/close it e.g. via `clearMarks`. My assumption is that that they don't persist beyond the page (and are discarded), but I don't know for sure, nor can I find the answer in the documentation.

liywjl: I think if you have started a performance mark, and are not going to end it, it makes sense to close it in the catch block. I doubt this really has any major performance benefit, but I guess its just good practice to stop unnecessary background calculations.

tjmw: Hey I agree with @liywjl that it does _feel_ cleaner to clear the mark in the `catch` block, though also agree that there's probably not a significant performance benefit to doing so (I'm not too sure how the performance API is implemented under the hood - but presumably not clearing would just mean an unused mark entry stored somewhere?) There's probably also an argument that clearing the mark makes the intent clearer to future developers reading the code. I reckon let's go ahead an add this since it seems like a pretty straightforward change.

rhystmills: Cool, I've added a .clear() method along the lines that @liywjl suggested, which we can call to tidy up marks when we don't call .end() - i.e. when we catch an error.

liywjl: Thanks @rhystmills I just noticed you have a lint issue because `// eslint-disable-next-line no-console` should be one line down. Bar that all looks good to merge

rhystmills: So it is, thanks for catching that.

PR #2014, opened 2020-10-29
Add Newsletter merch unit AB test

buck06191:
## What does this change?
DCR implementation of: https://github.com/guardian/frontend/pull/23187
In order to test the impact of a newsletters merch unit across
lighthouse segments we want to roll out a 50/50 AB test. The variant
will then be targeted with the newsletters merch unit.

## Why?
Newsletter engagement OKR

PR #2037, opened 2020-11-11
Reporting the AB test from dotcom

alexduf:
### What does this change?
See https://github.com/guardian/frontend/pull/23230

Adding the newsletter merch a/b test to match the dotcom implementation

Requires testing to confirm the opt-in and tracking

### Before
N/A
### After
N/A

## Why?
This A/B test should help us understand if there's value in using the merch space to promote our own newsletters

### DIFF
--- a/src/web/experiments/tests/newsletter-merch-unit-test.ts
+++ b/src/web/experiments/tests/newsletter-merch-unit-test.ts
@@ -26,9 +26,10 @@ export const newsletterMerchUnitLighthouseVariants: ABTest = {
     start: '2020-11-11',
     expiry: '2020-12-01',
     author: 'Josh Buckland & Alex Dufournet',
-    description: 'Show a newsletter advert in the merchandising unit to 25% of users. ' +
-        'These two variants test value of showing newsletter merch units instead of reader revenue ones. ' +
-        'This test needs to run at the same time as NewsletterMerchUnitLighthouseControl',
+    description: 'Show a newsletter advert in the merchandising unit to 25% of users and reader revenue' +
+        'to another 25%. The remaining 50% are covered by NewsletterMerchUnitLighthouseControl ' +
+        'which needs to run at the same time. These two variants test the value of showing ' + 
+        'newsletter merch units instead of reader revenue ones. ',
     audience: 0.5,
     audienceOffset: 0.5,
     successMeasure: 'We see increased engagement from users shown the Newsletters ad unit',
--- a/src/web/experiments/tests/newsletter-merch-unit-test.ts
+++ b/src/web/experiments/tests/newsletter-merch-unit-test.ts
@@ -1,12 +1,12 @@
 import { ABTest } from '@guardian/ab-core';
 
-export const newsletterMerchUnitLighthouse: ABTest = {
-    id: 'NewsletterMerchUnitLighthouse',
+export const newsletterMerchUnitLighthouseControl: ABTest = {
+    id: 'NewsletterMerchUnitLighthouseControl',
     start: '2020-11-11',
     expiry: '2020-12-01',
     author: 'Josh Buckland & Alex Dufournet',
     description: 'Show a newsletter advert in the merchandising unit to 50% of users',
-    audience: 1.0,
+    audience: 0.5,
     audienceOffset: 0.0,
     successMeasure: 'We see increased engagement from users shown the Newsletters ad unit',
     audienceCriteria: 'Website users only.',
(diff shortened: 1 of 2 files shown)


src/web/experiments/tests/newsletter-merch-unit-test.ts
```diff
@@ -0,0 +1,38 @@
+import { ABTest } from '@guardian/ab-core';
+
+export const newsletterMerchUnitLighthouse: ABTest = {
+    id: 'NewsletterMerchUnitLighthouse',
+    start: '2020-11-11',
+    expiry: '2020-12-01',
+    author: 'Josh Buckland & Alex Dufournet',
+    description: 'Show a newsletter advert in the merchandising unit to 50% of users',
+    audience: 1.0, // 0.01%
```
gtrufitt, line 9: Comment and audience doesn't match, 1.0 is 100% :)

src/web/experiments/tests/newsletter-merch-unit-test.ts
```diff
@@ -0,0 +1,38 @@
...
+    variants: [
+        {
+            id: 'control',
+            test: (): void => {},
+            impression: (impression) => {
+                impression();
+            },
+            success: (success) => {
+                success();
+            },
```
gtrufitt, line 25: This will immediately fire impression and success for your variants. 

Normally how you use this is by setting up a listener to call the callback like this:

```
 success: (success) => {
                // https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events
                window.addEventListener('myTestSuccess', (e) => {
					success()                
				});

            },
```

Which you can then fire as a custom event like `window.dispatchEvent(new CustomEvent('myTestSuccess')`

buck06191: I'm not sure we need success and impression to fire in this case. The test is being used to pass key-values to ad manager which will then use it for targeting. If memory serves, we should be getting all impressions and success/conversion data from the existing tracking around the merch unit

alexduf: So given this is going to be handled by ad-manager, should I leave the implementation empty?
```
            impression: (impression) => {},
```

src/web/experiments/tests/newsletter-merch-unit-test.ts
```diff
@@ -0,0 +1,49 @@
...
+};
+
+export const newsletterMerchUnitLighthouseVariants: ABTest = {
+    id: 'NewsletterMerchUnitLighthouseVariants',
+    start: '2020-11-11',
+    expiry: '2020-12-01',
+    author: 'Josh Buckland & Alex Dufournet',
+    description: 'Show a newsletter advert in the merchandising unit to 25% of users. ' +
+        'These two variants test value of showing newsletter merch units instead of reader revenue ones. ' +
+        'This test needs to run at the same time as NewsletterMerchUnitLighthouseVariants',
```
buck06191, line 31: I think this is meant to be
```suggestion
        'This test needs to run at the same time as NewsletterMerchUnitLighthouseControl',
```

src/web/experiments/tests/newsletter-merch-unit-test.ts
```diff
@@ -0,0 +1,49 @@
+import { ABTest } from '@guardian/ab-core';
+
+export const newsletterMerchUnitLighthouseControl: ABTest = {
+    id: 'NewsletterMerchUnitLighthouseControl',
+    start: '2020-11-11',
+    expiry: '2020-12-01',
+    author: 'Josh Buckland & Alex Dufournet',
+    description: 'Show any merch unit to 50% of users. This is the control for the NewsletterMerchUnitLighthouseVariants test.',
```
buck06191, line 8:
```suggestion
    description: 'Show BAU merch unit to 50% of users. This is the control for the NewsletterMerchUnitLighthouseVariants test.',
```

buck06191 (approved):
Couple of changes to make for clarity and then it  can be merged. We need to also remember to do the same on https://github.com/guardian/frontend/pull/23230

src/web/experiments/tests/newsletter-merch-unit-test.ts
alexduf: Thanks

src/web/experiments/tests/newsletter-merch-unit-test.ts
```diff
@@ -0,0 +1,49 @@
...
+};
+
+export const newsletterMerchUnitLighthouseVariants: ABTest = {
+    id: 'NewsletterMerchUnitLighthouseVariants',
+    start: '2020-11-11',
+    expiry: '2020-12-01',
+    author: 'Josh Buckland & Alex Dufournet',
+    description: 'Show a newsletter advert in the merchandising unit to 25% of users. ' +
+        'These two variants test value of showing newsletter merch units instead of reader revenue ones. ' +
+        'This test needs to run at the same time as NewsletterMerchUnitLighthouseControl',
```
buck06191, line 31:
```suggestion
    description: 'Show a newsletter advert in the merchandising unit to 25% of users and reader revenue' +
        'to another 25%. The remaining 50% are covered by NewsletterMerchUnitLighthouseControl ' +
        'which needs to run at the same time. These two variants test the value of showing ' + 
        'newsletter merch units instead of reader revenue ones. ',
```

Updating this for clarity after looking at it in the admin panel on frontend.

buck06191 (commented):
Small copy change

buck06191:
Closes #2014

PR #2040, opened 2020-11-12
Support posterImage variable for youtube overlay

liywjl:
## What does this change?
Adds support for `posterImage`,  `overlayImage` still takes precedence over `posterImage` if present

example article: https://www.theguardian.com/us-news/2020/oct/29/lordstown-ohio-trump-gm-plant-election

### Before
Screenshot 2020-11-12 at 14 55 05

### After
Screenshot 2020-11-12 at 14 52 48

## Why?
Parity


src/web/components/elements/YoutubeBlockComponent.tsx
```diff
@@ -76,6 +78,35 @@ export const YoutubeBlockComponent = ({
...
+                ): PosterImageType => {
+                    if (wrapperWidth < cur.width && cur.width < acc.width)
+                        return cur;
+                    return acc;
+                },
+                posterImage[0],
+            );
+            setSelectedPosterImage(bestFitPosterImage);
+        }
+    }, [wrapperRef, posterImage, setSelectedPosterImage]);
```
oliverlloyd, line 108: I'm not an expert on this but does the `picture` tag or maybr the `Image` component perhaps already do this?

shtukas (approved):

src/web/components/elements/YoutubeBlockComponent.tsx
liywjl: yes it does!!! Ill split it out into `lib`. Great catch!

oliverlloyd: I was more thinking how the `Picture` component passes the array of sources to the `picture` and lets - I guess 🤷 - the browser work out sizes.

This:

```
const forSource: (source: PictureSource) => string = (source) =>
    ` <source media="${mq(source)}" sizes="${source.width}px" srcset="${
        source.srcset
    }" />`;

export const Picture: React.FC<{
    sources: PictureSource[];
    alt: string;
    src: string;
    height: string;
    width: string;
    isLazy?: boolean;
}> = ({ sources, alt, src, height, width, isLazy = true }) => {
    return (
        // https://stackoverflow.com/questions/10844205/html-5-strange-img-always-adds-3px-margin-at-bottom
        // why did we put `style="vertical-align: middle;"` inside the img tag
        <picture
            dangerouslySetInnerHTML={{
                __html: `<!--[if IE 9]><video style="display: none;"><![endif]-->${sources
                    .map(forSource)
                    .join(
                        '',
                    )}<!--[if IE 9]></video><![endif]--><img style="vertical-align: middle;" itemprop="contentUrl" alt="${alt}" src="${src}" height="${height}" width="${width}" ${
                    isLazy ? 'loading="lazy"' : ''
                } />`,
            }}
        />
    );
};
```

Could we do something similar here so we don't have to have the rerender from useEffect?

oliverlloyd: Not sure if we still need the IE9 workaround though 🤔

gtrufitt: +1, I think it's something we should avoid doing client side if possible!

liywjl: Totally agree with @gtrufitt, I have always been reluctant to be doing this client side. The only issue is that if we have a list of candidate overlay images with different widths, there is no way not to have logic for the best match.

We could however lift out the Overlay logic out of the youtube atom package and into DCR, that would allow us to tinker with the logic without having to do package updates. But fundamentally I do not see any easy work around.

The examples used here https://github.com/guardian/dotcom-rendering/pull/2040/files#diff-58a048c4d236ea52a26db21d08e39d97e2774b1e9547884b6b519034133fe0d0R3-R24 actually come from real JSON examples

gtrufitt: This is what srcset is for though I think? The image array that you get for a poster is in the same (I think) format as any other images, it seems that we should probably be passing the return value of the `Picture` component (given that array) to the atom and using that rather than having a background image in the atom? You'd need to server render that image, but I think that's the correct approach anyway? 

It's a bit gnarly and I don't 100% understand it so happy to have a chat / mob with the three of us on Monday though!

liywjl:
https://github.com/guardian/dotcom-rendering/pull/2298 poster image was determined within Atomm

PR #2042, opened 2020-11-12
Introduce lighthouse-ci performance assertions

jorgeazevedo:
## What does this change?

Last quarter we introduced lighthouse-ci as a github action, which generates lighthouse reports for every PR. We hoped we could use it to detect regressions in performance before they ship to production, but it turned the measurements were too unstable.

We try to address this in this PR by enabling caching so as to remove the instability of the network from the picture. This significantly reduced how much the measurements oscillate. The following table contains the minimum and maximum value measured for different metrics over the course of 10 runs.

| Metric  | Min | Max | Threshold |
| ------------- | ------------- | ------------- | ------------- |
| First Contentful Paint  | 0.6s  | 1.0s | **1.0s** |
| Largest Contentful Paint  | 2.3s | 3.0s | **3.0s** |
| Time to Interactive  | 2.6s | 3.3s | **3.5s** |
| Cumulative Layout Shift  | 0  | 0.001 | **0.001** |

The table also contains a set of proposed thresholds that we can base assertions on. Me and @jamie-lynch paired on writing assertions for these, which means making the lighthouse-ci status check **fail** if the measurements go above the threshold. 

## Why?

Hopefully this will help us detect significant downgrades in performance before they hit production!


oliverlloyd (approved):
This looks great. 👏  for the work that's gone into this!

I'm very curious to see how these assertions play out. Is the plan to turn these into errors at some point?

jorgeazevedo:
@oliverlloyd thanks! So this will make the status check fail, but lighthouse is not a "required check" so it won't block merging.

PR #2043, opened 2020-11-12
Switch PR Deployments to working on demand

jorgeazevedo:
## What does this change?

This makes the PR Deployment action run on demand, by labelling the PR with the "PR Deployment" label, rather than automatically on PR open and push to branch.

Pull_requests_·_guardian_dotcom-rendering

## Why?

The idea of having a PR easily deployed like this has proven useful, especially the public url bit, but it's not used very frequently. The system as it stands has a number of annoyances that can become quite grating when it runs automatically. On demand though, it's less annoying:
- You won't get a failure email when you push to the branch
- It won't block dependabot
- You won't get an ever running action if you don't request it


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,9 +1,3 @@
-<!-- YOU SHOULD KNOW: dotcom is running an experiment called PR Deployments. -->
```
gtrufitt, line 1: Might be useful to leave a message here to explain how to get a live link using the label?

gtrufitt (approved):
Great!

.github/PULL_REQUEST_TEMPLATE.md
jorgeazevedo: Can do!

jorgeazevedo:
@gtrufitt it's a bit confusing that we're getting lighthouse reports from both the github action and PRbuilds. We could deactive PRbuilds for the repo if we could implement the amp validation as a github action? Or maybe we don't really need that validation anymore?

Switch_PR_Deployments_to_working_on_demand_by_jorgeazevedo_·_Pull_Request__2043_·_guardian_dotcom-rendering

gtrufitt:
@jorgeazevedo I think that amp test is not really sufficient enough to give us confidence so I agree deactivating PR builds now makes sense!

jorgeazevedo:
@gtrufitt ok cool! I've deactivated the webhook and removed the config files.

PR #2049, opened 2020-11-13
Remove split on guardian and non guardian dependabot

gtrufitt:
## What does this change?

You can't have two targets of the same npm in dependabot, so setting back to just the one for all npm modules


liywjl (approved):
👍

shtukas (approved):

PR #2070, opened 2020-11-16
DropCap Alignment

oliverlloyd:
A reader @blieque very kindly [pointed out](https://twitter.com/Blieque/status/1328131736060956673) that our DropCap looks, well, a bit rubbish really so this PR implements their suggested improvement of aligning it better with the surrounding text

### Before
Screenshot 2020-11-16 at 13 56 16

### After

Props to @blieque for not only pointing the problem out but actually providing a fix as well

### DIFF
--- a/src/web/components/DropCap.tsx
+++ b/src/web/components/DropCap.tsx
@@ -62,7 +62,7 @@ const outerStyles = (pillar: Pillar, designType: DesignType) => {
 const innerStyles = (designType: DesignType) => {
     const baseStyles = css`
         ${headline.large({ fontWeight: 'bold' })}
-        font-size: 121px;
+        font-size: 118px;
         line-height: 99px;
         vertical-align: text-top;
         pointer-events: none;
--- a/src/web/components/DropCap.tsx
+++ b/src/web/components/DropCap.tsx
@@ -63,7 +63,7 @@ const innerStyles = (designType: DesignType) => {
     const baseStyles = css`
         ${headline.large({ fontWeight: 'bold' })}
         font-size: 121px;
-        line-height: 0.82;
+        line-height: 99px;
         vertical-align: text-top;
         pointer-events: none;
         margin-right: ${space[1]}px;
--- a/src/web/components/DropCap.tsx
+++ b/src/web/components/DropCap.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { css } from 'emotion';
 import { headline } from '@guardian/src-foundations/typography';
 import { opinion } from '@guardian/src-foundations/palette';
+import { space } from '@guardian/src-foundations';
 import { pillarPalette } from '@frontend/lib/pillars';
 
 type Props = {
@@ -18,7 +19,7 @@ const outerStyles = (pillar: Pillar, designType: DesignType) => {
         float: left;
         text-transform: uppercase;
         box-sizing: border-box;
-        margin-right: 4px;
+        margin-right: ${space[1]}px;
     `;
 
     /*
@@ -60,12 +61,12 @@ const outerStyles = (pillar: Pillar, designType: DesignType) => {
 
 const innerStyles = (designType: DesignType) => {
     const baseStyles = css`
-        font-size: 109px;
-        line-height: 90px;
-
-        display: inline-block;
+        ${headline.large({ fontWeight: 'bold' })}
+        font-size: 121px;
+        line-height: 0.82;
         vertical-align: text-top;
         pointer-events: none;
+        margin-right: ${space[1]}px;
     `;
 
     switch (designType) {
--- a/src/web/components/Dropcap.stories.tsx
+++ b/src/web/components/Dropcap.stories.tsx
@@ -29,15 +29,15 @@ export const Article = () => {
                     ${body.medium()};
                 `}
             >
-                <DropCap designType="Article" letter="O" pillar="news" />
-                nce upon a time there was a dropcap. Lorem ipsum dolor sit amet,
-                consectetur adipiscing elit, sed do eiusmod tempor incididunt ut
-                labore et dolore magna aliqua. Ut enim ad minim veniam, quis
-                nostrud exercitation ullamco laboris nisi ut aliquip ex ea
-                commodo consequat. Duis aute irure dolor in reprehenderit in
-                voluptate velit esse cillum dolore eu fugiat nulla pariatur.
-                Excepteur sint occaecat cupidatat non proident, sunt in culpa
-                qui officia deserunt mollit anim id est laborum.
+                <DropCap designType="Article" letter="T" pillar="news" />
+                here once was a dropcap. Lorem ipsum dolor sit amet, consectetur
+                adipiscing elit, sed do eiusmod tempor incididunt ut labore et
+                dolore magna aliqua. Ut enim ad minim veniam, quis nostrud
+                exercitation ullamco laboris nisi ut aliquip ex ea commodo
+                consequat. Duis aute irure dolor in reprehenderit in voluptate
+                velit esse cillum dolore eu fugiat nulla pariatur. Excepteur
+                sint occaecat cupidatat non proident, sunt in culpa qui officia
+                deserunt mollit anim id est laborum.
             </p>
         </Container>
     );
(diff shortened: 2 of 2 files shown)


oliverlloyd:
Some useful [visual snapshots](https://www.chromatic.com/build?appId=5dfcbf3012392c0020e7140b&number=3444) on this one

liywjl (approved):
LGTM 👍

oliverlloyd:
@paperboyo just wanted to make sure I'm not missing any strange use cases with this change?

blieque:
Many thanks! Didn't realise the source was public; nice!

On a completely different note, I think something's wrong with your responsive image markup. A lot of the markup is OK, so a lot of the image scales probably still work, but perhaps not as intended. Looking at the image again, I think the HTML entities might be OK but the other parts are still odd. The example is from the homepage. Happy to file an issue if that's easier.

Screenshot 2020-08-19 at 22 45 20

gtrufitt:
> On a completely different note, I think something's wrong with The Guardian's responsive image markup. A lot of the markup is OK, so a lot of the image scales probably still work, but perhaps not as intended. Looking at the image again, I think the HTML entities might be OK but the other parts are still odd. The example is from the homepage. Happy to file an issue if that's easier.

Thanks for this @blieque - we can pull into our internal Trello board. I can see the issue on our homepage, appreciate you raising it!

paperboyo:
Thanks for that @oliverlloyd and, especially, thank you @blieque for raising! This design wasn’t adjusted for when [new fonts](https://github.com/guardian/frontend/issues/19550) with [new vertical metrics](https://github.com/guardian/dotcom-rendering/pull/1439#issuecomment-624362894) were introduced – hence the ugliness.

As `O` has overshoots, so its shape extends beyond baseline and cap height, it’s hard to see from screenshot if it aligns perfectly (`T` would be a better test), but it’s close enough, even if not perfect. This should apply to all uses of drop caps, but the only odd one likely to be forgotten about would be the usual suspect Keyser Söz… Photo Essay.

Oh, and `drop-cap__inner`, see middle copy eg [here](https://www.theguardian.com/news/2020/nov/12/western-travel-influencers-social-media-pakistan-politics).

oliverlloyd:
> As `O` has overshoots, so its shape extends beyond baseline and cap height, it’s hard to see from screenshot if it aligns perfectly (`T` would be a better test), but it’s close enough, even if not perfect. This should apply to all uses of drop caps, but the only odd one likely to be forgotten about would be the usual suspect Keyser Söz… Photo Essay.
> 

You're right. Ts and Ls are much better examples for testing. I updated the stories to include these and from doing that I noticed that we were still slightly off so I've pushed a tweak to fix that, thanks for the pointer 👍

If you want to see the visual before / after snapshots they're [here](https://www.chromatic.com/build?appId=5dfcbf3012392c0020e7140b&number=3458)

 
> Closes https://trello.com/c/Hpp7Lg25/1854-change-dropcap-position.

I've archived that card ^^

oliverlloyd:
@blieque thanks again for your help. This is merged now and should be live in 10 mins or so. Be aware that this is fixed on our new React version of the site (dotcom-rendering) which you may or may not be in a test bucket to see. You can force it though, just add `?dcr=true` to the end of any article url eg.

https://www.theguardian.com/tv-and-radio/2020/nov/16/i-feel-joy-bisha-k-ali-struggling-standup-ms-marvel-maestro-muslim-superhero?dcr=true

paperboyo:
> If you want to see the visual before

Thx! It looks, like they can still be (hopefully, consistently) raised a tad? Ideally, the cap height of a drop cap would be aligned to an x height of first line of text at the top and the baselines of a drop cap and fourth line of body should be aligned too. But no biggie if it’s fiddly or a whack-a-mole!

image

oliverlloyd:
Hmm, you know it does seem to be slightly off in Chrome and Safari but not in Firefox. Only by a pixel or two but you see it with a T. I'll have another look tomorrow.

paperboyo:
You’re right! Seems to be off for some zoom levels only too. On PC and Mac 125% zoom in Chrome is consistently off for me (Firefox good all the way). Cannot be the usual different interpretation of fonts’ vertical metrics, as we had them fixed, so platforms should be using the same ones and zoom level thingie suggests something else anyway.

May be a bug in Chrome (?). May be that we will need to wait for [`leading-trim`](https://www.w3.org/TR/css-inline-3/#propdef-leading-trim) (internally discussed [here](https://chat.google.com/u/1/room/AAAAWwBdSMs/v9unN7b0cTU)). [This](https://opentype.js.org/) doesn’t apply, I suppose.

blieque:
Bug report to deployment in under 24 hours – nice work! Pleasure to help.

@paperboyo What you mentioned about alignment at the top is interesting. Is it more traditional to align the top of a drop cap to the top of a lowercase *x* than an uppercase *X*? Most of the examples I can find seem to opt for the latter:

||||
|-|-|-|
| One | Two | Three) |

That said, this one opts for the former:

Four

There's a particularly-edgy edge case of articles starting with an initialism and drop cap, in which the latter would probably be preferable. Not sure how often "NASUWT" introduces a Lifestyle article though. 🙃

Five

It looks like the `margin-right` is doubled up now with the two elements (`.drop-cap`, `.drop-cap__inner`) – not sure if that's a mistake or not.

Other than that, Chromium and Firefox on macOS look good to me. Safari seems to be jumping around a bit though (drop cap is always a little too high or low). I might have been optimistic to suggest just relying on `line-height`. `leading-trim` can't come soon enough.

paperboyo:
My kinda comment, 👏, thank you!

> Is it more traditional …

When first spotted, I went with what I remembered being an original design, looked for some examples and queried design here. But, if I’m honest, I wouldn’t know, I’m no designer. I can post-rationalise, though, no problemo.

For me, it’s about aligning to the optically more pronounced line: so would the first line always be capped, it would be cap height. Since it’s most often not – it’s x-height. For the same reason, the bottom isn’t aligned to descenders. A bit like hanging punctuation. (So: yes, would it be possible, I would maybe align horizontally to the main stems, not to serifs, like in your third _Pexperci_ example above.)

Guardian style for acronyms is to cap only the initial letter ([here](https://www.theguardian.com/guardian-observer-style-guide-a#abbreviations%20and%20acronyms); anchor link won’t work, sadly), so the example you gave happens even less often, but when it does, it’s an acceptable compromise for it looking better most of the time, IMHO.

I worry a bit that leading-trim still seems to depend on legacy and often confused font metadata (see [The Webfont Strategy](https://glyphsapp.com/tutorials/vertical-metrics)) of [text-edge](https://www.w3.org/TR/css-inline-3/#ascent-descent), but that’s a deeper OpenType issue with a looong beard. [Still though](https://typedrawers.com/discussion/3764/issues-related-to-new-css-leading-trim-feature), it’s gonna be great to have it, I’m excited.

Oh, and from my esteemed engineering colleagues, I wouldn’t expect a deployment longer than 5m 😃.

blieque:
That all makes sense. As you mention, two uppercase letters at the start will be very rare. Optimise for the common case.

Thanks for the links; I'll have a read!

blieque:
Like any good comic-book villain, *I return.*

I think this one reappeared a week or two ago. The sizing is different this time, so it's not strictly a regression. The change doesn't seem to have been in the DropCap component, but perhaps some font files or `@font-face` rules were altered recently. Examples below – the drop-cap should be slightly smaller, nudged slightly left, and possibly nudged up or down a bit. An uppercase drop-cap "X" should align with the top of a lowercase "x" on the first line and the bottom of a lowercase "x" on the fourth line.

|[Article](https://www.theguardian.com/film/2023/aug/20/adele-exarchopoulos-actor-passages-interview-whishaw-blue-warmest-colour)|[Article](https://www.theguardian.com/sport/2023/aug/18/cricket-adil-rashid-interview-england-world-cup-northern-superchargers-the-hundred)|[Article](https://www.theguardian.com/food/2023/aug/20/luke-selby-raymond-blanc-le-manoir-aux-quatsaisons)|
|:-|:-|:-|
|Drop-cap example 01|Drop-cap example 02|Drop-cap example 03|

paperboyo:
thorgal-alinoe-2

Yep, line height changes overlooked this, good spot! Issue: https://github.com/guardian/dotcom-rendering/pull/8594.

blieque:
You're already on it – should have known! 😄

PR #2072, opened 2020-11-16
Use VideoAtom in DCR

oliverlloyd:
With this PR, if DCR is sent:

```typescript
interface MediaAtomBlockElement {
    _type: 'model.dotcomrendering.pageElements.MediaAtomBlockElement';
    id: string;
    assets: VideoAssets[];
    posterImage?: {
        url: string;
        width: number;
    }[];
    title?: string;
    duration?: number;
}
```

then it will

```typescript
                case 'model.dotcomrendering.pageElements.MediaAtomBlockElement':
                    return (
                        <VideoAtom
                            assets={element.assets}
                            poster={
                                element.posterImage &&
                                element.posterImage[0].url
                            }
                        />
                    );
```

See: https://github.com/guardian/atoms-rendering/pull/109

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,11 +33,11 @@
     },
     "dependencies": {
         "@braze/web-sdk-core": "^3.0.0",
-        "@emotion/core": "^10.0.35",
         "@cypress/skip-test": "^2.5.0",
+        "@emotion/core": "^10.0.35",
         "@guardian/ab-core": "^2.0.0",
         "@guardian/ab-react": "^2.0.1",
-        "@guardian/atoms-rendering": "^2.0.9",
+        "@guardian/atoms-rendering": "^2.4.1",
         "@guardian/automat-client": "^0.2.16",
         "@guardian/braze-components": "^0.0.14",
         "@guardian/consent-management-platform": "^6.6.0",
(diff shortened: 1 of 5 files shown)


oliverlloyd:
[Support non-Youtube (media atom) video](https://trello.com/c/tqlZ6kRw/2117-support-non-youtube-media-atom-video)

gtrufitt:
I have a slight concern with the use of `MediaAtomBlockElement` which is more of a tidy-up job outside of this PR but the internal terminology of `Media Atom` will suggest to most people what we've called `YoutubeBlockElement`. I think probably we need to call *both* Media atom - one MediaAtomNonYoutube and MediaAtomYoutube or something so that we're a bit clearer. cc/ @jfsoul @shtukas

liywjl (approved):

gtrufitt (approved):

jfsoul:
@gtrufitt yes I agree a review of the naming might be worthwhile, I still find it a bit confusing as someone who casually dips into the code without spending much time actually working on it.

oliverlloyd:
Re, naming. I completely agree. But it's not just here, all of the components around videos are confusingly named!

+1 for a comprehensive review

oliverlloyd:
I've added https://trello.com/c/OmM3PmDE/2158-review-video-naming-conventions to capture the naming review work

PR #2091, opened 2020-11-18
Put GA tracking behind consents

amyhughes:
## What does this change?

Here we use the [consent-management-platform](https://github.com/guardian/consent-management-platform) to determine whether to load tracking scripts for sending events to Google Analytics based on a reader's consent status.

### Before

Tracking events are sent to GA before readers have consented to tracking through the banner:
before_before_consenting

After consenting through the banner, more tracking events sent (timing events):
before_after_consenting

### After

No tracking events are sent to GA before readers have consented to tracking through the banner:
after_before_consents

After consenting through the banner, tracking events sent matching those in PROD:
after_after_consenting

After removing consents (in the footer, `Privacy settings` -> `Reject all`) and refreshing the page (mimicking a next pageview), no tracking events are sent to GA:
after_after_removing

NOTE: tracking scripts are not removed on consent change from `consent given` -> `consent not given`, rather we check the consent status on the next page load when deciding which scripts to load, and do not load the scripts accordingly.

## Why?

This ensures we respect the consent given by readers for GA tracking. The corresponding frontend PR can be found [here](https://github.com/guardian/frontend/pull/23203).


src/web/components/App.tsx
```diff
@@ -399,6 +399,38 @@ export const App = ({ CAPI, NAV }: Props) => {
...
+                gaCustomScript.src = window.guardian.gaPath;
+                document.head.appendChild(gaCustomScript);
+
+                console.log(`gaConsent state: ${consentGiven}`);
+            } else {
+                const gaScriptElement = document.getElementById('ga-script');
+                gaScriptElement?.parentNode?.removeChild(gaScriptElement);
+                const gaCustomScriptElement = document.getElementById('ga-custom-script');
+                gaCustomScriptElement?.parentNode?.removeChild(gaCustomScriptElement);
+
```
gtrufitt, line 428: This doesn't really do anything, because by the time you're running this script the contents of  is loaded into the JS engine and it is divorced from the script tag on the page. You probably want to no-op window.ga like in frontend?

src/web/server/htmlTemplate.ts
```diff
@@ -199,6 +201,14 @@ export const htmlTemplate = ({
                     })(window, document);
                 </script>
 
+                <script>
+                    if (window.guardian.mustardCut === false) {
+                        window.guardian.gaPath = "${gaPath.legacy}"
```
gtrufitt, line 206: Maybe you could just move these to within the two script tags that set the mustard cutting? https://github.com/guardian/dotcom-rendering/pull/2091/files#diff-2767f0a6851c0895c23f2fd1528f15726d84dbd13203a3d96247f5dd1e107564R146

src/web/components/App.tsx
```diff
@@ -399,6 +399,38 @@ export const App = ({ CAPI, NAV }: Props) => {
...
+                gaScript.id = 'ga-script';
+                gaScript.type = 'text/javascript';
+                gaScript.src = 'https://www.google-analytics.com/analytics.js';
+                document.head.appendChild(gaScript);
+
+                const gaCustomScript = document.createElement('script');
+                gaCustomScript.id = 'ga-custom-script';
+                gaCustomScript.type = 'text/javascript';
+                gaCustomScript.src = window.guardian.gaPath;
+                document.head.appendChild(gaCustomScript);
```
gtrufitt, line 420: I've just remembered that there is a `lib` function that does this for you! https://github.com/guardian/libs/blob/main/src/loadScript.ts

src/web/components/App.tsx
oliverlloyd: I think I had the same doubts as you but I seem to remember when we were testing that this code proved itslef out, that no further requests were sent. But maybe we didn't go back and test the negative...

src/web/server/htmlTemplate.ts
oliverlloyd: We thought about that but then we saw another use case [further up](https://github.com/guardian/dotcom-rendering/blob/main/src/web/server/htmlTemplate.ts#L158) where there was a conditional around `mustardCut` like this. That said, I see that this use case is inside a function so yeah, could be moved.

src/web/components/App.tsx
oliverlloyd: Nice!

I guess if we don't need to delete them then the id is redundant

oliverlloyd:
We've got some interesting [visual regression diffs here](https://www.chromatic.com/snapshot?appId=5dfcbf3012392c0020e7140b&id=5fb5142c5dd5630021df53ad). It looks like an embed is no longer loading.

src/web/components/App.tsx
gtrufitt: Here's an example of what I think you're doing - run some JS (i.e ping it on the window) and then remove the script - I don't think the script removal does anything once the JS is loaded  https://codesandbox.io/s/hungry-montalcini-z5ped?file=/src/index.js

amyhughes: Thanks both, we took another look at this and sure enough we weren't removing the scripts, but the script removal was redundant since on a page reload we check the consent status and either add the scripts or not depending on consent. We've removed the attempt to remove the scripts now.

src/web/components/App.tsx
amyhughes: Excellent, we were having type compatibility problems when pulling in @guardian/libs, so we've made our own copy of the `loadScript` function for now. Much nicer!

src/web/server/htmlTemplate.ts
amyhughes: Excellent, done

cypress/integration/e2e/consent.spec.js
```diff
@@ -0,0 +1,116 @@
...
+        // Accept tracking cookies
+        privacySettingsIframe().contains('Privacy settings');
+        privacySettingsIframe().find("[title='Accept all']").click();
+        // Make a second page load now that we have the CMP cookies set to accept tracking
+        cy.visit(`Article?url=${secondPage}`, fetchPolyfill);
+        // Wait for a call to Google Analytics to be made - we expect this to happen
+        cy.wait('@theCallToGoogle', { timeout: 3000 });
+    });
+
+    /**
```
amyhughes, line 55: @oliverlloyd went down a rabbit hole trying to write Cypress tests we were happy with for the case we care most about (not tracking users who have not yet given their consent, or have removed their consent). We'd like to include the tests as comments in case someone wants to build on these at a later date. They work, but as outlined here they are slow, and possibly fail to verify that events do not fire. Any objections?

src/web/components/App.tsx
```diff
@@ -399,6 +400,19 @@ export const App = ({ CAPI, NAV }: Props) => {
         CAPI.config.isSensitive,
     ]);
 
+    // ************************
```
amyhughes, line 403: This is the key change

src/web/server/document.tsx
```diff
@@ -104,8 +104,7 @@ export const document = ({ data }: Props) => {
      */
     const lowPriorityScriptTags = generateScriptTags(
         [
-            { src: 'https://www.google-analytics.com/analytics.js' },
```
amyhughes, line 107: ... along with this

src/web/server/document.tsx
```diff
@@ -145,6 +155,7 @@ export const document = ({ data }: Props) => {
         title,
         description: CAPI.trailText,
         windowGuardian,
+        gaPath,
```
amyhughes, line 158: we put this on windowGuardian so that we can construct the correct paths when we do want to load GA scripts

oliverlloyd (approved):
I'm happy to thumbs up here in terms of a sanity check but I paired on this with Amy so it'd be great to get some other thoughts on the approach

@gtrufitt @jfsoul ?

oliverlloyd:
Just had a random thought. Re this problem where we want to make sure that ga _isn't_ loaded, is there a way to manually try to call the lib? Such that if it fails we can pass the test?

src/web/components/App.tsx
```diff
@@ -399,6 +400,19 @@ export const App = ({ CAPI, NAV }: Props) => {
...
+    // *   Google Analytics   *
+    // ************************
+    useEffect(() => {
+        onConsentChange((state: any) => {
+            const consentGiven = getConsentFor('google-analytics', state);
+            if (consentGiven) {
+                loadScript('https://www.google-analytics.com/analytics.js');
+                loadScript(window.guardian.gaPath);
+            }
```
jfsoul, line 412: I think we ought to have an `else` condition here that sets `window.ga` to a no-op like we have done in the frontend PR. We should stop sending any data to GA as soon as someone removes their consent.

jfsoul:
> Just had a random thought. Re this problem where we want to make sure that ga isn't loaded, is there a way to manually try to call the lib? Such that if it fails we can pass the test?

Sounds about right. There's a function `window.ga` which you can try to call, but if we take my other suggestion I just made then it won't fail, it just won't do anything.

src/web/components/App.tsx
oliverlloyd: As we have it setup now, if the reader removes consent then any subsequent _page views_ will not send data to GA but if, as you suggest, we set `window.ga = () => {}` in the else then this should block _all_ future calls, which is much better.

Sematically, it might be nice to use `() => false`? This will also make testing for this use case clearer

oliverlloyd:
[Put GA tracking behind consents (DCR)](https://trello.com/c/RtcjeoIF/2138-put-ga-tracking-behind-consents-dcr)

cypress/integration/e2e/consent.spec.js
```diff
@@ -0,0 +1,93 @@
...
+        cmpIframe().find("[title='Manage my cookies']").click();
+        // Reject tracking cookies
+        privacySettingsIframe().contains('Privacy settings');
+        privacySettingsIframe().find("[title='Reject all']", {timeout: 30000}).click();
+        // Make a second page load now that we have the CMP cookies set to reject tracking and check
+        // to see if the ga property was set by Google on the window object
+        cy.visit(`Article?url=${secondPage}`);
+        waitForAnalyticsToInit();
+        // We force window.ga to be null on consent rejection to prevent subsequent requests
+        cy.window().its('ga').should('equal', null);
```
amyhughes, line 73: Looks good to me, great that we can test this now!

cypress/integration/e2e/consent.spec.js
```diff
@@ -0,0 +1,93 @@
...
+
+    const privacySettingsIframe = () => {
+        return cy
+            .get('[src*="https://cdn.privacy-mgmt.com/privacy-manager"]')
+            .its('0.contentDocument.body')
+            .should('not.be.empty')
+            .then(cy.wrap);
+    };
+
+    const waitForAnalyticsToInit = () => {
```
amyhughes, line 29: Okay, yes, this makes sense. At some point we just need to pick an amount of time to wait before declaring that something hasn't happened. This 300ms is much nicer than the seconds (odd) we had before

PR #2092, opened 2020-11-18
Use opinion pillar if designType is Comment

oliverlloyd:
We're using the wrong pillar for popular or frequently shared comment pieces, we should override the given pillar with `opinion` in these cases.

### Before
Screenshot 2020-11-18 at 21 02 27

### After
Screenshot 2020-11-18 at 21 11 47


oliverlloyd:
[MostPopular Card using wrong pillar](https://trello.com/c/phYEL5UN/2153-mostpopular-card-using-wrong-pillar)

shtukas (approved):

PR #2106, opened 2020-11-20
Move Ophan to priority scripts

nitro-marky:
This moves Ophan to priority scripts to hopefully stop `undefined` errors on the sign in gate.

### DIFF
--- a/src/web/server/document.test.tsx
+++ b/src/web/server/document.test.tsx
@@ -121,12 +121,12 @@ test('Sample of script tags have the correct attributes', () => {
     );
     expect(result).toEqual(
         expect.stringContaining(
-            `<script async type="module" src="/assets/ophan.js"></script>`,
+            `<script defer type="module" src="/assets/ophan.js"></script>`,
         ),
     );
     expect(result).toEqual(
         expect.stringContaining(
-            `<script async nomodule src=\"/assets/ophan.js\"></script>`,
+            `<script defer nomodule src=\"/assets/ophan.js\"></script>`,
         ),
     );
 });
--- a/src/web/server/document.tsx
+++ b/src/web/server/document.tsx
@@ -86,6 +86,7 @@ export const document = ({ data }: Props) => {
     const priorityScriptTags = generateScriptTags(
         [
             { src: polyfillIO },
+            { src: 'ophan.js', module: true },
             CAPI.config && { src: CAPI.config.commercialBundleUrl },
             { src: 'sentry.js', module: true },
             { src: 'dynamicImport.js', module: true },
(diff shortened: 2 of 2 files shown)


oliverlloyd:
I'm not sure the use case here is strong enough to justify adding a new high priority script tag?

I'm coming at this problem cold so I could be way off but is there benefit in looking at loading the script dynamically on the client? @amyhughes and I just did this for `ga.js` and the process to move `ophan.js` would be identical. Then, if we say, added a callback to the `ophan` code to shout when the lib isLoaded, then we can hook into that and trigger the next steps in a controlled fashion. Sorry, if I'm way off base with that random thought!

oliverlloyd:
> trigger the next steps in a controlled fashion

Eg. We could wrap the render for the `<SignInGate>` component in `App.tsx` with an `isOphanLoaded` state boolean so it would only appear once `ophan` is ready.

That said, I'm not sure the best way for Ophan to call back tbh. Ideally, we'd have a 'react' version of Ophan we could import, like with 'ab rendering'.

nitro-marky:
It's not ideal long term but it looks to be a necessary evil at the moment, there's some really tight coupling and assumptions that ophan will be on the page, so when we alias it in the commercial bundle those assumptions cause errors. Is there another way perhaps @jfsoul ?

oliverlloyd:
> It's not ideal long term but it looks to be a necessary evil at the moment, there's some really tight coupling and assumptions that ophan will be on the page, so when we alias it in the commercial bundle those assumptions cause errors. Is there another way perhaps @jfsoul ?

Yeah, I think the root evil here is loading code via script tags and then trying to manage the order of when those async calls complete - you can't. Even if you do put ophan higher up the page that's really no guarantee it will solve this problem

oliverlloyd:
In the short term, you could put ophan into an NPM package (just as it, no fanciness). And then we load it by importing it and we're then in a position to manage order and dependencies. Like, we could:

```typescript
import('@guardian/ophan').then(() => render <SignInGate />)
```

oliverlloyd:
Just putting this out there: https://www.npmjs.com/package/ophan-tracker-js

jfsoul:
My 2 cents...

We're predominantly removing ophan from the commercial bundle to ascertain whether that is affecting attention time tracking. The attention time tracking issue is concerning and it'd be great to either confirm or exclude ophan in the commercial bundle as a cause.

It's a fairly small package that we are currently inadvertently loading as priority via the commercial bundle anyway. My instinct would be to do the minimum now, get a few days of data with only a single version of ophan on the page, and then either revert or fix forward. However, I think the npm module is a possibility - as it's already provided that way!

gtrufitt (approved):
Happy to approve this considering there's no size difference in loaded scripts on the assumption that we'll review the other approaches after.

Defer ordering is guaranteed, so ophan.js will always execute before commerical bundle here (less guaranteed on the other scripts but they're loaded later and much larger so likelihood of loading before is miniscule).

gtrufitt:
> Just putting this out there: [npmjs.com/package/ophan-tracker-js](https://www.npmjs.com/package/ophan-tracker-js)

I think part of our problem is ophan-tracker-js has a sideffect that means it runs when loaded - so an import promise will be side effecty, everywhere we wanted to use ophan. I think ophan would need a bit of a re-jig

PR #2120, opened 2020-11-23
Better Blockquotes

oliverlloyd:
### Better Blockquotes
Turns out there are three possible states for a blockquote but DCR was only supporting two. Here we add support for the third case and refactor the code along the way to manage the additional complexity it adds.

A blockquote can be:

1) Quoted (has `quoted` class)
Screenshot 2020-11-23 at 09 36 03

2) Highlighted (Photo Essays only)
Screenshot 2020-11-23 at 09 35 31

3) Simple (neither of the above) - __**NEW*__
Screenshot 2020-11-23 at 10 00 05

### What was refactored
I've replaced the `add-highlights.ts` model enhancer with `enhance-blockquotes.ts`. With the addition to the logic of adding a `quoted` prop to the `BlockquoteBlockComponent` to support showing simple block quotes, it felt more natural to rename the file and restructure the logic

I have also added method chaining to `render.ts` to better represent executing the different steps of enhancement

I lifted the action to decide the `isDev` flag into the enhancers (`set-is-dev.ts`) because even though it's not an edit to the blocks it still felt like a model enhancement the same as the other steps.

### Quote icon alignment question
In `frontend` the bottom of the quote icon aligns with the bottom of the first line of text, with the top of the icon sitting proud. In DCR, we align the top of the icon with the top of the text.

@HarryFischer Making DCR support the align to top approach is a bit fiddly so before diving in I wanted to check if we maybe want to stick with bottom alignment?

#### Frontend
Screenshot 2020-11-23 at 10 04 55
Screenshot 2020-11-23 at 10 04 14

#### DCR
Screenshot 2020-11-23 at 10 06 25
Screenshot 2020-11-23 at 10 05 45


oliverlloyd:
[Blockquotes not displaying properly](https://trello.com/c/j5MXRu1U/2162-blockquotes-not-displaying-properly)

liywjl (approved):
great work!!

PR #2090, opened 2020-11-18
MJ - Add Goal Reached Behaviour to AMP Ticker

michaelbjacobson:
### What does this PR do?
As per the linked Trello card, this allows tickers in AMP epics to handle the exceeding of our goal, in the same way that our tickers do elsewhere. See screenshots below.

### Screenshots
_Goal exceeded_ 🎉 
Screenshot 2020-11-18 at 11 02 33

_Goal not **yet** exceeded_
Screenshot 2020-11-18 at 11 09 38

PR #2121, opened 2020-11-23
Mj fix amp ticker

michaelbjacobson:
### What does this PR do?
Fixes AMP problems caused by #2090, by adding a function to conditionally include an AMP `custom-component` depending on whether it's used in the page.


liywjl:
LGTM, but knowing we had an AMP issue not that log ago I am a little causious. @gtrufitt mentioned about having some AMP validator on CI, that was removed because it was acting flaky, but that might be worth revisiting to avoid future issues. This would give me more confidence before 👍  this PR

oliverlloyd:
I remember @gtrufitt mentioning that it wasn't a lot of work to update the tests to add more confidence for AMP. Maybe we can make these changes as part of this PR?

Obviously I don't have a view on priorities though and I'm blatantly speaking for other people here so I understand if this needs to go out sooner.

oliverlloyd:
Closing as stale

PR #2123, opened 2020-11-23
Index update with YoutubeBlockElement

shtukas:
Update the index with a new url for YoutubeBlockElement (the old one is dead) and also add a case for YoutubeBlockElement (in Main media position), which is a slightly different case.


liywjl (approved):
👍

PR #2124, opened 2020-11-23
Disable lazy loading images in Chromatic

oliverlloyd:
### What?
In the same way that we don't lazy render content at the bottom of the page for Chromatic snapshots, this PR prevents images being lazy loaded by the browser.

### Why?
Since making images lazy, Chromatic has been routinely throwing up false negatives where images are sometimes rendered at the bottom of page layouts and sometimes not.


liywjl (approved):
🙏 thank you for this

PR #2125, opened 2020-11-23
Support QuizAtomBlockElement

liywjl:
## What does this change?
Adds support for QuizAtom by creating QuizAtomBlockElement. Adds DCR support for personalised and knowledge quiz blocks

### Knowledge Quiz
Screenshot 2021-01-07 at 16 58 26

### Personality Quiz
Screenshot 2021-01-07 at 17 02 00


shtukas (approved):

src/web/lib/ArticleRenderer.tsx
```diff
@@ -136,6 +137,17 @@ export const ArticleRenderer: React.FC<{
                     );
                 case 'model.dotcomrendering.pageElements.ChartAtomBlockElement':
                     return <ChartAtom id={element.id} html={element.html} />;
+                case 'model.dotcomrendering.pageElements.QuizAtomBlockElement':
+                    return (
+                        <div id={`quiz-atom-${i}`}>
```
oliverlloyd, line 142: Can we change this to use Figure? That component will also set the id but in addition it manages role, and vertical spacing.

The quiz atom element will need `role` too. This should be available from CAPI

src/web/components/elements/QuizBlockComponent.tsx
```diff
@@ -0,0 +1,38 @@
...
+    id,
+    questions,
+    resultBuckets,
+    quizType,
+}: Props) => (
+    <div
+        className={css`
+            padding-bottom: ${space[2]}px;
+        `}
+    >
```
oliverlloyd, line 27: This div can be deleted in favour of Figure in `ArticleRenderer`

src/web/layouts/Standard.stories.tsx
```diff
@@ -9,7 +9,7 @@ import { Analysis } from '@root/fixtures/articles/Analysis';
 import { Feature } from '@root/fixtures/articles/Feature';
 import { GuardianView } from '@root/fixtures/articles/GuardianView';
 import { Interview } from '@root/fixtures/articles/Interview';
-import { Quiz } from '@root/fixtures/articles/Quiz';
+import { KnowledgeQuiz } from '@root/fixtures/articles/KnowledgeQuiz';
```
oliverlloyd, line 12: 👍  for the story but it might make more sense here to keep the name `Quiz` here as that is the `designType` value and the rest of these articles are named based on design type.

src/web/components/App.tsx
```diff
@@ -364,6 +365,20 @@ export const App = ({ CAPI, NAV }: Props) => {
...
+                    key={index}
+                    root="quiz-atom"
+                    index={quizAtoms.quizIndex}
+                >
+                    <QuizBlockComponent
+                        id={quizAtoms.id}
+                        questions={quizAtoms.questions}
+                        resultBuckets={quizAtoms.resultBuckets}
+                        quizType={quizAtoms.quizType}
+                    />
```
oliverlloyd, line 379: If we remove spacing from `QuizBlockComponent` then maybe we don't need it at all? Without the css it is just a thing that makes a choice about which atom to use. Maybe we can inline that here? This has the added bonus of maintaining a similar pattern to the other atoms and not having to use the *Component conovention

src/lib/content.d.ts
```diff
@@ -356,6 +364,7 @@ type CAPIElement =
     | CaptionBlockElement
     | CalloutBlockElement
     | ChartAtomBlockElement
+    | QuizAtomBlockElement
```
oliverlloyd, line 367: This is actually the only place where I find ordering things alphabetically useful (and I _know_ this will make @shtukas happy) so can we move this line and the declaration above to be better ordered. Hopefully the remaining code is also inline 🤞

index.d.ts
```diff
@@ -407,6 +408,35 @@ type HeadlineLink = {
...
+    imageUrl?: string;
+    resultBuckets?: QuizAtomResultBucketType[]
+}
+
+type ResultBucketsType = {
+    id: string;
+    title: string;
+    description: string;
+}
+
```
oliverlloyd, line 439: I think the definitions are a bit loose but maybe these would live better in `content.d.ts`? This might just be me but I've always thought of content as stuff around elements and `index` as more general stuff.

oliverlloyd (commented):
Looks really good, lots of work went into this!

Just a few comments - mostly about moving things around

src/web/components/App.tsx
liywjl: I originally did that, but I was unsure how to wrapp CSS to resolve padding bottom issue. Originally I thought of putting the CSS in `src/web/lib/ArticleRenderer.tsx` but felt this wasnt the right place for CSS. Adding CSS in `App.tsx` also doesnt feel right either and I suspected that we would of had some issues with hydration (if we wrap the styles outside hydration). I guess my question is where to put the CSS?

rcrphillips:
[Support other quiz types(personality types)](https://trello.com/c/aFOlvcf1/2084-support-other-quiz-typespersonality-types)

src/web/lib/ArticleRenderer.tsx
liywjl: Quiz pages doe not have a role, so are we ok to keep this?

src/web/components/elements/QuizBlockComponent.tsx
liywjl: removed in favor of `<Figure>`

src/web/lib/ArticleRenderer.tsx
oliverlloyd: Figure also does things like adding the `figure` tag for sematics as well as margins which might be useful. It's also just nice to be consistent, I think.

I suspect if you merge in `main` (where Figure is now widely used) this will make more sense

src/lib/content.d.ts
```diff
@@ -544,3 +553,32 @@ interface CampaignFieldSelect extends CampaignField {
 		value: string;
 	}[];
 }
+
```
oliverlloyd, line 556: Shall we add a label defining where these types are used? Right not they've been put under the 'Callout section' but I think we should create their own one.

src/web/components/App.tsx
```diff
@@ -386,6 +388,27 @@ export const App = ({ CAPI, NAV }: Props) => {
...
 				</HydrateOnce>
 			))}
+			{CAPI.quizAtoms.map((quizAtoms, index) => (
+				<HydrateOnce
+					key={index}
+					root="quiz-atom"
+					index={quizAtoms.quizIndex}
+				>
+					{quizAtoms.quizType === 'personality' &&
+					quizAtoms.resultBuckets ? (
```
oliverlloyd, line 398: This logic could result in a quiz with type 'personality' being rendered by the KnowledgeQuizAtom which we proabbly don't want. Maybe we can handle resultBuckets being undefined in the PersonalityQuizAtom itslef? Or, and this might be better, if it really is required we should make it so in the type system so we don't need to have this guard

src/web/components/elements/QuizBlockComponent.tsx
```diff
@@ -0,0 +1,31 @@
...
+	quizType: 'personality' | 'knowledge';
+};
+export const QuizBlockComponent = ({
+	id,
+	questions,
+	resultBuckets,
+	quizType,
+}: Props) => (
+	<>
+		{quizType === 'personality' && resultBuckets ? (
```
oliverlloyd, line 21: Ditto on 'personality' quizzes being sent to the KnowledgeQuizAtom

src/lib/content.d.ts
```diff
@@ -544,3 +553,32 @@ interface CampaignFieldSelect extends CampaignField {
...
+	revealText?: string;
+	isCorrect: boolean;
+	answerBuckets: string[];
+};
+
+type QuizAtomResultBucketType = {
+	id: string;
+	title: string;
+	description: string;
+	// assets TODO:
```
oliverlloyd, line 569: Should this remain?

oliverlloyd (commented):
One question, and this is a bit outside the scope of this DCR PR, but when I tested this locally with

http://localhost:3030/Article?url=https://www.theguardian.com/lifeandstyle/2020/dec/29/the-tv-quiz-of-the-year-from-a-schitts-creek-wedding-to-an-explosive-homeland-finale

I was surprised that the Submit button revealed the answer for all questions, even if no answer had been given.

I wonder if it would make more sense to have a reveal button against each question, more in line with how these quizzes work in Frontend (where each question is revealed separately, rather than a single 'Submit' action? Appreciate though that there could have already been discussion around this elsewhere!

liywjl:
> One question, and this is a bit outside the scope of this DCR PR, but when I tested this locally with
> 
> http://localhost:3030/Article?url=https://www.theguardian.com/lifeandstyle/2020/dec/29/the-tv-quiz-of-the-year-from-a-schitts-creek-wedding-to-an-explosive-homeland-finale
> 
> I was surprised that the Submit button revealed the answer for all questions, even if no answer had been given.
> 
> I wonder if it would make more sense to have a reveal button against each question, more in line with how these quizzes work in Frontend (where each question is revealed separately, rather than a single 'Submit' action? Appreciate though that there could have already been discussion around this elsewhere!

Yes this was something I did raise a while back with the team. I asked if a quiz submission should be prevented unless all answers were responded to, and I think it was informally agreeded to that no there isn't really a reason as to why we should prevent submission. But grated it might be worth raising this point again.

However, it might be better to implement `required` using Source . I raised the point that we cannot enforece `optional`/`required` in Source `RadioGroup` component https://github.com/guardian/source/issues/644 . I raised a PR to add this feature https://github.com/guardian/source/pull/678 but it is on hold until Source v3 is fully deployed.

**TLDR;** yes it is something we could add, and is probably worth raising with the team, but it might be best to enforce it through a Source update rather than through JS logic.

oliverlloyd:
> if a quiz submission should be prevented unless all answers were responded to

I think that's a subtly separate question. My point was should we reveal answers when the question hasn't been responded to? This is related to if we should force responses but not quite the same. So, we can decide to not force responses but then that can mean that we don't reveal the answers.

There's also the question of vocabulary here. If you say 'Reveal All' then it's natural to expect all answers to show. But 'Submit' suggests we only act on the inputs

src/web/components/App.tsx
liywjl: this is a good point, what do you think of removing the turnary and instead having something like this?
```
					{quizAtoms.quizType === 'personality' &&
						quizAtoms.resultBuckets && (
							<PersonalityQuizAtom
								id={quizAtoms.id}
								questions={quizAtoms.questions}
								resultBuckets={quizAtoms.resultBuckets}
							/>
						)}
					{quizAtoms.quizType === 'knowledge' && (
						<KnowledgeQuizAtom
							id={quizAtoms.id}
							questions={quizAtoms.questions}
						/>
					)}
```

I am reluctant to have to handled `quizAtoms.resultBuckets` within Quiz mainly because without it we do not have a way of displaying awnsers to users (thus rendering the quiz pointless).

liywjl:
> There's also the question of vocabulary here. If you say 'Reveal All' then it's natural to expect all answers to show. But 'Submit' suggests we only act on the inputs

You are right! I guess it's not something I thought about because of the issues we had with making the Quiz accessible. Remember before, when you would select an answer you would get a result back right away (referring to Knowledge quizzes). I am happy to add this extra button, what do you think of "Reveal All" and "Reveal Selected"?

src/web/components/App.tsx
oliverlloyd: So that's better in that it prevents 'personality' quizzes being rendered as Knowledge quizzes. But it still opens up the possibility of null results when really we want to prevent this with typing.

It feel like resultBuckets should be a required non nullable list and we should not accept from CAPI anything outside of that. Does that sound right?

oliverlloyd:
> > There's also the question of vocabulary here. If you say 'Reveal All' then it's natural to expect all answers to show. But 'Submit' suggests we only act on the inputs
> 
> You are right! I guess it's not something I thought about because of the issues we had with making the Quiz accessible. Remember before, when you would select an answer you would get a result back right away (referring to Knowledge quizzes). I am happy to add this extra button, what do you think of "Reveal All" and "Reveal Selected"?

I think Reveal Selected would be too confusing. My preference would be a separate Reveal for each question and then maybe a Reveal All at the end if that felt right but maybe we're straying into a place where more UX/Design input is needed?

liywjl:
> I think Reveal Selected would be too confusing. My preference would be a separate Reveal for each question and then maybe a Reveal All at the end if that felt right but maybe we're straying into a place where more UX/Design input is needed

Yeah that is more of a UX/Design issues in my view. Unless we just make every question required? that would remove the need for any of this. There was a thread about this issue that never was properlly concluded bar mentioning we should timebox. I think maybe it might be worth creating a new ticket to think how to design the quiz to be both accessible and good UX. I am not keen on the idea of a submit button for each question I think it would greatly penalise mouse users to have to click for every answer.

Are you ok to move this discussion out of the scope of the PR?

src/web/components/App.tsx
liywjl: > It feel like resultBuckets should be a required non nullable list and we should not accept from CAPI anything outside of that. Does that sound right?

the issue is that we have 1 quiz type `model.dotcomrendering.pageElements.QuizAtomBlockElement` that technically splits out into Personality Quiz types (that require `resultBuckets`) and Knowledge Quiz types that do not require that prop. Should this seperation be done in CAPI to avoid this possibility?

oliverlloyd:
> Are you ok to move this discussion out of the scope of the PR?

My concern is we need a place to have the discussion that blocks it being released. If we pass through this PR then we're essentially saying DCR supports quizzes as they stand, which I'm not sure we want to do without more work around the UX

oliverlloyd:
> we just make every question required

This feels like we'd be treating the quiz like a form that has to be submitted. But actually, at least for Knowledge quizzes, each question is it's own stand alone thing, with no relation to the others

src/web/components/App.tsx
liywjl: From internal discussions there seems little appetite to split the `QuizAtomBlockElement` into 2 blocks. I will close this thread as a result

liywjl:
After a lot of internal discussion we have decided to add a submit button for each Knowledge Quiz question to fix accessibility and maintain some UX on this PR https://github.com/guardian/atoms-rendering/pull/177 . We have also decided not to split out `QuizAtomBlockElement` into `PersonalityQuizBlockElement`  and `KnowledgeQuizBlockElement` to enforce `resultBuckets` is defined for all `PersonalityQuizBlockElement.

src/web/components/App.tsx
```diff
@@ -390,6 +392,30 @@ export const App = ({ CAPI, NAV }: Props) => {
...
 			))}
+			{CAPI.quizAtoms.map((quizAtoms, index) => (
+				<HydrateOnce
+					key={index}
+					root="quiz-atom"
+					index={quizAtoms.quizIndex}
+				>
+					<>
+						{quizAtoms.quizType === 'personality' &&
+							quizAtoms.resultBuckets && (
```
oliverlloyd, line 403: I think when we discussed this with Jon we spoke about how it's better to render the quiz even without results, rather than rendering nothing at all. In both cases the quiz is 'broken' because the authors have failed too provide the answers but still rendering the quiz feels like a more graceful error state?

It's also more semantic I think to handle the lack of results inside the component rather than here, in DCR

liywjl: The JSON will always return `resultBuckets` even if the quiz is not a personality quiz. I have changed for this prop to not be optional

PR #2132, opened 2020-11-24
Cypress v6.0.0

oliverlloyd:
Upgrade Cypress to [version 6.0.0](https://github.com/cypress-io/cypress/releases/tag/v6.0.0)

See: [Migration Guide](https://on.cypress.io/migration-guide)

- Replaced `is.not.visible` with `should.exist` in places
- Migrated to use `cy.intercept` instead of `cy.route`
- Added some longer timeouts in places to reduce issues on slow connections
- Added an `$el.is(':visible')` filter when checking element width
- deleted `fetchPolyfill` because this is no longer needed with v6
- Moved `intercept` logic inline because we only use it in one place


liywjl (approved):
Great work!

oliverlloyd:
[Cypress v6](https://trello.com/c/JpevclTA/2174-cypress-v6)

PR #2135, opened 2020-11-25
Footer Styling

oliverlloyd:
## What does this change?
Makes a few small styling changes to the page footer

1. Removes double borders.
The grey side borders were being set both by the `Section` and also by the code inside the section. We have a prop `showSideBorders` that we can use in this situation to prevent this.

2. Line Height
On Frontend the footer is not so deep, mainly because the text has a different `line-height`. Here I copy over the value used in Frontend for parity

3. Vertical alignment
I've increased the left margin on the copyright text to bring it into alignment with the rest of the page.

### Before
Screenshot 2020-11-25 at 13 36 08

### After
Screenshot 2020-11-25 at 13 35 08

### What does this not do?
There's also a footer styling issue around the padding for the Sign Up and another issue with how the offter displays between the `desktop` and `leftCol` breakpoints. These are not covered in this PR but there's [a ticket here](https://trello.com/c/QGf838H4/2175-footer-responsive-design)

### DIFF
--- a/src/web/components/Footer.tsx
+++ b/src/web/components/Footer.tsx
@@ -146,7 +146,10 @@ const readerRevenueLinks = css`
 
 const copyright = css`
     ${textSans.xsmall()};
-    padding: 12px;
+    padding-left: 20px;
+    padding-right: 12px;
+    padding-top: 12px;
+    padding-bottom: 12px;
 
     ${until.tablet} {
         margin-top: 11px;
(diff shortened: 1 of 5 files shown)


jfsoul (approved):
Change looks good to me if Harry is comfortable design wise.

HarryFischer:
As mentioned on chat - would be good to align newsletter sign up and Support Guardian messaging but other changes look good!

liywjl (approved):

PR #900, opened 2019-10-31
Remove Yarn Workspaces

oliverlloyd:
## What does this change?
Now that the work to extract the `guui` and `design` packages out, using `src-*` is complete, our `dotcom-rendering` repository is a monorepo with only 1 package. This doesn't make sense and because we have no intention to add more packages this PR removes the Yarn Workspaces configuration, returning us to a standard yarn repo.

## What work was done
- Removed `private` and `packages` properties from `/package.json`
- Merged `/frontend/package.json` > `/package.json`
- Moved contents of `/frontend` to `src`
- Updated all refs to `packages/frontend` > `src`
- Moved globals in `/index.d.ts` > `/window.guardian.d.ts`
- Moved `/frontend/index.d.ts` > `/index.d.ts`

## Steps to verify
`git checkout oliver/remove-monorepo && make dev`
- [x] `make dev` works
- [x] `make validate` works
- [x] `make cypress` works
- [x] `make percy` works
- [x] `make build` works
- [x] Deploys to code

## Why?
A simple yarn repo is easier for new members of the team to get started on and is generally simpler.

## Link to supporting Trello card
https://trello.com/c/jQeYasfH/827-remove-monorepo

PR #2136, opened 2020-11-25
Correct path of static assets

davidfurey:
I noticed that the last-modified time for the fonts in S3 is stuck on November 2019 (`PROD/frontend-static/static/frontend/fonts`).  
After a bit of investigation, I discovered that #900 changed the destination for static assets to PROD/frontend-static/**src**/static/frontend/fonts.  Since the [helpful directory layout diagram](https://github.com/guardian/dotcom-rendering/blob/4ccbfb15f86d1fcbe83a9d354f50faa36ba091aa/scripts/deploy/build-riffraff-bundle.js#L22) wasn't updated at the same time and DCR still references the old path,  I presume this was a mistake.

Fortunately these files are rarely actually changed so it hasn't caused a problem.  Hopefully we aren't reading from this incorrect path, and can delete these files.

### DIFF
--- a/scripts/deploy/build-riffraff-bundle.js
+++ b/scripts/deploy/build-riffraff-bundle.js
@@ -37,7 +37,7 @@ const copyStatic = () => {
     log(' - copying static');
     return cpy(
         ['**/*'],
-        path.resolve(target, `${siteName}-static`, 'src/static', siteName),
+        path.resolve(target, `${siteName}-static`, 'static', siteName),
         {
             cwd: path.resolve(root, 'src/static'),
             parents: true,


oliverlloyd:
@davidfurey I've added the PR Deployment label to this PR so we get a production build of the code that we can validate. That should pop up in the Checks section soon 👇

oliverlloyd:
PS. Amazing spot!

oliverlloyd:
I tested using the PR Deployment endpoint but I'm not 100% certain that ngrok server follows the exact same code path as production for builds so it would be nice to deploy and test this on CODE too, to be sure

davidfurey:
Deploying to CODE worked with the static files uploading to the correct place in S3, but I'm not really sure what I should be looking for to know if it has broken anything else.  Any suggestions?

oliverlloyd:
> Deploying to CODE worked with the static files uploading to the correct place in S3, but I'm not really sure what I should be looking for to know if it has broken anything else. Any suggestions?

One thing you could look at is the path used to get other static assets, like js. If you look at the bottom of the `head` tag in the page source you should see something similar to

Screenshot 2020-11-26 at 18 02 37

davidfurey:
Yes, these look fine, e.g. `//assets-code.guim.co.uk/assets/react.legacy.af771b1e90fc4946a421.js`.  And the page renders correctly.

shtukas (approved):

PR #2137, opened 2020-11-25
Hoist decidePosition

oliverlloyd:
## What does this change?
Moves `decidePosition` up out of `ImageComponent` into `/src/libs`

## Why?
This is the function we use to position elements based on their `role`. Up until now we've only used this for images but there is a requirement to also position other types of elements in the same way.


liywjl (approved):
👍

oliverlloyd:

PR #2138, opened 2020-11-25
Move fonts into a separate project

davidfurey:
# What does this change?

Removes fonts from this repo, and simplifies logic to load fonts from dev server, code or prod with a single location (prod)

## Why?

Fonts should be shared across all of our estate (e.g. dcr, interactives and reader revenue sites), so it does not make sense for them to live in the DCR repo.  They can now be found at https://github.com/guardian/fonts, which deploys to assets.guim.co.uk.


oliverlloyd:
Would this allow us to remove [the step in Storybook](https://github.com/guardian/dotcom-rendering/blob/main/.storybook/preview-head.html) where we manually load the fonts there?

davidfurey:
I'm not really clear what that step is doing, but I hope we can remove it since one of the goals is to stop fetching any fonts from pasteup.guim.co.uk and interactive.guim.co.uk

oliverlloyd:
> I'm not really clear what that step is doing, but I hope we can remove it since one of the goals is to stop fetching any fonts from pasteup.guim.co.uk and interactive.guim.co.uk

It was there because the getDist() funciton failed for Storybook (which runs it's own server to render pages for us to snapshot). I've made a change to remove it now because we can use `getFontsCss()` now 👍

oliverlloyd (approved):
This is great. It not only simplifies how we load fonts on the site but it also solves a long standing issue where we weren't using a complete font set for Storybook meaning some stories looked a bit off

👍

src/web/server/document.tsx
```diff
@@ -60,17 +60,17 @@ export const document = ({ data }: Props) => {
...
-        'fonts/guardian-textsans/noalts-not-hinted/GuardianTextSans-Bold.woff2',
+        // 'http://assets.guim.co.uk/static/frontend/fonts/guardian-headline/noalts-not-hinted/GHGuardianHeadline-Light.woff2',
+        // 'http://assets.guim.co.uk/static/frontend/fonts/guardian-headline/noalts-not-hinted/GHGuardianHeadline-LightItalic.woff2',
+        'http://assets.guim.co.uk/static/frontend/fonts/guardian-headline/noalts-not-hinted/GHGuardianHeadline-Medium.woff2',
+        'http://assets.guim.co.uk/static/frontend/fonts/guardian-headline/noalts-not-hinted/GHGuardianHeadline-MediumItalic.woff2',
+        'http://assets.guim.co.uk/static/frontend/fonts/guardian-headline/noalts-not-hinted/GHGuardianHeadline-Bold.woff2',
+        'http://assets.guim.co.uk/static/frontend/fonts/guardian-textegyptian/noalts-not-hinted/GuardianTextEgyptian-Regular.woff2',
+        // 'http://assets.guim.co.uk/static/frontend/fonts/guardian-textegyptian/noalts-not-hinted/GuardianTextEgyptian-RegularItalic.woff2',
+        'http://assets.guim.co.uk/static/frontend/fonts/guardian-textegyptian/noalts-not-hinted/GuardianTextEgyptian-Bold.woff2',
```
gtrufitt, line 70: Should these all be `https`?

src/lib/fonts-css.ts
```diff
@@ -268,6 +267,9 @@ const fontList: FontDisplay[] = [
     },
 ];
 
+const assetsUrl = (path: string): string =>
+    `//assets.guim.co.uk/static/frontend/${path}`;
```
gtrufitt, line 271: I think we should just use `https` rather than protocol-less

src/web/server/document.tsx
davidfurey: They should! Thank you

src/lib/fonts-css.ts
davidfurey: I copied this from `src/lib/assets.ts`, but I'm happy to change if we prefer specifying `https`

gtrufitt: I believe nowadays it's an anti-pattern and browsers are starting to upgrade all requests by default to `https` so makes sense to just hard code it here. 👍

gtrufitt (approved):
Great!

davidfurey:
@gtrufitt or @oliverlloyd - the UI test seems to show a slightly different text flow, which I think is unrelated to this PR.  Should I be concerned or just accept these as new baselines?

gtrufitt:
> @gtrufitt or @oliverlloyd - the UI test seems to show a slightly different text flow, which I think is unrelated to this PR. Should I be concerned or just accept these as new baselines?

No need for concern, like you say, looks unrelated to this PR - I have accepted them

davidfurey:
Tested on code

PR #2141, opened 2020-11-26
Figure

oliverlloyd:
### What does this change?
Adds a new component `Figure` to position elements within an article based on role

```typescript
type Props = {
    role?: RoleType; // Defaults to 'inline'
    children: JSXElements;
    id?: string;
};

type RoleType =
    | 'immersive'
    | 'supporting'
    | 'showcase'
    | 'inline'
    | 'thumbnail'
    | 'halfWidth';
```

It does three things:

1. Wraps its children in a `figure` tag. These tags are used to semantically identify elements in an article, esp. atoms and embeds but also images.

2. Set an `id` attribute. We often want to mark the root of a component with an id so that we can find it later when hydrating but we don't want to have to add another div layer to do this. Putting that logic here saves on the extra div and means that - hopefully - we only set these divs in one place.

3. Adds position styling based on the given `role`. By having a generic styling wrapper we can centralise how we position content based on the chosen role for that content.


liywjl (approved):
Great abstraction 👍  makes a lot more sense to see the differences visually

oliverlloyd:

src/web/components/Figure.tsx
```diff
@@ -1,22 +1,30 @@
...
+
 const roleCss = {
     inline: css`
         margin-top: 12px;
         margin-bottom: 12px;
     `,
 
     supporting: css`
+        margin-top: 12px;
+        margin-bottom: 12px;
```
gtrufitt, line 20: I know they weren't before, but should this margins be using the `space` token from the design system?

gtrufitt (approved):
Like!

src/web/components/Figure.tsx
oliverlloyd: Yes, they should. On it.

PR #2143, opened 2020-11-26
Use auto-foft to load fonts

sndrs:
## What does this change?

adds https://github.com/guardian/auto-foft to prioritise font loading and minimise reflows

## Why?

currently fonts are loaded on-demand by the browser and used as soon as they're ready. each new font triggers a reflow and requests are not prioritised

`auto-foft` allows you specify which of your fonts are critical, and which can wait. 

it renders immediately with fallbacks while it downloads the critical ones then applies them all in one go, falling back to faux styles for missing fonts. 

then it gets the rest and applies them all in one go too.

if the files are cached it happens as fast as the browser can manage (as with untouched CSS `@font-face`)

this means:

- initial bandwidth is used for the most important stuff first
- only 2 reflows on a cold cache

## frontend

if this gets a 👍 we can do the same for frontend and the coronavirus interactive widget, then we can start sharing font files better but still have control over how they load

attached to #2138

### DIFF
--- a/src/lib/fonts-css.ts
+++ b/src/lib/fonts-css.ts
@@ -1,4 +1,5 @@
 import minifyCssString from 'minify-css-string';
+import { Display } from '@root/src/lib/display';
 import { getStatic } from './assets';
 
 type FontFamily =
@@ -16,7 +17,7 @@ interface FontDisplay {
     woff2: string;
     woff: string;
     ttf: string;
-    weight: number;
+    weight: string | number;
     style: FontStyle;
 }
 
@@ -267,6 +268,174 @@ const fontList: FontDisplay[] = [
         style: 'normal',
     },
 ];
+
+export const isCritical = ({
+    font,
+    designType,
+    display,
+}: {
+    font: FontDisplay;
+    designType: DesignType;
+    display: Display;
+}): boolean => {
+    switch (display) {
+        case Display.Immersive:
+            switch (designType) {
+                case 'Comment':
+                case 'GuardianView':
+                    switch (font.family) {
+                        case 'GuardianTextEgyptian':
+                        case 'Guardian Text Egyptian Web':
+                        case 'GuardianTextSans':
+                        case 'Guardian Text Sans Web':
+                            return (
+                                font.weight === '400' && font.style === 'normal'
+                            );
+                        case 'GH Guardian Headline':
+                        case 'Guardian Egyptian Web':
+                            return (
+                                (font.weight === '200' ||
+                                    font.weight === '300' ||
+                                    font.weight === '700') &&
+                                font.style === 'normal'
+                            );
+                        default:
+                            return false;
+                    }
+                case 'Review':
+                case 'Recipe':
+                case 'Feature':
+                case 'Analysis':
+                case 'Interview':
+                case 'Live':
+                case 'Media':
+                case 'PhotoEssay':
+                case 'Article':
+                case 'SpecialReport':
+                case 'MatchReport':
+                case 'GuardianLabs':
+                case 'Quiz':
+                case 'AdvertisementFeature':
+                case 'Immersive':
+                default:
+                    switch (font.family) {
+                        case 'GuardianTextEgyptian':
+                        case 'Guardian Text Egyptian Web':
+                        case 'GuardianTextSans':
+                        case 'Guardian Text Sans Web':
+                            return (
+                                font.weight === '400' && font.style === 'normal'
+                            );
+                        case 'GH Guardian Headline':
+                        case 'Guardian Egyptian Web':
+                            return (
+                                (font.weight === '300' ||
+                                    font.weight === '700') &&
+                                font.style === 'normal'
+                            );
+                        default:
+                            return false;
+                    }
+            }
+        case Display.Showcase:
+        case Display.Standard:
+        default: {
+            switch (designType) {
+                case 'Comment':
+                case 'GuardianView':
+                    switch (font.family) {
+                        case 'GuardianTextEgyptian':
+                        case 'Guardian Text Egyptian Web':
+                        case 'GuardianTextSans':
+                        case 'Guardian Text Sans Web':
+                            return (
+                                font.weight === '400' && font.style === 'normal'
+                            );
+                        case 'GH Guardian Headline':
+                        case 'Guardian Egyptian Web':
+                            return (
+                                font.weight === '300' &&
+                                (font.style === 'normal' ||
+                                    font.style === 'italic')
+                            );
+                        default:
+                            return false;
+                    }
+                case 'Review':
+                case 'Recipe':
+                case 'Feature':
+                    switch (font.family) {
+                        case 'GuardianTextEgyptian':
+                        case 'Guardian Text Egyptian Web':
+                        case 'GuardianTextSans':
+                        case 'Guardian Text Sans Web':
+                            return (
+                                font.weight === '400' && font.style === 'normal'
+                            );
+                        case 'GH Guardian Headline':
+                        case 'Guardian Egyptian Web':
+                            return (
+                                (font.weight === '300' ||
+                                    font.weight === '700') &&
+                                font.style === 'normal'
+                            );
+                        default:
+                            return false;
+                    }
+                case 'Interview':
+                    switch (font.family) {
+                        case 'GuardianTextEgyptian':
+                        case 'Guardian Text Egyptian Web':
+                        case 'GuardianTextSans':
+                        case 'Guardian Text Sans Web':
+                            return (
+                                font.weight === '400' && font.style === 'normal'
+                            );
+                        case 'GH Guardian Headline':
+                        case 'Guardian Egyptian Web':
+                            return (
+                                (font.weight === '700' &&
+                                    font.style === 'normal') ||
+                                (font.weight === '400' &&
+                                    font.style === 'italic')
+                            );
+                        default:
+                            return false;
+                    }
+                case 'Analysis':
+                case 'Live':
+                case 'Media':
+                case 'PhotoEssay':
+                case 'Article':
+                case 'SpecialReport':
+                case 'MatchReport':
+                case 'GuardianLabs':
+                case 'Quiz':
+                case 'AdvertisementFeature':
+                case 'Immersive':
+                default:
+                    switch (font.family) {
+                        case 'GuardianTextEgyptian':
+                        case 'Guardian Text Egyptian Web':
+                        case 'GuardianTextSans':
+                        case 'Guardian Text Sans Web':
+                            return (
+                                font.weight === '400' && font.style === 'normal'
+                            );
+                        case 'GH Guardian Headline':
+                        case 'Guardian Egyptian Web':
+                            return (
+                                (font.weight === '500' ||
+                                    font.weight === '700') &&
+                                font.style === 'normal'
+                            );
+                        default:
+                            return false;
+                    }
+            }
+        }
+    }
+};
 
 const template: (_: FontDisplay) => string = ({
     family,
--- a/src/web/server/htmlTemplate.ts
+++ b/src/web/server/htmlTemplate.ts
@@ -204,8 +204,9 @@ export const htmlTemplate = ({
                     <img src="https://sb.scorecardresearch.com/p?c1=2&c2=6035250&cv=2.0&cj=1&cs_ucfr=0&comscorekw=${keywords}" />
                 </noscript>
                 ${[...priorityScriptTags].join('\n')}
-                <style class="webfont" data-auto-foft-fonts>${getFontsCss()}</style>
+                <style data-auto-foft-fonts>${getFontsCss()}</style>
                 <script>
+                // mark: PswXqO - keep these in sync with preloads
                 window.autoFoft = {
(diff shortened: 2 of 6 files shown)


SiAdcock (approved):
This is exciting, great work @sndrs! I guess to test, we should keep an eye on start render times and make sure that at least it's not negatively impacting the first contentful paint?

src/web/server/htmlTemplate.ts
```diff
@@ -203,8 +204,29 @@ export const htmlTemplate = ({
...
+                <script>
+                // mark: PswXqO - keep these in sync with preloads
+                window.autoFoft = {
+                    isCritical: function (font) {
+                        switch (font.family) {
+                            case 'GuardianTextEgyptian':
+                            case 'Guardian Text Egyptian Web':
+                            case 'GuardianTextSans':
+                            case 'Guardian Text Sans Web':
+                                return font.weight === 'normal' || font.weight === '400';
```
gtrufitt, line 217: should this be `font.style === 'normal'`?

src/web/server/htmlTemplate.ts
```diff
@@ -203,8 +204,29 @@ export const htmlTemplate = ({
                     <img src="https://sb.scorecardresearch.com/p?c1=2&c2=6035250&cv=2.0&cj=1&cs_ucfr=0&comscorekw=${keywords}" />
                 </noscript>
                 ${[...priorityScriptTags].join('\n')}
-                <style class="webfont">${getFontsCss()}${resetCSS}${css}</style>
-
+                <style data-auto-foft-fonts>${getFontsCss()}</style>
+                <script>
+                // mark: PswXqO - keep these in sync with preloads
```
gtrufitt, line 209:
```suggestion
                // mark: PswXqO - keep these in sync with preloads
                // Using https://github.com/guardian/auto-foft
                // and fonts from https://github.com/guardian/fonts
                // We load the regular weight, non-italic versions of body
                // fonts, and the bold and medium weights of headlines.
                // The weightings are described in 
 				// https://theguardian.design/2a1e5182b/p/930d69-typography
```

src/web/server/htmlTemplate.ts
```diff
@@ -203,8 +204,29 @@ export const htmlTemplate = ({
...
+                    isCritical: function (font) {
+                        switch (font.family) {
+                            case 'GuardianTextEgyptian':
+                            case 'Guardian Text Egyptian Web':
+                            case 'GuardianTextSans':
+                            case 'Guardian Text Sans Web':
+                                return font.weight === 'normal' || font.weight === '400';
+                            case 'GH Guardian Headline':
+                            case 'Guardian Egyptian Web':
+                                return (font.weight === '500' || font.weight === '700') &&
```
gtrufitt, line 220: I think this will miss opinion articles for headlines, maybe that's ok? https://www.theguardian.com/commentisfree/2020/nov/27/brexit-festival-eu-national

gtrufitt (commented):
Very cool idea! I dug into the approach and it seems great, I think I'd like to see comparisons of prev vs after across a number of article types (i.e I don't think opinion right now will have critical loaded for headline), but this seems like an excellent addition to the perf box of tricks! ⭐ 🚀

src/web/server/htmlTemplate.ts
sndrs: ooh yeah good spot

src/web/server/htmlTemplate.ts
paperboyo: Yep. But here we enter the territory of having different defaults depending on where one lands (in their very first visit). If that could be automated…

sndrs: yeah it does – do we know about content type in htmlTemplate?

sndrs:
@SiAdcock I did a [webpage test comparison of code and prod](https://www.webpagetest.org/video/compare.php?tests=201127_DiWW_1ebf8d1160bc51a71343160d0967d166%2C201127_DiEC_a8a7decf0eed4caab67dab0576b9d2c3&thumbSize=600&ival=500&end=visual)

the video comparison is really good https://www.webpagetest.org/video/view.php?id=201127_Di17ca1a8b98412e1aeda7947d1367e9ec0ae5dc05

src/web/server/htmlTemplate.ts
gtrufitt: We do, yeah, we can generate it in document.tsx from - I think we want to do it based on `designType` and `Display`, mirroring, at least for headline the logic here: https://github.com/guardian/dotcom-rendering/blob/main/src/web/components/ArticleHeadline.tsx#L176

What do you reckon @oliverlloyd ? Do we need a `decideFontStyle` mechanism?

sndrs: I wonder if that is a bit overkill currently? will there be more than the comment difference?

gtrufitt:
> @SiAdcock I did a [webpage test comparison of code and prod](https://www.webpagetest.org/video/compare.php?tests=201127_DiWW_1ebf8d1160bc51a71343160d0967d166%2C201127_DiEC_a8a7decf0eed4caab67dab0576b9d2c3&thumbSize=600&ival=500&end=visual)
> 
> the video comparison is really good [webpagetest.org/video/view.php?id=201127_Di17ca1a8b98412e1aeda7947d1367e9ec0ae5dc05](https://www.webpagetest.org/video/view.php?id=201127_Di17ca1a8b98412e1aeda7947d1367e9ec0ae5dc05)

Oh yeah, that's brilliant, really great example! Love the little `music` popping in bold, that's a great improvement.

src/web/server/htmlTemplate.ts
sndrs: although if that could pick critical fonts per content type it could also be used for preloads too, one place to define them

sndrs:
it's doing something v weird on CODE in safari (mobile and desktop) which it doesn't in dev, will hold on for #2138 so the env are more similar

src/web/server/htmlTemplate.ts
oliverlloyd: This PR is a really clever idea! (Sorry, missed it earlier)

So I've added a function `isCritical` which has a big, wandering decision tree to decide if a particular font should be deemed critical or not depending on `designType` and `display`. It's just stand alone, I've not wired it in. I took the logic from browsing our [layout stories](https://www.chromatic.com/pullrequest?appId=5dfcbf3012392c0020e7140b&number=2138&view=components&groupPrefix=Layouts) (I used [David's branch](https://github.com/guardian/dotcom-rendering/pull/2138) to do this as that PR fixes a bug we had before where we weren't loading fonts in Storybook properly)

#### Types need reviewing
I noticed we have a bit of a typing clash where `font.weight` can be either a string or a number. Looks like our existing types expected a number but here we're sometimes checking for `font.weight === '400' || font.weight === 'normal'`. I've left of this check for now but this will need reviewing.

#### Verbose vs. Abstraction
I initially started to abstract the repeated code here but I pulled back because I've noticed where we've used this pattern elsewhere that keeping it verbose like this makes edits later on a lot easier and I can see this being an area where edits are likely to happen as we start to change and add more page layouts (we already have open tickets for this work).

paperboyo: > critical or not depending on `designType` and `display`

Superb! That’s articles-only, right? I wonder if fronts differ enough to warrant having their default(s) too?

sndrs: I'm actually wondering about changing the API, so:

```html
<style data-auto-foft-fonts="0">
@font-face etc
</style>
<style data-auto-foft-fonts="100">
@font-face etc
</style>
```

then load each font found in a `data-auto-foft-fonts` stylesheet in order of the index (attribute value). allows you to customise declaratively rather than in script, and to create multiple batches if your page could benefit from it

sndrs: (and then prob not call it `auto-foft` any more ¯\\_(ツ)_/¯ )

src/lib/fonts-css.ts
```diff
@@ -16,7 +17,7 @@ interface FontDisplay {
     woff2: string;
     woff: string;
     ttf: string;
-    weight: number;
+    weight: string | number;
```
liywjl, line 20: if `weight` can be `string` or `number` would you not need to cast `weight` to a `string` to get conditions like `font.weight === '400'` to work?

src/web/server/htmlTemplate.ts
oliverlloyd: I wanted to return to this as I can see this being a useful feature to have when we adopt support for Labs articles, where an entirely different font is used. There would need to be a bit of work to bring this code up to date with the new format for `design` but that should be straight forward, happy to help with that.

@sndrs Apologies for the huge delay in responding. I didn't follow your proposed changes before and meant to ask some questions. Regarding the different api you proposed, how would that work with focussing on different fonts based on design type?

oliverlloyd:
[Help release Alex S's change to remove some fonts from initial load](https://trello.com/c/kWGzDfXW/2338-help-release-alex-ss-change-to-remove-some-fonts-from-initial-load)

src/web/server/htmlTemplate.ts
sndrs: I plan to revisit the API after this week, once I'm on devex, but I'm pretty certain it will very html-based, rather than react/dcr etc oriented

so I guess you'd have to move some decision making to the templating? applications would still need to work out which fonts they need and in what order, and provide them as `style` elements defining `@font-face`s for them

oliverlloyd: > it will very html-based

Is this because we don't want to commit too much to React and keep a door open for a non react world in the future? If so, I can appreciate the benefit to this but I worry that having similar decision logic in two different places could be a pain.

We're doing a lot of work to [centralise these types](https://github.com/guardian/types/blob/main/src/format.ts) of decisions, [for colours](https://docs.google.com/document/d/1nkntFJUunqrwNYszTrc2RBfyYUd1bEAeQh76yVFOLjo/edit#), for layouts, [for design](https://github.com/guardian/dotcom-rendering/blob/fdac1eb8d5a31b237c90a1d334620cf5a0a36681/src/web/lib/decideDesignType.ts) and I'd ❤️ to add fonts to that process!

Of course, if it's for other reasons then 👍

sndrs: > Is this because we don't want to commit too much to React and keep a door open for a non react world in the future?

100% this. 

this is a mechanism for improving font loading anywhere, not for deciding which fonts a platform needs

the more those decisions are tied up in the implementation details of the platform, the more likely it is the platform will stagnate in the future, as it becomes harder and harder to port common behaviours to new implementations 

I appreciate it's easier to say that than do it, but to me DCR itself is proof of the value of rearchitecting parts of a platform in situe, and frontend font loading is an example of how limiting it becomes to bind performance enhancements to a single platform.

whatever happens with dotcom design and information architecture in the future, it's very probable we'll want to prioritise font resource loading, and that's all this is intended to do

sndrs: I'm picturing something like this in the document template

```js
// document.tsx

export default () =>
    getFontsForThisPageSomehow.map(
        (fontGroup, i) => `
            <style data-auto-foft-fonts=${i}>
            ${fontGroup.map(
                (font) => `
                @font-face {
                    font-family: ${font.family};
                    font-style: ${font.style};
                    font-weight: ${font.weight};
                    src: url(${font.files.woff2}) format('woff2');
                }
            `,
            )}
            </style>
            <script>
            ${fontLoadingJS}
            </script>`,
    )
```

how `getFontsForThisPageSomehow` would be totally DCR-specific – does that make sense?

oliverlloyd: > this is a mechanism for improving font loading anywhere, not for deciding which fonts a platform needs

Gotcha. And thanks loads for the example, that really helps!

I think you're basically right, we should be thinking about where we want to be in 5 years time, not 6 months. Longer delays to rebuilds and making them smoother represent huge cost savings.

Doing this though is hard, but definately worth the effort, esp. in these higher level areas, like fonts.

sndrs:
closing in favour of a `font-display` based approach – to come...

oliverlloyd:
Now that we're starting in on Fronts is now a good time to revisit this area?

I had a chat with @HarryFischer about this subject earlier and he pointed me to [this document](https://docs.google.com/document/d/1AiHKjq5LUqGkELeEXLLUXsySdN-jhOChVh4BHyO5RcU/edit#heading=h.fceh6iz3gm9z) which reminded me of this PR. There's some great stuff here.

sndrs:
yeah we never finished our fonts work in CSTI, which would have picked this up, really only due to prioritising. i would love to see this through though – do you think it's possible dotcom would look at this now?

paperboyo:
Harry and Bryony also wanted to be able to use OpenType features available only in Full subsets. And we have these reduced-size Latin ones, too. Maybe declaring **all** would mean more performant site and wider subsets would get loaded **only** when necessary (and hopefully, only those would be cached)?

sndrs:
we would need to get some numbers i think on the effect of those wider subsets. i fully understand the desire to use them, but they don't come for free (as i know you know well @paperboyo hence all that subsetting work). we can try and understand better exactly what those costs are though

PR #2147, opened 2020-11-27
Position Embeds based on their role

oliverlloyd:
## What does this change?
Positions the `EmbedBlockComponent` and `UnsafeBlockComponent` based on the elements `role`. We use `Figure` to do this.

### A note on vertical margins
I've removed the `margin-bottom` css from the embed component. I'm planing on doing this each time we wrap a component in `Figure`. I think we should start to centralise to one location the place where we set the vertical spacing between elements on the page.

### Next Steps
We won't see non default (`role` = `inline`) behaviour until the `role` property is passed down from CAPI into the DCR model @shtukas ? Once we have the `role` property available on the model, assuming it is a mandatory field, then we should update our types to show `role` as required.

## Why?
Because embeds can have a role and we should respect this when deciding where to place them in an article.

### DIFF
--- a/fixtures/articles/AdvertisementFeature.ts
+++ b/fixtures/articles/AdvertisementFeature.ts
@@ -2326,6 +2326,7 @@ export const AdvertisementFeature: CAPIType = {
                     safe: false,
                     alt:
                         'Do you have a pressing question about electric cars? Submit it below for your chance to have your question answered by an EV expert...',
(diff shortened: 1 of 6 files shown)


oliverlloyd:

liywjl (approved):
I think this reshuffling is a lot better than before 👍

PR #2148, opened 2020-11-27
Restrict where we support force-braze-message

tjmw:
Restrict the domains where we support the `force-braze-message` query string arg. Marketing can use this to force the Braze banner to appear. As we make the banner configuration more flexible it doesn't make sense to support this on the main website, so restrict to preview (and dev environments).

### Before

The `force-braze-message` query string arg can be used in all environments.

### After

The `force-braze-message` can only be used in preview and dev.

### DIFF
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -150,10 +150,24 @@ const getMessageFromBraze = async (
     return canShowPromise;
 };
 
+const FORCE_BRAZE_ALLOWLIST = [
+    'preview.gutools.co.uk',
+    'preview.code.dev-gutools.co.uk',
+    'localhost',
+    'm.thegulocal.com',
+];
+
 const getBrazeMetaFromQueryString = (): Meta | null => {
     if (URLSearchParams) {
-        const params = new URLSearchParams(window.location.search);
         const qsArg = 'force-braze-message';
+
+        if (!FORCE_BRAZE_ALLOWLIST.includes(window.location.hostname)) {
+            // eslint-disable-next-line no-console
+            console.log(`${qsArg} is not supported on this domain`);
+            return null;
+        }
+
+        const params = new URLSearchParams(window.location.search);
         const value = params.get(qsArg);
         if (value) {
             try {


rhystmills (approved):
Looks good to me.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -150,10 +150,24 @@ const getMessageFromBraze = async (
...
+    'localhost',
+    'm.thegulocal.com',
+];
+
 const getBrazeMetaFromQueryString = (): Meta | null => {
     if (URLSearchParams) {
-        const params = new URLSearchParams(window.location.search);
         const qsArg = 'force-braze-message';
+
+        if (!FORCE_BRAZE_ALLOWLIST.includes(window.location.hostname)) {
```
oliverlloyd, line 164: Maybe use `FORCE_BRAZE_DENYLIST` to remove the negation?

oliverlloyd (approved):
I can't speak to the overall logic but this js looks sound 👍

src/web/components/StickyBottomBanner/BrazeBanner.tsx
gtrufitt: Probably safer to allow specifics, than deny specifics?

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -150,10 +150,24 @@ const getMessageFromBraze = async (
     return canShowPromise;
 };
 
+const FORCE_BRAZE_ALLOWLIST = [
+    'preview.gutools.co.uk',
+    'preview.code.dev-gutools.co.uk',
```
gtrufitt, line 155: Do you want m.code too?

gtrufitt (approved):
Nice, it's great to see stuff like this!

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: Good question! The goal was to not support anywhere public facing which I guess technically m.code is. I think I'll leave it for now, but if it's annoying to do without we can always add down the line.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: Yeah that was my thinking, that allowing specifics makes more sense.

tjmw:
Thanks folks!

PR #2149, opened 2020-11-27
Ipsos MORI / Dotmetrics tag technical review

jfsoul:
## What does this change?
Adds a technical review for a new audience measurement tag for Ipsos MORI / Dotmetrics. The context is best read from the review document.

## Why?
Best practice to do a thorough review before implementation.


jfsoul:
Merging as this was reviewed on #1961

PR #2150, opened 2020-11-27
Cypress Intercept Improvements

oliverlloyd:
### What does this change?
Replaces the use of `req.response.body` with `req.body` inside a `req.reply` callback. Looks like the upgrade to v6 the other day caused these checks to become a bit flakey and I noticed that the examples in the docs are using a newer structure to access the body so I've implemented that instead.

See: https://docs.cypress.io/api/commands/intercept.html#Intercepting-a-response


oliverlloyd:
[Cypress Improvements](https://trello.com/c/xbJYvqtI/2181-cypress-improvements)

liywjl (approved):
👍

PR #2152, opened 2020-11-29
Weird Dropcaps

oliverlloyd:
## What does this change?
One of the rules for deciding when a dropcap appears is that the length of the text it proceeds should be greater than 200 characters. We had a bug where when we tried to decide the length of this text we ended up also counting any html markup characters as well, incorrectly returning a longer value than what the reader actually sees.

In [this article](https://www.theguardian.com/sport/2020/nov/29/chess-world-hails-queens-gambit-fuelled-boom) there is a link in the first paragraph which leads to two possible outcomes when counting length

### Before (with markup)
```
Since its arrival on Netflix last month, The Queen’s Gambit has attracted a staggering <a href="https://www.theguardian.com/tv-and-radio/2020/nov/26/the-queens-gambit-netflix-most-watched-series-hit-chess">62 million</a> viewers – making it the streaming service’s most-watched scripted limited series.
```

Length: `302` - Weird DropCap

### After
```
Since its arrival on Netflix last month, The Queen’s Gambit has attracted a staggering 62 million viewers – making it the streaming service’s most-watched scripted limited series.
```

Length: `179` - No DropCap

## Why?
image


shtukas (approved):

oliverlloyd:
[Weird DropCaps](https://trello.com/c/8qU0I8r3/2186-weird-dropcaps)

PR #2158, opened 2020-11-30
ContainerLayout

oliverlloyd:
##  What does this change?
Adds the new `ContainerLayout` component. This is currently a story-only component and is not yet imported into any live code.

```typescript
type Props = {
    title?: string; // Section title
    fontColour?: string; // Title colour
    description?: string; // Title description
    url?: string; // Title link
    sectionId?: string; // Root `id`
    sideBorders?: boolean; // If we should show the side borders or not
    centralBorder?: 'partial' | 'full'; // If the left column border should be full, partial or not present
    showTopBorder?: boolean; // Helps to prevent duplicate borders
    padSides?: boolean; // Should pad section sides or not
    padContent?: boolean; // If we should pad the container where content (children) is placed
    backgroundColour?: string;
    borderColour?: string;
    children?: React.ReactNode; // What lands in the container
};
```

This component is an abstraction of an existing pattern used for onwards content where page sections were built using a combination of other components such as `Section`, `LeftColumn`, `OnwardsTitle`, `OnwardsContainer`, etc. By abstracting this logic we make the process of building more onwards content simpler and make page layouts more readable.

To reach this abstraction I've inlined the `Container` logic and hoisted `OnwardsTitle`, renaming it `ContainerTitle`.

## Future Steps
1. Implement the `ContainerLayout` component in out onwards code
2. Refactor out the use of `OnwardsTitle`
3. Investigate if there's benefit in merging `ArticleTitle` and `ContainerTitle`. Consider all possible use cases: Article titles, onwards titles but also front titles

## Why?
Because we want to make it easier for us and other developers to use DCR layout components to construct pages


shtukas (approved):

liywjl (approved):
👍  cool stuff

oliverlloyd:
[PageSection](https://trello.com/c/iqGqDPnC/2184-pagesection)

src/web/components/SectionTitle.tsx
```diff
@@ -31,9 +31,9 @@ const headerStyles = css`
     }
 `;
 
-const descriptionStyles = css`
+const descriptionStyles = (fontColour?: string) => css`
```
gtrufitt, line 34: Is there a type for palette text ?

oliverlloyd: You mean a type from Source? That would be cool. @SiAdcock ?

src/web/components/PageSection.stories.tsx
```diff
@@ -0,0 +1,199 @@
...
+            sideBorders={true}
+            centralBorder="full"
+            padSides={false}
+            padContent={false}
+        >
+            <Grey />
+        </PageSection>
+    );
+};
+SidesStory.story = { name: 'with a partial border divider' };
```
gtrufitt, line 124: Name is same as above

gtrufitt:
This is aces, interesting to see the different variations we have. I think the nomenclature should probably remain `container layout` or similar - I had no idea what 'page section' was going to be, and while it makes sense in isolation, we should try to work within the current namings.

src/web/components/PageSection.stories.tsx
oliverlloyd: Great spot! Fixed now

src/web/components/SectionTitle.tsx
oliverlloyd: I had a chat with Simon about this and he's going to create a ticket for this

PR #2160, opened 2020-11-30
Identity | Sign in gate | Stop scroll on dismiss

coldlink:
## What does this change?
- Remove scrolling to the top of the article body on sign in gate dismiss
- Extend sign in gate test expiry

## Why?
 Cumulative Layout Shift Score of 0.43 (https://web.dev/cls/) - 'Good' total CLS is below 0.1, and because CrUX reports are used for Google search rankings now, this will have a big impact on our rankings.

We should allow the sign in gate just to expand from its current position, rather than shifting back to the top of the page.

### DIFF
--- a/src/web/components/SignInGate/SignInGateSelector.tsx
+++ b/src/web/components/SignInGate/SignInGateSelector.tsx
@@ -55,13 +55,6 @@ const dismissGate = (
         currentAbTestValue.variant,
         currentAbTestValue.name,
     );
-
-    // When the user closes the sign in gate, we scroll them back to the main content
-    const articleBody = document.querySelector(
-        '.article-body-commercial-selector',
-    );
-
-    articleBody?.scrollIntoView(true);
 };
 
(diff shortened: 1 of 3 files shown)


oliverlloyd (approved):
Removing this code seems 👍

oliverlloyd:
Side question: Would we get a perf win if we move this show hide logic into React? As in, use props and state to decide when to show paragraphs?

gtrufitt (approved):
Amazing, thanks for the quick fix!

coldlink:
> Side question: Would we get a perf win if we move this show hide logic into React? As in, use props and state to decide when to show paragraphs?

theres some longer term work to rethink the way we insert the sign in gate into articles, but we haven't made any progress on that yet

oliverlloyd:
> Side question: Would we get a perf win if we move this show hide logic into React? As in, use props and state to decide when to show paragraphs?

Had chat offline. Result: Ignore me.

PR #2169, opened 2020-12-01
Add missing google-analytics.com to prefetch list

shtukas:
## What does this change?

Add missing `google-analytics.com` to prefetch list. This brings DCR on par with the list of preconnect and prefetch URLs on frontend.


gtrufitt (changes requested):
We don't want to do this because of consent

shtukas:
Turns out we can preconnect / prefetch to GA without user consent.

PR #2170, opened 2020-12-01
Add larger image preview for Google Discover

nitro-marky:
## What does this change?

Adding this meta tag allows Google Discover to include larger article previews where the main image is >=1200px. 

[https://developers.google.com/search/docs/advanced/mobile/google-discover?hl=en&visit_id=637424198370039526-3805703503&rd=1](url)

This is already implemented on frontend so this essentially brings DCR to parity.

### DIFF
--- a/src/web/server/htmlTemplate.ts
+++ b/src/web/server/htmlTemplate.ts
@@ -68,10 +68,6 @@ export const htmlTemplate = ({
 
     const twitterMetaTags = generateMetaTags(twitterData, 'name');
 
-    // This tag enables pages to be featured in Google Discover as large previews
-    // See: https://developers.google.com/search/docs/advanced/mobile/google-discover?hl=en&visit_id=637424198370039526-3805703503&rd=1
-    const googleDiscoverMetaTag = '<meta name="robots" content="max-image-preview:large">';
-
     // Duplicated prefetch and preconnect tags from DCP:
     // Documented here: https://github.com/guardian/frontend/pull/12935
     // Preconnect should be used for the most crucial third party domains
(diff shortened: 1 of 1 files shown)


shtukas (approved):

liywjl (approved):
👍

src/web/server/htmlTemplate.ts
```diff
@@ -68,6 +68,10 @@ export const htmlTemplate = ({
 
     const twitterMetaTags = generateMetaTags(twitterData, 'name');
 
+    // This tag enables pages to be featured in Google Discover as large previews
+    // See: https://developers.google.com/search/docs/advanced/mobile/google-discover?hl=en&visit_id=637424198370039526-3805703503&rd=1
```
oliverlloyd, line 72: Useful description!

src/web/server/htmlTemplate.ts
```diff
@@ -68,6 +68,10 @@ export const htmlTemplate = ({
...
+    const googleDiscoverMetaTag = '<meta name="robots" content="max-image-preview:large">';
```
oliverlloyd, line 73: We could just inline the string rather than assigning it to a variable?

oliverlloyd (approved):
Just a comment on maybe not using a var?

I had a quick chat with @gtrufitt about this. Seems 👍

src/web/server/htmlTemplate.ts
nitro-marky: No problem, I was just following the style there, but yeah it'll be fine as a string.

PR #2180, opened 2020-12-02
Extend Test Test

gtrufitt:
## What does this change?

Extends the ab-test used in Cypress to far-future.

## Why?
Cypress is failing


oliverlloyd (approved):
Setting a project end date there I see

PR #2183, opened 2020-12-02
Preview external links should open in a new tab

jfsoul:

### DIFF
--- a/src/web/components/ArticleBody.tsx
+++ b/src/web/components/ArticleBody.tsx
@@ -15,6 +15,7 @@ type Props = {
     designType: DesignType;
     adTargeting: AdTargeting;
     host?: string;
+    isPreview: boolean;
 };
 
 const pillarColours = pillarMap(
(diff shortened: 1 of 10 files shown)


oliverlloyd:
Not saying this _isn't_ the right way to go but this feels like a lot of complexity to support people being able to open links a certain way in preview.

Is there a way to solve the problem closer to where it's happening? Like a config setting people can enable in their browsers?

oliverlloyd:
Thinking about this around our earlier discussion about sand... it'd be great to be able to assign a cognitive cost to adding a prop to a component. I think it's actually quite high. There's a limit somewhere, say 20 props, where after that the component starts to become hard to reason about and maintenance slows down and the chances of bugs go up. Sure, it depends on a lot and is also quite subjective but I think it's true to say Every Prop is Precious. This is now my new motto.

jfsoul:
I tend to agree, it did end up being a bit more complex than I'd like and really this is only needed due to requirements of `viewer` which is probably better handled closer to the problem. V2 coming... closing this one.

PR #2191, opened 2020-12-03
Disabling CMP in Cypress

oliverlloyd:
This PR disables execution of the CMP tests on Team City and Github.

The CMP sourcepoint module acts differently depending on the location that the code is executed in. If the context location is the US - as can happen when using servers provided by TC and GH - then the expected CMP iframe will not appear and this breaks these tests.


mxdvl:
It’s a shame about 81f35109aba82e0da44efa409bb767ef1e81f29e, but I don’t think we can reliably run these on CI yet…

shtukas (approved):

PR #2192, opened 2020-12-03
Add documentation of contracts and a new Viewer article body selector contract

jfsoul:
## What does this change?
Adds an additional selector to the article body so that the viewer in composer can select this element.  I've also added some documentation for contracts because I feel these are very important to document for loose contracts such as this.

## Why?
Allows viewer to continue treating external links in preview articles specially. We could add this functionality in DCR directly, but it turned out to be quite complicated for a requirement of another system.

### DIFF
--- a/docs/contracts/000-contracts.md
+++ b/docs/contracts/000-contracts.md
@@ -1,0 +1,15 @@
+# Contracts of Dotcom Rendering
+
+## Context
+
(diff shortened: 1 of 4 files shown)


shtukas (approved):

oliverlloyd (approved):
👏

PR #2197, opened 2020-12-04
Use correct RR links in mobile header

tomrf1:
To be consistent with the header ctas in frontend here:
https://github.com/guardian/frontend/blob/main/common/app/views/fragments/header.scala.html#L92

For non-UK, cta is 'Contribute' and links to /contribute
Screen Shot 2020-12-04 at 11 01 16

For UK, cta is 'Subscribe and links to /subscribe
Screen Shot 2020-12-04 at 10 52 10


TomPretty (approved):
I'm not really sure why we do this, but the code looks good! 👍

PR #2198, opened 2020-12-04
SubNav is 14px when under tablet and 16px when over

oliverlloyd:
## What does this change?
Adjusts the font sizes used in the SubNav above and below `tablet`.

### Why aren't we using Source?
Because Source does not provide a 14px or a 16px `textSans` font.

## Why?
For design parity

### DIFF
--- a/src/web/components/SubNav/SubNav.tsx
+++ b/src/web/components/SubNav/SubNav.tsx
@@ -49,7 +49,12 @@ const collapsedStyles = css`
 `;
 
 const fontStyle = css`
-    ${textSans.medium()};
+    ${textSans.small()};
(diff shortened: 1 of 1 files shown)


oliverlloyd:
@SiAdcock @akemitakagi @HarryFischer we did already discuss this offline which this PR is based on but just for the record (and for others needing to make decisions similar to this) are we

1. happy to override Source like this or do we want to
2. adapt Source to fit the SubNav design or
3. adapt the SubNav design to fit Source?

HarryFischer:
I think 14/16px looks best to me - not sure about the Source integration though

liywjl (approved):
👍

PR #2200, opened 2020-12-04
EOY header test

tomrf1:
A/B test different copy in the header cta bar.
frontend PR: https://github.com/guardian/frontend/pull/23340

Also removed an old AU-only feature in the header cta bar.

### Control
Screen Shot 2020-12-04 at 15 27 07

### Variant
Screen Shot 2020-12-04 at 15 27 26

### Variant in January
Screen Shot 2020-12-04 at 15 29 17

### DIFF
--- a/src/web/components/ReaderRevenueLinks.tsx
+++ b/src/web/components/ReaderRevenueLinks.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { css, cx } from 'emotion';
 
 import ArrowRightIcon from '@frontend/static/icons/arrow-right.svg';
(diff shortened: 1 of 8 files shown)


src/web/lib/acquisitions.ts
```diff
@@ -0,0 +1,49 @@
+import {OphanComponentType} from "@root/src/web/browser/ophan/ophan";
+import {constructQuery} from "@root/src/lib/querystring";
```
liywjl, line 2: noticed you have some styling issues that are not handled by prettier. I had this issue as well, all i needed was to update VSC and just enable prettier on save

src/web/components/ReaderRevenueLinks.tsx
```diff
@@ -114,41 +120,77 @@ const subMessageStyles = css`
...
+                campaignCode: 'header_support',
+                componentType: 'ACQUISITIONS_HEADER',
+            });
         }
+    }, [variantName]);
+
+    const getHeading = (): string | JSX.Element => {
+        if (variantName === 'variant') return month === 12 ? `Support us this December` : 'Support us for 2021';
+        return <span>Support The&nbsp;Guardian</span>;
+    };
```
liywjl, line 168: is there a reason why we have a `span` for some text, and not for others?

src/web/lib/acquisitions.ts
```diff
@@ -0,0 +1,49 @@
...
+    componentType: OphanComponentType,
+    componentId: string,
+    campaignCode?: string,
+    abTest?: { name: string, variant: string },
+    pageViewId: string,
+    referrerUrl: string,
+};
+
+// Adds acquisition tracking codes if it is a support url
+export const addTrackingCodesToUrl = ({
```
tomrf1, line 15: adapted from frontend version https://github.com/guardian/frontend/blob/126dfcbc1aa961650b7f7ff41ee50a12782bb62e/static/src/javascripts/projects/common/modules/commercial/acquisitions-ophan.js#L105

liywjl (approved):
Just some minor things but overall looks good 👍

src/web/components/ReaderRevenueLinks.tsx
tomrf1: this line needs to keep the `&nbsp;` from the original, whereas just a string is fine for the variant copy

src/web/lib/acquisitions.ts
tomrf1: I ran `make fix`, should this do all the formatting? I don't use VSC

PR #2204, opened 2020-12-07
Immersive Headline Rebuild

oliverlloyd:
## What does this change?
This PR rebuilds how we layout immersive headlines.

## Why?
We have had a couple of issues with these inverted headlines, in particular with the pseudo black box we use to stretch the blackness to the right not being the right height or overalying the headline text.

## Principles
Our issues were caused by how we structured the headline html. This PR restructures things to respect the principles we need to make things work. Namely:

> 1. That the pseudo black box css needs to wrap the headline so that it will inherit the correct height

> 2. That it also needs to be high enough up the dom so that it isn't contained by anything with a margin that will prevent it stretching all the way to the right of the viewport.

In addition, while working in this area, I had to adjust and move some margins so I took the opportunity to set another principle:

> 3. That text should align vertically

### Other changes
I introduced the `ContainerLayout` component to manage the left colum logic. I also fixed some spacing around the `ArticleMeta` component when rendered inline. See below

Before                                |  After
:-------------------------:|:-------------------------:
  |


liywjl (approved):
👍

nitro-marky:
Banging! 🎉

oliverlloyd:
[Article Headline Simplification](https://trello.com/c/flf4E0Z4/2097-article-headline-simplification)

PR #2067, opened 2020-11-16
Use img tag instead of picture

oliverlloyd:
## What?
This PR proposes replacing the [picture](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture) tag with an [img](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img) tag as a stopgap ahead of using [image-rendering](https://github.com/guardian/image-rendering)

## TLDR
Surprisingly, tests for this change showed a lot of deviation between Frontend vs. DCR, both DCR with a picture tag and DCR with an img tab. More surprisingly, DCR with an img tag often outperformed Frontend. These variations could be for a lot of reasons though, not purely because an `img` tag was used.

## How?
There are two main code changes here:

1. Moving all image manipulation related logic into the `Picture` (now called `Img`) component
2. Removing the `hdipi` related logic and adding the ability to create sizes.

## Why?
DCR has known issues where it can sometimes render substantially lower quality images than appropriate in the context of position, screen width and resolution. In the short term, this change brings us into line with Frontend in most cases, makes us actually do better than frontend in some cases, and leaves us at the same or an acceptable level below frontend in the rest. 

## Why is this a stopgap?
See this [ongoing discussion](https://docs.google.com/document/d/1LfSCS0e9KJFxQuehguEdYdbns_afDSrtxx8MI9zgiMg/edit#heading=h.w4spq6y4l65b) for more detail on our medium to longer term goals. But to summarise we want to:

1. Solve the [Wealthy Guy problem](https://github.com/guardian/frontend/pull/11400). Mobile devices with small screens but high dpr values can end up being served excessively large files, wasting bandwidth.

2. [Finesse the quality](https://blog.imgix.com/2016/03/30/dpr-quality). We want to be able to specify a different set of images to use above a certain resolution. This is because we've found that for high resolution screens, by using `dpr=2&quality=45` together we can reduce the file size without noticeably impacting quality.

Both of these need a `picture` tag based solution.

## Test methodology
I rendered different images using a variety of simulated devices (via Firefox and Chrome dev tools) and then checked the resulting src chosen by the browser. Test were done using code from Frontend, new DCR and current DCR.

### Example urls:
https://www.theguardian.com/artanddesign/2020/may/11/cross-border-love-a-photo-essay?dcr=false
http://localhost:3030/Article?url=https://www.theguardian.com/artanddesign/2020/may/11/cross-border-love-a-photo-essay
[...]

PR #2205, opened 2020-12-07
The Portrait Immersive Main Media Width Problem

oliverlloyd:
## What does this change?
This is a proposal for discussion around a problem where immersive main media images are being chosen by the browser poorly on portrait devices. This happens because we size the main media content based on 100vh but for portrait aspect ratios this results in us loading much smaller images than we need.

Before                                |  After
:-------------------------:|:-------------------------:
  |  

## What was done in Frontend?
This problem has already been raised, discussed and solved 5 years ago, see:

https://github.com/guardian/frontend/pull/12811
and then
https://github.com/guardian/frontend/pull/12847

That work resulted in this excellent code:

```scala
@immersivePortraitSource(largestImage: model.ImageAsset, hidpi: Boolean = false) = {
    <source media="@if(hidpi) {
        (orientation: portrait) and (-webkit-min-device-pixel-ratio: 1.25), (orientation: portrait) and (min-resolution: 120dpi)
    } else {
        (orientation: portrait)
    }"

    sizes="@{100 * largestImage.ratioDouble}vh"
    @* This adds 15 src options to the srcset from 500 to the largest image width incrementing by 250px *@
    srcset="@{(500 to largestImage.width by 250)
        .toList
        .map(width => ImgSrc.srcsetForProfile(views.support.ImageProfile(width = Some(width)), picture, hidpi).asSrcSetString)
        .mkString(", ")}" />
}
```

where `ratioDouble` is

```scala
lazy val ratioDouble: Double = width.toDouble / height
```

## Why are we not doing the same thing?
We must ensure that this work is not lost when transitioning to `image-rendering` but before we adopt the IR picture tag approach, this PR  applies a simpler method via the `img` tag.

So instead of using an `imageRatio` based on an images `height` and `width` to calculate the correct width to set in `sizes` this PR is simply using the viewport height and an assumed image ration of 5:3 to decide what width to ask for.

## Shouldn't we be more explicit?
There's [a case](https://twitter.com/tabatkins/status/1265642071920017418) for being [more chill](https://twitter.com/tabatkins/status/1265642071920017418). The argument goes along the lines of: it's [hard to get this right](https://css-tricks.com/a-guide-to-the-responsive-images-syntax-in-html/) and when you aim for absolute perfection it's very easy to make mistakes because the logic is complex. Keeping it chill makes it easy to see what's going on.

Of course, we can do better and we should. Even small perf gains at scale are worth the extra complexity. But a staged approach to getting there, starting off simple, might be nice?

### DIFF
--- a/src/web/components/Img.tsx
+++ b/src/web/components/Img.tsx
@@ -48,6 +48,17 @@ const buildSourcesString = (srcSets: SrcSetItem[]): string => {
     return srcSets.map((srcSet) => `${srcSet.src} ${srcSet.width}w`).join(',');
 };
 
+/**
+ *       mobile: 320
+ *       mobileMedium: 375
+ *       mobileLandscape: 480
+ *       phablet: 660
+ *       tablet: 740
+ *       desktop: 980
+ *       leftCol: 1140
+ *       wide: 1300
+ */
+
 const buildSizesString = (role: RoleType, isMainMedia: boolean): string => {
     switch (role) {
         case 'inline':
@@ -57,8 +68,25 @@ const buildSizesString = (role: RoleType, isMainMedia: boolean): string => {
         case 'thumbnail':
             return '140px';
         case 'immersive':
+            // Immersive MainMedia elements fill the height of the viewport, meaning
+            // on mobile devices even though the viewport width is small, we'll need
+            // a larger image to maintain quality. Deciding which width image to ask
+            // for is difficult because aspect ratios and orientations between
+            // different devices varies so the values given below are best estimates
+            // based on the most popular configurations
+
+            // Immersive body images stretch the full viewport width below wide,
+            // but do not stretch beyond 1300px after that.
             return isMainMedia
-                ? '100vw'
+                ? `
+                    (min-width: ${breakpoints.leftCol}px) 1600px,
+                    (min-width: ${breakpoints.desktop}px) 1900px,
+                    (min-width: ${breakpoints.tablet}px) 1900px,
+                    (min-width: ${breakpoints.phablet}px) 1600px,
+                    (min-width: ${breakpoints.mobileLandscape}px) 1300px,
+                    (min-width: ${breakpoints.mobileMedium}px) 900px,
+                    1300px
+                `
                 : `(min-width: ${breakpoints.wide}px) 1300px, 100vw`;
         case 'supporting':
             return `(min-width: ${breakpoints.wide}px) 380px, 300px`;
--- a/src/web/components/Img.tsx
+++ b/src/web/components/Img.tsx
@@ -78,15 +78,7 @@ const buildSizesString = (role: RoleType, isMainMedia: boolean): string => {
             // Immersive body images stretch the full viewport width below wide,
             // but do not stretch beyond 1300px after that.
             return isMainMedia
-                ? `
-                    (min-width: ${breakpoints.leftCol}px) 1600px,
-                    (min-width: ${breakpoints.desktop}px) 1900px,
-                    (min-width: ${breakpoints.tablet}px) 1900px,
-                    (min-width: ${breakpoints.phablet}px) 1600px,
-                    (min-width: ${breakpoints.mobileLandscape}px) 1300px,
-                    (min-width: ${breakpoints.mobileMedium}px) 900px,
-                    1300px
-                `
+                ? `(orientation: portrait) 167vh, 100vw`
                 : `(min-width: ${breakpoints.wide}px) 1300px, 100vw`;
         case 'supporting':
             return `(min-width: ${breakpoints.wide}px) 380px, 300px`;
--- a/src/web/components/Img.tsx
+++ b/src/web/components/Img.tsx
@@ -74,6 +74,9 @@ const buildSizesString = (role: RoleType, isMainMedia: boolean): string => {
             // for is difficult because aspect ratios and orientations between
             // different devices varies so the values given below are best estimates
             // based on the most popular configurations
+            //
+            // The value of 167vh relates to a 5:3 image which is equal to 167 (viewport height)  : 100 (viewport width).
+            //
 
             // Immersive body images stretch the full viewport width below wide,
             // but do not stretch beyond 1300px after that.
--- a/src/web/components/Img.tsx
+++ b/src/web/components/Img.tsx
@@ -70,13 +70,10 @@ const buildSizesString = (role: RoleType, isMainMedia: boolean): string => {
         case 'immersive':
             // Immersive MainMedia elements fill the height of the viewport, meaning
             // on mobile devices even though the viewport width is small, we'll need
-            // a larger image to maintain quality. Deciding which width image to ask
-            // for is difficult because aspect ratios and orientations between
-            // different devices varies so the values given below are best estimates
-            // based on the most popular configurations
-            //
-            // The value of 167vh relates to a 5:3 image which is equal to 167 (viewport height)  : 100 (viewport width).
-            //
+            // a larger image to maintain quality. To solve this problem we're using
+            // the viewport height (vh) to calculate width. The value of 167vh
+            // relates to an assumed image ratio of 5:3 which is equal to
+            // 167 (viewport height)  : 100 (viewport width).
 
             // Immersive body images stretch the full viewport width below wide,
             // but do not stretch beyond 1300px after that.
--- a/src/web/components/elements/ImageBlockComponent.stories.tsx
+++ b/src/web/components/elements/ImageBlockComponent.stories.tsx
@@ -15,7 +15,7 @@ export default {
     component: ImageBlockComponent,
     title: 'Components/ImageBlockComponent',
     parameters: {
-        chromatic: { diffThreshold: 0.2 },
+        chromatic: { diffThreshold: 0.4 },
     },
 };
 
--- a/src/web/components/elements/ImageBlockComponent.stories.tsx
+++ b/src/web/components/elements/ImageBlockComponent.stories.tsx
@@ -14,6 +14,9 @@ import { image } from './ImageBlockComponent.mocks';
 export default {
     component: ImageBlockComponent,
     title: 'Components/ImageBlockComponent',
+    parameters: {
+        chromatic: { diffThreshold: 0.2 },
+    },
 };
 
 const Container = ({ children }: { children: JSX.Element | JSX.Element[] }) => (


oliverlloyd:
@paperboyo @sndrs @JamieB-gu @mxdvl thoughts?

JamieB-gu:
> @paperboyo @sndrs @JamieB-gu @mxdvl thoughts?

Interestingly on AR it looks like we've come up with a [similar solution](https://github.com/guardian/apps-rendering/blob/986336e5970154754f4f9a294961a4e17569667b/src/components/headerImage.tsx#L105) to the frontend one you described above (i.e. using `(imageRatio * 100)vh`). I've also noticed that our `HeaderImage` component in question is still using the pre-IR implementation, we'll need to migrate that.

> but before we adopt the IR picture tag approach, this PR applies a simpler, more arbitrary method

Are you saying this solution is temporary until you introduce IR into DCR?

liywjl (approved):
Really comprehensive! great work!

paperboyo:
Related to (as Lightbox would be another portrait situation): https://github.com/guardian/frontend/issues/12216#issuecomment-196869603

> but before we adopt the IR picture tag approach, this PR applies a simpler, more arbitrary method via the `img` tag

👏 

> Of course, we can do better and we should

👏 👏

Apart from using Picture (being able to prevent phones from downloading higher res for desktops is most important in immersive, as we currently satisfy quite high quality desktop displays), I wonder what more we could do. If we would to forego being able to resize the browser (and a picture following fluidly) only on phones and be OK with another request when readers change orientation (how often?), maybe we could use the resizer to [actually cut off](https://docs.fastly.com/en/image-optimization-api/crop) sides of these (with some leeway for differently ratioed phone displays)? That should shave a lot for phone-only downloads. Unless I haven’t thought of something?

Whenever we gave any thought to Grid supporting “mobile” crops explicitly, we always ended up thinking that setting point(s) of interest would be preferable to actually providing (and requiring users to set) two different physical crops (art direction-like). But whatever we will choose one day (if we ever do it), currently all these are just always cropped off equally relative to the image centre.

oliverlloyd:
> Are you saying this solution is temporary until you introduce IR into DCR?

Yes, we've taken the approach in DCR of seeting some short term goals of improving image quality to give readers a better experience - which we're doing by using an `img` tag and things like this PR. But we maintain our medium to long term goal of migrating to `image-rendering`

oliverlloyd:
> use the resizer to actually cut off sides of these

This sounds like a good trade off. Small images giving us a performance win vs. unexpected behaviour when resizing the screen. +1 for the perf. gain here, screen resizing must be very rare (outside the second Floor).

An automated solution like this would obviously be easier that asking people to take additional steps in Grid but there's possibly a lot of benefit to having a human choose where that portrait crop happens? I guess though that Viewer is the sanity check here so any strange mobile cropping will be seen at this stage.

paperboyo:
> screen resizing must be very rare

I think so. Especially on mobile!

But this is really unrelated directly to the second point (sorry for introducing confusion). Currently all images coming from Grid are the same, just of different size. DCR/AR/IR could unilaterally decide to cut off invisible sides (if only for mobile). And nobody would loose anything and only gain some pocket change from smaller phone bill (even orientation change probably fires off another request?).

Regarding us supporting aided multiple-ratio cropping for fluid scenarios **one day**: yes, users now sometimes crop source images in Grid several times, before they arrive at the satisfactory result on different ratios (doesn’t help that they all seem to think immersive must be 5:3 for some [unknown reason](https://www.theguardian.com/music/2020/nov/07/kylie-minogue-its-time-to-dress-in-sequins-and-glitter-through-the-darkness)!). I only mentioned it, because **then** we would need stop this arbitrary cropping off sides relative to the centre, described in the previous paragraph, and start doing it relative to these point(s) of interest that arrived through CAPI. I give us 5 years, though, so a non-issue.

oliverlloyd:
> I give us 5 years, though

👋 to anyone reading this in 5 years

oliverlloyd:
[Low quality images on DCR](https://trello.com/c/zUPi89R9/1817-low-quality-images-on-dcr)

src/web/components/Img.tsx
```diff
@@ -48,6 +48,17 @@ const buildSourcesString = (srcSets: SrcSetItem[]): string => {
     return srcSets.map((srcSet) => `${srcSet.src} ${srcSet.width}w`).join(',');
 };
 
+/**
+ *       mobile: 320
+ *       mobileMedium: 375
+ *       mobileLandscape: 480
```
mxdvl, line 54: how accurate is that, nowadays?

src/web/components/Img.tsx
```diff
@@ -57,8 +68,25 @@ const buildSizesString = (role: RoleType, isMainMedia: boolean): string => {
...
-                ? '100vw'
+                ? `
+                    (min-width: ${breakpoints.leftCol}px) 1600px,
+                    (min-width: ${breakpoints.desktop}px) 1900px,
+                    (min-width: ${breakpoints.tablet}px) 1900px,
+                    (min-width: ${breakpoints.phablet}px) 1600px,
+                    (min-width: ${breakpoints.mobileLandscape}px) 1300px,
+                    (min-width: ${breakpoints.mobileMedium}px) 900px,
+                    1300px
+                `
```
mxdvl, line 89:
```suggestion
                ? `
                    (orientation: portrait) 167vw,
                    100vw
                `
```

mxdvl (commented):
> @paperboyo @sndrs @JamieB-gu @mxdvl thoughts?

Could we use the same `(orientation: portrait)` and use that set the size to `167vh`, assuming 5:3 is the most common ratio? It seems that this change would limit the resolution of images to 1300px even though they stretch to the whole viewport on large devices.

```
(orientation: portrait) 167vw,
100vw
```

src/web/components/Img.tsx
oliverlloyd: Probably not very, the name is certainly confusing for me. We design to it though so it would be hard to remove.

oliverlloyd:
> It seems that this change would limit the resolution of images to 1300px even though they stretch to the whole viewport on large devices

I think that the default to `1300px` is an oversight, that should indeed be `100vw`. Great spot!

oliverlloyd:
> Could we use the same `(orientation: portrait)` and use that set the size to `167vh`

This is an inspired idea. It's so simple but does exactly what we want! 👏 

The reason that I used varying sizes at different breakpoints is because device screen ratios actually vary quite a bit but by utilizing the viewport height to set the desired width you nicely sidestep that problem. I also ❤️ how the code is so much simpler using this approach.

Hoowever, I tried it out locally and it seems to be returning the expected images in Chrome but Firefox is stubonly set on always returning the largest image so I want to investigate this some more.

paperboyo:
> Firefox is stubonly set

Firefox again? Maybe it’s somehow related to [this](https://github.com/guardian/dotcom-rendering/pull/2067#issuecomment-739246214)?

oliverlloyd:
> Firefox again? Maybe it’s somehow related to [this](https://github.com/guardian/dotcom-rendering/pull/2067#issuecomment-739246214)?

I think you're right. I think it was using an image from it's cache rather than fetching a new one. Seems to have fixed itself now though after I disabled the cache 🎉 . I've read about these kind of problems when testing images like this. Each browser has implemented the spec in their own way and there are some factors which can make things unpredicable (for instance, I _think_ Chrome will choose a different image based on network conditions).

src/web/components/Img.tsx
oliverlloyd: Just to be clear, I think `(orientation: portrait) 167vw` above should read as `(orientation: portrait) 167vh` (`vh`) so I'm leaving this commit suggestion alone and I'll push this change up manually

paperboyo:
Oh, boo. I thought there is something in common to both behaviours of Firefox. The other one still consistently chooses 445px in place of 620px cache or not. Nevermind, here was hoping, this is unrelated.

mxdvl (approved):
I like the clarity around portrait mode triggering a large image 😉

src/web/components/Img.tsx
```diff
@@ -57,8 +68,17 @@ const buildSizesString = (role: RoleType, isMainMedia: boolean): string => {
         case 'thumbnail':
             return '140px';
         case 'immersive':
+            // Immersive MainMedia elements fill the height of the viewport, meaning
+            // on mobile devices even though the viewport width is small, we'll need
+            // a larger image to maintain quality. Deciding which width image to ask
+            // for is difficult because aspect ratios and orientations between
+            // different devices varies so the values given below are best estimates
+            // based on the most popular configurations
```
gtrufitt, line 76: Something for clarity on the 167vh value.

```suggestion
            // based on the most popular configurations
            //
            // The value of 167vh relates to a 5:3 image which is equal to 167 (viewport height)  : 100 (viewport width).
            //
```

gtrufitt (approved):

src/web/components/Img.tsx
paperboyo: Just to note that while it’s definitely true that most main medias are 5:3, there is **nothing** stopping them being any ratio. As I understand, this (`167vh`) will control the number of steps, so if it’s gonna be off, nothing terrible will happen. Still, curious what [Kylie’s](https://www.theguardian.com/music/2020/nov/07/kylie-minogue-its-time-to-dress-in-sequins-and-glitter-through-the-darkness) gonna look like.

src/web/components/Img.tsx
JamieB-gu: The names are somewhat arbitrary. We need to specify breakpoints _somewhere_ - we could probably name them 1-7 if we wanted to 🤷‍♂️.

JamieB-gu:
> maybe we could use the resizer to [actually cut off](https://docs.fastly.com/en/image-optimization-api/crop) sides of these

Is this the ["Art Direction" problem](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Art_direction)?

> Could we use the same `(orientation: portrait)`

How is `orientation: portrait` defined? What happens if the user is on a desktop browser using a narrow window? Or a landscape iPad in split screen?

paperboyo:
> Is this the ["Art Direction" problem](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Art_direction)?

Not AFAIU. Kinda opposite of it. Art Direction means you are explicitly providing a different image for mobile. Here, knowing upfront that you are displaying this image on mobile, you just blindly cutting off the parts you know aren’t going to show up. And you keep quite about it in front of your Art Director, coz you know he won’t even notice.

PR #2212, opened 2020-12-09
Thank you header for supporters

tomrf1:
Adds a message for supporters to the header (excluding US edition).

Screen Shot 2020-12-09 at 07 28 55
Screen Shot 2020-12-09 at 07 29 09
Screen Shot 2020-12-09 at 07 29 23

### DIFF
--- a/src/web/components/ReaderRevenueLinks.test.tsx
+++ b/src/web/components/ReaderRevenueLinks.test.tsx
@@ -32,14 +32,14 @@ describe('ReaderRevenueLinks', () => {
     };
     const edition: Edition = 'UK';
 
-    it('should not render if shouldHideSupportMessaging() returns true', async () => {
+    it('should not render if shouldHideSupportMessaging() returns true and edition is US', async () => {
         shouldHideSupportMessaging.mockReturnValue(true);
 
         const { container } = render(
             <AbProvider>
                 <ReaderRevenueLinks
                     urls={urls}
-                    edition={edition}
+                    edition="US"
                     dataLinkNamePrefix="nav2 : "
                     inHeader={true}
                     pageViewId="1234"
--- a/src/web/components/ReaderRevenueLinks.tsx
+++ b/src/web/components/ReaderRevenueLinks.tsx
@@ -191,6 +191,23 @@ export const ReaderRevenueLinks: React.FC<Props> = ({
     };
 
     if (shouldHideSupportMessaging()) {
+        if (edition !== 'US') {
+            return (
+                <div className={cx(inHeader && paddingStyles)}>
+                    <div
+                        className={cx({
+                            [hiddenUntilTablet]: inHeader,
+                        })}
+                    >
+                        <div className={messageStyles}> Thank you for your support </div>
+                        <div className={subMessageStyles}>
+                            <div> You've powered our journalism </div>
+                            <div> through a historic year </div>
+                        </div>
+                    </div>
+                </div>
+            )
+        }
         return null;
     }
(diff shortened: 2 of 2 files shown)


src/web/components/ReaderRevenueLinks.test.tsx
```diff
@@ -32,14 +32,14 @@ describe('ReaderRevenueLinks', () => {
     };
     const edition: Edition = 'UK';
 
-    it('should not render if shouldHideSupportMessaging() returns true', async () => {
+    it('should not render if shouldHideSupportMessaging() returns true and edition is US', async () => {
```
TomPretty, line 35: should this be `or edition is US`?

tomrf1: while the campaign is running, all non-US supporters will see the "thank you" message.
so this test now also needs to set edition to US

TomPretty: oh okay - so we're actually ignoring `shouldHideSupportMessaging()` outside of the US?

tomrf1: kind of - but we're not showing a cta, just some text!

oliverlloyd: Could this mean we might show these messages to signed in readers in the US?

tomrf1: No supporters on the US edition will see this message.
Supporters on other editions will see it.

oliverlloyd: Ahhh, is this perhaps more a problem with the _name_ of the function `shouldHideSupportMessaging`? Looking into it it, this function returns true if:

```typescript
!shouldShowSupportMessaging() || isRecurringContributor(isSignedIn) || isRecentOneOffContributor();
```

`shouldShowSupportMessaging` resolves to `getCookie(HIDE_SUPPORT_MESSAGING_COOKIE) === 'true';`

and that cookie is set based on the value of `showSupportMessaging` in the server response, which gets set based on

```scala
  // show support messaging (in app & on dotcom) if they do NOT have any active products
  // TODO in future this could become more sophisticated (e.g. two weeks before their products expire)
  lazy val showSupportMessaging = !(
    isPaidTier
      || isRecurringContributor
      || isRecentOneOffContributor
      || digitalSubscriberHasActivePlan
      || isPaperSubscriber
      || isGuardianWeeklySubscriber
      || isPremiumLiveAppSubscriber
    )
```

I think there's an argument here to rename `shouldShowSupportMessaging` to something like `isSupporter` / `isNotSupporter`, which might better represent what it is saying and remove any confusion for people landing on this code in the future. What do you think?

tomrf1: I'm up for renaming that function, though I'd rather not do that in this PR as it's used in a few unrelated places!

oliverlloyd: Sounds good! I was thinking, reading that TODO comment, maybe something like `showCTA` makes more sense? But I'm totally happy to let you think of the best name, you have more knowledge around the context and lexicon in this area

oliverlloyd (approved):
👍

src/web/components/ReaderRevenueLinks.test.tsx
tomrf1: 👍 thanks @oliverlloyd !

PR #2216, opened 2020-12-10
Enabling the do-not-track feature of the vimeo video embed for amp articles

frj:
## What does this change?
This change enables the do-not-track feature for Vimeo videos embedded in amp articles.

For more details see the dnt flag in the Vimeo docs:

https://developer.vimeo.com/api/oembed/videos

> Whether to prevent the player from tracking session data, including cookies. Keep in mind that setting this argument to true also blocks video stats.

This change is dependent on a PR to the amphtml project:

https://github.com/ampproject/amphtml/pull/31406

Which has been merged but has not as yet been released.

In theory the PR should be tagged with 'PR Use: In Stable' once it is release for more details see:

https://amp.dev/documentation/guides-and-tutorials/learn/spec/release-schedule/#determining-if-your-change-is-in-a-release

## Why?

This is part of an ongoing effort to improve the privacy of users using the guardian products, in this case by stopping vimeo from collecting information on them through embedded vimeo videos.


src/amp/components/elements/VideoVimeoBlockComponent.tsx
```diff
@@ -15,6 +15,7 @@ export const VideoVimeoBlockComponent: React.FC<{
                 layout="responsive"
                 width={element.width}
                 height={element.height}
+                do-not-track="do-not-track"
```
oliverlloyd, line 18: I presume sending this attribute even if it's not supported won't have any adverse side effects?

frj: It shouldn't, but Im intending to hold off merging this change until the PR for the changes to amp-html are live.

oliverlloyd (approved):

src/model/json-schema.json
```diff
@@ -2490,7 +2490,7 @@
                 "type"
             ]
         },
-        "Pillar": {
+        "CAPIPillar": {
```
gtrufitt, line 2493: Should this change be a part of this PR? It looks like a breaking change if not matched with Frontend cc/ @oliverlloyd

gtrufitt (commented):
Looks like the AMP change has gone in so LGTM bar the query around CAPIPillar

src/model/json-schema.json
oliverlloyd: Not sure how this came about to be honest but I suspect if you merge in `main` it should go away?

frj: Yes, Ive merged main and its gone away.

frj:
The changes to amp-html are live, testing the changes locally and the do-not-track flag is seems to be working:

Screenshot 2021-01-12 at 15 55 03

shtukas (approved):

PR #2219, opened 2020-12-11
Fix some HTML validation errors

jfsoul:
## What does this change?
I ran a DCR article through an HTML validator yesterday for an unrelated reason and there were quite a few errors that cropped up. This PR resolves a few of them (the one's that seemed easy to fix).

### Before
Before

### After
Same as before but with more valid HTML

## Why?
While most browsers will handle small validation errors seamlessly, it doesn't hurt to be closer to valid HTML.


oliverlloyd (approved):
Very happy we've got Chromatic running on this one!

liywjl (approved):

PR #2220, opened 2020-12-11
Discussion Cypress Tests

oliverlloyd:
Ahead of some work to refactor discussion out of App I wanted to add some tests to our Cypress suite for the things that I might break when I do that work. Specifically, these are tests for those features that depending on shared state.

Adding them here in a new PR because 1) it's lunch so why not get them out in their own PR and 2) because I want to be 100% confident they will work hapilly on CI before I start depending on them


liywjl (approved):
LGTM 👍

PR #2221, opened 2020-12-11
Use 100vh for height of immersive image

gtrufitt:
## What does this change?

Add 100vh to immersive main media to fix Safari bug.

## Overflow hidden

Avoid spilling out into the main content. Before and after shows identical styling.

### Before 

Screen Shot 2020-12-11 at 14 47 15

### AFter

Screen Shot 2020-12-11 at 14 47 27

### DIFF
--- a/src/web/components/elements/ImageComponent.tsx
+++ b/src/web/components/elements/ImageComponent.tsx
@@ -248,7 +248,7 @@ export const ImageComponent = ({
                     /* These styles depend on the containing layout component wrapping the main media
                     with a div set to 100vh. This is the case for ImmersiveLayout which should
                     always be used if display === 'immersive' */
-                    height: 100%;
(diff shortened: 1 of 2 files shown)


liywjl (approved):
LGTM, is this only a safari issue? shame we cant pick this up in storybook

src/web/components/elements/ImageComponent.tsx
```diff
@@ -248,7 +248,7 @@ export const ImageComponent = ({
                     /* These styles depend on the containing layout component wrapping the main media
                     with a div set to 100vh. This is the case for ImmersiveLayout which should
                     always be used if display === 'immersive' */
```
oliverlloyd, line 250: More a note so we don't forget. This comment is now out of date

PR #2222, opened 2020-12-11
Change Print Layout so it can actually be read!

nitro-marky:
### Before
DCR page prints looked quite janky, hopefully this will remove a lot of unnecessary elements.
It's not perfect but is a lot better

### After
Screenshot 2020-12-11 at 14 59 43

### DIFF
--- a/src/web/browser/coreVitals/coreVitals.ts
+++ b/src/web/browser/coreVitals/coreVitals.ts
@@ -44,7 +44,6 @@ const getValue = ({ name, value }: coreVitalsArgs): void => {
     console.log(jsonData);
 }
 
-
     document.addEventListener("DOMContentLoaded", function() {
 
         if(window.guardian && window.guardian.ophan)
@@ -55,11 +54,12 @@ const getValue = ({ name, value }: coreVitalsArgs): void => {
         getLCP(getValue);
         getFCP(getValue);
         getTTFB(getValue);
-        getCLS(getValue);
-    });
 
-    window.addEventListener('unload', function(event) {
-        getCLS(console.log);
+        // At the moment a final CLS score is only reported on a 'visiblitychange' event (which is far from ideal)
+        // https://github.com/WICG/layout-instability#computing-dcls-with-the-api
+        // We need to find a more reliable way to record CLS without switching tabs.
+        // I looked at using the 'unload' event but it only returns the cls delta for that event, which is incorrect.
+        getCLS(getValue);
     });
 
 };
--- a/src/web/browser/coreVitals/coreVitals.ts
+++ b/src/web/browser/coreVitals/coreVitals.ts
@@ -1,56 +1,63 @@
 import { getCLS, getFID, getLCP, getFCP, getTTFB } from 'web-vitals';
 
+
+const jsonData = {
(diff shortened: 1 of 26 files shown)


nitro-marky:
[Print Layout - Hide elements](https://trello.com/c/HeNfqVlw/1778-print-layout-hide-elements)

src/web/components/AdSlot.tsx
```diff
@@ -150,12 +173,13 @@ export const AdSlotCore: React.FC<{
     const sizeMappings = makeInternalSizeMappings(sizeMapping);
     return (
         <div
+            data-print-layout='hide'
             id={`dfp-ad--${optId || name}`}
-            className={`${makeClassNames(
+            className={`${cx(makeClassNames(
```
liywjl, line 178: no need for cx

src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -0,0 +1,71 @@
...
+        getCLS(getValue);
+    });
+
+};
+
+
+
+
+
+
```
liywjl, line 71: looks like prettier is not formatting on save. I had a similar issue on VSC that was solved by updating VSC (just redownload it and launch the new download, that fixed issues for me)

src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -0,0 +1,71 @@
...
+    type coreVitalsArgs = {
+        name: string;
+        value: number;
+    };
+
+
+const getValue = ({ name, value }: coreVitalsArgs): void => {
+    switch(name){
+        case('FCP'):
+            console.log(`FCP: ${  value}`);
```
liywjl, line 24: You probably want to as ESLint disable for console log to avoid getting build errors

src/web/components/AdSlot.tsx
```diff
@@ -20,6 +20,9 @@ export const labelStyles = css`
         color: ${text.supporting};
         text-align: left;
         box-sizing: border-box;
+        @media print{
+        display: none !important;
+        }
```
liywjl, line 25: I am relucatnt to use `!important` in general (maybe we can check that page for print and just not render?) also I do not know if we want to remove adds for print layout? @gtrufitt @oliverlloyd @jfsoul thoughts?

src/web/components/Logo.tsx
```diff
@@ -52,8 +52,12 @@ const style = css`
         fill: ${brandText.primary};
     }
 `;
-
-const SVG = () => <TheGuardianLogoSVG className={style} />;
+const blackLogo = css`
+    @media print{
+        colour: #000000 !important;
```
liywjl, line 57: avoid using `!important`, maybe look at using source for colour?

src/web/components/AdSlot.tsx
liywjl: You use `data-print-layout='hide'` in a lot of HTML to hide in print. Maybe that could be a better idea than using CSS `!important` ?

liywjl (commented):
There are a couple things to change ^^ but I see you have just closed it :P

PR #2223, opened 2020-12-11
Wrapped Series Tag

oliverlloyd:
## What does this change?
Fixes a styling glitch seen when longer series tags were wrapping on smaller screens

### Before
Screenshot 2020-12-11 at 15 24 07

### After

Screenshot 2020-12-11 at 15 25 37

### DIFF
--- a/src/web/components/SeriesSectionLink.tsx
+++ b/src/web/components/SeriesSectionLink.tsx
@@ -75,8 +75,14 @@ const invertedStyle = (pillar: Pillar) => css`
     white-space: pre-wrap;
     box-shadow: -6px 0 0 ${pillarPalette[pillar].main};
     box-decoration-break: clone;
-    line-height: 26px;
+    line-height: 28px;
+    ${from.leftCol} {
+        line-height: 28px;
+    }
 
+    padding-right: ${space[3]}px;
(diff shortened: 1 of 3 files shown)


src/web/components/SeriesSectionLink.tsx
```diff
@@ -71,17 +71,25 @@ const invertedStyle = (pillar: Pillar) => css`
     color: ${neutral[100]};
     background-color: ${pillarPalette[pillar].main};
 
+    /* Handle text wrapping onto a new line */
+    white-space: pre-wrap;
+    box-shadow: -6px 0 0 ${pillarPalette[pillar].main};
+    box-decoration-break: clone;
```
gtrufitt, line 77: Nice solution. IE11 might look odd with no support (though maybe just one big block behind?)

liywjl (approved):

src/web/components/SeriesSectionLink.tsx
oliverlloyd: Without this clone the second line has no left padding but we've fixed the overlap problem so IE11 experience will still be better than before

HarryFischer:
😎 thank you!

PR #2225, opened 2020-12-11
Platform support for new Braze banner campaign

tjmw:
## What does this change?

This PR includes a few changes to the platform code around showing Braze banners, to support a new campaign which will run alongside the existing one.

In particular:

* Because users might now be eligible for > 1 message from Braze we now unsubscribe from in-app message events after we receive an event. This is so that we don't have multiple messages delivered from Braze in the same page request (we wouldn't actually _show_ > 1 message because we wrap the callback logic up in a promise which can only ever resolve once).

* Update the pre-check to look at the `gu_hide_support_messaging` cookie instead of `gu_digital_subscriber`. The new campaign has a wider audience, though we'd still like to restrict the number of readers we load the Braze SDK for. Broadening to users who _shouldn't_ see support messaging (which means they have a product, or otherwise subscribed/contributed).

Also includes a bonus fix to move the `force-braze-message` host check further down so we only ever run it if the query string arg was specified (right now we're console.log-ing that the current host isn't valid on every page view).

### DIFF
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -161,15 +161,15 @@ const getBrazeMetaFromQueryString = (): Meta | null => {
     if (URLSearchParams) {
         const qsArg = 'force-braze-message';
 
-        if (!FORCE_BRAZE_ALLOWLIST.includes(window.location.hostname)) {
-            // eslint-disable-next-line no-console
-            console.log(`${qsArg} is not supported on this domain`);
-            return null;
-        }
-
         const params = new URLSearchParams(window.location.search);
         const value = params.get(qsArg);
         if (value) {
+            if (!FORCE_BRAZE_ALLOWLIST.includes(window.location.hostname)) {
+                // eslint-disable-next-line no-console
+                console.log(`${qsArg} is not supported on this domain`);
+                return null;
+            }
+
             try {
                 const dataFromBraze = JSON.parse(value);
 
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -48,22 +48,26 @@ export const hasRequiredConsents = (): Promise<boolean> =>
 type PreCheckArgs = {
     brazeSwitch: boolean;
     apiKey?: string;
-    isDigitalSubscriber?: boolean;
+    shouldHideSupportMessaging?: boolean;
     pageConfig: { [key: string]: any };
 };
 
 export const canShowPreChecks = ({
     brazeSwitch,
     apiKey,
-    isDigitalSubscriber,
+    shouldHideSupportMessaging,
     pageConfig,
-}: PreCheckArgs) =>
-    Boolean(
-        brazeSwitch &&
-            apiKey &&
-            isDigitalSubscriber &&
-            !pageConfig.isPaidContent,
+}: PreCheckArgs) => {
+    // Currently all active web canvases in Braze target existing supporters,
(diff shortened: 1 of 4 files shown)


src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -13,15 +13,15 @@ type Props = {
     asyncCountryCode?: Promise<string>;
     CAPI: CAPIBrowserType;
     asyncBrazeUuid?: Promise<null | string>;
-    isDigitalSubscriber?: boolean;
+    shouldHideSupportMessaging?: boolean;
```
adamnfish, line 16: I wonder if we could encapsulate our logic a bit better here by naming this something that *we* care about, rather than having the name be based on the (current) source for the data?

The actual `ts` logic is where we do our business logic to decide how to render the component, and the `tsx` component only needs to be given that information. This feels a bit like leaking the implementation details into the template. Is the actual concept we're trying to convey here something along the lines of "this impression is eligible for Braze Web Messages", and if so would a name along those lines make it:
- easier to understand these template files in isolation
- easier to refactor the "can show" logic to consider other cases in the future

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -48,22 +48,22 @@ export const hasRequiredConsents = (): Promise<boolean> =>
...
-    isDigitalSubscriber?: boolean;
+    shouldHideSupportMessaging?: boolean;
     pageConfig: { [key: string]: any };
 };
 
 export const canShowPreChecks = ({
     brazeSwitch,
     apiKey,
-    isDigitalSubscriber,
+    shouldHideSupportMessaging,
```
adamnfish, line 58: As with the Frontend version, it might be nice to document our intent here, so that the reader can see whether `shouldHideSupportMessaging` meets that requirement now and in the future?

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
tjmw: Hey yeah this is a great point, thanks for raising. Rhys and I were discussing something similar on Friday. I agree that at some point we need to map the source of the data `shouldHideSupportMessaging` to the business logic we care about. This component, `StickyBottomBanner`, is the orchestrator of which banner actually gets shown and part of its responsibility is gathering and passing on the inputs which allow each of the banner `canShow` functions to make a decision. So to me it seems like the `canShow` of the relevant banner is the place where the mapping should take place. I think that's consistent with how the reader revenue `canShow` works for example. Happy to chat about this!

tjmw: I've added a comment to the preChecks in `BrazeBanner.tsx` which hopefully makes things clearer to future developers (including ourselves!)

tjmw:
I've marked this as ready for review. Ideally I'd like to get confirmation from Braze that unsubscribing from in-app messages in the callback won't have any hidden side effects.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -48,22 +48,26 @@ export const hasRequiredConsents = (): Promise<boolean> =>
...
     pageConfig,
-}: PreCheckArgs) =>
-    Boolean(
-        brazeSwitch &&
-            apiKey &&
-            isDigitalSubscriber &&
-            !pageConfig.isPaidContent,
+}: PreCheckArgs) => {
+    // Currently all active web canvases in Braze target existing supporters,
```
adamnfish, line 61: Great!

adamnfish (approved):

shtukas (approved):

PR #416, opened 2019-02-11
Amp ad element

nicl:
## What does this change?

Populate ad placeholders with actual ads! The was started by @philmcmahon with a bit of finishing by me.

I'm sure there are bugs here so I'm going to chat with commercial to get more feedback, but I think it will probably be easiest to discover most issues on prod.

## Why?

Crucial for go-live.

screenshot 2019-02-11 at 11 48 17

PR #2227, opened 2020-12-11
Teads: multi-size Ads

mxdvl:
## What does this change?

This change enables multi-size ads. It also switches the layout to `fixed`.

**Teads** offers video formats, where a lot of content has a 16:9 ratio, roughly translating to 350x170 pixel dimensions. They should be able to resize the ad slot without having grey bands top and bottom. The new array of sizes is thus:

- _300x250 (MPU: what we’ve always had)_
- ~~300x169 (16:9)~~
- 250x250 (square)

Setting the dimensions of the ad in CSS is redundant, as AMP also sets these values inline. `min-content` ensures that if an ad is narrower than the largest size, it the top `Advertisement` banner will shrink.

## Screenshots

> (Ads are different, but their sizes is what should be compared)

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/102130412-d6e1df00-3e1e-11eb-94cf-c49a84808139.png
[after]: https://user-images.githubusercontent.com/76776/102130346-bdd92e00-3e1e-11eb-8a17-48d70ebc46c0.png

## Why?

Since their inceptions in #416, it seems some attributes of `amp-ad` have not been challenged:
- size: `300x250`
- layout: `responsive`

## Todo:

- [X] Follow-up with Teads on implementation
- [X] Confirm the behaviour is parsed correctly by Prebid
- [ ] Monitor performance & revenue


ashishpuliyel (approved):
This is more or less exactly what I thought this change needed, when I looked at the ask originally and said something dismissive like 'easy job sure we can do it'. I was later worried  this had become much more complicated. I'm relieved and happy it's come back to a nice and straightforward change.

Having said that, putting the sizes in an array at the top and mapping them back into a string is much cuter than anything I would have done. Also the PR is beautifully explained. Very helpful for the future developer if they ever need to change this array.
Good job!

liywjl (approved):
LGTM 👍

mxdvl:
PR deployment that lasts on https://guardian-amp-teads.netlify.app/

src/model/json-schema.json
```diff
@@ -2741,6 +2741,9 @@
                 "brazeApiKey": {
                     "type": "string"
                 },
+                "ipsosTag": {
+                    "type": "string"
+                },
```
mxdvl, line 2746: Not sure why this crept in…?

liywjl: if you change the TS types, we auto generate a schema based on those types to use to validate CAPI

mxdvl: No sure where the TS type would have changed, in this case.

liywjl: I think it was @jfsoul who forgot to push it as part of his PR https://github.com/guardian/dotcom-rendering/pull/2396

I think its fine

PR #2228, opened 2020-12-11
Fix for encoding comscore keywords

ashishpuliyel:
## What does this change?
DCR didn't correctly encode keywords for comscore keywords the same way that frontend did.
Keywords are passed as part of the src of an `img` tag (in a `<noscript>` tag) and the spaces in the `src` field meant the page didn't validate.

This change replicates the encoding from frontend (which includes `+` instead of spaces)

### Before DCR
before

### After DCR
after

### Frontend (DCP) for comparison:
DCP

## Why?
Spaces in the `img` src aren't valid.

### DIFF
--- a/src/web/server/htmlTemplate.ts
+++ b/src/web/server/htmlTemplate.ts
@@ -206,7 +206,7 @@ export const htmlTemplate = ({
                 </script>
 
                 <noscript>
-                    <img src="https://sb.scorecardresearch.com/p?c1=2&c2=6035250&cv=2.0&cj=1&cs_ucfr=0&comscorekw=${keywords}" />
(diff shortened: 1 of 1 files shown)


mxdvl (approved):
LGTM 👍 

See original encoding in `frontend`: https://github.com/guardian/frontend/blob/94e3b7a8040d188d4c3a4e5ec64cecc6a9f0c34a/common/app/views/fragments/analytics/comscore.scala.html#L9

mxdvl:
Converting spaces to `+` is required to comply with `application/x-www-form-urlencoded`. See [the docs][] and [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent).

[the docs]: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#application/x-www-form-urlencoded-encoding-algorithm

jfsoul (approved):

PR #2232, opened 2020-12-14
Move Comments logic out of App

oliverlloyd:
## What?
1. Refactors `App.tsx` to move all discussion related logic down and into a new `Discussion.tsx` component.
2. Moves `CommentsLayout` logic into the `Discussion.tsx` component

## Why?
The `CommentCount` and `CommentsLayout` components share state. When the reader clicks on the count of comments we need to tell the nested `CommentsLayout` to expand. However, having this shared state at the same level as `App` is undesirable because it causes double hydration. By moving the state down we can hydrate it once and let the state be shared with normal rerenders.

Once a `Discussion` level container existed, it became logical to make that the home for the code in `CommentsLayout` allowing that component to be deleted. As part of that move I:

a) refactored the way we insert comments into the page to use `ContainerLayout`
b) removed and simplified the height calculations we were using for the ad slot (no more hackity hack hack 🎉 )

### Why does this PR also change Counts?
This move was made more complicated because the `CommentCount` component needed to be broken out. Before we set the `Counts` component as a portal and this managed getting and displaying both share count and comment count. But now we need to be able to set the share count by itself even if the `Discussion` component is never rendered (non opinion pieces) so I replaced the single portal root of `share-comment-counts` with `share-count-root` and `comment-count-root` and moved all the logic for each down.

To make this work I deleted the `NumbersBorder` component that was being used to display the vertical divider line and instead used css to decide when to set the border:

```css
   aside:not(:empty) + aside:not(:empty) {
     border-left: ...;
   }
```

As part of this refactor I also moved a lot of the tests around but made sure to maintain coverage


liywjl (approved):
Amazing work!! 👍

src/web/components/Discussion.tsx
```diff
@@ -0,0 +1,138 @@
...
+    return (
+        <>
+            <Portal root="comment-count-root">
+                <CommentCount
+                    isCommentable={isCommentable}
+                    commentCount={commentCount}
+                    pillar={pillar}
+                    setIsExpanded={setIsExpanded}
+                />
+            </Portal>
```
gtrufitt, line 116: This actually feels quite nice to me from a 'keep relevant components together' POV. Nicer than a big bucket in App? Maybe we think about how we do Portals in general...

src/web/components/Counts.stories.tsx
```diff
@@ -41,13 +43,22 @@ export const Both = () => {
...
     return (
         <Container>
-            <Counts
-                ajaxUrl="https://api.nextgen.guardianapps.co.uk"
-                commentCount={239}
-                pageId="/lifeandstyle/2020/jan/25/deborah-orr-parents-jailers-i-loved"
-                pillar="news"
-                setIsExpanded={() => {}}
-            />
+            <Counts>
```
gtrufitt, line 46: I'm not too fussed, but feels like we could put `Counts` in meta now as it doesn't do a whole lot - it feels like it exists as a component for stories and we could move those stories to meta and keep the render tree shallower. I don't think for this PR though!

src/web/components/Counts.tsx
```diff
@@ -3,97 +3,25 @@ import { css } from 'emotion';
...
-    const { data: shareData, error: shareError } = useApi<ShareCountType>(
-        shareUrl,
-    );
-
-    if (shareError) {
-        window.guardian.modules.sentry.reportError(
-            shareError,
-            'share-comment-counts',
-        );
+    aside:not(:empty) + aside:not(:empty) {
```
gtrufitt, line 13: As we're relying on html structure and elements here, would it be worth adding classes and targeting them rather than `aside` to imply more of a contract?

gtrufitt (approved):
Excellent!

src/web/components/Counts.stories.tsx
oliverlloyd: I had the same though but, indeed, once you move it into meta then it's harder to test, you'd have to export it, or duplicate the border css, which felt messy so I held off.

I suppose a better solution would be to incorporate the counts tests into `ArticleMeta`

src/web/components/Counts.tsx
oliverlloyd: This css grew logically but I think you're right, looking at this fresh a simple class would be more readable

shtukas (approved):

oliverlloyd:
[Move Comments logic out of App](https://trello.com/c/HfoXd06n/2228-move-comments-logic-out-of-app)

PR #2234, opened 2020-12-14
Improve Print Layout

nitro-marky:
## What does this change?
This makes the print layout a little more acceptable by removing certain page elements. (There is potentially more to do but this is a start). There are still edge cases, like atoms-rendering but this is a start.

### Before
Screenshot 2020-12-14 at 11 30 30

### After
Screenshot 2020-12-14 at 11 30 11

### DIFF
--- a/src/web/layouts/StandardLayout.tsx
+++ b/src/web/layouts/StandardLayout.tsx
@@ -340,14 +340,12 @@ export const StandardLayout = ({
 
                 <Stuck>
                     <Section
-                        data-print-layout='hide'
                         showTopBorder={false}
                         showSideBorders={false}
                         padded={false}
                         shouldCenter={false}
                     >
                         <HeaderAdSlot
-                            data-print-layout='hide'
                             isAdFreeUser={CAPI.isAdFreeUser}
                             shouldHideAds={CAPI.shouldHideAds}
                         />
@@ -355,7 +353,6 @@ export const StandardLayout = ({
                 </Stuck>
                 <SendToBack>
                     <Section
-                        data-print-layout='hide'
                         showTopBorder={false}
                         showSideBorders={false}
                         padded={false}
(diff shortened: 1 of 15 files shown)


src/web/layouts/StandardLayout.tsx
```diff
@@ -336,22 +336,26 @@ export const StandardLayout = ({
     const { branding } = CAPI.commercialProperties[CAPI.editionId];
     return (
         <>
-            <div>
+            <div data-print-layout='hide'>
```
liywjl, line 339: do you need to repeat `data-print-layout='hide'` on child DOM nodes?

src/web/layouts/StandardLayout.tsx
```diff
@@ -559,24 +568,25 @@ export const StandardLayout = ({
             </Section>
 
             <Section
+                data-print-layout='hide'
```
liywjl, line 571: do you need to repeat data-print-layout='hide' on child DOM nodes?

liywjl (changes requested):
I do not think you need to repeat `data-print-layout='hide'` on child DOM nodes

src/web/layouts/StandardLayout.tsx
nitro-marky: I wasn't sure as it was outside of a div, I'll give it a try. I've already removed some of the nesting

src/web/layouts/StandardLayout.tsx
nitro-marky: Where does standard layout get that attribute from?

nitro-marky: Where would you recommend putting that attribute?

nitro-marky:
[Print Layout - Hide elements](https://trello.com/c/HeNfqVlw/1778-print-layout-hide-elements)

src/web/layouts/StandardLayout.tsx
nitro-marky: I tried changing the attribute up the hierarchy and it failed to work. Unless there are other methods I'm happy that this works for now.

liywjl (approved):
👍

PR #2239, opened 2020-12-15
RightColumn padding

oliverlloyd:
## What does this change?
Hoists the 6px padding that lives above the RightColumn out into the layouts so it's set in context

## Why?
So we can use `RightColumn` without it being padding

### DIFF
--- a/src/web/layouts/StandardLayout.tsx
+++ b/src/web/layouts/StandardLayout.tsx
@@ -336,8 +336,7 @@ export const StandardLayout = ({
     const { branding } = CAPI.commercialProperties[CAPI.editionId];
     return (
         <>
-            <div data-print-layout='hide'>
-
+            <div data-print-layout="hide">
                 <Stuck>
                     <Section
                         showTopBorder={false}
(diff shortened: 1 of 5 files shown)


shtukas (approved):

src/web/layouts/StandardLayout.tsx
```diff
@@ -548,37 +551,56 @@ export const StandardLayout = ({
...
             >
-                <AdSlot data-print-layout='hide' asps={namedAdSlotParameters('merchandising-high')} />
+                <AdSlot
+                    data-print-layout="hide"
+                    asps={namedAdSlotParameters('merchandising-high')}
+                />
             </Section>
 
             {!isPaidContent && (
                 <>
```
liywjl, line 589: unrelated but maybe worth moving ` data-print-layout="hide"` here?

liywjl (approved):
LGTM

src/web/layouts/StandardLayout.tsx
oliverlloyd: That change popped up here because of Prettier glitches elsewhere. I'm only vaguely aware what it's about tbh!

PR #2242, opened 2020-12-15
Add ARD for Script Loading methods on DCR

gtrufitt:
## What does this change?

Add an ARD for our Script Loading mechanisms, module/nomodule, preloading approach (without recommendations for technical implementation).

Read here: https://github.com/guardian/dotcom-rendering/blob/gtrufitt/module-no-moduke/docs/architecture/ScriptLoading.md

## Why?

We want to try and implement modern script loading methods, and ensure that we're loading critical path early and deferring low priority scripts.


liywjl (approved):
great writeup!!! 👍

PR #2247, opened 2020-12-15
Move to block times for AMP

gtrufitt:
## What does this change?

Follows: https://github.com/guardian/frontend/pull/23384

See https://github.com/guardian/dotcom-rendering/commit/219a56ee084e80ac2ef4c68493245c130c3c52be for changes, the other commit is fixture updates.

We move to using the block times, rather than the article (we *only* use these fields in AMP, where the times used are for the *block*). See https://github.com/guardian/frontend/pull/23384 for screenshots.

## Why?

We were showing the wrong block times in AMP.


liywjl (approved):
LGTM

jfsoul (approved):

PR #2248, opened 2020-12-15
Mj explore amp multivariant testing and analytics

michaelbjacobson:
### AMP A/B testing and analysis
`support-dotcom-components` exposes an endpoint which returns a JSON object listing the names of all A/B tests, as well as the names of all their variants, along with the percentage of the audience should be assigned to each variant.

`dotcom-rendering` polls this endpoint every two minutes and caches the data in its local memory. 

When an AMP page is requested from `dotcom-rendering` the test data is retrieved from the cache and injected into the `amp-experiment` component on the document before being returned to the client.

AMP reads this test data and assigns users to a variant on each of the tests. This assignment is sticky, and AMP will maintain it on subsequent visits to the article.

As the user scrolls down the article, and begins to approach the epic, AMP sends a request to `support-dotcom-components` for the epic copy and data. This request includes a list of all the variant assignments for the user. 

`support-dotcom-components` receives the request and works out which epic test to use, based on geolocation and other factors. Then, it checks which variant has been assigned for the test, and returns the relevant copy to the client. The client renders the epic.

When the user views the epic a request is sent, from our Ophan `amp-analytics` component, to `support-dotcom-components` to record an impression. Once again, this request includes the list of all variant assignments for the user, and once again `support-dotcom-components` works out which test is in use. The relevant test, and variant, are then sent on to `ophan` as an event.

Epic tests are configured via the `support-admin-console`. This tool will be updated shortly to allow the creation of multivariate tests.

• [Adjustment to Ophan analytics script](https://github.com/guardian/ophan/pull/3983)
• [SDC changes](https://github.com/guardian/support-dotcom-components/pull/320)
• [Support Admin Console changes](https://github.com/guardian/support-admin-console/pull/189)
• [AMP experiment docs](https://amp.dev/documentation/components/amp-experiment/)
• [AMP analytics docs](https://amp.dev/documentation/components/amp-analytics/)


src/amp/server/render.tsx
```diff
@@ -48,13 +50,18 @@ export const render = ({ body }: express.Request, res: express.Response) => {
             canonicalURL: CAPI.webURL,
         };
 
+        const experimentsData: AmpExperiments =
+            await fetch(ampExperimentsDataUrl)
```
tomrf1, line 54: This will make a request for each page render. I don't think we want to introduce a network request here as it slows down rendering (and we have to go back to the event loop).

Instead, let's independently poll the endpoint and cache the data. This can be done with `setTimeout`, and it's ok if `experimentsData` is empty when the server first starts up

michaelbjacobson: Good point, thanks @tomrf1!

index.d.ts
```diff
@@ -755,6 +755,8 @@ declare module 'dynamic-import-polyfill' {
 declare namespace JSX {
     /* eslint-disable @typescript-eslint/no-explicit-any */
     interface IntrinsicElements {
+        'amp-experiment': any;
+        'amp-script': any;
```
tomrf1, line 759: `amp-script` can be removed?

src/amp/server/render.tsx
```diff
@@ -9,8 +9,9 @@ import { validateAsCAPIType as validateV2 } from '@root/src/model/validate';
 import { findBySubsection } from '@root/src/model/article-sections';
 import { bodyJSON } from '@root/src/model/exampleBodyJSON';
 import { generatePermutivePayload } from '@root/src/amp/lib/permutive';
+import { getAmpExperimentCache } from '@root/src/amp/server/ampExperimentCache';
 
-export const render = ({ body }: express.Request, res: express.Response) => {
+export const render = async ({ body }: express.Request, res: express.Response) => {
```
tomrf1, line 14: no longer needs to be `async`

index.d.ts
michaelbjacobson: Good spot, I'll remove that now!

src/amp/server/render.tsx
michaelbjacobson: 👍🏼

src/amp/components/AmpScript.tsx
```diff
@@ -0,0 +1,16 @@
+// import React from 'react';
+//
+// export const AmpScript: React.FC<{id: string, script: string}> = ({id, script}) => {
```
liywjl, line 3: no longer need this file? everything is commented out

michaelbjacobson: Yup, thanks for the catch! I'll remove it now :)

src/amp/components/Epic.tsx
```diff
@@ -9,7 +9,7 @@ import {
 } from '@root/src/amp/components/moustache';
 import { palette } from '@guardian/src-foundations';
 import { headline, body, textSans } from '@guardian/src-foundations/typography';
-import { Ticker } from '@root/src/amp/components/Ticker';
+import { Ticker } from "@root/src/amp/components/Ticker";
```
liywjl, line 12: you have some prettier issues. A fix to force prettier to run on push has been added to main so maybe worth pulling from `main`

src/amp/server/ampExperimentCache.ts
```diff
@@ -0,0 +1,40 @@
...
+    console.log(`------------------ called fetchExperimentsData at ${Date.now()} -----------------`)
+
+    return fetch(ampExperimentsDataUrl)
+        .then(rawResponse => rawResponse.json())
+        .then(json => {
+            ampExperimentCache = json
+        })
+}
+
+const oneMinute = 60_000;
```
liywjl, line 33: 🤯  never knew we could do that!!! 🎊

liywjl (approved):
need to resolve conflicts with main over overall looks good to me 👍

src/amp/components/Epic.tsx
michaelbjacobson: Sorted 💅🏼

src/amp/server/ampExperimentCache.ts
michaelbjacobson: Which thing @liywjl?

liywjl: using `_` within numbers in JS, never knew you could do that 😃

michaelbjacobson: Oh right, yeah, it’s great for readability! 🙂

src/amp/components/Epic.tsx
```diff
@@ -187,117 +187,123 @@ const buildUrl = (
...
+								<MoustacheVariable name="." />
+							</p>
+						</MoustacheSection>
+						<span className={highlightedTextStyle}>
+							<MoustacheVariable name="highlightedText" />
+						</span>
+						<br />
+						<MoustacheSection name="cta">
+							<a
+								href={buildUrl(
```
tomrf1, line 274: the `buildUrl` function now needs to add the `abTest` field, so we can track acquisitions against the test

tomrf1: One other thing - is it possible to add the `referrerPageviewId` field in as well? It's not essential, but if possible it would be good to pass through the ophan page view id

michaelbjacobson: Sorted ✅

tomrf1 (approved):
great stuff! 🎉

PR #2251, opened 2020-12-15
Nest AudioAtom a level down from App to avoid App rerenders

gtrufitt:
## What does this change?

Updating state in App.tsx is an anti-pattern within DCR, as it causes a re-hyrdation of the hydrate components within App. 

Therefore we need to push our state for `shouldUseAcast` a level lower in order to only re-render the audio atom.

### DIFF
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -431,7 +431,9 @@ export const App = ({ CAPI, NAV }: Props) => {
                         kicker={audioAtom.kicker}
                         title={audioAtom.title}
                         pillar={pillar}
-                        CAPI={CAPI}
+                        contentIsNotSensitive={!CAPI.config.isSensitive}
+                        aCastisEnabled={CAPI.config.switches.acast}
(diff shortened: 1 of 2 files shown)


liywjl (approved):
LGTM

src/web/components/App.tsx
```diff
@@ -454,13 +425,15 @@ export const App = ({ CAPI, NAV }: Props) => {
...
                         id={audioAtom.id}
                         trackUrl={audioAtom.trackUrl}
                         kicker={audioAtom.kicker}
                         title={audioAtom.title}
-                        pillar={toTypesPillar(pillar)}
-                        shouldUseAcast={shouldUseAcast}
+                        pillar={pillar}
+                        contentIsNotSensitive={!CAPI.config.isSensitive}
+                        aCastisEnabled={CAPI.config.switches.acast}
+                        readerCanBeShownAds={!CAPI.isAdFreeUser}
```
oliverlloyd, line 436: 👍 for the name

shtukas (approved):

oliverlloyd (approved):

PR #2254, opened 2020-12-16
Use diffThreshold to reduce false negatives

oliverlloyd:
## What does this change?
I'd disabled Chromatic on some stories to try to prevent some false negatives but we're still seeing them on similar stories in the same file so this PR enables them again and reduces the `diffThreshold` for the whole file


jfsoul (approved):
What's the default threshold?

oliverlloyd:
> What's the default threshold?

`.063` where 0 is super accurate and 1 really isn't

https://www.chromatic.com/docs/threshold

PR #2255, opened 2020-12-16
Don't pass classnames to stick ads, use props instead

oliverlloyd:
## What?
We're passing classnames into `AdSlot` as a way of changing it's appearance and how it behaves. We use `localStyles`, `labelStyles`, `optClassNames` and then calculate a few more.

The problem with this approach is it's very hard to prevent one classname overriding styles from another. An example of this is when we want a slot to be sticky we should use `position: sticky`, but this is being wiped out by other classNames trying to set `position: relative`.

Instead of trying to escalate this css classname war, here I am leaning into React and using props instead.

## Depends on
This change depends on the Comments layout work because we need the fix to how we set the ad container  height that's [in that PR](https://github.com/guardian/dotcom-rendering/pull/2232)

## Going forward
I think we should look to do this more, both in general, but also specifically in this component. Ideally, we can end up with no need for any classnames to be passed as props


liywjl (approved):
LGTM

oliverlloyd:
This code was merged as part of https://github.com/guardian/dotcom-rendering/pull/2232 because they had interdependant changes

PR #2258, opened 2020-12-16
Refactor AdSlot

oliverlloyd:
## What?
This is a refactor of the `AdSlot` component and associated logic. I've mostly removed abstractions in favour of inline code.

## Why?
This is admittedly subjective, but I find it considerably easier to read code like this. But more than that I've noticed that this code has grown hard to work with over time. In particular, one issue I've hit several times is with styling. This component grew to accept several different classnames as props which made it difficult to change or understand the eventual styles. It also created the risk of styles overriding each other - which happened in the wild at least once. Hopefully by purely using React props and inlining all the abstractions we can reset the logic to make it easier to manage going forward.

Finally, in terms of why now, this is an area where we will 🤞 be making larger changes soon and simple, inline code like this should hopefully make that task easier

### Why was display added as a prop?

This is a pattern used a lot in DCR. The idea is that by being declarative, by giving the component the context, it is free to make the choices it needs based on this information. The alternative is to be more imperative and say `beSticky` or `hideWrapper`, which isn't _wrong_ but it tends to scale less well. Where possible, and in particular around `display`, `pillar` and `design`, we try to take the declarative approach.

Sorry about all the prop drilling poluting the diffs!


shtukas (approved):

src/web/components/AdSlot.tsx
```diff
@@ -87,114 +127,210 @@ const mobileStickyAdStyles = css`
...
+                                css`
+                                    position: sticky;
+                                    top: 0;
+                                `,
+                                labelStyles,
+                            )}
+                            data-link-name="ad slot right"
+                            data-name="right"
+                            // mark: 01303e88-ef1f-462d-9b6e-242419435cec
+                            data-mobile={`${Size.outOfPage}|${Size.empty}|${Size.mpu}|${Size.googleCard}|${Size.halfPage}|fluid`}
```
mxdvl, line 155: For consistency:

```suggestion
                            data-mobile={`${Size.outOfPage}|${Size.empty}|${Size.mpu}|${Size.googleCard}|${Size.halfPage}|${Size.fluid}`}
```

It’s also used in other locations in this file.

mxdvl (commented):
LGTM, except for the long ad sizes chains such as:

`{`${Size.outOfPage}|${Size.empty}|${Size.leaderboard}|940,230|900,250|${Size.billboard}|${Size.fabric}|fluid`}`

I would favour an array that’s then `.join("|")`, like it was in with `adSizeNamesToString`. Especially if each size is on a different line, it’s easier to parse when and why these arrays change in the future.

src/web/components/AdSlot.tsx
oliverlloyd: I think you're right. It's hard to read these data attributes with all the variables jumbled up in there, I'm not sure though if there's a good solution here, string manipulations are rarely easy to read! Maybe this code shouldn't be simplified even further and just written out as it will end up in the browser?

So instead of

```html
      data-name="right"
      // mark: 01303e88-ef1f-462d-9b6e-242419435cec
      data-mobile={`${Size.outOfPage}|${Size.empty}|${Size.mpu}|${Size.googleCard}|${Size.halfPage}|fluid`}
      aria-hidden="true"
```

we have:


```html
      data-name="right"
      // mark: 01303e88-ef1f-462d-9b6e-242419435cec
      data-mobile="1,1|2,2|300,250|300,274|300,600|fluid"
      aria-hidden="true"
```

I quite like this simplicity; it's certainly clear and simple to read. The disadvantage though is that you lose the intrinsic naming of the different sizes. (You could also say that it's nice to change the code in one place and have that propogate but I text editors today make that kind of work trivial.)

What do you think? I'm happy to defer to your opinion here. You probably have a better idea on the importance of these size names so if you think they're worth keeping I'll try to tidy things up to be more consistent. But if you think they can be lost I'll par things down to pure html!

liywjl: another option that could be more readable could be something like this:
```
data-mobile={[
    `${Size.outOfPage}`
    `${Size.empty}`
    `${Size.mpu}`
    `${Size.googleCard}`
    `${Size.halfPage}`
    `fluid`
  ].join("|")
}
```

liywjl (approved):

src/web/components/AdSlot.tsx
oliverlloyd: This is much nicer! Now it's an array, prettier is able do its thing

src/web/components/AdSlot.tsx
```diff
@@ -255,8 +289,21 @@ export const AdSlot: React.FC<Props> = ({ position, display }) => {
...
+                        `${Size.outOfPage}`,
+                        `${Size.empty}`,
+                        `${Size.fabric}`,
+                        `${Size.fluid}`,
+                        `${Size.leaderboard}`,
+                    ].join('|')}
+                    data-desktop={[
+                        `${Size.outOfPage}`,
+                        `${Size.empty}`,
+                        `${Size.leaderboard}|940,230|900,250`,
```
mxdvl, line 302:
```suggestion
                        `${Size.leaderboard}`,
                        `940,230`,
                        `900,250`,
```

mxdvl (approved):
I’m happy now! 🎉 

(maybe go all-in the array format?)

oliverlloyd:
> I’m happy now! 🎉
> 
> (maybe go all-in the array format?)

Oh thanks! I missed those when moving to array

PR #2259, opened 2020-12-17
Prettier updates

oliverlloyd:
## What does this change?
This PR makes it so that you can't push to DCR until your code passes `prettier` validation.

It also runs prettier from the CLI to fix all the existing non prettiness.

## Why does this happen?
Because 1) not everyone is a VSCode user and 2) even if you are, the checks can often get stuck and stop formatting on save. Rather than punish people here, hopefully the checks show them useful messages to help them progress. In addition, I've added a `yarn prettier:fix` command in yarn and updated `make fix` as well to make dealing with these warning easier and also updated the README to help highlight what to do.

## Why do we care?
The problem with allowing code on `main` that isn't pretty is that these changes will get picked up in unrelated PRs, cluttering things and making it hard to review.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,13 @@
+# Exclude everything by default
+/**/*
+
+# Include all folders
+!/**/*/
+
+# But exclude node_modules folders anywhere
+node_modules/
+
+# Include specific file extensions if they appear under a "src" folder
+!/src/**/*.ts
+!/src/**/*.tsx
+!/src/**/*.js
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,6 @@
+# See: https://github.com/prettier/prettier/issues/6280#issuecomment-510293569
+# This config works alongside `prettier . --check`
+
 # Exclude everything by default
 /**/*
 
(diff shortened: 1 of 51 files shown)


shtukas (approved):

oliverlloyd:
[Must be pretty](https://trello.com/c/0ZWpSIAr/2234-must-be-pretty)

liywjl (approved):
I like pretty things 👍

.prettierignore
```diff
@@ -0,0 +1,13 @@
...
+# Include all folders
+!/**/*/
+
+# But exclude node_modules folders anywhere
+node_modules/
+
+# Include specific file extensions if they appear under a "src" folder
+!/src/**/*.ts
+!/src/**/*.tsx
+!/src/**/*.js
```
gtrufitt, line 13: When you include all folders, do you not get these included by default?

gtrufitt (approved):

.prettierignore
oliverlloyd: it's a little counter intuitive for sure but no, you need _both_ statements (include all folders and don't exclude these files) to get the result we want 🤷

oliverlloyd: I've added a comment to `.prettierignore` with a link to a Github discussion talking about how this file works

PR #2262, opened 2020-12-17
A star flag can mean a stand alone divider

oliverlloyd:
## What does this change?
Adds support for stand alone dividers in the page based on `* * *`

### Before
We only inserted a divider if the subsequent element was a text element (which we make a drop cap)

### After
We now allow a `* * *` flag to exist by itself and insert a divider, regardless of whether the subsequent element is text or not


shtukas (approved):

liywjl (approved):
👍

PR #2263, opened 2020-12-17
[CMP] add platform to pubData

mxdvl:
## What does this change?

Add `platform` to `pubData`, so we can pinpoint which consent signals are missing a `pageView`.

## Why?

More self-sufficient data.

### DIFF
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -261,6 +261,7 @@ export const App = ({ CAPI, NAV }: Props) => {
         // handled in here.
         if (CAPI.config.switches.consentManagement && countryCode) {
             const pubData = {
+                platform: 'next-gen',
                 browserId: getCookie('bwid') || undefined,
                 pageViewId,
             };


liywjl (approved):
LGTM

src/web/components/App.tsx
```diff
@@ -261,6 +261,7 @@ export const App = ({ CAPI, NAV }: Props) => {
         // handled in here.
         if (CAPI.config.switches.consentManagement && countryCode) {
             const pubData = {
+                platform: 'next-gen',
```
oliverlloyd, line 264: Just a nitpick. Next gen sounds way cooler but would 'dcr' be closer to convention? Of course, if there's another constraint elsewhere to make this the right choice then that's fine.

oliverlloyd (approved):
👍

src/web/components/App.tsx
gtrufitt: `next-gen` is Frontend's codename (just for context if that was missing).

oliverlloyd: Fairs. It's very cool.

mxdvl: I went with Ophan's platform!

sndrs: it does seem like the standard name is the wrong name 

https://github.com/guardian/frontend/pull/23405/files#diff-a439a71607601877fb7c2907fa1c68efd30127d3b869953461087627665685aeR45

sndrs: it's hard though – even `dcr` would be too generic I think (what do you call whatever replaces dcr eventually? it might still live as a rendering service just totally different code)

sndrs: `web-rendering@1`

mxdvl: Currently, the aim of this is to be able to join to the existing `pageview` table, so I’ll stick with the [existing platforms][]. It looks like the difference between `frontend` and `dotcom-rendering` is called an [experience][].

[existing platforms]: https://github.com/guardian/ophan-data-lake/blob/main/ophan_event.avsc#L128
[experience]: https://github.com/guardian/ophan-data-lake/blob/main/etl/pageview/src/test/scala/etl/ophanstream/dev/PageViewTest.scala#L509

PR #2264, opened 2020-12-17
Lazy Load Sentry

oliverlloyd:
## What?
Delays loading the Sentry bundle until an error occurs.

Instead of loading the 20Kb Sentry bundle at the initial page load, here we make use of Sentry's [lazy loading solution](https://docs.sentry.io/platforms/javascript/install/lazy-load-sentry/). When the page first loads we immediately download a [small 1Kb script](https://github.com/getsentry/sentry-javascript/blob/master/packages/browser/src/loader.js) whose job is to capture errors, queue them and then trigger the download of the main bundle. If there are no errors on the page then this script does nothing and the main Sentry code is never downloaded.

## Why?
We reduce page size by nearly 20Kb 🎉 

## Why not?
The downside here is we don't capture all the meta data for those errors that happen prior to the Sentry bundle being downloaded.

## Tests
I put a forced error at the bottom of the page so it only fires when you scroll down. I then recorded doing a refresh and scroll

### With adblocker installed
withadblock

### With adblocker installed, but Sentry whitelisted
withabblock-whitelisted


src/web/server/htmlTemplate.ts
```diff
@@ -29,7 +29,7 @@ export const htmlTemplate = ({
     html: string;
     fontFiles?: string[];
     windowGuardian: string;
-    gaPath: { modern: string, legacy: string };
```
oliverlloyd, line 32: Prettier / Eslint did this

oliverlloyd:
I have a couple of things I'm unsure about with this PR:

1. Should these blobs of text in htmlTemplate.ts be extracted somehow so they can be linted or is there value in keeping them inline?
2. Is window.guardian.adBlockEnabled okay?

and something I noticed along the way is that, now that we load the lib this way,  adblockers will block the call to download the Sentry lib so perhaps the check we make to block posting data to Sentry when if there is an adblocker is moot. I've kept it in though.

rcrphillips:
[Lazy load Sentry](https://trello.com/c/GMAsM9a5/760-lazy-load-sentry)

liywjl (approved):
👍  great work. Have you tested this on CODE?

oliverlloyd:
Closed in favour of #2289

PR #2266, opened 2020-12-18
Add "Gift options" link to main navigation

GHaberis:
Trello: https://trello.com/c/2PwBrI7H/3524-gifting-test-add-a-link-to-subscriptions-from-the-global-masthead

## What does this change?

_**Note:** The vast majority of files changed here are test fixtures. I had to update them to include the new `gifting` property that's been added to the `ReaderRevenueCategories` type in this PR, you can therefore skip over any `fixtures/*` changes._

This changes the main navigation for logged in subscribers and recurring contributors, so they can see a link to "Gift options", which link will take them through to the support.theguardian.com site. We render the link in the client once we've verified whether the user is a subscriber/contributor. The "Gift options" link will only appear from the desktop breakpoint up. 

### Additional changes

Like other Reader Revenue links the "Gift options" link needs it's own set of tracking properties so we can measure user interactions with this element. The code that generates Reader Revenue URLs with the correct tracking properties exists on `frontend`, the constructed URLs are passed in the data model to `dotcom-rendering`. I've made the changes on `frontend` to pass the `gifting` property in the data model in this https://github.com/guardian/frontend/pull/23407. In this PR I had to update the `ReaderRevenueCategories` type to include this new `gifting` property.

### desktop breakpoint (980px)
gifting-dcr-desktop-980

### leftCol breakpoint (1140px)
gifting-dcr-leftCol-1140

### wide breakpoint (1300px)
gifting-dcr-wide-1300

### Tested

- [x] Locally
- [x] On CODE (optional)


src/web/components/App.tsx
```diff
@@ -166,6 +168,14 @@ export const App = ({ CAPI, NAV }: Props) => {
...
 
+    useEffect(() => {
+        setIsRecurringContributor(getCookie('gu_recurring_contributor') === 'true');
+    }, []);
+
+    useEffect(() => {
+        setIsDigitalSubscriber(getCookie('gu_digital_subscriber') === 'true');
+    }, []);
+
     useEffect(() => {
```
gtrufitt, line 179: We're in the process of moving state down from App to child components due to the architecture of partial hydration (essentially setting state causes re-hydration, which is bad...). Could you please move the setState and useEffect into `Links`?

src/static/icons/gifting.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="16" height="18" viewBox="0 0 16 18" fill="none" xmlns="http://www.w3.org/2000/svg">
```
gtrufitt, line 1: I imagine it doesn't exist, but ideally icons would come from Source.

gtrufitt (commented):
Couple of comments from platform perspective otherwise LGTM. I noticed we don't have a story that shows these links - any chance you could add one for chromatic snapshots?

oliverlloyd (approved):
This PR underscores the need for a more automated way to define fixtures

This is great work. +1 for the considerate leftCol breakpoint font size change

👍

src/static/icons/gifting.svg
GHaberis: Yeah that'd be ideal, unfortunately it doesn't exist yet and is a bit of a last minute addition.

src/web/components/App.tsx
GHaberis: OK I've made the change as suggested

gtrufitt: Fab, thanks!

GHaberis:
> Couple of comments from platform perspective otherwise LGTM. I noticed we don't have a story that shows these links - any chance you could add one for chromatic snapshots?

Happy to do this, but can I do it as a follow-up PR? The stakeholders are keen for this link to be out today.

gtrufitt:
> > Couple of comments from platform perspective otherwise LGTM. I noticed we don't have a story that shows these links - any chance you could add one for chromatic snapshots?
> 
> Happy to do this, but can I do it as a follow-up PR? The stakeholders are keen for this link to be out today.

Aye!

src/web/components/ReaderRevenueLinks.tsx
```diff
@@ -43,7 +43,7 @@ const paddingStyles = css`
     }
 `;
 
-const messageStyles = css`
+const messageStyles = (isThankYouMessage: boolean) => css`
```
GHaberis, line 46: @oliverlloyd the change to `messageStyles` was only warranted when showing the "Thank You..." message, I've therefore adjusted this so we use the original `headline.medium` if `isThankYouMessage` is `false`.

oliverlloyd: To be honest, I quite liked the smaller version for everything - I think in DCR we have a larger font in general here. But there's an argument to say it's touchable so I think your approach is best

GHaberis:
@gtrufitt @oliverlloyd what actions should I take regarding https://www.chromatic.com/build?appId=5dfcbf3012392c0020e7140b&number=4121. I've taken a look and the diffs don't appear to be related to this PR.

oliverlloyd:
> @gtrufitt @oliverlloyd what actions should I take regarding https://www.chromatic.com/build?appId=5dfcbf3012392c0020e7140b&number=4121. I've taken a look and the diffs don't appear to be related to this PR.

I've approved those (unrelated) diffs. You should be good to go

paperboyo:
Yay, presents! (although icon should highlight on hover together with the text)

Marry Xmas, all!

PR #2269, opened 2020-12-18
Add monkey patch for performance method that aren't avaliable

gtrufitt:
## What does this change?

Monkey patches the `performance.mark` and `performance.measure` methods in browsers where it is unsupported.

```
  // Noop monkey patch perf.mark and perf.measure if not supported
if(window.performance.mark === undefined) {
    window.performance.mark = function(){};
    window.performance.measure = function(){};
}
```

### Before

Application fails miserably

image

### After

Application runs

image

### DIFF
--- a/src/web/server/htmlTemplate.ts
+++ b/src/web/server/htmlTemplate.ts
@@ -156,7 +156,7 @@ export const htmlTemplate = ({
 
                 <script>
                     // Noop monkey patch perf.mark and perf.measure if not supported
-                    if(window.performance.mark === undefined) {
+                    if(window.performance !== undefined && window.performance.mark === undefined) {
                         window.performance.mark = function(){};
                         window.performance.measure = function(){};
                     }
--- a/src/web/server/htmlTemplate.ts
+++ b/src/web/server/htmlTemplate.ts
@@ -155,6 +155,14 @@ export const htmlTemplate = ({
                 </script>
 
                 <script>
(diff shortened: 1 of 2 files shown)


shtukas (approved):

src/web/server/htmlTemplate.ts
```diff
@@ -154,6 +154,14 @@ export const htmlTemplate = ({
                     window.guardian.gaPath = "${gaPath.legacy}";
                 </script>
 
+                <script>
+                    // Noop monkey patch perf.mark and perf.measure if not supported
+                    if(window.performance.mark === undefined) {
```
oliverlloyd, line 159: What happens here if `window.performance` is undefined?

liywjl (approved):

oliverlloyd (approved):
Shame this isn't polyfilled.

Just a drive by comment

rhystmills:
This looks great.

Tangentially related - I wonder if we could use the new [`libs/logger`](https://github.com/guardian/libs/blob/main/src/logger.ts) to make the logs team-specific. Maybe one for a future PR.

src/web/server/htmlTemplate.ts
shtukas: https://github.com/guardian/dotcom-rendering/pull/2269/commits/5adeb76410ca78179756ce68fa20f96de614d57c

PR #2270, opened 2020-12-18
Update atoms-rendering to v4.1.0

liywjl:
## What does this change?
Updates atoms-rendering to  v4.1.0 and makes the nessesary changes

## Why?
This will pave the way to being able to support Quiz Atoms

## Atoms Rendering Versions:
v4.1.0 https://github.com/guardian/atoms-rendering/releases/tag/untagged-8dbf29db9fe142660a4d
v4.0.1 https://github.com/guardian/atoms-rendering/releases/tag/v4.0.1
v4.0.0 https://github.com/guardian/atoms-rendering/releases/tag/untagged-6ee45135a8ab8e49bcf2
v3.0.1 https://github.com/guardian/atoms-rendering/releases/tag/v3.0.1
v3.0.0 https://github.com/guardian/atoms-rendering/releases/tag/v3.0.0

## Changes:
- remove optional chaining in google analytics:
Optional chaining is not supported in the version of node we are running on CI. So for unit tests it is failing, and the `transformIgnorePatterns` that we would use in Jest to compile to previous versions doesnt solve our problem (other errors appear instead). It just better we avoid optional chaining and other more modern JS features until we bump Node on CI

**Youtube**
- Update youtube atom to use `youtube-player` (nicer way of interating with youtube API)
- Pass down `altText` to youtube atom (in case we have an image overlay its advisable we add alt to any image)


oliverlloyd:
Ah, optional chaining in Jest...

Also, it'd be great to see what changes are in this version, can we add a link to the release page?

shtukas (approved):

src/web/browser/ga/ga.ts
```diff
@@ -212,3 +212,24 @@ export const trackPerformance = (
         );
     }
 };
+
+export const videoTracking = ({
```
oliverlloyd, line 216: Can re rename this to something more imperative? A name that is more of a verb really helps my brain figure out what thins actually _do_

oliverlloyd: Ideally, the naming would also follow the convention already established in the file. Maybe `trackVideos`? Or `trackVideoInteraction`?

src/web/components/elements/YoutubeBlockComponent.tsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import { css } from 'emotion';
 
+import { videoTracking as gaRecord } from '@root/src/web/browser/ga/ga';
+import { record as ophanRecord } from '@root/src/web/browser/ophan/ophan';
```
oliverlloyd, line 5: This is a take it or leave it because I don't have the fill picture but do we need to rename these? I think using `as` can be useful in certain situations (which this might be) but is best avoided because it breaks continuity and can hide results when searching

src/web/components/elements/YoutubeBlockComponent.tsx
```diff
@@ -125,17 +127,42 @@ export const YoutubeBlockComponent = ({
         );
     }
 
+    const ophanTracking = (trackingEvent: string) => {
```
oliverlloyd, line 130: As before, take it or leave it based on context, but is there a more imperative name here to represent what the function _does_?

oliverlloyd (commented):
This looks solid.

Just a couple of points around naming.

src/web/components/elements/YoutubeBlockComponent.tsx
```diff
@@ -129,13 +129,21 @@ export const YoutubeBlockComponent = ({
...
+                          }
+                        : undefined
+                }
                 adTargeting={adTargeting}
                 height={height}
                 width={width}
                 title={title}
                 duration={duration}
                 origin={origin}
+                eventEmitters={[]}
```
liywjl, line 146: event emitter are empty until we resolve hydration issue

oliverlloyd (approved):
👍

PR #2275, opened 2020-12-21
Added CodeBlockComponent

oliverlloyd:
## What?
Adds the `CodeBlockElement` to display code snippets inside articles.

Support for code snippets has been fairly basic up to now on Frontend, we set a background colour, the font, some spacing and that's about it. In this PR I wanted to update these styles and also add syntax highlighting.

## Why?
More readability, better accessibility and so that our articles in the [Guardian Digital Blog](https://www.theguardian.com/info/series/engineering-blog) look nicer

### Before
CodeBlockElement - Before

### After
CodeBlockElement - After

## Library choice
Given code formatting is a specialist field with mature solutions already available, I decided to use a library. There are three main options available:

[highlight.js](https://highlightjs.org/)
[Google prettier](https://github.com/googlearchive/code-prettify)
and
[Prism](https://prismjs.com/)

#### Highlight.js
_Pros:_
- Used by Stackoverflow
- Feature rich
- Auto-detection
- Supports SSR

_Cons:_
- [Large bundle size](https://github.com/highlightjs/highlight.js/issues/2846)
- Typescript is not a 1st class citizen

#### Google
I list this here because I see references to this in Frontend and it's well known but it is now deprecated.

#### Prism
_Pros:_
- Lightweight
- JS/TS focussed
- Supports SSR

_Cons:_
- You have to specify the language

### Decision
Use Prism. Bundle size is critical for us and the result even when using the wrong language is still acceptable.


package.json
```diff
@@ -65,6 +65,7 @@
         "@typescript-eslint/eslint-plugin-tslint": "^4.1.1",
         "ajv": "^6.12.6",
         "aws-sdk": "^2.804.0",
+        "babel-plugin-prismjs": "^2.0.1",
```
liywjl, line 68: shouldnt this be in `devDependencies`?

shtukas (approved):

liywjl (approved):
LGTM, though unsure about babel being in `dependencies` rather than `devDependenciesz`

oliverlloyd:
[CodeBlockElement](https://trello.com/c/BXlktyFL/1767-codeblockelement)

package.json
oliverlloyd: Great spot! Fixed 👍

babel.config.js
```diff
@@ -24,6 +24,10 @@ module.exports = {
         ],
         'babel-plugin-px-to-rem',
         '@babel/plugin-transform-runtime',
+        ["prismjs", {
+            // This list should match those defined in typescript type Language
+            "languages": ["typescript", "javascript", "css", "markup", "scala", "elm"]
```
jfsoul, line 29: For the record, here's the list currently supported by composer.

image

oliverlloyd: We should probably add `typescript` to that list. Is that easy to do?

jfsoul: Not sure, first implementation in composer is here: https://github.com/guardian/flexible-content/pull/459. That makes it look like it's using [pretiffy](https://github.com/googlearchive/code-prettify) which may or may not support TS. To be fair, it might not matter that much.

oliverlloyd: Prettify is [deprecated](https://github.com/googlearchive/code-prettify) and I'm not sure if the implementation we have with it on Frontend is working anymore - the code blocks I see there don't seem to have much styling other than a basic pass.

PR #2276, opened 2020-12-21
Added ClickToView component

oliverlloyd:
## What?
This PR makes a proposal for a generic `ClickToView` interstitial component so that multiple teams can uniformly overlay content where we want to warn about tracking status. The copy and designs here are arbitrary, ideally this component will be hoisted to Source and fully implemented there.

```typescript
type Props = {
    children: React.ReactNode;
    width: number;
    height: number;
};

```

Screenshot 2020-12-21 at 10 42 31

 ## Still pending

- [ ] Add copy
- [x] Do we need to change size at other widths?
- [x] Is `width` and `height` always going to be known, what if not?
- [ ] Should we encapsulate this logic in Figure?
- [ ] Do we need more buttons? Such as Reject or Accept all?
- [ ] Should we encapsulate any additional actions in a hamburger menu?


src/web/components/ClickToView.stories.tsx
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable jsx-a11y/aria-role */
+import React from 'react';
+// import { css } from 'emotion';
```
liywjl, line 3: remove comment

JamieB-gu:
This looks like a sensible solution to me 👍.

Do you think it's possible that the overlay copy will vary by embed type, rather than being generic for all embeds? 🤔

oliverlloyd:
> Do you think it's possible that the overlay copy will vary by embed type, rather than being generic for all embeds?

Great question. My guess would be no, but I don't think any decisions have been taken here so it's certainly possible.

paperboyo:
Now here’s an absurd and almost certainly unrelated thought: maybe one day this (or some of it) may replace a custom element that provides [“censoring” an image](https://interactive.guim.co.uk/tools/embed-tool/censored-image/index.html)? Because custom elements should die a painful and sudden death. And, maybe, some day later, Grid could signal via Composer and CAPI which images should be hidden too?

oliverlloyd:
[Add context to ClickToView](https://trello.com/c/QMhItPII/2287-add-context-to-clicktoview)

frj (approved):

PR #2277, opened 2020-12-21
Remove inline fill/stroke colour from gifting svg

GHaberis:
## What does this change?

The gift icon svg included the fill/stroke colours inline, which meant the CSS overrides on `:hover` weren't applied, so the icon remained white on hover state. Removing these inline attributes fixes this issue so on hover state the icon turns yellow.

I'll need to do the same to the SVG on DCR.

## Screenshots after fix
gifting_icon


oliverlloyd (approved):

oliverlloyd:
Be good to get this svg into Source. Might be worth opening an issue over on that repo so we don't lose this?

GHaberis:
> Be good to get this svg into Source. Might be worth opening an issue over on that repo so we don't lose this?

Agreed, I'll do that now 👍

mxdvl:
These icons are lovely! I’m convinced they could look a bit sharper, especially on lower DPI devices, by snapping their geometries to pixels.

I also think `viewBox` should be used, and `width`/`height` avoided. May I raise a PR to make this uniform across `./src/static/icons`?

PR #2281, opened 2020-12-21
Callout file error fix

liywjl:
## What does this change?
Stop error from throwing on callout file change

## Why?
On file change, we would always throw an error. This was added for debugging reasons and forgotten about. It prevented the file from being sent to the server


src/web/components/Callout/FileUpload.tsx
```diff
@@ -38,7 +39,9 @@ export const FileUpload = ({ formField, formData, setFormData }: Props) => {
                     [formField.id]: stringifiedFile,
                 });
             } catch (e) {
-                setError(e);
+                setError(
```
liywjl, line 42: unable to use `e` to set the error string as `e` represents the stacktrack

shtukas (approved):

PR #2282, opened 2020-12-21
Hydrate Once

oliverlloyd:
## What does this change?
Adds logic to the `Hydrate` component to ensure that it only executes the `hydrate()` call once, and only after all dependencies are defined

## Why?
We never want to _re_ hydrate a component

## Why not use the useOnce function?
Because that contains a `useEffect` to ensure the given task only runs after rendering has completed but this is not the desired behaviour here.

## What else?
I also made a drive-by change to fix the names used to log out performance stats for lazy loaded components. I made this change here in this PR becuase the addition of these extra hydration perf logs really higlighted the fact that these calls were using the wrong naming convention


shtukas (approved):

src/web/components/Hydrate.tsx
```diff
@@ -1,22 +1,31 @@
...
+};
+
+const isReady = (dependencies: unknown[]): boolean => {
+    return dependencies.every((dep) => dep !== undefined);
 };
 
-export const Hydrate = ({ root, index, children }: Props) => {
-    const rootWithIndex = index === 0 || index ? `${root}-${index}` : root;
-    const element = document.getElementById(rootWithIndex);
+export const Hydrate = ({ root, children, index, waitFor = [] }: Props) => {
```
gtrufitt, line 17: `HydrateOnce`? Might make it clearer and align with `useOnce`

oliverlloyd: Done

oliverlloyd:
[Only Hydrate once](https://trello.com/c/FO7VpZFc/2229-only-hydrate-once)

liywjl (approved):
LGTM

gtrufitt (approved):

PR #2285, opened 2020-12-22
Youtube analytics

liywjl:
## What does this change?
Updates atoms rendering and adds analytics for youtube atom

## Why?
Parity with frontend


shtukas (approved):

src/web/browser/ga/ga.ts
```diff
@@ -212,3 +212,24 @@ export const trackPerformance = (
         );
     }
 };
+
+export const trackVideosGA = ({
```
oliverlloyd, line 216: Sorry to be picky, but the other functions in this file are called `trackNonClickInteraction` and `trackPerformance` so how about `trackVideoInteraction`?

oliverlloyd (approved):
Just a nitpick around a name

PR #2289, opened 2020-12-23
Lazy Load Sentry

oliverlloyd:
## What?
Delays loading the Sentry bundle until an error occurs.

Although the job that Sentry does seems simple it has to handle a lot of edge cases and deal with cross browser issues causing the bundle size to reach around 20Kb. There have been [improvements in this area](https://github.com/getsentry/sentry-javascript/issues/1552) but even then we should not be loading this much code on every page, especially when we so rarely use it.

Here, instead of loading the full Sentry lib in the document head, we load a custom `sentryLoader` script that listens for errors and, if they happen, only then loads the full script. This custom script has the following features:

### Ad Block detection
Ad blockers will block calls out to Sentry to report errors so - if we know an ad blocker is in use - there's no point trying to make them. This needlessly uses cpu cycles and messes up the console.

### Queueing
The call to download Sentry is asynchronous so we need a mechanism to capture any errors that occur while this download is happening. While waiting for Sentry to load and initialise, errors are written to an array. Once the load is complete the array is cleared and each error sent to Sentry.

### Sampling
Previously we used Sentry's own built in [sampleRate configuration](https://docs.sentry.io/platforms/node/performance/sampling/) option. But if we're lazing loading Sentry then it makes no sense to download 20Kb of code that we will only use 1% of the time. So here we move this sampling logic up into the `sentryLoader` script so that we only download Sentry 1% of the time and even then only when an error occurs.

## Why?
To reduce the page size by nearly 20Kb
Sizes

## Why not?
The downside here is we don't capture all the meta data for those errors that happen prior to the Sentry bundle being downloaded

## Tests
I put a forced error at the bottom of the page so it only fires when you scroll down. I then recorded doing a refresh and scroll

https://user-images.githubusercontent.com/1336821/102981611-20a68580-4501-11eb-8788-c040ee1f24c0.mov


src/web/browser/sentryLoader/sentry.ts
```diff
@@ -0,0 +1,62 @@
+import * as Sentry from '@sentry/browser';
```
oliverlloyd, line 1: This file was broadly unchanged. I removed the `initialiseSentry` function that the code was wrapped in and we no longer do anything with samplig or ad block detection here

shtukas (approved):

liywjl (approved):
Amazing work!!!

src/web/browser/sentryLoader/init.ts
```diff
@@ -0,0 +1,77 @@
...
+    // Sentry lets you configure sampleRate to reduce the volume of events sent
+    // but this filter only happens _after_ the library is loaded. The Guardian
+    // measures page views in the billions so we only want to log 1% of errors that
+    // happen but if we used sampleRate to do this we'd be needlessly downloading
+    // Sentry 99% of the time. So instead we just do some basic math here
+    // and use that to prevent the Sentry script from ever loading.
+    const randomNumber = Math.floor(Math.random() * 100) + 1;
+    if (randomNumber !== 42) return Promise.resolve(); // 42 is arbitrary
+
+    return Promise.resolve().then(() => {
```
jonathonherbert, line 16: Just out of interest – is there a reason to use `Promise.resolve().then` rather than `new Promise` here? Possibly to defer evaluation a tick?

mxdvl: You’d have to change the `return`s to `resolve` If you’re going with `new Promise((resolve, reject)=> {})`

oliverlloyd: I paired on this with @jonathonherbert and we made a change where we simply mark the `init()` as async so there's no need to explicitly return promises anymore.

It took a while before I was able to mangle my mental model of how async / await works into something more aligned with reality but the result is the code is a lot simpler and I now understand `async` a lot better!

src/web/browser/sentryLoader/init.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const init = async (): Promise<void> => {
+    // Sentry lets you configure sampleRate to reduce the volume of events sent
+    // but this filter only happens _after_ the library is loaded. The Guardian
+    // measures page views in the billions so we only want to log 1% of errors that
+    // happen but if we used sampleRate to do this we'd be needlessly downloading
+    // Sentry 99% of the time. So instead we just do some basic math here
+    // and use that to prevent the Sentry script from ever loading.
+    const randomNumber = Math.floor(Math.random() * 100) + 1; // A number between 1 - 100
+    if (randomNumber === 1) return; // 99% of the time we do nothing here
```
mxdvl, line 14: I think you’re doing this the wrong way around.

```suggestion
    if (randomNumber !== 1) return; // 99% of the time we do nothing here
```

mxdvl (changes requested):
You’re sampling 99% of users!

src/web/browser/sentryLoader/init.ts
mxdvl: I find quite clear, too:

```
if (Math.random() < 0.99) return; // 99% of the time we do nothing here.
```

src/web/browser/sentryLoader/sentry.ts
mxdvl: It seems git can find it easier to identify a file rename if you separate the change and move into distinct commits.

https://stackoverflow.com/questions/433111/how-to-make-git-mark-a-deleted-and-a-new-file-as-a-file-move

mxdvl (approved):

src/web/browser/sentryLoader/init.ts
oliverlloyd: Really great spot! Thanks!

jonathonherbert (approved):
✨ ✨ ✨

oliverlloyd:
[Lazy load Sentry](https://trello.com/c/GMAsM9a5/760-lazy-load-sentry)

oliverlloyd:
[Lazy load Sentry](https://trello.com/c/GMAsM9a5/760-lazy-load-sentry)

src/web/browser/sentryLoader/init.ts
```diff
@@ -0,0 +1,75 @@
...
+            // window.onunhandledrejection listeners but we need to
+            // manually redefine our own custom error reporting function
+            window.guardian.modules.sentry.reportError = reportError;
+
+            // Now that we have the real reportError function available,
+            // send any queued errors
+            while (queue.length) {
+                const queuedError = queue.pop();
+                if (queuedError) reportError(queuedError);
+            }
```
gtrufitt, line 58: Feel free to ignore, maybe typescript requires the if statement but I think the `while` on `length` should take care of it. Also, I have no idea if it matters, but maybe we want to send the errors in the order they were received?

```suggestion
	while (queue.length) {
		reportError(queue.shift());
	}
```

gtrufitt (approved):
Take those savings to the bank

src/web/browser/sentryLoader/init.ts
oliverlloyd: 👍 for shift. We need the guard though because, even with the while, TS worries that `.shift()` might be undefined

PR #2290, opened 2020-12-23
Hydrate Once

oliverlloyd:
## What does this change?
Adds logic to the `Hydrate` component to ensure that it only executes the `hydrate()` call once, and only after all dependencies are defined. It has also been renamed to `HydrateOnce`

```typescript
type Props = {
    root: IslandType;
    children: JSX.Element;
    index?: number;
    waitFor?: unknown[]; // new
};
```

## Why?
We never want to _re_ hydrate a component

## What else?
The [last attempt at this change](https://github.com/guardian/dotcom-rendering/pull/2282) caused an issue in production where the reader's signed in state was not being displayed correctly. This was caused because the hydration of `Discussion` and `Links` happened too soon before we knew if the reader had a `GU_U` cookie and before we had fetched their profile. And because we only hydrate once, we didn't update the UI once this information was available.

To solve this problem, this version of this change now includes the use of the `waitFor` param so that `Discussion` and `Links` will only hydrate after the `user` state value is defined (`null | UserProfile`). Furthermore, because _that_ state value depends on `isSignedIn` we use a [useOnce](https://github.com/guardian/dotcom-rendering/pull/2271) function to ensure that the code to fetch the user profile only runs once and only after the `isSignedIn` state value is defined.

I also added some tests to ensure that we always validate that the signed in state is shown correctly.

## Why not use the useOnce function inside HydrateOnce to do the same thing?
Because `useOnce` contains a `useEffect` to ensure the given task only runs after rendering has completed but this is not the desired behaviour for HydrateOnce where there is no DOM render in scope.


liywjl (commented):
If the user signs out, would we still show them as signed in?

oliverlloyd:
> If the user signs out, would we still show them as signed in?

Good question! At the moment, we trigger a refresh of the page on sign out so no, it's not an issue but ideally this would feel more like a single page app where the state is the only thing to refresh

liywjl (approved):
👍

oliverlloyd:
[Only Hydrate once](https://trello.com/c/FO7VpZFc/2229-only-hydrate-once)

gtrufitt:
LGTM 🎊

PR #2295, opened 2020-12-24
useOnce in StickyBottomBanner

oliverlloyd:
## What
Adds `useOnce` to `StickyBottomoBanner` to simplify the code.

This component was a good example of the use case where there was code that we only want to run once but only after a set of dependencies were defined. To achieve this we had to use 2 components and some if statements. By using `useOnce` these can be removed and the code is - hopefully! - a lot easier to reason about.

## Typescript
This PR highlights an issue we have with `useOnce` where typescript is noto aware that thhe dependencies are sure to be defined when the task is run. Instead it thinks they might be undefined and complains.

See: https://codesandbox.io/s/stoic-wilson-xe5uu?file=/src/App.tsx

This issue blocks this PR


oliverlloyd:
[Refactor StickyBottomBanner](https://trello.com/c/OGif4vpQ/2257-refactor-stickybottombanner)

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -149,36 +105,41 @@ export const StickyBottomBanner = ({
...
-        asyncBrazeUuid === undefined ||
-        shouldHideSupportMessaging === undefined
-    ) {
-        return null;
+    useOnce(() => {
+        const CMP = buildCmpBannerConfig();
+        const readerRevenue = buildReaderRevenueBannerConfig(
+            CAPI,
+            isSignedIn || false, // TODO: Remove this
+            asyncCountryCode || Promise.resolve('GB'), // TODO: Remove this
```
gtrufitt, line 121: RE the TODO's - removal before or after this PR?

gtrufitt (commented):
This is a really nice pattern, encases the existing complexities into something that feels familiar and React-y.

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
oliverlloyd: Based on some tests and discussions outisde of this PR, I think these fallbacks might need to remain to keep typescript happy. I wasn't able to find a way to tell TS that these variables are certain to be defined

oliverlloyd: So they should be removed as part of this PR (assuming we can't think of a better solution)

tjmw: As discussed offline would be great to figure out a way of not needing these defaults, but keeping them for the purposes of moving forward with this sounds good.

tjmw (commented):
This is looking really good, nice work!

It's be good to ensure that the contributions service/cmp/braze are still called in the same way after this change (i.e. only ever once, and only when all the dependencies are available). I'd previously added a cypress test to ensure that the contributions banner service was called exactly once (after briefly accidentally breaking it meaning it wasn't ever called) but I think this got removed at some point as it had started failing but it wasn't clear why. I can help with this if needed, probably easiest to do manually on CODE.

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -149,36 +105,41 @@ export const StickyBottomBanner = ({
...
+
+        pickBanner(bannerConfig).then((PickedBanner: () => MaybeFC) =>
+            setSelectedBanner(PickedBanner),
+        );
+    }, [
+        isSignedIn,
+        asyncCountryCode,
+        asyncBrazeUuid,
+        shouldHideSupportMessaging,
+        CAPI,
```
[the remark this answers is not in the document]
oliverlloyd: You're right, it's not strictly needed but I technically we want it to be defined so even though we onviously know it always will be, the convention is to define the array of all deps needed so I think leaving it makes sense.

oliverlloyd:
Marking this as ready even with the typescript issue. I'm not sure there's a good way to fix this, not if we want to preserve the useX pattern. I think if we use casting instead of fallbacks then at least the intentions are preserved and there are no strange fallback paths that look like they're possible (but really aren't).

oliverlloyd:
> It'd be good to ensure that the contributions service/cmp/braze are still called in the same way after this change

Agreed, lets pair on this tomorrow in CODE. Maybe we can even resurect that old test / add a new one?

liywjl (approved):
The code looks good to me 👍

oliverlloyd:
I just paired with @tjmw to check this in CODE and it all looks 👍

PR #2296, opened 2020-12-29
Remove Immersive designType

oliverlloyd:
## What does this change?
In DCR we have a legacy `DesignType` value of `Immersive` but this was made redundant when we introduced the `Display` type of `Immersive`. This PR cleans up from this by removing the unused design type.

## Why?
Because we want to start using the [Design type](https://github.com/guardian/types/blob/main/src/format.ts#L17) from https://github.com/guardian/types/ but there are some preparation steps required before we can do that migration, this is one of them.


oliverlloyd:
[Remove Immersive DesignType](https://trello.com/c/XJQcrPmF/2258-remove-immersive-designtype)

liywjl (approved):
LGTM 👍

PR #2297, opened 2020-12-29
move ticker into epic and add goal reached behaviour

michaelbjacobson:
<h3>What does this PR do?</h3>
Implements simple goal reached behaviour on AMP epic tickers.
<br>
<br>
<h3>Screenshots</h3>
<i>Goal reached example</i>
goal_reached
<br>
<i>Goal not yet reached example</i>
goal_not_reached


liywjl (approved):
LGTM

src/amp/components/Ticker.tsx
```diff
@@ -1,85 +0,0 @@
...
-    ${headline.xsmall({ fontWeight: 'bold' })}
-`;
-const goalAmountStyle = css`
-    ${headline.xxxsmall({ fontWeight: 'bold' })}
-`;
-const amountCaptionStyle = css`
-    ${body.small({ fontStyle: 'italic' })};
-`;
-
-export const Ticker: React.FC<{
```
tomrf1, line 56: is there a reason for moving this into the other file?

michaelbjacobson: It's going to be necessary to do this when we implement the animation. It's not vital at the moment, but I think is probably preferable to have a single component for the epic (including the ticker) as all the data comes from the same HTTP request and populates the same moustache template.

tomrf1 (approved):
💯

PR #2299, opened 2020-12-29
Use @guardian/types for Display

oliverlloyd:
## What does this change?
Replaces our local const enum version of `Display` with [the version](https://github.com/guardian/types/blob/main/src/format.ts#L17) exported from `@guardian/types/Format`

## Why?
This is part of a general move to use @guardian/types more so the definition of our content is better, and properly shared.


src/lib/content.d.ts
```diff
@@ -33,10 +33,7 @@ interface BlockquoteBlockElement {
 
 interface CaptionBlockElement {
     _type: 'model.dotcomrendering.pageElements.CaptionBlockElement';
-    display: Display;
-    designType: DesignType;
     captionText?: string;
-    pillar: Pillar;
```
oliverlloyd, line 39: These were never used and should not have been here

src/lib/format.ts
```diff
@@ -23,63 +17,3 @@ export const toTypesPillar = (p: Pillar): TypesPillar => {
             return TypesPillar.News;
     }
 };
-
-const toTypesDisplay = (d: Display): TypesDisplay => {
```
oliverlloyd, line 27: This function wasn't used and is no longer needed now that we import directly from /types

src/lib/format.ts
```diff
@@ -23,63 +17,3 @@ export const toTypesPillar = (p: Pillar): TypesPillar => {
...
-        case Display.Immersive:
-            return TypesDisplay.Immersive;
-        case Display.Showcase:
-            return TypesDisplay.Showcase;
-        default:
-            return TypesDisplay.Standard;
-    }
-};
-
-const toTypesDesign = (d: DesignType): TypesDesign => {
```
oliverlloyd, line 40: This function wasn't used and is no longer needed now that we import directly from /types

src/lib/format.ts
```diff
@@ -23,63 +17,3 @@ export const toTypesPillar = (p: Pillar): TypesPillar => {
...
-        case 'AdvertisementFeature':
-            return TypesDesign.AdvertisementFeature;
-
-        // note some are unsupported
-        default:
-            return TypesDesign.Article;
-    }
-};
-
-export const asFormat = (
```
oliverlloyd, line 75: This function wasn't used and is no longer needed now that we import directly from /types

src/web/components/AdSlot.tsx
```diff
@@ -200,8 +200,9 @@ export const AdSlot: React.FC<Props> = ({ position, display }) => {
                         </div>
                     );
                 }
+                default:
+                    return null;
```
oliverlloyd, line 204: Undefined is no longer an acceptable return type

liywjl (approved):
👍

oliverlloyd:
[Use Display from @guardian/types](https://trello.com/c/vgPxdgyi/2264-use-display-from-guardian-types)

PR #2300, opened 2020-12-29
Remove SpecialReport type

oliverlloyd:
## What does this change?
Here we remove the `SpecialReport` design type from DCR

## Why?
Instead of having `SpecialReport` as a design type we want to use [Special](https://github.com/guardian/types/blob/main/src/format.ts#L11) as a pillar value from @guardian/types

We're not _using_ `SpecialReport` anywhere in DCR specifically so this initial preparation PR (ahead of importing `Special`) is made as a convenience to better break up this work.


oliverlloyd:
[Remove SpecialReport type from DCR](https://trello.com/c/VFjcmIVN/2261-remove-specialreport-type-from-dcr)

liywjl (approved):
👍

PR #2302, opened 2020-12-30
Update atoms rendering v6

liywjl:
## What does this change?
- Updates atoms rendering to v6
- Changes the props for the youtube atom
- Add storybook for each youtube atom case

## Why?
- Stay up to date with atoms rendering
- The props for youtube atom are not nested anymore, instead they are flatterned to be more explicit
- Update contains some fixes for quiz atom and audio atom accesibility


oliverlloyd (approved):
👍

oliverlloyd:
Code looks good but the stories seem to highlight an issues with vertical margins and also a missing caption?

PR #2306, opened 2020-12-31
fix for comment slot sizing not matching DCP

ashishpuliyel:
## What does this change?
Matches the ad sizes supported by the DCR ad-slot to the sizes supported by the DCP ad-slot.

### Before
On DCR:

```
document.getElementById('dfp-ad--comments').getAttribute('data-desktop')
"1,1|2,2|620,1|620,350|550,310|fluid|300,600|160,600"
```

### After
Now matches DCP:

```
document.getElementById('dfp-ad--comments').getAttribute('data-desktop')
"1,1|2,2|300,250|300,274|620,1|620,350|550,310|fluid|300,600|160,600"
```
## Why?
DCR and DCP should always have been the same

### DIFF
--- a/src/web/components/AdSlot.tsx
+++ b/src/web/components/AdSlot.tsx
@@ -231,6 +231,7 @@ export const AdSlot: React.FC<Props> = ({ position, display }) => {
                         data-mobile={[
                             `${Size.outOfPage}`,
                             `${Size.empty}`,
+                            `${Size.halfPage}`,
                             `${Size.outstreamMobile}`,
                             `${Size.mpu}`,
                             `${Size.googleCard}`,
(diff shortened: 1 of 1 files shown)


oliverlloyd:
I think we delierately made this slot larger on DCR. The problem we have is that in Frontend there is javascript which checks the height of the container and then, based on this value, dynamically refreshes the adslot to be larger. So basically, if there is enough space, a double height slot is used, else a single height slot is placed.

In DCR, because we don't have the same tight coupling to commercial, we can't trigger this dynamic refresh so we either have single always or double always - we went with double because in most cases the container has sufficient space and in thos cases where it doesn't the impact is a  few lines of white space, which we felt was an okay trade off vs. the potential revenue drop from only using a single mpu.

I am going to caveat this comment with the fact I'm assuming that these `sizes` are where we actually did this single/double decision.

shtukas (approved):

liywjl (approved):

ashishpuliyel:
@oliverlloyd Yeah. As I was working on this I connected it with what you've mentioned before in meetings. You're referring to 300,600|160,600 at the end of the array – tall ads. This change is specifically to add "300,250|300,274" to the array – two regular sized ads that were, for some reason, left out.

oliverlloyd (approved):

PR #2307, opened 2020-12-31
Cypress video atom test

liywjl:
## What does this change?

**Adds Cypress test for video atoms**
- is the video visible?
- does the overlay image disapear on click?
- does the video dispatch a play event to 

** Host/origin prop**
Passed down host prop to origin to both Main media and embed elements.
This adds `origin` to the youtube iframe URL and prevents the video from playing on specific URLs.
Make sure its not applied to localhost and to CODE URLs as this would prevent videos from playing on those URLs


cypress/integration/e2e/atom.video.spec.js
```diff
@@ -0,0 +1,99 @@
...
+    describe('Main media youtube video', function () {
+        beforeEach(function () {
+            disableCMP();
+        });
+        it('should render', function() {
+            cy.visit(`/Article?url=${mainMediaVideo}`);
+            cy.get(`#youtube-block-main-media-0`).should('be.visible');
+        })
+
+        it('should hide overlay on click', function() {
```
oliverlloyd, line 19: I think these tests could be better structured with fewer 'tests' but more assertions. See: https://docs.cypress.io/guides/references/best-practices.html#Creating-%E2%80%9Ctiny%E2%80%9D-tests-with-a-single-assertion

Generally, although this isn't a rule, there's a pattern where each test starts with a visit, and then some actions, and then the assertions. Having a test with no visit means it is dependant on previous tests, but we want each test to be modular and stand on it's own as much as possible.

cypress/integration/e2e/atom.video.spec.js
```diff
@@ -0,0 +1,99 @@
...
+            }, (req) => {
+                const url = new URL(req.url)
+                const videoValue = url.searchParams.get('video')
+                if (videoValue) listOfOphanVideoValues.push(JSON.parse(videoValue))
+            })
+
+            cy.visit(`/Article?url=${mainMediaVideo}`);
+            cy.get(`#youtube-block-main-media-0`).should('be.visible');
+
+
```
oliverlloyd, line 43: We can remove this space?

cypress/integration/e2e/atom.video.spec.js
```diff
@@ -0,0 +1,99 @@
...
+            cy.get(`#youtube-overlay-S0CE1n-R3OY`).should('be.visible');
+            cy.get(`#youtube-overlay-S0CE1n-R3OY`).click();
+            cy.get(`#youtube-overlay-S0CE1n-R3OY`).should('not.be.visible');
+
+            // eslint-disable-next-line cypress/no-unnecessary-waiting
+            cy.wait(2000).then(function(){
```
oliverlloyd, line 49: Usually, when you have wait statements, it's a bad idea. The problem is we're often depending on the speed of the server running the test, at that moment in time. If that CI instance happens to be busy then the wait period might not be sufficient and this can cause the test to fail, leading to flakiness.

I think it the thing we want to test is that an Ophan event was sent then we should only assert for that, maybe we can just use intercept statements rather than these types of assertions?

cypress/integration/e2e/atom.video.spec.js
```diff
@@ -0,0 +1,99 @@
...
+                const url = new URL(req.url)
+                const videoValue = url.searchParams.get('video')
+                if (videoValue) listOfOphanVideoValues.push(JSON.parse(videoValue))
+            })
+
+            cy.visit(`/Article?url=${mainMediaVideo}`);
+            cy.get(`#youtube-block-main-media-0`).should('be.visible');
+
+
+            cy.get(`#youtube-overlay-S0CE1n-R3OY`).should('be.visible');
```
oliverlloyd, line 44: The value `S0CE1n-R3OY` looks suspiciously like it will change in the future?

Best practice [is to use](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements) data-* selectors

cypress/integration/e2e/atom.video.spec.js
```diff
@@ -0,0 +1,99 @@
...
+                expect(listOfOphanVideoValues).to.deep.include(mainMediaPlayEvent);
+            })
+
+        })
+    })
+
+    describe('Embed youtube video', function () {
+        it('should render', function() {
+            cy.visit(`/Article?url=${embedMediaVideo}`);
+            cy.get(`#youtube-block-41`).scrollIntoView();
```
oliverlloyd, line 59: Ditto selectors

cypress/integration/e2e/atom.video.spec.js
```diff
@@ -0,0 +1,99 @@
...
+                url: /^http:\/\/ophan\.theguardian\.com\/img\/2/,
+                query: {
+                    viewId: /^\.*/,
+                    video: /^\.*/,
+                }
+            }, (req) => {
+                const url = new URL(req.url)
+                const videoValue = url.searchParams.get('video')
+                if (videoValue) listOfOphanVideoValues.push(JSON.parse(videoValue))
+            })
```
oliverlloyd, line 81: There might be a better way to write this. In other parts of the code we use this pattern:

```
                    cy.intercept('GET', '**/most-read-geo**', (req) => {
                        req.reply((res) => {
                            expect(res.body).to.have.property('heading');
                            expect(res.statusCode).to.be.equal(200);
                        });
                    });
```

oliverlloyd: Note, this example is asserting on the response but think we should be mocking the response and only asserting on the request

src/web/components/MainMedia.tsx
```diff
@@ -61,6 +61,7 @@ function renderElement(
     hideCaption?: boolean,
     adTargeting?: AdTargeting,
     starRating?: number,
+    host?: string,
```
oliverlloyd, line 64: Can we move these live changes out into a separate PR. These looks like actual code changes to how thing function rather than being related to the test.

If they _are_ only needed for the test then we should question if we want to add these props. As a rule it is never great to add props purely for testing. Each prop that gets added comes at a cost and ideally we should limit them as much as possible.

src/web/components/elements/YoutubeBlockComponent.tsx
```diff
@@ -192,7 +192,13 @@ export const YoutubeBlockComponent = ({
...
                 title={mediaTitle}
                 duration={duration}
-                origin={origin}
+                origin={
+                    typeof window !== 'undefined' &&
+                    window.location.hostname !== 'localhost' &&
+                    window.location.hostname !== 'm.code.dev-theguardian.com'
+                        ? origin
+                        : ''
+                }
```
oliverlloyd, line 201: This code looks like it only gets added to make the test work? Ideally, we keep test implementations out of the live code as it adds a cognitive overload that we really don't want. That is, it will be hard for people in the future to know what this code is for and if they need to worry about it when refactoring / adding features.

oliverlloyd (changes requested):
I think we should look at trying to make this test more stable by focussing on asserting on the ophan requests rather than the event array.

There's also some changes to the live code that should be broken out into a separate PR so that we can think about it in isolation. Eg. Do we want/need `origin` / `host` to be a value that gets passed in? I already find it difficult to follow the logic of origin / host and this makes that even harder so maybe there's a simpler solution?

cypress/integration/e2e/atom.video.spec.js
liywjl: good catch, must be an issue with my VSC and prettier being applied on save

liywjl: It looks like prettier ignores cypress code, do we want that?

oliverlloyd: No, I can't write anything without Prettier!

cypress/integration/e2e/atom.video.spec.js
liywjl: a lot of the complexity from these tests arise from `cy.intercept`. The issue with the way I currently set it up was that we catch all ophan events (yet we only want to assert 1). 

A true solution would be to make sure I only intercept the right API call, but I have had difficulty only matching that query. The URL is the same (so we can easily turn that into a string to make it look nicer) yet the queries are always a little different:
- viewId: randomly generated ID (so we match everything)
- video: `encodeURIComponent` of an object. Yet somehow it never matches when I try to match it. Here is what the object looks like:
`{"id":"gu-video-youtube-2b33a7b7-e639-4232-9ecd-0fb920fa8147","eventType":"video:content:play"}`

src/web/lib/ArticleRenderer.tsx
```diff
@@ -462,7 +463,6 @@ export const ArticleRenderer: React.FC<{
 								duration={element.duration}
 								mediaTitle={element.mediaTitle}
 								altText={element.altText}
-								origin={host}
```
liywjl, line 465: removing origin for now, will add it in a future PR

src/web/components/App.tsx
```diff
@@ -357,7 +358,6 @@ export const App = ({ CAPI, NAV }: Props) => {
 						duration={youtubeBlock.duration}
 						mediaTitle={youtubeBlock.mediaTitle}
 						altText={youtubeBlock.altText}
-						origin={CAPI.config.host}
```
liywjl, line 360: removing origin for now, will add it in a future PR

src/web/components/Figure.tsx
```diff
@@ -8,6 +8,7 @@ type Props = {
 	role?: RoleType;
 	children: JSXElements;
 	id?: string;
+	dataCy?: string;
```
oliverlloyd, line 11: I'm hesitant to add props for test purposes. I find that there's a finite number of pops that you can have before code gets hard to reason about and brittle and because of this adding new ones comes with a cost and I'm not sure we want to make that spend just for Cypress?

oliverlloyd: I think if we need an element to have a data attribute then we should just set it manually. Shall we just add this `data-cy` attribute to the component directly?

oliverlloyd (changes requested):

oliverlloyd (approved):

PR #2310, opened 2021-01-04
Pillar > CAPIPillar

oliverlloyd:
## What?
Renames the type `Pillar` to `CAPIPillar`

**From hereon, `Pillar` refers to the a const enum version and CAPIPillar represents strings.**

## Why?
We are already importing the const enum version of `Pillar` in the codebase and plan to move everything over to this soon. But because we have two version of Pillar it creates the possibility for auto import errors and also is just generally confusing.

The concept of a string pillar type won't go away. CAPI us unlikely to move over to the types lib version so by explicitly defining this string version as `CAPIPillar` it's a lot clearer what the intentions are.

### DIFF
--- a/cypress/integration/e2e/article.interactivity.spec.js
+++ b/cypress/integration/e2e/article.interactivity.spec.js
@@ -65,7 +65,7 @@ describe('Interactivity', function () {
         });
     });
 
-    describe('Navigating the Pillar menu', function () {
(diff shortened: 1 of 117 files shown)


oliverlloyd:
[CAPIPillar](https://trello.com/c/cfzANxAU/2275-capipillar)

liywjl (approved):
LGTM

PR #2311, opened 2021-01-04
Refactor how AMP handles pillar

oliverlloyd:
## What?
This PR refactors the `pillar` prop in the AMP stack to:

Replace `getPillar` with `decidePillar`
AMP had its own function here but it was doing exactly the same thing

Always use the result of `decidePillar`
AMP doesn't use explicit props and often passes the whole data object down the stack. Because of this it was possible for the raw, unmodified pillar value to be read from the data object. This PR removes these cases and adds an explicit `pillar: Pillar;` prop instead.

## Why?
By centralising and simplifying this logic it makes future refactors (specifically moving towards const enums) easier

In addition, by removing direct access to the data object to get pillar we prevent incorrect use of unmodified pillars.

### DIFF
--- a/cypress/integration/e2e/article.interactivity.spec.js
+++ b/cypress/integration/e2e/article.interactivity.spec.js
@@ -65,7 +65,7 @@ describe('Interactivity', function () {
         });
     });
 
-    describe('Navigating the Pillar menu', function () {
+    describe('Navigating the pillar menu', function () {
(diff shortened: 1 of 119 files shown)


oliverlloyd:
[Refactor AMP Pillar](https://trello.com/c/j69dKl1r/2277-refactor-amp-pillar)

shtukas (approved):

PR #2312, opened 2021-01-04
EOY header test: new variant

tomrf1:
We've been testing a variant with different heading + subheading.
Now we want another variant with different heading but the same subheading as the control.

The variant names here aren't great, but changing the original variant name will mess up the analysis at this point.

Screen Shot 2021-01-04 at 11 40 45


liywjl (approved):

michaelbjacobson (approved):
✅

PR #2315, opened 2021-01-04
Peachy

oliverlloyd:
## What?
Here we propose extending the comment background colour to more of the page.

Screenshot 2021-01-04 at 15 33 41

## Why?
Because why not? This type of change is easy in DCR and has zero cost in terms of css complexity.

There's an argument that the background implies some kind of 'opinion' quality to the content and so we should limit it purely to the comment piece itself, and that applying it to below the fold content like Most Popular changes it. But we already change the background colour of Most Viewed (in the right column) and in fact the context of Most Popular _does_ change for opinion pieces.

My feeling is that it is desirable to have a consistent background colour and that sudden switches on the same page are jarring.

Another reason to move towards more consistency is to help with future work like new layout designs or for dark mode. If we already start to think about page level colours this will help a lot with these things.

## Why not Onwards?
The cards in onwards use their background colour to distinguish their borders so using the same colour for the container background will wash out this distinction. If we did want to fully set the entire page background then we could perhaps use a special border colour for comment cards to prevent this wash out?

### DIFF
--- a/src/web/layouts/CommentLayout.tsx
+++ b/src/web/layouts/CommentLayout.tsx
@@ -562,7 +562,10 @@ export const CommentLayout = ({
 					)}
 
 					{showComments && (
-						<Section sectionId="comments">
+						<Section
+							sectionId="comments"
+							backgroundColour={opinion[800]}
+						>
 							<Discussion
 								discussionApiUrl={CAPI.config.discussionApiUrl}
 								shortUrlId={CAPI.config.shortUrlId}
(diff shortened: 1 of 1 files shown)


liywjl (approved):
¿Por Qué No?

oliverlloyd:
Closing after discussions with @HarryFischer. We don't want to apply this background colour to Most Popular because that risks weighting the non comment content the wrong way and if we can't use it for onwards then that would only leave Comments. Comments in isolation could theoretically have the background colour changed but in the context of an article it results in something of a stripy result which isn't ideal.

PR #2318, opened 2021-01-05
Remove GuardianLabs

oliverlloyd:
## What?
Removes the `GuardianLabs` designType.

If Frontend _does_ send us an article with this design type we now replace it with `AdvertisementFeature`

## Why?
It isn't used. [Searching the Frontend codebase](https://github.com/guardian/frontend/search?q=GuardianLabs&unscoped_q=GuardianLabs) only provides one reference, the one where it is passed into DCR.

We don't branch anywhere with it in DCR and paid content is already represented using `AdvertisementFeature`

### DIFF
--- a/docs/patterns/switch-on-display-design.md
+++ b/docs/patterns/switch-on-display-design.md
@@ -25,7 +25,6 @@ const shouldShowAvatar = (designType: DesignType, display: Display) => {
                 case 'Article':
                 case 'MatchReport':
                 case 'GuardianView':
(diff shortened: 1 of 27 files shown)


shtukas (approved):

liywjl (approved):

oliverlloyd:
[Remove GuardianLabs from DCR](https://trello.com/c/SgyKo7et/2279-remove-guardianlabs-from-dcr)

PR #2322, opened 2021-01-05
Upgrade atoms-rendering to v7.1

oliverlloyd:
## What?
Bump @guardian/atoms-rendering to v7

As par to this upgrade we convert the `pillar` value from a string to a const enum

## Why?
To enable the latest features


shtukas (approved):

liywjl (approved):
👍

PR #2324, opened 2021-01-05
Adds Deeply Read Test to DCR

nitro-marky:
## What does this change?
This sets up an AB test for the Deeply Read hypothesis, which replaces the second tab in the Most Popular container with articles that have achieved the longest read time/engagement.

Control:
Screenshot 2021-01-06 at 11 51 53

Variant:
Screenshot 2021-01-06 at 15 21 39

After a conversation with Brendan when tabs are clicked on an event is sent which should 🤞  be easy enough to query from the datalake. 

When clicking on a tab:

Screenshot 2021-01-06 at 15 17 21

When clicking on an article:
Screenshot 2021-01-06 at 15 17 48

Both contain Deeply Read so should be queryable.

### DIFF
--- a/src/web/components/MostViewed/MostViewedFooter/MostViewedFooter.tsx
+++ b/src/web/components/MostViewed/MostViewedFooter/MostViewedFooter.tsx
@@ -138,7 +138,7 @@ export const MostViewedFooter = ({
 			>
 				<section className={asideWidth}>
 					{inDeeplyReadTestVariant ? (
-						<h2 className={headingStyles}>Accross The Guardian</h2>
+						<h2 className={headingStyles}>Across The Guardian</h2>
 					) : (
 						<h2 className={headingStyles}>Most popular</h2>
 					)}
--- a/src/web/components/MostViewed/MostViewedFooter/MostViewedFooter.tsx
+++ b/src/web/components/MostViewed/MostViewedFooter/MostViewedFooter.tsx
@@ -140,7 +140,7 @@ export const MostViewedFooter = ({
 					{inDeeplyReadTestVariant ? (
 						<h2 className={headingStyles}>Accross The Guardian</h2>
 					) : (
-						<h2 className={headingStyles}>Most Popular</h2>
+						<h2 className={headingStyles}>Most popular</h2>
 					)}
 				</section>
 				<section className={stackBelow('desktop')}>
--- a/src/web/components/MostViewed/MostViewedFooter/MostViewedFooter.tsx
+++ b/src/web/components/MostViewed/MostViewedFooter/MostViewedFooter.tsx
@@ -137,11 +137,10 @@ export const MostViewedFooter = ({
 				data-cy-ab-runnable-test={variantFromRunnable}
 			>
 				<section className={asideWidth}>
-					{!inDeeplyReadTestVariant && (
-						<h2 className={headingStyles}>Most popular</h2>
-					)}
-					{inDeeplyReadTestVariant && (
+					{inDeeplyReadTestVariant ? (
 						<h2 className={headingStyles}>Accross The Guardian</h2>
+					) : (
+						<h2 className={headingStyles}>Most Popular</h2>
 					)}
 				</section>
 				<section className={stackBelow('desktop')}>
--- a/src/web/components/MostViewed/MostViewedFooter/MostViewedFooter.tsx
+++ b/src/web/components/MostViewed/MostViewedFooter/MostViewedFooter.tsx
@@ -119,6 +119,11 @@ export const MostViewedFooter = ({
 	const variantFromRunnable =
 		(runnableTest && runnableTest.variantToRun.id) || 'not-runnable';
 
+	const inDeeplyReadTestVariant = ABTestAPI.isUserInVariant(
+		'DeeplyReadTest',
+		'variant',
+	);
+
 	return (
 		<div
 			data-print-layout="hide"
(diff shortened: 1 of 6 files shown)


src/web/components/MostViewed/MostViewedFooter/MostViewedFooter.tsx
```diff
@@ -132,7 +137,12 @@ export const MostViewedFooter = ({
 				data-cy-ab-runnable-test={variantFromRunnable}
 			>
 				<section className={asideWidth}>
-					<h2 className={headingStyles}>Most popular</h2>
+					{!inDeeplyReadTestVariant && (
+						<h2 className={headingStyles}>Most popular</h2>
+					)}
+					{inDeeplyReadTestVariant && (
+						<h2 className={headingStyles}>Accross The Guardian</h2>
+					)}
```
oliverlloyd, line 145:
```suggestion
					{inDeeplyReadTestVariant ? (
						<h2 className={headingStyles}>Accross The Guardian</h2>
					) : (
						<h2 className={headingStyles}>Most popular</h2>
					)}
```

oliverlloyd: This is perhaps a subjective thing but I think this ternary would be nicer here?

src/web/components/MostViewed/MostViewedFooter/MostViewedFooterData.tsx
```diff
@@ -50,13 +50,26 @@ function buildSectionUrl(
...
 	design,
 }: Props) => {
-	const url = buildSectionUrl(ajaxUrl, pillar, sectionName);
+	const ABTestAPI = useAB();
+
+	const inDeeplyReadTestVariant = ABTestAPI.isUserInVariant(
+		'DeeplyReadTest',
+		'variant',
+	);
+	// const url = buildSectionUrl(ajaxUrl, pillar, sectionName);
```
oliverlloyd, line 69: Comment

src/web/components/MostViewed/MostViewedFooter/MostViewedFooterData.tsx
```diff
@@ -50,13 +50,26 @@ function buildSectionUrl(
...
+	const url = inDeeplyReadTestVariant
+		? buildDeeplyReadUrl(ajaxUrl, pillar)
+		: buildSectionUrl(ajaxUrl, pillar, sectionName);
```
oliverlloyd, line 72: This does read really nice so I'm hesitant if this works or not but ccould we maybe inline the joinUrl call? I guess it's a trade of of the abstraction meaning you have too move around the file to see what happening (plus you don't know if there's anything else using that code) vs. it being immediately clear what happening but not as pretty?

oliverlloyd (approved):
Really nice! I really like how this is super easy to review and understand!

Just a couple of take it or leave them comments

oliverlloyd:
Note. The jest error `useAB must be used within the ABProvider` means there is a missing context provider. I _think_ this is a test issue

src/web/components/MostViewed/MostViewedFooter/MostViewedFooterData.tsx
nitro-marky: I guess my main reasoning for abstracting this way is that building a section URL is a bigger process and when/if it comes to removing the test logic it should be much easier to understand which parts relate to deeply read and which do not. Ultimately they both use joinUrl but I thought the difference between them merited having them split into their own functions.

src/web/components/MostViewed/MostViewedFooter/MostViewedFooter.tsx
nitro-marky: 👍

src/web/components/MostViewed/MostViewedFooter/MostViewedFooterData.tsx
oliverlloyd: > when/if it comes to removing the test logic it should be much easier to understand which parts relate to deeply read and which do not

👍 that's a really sound reason - great point

oliverlloyd:
I noticed we've added capitalisation to `Popular` but I think the intention is to have this as sentance case
Screenshot 2021-01-06 at 10 35 44

nitro-marky:
> I noticed we've added capitalisation to `Popular` but I think the intention is to have this as sentance case
> Screenshot 2021-01-06 at 10 35 44

Do we want 'The' in 'Accross The Guardian to be lower case as well?

oliverlloyd:
> Do we want 'The' in 'Accross The Guardian to be lower case as well?

My guess would be yes, for consistency - might be worth checking with @HarryFischer though

oliverlloyd:
> > Do we want 'The' in 'Accross The Guardian to be lower case as well?
> 
> My guess would be yes, for consistency - might be worth checking with @HarryFischer though

That said. Actually, no, I think The Guardian is a noun so should be capitalised.

nitro-marky:
> > > Do we want 'The' in 'Accross The Guardian to be lower case as well?
> > 
> > 
> > My guess would be yes, for consistency - might be worth checking with @HarryFischer though
> 
> That said. Actually, no, I think The Guardian is a noun so should be capitalised.

Yeah, you're right. I'll leave that then

HarryFischer:
Yep to 'Most popular'
And it should be 'Across The Guardian'

PR #2326, opened 2021-01-06
Refactor decideDesignType

oliverlloyd:
## What?
Refactor decideDesignType to accept a single prop rather than the whole CAPI object. I also refactored the input type to be declared as `CAPIDesign` instead of the (growing) union which was repeated a lot

## Why?
This makes it easier to use this function as we don't need to have a union type for multiple data objects


shtukas (approved):

liywjl (approved):
👍

PR #2328, opened 2021-01-06
Ensure local SDK data is wiped on logout

tjmw:
## What does this change?

Ensures that any local SDK data is wiped when a user logs out.

If we’ve recorded in localStorage that there was a braze user, but we haven’t got a brazeUuid, then we treat that as meaning that the user has logged out.

### DIFF
--- a/src/web/components/StickyBottomBanner/BrazeBanner.test.ts
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.test.ts
@@ -93,37 +93,9 @@ describe('hasRequiredConsents', () => {
 });
 
 describe('canShowPreChecks', () => {
-	describe('when the switch is off', () => {
-		it('returns false', () => {
-			const result = canShowPreChecks({
-				brazeSwitch: false,
-				apiKey: 'abcde',
-				shouldHideSupportMessaging: true,
-				pageConfig: { isPaidContent: false },
-			});
-
-			expect(result).toBe(false);
-		});
-	});
-
-	describe('when the api key is empty', () => {
-		it('returns false', () => {
-			const result = canShowPreChecks({
-				brazeSwitch: true,
-				apiKey: '',
-				shouldHideSupportMessaging: true,
-				pageConfig: { isPaidContent: false },
-			});
-
-			expect(result).toBe(false);
-		});
-	});
-
 	describe('when not a supporter', () => {
 		it('returns false', () => {
 			const result = canShowPreChecks({
-				brazeSwitch: true,
-				apiKey: 'abcde',
 				shouldHideSupportMessaging: false,
 				pageConfig: { isPaidContent: false },
 			});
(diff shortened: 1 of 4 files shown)


src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -249,10 +264,21 @@ export const canShow = async (
 		hasRequiredConsents(),
 	]);
 
+	await maybeWipeUserData(apiKey as string, brazeUuid);
```
liywjl, line 267: personal preference would be to have `as string` defined on the line with `const apiKey`

liywjl (approved):
LGTM, this is more personal opinion but `hasCurrentBrazeUser` was a little confusing to me, maybe something like `hasBrazeUserInLocalStorage` would be better?

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: Ah the issue here is that where it's defined `apiKey` isn't necessarily a string (it might not be set) but we'll return early if it's not truthy. So we'll only get to this line if it's a string, but TypeScript for some reason doesn't get that, hence the casting.

adamnfish, line 267: It looks to me like this line won't be reached if Braze is switched off. I can see arguments both ways on this one, so I just want to check that this is the behaviour we expect here.

adamnfish, line 267: Similarly, looking around this function I can see that we're fetching and using the consent that is determined for Braze, which looks just right. However, I want to check whether there are cases where a lack of consent means none of this code will execute? If no consents are given will the commercial bundle not load any of the Braze logic, or is it separately loaded and fully taking its own responsibility for consent?

src/web/components/StickyBottomBanner/BrazeBanner.test.ts
```diff
@@ -93,37 +93,9 @@ describe('hasRequiredConsents', () => {
 });
 
 describe('canShowPreChecks', () => {
-	describe('when the switch is off', () => {
```
adamnfish, line 96: While the refactor looks correct, it means the logic for checking the switch is now all intertwined with the side effects of the integration. It follows that we no longer have test coverage of these cases. Are we happy with that?

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: Discussed offline but summarising her: yep it's true and I think there are arguments either way here so I'll stick with this for now.

tjmw: My understanding is that this will be loaded either way so I don't think there are cases where none of this code will load because of consent.

src/web/components/StickyBottomBanner/BrazeBanner.test.ts
tjmw: Also discussed offline, but yep I think it's acceptable (the switch/apiKey check is super simple).

PR #2329, opened 2021-01-06
Get identity and mma base URL from config

tjmw:
Previously these URLs were always hardcoded to prod. This made it hard, for example, to sign in/out on DCR in the CODE environment, because the identity URL used was profile.theguardian.com which is a different domain to CODE.

Now the URLs are pulled from config, so should reflect the current environment.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -565,6 +565,7 @@ interface ConfigType extends CommercialConfigType {
     references?: { [key: string]: string }[];
     host?: string;
     idUrl?: string;
+    mmaUrl?: string;
     brazeApiKey?: string;
 }
 
@@ -600,6 +601,7 @@ interface ConfigTypeBrowser {
     switches: CAPIType['config']['switches'];
     host?: string;
     idUrl?: string;
+    mmaUrl?: string;
 }
 
 interface GADataType {
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -317,6 +317,8 @@ export const App = ({ CAPI, NAV }: Props) => {
 				<Links
 					giftingURL={CAPI.nav.readerRevenueLinks.header.gifting}
 					userId={user ? user.userId : undefined}
+					idUrl={CAPI.config.idUrl}
+					mmaUrl={CAPI.config.mmaUrl}
 				/>
 			</HydrateOnce>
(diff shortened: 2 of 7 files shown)


liywjl (approved):
LGTM 👍

twrichards:
this is great, hardcoded prod urls are nasty.

Would they (or at least the mma one) be better being derived from the `Stage` that DCR is running on instead - since they share the same top level domain (i.e. `gulocal.com`, `code.dev-theguardian.com` & `theguardian.com`)? Just to avoid extra config.

You might want to do the same for the `members-data-api` url (if it isn't already / if that is a part of DCR at this point - not sure if it still just relies on the cookies made by `frontend` [as it calls `members-data-api` for various things about the user and drops various cookies]).

QuarpT:
looks good, is there default config for local testing?

tjmw:
Hi @twrichards @QuarpT thanks for taking a look!

> looks good, is there default config for local testing?

Locally, using the dev server, this config will come from whichever instance of frontend you're getting data from. For example http://localhost:3030/Article?url=https://m.code.dev-theguardian.com/sport/2019/jul/28/tour-de-france-key-moments-egan-bernal-yellow-jersey will have code config (you could also point to prod frontend, or a locally running frontend instance which you've configured in a particular way).

> Would they (or at least the mma one) be better being derived from the Stage that DCR is running on instead - since they share the same top level domain (i.e. gulocal.com, code.dev-theguardian.com & theguardian.com)? Just to avoid extra config.

That's a good question. I think I'd lean towards getting these from config like I'm doing since it gives us a single source of truth - this is now coming from the same place that frontend gets its values from (and just to be clear, the config already exists, I've not had to newly add the config itself anywhere). I suppose conceptually I'd also lean towards being explicit rather than deriving, all else being equal. Like maybe there's a valid reason you might want to temporarily point CODE to prod identity for example, which is hard when you're deriving.

> You might want to do the same for the members-data-api url (if it isn't already / if that is a part of DCR at this point - not sure if it still just relies on the cookies made by frontend [as it calls members-data-api for various things about the user and drops various cookies]).

I had a quick look and I don't _think_ dcr makes requests directly to the members data api anywhere. But agreed it would make sense to follow the same pattern if/when that does happen 👍

PR #2331, opened 2021-01-06
Discussion v4.4

oliverlloyd:
## What does this change?
Bumps:

`@guardian/types` to v0.5.2*
`@guardian/discussion-rendering` to v4.4.0

*_this was needed because we use the older version was exporting a subtly different version of Pillar that wasn't compatible_
## Why?
This brings in the version of discussion using Pillar from @guardian/types which is the one we want to start using soon in DCR


shtukas (approved):

liywjl (approved):
LGTM, but CI isnt happy

PR #2332, opened 2021-01-06
Remove labs pillar

oliverlloyd:
## What does this change?
Removes support for the `labs` pillar in favour of using the `AdvertisementFeature` design type

## Why?
Because we only want to have one method to represent paid for content

## Why Not?
Because now we can't ever have paid for content that uses a different design type and the other pillar options won't apply here (you won't ever have paid for `culture` articles I assume)


oliverlloyd:
Closing in favour of [embracing labs as a pillar](https://github.com/guardian/types/pull/42)

PR #2342, opened 2021-01-07
Set max-width of Flex component child to 100%

buck06191:
## What does this change?
Sets the `max-width` of the first child in the `Flex` component to be `100%`. This stops horizontal scroll due to long usernames expanding the screen width when done in conjunction with https://github.com/guardian/discussion-rendering/pull/383

Will need to have the version of discussion-rendering with the accompanying bug fix released and then the version bumped in here

### Before
image

### After
image

## Why?
Poor UX


buck06191:
[Comments not properly fitting on the screen on iPhones](https://trello.com/c/3MZnq43Z/2281-comments-not-properly-fitting-on-the-screen-on-iphones)

liywjl (approved):
👍  great work!

PR #2344, opened 2021-01-08
Youtube atom host prop

liywjl:
## What does this change?
Pass down `host` prop to youtube atom's `origin` prop.

## Why?

https://developers.google.com/youtube/player_parameters

**origin**
This parameter provides an extra security measure for the IFrame API

NOTE: for localhost, origin should be empty


shtukas (approved):

nitro-marky (approved):
👍

src/web/components/elements/YoutubeBlockComponent.tsx
```diff
@@ -194,9 +194,14 @@ export const YoutubeBlockComponent = ({
...
 				duration={duration}
-				origin={origin}
 				eventEmitters={[ophanTracking, gaTracking]}
 				pillar={toTypesPillar(pillar)}
+				origin={
+					typeof window !== 'undefined' &&
+					window.location.hostname !== 'localhost'
+						? origin
+						: ''
+				}
```
oliverlloyd, line 204: Is this going to work on older browsers? Maybe there's another way to determine if the code is running on a dev environment?

Also, would it make more sense to put this logic in the you tube atom, rather than needing each consumer to repeat the same logic?

liywjl: Ahhh yes `process.env.NODE_ENV`, would that be a better case here? we would not need to check if the  `window` object exists anymore.

Yeah I don't see why the logic shouldnt be in atoms 👍

liywjl: https://github.com/guardian/atoms-rendering/pull/180 I have moved this conditional logic over to the Atom

liywjl: This PR has been reverted, instead we will be using `process.env.GU_STAGE` in DCR to make this determination

liywjl: as `process.env.GU_STAGE` is undefined when running locally I will be using `process.env.NODE_ENV === 'development'`

oliverlloyd (approved):

PR #2345, opened 2021-01-08
Atom base type has optiona url prop

liywjl:
## What does this change?
Changes URL key in `InteractiveAtomBlockElementBase` from being mandatory to being optional.

## Why?
Quiz atoms do not return a URL


liywjl:
Quiz types should not extend `InteractiveAtomBlockElementBase`

PR #2346, opened 2021-01-08
Remove Pillar from endpoint

nitro-marky:
## What does this change?

Removes the pillar from the deeply read api endpoint call.

### DIFF
--- a/src/web/components/MostViewed/MostViewedFooter/MostViewedFooterData.tsx
+++ b/src/web/components/MostViewed/MostViewedFooter/MostViewedFooterData.tsx
@@ -73,7 +73,7 @@ export const MostViewedFooterData = ({
 	const { data, error } = useApi<
 		MostViewedFooterPayloadType | TrailTabType[]
 	>(url);
-
+	console.log(url);
 	if (error) {
 		window.guardian.modules.sentry.reportError(error, 'most-viewed-footer');
 		return null;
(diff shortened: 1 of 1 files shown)


src/web/components/MostViewed/MostViewedFooter/MostViewedFooterData.tsx
```diff
@@ -50,8 +50,8 @@ function buildSectionUrl(
 	return joinUrl([ajaxUrl, `${endpoint}?dcr=true`]);
 }
 
-function buildDeeplyReadUrl(ajaxUrl: string, pillar: CAPIPillar) {
-	return joinUrl([ajaxUrl, 'most-read-deeply-read', `${pillar}.json?dcr`]);
+function buildDeeplyReadUrl(ajaxUrl: string) {
+	return joinUrl([ajaxUrl, 'most-read-deeply-read.json']);
```
oliverlloyd, line 54: We're also removing `?dcr` - is that intentional?

nitro-marky: Yeah, that should only be needed on endpoints that are used on frontend and DCR, this is just DCR

oliverlloyd (approved):

liywjl (approved):

PR #2349, opened 2021-01-11
Add ophan tracking calls to the blocked hosts for Cypress

buck06191:
## What does this change?
Blocks calls out to Ophan from inside Cypress tests. 

### Screenshot
image
**Shows a blocked Ophan tracking call and the header from cypress (`x-cypress-matched-blocked-host`)  to indicate that it was blocked for that reason**

## Why?
Causes data quality issues as it throws off the number of positive/negative clicks for atoms.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -4,5 +4,6 @@
     "baseUrl": "http://localhost:3030/",
     "video": false,
     "chromeWebSecurity": false,
+    "blockHosts": "*ophan.theguardian.com",
     "retries": 2
 }


liywjl (approved):
👍 youtube atom ophan test seems to with this change

oliverlloyd (approved):

buck06191:
> 👍 youtube atom ophan test seems to with this change

Works or doesn't work?

liywjl:
> > 👍 youtube atom ophan test seems to with this change
> 
> Works or doesn't work?

it works! ^^ sorry my bad

PR #2352, opened 2021-01-11
Show full navigation menu on top of top-above-nav slot

ioanna0:
## What does this change?

On tablets the navigation menu was hidden behind top-above-nav ad slot. Changes top above nav z-index so that the navigation menu appears on top of it.
Also reducing right margin to make the menu larger and meet the burger button as in DCP

### Before

Screenshot 2021-01-11 at 15 14 49

### After

Screenshot 2021-01-11 at 20 39 03

### DCP for reference:

**Note to dotcom team: Another card might be needed to give the same menu overlay as in DCP** 

Screenshot 2021-01-11 at 20 39 12

## Why?
Will give full visibility of navigation menu on tablets

### DIFF
--- a/src/web/components/Nav/ExpandedMenu/ExpandedMenu.tsx
+++ b/src/web/components/Nav/ExpandedMenu/ExpandedMenu.tsx
@@ -86,7 +86,7 @@ const mainMenuStyles = css`
 		margin-right: 70px;
 	}
 	${from.tablet} {
-		margin-right: 100px;
+		margin-right: 53px;
 	}
 `;
 
--- a/src/web/lib/getZIndex.test.tsx
+++ b/src/web/lib/getZIndex.test.tsx
@@ -1,9 +1,9 @@
 import { getZIndex } from './getZIndex';
 
 describe('getZIndex', () => {
-    it('gets the correct zindex for group and sibling', () => {
+	it('gets the correct zindex for group and sibling', () => {
 		expect(getZIndex('headerWrapper')).toBe('z-index: 9;');
-        expect(getZIndex('stickyAdWrapper')).toBe('z-index: 8;');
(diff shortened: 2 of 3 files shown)


gtrufitt:
I have a feeling this may cause issues on desktop with the positioning of the sticky ad over the nav being what you want on desktop?

ioanna0:
> I have a feeling this may cause issues on desktop with the positioning of the sticky ad over the nav being what you want on desktop?

@gtrufitt I don't see any impact on desktop on top-above nav or on the right hand sticky ad, are you referring to something else?
Screenshot 2021-01-12 at 13 52 02

liywjl (approved):
Looks good, seems to fix the top above Ad issue 👍 

We may need to do some work with the burger button as it doesnt seem to aline correctly, ill add a task for that

gtrufitt:
> @gtrufitt I don't see any impact on desktop on top-above nav or on the right hand sticky ad, are you referring to something else?

This is what I was thinking yes, I'm just surprised as the sticky ad and the header z-indexes are tightly coupled to enable to sticky ad to be above and this seems to invert that behaviour... 

If nothing else, we'll need to review the naming here: https://github.com/guardian/dotcom-rendering/blob/main/src/web/layouts/lib/stickiness.tsx#L47

cc/ @oliverlloyd

ashishpuliyel:
@ioanna0 Looks good to me. I've accepted the chromatic changes which were all false-positives.

ioanna0:
@gtrufitt You were right, there was a use case I haven't check which was when scrolling the z-index previously was higher and thus giving a sticky effect so I have reverted this PR for now https://github.com/guardian/dotcom-rendering/pull/2364

Will need to reconsider a fix for this

PR #2355, opened 2021-01-12
Snapshot Examples

oliverlloyd:
## What does this change?
Add these examples to Chromatic

## Why?
We were disabling them before to reduce the snapshot count but having these stories in a sharable format is useful


liywjl (approved):
👍

PR #2356, opened 2021-01-12
Remove secondary capital letters from deeply read tab

nitro-marky:
## What does this change?
Changes the capitalisation of the second word of the tab to fit in with design standards (note, it is formatted properly in frontend but DCR is capitalising erroneously). 

### Before
Screenshot 2021-01-12 at 16 08 30

### After
Screenshot 2021-01-12 at 16 09 40

## Why?
Design

### DIFF
--- a/src/web/components/MostViewed/MostViewedFooter/MostViewedFooterGrid.tsx
+++ b/src/web/components/MostViewed/MostViewedFooter/MostViewedFooterGrid.tsx
@@ -4,6 +4,7 @@ import { neutral, border } from '@guardian/src-foundations/palette';
 import { headline } from '@guardian/src-foundations/typography';
 import { from, until } from '@guardian/src-foundations/mq';
 import { visuallyHidden } from '@guardian/src-foundations/accessibility';
+import { useAB } from '@guardian/ab-react';
 
 import { pillarPalette } from '@frontend/lib/pillars';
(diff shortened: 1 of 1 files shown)


oliverlloyd:
The diff here seems to contain different changes to the capitalisation - merge glitch?

gtrufitt:
If this is incorrect in the design standards, can we not change it in the data in Frontend?

nitro-marky:
@gtrufitt It is formatted correctly in frontend but DCR is mutating it incorrectly so unfortunately this is a DCR problem

oliverlloyd (approved):
❤️

liywjl (approved):
👍

oliverlloyd:
> @gtrufitt It is formatted correctly in frontend but DCR is mutating it incorrectly so unfortunately this is a DCR problem

I think this is a valid point. Frontend should just send us the correctly capitalised string; this is how capitalisation works _everywhere_ else

nitro-marky:
> > @gtrufitt It is formatted correctly in frontend but DCR is mutating it incorrectly so unfortunately this is a DCR problem
> 
> I think this is a valid point. Frontend should just send us the correctly capitalised string; this is how capitalisation works _everywhere_ else

I agree, unfortunately this is one of those edge cases where regardless of what frontend sends it will be changed here. Perhaps one for a refactor at a later date?

PR #2364, opened 2021-01-13
Revert "Show full navigation menu on top of top-above-nav slot"

ioanna0:
Reverts guardian/dotcom-rendering#2352

This has cause issue on scrolling the top-above-nav hides behind the navbar

### DIFF
--- a/src/web/components/Nav/ExpandedMenu/ExpandedMenu.tsx
+++ b/src/web/components/Nav/ExpandedMenu/ExpandedMenu.tsx
@@ -86,7 +86,7 @@ const mainMenuStyles = css`
 		margin-right: 70px;
 	}
 	${from.tablet} {
-		margin-right: 53px;
+		margin-right: 100px;
 	}
 `;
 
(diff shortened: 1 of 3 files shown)


ashishpuliyel (approved):
Looks like the opposite of what you did yesterday! So.. LGTM

ioanna0:
@gtrufitt @oliverlloyd @liywjl , I have been looking for another fix on this, with trying to change z-index dynamically if nav menu is expanded on tablet. This appeared to a bit tricky though as the `decideIndex` method will need to get the breakpoint as a parameter but also the action of the `click` event. Happy to pair with one of you to discuss this solution, or you have an alternative solution to share please let me know on how we can many the nav menu fully expanded on top of the ad on tablets

gtrufitt:
> if you have an alternative solution to share please let me know

Could you have `headerWrapperDesktop` and `headerWrapperMobile` as two seperate z-index IDs and then call `getZIndex('headerWrapperDesktop')` within a greater-than-tablet (or whatever sticky nav shows on) media query?

ioanna0:
> Or does the ordering problem happen on mobile too?

Yeah that's the problem it happens also on tablets / mobiles. If we separate z-indexes via desktop and mobile we will only reduce the issue down to tablets and mobiles. We have to also consider that the menu is expanded and dynamically change the z-index to a higher one only for the duration the menu is expanded.

PR #2365, opened 2021-01-13
Design (as a const enum)

oliverlloyd:
## What does this change?
Replaces `designType: CAPIDesign;` which was defined locally and based on strings with `design: Design` which is a const enum [imported from @guardian/types](https://github.com/guardian/types/blob/main/src/format.ts#L18)

## Why?
To align our definition of what an artical is with other teams


shtukas (approved):

oliverlloyd:
[Use Design from @guardian/types](https://trello.com/c/facBrUeP/2260-use-design-from-guardian-types)

tsconfig.test.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "./tsconfig.json",
+    "compilerOptions": {
+        "target": "ES2018" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */,
```
oliverlloyd, line 4: Use target 2018 for jest because we're running on Node 10.15 which can't handle optional chaining

src/web/lib/decideDesignType.ts
```diff
@@ -1,6 +1,39 @@
-export const decideDesignType = (designType: CAPIDesign): DesignType => {
-	if (designType === 'Immersive') return 'Article';
-	if (designType === 'SpecialReport') return 'Article';
-	if (designType === 'GuardianLabs') return 'AdvertisementFeature';
-	return designType;
+import { Design } from '@guardian/types';
+
+export const decideDesignType = (designType: CAPIDesign): Design => {
```
oliverlloyd, line 3: Here is where we transform the CAPIDesign, string value into a Design enum

package.json
```diff
@@ -246,7 +247,7 @@
         "globals": {
             "ts-jest": {
                 "diagnostics": false,
-                "tsConfigFile": "tsconfig.json"
+                "tsConfigFile": "tsconfig.test.json"
```
oliverlloyd, line 250: We use a special test config file that compiles down to 2018 js

package.json
```diff
@@ -226,14 +226,15 @@
...
+        "webpack-filter-warnings-plugin": "^1.2.1",
         "webpack-hot-middleware": "^2.22.0",
         "webpack-hot-server-middleware": "^0.6.1",
         "webpack-merge": "^4.1.2",
         "webpack-node-externals": "^1.7.2",
         "webpack-sources": "^1.1.0"
     },
     "jest": {
-        "preset": "ts-jest/presets/js-with-babel",
+        "preset": "ts-jest/presets/js-with-ts",
```
oliverlloyd, line 237: Don't use babel to transpile

scripts/webpack/frontend.js
```diff
@@ -34,6 +35,9 @@ const commonConfigs = ({ platform }) => ({
         new webpack.DefinePlugin({
             'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
         }),
+        new FilterWarningsPlugin({ 
+            exclude: /export .* was not found in/ 
+        }),
```
oliverlloyd, line 40: As per [this section](https://github.com/TypeStrong/ts-loader/blob/3cc78b8874848c4485250d084a7ffbe3b3fee4ff/README.md#transpileonly) of `ts-loader`'s README

> ... webpack 4 will give you "export not found" warnings any time you re-export a type

Eg.

```bash
$ "export 'OphanABEvent' was not found in '@guardian/types'
```

To prevent thiis problem we are disabling warnings using a regex `/export .* was not found in/` via the `webpack-filter-warnings-plugin`

### Why use this plugin?
Because after several hours of trying different config combinations I _could not_ get webpack 4 to stop showing these warnings. Only this lib seemed to work

src/amp/components/BodyArticle.tsx
```diff
@@ -34,20 +35,19 @@ const bulletStyle = (pillar: CAPIPillar) => css`
...
-	};
+const decideBackground = (design: Design, pillar: CAPIPillar): string => {
+	if (pillar === 'labs') return palette.neutral[86];
+	switch (design) {
+		case Design.Comment:
+			return palette.opinion[800];
+		default:
+			return palette.neutral[100];
+	}
+};
```
oliverlloyd, line 46: `designTypeDefault` was throwing odd typescript errors so to fix these, and also because we want to also use `pillar` in this decision now too, I refactored this logic away from `designTypeDefault`

This will also make moving over to an Editorial Palette easier

src/amp/components/topMeta/TopMeta.tsx
```diff
@@ -1,33 +1,30 @@
...
+			return <TopMetaOpinion articleData={data} pillar={pillar} />;
+		default:
+			return (
+				<TopMetaNews
+					articleData={data}
+					adTargeting={adTargeting}
+					pillar={pillar}
+				/>
+			);
+	}
```
oliverlloyd, line 29: I refactored away from `designTypeDefault` to a classic switch here

fixtures/onwards.mocks.ts
```diff
@@ -112,6 +115,24 @@ const trails: TrailType[] = [
...
+		byline: thisTrail.byline,
+		image: thisTrail.image,
+		isLiveBlog: thisTrail.isLiveBlog,
+		webPublicationDate: thisTrail.webPublicationDate,
+		mediaType: thisTrail.mediaType,
+		mediaDuration: thisTrail.mediaDuration,
+		pillar,
+		design,
+	};
+});
```
oliverlloyd, line 134: This represents the transformation between the raw api response - with CAPIDesign strings - and the typed object used in the stack - with `Design`

liywjl (approved):
👍 LGTM. A lot is going on, but for reasons discussed I understand why.

PR #2369, opened 2021-01-14
Support YoutubeEmbedBlockComponent in main media

liywjl:
## What does this change?
Adds support for `YoutubeEmbedBlockComponent` as main media

## Article used
https://www.theguardian.com/film/2021/jan/13/netflix-new-films-movie-promo-video-disney-streaming-wars

### Before
Screenshot 2021-01-14 at 11 42 39

### After
Screenshot 2021-01-14 at 11 42 01

## Why?
frontend parity

### DIFF
--- a/src/web/components/MainMedia.tsx
+++ b/src/web/components/MainMedia.tsx
@@ -55,7 +55,6 @@ const immersiveWrapper = css`
 `;
 
 const youtubeWrapperStyles = css`
-	margin-top: 12px;
 	margin-bottom: 12px;
 `;
 
--- a/src/web/components/MainMedia.tsx
+++ b/src/web/components/MainMedia.tsx
@@ -8,7 +8,6 @@ import { ImageComponent } from '@root/src/web/components/elements/ImageComponent
 import { YoutubeBlockComponent } from '@root/src/web/components/elements/YoutubeBlockComponent';
 import { YoutubeEmbedBlockComponent } from '@root/src/web/components/elements/YoutubeEmbedBlockComponent';
 import { EmbedBlockComponent } from '@root/src/web/components/elements/EmbedBlockComponent';
-import { Figure } from '@root/src/web/components/Figure';
 
 import { getZIndex } from '@frontend/web/lib/getZIndex';
 
@@ -83,7 +82,7 @@ function renderElement(
 			);
 		case 'model.dotcomrendering.pageElements.YoutubeBlockElement':
 			return (
-				<div
+				<figure
 					key={i}
 					id={`youtube-block-main-media-${i}`}
 					data-cy="main-media-youtube-atom"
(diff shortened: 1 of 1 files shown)


rcrphillips (approved):

src/web/components/MainMedia.tsx
```diff
@@ -106,6 +109,22 @@ function renderElement(
 					/>
 				</div>
 			);
+		case 'model.dotcomrendering.pageElements.VideoYoutubeBlockElement':
+			return (
+				<Figure role={element.role}>
```
oliverlloyd, line 114: Not sure we want to use Figure in main media

liywjl: - Frontend wraps main media in this case in `figure`
- We use `figurecaption` for the caption, which semantically feels like it should belong in `figure`

For me, this raises the question should we use figure for `YoutubeBlockElement` in main Media as well?

src/web/components/MainMedia.tsx
```diff
@@ -2,11 +2,14 @@ import React from 'react';
 import { css, cx } from 'emotion';
 
 import { until } from '@guardian/src-foundations/mq';
+import { Display } from '@guardian/types/Format';
 
 import { ImageComponent } from '@root/src/web/components/elements/ImageComponent';
 import { YoutubeBlockComponent } from '@root/src/web/components/elements/YoutubeBlockComponent';
+import { YoutubeEmbedBlockComponent } from '@root/src/web/components/elements/YoutubeEmbedBlockComponent';
 import { EmbedBlockComponent } from '@root/src/web/components/elements/EmbedBlockComponent';
-import { Display } from '@guardian/types/Format';
```
buck06191, line 9: Was `Display` unused here? Or have you removed something that was using it?

liywjl: no I just moved it a couple lines (see line 5), I like to seperate imports from external packages from internal imports

buck06191: ah yeah. I see it now. I prefer that too.

src/web/components/MainMedia.tsx
oliverlloyd: We use `Figure` (capital F) for spacing, which does not apply here. The semantics of using a `figure` element is perhaps a separate question

src/web/components/MainMedia.tsx
```diff
@@ -52,6 +54,10 @@ const immersiveWrapper = css`
     overflow: hidden;
 `;
 
+const youtubeWrapperStyles = css`
+	margin-bottom: 12px;
+`;
+
```
oliverlloyd, line 60: Is this needed?

liywjl: frontend has a little more spacing between main media and article, I am happy to revert however and check with Harry?

oliverlloyd: The risk is this margin might look right in this context but break thiings in other contexts. MainMedia can be used in lots of places.

oliverlloyd: `Figure` is an attempt to rationalise spacing for in body elements, removing it from individual components and making it so that vertical spacing happens in one place, and one place only. But to implement it we had to review all the places where margins were getting set to make sure we didn't just make the problem worse. It could be we want to use it here for `MainMedia` but we need to do the same review beforehand.

oliverlloyd (approved):

PR #2374, opened 2021-01-15
Bumped Typescript version to 4.1.3 and fixed emerging bug

rhystmills:
## What does this change?
Bumped Typescript in anticipation of the inclusion of `@guardian/libs` which was causing some type issues when used with Typescript 3.9.7.

Also fixed a bug related to `loadScript.ts` - where the Promise output needed to include a `void` type.

### Before
Typescript at version 4.1.3.

### After
Typescript at version 3.9.7.

## Why?
To allow the integration of `@guardian/libs`

### DIFF
--- a/package.json
+++ b/package.json
@@ -219,7 +219,7 @@
         "tslint-plugin-prettier": "^2.3.0",
         "tslint-react": "^4.0.0",
         "tslint-react-a11y": "^1.1.0",
-        "typescript": "^3.9.7",
+        "typescript": "^4.1.3",
         "typescript-json-schema": "^0.43.0",
         "webpack": "^4.44.2",
         "webpack-assets-manifest": "^3.1.1",
--- a/src/web/lib/loadScript.ts
+++ b/src/web/lib/loadScript.ts
@@ -4,7 +4,7 @@
 export const loadScript = (
 	src: string,
 	props?: Omit<Partial<HTMLScriptElement>, 'src' | 'onload' | 'onerror'>,
-): Promise<Event | undefined> =>
+): Promise<Event | void> =>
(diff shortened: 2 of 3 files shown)


liywjl (approved):
👍

rhystmills:
Bit of an update - bumping `typescript` caused a few linting issues which I attempted to resolve. Bumping the version of related libraries was suggested as a fix - unfortunately bumping `@typescript-eslint/eslint-plugin` (which is two major versions behind the latest version) - and/or `eslint` itself (currently a year-old version) caused thousands of additional linting errors. Adding to this, `node` remains on version 10.15 which is 2 years old. Bumping one or all of these dependencies seems to have the knock-on effect of creating thousands of linting errors, perhaps due to other packages being dependent on these versions.

I've spent several hours on the problem and haven't made any progress - as our team can work with alternatives to the logging utility`@guardian/libs` I'm parking this work for now, rather than sinking more time into it.

I think it is a problem that `libs` and our other general-purpose libraries like `types` can't work with DCR - given that it is responsible for rendering an increasingly large segment of the site, and it seems that a bump to the core DCR technologies would help it play better with libraries on more modern versions.

gtrufitt:
Thanks Rhys, and apologies for the pain here, we're going to be looking at bumping these dependencies asap.

oliverlloyd:
It looks like @liywjl 's great work around bumping our dev dependencies could take over from this PR. William, there might still be some useful learnings in here that you can absorb?

liywjl:
Yes we have updated TS and its related linting dependencies. We ended up turing them off and slowly re-enabling them with a bunch of fixes. But right now TS 4.1 is up and running. How that helps 🙂

PR #801, opened 2019-10-04
Use Context to serve the `edition` prop

oliverlloyd:
## What does this change?
Adds an `EditionContext` provider that removes the need to keep passing `edition` around through multiple levels and replaces it with the Provider / Consumer model so it can be accessed as and when required.

https://reactjs.org/docs/context.html

## Note on Hooks
React Hooks were added in 16.8 and include the `useContext` hook which means the syntax for using context is now very clean and readable. There's no longer any need to wrap components in `<Providers>` and instead you can just use `const edition = useContext(EditionContext);` ❤️ 

## How to use
If your component wants to access the `edition` then you need 2 lines of code:

```
import { EditionContext } from '@frontend/web/context/EditionContext';
```

and then, in your function:
```
const edition = useContext(EditionContext);
```

Hooks allow classes to be replaced by functional components so ☝️ should always be sufficient but if you didn't want to migrate to hooks and needed to access `edition` from a class you can use the, still perfectly good, render props pattern:

```
<EditionContext.Consumer>
  {edition => /* render something based on the edition value */}
</EditionContext.Consumer>
```

## How to test context
If your test depends on the `edition` value then you can provide it to your test html stack in the same way as it's given to the actual app:

```
        const { container } = render(
            <EditionContext.Provider value={'AUS'}>
                {/* and now the useContext call inside ReaderRevenueLinks will return AUS */}
                <ReaderRevenueLinks
                    urls={urls}
                    dataLinkNamePrefix={'nav2 : '}
                    noResponsive={false}
                />
            </EditionContext.Provider>,
        );
```

## Link to supporting Trello card
https://trello.com/c/2C74DiyP/744-spike-react-context

PR #2378, opened 2021-01-18
Introduce `decidePalette` to DCR

nitro-marky:
## What does this change?

Work with @oliverlloyd to begin the implementation of a `decidePalette` function that, given `format: Format`, will return the correct colours to be used throughout DCR simply by calling the function `decidePallet(format)` where `format` equals:

```typescript
type Format = { display: Display; design: Design; theme: Theme };
```

_See: https://github.com/guardian/types/blob/main/src/format.ts_

We also add the definition of `format` to `decideLayout`. This is where we build the `format` object for an article, based on the CAPI response.
```typescript
const format: Format = { display, design, theme: pillar };
```

Any thoughts would be greatly appreciated.


src/web/components/ArticleHeadline.tsx
```diff
@@ -211,7 +206,15 @@ export const ArticleHeadline = ({
...
 					return (
 						<>
-							<h1 className={cx(lightFont, invertedText)}>
+							<h1
+								className={cx(
+									lightFont,
+									invertedText,
+									css`
+										color: ${palette.headline.colour};
+									`,
```
liywjl, line 215: might be nice to have all the colours using a const like `headlineColourStyles` as its reused through the file?

nitro-marky: This is old code at the moment and will be removed soon (hopefully!)

liywjl: Just noticed this depends on `palette` that comes from prop, so happy with its current setup

oliverlloyd: Yeah, we thought about that but we decided to wait until background colour comes in because that will likely trigger a larger refactor which will be easier with things inline

index.d.ts
```diff
@@ -35,6 +35,8 @@ type CAPIDesign = DesignType | "Immersive" | "SpecialReport" | "GuardianLabs";
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type DesignTypesObj = { [key in DesignType]: any };
 
+type Palette = {[key: string]: any};
```
liywjl, line 38: prettier issues? the spacing isnt quite right.

Also maybe better to call this `PaletteType` to make it more obvious that it is a type?

liywjl (approved):
Great work 👍  just some minor things to clear up

oliverlloyd:
This https://github.com/guardian/types/pull/61, along with this https://github.com/guardian/dotcom-rendering/pull/2365 will unblock this PR

oliverlloyd:
[Editorial Colour Palettes](https://trello.com/c/tJ84wc3z/1605-editorial-colour-palettes)

oliverlloyd:
We need to replace `white` and `black` with values from Source. Perhaps, we can assign them to named variables so it's easier to read?

JamieB-gu (commented):
This is exciting!

It got me curious about a couple of design decisions that you've made: prop-drilling and the API. I've rambled a bit below, sketching out my understanding of how you've gone about these and asking a couple questions. Let me know if I've understood correctly and what you think!

### Prop-drilling

By this I mean passing the palette down from the "top-level" (i.e. `DecideLayout`). As far as I can tell this is one of three options, the other two being:

1. Pass `Format` down and derive subsets of the palette at the component level, which is what AR has [tentatively started doing](https://github.com/guardian/apps-rendering/blob/70335d436279032fb503d082edb57d31a38bd0a8/src/components/headline.tsx#L25).
2. Use a [Theme](https://emotion.sh/docs/theming) to dependency inject the palette without prop-drilling, which I believe was the direction @oliverlloyd was [originally leaning in](https://github.com/guardian/apps-rendering/pull/313#pullrequestreview-406534014).

What made you settle on prop-drilling in the end?

### The API

The palette here is a single large object, derived from `Format`, with the colours pulled from nested fields. We [discussed the fact](https://github.com/guardian/apps-rendering/pull/313#discussion_r421421338) that DCR would probably go this route while AR used individual functions, but I can't remember why 🤔.

I'm also interested in the naming system and how that corresponds to what Source does. There's a discussion [about this here](https://github.com/guardian/apps-rendering/pull/313#discussion_r423819920) (featuring @SiAdcock), with the hierarchy defined by Source as:

```
[theme][property].[component][variant][state][mode]
```

For a headline text colour I believe this maps to:

- `property`: `text`
- `component`: `headline`
- `variant`: `primary`
- `mode`: `inverse` (only used for dark mode)

and this is why the [AR version of this](https://github.com/guardian/apps-rendering/blob/70335d436279032fb503d082edb57d31a38bd0a8/src/editorialPalette.ts#L115) looks something like:

```ts
{
  text: {
    headlinePrimary: text.headlinePrimary(format),
    headlinePrimaryInverse: text.headlinePrimaryInverse(format),
  },
  background: {
    headlinePrimary: background.headlinePrimary(format),
    headlinePrimaryInverse: background.headlinePrimaryInverse(format),
  },
}
```

I notice that the API defined here looks like this: `palette.headline.colour`. I think `colour` corresponds to `property` and therefore this flips the order of `property` and `component` from what I outlined above? Is that right?

oliverlloyd:
> What made you settle on prop-drilling in the end?

We've looked at [doing less prop drilling before](https://github.com/guardian/dotcom-rendering/pull/801) but we always end up keeping it. I don't think there's anything wrong with using a `theme` and we may well go down that route, but the predictability and simplicity of prop drilling is really nice.

I think this is definately are area where there's a case to be made for it but I think we want to be cautious and approach it in a new PR at a later stage when we better understand our use case and have a known api. I'm also wondering if using a `theme` via Emotion might have benefits in terms of efficiencies when flipping between dark and light mode?

It's perhaps worth pointing out that I'm - so far - grateful that my previous attempts to add Context failed and I'm actually enjoying good old prop drilling (not all the file changes, that bit is less fun, but the simplicity of it)

src/web/lib/decidePalette.ts
```diff
@@ -0,0 +1,39 @@
...
+				default:
+					return neutral[100];
+			}
+		case Display.Showcase:
+		case Display.Standard: {
+			switch (format.design) {
+				case Design.Review:
+				case Design.Recipe:
+				case Design.Feature:
+					return pillarPalette[format.theme].dark;
```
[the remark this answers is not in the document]
oliverlloyd: Good question! The `palette` numbers, 300, 400, etc. are mapped to strings. Seems to be for legacy reasons.

```typescript
const [news, opinion, sport, culture, lifestyle] = [
	_news,
	_opinion,
	_sport,
	_culture,
	_lifestyle,
].map((p) => ({
	// maps legacy colour names to new names in Source
	dark: p[300],
	main: p[400],
	bright: p[500],
	pastel: p[600],
	faded: p[800],

	// continue to expose the new names too!
	300: p[300],
	400: p[400],
	500: p[500],
	600: p[600],
	800: p[800],
}));
```

src/web/lib/decidePalette.ts
```diff
@@ -0,0 +1,39 @@
+import { Display, Design, Format } from '@guardian/types';
```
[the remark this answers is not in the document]
oliverlloyd: Can you elaborate on why this is better? (Genuine question, I've not used this approach and would love to learn more!)

oliverlloyd:
> The palette here is a single large object

This was, I think, related to how we prefer prop drilling in DCR. It certainly feels more natural to me to take this appraoch and I'm also a fan of the discoverability of being able to autocomplete on `palette.`. But I'm not wedded to it and I'm happy to take direction from the team.

oliverlloyd:
> the naming system

This is up for grabs. I definately think it makes sense to align on a pattern and it sounds like you've got a well established one already. I not sure though about having function names the same as property names. I worry this would make refactoring difficult (hard to search or do find and replaces). Maybe instead of

```typescript
headlinePrimary: text.headlinePrimary(format),
```

you could

```typescript
headlinePrimary: text.getHeadlinePrimary(format),
```

The `get` prefix highlights that the function and the name of the property are two different things. What do you think?

oliverlloyd:
> headlinePrimary

Is there a need for `primary`? We don't have a concept in DCR of a secondary. We decide the colour based on `format` and that's that

oliverlloyd:
> ```ts
> {
>   text: {
>     headlinePrimary: text.headlinePrimary(format),
>     headlinePrimaryInverse: text.headlinePrimaryInverse(format),
>   },
>   background: {
>     headlinePrimary: background.headlinePrimary(format),
>     headlinePrimaryInverse: background.headlinePrimaryInverse(format),
>   },
> }
> ```

Would this pattern mean the calling component has to decide between dark and light mode? What about having this logic inside `decidePalette` so you have a conditional, a bit like:

```typescript
text: {
   headlinePrimary: dark ? text.headlinePrimary(format) : text.headlinePrimaryInverse(format),
},
```

(or similar, that api could be better!)

src/web/lib/decidePalette.ts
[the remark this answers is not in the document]
nitro-marky: It can be changed to the numerical value if that's the current trend.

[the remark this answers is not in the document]
nitro-marky: `pillarPalette[format.theme][300]`

gtrufitt:
> Would this pattern mean the calling component has to decide between dark and light mode?

The palette object would be one big object and the getter functions would handle the light and dark I think?

Just my input reading through:

- I originally likes the getter in components approach *in isolation* but within DCR, the prop drilling approach is architecturally consistent. I think stick with it! 
- I think working out and sticking closely to what @SiAdcock is planning in the design system makes sense as the pillar object will likely sit here at some point. I think that's why in [the doc](https://docs.google.com/document/d/1nkntFJUunqrwNYszTrc2RBfyYUd1bEAeQh76yVFOLjo/edit#) we defined as @JamieB-gu mentions

I think we should try and create a type for the palette in @guardian/types and cascade that into AR and DCR - it may be that there are a number of properties on the type that DCR doesn't use in relation particularly to dark mode (or as Oliver mentions the 'secondary' headline type), but having a split in naming or even where the palette is used between the platforms will make it difficult to maintain.

src/web/lib/decidePalette.ts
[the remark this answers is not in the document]
JamieB-gu: I think the **Background** section in this PR description gives a nice summary: https://github.com/microsoft/TypeScript/pull/35200

[`@guardian/eslint-config-typescript` lints for this](https://github.com/guardian/configs/blob/b8ac5cc5c6130258d728f657d34d14c8e3c746f2/packages/eslint-config-typescript/index.js#L41), but tbf DCR doesn't use that yet - maybe I shouldn't have touched on this and instead left it to be fixed when that gets added. cc @sndrs who added that to the config.

oliverlloyd:
> I think we should try and create a type for the palette in @guardian/types and cascade that into AR and DCR

I think we definately want this 👍  But maybe not straight away. The chore of having to make changes to two repos could be a drag on development when initially creating this object. Or we use a monorepo 😄

JamieB-gu:
> the predictability and simplicity of prop drilling is really nice

> This was, I think, related to how we prefer prop drilling in DCR.

> I originally likes the getter in components approach in isolation but within DCR, the prop drilling approach is architecturally consistent. I think stick with it!

Fair, I think that's a reasonable preference. What are your thoughts about passing down the whole `palette` vs passing down `Format` and deriving subsets of the palette at the component level?

> I not sure though about having function names the same as property names

By this do you mean that in the `palette` object the field is a property but in the standalone `text`, `background` etc objects the field is a function? I personally don't have an issue with it because the types tell you what it does (alternatively you can think of a constant as a function with arity 0), but happy to discuss.

> Is there a need for `primary`?

Good question, I don't know. Perhaps one for @SiAdcock?

> Would this pattern mean the calling component has to decide between dark and light mode? What about having this logic inside `decidePalette` so you have a conditional

It does, but typically it already does that because there's [often more than one property](https://github.com/guardian/apps-rendering/blob/4eddab5d666dbecee9ff4066385390bac6adefe8/src/components/anchor.tsx#L25) to change for dark mode. Regarding having it inside `decidePalette`, I think you'd have to pass whether the user is in dark mode or not when you call it? I can see two potential issues with that:

1. It won't work server-side because there's no concept of a user with dark mode set at that point.
2. On the client you'll have to look up whether the user has dark mode set in JS and re-render everything if they do, and also listen for changes to re-render again if they change it. I think it's easier to use the CSS media query?

> I think we should try and create a type for the palette in @guardian/types and cascade that into AR and DCR

> I think we definately want this 👍 But maybe not straight away

Agree with both of these statements. We started incubating it in AR as well but haven't progressed too far because of other priorities.

oliverlloyd:
>  What are your thoughts about passing down the whole `palette` vs passing down `Format` and deriving subsets of the palette at the component level?

I think for DCR, it makes sense to pass down the whole `palette`, at least for now. Mainly because we're used to this pattern but also because I like the simplicity of it, the way `decidePalette` has one export and does one thing. I'm also afraid that we might reduce discoverability if you have to import a function rather than use an object property to get the value.

I'm very much of the opinion that we should keep an open mind here though and this could well be an area we come back to later after we've spent some time doing things one way and learnt more about how that felt

> By this do you mean that in the `palette` object the field is a property but in the standalone `text`, `background` etc objects the field is a function?

I think I just mean that the `headlineBackgroundColour` property and `headlineBackgroundColour()` function are named the same thing. To be honest I always thought of this as big no, no. But that's just because I've been conditioned to think this way by compilers and linters shouting at me when I make the mistake of using the same name in two places. Because of this I see it as _a mistake_ but it seems like - and I'm surprised by this - the linting rules allow it in this case so maybe my mental model is wrong? 🤷 

One problem with this could be if I am using find and replace or `CMD-D` to multi select I'll be highlighting both names, but then maybe that's desirable?

Another is, like this then we're constrained to always have one function per property, you can't have one generic function used in multiple places. Is that maybe intentional though?

> I think it's easier to use the CSS media query

Yeah, I think you're right, CSS would be a good solution here. But lets take this one out of this PR and discuss it seperately

oliverlloyd:
@JamieB-gu after discussing the api (same name business) with @nitro-marky we agreed that the approach used in AR is actually fine and so I've modified the code to align on this

I think for drilling `palette` this works well for DCR at the moment but is something we can revisit.

And let's talk dark mode, but not here.

index.d.ts
```diff
@@ -41,354 +41,356 @@ type Pillar = Theme;
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type DesignTypesObj = { [key in Design]: any };
 
+type Palette = { [key: string]: any };
```
JamieB-gu, line 44: Do you not know the type of these fields?

JamieB-gu:
> I'll be highlighting both names, but then maybe that's desirable?

I think in this case it probably is because you'll want to keep them in sync?

> have one function per property, you can't have one generic function used in multiple places. Is that maybe intentional though?

Maybe? In this case we're defining two APIs that do the same thing:

1. The palette object for DCR to use: `palette.text.headlinePrimary`
2. The functions for AR to use: `text.headlinePrimary(format)`

We probably want every property to have an equivalent function, and to have the property and function provide the same values and be kept in sync?

Again, if you accidentally use the wrong one (e.g. try to use a function where you're supposed to be using a property) the compiler should tell you.

> the approach used in AR is actually fine and so I've modified the code to align on this

> I think for drilling `palette` this works well for DCR at the moment but is something we can revisit.

> And let's talk dark mode, but not here.

👍

index.d.ts
oliverlloyd: We actively decided to not type this object to reduce the time needed to expand it's usage but, for sure, at some point the advantages of autocomplete will outweigh the cost of updating two things to add a new property

index.d.ts
oliverlloyd: I went with `Pallete` to match the naming convention from `@guardian/types` where you have `Design`, `Theme`, etc.

oliverlloyd: I _do_ think a Type suffix is useful though, especially when  dealing with prop data being used for a component where the namespace can be shared a lot

src/web/layouts/DecideLayout.tsx
```diff
@@ -2,8 +2,9 @@ import React from 'react';
 
 import { decidePillar } from '@root/src/web/lib/decidePillar';
 import { decideDisplay } from '@root/src/web/lib/decideDisplay';
-import { Display, Design } from '@guardian/types';
+import { Display, Design, Format } from '@guardian/types';
```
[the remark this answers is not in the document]
oliverlloyd: I added a commit to move to this pattern

src/web/lib/decidePalette.ts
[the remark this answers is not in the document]
oliverlloyd: I added a commit for this

src/web/lib/decidePalette.ts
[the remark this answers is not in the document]
oliverlloyd: This is an area we will need to look at. I'd like to see us align more with Source at some point.

PR #2381, opened 2021-01-19
Refactor dependency handling for the Braze SDK

tjmw:
## What does this change?

Refactors the way dependencies for loading the Braze web SDK are handled. By dependencies we mean the information we need and the conditions which need to be true before we can load the SDK. Currently these are:

* That the switch is on
* The API key is set
* The relevant consents have been given
* That the page isn't paid (GLabs) content
* That the user is signed in and we have their uuid
* That the user _is_ a Gu supporter

### Before

Dependency checks were all intermingled with other code in `BrazeBanner`.

### After

Dependency checks have been extracted out to `checkBrazeDependencies.ts` and are configured in a way that we think makes them much clearer. We're using the new (generic) `dependecyChecker` which we've added in this PR, which takes an array of async dependencies/conditions and processes them until one fails. Note that this is a slight change from the original code where we would make each check in turn in serial, now we're kicking them all off "concurrently" (in quotes because JS is single threaded. Some of the work _may_ be parallelisable, for example the request to idapi) and processing the results serially, until we come across a check which fails.

## Why?

* It's hopefully clearer to a new/returning reader of the code what the dependencies actually are.
* It's easier to write tests around the dependencies.
* We'd like to extract more Braze related code out to its own NPM package, which could be shared across DCR and frontend. Modularising the existing code is a first step along that path.
* It's much easier to have functionality like common logging for when dependencies aren't all met which is really useful when developing.

### DIFF
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -2,10 +2,6 @@ import React, { useEffect, useState } from 'react';
 import * as emotion from 'emotion';
 import * as emotionCore from '@emotion/core';
 import * as emotionTheming from 'emotion-theming';
-import {
-	getConsentFor,
-	onConsentChange,
-} from '@guardian/consent-management-platform';
 import { getZIndex } from '@root/src/web/lib/getZIndex';
 import { Props as BrazeBannerProps } from '@guardian/braze-components';
 import {
@@ -18,6 +14,7 @@ import {
 	setHasCurrentBrazeUser,
 	clearHasCurrentBrazeUser,
 } from '@root/src/web/lib/hasCurrentBrazeUser';
+import { checkBrazeDependencies } from '@root/src/web/lib/braze/checkBrazeDependencies';
 import { CanShowResult } from './bannerPicker';
 
 type Meta = {
@@ -38,35 +35,6 @@ const containerStyles = emotion.css`
     width: 100%;
     ${getZIndex('banner')}
 `;
-
-export const hasRequiredConsents = (): Promise<boolean> =>
-	new Promise((resolve, reject) => {
-		onConsentChange((state) => {
-			try {
-				resolve(getConsentFor('braze', state));
-			} catch (e) {
-				reject(e);
-			}
-		});
-	});
-
-type PreCheckArgs = {
-	shouldHideSupportMessaging?: boolean;
-	pageConfig: { [key: string]: any };
-};
-
-export const canShowPreChecks = ({
-	shouldHideSupportMessaging,
-	pageConfig,
-}: PreCheckArgs) => {
-	// Currently all active web canvases in Braze target existing supporters,
-	// subscribers or otherwise those with a Guardian product. We can use the
-	// value of `shouldHideSupportMessaging` to identify these users, limiting
-	// the number of requests we need to initialise Braze on the page:
-	const userIsGuSupporter = shouldHideSupportMessaging;
-
-	return Boolean(userIsGuSupporter && !pageConfig.isPaidContent);
-};
 
 const SDK_OPTIONS = {
 	enableLogging: false,
@@ -207,10 +175,10 @@ const getBrazeMetaFromQueryString = (): Meta | null => {
 };
 
 const maybeWipeUserData = async (
-	apiKey: string,
-	brazeUuid: null | string,
+	apiKey?: string,
+	brazeUuid?: null | string,
 ): Promise<void> => {
-	if (!brazeUuid && hasCurrentBrazeUser()) {
+	if (apiKey && !brazeUuid && hasCurrentBrazeUser()) {
 		const { default: appboy } = await import(
 			/* webpackChunkName: "braze-web-sdk-core" */ '@braze/web-sdk-core'
 		);
@@ -238,8 +206,8 @@ const maybeWipeUserData = async (
 // OR
 // - The force-braze-message query string arg is passed
 export const canShow = async (
-	asyncBrazeUuid: Promise<null | string>,
-	shouldHideSupportMessaging: undefined | boolean,
+	isSignedIn: boolean,
+	idApiUrl: string,
 ): Promise<CanShowResult> => {
 	const bannerTiming = initPerf('braze-banner');
 	bannerTiming.start();
@@ -252,35 +220,35 @@ export const canShow = async (
 		};
 	}
 
-	const { brazeSwitch } = window.guardian.config.switches;
-	const apiKey = window.guardian.config.page.brazeApiKey;
-	const isBrazeConfigured = brazeSwitch && apiKey;
-	if (!isBrazeConfigured) {
+	const dependenciesResult = await checkBrazeDependencies(
+		isSignedIn,
+		idApiUrl,
+	);
+
+	if (!dependenciesResult.isSuccessful) {
+		const { failure, data } = dependenciesResult;
+		if (SDK_OPTIONS.enableLogging) {
+			// eslint-disable-next-line no-console
+			console.log(
+				`Not attempting to show Braze messages. Dependency ${failure.field} failed with ${failure.data}.`,
+			);
+		}
+
+		await maybeWipeUserData(
+			data.apiKey as string | undefined,
+			data.brazeUuid as string | null | undefined,
+		);
+
 		return { result: false };
 	}
 
-	const [brazeUuid, hasGivenConsent] = await Promise.all([
-		asyncBrazeUuid,
-		hasRequiredConsents(),
-	]);
-
-	await maybeWipeUserData(apiKey as string, brazeUuid);
-
-	if (!(brazeUuid && hasGivenConsent)) {
-		return { result: false };
-	}
-
-	if (
-		!canShowPreChecks({
-			shouldHideSupportMessaging,
-			pageConfig: window.guardian.config.page,
-		})
-	) {
-		return { result: false };
-	}
+	const { data } = dependenciesResult;
 
 	try {
-		const result = await getMessageFromBraze(apiKey as string, brazeUuid);
+		const result = await getMessageFromBraze(
+			data.apiKey as string,
+			data.brazeUuid as string,
+		);
 
 		const timeTaken = bannerTiming.end();
 		record({
--- a/src/web/lib/braze/checkBrazeDependencies.test.ts
+++ b/src/web/lib/braze/checkBrazeDependencies.test.ts
@@ -1,0 +1,318 @@
+import { checkBrazeDependencies } from './checkBrazeDependencies';
+
+let mockBrazeUuid: string | null;
+jest.mock('@root/src/web/lib/getBrazeUuid', () => ({
+	getBrazeUuid: () => {
+		return Promise.resolve(mockBrazeUuid);
+	},
+}));
+
+let mockConsentsPromise: Promise<boolean>;
+jest.mock('./hasRequiredConsents', () => ({
+	hasRequiredConsents: () => {
+		return mockConsentsPromise;
+	},
+}));
+
+let mockHideSupportMessaging: boolean;
+jest.mock('./hideSupportMessaging', () => ({
+	hideSupportMessaging: () => {
+		return mockHideSupportMessaging;
+	},
+}));
+
+describe('checkBrazeDependecies', () => {
+	let windowSpy: jest.SpyInstance<any>;
+
+	beforeEach(() => {
+		windowSpy = jest.spyOn(window, 'window', 'get');
+	});
+
+	afterEach(() => {
+		windowSpy.mockRestore();
+
+		// Wait for any unsettled promises to complete at the end of each test. Once
+		// we encounter a failure in our list of checks we don't need to wait on
+		// subsequent operations to complete which is why there might be unsettled
+		// promises.
+		const flushPromises = new Promise(setImmediate);
+		return flushPromises;
+	});
+
+	const setWindow = (windowData: { [key: string]: any }) =>
+		windowSpy.mockImplementation(() => windowData);
+
+	it('succeeds if all dependencies are fulfilled', async () => {
+		setWindow({
+			guardian: {
+				config: {
+					switches: {
+						brazeSwitch: true,
+					},
+					page: {
+						brazeApiKey: 'fake-api-key',
+						isPaidContent: false,
+					},
+				},
+			},
+		});
+		mockBrazeUuid = 'fake-uuid';
+		mockConsentsPromise = Promise.resolve(true);
+		mockHideSupportMessaging = true;
+
+		const isSignedIn = true;
+		const idApiUrl = 'https://idapi.example.com';
+		const got = await checkBrazeDependencies(isSignedIn, idApiUrl);
+
+		expect(got.isSuccessful).toEqual(true);
+		expect(got.data).toEqual({
+			brazeSwitch: true,
+			apiKey: 'fake-api-key',
+			consent: true,
+			isNotPaidContent: true,
+			brazeUuid: 'fake-uuid',
+			userIsGuSupporter: true,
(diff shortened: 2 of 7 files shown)


src/web/lib/braze/checkBrazeDependencies.test.ts
```diff
@@ -0,0 +1,242 @@
...
+afterEach(() => {
+	// Wait for any unsettled promises to complete at the end of each test. Once
+	// we encounter a failure in our list of checks we don't need to wait on
+	// subsequent operations to complete which is why there might be unsettled
+	// promises.
+	const flushPromises = new Promise(setImmediate);
+	return flushPromises;
+});
+
+let mockBrazeUuid: string | null;
```
tjmw, line 12: I don't ❤️  all the mocking we're doing here but I think it's worth it for the confidence that the wiring etc is all correct, and at least it's all confined to this one test. This is much more of an integration test than a unit test.

src/web/lib/dependencyChecker.ts
```diff
@@ -0,0 +1,68 @@
...
+				};
+			}
+
+			return {
+				isSuccessful: true,
+				data: {
+					[cur.name]: result,
+					...syncAcc.data,
+				},
+			};
```
liywjl, line 59: feels like it would be clearer if this retrun was in the `try` block

src/web/lib/braze/checkBrazeDependencies.ts
```diff
@@ -0,0 +1,63 @@
...
+import { getBrazeUuid } from '@root/src/web/lib/getBrazeUuid';
+import { hasRequiredConsents } from './hasRequiredConsents';
+import { hideSupportMessaging } from './hideSupportMessaging';
+
+const getCompleteDependencies = async (
+	isSignedIn: boolean,
+	idApiUrl: string,
+): Promise<Array<DependencyConfig>> => {
+	return [
+		{
```
liywjl, line 15: I suspect prettier is not being applied correctly here.

src/web/lib/braze/checkBrazeDependencies.ts
```diff
@@ -0,0 +1,63 @@
+import {
+	checkDependencies,
+	DependencyConfig,
+	DependencyResult,
+} from '@root/src/web/lib/dependencyChecker';
+import { getBrazeUuid } from '@root/src/web/lib/getBrazeUuid';
+import { hasRequiredConsents } from './hasRequiredConsents';
+import { hideSupportMessaging } from './hideSupportMessaging';
+
+const getCompleteDependencies = async (
```
liywjl, line 10: the function is using `async`, but you are not using `await` anywhere

liywjl (commented):
I fully agree in the direction you are talking the Braze implementaion, I think it will help simplilfy and make it more testable in the future. 

However I think we could look at simplifying some of the logic around `checkBrazeDependencies.ts` and `dependencyChecker.ts`. I feel like a lot of what you are trying to achieve could be handled by `Promise.all`. Doing so would be a lot easier to understand to new devs looking at the code as they would likely understand the mental model quicker that using a `reduce` to handle each case. 

I am unsure what is the best approch would be, but if Braze requires all conditions to be met before displaying, maybe encapsulating each condtion in a promise wrapped by `Promise.all` might help.

tjmw:
Hi @liywjl thank you for the feedback! We've also had a discussion within the team about whether the `dependencyChecker` code is worthwhile given the complexity it adds, so I think your concerns are totally valid. I've just pushed up a commit which uses a simpler approach, removing `dependencyChecker` completely. I'd be interested in your thoughts on the change (we'll also discuss within TX).

src/web/lib/braze/checkBrazeDependencies.test.ts
```diff
@@ -0,0 +1,242 @@
...
+
+let windowSpy: jest.SpyInstance<any>;
+
+beforeEach(() => {
+	windowSpy = jest.spyOn(window, 'window', 'get');
+});
+
+afterEach(() => {
+	windowSpy.mockRestore();
+});
```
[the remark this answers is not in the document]
tjmw: Good spot, I've got that on a mental todo list of things to fix! Makes sense to refactor now.

src/web/lib/braze/checkBrazeDependencies.ts
tjmw: Yeah I wondered the same thing, and had a conversation with Alex S on Friday about it. I _think_ it's just GitHub rendering the diff weirdly. Locally it looks fine and I see the correct number of tab characters. I'll check again to make sure.

src/web/lib/braze/checkBrazeDependencies.ts
tjmw: Good spot, this is fixed in the latest commit I pushed.

tjmw:
Hi @liywjl I think I've addressed the feedback you left, and have moved away from using `dependencyChecker.ts` to a simpler approach. I'd be interested to hear what you think about the latest changes, particularly the stuff in 105277b.

src/web/lib/braze/checkBrazeDependencies.ts
```diff
@@ -1,63 +1,96 @@
...
+	const data: ResultData = {};
+
+	{
+		const name = 'brazeSwitch';
+		const value = window.guardian.config.switches.brazeSwitch;
+		if (!value) {
+			return buildFailureResponse(name, value, data);
+		}
+		data[name] = value;
+	}
```
liywjl, line 41: The structure I has in mind was more on the lines of:

```
const brazeSwitch = new Promise((resolve, reject) => {
    const value = window.guardian.config.switches.brazeSwitch;
    if (!value) {
        reject(buildFailureResponse(name, value, data))
    }
    resolve(value)
});

[...]

return [brazeSwitch, ...]
```

That means when you call `checkBrazeDependencies` you can do something like:

```
const got = await Promise.all[checkBrazeDependencies(isSignedIn, idApiUrl)}
    .then(results => {
        ...
    })
    .catch(err => {
        ...
    });
```

Also in thise case it might be worth renaming `checkBrazeDependencies` to something like `arrayBrazeChecks` (not fully convinced by the my suggestion either)

The example I based on a post I found really useful https://www.javascripttutorial.net/es6/javascript-promise-all/
Thoughts on this?

liywjl (commented):
Awesome work! This is definitely an improvement, still had another suggestion on how `checkBrazeDependencies`, curious to know your thoughts on my suggestion?

src/web/lib/braze/checkBrazeDependencies.ts
tjmw: Thanks for the detail! I considered `Promise.all` before, but ended up discounting it as the behaviour is _slightly_ different to what we have now. Currently the failure (if any) will be for the first item in the ordered list of dependencies. Using `Promise.all` changes that slightly to be the first item which fails chronologically, which _might_ not be the same. Especially if some operations are longer running than others (for example an API call). I'm now wondering how relevant or important that is though.

The other piece I'm not sure about with `Promise.all` is that in the event that one of the checks fails, we want to be able to access the data from any other successful checks. I don't think `Promise.all` lets you access that data (which is why we were reducing a list of promises in the `dependencyChecker`).

I definitely agree that it would be nice to be using common JavaScript APIs and building blocks here. I'll try some things out and report back.

liywjl: I see, maybe it might be worth looking at using `Promise.allSettled()` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

From what I gather that might be a better than `Promise.all ` for what you are hoping to achieve?

tjmw: Hi @liywjl I had a play with `Promise.allSettled` but wasn't super happy with the result (it turns out it isn't supported in the version of Node we're on which adds some additional friction). I pushed up a few more commits, including f160afb, which is kind of a compromise approach where I iterate over the conditions rather than pass them to `Promise.allSetted`. I think it could be a nice happy medium but interested in your take on it.

I left the `allSetted` approach in there but commented out so you can see how it looked. The part where we iterate the results to build up the final `DependenciesResult` object wasn't very nice.

tjmw: Hi @liywjl I've had a bit more of a play around with refactoring this but didn't get to a point where I was happier with the result than what's there now. I've removed the commented `Promise.allSettled` experimental code and have left a comment on using `Promise.allSettled` in future when it's more reliably available. What are your thoughts on me merging this as it is?

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -252,35 +220,35 @@ export const canShow = async (
...
 
-	const [brazeUuid, hasGivenConsent] = await Promise.all([
-		asyncBrazeUuid,
-		hasRequiredConsents(),
-	]);
+	if (!dependenciesResult.isSuccessful) {
+		const { failureData, failureField, data } = dependenciesResult;
+		if (SDK_OPTIONS.enableLogging) {
+			// eslint-disable-next-line no-console
+			console.log(
```
liywjl, line 232: `console.error` might be better?

src/web/lib/braze/checkBrazeDependencies.ts
```diff
@@ -0,0 +1,100 @@
...
+	for (const { name, value } of dependencies) {
+		try {
+			// eslint-disable-next-line no-await-in-loop
+			const result = await value;
+
+			if (result) {
+				data[name] = result;
+			} else {
+				return buildFailureResponse(name, result, data);
+			}
```
liywjl, line 84: maybe just throw an error here, so `buildFailureResponse` only ever used in catch block?

src/web/lib/braze/checkBrazeDependencies.ts
```diff
@@ -0,0 +1,100 @@
...
+			name: 'brazeSwitch',
+			value: Promise.resolve(window.guardian.config.switches.brazeSwitch),
+		},
+		{
+			name: 'apiKey',
+			value: Promise.resolve(window.guardian.config.page.brazeApiKey),
+		},
+		{
+			name: 'brazeUuid',
+			value: isSignedIn ? getBrazeUuid(idApiUrl) : Promise.resolve(null),
```
liywjl, line 47: worth wrapping in a Promise?

src/web/lib/braze/checkBrazeDependencies.ts
liywjl: Also if `buildFailureResponse` is only defined once, then maybe we don't need it as a seperate function? we could return the object here?

src/web/lib/braze/checkBrazeDependencies.ts
```diff
@@ -0,0 +1,100 @@
...
+type DependencyResult = {
+	name: string;
+	value: Promise<string | boolean | null | undefined>;
+};
+const buildFailureResponse = (name: string, value: any, data: ResultData) => ({
+	isSuccessful: false,
+	failureField: name,
+	failureData: value,
+	data,
+});
```
liywjl, line 30: personal preferrence but maybe the response could be structured like:

```
{
    isSuccessful: false,
    error: {
         failureField: name,
	failureData: value,
    },
    data
}
```

liywjl: that way if there are no errors we could have something that looks like:

```
{
    isSuccessful: true,
    error: null,
    data
}
```

Its nicer to not have to check if an object key exists and always assume it does exist  but is not set

src/web/lib/braze/checkBrazeDependencies.ts
```diff
@@ -0,0 +1,100 @@
...
+				e && e.message ? e.message : e,
+				data,
+			);
+		}
+	}
+
+	return {
+		isSuccessful: true,
+		data,
+	};
```
liywjl, line 97: ditto about returning something in the structure like:

```
{
    isSuccessful: true,
    error: null,
    data
}
```

liywjl (approved):
LGTM 👍 all the comments are just personal preferences, so feel free to ignore if you disagree

src/web/lib/braze/checkBrazeDependencies.ts
oliverlloyd: I might be wrong but I think you can just use the `async` key to do this?

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: It's actually not super exceptional to end up here, so an error doesn't feel quite right to me (for example, any user who isn't logged in won't meet all the dependencies, so will end up here). This is mostly useful to a developer working on the code (also in the wild `SDK_OPTIONS.enableLogging` won't be enabled).

src/web/lib/braze/checkBrazeDependencies.ts
tjmw: Ah interesting question! This did actually cross my mind, but I decided against (using exceptions for flow control never feels quite right to me). Maybe the scope here is tight enough that it's not a problem though. It would be nice to have a single failure case build in one place only. I'll give it a try and see how it looks!

src/web/lib/braze/checkBrazeDependencies.ts
tjmw: So I guess wrapping in a promise would look like this, is this what you had in mind?

```js
new Promise((resolve) => {
  if (isSignedIn) {
    getBrazeUuid(idApiUrl).then(resolve);
  } else {
    resolve(null);
  }
})
```

> I might be wrong but I think you can just use the async key to do this?

I'm interested in this! I _thought_ you could only use `async` to mark a function as async (which as you say means the function will always return a promise). Do you mean use an immediately invoked function in combination with `async` here?

src/web/lib/braze/checkBrazeDependencies.ts
tjmw: Oh interesting, yeah combining the two failure fields together is nice. Maybe the types as they are right now are overdoing it a bit. Currently I've got:

```js
type SuccessResult = {
	isSuccessful: true;
	data: ResultData;
};

type FailureResult = {
	isSuccessful: false;
	failureField: string;
	failureData: any;
	data: ResultData;
};

type DependenciesResult = SuccessResult | FailureResult;
```

But given a single nullable error key this could become:

```js
type DependencyError = {
        failureField: string;
	failureData: any;
};

type DependenciesResult = {
        error: DependencyError | null;
	data: ResultData;
}
```

The consumer would then just check if `error` is truthy to know the overall outcome.

> Its nicer to not have to check if an object key exists and always assume it does exist but is not set

I think in this case TypeScript is helping us, since if `isSuccessful` is `false` the type of `FailureResult` specifies that the `failureField` and `failureData` fields will always be defined. (Interesting blog post describing the approach I used here: https://mariusschulz.com/blog/more-literal-types-in-typescript#boolean-literal-types)

src/web/lib/braze/checkBrazeDependencies.ts
oliverlloyd: It was a drive by comment without fully parsing the code in my head but yes, `async` is like shorthand for 'this function always returns a promise' which I though might be handy here but take my idea with a lot of salt!

src/web/lib/braze/checkBrazeDependencies.ts
tjmw: I did try this but it felt clearer to me keeping the falsy case and the catch separate.

src/web/lib/braze/checkBrazeDependencies.ts
tjmw: I went ahead and combined the two failure fields into one object which I agree is nicer, thanks for the suggestion! For now I've kept separate `SuccessResult` and `FailureResult` types.

PR #2383, opened 2021-01-19
Add overflow-wrap: break-word to textBlockCompoment paraStyles

gtrufitt:
## What does this change?

Adds `overflow-wrap: anywhere` to `textBlockComponent` to allow the line to break and not flow out of the browser window on small screens.
| Before | After |
| --- | ----------- |
| image  |image|

## Why?
To break long lines.

### DIFF
--- a/src/web/components/elements/TextBlockComponent.tsx
+++ b/src/web/components/elements/TextBlockComponent.tsx
@@ -137,6 +137,7 @@ export const TextBlockComponent: React.FC<Props> = ({
 	const paraStyles = css`
 		margin-bottom: 16px;
 		${body.medium()};
+        line-break: anywhere;
 
 		ul {
 			margin-bottom: 12px;
--- a/src/web/components/elements/TextBlockComponent.tsx
+++ b/src/web/components/elements/TextBlockComponent.tsx
@@ -137,7 +137,7 @@ export const TextBlockComponent: React.FC<Props> = ({
 	const paraStyles = css`
 		margin-bottom: 16px;
 		${body.medium()};
-        line-break: anywhere;
+		overflow-wrap: anywhere;
 
 		ul {
 			margin-bottom: 12px;


liywjl (approved):
👍  LGTM, sounds like something that we would benefit from having a visual regression test for

oliverlloyd:
What about https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap? I've not tested it but it _sounds_ good vs. https://developer.mozilla.org/en-US/docs/Web/CSS/line-break

src/web/components/elements/TextBlockComponent.tsx
```diff
@@ -137,6 +137,7 @@ export const TextBlockComponent: React.FC<Props> = ({
 	const paraStyles = css`
 		margin-bottom: 16px;
 		${body.medium()};
+        line-break: anywhere;
```
oliverlloyd, line 140: Nitpick: This might be a syptom of us changing the style linter recently but that tabbing seems off.

gtrufitt: Ah gawd, does that mean I can't edit on Github anymore

oliverlloyd:
I can't immediately tell the difference but `anywhere` has less browser support so can we get away with `break-word`? Or will polyfill handle this?

Screenshot 2021-01-19 at 16 21 33

gtrufitt:
> I can't immediately tell the difference but `anywhere` has less browser support so can we get away with `break-word`? Or will polyfill handle this?

`break-word` doesn't work because our `grid-area: body` flows out of the container. If I can restrain the width of that to *its* container, it might fix all manner of sins and allow us to use `break-word`.

Quick test is using `max-width: calc(100vw - 20px)` (20px is margin width). Seems hacky, needs more investigation.

src/web/components/elements/TextBlockComponent.tsx
paperboyo: `line-break: anywhere` Oh! That may be better that `break-word` we applied somewhere recently (ie. will break URLs at hyphens and punctuation instead of middle of the word?). Asking, coz I don’t understand [the definition](https://developer.mozilla.org/en-US/docs/Web/CSS/line-break).

gtrufitt: @paperboyo here's some examples: https://codepen.io/gtrufitt/pen/NWRJRQo

gtrufitt: `line-break: anywhere` does seem to fair up better in those examples?

oliverlloyd (approved):
Thanks for the examples - that really helped

gtrufitt:
Discussing with @paperboyo and reviewing Frontend's implementation, will need to fix the bug and then use `overflow-wrap: break-word`

oliverlloyd:
Is this one good to go?

gtrufitt:
> Is this one good to go?

Nope, got a bug to fix with grid and then implement `overflow-wrap: break-word`

gtrufitt:
Fixed elsewhere

PR #2386, opened 2021-01-20
Add clear: left; to supporting content

gtrufitt:
## What does this change?

Clears supporting content on the left to prevent getting caught on floated content above. This matches the 'thumbnail' role which already has `clear: left`.

| Before | After |
| --- | ----------- |
|Screen Shot 2021-01-20 at 11 08 29 |Screen Shot 2021-01-20 at 11 08 16|

### DIFF
--- a/src/web/components/Figure.tsx
+++ b/src/web/components/Figure.tsx
@@ -17,6 +17,7 @@ const roleCss = {
 	`,
 
 	supporting: css`
+		clear: left;
 		margin-top: ${space[3]}px;
 		margin-bottom: ${space[3]}px;
 		${from.tablet} {


shtukas (approved):

liywjl (approved):
Maybe worth adding a story?

oliverlloyd (approved):
Nice. I saw this and thought it was an incorrect margin but got confused because it made no sense they would have changed...

gtrufitt:
> Nice. I saw this and thought it was an incorrect margin but got confused because it made no sense they would have changed...

Defo the route I investigated first!

PR #2389, opened 2021-01-20
Fix carousel index marker and behaviour

buck06191:
Fix broken progress indicator when using buttons. Prior to this the
index was often off by 1, seemingly because the state was captured prior
to updating the carousel.

Here we have moved to using `useReducer` to dispatch events related to
particular actions. This means that we can capture desired and current
behaviour/state, making the progress indicator more reliable.

TODO:
- [ ] Capture scroll progress and update progress bar.
- [ ] Extract adjusted dots logic to be usable in both CSS and JS (depends
on viewport size).

Co-authored-by: nitro-marky <mark.addis@guardian.co.uk>

### DIFF
--- a/src/web/components/Onwards/Carousel/Carousel.tsx
+++ b/src/web/components/Onwards/Carousel/Carousel.tsx
@@ -1,16 +1,17 @@
-import React, { useCallback, useEffect, useReducer, useRef } from 'react';
+import React, { useEffect, useReducer, useRef } from 'react';
 import {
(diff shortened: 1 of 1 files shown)


buck06191:
[📦 Debug existing carousel](https://trello.com/c/U3hBZJ5N/2333-%F0%9F%93%A6-debug-existing-carousel)

buck06191:
Bug found to only occur in Storybook and is fixed by #2411

PR #2393, opened 2021-01-21
Fix prettier formatting for cypress tests

gtrufitt:
## What does this change?

Adds prettier for Cypress and updates all matched files


oliverlloyd (approved):
💎

PR #2394, opened 2021-01-21
Fix prettier formatting for cypress tests

gtrufitt:
## What does this change?

Migrates the document.test Jest tests to Cypress because they're in realistic environment and easier to add and debug.


cypress/integration/e2e/article.metaAndOphan.e2e.spec.js
```diff
@@ -0,0 +1,118 @@
...
+	before(getPolyfill);
+
+	beforeEach(function () {
+		disableCMP();
+		setLocalBaseUrl();
+	});
+
+	const getExists = (selector) => cy.get(selector).should('have.length', 1);
+
+	describe('for WEB the document renders with the correct meta and analytics info', function () {
```
oliverlloyd, line 15: These pieces of describe text with WEB and E2E don't really make sense anymore so you could just write whatever is useful in this context - not really a big deal, I guess it only helps when scrolling the results or looking at errors

cypress/integration/e2e/article.metaAndOphan.e2e.spec.js
```diff
@@ -0,0 +1,118 @@
...
+		it(`Has a head element`, function () {
+			cy.visit(
+				`/Article?url=https://www.theguardian.com/lifeandstyle/2021/jan/21/never-conduct-any-business-naked-how-to-work-from-bed-without-getting-sacked`,
+			);
+			// cy.get(`head`).should('have.length', 1);
+			getExists(`head`);
```
oliverlloyd, line 21: I prefer to have as much code inline for tests so that they're easier to scan when debugging. I find abstractions in tests especially costly but this is a subjective thing so 🤷

cypress/integration/e2e/article.metaAndOphan.e2e.spec.js
```diff
@@ -0,0 +1,118 @@
...
+					value:
+						'From warming up your voice to avoiding spillages, here are some tips for keeping up professional appearances',
+				},
+				{
+					name: 'viewport',
+					value: 'width=device-width,minimum-scale=1,initial-scale=1',
+				},
+			];
+
+			metaTags.map((meta) =>
```
oliverlloyd, line 36: Do we need the map? You could just have two sets of assertions?

cypress/integration/e2e/article.metaAndOphan.e2e.spec.js
```diff
@@ -0,0 +1,118 @@
...
+			const names = [
+				'og:url',
+				'og:description',
+				'og:title',
+				'og:type',
+				'og:image',
+				'article:author',
+			];
+
+			names.map((name) => getExists(`head meta[property="${name}"]`));
```
oliverlloyd, line 53: Okay, so now I'm reading on I start to like this pattern more!

oliverlloyd (approved):
So I made a few comments about abstractions but after reading the code through to the end I'm less concerned if I'm honest.

One point though is that we can probably lift all of these smaller tests into one larger one. Cypress [recomends this](https://docs.cypress.io/guides/references/best-practices.html#Creating-%E2%80%9Ctiny%E2%80%9D-tests-with-a-single-assertion) approach. Assuming that, then 👍

cypress/integration/e2e/article.metaAndOphan.e2e.spec.js
oliverlloyd: Actually, maybe not one, perhaps several, but ideally fewer!

cypress/integration/e2e/article.metaAndOphan.e2e.spec.js
gtrufitt: Yeah, I tend to agree but I think in this case as there's so many it actually makes the test much easier to read as a whole

cypress/integration/e2e/article.metaAndOphan.e2e.spec.js
gtrufitt: IMO, I think it's clearer that *all* the values of the array do the *exact* same thing - that way it's easier to parse quickly and understand a single test than if they were separate and you had to carefully parse each assertion

oliverlloyd: That's true. One point though is that people will only ever land here when something fails and they will likely be on a how can I fix this fast ride so the less context they need to absorb the best. Also, if you need to fix a failing test, you don't want to have to check the remaining tests that are sharing that code.

I'm probably overthinking it though, these cases are pretty simple, it's more a general principle thing which sounds a lot like I'm being dogmatic!

oliverlloyd: And to be clear, I think in this case - the array map - I'm actuially in favour of the way it is here!

gtrufitt:
> One point though is that we can probably lift all of these smaller tests into one larger one.

Reading that doc, Cypress suggests that you can chain your tests together when you're looking for content, an attribute value etc in the same element which isn't quite the case here. I do understand that there is an advantage to performance, but I am also looking at the test run and it feels quite clear and simple to understand? 

Maybe a little unit-testy but there doesn't seem to be a huge overhead:



If you're concerned that this would be a poor example and cause issues in the future and think it definitely needs a change please shout, I'll fold to how strongly you feel!

cypress/integration/e2e/article.metaAndOphan.e2e.spec.js
gtrufitt: Yeah, definitely a good general principle, I think this is an odd case where there's no user interaction, just a 'did we render everything ok' where this is probably fairly clear (I had a failed cypress test in Github on this PR and it was quick to understand 😄 )

oliverlloyd:
> you can chain your tests together when you're looking for content, an attribute value etc

Yeah, I noticed the same afterwards 👍

gtrufitt:
Let me know if it needs refactor (or we can look at cypress tests holistically in health week!)

PR #2395, opened 2021-01-21
Identity | Sign in gate | Fix whitespace fade on sign in gate design test variant 3

coldlink:
## What does this change?
Due to the change in whitespace for this test, the fade would mess up on certain articles. This corrects this by setting the correct margin to match the new whitespace height.

### Before
chrome_RAJACPr0mJ

### After
chrome_ia2ztpAT7a


alexduf (approved):

gtrufitt:
Thanks for sorting the fade height thing!

PR #2396, opened 2021-01-21
Adds Ipsos MORI tag to AMP

jfsoul:
## What does this change?
Adds Ipsos MORI analytics (i.e. dotmetrics) to AMP. We have followed their simple instructions except for 2 differences:

1. We are hardcoding `www.theguardian.com` instead of using `AMPDOC_HOST` to populate the `dom` URL parameter. I believe this will be needed because dotmetrics has been configured to understand our domain as www.theguardian.com and will not return a config for amp.theguardian.com. It shouldn't affect the tracking requests anyway.
2. I've added a geolocation restriction so the tracking is only fired in the UK.

One thing that is difficult to know is whether this will work correctly in production before release. Dotmetrics do domain matching and so it's impossible to test until served from theguardian.com. Presumably they've worked this out so that it'll work either direct to amp.theguardian.com or through the google cache, but I haven't been able to verify yet.

### Before
No Ipsos MORI analytics

### After (In the UK only)
image

## Why?
See #1961 for more details.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -571,6 +571,7 @@ interface ConfigType extends CommercialConfigType {
     idUrl?: string;
     mmaUrl?: string;
     brazeApiKey?: string;
+    ipsosTag?: string;
 }
 
 interface ConfigTypeBrowser {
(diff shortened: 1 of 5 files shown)


liywjl (approved):
LGTM 👍

PR #2397, opened 2021-01-21
Set up large image variant styles

buck06191:
## What does this change?
Applies the styling for the "large image variant" in the upcoming curated content carousel test.

For design example, see [relevant Figma link](https://www.figma.com/file/j7Y1d0wXpnmoCW3j0u1dF4/DotCom---Onward-Journeys-OKR?node-id=1066%3A19998)

### After

image

### DIFF
--- a/src/web/components/Card/components/CardAge.tsx
+++ b/src/web/components/Card/components/CardAge.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { css, cx } from 'emotion';
+import { css } from 'emotion';
 
 import { neutral } from '@guardian/src-foundations/palette';
 import { textSans } from '@guardian/src-foundations/typography';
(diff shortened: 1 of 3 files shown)


buck06191:
[📦 Create variant A carousel](https://trello.com/c/Q3XJVrJG/2335-%F0%9F%93%A6-create-variant-a-carousel)

oliverlloyd:
The changes to CardAge are affecting other versions of the Card in different places

Screenshot 2021-01-22 at 12 40 08

Looking at these designs, they certainly _look_ like this is a new version of the Card and that we should be branching within the model (using `Format`) to accommodate them. The problem though is I don't think we've really thought about how we manage Cards within the model yet. I guess you could say this is the `Display.Immersive` version of a Card? @HarryFischer ?

Also, we probably want to do the same with `CardHeadline` - use that with `Display.Immersive` to position and style the text differently.

Ideally, and I think this should be our goal here, we want `Carousel` to be a container that manages the furniture around this type of onwards, the dots and chevrons, etc. And then `Card` is just a component the we pass in with the relevant `Format` props. This would mean that we're happy with these new designs of Card but given we don't _have_ an immersive version of a Card yet, this would have no impact so it seems like a reasonable thing to do?

buck06191:
Closing this to open a fresh branch that builds off current `main`. This should keep it in sync with the changes seen in #2416

PR #2399, opened 2021-01-22
Move appboy initialisation and message retrieval into a shared context

rhystmills:
## What does this change?

- Moves the `appboy` initialisation to a shared function (in `initialiseAppboy.ts`).
- Moves the retrieval of messages to shared functions (in `BrazeMessages.ts`), with tests.
- Allows the retrieval of slot-specific messages in preparation for the new `endOfArticle slot`.

## Why?

A step towards shared Braze web-messaging code between two slots, supporting `endOfArticle` messaging for the Marketing team.

## Frontend?

We're focusing on DCR for now as that's where the majority of eligible page views are for the `endOfArticle` messaging.


src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -96,14 +88,16 @@ const getMessageFromBraze = async (
 	const appboyTiming = initPerf('braze-appboy');
 	appboyTiming.start();
 
-	appboy.initialize(apiKey, SDK_OPTIONS);
+	const brazeMessages = new BrazeMessages(appboy);
 
-	const canShowPromise: Promise<CanShowResult> = new Promise((resolve) => {
-		let subscriptionId: string | undefined;
+	const messages = brazeMessages.getMessagesForBanner();
```
tjmw, line 93: Given that this method returns a promise which _may_ resolve with a single message what do you think about renaming to `getMessageForBanner`, singular? (Sorry I think this may have been my naming in the first place!)

src/web/lib/braze/FakeAppBoy.ts
```diff
@@ -0,0 +1,21 @@
+import { createNanoEvents, Emitter } from 'nanoevents';
```
tjmw, line 1: Now that we're only using this in one place, I almost wonder if it would be better defined within the test where it's used. Mostly just to make it super clear that it's a testing utility. Any thoughts on that?

src/web/lib/braze/BrazeMessages.ts
```diff
@@ -0,0 +1,49 @@
...
+	subscribeToInAppMessage: (callback: Callback) => string;
+}
+
+interface Message {
+	extras?: Extras;
+	triggerId?: string; // This keeps Typescript happy, because Message could be an InAppMessage or a ControlMessage
+}
+
+class BrazeMessages {
+	// emitter: Emitter;
```
tjmw, line 15: Let's remove this commented line

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -114,57 +108,36 @@ const getMessageFromBraze = async (
...
 				value: appboyTimeTaken,
 			});
+
+			return outcome;
 		})
 		.catch(() => {
 			appboyTiming.clear();
 			// eslint-disable-next-line no-console
 			console.log('Appboy Timing failed.');
+			return { result: false };
```
tjmw, line 139: I think returning `{ result: false }` here is the right thing to do. It occurs to me that the `console.log` just above isn't entirely correct is it? The promise chain could have rejected for any reason, not just that the timing failed, isn't that right?

tjmw (commented):
This is looking good to me, I left a few minor comments. It'd be good to get a resolution on whether it's okay to no longer unsubscribe after receiving the first banner message. I think it's probably fine since I believe there's no overlap in audiences for current canvases (so you'd never be eligible for > 1 message).

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -114,57 +108,36 @@ const getMessageFromBraze = async (
...
+			return { result: true, meta };
+		})
+		.then((outcome) => {
 			const appboyTimeTaken = appboyTiming.end();
 
 			record({
 				component: 'braze-appboy-timing',
 				value: appboyTimeTaken,
 			});
+
```
adamnfish, line 132: I realise this was here before, but should we call `appboyTiming.clear` from a finally block to make sure it gets called either way? It's not clear to me from the docs whether calling clear after you are done with the timer is required to:
* prevent the browser keeping its performance monitoring apparatus engaged
* make sure we don't accidentally use a previous measurement

src/web/lib/braze/initialiseAppboy.ts
```diff
@@ -0,0 +1,19 @@
+const SDK_OPTIONS = {
+	enableLogging: false,
+	noCookies: true,
+	baseUrl: 'https://sdk.fra-01.braze.eu/api/v3',
+	sessionTimeoutInSeconds: 1,
+	minimumIntervalBetweenTriggerActionsInSeconds: 0,
+};
+
+const getInitialisedAppboy = async (apiKey: string): Promise<typeof appboy> => {
```
adamnfish, line 9: Is this memoized as written (and do we want it to be)? It looks like if the Epic and Banner were both to request this, we'd do the import and initialisation code twice?

adamnfish (commented):
It looks from this design (and the tests) like we're planning on having a single instance of BrazeMessages in the platform, that's shared between the relevant slots. At the moment it gets instantiated directly from the BrazeBanner code, which is a little at odds with this. This might just be because the BrazeBanner is currently the only slot and there's an intention to move this instantiating code to a new central location when it comes up? The risk there is that we forget that and someone ends up duplicating the instantiation code into the epic banner instead of moving it to a shared location. At that point, the test for `supports multiple calls to the same slot, returning separate promises` becomes irrelevant, since that's not the expected usage.

No harm in any of this, but it's good to try and keep in mind that the tests should represent documentation for the way the code should actually be used <- I'm sure this is already the case which is great. The other side of the coin is that the code itself obeys this expected usage, and ideally railroads future users towards sticking with the intended design. Right now the code doesn't encourage current or future users to create a single instance shared across slots, which jars with the expected usage I'm inferring from what I've read here. It's especially risky because as written (to my eyes) if that happens it looks like we'd end up creating multiple instances of the Braze SDK, each with their own subscription? Maybe this works fine, but it feels like a moving part we will be happy to do without in future when debugging why messages aren't being delivered.

Assuming this is a just a temporary state of affairs as one step in a refactor towards our goal then no harm at all! This is a relatively minor point too, so apologies for spending so many words on it.

tjmw:
Hi @adamnfish thanks for the comments on this.

> It looks from this design (and the tests) like we're planning on having a single instance of BrazeMessages in the platform, that's shared between the relevant slots

Yep that's correct. The only reason we're still creating a `BrazeMessages` instance directly in `BrazeBanner` is because that's the only place it's used currently. So this PR is a stepping stone to get to where we want to be. As you say the expected behaviour if we were to import the Braze SDK (and initialize the session etc) multiple times in a single page view seems ambiguous, and almost certainly not the expected usage. As part of integrating with the end of article slot we'll pull the creation of both the SDK and `BrazeMessages` up out of `BrazeBanner` to be shared.

src/web/lib/braze/initialiseAppboy.ts
tjmw: Good spot, no it's not memoized at this point. The plan is to do that at the point where we need it.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: Hmm I don't remember the detail here. @rhystmills are you able to shed any light?

src/web/lib/braze/FakeAppBoy.ts
rhystmills: Good point - I've moved it into the test.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
rhystmills: Yes true. I'm not sure there's anything particularly useful we can log here given that - so I've removed it.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
rhystmills: Yes, that makes perfect sense - all updated now.

src/web/components/StickyBottomBanner/BrazeBanner.tsx
rhystmills: Yes that makes more sense than it being in the catch, so I'll move it.

I was under the impression that performance marks were scoped to the page, but I'm also having a hard time finding conclusive documentation that states that. Perhaps this is one to look into for a future PR? It would apply wherever else `initPerf` is used across the website.

shtukas (approved):

PR #2403, opened 2021-01-22
Fix typo on Chromatic config

oliverlloyd:
## What does this change?
Replaces `disabled` with `disable`

## Why?
Because it's `disable` and, for context, we don't want to keep seeing this:

Screenshot 2021-01-22 at 14 31 43


liywjl (approved):
🤣  great catch!

nitro-marky (approved):

nitro-marky:
That image is almost burnt into my retinas now.

PR #2408, opened 2021-01-22
Embedded content prototyping

frj:
This PR contains the prototyping for the click to view component.

image

This component is designed to deal with the problems of embedding third party content on theguardian.com in respect to the users privacy. Requests made to the third party during the process of embedding their content give them the opportunity to 'track' the user viewing the content.

We have categorised the third parties based on whether they explicitly support a 'do-not-track' feature.

For those that do we will continue to display their embedded content in the normal way.

For those that don't we will use this ClickToView component to prevent the embedded content from being loaded on the page requiring the user to click to consent to be being tracked by the third party, at which point we will load the embedded content.

 ## Overlay Size

For the best user experience the click to view 'overlay' would match the exact dimensions of the component that is rendered once the overlay is clicked. This would avoid the page shifting around on click. 

It turns out that in many cases it would be difficult to calculate this size accurately. 

As an example the following links

A Facebook post:
https://www.theguardian.com/uk-news/2019/aug/16/police-officers-death-sparks-sweeping-inquiry

A vimeo video:
https://www.theguardian.com/film/2020/oct/12/ultraviolence-ken-fero-documentary-injustice-deaths-police

Notice the facebook post will expand to fill the width of the page and expands vertically to fit the post content. The vimeo video scales to maintain its aspect ratio.

These components are 'embed' elements with very similar markup with no obvious way to derive their eventual dimensions.

Hence in this first cut the overlay will render at its natural size and the page will expand/shrink to fit the embedded content when it is displayed.

## Page score

There are concerns of the negative effect on page score this layout shift of the page will have. This is not the case as the layout shift is in response to a user action:

> Layout shifts that occur within 500 milliseconds of user input will have the hadRecentInput flag set, so they can be excluded from calculations.

https://web.dev/cls/#user-initiated-layout-shifts

In some cases there may actually be an improvement to page score as some embeds cause the layout shift on page load ( eg instagram), which will be deferred until after a user interaction. See:

https://www.theguardian.com/culture/2019/apr/27/idris-elba-marries-sabrina-dhowre-in-morocco

## Storybooks

This PR will not impact live code but provides a set of storybooks showing examples of real world embedded content and the ClickToView component. 

Just to avoid confusion not all of the example embeds are considered to be 'tracking', but are included to test the broadest range of embeds and their interaction with the ClickToView component.

The number of storybooks is probably excessive, they will be rationalised in subsequent interations. 

## Next Steps

The next steps will be to:
- Integrate the ClickToView component in to actual page rendering. This will be done at the ArticleRendered level.
- Temporarily Integrate a 0% A/B test into the ClickToView component so that this feature can be demoed to the Journalists in a realistic way in the context of the articles they are publishing, prior to full rollout.


oliverlloyd:
These stories will work better if broken out into separate stories, instead of one larger one. In general, this makes it easier to view the content - things have titles, you can arrow through them, and easier to test - errors on one component don't fail them all. This is doubly true when using regression testing where we want to test component by component.

frj:
> These stories will work better if broken out into separate stories, instead of one larger one. In general, this makes it easier to view the content - things have titles, you can arrow through them, and easier to test - errors on one component don't fail them all. This is doubly true when using regression testing where we want to test component by component.

Agreed. I have made a reasonable start on creating the overlay component and the examples are living in the story for that component for the moment. 

As we integrate the overlay into the various different components that are going to use it, I can move the associated stories to the stories for those components.

oliverlloyd (commented):
This is great work!

I think though we can change the api so specific components have specific tasks, does that make sense? So more:

```typescript
const ThirdParty = (children, isTargetted) => {
  if (isTargetted) { 
     return (<ClickToView>{children}</ClickToView>)
   }
   else return {children}
}
```

That makes ThirdParty a lot simpler and keeps all the ClickToView type stuff happening inside that component

src/web/components/ClickToView.tsx
```diff
@@ -12,8 +12,11 @@ import { Lines } from '@guardian/src-ed-lines';
 
 type Props = {
 	children: React.ReactNode;
-	width: number;
-	height: number;
+	width?: number;
+	height?: number;
```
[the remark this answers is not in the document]
frj: Given that we dont think we can easily match the size of the overlay to the size of the embed it 'covers', I am now using width: 100% and not setting a height.

You had some concern that not setting the height would cause a double reflow ie when the over lay was removed and when the embed is inserted.

This doesnt seem to be the case, see:

https://5dfcbf3012392c0020e7140b-ymnejlgtya.chromatic.com/?path=/story/components-clicktoview--small-story

Not having a height definitely simplifies things, what do you think?

src/web/components/ClickToView.tsx
```diff
@@ -12,8 +12,11 @@ import { Lines } from '@guardian/src-ed-lines';
...
+	overlayHeadline: string;
+	overlayBody: React.ReactNode;
+	overlayButtonText: string;
```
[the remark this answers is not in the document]
frj: As discussed Ive collapsed the ThirdPartyContent into the ClickToView so I think this is covered by that.

src/web/components/ClickToView.tsx
```diff
@@ -92,59 +106,59 @@ const Body = ({
...
 		setShowOverlay(false);
-		onAccept();
 	};
 
+	useEffect(() => {
+		if (!showOverlay) {
+			onAccept();
+		}
+	});
+
```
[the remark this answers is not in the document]
frj: I did this because for the specific use case of the UnsafeEmbedBlockComponent.

It requires that the embedIframe() function gets called after the associated iframe has been written into the DOM, this resizes the iframe.

Calling that method in the onClick handler doesnt work as the iframe has not been written into the dom at that point.

I need to do some more digging into this as there are still some issues and it doesn't work perfectly.

src/web/components/ThirdPartyEmbeddedContent.tsx
```diff
@@ -0,0 +1,78 @@
...
+		const noOp = () => {};
+
+		return (
+			<ClickToView
+				width={width}
+				height={height}
+				overlayHeadline={headline}
+				overlayBody={body}
+				overlayButtonText={buttonText}
+				onAccept={onAccept || noOp}
```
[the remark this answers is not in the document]
frj: I think I was just over complicating things its gone now.

src/web/components/ThirdPartyEmbeddedContent.tsx
```diff
@@ -0,0 +1,78 @@
+/* eslint-disable jsx-a11y/label-has-associated-control */
```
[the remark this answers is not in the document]
frj: Gone

src/web/components/ClickToView.tsx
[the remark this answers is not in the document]
frj: You can seen this in action in the storybook:

https://5dfcbf3012392c0020e7140b-ymnejlgtya.chromatic.com/?path=/story/components-clicktoview--unsafe-embed-block-component-story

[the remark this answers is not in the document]
oliverlloyd: Ah, I see. In this case you can wrap the call to the callback in a setTimeout, like `setTimeout(() => onMyThing())}` where the second param to setTimeout there is empty (so zero). This causes the execution to be put at the end of the js event stack.

Using useEffect is undesirably because it will fire the callback multiple times. You _couold_ use the custom hook we creted called `useOnce` to prevent this but the setTimeout is hopefully sufficient

src/web/components/ClickToView.tsx
```diff
@@ -12,9 +12,68 @@ import { Lines } from '@guardian/src-ed-lines';
...
+		case 'inline':
+		case 'showcase': {
+			return 'small';
+		}
+		case 'halfWidth':
+		case 'supporting':
+		case 'thumbnail': {
+			return 'xsmall';
+		}
+	}
```
oliverlloyd, line 76: These functions are really nice! I like how it's so decalrative.

src/web/components/ClickToView.tsx
```diff
@@ -89,57 +149,101 @@ const Headline = ({
...
+		>
+			{children}
+		</div>
+	);
+};
 
-export const ClickToView = ({ children, width, height, onAccept }: Props) => {
+export const ClickToView = ({
+	children,
+	role: roleProp,
```
oliverlloyd, line 174: Ha, yeah, this is a perenial problem!

src/web/components/ClickToView.tsx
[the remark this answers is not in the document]
frj: Great that works.

src/web/components/ClickToView.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+				sourceDomain="athirdparty.com"
+			>
+				<div id="third-party-content" />
+			</ClickToView>,
+		);
+
+		expect(getByText('Allow A Third Party content?')).toBeInTheDocument();
+		expect(
+			getByText(
+				'This article includes content provided by A Third Party. We ask for your permission before anything is loaded, as they may be using cookies and other technologies.',
```
oliverlloyd, line 37: If we reduce the length of this string we're still testing that the element shows but perhaps remove some maintainance work for the future?

src/web/components/ClickToView.tsx
```diff
@@ -89,74 +113,115 @@ const Headline = ({
...
+		if (onAccept) {
+			setTimeout(() => onAccept());
+		}
 	};
 
-	if (showOverlay) {
+	let headlineText;
+	let body;
+
+	const roleWithDefault: RoleType = role || 'inline';
```
oliverlloyd, line 178: We can replace this with `role = 'inline'` in the props on line 160. But maybe we can remove it completely by making `role` required? It _is_ required and should always be provided. If the parent component really doesn't know then that might be the better place to default to 'inline'

src/web/components/ClickToView.tsx
```diff
@@ -89,74 +113,115 @@ const Headline = ({
...
-							>
-								Allow and continue
-							</Button>
-						</Bottom>
-					</Inner>
-					<Lines />
-				</Outer>
+			<Container>
+				<Headline role={roleWithDefault}>{headlineText}</Headline>
+				<Body role={roleWithDefault}>{body}</Body>
```
oliverlloyd, line 217: My preference here would be to inline this html instead of having the `headlineText` and `body` variables. The reason  for this is to make future maintenance or refactors easier. In dotcom we strive to try and keep the code as simple as possible because we know we have a long path ahead of us involving lots of changes and refactors so the fewer abstractions and more inline code there is, the easier this will be. Obviously, this case is about as simple as abstractions get so it's true that the difference is very slight. The issue though is how the code might grow. Abstraction have a way of being built upon and this is the real thing we want to avoid.

src/web/components/ClickToView.tsx
```diff
@@ -89,74 +113,115 @@ const Headline = ({
...
+		}
 	};
 
-	if (showOverlay) {
+	let headlineText;
+	let body;
+
+	const roleWithDefault: RoleType = role || 'inline';
+
+	if (isTracking && showOverlay) {
```
oliverlloyd, line 180: 👍

src/web/components/ClickToView.tsx
```diff
@@ -89,74 +113,115 @@ const Headline = ({
...
+		if (source) {
+			headlineText = `Allow ${source} content?`;
+			body = (
+				<div>
```
oliverlloyd, line 184: Ideally we wouldn't need this div. I think by removing the abstraction it becomes moot but in any case it could be replaced with a `<></>` fragment

src/web/components/ClickToView.tsx
```diff
@@ -89,74 +113,115 @@ const Headline = ({
...
+
+const AcceptButton = ({
+	role,
+	handleClick,
+}: {
+	role: RoleType;
+	handleClick: Function;
+}) => {
+	return (
+		<div>
```
oliverlloyd, line 144: Do we need this div?

src/web/components/ClickToView.tsx
```diff
@@ -89,74 +113,115 @@ const Headline = ({
...
 			</Container>
 		);
 	}
 
-	return (
-		<Container width={width} height={height}>
-			{children}
-		</Container>
-	);
+	return <div>{children}</div>;
```
oliverlloyd, line 226: We can remove these divs (or use fragments if needed)

oliverlloyd (approved):
Looks great! 👍 

There are some divs we can remove and some thoughts on making `role` required plus inlining some html but other than that it's really nice!

src/web/components/ClickToView.tsx
frj: Yes, otherwise it seems the button expands to be the full width of the component.

src/web/components/ClickToView.tsx
frj: 👍

src/web/components/ClickToView.tsx
frj: 👍

src/web/components/ClickToView.tsx
frj: 👍

src/web/components/ClickToView.test.tsx
frj: 👍

src/web/components/ClickToView.tsx
oliverlloyd: Souonds good.

Worth noting that the concept of `role` changed recently. There was a change in the CAPI code to add this property to a lot more elements by default and a reindex of all the articles was run to bring it though. But even then, perhaps it's not that common?

oliverlloyd (approved):

PR #2410, opened 2021-01-22
Support witness block elements

liywjl:
## What does this change?

Add support for witness text/image/video block elements

### Example articles
- youtube/image: https://www.theguardian.com/lifeandstyle/2015/aug/28/20-awesome-board-games-you-may-never-have-heard-of
- text: https://www.theguardian.com/cities/2016/feb/29/ring-ask-milky-bar-kid-your-memories-stories-red-phone-boxes

### After
Screenshot 2021-01-22 at 16 42 39
Screenshot 2021-01-22 at 16 42 47
Screenshot 2021-01-22 at 16 42 54

## Why?
Support witness articles

## Note:
Witness articles are depricated, we therefore have removed links.

## TODO: 
- [ ] Text does not have newline char

on frontend:
Screenshot 2021-01-22 at 16 36 24

on DCR:
Screenshot 2021-01-22 at 16 42 47

- [ ] date format different

Screenshot 2021-01-22 at 16 37 45

Screenshot 2021-01-22 at 16 44 18


src/web/components/elements/WitnessImageBlockComponent.tsx
```diff
@@ -0,0 +1,68 @@
...
+		authorName={authorName}
+		dateCreated={dateCreated}
+		pillar={pillar}
+	>
+		<>
+			<img
+				className={css`
+					width: 100%;
+				`}
+				src="https://gu-witness.s3.amazonaws.com/image/1440505256168e17c87ce879c40d3f575b6625340fb7c-mediumoriginalaspectdouble.jpg"
```
oliverlloyd, line 47: Hardcoded to Mars guy?

Screenshot 2021-01-22 at 16 38 25

src/web/components/elements/WitnessVideoBlockComponent.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+		'https://witness.theguardian.com/user/Gregg Lewis-Qualls',
+	height: 330,
+	youtubeTitle: 'Warhammer Quest Timelapse',
+};
+
+export const WitnessVideoBlockComponentDefault = () => (
+	<div
+		className={css`
+			margin: 15px;
+			width: 620px;
```
oliverlloyd, line 49: If you use `ContainerLayout` this story might be easier and you won't need these

src/web/components/elements/WitnessVideoBlockComponent.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+	>
+		<WitnessVideoBlockComponent
+			title={witnessTypeData.title}
+			description={witnessTypeData.description}
+			authorUsername={witnessTypeData.authorUsername}
+			width={witnessTypeData.width}
+			height={witnessTypeData.height}
+			youtubeUrl={witnessTypeData.youtubeUrl}
+			dateCreated={witnessTypeData.dateCreated}
+			pillar={Pillar.Sport}
```
oliverlloyd: Maybe we can just inline these prop values? Pulling them in from an object doesn't add anything

src/web/components/elements/WitnessVideoBlockComponent.tsx
```diff
@@ -0,0 +1,77 @@
...
+	description,
+	authorUsername,
+	width,
+	height,
+	youtubeUrl,
+	dateCreated,
+	pillar,
+}: Props) => {
+	const youtubeVideoId = youtubeUrl.split('?v=')[1];
+	const parsedURL = `https://www.youtube-nocookie.com/embed/${youtubeVideoId}?wmode=opaque&feature=oembed`;
```
oliverlloyd, line 42: We're no longer using `nocookie` as far as I'm aware

src/web/components/elements/WitnessVideoBlockComponent.tsx
```diff
@@ -0,0 +1,77 @@
...
+	title,
+	description,
+	authorUsername,
+	width,
+	height,
+	youtubeUrl,
+	dateCreated,
+	pillar,
+}: Props) => {
+	const youtubeVideoId = youtubeUrl.split('?v=')[1];
```
oliverlloyd, line 41: This seems like it might cause issues. Is there a better way to do this?

src/web/lib/ArticleRenderer.tsx
```diff
@@ -519,6 +522,75 @@ export const ArticleRenderer: React.FC<{
 							language={element.language}
 						/>
 					);
+				case 'model.dotcomrendering.pageElements.WitnessBlockElement': {
+					const witnessType = element.witnessTypeData._type;
+					switch (witnessType) {
+						case 'model.dotcomrendering.pageElements.WitnessTypeDataImage':
+							const witnessTypeDataImage = element.witnessTypeData as WitnessTypeDataImage;
+							return (
+								<figure itemType="http://schema.org/ImageObject">
```
oliverlloyd, line 531: We should be using `Figure` here. That will set the top and bottom margins which don't seem to exist here

oliverlloyd (commented):
A few things and one question: What do all these `itemProp`s do?

src/web/components/elements/WitnessVideoBlockComponent.tsx
```diff
@@ -0,0 +1,77 @@
...
+	${headline.xxxsmall()}
+`;
+
+export const WitnessVideoBlockComponent = ({
+	title,
+	description,
+	authorName,
+	width,
+	height,
+	youtubeUrl,
```
oliverlloyd, line 37: If we're not using an atom here and if we have zero control over what this url is, isn't the most we can do just use it? Rather than trying to parse it?

src/web/components/elements/WitnessVideoBlockComponent.tsx
liywjl: The only placed where this is defined would be the `html` or `youtubeUrl` props. The `html` option isnt ideal for styling. The `youtubeUrl` has some issues, the example in storybooks (extracted from CAPI) retrurns a string like: `'http://www.youtube.com/watch?v=XRq83txBgxw'`. It appears when you try load a video in an iframe using `/watch` that it will not load, instead you need to use `/embed/${videoId}` to get it to work.

The only way I see of extracting the youtuve video ID is to extract it via `youtubeUrl`.

src/web/components/elements/WitnessVideoBlockComponent.stories.tsx
liywjl: I tried using `ContainerLayout` but it seems to make each block a lot langer than it acctually appears

In storybook:
Screenshot 2021-01-25 at 11 10 43

In DCR (running locally)
Screenshot 2021-01-25 at 12 26 59

shtukas:
Nice one!

oliverlloyd (approved):
Looks good. As discussed, it might be easier to have fewer files but only if it's an easy thing to change

PR #2411, opened 2021-01-24
Only mock the date when building for Chromatic

oliverlloyd:
## What?
Wraps the call to MockDate in a check to see if we're building the stories for Chromatic and only executes the mock in this context.

## Why?
The reason we want to mock the date is to prevent false negatives in Chromatic. But doing so has started to cause issues when running Storybook locally where we're using code that is dependant on time advancing. Because `MockDate` freezes time, functions like `setTimeout` or `setInterval` won't work as expected.


shtukas (approved):

gtrufitt (approved):
LGTM

PR #2413, opened 2021-01-25
Fewer Cases

oliverlloyd:
## What?
Replaces the pattern where we were explicitly declaring all possible `Design` types when switching on that type with a pattern where we only declare the cases that we actually switch on.

## Why?
Because by doing this we make it easier to add or remove `Design` types in future and also because this ensures that we always include a `default` case


shtukas (approved):

oliverlloyd:
[Fewer Cases](https://trello.com/c/4Rx2zSfn/2368-fewer-cases)

liywjl (approved):

PR #2416, opened 2021-01-25
Use card component for carousel

buck06191:
## What does this change?
Swaps to using the `Card` component in the Carousel.

In order to do this it adds some additional props to the `Card` component to enforce a minimum width and flex direction.
It also adds a prop to `LI` in order to snap the scroll to the start of cards in the carousel.

### Before
image

### After
image

## Why?
For AB test around onwards content


buck06191:
[📦 Create variant B carousel](https://trello.com/c/JKYc6PX5/2336-%F0%9F%93%A6-create-variant-b-carousel)

shtukas (approved):

src/web/components/Onwards/Carousel/Carousel.tsx
```diff
@@ -77,7 +72,7 @@ const carouselStyle = css`
 
 	position: relative; /* must set position for offset(Left) calculations of children to be relative to this box */
 
-	overflow-x: scroll; /* Scrollbar is less intrusive visually on non-desktop devices typically */
+	overflow-x: auto; /* Scrollbar is less intrusive visually on non-desktop devices typically */
```
oliverlloyd, line 75: Not a major concern but I think we switched to `auto` when playing with this but I'm not sure if this is preferable over `scroll`. I read the docs on auto and it says

> Depends on the user agent. If content fits inside the padding box, it looks the same as visible, but still establishes a new block-formatting context. Desktop browsers provide scrollbars if content overflows.

So, yeah, I'm not sure I'm any wiser after reading that!

src/web/components/Card/components/ImageWrapper.tsx
```diff
@@ -5,18 +5,23 @@ import { until } from '@guardian/src-foundations/mq';
...
+}: Props) => {
 	return (
 		<div
 			className={css`
 				/* position relative is required here to bound the image overlay */
 				position: relative;
 
 				flex-basis: ${percentage && percentage};
-				${until.tablet} {
+				${!alwaysVertical && until.tablet} {
```
oliverlloyd, line 24: This reads really nice!

src/web/components/Card/components/LI.tsx
```diff
@@ -56,6 +61,7 @@ type Props = {
 	padSides?: boolean; // If this LI directly wraps a card this should be true
 	bottomMargin?: boolean; // True when wrapping a card in a column and not the last item
 	showTopMarginWhenStacked?: boolean;
+	snapAlignStart?: boolean; // True when snapping card when scrolling e.g. in carousel
```
oliverlloyd, line 64: +1 for the comment here and +2 for keeping it generic so it works for other use cases

src/web/components/Onwards/Carousel/Carousel.tsx
```diff
@@ -420,7 +289,7 @@ export const Carousel: React.FC<OnwardsType> = ({
 	});
 
 	const cards = trails.map((trail, i) => (
-		<Card trail={trail} isFirst={i === 0} />
+		<CarouselCard key={trail.url + i} trail={trail} />
 	));
```
oliverlloyd, line 293: It might save a bit of scrolling around if this was just inlined inside the `ul`. I think now that it is just that `ul` tag there's no more need to extract it out?

src/web/components/Onwards/Carousel/CarouselCard.tsx
```diff
@@ -0,0 +1,37 @@
...
+				design={design}
+				headlineText={headlineText}
+				webPublicationDate={webPublicationDate}
+				kickerText={kickerText || ''}
+				imageUrl={imageUrl || ''}
+				showClock={true}
+				alwaysVertical={true}
+				minWidthInPixels={258}
+			/>
+		</LI>
```
oliverlloyd, line 35: A take it leave it thought but you could just inline this code directly inside the `ul`. It would make that file a big bigger but it would reduce the need to swap files. But it's true `Carousel` is getting big so I'll leave it up to you.

One thing though, and this applies in both cases. On the first `LI` if you set `showDivider` to false then you don't get the leading vertical divider.

You could achieve this with some sort of `isFirst` prop as it stands but if it were to inline the map then you could just `showDivider={index !== 0}`?

oliverlloyd (approved):
Nicely done!

I made some comments around maybe inlining some code and I also noticed we can remove that leading vertical divider on the first card but I'll leave it up to you

src/web/components/Onwards/Carousel/Carousel.tsx
buck06191: I think that (based on a StackOverflow dive) the difference is that 
> Auto will only show a scrollbar when any content is clipped.
> Scroll will however always show the scrollbar even if all content fits and you cant scroll it.

So for our use case it should probably be `auto` even if we expect the content to always clip. If there is a situation where there's a single card or something, we don't allow scrolling and so we should reduce the risk of errors in the progress indicator.

src/web/components/Card/components/CardLayout.tsx
```diff
@@ -22,16 +24,50 @@ const decideDirection = (imagePosition?: ImagePositionType) => {
 	}
 };
 
-export const CardLayout = ({ children, imagePosition }: Props) => (
+const decideWidth = (minWidthInPixels?: number) => {
+	if (minWidthInPixels) {
+		return css`
+			${`min-width: ${minWidthInPixels}px`};
```
liywjl, line 30:
```suggestion
			min-width: ${minWidthInPixels}px;
```

src/web/components/Onwards/Carousel/CarouselCard.tsx
```diff
@@ -0,0 +1,37 @@
...
+}: CarouselCardProps) => {
+	const {
+		pillar,
+		design,
+		url: linkTo,
+		image: imageUrl,
+		headline: headlineText,
+		webPublicationDate,
+		kickerText,
+	} = trail;
```
liywjl, line 20: I feel like these should be individual props rather than all just in a `trail` prop

liywjl (approved):
👍  Awesome work!!

src/web/components/Onwards/Carousel/CarouselCard.tsx
oliverlloyd: Yeah, I think you're right. We have a bit of a (unwritten) rule that explicit props is better. Eg. we ban prop spreading.

It'd be nice to be able to _know_ what values are actually being used from `trail` and which are required without having to dive in

shtukas (approved):

src/web/components/Onwards/Carousel/Carousel.tsx
```diff
@@ -269,63 +188,53 @@ export const Title = ({
...
+			}}
+			headlineText={headlineText}
+			webPublicationDate={webPublicationDate}
+			kickerText={kickerText || ''}
+			imageUrl={imageUrl || ''}
+			showClock={true}
+			alwaysVertical={true}
+			minWidthInPixels={258}
+		/>
+	</LI>
```
oliverlloyd, line 228: I guess now that this component doesn't do much, you could just move this code to line 394 and not have the component at all?

PR #2417, opened 2021-01-26
Format

oliverlloyd:
## What does this change?
The partially adds usage of the `format: Format` type

```typescript
type Format = {
  display: Display;
  design: Design;
  theme: Theme;
}
```

## Why?
For consistency. `Format` is now used as the way to 'describe' an article but in our code we are still passing `pillar`, `design` and `display` by themselves

## Why not everything?
Because I want to avoid painful merge conflicts and there's no changes to any external imports or exports here so we can safely do this work piecemeal


shtukas (approved):

src/web/components/App.tsx
```diff
@@ -297,6 +297,12 @@ export const App = ({ CAPI, NAV }: Props) => {
 		design,
 	});
 
+	const format: Format = {
+		display,
+		design,
+		theme: pillar,
+	};
```
oliverlloyd, line 304: We don't get passed this value from CAPI so we build it here

src/web/components/MostViewed/MostViewedFooter/SecondTierItem.tsx
```diff
@@ -98,7 +98,12 @@ export const SecondTierItem = ({
 	} = trail;
 
 	const avatarToShow = avatarUrl || image;
-	const pilarToUse = design === Design.Comment ? Pillar.Opinion : pillar;
+	const pillarToUse = design === Design.Comment ? Pillar.Opinion : pillar;
+	const formatToUse: Format = {
+		design,
+		display: Display.Standard,
```
oliverlloyd, line 104: The `Card` data we're sent from CAPI includes nothing about `display` so we default to `Standard`

oliverlloyd:
[Use Format from @guardian/types in DCR](https://trello.com/c/Qh45JBQP/2367-use-format-from-guardian-types-in-dcr)

PR #2418, opened 2021-01-26
ClickToView Designs

oliverlloyd:
## What does this change?
Brings the `ClickToView` component inline with the designs by updating the copy and adding a grey background

### Before
Screenshot 2021-01-26 at 11 00 23

### After
Screenshot 2021-01-26 at 11 15 49

### DIFF
--- a/src/web/components/ClickToView.tsx
+++ b/src/web/components/ClickToView.tsx
@@ -122,16 +122,17 @@ export const ClickToView = ({ children, width, height, onAccept }: Props) => {
 					<Inner>
 						<Top>
 							<Headline width={width}>
-								The colourful beak is very large
+								Allow content provided by a third party?
 							</Headline>
 							<Body width={width}>
-								Quaerat quaerat ex nihil autem consequatur.
-								Velit rerum at ad dignissimos aut excepturi
-								ratione excepturi. Quaerat ipsam natus totam et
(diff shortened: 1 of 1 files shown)


src/web/components/ClickToView.tsx
```diff
@@ -121,17 +121,19 @@ export const ClickToView = ({ children, width, height, onAccept }: Props) => {
...
-								Velit rerum at ad dignissimos aut excepturi
-								ratione excepturi. Quaerat ipsam natus totam et
-								aut distinctio eaque voluptatem.
-								<Link href="https://theguardian.com">
-									{' '}
-									Quaerat ipsam
-								</Link>
+								<p>
+									This article includes content hosted on
+									other.com. We ask for your permission before
```
jfsoul, line 129: One thing I was wondering in our last meeting was where we'd get `other.com` from. Perhaps it'd make sense to be in the CAPI response as `trackingDomain` or something?

jfsoul (approved):

src/web/components/ClickToView.tsx
oliverlloyd: `other.com`_should_ eventually be replaced by a prop sent in from CAPI but we might need some fallback copy

liywjl (approved):
LGTM

PR #2428, opened 2021-01-28
ArticleTitle uses palette

oliverlloyd:
## What does this change?
Replaces local definition of colours with the `palette` object

## Why?
To make the code more scalable and easier to maintain.

## Why do I see these visual diffs?
By doing this work we fix some incorrect settings for things like Immersive Comment pieces. We don't have any of these so the bug was never surfaced by by simplifying the logic we fixed it by accident!

### DIFF
--- a/src/web/lib/decidePalette.ts
+++ b/src/web/lib/decidePalette.ts
@@ -4,14 +4,17 @@ import { neutral, text } from '@guardian/src-foundations';
 
 import { pillarPalette } from '@root/src/lib/pillars';
 
+const WHITE = neutral[100];
+const BLACK = text.primary;
+
 const textHeadline = (format: Format): string => {
 	switch (format.display) {
 		case Display.Immersive:
 			switch (format.design) {
 				case Design.PrintShop:
-					return text.primary;
+					return BLACK;
 				default:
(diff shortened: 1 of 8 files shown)


oliverlloyd:
[ArticleTitle uses palette](https://trello.com/c/qbUoZhJO/2384-articletitle-uses-palette)

src/web/lib/decidePalette.ts
```diff
@@ -21,20 +24,58 @@ const textHeadline = (format: Format): string => {
 				case Design.Feature:
 					return pillarPalette[format.theme].dark;
 				case Design.Interview:
-					return neutral[100];
+					return WHITE;
```
nitro-marky, line 27: Should these be better as palette colours? I'm not sure with source these days!

nitro-marky (approved):

src/web/lib/decidePalette.ts
oliverlloyd: I think the purest definition of white from `source` is this `neutral[100]`. The rest seem very context oriented, like `background.primary` or `text.cta...`, neither of which is semantically true here

PR #2430, opened 2021-01-28
Remove deprecated settings

buck06191:
## What does this change?
Gets rid of a bunch of deprecated settings. VS Code kept warning me about them and it added noise when I was updating `settings.json`.

Tested it locally by modifying and then saving a few files and it seems to all be working the same.

## Before
image

## After
image


shtukas (approved):

oliverlloyd (approved):
Love the small tagetted PR 👍

PR #2433, opened 2021-01-28
Loadable: Use the correct url proprety from preAssets, not chunkname

gtrufitt:
## What does this change?

A chunkname will have multiple paths with webpack code splitting. I.e Atoms split into a small bundle and bigger atoms bundle - we need to load both but the `chunk` will always be just the original chunk, whereas we need the 'url' of the asset that is required.

This fixes a bug that prevented splitting on atoms.

## Also

Moving the url lib to main lib, not web lib as want to use it in assets.


oliverlloyd (approved):
👍

PR #2434, opened 2021-01-29
Special ArticleTitle

oliverlloyd:
## What?
Adds support for special reports to `ArticleTitle`

### Before
Screenshot 2021-01-29 at 09 13 53

### After
Screenshot 2021-01-29 at 09 10 30

### DIFF
--- a/src/web/components/SeriesSectionLink.tsx
+++ b/src/web/components/SeriesSectionLink.tsx
@@ -26,9 +26,10 @@ const sectionLabelLink = css`
 `;
 
 const rowBelowLeftCol = css`
-	display: flex;
 	flex-direction: column;
+	display: inline-block;
 	${until.leftCol} {
+		display: flex;
 		flex-direction: row;
 	}
 `;
@@ -43,8 +44,9 @@ const primaryStyle = css`
 	${from.leftCol} {
 		${headline.xxsmall({ fontWeight: 'bold' })};
 	}
-
-	padding-right: ${space[2]}px;
+	${until.leftCol} {
+		margin-right: ${space[2]}px;
+	}
 `;
 
 const invertedStyle = css`
@@ -126,6 +128,12 @@ export const SeriesSectionLink = ({
 											color: ${palette.text.seriesTitle};
 											background-color: ${palette
 												.background.seriesTitle};
+											box-shadow: -6px 0 0 0
+													${palette.background
+														.seriesTitle},
+												6px 0 0 0
+													${palette.background
+														.seriesTitle};
 										`,
 									)}
 									data-component="series"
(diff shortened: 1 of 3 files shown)


src/web/components/SeriesSectionLink.tsx
```diff
@@ -26,9 +26,10 @@ const sectionLabelLink = css`
 `;
 
 const rowBelowLeftCol = css`
-	display: flex;
 	flex-direction: column;
+	display: inline-block;
```
oliverlloyd, line 30: This prevents the yellow background stretching to the right

src/web/components/SeriesSectionLink.tsx
```diff
@@ -43,8 +44,9 @@ const primaryStyle = css`
 	${from.leftCol} {
 		${headline.xxsmall({ fontWeight: 'bold' })};
 	}
-
-	padding-right: ${space[2]}px;
+	${until.leftCol} {
+		margin-right: ${space[2]}px;
+	}
```
oliverlloyd, line 49: This spacing is much more conspicuous when there's a background on the series title but not the section title, hence margin, and hence why we only want it if these two things are on one row

src/web/components/SeriesSectionLink.tsx
```diff
@@ -126,6 +128,12 @@ export const SeriesSectionLink = ({
 											color: ${palette.text.seriesTitle};
 											background-color: ${palette
 												.background.seriesTitle};
+											box-shadow: -6px 0 0 0
+													${palette.background
+														.seriesTitle},
+												6px 0 0 0
+													${palette.background
+														.seriesTitle};
```
oliverlloyd, line 136: I'm applying this everywhere so that we sematically stretch all backgrounds 6px either side but when transparent this has no noticable effect.

I'm also keeping it verbose and inline because I expect this component to be refactored soon and want to make that easier

src/web/lib/decidePalette.ts
```diff
@@ -13,6 +14,7 @@ const WHITE = neutral[100];
 const BLACK = text.primary;
 
 const textHeadline = (format: Format): string => {
+	if (format.theme === Special.SpecialReport) return specialReport[300];
```
oliverlloyd, line 17: This could grow into a nested switch to make it more scalable bot for now this simple exception seems logical

src/web/lib/decidePalette.ts
```diff
@@ -76,7 +81,16 @@ const backgroundSeriesTitle = (format: Format): string => {
 	}
 };
 
-const backgroundSectionTitle = backgroundSeriesTitle;
+const backgroundSectionTitle = (format: Format): string => {
```
oliverlloyd, line 84: The section title never has a yello background, just the series title

nitro-marky (approved):

liywjl (approved):
👍

gtrufitt:
Inline comment explanations are useful, thanks 👍

PR #2435, opened 2021-01-29
Change Onward Description link colour and add story

nitro-marky:
## What does this change?

Changes the link colour in the onwards description from blue to text.primary. It also removes text decoration. 

A story has also been added for confirmation.

### Before

Screenshot 2021-01-29 at 14 04 33

### After

Screenshot 2021-01-29 at 13 58 20

## Why?
Parity

### DIFF
--- a/fixtures/onwards.mocks.ts
+++ b/fixtures/onwards.mocks.ts
@@ -217,7 +217,7 @@ export const withLongDescription: OnwardsType = {
 	pillar: Pillar.News,
 };
 
-export const withLongDescriptionAndLink: OnwardsType = {
+export const withLink: OnwardsType = {
 	description:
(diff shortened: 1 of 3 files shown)


fixtures/onwards.mocks.ts
```diff
@@ -216,3 +216,12 @@ export const withLongDescription: OnwardsType = {
 	ophanComponentName: 'more-on-this-story',
 	pillar: Pillar.News,
 };
+
+export const withLongDescriptionAndLink: OnwardsType = {
```
oliverlloyd, line 220: Maybe just call this `withLink`? Simpler. You can perhaps make the text shorter to be accurate!

oliverlloyd (approved):
Nice!

fixtures/onwards.mocks.ts
nitro-marky: Good Shout

PR #2436, opened 2021-01-29
Refactor ShareIcons

oliverlloyd:
## What does this change?
Moves logic out of abstracted functions using loops / reduce into a more explicit pattern where we just inline the html we need to render the share icons.

## Why?
There's nothing wrong with loops but more explicit code is easier to maintain and move around and some of the work we've been doing on atoms-rendering and special reports highlighted that a simpler style here might be more suited to our use case

### What happened to clear left?
The icons in the sub meta were using this css but it doesn't appear to have any effect and passing class names is problematic (because they are unknowns) so I've removed the `className` prop and the use of this css

```typescript
const subMetaSharingIcons = css`
	:after {
		content: '';
		display: block;
		clear: left;
	}
`;
```

### DIFF
--- a/src/web/components/ShareIcons.tsx
+++ b/src/web/components/ShareIcons.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { css, cx } from 'emotion';
+import { css } from 'emotion';
 
 import { border } from '@guardian/src-foundations/palette';
 import { from } from '@guardian/src-foundations/mq';
@@ -17,7 +17,7 @@ import MessengerIcon from '@frontend/static/icons/messenger.svg';
 
 import { Hide } from './Hide';
 
-const shareIconList = css`
+const ulStyles = css`
 	float: left;
 	${from.wide} {
 		flex: auto;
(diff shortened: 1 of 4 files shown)


oliverlloyd:
[Refactor SharingIcons](https://trello.com/c/oXPPbRgt/2391-refactor-sharingicons)

liywjl (approved):
👍

src/web/components/ShareIcons.tsx
```diff
@@ -59,102 +59,131 @@ const shareIcon = (colour: string) => css`
...
+				</li>
+			)}
+
+			{displayIcons.includes('twitter') && (
+				<li className={liStyles} key="twitter">
+					<a
+						href={`https://twitter.com/intent/tweet?text=${webTitle.replace(
+							/Leave.EU/gi,
+							'Leave. EU',
+						)}&url=https://www.theguardian.com/&CMP=share_btn_tw`}
```
buck06191, line 97: I think the URL here is broken. It's always linking back to the homepage

oliverlloyd: 👏 for spotting this, that turned out to be an issue on the site. Wish I hadn't merged this in so quickly!

PR #2438, opened 2021-01-29
isSpecialReport

oliverlloyd:
## What does this change?
Makes everything grey with occasional bursts of yellow

## Why?
Because it's special

### DIFF
--- a/fixtures/CAPI/CAPI.ts
+++ b/fixtures/CAPI/CAPI.ts
@@ -3742,4 +3742,5 @@ export const CAPI: CAPIType = modifyDCRJsonResponse({
 		isNumberedList: false,
 	},
 	sectionLabel: 'Ticket prices',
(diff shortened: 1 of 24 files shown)


oliverlloyd:
[isSpecialReport in DCR](https://trello.com/c/KmiCabIF/2395-isspecialreport-in-dcr)

liywjl (approved):
a special thank you, goes to ... @oliverlloyd  ^^ 👍

oliverlloyd:
It should be noted that we want to migrate away from this `isSpecialReport` property on the CAPI model to having `Format` defined higher up the stack so that we already know if an article is a special report or not. This could initially be set on Frontend but would ideally be done by CAPI. Once this work is done then most of these changes become redundant and we can remove `isSpecialReport` from the model.

PR #2445, opened 2021-02-01
Encode title and url in share urls

oliverlloyd:
## What does this change?
Adds functions to encode the title and url params of share urls

## Why?
So that characters like `|` don't cause problems Eg. https://www.theguardian.com/football/2021/feb/01/mattia-destro-genoa-journeyman-italy-new-schillaci-euros-serie-a

### DIFF
--- a/src/web/components/ShareIcons.tsx
+++ b/src/web/components/ShareIcons.tsx
@@ -65,6 +65,13 @@ const iconStyles = (pillar: Theme) => css`
 	}
 `;
 
+const encodeUrl = (pageId: string): string => {
+	return encodeURIComponent(`https://www.theguardian.com/${pageId}`);
+};
+
+const encodeTitle = (webTitle: string): string => {
+	return encodeURIComponent(webTitle.replace(/Leave.EU/gi, 'Leave. EU'));
+};
 export const ShareIcons: React.FC<{
 	pageId: string;
 	webTitle: string;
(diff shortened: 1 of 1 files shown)


src/web/components/ShareIcons.tsx
```diff
@@ -109,7 +117,9 @@ export const ShareIcons: React.FC<{
 			{displayIcons.includes('email') && (
 				<li className={liStyles} key="email">
 					<a
-						href={`mailto:?subject=${webTitle}&body=https://www.theguardian.com/${pageId}&CMP=share_btn_link`}
+						href={`mailto:?subject=${webTitle}&body=${encodeUrl(
```
jfsoul, line 120: `encodeTitle` here?

shtukas (approved):

src/web/components/ShareIcons.tsx
shtukas: Seconded

shtukas:
+1

src/web/components/ShareIcons.tsx
oliverlloyd: I've added encoding for the subject line of emails as well. I tested this locally and it worked before, but also works after and I think the added security of encoding is definately what we want

src/web/components/ShareIcons.tsx
```diff
@@ -65,6 +65,13 @@ const iconStyles = (pillar: Theme) => css`
 	}
 `;
 
+const encodeUrl = (pageId: string): string => {
+	return encodeURIComponent(`https://www.theguardian.com/${pageId}`);
+};
+
+const encodeTitle = (webTitle: string): string => {
+	return encodeURIComponent(webTitle.replace(/Leave.EU/gi, 'Leave. EU'));
```
mxdvl, line 73: Could we opt for `&period;` &period; instead of an invisible character?

PR #2446, opened 2021-02-01
Special Icons

oliverlloyd:
## What does this change?
We now use palette for colouring share icons, which means we support icons with special report styling

### Before
Screenshot 2021-02-01 at 15 42 08

### After
Screenshot 2021-02-01 at 15 26 19


oliverlloyd:
[Special Icons](https://trello.com/c/WZamTena/2397-special-icons)

shtukas (approved):

PR #2447, opened 2021-02-01
Refactor pillarPalette to support specialReport

oliverlloyd:
## What does this change?
Makes `pillarPalette` a little more explicit and verbose by inlining and repeating some code

## Why?
Although this creates a larger, more repetitive piece of code, I hope that this newer pattern will be a lot easier to maintain and edit. The was already an attempt to make the original abstraction handle `labs` colours, which are different, and then, when trying to add support for `specialReport` I got stuck because the Source repo does no expose a value for the `600` key on this particular palette.

Handling this edge case inside the abstraction was confusing, with typescript making overrides difficult. Instead, by moving the abstraction back inline I was able to handle special reports being different.


liywjl (approved):
👍

shtukas (approved):

PR #2448, opened 2021-02-01
Use an 'immersive' card style for Carousel test

buck06191:
This adds a `isFullCardImage` prop that is used to create an 'immersive'
card component for use in the Carousel AB test.

This will definitely need refactoring  at a future date to make this
much easier to read and handle, but for an AB test this _should_
consolidate all of the design changes required into a single commit that
can be reverted if not needed. If this gets taken into production, we
will probably want to update this as part of the work to rationalise
card designs.

### Before
image

### After
image

## Why?
Required for Carousel AB test

### DIFF
--- a/src/web/components/Card/components/CardAge.tsx
+++ b/src/web/components/Card/components/CardAge.tsx
@@ -9,32 +9,47 @@ import ClockIcon from '@frontend/static/icons/clock.svg';
 
 import { makeRelativeDate } from '@root/src/web/lib/dateTime';
 import { decidePillarLight } from '@root/src/web/lib/decidePillarLight';
+import { space } from '@guardian/src-foundations';
+import { until } from '@guardian/src-foundations/mq';
 
-const ageStyles = (design: Design, pillar: Theme) => css`
-	${textSans.xsmall()};
-	color: ${design === Design.Live ? decidePillarLight(pillar) : neutral[60]};
+const ageStyles = (
+	design: Design,
+	pillar: Theme,
+	isFullCardImage?: boolean,
+) => {
+	// This is annoying but can't apply SVG color otherwise
+	const smallImageSvgColor =
+		design === Design.Live ? decidePillarLight(pillar) : neutral[46];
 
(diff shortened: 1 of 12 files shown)


buck06191:
[📦 Create variant A carousel](https://trello.com/c/Q3XJVrJG/2335-%F0%9F%93%A6-create-variant-a-carousel)

shtukas (approved):

src/web/components/Card/components/CardAge.tsx
```diff
@@ -9,32 +9,47 @@ import ClockIcon from '@frontend/static/icons/clock.svg';
...
-	${textSans.xsmall()};
-	color: ${design === Design.Live ? decidePillarLight(pillar) : neutral[60]};
+const ageStyles = (
+	design: Design,
+	pillar: Theme,
+	isFullCardImage?: boolean,
+) => {
+	// This is annoying but can't apply SVG color otherwise
+	const smallImageSvgColor =
+		design === Design.Live ? decidePillarLight(pillar) : neutral[46];
```
jfsoul, line 22: there's some similar logic for color on line 26 which we could pull out like this.

I also spot some additional logic on line 39 which might make sense to include here so it's in one place?

src/web/components/Card/components/CardFooter.tsx
```diff
@@ -54,7 +69,9 @@ export const CardFooter = ({ design, age, mediaMeta, commentCount }: Props) => {
 
 	if (age) {
 		return (
-			<footer className={spaceBetween}>
+			<footer
+				className={isFullCardImage ? fullCardImageAge : spaceBetween}
```
jfsoul, line 73: Not entirely sure, but does the `spaceBetween` name still make sense when it's an alternative to `fullCardImageAge`?

jfsoul (approved):
I think this makes sense for a test, but agree that we could branch at a higher level if we give some more thought to the model for cards in general. I don't think this is necessarily the time to do it in this PR though.

PR #2449, opened 2021-02-01
Add basic sticky nav

nicl:
The nav is sticky until the end of the main content area (onwards begins).

This is a basic implementation but on the plus side, it is little code and no custom JS.

## What does this change?

This is a WIP PR to showcase a basic sticky nav. Note, it is not ready for review.

### Before

(Nav stays at top.)

### After

Kapture 2021-02-01 at 15 12 34

Update, also a JS version that handles scrolling backwards (see nicl/sticky-nav-2 branch for code):

Kapture 2021-02-02 at 14 51 51

There are some disadvantages to this version though:

* custom JS to maintain (and perf impact)
* difficult to work out when initial header is out of viewport (I've fudged this for now but we could attempt a more correct version if we want to go down this route)

## Why?

https://trello.com/c/fy5HsbwD/2347-sticky-nav-test (Idea is that sticky nav will increase nav engagement)

The idea is that the this will aid discovery. A key question then is which content to place in the sticky navbar.

### DIFF
--- a/src/web/layouts/StandardLayout.tsx
+++ b/src/web/layouts/StandardLayout.tsx
@@ -51,6 +51,7 @@ import {
 	SendToBack,
 	BannerWrapper,
 } from '@root/src/web/layouts/lib/stickiness';
+import { neutralBorder } from '@root/src/lib/pillars';
 
 const gridTemplateWide = css`
 	grid-template-areas:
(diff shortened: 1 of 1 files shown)


nicl:
Closing as going to do multiple PRs and merge with JS version.

PR #2453, opened 2021-02-02
Special Caption

oliverlloyd:
## What does this change?
Refactors `Caption` to colour things using `pallete`.

Screenshot 2021-02-02 at 12 26 05

### Along the way
To make this possible, we're prop drilling `format` and `palette` and because a lot of these files haven't been migrated to this new pattern yet there's a lot of diffs here, but hopefully this will calm down as we add these props more


oliverlloyd:
[Special Caption](https://trello.com/c/RjuY0N8k/2405-special-caption)

jfsoul (approved):
🙏  nice one

liywjl (approved):
👍

PR #2454, opened 2021-02-02
Make the `Card.stories.tsx` url much more palatable

buck06191:
## What does this change?
Change the URL in the stories for `Card` to something nicer

## Why?
Because the world is dark enough as it is. Eat more tasty food


oliverlloyd (approved):
Yum 👍

gtrufitt (approved):
🥦

PR #2455, opened 2021-02-02
Track subnav more and add basic tracking doc

nicl:
### Before

No DCR tracking of the subnav 'More' toggle.

Also adds some basic docs on tracking. The real solution here though is to document `tracker-js`.

### After

Screenshot 2021-02-02 at 16 12 58

## Why?

We want to understand how many users click this on DCR.

### DIFF
--- a/docs/development/track-user-behaviour.md
+++ b/docs/development/track-user-behaviour.md
@@ -1,0 +1,26 @@
+# Track user behaviour
+
+*Note, the key library here - `tracker-js` is not well-documented at the time of
+writing, but the hope is that this will improve soon.*
+
+## Track clicks:
+
+* add a data-link-name attribute to your component and the `ophan-tracking.js`
+  script will automatically send click data to Ophan.
+
+The code that does this is
+[here](https://github.com/guardian/ophan/blob/75b86abcce07369c8998521399327d436246c016/tracker-js/assets/coffee/ophan/click-path-capture.coffee)
+if you're interested in seeing the nitty gritty.
+
+## Track anything else
+
+Have a look at
+[tracker-js](https://github.com/guardian/ophan/tree/master/tracker-js/assets/coffee/ophan)
+as there are a few existing helpers which you can use. For example, the
+`data-component` attribute can be used to track attention time.
+
+Unfortunately, this library is not properly documented (yet) so it may be
+easiest to ask around to find out what is available.
+
+If `tracker-js` doesn't do what you need, you can fire your own Ophan events
+using the helpers in `ophan.ts`.
--- a/src/web/components/SubNav/SubNav.tsx
+++ b/src/web/components/SubNav/SubNav.tsx
@@ -163,7 +163,7 @@ export const SubNav = ({ subNavSections, format, currentNavLink }: Props) => {
 		} else {
 			setShowMore(false);
 		}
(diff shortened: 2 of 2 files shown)


gtrufitt (approved):

docs/development/track-user-behaviour.md
```diff
@@ -0,0 +1,26 @@
+# Track user behaviour
+
+*Note, the key library here - `tracker-js` is not well-documented at the time of
+writing, but the hope is that this will improve soon.*
+
+## Track clicks:
+
+* add a data-link-name attribute to your component and the `ophan-tracking.js`
```
buck06191, line 8: I think you need `data-component` as well to make sure that you get the [`referringComponent`](https://github.com/guardian/ophan/blob/75b86abcce07369c8998521399327d436246c016/tracker-js/assets/coffee/ophan/click-path-capture.coffee#L45) as well.

src/web/components/SubNav/SubNav.tsx
```diff
@@ -220,6 +220,7 @@ export const SubNav = ({ subNavSections, format, currentNavLink }: Props) => {
 				<button
 					onClick={() => setIsExpanded(!isExpanded)}
 					className={showMoreStyle}
+					data-link-name="nav2 : subnav-toggle"
```
buck06191, line 223: Could be wrong here, but it might be worth sticking a `data-component` on here as well. It should send a component event up to Ophan as well then. I think it depends if it's already declared somewhere else higher up the tree.

buck06191 (approved):
LGTM but I've added some notes around `data-component`. I can't remember exactly when and where it's used, but if you try adding it in and checking the network tab, you should see the component information going up to Ophan which you can then potentially compare against the [thrift docs](https://dashboard.ophan.co.uk/docs/thrift/componentevent.html#Struct_ComponentEvent)

src/web/components/SubNav/SubNav.tsx
```diff
@@ -163,7 +163,7 @@ export const SubNav = ({ subNavSections, format, currentNavLink }: Props) => {
 		} else {
 			setShowMore(false);
 		}
-	}, [ulRef, setShowMore]);
+	}, [ulRef]);
```
oliverlloyd, line 166: Good call. I _think_ we can also remove `ulRef` from the deps here too? I don't have a complete mental model for what should go here though but I notice the lint rule is okay with it not being present

oliverlloyd (approved):
👏  for the extra work on the documentation

src/web/components/SubNav/SubNav.tsx
nicl: ah yes - I think these are guaranteed by React to not change so both can be removed.

src/web/components/SubNav/SubNav.tsx
nicl: Thanks. Yeah doesn't look like one on Frontend, but I've added it now as does seem useful.

There's also a `data-cy` attribute, but I'm not sure what that does.

buck06191: I think `data-cy` is used in [Cypress testing](https://docs.cypress.io/guides/references/best-practices.html#How-It-Works) so it would come from that I think

PR #2456, opened 2021-02-02
New carousel layout

buck06191:
## What does this change?
Implements the carousel layout changes needed for the AB test.

### Before
image

image

### After
image

image

## Why?

Required to work with the new designs implemented in previous PRs: #2448 #2416


src/web/components/CardHeadline.tsx
```diff
@@ -131,7 +131,9 @@ export const CardHeadline = ({
 
 				<span
 					className={css`
-						color: ${palette.text.cardHeadline};
+						color: ${isFullCardImage
+							? neutral[100]
+							: palette.text.cardHeadline};
```
oliverlloyd, line 136: Can we make this change in `decidePalette`? Sorry to nag and I appreciate that we originally started with `isFullCardImage` and the timing for introducing `palette` at the same time as you're doing this work isn't ideal. But I'm actually all over the codebase doing the same thing and I'm quite keen for future changes to adopt this pattern.

In `Carousel.tsx` I added the line

```
if (isFullCardImage) trails = convertToImmersive(trails);
```

which means that each `trail` after that point has `format.display` set to `Display.Immersive` and this is the value used to decide the `palette` property. If you then look at `decidePalette` there's a line

```
if (format.display === Display.Immersive) return WHITE;
```

which is where this decision about the headline colour to use for immersive cards.

oliverlloyd: By the way, I _think_ this is already set to white so I'm not sure if this change does anything?

HarryFischer:
Nice!

oliverlloyd (approved):
I think as a post step we should be able to refactor away most of these `isFullCardImage` props by using `format` and `format.display === Display.Immersive` but in the interests of getting you moving Josh lets worry about that later when we have more idea where we're going with Carousels and immersive cards.

src/web/components/CardHeadline.tsx
buck06191: Ah yes. I added that line in before running storybook so I added it in and didn't realise it was superfluous. I've removed it now which makes it much neater

buck06191:
> I think as a post step we should be able to refactor away most of these `isFullCardImage` props by using `format` and `format.display === Display.Immersive` but in the interests of getting you moving Josh lets worry about that later when we have more idea where we're going with Carousels and immersive cards.

Cheers. I definitely think that once we've determined what's best here we'll need to go in and tidy up a lot of this. _Hopefully_, the decisions around Immersive style cards will align with what's needed for a carousel if that's the direction we go, but we'll handle it then and there when we come to it.

PR #2463, opened 2021-02-03
Fix merge issue on ClickToView stories

oliverlloyd:
## What does this change?
After a merge, typescript started complaining about issues in the `ClickToView` stories on `main`. I suspect this was where we don't rerun our tests when we merge in `main` on PRs. We know about this risk and accept it in return for the faster turnaround in 99% of cases.

As part ofthe fix for the typescript errors I removed some abstractions from these stories because, upon checking, I saw that we actually don't use these functions (the props are always the same and each function is only used once) and by inlining them it makes maintenance easier going forwards. We're actively making changes and often  prop drilling so we want these kinds of updates to be as smooth as possible.

### DIFF
--- a/src/web/components/ClickToView.stories.tsx
+++ b/src/web/components/ClickToView.stories.tsx
@@ -2,11 +2,13 @@
 import React from 'react';
 import { css } from 'emotion';
 import { textSans } from '@guardian/src-foundations/typography';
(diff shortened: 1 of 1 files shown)


shtukas (approved):

gtrufitt (approved):

liywjl (approved):
👍

PR #2464, opened 2021-02-03
Add new URLs for AMP Validation Action

gtrufitt:
## What does this change?

We want to remove the AMP validation TC runner, but increase the coverage at PR time.


oliverlloyd (approved):
More coverage 👍

PR #2465, opened 2021-02-03
Sticky nav 2 (JS version with backscroll behaviour)

nicl:
WIP - still to add test harness code.

### Before

(No sticky nav behaviour on scroll.)

### After

Kapture 2021-02-03 at 10 34 25

## Why?

Increase engagement through nav visibility. Idea is that backwards scroll shows intent to access nav and/or journey to another piece of content.

## Tech questions

*Update:* on the increased size of `window.guardian` I've measured this, and on a standard article (https://www.theguardian.com/world/2020/feb/10/irish-general-election-everything-you-need-to-know) we see:

    before: 46kb
    after: 67kb

which is quite a big difference. However, it doesn't change the lighthouse score (81 when I ran it locally for both).

* is the increase in data on client justified (we have to send entire nav now)?
* is the user-benefit enough to justify extra complexity compared to basic CSS version?
* more generally, noticed that our client-side and server-side models often differ and this causes difficulties when writing code that must work with both (typically you have to use the client-side models and then adapt, but a lot of our utility functions rely on the server-side models).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -698,7 +698,7 @@ interface DCRBrowserDocumentData {
 	page: string;
 	site: string;
 	CAPI: CAPIBrowserType;
-	NAV: SubNavBrowserType;
+	NAV: NavType;
 	GA: GADataType;
 	linkedData: object;
 }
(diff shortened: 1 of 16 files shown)


nicl:
Closing as splitting some of this out into separate PRs now I have a better sense of things (+feedback).

PR #2466, opened 2021-02-03
Ever so slightly darker special text

oliverlloyd:
## What does this change?
Makes some text very slightly darker. `specialReport[100]`  instead of `specialReport[300]` or `specialReport[200]`

### Before
Screenshot 2021-02-03 at 10 50 49

### After
Screenshot 2021-02-03 at 10 49 18

## Why?
Better contrast and parity


oliverlloyd:
[Darker Special Text](https://trello.com/c/n69pSe9n/2417-darker-special-text)

shtukas (approved):

buck06191:
Just looking at the screenshots above, do you think we can make the caption text slightly darker? The contrast there is almost definitely not accessible

oliverlloyd:
> Just looking at the screenshots above, do you think we can make the caption text slightly darker? The contrast there is almost definitely not accessible

Yeah, there's an open thread on this in the * rendering channel. I'd like to see us review special reports to add more contrast in general, esp. on in body links. I think you're right here  though, this should be darker

oliverlloyd:
> > Just looking at the screenshots above, do you think we can make the caption text slightly darker? The contrast there is almost definitely not accessible
> 
> Yeah, there's an open thread on this in the * rendering channel. I'd like to see us review special reports to add more contrast in general, esp. on in body links. I think you're right here though, this should be darker

I've  fixed this in https://github.com/guardian/dotcom-rendering/pull/2469. There was actually a small bug here so thanks for pointing this put!

PR #2468, opened 2021-02-03
Special SubNav Palette

oliverlloyd:
## What does this change?
Adds the `palette` pattern to `SubNav`

## Why?
To continue the [editorial palette](https://docs.google.com/document/d/1nkntFJUunqrwNYszTrc2RBfyYUd1bEAeQh76yVFOLjo/edit#) work and, nominally, to support special reports although in this case the support is already there because we're using `pillarPalette` via `format` and we don't do anything different so specials other than use their `main` colour

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -64,6 +64,9 @@ type Palette = {
 		shareIcon: Colour;
 		captionTriangle: Colour;
 	},
+	border: {
+		subNav: Colour;
+	}
 };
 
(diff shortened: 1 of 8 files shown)


shtukas (approved):

oliverlloyd:
[Special Nav](https://trello.com/c/wsltQqVk/2403-special-nav)

PR #2470, opened 2021-02-03
Add zero width space before list-style: none list items to fix accessiblity issue in safari

gtrufitt:
## What does this change?

As per https://developer.mozilla.org/en-US/docs/Web/CSS/list-style#accessibility_concerns there is an issue with Safari where they don't recognise `ul`'s with `list-style: none` as lists for assistive devices. This PR fixes that. I *think* this would be the correct approach (mainly applying to the navigation, where we want a 'list' of 'links'.

cc/ @SiAdcock @sndrs


oliverlloyd (approved):

oliverlloyd:
Those tests seem unfairly upset

gtrufitt:
> Those tests seem unfairly upset

they are being a little dramatic yes...

gtrufitt:
This seems to be related to the tooling setup https://github.com/emotion-js/emotion/issues/1354. Rather than spend time trying to fix now, I am going to return to this after the TS and Emotion upgrades.

oliverlloyd:
> This seems to be related to the tooling setup [emotion-js/emotion#1354](https://github.com/emotion-js/emotion/issues/1354). Rather than spend time trying to fix now, I am going to return to this after the TS and Emotion upgrades.

I have a feeling @liywjl knows why this is happening - I think he saw something similar earlier this week?

sndrs:
i think it should _really_ be double dots https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements#syntax

gtrufitt:
> i think it should _really_ be double dots [developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements#syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements#syntax)

Yeah, I thought that was what emotion was choking on for some reason, will revert back to the double dots

PR #2472, opened 2021-02-03
Refactor SubMeta

oliverlloyd:
## What does this change?
Refactors `SubMeta` to open up using the `palette`, and then actually uses `palette`.

I've inlined both `SyndicationButton` and `SubMetaLinksList` to make adding palette easier and also because I wanted to use Source for the button as a drive by improvement and once we have that in place the extra component wasn't really needed anymore.

Another change I made here was to replace `neutral[60]` with the darker `text.supporting`. This is both an improvement semantically (the name is descriptive) but the darker shade makes this text more accessible. This can be seen on 'Topics' and on the '/'s between links.

### Before
Screenshot 2021-02-03 at 12 36 36

### After
Screenshot 2021-02-03 at 12 36 09

and

Screenshot 2021-02-03 at 12 49 14


oliverlloyd:
[Special SubMeta](https://trello.com/c/ebkmRGt8/2399-special-submeta)

shtukas (approved):

PR #2473, opened 2021-02-03
Special Links

oliverlloyd:
## What does this change?
Uses the `palette`  pattern to decide colours for in bodt article links

## Special Reports
The links for these have been restyled to give more contrast and improve accessibility


oliverlloyd:
[Special body links](https://trello.com/c/tbz611v7/2398-special-body-links)

shtukas (approved):

PR #2474, opened 2021-02-03
Embedded overlay for DocumentBlockComponents

frj:
This PR enables the ClickToView component for the DocumentBlockElements. Requiring the user to click on the component to 'consent' to being tracked by the third party provider that hosts the document, eg Scribd.

For more details on click to view see the PR:

https://github.com/guardian/dotcom-rendering/pull/2408

The ClickToView will initially be behind a feature switch. This has been implemented as a 0% server side ab test. 

This was implemented a server side test as each variant requires a different page layout. and running this AB test in the client would cause layout changes during page load if the user is in a segment that requires a different layout to the one that was used to render the page server side.

### Before

Screenshot 2021-02-08 at 16 27 48

### After
Screenshot 2021-02-08 at 16 24 51

## Why?
This is ultimately part of the Embeds Consent project that is aimed at giving our users choice about when they get tracked by third party providers of embedded content.


src/web/components/ClickToView.test.tsx
```diff
@@ -20,12 +30,28 @@ describe('ClickToView', () => {
...
+				sourceDomain="athirdparty.com"
+				abTests={abTests(false)}
+			>
+				<div data-testid="third-party-content" />
+			</ClickToView>,
+		);
+
+		expect(getByTestId('third-party-content')).toBeInTheDocument();
+	});
+
```
oliverlloyd, line 47: Really readable tests!

src/web/components/ClickToView.tsx
```diff
@@ -14,20 +14,21 @@ type Props = {
...
 		case 'inline':
 		case 'showcase': {
-			return textSans.small();
+			return textSans.medium();
 		}
 		case 'halfWidth':
 		case 'supporting':
 		case 'thumbnail': {
-			return textSans.xsmall();
+			return textSans.small();
```
oliverlloyd, line 31: Is it worth making these design changes separately? I'd be interested to check how these new font sizes render with different roles and with different sizes

src/web/components/ClickToView.tsx
```diff
@@ -77,26 +78,51 @@ const roleButtonText = (role: RoleType) => {
...
+	isOverlayClicked: boolean,
+	isInABTestVariant: boolean,
+) => {
+	if (!isTracking) {
+		return false;
+	}
+	if (isOverlayClicked) {
+		return false;
+	}
+	return isInABTestVariant;
```
oliverlloyd, line 92: This reads really nice 👍

src/web/components/ClickToView.tsx
```diff
@@ -77,26 +78,51 @@ const roleButtonText = (role: RoleType) => {
...
 	role = 'inline',
 	onAccept,
 	isTracking,
 	source,
-	sourceDomain,
+	sourceDomain = 'unknown',
+	abTests,
 }: Props) => {
-	const [showOverlay, setShowOverlay] = useState<boolean>(true);
+	const [isOverlayClicked, setOverlayClicked] = useState<boolean>(false);
```
oliverlloyd, line 108: There's a loose convention I think where for state hooks you have `thing`, `setThing` so maybe here we can do `isOverlayClicked`, `setIsOverlayClicked`? Or maybe use `overlayClicked`? But, to be honest, I can't think of a good reason why this should be so, other than this just seems to be where we've ended up with hook naming 🤷 , so this is very much a take it or leave it comment

src/web/lib/ArticleRenderer.tsx
```diff
@@ -167,17 +169,29 @@ export const ArticleRenderer: React.FC<{
 							</>
 						</Figure>
 					);
-				case 'model.dotcomrendering.pageElements.DocumentBlockElement':
+				case 'model.dotcomrendering.pageElements.DocumentBlockElement': {
```
oliverlloyd, line 172: Can we remove the enclosing braces here? Not because I especially care about this formatting but because all the other case statements in this switch don't use them and so this could present an issue later on if someone did a multi select during a refactor and wasn't expecting this pattern.

oliverlloyd (approved):
This is amazing work! This is a really complicated feature to add but the code ended up looking really simple which is an impressive result.

I also really like how it will be simple and easy to remove the ab test code later 👍

src/lib/content.d.ts
```diff
@@ -114,6 +114,10 @@ interface DocumentBlockElement {
 	width: number;
 	title?: string;
 	role?: RoleType;
+	isThirdPartyTracking: boolean;
+	source?: string;
+	sourceDomain?: string;
+	documentIndex?: number;
```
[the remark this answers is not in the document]
frj: I know what you mean just adding this field onto the block element seems a bit clunky but it follows the pattern for all of the other block elements that get hydrated, eg AudioAtomBlockElement has an AudioIndex field that does not exist in the json from frontend and gets populated by this function:

https://github.com/guardian/dotcom-rendering/blob/faadaac386a6785805ff27665dc17a0e64f54e5f/src/model/window-guardian.ts#L72

src/web/components/ClickToView.tsx
frj: So these changes actually implement the original designs, there was some inline styling in the story books which was interfering with the styles from the component. 

Ive reworked the storybooks to fix that problem so you should be able to see the storybooks for the different roles as they would appear in rendered articles.

Without these changes the overlay looks bad :)

src/lib/content.d.ts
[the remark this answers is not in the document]
oliverlloyd: Ah, I see. I agree if this is the pattern then it makes sense to follow it 👍

frj:
If you want to try it out you will need to opt in to the AB test using this link:

https://www.theguardian.com/opt/in/click-to-view

You can then see an example of an embedded Scribd 'Document' here:

https://www.theguardian.com/politics/2020/jul/21/what-does-the-russia-report-mean-for-british-people-and-politics

PR #2475, opened 2021-02-03
Added Higher order timer function

rhystmills:
## What does this change?

Adds a higher order `Timer` class, which can be used to process real user monitoring (RUM) data. The class is agnostic to how the measurements are used - this is specified in a callback passed to the class, for example logging it to the console or sending data to Ophan.

The `Timer` can be used to measure a continuous timeline of program execution, and is initialised with a name prefix and a callback function to be used on measurements. First, the `start()` method is run, then `pip(name: string)` is run each time the developer wants to measure a section of the program and process the data in the manner specified by the callback.

### Before

Braze RUM data is collected using `initPerf` and sent to Ophan with each measurement.

### After

Braze RUM data is collected and set to Ophan using an instance of `Timer`.

## Why?

We had tried to use `initPerf` to generate RUM data for the Braze web integration. However, it was opinionated on how to process user data (logging it to the console) and was geared towards making single measurements - consequently we had to write a lot of code and repeat ourselves in the consuming functions in order to use it for our purpose. We also didn't want to log to the console for other users.

We needed a utility which made a series of measurements, constituting a bigger overall timeline for our program execution, the output of which could be used to monitor the overall flow of the program and help us measure which parts of our program were taking the most time to execute (e.g. calls to Identity or Braze's methods themselves).

Given our experience with

### DIFF
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -104,10 +104,10 @@ const getMessageFromBraze = async (
 			return outcome;
 		})
 		.catch(() => {
-			appboyTiming.clear();
-			// eslint-disable-next-line no-console
-			console.log('Appboy Timing failed.');
 			return { result: false };
+		})
+		.finally(() => {
+			appboyTiming.clear();
 		});
 };
 
--- a/src/web/components/StickyBottomBanner/BrazeBanner.tsx
+++ b/src/web/components/StickyBottomBanner/BrazeBanner.tsx
@@ -4,11 +4,8 @@ import * as emotionCore from '@emotion/core';
 import * as emotionTheming from 'emotion-theming';
 import { getZIndex } from '@root/src/web/lib/getZIndex';
 import { Props as BrazeBannerProps } from '@guardian/braze-components';
-import {
-	submitComponentEvent,
-	record,
-} from '@root/src/web/browser/ophan/ophan';
-import { initPerf } from '@root/src/web/browser/initPerf';
+import { submitComponentEvent } from '@root/src/web/browser/ophan/ophan';
+import { createOphanTimer } from '@root/src/web/lib/timerOphan';
 import {
 	hasCurrentBrazeUser,
 	setHasCurrentBrazeUser,
@@ -41,23 +38,17 @@ const containerStyles = emotion.css`
     ${getZIndex('banner')}
 `;
 
+const timer = createOphanTimer('brazeBanner');
+
 const getMessageFromBraze = async (
 	apiKey: string,
 	brazeUuid: string,
 ): Promise<CanShowResult> => {
-	const sdkLoadTiming = initPerf('braze-sdk-load');
-	sdkLoadTiming.start();
+	timer.pip('canShowComplete');
 
 	const appboy = await getInitialisedAppboy(apiKey);
 
-	const sdkLoadTimeTaken = sdkLoadTiming.end();
-	record({
-		component: 'braze-sdk-load-timing',
-		value: sdkLoadTimeTaken,
-	});
-
-	const appboyTiming = initPerf('braze-appboy');
-	appboyTiming.start();
+	timer.pip('sdkLoaded');
 
 	const brazeMessages = new BrazeMessages(appboy);
 
@@ -94,12 +85,7 @@ const getMessageFromBraze = async (
 			return { result: true, meta };
 		})
 		.then((outcome) => {
-			const appboyTimeTaken = appboyTiming.end();
-
-			record({
-				component: 'braze-appboy-timing',
-				value: appboyTimeTaken,
-			});
+			timer.pip('appboyMethodsComplete');
 
 			return outcome;
 		})
@@ -107,7 +93,7 @@ const getMessageFromBraze = async (
 			return { result: false };
 		})
 		.finally(() => {
-			appboyTiming.clear();
+			timer.clear();
 		});
 };
 
@@ -181,8 +167,7 @@ export const canShow = async (
 	isSignedIn: boolean,
 	idApiUrl: string,
 ): Promise<CanShowResult> => {
-	const bannerTiming = initPerf('braze-banner');
-	bannerTiming.start();
+	timer.start();
 
 	const forcedBrazeMeta = getBrazeMetaFromQueryString();
 	if (forcedBrazeMeta) {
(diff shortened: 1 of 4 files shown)


oliverlloyd (commented):
Isthis similar to [initPerf](https://github.com/guardian/dotcom-rendering/blob/7ae84661b3b6ac357b2720c6cc4aea34932afd56/src/web/browser/initPerf.ts)?

EDIT: Sorry, I just saw your comment about this!

oliverlloyd:
This seems like there's a good use case to extend the existing initPerf function. To help with long term maintainence we probably only want one way to measuer performance in the browser and we've already had issues around using the `window.performance` api in legacy broswers that we need to be careful about.

I think being able to measure client side performance is a hugely important area that we should definately invest more effort in though and I'd be really keen to udnerstand what we're missing and what features we could add to improve things? For instance, I suspect the console logging this function does is redundant and could be removed or put behind a flag and I'm actually raelly keen to think of better ways to surface these numbers so we can track them!

rhystmills:
> This seems like there's a good use case to extend the existing initPerf function. To help with long term maintainence we probably only want one way to measuer performance in the browser and we've already had issues around using the `window.performance` api in legacy broswers that we need to be careful about.
> 
> I think being able to measure client side performance is a hugely important area that we should definately invest more effort in though and I'd be really keen to udnerstand what we're missing and what features we could add to improve things? For instance, I suspect the console logging this function does is redundant and could be removed or put behind a flag and I'm actually raelly keen to think of better ways to surface these numbers so we can track them!

Hey Oliver, thanks for looking over this. I wasn't quite done with the PR yet (hence the unfinished description) and was planning to add some bits to it (including checking for `window.performance` - the purpose was indeed to get feedback rather than launch it straight into production code so thanks for your feedback!

I do feel quite strongly that the use case here is different to what `initPerf` does. `initPerf` is built to take a single measurement - in our application, making a series of linked measurements and processing them ended up looking something like this:

```js
const sdkLoadTiming = initPerf('braze-sdk-load');
sdkLoadTiming.start();

// ... code ...

const sdkLoadTimeTaken = sdkLoadTiming.end();
record({
    component: 'braze-sdk-load-timing',
    value: sdkLoadTimeTaken,
});

const appboyTiming = initPerf('braze-appboy');
appboyTiming.start();

// ... code ...

record({
    value: appboyTimeTaken,
});

const bannerTiming = initPerf('braze-banner');
bannerTiming.start();

// ... code ...

const timeTaken = bannerTiming.end();
record({
    component: 'braze-banner-timing',
    value: timeTaken,
});

sdkloadTimeTaken.clear();
appboyTiming.clear();
bannerTiming.clear();

```
i.e. there was **a lot** of boilerplate to do the same thing three times, and it also didn't achieve what we actually we wanted, which was a taking a series of clearly structured, ordered and related measurements without introducing lots of noise into our code. Using this `Timer` utility, the boilerplate in our consuming code looks like this:

```js
const timer = createOphanTimer('brazeBanner');

// ... code ...

timer.pip('canShowComplete');

// ... code ...

timer.pip('sdkLoaded');

// ... code ...

timer.pip('appboyMethodsComplete');

// ... code ...

timer.clear();
```

I did attempt to morph `initPerf` towards doing what we wanted when we first started making these measurements (leading to it returning a measurement value rather than only logging to the console), but encountered some friction because it is already used in several places to achieve its existing purpose (logging a measurement to the console) and we were reducing the extent to which it had a single responsibility. This PR is really a consequence of my experiences with that, and feeling that our aim was similar (involving performance measurement) but also different enough to warrant a separate solution.

There is shared logic that could be abstracted out of both functions (e.g. checking that the performance API is supported) and I also suspect that `initPerf` could be implemented as a subtype of `Timer`.

rhystmills:
> This seems like there's a good use case to extend the existing initPerf function. To help with long term maintainence we probably only want one way to measuer performance in the browser and we've already had issues around using the `window.performance` api in legacy broswers that we need to be careful about.
> 
> I think being able to measure client side performance is a hugely important area that we should definately invest more effort in though and I'd be really keen to udnerstand what we're missing and what features we could add to improve things? For instance, I suspect the console logging this function does is redundant and could be removed or put behind a flag and I'm actually raelly keen to think of better ways to surface these numbers so we can track them!

Adding to what I wrote earlier; I would agree that it would be good to at least share functionality between the two since they have a lot of overlap. I think directly extending `initPerf` is undesirable because it's used in a lot of places for its existing purpose (console logging) - and our previous attempts to coerce it to our use case has prevented it having a single responsibility. 

I think `initPerf` is really a specific 'logging' implementation of a higher order function which can encapsulate working with (and checking support for) the Performance API. 

From my own work I've found `initPerf` works well for an individual measurement, but not for a timeline of related measurements which we would like to collectively analyse (in our case a waterfall-style breakdown of processes making up the Braze SDK initialisation). So perhaps the higher order function needs to support both the 'individual measurement' and 'timeline' paradigms.

I think the `Timer` class here does some of this, but could do with some tweaking.

oliverlloyd:
Hey @rhystmills Apologies for the premature review, I know that's annoying when someone dumps on your PR before you're ready!

But I was interested to hear that it was the console logging that might have been the reason to move away from `initPerf`  because I don't belive that is especially important, so if it were changed or removed I don't ) _think_ that would impact anything and because of that I thought it worth having the discussion earlier.

Your examples (thanks for those 👍 ) clearly show there's a strong use case for having the ability to record performance in a more advanced way, not just start and end, and I feel like other parts of DCR could benefit a lot from the same approach. Maybe we could go the other way and migrate `initPerf` to use this function? Am I right in thinking that you are logging this performance data to ophan? I wonder if there's any cross-over here with the work @nitro-marky and @gtrufitt are doing to record core vitals performance data in the browser? It would be amazing if we had real user performance data, around hydration for example, stored in a location we could query.

I think we all share a similar goal here (being able to track client side performance in the wild) but it's a tricky thing to get right and it might be something we want to work together on. I get though that sometimes it's just easier to have code that works and that adding the complexity of shared requirements can be a messy pain!

rhystmills:
No worries at all, getting feedback on the idea was pretty much the point of this PR!

We are indeed logging to Ophan, not in a particularly elegant way (I think it's basically dumped into the component tracking on the pageview table each time the Braze SDK is loaded - which only happens for logged in supporters currently) - and we're probably doing this frequently than we need to - I think aggregated/sampled data would probably work for our purposes.

I've been imagining getting to a point where it's really easy to get a breakdown of execution times for the Braze integration - essentially we exceed the 2000ms timeout set by the platform fairly regularly, and it's good to know how much is due to our code (which can be refactored to improve time taken) or Braze's code (in which case we need to chase them, or maybe work with their system in a different way to see improvements).

It might be that we end up with structured data like this:
```json
[
  {"brazeBanner-1-sdkLoad":1000},
  {"brazeBanner-2-appboyInit":500},
  {"brazeBanner-3-bannerRendered":600},
  {"brazeBanner-total":2100},
]
```
... or maybe raw data which we can feed into something providing data visualisation like a waterfall graph.

So that's our context - but it had made me think about measuring execution times in general, and I can see that a detailed breakdown of performance could help developers identify which parts of their code tends to take the longest to execute on user machines.

That work sounds really interesting - certainly another part of this problem is where the data actually gets stored and how we can access/use it. It would be great to work together/pair on this!

PR #2476, opened 2021-02-03
Explicitly type CAPIType.nav

nicl:
This provides extra type safety around fixtures and request body handling. (I hit some issues around typing recently when working on sticky nav that would have been easier to debug with this.)

Note, as we validate request data, it's possible we'll see production errors from this if our model doesn't match data sent. This should show up in logs as `Unable to validate request body for url...` so we can monitor for these on deploy and revert if necessary.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -132,24 +132,12 @@ interface Branding {
 	};
 }
 
-interface CAPILinkType extends SimpleLinkType {
-	longTitle: string;
-	children?: LinkType[];
-	mobileOnly?: boolean;
-	pillar?: CAPIPillar;
-	more?: boolean;
-}
-
 interface LinkType extends SimpleLinkType {
 	longTitle: string;
 	children?: LinkType[];
 	mobileOnly?: boolean;
 	pillar?: Pillar;
 	more?: boolean;
-}
-
-interface CAPIPillarType extends CAPILinkType {
-	pillar: CAPIPillar;
 }
 
 interface PillarType extends LinkType {
@@ -193,10 +181,6 @@ interface BaseNavType {
 
 interface NavType extends BaseNavType {
 	pillars: PillarType[];
-}
-
-interface CAPINavType extends BaseNavType {
-	pillars: CAPIPillarType[];
 }
 
 interface SubNavBrowserType {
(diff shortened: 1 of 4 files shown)


oliverlloyd:
I approved that Vimeo diff, that pops up sometimes - we should think about disabling snapshots on this component.

oliverlloyd:
Ideally, if we had the data typed with `format` already when sent from Frontend, we wouldn't need to have these different sets of types (the 'CAPI' version and then the actualy version with `pillar` in. It gets quite painful trying to track everything because we have so many entrypoints (not just the DCR data but all the api endpoints too).

Agreed though that NAV is one area where typing could be better so 👍

nicl:
@oliverlloyd agreed - it's not great atm. This is a small(ish) improvement but doesn't really tackle the root issues.

index.d.ts
```diff
@@ -275,6 +259,37 @@ type PageTypeType = {
...
+	otherLinks: CAPILinkType[];
+	brandExtensions: CAPILinkType[];
+    currentNavLink?: CAPILinkType;
+    currentParent?: CAPILinkType;
+    currentPillar?: CAPILinkType;
+	subNavSections?: {
+        parent?: CAPILinkType;
+        links: CAPILinkType[];
+    };
+	readerRevenueLinks: ReaderRevenuePositions;
```
liywjl, line 290: some unusual spacing here

index.d.ts
```diff
@@ -275,6 +259,37 @@ type PageTypeType = {
 	isSensitive: boolean;
 };
 
+// Data types for the API request bodies from clients. The 'CAPI' prefix
+// convention doesn't really make sense anymore (it is not the CAPI model,
+// though closely related) but is not worth changing now.
```
oliverlloyd, line 264: This might be worth revisiting. The CAPI prefix is meant to denote data that came from CAPI (or rather, Frontend) and that has a _different_ type to the one eventually passed down the react stack. We now use `decideThing` functions to transform any data we get from api calls. This is mostly related to `CAPIPillar`; the idea was to try to make any later refactor easier if and when Frontend adopts `Format`.

I _think_ that this is also the case here? I just wanted to double check

index.d.ts
nicl: ooh yeah! Thanks. Will fix. Strange my auto-formatting didn't rescue me.

index.d.ts
nicl: @oliverlloyd okay thanks. I'll tweak the comment to include the 'and has a different type to our internal use' caveat.

index.d.ts
nicl: It turned out to be hidden tabs that appeared the same width as 4 spaces locally 😱

oliverlloyd (approved):

liywjl (approved):

index.d.ts
rcrphillips: I've had the same issue with auto-formatting recently. `yarn prettier:fix` should help

PR #2477, opened 2021-02-03
Hydrate using correct nav in storybook layouts

nicl:
This doesn't currently affect things as we don't hydrate the nav, but the (in progress) sticky nav test requires this.


oliverlloyd (approved):
Nice

oliverlloyd:
The storybook bug is because there VSCode has auto inserted an import statement without the `@root` prefix, which is required. Search for `from 'src/`

PR #2478, opened 2021-02-03
Rename BannerPicker to MessagePicker

tjmw:
## What does this change?

This PR renames `BannerPicker` to `MessagePicker`.

### Before

`pickBanner`

### After

`pickMessage`

## Why?

There's nothing banner specific about the picker, it's just that's the only use case currently. However, soon we'd like to start showing end-of-article messages from Braze. This means we'll need something to broker who gets the slot impression, support-dotcom-components epics or Braze (realistically for the first scenarios there won't be any overlap in audiences).


liywjl (approved):
👍

src/web/lib/messagePicker.ts
```diff
@@ -0,0 +1,103 @@
...
+	show: ShowMessage;
+	canShow: () => Promise<CanShowResult>;
+	timeoutMillis: number | null;
+};
+
+type MessageWithTimeout = Message & {
+	cancelTimeout: () => void;
+};
+
+export type MessageConfig = {
```
adamnfish, line 22: Is this really `MessageConfig`, or is it something more like `SlotConfig`? It could even be something quite different like `Channel`?

src/web/lib/messagePicker.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export type MaybeFC = React.FC | null;
+type ShowMessage = (meta?: any) => MaybeFC;
+
+export type CanShowResult = {
+	result: boolean;
+	meta?: any;
+};
+
+export type Message = {
```
adamnfish, line 11: Is this really a `Message`, or is this more like one of the following:
* channel
* message handler
* or even more slot configuration?

src/web/lib/messagePicker.ts
```diff
@@ -0,0 +1,103 @@
+import { record } from '@root/src/web/browser/ophan/ophan';
+
+export type MaybeFC = React.FC | null;
```
adamnfish, line 3: Is this alias offering us enough value to justify its existence? Would the code be clearer if we wrote `React.FC | null` everywhere this appears?

src/web/lib/messagePicker.ts
```diff
@@ -0,0 +1,103 @@
...
+	});
+
+const timeoutify = (message: Message, type: string): MessageWithTimeout => {
+	let timer: number | undefined;
+
+	const canShow = (): Promise<CanShowResult> =>
+		new Promise((resolve) => {
+			if (message.timeoutMillis) {
+				timer = window.setTimeout(() => {
+					recordMessageTimeoutInOphan(message.id, type);
```
adamnfish, line 40: Similarly here, is it the `message` that times out, or is it the provider? I don't feel strongly here but the other comments have made me don my eagle eyes...

adamnfish (commented):
While this is a great first change (and thanks for doing it separately so its a nice PR!) I feel like this brushes past some nomenclature thinking that might be worth doing up front.

As it stands it's a bit harder to follow functions like `pickMessage` then it ought to be, because it's hard to see from the code which bits refer to which concepts - they're mostly all called `message`.

Previously this was just the banner, so everything was just called banner, with no distinction between the concept of the banner, the slot on the page, and the contents that actually get put on the page. This was ok when it really was all "the banner" but gets a bit less clear cut when we're talking about a general "slot filler".

I am super eager for this to not become Automat take-two, but I wonder whether from your experience with that project you're aware of any agreed naming conventions we have that might help make this a bit clearer. We already certainly have "slot" for the space on the page, and I think "message" is probably uncontroversial as the content that gets put in a slot. We seem to lack a word for the logic from competing business units that target a slot, perhaps something like "channel"?

What did Automat call these things, can we use the same here to give more meaningful names to the abstractions we are using?

tjmw:
Hey thanks for the comments @adamnfish! I think you're right, the translation of Banner -> Message in all cases is too simplistic. I think the overall config for the picker, currently called `MessageConfig`, should be `SlotConfig` or similar. I'll definitely take another pass at this to see if the naming can be improved.

Looking back to Automat I don't think we had a single term for the concept of the business unit filling the slot, other than just that ("business unit") or perhaps "provider". Thinking it over I'm not 100% on whether that distinction is useful in the code - i.e. the message vs the business unit. Ultimately it's the message that gets the slot (or not) - is the business unit a useful concept to represent in the code? There's nothing in place to say that a given business unit couldn't have >1 message up for taking the slot, and the picking logic wouldn't need to know that either way.

src/web/lib/messagePicker.ts
tjmw: Yeah agreed I think it's marginal. The reason for it is that there are types later on which build on top of `MaybeFC` which to me look slightly unwieldy when we expand `MaybeFC` i.e.:

`type ShowMessage = (meta?: any) => MaybeFC;` vs `type ShowMessage = (meta?: any) => React.FC | null;`

and

`Promise<() => MaybeFC>` vs `Promise<() => React.FC | null>`

Definitely subjective though.

src/web/lib/messagePicker.ts
tjmw: I really like `SlotConfig` here. I think that reads nicely - the configuration for the slot includes the possible messages we might show in that slot.

src/web/lib/messagePicker.ts
tjmw: Yeah I see what you mean, `Message` isn't quite right. It represents the possibility of showing a message, in this slot.

tjmw:
Hey @adamnfish I've tried to improve some of the naming here to make things clearer. I'd be interested in your thoughts if you have time. Thanks!

src/web/lib/messagePicker.ts
tjmw: Discussed offline that we think we can remove the `null` part of this type completely. It's this way because the CMP rendering exists outside of React's world, and doesn't return a component. However, returning a no-op component from the CMP show method would allow us to simplify the type, which is probably worth doing (in a separate PR though).

src/web/lib/messagePicker.ts
```diff
@@ -0,0 +1,123 @@
...
+
+		const winner: Promise<WinningMessage> = results.reduce(
+			async (winningMessageSoFar, canShow, idx) => {
+				const runningIdx = (await winningMessageSoFar).idx;
+				if (runningIdx >= 0) {
+					return winningMessageSoFar;
+				}
+
+				const result = await canShow;
+				candidateConfigsWithTimeout[idx].cancelTimeout();
```
adamnfish, line 102: This looks like a real "cancel timeout", I'm happy here.

(although as above, it feels like this might be better considered as an internal concern that gets cancelled from the timeoutify wrapper when the canShow completes)

src/web/lib/messagePicker.ts
```diff
@@ -0,0 +1,123 @@
...
+					resolve({ result: false });
+				}, candidateConfig.timeoutMillis);
+			}
+
+			candidateConfig.candidate
+				.canShow()
+				.then((result) => resolve(result));
+		});
+
+	const cancelTimeout = () => timer && clearTimeout(timer);
```
adamnfish, line 60: Here too, I'm a little confused (sorry!)

This cancels the timeouts, but not the actual logic. So all the pending candidates will continue executing as normal. But since the cancel handle is called when a message is selected from the available candidates, should this not also resolve the can show promise with a failure?

Reading ahead, I think my issue is that I followed this from the "clearAllTimeouts" invocation. This makes sense from a "cancel the timeout if `canShow` succeeded" perspective. If *this* is the goal though, why not cancel the timeout in the `then` on line 57? Why are we depending on our caller to tidy up this piece of internal state for us?

src/web/lib/messagePicker.ts
```diff
@@ -0,0 +1,123 @@
...
+					return { idx, meta: result.meta };
+				}
+
+				return winningMessageSoFar;
+			},
+			Promise.resolve({ idx: -1 }),
+		);
+
+		winner.then(({ idx, meta }: WinningMessage) => {
+			clearAllTimeouts(candidateConfigsWithTimeout);
```
adamnfish, line 113: This is the bit that led to my confusion, and caused me to drill in on this. Is it not true at this point that all the `canShows` have already timed out or succeeded?

src/web/lib/messagePicker.ts
```diff
@@ -0,0 +1,123 @@
...
+	};
+};
+
+const clearAllTimeouts = (messages: CandidateConfigWithTimeout[]) =>
+	messages.map((m) => m.cancelTimeout());
+
+const defaultShow = () => null;
+
+type WinningMessage = {
+	idx: number;
```
adamnfish, line 78: Would this be easier to use if it returned the winning candidate, rather than the index of the winning candidate? Typically it's nice to return information directly where possible, rather than the means to look it up. Apart from anything else, it removes the need to add code that checks that the index exists (or in general where the lookup fails).

adamnfish (commented):
Much easier to follow with the new names!

I got a bit mired in the timeout logic this time, apologies. Some comments there we can chat about if you like?

adamnfish (approved):
We discussed this in detail IRL, and went through some details of the code that explain the way it is. This did raise a few more Qs which we'll look into, but this looks like a great change. Very happy to take on the more detailed thoughts as and when we're able to get a bit more clarity on how this should work.

tjmw:
Thanks @adamnfish 👍  I'm going to go ahead and merge this to avoid blocking other dependent work and follow up on the most recent round of feedback in a separate PR. I think that's an okay route to take because the feedback now mostly relates to the timeout behaviour which hasn't changed here.

PR #2479, opened 2021-02-03
Consistent SVG star icon sizing ★★★☆☆

mxdvl:
## What does this change?

Makes the stroke of the star icons `inside` rather than `centred` for visual consistency.

By using the filled shape as a `clip-path` to the stroke-shape, and doubling the stroke width, we achieve the same visual result as if the stroke was inside the shape.

image

<details>
    <summary>Example SVG code</summary>

```xml
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="600" height="200"
	viewBox="-2 -2 60 20">
	<path id="star" paint-order="stroke"
		d="M0 5.2L3.7 8l-1.4 4.6.5.4 3.7-2.8 3.7 2.8.5-.4L9.3 8 13 5.2l-.2-.6H8.2L6.8 0h-.6L4.8 4.6H.2l-.2.6z">
	</path>
	<clipPath id="clip">
		<use xlink:href="#star" stroke="none" />
	</clipPath>

	<use xlink:href="#star" fill="rgba(255,100,0,0.5)" />
	<use x="15" xlink:href="#star" stroke="black" stroke-width="1" fill="rgba(255,100,0,0.5)" />
	<use x="30" xlink:href="#star" stroke="black" stroke-width="2" fill="rgba(255,100,0,0.5)" />
	<use x="45" xlink:href="#star" stroke="black" stroke-width="2" fill="rgba(255,100,0,0.5)" clip-path="url(#clip)" />

	<g text-anchor="middle" font-size="2px" font-family="sans-serif">
		<text y="16" x="7">baseline</text>
		<text y="16" x="7" dx="15">centred 1px</text>
		<text y="16" x="7" dx="30">centred 2px</text>
		<text y="16" x="7" dx="45">2px clipped</text>
	</g>
</svg>
```

</details>

## Screenshot

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/106797722-fe894280-662a-11eb-851d-08329fda52b2.png
[after]: https://user-images.githubusercontent.com/76776/106797762-0b0d9b00-662b-11eb-9199-27cb32c3181c.png

## Why?

It looks weird that empty stars are bigger than filled ones.

### DIFF
--- a/src/static/icons/star.svg
+++ b/src/static/icons/star.svg
@@ -1,3 +1,8 @@
 <svg width="14" height="13" viewBox="0 0 14 13">
-    <path d="M0 5.2L3.7 8l-1.4 4.6.5.4 3.7-2.8 3.7 2.8.5-.4L9.3 8 13 5.2l-.2-.6H8.2L6.8 0h-.6L4.8 4.6H.2l-.2.6z"></path>
+    <path id="star" clip-path="url(#clip)" stroke-width="2"
+        d="M0 5.2L3.7 8l-1.4 4.6.5.4 3.7-2.8 3.7 2.8.5-.4L9.3 8 13 5.2l-.2-.6H8.2L6.8 0h-.6L4.8 4.6H.2l-.2.6z">
+    </path>
+    <clipPath id="clip">
+        <use xlink:href="#star" stroke="none" />
+    </clipPath>
 </svg>


oliverlloyd (approved):
Oh, well done, this is @paperboyo level perfection. I hadn't even noticed this but as soon as you pointed it out it's suddenly shouting at me that the empty star is too big!

Just a thought and not a blocker for this PR, but is this improvement also worth adding to the star icon in Source?

https://theguardian.design/2a1e5182b/p/3892a2-icons/i/13404924

paperboyo:
Mega! Also, pointy not tempered corners: 👏 .

oliverlloyd:
@liywjl do you think we need to check this against your special 6th star?

mxdvl:
I still think that it would be best for Source to provide two distinct icons for the star. The benefir of one `filled` and one `empty` star is that Figma will [automagically adjust the path][1] so it works with a centred stroke (emphasis mine):

> #### Inside/Outside Strokes
>
> Because SVG only supports center strokes, design tools have had to devise various workarounds to simulate inside and outside strokes. Previously, we did this by using center strokes with a doubled stroke width. That is visually equivalent to having both an inside and outside stroke so we used a mask to hide the unnecessary half. This approach greatly inflated the file size and complicated the markup.
> 
> **We now try to adjust the points in the path such that the visual result will resemble an inside or outside stroke while using center strokes with the original stroke width.** For example, a rectangle with an inner stroke can be represented as smaller rectangle with a center stroke.

cc. @SiAdcock @akemitakagi 

[1]: https://www.figma.com/blog/with-figmas-new-svg-exports-less-more/

mxdvl:
Added a visual explanation of what is happening with the clip-path:

```svg
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="600" height="200"
	viewBox="-2 -2 60 20">
	<path id="star" paint-order="stroke"
		d="M0 5.2L3.7 8l-1.4 4.6.5.4 3.7-2.8 3.7 2.8.5-.4L9.3 8 13 5.2l-.2-.6H8.2L6.8 0h-.6L4.8 4.6H.2l-.2.6z">
	</path>
	<clipPath id="clip">
		<use xlink:href="#star" stroke="none" />
	</clipPath>

	<use xlink:href="#star" fill="rgba(255,100,0,0.5)" />
	<use x="15" xlink:href="#star" stroke="black" stroke-width="1" fill="rgba(255,100,0,0.5)" />
	<use x="30" xlink:href="#star" stroke="black" stroke-width="2" fill="rgba(255,100,0,0.5)" />
	<use x="45" xlink:href="#star" stroke="black" stroke-width="2" fill="rgba(255,100,0,0.5)" clip-path="url(#clip)" />

	<g text-anchor="middle" font-size="2px" font-family="sans-serif">
		<text y="16" x="7">baseline</text>
		<text y="16" x="7" dx="15">centred 1px</text>
		<text y="16" x="7" dx="30">centred 2px</text>
		<text y="16" x="7" dx="45">2px clipped</text>
	</g>
</svg>
```

mxdvl:
I think this broke `main`?

ashishpuliyel:
I rate this PR
:star2::star2::star2::star2::star2:

gtrufitt:
We had reports that on AMP, we're seeing star that have gone walkabout on Safari and Chrome on IOS 14.4:

image

I can't replicate it *but* I can see the last star render first and the others pop in later on my phone.

I'm not sure whether this PR is related at all!

However, I wondered whether the clip path using and href of #star and an id of star matters or whether the svg essentially creates a shadow-dom where the same ID for each of the svgs doesn't matter?

Any other thought that might be causing this behaviour (outside of this change!) @mxdvl ?

mxdvl:
Hey @gtrufitt — I can’t replicate this first star showing, but can only say that the SVG _does not_ create a shadow DOM. That only happens with `<use xlink:href="#star">` and indeed the first instance of id `star` on the page will be used as the reference.

What is most peculiar is that these SVGs are inlined, so they should already be fully loaded when the HTML is parsed.

gtrufitt:
Interesting, thanks

> What is most peculiar is that these SVGs are inlined, so they should already be fully loaded when the HTML is parsed.

yeah... I *think* maybe all the star end up in the same position, i.e a layout issue rather than a rendering issue

mxdvl:
That would make more sense, actually, but I was unable to reproduce this issue, even with the slowest throttle on my device.

mxdvl:
@gtrufitt, might be worth investigating if #2732 helped at all with the issues you’ve seen.

PR #2485, opened 2021-02-04
Add spacing and border to both sets of links

oliverlloyd:
## What does this change?
Moves the css for padding and a bottom border away from the keyword links and adds it to a wrapper around both the section _and_ keyword links.

### Before
Screenshot 2021-02-04 at 09 48 38

### After
Screenshot 2021-02-04 at 09 48 11

## Why?
Because if we only apply this css to the bottom of the keyword links then we miss it entirely if there are none of them

### DIFF
--- a/src/web/components/SubMeta.tsx
+++ b/src/web/components/SubMeta.tsx
@@ -32,7 +32,7 @@ const listStyleNone = css`
 	list-style: none;
 `;
 
-const keywordListStyles = css`
+const listWrapper = css`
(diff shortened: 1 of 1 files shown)


oliverlloyd:
[Zero keyword links spacing issue](https://trello.com/c/hjPTPMKt/2427-zero-keyword-links-spacing-issue)

liywjl (approved):
LGTM

PR #2486, opened 2021-02-04
Remove pillarToUse

oliverlloyd:
## What?
Deletes the `pillarToUse` override logic

## Why?
Because we do the same thing inside `decideTheme` so these checks are no longer needed


oliverlloyd:
[Remove pillarToUse from Card](https://trello.com/c/GYYmPt9h/2430-remove-pillartouse-from-card)

liywjl (approved):
👍

PR #2489, opened 2021-02-04
decideTrail

oliverlloyd:
## What does this change?
Hoists the logic to transform the `CAPITrailType` into `TrailType` up into a shared decide function called `decideTrail`. This now lives with the other decide family of functions, all serving to transform the old school CAPI data into the new and cool enums used to define `Format`

```typescript
export const decideTrail = (trail: CAPITrailType): TrailType => {
	// We don't have tags here so we send an empty array
	const design = decideDesign(trail.designType, []);
	return {
		...trail,
		format: {
			// We don't have enough data from CAPI to know what display should be
			display: Display.Standard,
			theme: decideTheme({ pillar: trail.pillar, design }),
			design,
		},
	};
};
```

### Hardcoding display
We currently need to hardcode `display` because card data from CAPI has no information that lets us decide this. We were doing this before but in multiple places, this PR centralises this hardcoded decision, making it easier to fix in the future.

### `format`
Once we had, `display`, `design` and `pillar` as props for `Card` and friends, it was logical to upgrade `TrailType` to use `format`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -30,6 +30,7 @@ type CAPIDesign =
 	| 'SpecialReport'
 	| 'GuardianLabs';
 
+type Display = import('@guardian/types').Display;
 type Design = import('@guardian/types').Design;
 type Theme = import('@guardian/types').Theme;
 type Format = import('@guardian/types').Format;
@@ -806,6 +807,7 @@ interface BaseTrailType {
     linkText?: string;
 }
 interface TrailType extends BaseTrailType {
+    display: Display;
 	design: Design;
 	pillar: Theme;
 }
--- a/index.d.ts
+++ b/index.d.ts
@@ -807,9 +807,7 @@ interface BaseTrailType {
     linkText?: string;
 }
 interface TrailType extends BaseTrailType {
-    display: Display;
-	design: Design;
-	pillar: Theme;
+    format: Format;
 }
 
 interface CAPITrailType extends BaseTrailType {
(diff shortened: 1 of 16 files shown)


oliverlloyd:
[Hoist decision for Display](https://trello.com/c/BAUmNrqQ/2431-hoist-decision-for-display)

index.d.ts
```diff
@@ -806,8 +807,7 @@ interface BaseTrailType {
     linkText?: string;
 }
 interface TrailType extends BaseTrailType {
-	design: Design;
-	pillar: Theme;
+    format: Format;
```
oliverlloyd, line 810: This is the heart of this PR. All the other changes exist to service this change

src/web/components/MostViewed/MostViewedRight/MostViewedRight.tsx
```diff
@@ -46,26 +35,22 @@ export const MostViewedRight = ({ pillar, limitItems = 5 }: Props) => {
...
-								mostViewedItemIndex: number,
-							) => (
-								<MostViewedRightItem
-									key={trail.url}
-									trail={transformTrail(trail)}
-									mostViewedItemIndex={mostViewedItemIndex}
-								/>
-							),
-						)}
+					{(trails || []).map((trail, index) => (
```
liywjl, line 47: if `data.trails` is not an array it would fail at line 38, maybe do the cheak for its existance before? or remove ` || []` if you know its defined

liywjl (approved):
LGTM

src/web/components/MostViewed/MostViewedRight/MostViewedRight.tsx
oliverlloyd: Yeah, I was tempted to remove this. I suspect the reason it was put there to start with was typescript related

PR #2492, opened 2021-02-05
Use beige for editorials as well

oliverlloyd:
## What does this change?
Fixes an issue where the background colour for editorials was being set as white, not beige

### Before
Screenshot 2021-02-05 at 09 10 22

### After
Screenshot 2021-02-05 at 09 10 42

## Why?
In response to an issue raised to cp

## Why didn't we capture this in our stories?
We do snapshot GuardianView articles so we _should_ have caught this but the quality of that fixture data isn't great which meant we were not rendering a proper representation of an editorial article. To fix this, I've updated that fixtures data in this PR but I also added https://trello.com/c/NYD6Pzjm/2435-make-updating-article-fixtures-easier so that we prevent similar problems with other articles and ensure we have up to date data moving forwards

### DIFF
--- a/src/web/lib/decidePalette.ts
+++ b/src/web/lib/decidePalette.ts
@@ -144,9 +144,18 @@ const textArticleLinkHover = (format: Format): string => {
 
 const backgroundArticle = (format: Format): string => {
 	// Order matters. We want comment special report pieces to have the opinion background
-	if (format.design === Design.Comment) return opinion[800];
-	if (format.theme === Special.SpecialReport) return specialReport[800];
-	return 'transparent';
+	if (
+		format.design !== Design.Comment &&
+		format.theme === Special.SpecialReport
+	)
+		return specialReport[800];
+	switch (format.design) {
+		case Design.Comment:
+		case Design.GuardianView:
+			return opinion[800];
+		default:
+			return 'transparent';
+	}
 };
 
 const backgroundSeriesTitle = (format: Format): string => {
--- a/src/web/lib/decidePalette.ts
+++ b/src/web/lib/decidePalette.ts
@@ -144,18 +144,10 @@ const textArticleLinkHover = (format: Format): string => {
 
 const backgroundArticle = (format: Format): string => {
 	// Order matters. We want comment special report pieces to have the opinion background
-	if (
-		format.design !== Design.Comment &&
-		format.theme === Special.SpecialReport
-	)
-		return specialReport[800];
-	switch (format.design) {
-		case Design.Comment:
-		case Design.GuardianView:
-			return opinion[800];
-		default:
-			return 'transparent';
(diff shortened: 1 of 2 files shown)


src/web/lib/decidePalette.ts
```diff
@@ -144,9 +144,18 @@ const textArticleLinkHover = (format: Format): string => {
 
 const backgroundArticle = (format: Format): string => {
 	// Order matters. We want comment special report pieces to have the opinion background
-	if (format.design === Design.Comment) return opinion[800];
-	if (format.theme === Special.SpecialReport) return specialReport[800];
-	return 'transparent';
+	if (
```
nicl, line 147: This whole bit still feels a bit (though mostly because the underlying logic is just confusing rather than the code itself). But what about:

    if (format.design === Design.Comment) return opinion[800];
    if (format.theme === Special.SpecialReport) return specialReport[800];
    if (format.design === Design.GuardianView) return opinion[800];
    return 'transparent';

(it's not a major thing either way but seems slightly clearer as only one type of control flow and also the code mirrors the actual priority order.)

nicl (approved):

src/web/lib/decidePalette.ts
```diff
@@ -144,9 +144,18 @@ const textArticleLinkHover = (format: Format): string => {
...
+		format.theme === Special.SpecialReport
+	)
+		return specialReport[800];
+	switch (format.design) {
+		case Design.Comment:
+		case Design.GuardianView:
+			return opinion[800];
+		default:
+			return 'transparent';
+	}
```
jfsoul, line 158: is this the same as?
```
	if (format.design === Design.Comment) return opinion[800];
	if (format.theme === Special.SpecialReport) return specialReport[800];
	if (format.design === Design.GuardianView) return opinion[800];
	return 'transparent';
```
Neater, but debatable which is easier to read

jfsoul (approved):

src/web/lib/decidePalette.ts
nicl: I was first! 😆

src/web/lib/decidePalette.ts
jfsoul: oh snap!

oliverlloyd: Your code is _much_ cleaner. Adopted. (I reversed special and guardianview though as I think we want to prioritse guardianview over special)

oliverlloyd: Sorry, Nic won here. I read top to bottom

liywjl (approved):

src/web/lib/decidePalette.ts
jfsoul: 😭

PR #2493, opened 2021-02-05
Bump Discussion to 6.1.0

nitro-marky:
## What does this change?

Bumps to the latest version of discussion-rendering, which should fix the 'Access Denied' error for new commenters. It is worth noting local testing has not been possible so patterns used on other functions have been copied.


oliverlloyd (approved):
Looks like Cypress ran okay on TC so (👍 ). Not sure why it failed here on GH

PR #2495, opened 2021-02-05
Sticky nav engagement test

nicl:
**UPDATE:** now lazy-loads nav data on client when reader expands the main menu to avoid increasing the size of `window.guardian` data. A new Frontend API endpoint is used for this (https://github.com/guardian/frontend/pull/23543).

## What does this change?

Adds an AB test `stickyNavTest` with three variants:

    control // the status quo
    sticky-nav-simple // a CSS-only sticky nav implementation
    sticky-nav-backscroll // a JS version that shows the nav on backscroll

### Before

(Nav stays put when scrolling down the page and does not track the viewport.)

### After

CSS-only version:
Kapture 2021-02-01 at 15 12 34

JS-backscroll version:
Kapture 2021-02-03 at 10 34 25

## Why?

Increased discoverability of the nav menu and pillars is hypothesised to increase engagement.

The Trello card collects together some of the relevant docs here: https://trello.com/c/fy5HsbwD/2347-sticky-nav-test for further reading.

### DIFF
--- a/src/lib/reset-css.js
+++ b/src/lib/reset-css.js
@@ -32,11 +32,6 @@ const defaults = `
     em {
         font-style: italic;
     }
-
-    *:focus {
-        outline-color: red;
-        outline-style: solid;
-    }
 `;
 
 module.exports = [resetCSS, defaults].join('').replace(/\s/g, '');
--- a/src/lib/reset-css.js
+++ b/src/lib/reset-css.js
@@ -32,6 +32,11 @@ const defaults = `
     em {
         font-style: italic;
     }
+
+    *:focus {
+        outline-color: red;
+        outline-style: solid;
+    }
 `;
 
 module.exports = [resetCSS, defaults].join('').replace(/\s/g, '');
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -66,6 +66,7 @@ import { InstagramBlockComponent } from '@root/src/web/components/elements/Insta
 import { MapEmbedBlockComponent } from '@root/src/web/components/elements/MapEmbedBlockComponent';
 import { SpotifyBlockComponent } from '@root/src/web/components/elements/SpotifyBlockComponent';
 import { VideoFacebookBlockComponent } from '@root/src/web/components/elements/VideoFacebookBlockComponent';
+import { StickyNav } from '@root/src/web/components/StickyNav/StickyNav';
 import {
 	submitComponentEvent,
 	OphanComponentEvent,
@@ -472,6 +473,15 @@ export const App = ({ CAPI, NAV }: Props) => {
 					</>
 				</HydrateOnce>
 			))}
+			<HydrateOnce root="sticky-nav-root">
+				<StickyNav
+					CAPI={CAPI}
+					NAV={NAV}
+					format={format}
+					palette={palette}
+				/>
+			</HydrateOnce>
+
 			{NAV.subNavSections && (
 				<HydrateOnce root="sub-nav-root">
 					<>
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -66,7 +66,11 @@ import { InstagramBlockComponent } from '@root/src/web/components/elements/Insta
 import { MapEmbedBlockComponent } from '@root/src/web/components/elements/MapEmbedBlockComponent';
 import { SpotifyBlockComponent } from '@root/src/web/components/elements/SpotifyBlockComponent';
 import { VideoFacebookBlockComponent } from '@root/src/web/components/elements/VideoFacebookBlockComponent';
-import { StickyNav } from '@root/src/web/components/StickyNav/StickyNav';
+import {
+	StickyNavSimple,
+	StickyNavBackscroll,
+} from '@root/src/web/components/StickyNav/StickyNav';
+
 import {
 	submitComponentEvent,
 	OphanComponentEvent,
@@ -128,7 +132,7 @@ const GetMatchStats = React.lazy(() => {
 
 type Props = {
 	CAPI: CAPIBrowserType;
-	NAV: NavType;
+	NAV: BrowserNavType;
 };
 
 const componentEventHandler = (
@@ -326,6 +330,16 @@ export const App = ({ CAPI, NAV }: Props) => {
 
 	const adTargeting: AdTargeting = buildAdTargeting(CAPI.config);
 
+	// sticky nav test status
+	const inStickyNavBackscroll = ABTestAPI.isUserInVariant(
+		'stickyNavTest',
+		'sticky-nav-backscroll',
+	);
+	const inStickyNavSimple = ABTestAPI.isUserInVariant(
+		'stickyNavTest',
+		'sticky-nav-simple',
+	);
+
 	// There are docs on loadable in ./docs/loadable-components.md
 	const YoutubeBlockComponent = loadable(
 		() => {
@@ -473,14 +487,28 @@ export const App = ({ CAPI, NAV }: Props) => {
 					</>
 				</HydrateOnce>
 			))}
-			<Portal root="sticky-nav-root">
-				<StickyNav
-					CAPI={CAPI}
-					NAV={NAV}
-					format={format}
-					palette={palette}
-				/>
-			</Portal>
+
+			{inStickyNavBackscroll && (
+				<Portal root="sticky-nav-root">
+					<StickyNavBackscroll
+						capiData={CAPI}
+						navData={NAV}
+						format={format}
+						palette={palette}
+					/>
+				</Portal>
+			)}
+
+			{inStickyNavSimple && (
+				<Portal root="sticky-nav-root">
+					<StickyNavSimple
+						capiData={CAPI}
+						navData={NAV}
+						format={format}
+						palette={palette}
+					/>
+				</Portal>
+			)}
 
 			{NAV.subNavSections && (
 				<HydrateOnce root="sub-nav-root">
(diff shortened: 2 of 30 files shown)


shtukas (approved):

src/lib/reset-css.js
```diff
@@ -32,6 +32,11 @@ const defaults = `
     em {
         font-style: italic;
     }
+
+    *:focus {
+        outline-color: red;
+        outline-style: solid;
+    }
```
gtrufitt, line 39: Was this testing code?

oliverlloyd (commented):
Some general comments, often things you can make your own mind up about. My larger concern is the bundle bump. Not sure how we deal with that though

src/web/components/StickyNav/StickyNav.tsx
```diff
@@ -0,0 +1,208 @@
...
+	NAV: NavType;
+	format: Format;
+	palette: Palette;
+	ID?: string;
+}
+
+const stickyStyle = (theme: Theme) => css`
+	position: sticky;
+	top: 0;
+	z-index: 900;
```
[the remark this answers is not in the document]
nicl: Ah yes, thanks. I meant to clear that up once I'd got things working.

src/web/components/StickyNav/StickyNav.tsx
```diff
@@ -0,0 +1,208 @@
...
+	background-color: white;
+	box-shadow: 0 0 transparent, 0 0 transparent,
+		1px 3px 6px ${neutralBorder(theme)};
```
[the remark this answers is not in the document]
nicl: @oliverlloyd would be keen to decouple that from this test but I'm very happy to do that work!

src/web/components/StickyNav/StickyNav.tsx
```diff
@@ -0,0 +1,208 @@
...
+`;
+
+const fixedStyle = (theme: Theme, display: boolean) => css`
```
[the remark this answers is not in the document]
nicl: 👍 I'll change to `show` or `shouldDisplay`

src/web/components/App.tsx
```diff
@@ -455,6 +456,15 @@ export const App = ({ CAPI, NAV }: Props) => {
 					</>
 				</HydrateOnce>
 			))}
+			<Portal root="sticky-nav-root">
+				<StickyNav
+					CAPI={CAPI}
```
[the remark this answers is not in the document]
nicl: Yep good points. I'll look into being more explicit here as agree that is generally the better approach.

src/web/components/App.tsx
```diff
@@ -472,6 +487,29 @@ export const App = ({ CAPI, NAV }: Props) => {
 					</>
 				</HydrateOnce>
 			))}
+
+			{inStickyNavBackscroll && (
```
nicl, line 491: Key parts here - depending on which variant (if any), we hydrate the relevant component.

src/web/components/Nav/Nav.tsx
```diff
@@ -127,6 +127,7 @@ export const Nav = ({ format, nav, subscribeUrl, edition }: Props) => {
                         var keydownToggleMainMenu = function(e){
                           // keyCode: 13 => Enter key | keyCode: 32 => Space key
                           if (e.keyCode === 13 || e.keyCode === 32) {
+							  console.log('WOOOOW');
```
liywjl, line 130: remove

src/web/components/StickyNav/ExpandedMenu/Column.tsx
```diff
@@ -0,0 +1,269 @@
...
+				columnInputId={columnInputId}
+				ariaControls={`${column.title.toLowerCase()}Links`}
+			/>
+
+			{/* ColumnLinks */}
+			<ul
+				className={cx(
+					columnLinks,
+					{ [firstColumnLinks]: index === 0 },
+					{ [pillarColumnLinks]: !!column.pillar },
```
liywjl, line 236:
```suggestion
					index === 0 && firstColumnLinks,
					!!column.pillar && pillarColumnLinks,
```

src/web/components/StickyNav/ExpandedMenu/Column.tsx
```diff
@@ -0,0 +1,269 @@
...
+				role="menu"
+				id={`${column.title.toLowerCase()}Links`}
+				data-cy={`${column.title.toLowerCase()}Links`}
+			>
+				{(column.children || []).map((link) => (
+					<li
+						key={link.title.toLowerCase()}
+						className={cx(mainMenuLinkStyle, {
+							[hideDesktop]: !!link.mobileOnly,
+						})}
```
liywjl, line 248:
```suggestion
						className={cx(mainMenuLinkStyle, !!link.mobileOnly && hideDesktop)}
```

src/web/components/StickyNav/ExpandedMenu/MoreColumn.tsx
```diff
@@ -0,0 +1,211 @@
...
+	};
+	return (
+		<li
+			className={cx(columnStyle, pillarDivider, pillarDividerExtended)}
+			role="none"
+		>
+			<ul
+				className={cx(columnLinks, {
+					[pillarColumnLinks]: !!moreColumn.pillar,
+				})}
```
liywjl, line 181:
```suggestion
				className={cx(columnLinks, !!moreColumn.pillar && pillarColumnLinks)}
```

src/web/components/StickyNav/ExpandedMenu/MoreColumn.tsx
```diff
@@ -0,0 +1,211 @@
...
+				role="menu"
+				id={subNavId}
+			>
+				{(moreColumn.children || []).map((link) => (
+					<li
+						key={link.title.toLowerCase()}
+						className={cx(mainMenuLinkStyle, {
+							[hideDesktop]: !!link.mobileOnly,
+						})}
```
liywjl, line 190:
```suggestion
						className={cx(mainMenuLinkStyle, !!link.mobileOnly && hideDesktop)}
```

src/web/components/StickyNav/ExpandedMenu/ReaderRevenueLinks.tsx
```diff
@@ -0,0 +1,102 @@
...
+	];
+
+	return (
+		<ul className={hideDesktop}>
+			{links.map((link) => (
+				<li
+					key={link.title.toLowerCase()}
+					className={cx(mainMenuLinkStyle, {
+						[hideDesktop]: !!link.mobileOnly,
+					})}
```
liywjl, line 85:
```suggestion
					className={cx(mainMenuLinkStyle, !!link.mobileOnly && hideDesktop)}
```

src/web/components/StickyNav/LazyExpandedMenu.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+const mainMenuStyles = (ID: string) => css`
+	background-color: ${brandBackground.primary};
+	box-sizing: border-box;
+	${textSans.large()};
+	left: 0;
+	margin-right: 29px;
+	padding-bottom: 24px;
+	top: 0;
+	z-index: 1070;
```
liywjl, line 27: I remember having issues with `z-index` on the Nav bar, but have you tried using our zIndex function? (I think its unlikely to work but worth a shot)

oliverlloyd:
There's a lot of code here so it's difficult to give a more thorough review but I wanted understand the bundle size bump better. Is it the case that adding this feature will add 8Kb to the page size because of a change to how we render/hydrate? Or is it that we're just adding a lot of code and this is where the bump comes from?

If the cause is structural, then _should_ we be doing this? We might gain a sticky nav but the potential cost (and admimtedly this is hard to measure) in terms of traffic, search position (and resulting loss of revenue) plus the impacct to reader UX, might not be worth it?

If it's just extra code then would this reduce when we remove the test? I presume we're doubling up on a lot of things?

liywjl (approved):
There is a lot going on here, so its hard to know if I have cought every case, but over seems good 👍

oliverlloyd:
Also, it seems like we're intoducing two additional versions of the menu? Maybe we can break these out into separate PRs to make it easier to review?

It's difficult though and I get that we just want something that works that we can test. My concern though is if we pass through code under the 'just a test' banner it has the effect of giving it an easier ride into the codebase and then, later on, if we remove the test and actually enable the feature for everyone, there's no opportunity to review the code properly or if you do you have to go outside of the git tools and processes that we're so used to. For this reason I think we should treat test code the same as a team policy.

Crazy idea? @guardian/dotcom-platform ?

oliverlloyd:
> For this reason I think we should treat test code the same as a team policy

Alternatively, maybe we should embrace this 'test only' swimlane into production and say that we don't worry about time consuming reviews and make the lighter approach to code checks (and testing, and models, etc.) policy _but_ we also enforce an approach where the code must be rewritten and reintroduced again when the test is complete. 

Maybe we could do that by using folders or naming? Not sure

gtrufitt:
> but we also enforce an approach where the code must be rewritten and reintroduced again when the test is complete.

Yeah, I've seen this before in Frontend a few times. Really we want a `remove` PR that rips out everything and then bring it all back in as a real PR. In Frontend JS that was an 'experiement' lived in the `experiements` folder 90% of time, maybe we should adopt that.

src/web/components/StickyNav/LazyExpandedMenu.tsx
nicl: Oh @liywjl that is something I missed. I am using the helper function elsewhere and should have updated this too. Thanks.

nicl:
Closing in favour of https://github.com/guardian/dotcom-rendering/pull/2592 which should be easier to review.

PR #2497, opened 2021-02-05
Html is an Option in pullquote

gtrufitt:
## What does this change?

As per https://www.theguardian.com/world/2020/sep/20/the-qanon-conspiracy?dcr=false there is a pullquote as the first element in CAPI with no `html` field. This is an Option(html) type in Frontend, so needs handling in DCR.

### Before

Pages with this element throw 500

### After

Pages with this element do not throw 500

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
         "@types/lodash.debounce": "^4.0.6",
         "@types/lodash.escape": "^4.0.6",
         "@types/node-fetch": "^2.5.8",
-        "@types/sanitize-html": "^1.18.3",
+        "@types/sanitize-html": "^1.27.1",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -65,7 +65,7 @@
         "@types/lodash.debounce": "^4.0.6",
         "@types/lodash.escape": "^4.0.6",
         "@types/node-fetch": "^2.5.8",
-        "@types/sanitize-html": "^1.18.3",
+        "@types/sanitize-html": "^1.27.1",
```
gtrufitt, line 68: Transitive dependency of sanitize-html caused a failure in generating the schema.

liywjl (approved):
LGTM

PR #2499, opened 2021-02-05
Use palette in Card to make them Special

oliverlloyd:
## What does this change?
Refactors Card and friends to use `palette` and `format`

### `Design.Line`
Screenshot 2021-02-05 at 20 37 17

### `Design.Comment`
Screenshot 2021-02-05 at 17 47 30

## Why?
So that we can bring styling decisions around colour into `decidePalette`

### Why is the footer text hard to read?
Because that change (to white) - including passing through `palette to the footer components - will be done in a seperate PR


oliverlloyd:
[Use palette in Card](https://trello.com/c/LjZNngDq/2432-use-palette-in-card)

oliverlloyd:
@buck06191 just a heads up, to allow the headline colour to be set in `decidePalette` I had to create the concept of an immersive card. That is, `decidePalette` will return different colours when `format` contains `Display.Immersive`. I only took it as far as this.

This moves away from the boolean of `isFullCardImage` but I didn't want to add a new prop to the model for this use case.

@HarryFischer I think having an `Immersive` card is a reasonable style choice based on how the carousel looks and based on the Format model but I'm not sure if there are other intentions here?

paperboyo:
> I had to create the concept of an immersive card

Just a heads up, this is (was?) a [thing already once](https://github.com/guardian/frontend/pull/19647) (couldn’t find the initial PR). There were some image issues around them and then editors forgotten what obscure combination of switches activates them and then those who came up with that combination forgot that too. And now I forgot, so trying to trigger them in fronts tool doesn’t work for me. Maybe they were properly killed so the name is for the taking?

oliverlloyd:
> Just a heads up, this is (was?) a [thing already once](https://github.com/guardian/frontend/pull/19647)

Ah, thanks for the link, this is really useful!

So this raises a question: should we change the designs for the 'immersive' card to align with these already existing designs? Perhaps we can come back to that later though in a separate ticket? I think this an important area to work through but I don't think we should block the work @buck06191 is doing for the Carousel test because of it.

It also reases the question of how we model these designs. I had a discussion with @gtrufitt and @JamieB-gu yesterday around this issue. We very roughly drew out a type that looked a bit like:

```typescript
type Card = {
  // Some imperative props to model a card
  palette: Palette; // Built based on a combination of Format AND Container
}

enum ContainerThene = {
  Dynamo
  Live
}

enum ContainerLayout = {
  2HQ1Q1
  Spotlight
  Carousel
  Q1Q1Q1Q1
  Blah
}

type Container = {
  layout: ContainerLayout;
  theme: ContainerThene;
  cards: Card[];
}
```

Where the idea would be that when deciding `palette` for a Card, we include the `Container` type as part of the decision so that we have the context of the card in scope. Eg. If `ContainerLayout` were `Carousel`, then we can always use immersive cards, regardless of what might be set in `Display`.

liywjl (approved):
LGTM

HarryFischer:
@oliverlloyd I think as the carousel immersive cards are just a test we will probably further develop still if it performs well. Are Zefs immersive cards only available on front at the moment? Or can they be used in the end of article component?

paperboyo:
> Are Zefs immersive cards only available on front at the moment? Or can they be used in the end of article component?

They were front-only. But they may not be a thing any more: I can’t get them to work in TEST. Maybe @gtrufitt will remember if they were killed (I can’t inspect codebase to check)?

oliverlloyd:
> But they may not be a thing any more

If they were killed, I'd be interested to understand the reason why. It could well still be valid

gtrufitt:
> > Are Zefs immersive cards only available on front at the moment? Or can they be used in the end of article component?
> 
> They were front-only. But they may not be a thing any more: I can’t get them to work in TEST. Maybe @gtrufitt will remember if they were killed (I can’t inspect codebase to check)?

I don't remember the designs! Think it was the time when I wasn't on the team

PR #2500, opened 2021-02-06
Dinkus

oliverlloyd:
## What does this PR do?
Widens the definition of a dinkus to include both the spaced `* * *` and the unspaced `***`. We're also removing the requirement for a dinkus to be marked as a subheading.

## Before
Screenshot 2021-02-10 at 09 13 16

## After
fancy dinkus

## What is a dinkus?
A dinkus is a line of 3 asterisks used to define a section break between pieces of text. It's the same as an [asterism](https://en.wikipedia.org/wiki/Asterism_%28typography%29) (⁂) but easier to type.

The etymology of dinkus is possibly [derived from the dutch](https://www.theparisreview.org/blog/2018/06/08/ode-to-the-dinkus/) [ding](https://en.wiktionary.org/wiki/ding#Dutch), meaning thing, or maybe its cuter diminutive, dingetje ('ding-et-cher') meaning thingy.

For a long time I thought that these asterisks were a hack, a Guardian quirk invented a few years ago to badly solve a problem in photo essays. Turns out they date back to the [Illiad](https://en.wikipedia.org/wiki/Iliad) so they're basically as old as literature itself. My bad.

## How do we represent a dinkus?
Before DCR, if the rendering code came across `<h2><p>* * *</p></h2>` then it would replace this html with a `<hr />` (a section break) and mark the subsequent paragraph as a drop cap. But only if the article were `Immersive`.

DCR [widened this net](https://github.com/guardian/dotcom-rendering/pull/2262) to include any article, not just immersives, but we still required the asterisks to be in a `h2` tag.

This PR widens that net further. 

```html
<p>***</p>
<p>* * *</p>
<h2><p>***</p></h2>
<h2><p>* * *</p></h2>
```
All of these are now valid dinkus.

## When do drop caps appear?
If you insert a dinkus then an attempt to add a drop cap on the subsequent text block is made. This attempt is subject to the [rules of drop caps](https://github.com/guardian/dotcom-rendering/blob/c6147562240b52ef1906a3fde1c4e8f5abd2c4ac/src/web/components/elements/TextBlockComponent.tsx#L52):

1. The article type must be either Feature, Comment, Interview, PhotoEssay, Recipe or Review **_or_** the article must be Immersive.
2. A dropcap will only appear if the paragraph has more than 200 charactors.

If both these rules are true then a dropcap is always set on the first paragraph and also set on other in body paragraphs if a dinkus was used.

## What if we want the dinkus to mean something else at the Guardian?
Right now we use a simple grey divider about 150 pixels long, left aligned. But there's nothing stopping us having centre aligned dinki for Interviews, or having a fancy flower gliph (aka a [fleuron](https://en.wikipedia.org/wiki/Fleuron_(typography))) for Lifestyle pieces, these types of exceptions are easy to do in DCR.

Or we can do what [this person](https://lithub.com/i-reject-your-asterisks-and-your-dinkus-too/) wants and just replace everything with double white space.


paperboyo:
💎! Very informative, thank you!

My personal opinion: it makes all the sense to widen the net to have control over all manner of dinkuses (dinki?). I guess, they just **mean** a section break. Now: how should the section break look for any given template is a question for Design. I guess a drop cap may be a tad too much for a short news article, but this would only be my hunch. Also: why would anyone use a dinkus in a short news piece?

I guess CAPI doesn’t offer a way of escaping stemmers and analyzers and allows searching for `***` (a common problem with Guardian’s Elasticsearch implementations)? Worth checking, though. Trawiing existing corpus might give some idea about current use (and reveal possible other dinkuses).

After Design decides how (and if) to differ the appearance of section breaks, this needs signalling to Central Production who may decide to let editors know. As always, exercising caution and restraint, otherwise we will see a drop cap in every second article…

oliverlloyd:
[Dinkus](https://trello.com/c/TLSlNCXF/2436-dinkus)

shtukas (approved):

liywjl (approved):
Great work, also super insightful.

oliverlloyd:
@rcrphillips I think the questions for editorial here are:

1. Do we _always_ want to replace the text *** with a section break or are there cases where we intentionally want to display the classic *** dinkus?
Screenshot 2021-02-08 at 12 17 04

2. Do we want to use _other_ designs for section breaks? There are many examples of asterisms or fleurons on the web and it's reasonably easy for us to offer such variations based on the type of article.

In regards to dropcaps, I think we're agreed that these should only be used if the article is a Feature, Interview, Comment, PhotoEssay, Review, Recipe or Immersive? This means if you insert a section break in, say, a News piece, then we would replace it with a grey section break line but we wouldn't show a dropcap on the subsequent paragraph.

paperboyo:
Even if the answer to pt. 1 is that we sometimes would like to keep them asterisks (which I hope won’t happen), then this logic should stay so that we will clean all sorts of asterisks (h2ed, speaced out) int oa consistent view of them.

rcrphillips:
Thanks @oliverlloyd I might leave question 2 out to avoid adding extra work for now.....

JamieB-gu:
> What is a dinkus?

> or maybe its cuter diminutive, dingetje

> Turns out they date back to the [Illiad](https://en.wikipedia.org/wiki/Iliad) so they're basically as old as literature itself.

> if it worked for Homer, maybe it's good enough for the Guardian?

I love everything about this PR description 😂

Reading through this I think the presence of the `***` alone to mark a section is fine. The issue is that it does different things depending on the surrounding elements, so we have to backtrack/look ahead in our parsing to figure out what it actually represents.

If it's being used to create drop-caps then it no longer represents a simple section divider. Therefore perhaps we should parse these compound elements into a single `DropCap` element further up the tree (i.e. before it gets to CAPI)?

paperboyo:
> If it's being used to create drop-caps then it no longer represents a simple section divider.

Not necessarily. It’s akin to what editors are used to from InDesign: Paragraph Style. It can contain all sorts of stuff web will take 60 years to get: space before, after, background, all font styling in the world, nested styles, drop caps, you name it.

If we (or, rather, Design) decide that a section break on features template is a line plus a drop cap and a section break on news is simply a line, it’s OK, I think, if both are signalled by a dinkus, no? The dinkus does not controll how it will look in different contexts. It’s simply a signal of a section break. Platforms knowing it happened is useful. Completely orthogonal how it should look like in a specific context (may even look like a dinkus, just consistent without, or with, spaces).

oliverlloyd:
I've updated this PR to reflect the decisions that came out of a conversation with editorial. Those decisions were:

1. All instances of * * * (spaced) should be matched and reformatted
2. Any instance of *** (unspaced) should be ignored

I've also changed the logic for when dropcaps should show to make it impossible to force a dropcap if an artilce is not either Immersive or one of the allowed design types. This means we can allow * * * to create a divider in any article, but it will only trigger dropcaps if the rules are met.

gtrufitt:
> Any instance of *** (unspaced) should be ignored

Even if we do not change the style of the element, we should semantically mark it up.

gtrufitt (approved):
This is a dinkus of a PR! The power of EnhanceCAPI strikes again ⚡

PR #2506, opened 2021-02-08
Missing types

oliverlloyd:
## What?
Adds some missing types to `CAPIType` as well as to some element types

## Why?
Because the latest data being sent from CAPI includes this data but we're not typing it causing errors when we try to update our fixtures.


oliverlloyd:
[Missing Types](https://trello.com/c/VN8BXPhT/2437-missing-types)

liywjl (approved):

PR #2508, opened 2021-02-08
Correct BylineLink

shtukas:
Correct a bug in BylineLink. We now pick up the correct associated tags. This prevent bylines to point to tags pages when they should point to contributor pages.

Before:

Screenshot 2021-02-08 at 12 42 41

After:

Screenshot 2021-02-08 at 12 43 11


liywjl (approved):
LGTM

PR #2509, opened 2021-02-08
Refactor fixtures

oliverlloyd:
## What?
Refactors fixture usage so that all article level test data is read from the /articles directory, removing `CAPI.ts` and `article.ts`

I've also added `config.js` and `switches.js` as pure javascript files using the classic modules.export pattern so it's easier to use them in node later.

## Why?
Because we intend to write a script to refresh this data so that we're testing from the latest, most up to date sets and we don't want any files being used from outside this directory


shtukas (approved):

liywjl (approved):

oliverlloyd:
[Refactor Fixtures](https://trello.com/c/Sr45MMHx/2438-refactor-fixtures)

PR #2513, opened 2021-02-09
Update Witness test video

oliverlloyd:
## What does this change?
Replaces the test url used for the witness video with a more up to date one

### Before
Screenshot 2021-02-09 at 13 36 44

### After
Screenshot 2021-02-09 at 13 35 15

## Why?
The older style video was missing the Guardian logo allowing the 'Watch on YouTube' message to appear. But this message is intermittent and causes false nagatives in our visual regression suite. By using a more modern link we reduce these.

### DIFF
--- a/src/web/components/elements/WitnessBlockComponent.stories.tsx
+++ b/src/web/components/elements/WitnessBlockComponent.stories.tsx
@@ -135,7 +135,7 @@ export const WitnessVideoBlockComponentDefault = () => (
 			title="Warhammer Quest Timelapse"
 			description="Tell me if you&#39;ve heard this one before: A Dwarf, a thief, a fighter and a priest walk into a dungeon ... Warhammer Quest is the quintessential &#39;Old School&#39; dungeon-crawler."
 			authorName="Gregg Lewis-Qualls"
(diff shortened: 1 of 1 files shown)


shtukas (approved):

liywjl:
wouldn't just setting `modestbranding` to 1 hide the YT logo?   https://developers.google.com/youtube/player_parameters#modestbranding

no need to nessesary change the URL?

oliverlloyd:
> wouldn't just setting `modestbranding` to 1 hide the YT logo? https://developers.google.com/youtube/player_parameters#modestbranding
> 
> no need to nessesary change the URL?

My reading of the YT docs makes me think this option applies to the logo in the control bar

> Set the parameter value to 1 to prevent the YouTube logo from displaying in the control bar

That doesn't mean it doesn't _also_ affect the first page logo but with this new video url we can be more confident it addresses the issue

oliverlloyd:
[Update YouTube urls used for snapshots](https://trello.com/c/NNrxo7Qf/2441-update-youtube-urls-used-for-snapshots)

gtrufitt (approved):

liywjl (approved):
👍

PR #2521, opened 2021-02-11
Use loadable-assets-manifest for all asset paths

gtrufitt:
## What does this change?

Unfortunately a previous change left us loading the non-legacy version of our App JS on older browsers. This caused that JS not to run on these browsers. 

This happened because the `script.src` on prod was the chunk filename, and thus was not fetched from the asset manifest - rather than just introduce yet another manifest into the mix as was required, I removed the other asset manifests. This should also have a slight compile speed benefit as generating manifests is slow. This removes some complexity from the app (though, granted, adds some other complexities). 

### Before
Load non-legacy script in the nomodule script tags.

### After
Load legacy script in the nomodule script tags.


shtukas (approved):

src/web/server/document.tsx
```diff
@@ -109,27 +106,31 @@ export const document = ({ data }: Props): string => {
...
+		// {
+		//     filename: 'elements-RichLinkComponent.js',
+		//     scriptType: 'script',
+		//     chunk: 'elements-RichLinkComponent',
+		//     url: '/elements-RichLinkComponent.js',
+		//     path: '/Users/gareth_trufitt/code/dotcom-rendering/dist/elements-RichLinkComponent.js',
+		//     type: 'mainAsset',
+		//     linkType: 'preload'
+		// }
+		.reduce(
```
oliverlloyd, line 126: Whenever I see reduce my brain is like whelp 🤷‍♂️ , something with adding stuff? I know that there's an accumulator and that we're interating but I trip up because it's not explicitly written out whats happening and because I'm not used to using this function I really have to _think_ to see it, it's like javascript written in French for me, I _can_ read it but I'm left wondering if I missed some subtleties...

I think that this is probably taking one array, iterating it and producing another one after doing a transformation? Maybe map works?

On the other hand, this might be a perfectly valid use case for reduce and maybe I need to study it more!

src/web/server/htmlTemplate.ts
```diff
@@ -221,9 +219,8 @@ export const htmlTemplate = ({
 						keywords,
 					).replace(/%20/g, '+')}" />
                 </noscript>
-                ${priorityScriptTags.join('\n')}
                 ${loadableConfigScripts.join('\n')}
-                ${loadableScripts.join('\n')}
+                ${priorityScriptTags.join('\n')}
```
oliverlloyd, line 223: I presume it was but just in case: this change in order was intentional?

oliverlloyd (approved):
Well done for spotting this, really nice work 👍 This was the iPad issue, right?

To be honest, I do struggle a bit to follow along reviewing these changes in context; the pace of change in this area has been rapid recently (which is great!) and I've not had a chance to absorb it all so I lack the wider picture and can't provide as complete a review as I'd want.

I like how we're not taking a similar approach for both legacy and not, that makes a lot of sense, but, as you say, this appraoch is now a bit more complex itself.

One thing that might help, the name `getByChunkName`. I can see this is taking the chunk but I have to dig into it to see it is returning the legacy and non legacy paths to a script. I wonder if this can be refactored. Maybe `getScriptArrayFromChunkName`? Or something like that. But maybe it's better to look at this outside of this PR, I'm concious I don't have a complete mental model so I can't really judge this properly.

src/web/server/htmlTemplate.ts
gtrufitt: Yeah, I removed `loadableScripts`, it's now a part of the `priorityScriptTags` array

gtrufitt:
> One thing that might help, the name getByChunkName. I can see this is taking the chunk but I have to dig into it to see it is returning the legacy and non legacy paths to a script. I wonder if this can be refactored. Maybe getScriptArrayFromChunkName? Or something like that.

I'll do this 👍 I'm going to follow up with that though as I've fully tested this PR as is and want to get a fix out and then follow up with tests.

>  I do struggle a bit to follow along reviewing these changes in context; the pace of change in this area has been rapid recently (which is great!) and I've not had a chance to absorb it all so I lack the wider picture and can't provide as complete a review as I'd want.

It's complex, I've been working on it and struggle myself! I am going to diagram it out.

src/web/server/document.tsx
gtrufitt: Yea, valid. I don't think I can use map because I need to `merge` the arrays, but maybe it would help with a simple `let` and `forEach`. Again, I'll give it a go after this PR with tests and a refactoring PR.

gtrufitt:
> This was the iPad issue, right?

I don't think it is the ipad issue, but it is the Chrome 49 issue 😆

PR #2522, opened 2021-02-11
Orange Kicker

oliverlloyd:
## What does this change?
Makes the kicker orange for special report comment pieces

### Before
Screenshot 2021-02-11 at 08 41 45

### After
Screenshot 2021-02-11 at 08 40 14

## Why?
So that we correctly represent both the fact the piece is opinion but also a special report


liywjl (commented):
👍

liywjl (approved):

PR #2523, opened 2021-02-11
Run file through prettier:fix

tjmw:
### Before

Prettier complains about the formatting of `gen-fixtures.js`.

### After

Prettier is happy.


liywjl (approved):
might be worht exploring if husky can fix this before commit

oliverlloyd (approved):
Thanks for fixing this

PR #2525, opened 2021-02-11
Click to view for all element types

frj:
## What does this change?

This change is rollout the ClickToView wrapper for all components that need it. 

This builds on the following PR:

https://github.com/guardian/dotcom-rendering/pull/2474
https://github.com/guardian/dotcom-rendering/pull/2408

This feature is currently behind a feature switch, you can turn this on using the link:

https://www.theguardian.com/opt/in/click-to-view

## Details

The ClickToView implements the 'click to view' behaviour and is wrapped around the elements component in ArticleRender. As click to view requires javascript in client to work, any component wrapped in needs to be hydrated, in App.

Ideally changes to the 'tracking' status of an element in CAPI would immediately be reflected in the rendering of
that component.

However due to the overhead associated with hydration we chose not to hydrate elements that are not _currently_ classified as tracking. For example, vimeo might stop supporting their 'do-not-track' feature as some point, but we do not wrap the VimeoBlockComponent in a ClickToView component to avoid unnecessary hydration of that component.

In addition the lists of elements stored in window.guardian used to drive hydration in the client are filtered on the 'tracking' status provided by CAPI. This handles the case where a element type can contain both tracking and non-tracking content. In the case where an element is not tracking and the 'click to view' overlay is not rendered, there is no need to hydrate the component.

Here is a list of elements in the 'frontend' article json model that can contain third party content with details of the decision to wrap them in ClickToView or not.

Element | Wrapped in Click to View | Details
--------|--------|-------
DocumentBlockElement | yes | Can contain both Scribd (tracking) and DocumentCloud (non-tracking)
EmbedBlockElement | yes | Can contain almost anything including content which has its own specific element type eg youtube, twitter
InstagramBlockElement | yes | Facebook
MapBlockElement | yes | maps are provided by google who track
PullquoteBlockElement | yes | It is possible to include an iframe in a pullquote, its arguable if this should be wrapped in click to view as this rarely if ever happens
SoundcloudBlockElement | no | Soundcloud claim to not track users
SpotifyBlockElement | yes | Spotify track users
TweetBlockElement | no | twitter support a [do-not-track](https://developer.twitter.com/en/docs/twitter-for-websites/privacy) feature
VideoBlockElement | no | These contain videos from random providers which may track. However we do not currently render these videos 😱
VideoFacebookBlockElement | yes | Facebook
VideoVimeoBlockElement | no | Vimeo provide a 'do-not-track' feature
VideoYoutubeBlockElement | no | Youtube provide a 'do-not-track' feature
WitnessTypeBlockElement | no | Witness is part of the Guardian, post can contain Youtube videos which supports a 'do-not-track' feature.

### Before
Screenshot 2021-02-08 at 16 27 48

### After
Screenshot 2021-02-10 at 14 26 48

## Why?

This is part of the work to deal with concerns over users privacy over the ability of third parties to track the users via content embedded in the guardians products


gtrufitt:
Looks good other than:

```
PullquoteBlockElement | yes | It is possible to include an iframe in a pullquote, its arguable if this should be wrapped in click to view as this rarely if ever happens
```

Wrapping every pullquote has to be a no go for us. 

Can it be restricted on the tools side to avoid iframes in pullquotes? If not, we'll have to check if it contains and iframe *before* wrapping it.

src/web/components/App.tsx
```diff
@@ -629,24 +637,181 @@ export const App = ({ CAPI, NAV }: Props) => {
...
+					index={element.pullquoteIndex}
+				>
+					<ClickToView
+						role={element.role}
+						isTracking={element.isThirdPartyTracking}
+						source={element.source}
+						sourceDomain={element.sourceDomain}
+						abTests={CAPI.config.abTests}
+					>
+						<PullQuoteBlockComponent
```
gtrufitt, line 758: As mentioned, this seems wild, but does this *only* hydrate when Composer *knows* there is an iframe?

docs/architecture/025-click-to-view.md
```diff
@@ -0,0 +1,74 @@
+# Click To View
+
+## Overview
+
+We support embedding may types of third party content in our products. There is the potential for the third party
```
oliverlloyd, line 5:
```suggestion
We support embedding many types of third party content in our products. There is the potential for the third party
```

docs/architecture/025-click-to-view.md
```diff
@@ -0,0 +1,74 @@
...
+to do a certain amount of tracking of our users activity of their use of our product using their embedded content.
+
+In an attempt to give the use choice about being tracked by the third parties we have implemented a
```
oliverlloyd, line 8:
```suggestion
In an attempt to give the reader choice about being tracked by the third parties we have implemented a
```

docs/architecture/025-click-to-view.md
```diff
@@ -0,0 +1,74 @@
+# Click To View
+
+## Overview
+
+We support embedding may types of third party content in our products. There is the potential for the third party
+to do a certain amount of tracking of our users activity of their use of our product using their embedded content.
```
oliverlloyd, line 6:
```suggestion
to do a certain amount of tracking of our reader's activity of their use of our product using their embedded content.
```

src/lib/content.d.ts
```diff
@@ -1,6 +1,12 @@
 // -------------------------------------
 // Elements
 // -------------------------------------
+interface ThirdPartyEmbeddedContent {
```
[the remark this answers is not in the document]
oliverlloyd: Ah, yes, I added some of these recently - we should be able to remove them now

src/lib/content.d.ts
```diff
@@ -252,12 +266,15 @@ interface ProfileAtomBlockElement {
 	role?: RoleType;
 }
 
-interface PullquoteBlockElement {
+interface PullquoteBlockElement extends ThirdPartyEmbeddedContent {
 	_type: 'model.dotcomrendering.pageElements.PullquoteBlockElement';
 	html?: string;
-	role: string;
+	role: RoleType;
```
oliverlloyd, line 272: I'd be careful with doing this, it's very possible that we are being sent data that does not confirm to `RoleType` which is why this is a string. I recall a problem where `halfWidth` was being sent as `halfwidth` although that may have just been for RichLinks

src/web/components/App.tsx
```diff
@@ -629,24 +637,181 @@ export const App = ({ CAPI, NAV }: Props) => {
...
+							embedUrl={element.embedUrl}
+							height={element.height}
+							width={element.width}
+							title={element.title}
+						/>
+					</ClickToView>
+				</HydrateOnce>
+			))}
+			{CAPI.embedBlockElements.map((element) => (
+				<HydrateOnce
```
oliverlloyd, line 662: Can't recall if this is a thng still (does React 17 fix this?) but we might need a `key` prop here (and elsewhere)

src/web/lib/ArticleRenderer.tsx
```diff
@@ -194,23 +194,45 @@ export const ArticleRenderer: React.FC<{
...
+						>
+							<ClickToView
+								role={element.role}
+								isTracking={element.isThirdPartyTracking}
+								source={element.source}
+								sourceDomain={element.sourceDomain}
+								abTests={abTests}
+							>
+								<EmbedBlockComponent
+									key={i}
```
oliverlloyd, line 231: I think this key can move up now to the `Figure`? This was quite possibly already an issue prior to this PR to be fair!

oliverlloyd (approved):
Great work and 👍 for the useful description

I would echo what @gtrufitt said about pull quotes, that seems unlikely to be a thing and they get used a lot so hydrating them would be a shame

src/model/window-guardian.ts
```diff
@@ -88,6 +88,14 @@ export const makeGuardianBrowserCAPI = (CAPI: CAPIType): CAPIBrowserType => {
 		);
 	};
 
+	const thirdPartyTrackingElementsOnly = <
```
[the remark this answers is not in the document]
frj: Yes, sounds good, ill add that.

src/web/components/App.tsx
frj: Yes so pullquotes will only be marked as 'tracking' if it contains an iframe with a src domain that isnt in our white list of non tracking domains. As such the overlay wont get rendered and it wont get hydrated.

src/web/components/App.tsx
frj: Ah ok, interesting, I did wander what those ids were for.

frj:
There are a couple of issues to do with pull quotes.

Although its technically possible to put embedded content into a pullquote, as far as I can tell its never actually happened.

So Im thinking maybe I take out the ClickToView code for pullquotes, Ill discuss with David B about putting some restrictions into composer to stop this from happening.

liywjl (approved):
👍  LGTM

jfsoul:
> These contain videos from random providers which may track. However we do not currently render these videos 😱

^ Regarding VideoBlockElement, do you have an example? I thought we did render all videos nowadays.

frj:
@jfsoul I was referring to this line of code:

https://github.com/guardian/dotcom-rendering/blob/main/src/web/lib/ArticleRenderer.tsx#L600-L601

Here is an example one of these in live:

https://www.theguardian.com/music/2016/sep/06/oasis-documentary-supersonic-first-trailer-amy-liam-noel-gallagher.json?dcr=true

PR #2526, opened 2021-02-11
More Card Stories

oliverlloyd:
## What does this change?
Adds more stories for `Card`, showing a range of variations for different `format` values

Eg. For `Design.Live` we see
Screenshot 2021-02-11 at 12 42 24


liywjl (approved):
👍

PR #2527, opened 2021-02-11
Fix no unsafe member access part 2

liywjl:
## What does this change?
- enable `@typescript-eslint/no-unsafe-member-access`
- fix linting issues

## What is `no-unsafe-member-access`?

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unsafe-member-access.md

We should not be using keys on objects that may not nessesarly exist. A common cause of issues is when using `any` as a type. The result being that if you try get a nested property of that key, TS will throw an error saying it has no assurance that the key is present in the object


scripts/jest/setup.ts
```diff
@@ -60,6 +60,7 @@ const windowGuardian = {
 };
 
 // Stub global Guardian object
+// @ts-ignore
 window.guardian = windowGuardian;
```
liywjl, line 64: we shouldn't be allowed to do this, but know its only in the case of Jest, I think its fine to ignore

src/amp/components/Ad.test.tsx
```diff
@@ -47,28 +49,34 @@ describe('AdComponent', () => {
 		expect(ampAdElement).not.toBeNull();
 
 		if (ampAdElement) {
-			const usRtcAttribute = ampAdElement[0].getAttribute('rtc-config');
-			const auRtcAttribute = ampAdElement[1].getAttribute('rtc-config');
-			const rowRtcAttribute = ampAdElement[2].getAttribute('rtc-config');
+			const usRtcAttribute: ObjectType = JSON.parse(
```
liywjl, line 52: defining what is parsed as `ObjectType` means we can try access `.urls` from that object. You will notice this is a common work around in the PR for anything that had implicitly `any`

src/amp/components/WithAds.tsx
```diff
@@ -49,6 +49,9 @@ export const WithAds = ({ items, adSlots, adClassName, adInfo }: Props) => {
 			return (
 				<>
 					{item}
+					{/* TODO: we should not be assuming the JSX has an ID property and using it as such, we should update `items` to contain IDs and JSX elements */}
+					{/*
+                    // @ts-ignore */}
```
liywjl, line 54: !!!! 🔴   This will need to be fixed in a future PR (there is already a lot happening here) 🔴  !!!!

We are accessing `id` as a prop from a JSX Element. I am not sure why this currently works, but we should instead send down an array continaing `ID` and `JSX.Elements[]`

src/amp/lib/scripts.ts
```diff
@@ -29,8 +29,10 @@ export const extractScripts: (
 					return null;
 			}
 		})
-		.filter(notEmpty)
-		.filter(unique) as string[];
```
liywjl, line 33: https://jakearchibald.com/2021/function-callback-risks/

src/amp/server/render.tsx
```diff
@@ -72,6 +72,7 @@ export const render = ({ body }: express.Request, res: express.Response) => {
 		if (e instanceof TypeError) {
```
liywjl, line 72: not sure if this is the best way of determining the difference between `400` and `500` error?

src/app/aws/aws-parameters.ts
```diff
@@ -47,16 +47,18 @@ const getAllParams = function getGuardianConfigurationRecursiveStep(
...
-			return params;
-		}
-		return getAllParams(
-			stage,
-			params.concat(response.Parameters),
-			response.NextToken === '' ? undefined : response.NextToken,
-		);
-	});
+	return getParams(stage, token).then(
+		(response: Response & { NextToken?: any; Parameters?: any }) => {
```
liywjl, line 51: I am assuming that what is returned is a `Response` type with some AWS caviats `{ NextToken?: any; Parameters?: any }`

src/app/server.ts
```diff
@@ -37,7 +37,8 @@ const buildUrlFromQueryParam = (req: Request) => {
 	// Note. This is the same as how dev-server.js works
 	const DEFAULT_URL =
 		'https://www.theguardian.com/money/2017/mar/10/ministers-to-criminalise-use-of-ticket-tout-harvesting-software';
-	const url = new URL(req.query.url || DEFAULT_URL);
+	const query = req.query as { url?: string } | undefined;
```
liywjl, line 40: by splitting out `req.query` I can make type assertions to avoid linting erros

src/model/clean.ts
```diff
@@ -20,5 +20,6 @@ export const clean = compose(
 			minifyCSS: true,
 			minifyJS: true,
 		}),
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
 	DOMPurify.sanitize,
```
liywjl, line 24: was unable to get types to be pulled from `@types/dompurify`

src/model/enhance-blockquotes.test.ts
```diff
@@ -10,10 +10,14 @@ describe('Enhancing blockquotes', () => {
 	});
 
 	it('adds the quoted prop when the quoted class was found', () => {
-		const input = {
+		const input: CAPIType = {
```
liywjl, line 13: `enhanceBlockquotes` expect param to be of type `CAPIType`, so we need to add type definition. But doing so means we have some missing keys in the object (`id`, `primaryDateLine ` and `secondaryDateLine`)

src/web/components/App.tsx
```diff
@@ -294,7 +294,8 @@ export const App = ({ CAPI, NAV }: Props) => {
 				loadScript('https://www.google-analytics.com/analytics.js');
 				loadScript(window.guardian.gaPath);
 			} else {
-				(window as any).ga = null;
+				// @ts-ignore
+				window.ga = null;
```
liywjl, line 298: if `consentGiven` is false we set `window.ga = null`, even though this is an illigal operation in TS, so using `// @ts-ignore` makes sense here

shtukas (approved):

scripts/jest/setup.ts
nicl: I think it's worth commenting here why, or possibly casting to the required type (with comment) as @ts-ignore is a bit cryptic.

src/app/aws/aws-parameters.ts
nicl: Parameters can be narrowed to `AWSParameter[]` here I think.

nicl: Getting really specific, we could just use the AWS types here

```
import { Parameter } from 'aws-sdk/clients/ssm';
import { PromiseResult } from 'aws-sdk/lib/request';

// ...

const getParams = function getAWSParameterStoreParameters(
	stage: string,
	token: string | undefined = undefined,
): Promise<PromiseResult<AWS.SSM.GetParametersByPathResult, AWS.AWSError>> {
	const params = {
		Path: `/${STACK}/${stage}/`,
		Recursive: true,
		WithDecryption: true,
		NextToken: token,
	};

	return ssm.getParametersByPath(params).promise();
};

// a recursive function to retrieve all pages of guardian configuration

const getAllParams = function getGuardianConfigurationRecursiveStep(
	stage: string,
	params: Parameter[] = [],
	token: string | undefined = undefined,
): Promise<Parameter[]> {
	return getParams(stage, token).then((response) => {
		if (!response.NextToken) {
			return params;
		}
		return getAllParams(
			stage,
			params.concat(response.Parameters || []),
			response.NextToken === '' ? undefined : response.NextToken,
		);
	});
};```

File: src/web/lib/getBrazeUuid.ts [hunk 11]
```diff
@@ -1,12 +1,23 @@
...
+			(json: {
+				user?: {
+					privateFields?: {
+						brazeUuid?: string;
+					};
+				};
+			}) =>
+				json.user &&
+				json.user.privateFields &&
+				json.user.privateFields.brazeUuid,
```
nicl, line 19: Can we still use `?.` for indexing here?

nicl (approved):
Looks good! There are a lot of cases so being pragmatic makes sense.

src/app/logging.ts
```diff
@@ -1,12 +1,18 @@
...
+	LoggingEvent,
+} from 'log4js';
 
 const logLocation =
 	process.env.NODE_ENV === 'production'
 		? '/var/log/dotcom-rendering/dotcom-rendering.log'
 		: `${path.resolve('logs')}/dotcom-rendering.log`;
 
-const logFields = (logEvent: any): any => {
+const logFields = (logEvent: LoggingEvent): any => {
```
oliverlloyd, line 15: 👍 nice

src/model/enhance-blockquotes.test.ts
```diff
@@ -10,10 +10,14 @@ describe('Enhancing blockquotes', () => {
...
 			...example,
 			blocks: [
 				{
+					id: '123',
+					primaryDateLine: 'Wed 9 Dec 2020 06.30 GMT',
+					secondaryDateLine:
+						'Last modified on Wed 9 Dec 2020 13.40 GMT',
```
oliverlloyd, line 20: Be good to break these out and spread them so that the intentions of the tests are kept pure

oliverlloyd (approved):
Huge amounts of great work!

Assuming you're okay with updating those tests and with Nic's comments then 👍

src/web/lib/getBrazeUuid.ts
liywjl: yes I forgot we can since the node update thx 👍

PR #2528, opened 2021-02-11
SlotBodyEnd to use generalised messagePicker

rhystmills:
Co-authored with @tjmw 

## What does this change?
SlotBodyEnd (the end of article slot) has previously used a single category of Epic - the reader revenue Epic. However - the Braze web integration is set to establish a second category of Epic delivered via Braze.

In order to support these two categories, we have set the ReaderRevenueEpic to be a candidate for the Epic slot chosen by a generalised `messagePicker` - already used similarly for reader revenue banners elsewhere on DCR. 

This PR moves control of messages rendered in SlotBodyEnd to the generalised `messagePicker` system to a future PR, where we will add the Braze-delivered Epic category. The `messagePicker` will check whether multiple Epics are eligible to display and return the highest priority eligible Epic.

There are some additional changes relating to tidying up existing issues in the code, namely:
- Removal of the unused 'isSensitive' prop
- The addition of type information added in response to lingering `TODO` comments in the code

Epic rendering and Ophan logs have been tested locally.

### Before
SlotBodyEnd is responsible only for delivering the ReaderRevenueEpic to App.

### After
Selection of the ReaderRevenueEpic is controlled by the `messagePicker` in SlotBodyEnd, allowing other Epic types to be added alongside the ReaderRevenueEpic.

## Why?
To allow us to add the Braze Epic in a future PR.


src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx
```diff
@@ -108,25 +128,72 @@ const buildPayload = async (props: Props) => {
...
+				contributionsPayload,
+				`${contributionsServiceUrl}/epic${queryString}`,
+			),
+		)
+		.then((response) => {
+			dataPerf.end();
+			return checkForErrors(response);
+		})
+		.then((response) => response.json())
+		.then((json: { data: PreEpicConfig }) => {
```
tjmw, line 177: So `PreEpicConfig` describes what we get back from the support-dotcom-components service? I think when the contributions service has nothing to show an empty object is returned (also suggested by the `json.data` check a couple of lines below), so I think data should be optional here:

```js
.then((json: { data?: PreEpicConfig }) => {
```

src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx
```diff
@@ -142,132 +209,39 @@ const MemoisedInner = ({
...
-	contentType,
-	sectionName,
-	shouldHideReaderRevenue,
-	isMinuteArticle,
-	isPaidContent,
-	isSensitive,
-	tags,
-	contributionsServiceUrl,
-}: Props) => {
-	if (isSignedIn === undefined || countryCode === undefined) {
```
tjmw, line 248: Just a note for future readers because it's not immediately clear where this check lives in the new code: waiting for `isSignedIn` and `countryCode` to be defined before calling the support-dotcom-components service is now handled by the `useOnce` in `SlotBodyEnd`.

src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx
rhystmills: Ah great shout - I'll make that change.

src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx
```diff
@@ -108,25 +128,72 @@ const buildPayload = async (props: Props) => {
 			mvtId: Number(getCookie('GU_mvt_id')),
 			countryCode: props.countryCode,
 		},
-	};
+	} as Metadata; // Metadata type incorrectly does not include required hasOptedOutOfArticleCount property
```
tomrf1, line 131: we should update this model - but the question is whether we want to keep maintaining automat-client or publish from dotcom-components instead...

tomrf1 (approved):
looks good!

src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx
rhystmills: Thanks for looking over this! I think ultimately it makes sense for `support-dotcom-components` to be the home for the types.

If we collectively think it's worthwhile I can make an intermediate fix for `automat-client` since I have a related PR open there already.

PR #2529, opened 2021-02-11
Special CardAge

oliverlloyd:
## What does this change?
Decides the colour used for the card age using `palette`

| Before        | After           |
| ------------- |:-------------:|
| b1 |  Screenshot 2021-02-11 at 15 48 08      |


shtukas (approved):

liywjl (approved):
👍

oliverlloyd:
[Special CardAge](https://trello.com/c/LA6HVBnu/2444-special-cardage)

gtrufitt:
Is lifestyle right? Looks like we'll need a specific colour. 

I'm confused what this PR is dealing with? It goes beyond special report card doesn't it? But there's not screenshots of all pillars?

oliverlloyd:
> Is lifestyle right? Looks like we'll need a specific colour.
> 
> I'm confused what this PR is dealing with? It goes beyond special report card doesn't it? But there's not screenshots of all pillars?

I think my description before was misleading. The _code change_ here is simply to add palette into Card and use it to decide the CardAge colour. The change resulting from this is around special reports. But in addition to this I've been having a conversation with Harry about Card colours in general and so I put those additional screenshots in the description to highlight that there is other work and that this PR _isn't_ fixing it - I should have made that clearer! In the end, we had the discussion using the visual diffs so I've update the PR description to be more forcussed.

PR #2530, opened 2021-02-11
No more Trump (video diffs)

oliverlloyd:
## What?
Turns off Chromatic snapshots for the Witness YouTube story.

## Why?
These stories are causing false negatives because a Watch on YourTube logo is appearing intermittently. We're already disabling Chromatic on other youtube stories where we're not using the poster overlay.


jfsoul (approved):

liywjl (approved):
could revert them back to original Videos? (just a preference, not a necessity)

PR #2531, opened 2021-02-11
Create a buildCheck action as a basic sense check on the build files

gtrufitt:
## What does this change?

Adds a github action that checks the build and build manifest files are as expected.

Some safety that bundles are transpiled correctly, and that we'll be able to access the JS files as expected.


liywjl (approved):
awesome work!!!

PR #2535, opened 2021-02-11
Add basic Cypress test to cover RR banner request

tjmw:
## What does this change?

Adds a basic test to ensure that a request is made to the support-dotcom-components service to find out if there's a banner to show.

## Why?

To give us confidence that platform changes haven't affected the banner logic (for example changes to the message picker logic). This is a follow up to #2524 as I think it would have caught the issue which that PR fixed.


liywjl (approved):
👍 LGTM

PR #2536, opened 2021-02-12
Carousel A/B test set up

buck06191:
## What does this change?
Adds the A/B test for the new carousel designs.

## Why?
Compare two new carousel designs and the existing fixed content container. Should see an increase in CTR for the carousel (hopefully)


buck06191:
[Carousel Test: Next curated container design test](https://trello.com/c/e1WPS9hm/1969-carousel-test-next-curated-container-design-test)

shtukas (approved):

PR #2539, opened 2021-02-12
Carousel: final design changes

buck06191:
## What does this change?
Fixes small design requirements before the AB test goes live.

Requires https://github.com/guardian/frontend/pull/23548 to be live so that we have the image srcset.

### Before
image

### After
image

## Why?
Final design changes for AB test


HarryFischer (approved):
👌

index.d.ts
```diff
@@ -532,6 +532,7 @@ type OnwardsType = {
 	url?: string;
 	ophanComponentName: OphanComponentName;
     pillar: Theme;
+	isCuratedContent?: boolean;
     isFullCardImage?: boolean
```
liywjl, line 536: prettier indentation issue

liywjl (approved):
LGTM 👍

oliverlloyd (commented):
This comment is outside this PR and not a reflection on this work at all, I'm just using it as a base for dicsussion.

Maybe we need to rethink our approach to adding features for testing.

@gtrufitt @JamieB-gu?

These changes are being done outside of our model - they're not part of `format` - and this causes a lot of problems because they only way to make that work is to add additional props. This means that we're going to have to either:

1) Decide to continue with this approach of working around `Format` or
2) We'll have to refactor away these props and return to using the model

I think there's strong backing for using `format` and that we want to maintain it so that means we will need to go with option 2.

I think that there's a lot of value in trying out new things and this Carousel is a great achievement with some really awesome work behind it. I know @buck06191 and @HarryFischer both have really gone the extra mile here. But if we had tried to fit these changes into the model prior to implementing the test their jobs would have been a lot easier.

As a way to improve things, and to make all our jobs easier, perhaps in future new designs would need to be worked into the model as a pre-requisit for being implemented? If the design _can't_ be modelled, then the question should be: if we can't model it, do we still want to test it?

buck06191:
Looks like we need to work on the image size. I think we're only pulling through a `thumbnail` image from CAPI inside `com.gu.contentapi.client.model.v1.ContentFields`
https://github.com/guardian/frontend/blob/3e57b62e71ed6c4f164e9d2eff94241b44a63d2c/onward/app/feed/DeeplyReadAgent.scala#L152

JamieB-gu:
> As a way to improve things, and to make all our jobs easier, perhaps in future new designs would need to be worked into the model as a pre-requisit for being implemented?

Looking at the code it seems the if something is "curated content" we change the copy and the text colour, is that correct? What determines whether something is "curated content"? Does that come from CAPI or the fronts tool?

On whether this should be worked into the model, I agree that it should @oliverlloyd. Only thing I'm unsure about is how we handle that for tests? Should we be very strict and say we _have_ to update the top-level model as a pre-requisite for running this kind of test? Or do we run the test using a workaround, and if the test is successful and we want to implement it going forward only then do we refactor it into the model?

src/web/components/Onwards/Carousel/Carousel.tsx
```diff
@@ -607,9 +621,14 @@ export const Carousel: React.FC<OnwardsType> = ({
...
 							webPublicationDate,
-							image: imageUrl,
+							image: fallbackImageUrl,
+							carouselImages,
 							kickerText,
 						} = trail;
+						const imageUrl =
+							isFullCardImage && carouselImages
+								? carouselImages['460']
+								: fallbackImageUrl;
```
oliverlloyd, line 631: 👍 this makes sense. It's basically correct in that this represents exactly what we're doing

src/web/lib/decidePalette.ts
```diff
@@ -347,11 +347,13 @@ const fillCaptionTriangle = (format: Format): string => {
 };
 
 const fillCardIcon = (format: Format): string => {
+	// Setting Card clock colour for immersive cards to all be dark grey
+	if (format.display === Display.Immersive) return neutral[60];
```
oliverlloyd, line 351: I think this is a merge conflict gone awry? These lines aren't on `main` and don't seem to be related

oliverlloyd (approved):
Just an issue with a merge conflict where we're adding code wrongly I think but otherwise 👍

oliverlloyd:
> Just an issue with a merge conflict where we're adding code wrongly I think but otherwise 👍

Oh actually, looking at the visual diffs maybe this _is_ intentional? If so, ignore me!

src/web/lib/decidePalette.ts
buck06191: I think these should be there from this PR rather than  `main`. There's the pictures stuff in here as well as a few small design fixes. PR for this should be here: https://github.com/guardian/dotcom-rendering/pull/2539/commits/800012b14819b27f71e18630ff8f3da46396de0c

buck06191:
[📦 Last minute design changes/checks](https://trello.com/c/MCBxrbDZ/2458-%F0%9F%93%A6-last-minute-design-changes-checks)

jamesgorrie:
> Should we be very strict and say we have to update the top-level model as a pre-requisite for running this kind of test?

My 2¢ on this is to ask the question: do we _need_ to work outside our model for a what we are trying to test? I know this is sometimes true, but might be a hint that your test is too complicated. It can also make it hard to unwind if it's not successful.

Saying this I am outside of understanding what this test is / was. Where did the `isCuratedContent` prop come from?

PR #2540, opened 2021-02-12
Wrap Instagram Embed in and iframe

frj:
## What does this change?
This change puts the embed html from InstagramBlockElements into the srcdoc of an iframe, rather than just writing it into the article body.

There are some pros and cons to doing this described below:

### Pros
- This fixes a problem where the instagram embed js did not work when the embed was wrapped in a ClickToView component. Which injects the embed code into the page when the component is clicked rather the embed code being there at page load
- This somewhat isolates the instagram java script from the rest of the article page.
- Many instagram embeds are embedded as 'embed' type elements and get rendered using the UnsafeEmbedBlockComponent, which renders the embed in this way, making the rendering of Instagram embeds more consistent.

### Cons
- This change requires some additional javascript to be run to resize the iframe to match the underlying embed. This is already being run for the iframes created by the UnsafeEmbedBlockComponent. This is no big deal but is just some additional complexity. There is a layout shift caused by this resizing, but there is also a layout shift without this change.

### Before

Screenshot 2021-02-12 at 15 20 24

### After
Screenshot 2021-02-12 at 15 20 24


src/web/components/elements/InstagramBlockComponent.tsx
```diff
@@ -1,24 +1,29 @@
...
+}> = ({ element, index }) => {
 	return (
-		<div className={widthOverride}>
-			<div
-				data-cy="instagram-embed"
-				dangerouslySetInnerHTML={{ __html: unescapeData(element.html) }}
-			/>
-		</div>
+		<iframe
+			className={`${fullWidthStyles} js-embed__iframe`}
```
oliverlloyd, line 20: I don't think we need this class `js-embed__iframe`

src/web/components/elements/InstagramBlockComponent.tsx
```diff
@@ -1,24 +1,29 @@
...
-import { unescapeData } from '@root/src/lib/escapeData';
 import { css } from 'emotion';
 
-const widthOverride = css`
-	iframe {
-		/* The instagram embed js hijacks the iframe and calculated an incorrect width, which pushed the body out */
-		min-width: 300px !important;
-	}
+const fullWidthStyles = css`
+	width: 100%;
```
oliverlloyd, line 6: What was the use case for having this before. Will this still solve that?

src/web/components/elements/InstagramBlockComponent.tsx
```diff
@@ -1,24 +1,29 @@
 import React from 'react';
 
-import { unescapeData } from '@root/src/lib/escapeData';
 import { css } from 'emotion';
 
-const widthOverride = css`
-	iframe {
-		/* The instagram embed js hijacks the iframe and calculated an incorrect width, which pushed the body out */
```
oliverlloyd, line 8: This comment suggests we might need to handle the instagram js doing something to override our width styles

oliverlloyd (commented):
This seems like a sensible approach. Indeed, it feels like an improvement so if it also fixes the loading issue then all the better!

I think we need to give some thought to that width issue though. I'd be keen to understand that more.

src/web/components/elements/InstagramBlockComponent.tsx
frj: Well... In the case where the click to view is not in operation eg if we just roll out this PR without the ClickToView changes, this class is required as it gets picked up by this piece of code:

https://github.com/guardian/dotcom-rendering/blob/main/src/web/browser/embedIframe/embedIframe.ts#L3-L4

This is how the UnsafeEmbedBlockComponent work which Im basing this off. There is most probably a better way of doing this.

src/web/components/elements/InstagramBlockComponent.tsx
frj: I think previously there was an issue where the instagram js was creating an iframe that was too big for the article.

In this new approach I think the instagram code will be 'trapped' inside the our iframe which this sets to 100% width.

I guess in the event instagram creates something that is too big it would get clipped, which isnt ideal but wouldnt break anything outside the component.

TBH Ive never been able to reproduce anything like that. 

Also a bunch of Instagram embeds get rendered like this all the time with no complaints, ie the ones in 'embed' type elements.

src/web/components/elements/InstagramBlockComponent.tsx
frj: I think this is talking about the iframe that the instagram js creates its-self, Im not sure this is a problem with these changes.

src/web/components/elements/InstagramBlockComponent.tsx
oliverlloyd: 😱 that's the first time I've seen a classname used as a selector in DCR! But there it is.

src/web/components/elements/InstagramBlockComponent.tsx
oliverlloyd: I think you're right, now that we create our own iframe we should maintain control. In fact, I wonder if we need any styling here?

oliverlloyd (approved):
Nice work 👍

Lets ship this next week

liywjl (approved):
LGTM 👍

src/web/components/elements/InstagramBlockComponent.tsx
frj: Without this stying the iframe defaults to ~300px wide so I think we do need this.

PR #2546, opened 2021-02-15
Respect Accessibility Preferences

oliverlloyd:
## What does this change?
Fixing an issue where we were looking for the reader's flashing preference as a class on the page when it is actually being saved in localStorage

## Support for `prefers-reduced-motion`
The css media query [prefers reduced motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) is also supported

### I've added the `storage` lib
Reading from local storage can typically involve a lot of boilerplate but this is even more so on dotcom where there is a convention of storing data using the `{ value: obj }` pattern where data needs to be serialised and parsed. To address this I'm using [storage](https://github.com/guardian/libs/blob/main/src/storage.README.md) from `@guardian/libs` which expects this object pattern and contains all the required boilerplate.

### I've removed a test
The previous version of this code was removing the dot completely and there was a test to check for this. I've removed that test because we now continue to render the dot, we just stop it flashing. I did briefly look at testing for if this animation was occurring or not but there didn't seem a good way to do this and I don't think this use case warrants a complex, potentially brittle, solution so I've left it out

### DIFF
--- a/src/web/components/PulsingDot.tsx
+++ b/src/web/components/PulsingDot.tsx
@@ -28,6 +28,9 @@ const livePulse = keyframes`{
 
 const animate = css`
 	animation: ${livePulse} 1s infinite;
+	@media (prefers-reduced-motion: reduce) {
+		animation: none;
+	}
 `;
 
 interface Props {
--- a/src/web/components/PulsingDot.tsx
+++ b/src/web/components/PulsingDot.tsx
@@ -1,17 +1,10 @@
 import React from 'react';
-import { css } from 'emotion';
+import { css, cx } from 'emotion';
 import { keyframes } from '@emotion/core';
 
-export const DISABLE_FLASHING_ELEMENTS_CLASS = 'disable-flashing-elements';
+import { storage } from '@guardian/libs';
 
-const livePulse = keyframes`{
-    0% {opacity: 1;}
-    10% {opacity: .25;}
-    40% {opacity: 1;}
-    100% {opacity: 1;}
-}`;
-
-const pulsingDot = (colour?: string) => css`
+const dotStyles = (colour?: string) => css`
 	color: ${colour && colour};
 	::before {
 		border-radius: 62.5rem;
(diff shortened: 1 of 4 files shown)


oliverlloyd:
[Respect Flashing Preference](https://trello.com/c/L9iS5Z6x/2460-respect-flashing-preference)

liywjl (approved):
LGTM

src/web/components/PulsingDot.tsx
```diff
@@ -23,23 +16,36 @@ const pulsingDot = (colour?: string) => css`
...
-
-	return <span className={pulsingDot(colour)} />;
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+	const flashingPreference = storage.local.get(
+		'gu.prefs.accessibility.flashing-elements',
+	);
+	// flashingPreference is null if no preference exists and explicitly
+	// false when the reader has said they don't want flashing
+	const flashingEnabled = flashingPreference !== false;
+	console.log('flashingEnabled', flashingEnabled);
```
gtrufitt, line 47:
```suggestion

```

src/web/components/PulsingDot.tsx
```diff
@@ -23,23 +16,36 @@ const pulsingDot = (colour?: string) => css`
...
+	return (
+		<span className={cx(dotStyles(colour), flashingEnabled && animate)} />
```
gtrufitt, line 49: Wondering if we should also use the opportunity to respect -prefers-reduced-motion? Either via the media query (and refactor the css bit) or implement a JS helper that you can then pass in as here, something like:

```js
function getPrefersReducedMotion() {
  const QUERY = '(prefers-reduced-motion: no-preference)';
  const mediaQueryList = window.matchMedia(QUERY);
  const prefersReducedMotion = !mediaQueryList.matches;
  return prefersReducedMotion;
}
```

https://www.joshwcomeau.com/react/prefers-reduced-motion/#in-js-land

gtrufitt (approved):
+++++ Great work!

src/web/components/PulsingDot.tsx
oliverlloyd: Great idea! I'll add this 👍

oliverlloyd: I added support for `prefers-reducced-motion` using the media query in the css directly as that seems to work well and is simple

gtrufitt: ❤️

PR #2552, opened 2021-02-15
Emotion 11 source 3 react update

liywjl:
## What does this change?
- Update to Emotion 11
- Update to React 17
- Use `@emotion/react` jsx factory
- Change `className` to `css` to use Emtion 11 factory
- Update to Source v3 (which depends on Emotion 11)
- Use `SerializedStyles` instead of `string` for passing down CSS props
- replace `composeLabsCSS`

## TODO:
- Update Discussion Rendering
- Update Atoms Rendering
- Check if others dependences need updating (Braze?)
- Remove `React` as import

## Files to double check
- src/web/components/elements/RewrappedComponent.tsx:
We are using `React.createElement` but we should probably be using the JSX conversion instead
- src/web/components/SlotBodyEnd.tsx: we hoised package from emotion to the global object. However as these packages names have changed, their contents have likely changed and there is no explicite contract between both systems


liywjl:
closing this PR in favor of https://github.com/guardian/dotcom-rendering/pull/2994

PR #2553, opened 2021-02-15
Record Google Core Vitals and store in BigQuery

nitro-marky:
## What does this change?
This uses the [web vitals library](https://github.com/GoogleChrome/web-vitals) to record core vitals from a page, via Fastly which is then sent to BigQuery, these metrics are:

- First Input Delay
- Cumulative Layout Shift
- Largest Contentful Paint
- First Contentful Paint
- Time To First Byte

[For more info on these check this.](https://web.dev/vitals/)

Once CLS has been collected (one of the key metrics) a Post is sent to Fastly, which then stores the results in BigQuery for later analysis.

**Things to note**

- If a user does not click on the page first input delay will be null
- The ID in the example below is from the localhost, which is why it is wrong

Example data:
Screenshot 2021-02-15 at 14 36 01

### DIFF
--- a/src/model/window-guardian.ts
+++ b/src/model/window-guardian.ts
@@ -27,6 +27,7 @@ export interface WindowGuardianConfig {
 	tests?: { [key: string]: string };
 	ophan: {
 		pageViewId: string;
+		browserId: string;
 	};
 }
 
@@ -62,6 +63,7 @@ const makeWindowGuardianConfig = (
 		tests: config.abTests || {},
 		ophan: {
 			pageViewId: '',
+			browserId: ''
 		},
 	} as WindowGuardianConfig;
 };
--- a/src/model/window-guardian.ts
+++ b/src/model/window-guardian.ts
@@ -63,7 +63,7 @@ const makeWindowGuardianConfig = (
 		tests: config.abTests || {},
 		ophan: {
 			pageViewId: '',
-			browserId: ''
+			browserId: '',
 		},
 	} as WindowGuardianConfig;
 };
--- a/src/web/browser/coreVitals/coreVitals.ts
+++ b/src/web/browser/coreVitals/coreVitals.ts
@@ -74,7 +74,7 @@ export const coreVitals = (): void => {
 				redirect: 'follow',
 				referrerPolicy: 'no-referrer', // no-referrer, *no-referrer-w
 				body: JSON.stringify(jsonData),
-			});
+			}).catch(() => {});
 		}
 	};
 
--- a/src/web/browser/coreVitals/coreVitals.ts
+++ b/src/web/browser/coreVitals/coreVitals.ts
@@ -54,23 +54,27 @@ export const coreVitals = (): void => {
 				break;
 		}
 
+		const endpoint =
+			window.location.hostname === 'm.code.dev-theguardian.com' ||
+			window.location.hostname === 'localhost' ||
+			window.location.hostname === 'preview.gutools.co.uk'
+				? 'http://performance-events.code.dev-guardianapis.com/core-web-vitals'
+				: 'https://performance-events.guardianapis.com/core-web-vitals';
+
 		// If CLS has been calculated
 		if (jsonData.cls !== null) {
-			fetch(
-				'https://performance-events.guardianapis.com/core-web-vitals',
-				{
-					method: 'POST', // *GET, POST, PUT, DELETE, etc.
-					mode: 'cors', // no-cors, *cors, same-origin
-					cache: 'no-cache', // *default, no-cache, reload, force-cache, only-if-cached
-					credentials: 'same-origin', // include, *same-origin, omit
-					headers: {
-						'Content-Type': 'application/json',
-					},
-					redirect: 'follow',
-					referrerPolicy: 'no-referrer', // no-referrer, *no-referrer-w
-					body: JSON.stringify(jsonData),
+			fetch(endpoint, {
+				method: 'POST', // *GET, POST, PUT, DELETE, etc.
+				mode: 'cors', // no-cors, *cors, same-origin
+				cache: 'no-cache', // *default, no-cache, reload, force-cache, only-if-cached
+				credentials: 'same-origin', // include, *same-origin, omit
+				headers: {
+					'Content-Type': 'application/json',
 				},
-			);
+				redirect: 'follow',
+				referrerPolicy: 'no-referrer', // no-referrer, *no-referrer-w
+				body: JSON.stringify(jsonData),
+			});
 		}
 	};
 
--- a/src/web/browser/coreVitals/coreVitals.ts
+++ b/src/web/browser/coreVitals/coreVitals.ts
@@ -1,12 +1,10 @@
 import { getCLS, getFID, getLCP, getFCP, getTTFB } from 'web-vitals';
-import { getCookie } from '../cookie';
-
 
 const jsonData = {
 	'page_view_id': '',
 	'received_timestamp': `${new Date().toISOString()}`,
 	'id': 'prod',
-	'received_date': '2021-02-03',
+	'received_date': new Date().toISOString().slice(0, 10),
     'fid': null,
     'cls': null,
     'lcp': null,
@@ -14,16 +12,14 @@ const jsonData = {
     'ttfb': null,
 }
 
-
-
 export const coreVitals = (): void =>
 {
-    type coreVitalsArgs = {
+    type CoreVitalsArgs = {
         name: string;
         value: number;
     };
 
-const getValue = ({ name, value }: coreVitalsArgs): void => {
+	const getValue = ({ name, value }: CoreVitalsArgs): void => {
 
     switch(name){
         case('FCP'):
@@ -41,7 +37,9 @@ const getValue = ({ name, value }: coreVitalsArgs): void => {
         case('TTFB'):
             jsonData.ttfb = parseFloat(value.toFixed(6));
             break;
-    }
+	}
+
+	// If CLS has been calculated
     if(jsonData.cls !== null)
     {
         fetch('https://performance-events.guardianapis.com/core-web-vitals', {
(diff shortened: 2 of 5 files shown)


src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -0,0 +1,74 @@
...
+export const coreVitals = (): void => {
+	type CoreVitalsArgs = {
+		name: string;
+		value: number;
+	};
+
+	const getValue = ({ name, value }: CoreVitalsArgs): void => {
+		switch (name) {
+			case 'FCP':
+				jsonData.fcp = parseFloat(value.toFixed(6));
```
oliverlloyd, line 27: What about `Math.round(value * 100000) / 100000`? I like how this is just math and there's no type conversion needed.

src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -0,0 +1,74 @@
...
+	ttfb: temp,
+};
+
+export const coreVitals = (): void => {
+	type CoreVitalsArgs = {
+		name: string;
+		value: number;
+	};
+
+	const getValue = ({ name, value }: CoreVitalsArgs): void => {
```
oliverlloyd, line 24: Is `getValue` the right name here? From reading this function it's not returning anything and seems to instead be sending something?

src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -0,0 +1,74 @@
+import { getCLS, getFID, getLCP, getFCP, getTTFB } from 'web-vitals';
+
+// Stops ts assuming json values are null
+const temp: any = null;
```
oliverlloyd, line 4: Will this change at some point (via refactor) or will it be `temp` by design? As in, is this the initial value before it gets set?

src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -0,0 +1,74 @@
...
+					cache: 'no-cache', // *default, no-cache, reload, force-cache, only-if-cached
+					credentials: 'same-origin', // include, *same-origin, omit
+					headers: {
+						'Content-Type': 'application/json',
+					},
+					redirect: 'follow',
+					referrerPolicy: 'no-referrer', // no-referrer, *no-referrer-w
+					body: JSON.stringify(jsonData),
+				},
+			);
```
oliverlloyd, line 59: Do we need to check the response or catch an errors from this call?

src/web/browser/coreVitals/init.ts
```diff
@@ -0,0 +1,10 @@
+import '../webpackPublicPath';
+import { startup } from '@root/src/web/browser/startup';
+import { coreVitals } from './coreVitals';
+
+const init = (): Promise<void> => {
+	coreVitals();
+	return Promise.resolve();
+};
```
oliverlloyd, line 8: You could use [async](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) here

```typescipt
  coreVitals();
  return;
};
```

oliverlloyd: I guess this also where we would set sampling?

oliverlloyd (commented):
This PR seems really simple which I think is hugely impressive considering that this is in fact a really complex area, so well done!

I think it might make things read a bit better if the naming of things was a little bit more indicative to what's happening. When you're been working on something for a while you don't see it but reading this cold I got a bit confused

src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -0,0 +1,74 @@
...
+		}
+	};
+
+	// Set page view and browser ID
+	if (window.guardian && window.guardian.ophan) {
+		jsonData.page_view_id = window.guardian.ophan.pageViewId;
+		jsonData.id = window.guardian.config.ophan.browserId;
```
jfsoul, line 66: it would be great if we could change the `id` column name to `browser_id` before releasing this since it'll be harder to do seamlessly once we have live data coming through

src/web/browser/coreVitals/init.ts
jfsoul: > I guess this also where we would set sampling?

+1

src/web/server/document.tsx
```diff
@@ -176,6 +176,7 @@ export const document = ({ data }: Props): string => {
 			CAPI.config && { src: CAPI.config.commercialBundleUrl },
 			...getByChunkName('sentryLoader'),
 			...getByChunkName('dynamicImport'),
+			...getByChunkName('coreVitals'),
```
jfsoul, line 179: should this be high priority or low priority? (genuine question)

src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -0,0 +1,74 @@
...
+				break;
+			case 'TTFB':
+				jsonData.ttfb = parseFloat(value.toFixed(6));
+				break;
+		}
+
+		// If CLS has been calculated
+		if (jsonData.cls !== null) {
+			fetch(
+				'https://performance-events.guardianapis.com/core-web-vitals',
```
jfsoul, line 46: I suppose we ought to configure this value so it works seamlessly between PROD and CODE/DEV. Not immediately sure where because typically this kind of thing comes from the CAPI data from frontend, but that doesn't seem wise for this. Maybe there's already somewhere sensible to do this in DCR directly. Doesn't seem worth putting in AWS parameter store, just for this.

src/web/browser/coreVitals/coreVitals.ts
nitro-marky: I tried setting a type in the JSON but couldn't so this was a quick way to solve it, I'm sure it can be refactored but it's minimal and works for now.

src/web/server/document.tsx
nitro-marky: I guess Gareth mentioned logging immediately so perhaps I'll move it to the top.

oliverlloyd: As I understand it, these scripts block the page, so my take would be we should only have performance monitoring there if it _needs_ to load ahead of everything else. But even then I'm not sure we guarentee that here?

src/web/browser/coreVitals/init.ts
nitro-marky: Yeah, perhaps I'll add that in and comment it out until it goes fully live

nitro-marky: @oliverlloyd I didn't use async as none of the other init files do, so I was keeping to the pattern. I can add it in if you think it would be better?

oliverlloyd: I used it in the Sentry init file based on a PR review I got there. I don't think it matters a great deal if you use it or not but the review comment I got improved my knowledge of what `async` is and does!

src/web/browser/coreVitals/coreVitals.ts
nitro-marky: The problem with Math.round is we loose the decimals which some metrics need.

oliverlloyd: Same same
Screenshot 2021-02-16 at 10 37 27

rcrphillips:
[Core web vitals - DCR usage of new performance event service](https://trello.com/c/WXXPGQJk/2419-core-web-vitals-dcr-usage-of-new-performance-event-service)

src/web/browser/coreVitals/coreVitals.ts
nitro-marky: Fair enough 👍

src/web/browser/coreVitals/init.ts
nitro-marky: 👍

src/web/server/document.tsx
nitro-marky: Yeah, I'm sure the order is not guaranteed

oliverlloyd: Yeah, the order is insignificant if these are asyncronous actions. But maybe we want to drop `coreVitals` down to:

```
	const lowPriorityScriptTags = generateScriptTags([
		...getScriptArrayFromChunkName('lotame'),
		...getScriptArrayFromChunkName('atomIframe'),
		...getScriptArrayFromChunkName('embedIframe'),
		...getScriptArrayFromChunkName('newsletterEmbedIframe'),
	]);
```

where it will be loaded later in the bottom of the page (not the head, which blocks).

But if we do this it then means this code will execute after the initial page load which might break things here?

nitro-marky: Ideally we need the script running as soon as possible

src/web/browser/coreVitals/coreVitals.ts
nitro-marky: Not really, if the person leaves the page a response is not necessary, even if they stay on the page and there is an error sentry should catch it

src/model/window-guardian.ts
```diff
@@ -27,6 +27,7 @@ export interface WindowGuardianConfig {
 	tests?: { [key: string]: string };
 	ophan: {
 		pageViewId: string;
+		browserId: string;
```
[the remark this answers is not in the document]
nitro-marky: So was I!

jfsoul (approved):
Apart from adding the sampling, waiting for the browser_id column name change and working out the endpoint to use for DEV/CODE/PROD - I think this looks ready, nice one!

src/web/server/document.tsx
oliverlloyd: The comment for high priority scripts is

```
	/**
	 * The highest priority scripts.
	 * These scripts have a considerable impact on site performance.
	 * Only scripts critical to application execution may go in here.
	 * Please talk to the dotcom platform team before adding more.
	 * Scripts will be executed in the order they appear in this array
	 */
```

So we should be *very* cautious when adding to it as we are here.

But I'm unsure how the lib works so I can't say this for sure where it needs to be. Eg. Does this lib need to be loaded first thing to start recording information that it sends later or can it be loaded at a later stage and still capture what it needs? My instinct is that it can come later but I think we should run some tests for this.

oliverlloyd: Could we maybe?


```typescript
// App.tsx
import {getCLS, getFID, getLCP} from 'web-vitals';

const reportStuff = () => {}

useOnce(() => {
  getCLS(reportStuff);
  getFID(reportStuff);
  getLCP(reportStuff);
});
```

No idea if that's valid but might be worth a test

nitro-marky: We can test that impact with this https://github.com/guardian/dotcom-rendering/pull/2559 when the tables are re-created

PR #2554, opened 2021-02-15
Addon Essentials

oliverlloyd:
## What does this change?
Removes some unused addons as well as `@storybook/addon-viewer`, replacing them all with [addon-essentials](https://storybook.js.org/docs/react/essentials/introductionhttps://storybook.js.org/docs/react/essentials/introduction)

## Why?
Now that we're using v6, we can take advantage of ccore addons like [Docs](https://storybook.js.org/docs/react/writing-docs/introduction)


buck06191 (approved):
+1

PR #2558, opened 2021-02-16
Identity | Stop sign in gate design test

coldlink:
## What does this change?
Remove the sign in gate design optimisation test as it's finished.

- Remove the test definition
- Update the `SignInGateMainVariant` audience
- Remove the test from the `SignInGateSelector`
- Remove the test from Cypress tests

~We're keeping the designs so that once the winner is determined we are able to easily roll that out.~

There was no winner so removing all design test related code.


oliverlloyd (approved):
👏  for making the work to remove this test so clean

coldlink:
Updated the PR to actually remove the test designs too, since there was no clear winner, so sticking with the `MainVariant` design.

QuarpT (approved):

PR #2561, opened 2021-02-17
Set DropCap when length is exactly 199

oliverlloyd:
## What does this change?
Allows a dropcap when the number of charators in a paragraph is greater than, _or equal_, to 199.

### Before
Screenshot 2021-02-17 at 08 58 57

### After
Screenshot 2021-02-17 at 08 58 34

## Why?
An issue was raised by CP highlighting that an article (which had 199 charaters exactly in the first paragraph) was showing inconsistent results. The [code in Frontend](https://github.com/guardian/frontend/blob/9468d4bc3e35dc9213f8f5c18d4b88c70b8667cc/common/app/views/support/HtmlCleaner.scala#L539) does specify 199 as the limit but it appears the `text` value being used is calculated in a different way to DCR, giving rise to this difference.


nicl (approved):
CP miss nothing!

jfsoul (approved):
👍  thanks

liywjl (approved):

PR #2563, opened 2021-02-17
Add Badges to Immersive articles on DCR

buck06191:
## What does this change?
Add series badges where display type is `Display.Immersive` and badge is available. Styling done to try match up against current Frontend.

### Before
image

### After
image

## Why?
https://trello.com/c/LqveAsMD/1550-immersive-badges

### DIFF
--- a/src/web/components/SeriesSectionLink.tsx
+++ b/src/web/components/SeriesSectionLink.tsx
@@ -96,7 +96,7 @@ const titleBadgeWrapper = css`
 	margin-right: ${space[2]}px;
 `;
 
-const immersiveTitleBadgeStyle = css`
+const immersiveTitleBadgeStyle = (palette: Palette) => css`
 	display: flex;
 	flex-direction: row;
 	align-items: center;
@@ -105,6 +105,9 @@ const immersiveTitleBadgeStyle = css`
 	line-height: 1.15;
 	/* Offset parent container margins when Immersive */
 	margin-bottom: -10px;
+	background-color: ${palette.background.seriesTitle};
+	box-shadow: -6px 0 0 0 ${palette.background.seriesTitle},
+		6px 0 0 0 ${palette.background.seriesTitle};
 `;
 
 export const SeriesSectionLink = ({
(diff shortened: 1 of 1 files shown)


buck06191:
[Immersive badges](https://trello.com/c/LqveAsMD/1550-immersive-badges)

oliverlloyd (commented):
Does this also need to apply to Comment or GuardianView immersive articles?

buck06191:
> Does this also need to apply to Comment or GuardianView immersive articles?

Good question. I assumed not because the problem seemed to be with Series Badges. Do we have those on Comment and GuardianView pieces?

jfsoul (approved):

shtukas (approved):

oliverlloyd:
> > Does this also need to apply to Comment or GuardianView immersive articles?
> 
> Good question. I assumed not because the problem seemed to be with Series Badges. Do we have those on Comment and GuardianView pieces?

I guess as a rule, you don't see many (any?) immersive comment pieces in a series but they are part of our model and there are designs to make more use or them in the future so I guess the answer is yes and no 🤷 

The problem is this component could do with refactoring. As the use cases it serves have grown, it's got quite jumbled so it's hard to reason through what's going on, and this makes changes like this hard to do

buck06191:
> > > Does this also need to apply to Comment or GuardianView immersive articles?
> > 
> > 
> > Good question. I assumed not because the problem seemed to be with Series Badges. Do we have those on Comment and GuardianView pieces?
> 
> I guess as a rule, you don't see many (any?) immersive comment pieces in a series but they are part of our model and there are designs to make more use or them in the future so I guess the answer is yes and no 🤷
> 
> The problem is this component could do with refactoring. As the use cases it serves have grown, it's got quite jumbled so it's hard to reason through what's going on, and this makes changes like this hard to do

Maybe a good approach here then is to add a card to the backlog to revisit this. I feel like refactoring this component at this point is outside of the scope of both the initial Trello card and this PR, but it does need revisiting.

PR #2567, opened 2021-02-17
Empty Blocks Work

nitro-marky:
## What does this change?

Allows empty block pages to load 

Screenshot 2021-02-17 at 17 22 41

Regarding the white space at the bottom this only seems to occur on localhost.

##Before

Screenshot 2021-02-18 at 14 32 20

##After
On `CODE`

Screenshot 2021-02-18 at 14 31 24

Screenshot 2021-02-18 at 14 31 05


oliverlloyd:
Hmm, I think this highlights a shortfall in the approach of just putting a comment at the top of the file as a way to prevent edits. I'm guessing if you're searching multiple files then you never get to see that comment. But anyway, you can't - or really shouldn't be - editing fixtures anymore. There's a script to refresh them from live (`make gen-fixtures`)

oliverlloyd:
By the way, if you need to add a property that currently doesn't exist (eg. it isn't brought in when you run `make gen-fixtures` then you can import the fixture where it's being used and and then spread it to add the new value

src/web/layouts/StandardLayout.tsx
```diff
@@ -507,9 +507,9 @@ export const StandardLayout = ({ CAPI, NAV, format, palette }: Props) => {
 								webTitle={CAPI.webTitle}
 								author={CAPI.author}
 								tags={CAPI.tags}
-								primaryDateline={CAPI.blocks[0].primaryDateLine}
+								primaryDateline={CAPI.webPublicationDateDisplay}
 								secondaryDateline={
-									CAPI.blocks[0].secondaryDateLine
+									CAPI.webPublicationSecondaryDateDisplay
```
jfsoul, line 512: I suppose we need to make this change for all the other layouts as well? E.g. Immersive

src/model/window-guardian.ts
```diff
@@ -74,18 +74,21 @@ export const makeGuardianBrowserCAPI = (CAPI: CAPIType): CAPIBrowserType => {
...
+							...element,
+							[indexName]: index,
+						} as T);
+					}
+					return acc;
+				},
+				[] as T[],
+			);
+		}
+		return [];
```
jfsoul, line 91: 👍 Minor style thing - I tend to prefer early return for the edge/error case rather than for the happy case (and I think this is fairly common). So this would instead be:

```
if(blocks.length === 0) return [];

return blocks[0].elements...
```

src/model/json-schema.json
```diff
@@ -174,6 +174,9 @@
         "webPublicationDateDisplay": {
             "type": "string"
         },
+        "webPublicationSecondaryDateDisplay": {
```
jfsoul, line 177: Is this a new field we're adding to the model in frontend?

nitro-marky: Yes, I've already added it

src/web/layouts/StandardLayout.tsx
nitro-marky: Done

src/lib/content.d.ts
```diff
@@ -78,6 +78,12 @@ interface QuizAtomBlockElement {
 	id: string;
 	questions: QuestionType[];
 	resultBuckets: ResultBucketsType[];
+	resultGroups: {
+		id: string;
+		title: string;
+		shareText: string;
+		minScore: number;
+	}[];
```
oliverlloyd, line 86: 👍 Nice

oliverlloyd (approved):
Great work, very thorough!

nitro-marky:
[Support empty article body](https://trello.com/c/fwVQA2Np/2240-support-empty-article-body)

fixtures/articles/Analysis.ts
```diff
@@ -12,9 +12,7 @@
  */
```
gtrufitt, line 12: Should these changes to fixtures be part of your PR?

shtukas (approved):

PR #2573, opened 2021-02-18
Remove restrict-plus-operands exception

liywjl:
## What does this change?
- Remove `@typescript-eslint/restrict-plus-operands` from TS lint exception rules
- Fix related linting errors

## Why?
It was turned off after an update to TS and TS linting packages, we are progressively removing these linting rules exceptions


shtukas (approved):

oliverlloyd (approved):
👍

PR #2581, opened 2021-02-18
Use a defined ID for AMP ad wrapper divs

nicl:
### Before

`ad-undefined` as ID for all AMP ad wrappers

### After

`ad-1` (or 2, 3, etc)

## Why?

It doesn't seem to matter either way from a reader perspective, but it allows us to remove some mysterious and (type) unsafe code.

https://trello.com/c/G4Bi9N6h/2459-fix-scary-jsxelement-extraction-of-id


shtukas (approved):

oliverlloyd (approved):
Code mysteries are bad

PR #2587, opened 2021-02-19
ElementRenderer

oliverlloyd:
## What does this change?
Introdues the `ElementRenderer` component which is an abstraction of our switch statement we use to render elements.

### Before
We had two switch statements. One for the article body, and one for main media

### After
There is just one single switch which we use for both main media and article body elements

## Why?
Because like this we now support _all_ elements as main media. You can't necessarily insert,  say, an atom as main media because Composer won't allow this but there are a lot of strange things that can be put there as embeds and there's no reason that we shouldn't do this so now DCR offers broad support instead of having hardcoded limits.

Another side benefit to this approach is that we can also use this abstraction for live blogs, once we start working on those (maybe soon!)

### What about ClickToView?
This has been disabled for main media. @frj I'll raise this in the channel to confirm this is the approach we want

### So now we can set `starRating` for all images?
We don't do this. We only overlay stars for `Design.Review` and only on main media. But now we _can_

### And we can `hideCaption` for in body images?
Again, not something we do but it happened for `Design.Immersive` main media and now it's part of the api if we need it

### What about Figure?
Previously `Figure` was a way to centralise certain element features:

- Adding an `id` for hydration
- Wrapping children in a `figure` element 
- Applying position and margin styles based on role

Now that we're also using this component for main media elements we had to add the `isMainMedia` prop and conditionally apply some of these features. For main media we only add the `id` and wrap in a `figure` tag, we don't add any styling.

At a later stage, we could come back to this. Main media does get position styling based on role (`showcase` and `immersive` are both valid) but at the moment this logic happens inside the separate coponents using `isMainMedia` and `role` depending on context. We ccould potentially refactor this too happen in `Figure` so that we apply these styles universally, regardless of element type.

### Why have we bumped the version of Prismjs here?
Because there was an issue in Storybook where we were getting an error on our page layout stories complaining about how we're importing things. I spent a lot of time on this pairing with @buck06191 and @nicl and we think it was a similar issue to the [one reported here](https://github.com/storybookjs/storybook/issues/9516). Using bisection, we were able to narrow down the error further to relate to the `CodeBlockElement`, specifcally the import of the (actually quite nice) [Prism](https://www.npmjs.com/package/prismjs) lib. By bumping from [v1.22.0](https://github.com/PrismJS/prism/releases/tag/v1.22.0) to [v1.23.0](https://github.com/PrismJS/prism/releases/tag/v1.23.0) of that lib the problem went away 🤷. I don't have the root cause for why this should be and can't see anything obvious in the release notes, but sometimes you just need to move on.


package.json
```diff
@@ -112,7 +112,7 @@
         "preact-render-to-string": "^5.1.12",
         "prettier-eslint": "^11.0.0",
         "pretty-bytes": "^5.5.0",
-        "prismjs": "^1.22.0",
+        "prismjs": "^1.23.0",
```
oliverlloyd, line 115: See PR description for why this is happening

src/web/components/ClickToView.tsx
```diff
@@ -117,11 +125,12 @@ export const ClickToView = ({
...
 	if (
-		shouldDisplayOverlay(
+		shouldDisplayOverlay({
 			isTracking,
 			isOverlayClicked,
-			isInABTestVariant(abTests),
-		)
+			isInABTestVariant: isInABTestVariant(abTests),
+			isMainMedia,
+		})
```
oliverlloyd, line 133: I converted this function to accept an object to prevent errors/confusion with prop order now that it is accepting more and more props

src/web/components/Figure.tsx
```diff
@@ -126,7 +127,20 @@ const decidePosition = (role: RoleType) => {
...
+	children,
+	id,
+	isMainMedia,
+}: Props) => {
+	if (isMainMedia) {
+		// Don't add in-body styles for main media elements
+		// TODO: If we want to support other element types having role position, such
+		// as showcase twitter embeds, then we should remove the role positioning which
+		// currently lives in ImageComponent and hoist it up to here, the same as we're
+		// doing using decidePosition for in-body elements
```
oliverlloyd, line 141: See PR comment for more information

src/web/components/MainMedia.tsx
```diff
@@ -55,93 +50,6 @@ const immersiveWrapper = css`
     overflow: hidden;
 `;
 
-function renderElement(
```
oliverlloyd, line 58: This code is moved in `Element Renderer` with some additional props to support all use cases

oliverlloyd:
[Support main media embeds](https://trello.com/c/KQWwNgIP/1759-support-main-media-embeds)

liywjl (approved):
Great work! 👍

cypress/integration/e2e/atom.video.spec.js
```diff
@@ -21,7 +21,7 @@ describe('Video', function () {
 		});
 		it('should render', function () {
 			cy.visit(`/Article?url=${mainMediaVideo}`);
-			cy.get(`[data-cy="main-media-youtube-atom"]`).should('be.visible');
+			cy.get(`[id^="youtube-block"]`).should('be.visible');
```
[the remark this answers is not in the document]
oliverlloyd: `id^=` means starts with. We need this while we're still using indexes for ids

gtrufitt:
tenor-202007429

Looks like a lot of code moved around, and I assume nothing has changed within that. Do we have the test coverage on chromatic for main media block elements, or will we add that when accepting them in the article picker?

gtrufitt (approved):

oliverlloyd:
> tenor-202007429
> 
> Looks like a lot of code moved around, and I assume nothing has changed within that. Do we have the test coverage on chromatic for main media block elements, or will we add that when accepting them in the article picker?

The only code changes were to add some props as detailed. If you follow the commits you can see the original move with no changes.

Re stories, as you say, we can add these in as we enable. I don't actually think we changing much here though

PR #2590, opened 2021-02-19
Braze Epic

rhystmills:
## What does this change?

Adds the BrazeEpic to DCR.

### Before

There was only the ReaderRevenueEpic.

### After

The BrazeEpic is also available for the end of article slot. Copy can be passed in from Braze via key-value pairs on an InAppMessage campaign.

The `BrazeMessages` instance, which we wrap the Braze SDK with is now shared with `SlotBodyEnd` and `StickyBottomBanner` via state held in `App.tsx`. We found that this was a better fit, and gave us more control over the session start call etc than our initial approach of memoizing the Braze SDK instance.

## Why?

To allow Marketing to target custom Epics to supporters.

## Other information

As is this isn't _quite_ ready to be used in production. We're currently missing some tracking events for the epic. We'll follow up in another PR to add these. Epic's won't be configured in Braze until this work is done.

The actual epic component is dynamically imported from the `support-dotcom-components` service and we translate the data we get from Braze to the format that component expects for its props. For the paragraphs, which can vary in number, we've limited to 9 for simplicity (the component itself has no limit in theory). This should cover most scenarios and can always be tweaked in future.

`cta` is derived from two key-value pairs from Braze - 'buttonText' and 'buttonUrl'.

The current key-value pairs are:

```typescript
type DataFromBraze = {
  name?: string;
  heading?: string; 
  paragraph1?: string;
  paragraph2?: string;
  paragraph3?: string;
  paragraph4?: string;
  paragraph5?: string;
  paragraph6?: string;
  paragraph7?: string;
  paragraph8?: string;
  paragraph9?: string;
  highlightedText?: string;
  buttonText?: string;
  buttonUrl?: string;
};
```

I chose an essentially arbitrary timeout of 10,000ms. It seems useful to have a limit (to prevent the epic loading a long time after the rest of the page does) - but I don't think we need a short timeout like the brazeBanner does as users will probably not reach the endOfArticle slot as quickly.

### DIFF
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -71,12 +71,14 @@ import {
 	StickyNavSimple,
 	StickyNavBackscroll,
 } from '@root/src/web/components/Nav/StickNavTest/StickyNav';
+import { BrazeMessagesInterface } from '@root/src/web/lib/braze/BrazeMessages';
 import {
 	submitComponentEvent,
 	OphanComponentEvent,
 } from '../browser/ophan/ophan';
 import { trackPerformance } from '../browser/ga/ga';
 import { decidePalette } from '../lib/decidePalette';
+import { buildBrazeMessages } from '../lib/braze/buildBrazeMessages';
 
 // *******************************
 // ****** Dynamic imports ********
@@ -164,6 +166,10 @@ export const App = ({ CAPI, NAV }: Props) => {
 	// executing their canShow logic until countryCode is available):
 	const [asyncCountryCode, setAsyncCountryCode] = useState<
 		Promise<string | void>
+	>();
+
+	const [brazeMessages, setBrazeMessages] = useState<
+		Promise<BrazeMessagesInterface>
 	>();
 
 	const pageViewId = window.guardian?.config?.ophan?.pageViewId;
@@ -329,6 +335,12 @@ export const App = ({ CAPI, NAV }: Props) => {
 			}
 		});
 	}, []);
+
+	useOnce(() => {
+		setBrazeMessages(
+			buildBrazeMessages(isSignedIn as boolean, CAPI.config.idApiUrl),
+		);
+	}, [isSignedIn, CAPI.config.idApiUrl]);
 
 	const display: Display = decideDisplay(CAPI);
 	const design: Design = decideDesign(
@@ -958,6 +970,7 @@ export const App = ({ CAPI, NAV }: Props) => {
 					isPaidContent={CAPI.pageType.isPaidContent}
 					tags={CAPI.tags}
 					contributionsServiceUrl={CAPI.contributionsServiceUrl}
+					brazeMessages={brazeMessages}
 				/>
 			</Portal>
 			<Portal
@@ -1049,7 +1062,7 @@ export const App = ({ CAPI, NAV }: Props) => {
 					isSignedIn={isSignedIn}
 					asyncCountryCode={asyncCountryCode}
 					CAPI={CAPI}
-					idApiUrl={CAPI.config.idApiUrl}
+					brazeMessages={brazeMessages}
 				/>
 			</Portal>
 		</React.StrictMode>
--- a/src/web/components/SlotBodyEnd/BrazeEpic.tsx
+++ b/src/web/components/SlotBodyEnd/BrazeEpic.tsx
@@ -1,0 +1,174 @@
+import React, { useState } from 'react';
+import * as emotion from 'emotion';
+import * as emotionCore from '@emotion/core';
+import * as emotionTheming from 'emotion-theming';
+import {
+	parseBrazeEpicParams,
+	EpicDataFromBraze,
+	Variant,
+} from '@root/src/web/lib/braze/parseBrazeEpicParams';
+import { BrazeMessagesInterface } from '@root/src/web/lib/braze/BrazeMessages';
+import { CanShowResult } from '@root/src/web/lib/messagePicker';
+import { useOnce } from '@root/src/web/lib/useOnce';
+import { joinUrl } from '@root/src/lib/joinUrl';
+
+const { css } = emotion;
+
+const wrapperMargins = css`
+	margin: 18px 0;
+`;
+
+const EPIC_COMPONENT_PATH = '/epic.js';
+
+type NoEpicDataFromBraze = {
+	componentName: 'NoEpic';
+};
+
+type DataFromBraze = EpicDataFromBraze | NoEpicDataFromBraze;
+
+type Meta = {
+	dataFromBraze?: DataFromBraze;
+	logImpressionWithBraze: () => void;
+	// logButtonClickWithBraze: (id: number) => void; // TODO: Handle button click tracking
+};
+
+type EpicConfig = {
+	meta: Meta;
+	contributionsServiceUrl: string;
+	countryCode: string;
+};
+
+type Tracking = {
+	ophanPageId: string;
+	platformId: string;
+	clientName: string;
+	referrerUrl: string;
+};
+
+type EpicProps = {
+	variant: Variant;
+	tracking: Tracking;
+	numArticles: number;
+	countryCode: string;
+};
+
+export const canShow = async (
+	brazeMessagesPromise: Promise<BrazeMessagesInterface>,
+): Promise<CanShowResult> => {
+	try {
+		const brazeMessages = await brazeMessagesPromise;
+		const message = await brazeMessages.getMessageForEndOfArticle();
+
+		return {
+			result: true,
+			meta: {
+				dataFromBraze: message.extras,
+				logImpressionWithBraze: () => {
+					message.logImpression();
+				},
+			},
+		};
+	} catch (e) {
+		return { result: false };
+	}
+};
+
+const buildEpicProps = (
+	dataFromBraze: EpicDataFromBraze,
+	countryCode: string,
+): EpicProps | null => {
+	const variant = parseBrazeEpicParams(dataFromBraze);
+	if (variant === null) return null;
+
+	const pageTracking = {
+		ophanPageId: window.guardian.config.ophan.pageViewId,
+		platformId: 'GUARDIAN_WEB',
+		clientName: 'dcr',
+		referrerUrl: window.location.origin + window.location.pathname,
+	};
+
+	const tracking = {
+		...pageTracking,
+	};
+
+	return { variant, tracking, numArticles: 0, countryCode };
+};
+
+const BrazeEpic = ({
+	contributionsServiceUrl,
+	meta,
+	countryCode,
+}: EpicConfig) => {
+	const [Epic, setEpic] = useState<React.FC<EpicProps>>();
+
+	useOnce(() => {
+		window.guardian.automat = {
+			react: React,
+			preact: React,
+			emotionCore,
+			emotionTheming,
+			emotion,
+		};
+
+		const componentUrl = joinUrl([
+			contributionsServiceUrl,
+			EPIC_COMPONENT_PATH,
+		]);
+
+		window
+			.guardianPolyfilledImport(componentUrl)
+			.then((epicModule: { ContributionsEpic: React.FC<EpicProps> }) => {
+				setEpic(() => epicModule.ContributionsEpic); // useState requires functions to be wrapped
+				// TODO: log the impression when the epic is in view (using hasBeenSeen probably)
+				meta.logImpressionWithBraze();
+			})
+			.catch((error) => {
+				window.guardian.modules.sentry.reportError(error, 'braze-epic');
+			});
+	}, [contributionsServiceUrl, meta]);
+
+	if (Epic && meta.dataFromBraze) {
+		// This will come from Braze via the meta from canShow
+		const props = buildEpicProps(
+			meta.dataFromBraze as EpicDataFromBraze,
+			countryCode,
+		);
+
+		if (props) {
+			return (
+				<div className={wrapperMargins}>
+					{/* eslint-disable-next-line react/jsx-props-no-spreading */}
+					<Epic {...props} />
+				</div>
+			);
+		}
+	}
+
+	return null;
+};
+
+const NoEpic = () => null;
+
+export const MaybeBrazeEpic = ({
+	contributionsServiceUrl,
+	meta,
+	countryCode,
+}: EpicConfig) => {
+	const componentName = meta.dataFromBraze?.componentName;
+
+	if (componentName === 'Epic') {
+		return (
+			<BrazeEpic
+				contributionsServiceUrl={contributionsServiceUrl}
+				meta={meta}
+				countryCode={countryCode}
+			/>
+		);
+	}
+
+	if (componentName === 'NoEpic') {
+		return <NoEpic />;
+	}
+
+	return null;
+};
--- a/src/web/components/SlotBodyEnd/SlotBodyEnd.tsx
+++ b/src/web/components/SlotBodyEnd/SlotBodyEnd.tsx
@@ -8,10 +8,12 @@ import {
 	CandidateConfig,
 } from '@root/src/web/lib/messagePicker';
 
+import { BrazeMessagesInterface } from 'src/web/lib/braze/BrazeMessages';
 import {
 	ReaderRevenueEpic,
 	canShow as canShowReaderRevenueEpic,
 } from './ReaderRevenueEpic';
+import { MaybeBrazeEpic, canShow as canShowBrazeEpic } from './BrazeEpic';
 
 type Props = {
 	isSignedIn?: boolean;
@@ -23,6 +25,7 @@ type Props = {
 	isPaidContent: boolean;
 	tags: TagType[];
 	contributionsServiceUrl: string;
+	brazeMessages?: Promise<BrazeMessagesInterface>;
 };
 
 const buildReaderRevenueEpicConfig = ({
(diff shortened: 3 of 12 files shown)


src/web/lib/braze/initialiseAppboy.ts
```diff
@@ -1,19 +1,33 @@
+import type appboy from '@braze/web-sdk-core';
```
tjmw, line 1: Note to ourselves to verify that this gives the behaviour we want!

tjmw: Discussed offline. Looking at the bundle size changes and visualisation I think this is fine.

src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -0,0 +1,212 @@
...
+			`Not attempting to show Braze messages. Dependency ${dependenciesResult.failure.field} failed with ${dependenciesResult.failure.data}.`,
+		);
+		return { result: false };
+	}
+
+	const appboy = await getInitialisedAppboy(
+		dependenciesResult.data.apiKey as string,
+	);
+
+	const brazeMessages = new BrazeMessages(appboy);
```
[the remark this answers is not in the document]
tjmw: We've ended up doing this work in this PR as we realised there were some events happening twice which were hard to control (setting the Braze user, opening the session). This work has now been pulled up out of the components and is kicked off in `App.tsx`.

shtukas (approved):

tjmw:
I just noticed that the bundle size report shows quite a big jump for `react.js`. I think that's slightly misleading. It looks like this is a comparison with where this branch started rather than current main and so doesn't account for the fact that other changes have increased the bundle size in the meantime. Before merging I'll make sure to rebased on `main` to check that this is the case.

src/web/components/App.tsx
```diff
@@ -326,6 +332,12 @@ export const App = ({ CAPI, NAV }: Props) => {
 		});
 	}, []);
 
+	useOnce(() => {
+		setBrazeMessages(
+			buildBrazeMessages(isSignedIn as boolean, CAPI.config.idApiUrl),
+		);
+	}, [isSignedIn, CAPI.config.idApiUrl]);
```
oliverlloyd, line 339: I think you're already aware but just to be sure this array are the things that must be defined before the body executes. So having `CAPI.config.idApiUrl` is a guard against that not being defined (which I presume is unlikely but good to have) and then `isSignedIn` is what is really being used here

src/web/components/App.tsx
```diff
@@ -869,6 +881,7 @@ export const App = ({ CAPI, NAV }: Props) => {
 					isPaidContent={CAPI.pageType.isPaidContent}
 					tags={CAPI.tags}
 					contributionsServiceUrl={CAPI.contributionsServiceUrl}
+					brazeMessages={brazeMessages}
```
oliverlloyd, line 884: We don't have the concept of, er, `PortalOnce` so these will be rerendered. I think we came to the conclusion that this is fine and that it should work like a normal react rerender.

The down side is everything in the file gets rerendered so we might want to think about memoisation at some point (which is what our ThingOnce pattern is essentially doing). Of course, sometimes you _want_ the rerender which is the case here.

src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -0,0 +1,169 @@
...
+		const componentUrl = `${contributionsServiceUrl}${EPIC_COMPONENT_PATH}`;
+		window
+			.guardianPolyfilledImport(componentUrl)
+			.then((epicModule: { ContributionsEpic: React.FC<EpicProps> }) => {
+				setEpic(() => epicModule.ContributionsEpic); // useState requires functions to be wrapped
+				// TODO: log the impression when the epic is in view (using hasBeenSeen probably)
+				meta.logImpressionWithBraze();
+			})
+			// eslint-disable-next-line no-console
+			.catch((error) => console.log(`epic - error is: ${error}`));
```
oliverlloyd, line 119: If this is a real error we probably want to send it to Sentry using `reportError`

eg.

```
window.guardian.modules.sentry.reportError(error, 'match-nav');
```

src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -0,0 +1,169 @@
...
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
oliverlloyd, line 120: Usually, this lint rule is right - what was the reason for disabling it

src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+const NoEpic = () => null;
+
+export const MaybeBrazeEpic = ({
+	contributionsServiceUrl,
+	meta,
+	countryCode,
+}: EpicConfig) => {
+	const componentName = meta.dataFromBraze?.componentName;
+	console.log('rendering the epic!', meta);
```
oliverlloyd, line 152: Test code can be removed

src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -0,0 +1,169 @@
...
+};
+
+const NoEpic = () => null;
+
+export const MaybeBrazeEpic = ({
+	contributionsServiceUrl,
+	meta,
+	countryCode,
+}: EpicConfig) => {
+	const componentName = meta.dataFromBraze?.componentName;
```
oliverlloyd, line 151: I suspect this code might be from another PR so this is more a nitpick but I find the name `componentName` confusing because it's not actually the name of a component 🤷

src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+		if (props) {
+			return (
+				<div className={wrapperMargins}>
+					{/* eslint-disable-next-line react/jsx-props-no-spreading */}
+					<Epic {...props} />
+				</div>
+			);
+		}
+		console.log('End of article props were not defined.');
```
oliverlloyd, line 138: Testing code?

src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+	useEffect(() => {
+		window.guardian.automat = {
+			react: React,
+			preact: React,
+			emotionCore,
+			emotionTheming,
+			emotion,
+		};
+		const componentUrl = `${contributionsServiceUrl}${EPIC_COMPONENT_PATH}`;
```
oliverlloyd, line 110: We have a utility function called `joinUrl` which might be useful here? It fixes the double slash problem that can arise in situations like this

src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -0,0 +1,169 @@
+import React, { useEffect, useState } from 'react';
+import * as emotion from 'emotion';
+import * as emotionCore from '@emotion/core';
+import * as emotionTheming from 'emotion-theming';
```
oliverlloyd, line 4: Is this pattern of importing emotion needed? We're about to migrate to Emotion 11 and I'm not sure how this fits into that

src/web/components/SlotBodyEnd/SlotBodyEnd.tsx
```diff
@@ -60,6 +62,28 @@ const buildReaderRevenueEpicConfig = ({
...
+			canShow: () => canShowBrazeEpic(isSignedIn, idApiUrl),
+			/* eslint-disable-next-line react/jsx-props-no-spreading */
+			show: (meta: any) => () => (
+				<BrazeEpic
+					meta={meta}
+					contributionsServiceUrl={contributionsServiceUrl}
+				/>
+			),
+		},
+		timeoutMillis: 10000,
```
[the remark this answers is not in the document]
oliverlloyd: What's the timeout for? Is there an api call being made here or?

src/web/lib/braze/BrazeMessages.ts
```diff
@@ -1,48 +1,88 @@
...
-	extras?: Extras;
-	triggerId?: string; // This keeps Typescript happy, because Message could be an InAppMessage or a ControlMessage
+class BrazeMessage {
+	appboy: typeof appboy;
+
+	message: appboy.InAppMessage;
+
+	constructor(message: appboy.InAppMessage, ab: typeof appboy) {
+		this.message = message;
+		this.appboy = ab;
```
oliverlloyd, line 19: I don't think there's an actual crossover but it might safer to not use the 'ab' name just because it denotes ab testing in other parts of the codebase. Perhaps later if an ab test were to be added here it could be tricky.

src/web/lib/braze/buildBrazeMessages.ts
```diff
@@ -0,0 +1,79 @@
...
+		idApiUrl,
+	);
+
+	if (!dependenciesResult.isSuccessful) {
+		const { failure, data } = dependenciesResult;
+
+		if (SDK_OPTIONS.enableLogging) {
+			console.log(
+				`Not attempting to show Braze messages. Dependency ${failure.field} failed with ${failure.data}.`,
+			);
```
oliverlloyd, line 44: Is this console test code?

oliverlloyd (approved):
I've left the review of the braze related functionality to others that know more about this than myself and instead just made lots of small nitpicky comments, sorry! 😉 

For the shared state in App though (the reason I landed here) I think this looks sound. We know from looking at [the Preact code](https://github.com/preactjs/preact/blob/df748d106fb78fbd46d14563b4712f921ccf0300/compat/src/portals.js) that Portals can rerender safely so the only reason we would want to limit this is normal perf optimisations. Which is a valid area to think about but out of scope for this PR.

src/web/components/App.tsx
tjmw: Good point. I _think_ this is fine, in both cases we use `useOnce` around the heavy lifting of figuring out if there's a message to show.

src/web/components/SlotBodyEnd/BrazeEpic.tsx
tjmw: This relates to how these core deps are shared with dynamically imported components via `window.guardian.automat` (also the case for reader revenue components). I think it's on the roadmap for the Emotion upgrade (I've chatted to Gareth and William about this in recent weeks).

src/web/components/SlotBodyEnd/BrazeEpic.tsx
tjmw: Good point, that's probably what we should be doing here 👍

src/web/components/SlotBodyEnd/BrazeEpic.tsx
tjmw: Ah good spot, yeah I think we can fill out the dependencies correctly and remove this commit.

src/web/components/SlotBodyEnd/BrazeEpic.tsx
tjmw: Good spot. Huh I wonder why the linting rule isn't picking this up 🤔

src/web/components/SlotBodyEnd/BrazeEpic.tsx
tjmw: Yeah you're right, this is a name we already use for the banner where it actually _is_ a component name (from `@guardian/braze-components`. That mapping isn't quite the same here because we're piggybacking on the reader revenue epic component from support-dotcom-components.

src/web/components/SlotBodyEnd/SlotBodyEnd.tsx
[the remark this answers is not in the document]
tjmw: Yeah behind the scenes we're communicating with Braze to know if there's a message to be shown in the epic slot. The reader revenue epic doesn't have a timeout. I wonder if we should remove this for consistency.

src/web/lib/braze/buildBrazeMessages.ts
tjmw: It's actually useful for local development. Since it'll only be output if that SDK_OPTIONS flag is set I think we'll leave it in (functionally this hasn't changed, just this line has moved).

src/web/components/SlotBodyEnd/BrazeEpic.tsx
tjmw: Ah yes I'll make that change 👍

src/web/components/SlotBodyEnd/BrazeEpic.tsx
oliverlloyd: You've beaten linting

src/web/components/SlotBodyEnd/BrazeEpic.tsx
oliverlloyd: Not a huge issue and not worth refactoring for unless it's easy

src/web/components/SlotBodyEnd/BrazeEpic.tsx
oliverlloyd: Sounds good

src/web/components/SlotBodyEnd/BrazeEpic.tsx
tjmw: Changing this reminded me that the reason for `[]` (and therefore the linting rule exception) was because we only want/need this to happen once. I've replaced with the true dependency list and `useOnce` which I think makes things clearer.

oliverlloyd: It's great to see this `useOnce` api working out!

src/web/components/SlotBodyEnd/SlotBodyEnd.tsx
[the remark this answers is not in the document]
tjmw: I chatted this over with the team and we decided to remove the timeout for now.

src/web/components/App.tsx
```diff
@@ -330,6 +336,12 @@ export const App = ({ CAPI, NAV }: Props) => {
 		});
 	}, []);
 
+	useOnce(() => {
+		setBrazeMessages(
+			buildBrazeMessages(isSignedIn as boolean, CAPI.config.idApiUrl),
```
liywjl, line 341: if you set the default `isSignedIn` to `false` in the useState you shouldnt need to case it here

src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -0,0 +1,174 @@
...
+	const pageTracking = {
+		ophanPageId: window.guardian.config.ophan.pageViewId,
+		platformId: 'GUARDIAN_WEB',
+		clientName: 'dcr',
+		referrerUrl: window.location.origin + window.location.pathname,
+	};
+
+	const tracking = {
+		...pageTracking,
+	};
```
liywjl, line 92: why not just reference `pageTracking`?

src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+	return { variant, tracking, numArticles: 0, countryCode };
```
liywjl, line 94: you could just set `tracking: pageTracking` as not to add `tracking`

src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -0,0 +1,174 @@
...
+import { useOnce } from '@root/src/web/lib/useOnce';
+import { joinUrl } from '@root/src/lib/joinUrl';
+
+const { css } = emotion;
+
+const wrapperMargins = css`
+	margin: 18px 0;
+`;
+
+const EPIC_COMPONENT_PATH = '/epic.js';
```
liywjl, line 21: if it is only used once, just set the string. Add a comment to add more detail if needed

src/web/components/App.tsx
tjmw: Ah right yeah that's true. This is needed because it starts as `undefined`. I'm not sure about changing the default from `undefined` to `false` as other things may be relying on the current behaviour. Perhaps worth revisiting in another PR?

src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -0,0 +1,174 @@
...
+} from '@root/src/web/lib/braze/parseBrazeEpicParams';
+import { BrazeMessagesInterface } from '@root/src/web/lib/braze/BrazeMessages';
+import { CanShowResult } from '@root/src/web/lib/messagePicker';
+import { useOnce } from '@root/src/web/lib/useOnce';
+import { joinUrl } from '@root/src/lib/joinUrl';
+
+const { css } = emotion;
+
+const wrapperMargins = css`
+	margin: 18px 0;
```
liywjl, line 18: could look at using Soure's spaceing params? Also do you need the `0`?

liywjl (approved):
Sorry for the late review, just had some minor things to cleanup but overall look good

src/web/components/App.tsx
oliverlloyd: I think there's a specific use case for it being undefined

`undefined` = we're checking 🤷 
`false` = definately not signed in
`true` = signed in

It's probably worth adding a comment about this

tjmw:
Thanks for the feedback @liywjl 👍  We're not done with changes in this area so will look at incorporating your suggestions in a future PR.

PR #2591, opened 2021-02-19
Revert restrict template expressions lint exception

liywjl:
## What does this change?
- Revert `@typescript-eslint/restrict-template-expressions`
- fix linting errors

A lot of this PR's diffs are disabling linting issues related to `console.log` for errors. Many times it is not obvious how to log errors (do we log the `error` itself? or `error.stack`? or `error.message`?)

Other changes are mainly focused on handling cases of `undefined` or `null` that were not handled before in template strings


src/web/components/Links.tsx
```diff
@@ -134,60 +134,66 @@ export const Links = ({
 	mmaUrl: mmaUrlFromConfig,
 }: Props) => {
 	const [showGiftingLink, setShowGiftingLink] = useState<boolean>();
-	const [userIsDefined, setUserIsDefined] = useState<boolean>();
+	// const [userIsDefined, setUserIsDefined] = useState<boolean>();
```
oliverlloyd, line 137: Nix

src/amp/components/Analytics.test.tsx
```diff
@@ -2,33 +2,34 @@ import React from 'react';
...
 		},
-		ipsosSectionName: 'section',
-	};
-
-	beforeEach(() => {
-		delete analytics.section;
-	});
+	},
+	ipsosSectionName: 'section',
+	section: '',
```
liywjl, line 23: added `section` key to object

src/amp/components/Analytics.test.tsx
```diff
@@ -2,33 +2,34 @@ import React from 'react';
...
+		apiKey: '42-2020',
+		payload: {
+			'properties.content.title': 'article title',
 		},
-		ipsosSectionName: 'section',
-	};
-
-	beforeEach(() => {
-		delete analytics.section;
-	});
```
liywjl, line 28: removed this pattern in favor of spreading `analyticsBase` and changing `section` when needed

src/amp/components/Analytics.tsx
```diff
@@ -6,7 +6,7 @@ export interface AnalyticsModel {
 	title: string;
 	fbPixelaccount: string;
 	comscoreID: string;
-	section?: string;
+	section: string;
```
liywjl, line 9: `section` is used in `Analytics` (src/amp/components/Analytics.tsx) and `Ad` (/src/amp/components/Ad.tsx). 

It would make sense for it to be always defined, but currently it is sometimes defined as `''`.

src/amp/server/render.tsx
```diff
@@ -24,7 +24,7 @@ export const render = ({ body }: express.Request, res: express.Response) => {
 
 		const scripts = [...extractScripts(elements, CAPI.mainMediaElements)];
 
-		const sectionName = CAPI.sectionName || '';
+		const sectionName = CAPI.sectionName || ''; // TODO: sectionName should be required but is missing from CAPI on some article types (ex: Interview fixture)
```
liywjl, line 27: Should investigate why `sectionName` is sometimes undefined

src/web/lib/getCommentContext.ts
```diff
@@ -101,9 +91,15 @@ export const getCommentContext = async (
 ): Promise<CommentContextType> => {
 	const url = joinUrl([ajaxUrl, 'comment', commentId.toString(), 'context']);
 	const filters = initFiltersFromLocalStorage();
-	const params = buildParams(filters);
+	const paramsObj = buildParams(filters);
+
+	const params = Object.keys(paramsObj)
+		// @ts-ignore
+		// eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+		.map((key) => `${key}=${paramsObj[key]}`)
```
liywjl, line 99: `key` will always be a key inferred from `paramsObj`, TS is not able to pick up this logic however so we have to ignore

src/web/lib/getZIndex.tsx
```diff
@@ -62,8 +62,12 @@ const decideIndex = (name: string): number | null => {
 	return decided || null;
 };
 
-export const getZIndex = (zIndex: ZIndex): string =>
-	`z-index: ${decideIndex(zIndex)};`;
+export const getZIndex = (zIndex: ZIndex): string => {
+	const selectedIndex = decideIndex(zIndex);
```
liywjl, line 66: as `decideIndex` can return `null` we need to handle that case

liywjl (commented):
A lot of this PR's diffs are disabling linting issues related to `console.log` for errors. Many times it is not obvious how to log errors (do we log the `error` itself? or `error.stack`? or `error.message`?)

Other changes are mainly focused on handling cases of `undefined` or `null` that were not handled before in template strings

shtukas (approved):

oliverlloyd:
There are a lot of refactors here that could perhaps aren't really linting changes and should be done as separate PRs?

liywjl:
I have split out all AMP related changes to this PR https://github.com/guardian/dotcom-rendering/pull/2603

Will revert AMP changes here

src/amp/components/Footer.tsx
```diff
@@ -134,7 +134,7 @@ const FooterLinks: React.FC<{
 		const ls = linkGroup
 			.filter((l) => isOnPlatform(l, LinkPlatform.Amp))
 			.map((l, index) => (
-				<li key={`${l.url}${index}`}>
+				<li key={`${l.title}${index}`}>
```
liywjl, line 137: `url` can be undefined so instead we use `title`

src/lib/assets.ts
```diff
@@ -35,10 +35,16 @@ export const getScriptArrayFromFilename = (
...
-		{ src: `${CDN}assets/${legacyFilename}`, legacy: true },
-	];
+
+	const scriptArray = [{ src: `${CDN}assets/${filename}`, legacy: false }];
+	// legacyFilename can be undefined, we therefor should only add it if defined
+	if (legacyFilename)
+		scriptArray.push({
+			src: `${CDN}assets/${legacyFilename}`,
+			legacy: true,
+		});
```
liywjl, line 45: as `legacyFilename` can be undefined, we need to add a condition before returning it

src/amp/components/Analytics.tsx
```diff
@@ -58,7 +58,7 @@ export const Analytics: React.FC<{
                      "vars": {
                        "title": "${title}",
                        "platform": "AMP",
-                       "sectionId": "${section}",
+                       "sectionId": "${section || 'undefined'}",
```
oliverlloyd, line 61: Do we want 'undefined' as a string? I would expect to either see `undefined` the value or '' as an empty string here

src/app/server.ts
```diff
@@ -138,15 +138,21 @@ if (process.env.NODE_ENV === 'production') {
             `);
 		} catch (e) {
 			const error = e as Error;
-			res.status(500).send(`<pre>${error.stack}</pre>`);
+			res.status(500).send(
+				`<pre>${
+					error.stack || `Undefined error: ${error.message}`
```
oliverlloyd, line 143: Will the text 'Undefined error' appear anywhere? I wonder if it's safer to just show the error message 'Error: Some error', that would result in 'Undefined error: Error: Some error'

src/lib/assets.ts
```diff
@@ -35,10 +35,16 @@ export const getScriptArrayFromFilename = (
 	const chunks: string[] | undefined =
 		loadableManifestLegacy.assetsByChunkName[chunkName];
 	const legacyFilename = chunks && chunks.length > 0 && chunks[0];
-	return [
-		{ src: `${CDN}assets/${filename}`, legacy: false },
-		{ src: `${CDN}assets/${legacyFilename}`, legacy: true },
-	];
+
+	const scriptArray = [{ src: `${CDN}assets/${filename}`, legacy: false }];
+	// legacyFilename can be undefined, we therefor should only add it if defined
```
oliverlloyd, line 40:
```suggestion
	// legacyFilename could be undefined
```

src/lib/content.d.ts
```diff
@@ -224,6 +224,7 @@ interface MapBlockElement extends ThirdPartyEmbeddedContent {
 	embedUrl: string;
 	originalUrl: string;
 	title: string;
+	source: string;
```
oliverlloyd, line 227: Is this change related?

src/web/components/Links.tsx
```diff
@@ -134,60 +134,65 @@ export const Links = ({
 	mmaUrl: mmaUrlFromConfig,
 }: Props) => {
 	const [showGiftingLink, setShowGiftingLink] = useState<boolean>();
-	const [userIsDefined, setUserIsDefined] = useState<boolean>();
+	const [identityLinks, setIdentityLinks] = useState<DropdownLinkType[]>();
```
oliverlloyd, line 137: Can we break this refactor out? This seems like a big change

src/web/components/MostViewed/MostViewedFooter/MostViewedFooterGrid.tsx
```diff
@@ -193,7 +193,9 @@ export const MostViewedFooterGrid = ({ data, sectionName, pillar }: Props) => {
 					aria-labelledby={`tabs-popular-${i}-tab`}
 					data-link-name={tab.heading}
 					data-testid={tab.heading}
-					data-link-context={`most-read/${sectionName}`}
+					data-link-context={
+						sectionName ? `most-read/${sectionName}` : ''
+					}
```
oliverlloyd, line 198: Is there any impact to setting `data-link-context` as an empty string? Are we happy that the code reading this will work in this situation?

src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -84,7 +84,7 @@ const generateSignInUrl = (
 	currentTest: CurrentABTest,
 ) => {
 	// url of the article, return user here after sign in/registration
-	const returnUrl = `${CAPI.config.host}/${CAPI.pageId}`;
+	const returnUrl = `${CAPI.config.host || ''}/${CAPI.pageId}`;
```
oliverlloyd, line 87: This looks like it could result in an invalid return url

src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -98,7 +98,7 @@ const generateSignInUrl = (
 	};
 
 	return `${
-		CAPI.config.idUrl
+		CAPI.config.idUrl || ''
```
oliverlloyd, line 101: Ditto on opening up a vector for an invalid url

src/web/lib/getCommentContext.ts
oliverlloyd, line 99: I think we want to move this refactor out so that we can verify it and maybe add some tests? Ideally we can work out a way too achieve this without having to use ts-ignore but lets see

src/web/lib/getZIndex.tsx
oliverlloyd: Is it possible to make it so `decideIndex` isn't able to return `null`?

src/web/lib/isLight.ts
```diff
@@ -8,7 +8,7 @@ function hexToRgb(
 } | null {
 	// Expand shorthand form (e.g. "03F") to full form (e.g. "0033FF")
 	const shorthandRegex = /^#?([a-f\d])([a-f\d])([a-f\d])$/i;
-	hex = hex.replace(shorthandRegex, (m, r, g, b) => {
+	hex = hex.replace(shorthandRegex, (_, r: string, g: string, b: string) => {
```
oliverlloyd, line 11: 👍

src/web/lib/readerRevenueDevUtils.ts
```diff
@@ -132,7 +132,9 @@ const changeGeolocation = (
 		.then((current) => {
 			/* eslint-disable no-alert */
 			const geo = window.prompt(
-				`Enter two-letter geolocation code (e.g. GB, US, AU). Current is ${current}.`,
+				`Enter two-letter geolocation code (e.g. GB, US, AU). Current is ${
+					current || 'undefined'
+				}.`,
```
oliverlloyd, line 137: We should not be designing in failure. If there's a use case where this is undefined then the message should speak to that in a useful way - eg. Please set your country etc.

src/web/server/render.ts
```diff
@@ -67,8 +67,10 @@ export const render = (
 
 		res.status(200).send(resp);
 	} catch (e) {
-		// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-		res.status(500).send(`<pre>${e.stack}</pre>`);
+		const error = e as Error;
+		res.status(500).send(
+			`<pre>${error.stack || `Unknown error: ${error.message}`}</pre>`,
```
oliverlloyd, line 72: I'm not sure we should be prefixing potentially reader facing error messages in the code

oliverlloyd (commented):
In terms of how we handle errors, I'm inclined to think Typescript is right here and we should not be console logging errors.

As developers we reach for `console.error` as a tool to help us when developing features because it is the most useful solution at that time. But for production code it has a negative impact. It surfaces errors to the reader, errors that they don't need to see and might not even be relevant or a problem.

Instead we should be reviewing each error and either ignoring it or passing it up to Sentry

src/amp/components/Analytics.tsx
gtrufitt: +1, that'd be classic showing `undefined` under section name on all articles 😆

gtrufitt (changes requested):
There's some good points from Oliver RE designing in failure, we should try to move towards not accepting null and increasing test coverage as there's a few cases here where scary strings would end up in the UI

oliverlloyd:
Closing as stale

PR #2593, opened 2021-02-22
Tiny headline

oliverlloyd:
## What?
Builds out better support for the `tiny` headline size, mainly around the `QuoteIcon`

## Why?
To ensure we fully support the api that exists and also so that we can show a tiny `Updated 7 min ago` headline with a pulsing dot on a liveblog


oliverlloyd:
[Tiny Headline](https://trello.com/c/WRSX43eS/2486-tiny-headline)

shtukas (approved):

PR #2594, opened 2021-02-22
Refactor layouts

oliverlloyd:
## What?
Refactors the layout components to use explicit props instead of CAPI and inline code more.

## Why?
To make future expansion and maintenance easier. For example, we will be adding new layouts for designs such as immersive opinion soon and these types of changes are easier when there are fewer abstractions to consider.


oliverlloyd:
[Refactor Layouts](https://trello.com/c/KMxtRYUX/2487-refactor-layouts)

shtukas (approved):

PR #2596, opened 2021-02-22
isLiveBlog

oliverlloyd:
## What?
Adds the properties `isLiveBlog` and `isLive` to the DCR model

## Why?
So that we an determine is an article is live or not, giving us the ability to decide to do something different with it


liywjl (approved):
👍

oliverlloyd:
[isLiveBlog](https://trello.com/c/icPIypNI/2491-isliveblog)

PR #2597, opened 2021-02-22
Reduce > map

oliverlloyd:
## What?
Replaces a reduce function with a map function

## Why?
Because I find it hard to read reduce functions

Plus, Twitter told me to do it
https://twitter.com/jaffathecake/status/1213077702300852224
https://twitter.com/dan_abramov/status/1338253118199508992

### Why are we even using indexes?
Because we need a way to identify an element between the server and client. A better root solution would be to have some form of `elementId` property on the model.

### DIFF
--- a/src/model/window-guardian.ts
+++ b/src/model/window-guardian.ts
@@ -68,25 +68,21 @@ const makeWindowGuardianConfig = (
 
 export const makeGuardianBrowserCAPI = (CAPI: CAPIType): CAPIBrowserType => {
 	// For some elements it is important to keep thier index in the `elements` array
-
 	const blockElementWithIndex = <T extends CAPIElement>(
 		blocks: { elements: CAPIElement[] }[],
 		blockElementType: CAPIElement['_type'],
 		indexName: string,
 	): T[] => {
 		if (blocks.length === 0) return [];
-		return blocks[0].elements.reduce(
-			(acc: T[], element: CAPIElement, index: number) => {
-				if (element._type === blockElementType) {
-					acc.push({
-						...element,
-						[indexName]: index,
-					} as T);
-				}
-				return acc;
-			},
-			[] as T[],
-		);
+		return blocks[0].elements.map((element: CAPIElement, index: number) => {
+			if (element._type === blockElementType) {
+				return {
+					...element,
+					[indexName]: index,
+				};
+			}
+			return element;
+		}) as T[];
 	};
 
 	// If our element type is one that can contain third party content that can track user, but hasn't been marked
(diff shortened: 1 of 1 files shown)


liywjl (approved):
👍

oliverlloyd:
[Reduce > map](https://trello.com/c/Exvn9nZM/2490-reduce-map)

buck06191:
I don't know about performance considerations, but I personally would never use `reduce` to produce the same data structure I'm starting with. That feels like what `map` is literally designed for.
I feel like this image clarifies my mental model
image
`reduce` should move thing to either another data structure or it should _reduce_ things down to a single value e.g. using `reduce` to get the sum of an array or something.

All that is to say, I prefer this PR over the original code. I would like to know the original motivation though

src/model/window-guardian.ts
```diff
@@ -68,25 +68,21 @@ const makeWindowGuardianConfig = (
...
-			[] as T[],
-		);
+		return blocks[0].elements.map((element: CAPIElement, index: number) => {
+			if (element._type === blockElementType) {
+				return {
+					...element,
+					[indexName]: index,
+				};
+			}
+			return element;
```
jfsoul, line 84: I'm not sure this will have the same behaviour as before will it? This is leaving elements unchanged if they aren't of `blockElementType` whereas the reduce actually filtered them out.  I think we'd need to do a flatMap / filter then map. I do think that would be easier to read than the reduce though.

jfsoul (changes requested):

src/model/window-guardian.ts
```diff
@@ -68,25 +68,21 @@ const makeWindowGuardianConfig = (
...
-			(acc: T[], element: CAPIElement, index: number) => {
-				if (element._type === blockElementType) {
-					acc.push({
-						...element,
-						[indexName]: index,
-					} as T);
-				}
-				return acc;
-			},
-			[] as T[],
```
buck06191, line 88: I feel like this is the key bit. If your accumulator is the same type as your original data structure, I wonder what the goal of using `reduce` instead of `map` is

buck06191 (approved):
LGTM

src/model/window-guardian.ts
buck06191: That's a great catch. `filter` and then `map` would definitely be more readable because the logic follows the function purposes

src/model/window-guardian.ts
buck06191: The goal is that this actually need `filter().map()`

shtukas (approved):

src/model/window-guardian.ts
oliverlloyd: I've made it `map` then `filter`, as per our discussion on chat. Thanks for helping here!

jfsoul (approved):

src/model/window-guardian.ts
mxdvl: Was this due to type predicate, as we’re running into #4582?

PR #2598, opened 2021-02-22
Remove webpack-cli

liywjl:
## What does this change?
Remove `webpack-cli` package, as its unused


shtukas (approved):

liywjl:
Looks like `webpack-cli` is acctually used implicitly in the `makefile`

PR #2599, opened 2021-02-22
GitLens

oliverlloyd:
## What?
Adds [GitLens](https://github.com/eamodio/vscode-gitlens) as a recommend workspace extensions

## Why?
Because we write such great PRs at the Guardianand it's really useful to be able to link through to them simply by hovering a line

### Some git config
I've added a `.git-blame-ignore-revs` file to include a large linting edit made a few months ago which touched most lines of the code.

There's an argument against making large linting changes on repos which says that it's bad because it destroys the git history. This is true but it's a problem that should be solved by tooling - which is what's happening here.


liywjl (approved):
👍  awesome!!! can't wait to be able to using git history on VSC again 🙂

gtrufitt (approved):

PR #2600, opened 2021-02-22
Live Stories

oliverlloyd:
## What?
Adds stories for when an article is a live or dead blog

## Why?
This renders a basic looking article with no special styles (because we've no added them yet) but now we have a baseline to compare changes against which is useful to understand the impact of the liveblog work going forwards


liywjl (approved):
👍

PR #2607, opened 2021-02-23
Add VineBlockElement

shtukas:
This is @nitro-marky's work presented in a different PR. 

## What does this change?

Implements Vine embeds (I live to Vine). 

Rather brillitantly at the time this was added the Vine archive was down, so it looks broken at the moment but it's like that on PROD too! It is hidden behind `ClickToView`.

It is also hidden by a `clickToView` for consent reasons, and to hide the broken parts straight away.

Before:

Screenshot 2021-02-19 at 15 39 10

After:

Screenshot 2021-02-22 at 17 09 05

## Why?

no-answer


buck06191 (approved):
Excellent work. Have a vine for your troubles.

giphy

PR #2608, opened 2021-02-23
InteractiveBlockElements in DCR 

gtrufitt:
## What does this change?

- Adds interactive block element to render a link on the server
- Adds a portal for interactive block elements, and loads this with loadable components to split out the component.
- Adds curl AMD loader when we have an interactive block element on the page. It is pasted directly in, the loader has not been updated since 2016.
- Loads the script from the `scriptUrl` via `window.require` which is accessible due to loading the curl loader

Depends on https://github.com/guardian/dotcom-rendering/pull/2616 and https://github.com/guardian/frontend/pull/23547 and (elementId) https://github.com/guardian/dotcom-rendering/pull/2615

## Screenshots

image

image

## Follow up work

As mentioned in the [documentation](https://github.com/guardian/dotcom-rendering/pull/2608/files#diff-7535f97e603cacda22b68f5923c7440a074dfa33e16324bfbabdd858ed59f75dR10-R65), there is work required to remove boot.js, which is used for 99% of articles with interactive embed elements. This means we don't need to use the curl loader for the majority of pageviews with an interactive embed. PR to follow.


src/web/server/htmlTemplate.ts
```diff
@@ -214,6 +214,15 @@ export const htmlTemplate = ({
...
                 </script>
 
+				<script>
+					window.curlConfig = {
+						baseUrl: '${CDN}assets',
+						apiName: 'require'
+					};
+					window.curl = window.curlConfig;
+				</script>
+
```
gtrufitt, line 224: This is the contract for the curl loader

src/web/server/document.tsx
```diff
@@ -173,22 +178,33 @@ export const document = ({ data }: Props): string => {
...
 			...getScriptArrayFromChunkName('ophan'),
 			CAPI.config && { src: CAPI.config.commercialBundleUrl },
 			...getScriptArrayFromChunkName('sentryLoader'),
 			...getScriptArrayFromChunkName('dynamicImport'),
+			pageHasInteractiveElements && {
+				src: `${CDN}static/frontend/js/curl-with-js-and-domReady.js`,
+			},
 			...arrayOfLoadableScriptObjects, // This includes the 'react' entry point
-		].filter(Boolean),
+		].filter(isDefined), // We use the TypeGuard to keep TS happy
```
gtrufitt, line 207: It wasn't happy with the previous falsy bouncer, had to add a specific Type Guard here for TS to pass

src/web/components/elements/InteractiveBlockComponent.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+*/
+
+export const InteractiveBlockComponent = ({ url, scriptUrl, alt }: Props) => {
+	const wrapperRef = useRef<HTMLDivElement>(null);
+	useEffect(() => {
+		if (scriptUrl) {
+			// We're going to use curl AMD loader to load the script that the
+			// interactive has given us.
+			window.require(
```
gtrufitt, line 76: Matches up with https://github.com/guardian/frontend/blob/e38da96c59eaaef143b9f89b2318eab9dcf2fc5a/common/app/views/fragments/inlineJSNonBlocking.scala.html#L52

.prettierignore
```diff
@@ -19,3 +19,6 @@ node_modules/
 !/cypress/**/*.js
 !/cypress/**/*.ts
 !/scripts/**/*.js
+
+# Ignore specific files
+/**/*/curl-with-js-and-domReady.js
```
gtrufitt, line 24: This file is already minifed and compiled, we don't want to touch it

src/web/components/elements/InteractiveBlockComponent.tsx
```diff
@@ -0,0 +1,102 @@
...
+	- navigate
+	- scroll-to
+	- get-location
+	- get-position
+	- monitor-position
+
+It has not been updated since 2016.
+
+MIGRATION FROM FRONTEND
+- TODO For the standard boot.js, we will re-write the behavior in modern JS to avoid the requirement of an AMD loader
```
buck06191, line 60: Is this still a TODO?

buck06191 (approved):
It LGTM, but want to check first on a `TODO` inside some inline documentation? Is this to do as part of this PR, or is it future work?

cypress/integration/e2e/article.elements.spec.js
```diff
@@ -117,12 +117,50 @@ describe('Elements', function () {
...
+					)
+					.its('0.contentDocument.body')
+					.should('not.be.empty')
+					.then(cy.wrap);
+			};
+
+			cy.visit(
+				'Article?url=https://www.theguardian.com/us-news/2017/jan/17/donald-trump-america-great-again-northampton-county-pennsylvania',
+			);
+			console.log(getIframeBody());
```
oliverlloyd, line 160: Can remove

index.d.ts
```diff
@@ -730,7 +731,8 @@ type IslandType =
 	| 'map-block-element'
 	| 'spotify-block-element'
 	| 'video-facebook-block-element'
-	| 'vine-block-element';
+	| 'vine-block-element'
+	| 'interactive-block';
```
oliverlloyd, line 735: Heads up: this will choke when elementsId lands but is simple to fix

src/lib/content.d.ts
```diff
@@ -215,13 +215,15 @@ interface InteractiveAtomBlockElement extends InteractiveAtomBlockElementBase {
...
 
+// Can't guarantee anything in interactiveBlockElement :shrug:
 interface InteractiveBlockElement {
 	_type: 'model.dotcomrendering.pageElements.InteractiveBlockElement';
 	url?: string;
 	isMandatory?: boolean;
 	scriptUrl?: string;
 	alt?: string;
 	role?: RoleType;
+	interactiveBlockIndex?: number;
```
oliverlloyd, line 226: Note to remove this when merging in elementId

src/web/lib/ElementRenderer.tsx
```diff
@@ -335,6 +336,16 @@ export const ElementRenderer = ({
 					/>
 				</Figure>
 			);
+		case 'model.dotcomrendering.pageElements.InteractiveBlockElement':
+			return (
+				<Figure role={element.role} id={`interactive-block-${index}`}>
```
oliverlloyd, line 341: Note to remove the id at merge

src/web/server/document.tsx
oliverlloyd, line 207: If we inline this does that make the code simpler or add a bunch of ts issues?

oliverlloyd (approved):
Thanks for keeping this understandable when its actually a very complex area. I've made some notes to try and make the merge of elementId easier

buck06191:
[interactiveBlockElement - Parity with curl AMD loader](https://trello.com/c/iVox5AwF/1849-interactiveblockelement-parity-with-curl-amd-loader)

src/web/components/elements/InteractiveBlockComponent.tsx
gtrufitt: Yup, will follow up PR.

src/web/server/document.tsx
gtrufitt: I think this is a rare occasion where inlining reduced comprehension because we lose the function name that specifically helps with the comprehension

gtrufitt:
> Thanks for keeping this understandable when its actually a very complex area. I've made some notes to try and make the merge of elementId easier

👍 Thanks, elementId in and those changes made

oliverlloyd (approved):

cypress/integration/e2e/article.elements.spec.js
```diff
@@ -117,12 +117,49 @@ describe('Elements', function () {
...
 			);
 
 			getIframeBody().contains('radiolab');
 		});
+
+		it('should render the interactive using a boot.js', function () {
+			const getIframeBody = () => {
+				return cy
+					.get(
+						'[data-cypress="interactive-element-LA%20Rams%20dead%20cap%20numbers"] > iframe',
```
oliverlloyd, line 132: Maybe you can simplify these data attributes to be easier to match?

src/web/components/elements/InteractiveBlockComponent.tsx
```diff
@@ -0,0 +1,105 @@
...
+						);
+					}
+				},
+			);
+		}
+	}, [scriptUrl]);
+
+	return (
+		<div
+			data-cypress={`interactive-element-${encodeURI(alt || '')}`}
```
oliverlloyd, line 97: This could probably be simpler, maybe just `interactive-element`?

gtrufitt: I need a specific interactive-element on the page, and I don't have an index?

oliverlloyd: The test article only has one interactive element so maybe it just works?

oliverlloyd: Alternatively, we could base64 encode the url so that the selector doesn't have weird characters in it?

gtrufitt: > The test article only has one interactive element so maybe it just works?

There's multiple in one of the test pages.

> Alternatively, we could base64 encode the url so that the selector doesn't have weird characters in it?

Sure, don't mind tbh, think maybe that that is 6-of-one-etc because for a selector for the test environment you either have an encoded uri or encoded base 64... I guess it comes to what's better on PROD? But base64 encoded URL will still be pretty long. Ideally there'd be a string -> short ID algo built in to JS..?

oliverlloyd: 🤔 I was only saying all these things becuase I saw your Cypress tests were failing on that selector but they seem to pass now so sure, whatever works

oliverlloyd:
Something off with Storybook, it can'd find `window.require`

buck06191:
> Something off with Storybook, it can'd find `window.require`

Is it to do with it running it all in an `iframe`?

gtrufitt:
> Something off with Storybook, it can'd find window.require
> Is it to do with it running it all in an iframe?

Locally storybook could find the static Curl loader file (a plain JS file that gets copied over and served as-is). I added this to `preview-head.html` which storybook uses as the `<head>` but Chromatic's storybook couldn't find it, 404ing on the request.

Hidden in the [Chromatic docs](https://www.chromatic.com/docs/setup#run-chromatic):

> We use the build-storybook script from your package.json by default. If you have customized the storybook script (for example, adding a static directory with -s), check that you've done the same for <script>build-storybook.

We *had* customised the script with a static directory for `storybook` but not `build-storybook`.

I added that here: [`055b49f` (#2608)](https://github.com/guardian/dotcom-rendering/pull/2608/commits/055b49fb93f83880e0dcc3e2ec269b7c36aecf2f)

src/web/components/elements/InteractiveBlockComponent.tsx
gtrufitt: > I was only saying all these things becuase I saw your Cypress tests were failing on that selector but they seem to pass now so sure, whatever works

Oh that was because I forgot to add `interactiveBlockElement` to `typesThatNeedHydrating` 😆

PR #2609, opened 2021-02-23
Refactoring: ordering

shtukas:
Simply maintain the cases in ElementRenderer in alphabetical order 📐

### DIFF
--- a/src/web/lib/ElementRenderer.tsx
+++ b/src/web/lib/ElementRenderer.tsx
@@ -102,6 +102,19 @@ export const ElementRenderer = ({
 					quoted={element.quoted}
 				/>
 			);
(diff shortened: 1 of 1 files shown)


nitro-marky (approved):

buck06191:
Interesting note here: when we were debugging `ElementRenderer` in #2587 one of the potential causes of the error we got, later found to be due to PrismJS, was the import order. The fact things were out of alphabetical order actually made it harder to debug and we had to spend a good 10 minutes moving around code blocks to see the impact it had

shtukas:
@buck06191 Oh wow. That's amazing. Thanks for sharing that. 

If it ever happens that the order becomes relevant one day, here or in another codebase, I would specify in the source code itself :)

PR #2613, opened 2021-02-25
Remove validation check

oliverlloyd:
## What does this change?
This removes a function that was being used to ensure that the `isThirdPartyTracking` property was present on certain elements

## Why?
Because the property was required so we already ensure it exists using json schema validation

### DIFF
--- a/src/model/window-guardian.ts
+++ b/src/model/window-guardian.ts
@@ -86,17 +86,6 @@ export const makeGuardianBrowserCAPI = (CAPI: CAPIType): CAPIBrowserType => {
 			.filter((element) => element._type === blockElementType) as T[];
 	};
 
-	// If our element type is one that can contain third party content that can track user, but hasn't been marked
-	// as 'tracking' by the back-end, we remove it from the array of elements that we'll hydrate on the client-side.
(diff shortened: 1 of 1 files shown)


shtukas (approved):

liywjl (approved):
👍

src/model/window-guardian.ts
```diff
@@ -86,17 +86,6 @@ export const makeGuardianBrowserCAPI = (CAPI: CAPIType): CAPIBrowserType => {
 			.filter((element) => element._type === blockElementType) as T[];
 	};
 
-	// If our element type is one that can contain third party content that can track user, but hasn't been marked
```
frj, line 89: As discussed this check is required in the case where element.isThirdPartyTracking === false.

This prevents unnecessary hydration of embedded content that is not 'tracking' and for which we dont display a click to view overlay.

oliverlloyd: Thanks @frj this (and our chat offline) clears things up. I'm going to close this PR and wrap this work up into another PR I'm working on with the larger refactor work around how we hydrate.

But from this the use case I will aim to support is:

`EmbedBlockElements` (who only track _sometimes_) will only by hydrated if `isThirdPartyTracking === true` - all other tracking elements always track so will always be hydrated

PR #2614, opened 2021-02-25
Correct the Sam Levin|Sam Levine problem

shtukas:
BEFORE: 

Screenshot 2021-02-25 at 12 31 29

Both pointing to: http://www.theguardian.com/profile/sam-levin

AFTER:

Screenshot 2021-02-25 at 12 32 03

Pointing to respectively
- http://www.theguardian.com/profile/sam-levin
- http://www.theguardian.com/profile/sam-levine

### DIFF
--- a/src/web/components/BylineLink.tsx
+++ b/src/web/components/BylineLink.tsx
@@ -9,6 +9,38 @@ const getContributorTags = (tags: TagType[]): TagType[] => {
 	return tags.filter((t) => t.type === 'Contributor');
 };
 
+const applyCleverOrderingForMatching = (titles: string[]): string[] => {
+
+	/*
+		Q: Why does this function exist ?
+
+		A: We had a case where the byline was "Sam Levin in Los Angeles and Sam Levine in New York",
+		which would lead to the regex: Sam Levin|Sam Levine. Unfortunately the first expression would take priority and we would end up
+		with a bylineAsTokens equal to [
+			'',
+			'Sam Levin',
+			' in Los Angeles and ',
+			'Sam Levin',
+			'e in New York'
+		]
+
+		The solution here is simply to order the titles in decreasing lenght. This ensures that in a case where one name is a substring or another
+		name, then the longest name is tried first, preventing the problem we had. We now have a bylineAsTokens equals to [
+			'',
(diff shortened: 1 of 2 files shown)


src/web/components/BylineLink.tsx
```diff
@@ -9,6 +9,42 @@ const getContributorTags = (tags: TagType[]): TagType[] => {
...
+
+		This function doesn't change any thing for the "standard" case, but solves the "Sam Levin|Sam Levine" case and similar cases.
+
+		ps: If one day another problem comes up whose solution would be incompatible with this correction, then a better solution for the matching
+		will have to be invented from the ground up, but for the moment, this will do :)
+	*/
+
+	return titles
+		.sort((a, b) => {
+			return a.localeCompare(b);
```
gtrufitt, line 43: the `localeCompare` thing is throwing me, would `a.length - b.length` work the same way?

shtukas: Yeah. StackOverflow copy-paste 😅 Let me correct that.

src/web/components/BylineLink.tsx
```diff
@@ -9,6 +9,42 @@ const getContributorTags = (tags: TagType[]): TagType[] => {
...
+		})
+		.reverse();
+};
+
 // This crazy function aims to split bylines such as
```
oliverlloyd, line 48: I think this comment was lifted from Frontend but it seems things are getting crazier!

But seriously, should be centralise the comments/descriptions to this one place?

src/web/components/BylineLink.tsx
oliverlloyd: +1 I don't know what localCompare is so possibly other swon't either

shtukas: @gtrufitt Done.

shtukas: @gtrufitt Done.

oliverlloyd (commented):
Shall we maybe add some tests for this? That might be easier to read rather than trying to explain this with text?

gtrufitt (approved):

oliverlloyd (approved):

PR #2615, opened 2021-02-25
elementId

oliverlloyd:
## What does this change?
Now that we have an `elementId` set on each element we can remove our usage of indexes to identify elements between the server and the client and use this property instead.

We use this pattern of identification between the server and client for the purposes of partial hydration. We server side render the whole page but only hydrate select pieces. We use ids that we mark divs with on the server and send these, along with the elements themsleves, down to the client.

## Why?
Because indexes are well known as a poor way to identify things, they depend on order and are subject to change in unpredictable ways. In addition, now that we're hydrating elements both in the article body but also in main media we have two arrays, so that creates index duplication, breaking our usage of it.

`elementId` is unique and much easier to work with.

### Highlights
With addition of `elementId` to the model we're both updating the json schema as well as our fixture files

#### How we're passing data for hydration
Previously, we had a separate property for each element type but now we're using a single `elementsToHydrate` property
Screenshot 2021-02-25 at 15 12 13

#### No more `IslandType`
This made sense when we have a small number of singular islands but now that we're using ids it makes sense to just type these root ids as simply strings

#### ClickToView and tracking
Some elements implement third party tracking and we overlay these with a ClickToView componet. This component needs hydration, but not always. For `EmbedBlockElement`s, these will sometimes track the user and sometimes not.

The logic to implement this was already in place and working but with the change to a single model property I had to refactor it

### DIFF
--- a/src/web/components/HydrateOnce.tsx
+++ b/src/web/components/HydrateOnce.tsx
@@ -15,11 +15,11 @@ const isReady = (dependencies: unknown[]): boolean => {
 };
 
 /**
- * Finds the element with the same id as `root` and calls `ReactDOM.hydrate(children, element)`. Only
+ * Finds the element with the same id as `rootId` and calls `ReactDOM.hydrate(children, element)`. Only
  * executes once and only after all variables in `waitFor` are defined.
- * @param {String} root - The id of the element to hydrate
+ * @param {String} rootId - The id of the element to hydrate
+ * @param {String} name - Used to log out performance data via `initPerf()`
  * @param children - The react elements passed to ReactDOM.hydrate()
- * @param {number} index - Used when there are multiple elements the same (eg. RichLinks)
  * @param {Array} waitFor - An array of variables that must be defined before the task is executed
  * */
 export const HydrateOnce = ({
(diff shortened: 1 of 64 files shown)


shtukas (approved):

src/web/components/HydrateOnce.tsx
```diff
@@ -22,12 +22,16 @@ const isReady = (dependencies: unknown[]): boolean => {
  * @param {number} index - Used when there are multiple elements the same (eg. RichLinks)
  * @param {Array} waitFor - An array of variables that must be defined before the task is executed
```
gtrufitt, line 23: Need to update these docs. I'm not sure what `name` is

gtrufitt (approved):
Huge reduction in complexity 👍

src/web/components/HydrateOnce.tsx
oliverlloyd: `name` is just me trying to preserve the performance output so that we log out the same information as before, eg. `hydrate-youtube`. I think I still need to run through and pass these strings in. I wonder if there's a better way of doing this, like grabbing the component name

oliverlloyd: I've gone with removing this `name` prop altogether. We were using it to log individual hydration times to the console but we don't do anything more with it yet so there's not really a use case here that justifies the extra complexity that wuold be needed to continue to support using a different log name for each hydration step

liywjl (approved):
👍

oliverlloyd:
[Indexes > elementId](https://trello.com/c/zcduwaPJ/2671-indexes-elementid)

PR #2616, opened 2021-02-25
Make InteractiveBlockElements their own component in AMP

gtrufitt:
## What does this change?

Breaks out interactiveBlockElement from interactiveAtomBlockElement.

We also throw an error if the interactive is mandatory to force the article not to appear in AMP.

### DIFF
--- a/src/amp/components/elements/InteractiveBlockComponent.tsx
+++ b/src/amp/components/elements/InteractiveBlockComponent.tsx
@@ -1,0 +1,49 @@
+import React from 'react';
+import { css } from 'emotion';
+import { ShowMoreButton } from '@root/src/amp/components/ShowMoreButton';
+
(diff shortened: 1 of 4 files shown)


src/amp/components/elements/InteractiveBlockComponent.tsx
```diff
@@ -0,0 +1,53 @@
...
+	&[overflow] {
+		position: absolute;
+		bottom: 0;
+		left: 0;
+		right: 0;
+	}
+`;
+
+export const InteractiveBlockComponent: React.SFC<{
+	url?: string;
```
gtrufitt, line 22: The majority of this is copied over from the interactive atom component that the interactive block element was already using

jfsoul (approved):

PR #2617, opened 2021-02-25
Update dot styling

nitro-marky:
## What does this change?

This fixes the bullet styling issue which mainly occurs in Opinion articles by assigning it the relevant pillar.  

On rare occasions these points appear in the `standfirst` as well so this also address that, again with the relevant pillar colour. This is why the Theme prop is is passed to `standfirst`.

Articles that have used a list instead of a dot will not be changed.

Stories have been updated with pillars. 

### Before

Screenshot 2021-02-25 at 16 05 23

### After

Screenshot 2021-02-25 at 11 11 16

Just to note, after speaking to @HarryFischer, the dots in `standfirst` will retain their grey colour, only the footer dot will be based on the Palette.

Screenshot 2021-03-01 at 11 32 59

### DIFF
--- a/src/model/add-dots.test.ts
+++ b/src/model/add-dots.test.ts
@@ -1,0 +1,191 @@
+import { addDots } from './add-dots';
+import { bodyJSON } from './exampleBodyJSON';
+
+const example = JSON.parse(bodyJSON);
+
+describe('Replacing Middot with fake dot', () => {
+	it('creates an identical but new object when no changes are needed', () => {
+		expect(addDots(example)).not.toBe(example); // We created a new object
+		expect(addDots(example)).toEqual(example); // The new object is what we expect
+	});
+
+	it('replace the Middot', () => {
+		const input = {
+			...example,
+			blocks: [
+				{
+					elements: [
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I am the first paragraph</p>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.SubheadingBlockElement',
+							html: '<p>* * *</p>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I should become a drop cap.</p>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>• I should have a dot.</p>',
+						},
+					],
+				},
+			],
+		};
+
+		const expectedOutput = {
+			...example,
+			blocks: [
+				{
+					elements: [
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I am the first paragraph</p>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.DividerBlockElement',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I should become a drop cap.</p>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html:
+								'<p><span data-dcr-style="bullet"></span> I should NOT have a dot.</p>',
+						},
+					],
+				},
+			],
+		};
+
+		expect(addDots(input)).toEqual(expectedOutput);
+	});
+
+	it('Should not make any changes without dots', () => {
+		const input = {
+			...example,
+			blocks: [
+				{
+					elements: [
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I am the first paragraph</p>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.SubheadingBlockElement',
+							html: '<p>***</p>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I should become a drop cap.</p>',
+						},
+					],
+				},
+			],
+		};
+
+		const expectedOutput = {
+			...example,
+			blocks: [
+				{
+					elements: [
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I am the first paragraph</p>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.DividerBlockElement',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I should become a drop cap.</p>',
+						},
+					],
+				},
+			],
+		};
+
+		expect(addDots(input)).toEqual(expectedOutput);
+	});
+
+	it('Should not insert fake dot', () => {
+		const input = {
+			...example,
+			blocks: [
+				{
+					elements: [
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I am the first paragraph</p>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.SubheadingBlockElement',
+							html: '<h2><strong>* * *</strong></h2>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I should become a drop cap.</p>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I should NOT become a drop cap.</p>',
+						},
+					],
+				},
+			],
+		};
+
+		const expectedOutput = {
+			...example,
+			blocks: [
+				{
+					elements: [
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I am the first paragraph</p>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.DividerBlockElement',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I should become a drop cap.</p>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html:
+								'<p><span data-dcr-style="bullet"></span> I should NOT have a dot.</p>',
+						},
+					],
+				},
+			],
+		};
+		expect(addDots(input)).not.toEqual(expectedOutput);
+	});
+});
--- a/src/model/add-dots.test.ts
+++ b/src/model/add-dots.test.ts
@@ -1,12 +1,12 @@
-import { addDots } from './add-dots';
+import { enhancedDots } from './add-dots';
 import { bodyJSON } from './exampleBodyJSON';
 
 const example = JSON.parse(bodyJSON);
 
 describe('Replacing Middot with fake dot', () => {
 	it('creates an identical but new object when no changes are needed', () => {
-		expect(addDots(example)).not.toBe(example); // We created a new object
-		expect(addDots(example)).toEqual(example); // The new object is what we expect
+		expect(enhancedDots(example)).not.toBe(example); // We created a new object
+		expect(enhancedDots(example)).toEqual(example); // The new object is what we expect
 	});
 
 	it('replace the Middot', () => {
@@ -70,7 +70,7 @@ describe('Replacing Middot with fake dot', () => {
 			],
 		};
 
-		expect(addDots(input)).toEqual(expectedOutput);
+		expect(enhancedDots(input)).toEqual(expectedOutput);
 	});
 
 	it('Should not make any changes without dots', () => {
@@ -123,7 +123,7 @@ describe('Replacing Middot with fake dot', () => {
 			],
 		};
 
-		expect(addDots(input)).toEqual(expectedOutput);
+		expect(enhancedDots(input)).toEqual(expectedOutput);
 	});
 
 	it('Should not insert fake dot', () => {
@@ -186,6 +186,6 @@ describe('Replacing Middot with fake dot', () => {
 				},
 			],
 		};
-		expect(addDots(input)).not.toEqual(expectedOutput);
+		expect(enhancedDots(input)).not.toEqual(expectedOutput);
 	});
 });
--- a/src/model/add-dots.test.ts
+++ b/src/model/add-dots.test.ts
@@ -22,7 +22,7 @@ describe('Replacing Middot with fake dot', () => {
 						},
 						{
 							_type:
-								'model.dotcomrendering.pageElements.SubheadingBlockElement',
+								'model.dotcomrendering.pageElements.TextBlockElement',
 							html: '<p>* * *</p>',
 						},
 						{
@@ -52,7 +52,8 @@ describe('Replacing Middot with fake dot', () => {
 						},
 						{
 							_type:
-								'model.dotcomrendering.pageElements.DividerBlockElement',
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>* * *</p>',
 						},
 						{
 							_type:
@@ -70,7 +71,7 @@ describe('Replacing Middot with fake dot', () => {
 			],
 		};
 
-		expect(enhancedDots(input)).toEqual(expectedOutput);
+		expect(enhancedDots(input)).not.toBe(expectedOutput);
 	});
 
 	it('Should not make any changes without dots', () => {
@@ -111,8 +112,10 @@ describe('Replacing Middot with fake dot', () => {
 						},
 						{
 							_type:
-								'model.dotcomrendering.pageElements.DividerBlockElement',
+								'model.dotcomrendering.pageElements.SubheadingBlockElement',
+							html: '<p>***</p>',
 						},
+
 						{
 							_type:
 								'model.dotcomrendering.pageElements.TextBlockElement',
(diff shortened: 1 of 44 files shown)


src/model/add-dots.ts
```diff
@@ -0,0 +1,60 @@
...
+			element._type ===
+				'model.dotcomrendering.pageElements.TextBlockElement' &&
+			hasDots(element)
+		) {
+			enhanced.push({
+				...element,
+			});
+		} else {
+			enhanced.push(element);
+		}
```
oliverlloyd, line 31: I think this is mutating the original element, which is something we generally want to avoid. The hasDots function should return a boolean and then a decision branched from there.

Eg. We can use the spread pattern to avoid mutation:

return {
  ...originalThing,
  someThing: newValue,
}

See: https://css-tricks.com/understanding-immutability-in-javascript/

Immutability is esp. useful in js because objects are passed as references, which leads to confusing errors.

Note. Some codebases will 'freeze' objects to make mutation illegal. We don't do this but we want out code to allow this without needing to be refactored.

shtukas (approved):

shtukas (approved):

src/model/add-dots.ts
```diff
@@ -0,0 +1,60 @@
+// Check if the block contains a dot, replace with an empty span with a bullet attribute
+const hasDots = (element: CAPIElement): boolean => {
+	if (element._type !== 'model.dotcomrendering.pageElements.TextBlockElement')
+		return false;
+
+	if (element.html.includes('•')) {
+		element.html = element.html.replace(
+			new RegExp('[•]', 'g'),
+			'<span data-dcr-style="bullet"></span>',
+		);
```
oliverlloyd, line 10: This _might_ work but string manipulation on html using regexes is _potentially_ brittle. For example, this doesn't deal with the html tags that might be wrapped around the dot. Eg. By using spans and inserting them inside the containing elements, we risk style conflicts where the browser might try to apply styles for a `em` or `li` tag but the span competes?

src/model/add-dots.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+
+const checkForDots = (elements: CAPIElement[]): CAPIElement[] => {
+	// Loop over elements and check if a dot is in the TextBlockElement
+	const enhanced: CAPIElement[] = [];
+	elements.forEach((element) => {
+		if (
+			element._type ===
+				'model.dotcomrendering.pageElements.TextBlockElement' &&
+			hasDots(element)
```
jfsoul, line 24: am I right in thinking this works because `hasDots` is side-effecting? I think now that we're explicitly changing the element in `hasDots` we should call it `transformDots` and have a return type of Element.

src/model/add-dots.ts
jfsoul, line 31: I think this `if` statement is redundant now. You can just do `elements.forEach(transformDots)`

src/model/add-dots.ts
```diff
@@ -0,0 +1,60 @@
...
+			});
+		} else {
+			enhanced.push(element);
+		}
+	});
+	return enhanced;
+};
+
+// On rare occasions the standfirst uses dots so it is also checked
+const checkStandFirstForDots = (standFirst: string) => {
```
jfsoul, line 37: I think you could consolidate this with the function that does the same thing but takes a `CAPIElement` - they're doing the same thing more or less.

src/web/components/ArticleHeadline.stories.tsx
```diff
@@ -232,6 +232,7 @@ export const Interview = () => (
 				<Standfirst
 					display={Display.Standard}
 					design={Design.Interview}
+					theme={Pillar.Culture}
```
oliverlloyd, line 235: Can we change this to pass `format`? This is the pattern we're moving towards.

src/web/components/Standfirst.tsx
```diff
@@ -42,6 +44,16 @@ const nestedStyles = css`
...
 	}
+
+	[data-dcr-style='bullet'] {
+		display: inline-block;
+		content: '';
+		border-radius: 0.375rem;
+		height: 0.75rem;
+		width: 0.75rem;
+		margin-right: 0.125rem;
+		background-color: ${pillarPalette[theme].main};
```
oliverlloyd, line 55: We have stopped using `pillarPalette` in favour of `palette` (which we build in `decidePalette`) - happy to pair on how we implement this!

src/web/components/elements/TextBlockComponent.tsx
```diff
@@ -149,6 +150,16 @@ const paraStyles = css`
 		position: relative;
 		vertical-align: baseline;
 	}
+
+	[data-dcr-style='bullet'] {
```
oliverlloyd, line 154: I'm a little unsure about this pattern for setting styles, I've not seen it before. I think we would be better off sticking to existing patterns?

If we want to communicate style we could use either semantic html (eg. style _all_ `<li>` tags the same or use react props for something more specific?

oliverlloyd (commented):
If it were me, I'd either use the react prop pattern here to communicate to the `TextBlockElement` that we want to style things differently or make a more generalised decision that we want to style `li` tags in text blocks with big dots.

Or a bit of both maybe? When `addDottiness` is true style 'li' tags with big pillar dots.

The other advantage to this is we offer Composer a route to enhance their own tool so that we can remove this cleaner completely

src/web/components/elements/TextBlockComponent.tsx
gtrufitt: In this case, my suggestion of this attribute selector pattern to be more explicit that they're dcr styles and to avoid clashes because we're dangerously setting inner html so don't have a prop to set but what's an 'existing pattern'? A class?

oliverlloyd: The danger here is we've started to build our own api for styles 😨 . Once you breach these kinds of lines other people tend to saunter along behind you and do the same thing, possibly on a larger scale.

Could we perhaps edit the html text to use an `li` element and then have a prop on TextBlockComponent telling it to style these `li` tags different - as we have here (but maybe using a `before` pseudo selector 🤷 )? That was we benefit in three ways:

1. Accessible - sematic html matters
2. Scalable, this will fit in with future work in Composer
3. Easier to reason about because dots in front of li tags is a thing

gtrufitt: `The danger here is we've started to build our own api for style`

I mean, it's just a selector?

> prop on TextBlockComponent telling it to style these li tags different - as we have here (but maybe using a before pseudo selector 🤷 

Yeah, totally. I think though a prop doesn't meet the goal - in fact we would want Composer to convert the usage (or provide differently) of a dot directly to the element. So we *do* need to mutate the html to get to remove the dot? 

I don't think these are semantically lists though, and are distinct for what *can* be marked up as a list in Composer. Maybe we need to convert these elements to a [`footer` element ](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/footer). 

So actually, maybe we're keen to introduce a new footnoteBlockComponent(??) element. It wouldn't make sense in standfirst though....

oliverlloyd: > we do need to mutate the html 

Oh sure, that's a given. The weird little (ascii?)dots need to be made bigger dots which means the html needs to be edited.

Maybe there's two things here. The dots you see in standfirsts, which really are lists, and these footer notes, which are also a thing in photo essays (of course they are).

In photo essays we replace them with their won special component, called HighlightBlockComponent

Screenshot 2021-02-25 at 17 46 10

You could use that or a variation of it?

You'd be creating a rule based on convention: That a standfirst can havve 1 or more bulletted items but the footer note can only have one

src/model/add-dots.ts
nitro-marky: Oliver and Pascal are keen to avoid mutation so I think this would help with that.

src/model/add-dots.ts
```diff
@@ -0,0 +1,56 @@
+// Wraps the last dot in a footer
+const transformLastDot = (element: string): string => {
+	const modified = `<footer>${element}</footer>`;
+	return modified.replace(
```
oliverlloyd, line 4: Will this get transpiled down for IE11?

src/model/add-dots.ts
```diff
@@ -0,0 +1,56 @@
...
+					...element,
+					html: transformLastDot(element.html),
+				});
+			} else {
+				enhanced.push({
+					...element,
+					html: element.html.replace(
+						new RegExp('[•]', 'g'),
+						'<span data-dcr-style="bullet"></span>',
+					),
```
oliverlloyd, line 30: We could rewrite this to

```
					html: `<footer>${transformDots(element.html)}</footer>`,
```

Which is a clearer

src/model/add-dots.ts
```diff
@@ -0,0 +1,56 @@
+// Wraps the last dot in a footer
+const transformLastDot = (element: string): string => {
```
oliverlloyd, line 2: Maybe `transformDots`? It's more descriptive and means the same function can be used in the body

src/model/add-dots.ts
```diff
@@ -0,0 +1,56 @@
...
+		}
+	});
+	return enhanced;
+};
+
+export const addDots = (data: CAPIType): CAPIType => {
+	data.standfirst = data.standfirst.replace(
+		new RegExp('[•]', 'g'),
+		'<span data-dcr-style="bullet"></span>',
+	);
```
oliverlloyd, line 44: This is mutating (sorry! 😉 ). I've added a suggestion in the return statement to show how we can spread the change in.

PS. You know you're mutating when you take the original object (`data` in this case) and assign something to it, as in `data.thing = newVal`.

src/model/add-dots.ts
```diff
@@ -0,0 +1,56 @@
...
+		return {
+			...block,
+			elements: checkForDots(block.elements),
+		};
+	});
+
+	return {
+		...data,
+		blocks: enhancedBlocks,
+	} as CAPIType;
```
oliverlloyd, line 55:
```
	return {
		...data,
        standfirst: transformDots(data.standfirst),
		blocks: enhancedBlocks,
	} as CAPIType;
```

src/web/components/Standfirst.tsx
```diff
@@ -5,16 +5,17 @@ import { neutral } from '@guardian/src-foundations/palette';
...
 
 type Props = {
 	display: Display;
 	design: Design;
+	theme?: Theme;
 	standfirst: string;
 };
 
-const nestedStyles = css`
+const nestedStyles = () => css`
```
oliverlloyd, line 18: This no longer needs to be a function

src/web/components/Standfirst.tsx
```diff
@@ -5,16 +5,17 @@ import { neutral } from '@guardian/src-foundations/palette';
...
 import { from } from '@guardian/src-foundations/mq';
 import { headline } from '@guardian/src-foundations/typography';
-import { Display, Design } from '@guardian/types';
+import { Display, Design, Theme } from '@guardian/types';
 import { sanitise } from '@frontend/lib/sanitise-html';
 
 type Props = {
 	display: Display;
 	design: Design;
+	theme?: Theme;
```
oliverlloyd, line 14: Are we using this?

src/web/components/elements/TextBlockComponent.tsx
```diff
@@ -99,7 +100,7 @@ const sanitiserOptions = {
 	},
 };
 
-const paraStyles = css`
+const paraStyles = (format: Format, palette: Palette) => css`
```
oliverlloyd, line 103: Are we using `format` here?

src/web/server/render.ts
```diff
@@ -52,8 +58,8 @@ export const render = (
 			.validateAsCAPIType()
 			.addDividers()
 			.enhanceBlockquotes()
+			.addDots()
```
oliverlloyd, line 61: I think `enhanceDots` is better in that it follows the established lexicon where we take an existing thing and enhance it

src/web/components/elements/TextBlockComponent.tsx
```diff
@@ -149,13 +150,24 @@ const paraStyles = css`
...
 	}
+
+	[data-dcr-style='bullet'] {
+		display: inline-block;
+		content: '';
+		border-radius: 0.375rem;
+		height: 0.75rem;
+		width: 0.75rem;
+		margin-right: 0.125rem;
+		background-color: ${palette.background.avatar};
```
oliverlloyd, line 161: `palette.background.avatar` is the wrong property to use. The risk with picking up values for other areas of the page is when they change (and they will) that decision will bleed into unrelated areas of the code. It's better to add in a specific decision function for this use case. Like `palette.fill.bullet` or maybe `palette.background.bullet`?

oliverlloyd (commented):
Almost there. Most of my changes are small so hopefully will be simple to carry through but shout if anything's not clear and I'm happy to go through it!

src/model/add-dots.test.ts
```diff
@@ -0,0 +1,131 @@
...
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>* * *</p>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I should become a drop cap.</p>',
+						},
```
oliverlloyd, line 32: This is from another test and might be confusing having it here

src/model/add-dots.test.ts
```diff
@@ -0,0 +1,131 @@
...
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html:
+								'<p><span data-dcr-style="bullet"></span> I should NOT have a dot.</p>',
```
oliverlloyd, line 67: It looks like the text after the dot has changed here which _should_ fail the test 🤔

src/model/add-dots.test.ts
```diff
@@ -0,0 +1,131 @@
...
+						},
+					],
+				},
+			],
+		};
+
+		expect(enhancedDots(input)).not.toBe(expectedOutput);
```
oliverlloyd, line 74: Ah, this is why it's passing, it's expecting a negative result. The name of this test suggests that it's  testing for a positive though so I think this test needs refactoring.

src/model/add-dots.test.ts
```diff
@@ -0,0 +1,131 @@
...
+						{
+							_type:
+								'model.dotcomrendering.pageElements.SubheadingBlockElement',
+							html: '<p>***</p>',
+						},
+						{
+							_type:
+								'model.dotcomrendering.pageElements.TextBlockElement',
+							html: '<p>I should become a drop cap.</p>',
+						},
```
oliverlloyd, line 97: I think it is valid to have a test with *s in but maybe make that an explicit test - eg. 'It does not incorrectly replace asterisks with dot spans'. Otherwise I think the generic test data should be lorem ipsum to make it clear it's not part of the test

src/model/add-dots.ts
```diff
@@ -0,0 +1,53 @@
...
+					...element,
+					html: `<footer>${transformDot(element.html)}</footer>`,
+				});
+			} else {
+				enhanced.push({
+					...element,
+					html: element.html.replace(
+						new RegExp('[•]', 'g'),
+						'<span data-dcr-style="bullet"></span>',
+					),
```
oliverlloyd, line 30: You can use `transformDot` here.

Eg.

```
				html: transformDots(element.html),
```

src/model/add-dots.ts
```diff
@@ -0,0 +1,53 @@
+// Current fix for incorrect use of Middot
+
+const transformDot = (html: string): string => {
```
oliverlloyd, line 3: The regex is using the `g` flag so it should really be `transformDots` as it will handle _muiltiple_ dots

oliverlloyd (approved):
Last few bits otherwise 👍

src/model/add-dots.test.ts
nitro-marky: I've done the test as a 'not to be equal', so the input is not equal to the output, therefore correct.

oliverlloyd: 🤔 Sorry to nag but the test is called `it('replace the Middot'` and the expected output says

```
'<p><span data-dcr-style="bullet"></span> I should NOT have a dot.</p>',
```

But it does have a dot so it's not super clear what this test is proving

oliverlloyd: One of the useful aspects of tests is they can act as a guide to what the code should do, so they form a sort of document for future developers to read and understand your intentions. For the benefit of these good people coming after us (and because these people may very well be _us_ after we've long forgotten about all this) we want these tests to be as clear and simple as possible

gtrufitt (changes requested):
I don't think you need to add the palette prop to `textblockComponent` with `Format` when you're already passing `format` in as a prop? You can just use `decidePalette` directly inside the component and remove the extra prop.

src/web/server/render.ts
```diff
@@ -22,6 +23,11 @@ class CAPIEnhancer {
 		return this;
 	}
 
+	enhancedDots() {
```
gtrufitt, line 26: Probably `enhanceDots` rather than `enhanced` for consistency?

src/model/add-dots.ts
```diff
@@ -0,0 +1,50 @@
...
+	elements.map((element, i) => {
+		if (
+			element._type ===
+				'model.dotcomrendering.pageElements.TextBlockElement' &&
+			element.html.includes('•')
+		) {
+			if (elements.length - 1 === i) {
+				enhanced.push({
+					...element,
+					html: `<footer>${transformDots(element.html)}</footer>`,
```
gtrufitt, line 22: 🎊 Nice!

oliverlloyd:
> I don't think you need to add the palette prop to `textblockComponent` with `Format` when you're already passing `format` in as a prop? You can just use `decidePalette` directly inside the component and remove the extra prop.

That's an interesting point. So at present, we only run `decidePalette` at the root of the stack in the same place we decide theme, design, etc. But there's no particular reason not to run it elsewhere. I guess the benefit is less prop drilling and the downside is more imports and execution.

gtrufitt:
> > I don't think you need to add the palette prop to `textblockComponent` with `Format` when you're already passing `format` in as a prop? You can just use `decidePalette` directly inside the component and remove the extra prop.
> 
> That's an interesting point. So at present, we only run `decidePalette` at the root of the stack in the same place we decide theme, design, etc. But there's no particular reason not to run it elsewhere. I guess the benefit is less prop drilling and the downside is more imports and execution.

Just seemed gnarly to have `format` prop next to a `pillar` prop that took the contents of that very same `format` prop?

I'd be happy passing `format` around everywhere and then calling `decideFunctions` where we need them inside components... It's potentially a nicer and reasonably understandable pattern?

gtrufitt:
I'll release Mark's PR though as I didn't realise it was more of a wider architectural question!

gtrufitt (approved):

oliverlloyd:
> > > I don't think you need to add the palette prop to `textblockComponent` with `Format` when you're already passing `format` in as a prop? You can just use `decidePalette` directly inside the component and remove the extra prop.
> > 
> > 
> > That's an interesting point. So at present, we only run `decidePalette` at the root of the stack in the same place we decide theme, design, etc. But there's no particular reason not to run it elsewhere. I guess the benefit is less prop drilling and the downside is more imports and execution.
> 
> Just seemed gnarly to have `format` prop next to a `pillar` prop that took the contents of that very same `format` prop?
> 
> I'd be happy passing `format` around everywhere and then calling `decideFunctions` where we need them inside components... It's potentially a nicer and reasonably understandable pattern?

Totally open to using decidePalette locally. Lets go with whatever works / feels best in each use case. I can see both approaches making sense in different scenarios.

PR #2622, opened 2021-02-26
Identity | Sign In Gate | Add SignInGate Copy Optimisation test

rebecca-thompson:
## What does this change?
This adds a new AB test, with **1x Control** (same as MainVariant) and **6x Copy Optimisation Variants** for the sign-in gate. Here are the 6 variants:

Variant - Transparency 1
var1

Variant - Transparency 2 
var2

Variant - Purpose
var3

Variant - Petition
var4

Variant - Belonging 1
var5

Variant - Belonging 2
var6


rebecca-thompson:
[Copy Test on the Sign In Gate (M/L)](https://trello.com/c/G15Alj3r/2822-copy-test-on-the-sign-in-gate-m-l)

src/web/components/SignInGate/displayRule.ts
```diff
@@ -72,3 +74,20 @@ export const isValidTag = (CAPI: CAPIBrowserType): boolean => {
 		CAPI.tags.map((tag) => tag.id).includes(invalidTag),
 	);
 };
+
+export const canShowCopyOpt = (
```
QuarpT, line 78: is this the same as the main-variant `canShow`?
If so we should export and use that one

src/web/experiments/tests/sign-in-gate-copy-opt.ts
```diff
@@ -0,0 +1,50 @@
+import { ABTest } from '@guardian/ab-core';
+
+export const signInGateCopyOpt: ABTest = {
+	id: 'SignInGateCopyOpt',
+	start: '2021-02-24',
+	expiry: '2021-12-01',
```
QuarpT, line 6: the expiry could be set at a couple months away, rather than 9 😄

src/web/components/SignInGate/gateDesigns/copy-opt-test/SignInGateCopyOptVar.tsx
```diff
@@ -0,0 +1,114 @@
...
+		<div
+			className={signInGateContainer}
+			data-cy={`sign-in-gate-${abTest?.variant}`}
+		>
+			<style>{hideElementsCss}</style>
+			<div className={firstParagraphOverlay(!!isComment)} />
+			<h1 className={headingStyles}>{signInGateCopy?.header}</h1>
+			<p className={bodyBold}>{signInGateCopy?.subHeader}</p>
+			{signInGateCopy?.paragraphs.map((paragraph) => {
+				return <p className={bodyText}>{paragraph}</p>;
```
QuarpT, line 42: Looking at `SignInGateMain.tsx ` the text `privacy settings` is a button.  Did you mean to change this in the copy test?

src/web/components/SignInGate/displayRule.ts
rebecca-thompson: It's slightly different as the main variant `canShow` does not pass in a count for the `hasUserDismissedGate` function. It could be that it's an error on the main variant.

src/web/experiments/tests/sign-in-gate-copy-opt.ts
rebecca-thompson: I assumed these would be updated once we knew when the test would run

coldlink: We usually just keep a long date, and manually turn it off/remove the test once we're done. Which is why I usually set a long date.

src/web/components/SignInGate/displayRule.ts
coldlink: Ah yeah, thats a mistake on the main variant. It should be the same as this, so if they've dismissed more than 5 times we stop showing the gate

src/web/components/SignInGate/gateDesigns/copy-opt-test/SignInGateCopyOptVar.tsx
rebecca-thompson: I have added a fix for this if you wouldn't mind re-reviewing?

src/web/components/SignInGate/displayRule.ts
rebecca-thompson: I have refactored the code so that the `canShow` function is only defined once

QuarpT (approved):

src/web/experiments/tests/sign-in-gate-copy-opt.ts
QuarpT: > We usually just keep a long date, and manually turn it off/remove the test once we're done. Which is why I usually set a long date.

I think the expiry parameter exists to remind developers to clean up / review tests that might have been forgotten. So it's probably best to set something realistic.

PR #2628, opened 2021-03-01
BootReact

oliverlloyd:
## What does this change?
Renames `HydrateApp` to `BootReact`

## Why?
Because it isn't hydrating anything anymore


liywjl (approved):
👍

PR #2629, opened 2021-03-01
Accessible Donuts

oliverlloyd:
## What?
Adds the `isLight` function to the `Donut` component so that we can use it to decide what colour to set the text relative to the contrast of the background colour

## Why?
This was raised as an issue where if the background is white and the text is white, you see nothing but we also want to make changes like this to improve accessibility

## Before
Screenshot 2021-03-01 at 12 21 07

## After
Screenshot 2021-03-01 at 12 07 32


liywjl (approved):
👍

rcrphillips (approved):

PR #2632, opened 2021-03-02
We are hiring message for DCR

shtukas:
Add the [We are hiring] message to DCR (thereby targeting all "modern" pages), with an updated ascii art :)


shtukas:
Q: Do you plan to retroactively update the "old" frontend version then ?

A: Maybe.

liywjl (approved):
👍

PR #2635, opened 2021-03-02
Fix Storybook build

gtrufitt:
## What does this change?

The storybook build was bugging out due to an (undiagnosed) problem with awesome-loader. This PR:

- Switches to the same webpack config pattern as the main bundle (ts-loader & babel-loader rather than typescript-awesome-loader)
- Moves the Webpack config to main.js as the docs suggest
- Adds prettier to the files and related file update
- Imports Ophan types as types to supress warnings 

The result being that Storybook and Hot Module Reloading works as expected.


.storybook/preview.js
```diff
@@ -18,8 +18,8 @@ Lazy.disabled = isChromatic();
 Picture.disableLazyLoading = isChromatic();
```
gtrufitt, line 18: This whole file is prettier, ignore.

.storybook/main.js
```diff
@@ -1,4 +1,71 @@
+const path = require('path');
+
 module.exports = {
-  "stories": ["../src/**/*.stories.@(tsx)"],
-   "addons": ["@storybook/addon-essentials"]
-}
+	stories: ['../src/**/*.stories.@(tsx)'],
+	addons: ['@storybook/addon-essentials'],
+	webpackFinal: async (config) => {
```
gtrufitt, line 6: As recommended by https://storybook.js.org/docs/react/configure/webpack moved from webpack to main.js

.storybook/main.js
```diff
@@ -1,4 +1,71 @@
...
+							],
+						],
+					},
+				},
+				{
+					loader: 'ts-loader',
+					options: {
+						configFile: 'tsconfig.build.json',
+						transpileOnly: true,
+					},
```
gtrufitt, line 46: Bring these in from our babel loader as a replacement for awesome-typescript-loader

src/web/browser/ophan/ophan.ts
```diff
@@ -1,4 +1,4 @@
-import {
+import type {
```
gtrufitt, line 1: This was causing warnings, as it is `export type` in guardian/types, importing as a `type` to removes the warnings.

oliverlloyd:
Just a random thought. I worked on a project once where I was able to import the webpack config into SB and use it that way. Not sure if that's a) possible or even b) useful?

gtrufitt:
> Just a random thought. I worked on a project once where I was able to import the webpack config into SB and use it that way. Not sure if that's a) possible or even b) useful?

Yeah, I thought similar. Will give it a try.

oliverlloyd:
> we have to split them out into different objects and combine them together differently

Ah so yes, this was the situation I had in mind that would make this a bad idea

oliverlloyd (approved):

liywjl (approved):
👍

mxdvl:
The dependency wasn’t removed from the `package.json`!

PR #170, opened 2018-09-17
Use correct fonts

SiAdcock:
## What does this change?

Use the correct fonts in dotcom-rendering. I have updated the names to those in the `fonts.css`  files constructed by our font designers, to ensure future compatibility.

This change does not introduce any clever font loading strategy.

sep-17-2018 16-51-46

## Why?

Keeps us up to date.

PR #2636, opened 2021-03-03
Add Text Egyptian and Text Sans Bold Italic fonts

paperboyo:
## What does this change?
This PR adds Bold Italic styles for both Text Egyptian and Text Sans font families. I believe the fonts are not requested if they are not used on the page, so this should not incur extra downloads in most circumstances. Having them for Text Egyptian is also for frontend parity.

This is a proposal, please shot it down if there are performance implications I haven’t thought of. OTOH, I won’t go down without a good fight ;-)
cc @sndrs 

### Before
Firefox, [URL](https://www.theguardian.com/politics/2020/sep/03/margaret-thatcher-wasnt-for-turning-but-boris-johnson-is):
image
Chrome:
image

### After (current frontend; `dcr=false`)
Firefox:
image
Chrome:
image

Text Sans Bold Italic is much less common, mostly in comments. Take a look at [this example](https://www.theguardian.com/commentisfree/2021/mar/02/healthcare-professionals-uk-moral-duty-covid-jab-vaccine#comment-147811921). No screenshots, but the discrepancy and odd rendering is even worse. Also, unrelatedly, note how we may have a bug in discussion-rendering (?) because only the first word (_ask_) is italicised in Firefox versus Chrome).

## Why?
Because faux styles are evil and naff. If we have content that calls for a font style, we must have a font style available for it. Also, because different browsers render them differently (Firefox, shown above, worse than eg. Chrome). Also, because they are used sparingly, so will be requested sparingly. And then they will stay in cache (if one believes in browser cache, that is).

### DIFF
--- a/src/lib/fonts-css.ts
+++ b/src/lib/fonts-css.ts
@@ -198,6 +198,28 @@ const fontList: FontDisplay[] = [
 		weight: 700,
 		style: 'normal',
 	},
+	{
+		family: 'GuardianTextEgyptian',
+		woff2:
+			'fonts/guardian-textegyptian/noalts-not-hinted/GuardianTextEgyptian-BoldItalic.woff2',
+		woff:
+			'fonts/guardian-textegyptian/latin1-not-hinted/GuardianTextEgyptian-BoldItalic.woff',
+		ttf:
+			'fonts/guardian-textegyptian/latin1-not-hinted/GuardianTextEgyptian-BoldItalic.ttf',
+		weight: 700,
+		style: 'italic',
+	},
+	{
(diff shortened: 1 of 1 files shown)


oliverlloyd (approved):
Some useful history on adding fonts to DCR [here](https://github.com/guardian/dotcom-rendering/pull/229) and [here](https://github.com/guardian/dotcom-rendering/pull/170), if you're into that particular kind of DCarchiology.

But from looking at how this text displays in DCR vs. Frontend today it is clear that this PR is a big improvement, this is a much nicer bold.

sndrs:
yay @paperboyo! first dcr pr?

PR #2645, opened 2021-03-03
Log Braze epic impression when in viewport

tjmw:
## What does this change?

We only want to record the in-app message impression with Braze when it is visible in the viewport. This uses the same configuration as the reader revenue epic impression tracking.

Note: Ophan tracking to come in a future PR.

### Before

The impression was recorded when the epic component was successfully imported.

### After

The impression is recorded when the epic is rendered and in the viewport.


adamnfish (approved):
Looks great.

PR #2646, opened 2021-03-04
Complete ConfigType

oliverlloyd:
## What does this change?
Adds all the missing types to `ConfigType`

I also refactored the way fixtures are built to make it clear that the config and switches objects we use to override values are indeed _overrides_.

## Why?
We weren't using these properties so we never typed them but this data is still sent to us and it throws errors in the fixture files when we try to type the data. To get around this we delete these properties but that's a poor solution so I've corrected the types here


oliverlloyd:
[Refactor fixtures](https://trello.com/c/ag4isw4x/2675-refactor-fixtures)

liywjl (approved):
👍 LGTM

PR #2647, opened 2021-03-04
Add capability to cache Braze messages in local storage

tjmw:
## What does this change?

Adds capability to cache Braze messages. The cache implementation used by `BrazeMessages` is configurable. This PR uses an in memory cache (`InMemoryCache`) so no data will persist across pageviews. We need to ensure that we're clearing the cache on logout and consent change on DCR and frontend before switching to the real localStorage backed cache (`LocalMessageCache`). This PR includes clearing data DCR, but we still need to make the change for frontend.

## Why?

The delivery behaviour of Braze messages to the browser don't exactly fit our use cases. Once Braze has given us a message via the `subscribeToInAppMessage` callback they consider it delivered, and it won't be delivered again. But there are many reasons why we might not be able to show a message in that same pageview. For example: banner timeout limit is reached, a higher priority message is shown instead, the page isn't suitable (e.g. paid content) or doesn't have the required slot (e.g. an epic message is delivered on a page without that slot).

## Notes

The TTL for messages in the cache is 1 day currently (though that's not so relevant for this PR which uses the in memory cache). We limit the cache size to two messages per slot. If a message is pushed to the queue for a slot which already has 2 messages it will be discarded. Given our current usage of Braze this should be plenty, and means we won't take up excessive local storage space.

### DIFF
--- a/src/web/lib/braze/LocalMessageCache.ts
+++ b/src/web/lib/braze/LocalMessageCache.ts
@@ -1,0 +1,241 @@
+/* eslint-disable class-methods-use-this */
+/* eslint-disable max-classes-per-file */
+
+import type appboy from '@braze/web-sdk-core';
+import { storage } from '@guardian/libs';
+import { SlotName, SlotNames } from './types';
+
+const localStorageKeyBase = 'gu.brazeMessageCache';
+export const millisecondsBeforeExpiry = 1000 * 60 * 60 * 24; // 24 hours: 60 seconds * 60 minutes
+
+type Message = appboy.InAppMessage;
+
+type MessageWithId = {
+	id: string;
+	message: Message;
+};
+
+type CachedMessage = {
+	message: {
+		id: string;
+		message: Record<string, any>; // From the serialized JSON
+	};
+	expires: number; // Expiry date in Unix time
+};
+
+const MAX_QUEUE_SIZE = 2;
+
+const keyFromSlotName = (slotName: SlotName): string =>
+	`${localStorageKeyBase}.${slotName}`;
+
+const hasNotExpired = (cachedMessage: CachedMessage) =>
+	cachedMessage.expires > Date.now();
+
+// setQueue is effectively private, but it's useful to expose it publicly
+// so that we can use it in the tests
+const setQueue = (slotName: SlotName, queue: CachedMessage[]) => {
+	const key = keyFromSlotName(slotName);
+	storage.local.set(key, queue);
+};
+
+const readQueue = (slotName: SlotName): CachedMessage[] => {
+	const key = keyFromSlotName(slotName);
+	const queue = storage.local.get(key);
+
+	if (Array.isArray(queue)) {
+		return queue;
+	}
+
+	return [];
+};
+
+const hydrateMessage = (
+	messageData: Record<string, any>,
+	appboyInstance: typeof appboy,
+): appboy.InAppMessage => {
+	/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+	const hydratedMessage = new appboyInstance.HtmlMessage(
+		messageData.message as string,
+		messageData.extras,
+		messageData.campaignId,
+		messageData.cardId,
+		messageData.triggerId,
+		messageData.dismissType,
+		messageData.duration,
+		messageData.animateIn,
+		messageData.animateOut,
+		messageData.frameColor,
+		messageData.htmlId,
+		messageData.css,
+		messageData.messageFields,
+	);
+	/* eslint-enable @typescript-eslint/no-unsafe-member-access */
+
+	return hydratedMessage;
+};
+
+const isValid = (m: CachedMessage): boolean => {
+	return (
+		m?.expires &&
+		Number.isFinite(m?.expires) &&
+		m?.message?.id &&
+		m?.message?.message?.triggerId &&
+		m?.message?.message?.extras
+	);
+};
+
+const getQueue = (slotName: SlotName): CachedMessage[] => {
+	const queue = readQueue(slotName);
+	const validQueue = queue.filter((i) => isValid(i));
+	const unexpiredQueue = validQueue.filter((i) => hasNotExpired(i));
+
+	if (queue.length !== unexpiredQueue.length) {
+		setQueue(slotName, unexpiredQueue);
+	}
+
+	return unexpiredQueue;
+};
+
+interface MessageCache {
+	peek: (
+		slotName: SlotName,
+		appboyInstance: typeof appboy,
+	) => MessageWithId | undefined;
+	remove: (slotName: SlotName, id: string) => boolean;
+	push: (slotName: SlotName, message: MessageWithId) => boolean;
+	clear: () => void;
+}
+
+class LocalMessageCache {
+	static peek(
+		slotName: SlotName,
+		appboyInstance: typeof appboy,
+	): MessageWithId | undefined {
+		const queue = getQueue(slotName);
+		const topItem = queue[0];
+
+		if (topItem) {
+			return {
+				id: topItem.message.id,
+				message: hydrateMessage(
+					topItem.message.message,
+					appboyInstance,
+				),
+			};
+		}
+	}
+
+	static remove(slotName: SlotName, id: string): boolean {
+		const queue = getQueue(slotName);
+		const idx = queue.findIndex((i) => i.message.id === id);
+
+		if (idx >= 0) {
+			const removedItem = queue.splice(idx, 1);
+
+			if (removedItem) {
+				setQueue(slotName, queue);
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	static push(slotName: SlotName, message: MessageWithId): boolean {
+		const queue = getQueue(slotName);
+
+		if (queue.length < MAX_QUEUE_SIZE) {
+			const expires = Date.now() + millisecondsBeforeExpiry;
+
+			const messageToCache: CachedMessage = {
+				message,
+				expires,
+			};
+
+			queue.push(messageToCache);
+
+			setQueue(slotName, queue);
+			return true;
+		}
+
+		return false;
+	}
+
+	static clear() {
+		// eslint-disable-next-line guard-for-in
+		for (const slotName in SlotNames) {
+			const key = keyFromSlotName(slotName as SlotName);
+			storage.local.remove(key);
+		}
+	}
+}
+
+type InMemoryCachedMessage = {
+	message: MessageWithId;
+	expires: number; // Expiry date in Unix time
+};
+
+const inMemoryQueue: Record<SlotName, InMemoryCachedMessage[]> = {
+	EndOfArticle: [],
+	Banner: [],
+};
+
+// Until we're ready to turn caching on we can use this as a swap in replacement
+// for LocalMessageCache.
+class InMemoryCache {
+	static peek(slotName: SlotName): MessageWithId | undefined {
+		const unexpiredMessages = inMemoryQueue[slotName].filter((i) =>
+			hasNotExpired(i),
+		);
+		return unexpiredMessages[0]?.message;
+	}
+
+	static remove(slotName: SlotName, id: string): boolean {
+		const idx = inMemoryQueue[slotName].findIndex(
+			(i) => i.message.id === id,
+		);
+
+		if (idx >= 0) {
+			const removedItem = inMemoryQueue[slotName].splice(idx, 1);
+
+			if (removedItem) {
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	static push(slotName: SlotName, message: MessageWithId): boolean {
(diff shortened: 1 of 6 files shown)


src/web/lib/braze/LocalMessageCache.ts
```diff
@@ -0,0 +1,76 @@
+import type appboy from '@braze/web-sdk-core';
+import { SlotName } from './BrazeMessages';
+
+const localStorageKeyBase = 'gu.brazeMessageCache';
+
+type Message = appboy.InAppMessage;
+
+interface Store {
```
adamnfish, line 8: Give that this isn't specialised to our own types, is it true that this LocalStorage interface doesn't already exist somewhere in DCR or the browser platform?

src/web/lib/braze/LocalMessageCache.ts
```diff
@@ -0,0 +1,76 @@
...
+		const queue = this.readQueue(key);
+		const topItem = queue.shift();
+
+		if (topItem) {
+			this.store.setItem(key, JSON.stringify(queue));
+			return topItem;
+		}
+	}
+
+	peek(slotName: SlotName): Message | undefined {
```
adamnfish, line 35: What's the motivation for `peek` returning the actual message? It seems to me that this interface makes it easy for clients to misuse this interface.

Peek allows client code to get the entire message and display it, without removing the message from the store. If peek instead only said that a message exists (`bool` - all ew technically need), the number of messages (`number` - probably the best balance of usability and strictness) or perhaps even just a summary of the messages (`array<string>` - I'm not a fan of this), then we'd be addressing the use case of "is there a message for me" while strictly enforcing the correct usage of the cache.

src/web/lib/braze/LocalMessageCache.ts
```diff
@@ -0,0 +1,76 @@
...
+type Message = appboy.InAppMessage;
+
+interface Store {
+	getItem: (key: string) => string | null;
+	setItem: (key: string, value: string) => void;
+	removeItem: (key: string) => void;
+}
+
+const keyFromSlotName = (slotName: SlotName): string =>
+	`${localStorageKeyBase}.${slotName}`;
```
adamnfish, line 15: Would this be simpler as a single queue of braze messages where each stores the slot, rather than one queue per slot? The interface would be identical (we could filter on the slot) but the storage would map more closely to how Braze sees these messages.

I'm very happy with it as-is, just wanted to check what you think. If we ever want to support a Braze message that is flexible in which slot it uses we'd be much better off with a single queue, but as it stands I think it's 6 and a half-dozen. This slot flexibility is purely hypothetical for now and I think it's unlikely we'll do that, so the current choice looks good.

src/web/lib/braze/LocalMessageCache.ts
```diff
@@ -0,0 +1,76 @@
...
+	push(slotName: SlotName, message: Message) {
+		const key = keyFromSlotName(slotName);
+		const queue = this.readQueue(key);
+
+		queue.push(message);
+		this.store.setItem(key, JSON.stringify(queue));
+	}
+
+	clear() {
+		const allSlots: SlotName[] = ['Banner', 'EndOfArticle'];
```
adamnfish, line 54: Does this line have exhaustivity checking from the type-system? As it stands, I think it would be easy for us to add a new `SlotName` in the file above, but forget to include it here. This would mean we're failing to clear Braze message data for one of the queues. This might be a reason to favour a single queue, as mentioned in a previous comment.

src/web/lib/braze/LocalMessageCache.ts
```diff
@@ -0,0 +1,76 @@
...
+interface Store {
+	getItem: (key: string) => string | null;
+	setItem: (key: string, value: string) => void;
+	removeItem: (key: string) => void;
+}
+
+const keyFromSlotName = (slotName: SlotName): string =>
+	`${localStorageKeyBase}.${slotName}`;
+
+class LocalMessageCache {
```
adamnfish, line 17: I wonder if this class is missing a way to clear out expired messages? We'll want to consider the time a message should be held on for, but unlike cookies LocalStorage does not have an in-built expiry mechanism.

Should we make sure that the message's date is part of the stored information, and include a method for deleting expired messages? That method doesn't need to necessarily be "public", it can be an internal implementation detail to keep the queue containing only relevant messages. In particular, call to `peek` and `shift` should never return expired messages so that client code can safely use the cache without the idea of expiry leaking out into the rest of the codebase.

src/web/lib/braze/LocalMessageCache.ts
tjmw: Good point, I'll check.

src/web/lib/braze/LocalMessageCache.ts
tjmw: Really good point, I agree that would be much better. In that case this could become `hasMessage(slotName: SlotName): boolean` which seems nicer. I can't remember if I did things this way because it was necessary because of how the rest of the message selection stuff is wired up, I'll investigate.

src/web/lib/braze/LocalMessageCache.ts
tjmw: To me keeping things separate felt simpler, since at the moment the slots can be blissfully ignorant of what's going on with the other. It's true, I can imagine a future world where a message is agnostic about it's slot, or at least is flexible about where it goes, but that feels like a long way off at the moment.

src/web/lib/braze/LocalMessageCache.ts
tjmw: No, but you're right that it should. I'm pretty sure this is possibly with enum types. I'll investigate.

src/web/lib/braze/LocalMessageCache.ts
tjmw: Yep I'm on the same page with all this, just didn't want to cram too much into this first PR.

tjmw: Having thought about this a little more, the expiration is such a fundamental part of the cache that I don't think it makes sense to consider it without it, even as a first step. So I'll work on adding that functionality to this PR.

src/web/lib/braze/LocalMessageCache.ts
tjmw: There's a global `Storage` TS interface we can use, so I've switched to that.

src/web/lib/braze/LocalMessageCache.ts
rhystmills: Would this complicate e.g. the Epic where we don't want to remove message from the queue until it's actually been seen? Do we want to decouple using a message from deciding it has been displayed - otherwise we will have to handle re-adding an undisplayed Epic to the front of the queue.

tjmw: Thanks @rhystmills this is a really good point. Yeah for the epic we probably want to render it into the DOM but not remove it from the queue until it's had an impression. The alternative would be to pop the message off and then push back on if it doesn't get seen, but it's tricky to know when that would happen (when the page gets unloaded?)

src/web/lib/braze/LocalMessageCache.ts
rhystmills: I started added expiry this afternoon but haven't finished the tests yet - I expect some checks may fail in the mean time.

src/web/lib/braze/LocalMessageCache.ts
tjmw: I've made this change, using a TypeScript enum. There are a couple of annoying TS things to get it working so it's not as clean as I'd like, but it gets the job done.

tjmw (commented):
Adding some feedback (on my own PR 😂) based on IRL conversation with @adamnfish and @rhystmills.

src/web/lib/braze/LocalMessageCache.ts
```diff
@@ -0,0 +1,230 @@
...
+	push: (slotName: SlotName, message: MessageWithId) => boolean;
+	clear: () => void;
+}
+
+const hydrateMessage = (
+	messageData: Record<string, any>,
+	appboyInstance: typeof appboy,
+): appboy.InAppMessage => {
+	/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+	const hydratedMessage = new appboyInstance.HtmlMessage(
```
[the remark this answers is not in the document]
tjmw: We're now validating the data above.

src/web/lib/braze/LocalMessageCache.ts
tjmw: This is now done.

src/web/lib/braze/buildBrazeMessages.ts
```diff
@@ -66,7 +77,7 @@ export const buildBrazeMessages = async (
 			value: sdkLoadTimeTaken,
 		});
 
-		const brazeMessages = new BrazeMessages(appboy);
+		const brazeMessages = new BrazeMessages(appboy, InMemoryCache);
```
tjmw, line 80: This line defines that we're using the in memory cache.

oliverlloyd (commented):
Do you think it would be possible to use something like [React Query](https://react-query.tanstack.com/overview) here instead? 

I've already been considering adopting this library for other reasons, we're starting to add more features to `useApi` like custom fetch options or`pollInterval` values, so a battle-tested, off the shelve solution starts to make a lot of sense.

Once (and if) such a lib were available in DCR, maybe it could be incorporated here?

oliverlloyd:
Answering my own question a little bit. At least as far as I understand the requirement which I admit is not that far!

So, yes, you probably could use [useQuery](https://react-query.tanstack.com/guides/queries) here, however, persistance is 'experimental'.

In React Query land a query is
> a declarative dependency on an asynchronous source of data that is tied to a unique key

This fact that React Query will work with _any_ promise based source of data, not just fetch calls, means you can cache anything. So long as you give it a unique id then any other calls to that same id will first read from the cache, based on the configuration options

But yeah, persistence is alpha, which does sound bad but I don't think we should discount it. If it were tested and shown to work then we can lock our version to guard against breaking patch changes

oliverlloyd:
This discussion is perhaps a little outside of this PR but I wrote up a [mini proposal](https://docs.google.com/document/d/1F7PdM7jxzuuD4s56ceSd1oi4oaw0zxPdeJ2PDDwJG9w/edit#) to bring React Query into DCR.

tjmw:
Hey @oliverlloyd thanks for the suggestion 👍  I'm not too familiar with React Query so I'll have a read. One sticking point _might_ be that we'd also like to have this functionality on frontend (next steps are to extract this out to an NPM package and use that in both places).

jfsoul:
Hi @tjmw - from a platform point of view, we don't want to block your work but we've also been having (somewhat unresolved) conversations about consolidating around a library which might solve some of these problems. What we have so far is our own API that is not too far away from something like React Query, so the hope would be that a migration wouldn't be too difficult in the future.

This PR goes further than what the platform provides at the moment, and wouldn't align neatly with the API of something like React Query, so migrating it would likely be more work (but something that would absolutely be sensible to do, if the platform was already consolidated on a library which gave us persistence). 

In order to unblock this I think our suggestion would be that you should go ahead, but it's an area than TX would need to be responsible for maintaining, we probably should discourage particularly widespread use, and there'd be some work to migrate once we have a consistent approach / best practice suggestion. If you're happy with that then 👍

tjmw:
Hey @jfsoul thanks for this. We'll proceed with this approach with the expectation that once the platform provided API is ready we'll integrate with it. I think it's getting a little late to deploy on a Friday but we'll go ahead and merge this early next week.

Having read a bit more about React Query I think it could play really nicely with the work in this PR rather than replace it. I think React Query provides a nice solution to combining server state with React components. But I think there continues to be a place for plain-old-JS classes like `BrazeMessages` which provide a promise based API (and so will play nicely with React Query) but are agnostic about the framework they're used with. One of the next steps for this is that we're planning on extracting some of this out to an NPM package so we can share between DCR and frontend, so in order to be able to use in both we'd want to not be tied to React too closely.

I'd be really happy to use some of the work TX have been doing as an integration test or POC for the new React Query like API you're working on. Exciting times!

nicl (commented):
Given this is experimental, happy to inline the code in Braze files and see how it goes.

The broader question I have is: does Braze expose a (Rest or other) API to do this rather than requiring the SDK?

I vaguely remember in the past that they didn't. But perhaps this has changed. If not, I think we should pressure them to introduce this as this SDK approach has quite a few downsides:

* package size is very large (34.3kb minified+gzipped) and may increase in future
* it is difficult to arbitrate the slot with other concerns without lots of inline logic that is often poorly documented and not so visible to product/other stakeholders)
* generally requires shifting this logic to client when it is normally better to do stuff server-side for maintenance and performance reasons

(I'm rehashing the Automat arguments here but the concerns are general.)

nicl (approved):

PR #2648, opened 2021-03-04
Removed all fixture overrides

oliverlloyd:
## What?
This deletes all config and switch overrides.

I found that if I removed all the overrides, our tests still pass, so I have reset us to a baseline of zero global overrides.

## Before
```javascript
module.exports = {
	configOverrides: {
		ajaxUrl: 'https://api.nextgen.guardianapps.co.uk',
                  // …
                  // … Plus LOADS of other stuff
         }
};
```

## After
```javascript
module.exports = {
	configOverrides: {}, // Empty!
};
```

## Why?
Because we want to be testing as close to live data as possible and now that we’re using a merge pattern we can be selective. And it turns out we currently don't need any overrides

### What about the reasons we introduced these overrides?
Switch overrides were first [introduced here](https://github.com/guardian/dotcom-rendering/pull/1876) to address a [specific problem](https://github.com/guardian/dotcom-rendering/pull/1861) where the `consentManagement` property wasn’t present in the fixtures. This issue is now addressed by the move to automatically regenerating the fixtures via `make gen-fixtures` so the original use case for overriding switches is no longer present.

Config overrides were added to get around the problem where we were only partially typing the config data. This was fixed in #2646 so that use case has also been removed.

### What if I want to override something for a test?
Please don’t use a global override, do it locally using a spread operator. We should not be changing the test data for _all_ tests just for one use case (unless we really need to do that in which case we should be clear about that.)


oliverlloyd:
@sndrs This PR overwrites some of [your earlier work](https://github.com/guardian/dotcom-rendering/pull/1876). I checked and I think this is fine but please do check to make sure I'm not missing something!

nitro-marky (approved):
Looks good

sndrs (approved):
lgtm – much nicer

oliverlloyd:
I made a change here to how we mock data in our stories because after updating some of the fixtures, the type of onwards content needed changed and we weren't mocking it. I updated Storybook's config to make use of the same fetch mocks as we have elsewhere (instead of having it's own list)

PR #2649, opened 2021-03-04
Drop box shadow on sticky navs

nicl:
cc @HarryFischer 

### Before
Screenshot 2021-03-04 at 12 43 49

### After
Screenshot 2021-03-04 at 12 43 05

## Why?

The drop-shadow was an improvisation shall we say that has since been corrected by design.


HarryFischer:
Looks good! Thanks

rcrphillips (approved):

PR #2650, opened 2021-03-04
Change how Vitals are collected

nitro-marky:
## What does this change?

The core vitals are now collected directly as they are provided.

### DIFF
--- a/src/web/browser/coreVitals/coreVitals.ts
+++ b/src/web/browser/coreVitals/coreVitals.ts
@@ -36,19 +36,19 @@ export const coreVitals = (): void => {
 	const addToJson = ({ name, value }: CoreVitalsArgs): void => {
 		switch (name) {
 			case 'FCP':
-				jsonData.fcp = Math.round(value * 1000000) / 1000000;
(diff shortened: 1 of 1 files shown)


gtrufitt (approved):
Makes sense, the reason for the previous method in GA is:

https://github.com/GoogleChrome/web-vitals#send-the-results-to-google-analytics

```js
 // Google Analytics metrics must be integers, so the value is rounded.
    // For CLS the value is first multiplied by 1000 for greater precision
    // (note: increase the multiplier for greater precision if needed).
    eventValue: Math.round(name === 'CLS' ? delta * 1000 : delta),
    ```

So without this restriction, we should send the data pure from the browser.

nitro-marky:
Yeah, for some reason I thought I should normalise all data!

jfsoul:
The reason we had the `Math.round(x * 1000000) / 1000000` is to ensure we don't try sending too high a precision number to big query - which is only set up to accept 9 decimal places in these columns (we got errors without this). In terms of scaling the numbers - we're not changing anything so we reverted this.

PR #2652, opened 2021-03-04
Go live on 1% initially for sticky-nav

nicl:
Initial go-live of this test. See the original Trello card for wider context: https://trello.com/c/fy5HsbwD/2347-sticky-nav-test


nitro-marky (approved):
👍

PR #2657, opened 2021-03-05
ARM/Graviton testing

nicl:
**Note: plan is to run this for a few days (assuming no issues) and then switch back to x86 for the rest of the month.**

## What?

The changes necessary to test Graviton for a few days:

* Amigo AMI changes (already done)
* cloudformation updates to instance type and AMI (via riffraff.yaml)
* disable `disksage` module (as breaks on ARM) 

Note the latter means we won't get diskusage stats for the duration of the test. I consider this a reasonable risk, as we rarely check these metrics and still have regular autoscaling, CPU, latency, errors etc. If the test is successful we can think of an alternative disk-monitoring solution that is ARM friendly.

In preliminary testing we see:

1. no change in CPU usage
Screenshot 2021-03-05 at 14 33 56
(light blue is the new ARM instance - you can see an initial CPU spike before settling at ~20% which is the same as our x86 instances)

2. an improvement in latency
Screenshot 2021-03-05 at 14 43 46
(note the dip on the right side of around 20% from 40ms to just over 30ms average around the 14:40 mark)

Screenshot 2021-03-05 at 15 20 27
(percentiles are always nicer :) - here we can see the gain is especially at the higher percentiles)

## Why?

To save $$$ (ARM instances are 20% cheaper) and possibly improve performance.

https://trello.com/c/bs16BYKE/2466-aws-reservations

### DIFF
--- a/scripts/deploy/riff-raff.yaml
+++ b/scripts/deploy/riff-raff.yaml
@@ -9,12 +9,12 @@ deployments:
       cloudFormationStackName: rendering
       templateStageParameters:
         CODE:
-          InstanceType: t3.micro
+          InstanceType: t4g.micro
         PROD:
-          InstanceType: t3.small
+          InstanceType: t4g.small
       amiParametersToTags:
         AMI:
-          Recipe: dotcom-rendering-node14-test
+          Recipe: dotcom-rendering-ARM
           BuiltBy: amigo
           AmigoStage: PROD
   rendering:
--- a/src/app/aws/metrics-baseline.ts
+++ b/src/app/aws/metrics-baseline.ts
@@ -1,4 +1,6 @@
-import os from 'os';
+// TODO re-enable after Graviton testing complete (the diskusage module doesn't work on ARM).
+
+/* import os from 'os';
 import disk from 'diskusage';
 import { BytesMetric, collectAndSendAWSMetrics } from './aws-metrics';
 
@@ -43,4 +45,4 @@ export const recordBaselineCloudWatchMetrics = () => {
 			}
 		}
 	});
-};
+}; */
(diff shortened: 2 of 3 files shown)


scripts/deploy/riff-raff.yaml
```diff
@@ -7,9 +7,14 @@ deployments:
...
       cloudFormationStackName: rendering
+      templateStageParameters:
+        CODE:
+          InstanceType: t4g.micro
+        PROD:
+          InstanceType: t4g.small
       amiParametersToTags:
         AMI:
-          Recipe: dotcom-rendering-node14-test
+          Recipe: dotcom-rendering-ARM
```
jfsoul, line 17: this saves us a job of switching back to the non-"testing" AMI 😆

src/app/aws/metrics-baseline.ts
```diff
@@ -43,4 +45,4 @@ export const recordBaselineCloudWatchMetrics = () => {
 			}
 		}
 	});
-};
+}; */
```
jfsoul, line 48: is this necessary if we're commenting out the import of this file?

src/app/server.ts
```diff
@@ -19,7 +19,7 @@ import {
 	getGuardianConfiguration,
 	GuardianConfiguration,
 } from './aws/aws-parameters';
-import { recordBaselineCloudWatchMetrics } from './aws/metrics-baseline';
+// import { recordBaselineCloudWatchMetrics } from './aws/metrics-baseline';
```
jfsoul, line 22: would be nice to keep the memory stats if we can at least

jfsoul (approved):

scripts/deploy/riff-raff.yaml
nicl: yep exactly

src/app/aws/metrics-baseline.ts
nicl: I've just commented out the whole file to be honest for clarity. There are some calls midway down too that fire off some scheduled jobs which this disables (in addition to the server.ts initialisation). Not actually sure why we have both so need to investigate that.

PR #2661, opened 2021-03-05
Identity | Sign In Gate | Mandatory Sign In Gate Test

QuarpT:
## What does this change?

Add new signin gate. We are testing the affects of a mandatory signin gate. This is a gate asking users to login for articles without a dismiss button.

Additional requirements are:
- UK only 
- Only if accepted specific CMP consents

## Why?

We are experimenting with ways of making the signin gate more effective without having a negative impact of other parts of the site.

image

### DIFF
--- a/src/web/components/SignInGate/SignInGateSelector.tsx
+++ b/src/web/components/SignInGate/SignInGateSelector.tsx
@@ -20,7 +20,13 @@ import { signInGateMainControl } from '@root/src/web/experiments/tests/sign-in-g
 // Sign in Gate Types
 import { signInGateComponent as gateMainVariant } from '@root/src/web/components/SignInGate/gates/main-variant';
 import { signInGateComponent as gateMainControl } from '@root/src/web/components/SignInGate/gates/main-control';
+import { signInGateComponent as gateMandatoryVariant } from '@root/src/web/components/SignInGate/gates/mandatory-gate';
 
+import { signInGateMandoryTest } from '@root/src/web/experiments/tests/sign-in-gate-mandatory';
+import {
+	canShow,
+	canShowMandatoryGate,
+} from '@root/src/web/components/SignInGate/displayRule';
 import {
 	ComponentEventParams,
 	submitViewEventTracking,
@@ -65,16 +71,20 @@ type GateTestMap = { [name: string]: SignInGateComponent };
 const tests: ReadonlyArray<ABTest> = [
 	signInGateMainVariant,
 	signInGateMainControl,
+	signInGateMandoryTest,
 ];
 
 const testVariantToGateMapping: GateTestMap = {
 	'main-control-3': gateMainControl,
-	'main-variant-3': gateMainVariant,
+	'main-variant-3': gateMainVariant(canShow),
+	'mandatory-gate-control': gateMainVariant(canShowMandatoryGate),
+	'mandatory-gate-variant': gateMandatoryVariant,
 };
 
 const testIdToComponentId: { [key: string]: string } = {
 	SignInGateMainVariant: 'main_variant_3',
 	SignInGateMainControl: 'main_control_3',
+	SignInGateMandatory: 'mandatory_gate_test',
 };
 
 // function to generate the profile.theguardian.com url with tracking params
@@ -150,7 +160,9 @@ export const SignInGateSelector = ({
 	isSignedIn,
 	CAPI,
 }: SignInGateSelectorProps) => {
-	const [showGate, setShowGate] = useState(true);
+	const [isGateDismissed, setIsGateDismissed] = useState(false);
+	const [canShowGate, setCanShowGate] = useState(false);
+
 	const [currentTest, setCurrentTest] = useState<CurrentABTest>({
 		name: '',
 		variant: '',
@@ -171,14 +183,14 @@ export const SignInGateSelector = ({
 
 	useEffect(() => {
 		// this hook will fire when the sign in gate is dismissed
-		// which will happen when the showGate state is set to false
+		// which will happen when the isGateDismissed state is set to true
 		// this only happens within the dismissGate method
-		if (showGate === false) {
+		if (isGateDismissed) {
 			document.dispatchEvent(
 				new CustomEvent('dcr:page:article:redisplayed'),
 			);
 		}
-	}, [showGate]);
+	}, [isGateDismissed]);
 
 	// check to see if the test is available on this render cycle
 	// required by the ab test framework, as we have to wait for the above
@@ -192,19 +204,25 @@ export const SignInGateSelector = ({
 	const gateVariant: SignInGateComponent | null =
 		testVariantToGateMapping?.[currentTest.variant];
 
+	if (gateVariant) {
+		// eslint-disable-next-line @typescript-eslint/no-floating-promises
+		gateVariant
+			.canShow(CAPI, !!isSignedIn, currentTest)
+			.then(setCanShowGate);
+	}
+
 	return (
 		<>
 			{/* Sign In Gate Display Logic */}
-			{showGate &&
-				gateVariant?.canShow(CAPI, !!isSignedIn, currentTest) && (
-					<ShowSignInGate
-						CAPI={CAPI}
-						abTest={currentTest}
-						setShowGate={setShowGate}
-						signInUrl={signInUrl}
-						gateVariant={gateVariant}
-					/>
-				)}
+			{!isGateDismissed && canShowGate && (
+				<ShowSignInGate
+					CAPI={CAPI}
+					abTest={currentTest}
+					setShowGate={(show) => setIsGateDismissed(!show)}
+					signInUrl={signInUrl}
+					gateVariant={gateVariant}
+				/>
+			)}
 		</>
 	);
 };
--- a/src/web/components/SignInGate/displayRule.ts
+++ b/src/web/components/SignInGate/displayRule.ts
@@ -6,6 +6,8 @@ import {
 import { getCountryCodeFromLocalStorage } from '@frontend/web/lib/getCountryCode';
 import { CurrentABTest } from '@root/src/web/components/SignInGate/gateDesigns/types';
 import { hasUserDismissedGateMoreThanCount } from '@root/src/web/components/SignInGate/dismissGate';
+import { onConsentChange } from '@root/node_modules/@guardian/consent-management-platform';
+import { ConsentState } from '@root/node_modules/@guardian/consent-management-platform/dist/types';
 
 // in our case if this is the n-numbered article or higher the user has viewed then set the gate
 export const isNPageOrHigherPageView = (n: number = 2): boolean => {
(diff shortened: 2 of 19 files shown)


src/web/experiments/tests/sign-in-gate-mandatory.ts
```diff
@@ -0,0 +1,30 @@
+import { ABTest } from '@guardian/ab-core';
+
+export const signInGateMandoryTest: ABTest = {
+	id: 'SignInGateMandatory',
+	start: '2021-03-04',
+	expiry: '2021-06-04',
+	author: 'Peter Colley',
+	description:
+		'Compare mandatory gate (an article sigin gate without the dismiss button) with the main signin gate',
+	audience: 0.2,
```
QuarpT, line 10: We require a sample size of 170k browsers seeing the variant. We'll need to set the audience appropriately to achieve this.

src/web/components/SignInGate/displayRule.ts
```diff
@@ -79,15 +83,53 @@ export const canShow = (
...
+			isNPageOrHigherPageView(3) &&
+			isValidContentType(CAPI) &&
+			isValidSection(CAPI) &&
+			isValidTag(CAPI) &&
+			!isIOS9(),
+	);
+};
+
+export const hasRequiredConsents = (): Promise<boolean> => {
+	const consentsRequiredFor: ['google-analytics', 'ophan', 'sentry'] = [
```
QuarpT, line 103: we need to confirm whether these are all the consents required for mandatory signin

liywjl (commented):
Overall the code looks good, I have just some minor improvement recommendations.

I am not how to test the PR myself, so maybe adding storybook to these new components could help generate a playgroup to test

src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -73,24 +79,28 @@ const tests: ReadonlyArray<ABTest> = [
...
+	'main-variant-3': gateMainVariant(canShow),
+	'copy-opt-control': gateMainVariant(canShow),
 	'copy-opt-variant-1': gateCopyOptVar1,
 	'copy-opt-variant-2': gateCopyOptVar2,
 	'copy-opt-variant-3': gateCopyOptVar3,
 	'copy-opt-variant-4': gateCopyOptVar4,
 	'copy-opt-variant-5': gateCopyOptVar5,
 	'copy-opt-variant-6': gateCopyOptVar6,
+	'mandatory-gate-control': gateMainVariant(canShowMandatoryGate),
+	'mandatory-gate-variant': gateMandatoryVariant,
```
coldlink, line 96: I'm not sure entirely what's happening here with the `gateMainVariant` method?

src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -208,19 +220,25 @@ export const SignInGateSelector = ({
...
-				)}
+			{!isGateDismissed && canShowGate ? (
+				<ShowSignInGate
+					CAPI={CAPI}
+					abTest={currentTest}
+					setShowGate={(show) => setIsGateDismissed(!show)}
+					signInUrl={signInUrl}
+					gateVariant={gateVariant}
+				/>
+			) : null}
```
coldlink, line 241:
```suggestion
			{!isGateDismissed && canShowGate &&
				<ShowSignInGate
					CAPI={CAPI}
					abTest={currentTest}
					setShowGate={(show) => setIsGateDismissed(!show)}
					signInUrl={signInUrl}
					gateVariant={gateVariant}
				/>
			}
```

should achieve the same thing without having to rely on the ternary operator

src/web/components/SignInGate/displayRule.ts
coldlink: We need to check for "accept all" consents as said in standup.
It would be useful if the CMP returned a list of id's to check, or a type that returns all the id's.
This will get real messy if we have to manually add all the ids.

Maybe one for @guardian/commercial-dev can help us with/advise?

src/web/components/SignInGate/displayRule.ts
```diff
@@ -79,15 +83,53 @@ export const canShow = (
...
+		'sentry',
+	];
+	return new Promise((resolve, reject) => {
+		onConsentChange((state) => {
+			try {
+				resolve(
+					consentsRequiredFor.every((c) => getConsentFor(c, state)),
+				);
+			} catch (e) {
+				reject(e);
```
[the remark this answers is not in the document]
coldlink: In this case, we'd probably want to "reject(false)", if there's any sort of error we want the article to show, anything else, like returning an error may cause unintended side effects

src/web/components/SignInGate/gateDesigns/SignInGateMandatory.tsx
```diff
@@ -0,0 +1,121 @@
...
+	headingStyles,
+	hideElementsCss,
+	privacyLink,
+	registerButton,
+	signInGateContainer,
+	signInHeader,
+	signInLink,
+} from './shared';
+
+export const SignInGateMandatory = ({
```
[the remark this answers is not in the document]
coldlink: Agreed, the gate should be added to: https://github.com/guardian/dotcom-rendering/blob/main/src/web/components/SignInGate/SignInGate.stories.tsx

src/web/experiments/tests/sign-in-gate-mandatory.ts
coldlink: Should match what we decide in https://github.com/guardian/frontend/pull/23587

src/web/components/SignInGate/SignInGateSelector.tsx
QuarpT: canShow is passed as a parameter to mainVariant because we need to have different canShow criteria for different tests

src/web/components/SignInGate/displayRule.ts
```diff
@@ -79,15 +83,53 @@ export const canShow = (
...
+				currentTest.variant,
+				currentTest.name,
+				5,
+			) &&
+			isNPageOrHigherPageView(3) &&
+			isValidContentType(CAPI) &&
+			isValidSection(CAPI) &&
+			isValidTag(CAPI) &&
+			!isIOS9(),
+	);
```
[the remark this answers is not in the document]
QuarpT: I'm using `Promise.resolve` to lift a boolean to a promise. I'm not sure how you're thinking of using `Promise.all` here

src/web/components/SignInGate/displayRule.ts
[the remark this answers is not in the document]
QuarpT: @coldlink are you suggesting `resolve(false)` on error? Using `reject(false)` doesn't help, unless you recover the error (which happens to be false).

@liywjl The Promise type parameter specifies the success type. The error type is always any.
```
    new <T>(executor: (resolve: (value: T | PromiseLike<T>) => void, reject: (reason?: any) => void) => void): Promise<T>;
```

src/web/components/SignInGate/gates/main-control.ts
```diff
@@ -15,14 +15,16 @@ const canShow = (
...
+): Promise<boolean> =>
+	Promise.resolve(
+		!isSignedIn &&
+			!hasUserDismissedGate(currentTest.variant, currentTest.name) &&
+			isNPageOrHigherPageView(3) &&
+			isValidContentType(CAPI) &&
+			isValidSection(CAPI) &&
+			isValidTag(CAPI) &&
+			!isIOS9(),
+	);
```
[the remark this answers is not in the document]
QuarpT: https://github.com/guardian/dotcom-rendering/pull/2661/files#r589458148

src/web/components/SignInGate/gates/mandatory-gate.tsx
```diff
@@ -0,0 +1,42 @@
...
+				<SignInGateMandatory
+					ophanComponentId={ophanComponentId}
+					dismissGate={dismissGate}
+					guUrl={guUrl}
+					signInUrl={signInUrl}
+					abTest={abTest}
+					isComment={isComment}
+				/>
+			</Suspense>
+		</Lazy>
```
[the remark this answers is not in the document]
QuarpT: Refactoring the gate tests was brought up recently on the Identity team. There's a lot of duplicated code for these signin gate tests which are frequently added and removed. 

I think we're pushing off refactoring/abstracting the code until the gate requirements have been confirmed.

src/web/components/SignInGate/SignInGateSelector.tsx
coldlink: Ah I see, that makes sense, just a note for myself: the `mandatory-gate-control` is the same design as the main variant test, but with different `canShow` rules, therefore the `gateMainVariant` method is used.

src/web/components/SignInGate/displayRule.ts
QuarpT: I've implemented a temporary solution https://github.com/guardian/dotcom-rendering/pull/2661/commits/4c1fb8853313f08114ff1f9bc643a5446eecc1d0 , but it's not ideal

src/web/components/SignInGate/displayRule.ts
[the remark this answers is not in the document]
liywjl: I assume the use of a `Promise` was because some of these functions are async? 
if so, you would benefit by using `Promise.all` to call them all at once rather than individually?
If none of them are async do you need to return a `Promise`?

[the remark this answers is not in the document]
QuarpT: `SignInGateComponent` requires a canShow function that returns a Promise. Some implementations of canShow are async. I have lifted the implementations that are not.

coldlink:
Would also be useful to add Cypress tests in for the mandatory gate https://github.com/guardian/dotcom-rendering/blob/main/cypress/integration/parallel-1/sign-in-gate.spec.js

QuarpT:
@coldlink 

> Would also be useful to add Cypress tests in for the mandatory gate https://github.com/guardian/dotcom-rendering/blob/main/cypress/integration/parallel-1/sign-in-gate.spec.js

I attempted writing a cypress test for this, but the CMP banner makes things difficult. After consenting, CMP JavaScript controls the 'articles viewed count', so we can't hack it manually from the test. Refreshing the page 3 times causes the test to timeout. If you think it's worth the effort, I can spend more time on this.

mxdvl:
If you want to keep the list of supported vendors updated, it might be worth importing the list directly from https://github.com/guardian/consent-management-platform/blob/273b854792509ceb5e390da3e6b092cb0aebe853/src/getConsentFor.ts#L7

coldlink:
> @coldlink
> 
> > Would also be useful to add Cypress tests in for the mandatory gate https://github.com/guardian/dotcom-rendering/blob/main/cypress/integration/parallel-1/sign-in-gate.spec.js
> 
> I attempted writing a cypress test for this, but the CMP banner makes things difficult. After consenting, CMP JavaScript controls the 'articles viewed count'. It seems to overwrite changes to local storage from the test. Refreshing the page 3 times causes the test to timeout. If you think it's worth the effort, I can spend more time on this.
> 
> edit:
> I've added some cypress tests [2642e22](https://github.com/guardian/dotcom-rendering/commit/2642e22b7a3bfb81851464e6bc5e8f245d14fe14)
> 
> The issue was the set localstorage function is synchronous, whereas we need to use the Cypress async version

Thanks, interesting that the cypress test on gh actions is failing, I checked out locally and they work fine. I have a feeling that it's something to do with the banners that appear after accepting all consents on the CMP.

QuarpT:
> If you want to keep the list of supported vendors updated, it might be worth importing `VendorIDs` directly from https://github.com/guardian/consent-management-platform/tree/main/src/getConsentFor.ts#L7

@mxdvl the VendorIDs enum is not currently exported, but we could change that. 

Also, not all of the consents are set to `true` after accepting all on the CMP banner.

coldlink (approved):
👍

mxdvl:
> > If you want to keep the list of supported vendors updated, it might be worth importing `VendorIDs` directly from [guardian/consent-management-platform@`main`/src/getConsentFor.ts#L7](https://github.com/guardian/consent-management-platform/tree/main/src/getConsentFor.ts#L7)
> 
> @mxdvl the VendorIDs enum is not currently exported, but we could change that.
> 
> Also, not all of the vendor ID consents are set to `true` after accepting all on the CMP banner. Do we have a list of vendor ID consents set by the banner?

Yes, scratch that. Looking at this again, I think that Identity should make a decision about which vendor consents it requires. If a specific vendor is not readily available, we can add it to the list.

The CMP library acts as an interface for consent, but it cannot make decisions on which consent you need for your app to run: that’s not something shared by all its consumers.

src/web/components/SignInGate/displayRule.ts
```diff
@@ -99,68 +97,28 @@ export const canShow = (
...
+			? Object.values(state.tcfv2.vendorConsents)
+			: [];
+		const isEmpty =
+			consentFlags.length === 0 || vendorConsentFlags.length === 0;
+
+		return (
+			!isEmpty &&
+			[...consentFlags, ...vendorConsentFlags].every(
+				(hasConsented) => hasConsented,
+			)
```
mxdvl, line 115: FWIW [we’ve used][] and liked `.every(Boolean)`. So did [membership][].

[we’ve used]: https://github.com/guardian/frontend/blob/89802f2230100a9b9c1dc42e9ff4c64f77ec6e3f/static/src/javascripts/projects/common/modules/commercial/build-page-targeting.js#L224
[membership]: https://github.com/guardian/membership-frontend/blob/43751df8baae6656cfd31fb6d695485c4ad3a19f/frontend/assets/javascripts/src/modules/analytics/cmp.es6#L27

src/web/components/SignInGate/displayRule.ts
```diff
@@ -99,68 +97,28 @@ export const canShow = (
...
-];
-
 export const hasRequiredConsents = (): Promise<boolean> => {
+	const hasConsentedToAll = (state: ConsentState) => {
+		const consentFlags = state.tcfv2?.consents
+			? Object.values(state.tcfv2.consents)
+			: [];
+		const vendorConsentFlags = state.tcfv2?.vendorConsents
+			? Object.values(state.tcfv2.vendorConsents)
+			: [];
```
mxdvl, line 107: You should not check this!

mxdvl (changes requested):
That’s the right idea, but after consideration, it’s actually better to only check the `consents`.

This is because if a new vendor is added to the vendor list, it will not be consented to, until the user re-opens the privacy manager dialog or is reconsented.

mxdvl (approved):
It’s better to be careful and only show the gate to users who explicitly consented to all. 😀 

Leave it as-is.

src/web/components/SignInGate/displayRule.ts
mxdvl: But if you prefer to be safe, it’s perfectly reasonable to do so!

PR #2663, opened 2021-03-06
Cypress ❤️ 

oliverlloyd:
## What
Gives Cypress a little love

### 🔨  Maintenance
- Migrated to new `cy.intercept` command
- Removed unused fetch mock

### 🔍  Coverage
- Deleted old tests we no longer need (we already have coverage elsewhere)
- Removed duplicate tests

### 🏎️  Speed
- We now execute tests in parallel using 6 seperate actions

## Before
Screenshot 2021-03-06 at 12 27 20

## After
Screenshot 2021-03-06 at 12 27 54

## Why?
Cypress is no longer the slowest Github Action we have and we now have quite a bit of headroom to add more tests without degrading the dev ex for builds


.github/workflows/cypress-parallel-2.yml
```diff
@@ -0,0 +1,20 @@
+name: cypress-2
```
gtrufitt, line 1: I have a feeling we can do this with the `matrix` stuff in github actions rather than breaking out into different workflows? https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategy

liywjl (approved):
The code looks good! would be great to use @gtrufitt 's suggestion of matrix for having parallel cypress tests

.github/workflows/cypress-parallel-2.yml
oliverlloyd: Amazing. Thanks for pointing out that feature, I've not seen it before and it is exactly what I needed!

oliverlloyd:
[Cypress Improvements](https://trello.com/c/xkh1jTIe/2701-cypress-improvements)

PR #2666, opened 2021-03-08
Remove carousel AB test

buck06191:
## What does this change?
Removes Carousel AB test - related to https://github.com/guardian/frontend/pull/23588

## Why?
Test is done!


buck06191:
[Tear down Carousel AB test](https://trello.com/c/tzGKQ16w/2700-tear-down-carousel-ab-test)

oliverlloyd:
Can we also remove the other test related code? We added a lot of logic as a temporary solution to get the test running so ideally this PR - where we remove the test - is a good place to prevent that code living on when it no longer has a use case?

buck06191:
> Can we also remove the other test related code? We added a lot of logic as a temporary solution to get the test running so ideally this PR - where we remove the test - is a good place to prevent that code living on when it no longer has a use case?

I can do that now, or I can extend the test in this PR and then handle the full tear down once we have results?

liywjl (approved):
👍

PR #2668, opened 2021-03-08
Remove padding-top of top-above-nav ad slot to match Frontend rendering

arelra:
## What does this change?

Remove `padding-top` of `top-above-nav` ad slot to match Frontend rendering

### Before

Screenshot 2021-03-08 at 11 58 50

### After - DCR

Screenshot 2021-03-08 at 12 11 54

## Why?

- Align DCR and Frontend rendering
- Remove whitespace


liywjl (approved):
Great catch!! 👍

PR #2669, opened 2021-03-08
Enable custom link styles in the article body

oliverlloyd:
## What?
Adds a pattern to allow the global link styles that we set on the article body to be ignored. Any element with the `data-ignore-global-link-styling` data attribute will not be selected and can have it's own a tag css

## Why?
Because the global selector has a high specificity and is hard to override but we sometimes need to. One upcoming example is if we place share icons inside each block on a live blog. These share icons use a tags and would pick up the global styles.

### What about making the global selector more targeted?
I could have used `a p {` but that would miss lists. So I'd have to have `a p, a li {` but... etc. It's perhaps just easier to use `:not`?

### Why not have different global styles for different contexts?
Another appoach is to have a separate or dynamic set of global styles depending on the type of article. But dynamic global styles probably isn't a path we want to go down.

### DIFF
--- a/src/web/components/ArticleBody.tsx
+++ b/src/web/components/ArticleBody.tsx
@@ -38,7 +38,7 @@ const bodyStyle = (display: Display) => css`
 `;
 
 const linkColour = (palette: Palette) => css`
-	a {
+	a:not([data-ignore-global-link-styling]) {
 		text-decoration: none;
 		border-bottom: 1px solid ${palette.border.articleLink};
 		color: ${palette.text.articleLink};
--- a/src/web/components/ShareIcons.tsx
+++ b/src/web/components/ShareIcons.tsx
@@ -86,6 +86,7 @@ export const ShareIcons: React.FC<{
 						role="button"
 						aria-label="Share on Facebook"
 						target="_blank"
+						data-ignore-global-link-styling={true}
 					>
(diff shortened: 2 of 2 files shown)


nicl:
I actually like this. Global styles are necessary but work against our scoped CSS; the data attribute is nice way to manage exceptions.

nicl (approved):

liywjl (approved):
Love this approach 👍

src/web/components/ArticleBody.tsx
```diff
@@ -38,7 +38,7 @@ const bodyStyle = (display: Display) => css`
 `;
 
 const linkColour = (palette: Palette) => css`
-	a {
+	a:not([data-ignore-global-link-styling]) {
```
gtrufitt, line 41: Is it worth following a key-valuation convention (like print.css https://github.com/guardian/dotcom-rendering/blob/main/src/static/css/print.css#L1)? I don't mind this, but I wonder if you can do `data-ignore="global-link-style"` so if you need to add `data-ignore="global-list-style` it follows the convention? Otherwise the value of `true` in the data attribute is not really meaningful (it could be `false` with the same effect)

oliverlloyd:
[data-ignore-global-link-styles](https://trello.com/c/xrCOn7Zp/2705-data-ignore-global-link-styles)

PR #2671, opened 2021-03-08
LiveBlock

oliverlloyd:
## What does this change?
Adds the `LiveBlock` coomponent

```typescript
type Props = {
	format: Format;
	block: Block;
	pageId: string;
	webTitle: string;
	abTests: CAPIType['config']['abTests'];
	adTargeting: AdTargeting;
	host?: string;
};
```

Screenshot 2021-03-08 at 16 21 50

## Pending
This work highlights the need to enhance `ShareIcons` to:
- Support a smaller variation of the icon
- To allow a hash value to be passed as part of any url being linked to
- Remove the padding that's being set inside the `ShareIcon` component and moving this to the context where this component is used (we don't want it in LiveBlocks)


oliverlloyd:
[LiveBlock](https://trello.com/c/wKqjBLLl/2706-liveblock)

src/web/components/LiveBlock.tsx
```diff
@@ -0,0 +1,336 @@
...
+export const LiveBlock = ({
+	format,
+	block,
+	pageId,
+	webTitle,
+	abTests,
+	adTargeting,
+	host,
+}: Props) => {
+	if (block.elements.length === 0) return null;
```
liywjl, line 215: Do we want to return `null` ?

nitro-marky (approved):
🚀

liywjl (approved):
LGTM

src/web/components/LiveBlock.tsx
nitro-marky: What are you thinking in particular? And empty object?...just curious 😄

oliverlloyd: I've learnt to always expect the unexpected at the Guardian, If a thing is possible, it will happen. But I didn't want to have guards all though the code and I'm confident we never want to render an empty block so by returning `null` it is signalling to React to ignore this component (React will pass over `null`s)

PR #2672, opened 2021-03-09
Convert className to adCss prop

liywjl:
## What does this change?
Convert `className` prop to `adCss`  prop

## Why?
Avoid prop names that could be confused for acctual DOM attributes

### DIFF
--- a/src/amp/components/Ad.test.tsx
+++ b/src/amp/components/Ad.test.tsx
@@ -40,7 +40,7 @@ describe('AdComponent', () => {
 				contentType={contentType}
 				config={commercialConfig}
 				commercialProperties={commercialProperties}
-				adCss=""
+				cssOverrides=""
 			/>,
 		);
 
@@ -94,7 +94,7 @@ describe('AdComponent', () => {
 				contentType={contentType}
 				config={commercialConfig}
 				commercialProperties={commercialProperties}
-				adCss=""
+				cssOverrides=""
 			/>,
 		);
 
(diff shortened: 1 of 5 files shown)


oliverlloyd (commented):
This is maybe worthy of a larger discussion around:

1. Do we want to use this pattern in DCR at all?
I woould argue passing class names as props is a useful pattern if you're a lib and want to give consumers the ability to override your api. You can't know their needs and perhaps shouldn't constrain them. But within DCR we have control and can and should sole variations in style requirements using the api. Passing classnames is always sub optimal as you can only ever type it as a string and have zero control over what gets passed.

2. If we do allow the pattern, then a joined up approach over what to name these props might make sense?
Perhaps a name that communicates something like `dangerouslyOverrideStyles`?

liywjl:
**1. Do we want to use this pattern in DCR at all?**

The logic behind this change was that in Emotion 11, we will be using `css` rather than `className`. Because `css` is picked up by a babel plugin to be converted into `className` it is probably preferable not to name things that could be construed as DOM attributes to avoid bugs. 

Furthermore, className expects a string, and if we will be passing down styles using `css` from Emotion 11, we wont be passing down strings but rather serialized styles that cannot be directly inserted into className.

Overall I do not favor one approche over the other but think we should just have an agreeded on standard to apply to the repo.

**2. If we do allow the pattern, then a joined up approach over what to name these props might make sense?
Perhaps a name that communicates something like dangerouslyOverrideStyles?**

Not sure about the use of `dangerously` in `dangerouslyOverrideStyles` but I agree we should find a standard if we go with this approach

liywjl:
I think we should embrace the same pattern as `Source` https://theguardian.design/2a1e5182b/p/2619e3-overriding-styles

Will convert the prop to `cssOverrides`

liywjl:
closed in favor of https://github.com/guardian/dotcom-rendering/pull/2676

PR #2674, opened 2021-03-09
Simplify AMP pagination logic

liywjl:
## What does this change?
- Convert `link` function to instead using `<a>` tag directly
- Overload styles rather than using a ternary

## Why?
Easier to understand


oliverlloyd (approved):
Looks great. As discussed irl, it'd be great to have a story for an amp page if that's simple to do?

PR #2675, opened 2021-03-09
Live layout styles

oliverlloyd:
## What does this change?
Takes the existing `LiveLayout` component and adds styling to it so the resulting article looks like a live blog

### Before
Screenshot 2021-03-09 at 12 13 17

### After
Screenshot 2021-03-09 at 12 12 57

## Why?
Chip chip

### Why have I removed the match report logic?
I deleted the layout logic relating to `Design.MatchReport` because the way these article types are represented in live blogs is different to a normal article and I think we'll end up with a better result if we start from scratch


src/web/lib/decidePalette.ts
```diff
@@ -55,6 +56,18 @@ const textSeriesTitle = (format: Format): string => {
...
 			switch (format.design) {
+				case Design.LiveBlog:
+					switch (format.theme) {
+						case Pillar.News:
+							return news[600];
+						case Pillar.Sport:
+						case Pillar.Lifestyle:
+						case Pillar.Culture:
+						case Pillar.Opinion:
+						case Special.Labs:
```
oliverlloyd, line 67: These may need to change depending on requirements

src/web/layouts/LiveLayout.tsx
```diff
@@ -480,6 +377,9 @@ export const LiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
 							/>
 						</div>
 					</GridItem>
+					<GridItem area="border">
+						<></>
+					</GridItem>
```
oliverlloyd, line 382: The same as Immersives, we don't show the grey border but we do still want the vertical spacing so we use this pattern

liywjl (approved):
LGTM

PR #2677, opened 2021-03-09
Live widths

oliverlloyd:
## What?
Adds support for different columns widths for LiveBlogs

## Why?
So that we give more space to key events (left column) and keep them on the page for longer

### What about main content width?
There are two divergences with the main column width here:

1. On frontend this is sometimes using a max width of `610px`. In DCR we have always used `620px` and I don't want to break that pattern.
2. Between `leftCol` and `wide` Frontend breaks out of the max width for article content and stretchs to the sides, up to 7 or 800px wide. This is outside of the conventional max width of `620px` so I wanted to approach this separately.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -473,6 +473,8 @@ type MediaType = 'Video' | 'Audio' | 'Gallery';
 
 type LineEffectType = 'squiggly' | 'dotted' | 'straight';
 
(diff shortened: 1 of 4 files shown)


oliverlloyd:
[Live Widths](https://trello.com/c/1D3LFo0k/2713-live-widths)

PR #2678, opened 2021-03-09
Automate npm package aliasing

liywjl:
## What does this change?
- Install `emotion` related packages used in automoate with aliases
- Adds `-legacy` suffix to the libraries (to indicate that we will be moving away from them in the future)

## Why?
As part of our Emotion 11 migration, we want to make sure we do not break implicite dependancies that use `window.guardian.automat`. To do this we reinstall the exported packages using aliases and add suffix name `-legacy` to inidcate that these are legacy version of emotion


liywjl:
`Braze` dependency is changing to use `peerDependencies` instead of `window.guardian.automat` to get React/Preact and Emotion packages.

Reader revenue is the only system that directly interfaces with `window.guardian.automat`, we will just need to coordinate on updating Emotion.

This PR is no longer needed as a workaround.

PR #2679, opened 2021-03-09
Don't use AMD loader for 99% of interactives

gtrufitt:
## What does this change?
- Brings in the behaviour of boot.js into the platform to avoid loading the AMD loader and the boot.js script.
- Adds the placeholder to reduce CLS, and avoid going from 0 height -> interactive height, not perfect but better than it was

### DIFF
--- a/src/web/components/elements/InteractiveBlockComponent.tsx
+++ b/src/web/components/elements/InteractiveBlockComponent.tsx
@@ -9,7 +9,7 @@ type Props = {
 	url?: string;
 	scriptUrl?: string;
 	alt?: string;
-	role?: string;
+	role?: RoleType;
 };
 
 /*
@@ -70,17 +70,21 @@ and to avoid loading the boot.js file.
 For the remaining few we dynamically load and AMD loader and support the contract as defined with curl AMD loader
 
 */
-const supportingMinHeight = 200;
-const inlineMinHeight = 500;
-const getMinHeight = (role: string, loaded: boolean) => {
+const decideHeight = (role: RoleType) => {
+	switch (role) {
+		case 'supporting':
+			return 200;
+		default:
+			return 500;
+	}
+};
+const getMinHeight = (role: RoleType, loaded: boolean) => {
 	if (loaded) {
 		return `auto`;
 	}
-	return role === 'supporting'
-		? `${supportingMinHeight}px`
-		: `${inlineMinHeight}px`;
+	return `${decideHeight(role)}px`;
 };
-const wrapperStyle = (role: string, loaded: boolean) => css`
+const wrapperStyle = (role: RoleType, loaded: boolean) => css`
 	${body.medium()};
 	background-color: ${neutral[100]};
 	font-weight: 300;
@@ -216,10 +220,7 @@ export const InteractiveBlockComponent = ({
 			const iframe = document.createElement('iframe');
 			iframe.style.width = '100%';
 			iframe.style.border = 'none';
-			iframe.height =
-				role === 'supporting'
-					? supportingMinHeight.toString()
-					: inlineMinHeight.toString(); // default height
+			iframe.height = decideHeight(role).toString();
 			iframe.src = url;
 
 			setupWindowListeners(iframe);
@@ -261,11 +262,7 @@ export const InteractiveBlockComponent = ({
 		>
 			{!loaded && (
 				<Placeholder
-					height={
-						role === 'supporting'
-							? supportingMinHeight
-							: inlineMinHeight
-					}
+					height={decideHeight(role)}
 					shouldShimmer={false}
 				/>
 			)}
--- a/src/web/components/elements/InteractiveBlockComponent.tsx
+++ b/src/web/components/elements/InteractiveBlockComponent.tsx
@@ -1,4 +1,5 @@
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useRef, useState } from 'react';
+import { useOnce } from '@root/src/web/lib/useOnce';
 import { css } from 'emotion';
 import { body } from '@guardian/src-foundations/typography';
 import { space } from '@guardian/src-foundations';
@@ -216,12 +217,7 @@ export const InteractiveBlockComponent = ({
 	const wrapperRef = useRef<HTMLDivElement>(null);
 	const placeholderLinkRef = useRef<HTMLAnchorElement>(null);
 	const [loaded, setLoaded] = useState(false);
-	useEffect(() => {
-		// Once loaded, we don't touch it.
-		if (loaded) {
-			return;
-		}
-
+	useOnce(() => {
 		// We've brought the behavior from boot.js into this file to avoid loading 2 extra scripts
 		if (
 			scriptUrl ===
@@ -264,7 +260,7 @@ export const InteractiveBlockComponent = ({
 
 			setLoaded(true);
 		}
-	}, [scriptUrl, role, url, loaded]);
+	}, [loaded]);
 
 	return (
 		<div
--- a/src/web/components/elements/InteractiveBlockComponent.tsx
+++ b/src/web/components/elements/InteractiveBlockComponent.tsx
@@ -245,7 +245,7 @@ export const InteractiveBlockComponent = ({
 
 			setLoaded(true);
 		}
-	}, [scriptUrl]);
+	}, [scriptUrl, role, url]);
 
 	return (
 		<div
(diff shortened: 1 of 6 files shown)


jfsoul:
Not one for this PR specifically, but it'd be nice to add a high-level "state of interactives" doc to DCR I think - and maybe a quick `contract` in the special directory for the AMD loader.

oliverlloyd:
Just a quick question before looking at the code: Is that red text on the placeholder desired?

src/web/components/elements/InteractiveBlockComponent.tsx
```diff
@@ -57,18 +63,171 @@ and is sent with all interactive elements in the scriptUrl from CAPI to do a two
...
+const supportingMinHeight = 200;
+const inlineMinHeight = 500;
+const getMinHeight = (role: string, loaded: boolean) => {
+	if (loaded) {
+		return `auto`;
+	}
+	return role === 'supporting'
+		? `${supportingMinHeight}px`
+		: `${inlineMinHeight}px`;
+};
```
oliverlloyd, line 82: This is maybe just a naming thing but the default here is called 'inline...' but it's also capturing all the other roles

gtrufitt:
> Just a quick question before looking at the code: Is that red text on the placeholder desired?

Hm, I think it just uses the article level styles for anchors. @HarryFischer do we want a different link colour?

src/web/components/elements/InteractiveBlockComponent.tsx
```diff
@@ -57,18 +63,171 @@ and is sent with all interactive elements in the scriptUrl from CAPI to do a two
...
+const wrapperStyle = (role: string, loaded: boolean) => css`
+	${body.medium()};
+	background-color: ${neutral[100]};
+	font-weight: 300;
```
oliverlloyd, line 86: Can we set this using `body.medium({fontWeight: '...'})`?

src/web/components/elements/InteractiveBlockComponent.tsx
```diff
@@ -57,18 +63,171 @@ and is sent with all interactive elements in the scriptUrl from CAPI to do a two
...
+	min-height: ${getMinHeight(role, loaded)};
+	position: relative;
+`;
+
+const placeholderLinkStyle = css`
+	position: absolute;
+	bottom: 1rem;
+	left: 1rem;
```
oliverlloyd, line 94: Are we doing rems now? I've been hearing talk about this but I'm still working to the assumption code should be written in pixels.

src/web/components/elements/InteractiveBlockComponent.tsx
```diff
@@ -57,18 +63,171 @@ and is sent with all interactive elements in the scriptUrl from CAPI to do a two
...
 
-export const InteractiveBlockComponent = ({ url, scriptUrl, alt }: Props) => {
+	window.addEventListener(
+		'message',
+		(event) => {
+			if (event.source !== iframe.contentWindow) {
+				return;
+			}
+
+			const message: Record<string, unknown> = JSON.parse(event.data);
```
oliverlloyd, line 114: Maybe set a try catch around that `parse`? I'm not sure what control we have oover `event.data` but if it were to contain a function or just not be valid in general...

src/web/components/elements/InteractiveBlockComponent.tsx
```diff
@@ -57,18 +63,171 @@ and is sent with all interactive elements in the scriptUrl from CAPI to do a two
...
+
+export const InteractiveBlockComponent = ({
+	url,
+	scriptUrl,
+	alt,
+	role = 'inline',
+}: Props) => {
 	const wrapperRef = useRef<HTMLDivElement>(null);
+	const placeholderLinkRef = useRef<HTMLAnchorElement>(null);
+	const [loaded, setLoaded] = useState(false);
```
oliverlloyd, line 202: Can we use `useOnce` here for this?

src/web/components/elements/InteractiveBlockComponent.tsx
oliverlloyd: I see this ternary is used in a few places. What about having a `decideHeight(role: RooleType)` function with a simple switch to better convey intentions and defaults?

src/web/components/elements/InteractiveBlockComponent.tsx
```diff
@@ -89,17 +248,36 @@ export const InteractiveBlockComponent = ({ url, scriptUrl, alt }: Props) => {
...
+			{!loaded && (
+				<a
+					ref={placeholderLinkRef}
+					data-name="placeholder"
+					className={placeholderLinkStyle}
+					href={url}
+				>
+					{alt}
+				</a>
+			)}
```
oliverlloyd, line 280: 🤔 I'm only half understanding this code but would it work if we server rendered the placeholder and then used a portal to replace it with the final content?

Or. Is it that we have 2 loading states. The SSR placeholder, _and_ also a client side placeholder for when we're making the api call to get boot.js? In which case... I think youo could have a more classic setup where you use `useApi` to fetch the script and render a placeholder based on the `loading` prop that this custom hook returns?

I feel like I'm missing an aspect to this though so I could be oversimplifying things. I can see there's something around needing to know the height of something after it's rendered...?

Sorry, that was a very rambling comment!

oliverlloyd (approved):
A few suggestions for some simple improvements and then me just dumping a stream of conciousness on the page trying to get my head around the loading order. Maybe we can talk it through so I understand it better?

src/web/components/elements/InteractiveBlockComponent.tsx
gtrufitt: There's a few difficulties:

- The script is an AMD module so we need to use an AMD loader to fetch it, we can't use `useAPI`
- I tried to use portal and it just added a new div element in the container and I wasn't sure why, hydration works as expected (I am 'hydrating' the placeholder and anchor with the client side fetching stuff essentially, there's not an empty `div` like portal expects)
- The steps are placeholder -> setup listeners -> switch out for iframe -> iframe calls out to. parent and sets its height. So there's a contract there that needs to be taken into account.

src/web/components/elements/InteractiveBlockComponent.tsx
gtrufitt: I reach for rems by default, have updated!

src/web/components/elements/InteractiveBlockComponent.tsx
gtrufitt: I don't think so because the state we would need to wait on is set within the `useEffect`?

src/web/components/elements/InteractiveBlockComponent.tsx
oliverlloyd: >     * added a new div element in the container

I think Portal is looking for a marker so it can remove the placeholder... This:

```
	const placeholderElement = element.querySelector(
		'[data-name="placeholder"]',
	);
```

oliverlloyd: This is the problem with weird contracts that are hard to discover... 😞

src/web/components/elements/InteractiveBlockComponent.tsx
oliverlloyd: `useOnce` uses `useEffect` so the effect is the same (if that makes sense?)

oliverlloyd: But you're using `loaded` to branch rendering too so...

gtrufitt: Ah yes, no that works 👍

src/web/components/elements/InteractiveBlockComponent.tsx
gtrufitt: Yeah, that's in there from where it was a portal before 🤔

gtrufitt: I'm going to leave this as is because I think I *want* server-rendered placeholder and if I was to use a portal I'd need to branch the interactiveBlockElement into a server-rendered element and client rendered element, whereas I can just handle that in the component itself.

oliverlloyd: I'll catch up with you directly. As far as I'm aware the Portal is designed for this purpose. You have a SSR placeholder and then that gets replaced by the actual content on the client following an api call.

Sorry, I've been tied with lots of things today so I haven't been able to dig into tis more (or respond sooner!)

gtrufitt: Yeah, no rush, we can look next week, I'd be keen to get to the bottom of the pattern and happy to update if it makes sense!

PR #2685, opened 2021-03-10
Small ShareIcons

oliverlloyd:
## What?
Introduces a new, smaller, share icons size.

## Why?
Because live blocks need these icons to have a 23px width and height

I also removed the bottom padding that was being set on these as this doesn't seem to be used anywhere in normal articles and interferes with live blocks.


oliverlloyd:
This completes some of the pending actions on #2671

oliverlloyd:
[Small ShareIcons](https://trello.com/c/VWC6MGGb/2714-small-shareicons)

PR #2687, opened 2021-03-10
Add a bool check to stop the data being sent more than once

nitro-marky:
## What does this change?

Adds a boolean to check if the post has already been fired, hopefully this will stop the duplicates we seem to be getting.

### DIFF
--- a/src/web/browser/coreVitals/coreVitals.ts
+++ b/src/web/browser/coreVitals/coreVitals.ts
@@ -63,16 +63,19 @@ export const coreVitals = (): void => {
 				break;
 		}
 
-		// Set page view and browser ID
-		if (window.guardian.config.ophan.browserId && window.guardian.ophan) {
-			jsonData.page_view_id = window.guardian.ophan.pageViewId;
-			jsonData.browser_id = window.guardian.config.ophan.browserId;
-		}
-
 		// Check if POST has been sent
 		if (!hasBeenSent) {
 			// If CLS has been calculated
 			if (jsonData.cls !== null) {
+				// Set page view and browser ID
+				if (
+					window.guardian.config.ophan.browserId &&
+					window.guardian.ophan
+				) {
+					jsonData.page_view_id = window.guardian.ophan.pageViewId;
+					jsonData.browser_id =
+						window.guardian.config.ophan.browserId;
+				}
 				fetch(endpoint, {
 					method: 'POST', // *GET, POST, PUT, DELETE, etc.
 					mode: 'cors', // no-cors, *cors, same-origin
--- a/src/web/browser/coreVitals/coreVitals.ts
+++ b/src/web/browser/coreVitals/coreVitals.ts
@@ -8,7 +8,7 @@ type CoreWebVitalsPayload = {
 	fid: null | number;
 	cls: null | number;
 	lcp: null | number;
-	fcp: null | number;
+	fcp: null | number; // Report as a DOMHighResTimeStamp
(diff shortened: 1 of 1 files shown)


src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -52,20 +63,16 @@ export const coreVitals = (): void => {
...
+		// Set page view and browser ID
+		if (window.guardian.config.ophan.browserId && window.guardian.ophan) {
+			jsonData.page_view_id = window.guardian.ophan.pageViewId;
+			jsonData.browser_id = window.guardian.config.ophan.browserId;
+		}
 
+		// Check if POST has been sent
+		if (!hasBeenSent) {
+			// If CLS has been calculated
+			if (jsonData.cls !== null) {
```
oliverlloyd, line 69: You can group these two conditional I presume?

src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -27,6 +27,16 @@ const jsonData: CoreWebVitalsPayload = {
...
+let hasBeenSent = false;
+
+// Can be tested locally or on prod
+const endpoint =
+	window.location.hostname === 'm.code.dev-theguardian.com' ||
+	window.location.hostname === 'localhost' ||
+	window.location.hostname === 'preview.gutools.co.uk'
+		? 'http://performance-events.code.dev-guardianapis.com/core-web-vitals'
+		: 'https://performance-events.guardianapis.com/core-web-vitals';
+
```
oliverlloyd, line 39: Did you mean to include these changes? They don't seem related to this PR?

src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -52,20 +63,19 @@ export const coreVitals = (): void => {
...
+			if (jsonData.cls !== null) {
+				// Set page view and browser ID
+				if (
+					window.guardian.config.ophan.browserId &&
+					window.guardian.ophan
+				) {
+					jsonData.page_view_id = window.guardian.ophan.pageViewId;
+					jsonData.browser_id =
+						window.guardian.config.ophan.browserId;
+				}
```
oliverlloyd, line 78: Unrelated changes?

src/web/browser/coreVitals/coreVitals.ts
nitro-marky: Sure, I just did it as an initial readability thing

src/web/browser/coreVitals/coreVitals.ts
nitro-marky: They were in before, I just moved them. It means it can be tested locally with no changes

nitro-marky:
Sending one piece of data is a lot easier in the long term than trying to remove duplicates in the SQL side

nitro-marky:
There are still moving pieces at the moment so it's not in a final form

oliverlloyd:
> Sending one piece of data is a lot easier in the long term than trying to remove duplicates in the SQL side

My feeling on this is that actually removing queries in an SQL query should be trivial? I could be wrong on that though, you would need some form of page view id to group on - but assuming you _can_ identify the page view, then it should be simple to group on that.

My fear here is we're opening ourselves up the kind of bugs where if the core vitals lib reports data in a different way to what we expect in the wild, then we lose data. By keeping the clientside code simple we reduce that exposure

nitro-marky:
I'll revert the changes and leave it to data

PR #2691, opened 2021-03-10
RichLinks use Figure

oliverlloyd:
## What?
Here we refactor the `RichLink` component to make it easier to adapt it for Liveblogs later as well as  wrapping `RichLink` in the `Figure` component for positioning

## Why?
Rich links very closely follow the positioning given to other element types based on `role` so instead of having duplicate css, I've migrated `RichLink` to using the `Figure` component.

Unfortunately, the `thubnail` role isn't exactly what we want, it's too narrow at some breakpoints, so I've created a supplementary `role` of `richLink`.

### DIFF
--- a/src/web/components/Figure.tsx
+++ b/src/web/components/Figure.tsx
@@ -7,7 +7,7 @@ import { space } from '@guardian/src-foundations';
 type Props = {
 	children: React.ReactNode;
 	isMainMedia?: boolean;
-	role?: RoleType;
+	role?: RoleType | 'richLink';
 	id?: string;
 };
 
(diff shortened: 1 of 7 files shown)


oliverlloyd:
[Inline RichLink](https://trello.com/c/QQXT1kLG/2715-inline-richlink)

src/web/components/Figure.tsx
```diff
@@ -7,7 +7,7 @@ import { space } from '@guardian/src-foundations';
 type Props = {
 	children: React.ReactNode;
 	isMainMedia?: boolean;
-	role?: RoleType;
+	role?: RoleType | 'richLink';
```
liywjl, line 10: curious to know why `richLink` isnt a Role?

src/web/components/RichLink.tsx
```diff
@@ -259,7 +219,7 @@ export const RichLink = ({
 			className={pillarBackground(format)}
 			data-name={(isPlaceholder && 'placeholder') || ''}
 		>
-			<div className={cx(richLinkContainer, neutralBackground)}>
+			<div className={cx(neutralBackground)}>
```
liywjl, line 222: no need for `cx`

liywjl (approved):
LGTM, just some minor things

src/web/components/Figure.tsx
oliverlloyd: I wanted to be explicit that `richLink` was being added as an exception here to handle the fact that rich links are a bit special. If I were to 'officially' add a new `role` type to the model then that would involve a lot more work to add support and testing for it plus it would no longer match the data being sent from CAPI.

It's not ideal, but based on our discussions in the chat channel, this seemed the best approach.

src/web/components/RichLink.tsx
oliverlloyd: Good spot!

PR #2692, opened 2021-03-11
LiveBlock RichLinks

oliverlloyd:
## What does this change?
All elements appearing inside `LiveBlock`s are now forced to be inline

### Before
Screenshot 2021-03-11 at 09 44 28

### After
Screenshot 2021-03-11 at 09 43 56

## Why?
This means `RichLink`s are no longer rendered with a negative left margin, putting them in the leftCol.

### DIFF
--- a/src/web/components/LiveBlock.tsx
+++ b/src/web/components/LiveBlock.tsx
@@ -277,7 +277,7 @@ export const LiveBlock = ({
 				{mainElements.map((element, index) => {
 					if (typesWeStretch.includes(element._type)) {
 						return (
-							<BlockMedia key={index}>
+							<BlockMedia key={`${element._type}-${index}`}>
 								<ElementRenderer
 									isMainMedia={false}
 									isLiveBlog={true}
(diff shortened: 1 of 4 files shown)


oliverlloyd:
[Inline RichLink](https://trello.com/c/QQXT1kLG/2715-inline-richlink)

src/web/lib/ElementRenderer.tsx
```diff
@@ -462,13 +491,13 @@ export const ElementRenderer = ({
 				</Figure>
 			);
 		case 'model.dotcomrendering.pageElements.RichLinkBlockElement':
+			console.log('isLiveBlog', isLiveBlog);
```
liywjl, line 494: to remove

liywjl (approved):
LGTM

src/web/components/LiveBlock.tsx
```diff
@@ -276,9 +277,10 @@ export const LiveBlock = ({
 				{mainElements.map((element, index) => {
 					if (typesWeStretch.includes(element._type)) {
 						return (
-							<BlockMedia>
+							<BlockMedia key={index}>
```
buck06191, line 280: As we chatted about a bit earlier, this is technically an [anti-pattern](https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318) but probably isn't a massive issue. Just maybe something we need to watch out for more to set good patterns in the codebase and because live blogs might have a different reaction to changing elements in a list

buck06191 (approved):
LGTM

PR #2693, opened 2021-03-11
LiveBlogRenderer

oliverlloyd:
## What does this change?
Adds a sibling to the `ArticleRenderer` component, `LiveBlogRenderer`

```typescript
type Props = {
	format: Format;
	blocks: Block[];
	adTargeting: AdTargeting;
	host?: string;
	abTests: CAPIType['config']['abTests'];
	pageId: string;
	webTitle: string;
};
```

Screenshot 2021-03-11 at 10 18 46

## Why?
Multiple block lovelyness

### What's happened with the global styles we apply to articles?
I've broken these blobs of global css down into smaller names chunks so that we have better control over what global styles go where. I've also renamed them so it's clear that they are global css.

### DIFF
--- a/src/web/components/ArticleBody.tsx
+++ b/src/web/components/ArticleBody.tsx
@@ -4,7 +4,8 @@ import { css, cx } from 'emotion';
 import { headline } from '@guardian/src-foundations/typography';
 import { between } from '@guardian/src-foundations/mq';
 import { ArticleRenderer } from '@root/src/web/lib/ArticleRenderer';
-import { Display } from '@guardian/types';
+import { LiveBlogRenderer } from '@root/src/web/lib/LiveBlogRenderer';
+import { Design, Display } from '@guardian/types';
 import type { Format } from '@guardian/types';
 
 type Props = {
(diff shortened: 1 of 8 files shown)


oliverlloyd:
[Multi Blocks](https://trello.com/c/W4kiFLjp/2720-multi-blocks)

liywjl (approved):
LGTM

shtukas (approved):

src/web/components/ArticleBody.tsx
```diff
@@ -57,9 +66,43 @@ export const ArticleBody = ({
...
+					blocks={blocks}
+					adTargeting={adTargeting}
+					host={host}
+					abTests={abTests}
+					pageId={pageId}
+					webTitle={webTitle}
+				/>
+			</div>
+		);
+	}
```
gtrufitt, line 95: I have a feeling we might find this branching to be happening too low down and we might want to do it at the route level. Probably fine for now, but something to be aware of

oliverlloyd: Yeah, I think there's a larger discussion about how we manage rendering liveblogs. Right now though, given how everything thing else currently works, this gets something working on the page

That said... even with a new route, we'd still be asking DCR to render this and there will need to be a branch based on `format` somewhere 🤔

PR #2694, opened 2021-03-11
Adding componentEvent tracking to the click to view component

frj:
The ClickToView component renders an placeholder for third party embedded contents like embedded instagram posts. Users need to click the 'accept' button, consenting to being tracked by the third party provider, before they can see the embedded content.

This PR contains the changes to add ophan tracking to help us understand if users are clicking on the 'accept' button, or if they are being put off viewing the content by the presence of the placeholder.

This is modelled using ophan componentEvents. 

An event is trigger with an 'View' action when the embed is rendered on the page, and then an event with a  'Click' action is sent when the user clicks on the 'accept' button.

The following additional fields are send with the componentEvent to provide additional context:

- component.componentType=CLICK_TO_VIEW_EMBED_OVERLAY
- component.labels=[$providerSourceDomain, $providerName]
- component.abTest={name: 'clickToViewVariant',variant: 'variant',}

TODO
- The types library needs to be updated to add the CLICK_TO_VIEW_EMBED_OVERLAY componentType

### DIFF
--- a/src/web/components/ClickToView.tsx
+++ b/src/web/components/ClickToView.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import { css } from 'emotion';
 
 import { border, background } from '@guardian/src-foundations/palette';
@@ -7,6 +7,12 @@ import { headline, textSans } from '@guardian/src-foundations/typography';
 import { Button } from '@guardian/src-button';
 import { SvgCheckmark } from '@guardian/src-icons';
 import { space } from '@guardian/src-foundations';
+import { OphanComponentType } from '@root/src/web/browser/ophan/ophan';
+import {
+	submitViewEventTracking,
+	submitClickEventTracking,
+	ABTestVariant,
+} from './SignInGate/componentEventTracking';
 
 type Props = {
 	children: React.ReactNode;
@@ -79,6 +85,13 @@ const roleButtonText = (role: RoleType) => {
 	}
 };
 
+const clickToViewCompnentType: OphanComponentType = 'CLICK_TO_VIEW_EMBED_OVERLAY' as OphanComponentType;
+
+const clickToViewABTestVariant: ABTestVariant = {
+	name: 'clickToViewVariant',
+	variant: 'variant',
+};
+
 const shouldDisplayOverlay = ({
 	isTracking,
 	isOverlayClicked,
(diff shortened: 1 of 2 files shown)


oliverlloyd (commented):
I think we may have some existing patterns in the codebase for tracking which are not exactly what we want. We've got a few issues in our backlog to look at these, possibly moving away from them.

Specifically, I hesitant to implement tracking in DCR in an imperative manner where the tracking lib requires us to initiate the call. I also think creating these couplings between /types and the `SignInGate` will create more complexity later.

To simplify things, I think the use case here is we want to know:

1. When a ClickToView component appears on the page and
2. when it is clicked?

There's not a lot of documentation out there on how to use Ophan but from looking around on Frontend and in Ophan itself I think we can use

`data-link-name=""` and
`data-component=""`

`data-link-name` provides click tracking and `data-component` provide visibility tracking (appears in the viewport).

Link tracking is well established but see [these comments](https://github.com/guardian/ophan/blob/75b86abcce07369c8998521399327d436246c016/tracker-js/assets/coffee/ophan/components.coffee) and also the [original PR](https://github.com/guardian/ophan/pull/1515) for visibility tracking

Both Frontend and DCR use these two attributes to capture tracking data and send it to Ophan.

Where more granular data is needed, the pattern is to separate it using `:`s. Eg.: 

`data-component="Interactive:standfirst_cta"`
`data-component="Competition:send email box"`

but it can also just be a simple string and the two attributes often appear together:

`data-link-name="breaking news" data-component="breaking-news">`
etc.

There is [an open PR](https://github.com/guardian/ophan/pull/4012) to optimise visibility tracking to use a more modern api but it should still serve our purposes as is. It's already working well on both Frontend and DCR.

In short, I think we don't need any of this react code and can just set data attributes and let Ophan do what it does

frj:
Thanks for taking a look @oliverlloyd.

I missed the fact that the data-component="xxxx" was getting sent in the renderedComponent fields of one of the requests to ophan. Eg:

```
http://ophan.theguardian.com/img/2?viewId=km4z8k3yc3jp1h4bizqi&performance=%7B%22dns%22%3A0%2C%22connection%22%3A0%2C%22firstByte%22%3A249%2C%22lastByte%22%3A3%2C%22domContentLoadedEvent%22%3A1294%2C%22loadEvent%22%3A335%2C%22navType%22%3A1%2C%22redirectCount%22%3A0%7D&renderedComponents=%5B%22nav2%22%2C%22sub-nav%22%2C%22section%22%2C%22meta-byline%22%2C%22auto-linked-tag%22%2C%22click-to-view-component%22%2C%22rich-link%22%2C%22geo-most-popular%22%2C%22most-popular%22%2C%22footer%22%5D
```

Thanks for bearing with me on this one, I think we are on the same page.

oliverlloyd:
> Thanks for bearing with me on this one, I think we are on the same page.

Not at all - it was an interesting learning exercise. We're talking about rebuilding `tracker-js` (the lib Ophan uses) to migrate it from Coffeescript to Typescript so it's good to get my head around this.

PR #2695, opened 2021-03-11
Change Browser ID collection order

nitro-marky:
## What does this change?

This checks for the browser and page view ID each time an item is added to the vitals object.

## Why?

Because sometimes browser_ID is not captured in the table.

### DIFF
--- a/src/web/browser/coreVitals/coreVitals.ts
+++ b/src/web/browser/coreVitals/coreVitals.ts
@@ -52,6 +52,13 @@ export const coreVitals = (): void => {
 				break;
 		}
 
+		// Some browser ID's are not caputured (and if they have no cookie there won't be on)
+		// but there are occassions of reoccuring users without a browser ID being sent
+		if (window.guardian && window.guardian.ophan) {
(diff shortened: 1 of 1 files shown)


src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -62,23 +69,19 @@ export const coreVitals = (): void => {
...
-					credentials: 'same-origin', // include, *same-origin, omit
-					headers: {
-						'Content-Type': 'application/json',
-					},
-					redirect: 'follow',
-					referrerPolicy: 'no-referrer', // no-referrer, *no-referrer-w
-					body: JSON.stringify(jsonData),
-				}).catch(() => {});
-			}
+			fetch(endpoint, {
```
jfsoul, line 73: So this now means that we'll send the data even if ophan hasn't loaded? (but with nulls for pageview and browser ID

jfsoul (approved):

src/web/browser/coreVitals/coreVitals.ts
nitro-marky: No, it still has the CSL check on line 70

PR #2696, opened 2021-03-11
Emotion 11 SSR

liywjl:
⚠️ DO NOT MERGE ⚠️ 
## What does this change?

Implements SSR using Emotion 11

## NOTE
This needs to be tested when we move the rest of the styling to Emotion 11. This PR is kept as a reference and base just to understand what changes we need to do on the server


oliverlloyd:
This PR is no longer needed

PR #2697, opened 2021-03-11
Use palette in Branding

gtrufitt:
## What does this change?
Removes usage of `pillarPalette` outside `decidePalette`

## Why?
Becasue we want to centralise colour decisions so we can share our `palette`


liywjl (approved):
Scrolling through, looks good (but there is a lot going on).

PR #2703, opened 2021-03-12
Log Braze Epic views to Ophan

rhystmills:
### Before
Braze Epic views are only logged to Braze.

### After
Logs a view `ComponentEvent` to Ophan for the Braze Epic. This new event was recently added to [Ophan](https://github.com/guardian/ophan/pull/4046) and [@guardian/types](https://github.com/guardian/types/pull/94).

@guardian/types has also been bumped to version 5.1.0.

## Why?
In order to facilitate analysis of upcoming marketing via the Braze Epic channel.

### DIFF
--- a/src/web/components/SlotBodyEnd/BrazeEpic.tsx
+++ b/src/web/components/SlotBodyEnd/BrazeEpic.tsx
@@ -12,7 +12,7 @@ import { CanShowResult } from '@root/src/web/lib/messagePicker';
 import { useOnce } from '@root/src/web/lib/useOnce';
 import { joinUrl } from '@root/src/lib/joinUrl';
 import { useHasBeenSeen } from '@root/src/web/lib/useHasBeenSeen';
-import { submitComponentEvent } from 'src/web/browser/ophan/ophan';
+import { submitComponentEvent } from '@root/src/web/browser/ophan/ophan';
 
 const { css } = emotion;
 
--- a/src/web/components/SlotBodyEnd/BrazeEpic.tsx
+++ b/src/web/components/SlotBodyEnd/BrazeEpic.tsx
@@ -12,6 +12,7 @@ import { CanShowResult } from '@root/src/web/lib/messagePicker';
 import { useOnce } from '@root/src/web/lib/useOnce';
 import { joinUrl } from '@root/src/lib/joinUrl';
 import { useHasBeenSeen } from '@root/src/web/lib/useHasBeenSeen';
+import { submitComponentEvent } from 'src/web/browser/ophan/ophan';
 
 const { css } = emotion;
 
@@ -127,8 +128,17 @@ const BrazeEpic = ({
 	}, [contributionsServiceUrl, meta]);
 
 	useEffect(() => {
-		if (hasBeenSeen && meta) {
+		if (hasBeenSeen && meta && meta.dataFromBraze) {
 			meta.logImpressionWithBraze();
+
+			// Log VIEW event with Ophan
+			submitComponentEvent({
+				component: {
+					componentType: 'RETENTION_EPIC',
+					id: meta.dataFromBraze.componentName,
+				},
+				action: 'VIEW',
+			});
 		}
 	}, [hasBeenSeen, meta]);
 
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
         "@guardian/src-radio": "^2.7.1",
         "@guardian/src-text-area": "^2.7.1",
         "@guardian/src-text-input": "^2.7.1",
-        "@guardian/types": "^5.0.0",
+        "@guardian/types": "^5.1.0",
         "@hkdobrev/run-if-changed": "^0.3.1",
         "@loadable/component": "^5.14.1",
         "@loadable/server": "^5.14.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2485,10 +2485,10 @@
     "@guardian/src-helpers" "^2.7.1"
     "@guardian/src-icons" "^2.7.1"
 
-"@guardian/types@^5.0.0":
-  version "5.0.0"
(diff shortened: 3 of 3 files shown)


src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -127,8 +128,17 @@ const BrazeEpic = ({
...
+
+			// Log VIEW event with Ophan
+			submitComponentEvent({
+				component: {
+					componentType: 'RETENTION_EPIC',
+					id: meta.dataFromBraze.componentName,
+				},
+				action: 'VIEW',
+			});
 		}
```
gtrufitt, line 142: I think we might be rebuilding behaviour that is already in tracker-js here https://github.com/guardian/ophan/pull/4012 ?

oliverlloyd: I noticed this in another PR, I think you can replace this logic with simply adding

```
data-link-name='RETENTION_EPIC'
data-component='RETENTION_EPIC'
```

`link-name` tracks clicks, `component` tracks visibility (how long it is visible in the viewport for)

gtrufitt: To clarify "component tracks visibility (how long it is visible in the viewport for)"

`data-component` tracks components that were rendered at the time we send a request to Ophan. It does not track time visible in viewport *or* lazy loaded components (though data-link-name will work to track clicks on lazy-loaded components). 

The linked PR will track attention time for components in the viewport at render-time and lazily-loaded. 

I believe the workaround we're seeing here is because we cannot track lazily-loaded components, whereas fixing attention time in tracker-js would meet this requirement. I am not sure why we have a new Component Event field, maybe it helps with tracking on the back-end but it does seem to be replicated behaviour in the client.

oliverlloyd: > `data-component` tracks components that were rendered at the time we send a request to Ophan (when the page has rendered). It does not track time visible in viewport

Ah, I thought it was sent to Ophan as time visible in the viewport (`attentionMs`)? But I'm only assuming that [based on this PR](https://github.com/guardian/ophan/pull/1515), which is old!

gtrufitt: > > `data-component` tracks components that were rendered at the time we send a request to Ophan (when the page has rendered). It does not track time visible in viewport
> 
> Ah, I thought it was sent to Ophan as time visible in the viewport (`attentionMs`)? But I'm only assuming that [based on this PR](https://github.com/guardian/ophan/pull/1515), which is old!

That's what you'd expect, yes, but in reviewing it to migrate to intersectionObserver it falls over somewhere!

oliverlloyd: Ha, I see!

It'd be great to expediate that PR you raised. It looks like a painful chore to add tracking like this to components when it really shouldn't be (sorry, you're obviously way ahead of us all here!)

rhystmills: Cheers for the feedback - to clarify, what behaviour do you believe we have duplicated here?

To add some context - `hasBeenSeen` is using `intersectionObserver` so that we can ensure that an impression is only logged with Braze (by running `appboy.logInAppMessageImpression` via `meta.logImpressionWithBraze`) and Ophan when the Epic is visible to the user.

Do you think it would be possible to do this with functionality that's already on the platform?

gtrufitt: > to clarify, what behaviour do you believe we have duplicated here?

The linked PR uses `intersectionObserver` to track 'attention time' which if `> 0` would equal the same has `hasBeenSeen` tracks. It's built into Ophan's tracker-js *but* I don't believe is a part of apps so if your solution is being implemented there too then you likely want to leave as-is. 

Otherwise, web-only, we should try and use (or based on the fact we don't currently have it, refactor in) the tracker-js behaviour *unless* there's a compelling reason not to from a data integrity perspective

tjmw: Thanks for pointing this out @gtrufitt, I'd not seen your tracker-js PR until now, that's great. So once that PR is merged the only thing we'd need to do to take advantage of it is specify the component type declaratively using `data-component`? Currently we also pass a component ID, do you know how that fits in? (Is there another data attribute we can use?)

> The linked PR uses intersectionObserver to track 'attention time' which if > 0 would equal the same has hasBeenSeen tracks

One piece which may not quite align is that `hasBeenSeen` supports debouncing, so a view isn't recorded until the component has intersected with the viewport for 200ms. It sounds like we could use attention time to get the same thing, but we may want to migrate other use cases over at the same time for consistency, just a thought.

This tracking is part of one of our KRs for this quarter, which we're trying to wrap up currently. So I think we may proceed with this as is and then revisit once the tracker-js PR is merged.

gtrufitt: > So once that PR is merged the only thing we'd need to do to take advantage of it is specify the component type declaratively using data-component? 

Potentially, though I feel like that should be the case now and it isn't so requires investigation.

> Currently we also pass a component ID, do you know how that fits in? (Is there another data attribute we can use?)

Ah, yes, I missed that requirement. That's interesting, and maybe something extra we would need to consider in the design of that API. Potentially you could send the ID as a part of the `data-component` and parse this out from the data-lake. 

> One piece which may not quite align is that hasBeenSeen supports debouncing, so a view isn't recorded until the component has intersected with the viewport for 200ms. 

I think the registration of components takes a visibility threshold, which is 'height of the component' - so 0.5 would be 50% of the component is intersecting rather than a `ms` time intersecting.

> This tracking is part of one of our KRs for this quarter, which we're trying to wrap up currently. So I think we may proceed with this as is and then revisit once the tracker-js PR is merged.

Makes sense, the ophan PR isn't tested (or even built).

tjmw (approved):

PR #2706, opened 2021-03-15
feat: added pollInterval option to useApi

oliverlloyd:
## What?
Adds the `pollInterval` option to `useApi`

## Usage
```typescript
    const { data, loading, error } = useApi<MyResponseType>(url, {
        pollInterval: 15000,
    });

```

## Why?
Because we want to be able to poll apis at a defined interval so that we automatically update the page with the latest data. Because `useApi` is a custom hook, returning react state, we automatically get rerenders out of the box so all we had to do here was refetch the data to make it live.

## Why this api?
The structure of an api matters. We need it to be intuitive and easy to use. Here I am following the [pattern used by the Apollo team](https://www.apollographql.com/docs/react/data/queries/#polling). They have some addition features, such as returning `startPolling` and `stopPolling` functions that we could decide to emulate if our use case were to grow in that direction.

## What about callApi
This function is exported and used by Braze for making POST calls. This is the only place in our codebase where this abstraction is used. I plan on breaking this function out from this file so that we can simplify the custom useApi hook and keep the pure api fetching logic separate. I'll do this in a follow up PR

### DIFF
--- a/src/web/lib/api.tsx
+++ b/src/web/lib/api.tsx
@@ -1,8 +1,6 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 
-// Not meant to be an exhaustive type definition of the fetch API,
-// just a starting point to get us going on 99% of our possible use cases
-interface FetchOptions {
+type FetchOptions = {
 	method?:
 		| 'GET'
 		| 'HEAD'
@@ -22,6 +20,10 @@ interface FetchOptions {
 	};
 	body?: string;
 	credentials?: 'omit' | 'include' | 'same-origin';
+};
+
+interface Options extends FetchOptions {
+	pollInterval?: number;
 }
 
 function checkForErrors(response: Response) {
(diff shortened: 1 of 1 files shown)


oliverlloyd:
[pollInterval](https://trello.com/c/aepCZg1v/2731-pollinterval)

src/web/lib/api.tsx
```diff
@@ -46,10 +48,19 @@ interface ApiResponse<T> {
...
+ * returning { loading, error, data }
+ * @param {String} url - The url to fetch
+ * @param {Object} options
+ * @param {Number} options.method - If supplied, the api will be called using this value
+ * @param {Number} options.headers - If supplied, the api will be called using this value
+ * @param {Number} options.body - If supplied, the api will be called using this value
+ * @param {Number} options.credentials - If supplied, the api will be called using this value
+ * @param {Number} options.pollInterval - If supplied, the api will be polled at this interval
+ * */
+export const useApi = <T,>(url: string, options?: Options): ApiResponse<T> => {
```
liywjl, line 63: I was unsure what the `,` was after the `T`, but found this https://stackoverflow.com/a/56989122/7378674

liywjl (approved):
LGTM, was confused by the `,` in generic, but seems to be a nicer workaround to `tsx`

src/web/lib/api.tsx
oliverlloyd: I didn't put that there, prettier did and prettier knows best 🙇

PR #2707, opened 2021-03-16
Fix 'show me the banner' helper

tomrf1:
The localstorage item was wrong. Also, it's helpful to delete the subcriptions banner item and `noRRBannerTimestamp`, which tells it not to make any more requests for a while.

I've also set the initial mvt value to 10, to avoid hitting the 'holdback' epic. (this is mainly for CODE testing)


TomPretty (approved):
LGTM 👍

PR #2708, opened 2021-03-16
feat: Added useTimeAgo custom hook

oliverlloyd:
## What?
Adds the `useTimeAgo` custom hook

This function creates a contract where `time` elements with the `data-relativeformat` and `datetime` attributes will be kept up to date in terms of the relative text they display

This custom hook depends on:

1. The `timeAgo()` function from `@guardian/libs`.
2. The `useInterval` custom hook (which is added as part of this PR)

### Usage
Set elements on the page such as
`<time datetime="2021-03-14T12:42:18.000Z" data-relativeformat='long'>17min ago</time>`

Where

`datetime` - an ISO date time string representing absolute time
`data-relativeformat` - 'short' | 'med' | 'long' - the format for the relative text

Call `useTimeAgo()` using:

```typescript
type Options = {
    interval?: number;
};
```

#### Example
```typescript
	// *****************************
	// *   Refresh relative time   *
	// *****************************
	useTimeAgo({ interval: 30000 });
```

## `useInterval`
We want to run the date check on an interval but Intervals in react hooks (and react in general) have some non intuitive behaviour which can cause issues. This [custom hook from Dan Abramov](https://overreacted.io/making-setinterval-declarative-with-react-hooks/) fixes those issues and nicely abstracts things to improve the dev ex.

## Why?
We don't want the page to lie. If a time element says 3m ago then want this to be as close to accurate as reasonably possible.

The absolute time for an event is set in the `datetime` property and does not change but the relative time we render needs to be updated but we shouldn't expect every place on the page where a relative time marker is used to reimplement the same interval logic. By grouping this into a shared lib we make it easy to improve reader UX


src/web/lib/updateRelativeDates.ts
```diff
@@ -0,0 +1,66 @@
...
+					element.dataset.interval &&
+					parseInt(element.dataset.interval, 10);
+				const interval: number =
+					elementInterval || options?.interval || DEFAULT_INTERVAL;
+
+				// Create a new interval for this element
+				intervalIds.push(
+					window.setInterval(() => {
+						let newTime;
+						switch (relativeFormat) {
```
liywjl, line 32: should we support default?

liywjl (commented):
I assume that dates are returned directly in HTML thus why we have to use DOM queries to interact with these elements?

I would prefer to just have a TS file that would return a number/string and that being used in JSX, but also understand that might not be possible.

One drawback is that we could have some layout shift caused by the innerHTML change?

src/web/lib/updateRelativeDates.ts
oliverlloyd: > I would prefer to just have a TS file that would return a number/string and that being used in JSX

I thought about that but the limitation of a doing this via jsx is I think you'd need all content hydrated? We would also not be supporting the existing contract that's in place for Frontend which is nice to do because then dates stay updated, even if the `time` elements land on the page via embeds or interatives.  Like this, so long as the contract is met, it works.

oliverlloyd:
> One drawback is that we could have some layout shift caused by the innerHTML change?

Possible. Hopefully though, given the formats being returned are all very similar and very short, this shouldn't be an issue. If it is though, I'd that's a design problem and that content needs more space given to it?

src/web/lib/updateRelativeDates.ts
oliverlloyd: > should we support default?

In this case I think ts prevents us ending up here with any other option. In any case, there's a [replacement function](https://github.com/guardian/libs/pull/134) for `makeRelativeDates` in the pipeline which will remove this model.

liywjl (approved):
👍  LGTM

nicl (commented):
Code looks really nice - tests thorough and the Abramov helper makes sense.

Because `updateTimeElements` is global, is it expected that we'd only use the `useRelativeDates` hook once at the top-level somewhere? If so, is it worth renaming to something like `initialiseRelativeDates`? (Although I know this breaks the hook naming convention.) Or perhaps just commenting that it should only be called once?

Is the plan to use this for Liveblogs?

oliverlloyd:
> Code looks really nice - tests thorough and the Abramov helper makes sense.
> 
> Because `updateTimeElements` is global, is it expected that we'd only use the `useRelativeDates` hook once at the top-level somewhere? If so, is it worth renaming to something like `initialiseRelativeDates`? (Although I know this breaks the hook naming convention.) Or perhaps just commenting that it should only be called once?

Yeah, sometimes the use* prefix requirement can lead to non perfect names. It might work better to say `useTimeAgo`? `timeAgo` is the name of the shared function which is what is _used_ here. What do you think?

> 
> Is the plan to use this for Liveblogs?

There's definately a lot of places in LiveBlogs where this will be useful but it's more of a global convention. The same convention exists already in Frontend so this is about bringing support for it into DCR.

Other places where it will be useful are on Cards and Comments.

nicl (approved):

src/web/lib/useTimeAgo.ts
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Sets up an interval in which the timeAgo function is run against every time element on the
+ * page which has the data-relativeformat attribute set
+ * @param {Number} interval - The length of time in-between time being updated
+ * */
+export const useTimeAgo = (options?: Options) => {
+	// Decide what interval to use
+	const interval: number = options?.interval || DEFAULT_INTERVAL;
+
+	useInterval(() => {
```
arelra, line 50: Would passing the function directly here work, rather than creating a new arrow function?

oliverlloyd: It would! Done 👍

PR #2710, opened 2021-03-17
Hf review padding headline

HarryFischer:
## What does this change?
Adds padding below the headline on review articles that don't have stars

### Before
image

### After
image

## Why?
Headline and standfirst too tight before


oliverlloyd (approved):
Nice

PR #2713, opened 2021-03-17
We are hiring — Use the right url

mchv:
Use the new name of the department


oliverlloyd (approved):
Good spot, I tested the url but missed the redirect to index.php

PR #2714, opened 2021-03-18
Adding link to original document for documentcloud.org embedded documents

frj:
This PR addresses a problem with the use of [DocumentCloud](https://www.documentcloud.org/) as a provider of embedded pdf documents.

We are asking central production to user DocumentCloud to host embedded pdfs instead of the current provider Scribd. This is part of the embeds consent work and we have determined that Scribd may be collecting data on users viewing their content that has been embedded in Guardian products. Using DocumentCloud will avoid the user seeing the 'click to view' placeholder that will be rendered in place of Scribd embedded content.

There is however a problem with the DocumentCloud embeds on Android, in that the gesture based scrolling and zooming doesnt work well in iframes on that platform.

Scribd works around this problem by having a button that links to the document on scribd.com where the scrolling and zooming works.

In order to achieve parity with Scribd this change adds a button that links out to DocumentCloud.

This is sadly not the whole story as on documentcloud.org scrolling works but zooming doesnt. DocumentCloud are 'working' on this.

To get round this problem we convert the url to the pdf viewer to the pdf its-self, meaning the pdf gets rendered using the native rendering on the browser which seems to work fine.

| Before | After |
|--|--|
| Screenshot 2021-03-18 at 10 15 00 | Screenshot 2021-03-18 at 10 15 11 |


HarryFischer:
Hi @frj - could we change the button to use 'subdued'? I think it looks cleaner:
image

frj:
@HarryFischer Ok sounds good.

src/web/components/elements/DocumentBlockComponent.tsx
```diff
@@ -24,6 +41,19 @@ export const DocumentBlockComponent: React.FC<{
 				allowFullScreen={true}
 				data-cy="document-embed"
 			/>
+			{source === 'DocumentCloud' ? (
+				<LinkButton
+					priority="subdued"
+					size="small"
+					href={getDocumentCloudAssetUrl(embedUrl)}
+					title={title}
+					download={true}
```
oliverlloyd, line 50: til 🎉 

I had no idea this was an option for LinkButton

oliverlloyd (approved):

liywjl (approved):
LGTM 👍

PR #2715, opened 2021-03-18
Prefill epic reminder email for signed in users

TomPretty:
## What does this change?

Prefill the epic reminder signup form for signed in users.

### DIFF
--- a/src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx
+++ b/src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx
@@ -13,6 +13,7 @@ import {
 	getLastOneOffContributionDate,
 	shouldHideSupportMessaging,
 	getArticleCountConsent,
+	getEmail,
 } from '@root/src/web/lib/contributions';
 import { getForcedVariant } from '@root/src/web/lib/readerRevenueDevUtils';
 import { CanShowResult } from '@root/src/web/lib/messagePicker';
@@ -43,10 +44,12 @@ type EpicConfig = {
 		url: string;
 		props: EpicProps;
 	};
+	email?: string;
 };
 
 type EpicProps = {
 	onReminderOpen: () => void;
+	email?: string;
 	// Also anything specified by support-dotcom-components
 };
 
@@ -97,6 +100,7 @@ type Props = {
 	isPaidContent: boolean;
 	tags: TagType[];
 	contributionsServiceUrl: string;
+	idApiUrl: string;
 };
 
 const buildPayload = async (props: Props): Promise<Metadata> => {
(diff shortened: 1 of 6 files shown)


src/web/lib/getIdapiUserData.ts
```diff
@@ -0,0 +1,20 @@
...
+		}).catch((error) => {
+			window.guardian.modules.sentry.reportError(
+				error,
+				'getIdapiUserData',
+			);
+		});
+		cache[ajaxUrl] = response;
+	}
+	return cache[ajaxUrl];
+};
```
TomPretty, line 20: @tjmw I wanted to get some feedback on this one. I've got two issues with how I've implemented it so far. 1 - should we add a better return type? Right now it's just `any` and both places this is used (`getBrazeUuid` and `getEmail` need to add some types. 2 - do we need to cache based on the URL? for a given webpage, the url is always going to be the same. Passing the url is more of a convenience for when running in dev/code/prod, but it just felt wrong to ignore this when caching.

tjmw: Hey @TomPretty nice work, I think this is looking good!

> 1 - should we add a better return type? Right now it's just any and both places this is used (getBrazeUuid and getEmail need to add some types. 

Yeah I think adding a return type here for the response data sounds good. If you wanted to avoid typing the entirety of the `/user/me` response you could create an interface containing just the pieces we're currently interested in (i.e. the brazeUuid and the email address). Because we catch errors here the return type will be something like `Promise<IdApiUser | undefined>` I think (as we may return a resolved promise with no data).

> 2 - do we need to cache based on the URL? for a given webpage, the url is always going to be the same.

Yep I think what you've done (i.e. separate cache entry for each url) is right. It's true that right now because the url comes from the CAPI data there'll only be one value for a given page request. But if that ever changes in future then this function will continue to work fine.

TomPretty: Yeah that makes a lot of sense. I've had a go at adding the return type for just the bits we're using. I tried to find the actual response schema from idapi but that was proving a little tricky! I've also dropped the `catch` as it was making the return type a little trickier to work with. It would make the promises be `Promise<IdApiUser | void>` which the subsequent chains have to deal with. 

Okay yeah nice! sounds good 👍

src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx
```diff
@@ -162,37 +167,37 @@ export const canShow = ({
...
-		});
+		idApiUrl,
+	});
+
+	const response = await getBodyEnd(
+		contributionsPayload,
+		`${contributionsServiceUrl}/epic${queryString}`,
+	);
+
+	checkForErrors(response);
```
tomrf1, line 178: `checkForErrors` throws if it's not a 200. Elsewhere in this `canShow` we return a `Promise.resolve(...)`, and also simply an object.
I'm sure it all works fine, but it would be nice to make this consistent thoughout the function.

TomPretty: discussed IRL - it's a little confusing, but okay for now!

tomrf1 (approved):
great!

PR #2716, opened 2021-03-18
Centralise logic to init automat

liywjl:
## What does this change?
- create `initAutomat` file
- replace all all `window.guardian.automat` logic with `initAutomat`
- add `module` and `meta` to useEffect as they are being explicitly used in use Effect

## Why?
To avoid setting automat several times, we should centralise it. Also it helps us have 1 central place to make changes (which is what we plan on doing shortly)


shtukas (approved):

oliverlloyd:
Closing as stale

PR #2719, opened 2021-03-18
Revert "Revert "ARM/Graviton testing""

nicl:
Reverts guardian/dotcom-rendering#2684. We have now reserved our instances so can turn this on.

We still don't have a solution to disk-usage monitoring but I don't think that's a blocker. Any solution we do adopt will likely be outside of the DCR node process itself (perhaps using Fluentbit).


jfsoul (approved):
Shall we just delete the commented out code now?

PR #2721, opened 2021-03-19
feat: Log renderTime time on the server

oliverlloyd:
## What does this change?
Logs the time taken to render each request. To enable this I've refactored `logger` to accept an object.

### Before
`logger.info(event: any)` // > This really only worked if event was a string

### After
`logger.info(event: string | Object)` // Detects the type of the input and logs the event accordingly, spreading the object to use all fields.

## Why?
So we can track how long DCR is taking to render articles


oliverlloyd:
[Node Performance SSR logging](https://trello.com/c/Ta73k5rj/2711-node-performance-ssr-logging)

src/app/logging.ts
```diff
@@ -13,15 +13,31 @@ const logLocation =
...
 		stack: 'frontend',
 		app: 'dotcom-rendering',
 		stage: 'CODE',
 		'@timestamp': logEvent.startTime,
 		'@version': 1,
 		level: logEvent.level.levelStr,
 		level_value: logEvent.level.level,
-		message: logEvent.data.join(','),
+	};
+	// Loagstash uses any[] to type data but we want to coerce it here
```
nicl, line 25: Minor, but it's not Logstash I think.

nicl (approved):
🐌

src/app/logging.ts
oliverlloyd: Oh, it's log4js - I was close!

package.json
```diff
@@ -202,8 +203,10 @@
         "node-fetch": "^2.6.1",
         "nodemon": "^2.0.7",
         "npm-run-all": "^4.1.5",
+        "perf_hooks": "^0.0.1",
```
liywjl, line 206: what is this for??

liywjl (changes requested):
LGTM, just remove the `perf_hooks` dependency

liywjl:
Has this been tested on CODE?

package.json
oliverlloyd: But for reference, it's the node lib that allows you to use a version of `window.performance` on the server. It was an option I was trying out before using the method here.

oliverlloyd:
> Has this been tested on CODE?

Yep. If you search the `frontend-*` index for 'renderTime' you should see entries from CODE

package.json
liywjl: The recent press on substitution attacks make me a little wary of such package names
https://github.blog/2021-02-12-avoiding-npm-substitution-attacks/

liywjl (approved):
👍

PR #2722, opened 2021-03-19
feat: Live counts

oliverlloyd:
## What does this change?
Comment counts are now "live". Where live is as live as you can get when discussion data is cached for 30 seconds

### Before
Counts were a snapshot in time when the article was rendered

### After
Counts are kept up to date

## Why?
Because the little things matter. And also because seeing that number change can draw the eye and engage readers more.


oliverlloyd:
[Live Counts](https://trello.com/c/twDqBTSA/2749-live-counts)

liywjl (approved):
👍

PR #2724, opened 2021-03-19
Main media unsafe EmbedBlockElement

liywjl:
## What does this change?
Add support for unsafe main media embed block elements

### Before
Screenshot 2021-03-19 at 17 18 35

### After
Screenshot 2021-03-19 at 17 06 32

## Why?
`EmbedBlockElement` iframe resize no longer needed as we know what main media height should be

### DIFF
--- a/src/web/lib/ElementRenderer.tsx
+++ b/src/web/lib/ElementRenderer.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { css } from 'emotion';
 
 import { BlockquoteBlockComponent } from '@root/src/web/components/elements/BlockquoteBlockComponent';
 import { CalloutBlockComponent } from '@root/src/web/components/elements/CalloutBlockComponent';
@@ -201,6 +202,25 @@ export const ElementRenderer = ({
 			);
 		case 'model.dotcomrendering.pageElements.EmbedBlockElement':
 			if (!element.safe) {
+				if (isMainMedia) {
+					return (
+						<Figure
+							isMainMedia={isMainMedia}
+							role={isLiveBlog ? 'inline' : element.role}
+							id={element.elementId}
+						>
+							<iframe
+								className={css`
+									width: 100%;
+									height: 100vh;
+								`}
+								title={element.alt || ''}
+								srcDoc={element.html}
+							/>
+						</Figure>
+					);
+				}
+
 				return (
 					<Figure
 						isMainMedia={isMainMedia}
--- a/src/web/lib/ElementRenderer.tsx
+++ b/src/web/lib/ElementRenderer.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import { css } from 'emotion';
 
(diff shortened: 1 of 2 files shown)


src/web/lib/ElementRenderer.tsx
```diff
@@ -201,6 +202,25 @@ export const ElementRenderer = ({
 			);
 		case 'model.dotcomrendering.pageElements.EmbedBlockElement':
 			if (!element.safe) {
+				if (isMainMedia) {
+					return (
+						<Figure
+							isMainMedia={isMainMedia}
+							role={isLiveBlog ? 'inline' : element.role}
```
oliverlloyd, line 209: You don't need this for main media content, you can just pass role

src/web/lib/ElementRenderer.tsx
```diff
@@ -201,6 +202,25 @@ export const ElementRenderer = ({
...
+							id={element.elementId}
+						>
+							<iframe
```
oliverlloyd, line 212: +1 for breaking this out of `UnsafeBlockComponent` - we don't need the iframe resizer here.

But I feel that semantically, ElementRenderer should be rendering an element _component_. Just because that's what it does everywhere else.

I suppose you could have a new type of element, like a MainMediaEmbedBlockComponent, but ooof, that's long!

Seems like the list of different Embeds is growing. We've got `UnsafeEmbedBlockComponent`, `EmbedBlockComponent` and now , potentially, `MainMediaEmbedBlockComponent`. I think there's a case here to refactor this somehow. Either we look at splitting types higher up, in Frontend or we could bring the variations down into `EmbedBlockComponent`. Not sure what the right approach is but I think we've hit the point when a refactor would be a good idea.

shtukas (approved):

src/web/lib/ElementRenderer.tsx
liywjl: In this case I do not feel we need to split this out into it own component, but I sense the main reason is not to import `css` here and give the idea that we should be donig styling in this file (so I will spin it out).

Personally I feel like `ElementRenderer.tsx` could be split, one file for Main media rendering, the other for body content, as there are lots of components that would not make sense to be rendered as Main media (ex: SpotifyBlockElement, TimelineBlockElement or QuizAtomBlockElement to name a few)

oliverlloyd: > Personally I feel like ElementRenderer.tsx could be split

Actually, the reason why [I grouped them](https://github.com/guardian/dotcom-rendering/pull/2587) was precisely because people _do_ use strange elements in MainMedia 🤷

oliverlloyd: And yeah, importing css into ElementRenderer feels like a coupling we don't want to make - it trues creating a component is excessive for just this small amount of jsx code but the pattern is strong in this file and once you break it it's lost and then css will slowly spread everywhere.

oliverlloyd (approved):

PR #2731, opened 2021-03-22
Increase mandatory gate audience to 30%

QuarpT:
## What does this change?

Increase mandatory gate audience to 30%

### Before

Audience was 20%

### After

Audience is 30%

## Why?

We want to reach 170K views of the variant by March 29

### DIFF
--- a/src/web/experiments/tests/sign-in-gate-mandatory.ts
+++ b/src/web/experiments/tests/sign-in-gate-mandatory.ts
@@ -7,8 +7,8 @@ export const signInGateMandoryTest: ABTest = {
 	author: 'Peter Colley',
 	description:
 		'Compare mandatory gate (an article sigin gate without the dismiss button) with the main signin gate',
-	audience: 0.2,
-	audienceOffset: 0.7,
+	audience: 0.3,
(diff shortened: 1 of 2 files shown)


liywjl (approved):
👍

src/web/experiments/tests/sign-in-gate-mandatory.ts
```diff
@@ -7,8 +7,8 @@ export const signInGateMandoryTest: ABTest = {
 	author: 'Peter Colley',
 	description:
 		'Compare mandatory gate (an article sigin gate without the dismiss button) with the main signin gate',
-	audience: 0.2,
-	audienceOffset: 0.7,
+	audience: 0.3,
+	audienceOffset: 0.6,
```
liywjl, line 11: just curious if it is intended that both these numbers do not reach 1?

QuarpT: it's intended, the main variant uses 0.0-0.6, the mandatory experiment uses 0.6-0.9 and the long running control uses 0.9-1.0

PR #2732, opened 2021-03-23
Reactive stars

mxdvl:
## What does this change?

Make star SVG component a first-class React citizen. Remove unused code and ensure that `id` is _more_ unique.

This fixes an issue in onward content cards where the `clip-path` might not be displayed.

### Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/112175694-d69b7180-8bcd-11eb-9855-5c25e08fe3ba.png
[after]: https://user-images.githubusercontent.com/76776/112175827-f337a980-8bcd-11eb-9357-d7dddc0d489b.png

## Why?

Follow-up and improvement from #2479.

### DIFF
--- a/src/amp/components/StarRating.tsx
+++ b/src/amp/components/StarRating.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { css, cx } from 'emotion';
-import Star from '@frontend/static/icons/star.svg';
+import { Star } from '@frontend/static/icons/Star';
 import { palette } from '@guardian/src-foundations';
 
 const ratingsWrapper = css`
@@ -24,11 +24,6 @@ const smallSize = css`
 	}
 `;
 
-const emptyStar = css`
-	fill: transparent;
-	stroke: ${palette.neutral[7]};
-`;
-
 export const StarRating: React.FC<{
 	rating: number;
 	size: string;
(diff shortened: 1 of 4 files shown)


src/amp/components/StarRating.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { css, cx } from 'emotion';
-import Star from '@frontend/static/icons/star.svg';
+import { Star } from '@frontend/static/icons/Star';
```
oliverlloyd, line 3: I was going to ask to move this into the /components folder but then I realised it's more discoverable here?

src/static/icons/Star.tsx
```diff
@@ -0,0 +1,37 @@
+import { neutral } from '@guardian/src-foundations/palette';
+import React from 'react';
+
+export const Star = ({
+	starId = '000',
```
oliverlloyd, line 5: I _think_ we don't need this default?

oliverlloyd (approved):
⭐ ⭐ ⭐ ⭐ ⭐ s!

src/amp/components/StarRating.tsx
mxdvl: Yes?

src/static/icons/Star.tsx
mxdvl: I think it would make it easier to debug, and allow consumers not to worry too much about what they’d use as an ID.

liywjl (approved):
LGTM

PR #2735, opened 2021-03-23
Add SP domain as preconnect link to the head of the page

ioanna0:
## What does this change?
This adds Sourcepoint's domain URL (`https://sourcepoint.theguardian.com`) as a preconnect link to the head of the page. 

### Before
N/A

### After
N/A

## Why?
We know we will definitely call this domain's URL for all SP requests in all regions. This should allow establish an early connection with this domain and will speed up the page load.


liywjl (approved):
👍

arelra (approved):

src/web/server/htmlTemplate.ts
```diff
@@ -110,32 +115,32 @@ export const htmlTemplate = ({
...
+            GGGGGGGG            GGGGGGGGGGGGG           GGGGGGGGGGGGG
+             GGGGGGGG            GGGGGGGGGGGG           GGGGGGGGGGGG
+              GGGGGGGGG           GGGGGGGGGGG           GGGGGGGGGGG
+                GGGGGGGGGG         GGGGGGGGGG           GGGGGGGGG
+                  GGGGGGGGGGG        GGGGGGGG        GGGGGGGGGG
+                    GGGGGGGGGGGGGG      GGGGG  GGGGGGGGGGGGGG
+                       GGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGG
+                            GGGGGGGGGGGGGGGGGGGGGGGGG
+                                    GGGGGGGGG
+
```
mxdvl, line 143: Won’t this break the rendering?

In any case it’s wonderfully lean.

mxdvl (approved):

src/web/server/htmlTemplate.ts
oliverlloyd: Looks okay to me 👍 

Screenshot 2021-03-23 at 17 33 42

oliverlloyd: That said, for the sake of a tidy git history we should probably remove these changes. Not sure how they got here, is prettier getting overzelous again?

mxdvl: I think @ioanna0 uses IntelliJ, could this be it?

ioanna0: Arghh, I also though it was because I'm using IntelliJ but VS also changes it on Save. Could it be some prettier settings have changed since the last time this file was edited?

src/web/server/htmlTemplate.ts
```diff
@@ -99,7 +99,12 @@ export const htmlTemplate = ({
 		`https://support.theguardian.com`,
 	];
 
-	const preconnectTags = staticPreconnectUrls.map(
+	const allStaticPreconnectUrls =
+		process.env.NODE_ENV === 'production'
```
jfsoul, line 103: What happens in other environments with sourcepoint?

ioanna0: another domain is used: `https://cdn.privacy-mgmt.com`

jfsoul: Ah ok. I don't suppose it's a major problem, but `NODE_ENV` will be `'production'` in CODE as well. You may already know, but the value may be interpreted by a number of frameworks and their behaviour changed accordingly. So it's only _not_ `'production'` in DEV - as we want CODE to largely mirror PROD in behaviour.

mxdvl:
I think this is still valuable, and should be merged. Ioanna is no longer with us, so we’ll have to merge it ourselves.

cc. @arelra, @liywjl do you see any issue?

liywjl (approved):
LGTM

arelra:
LGTM

PR #2736, opened 2021-03-23
Call puzzles banner for puzzles pages

lucymonie:
## What does this change?
This adds a call to the puzzles banner for puzzles and sudoku pages, with some checks as to whether the banner should be shown or not.

### Before
Screen Shot 2021-03-23 at 16 57 14

### After
Screen Shot 2021-03-23 at 16 55 43

## Why?
We are trying to get more puzzles users to discover the Guardian puzzles app.


jfsoul:
Hello! What's the rationale for this not being another version of the `ReaderRevenueBanner`?

lucymonie:
> Hello! What's the rationale for this not being another version of the `ReaderRevenueBanner`?

Good question @jfsoul! I did go back and forth a bit on this. The puzzles banner has its own endpoint, and needs to go ahead of the RR banner in priority order. It seemed as though it would make the existing banner code quite a bit more complicated because of the different circumstances in which it is called. However, there is an argument in the other direction, especially as (for the moment) it generates the same payload.

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
```diff
@@ -0,0 +1,238 @@
...
+
+			sendOphanComponentEvent('VIEW', meta);
+
+			// track banner view event in Google Analytics for subscriptions banner
+			if (componentType === 'ACQUISITIONS_PUZZLES_BANNER') {
+				trackNonClickInteraction('puzzles-banner : display');
+			}
+		}
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [hasBeenSeen]);
```
liywjl, line 217: you depend on `meta`

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
```diff
@@ -0,0 +1,238 @@
...
+			.catch((error) =>
+				// eslint-disable-next-line no-console
+				console.log(`banner - error is: ${error}`),
+			);
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, []);
+
+	// Should only run once
+	useEffect(() => {
+		if (hasBeenSeen && meta) {
```
liywjl, line 204: `meta` is set to always be defined
https://github.com/guardian/dotcom-rendering/pull/2736/files#diff-c71689108b52189dab75a54e7f3de25e677987bd0f09d0b30c1a841f136dfb83R164

therefore adding `if(meta)` doesn't make much sense here? unless it isnt always defined? (I am away this is an antipatter used elsewhere)

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
```diff
@@ -0,0 +1,238 @@
...
+	if (!response.ok) {
+		throw Error(
+			response.statusText ||
+				`SlotBanner | An api call returned HTTP status ${response.status}`,
+		);
+	}
+	return response;
+};
+
+type HasBeenSeen = [boolean, (el: HTMLDivElement) => void];
```
liywjl, line 34: ditto, would prob not add an extra type here

liywjl (commented):
I notice a lot of similarities with this and `ReaderRevenueBanner` that I think would make sense for us to try unifying.
We are busy updating to Emotion 11, which is causing use to have to change how we interact with `automat` as there is no way to communicate version changes. 

`ReaderRevenueBanner` now supports version numbers https://github.com/guardian/support-dotcom-components/pull/400 and this allows us some way of creating a contract between `automat` and other libs. 

The puzzel banner would need to support something similar or it might be simpler to integrate with `ReaderRevenueBanner`

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
```diff
@@ -0,0 +1,238 @@
...
+	getArticleCountConsent,
+} from '@root/src/web/lib/contributions';
+import {
+	sendOphanComponentEvent,
+	submitComponentEvent,
+} from '@root/src/web/browser/ophan/ophan';
+import { getCookie } from '@root/src/web/browser/cookie';
+import { getZIndex } from '@root/src/web/lib/getZIndex';
+import { trackNonClickInteraction } from '@root/src/web/browser/ga/ga';
+import { WeeklyArticleHistory } from '@root/node_modules/@guardian/automat-client/dist/types';
```
oliverlloyd, line 21: Is this right? This might be an automat thing but it sure looks odd so I thought I'd check!

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
```diff
@@ -0,0 +1,238 @@
...
+	hasConsentedToArticleCounts: boolean;
+};
+
+type CanShowProps = BaseProps & {
+	asyncCountryCode: Promise<string>;
+	remotePuzzlesBannerConfig: boolean;
+};
+
+// TODO specify return type (need to update client to provide this first)
+const buildPayload = (props: BuildPayloadProps) => {
```
oliverlloyd, line 63: If this is only used in one place we might as well inline it? Saves a bit of scrolling around

oliverlloyd, line 63: There's a convention of destructuring props in DCR. I get that it's subjective and perhaps not a huge issue but I think once a code style has been established it's good to maintain it.

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
```diff
@@ -0,0 +1,238 @@
...
+	return {
+		tracking: {
+			ophanPageId: window.guardian.config.ophan.pageViewId,
```
oliverlloyd, line 66: I'm not sure we shouold be accessing the `window` object here. `pageViewId` is a property on the CAPI object so it's available as a prop. The problem with not using props and starting to use the global `window` object is... well, I guess it's global state?

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
```diff
@@ -0,0 +1,238 @@
...
+): Promise<Response> => {
+	const json = JSON.stringify(meta);
+	return fetch(url, {
+		method: 'post',
+		headers: { 'Content-Type': 'application/json' },
+		body: json,
+	});
+};
+
+export const canShow = async ({
```
oliverlloyd, line 99: Would it be possible to export this function from it's own file? I think the `PuzzleBanner` component should only be that and then if you have another file calls `lib/canShow.ts` then that is it's own stand alone (perhaps more easilly tested?) thing

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
```diff
@@ -0,0 +1,238 @@
...
+	isSensitive,
+	tags,
+	contributionsServiceUrl,
+	alreadyVisitedCount,
+	engagementBannerLastClosedAt,
+	subscriptionBannerLastClosedAt,
+}: CanShowProps): Promise<CanShowResult> => {
+	const isPuzzlesPage =
+		window.guardian.config.page.section === 'crosswords' ||
+		window.guardian.config.page.series === 'Sudoku';
```
oliverlloyd, line 117: These should be props as they are part of the CAPI object

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
```diff
@@ -0,0 +1,238 @@
...
+			.then(checkForErrors)
+			.then((response) => response.json())
+			.then((json: { data?: any }) => {
+				if (!json.data) {
+					return { result: false };
+				}
+
+				const { module, meta } = json.data;
+
+				return { result: true, meta: { module, meta } };
```
oliverlloyd, line 156: This function is called `canShow` but it doesn't return a boolean. Maybe this logic can be broken down into  two steps, one function to check the logic at the start, returning a boolean, and then another which makes the fetch call?

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
```diff
@@ -0,0 +1,238 @@
...
+			subscriptionBannerLastClosedAt,
+			hasConsentedToArticleCounts,
+		});
+		return getPuzzlesBanner(
+			bannerPayload,
+			`${contributionsServiceUrl}/puzzles`,
+		)
+			.then(checkForErrors)
+			.then((response) => response.json())
+			.then((json: { data?: any }) => {
```
oliverlloyd, line 149: Can we replace this with `useApi`? There's a lot of logic being duplicated here that's already set there. We're already using other custom hooks here so I think the pattern should be okay?

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
```diff
@@ -0,0 +1,238 @@
...
+			})
+			.catch((error) =>
+				// eslint-disable-next-line no-console
+				console.log(`banner - error is: ${error}`),
+			);
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, []);
+
+	// Should only run once
+	useEffect(() => {
```
oliverlloyd, line 203: If you only want a useEffect to only run once you can use our custom hook `useOnce` 🎉

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -106,6 +146,11 @@ export const StickyBottomBanner = ({
 	const [SelectedBanner, setSelectedBanner] = useState<React.FC | null>(null);
 	useOnce(() => {
 		const CMP = buildCmpBannerConfig();
+		const puzzlesBanner = buildPuzzlesBannerConfig(
+			CAPI,
```
oliverlloyd, line 150: Not sure if this is withing the sccope of this PR or not but we have a convention in DCR of not passing `CAPI`. CAPI is Haauuuge, like Killobytes and Kilobytes of json. So passing it as a prop is a drain on performance and it's also a lot easier to reason about what  thing is doing when you can explicitly see what inputs it has.

oliverlloyd (commented):
I can't speak to the actual banner implementation but I made some ccomments around conventions. Sorry, I know these kinds of 'but we don't do it that way!' reviews are annoying but hopefully our conventions are based on sound principles - or at least we try to make sure they are!

lucymonie:
Thanks so much for all the feedback @liywjl - we are thinking all this over and will make some changes for you to review. 👍🏻

liywjl:
@lucymonie I have moved `automat` into its own hooks implementation (to avoid use re-setting it multiple times)
https://github.com/guardian/dotcom-rendering/pull/2775/files#diff-c2ac20e541fdf6a7348d335c5ce7cb9863f9654d82df540678f07c0a564821edR8

i-hardy:
Hi all, sorry we’ve let this one sit around for so long. We really appreciate all the feedback, but a lot of this relates to old code that’s essentially just been lifted from the existing `ReaderRevenueBanner.tsx` file- I’m taking this as an indication that we have too much repetition of code, so I’m working on moving everything into the same file. I’ll take into account as much of the feedback as I can, but some may end up out of scope for the changes we actually need to make and should probably be tackled a later refactor.

I should hopefully have something ready for review in another day or so.

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
i-hardy: I think it's a bit too far out of scope for this PR unfortunately- I had a look at it but it would involve changing so much of the code that it seems like it would be better as something for a separate PR.

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
i-hardy: You can't use hooks outside of a function component, unfortunately. It would be good to have a similar utility to be used in non-component contexts like this.

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
```diff
@@ -0,0 +1,238 @@
...
+		window.guardian.config.page.section === 'crosswords' ||
+		window.guardian.config.page.series === 'Sudoku';
+
+	if (shouldHideReaderRevenue) {
+		// We never serve Reader Revenue banners in this case
+		return { result: false };
+	}
+
+	if (isPuzzlesPage && remotePuzzlesBannerConfig) {
+		const countryCode = await asyncCountryCode;
```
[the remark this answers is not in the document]
i-hardy: `asyncCountryCode` is a `Promise<string>`, it's passed in from `App.tsx`.

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
i-hardy: Agreed, but I can't seem to find the equivalent of the `pageViewId` on the CAPI object.

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
i-hardy: I agree the names aren't great, but I feel like this would be seriously out of scope- it would involve changing a lot of code unrelated to the puzzles banner.

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -157,96 +168,164 @@ export const canShow = async ({
...
 
-type Props = {
+export type BannerProps = {
 	meta: any;
 	module: { url: string; name: string; props: any[] };
 };
 
-export const ReaderRevenueBanner = ({ meta, module }: Props) => {
-	const [Banner, setBanner] = useState<React.FC>();
+const createReaderRevenueBannerComponent = (
```
liywjl, line 255: I read this component more as just a Base Banner component than a Reader Revenue Banner? Maybe it might be better to use a different name to indicate it is a more generic component?

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -157,96 +168,164 @@ export const canShow = async ({
...
+	componentTypeName: ReaderRevenueComponent,
+	displayEvent: string,
+): React.FC<BannerProps> => {
+	return ({ meta, module }: BannerProps) => {
```
liywjl, line 259: I would favour using a classic component pattern over a `HOC` pattern in this case, as technically, we are just passing through props rather than enhancing a component.

Example:
```
const BaseBannerComponent = ({
	componentTypeName: ReaderRevenueComponent,
	displayEvent: string,
	meta: any,
	module: { url: string; name: string; props: any[] },
}) => ...
```

It means that `ReaderRevenueBanner` could be used like:
```
export const ReaderRevenueBanner = ({ meta, module }) => 
	<BaseBannerComponent
		componentTypeName="ReaderRevenueComponent"
		displayEvent="subscription-banner : display"
		meta={meta}
		module={module}
	/>
```

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
```diff
@@ -50,42 +60,64 @@ const buildCmpBannerConfig = (): CandidateConfig => ({
...
+						engagementBannerLastClosedAt: getBannerLastClosedAt(
+							'engagementBannerLastClosedAt',
+						),
+						subscriptionBannerLastClosedAt: getBannerLastClosedAt(
+							'subscriptionBannerLastClosedAt',
+						),
+						section: CAPI.config.section,
+					}),
+				/* eslint-disable-next-line react/jsx-props-no-spreading */
+				show: (meta: any) => () => <BannerComponent {...meta} />,
```
liywjl, line 100: isn't `meta` related to `BannerProps`? (which itself contains `meta` prop, which is a little confusing).

If we know what the props are, I think its best to favor directly using the props than spreading them

liywjl (commented):
The PR looks a lot clearer. I added some points that I think could benefit readability, but overall this looks great.

I haven't tested it locally; it would be cool to add a way to force the banner to appear to add cypress tests?

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
i-hardy: Makes sense! 👍

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
i-hardy: I don't want to make it _too_ generic in case that might create confusion with the Braze banners which work in a very different way, but this is the pattern we'd use for anything coming from the `support-dotcom-components` service. Maybe it's a `RemoteBanner`?

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -73,36 +78,42 @@ const buildPayload = (props: BuildPayloadProps) => {
...
+		.then((response: Response) => {
+			if (!response.ok) {
+				throw Error(
+					response.statusText ||
+						`SlotBanner | An api call returned HTTP status ${response.status}`,
+				);
+			}
+			return response;
+		})
+		.then((response) => response.json());
```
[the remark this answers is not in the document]
i-hardy: Errors in any candidate message's `canShow` function should be caught by the message picker- here https://github.com/guardian/dotcom-rendering/blob/53198874e700b66577727fe97f12de9f69b5aff5/src/web/lib/messagePicker.ts#L55

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
oliverlloyd: You're right. Sorry, I was thinking of `pageId`!

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -203,29 +285,23 @@ export const ReaderRevenueBanner = ({ meta, module }: Props) => {
...
-			);
-		// eslint-disable-next-line react-hooks/exhaustive-deps
+			.catch((error) => console.log(`banner - error is: ${error}`));
 	}, []);
 
-	// Should only run once
-	useEffect(() => {
-		if (hasBeenSeen && meta) {
+	useOnce(() => {
+		if (hasBeenSeen) {
```
oliverlloyd, line 292: You don't need this check here. The way `useOnce` works is it will not run until all dependencies are truthy. So if code execution reaches this point `hasBeenSeen` is guarenteed to be true.

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
oliverlloyd: Fair enough. You've got a better overview on what code is being used where so I'm happy to defer to you on that

src/web/components/StickyBottomBanner/PuzzlesBanner.tsx
oliverlloyd: Ah, I see. To be honest, I find it a little hard to follow the rendering logic for a banner, I'm more used to everything being declarative in the classic react style, but I can see that this isn't the case here so yes, you're right, a reat hook won't work

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
oliverlloyd: Sure, I don't want to hold up this PR if that refactor is out of scope. It would be great to move away from this pattern at some point though. `CAPI` weighs in at 17Kb...

The pattern we've used to date has been to extract all the required fields out of CAPI in `App.tsx` and then prop drill them down from there.

Looking at things, there are actually only two places where this rule is not followed, `StickyBottomBanner` and `SignInGateSelector` and I think it would be simple to replace the `CAPI` prop with some more explicit ones. I'm happy to pick this up if you like? I quite enjoy tidy up tasks 😃 Once this one is merged I'll pick it up

oliverlloyd (approved):
Really great stuff! And thanks for holding in there during the (hopefully useful!) review.

I left a couple of small comments, nothing major though.

Happy to merge this one once @liywjl is happy too 👍

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
liywjl: Remote banner sounds like a good name, I am just keen to drop `ReaderRevenue` from the name

liywjl (approved):
Thank you for all the updates, this looks great!

src/web/components/StickyBottomBanner/StickyBottomBanner.tsx
i-hardy: If you've got the time to take a look at it that would be great! Otherwise I'd be happy to stick it in our backlog as a health task.

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
i-hardy: Thanks!

PR #2737, opened 2021-03-23
Replace gifting CTA with supporter CTA in header

GHaberis:
## What does this change?

Do not merge until https://github.com/guardian/frontend/pull/23630/ has been released on `frontend`. (edit: this has now been released). 

**Note:** The vast majority of files changed here are test fixtures. I had to update them to replace the `gifting` property with a `supporter` property to reflect the change to the `ReaderRevenueCategories` type, you can therefore skip over any `fixtures/*` changes.

This PR replaces the "Gift options" link in the Header with a new Supporter call-to-action, a "Subscriptions" link.  The "Gift options" link was only ever shown to existing supporters, and we've retained this rule for the new Supporter call-to-action.

### Additional changes

Like other Reader Revenue links the Supporter CTA needs it's own set of tracking properties so we can measure user interactions with this element. The code that generates Reader Revenue URLs with the correct tracking properties exists on `frontend`, the constructed URLs are passed in the data model to `dotcom-rendering`. I've made the changes on `frontend` to pass the `supporter` property in the data model in this https://github.com/guardian/frontend/pull/23630. In this PR I therefore had to update the `ReaderRevenueCategories` type to replace the now redundant `gifting` property with the `supporter` property.

### Before

Screenshot 2021-03-23 at 17 08 06

### After

Screenshot 2021-03-23 at 17 12 00


oliverlloyd (commented):
Just to check, did you update those article fixtures using `make gen-fixtures` or manually? If manually then there's a good chance you changes will be lost (they will get overwritten next time they are auto updated)

GHaberis:
> Just to check, did you update those article fixtures using `make gen-fixtures` or manually? If manually then there's a good chance you changes will be lost (they will get overwritten next time they are auto updated)

@oliverlloyd I've actually just committed generated versions created using that script, although yes I originally updated them with a search & replace.

GHaberis:
> > Just to check, did you update those article fixtures using `make gen-fixtures` or manually? If manually then there's a good chance you changes will be lost (they will get overwritten next time they are auto updated)
> 
> @oliverlloyd I've actually just committed generated versions created using that script, although yes I originally updated them with a search & replace.

@oliverlloyd the generated fixtures were causing the validation steps to fail: `Type '{ contribute: string; subscribe: string; support: string; gifting: string; supporter: string; }' is not assignable to type 'ReaderRevenueCategories'.
  Object literal may only specify known properties, and 'gifting' does not exist in type 'ReaderRevenueCategories'.`

It looks like the generated fixtures are built using the data model output by `frontend` in PROD. The data model sent from `frontend` still contains a `gifting` property, although I intend to remove this, before I can remove this property I thought I'd need to remove the dependency `dotcom-rendering` has on the property being present in the model - which is partly what this PR did by removing it from the `ReaderRevenueCategories` interface, but doing this caused the validation issue.  I've therefore had to restore gifting prop in ReaderRevenueCategories but now as an optional string. This has fixed the validation issues and should mean I can remove it from the data model sent by `frontend` without causing issues.

oliverlloyd:
> Before I can remove this property I thought I'd need to remove the dependency `dotcom-rendering` has on the property being present in the model

Hey George,

Yeah, we have that classic dependency PR problem where if you change one thing on Frontend you need a sibling PR in DCR. (Monorepo? 🤷 ) Normally it's less painful because we're only adding new properties so we can get away with first making the change in Frontend and then updating DCR later. But if the property is changing then making the it optional seems like the best solution here 👍

GHaberis:
@oliverlloyd OK great, just tested this on `CODE` and it's working as expected 👍

oliverlloyd (approved):
Thanks for taking the time to implement this solution well. Having the right naming here will hopefully be more useful and flexible in future (less gift specific).

Sorry about the fixture fun! We tie our fixtures tightly to production which causes these sorts of problems but the benefits of having very realistic test coverage is - I think - worth it.

PR #2738, opened 2021-03-24
Labs Byline

oliverlloyd:
## What?
Use a different font for card headline bylines when the theme is `Special.Labs`

Screenshot 2021-03-24 at 08 09 27

## Why?
Labs are special

### DIFF
--- a/src/web/components/Byline.tsx
+++ b/src/web/components/Byline.tsx
@@ -1,41 +1,78 @@
 import React from 'react';
 import { css, cx } from 'emotion';
 
-import { headline } from '@guardian/src-foundations/typography';
+import { headline, textSans } from '@guardian/src-foundations/typography';
 import { until } from '@guardian/src-foundations/mq';
+import { Special } from '@guardian/types';
 
 type Props = {
 	text: string;
 	palette: Palette;
+	format: Format;
 	size: SmallHeadlineSize;
 };
 
-const bylineStyles = (size: SmallHeadlineSize) => {
+const bylineStyles = (size: SmallHeadlineSize, format: Format) => {
+	const baseStyles = css`
+		display: block;
+		font-style: italic;
+	`;
+
 	switch (size) {
-		case 'large':
+		case 'large': {
+			if (format.theme === Special.Labs) {
+				return css`
+					${baseStyles};
+					${textSans.large()};
+					font-size: 24px;
+					line-height: 24px;
+					${until.desktop} {
+						${textSans.large()};
+						line-height: 20px;
+					}
+				`;
+			}
 			return css`
-				display: block;
-				font-style: italic;
+				${baseStyles};
 				${headline.xsmall()};
 				${until.desktop} {
 					${headline.xxsmall()};
 				}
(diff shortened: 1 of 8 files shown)


oliverlloyd:
[Labs Font](https://trello.com/c/Sxbsdlka/2669-labs-font)

src/web/components/Byline.tsx
```diff
@@ -1,41 +1,78 @@
...
+	`;
+
 	switch (size) {
-		case 'large':
+		case 'large': {
+			if (format.theme === Special.Labs) {
+				return css`
+					${baseStyles};
+					${textSans.large()};
+					font-size: 24px;
```
jfsoul, line 27: is `textSans.large()` not giving us font size already?

src/web/components/Byline.tsx
```diff
@@ -1,41 +1,78 @@
...
+const bylineStyles = (size: SmallHeadlineSize, format: Format) => {
+	const baseStyles = css`
+		display: block;
+		font-style: italic;
+	`;
+
 	switch (size) {
-		case 'large':
+		case 'large': {
+			if (format.theme === Special.Labs) {
```
jfsoul, line 23: Would a LabsByline component be worth it on balance?

jfsoul (approved):

src/web/components/Byline.tsx
oliverlloyd: It's not large enough 😞 

The issue is Labs is using `textSans` like it's a headline, but it wasn't designed for that

jfsoul: 😞

src/web/components/Byline.tsx
oliverlloyd: Possibly, although we have this convention (passing in `format` and branching on that) in quite a few places now so it's probably easier for future refactors if this is consistent.

I agree though, it's not ideal. I think the problem here is that suddenly we're branching on font - something we _never_ had to think about before, so it's a bit jarring. Maybe we could think about some sort of `decideFont()` apprpoach? We'd need to account for the kinds of variations ^^ where there are multiple `sizes`, but perhaps that could be standardised?

One nice advantage of centralising font decisions is you could use that logic to work out _exactly_ the fonts you want on a particular page, allowing us to selectivly load them - which would be pretty cool

jfsoul: Yeah, selective loading would be pretty great - although realistically I wonder whether most views would have all the fonts cached already. Still a benefit though.

Doesn't need to block this going out, mind you.

oliverlloyd: > most views would have all the fonts cached already

Tangent: We typically solve the problem of loading the page with an empty cache and don't really think about the second page view experience. I guess we just assume it will be better. But maybe there are ways we can _make_ it better. Or maybe we need to think about how much of our traffic is from readers that have already visited the site? I'm not sure if even a large number here would or should change how we approach performance but it's perhaps nice to have the data to think about it.

PR #2740, opened 2021-03-24
Stop long word edge cases overflowing in narrow viewports

nitro-marky:
## What does this change?

In extremely rare cases (usually on mobile and when URLs are written in the article) they can break out of the viewport. This fix should hopefully stop that from happening.

This is only set until tablet at which point it shouldn't cause a problem.

### Before

Screenshot 2021-03-24 at 10 35 32

### After

Screenshot 2021-03-24 at 10 36 27


shtukas (approved):

nitro-marky:
[Really long words break layout](https://trello.com/c/ir4ceQbu/2522-really-long-words-break-layout)

PR #2741, opened 2021-03-24
Making requested design tweaks to click to view component

frj:
This PR contains some changes suggested after review by the designers.

For details see the figma file:

Screenshot 2021-03-24 at 14 37 15

https://www.figma.com/file/Ji1XENAJJjzcxwwwfWZnJc/Embed-consent?node-id=221%3A8

Before | After
---|---
Screenshot 2021-03-24 at 14 32 38 | Screenshot 2021-03-24 at 14 30 36


shtukas (approved):

src/web/components/ClickToView.tsx
```diff
@@ -137,19 +143,19 @@ export const ClickToView = ({
...
 					width: 100%;
 					background: ${background.secondary};
-					border: 1px solid ${border.primary};
+					border: 1px solid ${border.secondary};
 					display: flex;
 					flex-direction: column;
 					justify-content: space-between;
-					padding: ${space[3]}px;
-					margin-bottom: 8px;
+					padding: 6px 25px 12px;
```
oliverlloyd, line 150: `space` is essentially a base of 4 so these hard coded padding values could be replaced with, `space.1 space.6 space 3

Same for the other numbers

src/web/components/ClickToView.tsx
```diff
@@ -159,39 +165,34 @@ export const ClickToView = ({
...
-								This article includes content provided by{' '}
-								{source}. We ask for your permission before
-								anything is loaded, as they may be using cookies
-								and other technologies.
-							</p>
-							<p>
-								To view this content, click &apos;Allow and
-								continue&apos;.
-							</p>
+							This article includes content provided by {source}.
```
oliverlloyd, line 176: Is there an accessibility impact to removing the p tags? I'm not certain but I think it might be useful to maintain the semantics of text being in a p tag.

oliverlloyd (commented):
Not sure about this but semantically, it might be better to keep the p tags

src/web/components/ClickToView.tsx
frj: I guess I was trying to do a pixel perfect version of the figma, using the space would not be identical.

I happy to 'interpret' the design if thats normally acceptable. 

What do you think?

oliverlloyd:
> Not sure about this but semantically, it might be better to keep the p tags

Actually, I checked and we should always put paragraphs of text in `p` tags - both for accessibility but also things like the spacing in-between them is a design decision set site wide

src/web/components/ClickToView.tsx
oliverlloyd: Ideally, the designs should be using base 4px - this is 'the official' [spacing system](https://theguardian.design/2a1e5182b/p/449bd5-space) but that extra pixel 'tweak' is very tempting!

The `space` tokens are, loosely, designed to prevent this. That said, we do sometimes deviate but normally only because things need to line up or for other concrete reasons. If it's a matter of subjectivity, I would say the design system should win.

frj: Ive left the 6px in and changed the rest to use the 'space' values.

src/web/components/ClickToView.tsx
```diff
@@ -137,64 +144,59 @@ export const ClickToView = ({
...
 					className={css`
 						${textSize}
-						a {
-							${textSize}
-						}
-						p {
-							margin-bottom: 8px;
+						margin-bottom: ${space[5]}px;
+						strong {
+							font-weight: bold;
```
oliverlloyd, line 171: I think that this should already be applied by `globalStrongStyles` in `ArticleBody`. If not, that's a flag for a larger issue which might need investigating!

Oh... I'm guessing this is because Storybook doesn't have these globals styles! Myeh, that's annoying. We should look at setting these in config

src/web/components/ClickToView.tsx
```diff
@@ -137,64 +144,59 @@ export const ClickToView = ({
...
-				<div
+				<p
 					className={css`
 						${textSize}
-						a {
-							${textSize}
-						}
-						p {
-							margin-bottom: 8px;
+						margin-bottom: ${space[5]}px;
```
oliverlloyd, line 169: Okay, I'm delving deep into nitpick terratory here, sorry! But if we set this spacing around the Button then that kind of communicates the intention better (eg. we want more space above the button) and protects us against if new paragraphs were to be added later.

What do you think?

frj: Sounds good :)

src/web/components/ClickToView.tsx
frj: Ive taken this out and added it to the custom styling in the storybook and validated that it gets rendered correctly in the actual article.

src/web/components/ClickToView.stories.tsx
```diff
@@ -53,6 +53,9 @@ const RoleStory = ({
 				className={css`
 					max-width: 620px;
 					clear: left;
+					strong {
+						font-weight: bold;
+					}
```
oliverlloyd, line 58: Note to self: Look at a way to set this global css somewhere higher up in Storybook - I think there's a resets options or something similar.

oliverlloyd (approved):
Thanks for bearing with me. Great stuff!

PR #2743, opened 2021-03-25
Palette in GuardianLines

oliverlloyd:
## What does this change?
Refactors `GuardianLines` to accept `palette`

## Why?
Because we want to be able to set the colour of the lines based on `format`. In particular, this is useful when the article background i grey and we need the lines to pop more. Labs and SpecialReport, basically.

We will only be able to actually change the colour of Lines once we're using a release of Source that contains [this PR](https://github.com/guardian/source/pull/786)

## Why are we still being imperative?
You'd think we would be able to just pass in `format` and have this component work out what to do based on that. But that won't work because we shown different numbers of lines in different places on the page for the same article.


oliverlloyd:
[Palette in GuardianLines](https://trello.com/c/3vPmi7iV/2750-palette-in-guardianlines)

shtukas (approved):

PR #2744, opened 2021-03-25
Update Facebook App ID

jfsoul:
## What does this change?
Uses the PROD Facebook app ID - looks like we were using the DEV one by accident and that suddenly stopped working today.

### DIFF
--- a/src/web/components/ShareIcons.tsx
+++ b/src/web/components/ShareIcons.tsx
@@ -94,7 +94,7 @@ export const ShareIcons = ({
 			{displayIcons.includes('facebook') && (
 				<li className={liStyles(size)} key="facebook">
 					<a
-						href={`https://www.facebook.com/dialog/share?app_id=202314643182694&href=${encodeUrl(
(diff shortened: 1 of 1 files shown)


rcrphillips (approved):

oliverlloyd (approved):
Thanks. I've been trying to find the place in the code where this happened but things strangely trace back to [CAPI.sharingUrls](https://github.com/guardian/dotcom-rendering/commit/32a3f2b46f4c89c48d816172637fc5f9ed567413#diff-c75df329263bb473cdea40239b3479621ede29aa7fa17374b3fe5d5d17ca0f31R297) which doesn't exist anywhere anymore so I'm a bit perplexed.

oliverlloyd:
Ah ha.. This wasn't anything to do with my refactor. We have _always_ [used the wrong id](https://github.com/guardian/dotcom-rendering/blob/7ae84661b3b6ac357b2720c6cc4aea34932afd56/src/lib/sharing-urls.ts#L37)

Note. We are _still_ using this in AMP ^^

PR #2747, opened 2021-03-26
Add tracking to RR epics

TomPretty:
## What does this change?
Pass `submitComponentEvent` function to RR epics to allow them to track ophan events


tomrf1 (approved):
👍

PR #2748, opened 2021-03-26
`Labs Header`

oliverlloyd:
## What does this change?
Adds the teal header that youo see at the top of all Labs articles

### DIFF
--- a/src/web/components/Dropdown.test.tsx
+++ b/src/web/components/Dropdown.test.tsx
@@ -85,9 +85,9 @@ describe('Dropdown', () => {
 
 		const ulElement = container.querySelector('ul');
 
-		expect(ulElement).toHaveStyle('display: none');
+		expect(ulElement).not.toBeVisible();
 		fireEvent.click(getByRole('button'));
-		expect(ulElement).toHaveStyle('display: block');
+		expect(ulElement).toBeVisible();
 	});
 
 	it('should close the expanded menu when they click away', () => {
@@ -103,9 +103,9 @@ describe('Dropdown', () => {
 		const ulElement = container.querySelector('ul');
 
 		fireEvent.click(getByRole('button'));
-		expect(ulElement).toHaveStyle('display: block');
+		expect(ulElement).toBeVisible();
 		container.click();
-		expect(ulElement).toHaveStyle('display: none');
+		expect(ulElement).not.toBeVisible();
 	});
 
 	it('should close the expanded menu when blurred', () => {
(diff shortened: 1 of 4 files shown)


oliverlloyd:
[Labs 'about paid for' header](https://trello.com/c/PDmOFe5x/2667-labs-about-paid-for-header)

src/web/components/Dropdown.test.tsx
```diff
@@ -85,9 +85,9 @@ describe('Dropdown', () => {
 
 		const ulElement = container.querySelector('ul');
 
-		expect(ulElement).toHaveStyle('display: none');
+		expect(ulElement).not.toBeVisible();
 		fireEvent.click(getByRole('button'));
-		expect(ulElement).toHaveStyle('display: block');
+		expect(ulElement).toBeVisible();
```
liywjl, line 90: Nice!

src/web/components/LabsHeader.tsx
```diff
@@ -0,0 +1,185 @@
...
+const Right = ({ children }: { children: React.ReactNode }) => (
+	<div
+		className={css`
+			display: flex;
+			padding: ${space[1]}px 0;
+		`}
+	>
+		{children}
+	</div>
+);
```
liywjl, line 48: I have mixed feeling about these components, could we not just use thier css values to define them? or at least rename them to be a little more decriptive than Left and Right (maybe something like LeftCol?)

src/web/components/Dropdown.tsx
```diff
@@ -261,29 +271,29 @@ export const Dropdown = ({ id, label, links, dataLinkName }: Props) => {
...
+											})}
+											data-link-name={l.dataLinkName}
+										>
+											{l.title}
+										</a>
+									</li>
+								))}
+							</ul>
+						)}
+					</div>
```
liywjl, line 296: This component could benefit from some refactoring, we are getting quite nested turnaries that are become a little hard to follow

liywjl (approved):
Funtionally the PR is good 👍   I would just mabye abstract some things for reability

src/web/components/LabsHeader.tsx
oliverlloyd: LeftCol already has a strong meaning as a breakpoint. I've used Left and Right in a few places to denote content that is moved to the left and right in a few places. There's also Top and Bottom. I find it helps when reading the layout code to know where things will appear 🤷

src/web/components/Dropdown.tsx
oliverlloyd: Yeah, agreed. This marks the first time we're using this component in another place and I noticed things could be arranged better for this, the styles in `Dropdown` are too specific and could be moved up. I think as well there needs to be some investigation around the noJs route.

I'll ticket this up for another PR.

oliverlloyd: https://trello.com/c/YHajezB2/2763-refacctor-dropdown

PR #2751, opened 2021-03-29
Lab Cards With Badges

nitro-marky:
## What does this change?

This adds the ability to have Lab Cards with badges.

In fact, the badge prop is one of the most important as if it has
an image then the card is treated as a lab card, for example, the presence of a badge will hide the footer.

### After

Screenshot 2021-04-13 at 15 03 19

### DIFF
--- a/src/web/components/Card/Card.tsx
+++ b/src/web/components/Card/Card.tsx
@@ -26,6 +26,7 @@ import { CardFooter } from './components/CardFooter';
 import { TopBar } from './components/TopBar';
 import { CardLink } from './components/CardLink';
 import { CardAge } from './components/CardAge';
+import { LabBadge } from './components/LabBadge';
 
 type Props = {
 	linkTo: string;
@@ -57,12 +58,7 @@ type Props = {
 	// Ophan tracking
 	dataLinkName?: string;
 	// Labs
-<<<<<<< HEAD
 	badge?: BadgeType;
-=======
-	badgeForLab?: string;
-	brandName?: string;
->>>>>>> 01408f35f (Add visuallyHidden)
 };
 
 type ImageSizeType = 'small' | 'medium' | 'large' | 'jumbo';
@@ -184,12 +180,7 @@ export const Card = ({
 	alwaysVertical,
 	minWidthInPixels,
 	dataLinkName,
-<<<<<<< HEAD
 	badge,
-=======
-	badgeForLab,
-	brandName,
->>>>>>> 01408f35f (Add visuallyHidden)
 }: Props) => {
 	// Decide how we position the image on the card
 	let imageCoverage: CardPercentageType | undefined;
@@ -323,24 +314,13 @@ export const Card = ({
 										</AvatarContainer>
 									</Hide>
 								)}
-<<<<<<< HEAD
-
-								{badge && (
-									<div>
-										<div className={badgeStyle}>
-											<div>Paid for by</div>
-											<img
-												className={badgeImageStyle}
-												alt="test"
-												src="https://static.theguardian.com/commercial/sponsor/24/Jun/2020/c15a28c9-ab71-4819-ac5b-066f04cb41ba-Togetherlogo.png"
-=======
 								<CardFooter
 									format={format}
 									age={
 										webPublicationDate ? (
 											<CardAge
 												format={format}
-												palette={palette}
+												palette={cardPalette}
 												webPublicationDate={
 													webPublicationDate
 												}
@@ -353,7 +333,7 @@ export const Card = ({
 										format.design === Design.Media &&
 										mediaType ? (
 											<MediaMeta
-												palette={palette}
+												palette={cardPalette}
 												mediaType={mediaType}
 												mediaDuration={mediaDuration}
 											/>
@@ -364,68 +344,23 @@ export const Card = ({
 										longCount &&
 										shortCount ? (
 											<CardCommentCount
-												palette={palette}
+												palette={cardPalette}
 												long={longCount}
 												short={shortCount}
 											/>
 										) : undefined
 									}
 									labBadge={
-										badgeForLab ? (
-											<LabBadge
-												badgeForLab={badgeForLab}
-												palette={palette}
-<<<<<<< HEAD
-												badgeAlt={badgeAlt}
->>>>>>> d740832d0 (Feedback)
-=======
-												brandName={brandName}
->>>>>>> 01408f35f (Add visuallyHidden)
-											/>
-										</div>
-									</div>
-								)}
-								{!badge && (
-									<CardFooter
-										format={format}
-										age={
-											webPublicationDate ? (
-												<CardAge
-													format={format}
-													palette={palette}
-													webPublicationDate={
-														webPublicationDate
-													}
-													showClock={showClock}
+										badge ? (
+											<div>
+												<LabBadge
+													badgeForLab={badge.imageUrl}
+													palette={cardPalette}
 												/>
-											) : undefined
-										}
-										isFullCardImage={isFullCardImage}
-										mediaMeta={
-											format.design === Design.Media &&
-											mediaType ? (
-												<MediaMeta
-													palette={palette}
-													mediaType={mediaType}
-													mediaDuration={
-														mediaDuration
-													}
-												/>
-											) : undefined
-										}
-										commentCount={
-											showCommentCount &&
-											longCount &&
-											shortCount ? (
-												<CardCommentCount
-													palette={palette}
-													long={longCount}
-													short={shortCount}
-												/>
-											) : undefined
-										}
-									/>
-								)}
+											</div>
+										) : undefined
+									}
+								/>
 							</div>
 						</ContentWrapper>
 					</>
--- a/src/web/components/Card/Card.tsx
+++ b/src/web/components/Card/Card.tsx
@@ -335,13 +335,11 @@ export const Card = ({
 									}
 									labBadge={
 										badge ? (
-											<div>
-												<LabBadge
-													badgeForLab={badge.imageUrl}
-													palette={cardPalette}
-													brandName={brand}
-												/>
-											</div>
+											<LabBadge
+												badgeForLab={badge.imageUrl}
+												palette={cardPalette}
+												brandName={brand}
+											/>
 										) : undefined
 									}
 								/>
--- a/src/web/components/Card/Card.tsx
+++ b/src/web/components/Card/Card.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { css, cx } from 'emotion';
 
 import { Design, Special } from '@guardian/types';
-import { brandAltBackground, neutral } from '@guardian/src-foundations/palette';
+import { brandAltBackground } from '@guardian/src-foundations/palette';
 
 import { StarRating } from '@root/src/web/components/StarRating/StarRating';
 import { CardHeadline } from '@frontend/web/components/CardHeadline';
@@ -114,24 +114,6 @@ const labTitleOrNot = (format: Format) => {
 		line-height: 20px;
 	`;
 };
-
-const badgeImageStyle = css`
-	max-height: 3.75rem;
-	margin-left: 0.625rem;
-	vertical-align: middle;
-`;
-
-const badgeStyle = css`
-	font-size: 0.75rem;
-	line-height: 1rem;
-	${textSans.xsmall({ fontWeight: 'regular' })};
-	color: ${neutral[46]};
-	font-weight: bold;
-	margin-top: 0.375rem;
-	padding-right: 0.625rem;
-	padding-bottom: 0.625rem;
-	text-align: right;
-`;
 
 const StarRatingComponent: React.FC<{ rating: number }> = ({ rating }) => (
 	<>
--- a/src/web/components/Card/Card.tsx
+++ b/src/web/components/Card/Card.tsx
@@ -313,6 +313,7 @@ export const Card = ({
 										</AvatarContainer>
 									</Hide>
 								)}
+<<<<<<< HEAD
 
 								{badge && (
 									<div>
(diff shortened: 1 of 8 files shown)


src/web/components/CardHeadline.tsx
```diff
@@ -52,6 +52,34 @@ const fontStyles = (size: SmallHeadlineSize) => {
...
+			`;
+		case 'small':
+			return css`
+				${textSans.small()};
+			`;
+		case 'tiny':
+			return css`
+				${textSans.xsmall()};
+				font-size: 14px;
+			`;
```
oliverlloyd, line 79: Looking at the stories it seems some of these sizes are a bit off. 

We also need to set `{lineHeight: 'something'}` - maybe 'regular'?

Also looks like the standfirst needs the font setting. It might be easier to break these font changes out into a different PR/ticket?

src/web/components/Card/Card.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import { css, cx } from 'emotion';
 
-import { Design } from '@guardian/types';
-import { brandAltBackground } from '@guardian/src-foundations/palette';
+import { Design, Special } from '@guardian/types';
+import { brandAltBackground, neutral } from '@guardian/src-foundations/palette';
```
oliverlloyd, line 5: Favour `palette` for colours vs. accessing Source directly

src/web/components/Card/Card.tsx
```diff
@@ -99,6 +102,35 @@ const starWrapper = css`
...
+	if (format.theme === Special.Labs) {
+		return css`
+			${textSans.medium({ fontWeight: 'bold' })}
+		`;
+	}
+	return css`
+		${headline.xxxsmall({ fontWeight: 'regular' })}
+		line-height: 20px;
+	`;
+};
```
oliverlloyd, line 115: I think this is already set in `CardHeadline`?

oliverlloyd (commented):
I know this PR is still in draft so you might already have fixed half these things but I dropped some comments anyway just in case. Almost entirely small things.

I did have one thought about using the CardFooter component as a way to make the choice about what to show in the foooter. This is the established place where this decision is made now and I think it would be better to keep that the one place where this choice it made

src/web/components/Card/Card.tsx
nitro-marky: It's not set as `textsans` just `headline` in `CardHeadline`

src/web/components/Card/Card.tsx
```diff
@@ -263,43 +298,60 @@ export const Card = ({
...
+												src="https://static.theguardian.com/commercial/sponsor/24/Jun/2020/c15a28c9-ab71-4819-ac5b-066f04cb41ba-Togetherlogo.png"
 											/>
-										) : undefined
-									}
-								/>
+										</div>
+									</div>
+								)}
+								{!badge && (
+									<CardFooter
```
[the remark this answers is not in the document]
nitro-marky: Yeah, it's all in the footer now.

src/web/components/Card/Card.tsx
```diff
@@ -265,8 +269,10 @@ export const Card = ({
 								)}
 								<CardFooter
 									format={format}
+									badge={badge || undefined}
+									palette={palette}
 									age={
-										webPublicationDate ? (
+										webPublicationDate && !badge ? (
```
oliverlloyd, line 275: If you look at the other examples of use cases being passed into `CardFooter` there's a pattern where the thing to show, (`age`, `mediaMeta` or `commentCount`) is passed in as a prop and then the `CardFooter` uses a `footer` tag with some flex position styling to wrap it (you might not need any flex styling here - not sure).

So we could have another prop of `badge` and that could be set to the react component that represents how the badge renders. That way we keep the logic consistent, we only extend it, and we don't need to use this `!badge &&` check here.

Shout if I'm not making sense - happy to step though it!

src/web/components/Card/components/TopBar.tsx
```diff
@@ -12,7 +12,7 @@ export const TopBar = ({ children, palette, isFullCardImage }: Props) => (
 		className={css`
 			display: flex;
 			width: 100%;
-
+			line-height: 1;
```
oliverlloyd, line 15: Typo?

paperboyo:
Hiya, a code-blind person just passing by… These logos: will their size be decided in CSS and actual image dimension won‘t affect it at all? Asking coz they were giving us grief on some templates where their display size was dependent on image dimensions. cc @nicl FYI

nitro-marky:
@paperboyo the logo sizes [will be set here](https://github.com/guardian/dotcom-rendering/blob/fa66b0ad9cd34915b3865bf2bd6014be8fdc18ae/src/web/components/Card/components/CardFooter.tsx#L49) (unless this is going to be a problem)

src/web/components/Card/components/TopBar.tsx
liywjl, line 15:
```suggestion
			line-height: 1px;
```

src/web/components/LabBadge.tsx
```diff
@@ -0,0 +1,44 @@
...
+`;
+
+export const LabBadge = ({ badge, badgeAlt, palette }: Props) => (
+	<div className={badgeWrapperStyle}>
+		<div
+			className={cx(
+				paidForStyle,
+				css`
+					color: ${palette.text.branding};
+				`,
```
liywjl, line 37: you can merge this with `paidForStyle`?

src/web/components/Card/Card.tsx
```diff
@@ -265,8 +272,9 @@ export const Card = ({
 								)}
 								<CardFooter
```
liywjl, line 273: I feel like it might be better to have `!badge` at the `CardFooter` component instead of having it for each prop? You could then add use `LabBadge` component here instead of `CardFooter`?

src/web/components/Card/Card.tsx
```diff
@@ -205,6 +211,7 @@ export const Card = ({
 									isFullCardImage={isFullCardImage}
 								>
 									<CardHeadline
+										// Try and create parity with frontend, not working at the moment
```
liywjl, line 214: why does it not work? not sure if this comment is very useful

liywjl (changes requested):
I think Labs have thier own footer that means it makes sense to break out of the `CardFooter` logic IMO. Plus a couple other nice to haves

oliverlloyd (commented):
Sorry, quite a few comments still!

Happy to go through these tomorrow if you like.

And one final point but I wasn't sure where to put this but the size of the badge seems to be too small. Comparing Frontend:

Screenshot 2021-03-31 at 14 57 19

to this PR:

Screenshot 2021-03-31 at 15 00 58

Not a HUGE difference but I suspect logo sizing is important for paid content

src/web/components/Card/Card.tsx
nitro-marky: That was just a message to myself, forgot to take it out

src/web/components/Card/Card.tsx
nitro-marky: I was told to do it this way 🤷‍♂️

src/web/components/Card/Card.tsx
```diff
@@ -147,6 +151,8 @@ export const Card = ({
 	alwaysVertical,
 	minWidthInPixels,
 	dataLinkName,
+	badge,
+	badgeAlt,
```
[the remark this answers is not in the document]
nitro-marky: The linting is requiring an alt text

src/web/components/LabBadge.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { textSans } from '@guardian/src-foundations/typography';
+
+type Props = {
+	badge: string;
+	badgeAlt: string;
+	palette: Palette;
+};
+
+const badgeImageStyle = css`
+	max-height: ${space[12]}px;
```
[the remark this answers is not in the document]
nitro-marky: What do you mean?

src/web/components/Card/Card.tsx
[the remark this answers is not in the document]
oliverlloyd: I think this is worth more thought.

Alt text generally gets very little attention, I'm as guilty as anyone of seeing it as an annoyance, something that my linter is complaining about. Like most people I typically add something to make the squiggly line go away and move on. But I think we can do better here.

[This great article from the RNIB](https://www.rnib.org.uk/accessibility-guidelines-alt-text-what-you-need-know)  explains what alt text is and how it can be used.

The key things I learnt from reading this and other articles are:

Alt text is often annoying noise for people using screen readers. If it isn't adding value then you should _not_ set alt text.

If an image is key to the article, if it provides context that is relevant to the text around it, then alt text is essential to maintianing that meaning and context (alt text is required). If it is just a pretty addition with no particular relavance besides looking nice, then no alt text should be set (alt text is noise).

## What about this example here?
Screenshot 2021-04-05 at 19 13 03

In this particular case, we're communicating the message "Paid for by Vodafone". This is important context. Unfortunately, we're not communicating this in an accessible manner. We're assuming that everyone is sighted and can easily make the transition between text and image but that's not true and even if we do set alt text it's clunky and we're only giving the information in the card footer (long after the link was first read out and so it probably never gets accessed).

So the accessibility 'problem' that we need to solve is not what alt text should I use, but instead, it is how can we make it clear to someone using a screen reader that this content is paid for by Vodafone?

### What does good look like?
A good, accessible solution would be to have a prop called `brandName` and then to use that to create a, longer, screen reader accessible version of the 'Paid for by' text that reads a useful message: Eg. "This content was paid for by Vodafone as part of Guardian Labs". You then make this mini component hidden and only something consumed by screen readers. We do something similar for the age warning on articles and [there is a utility for this](https://theguardian.design/2a1e5182b/p/6691bb-accessibility/t/1285a8) in Source. We can then place it next to the headline so we give the information to people when they need it. If the prop `brandName` is noot provided we default to "This content has been paid for by a sponser as part of Guardian Labs".

By creating a separate mini component to read this text out we can also use the same thing when rendering the main article. Right now I'm not sure we're annoucing paid content very well to people using screen readers, it's in the article meta but sufferes from similar issues as here. We should probably place it in the same position that we do the accessible age warning, directly above or below the headline.

### Random side thought
@SiAdcock It'd maybe be useful to have a generic component for this kind of thing. The `visuallyHidden` css class is already 😍 but maybe something like 

```ts
<ScreenReaderOnly text="I am only read to screen readers" />
```

better communicates the intentions? Just a thought. It took me a while to absorb what `visuallyHidden` was about.

## Urgh, but I just want to finish my PR already!
I get it, I'm dragging this PR out, but I actually think this is a fantastic opportunity to examine what alt text is and make ourselves better developers. We care about accessibilty at the Guardian and that means that we sometimes put the extra time and effort in to create really great solutions. Should we invest this much time and effort into all code and PRs? Probably not, but hopefully by focussing in every now and then we all learn something.

Should I stop wrting this comment already? Yes. Was it too long? Possibly. Sorry.

rcrphillips:
[Support badges in Cards](https://trello.com/c/bFKmXhlR/2663-support-badges-in-cards)

rcrphillips:
[Support badges in Cards](https://trello.com/c/bFKmXhlR/2663-support-badges-in-cards)

rcrphillips:
[Support badges in Cards](https://trello.com/c/bFKmXhlR/2663-support-badges-in-cards)

rcrphillips:
[Support badges in Cards](https://trello.com/c/bFKmXhlR/2663-support-badges-in-cards)

src/web/components/Card/Card.tsx
oliverlloyd: This comment is still here 😉

src/web/components/Card/Card.tsx
```diff
@@ -313,6 +334,17 @@ export const Card = ({
 											/>
 										) : undefined
 									}
+									labBadge={
+										badge ? (
+											<div>
```
oliverlloyd, line 339: We can delete this div

src/web/components/Card/components/LabBadge.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { textSans } from '@guardian/src-foundations/typography';
+import { visuallyHidden } from '@guardian/src-foundations/accessibility';
+
+type Props = {
+	badgeForLab: string;
+	brandName?: string;
+	palette: Palette;
+};
+
+// max-height: ${space[12]}px; - Couldn't find a direct replacement in Source
```
oliverlloyd, line 13: I think we can remove this comment. As per our chat earlier `space` is used for padding and margins and isn't needed for things like the height of stuff

src/web/components/Card/components/LabBadge.tsx
```diff
@@ -0,0 +1,49 @@
...
+			className={css`
+				${visuallyHidden};
+			`}
+		>
+			{brandName
+				? `This content was paid for by ${brandName} as part of Guardian labs`
+				: 'This content has been paid for by a sponsor as part of Guardian Labs.'}
+		</span>
+		{/* eslint-disable-next-line jsx-a11y/alt-text */}
+		<img className={badgeImageStyle} src={badgeForLab} />
```
oliverlloyd, line 47: I think by convention you can just pass an empty string to alt. This means you don't need to ignore the lint rule and also communicates intent better.

src/web/components/Card/components/LabBadge.tsx
```diff
@@ -0,0 +1,49 @@
...
+		<div className={paidForStyle(palette)}>Paid for by</div>
+		<span
+			className={css`
+				${visuallyHidden};
+			`}
+		>
+			{brandName
+				? `This content was paid for by ${brandName} as part of Guardian labs`
+				: 'This content has been paid for by a sponsor as part of Guardian Labs.'}
+		</span>
```
oliverlloyd, line 45: 😻 😻 😻 😻 😻 😻

src/web/components/CardHeadline.tsx
```diff
@@ -111,7 +140,9 @@ export const CardHeadline = ({
 	<>
 		<h4
 			className={cx(
-				fontStyles(size),
+				format.theme === Special.Labs
+					? labTextStyles(size)
+					: fontStyles(size),
```
oliverlloyd, line 145: 👍

src/web/components/CardHeadline.tsx
```diff
@@ -52,6 +52,35 @@ const fontStyles = (size: SmallHeadlineSize) => {
...
+					${textSans.medium()};
+				}
+			`;
+		case 'medium':
+			return css`
+				${textSans.large({ lineHeight: 'tight' })};
+				${until.desktop} {
+					${textSans.medium({ lineHeight: 'tight' })};
+				}
+				padding-bottom: ${space[1]}px;
```
oliverlloyd, line 70: Noice

src/web/components/Card/components/CardFooter.tsx
```diff
@@ -47,7 +48,12 @@ export const CardFooter = ({
 	mediaMeta,
 	commentCount,
 	isFullCardImage,
+	labBadge,
 }: Props) => {
+	if (format.theme === Special.Labs && labBadge) {
+		return labBadge;
```
oliverlloyd, line 54:
```suggestion
		return (
			<footer>labBadge</footer>
		);
```

for the semantic html win

oliverlloyd (approved):
Nice!

Still a few points to clear up but looking great!

liywjl (approved):
LGTM 👍

PR #2752, opened 2021-03-29
Decouple braze automat

liywjl:
## What does this change?
- Moves `atomoat` to `lib`
- Update `@guardian/braze-components`

## Why?
Decouple Braze from Automat, `@guardian/braze-components - v0.0.20` uses peer dependencies so no longer need to define automat in global scope


src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -159,41 +158,30 @@ export const BrazeBanner = ({ meta }: Props) => {
...
-					setBrazeComponent(() => module.BrazeMessage);
-				})
-				.catch((error) =>
-					window.guardian.modules.sentry.reportError(
-						error,
-						'braze-banner',
-					),
-				);
-		}
-	}, [meta]);
```
liywjl, line 187: `meta` is never used in the `useEffect` and according to the componet prop it is always defined

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -159,41 +158,30 @@ export const BrazeBanner = ({ meta }: Props) => {
...
-
-	if (BrazeComponent && meta) {
```
liywjl, line 189: ditto `meta`

src/web/components/StickyBottomBanner/BrazeBanner.tsx
```diff
@@ -159,41 +158,30 @@ export const BrazeBanner = ({ meta }: Props) => {
...
+			{BrazeComponent ? (
+				<BrazeBannerWithSatisfiedDependencies
+					BrazeComponent={BrazeComponent}
+					meta={meta}
+				/>
+			) : (
+				<div />
+			)}
+		</>
+	);
```
liywjl, line 186: I feel its nicer to unify the rendering logic into one return statement

shtukas (approved):

src/web/components/StickyBottomBanner/BrazeBanner.tsx
tjmw: Huh yep you're right. I wonder why the condition exists, maybe in a previous version of this it was needed 🤔   Since we've now got a `useEffect` with an empty dependency array, causing it to only run once, perhaps `useOnce` would be better?

PR #2755, opened 2021-03-29
fix: Prevent layout shift for top above nav slot by using toggled ad …

arelra:
## What does this change?

DCR counterpart to: https://github.com/guardian/frontend/pull/23613

Removes the layout shift caused by the dynamic insertion of the advertisement label into the top banner (aka top-above-nav).

The change is to add the label to the top banner template.

This reserves its height so when shown it does not shift content downwards.

At runtime, when an ad is available, [render-advert-label](https://github.com/guardian/frontend/blob/1d462478295c13a2f3dbc47ad96609f824a8d9e8/static/src/javascripts/projects/commercial/modules/dfp/render-advert-label.ts#L68):
- checks if the label should be shown
- detects if a toggled label exists

If so the toggled label is:

- made visible
- has its width set to match the advert and centre it correctly

Fyi we are planning another look at the way labels are handled. Ideally we'd like to push everything into GAM and remove any custom code.

### Before

layout-shift-before-dcr

### After

layout-shift-after-dcr

## Why?

Reduces annoyance to users
Improves core web vitals scores

### DIFF
--- a/src/web/components/AdSlot.tsx
+++ b/src/web/components/AdSlot.tsx
@@ -1,10 +1,9 @@
-/* eslint-disable react/jsx-props-no-spreading */
 import React from 'react';
 import { css, cx } from 'emotion';
 
 import { border, neutral, text } from '@guardian/src-foundations/palette';
 import { textSans } from '@guardian/src-foundations/typography';
-import { from } from '@guardian/src-foundations/mq';
+import { from, until } from '@guardian/src-foundations/mq';
 import { Display } from '@guardian/types';
 
 type Props = {
@@ -46,18 +45,34 @@ enum Size {
 	empty = '2,2',
 }
 
+const adSlotLabelStyles = css`
+	${textSans.xsmall()};
+	position: relative;
+	height: 24px;
+	background-color: ${neutral[97]};
+	padding: 0 8px;
+	border-top: 1px solid ${border.secondary};
+	color: ${text.supporting};
+	text-align: left;
+	box-sizing: border-box;
+	&.visible {
+		visibility: initial;
+	}
+	&.hidden {
+		visibility: hidden;
+	}
+	&.ad-slot__label--toggle {
+		margin: 0 auto;
+		${until.tablet} {
+			display: none;
+		}
+	}
+`;
+
 export const labelStyles = css`
 	.ad-slot__label,
 	.ad-slot__scroll {
-		${textSans.xsmall()};
-		position: relative;
-		height: 24px;
-		background-color: ${neutral[97]};
-		padding: 0 8px;
-		border-top: 1px solid ${border.secondary};
-		color: ${text.supporting};
-		text-align: left;
-		box-sizing: border-box;
+		${adSlotLabelStyles}
 	}
 
 	.ad-slot__close-button {
@@ -126,6 +141,19 @@ const mobileStickyAdStyles = css`
 		${textSans.xsmall()};
 	}
 `;
+
+const AdSlotLabelToggled: React.FC = () => (
+	<div
+		className={cx(
+			'ad-slot__label',
+			'ad-slot__label--toggle',
+			'hidden',
+			adSlotLabelStyles,
+		)}
+	>
+		Advertisement
+	</div>
+);
 
 export const AdSlot: React.FC<Props> = ({ position, display }) => {
 	switch (position) {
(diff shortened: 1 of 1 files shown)


src/web/components/AdSlot.tsx
```diff
@@ -267,52 +295,55 @@ export const AdSlot: React.FC<Props> = ({ position, display }) => {
 		case 'top-above-nav': {
 			const adSlotAboveNav = css`
 				margin: 0 auto;
-				height: 151px;
+				min-height: 108px;
```
arelra, line 298: Previously this calculated height as:
`smallest ad height (90px) + 2 * padding  (2 * 18px) + ad label (24px) = 150px`

`padding-top` [previous PR](https://github.com/guardian/dotcom-rendering/pull/2668) and the ad label (this PR) no longer exist within `top-above-nav`.

Calculation is now:
`smallest ad height (90px) + 1 * padding  (1 * 18px) = 108px`

oliverlloyd (approved):
Really great work! This PR is focussed, clear and easy to understand.

This change is, I presume, dependent on the Frontend PR being deployed? Will it break anything if the FE one is deployed but the DCR one hasn't gone out?


`----------- Everything below 👇  is irrelevant to this PR --------------`

And (and this is just a general comment and nothing to do with this work here) this work really highlights the fact that we have these cross repo couplings where FE depends on these classnames existing with the right styles set on them and DCR is exposing this html and 🤞 that the commercial script is able to do what it needs.

There's nothing new about this but because DCR is it's own app the coupling is that bit more brittle.

This is an old subject but it'd be great to think about a more react focussed solution for rendering ads where there was a defined contract for the data to be passed in (props) and then this react component could return some functions like `toggleLabel()` or `refreshSLot()` to allow interactivity.

Am I dreaming? Is it a nice idea but too much work? Quite possibly. But I do think it's good to raise it every now and then, just as a thought experiement, really.

mxdvl (approved):
As good as its `frontend` counterpart. 👍

arelra:
> Really great work! This PR is focussed, clear and easy to understand.

Thanks 👍 

> This change is, I presume, dependent on the Frontend PR being deployed? Will it break anything if the FE one is deployed but the DCR one hasn't gone out?

They can be deployed independently.

If toggled labels are not detected the new commercial code will render the old dynamic way.

Ditto for the old commercial bundle by default.

However, it does make more sense for Frontend to go first so the updated commercial code can handle the toggled ad labels if present.

I tested this behaviour locally and on CODE.
 
> `----------- Everything below 👇 is irrelevant to this PR --------------`
> 
> And (and this is just a general comment and nothing to do with this work here) this work really highlights the fact that we have these cross repo couplings where FE depends on these classnames existing with the right styles set on them and DCR is exposing this html and 🤞 that the commercial script is able to do what it needs.

I agree. Wrt commercial code, the coupling is brittle and DCR is dependent on external behaviour and various dom hooks which break the encapsulation of the DCR components.

> There's nothing new about this but because DCR is it's own app the coupling is that bit more brittle.
> 

Killing Frontend would be a massive help/incentive of course.

> This is an old subject but it'd be great to think about a more react focussed solution for rendering ads where there was a defined contract for the data to be passed in (props) and then this react component could return some functions like `toggleLabel()` or `refreshSLot()` to allow interactivity.
> 
> Am I dreaming? Is it a nice idea but too much work? Quite possibly. But I do think it's good to raise it every now and then, just as a thought experiement, really.

Please keep raising :)

It would be great to expose commercial components and utilities that could then be consumed by DCR.

The intention is to move all our code out of Frontend into our own repos at which point we could start to encapsulate as components.

Our main issue (I believe) is we have layers of ~~sedimentary rock~~ client side code and behaviour in GAM which need picking apart.

We could start with something small and get some direction of travel.

A good topic to discuss further in our team catchups. Looping in the team...

@ashishpuliyel @ioanna0  @mxdvl

liywjl (approved):
Great work!! 👍

oliverlloyd:
> Killing Frontend would be a massive help/incentive of course.

We're actually quite close to this, at least for articles anyway, we plan on switching off the 90/10% AB test this week which means DCR will be serving something like 99% of all traffic (and we're actively working on the remaining bits)

oliverlloyd:
> The intention is to move all our code out of Frontend into our own repos at which point we could start to encapsulate as components.

Lots of thumbs up here. I notice you've already got commercial-core up and running which is awesome. I am very aware that this is really challenging work!

oliverlloyd:
> We could start with something small and get some direction of travel.

Perhaps at some point in the future once we've got the core logic broken out we could take a single slot, like top right, and POC something for that? That might be nice because (I think) it's only used by articles and is relatively simple.

This is all very loose but maybe it could work a bit like:

1. DCR server side renders a div that has the id `top-right` and then adds some local styles, makes it sticky maybe.
2. On the client, during hydration, we also 'hydrate' the slot using a custom hook

```typescript
  import { useAd } from '@guardian/commercial-react'

  const { refreshSlot, loading } = useAd({
	slotId: 'top-right',
	config: configStuff,
        onImpression: someCallback,
  });
```

That's all very loose but you'd have `refreshSlot()` being returned so that DCCR can use that to force a refresh (if needed, I think there are use cases for this) and maybe a `loading` boolean so we can perhaps control loading state while we wait for the slot to initialise.

This is all just crazy thinking and I'm certain to be missing a whole bunch of things but it's maybe a nice excercise to think about what the ideal api would look like in terms of the dev ex for consumers of the commercial lib? For DCR, custom hooks are a great dev ex and something that very easy to understand and train new people up on but maybe other teams want a more plain js approach? Perhaps you could also expose a non reacty way to set a ad slot?

```
  import { initSlot } from '@guardian/commercial-core'
```

arelra:
> > We could start with something small and get some direction of travel.
> 
> Perhaps at some point in the future once we've got the core logic broken out we could take a single slot, like top right, and POC something for that? That might be nice because (I think) it's only used by articles and is relatively simple.

A POC sounds like a great idea to get some movement.

> This is all very loose but maybe it could work a bit like:
> 
> 1. DCR server side renders a div that has the id `top-right` and then adds some local styles, makes it sticky maybe.
> 2. On the client, during hydration, we also 'hydrate' the slot using a custom hook
> 
> ```ts
>   import { useAd } from '@guardian/commercial-react'
> 
>   const { refreshSlot, loading } = useAd({
> 	slotId: 'top-right',
> 	config: configStuff,
>         onImpression: someCallback,
>   });
> ```

> That's all very loose but you'd have `refreshSlot()` being returned so that DCCR can use that to force a refresh (if needed, I think there are use cases for this) and maybe a `loading` boolean so we can perhaps control loading state while we wait for the slot to initialise.
> 
> This is all just crazy thinking and I'm certain to be missing a whole bunch of things but it's maybe a nice excercise to think about what the ideal api would look like in terms of the dev ex for consumers of the commercial lib? For DCR, custom hooks are a great dev ex and something that very easy to understand and train new people up on but maybe other teams want a more plain js approach? Perhaps you could also expose a non reacty way to set a ad slot?
> 
> ```js
>   import { initSlot } from '@guardian/commercial-core'
> ```

Thanks for taking the time to show a path forward. 🚀 

I agree that we should start from a desired outcome rather than doing the more intuitive steps of 1) porting existing code out of Frontend and then 2) componentising.

My first thought is, given that GAM is generally pulling the strings of ads I think Ad components will be a bit more black box than you've described but of course all that is yet to be determined.

I guess we could facade GAM via hooks which would then reference a GAM singleton.

I'll add a ticket to our backlog and see where we can start taking time to do a POC.

Thanks again

PR #2757, opened 2021-03-30
`DeadBlog` | Colours and spacing

oliverlloyd:
## What does this change?
Fixes the palette used for DeadBlogs

### Before
Screenshot 2021-03-30 at 11 42 11

### After
Screenshot 2021-03-30 at 11 40 38

## Why?
Now that we're using CAPI `format` we have the right types set but this exposed some issues with the stories.

### DIFF
--- a/src/web/components/ArticleHeadlinePadding.tsx
+++ b/src/web/components/ArticleHeadlinePadding.tsx
@@ -15,29 +15,29 @@ const determinePadding = ({
 	switch (design) {
 		case Design.Interview:
 			return css`
-				${space[1]}px;
+				padding-top: ${space[1]}px;
 			`;
 		case Design.Review:
 			if (starRating) {
 				return '';
 			}
 			return css`
-				${space[6]}px;
+				padding-bottom: ${space[6]}px;
 				${from.tablet} {
-					${space[9]}px;
+					padding-bottom: ${space[9]}px;
 				}
 			`;
 		case Design.LiveBlog:
 		case Design.DeadBlog:
 			return css`
-				${space[1]}px;
+				padding-top: ${space[1]}px;
 			`;
 		default:
 			return css`
-				${space[1]}px;
-				${space[6]}px;
+				padding-top: ${space[1]}px;
+				padding-bottom: ${space[6]}px;
 				${from.tablet} {
-					${space[9]}px;
+					padding-bottom: ${space[9]}px;
 				}
 			`;
 	}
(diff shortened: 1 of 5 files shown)


oliverlloyd:
@HarryFischer the spacing below the Headline and the Standfirst here is a bit looser than on Frontend. This is naturally how the spacing falls through so I thought I'd check if this was okay or if it really needs to be so tight (see below for comparison)

### Frontend
Screenshot 2021-03-30 at 11 45 04

### DCR (this PR)
Screenshot 2021-03-30 at 11 47 04

oliverlloyd:
[Update `decidePalette` to use correct colours for `DeadBlog` header now that we have Format available to use.](https://trello.com/c/gA4M6ScX/2762-update-decidepalette-to-use-correct-colours-for-deadblog-header-now-that-we-have-format-available-to-use)

src/web/components/ArticleHeadlinePadding.tsx
```diff
@@ -26,6 +26,11 @@ const determinePadding = ({
 					padding-bottom: 36px;
 				}
 			`;
+		case Design.LiveBlog:
+		case Design.DeadBlog:
+			return css`
+				padding-top: 3px;
```
nicl, line 32: Not related to this PR but more a question cc @HarryFischer : what are thoughts on changing these paddings to standard Guardian spacing/Source sizes?

src/web/lib/decidePalette.ts
```diff
@@ -624,6 +628,8 @@ const borderNavPillar: (format: Format) => string = (format) =>
 	pillarPalette[format.theme].bright;
 
 const borderArticle: (format: Format) => string = (format) => {
+	if (format.design === Design.LiveBlog || format.design === Design.DeadBlog)
+		return '#CDCDCD';
```
nicl, line 632: Again, more a future question cc @HarryFischer is it possible to use a Source/palette colour here. (Or if this exact shade required, should we add this to Source?)

nicl (approved):
🎈

src/web/lib/decidePalette.ts
nicl: (And for the other border colours we use.)

oliverlloyd: Yeah, there's a few of these where you really need the exact shade for the border to look right and Source doesn't have them. It brings us back to the idea of grouping colours. Each backgrond colour could have it's own border and text variations

src/web/components/ArticleHeadlinePadding.tsx
oliverlloyd: Actually, that really should be `space.1` I'll fix that!

src/web/lib/decidePalette.ts
oliverlloyd: Leaving these as hex strings for now as per the the conversation irl with @HarryFischer . I agree that they should ideally be values from Source but I think given that they're already abstracted behind `palette` there's less impact compared to just randomly dropping hex values in the code.

PR #2759, opened 2021-03-30
Simplify type relations

shtukas:
This mostly removes the confusing fact that `GenericAtomBlockElement` extends from `InteractiveAtomBlockElement(Base)`, also there is no reason why our basic types should have some some sort of hierarchy between them (if not to abstract something really common, which wasn't the case here).


nicl (approved):

shtukas:
Thanks @nicl

PR #2760, opened 2021-03-30
refactors docs/atoms.md into docs/important-files.md

shtukas:
This removes the atoms.md file I had created when we started the DCR atom implementation work a long time ago (well, before atom-rendering was born), which is no longer needed and not really maintained (and potentially misleading) but preserves the additions Mark had added to it.


nitro-marky (approved):

liywjl (approved):

nitro-marky:
The only thing to be aware of is I believe the intention is to move event handling into the atom at some point so it might need a teak in the future but looks good 👍

PR #2761, opened 2021-03-30
Remove placeholder.com dependency

nicl:
At the time of writing placeholder.com has a certificate error and requests are failing. In any case, it's better not to depend on a third-party for storybook and Chromatic where possible.

New look:

Screenshot 2021-03-30 at 13 54 24

### DIFF
--- a/src/web/components/LeftColumn.stories.tsx
+++ b/src/web/components/LeftColumn.stories.tsx
@@ -6,6 +6,7 @@ import { ArticleContainer } from '@root/src/web/components/ArticleContainer';
 import { Section } from '@frontend/web/components/Section';
 
 import { LeftColumn } from './LeftColumn';
+import { Placeholder } from './Placeholder';
 
 export default {
 	component: LeftColumn,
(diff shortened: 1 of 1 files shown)


tjmw (approved):
Nice, this LGTM (assuming Chromatic is happy)

src/web/components/LeftColumn.stories.tsx
```diff
@@ -1,17 +1,25 @@
 import React from 'react';
+import { css } from 'emotion';
 
 import { Flex } from '@root/src/web/components/Flex';
 import { RightColumn } from '@root/src/web/components/RightColumn';
 import { ArticleContainer } from '@root/src/web/components/ArticleContainer';
 import { Section } from '@frontend/web/components/Section';
 
+import { palette } from '@guardian/src-foundations';
```
oliverlloyd, line 9: We migrated away from this as we now have our own `palette`

oliverlloyd: Re this import this was deprecated by Simon et al a while back, I think for performance reasons, not 100% sure, but I remember doing a big mob to remove them all

nicl:
> There's actually a `Placeholder` component we could use here

Nice thanks. Have subbed it in.

PR #2765, opened 2021-03-30
Hack Cards to fix display issue

oliverlloyd:
## What does this change?
This PR introduces a hack that we're using to fix a display issue in Production instead of refactoring the Card component to fix the root issue properly

### The History
After https://github.com/guardian/dotcom-rendering/pull/2634 was merged we noticed that Cards for immersive articles were not appearing correctly

Screenshot 2021-03-30 at 17 48 45

This happened because we had previously put in some hacks to hardcode the `display` value for onwards Cards to `Standard`. We did this because the definition of an 'immersive' card had "moved" as part of the Carousel work. But it had only really kind of shifted and we still aren't sure what we will do with them. But that uncertainty was okay because it was just an AB test. Or so we thought.

However, as time moved on other work around `format` was progressing and one of the major steps was to hoist the decision about what `format` an article is up into CAPI. This was great but it meant we were no longer hard coding it, so we ended up again having conflicting model problems.

### The Hack
Is to, once again, hardcode the `display` value for all Cards in Onwards. This is scary and dangerous because it breaks the model. Any sub component below a Card will get an unexpected value.

### After
Screenshot 2021-03-30 at 17 57 50

### Why did we do this if it's so bad?
The real solution is to refactor how we model Cards and correctly incorporate Carousels and immersive cards into it. But we can't do that until we know 1) if we even want to have a Carousel or 2) what an 'immersive' card means. Those choices affect the direction any refactor will take.

### Where should we aim to be
We want to be able to correctly model Cards, in all their glorious variation, as part of the DCR data model. This can either be part of `format` or an extension to `format`.

The extra layer of variation is this:

Sometimes, when a Card is in a special container, it changes. It might get a different background colour, or different sizing and some elements might show or appear. The Container impacts the styling of the card so we need to model the containers. There's an ongoing piece of work to start looking at this but in this particular case you can say that `Carousel` is a type of container and when a Card appears inside it then in _that context_ `Display.Immersive` means one thing, but in the normal context, it might mean something else, or probably just nothing.

So we need to extend the logic where we make design decisions to include this new ContainerType variable and to also use that when making decisions.

### DIFF
--- a/src/web/components/Card/Card.Designs.stories.tsx
+++ b/src/web/components/Card/Card.Designs.stories.tsx
@@ -6,7 +6,7 @@ import { Card } from './Card';
 
 export default {
 	component: Card,
-	title: 'Components/Card/Themes',
+	title: 'Components/Card/Designs',
 	parameters: {
 		viewport: {
 			// This has the effect of turning off the viewports addon by default
--- a/src/web/components/Card/Card.Displays.stories.tsx
+++ b/src/web/components/Card/Card.Displays.stories.tsx
@@ -1,0 +1,45 @@
+import { Design, Display, Pillar } from '@guardian/types';
+
+import { Format } from './Card.Format.stories';
+
+import { Card } from './Card';
+
+export default {
+	component: Card,
+	title: 'Components/Card/Displays',
+	parameters: {
+		viewport: {
+			// This has the effect of turning off the viewports addon by default
+			defaultViewport: 'doesNotExist',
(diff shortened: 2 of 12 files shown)


shtukas (approved):

shtukas:
The difference between an ugly hack and a sensible approach to a pressing problem is the comments. Comments are gold, they let people in the future know what to do 🙂

PR #2767, opened 2021-03-31
Workaround for missing "Letter" design and incorrectly chosen theme

jfsoul:
## What does this change?
Adds a direct workaround for the fact we have no special treatment for the _Letters_ design, and that the theme is not adjusted for this design to be Opinion. We need to move some of this logic into the CAPI client directly (or create a specific treatment for letters if they're meant to be different) but this is the quick way to resolve the live issue.

### Before
image

### After
image

## Why?
To resolve a live issue.


shtukas (approved):

oliverlloyd (approved):
Be good to add a comment to these so that we help ourselves remove these later (once we done the respective tasks to enable removing the workarounds)

PR #2769, opened 2021-03-31
Better file names in docs/elements

shtukas:
Remove the inconsistent file prefix numbering in [docs/elements].

BEFORE:

Screenshot 2021-03-31 at 12 27 40

AFTER:

Screenshot 2021-03-31 at 12 27 26


liywjl (approved):
Nice 👍

PR #2771, opened 2021-03-31
Add some basic focus styling

nicl:
## What does this change?

A very crude (attempted) improvement on the status quo. There is a long way to go to really fix focus visuals. It would be great here to figure out a sensible default, which we can then refine for specific components. 

~Note, https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/32e9fb strongly recommends the blue halo around focused elements. I've tweaked the implementation slightly from box-shadow to outline as there appear to be some downsides to box-shadow (see https://stackoverflow.com/questions/52589391/css-box-shadow-vs-outline) but happy to be corrected on this.~ Going to use `focusHalo` for now and we will consider updating Source if/as necessary.

### Before

The issue differs across browsers. Chrome default focus styling is a bit better here, but FF out of the box is pretty useless alas (speaking as someone who loves FF). The examples below are for FF anyway.

Kapture 2021-03-31 at 16 30 27

### After

Kapture 2021-03-31 at 16 14 06

## Why?

https://theguardian.design/2a1e5182b/p/6691bb-accessibility/t/24562a is the starting point.

https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G195 is the kind of approach I think we are aiming for here.

### DIFF
--- a/src/web/layouts/DecideLayout.tsx
+++ b/src/web/layouts/DecideLayout.tsx
@@ -8,6 +8,8 @@ import { decideDisplay } from '@root/src/web/lib/decideDisplay';
 import { decidePalette } from '@root/src/web/lib/decidePalette';
 import { decideDesign } from '@root/src/web/lib/decideDesign';
 
+import { injectGlobal } from 'emotion';
+import { focusHalo } from '@guardian/src-foundations/accessibility';
 import { StandardLayout } from './StandardLayout';
 import { ShowcaseLayout } from './ShowcaseLayout';
 import { CommentLayout } from './CommentLayout';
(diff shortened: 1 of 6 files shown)


src/web/layouts/DecideLayout.tsx
```diff
@@ -19,6 +21,15 @@ type Props = {
 	NAV: NavType;
 };
 
+// eslint-disable-next-line @typescript-eslint/no-unused-expressions
+injectGlobal`
```
nicl, line 25: Not at all sure this is a good way to do. An alternative might be to create an `accessibilityStyles` export and include that directly in `htmlTemplate`. Requiring every component to customise focus styling doesn't seem realistic or efficient (though we should definitely look to customising focus for a lot of components as appropriate going forward).

paperboyo:
🎉 (does it use [this](https://theguardian.design/2a1e5182b/p/6691bb-accessibility/t/32e9fb) or similar to prevent ugly left-overs for mouse users?)

liywjl (commented):
idk if it wouldnt be better to instead use `focusHalo` from `'@guardian/src-foundations/accessibility'`?

nicl:
@paperboyo good point thanks. The Source implementation does, but my customised version broke things so now re-added.

@liywjl yeah agree. Awaiting @SiAdcock 's thoughts on the box-shadow vs outline question in description. Depending on that, we should either just use `focusHalo` directly, or override now but also update Source accordingly.

SiAdcock:
I would advise using Source's `focusHalo` for now, and if we discover problems with the current Source implementation, we can fix it there. There's definitely a point to be made about switching to `outline`, but I'd rather us tackle accessibility issues centrally than having divergent custom implementations of focus styling in different places.

shtukas (approved):

liywjl (approved):
LGTM

PR #2773, opened 2021-04-01
Update Cypress and fixtures

buck06191:
## What does this change?
- Generate share counts from a function rather than repeating the object.
  - Use this function to generate the Cypress mock response for `shareCount.json`
- Update Cypress version to improve speed.
- Use a JSON source for `most read` and `most read geo` fixtures to make sure Cypress and Storybook fixtures are kept up to date with each other.
- Fix the broken interactivity test that used the `most read` and `most read geo` endpoint.
- Update the rich link fixture to include `format` data.
- Adds ability to log messages to the Node terminal when running Cypress in headless mode (very useful for debugging where things are breaking or hanging if it's only happening on GH Actions)
## Why?
Keep the Cypress fixtures up to date and make sure that we're not failing to test updated code changes because fixtures fall out of date.


oliverlloyd (approved):
This made me think of some recent discussions around when to abstract

Thi article really nicely sums it up

https://kentcdodds.com/blog/aha-programming

In this particular case there are two arguments agains the abstraction, that one of the use cases might change in the future and that you have to open a new file to see what's happening. I don't think either is especially strong in this case so I'm happy either way.

I just thought this was a great opportunity to introduce the subject. I find it helps to have a good understanding of the pros and cons of abstracting things. Especially when it's an area with a lot of different ideas about it floating about!

buck06191:
[Automate cypress fixtures](https://trello.com/c/iKQlvd44/2764-automate-cypress-fixtures)

liywjl (approved):
LGTM :

cypress/integration/parallel-5/discussion.spec.js
```diff
@@ -17,12 +17,20 @@ describe('Discussion', function () {
 
 	it('should scroll the page to the comments section and expand it when the comment count link is clicked', function () {
 		cy.visit(`/Article?url=${articleUrl}`);
+		cy.task(
```
buck06191, line 20: Without these logs the tests for this just hang on GH Actions.

cypress/integration/parallel-3/article.interactivity.spec.js
```diff
@@ -46,23 +46,24 @@ describe('Interactivity', function () {
 				// which case this check should be updated
 				.should('eq', 2);
 		});
-		describe('When most viewed is mocked', function () {
+		// eslint-disable-next-line mocha/no-exclusive-tests
+		describe.only('When most viewed is mocked', function () {
```
oliverlloyd, line 50: We need to remove this `only`

cypress/integration/parallel-5/discussion.spec.js
oliverlloyd: I wonder if this is related to the `only` above?

scripts/test-data/gen-manual-fixtures.js
```diff
@@ -0,0 +1,95 @@
...
+
+const execa = require('execa');
+const fs = require('fs');
+const { resolve } = require('path');
+
+const root = resolve(__dirname, '..', '..');
+const jsonPath = resolve(__dirname, 'data');
+
+/**
+ * gen-manual-fixtures.js
```
oliverlloyd, line 14: I think we shouldn't use 'manual' in the name here. There's a lot of existing docs and naming where manual means the opposite of generated content, eg: fixture data that is manually created, rather than generated automatically

oliverlloyd (changes requested):

scripts/test-data/gen-manual-fixtures.js
buck06191: This is manually created though, it's just copied from a manual JSON to make sure we use the same data in both Storybook and Cypress. It doesn't actually generate it from any source other than a manually created JSON file. 
Is there a preferred term? Maybe `generated-from-manual-fixtures.js` or similar

cypress/integration/parallel-5/discussion.spec.js
buck06191: It shouldn't be because it's in a totally different test. I've removed anyway and we'll see what happens

scripts/test-data/gen-manual-fixtures.js
oliverlloyd: If this script is copying from manual data... why do we need the script? Could we not just use one file?

cypress/integration/parallel-3/article.interactivity.spec.js
```diff
@@ -46,23 +46,24 @@ describe('Interactivity', function () {
 				// which case this check should be updated
 				.should('eq', 2);
 		});
+		// eslint-disable-next-line mocha/no-exclusive-tests
```
oliverlloyd, line 49: Is this needed now?

cypress/integration/parallel-3/article.interactivity.spec.js
```diff
@@ -46,23 +46,24 @@ describe('Interactivity', function () {
...
 		describe('When most viewed is mocked', function () {
+			// eslint-disable-next-line mocha/no-mocha-arrows
```
oliverlloyd, line 51: Ditto here, is this linting exclusion required?

cypress/lib/mocks.js
```diff
@@ -1,10 +1,18 @@
...
 	// Mock most-read
-	cy.intercept('GET', '**/most-read-geo**', { fixture: 'mostReadGeo.json' });
+	cy.intercept('GET', '**/most-read-geo**', {
+		fixture: 'mostReadGeo.json',
+	}).as('getMostReadGeo');
 	// Mock most-read
-	cy.intercept('GET', '/embed/card/**', { fixture: 'richLink.json' });
+	cy.intercept('GET', '/embed/card/**', {
+		fixture: 'richLink.json',
+	}).as('getRichLink');
```
oliverlloyd, line 17: I suspect these aliases are not needed. I don't see them being used in the PR so I think they can be removed.

fixtures/manual/most-read-geo.ts
```diff
@@ -1,3 +1,16 @@
...
+ * This file was automatically generated using the gen-fixtures.js script. Any edits
+ * you make here will be lost.
+ *
+ * If the data in these fixtures is not what you expect then
+ *
+ * 1. Update the data in 'scripts/test-data/data', and then
+ * 2. Re-run node 'scripts/test-data/gen-fixtures.js'
+ *
+ */
+
```
oliverlloyd, line 13: Inserted by mistake?

fixtures/manual/most-read.ts
```diff
@@ -1,3 +1,16 @@
...
+ * This file was automatically generated using the gen-fixtures.js script. Any edits
+ * you make here will be lost.
+ *
+ * If the data in these fixtures is not what you expect then
+ *
+ * 1. Update the data in 'scripts/test-data/data', and then
+ * 2. Re-run node 'scripts/test-data/gen-fixtures.js'
+ *
+ */
+
```
oliverlloyd, line 13: Ditto. This data was manually created and lives in the /manual/ folder. Data that is autoamtically generated lives in the /generated/ folder and it is that which should not be edited

scripts/test-data/gen-manual-fixtures.js
buck06191: We're copying from JSON because Cypress needs JSON data for fixtures whilst Storybook/Jest needs a `.ts` file. We can't use one or the other because the scripts we use to run this step need to be in CommonJS. So we can't import the manual fixtures that are in Typescript because they use ECMAModules, and I want there to be a single source of truth because updating the JSON files in Cypress is a manual and horrific process. 

Doing it this way we have a single JSON file as a source of truth. When that file is updated the "manual" fixtures that are a Typescript file with a Javascript object exported get updated by converting the JSON file to that format. This is similar to the existing `gen-fixtures.js` step albeit with a manual JSON file rather than one obtained via a `GET ` request. Meanwhile, Cypress gets the JSON file as is but copied over to where it needs to exist.

It's not an ideal solution but it's the simplest one which preserves existing approaches to solving a problem, gives us a single source of truth for shared manual fixtures, and does so using JSON as an accepted way to created a shared data source that is compatible across multiple languages/module systems.

scripts/test-data/data/mostReadGeo.json
```diff
@@ -0,0 +1,225 @@
+{
```
oliverlloyd, line 1: Probably best to put this under the fixtures folder. The /scripts/ folder currently, more or less, only have scripts in it

cypress/integration/parallel-3/article.interactivity.spec.js
buck06191: Removed

cypress/integration/parallel-3/article.interactivity.spec.js
buck06191: Removed

cypress/lib/mocks.js
buck06191: They're needed for the tests to work. If you remove them it fails and that's why the test was previously failing. It couldn't find the aliases and that caused the tests to fail. I re-ran this multiple times with multiple configurations to check

fixtures/generated/most-read-geo.ts
```diff
@@ -0,0 +1,296 @@
...
+ *
+ * This file was automatically generated using the update-manual-fixtures.js script.
+ * Any edits you make here will be lost.
+ *
+ * If the data in these fixtures is not what you expect then
+ *
+ * 1. Update the data in 'fixtures/manual/data', and then
+ * 2. Re-run node 'scripts/test-data/update-manual-fixtures.js'
+ *
+ */
```
buck06191, line 12: @oliverlloyd I've updated the header comment so that it correctly identifies the script that is being used to generate this

cypress/plugins/index.js
```diff
@@ -9,6 +9,12 @@
 // ***********************************************************
 
 module.exports = (on, config) => {
+	on('task', {
+		log(message) {
+			console.log(message);
+			return null;
+		},
+	});
```
oliverlloyd, line 17: Is this needed?

fixtures/generated/most-read-geo.ts
oliverlloyd: I still feel this is breaking the model here.

the `/manual` folder only contains fixtures that were manaully created. Manual as in, not generated from a script file, that's the distinction.

But here we're explicitly inserting data that is being outpuit from a script and then adding the same script header that gets inserted by the other automatically generated data. You see?

makefile
```diff
@@ -159,6 +159,14 @@ gen-fixtures:
...
 	@node scripts/test-data/gen-fixtures.js
 
+gen-fixtures-and-schema:
+	$(call log, "Generating new article fixture data and new schema")
+	make gen-schema && make gen-fixtures
+
+update-manual-fixtures:
+	$(call log, "Updating all manual fixture data")
+	@node scripts/test-data/update-manual-fixtures.js
+
```
oliverlloyd, line 169: I think this represents my issue here. Manual fixutes are manual because they are not updated by scripts. That's the whole point of the `manual` folder. But what we're doing here is we're explicitly breaking that model by creating a automatied script to generate 'manual' data

cypress/plugins/index.js
buck06191: It's not explicitly needed here, but it's a useful script to have in place as without it you can't debug the output from the GH Actions. 

I'll remove it from this PR and move it into a separate one.

fixtures/generated/most-read-geo.ts
buck06191: The problem here is that I've not updated where this gets moved to. This file is actually in `generated/` rather than `manual/` so I just need to update the header again to reflect this.

makefile
buck06191: I'd argue that the distinction is that we're producing generated fixtures from manual data. Whereas the other script generates fixtures from automated data

buck06191:
Closing in favour of #2843

PR #2778, opened 2021-04-01
force-braze-message to use URL fragment

rhystmills:
## What does this change?

Moves `force-braze-message` to use the URL fragment rather than query strings because the new Epic exceeded the 1024 character limit set for query strings.

This also adds `force-braze-message` as an option for the Braze Epic.

Example local url [here](http://localhost:3030/Article#force-braze-message=%7B%22heading%22%3A%22Since%20you%E2%80%99re%20forcing%20a%20Braze%20Message...%22%2C%22highlightedText%22%3A%22Support%20The%20Guardian%20from%20as%20little%20as%20%25%25CURRENCY_SYMBOL%25%251%20-%20and%20it%20only%20takes%20a%20minute.%20Thank%20you.%22%2C%22buttonText%22%3A%22Support%20The%20Guardian%22%2C%22buttonUrl%22%3A%22https%3A%2F%2Fsupport.theguardian.com%2Fcontribute%22%2C%22componentName%22%3A%22Epic%22%2C%22slotName%22%3A%22EndOfArticle%22%2C%22paragraph0%22%3A%22...%20we%20have%20a%20small%20favour%20to%20ask.%20More%20people%2C%20like%20you%2C%20are%20reading%20and%20supporting%20the%20Guardian%E2%80%99s%20independent%2C%20investigative%20journalism%20than%20ever%20before.%20And%20unlike%20many%20news%20organisations%2C%20we%20made%20the%20choice%20to%20keep%20our%20reporting%20open%20for%20all%2C%20regardless%20of%20where%20they%20live%20or%20what%20they%20can%20afford%20to%20pay.%22%2C%22paragraph1%22%3A%22The%20Guardian%20will%20engage%20with%20the%20most%20critical%20issues%20of%20our%20time%20%E2%80%93%20from%20the%20escalating%20climate%20catastrophe%20to%20widespread%20inequality%20to%20the%20influence%20of%20big%20tech%20on%20our%20lives.%20At%20a%20time%20when%20factual%20information%20is%20a%20necessity%2C%20we%20believe%20that%20each%20of%20us%2C%20around%20the%20world%2C%20deserves%20access%20to%20accurate%20reporting%20with%20integrity%20at%20its%20heart.%22%2C%22paragraph2%22%3A%22Our%20editorial%20independence%20means%20we%20set%20our%20own%20agenda%20and%20voice%20our%20own%20opinions.%20Guardian%20journalism%20is%20free%20from%20commercial%20and%20political%20bias%20and%20not%20influenced%20by%20billionaire%20owners%20or%20shareholders.%20This%20means%20we%20can%20give%20a%20voice%20to%20those%20less%20heard%2C%20explore%20where%20others%20turn%20away%2C%20and%20rigorously%20challenge%20those%20in%20power.%22%2C%22paragraph3%22%3A%22We%20hope%20you%20will%20consider%20supporting%20us%20today.%20We%20need%20your%20support%20to%20keep%20delivering%20quality%20journalism%20that%E2%80%99s%20open%20and%20independent.%20Every%20reader%20contribution%2C%20however%20big%20or%20small%2C%20is%20so%20valuable.%20%22%7D).

### Before

- URL query parameters are used for `force-braze-message`
- The Braze Epic cannot be forced

### After

- URL fragment is used for `force-braze-message`
- The Braze Epic can be forced

### In use

image

### DIFF
--- a/src/web/lib/braze/forceBrazeMessage.ts
+++ b/src/web/lib/braze/forceBrazeMessage.ts
@@ -1,0 +1,50 @@
+const FORCE_BRAZE_ALLOWLIST = [
+	'preview.gutools.co.uk',
+	'preview.code.dev-gutools.co.uk',
+	'localhost',
+	'm.thegulocal.com',
+];
+
+type Meta = {
+	dataFromBraze: {
+		[key: string]: string;
+	};
+	logImpressionWithBraze: () => void;
+	logButtonClickWithBraze: (id: number) => void;
+};
+
+export const getBrazeMetaFromQueryString = (): Meta | null => {
+	if (URLSearchParams) {
+		const qsArg = 'force-braze-message';
+
+		const params = new URLSearchParams(window.location.search);
+		const value = params.get(qsArg);
+		if (value) {
+			if (!FORCE_BRAZE_ALLOWLIST.includes(window.location.hostname)) {
+				// eslint-disable-next-line no-console
+				console.log(`${qsArg} is not supported on this domain`);
+				return null;
+			}
+
+			try {
+				const dataFromBraze = JSON.parse(value);
+
+				return {
+					dataFromBraze,
+					logImpressionWithBraze: () => {},
+					logButtonClickWithBraze: () => {},
+				};
+			} catch (e) {
+				const error = e as Error;
+				// Parsing failed. Log a message and fall through.
+				// eslint-disable-next-line no-console
+				console.log(
+					`There was an error with ${qsArg}: `,
+					error.message,
+				);
+			}
+		}
+	}
+
+	return null;
+};
--- a/src/web/lib/braze/forceBrazeMessage.ts
+++ b/src/web/lib/braze/forceBrazeMessage.ts
@@ -38,8 +38,6 @@ export const getBrazeMetaFromUrlFragment = (): Meta | null => {
 					decodeURIComponent(forcedMessage),
 				);
 
-				console.log(dataFromBraze);
-
 				return {
 					dataFromBraze,
 					logImpressionWithBraze: () => {},
--- a/src/web/lib/braze/forceBrazeMessage.ts
+++ b/src/web/lib/braze/forceBrazeMessage.ts
@@ -14,20 +14,28 @@ type Meta = {
 };
 
 export const getBrazeMetaFromQueryString = (): Meta | null => {
-	if (URLSearchParams) {
-		const qsArg = 'force-braze-message';
+	if (window.location.hash) {
+		const key = 'force-braze-message';
 
-		const params = new URLSearchParams(window.location.search);
-		const value = params.get(qsArg);
-		if (value) {
+		const hashString = window.location.hash;
+		const forcedMessage = hashString.slice(
+			hashString.indexOf(`${key}=`) + key.length + 1,
+			hashString.length,
+		);
+
+		if (forcedMessage) {
 			if (!FORCE_BRAZE_ALLOWLIST.includes(window.location.hostname)) {
 				// eslint-disable-next-line no-console
-				console.log(`${qsArg} is not supported on this domain`);
+				console.log(`${key} is not supported on this domain`);
 				return null;
 			}
 
 			try {
-				const dataFromBraze = JSON.parse(value);
+				const dataFromBraze = JSON.parse(
+					decodeURIComponent(forcedMessage),
+				);
+
+				console.log(dataFromBraze);
 
 				return {
 					dataFromBraze,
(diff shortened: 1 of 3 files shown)


liywjl (commented):
Amazing work!!! I am super keen on increasing visual regression for Epic/Banner tests. A couple of questions:

- Would you be able to add a story for this Epic?
- Could you add a story for the Banner?
- As the logic is shared between the Banner and the Epic? is there a way to display one instead of the other?

rhystmills:
Thanks for your feedback! To answer your questions @liywjl :
> * Would you be able to add a story for this Epic?

We do have some stories for the banners themselves over at [braze-components.gutools.com](https://braze-components.gutools.co.uk/?path=/story/components-digitalsubscriberappbanner--default-story). The Epic is actually borrowed from [support-dotcom-components](https://github.com/guardian/support-dotcom-components/blob/main/src/components/modules/epics/ContributionsEpic.tsx), but we populate it with copy from Braze (provided by the marketing team).

> * Could you add a story for the Banner?

We have a story for the banner at the same url.

> * As the logic is shared between the Banner and the Epic? is there a way to display one instead of the other?

There is! Here is [an example link for the banner](https://preview.gutools.co.uk/sport/2019/jul/28/tour-de-france-key-moments-egan-bernal-yellow-jersey?force-braze-message=%7B%22header%22%3A%22A%20note%20to%20our%20digital%20subscribers%22%2C%22body%22%3A%22Hi%20John%2C%20did%20you%20know%20that%20as%20a%20Guardian%20digital%20subscriber%20you%20can%20enjoy%20an%20enhanced%20experience%20of%20our%20quality%2C%20independent%20journalism%20on%20all%20your%20devices%2C%20including%20The%20Guardian%20Live%20app.%22%2C%22componentName%22%3A%22DigitalSubscriberAppBanner%22%7D). (You may have to refresh a few times to get through any other higher priority messages before it will appear. Essentially we specify a `slotName` of  'Banner' rather than 'EndOfArticle' in the json encoded in the preview URL.

shtukas (approved):

src/web/lib/braze/forceBrazeMessage.ts
```diff
@@ -0,0 +1,56 @@
...
+		const hashString = window.location.hash;
+		const forcedMessage = hashString.slice(
+			hashString.indexOf(`${key}=`) + key.length + 1,
+			hashString.length,
+		);
+
+		if (forcedMessage) {
+			if (!FORCE_BRAZE_ALLOWLIST.includes(window.location.hostname)) {
+				// eslint-disable-next-line no-console
+				console.log(`${key} is not supported on this domain`);
```
oliverlloyd, line 32: Does this create a console log statement in production? Ideally (best endeavours) we should aim to have a clean console log for readers

rhystmills: It should log to the console - but only if the user has attempted to use a URL fragment including `force-braze-message` (and that will probably only happen intentionally).

I'm not quite sure I was properly checking for that before, so I've made another commit that ensures we will only log to the console if there is a URL fragment and that fragment includes 'force-braze-message'.

rhystmills: (I should add - I don't feel particularly strongly about keeping the log as I also like to keep the console clear - I think in this instance though it's unlikely anyone other than Guardian staff will see the log)

oliverlloyd: Is there a particular reason we _need_ to log this information? `console` feels like a tool for debugging.

I think it's generally best practice to remove console log statements from code that is being shipped to production unless there's a strong use case for keeping it.

adamnfish: Since this only works in preview, there's a pretty good chance someone in the future will try and drop a message elsewhere and be confused. This is exacerbated by there being loads of reasons in the code why a message might not show up (consent, login, type of subscription), which means that the journey the developer will go on to work out the problem might be a long one. At the time we thought it would be helpful to point out the user error when it occurs, and `console.log` is ideal here because the 'user' (and thus the audience for the message) will be a developer that see console messages.

Are there alternative patterns elsewhere in the codebase for warning developers when they've put things into a known invalid state, to save them having to go on a debugging journey to (re)discover that knowledge?

liywjl: I have mixed feelings about this console log, I am unsure we really want to log this on prod? Its not an error, nor something we should communicate to the user, so it just adds noise to the client IMO.

gtrufitt: FWIW, I don't hate this considering it's only there for developers who have tried to force something specifically. Maybe you want to use https://github.com/guardian/libs/blob/main/src/logger.README.md with your Team Colours?

rhystmills: We’re going to merge this for now given that:

- It will only log if the user is specifically using a URL hash containing `force-braze-message` and therefore shouldn’t impact other teams or our readers.
- It’s a pre-existing log which we’re moving to a new file.

`libs/logger` seems like a good solution but if we were to use it, we would probably want to implement it across all of our logs rather than just this one- so we will look into that in a future PR.

PR #2641, opened 2021-03-03
Special Headline Byline

oliverlloyd:
## What does this change?
Refactors `HeadlineByline` to use `palette`

### Before
Screenshot 2021-03-03 at 12 46 58

### After
Screenshot 2021-03-03 at 12 46 30

## Why?
To support the SpecialReport use case where the background is set to yellow. Eg: https://www.theguardian.com/world/commentisfree/2021/feb/14/brexit-britain-eu-covid-vaccination-fiasco

PR #2779, opened 2021-04-01
Re-add width adjustment for headline bylines

jfsoul:
## What does this change?
This adds back some width adjustment for contributor bylines that also have a contributor image, which was originally added to avoid overlapping of byline and image when the byline is a bit longer.

I think this was accidentally removed when we added `display:inline` (hence removing the width) to the div containing the byline in order to only highlight the words, not the entire width of the container in special reports. See https://github.com/guardian/dotcom-rendering/pull/2641

In order to preserve the intention of the fix for opinion special reports I've added an additional wrapper div which is what get's its width set explicitly, and the byline itself remains `display:inline` for highlighting goodness. (Hard to test this thorough because opinion special reports are currently formatted incorrectly)

I think there's still a better longer term fix where structurally the byline and contributor cutout sit alongside each other and the byline just wraps around the image... but that will require a bit more work.

### Before
image

### After
image

## Why?
To avoid overlapping contributor cutouts and bylines

### DIFF
--- a/src/web/components/HeadlineByline.tsx
+++ b/src/web/components/HeadlineByline.tsx
@@ -97,7 +97,8 @@ type Props = {
 	tags: TagType[];
 };
 
-const hasSingleContributor = (tags: TagType[]) => tags.filter((tag) => tag.type === 'Contributor').length === 1
+const hasSingleContributor = (tags: TagType[]) =>
+	tags.filter((tag) => tag.type === 'Contributor').length === 1;
 
 export const HeadlineByline = ({ format, byline, tags }: Props) => {
 	if (byline === '') {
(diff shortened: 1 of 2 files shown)


shtukas (approved):

src/web/components/HeadlineByline.tsx
```diff
@@ -126,14 +133,15 @@ export const HeadlineByline = ({ format, byline, tags }: Props) => {
...
-									: ''
-							}`}
+							className={cx(opinionWrapperStyles, {
+								[authorBylineWithImage]: hasSingleContributor(
+									tags,
+								),
+							})}
 						>
-							<BylineLink byline={byline} tags={tags} />
+							<div className={opinionStyles(palette, format)}>
```
oliverlloyd, line 142: Is there a way to merge this with the outer div? I suspect not from reading your description but if there were it's always nice to have fewer divs!

Great work by the way - thanks for fixing my glitch! Ideally one day we could use something really nice like `shape-outside` I think I did look at this way back and never fully discounted it, I just didn't have the weight of experience to know how useful it would have been and that it was worth the extra time to implement.

oliverlloyd (approved):

src/web/components/HeadlineByline.tsx
jfsoul: I think there might be if we could add the special report highlighting to the link itself rather than the wrapper - but the byline is not always just a link, it could have other text like `and` in the middle for multiple contributors.

jfsoul: Maybe a span in the `BylineLink` component, but I think you have an extra level of nesting somewhere no matter what?

PR #2780, opened 2021-04-02
`SWR` (React Hooks library for data fetching)

oliverlloyd:
## What?
Adds the SWR (https://swr.vercel.app/) library for fetching data.

Under the hood, this PR completely rebuilds the `useApi` custom hook, but it does this whilst still supporting, more or less*, the same api. Existing components that already use `useApi` don't need to change but they will now automatically benefit from additional features.

Eg. This code does not need to change

```typescript
  const { data, error } = useApi<CAPITrailTabType>(endpointUrl);
```

This still works but now if the reader backgrounds the tab and then returns to it the trail data will be automatically refetched and, if the response has changed, the component re-rendered

\*More or less means the api is the same but the `options` object is different. We only used it in one place though, to pass in `pollingInterval`. I changed it so that we can now pass the entire set of options [that swr allows](https://swr.vercel.app/docs/options).

## Why?
There's no one killer reason, just a series of small positives that, when combined, make a compelling case.

### ✅ Deduplication
If the same request is made twice, the second attempt will be served from the cache

### ✅ Fewer re-renders
Smart (deep) comparisons of api data responses prevent needless re-renders

In addition to comparisons, re-renders are only triggered if the component is actually using any changed state

### ✅ Smart re-fetching
Data is refreshed when readers return focus to the page or tab or when their network connection is restored

### ✅ Prefetching
Because SWR implements a cache, we can use techniques such as hidden divs, manually mutating the cache or link hovers to pre-fetch data that we think the reader may be about to need. The next page of comments, for example

### ✅ Smart polling
Data can be polled via `refreshInterval`. Polling is automatically disabled when the tab is backgrounded

### ✅ Retry logic
Errors are automatically retried after a defined period

### ✅ Less code
There are several places in DCR where we've written code to implement some of the features that are now provided by SWR. With this PR, we will have the opportunity to refactor that custom fetching logic, replacing it with `useApi`

### ✅ The future
If we ever wanted to adopt something like React Server Components then a precursor to that is to ensure our data fetching logic is neatly separated from our local state logic. This PR doesn't do that but it does offer a robust data fetching solution that hopefully meets all our use cases, making it easier for us to create that clean separation

## How does this change our traffic profile?
It both increases and decreases the volume of requests the page will make.

It increases requests because we will make additional calls as users refocus on tabs. It decreases requests because some calls will now be returned from the cache, and not result in a network roundtrip.

The exact profile change is dependent on reader behaviour but is unlikely to be make a big difference in either direction.

## What if I want to POST something?
Then just post it. We don't need any of the features talked about here for post requests, you don't cache posts. swr does allow [mutation of the cache](https://swr.vercel.app/docs/mutation) though which can be useful when editing data. 

## What about persistence?
SWR only uses an in memory cache by default. This is already a big improvement on what we have now. In order to persist this in memory cache over page reloads logic needs to be added to do this.

There are [examples](https://gist.github.com/derindutz/179990f266e25306601dd53b8fbd8c6a) of this being done by others online and also [a lib](https://github.com/sergiodxa/swr-sync-storage) that can be imported to offload this work but given we already have our own custom lib for accessing local storage ([@guardian/libs/storage](https://github.com/guardian/libs/blob/main/src/storage.ts)) it's perhaps logical, and probably better, for us to write this code ourselves.

The SWR cache is implemented here https://github.com/vercel/swr/blob/master/src/cache.ts and can be accessed using `import { cache } from 'swr'`. Essentially you want to subscribe to changes on the cache, writing them to the persistence store, and then implement something to read the store at boot time and populate the cache with it

## Why not React Query?
I originally looked at [React Query](https://react-query.tanstack.com/), which is also really great, but SWR has the features we need and comes in at a much smaller bundle size (5.1Kb vs. 11.5Kb). Both libs are active and popular and both expose very similar apis.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,6 +113,7 @@
         "prismjs": "^1.23.0",
         "reset-css": "^5.0.1",
         "sanitize-html": "^2.3.2",
+        "swr": "^0.5.5",
         "web-vitals": "^0.2.4",
         "youtube-player": "^5.5.2"
     },
--- a/src/web/lib/api.tsx
+++ b/src/web/lib/api.tsx
@@ -1,30 +1,8 @@
 import { useState, useEffect, useRef } from 'react';
 
-type FetchOptions = {
-	method?:
-		| 'GET'
-		| 'HEAD'
-		| 'POST'
-		| 'PUT'
-		| 'DELETE'
-		| 'CONNECT'
-		| 'OPTIONS'
-		| 'TRACE'
-		| 'PATCH';
-	headers?: {
-		'Content-Type':
-			| 'text/plain'
-			| 'multipart/form-data'
-			| 'application/json'
-			| 'application/x-www-form-urlencoded';
-	};
-	body?: string;
-	credentials?: 'omit' | 'include' | 'same-origin';
-};
-
-interface Options extends FetchOptions {
+type Options = {
 	pollInterval?: number;
-}
+};
 
 function checkForErrors(response: Response) {
 	if (!response.ok) {
@@ -36,8 +14,8 @@ function checkForErrors(response: Response) {
 	return response;
 }
 
-const callApi = (url: string, fetchOptions?: FetchOptions) => {
-	return fetch(url, fetchOptions)
+const callApi = (url: string) => {
+	return fetch(url)
 		.then(checkForErrors)
 		.then((response) => response.json());
 };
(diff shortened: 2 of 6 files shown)


oliverlloyd:
[SWR](https://trello.com/c/d8Bwljya/2782-swr)

gtrufitt (approved):
Thanks for the through write-up, this LGTM. Fetching data and the benefits of S-W-R and caching makes this a good area in which to reach for a library and this is a very reasonable size. 

The resultant code is clear to understand and being backed by the SWR docs will over time increase the user benefits, because it's more likely the benefits will be realised than trying to write much of this ourselves.

It's worth notifying @guardian/client-side-infra as I think we should look to standardise / add to the recommendations to allow peer-dependencies on modules that do data fetching.

liywjl (approved):
❤️  amazing work!! 

I am curious on your thoughts about using this with discussion rendering? We could even share it as a peer dependency?

oliverlloyd:
> I am curious on your thoughts about using this with discussion rendering? We could even share it as a peer dependency?

Great point! I haven't stepped through how it would work but this definately makes a lot of sense. We have a (purposefully and usefully) fragmented codebase where we use NPM to bring in logic and so we want to make sure we're extending any benefit we get from SWR cross lib

AWare:
swr:
image

mxdvl (commented):
Really interesting to see SWR make its way into dotcom-rendering. 👀

PR #2781, opened 2021-04-06
Remove unused code from getCountryCode

oliverlloyd:
## What does this change?
Removes some code that is no longer being used.

## Why?
Dead code makes it harder to understand what is happening in a file. Less code is simpler.

## What if we need this code in the future?
Sometimes it can be tempting to keep code which seems to be doing something useful because it just feels nice to offer a great api but our primary goals in DCR are maintainability and performance so we should only keep code if it is needed and used

### DIFF
--- a/src/web/components/SignInGate/displayRule.test.ts
+++ b/src/web/components/SignInGate/displayRule.test.ts
@@ -1,14 +1,12 @@
 import { incrementDailyArticleCount } from '@frontend/web/lib/dailyArticleCount';
 import { Article } from '@root/fixtures/generated/articles/Article';
 import { makeGuardianBrowserCAPI } from '@root/src/model/window-guardian';
-import { setCountryCodeSynchronous } from '@root/src/web/lib/getCountryCode';
 import {
 	isNPageOrHigherPageView,
 	isIOS9,
 	isValidContentType,
 	isValidSection,
 	isValidTag,
-	isCountry,
 } from './displayRule';
 
 const CAPI = Article;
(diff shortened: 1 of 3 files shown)


shtukas (approved):

oliverlloyd:
[Refactor getCountryCode](https://trello.com/c/61kqcqZj/2783-refactor-getcountrycode)

liywjl (approved):
👍

oliverlloyd:
@tjmw @vlbee @rhystmills @adamnfish @tomrf1 (+ anyone else I might have missed!)

I _think_ this change is okay, pretty sure this code isn't being touched, but I'm aware this is an area of the codebase where you all have a greater idea what needs to exist so I wanted to make sure I'm communicating this well.

### Why am I doing this?
Outside of the normal it's just nice to remove code (less code 🎉 ) I'm also doing it as a precursor to another change. I was thinking of having a custom hook to `useCountryCode`.

### What's this obsession with custom hooks?
[Custom hooks](https://reactjs.org/docs/hooks-custom.html) are more declarative and make jsx code easier to reason about.  But they're still just functions, they just have some addition conventions (enforced via lint rules and react warnings) that make them powerful tools to help scale things.

### Why is a custom hook useful here?
React is a declarative framework which means it's a real pain to make promises work, you constantly need to deal with the loading/undefined state, so any data fetched from an async source can be difficult to manage. Luckilly this is such a common problem that there are some really great solutions available, one of which, swr, we're [looking to implement](https://github.com/guardian/dotcom-rendering/pull/2780).

But to use that solution for fetching country code, we need to make a small refactor to how we do that in `App` such that we use a custom hook to call our existing `useApi` custom hook. Dont worry, composing hooks like this is common and encouraged. 

### TLDR
As part of looking into all that ^^ and working out the use cases I'd need to support, I noticed these functions were not used (so I don't need to support them, yay!)

### `asyncCountryCode`
There is, however, a use case where there's a requirement for `asyncCountryCode`. I don't plan on changing this and will make sure it is still supported. There's a question around how I do this regarding _when_ you get passed this prop, I suspect you don't want to wait for countryCode to be resolved before you get the `asyncCountryCode` promise. Don't worry, I see the issue and won't break that contract

shtukas (approved):

PR #2789, opened 2021-04-07
CommentCount: `...`

oliverlloyd:
## What does this change?
Shows an elipsis in place of the comment count when the data is still being fetched.

### Before
Screenshot 2021-04-07 at 09 05 40

### After
Screenshot 2021-04-07 at 09 04 33

## Why?
Two reasons:

1. It's not valid to show `0` if there are not truly zero comments.
2. It's more semantically correct to support an `undefined` state for `commentCount` while the data is loading rather than defaulting to `0`. It also creates simpler, more decarative code. Now `undefined` means undefined and zero means there are zero comments.

### DIFF
--- a/src/web/components/SignedInAs.tsx
+++ b/src/web/components/SignedInAs.tsx
@@ -99,7 +99,7 @@ export const SignedInAs = ({
 						color: ${neutral[60]};
 					`}
 				>
-					({commentCount || '...'})
+					({commentCount || '…'})
 				</span>
 			</h2>
 
--- a/src/web/components/SignedInAs.tsx
+++ b/src/web/components/SignedInAs.tsx
@@ -8,7 +8,7 @@ import { until } from '@guardian/src-foundations/mq';
 import { createAuthenticationEventParams } from '@root/src/lib/identity-component-event';
 
 type Props = {
-	commentCount: number;
+	commentCount?: number;
 	palette: Palette;
 	enableDiscussionSwitch: boolean;
 	user?: UserProfile;
@@ -99,7 +99,7 @@ export const SignedInAs = ({
 						color: ${neutral[60]};
 					`}
 				>
-					({commentCount})
+					({commentCount || '...'})
 				</span>
 			</h2>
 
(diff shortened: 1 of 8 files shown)


shtukas (approved):

oliverlloyd:
[`...` comments](https://trello.com/c/2ScahgOW/2790-comments)

src/web/components/SignedInAs.tsx
```diff
@@ -99,7 +99,7 @@ export const SignedInAs = ({
 						color: ${neutral[60]};
 					`}
 				>
-					({commentCount})
+					({commentCount || '...'})
```
gtrufitt, line 102: Are these three dots or an ellipsis... 😁

src/web/lib/formatCount.ts
```diff
@@ -3,7 +3,8 @@ import { integerCommas } from '@root/src/lib/formatters';
 export const formatCount = (
 	count?: number,
 ): { short: string; long: string } => {
-	if (!count) return { short: '0', long: '0' };
+	if (count === 0) return { short: '0', long: '0' };
+	if (!count) return { short: '...', long: '...' };
```
gtrufitt, line 7: Would `&hellip;` work?

src/web/components/SignedInAs.tsx
gtrufitt: ... vs …

gtrufitt (approved):
LGTM, other than being pedantic on the ellipsis (though, interestingly the style guide is less clear? https://www.theguardian.com/guardian-observer-style-guide-e#sign-in-gate:~:text=ellipsis)

src/web/components/SignedInAs.tsx
oliverlloyd: I went with the non dot dot dot version 👍

src/web/lib/formatCount.ts
oliverlloyd: Straight out …

paperboyo:
> style guide is less clear?

I’m sure that by _three dots_, style guide means a proper ellipsis glyph (even though it’s itself using three full stops there, haha). Typographically speaking, its contents could use some love… Composer and Typerighter auto-converts to ellipses, too.

PR #2793, opened 2021-04-07
Introduce localisation

shtukas:
This change is the DCR implementation of this frontend file/functionality ( https://github.com/guardian/frontend/blob/efb9c6a2909cd3a1eba01d06ce11e2db482aeba6/common/app/common/Localisation.scala ).

More exactly: 

1. We add `editionId` as additional optional parameters to `ArticleTitle` and `SeriesSectionLink`
2. Implement `localise`.

This is just the introduction of `localise.ts` and its use in `SeriesSectionLink`, there are other places we might have to apply the same function, in future PRs

### DIFF
--- a/src/lib/localisation.ts
+++ b/src/lib/localisation.ts
@@ -1,0 +1,60 @@
+/*
+	Date: April 2021
+	This is a reimplementation in DCR, for parity of the Localisation.scala file in frontend.
+	I realise that the replacement items are rather adhoc but again, we are just ensuring parity.
+*/
+
+// How to read the repacement instructions ?
+// ["US", "Film", "Movies"] , means: If the edition is US, then replace "Film" by "Movies"
+// Instructions in the caseInsensitive array must also work in lowecase, therefore replace "film" by "movies"
+
+type TranslationInstruction = [string, string, string];
+
+const caseInsensitive: TranslationInstruction[] = [
+	['US', 'Film', 'Movies'],
+	['US', 'Football', 'Soccer'],
+];
+
+const caseSensitive: TranslationInstruction[] = [
+	['US', 'in film', 'in movies'],
+	['US', 'in football', 'in soccer'],
+	[
+		'US',
+		'Football news, match reports and fixtures',
+		'Soccer news, match reports and fixtures',
+	],
+];
+
+const intructionToLowerCase = (
+	instr: TranslationInstruction,
+): TranslationInstruction => {
+	return [instr[0], instr[1].toLowerCase(), instr[2].toLowerCase()];
+};
+
+const instructionsToLowerCase = (
+	instrs: TranslationInstruction[],
+): TranslationInstruction[] => {
+	return instrs.map((i) => {
+		return intructionToLowerCase(i);
+	});
+};
+
+const allInstructions: TranslationInstruction[] = caseInsensitive.concat(
+	instructionsToLowerCase(caseInsensitive),
+	caseSensitive,
+);
+
+export const localise = (editionId: Edition | undefined, input: string) => {
+	const localiseReduceStep = (
+		str: string,
+		instr: TranslationInstruction,
+	): string => {
+		if (instr[0] === editionId) {
+			if (str === instr[1]) {
+				return instr[2];
+			}
+		}
+		return str;
+	};
+	return allInstructions.reduce(localiseReduceStep, input);
+};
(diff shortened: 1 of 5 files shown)


src/lib/localisation.ts
```diff
@@ -0,0 +1,60 @@
+/*
+	Date: April 2021
+	This is a reimplementation in DCR, for parity of the Localisation.scala file in frontend.
+	I realise that the replacement items are rather adhoc but again, we are just ensuring parity.
+*/
+
+// How to read the repacement instructions ?
+// ["US", "Film", "Movies"] , means: If the edition is US, then replace "Film" by "Movies"
+// Instructions in the caseInsensitive array must also work in lowecase, therefore replace "film" by "movies"
```
oliverlloyd, line 9: Not a big deal, and 👍 for the useful explanation, but if you use the js-docs pattern here and put this text above the `export localise` line then it will appear in a tooltip in VSCode! See [useOnce](https://github.com/guardian/dotcom-rendering/blob/44bf079a459f5c29a5c3a03d946bbd28dd5df65a/src/web/lib/useOnce.ts#L3) for an  example

src/web/components/ArticleTitle.tsx
```diff
@@ -14,6 +14,7 @@ type Props = {
 	sectionUrl: string;
 	guardianBaseURL: string;
 	badge?: BadgeType;
+	editionId?: Edition;
```
oliverlloyd, line 17:
```suggestion
	editionId: Edition;
```

Can we make this a required prop? Does that work?

src/web/components/SeriesSectionLink.tsx
```diff
@@ -17,6 +19,7 @@ type Props = {
 	sectionUrl: string;
 	guardianBaseURL: string;
 	badge?: BadgeType;
+	editionId?: Edition;
```
oliverlloyd, line 22:
```suggestion
	editionId: Edition;
```

As above

src/lib/localisation.ts
```diff
@@ -0,0 +1,60 @@
...
+		return intructionToLowerCase(i);
+	});
+};
+
+const allInstructions: TranslationInstruction[] = caseInsensitive.concat(
+	instructionsToLowerCase(caseInsensitive),
+	caseSensitive,
+);
+
+export const localise = (editionId: Edition | undefined, input: string) => {
```
oliverlloyd, line 47: I think by making the earlier props required, we can remove the need for `| undefined`, as in:

```suggestion
export const localise = (editionId: Edition, input: string) => {
```

Not sure if that siimplifies the later code here?

src/lib/localisation.ts
```diff
@@ -0,0 +1,60 @@
...
+	Date: April 2021
+	This is a reimplementation in DCR, for parity of the Localisation.scala file in frontend.
+	I realise that the replacement items are rather adhoc but again, we are just ensuring parity.
+*/
+
+// How to read the repacement instructions ?
+// ["US", "Film", "Movies"] , means: If the edition is US, then replace "Film" by "Movies"
+// Instructions in the caseInsensitive array must also work in lowecase, therefore replace "film" by "movies"
+
+type TranslationInstruction = [string, string, string];
```
oliverlloyd:
```suggestion
type Before = string;
type After = string;

type TranslationInstruction = [Edition, Before, After];
```

?

src/lib/localisation.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const caseSensitive: TranslationInstruction[] = [
+	['US', 'in film', 'in movies'],
+	['US', 'in football', 'in soccer'],
+	[
+		'US',
+		'Football news, match reports and fixtures',
+		'Soccer news, match reports and fixtures',
+	],
+];
```
oliverlloyd, line 26: Do we need two lists? Could you perhaps have one big list that contains the case variations and then you do case sensitive comparisons?

Not sure how that would work with the code but it seems like it might have a better developer experience in terms of adding more examples in future

oliverlloyd (approved):
Nice! This makes me think of Typewriter in Composer 😉 

Localisation is a big area (if you really get into it). Is this the start of something larger or should this be explicitly only about strings? If we just want to keep it about words then maybe we don't want to use the namespace `localisation`? Perhaps `localiseString`? I'm not super fussed though and I guess the decision here is based on where we see this area going.

shtukas:
Thanks for the review @oliverlloyd , but actually we are going to do this server side 🙂 I will close this once I am done.

shtukas:
Done server side: https://github.com/guardian/frontend/pull/23674

PR #2794, opened 2021-04-07
Fix CMP callbacks (v6.11.3)

mxdvl:
## What does this change?

Updated CMP library to fix `onConsentChange` callbacks.

https://github.com/guardian/frontend/pull/23673

## Why?

Callbacks were not fired on first pages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
         "@guardian/atoms-rendering": "^10.0.2",
         "@guardian/automat-client": "^0.2.16",
         "@guardian/braze-components": "^0.0.20",
-        "@guardian/consent-management-platform": "^6.11.2",
+        "@guardian/consent-management-platform": "^6.11.3",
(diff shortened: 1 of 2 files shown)


mxdvl:
The UI tests all seem related to Pinterest’s removal.

ashishpuliyel (approved):
Looks fine to me. I messaged dotcom just to be sure about pintrest – you'd think that release would have triggered the checks, not yours.

oliverlloyd:
> Looks fine to me. I messaged dotcom just to be sure about pintrest – you'd think that release would have triggered the checks, not yours.

These ghost diffs happen sometimes. It feels like a bug but it's actually for a good reason - I struggle to fully understand it though - if you really want to dig into it: https://www.chromatic.com/docs/branching-and-baselines

Or you can do what I do and just approve them (assuming they're not part of your PR - you can usually tell)

PR #2795, opened 2021-04-07
Set modulesVersion in requests to dotcom-components

tomrf1:
We now have versioning on the modules served by dotcom-components.
This change ensures that v1 is requested by DCR pages.
This will ensure backwards-compatibility when we do the upgrade to v2.

See [here](https://github.com/guardian/support-dotcom-components#module-versions)

[PR for v2 modules](https://github.com/guardian/support-dotcom-components/pull/361)

Epic request:
Screen Shot 2021-04-08 at 08 21 36

Banner request:
Screen Shot 2021-04-08 at 08 21 45

### DIFF
--- a/src/web/lib/contributions.ts
+++ b/src/web/lib/contributions.ts
@@ -24,6 +24,9 @@ const DAILY_ARTICLE_COUNT_KEY = 'gu.history.dailyArticleCount';
 const WEEKLY_ARTICLE_COUNT_KEY = 'gu.history.weeklyArticleCount';
 export const NO_RR_BANNER_TIMESTAMP_KEY = 'gu.noRRBannerTimestamp'; // timestamp of when we were last told not to show a RR banner
 
+// See https://github.com/guardian/support-dotcom-components/blob/main/module-versions.md
+export const MODULES_VERSION = 'v1';
+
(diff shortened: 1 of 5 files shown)


src/web/lib/contributions.ts
```diff
@@ -24,6 +24,9 @@ const DAILY_ARTICLE_COUNT_KEY = 'gu.history.dailyArticleCount';
 const WEEKLY_ARTICLE_COUNT_KEY = 'gu.history.weeklyArticleCount';
 export const NO_RR_BANNER_TIMESTAMP_KEY = 'gu.noRRBannerTimestamp'; // timestamp of when we were last told not to show a RR banner
 
+// See https://github.com/guardian/support-dotcom-components/blob/main/module-versions.md
+export const ModulesVersion = 'v1';
```
liywjl, line 28: As this is a static variable that never changes, we can follow the convention of the upper snake case
```suggestion
export const MODULES_VERSION = 'v1';
```

liywjl (approved):
LGTM 👍 

Could you update https://github.com/guardian/dotcom-rendering/blob/main/src/web/lib/setAutomat.ts by linking or adding some text related to the current contract?

We should document the current contract between Automat and RR here to help future developers understand where else to look.

PR #2788, opened 2021-04-07
useDiscussion (again)

oliverlloyd:
## What does this change?

This PR reverts [the revert](https://github.com/guardian/dotcom-rendering/pull/2787) to again reintroduce `useDiscussion` as a custom hook.

Original PR: https://github.com/guardian/dotcom-rendering/pull/2782

The difference here is we no longer return `true` if `isClosedForComments` is `false`

### Re-enabled Cypress test
Something of a drive-by change, I re-enabled a disabled Cypress test. I was trying to add more testing coverage but in the end I got blocked by restrictions with the identity api (not allowing logins from localhost) but this test does now seem to work and so I optimistically left it in.

PR #2806, opened 2021-04-08
Slower checks for View More button

oliverlloyd:
## What does this change?
This PR is a bit of a punt. I recently [re-enabled a failing Cypress](https://github.com/guardian/dotcom-rendering/pull/2788) test hoping that the test would continue to work well (it seemed to be okay) but looking at our recent PRs it again seems to be flakey in TeamCity (it works fine in Github 🤷 ).

So my hypothesis is that TeamCity is a slower environment and that by reducing the speed of the test and by checking things more I can add some resilience.

🤞


nicl (approved):
Go for it!

shtukas (approved):

PR #2811, opened 2021-04-08
Add Caption to interactiveBlockElement

gtrufitt:
## What does this change?

Adds Caption to the interactiveBlockElement as it is now passed from Frontend.

### Before

image

### After

image

### DIFF
--- a/src/web/components/elements/InteractiveBlockComponent.tsx
+++ b/src/web/components/elements/InteractiveBlockComponent.tsx
@@ -5,6 +5,7 @@ import { body } from '@guardian/src-foundations/typography';
 import { space } from '@guardian/src-foundations';
 import { neutral } from '@guardian/src-foundations/palette';
 import { Placeholder } from '@root/src/web/components/Placeholder';
+import { Caption } from '@root/src/web/components/Caption';
 import libDebounce from 'lodash.debounce';
 
 type Props = {
@@ -12,6 +13,9 @@ type Props = {
 	scriptUrl?: string;
 	alt?: string;
 	role?: RoleType;
+	caption?: string;
+	format: Format;
+	palette: Palette;
 };
 
 /*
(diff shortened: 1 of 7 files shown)


shtukas (approved):

src/web/components/elements/InteractiveBlockComponent.tsx
```diff
@@ -263,27 +270,36 @@ export const InteractiveBlockComponent = ({
...
+				ref={wrapperRef}
+				className={wrapperStyle(role, loaded)}
+			>
+				{!loaded && (
+					<Placeholder
+						height={decideHeight(role)}
+						shouldShimmer={false}
+					/>
+				)}
+				{!loaded && (
```
oliverlloyd, line 285: Can this !loaded check merge with the one above?

src/web/components/elements/InteractiveBlockComponent.tsx
```diff
@@ -263,27 +270,36 @@ export const InteractiveBlockComponent = ({
...
+						ref={placeholderLinkRef}
+						data-name="placeholder"
+						className={placeholderLinkStyle}
+						href={url}
+					>
+						{alt}
+					</a>
+				)}
+			</div>
+			{caption && (
```
oliverlloyd, line 296: This might be doing the right thing, not sure, but maybe the logic around `loaded` or not can be raised slightly so we `return`. That way we avoid the render falling through to here if we don't want it to?

oliverlloyd (approved):
Loving that nice blue triangle

src/web/components/elements/InteractiveBlockComponent.tsx
gtrufitt: The `loaded` is referencing loading from external to React, not the loading of the component itself. We always want to render the caption if it exists, so I think this is correctly structured.

oliverlloyd: Ah, gotcha 👍

PR #2815, opened 2021-04-08
Filter out blank links in `SubMeta`

oliverlloyd:
## What does this change?
Ensures that we don't try to render links with blank titles

### Before
Screenshot 2021-04-08 at 16 21 24

### After
Screenshot 2021-04-08 at 16 20 15

## Why?
Because the slash looks odd sitting down there by itself

## Why not?
Because this type of data manipulation would be better done on the server, instead of the client. I can't think why a blank link would be valid

### DIFF
--- a/src/web/components/SubMeta.tsx
+++ b/src/web/components/SubMeta.tsx
@@ -121,8 +121,12 @@ export const SubMeta = ({
 	showBottomSocialButtons,
 	badge,
 }: Props) => {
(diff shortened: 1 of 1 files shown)


liywjl (approved):
👍

mxdvl (commented):
Gone 😃

mxdvl (approved):

oliverlloyd:
Closing hapilly in favour of the same fix being implemented on the server https://github.com/guardian/frontend/pull/23688

PR #2838, opened 2021-04-12
AB test remote importing of RR links in header

tomrf1:
Currently it's difficult to change and AB test the RR links space in the header.
This PR introduces an AB test to measure the impact of serving this component as a module from support-dotcom-components. See https://github.com/guardian/frontend/pull/23703
We already use this mechanism for epics and banners.

This AB test does not include frontend. If the variant performs no worse then we can migrate frontend to use the same mechanism.

The PR also changes `submitComponentEvent` to take the `ophanRecord` function as a dependency. This is to enable storybook testing of the component, where `ophan` is not on the window object. Various places in this codebase also use ophan, but I didn't want to do the full refactor in this PR.

There are slight style differences because the variant uses src-button, but I don't think it'll make a big difference.
#### control
Screen Shot 2021-04-19 at 10 01 41

#### remote
Screen Shot 2021-04-19 at 10 18 31

#### control
Screen Shot 2021-04-19 at 10 02 05

#### remote
Screen Shot 2021-04-19 at 10 02 14

#### control
Screen Shot 2021-04-19 at 10 19 39

#### variant
Screen Shot 2021-04-19 at 10 19 15

#### control
Screen Shot 2021-04-19 at 10 20 42

#### variant
Screen Shot 2021-04-19 at 10 20 52

#### control
Screen Shot 2021-04-19 at 10 21 53

#### variant
Screen Shot 2021-04-19 at 10 21 45

### DIFF
--- a/cypress/integration/parallel-4/signedin.spec.js
+++ b/cypress/integration/parallel-4/signedin.spec.js
@@ -27,9 +27,9 @@ describe('Signed in readers', function () {
 			// this commercial error from failing this test
 			return false;
 		});
-		cy.visit(`Article?url=${articleUrl}`);
 		// Mock call to 'profile/me'
 		cy.intercept('GET', '**/profile/me', profileResponse);
+		cy.visit(`Article?url=${articleUrl}`);
 		// This text is shown in the header for signed in users
 		cy.contains('My account');
 	});
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -68,6 +68,8 @@ import { SpotifyBlockComponent } from '@root/src/web/components/elements/Spotify
 import { VideoFacebookBlockComponent } from '@root/src/web/components/elements/VideoFacebookBlockComponent';
 import { VineBlockComponent } from '@root/src/web/components/elements/VineBlockComponent';
 import type { BrazeMessagesInterface } from '@guardian/braze-components/logic';
+import { remoteRRHeaderLinksTestName } from '@root/src/web/experiments/tests/remoteRRHeaderLinksTest';
+import { OphanRecordFunction } from '@root/node_modules/@guardian/ab-core/dist/types';
 import {
 	submitComponentEvent,
 	OphanComponentEvent,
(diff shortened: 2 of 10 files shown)


cypress/integration/parallel-4/signedin.spec.js
```diff
@@ -27,13 +27,15 @@ describe('Signed in readers', function () {
 			// this commercial error from failing this test
 			return false;
 		});
-		cy.visit(`Article?url=${articleUrl}`);
 		// Mock call to 'profile/me'
 		cy.intercept('GET', '**/profile/me', profileResponse);
+
+		cy.visit(`Article?url=${articleUrl}`);
```
tomrf1, line 33: this test now breaks if the intercept isn't set up earlier

src/web/components/App.tsx
```diff
@@ -135,26 +137,10 @@ const GetMatchStats = React.lazy(() => {
 type Props = {
 	CAPI: CAPIBrowserType;
 	NAV: BrowserNavType;
+	ophanRecord: OphanRecordFunction;
```
tomrf1, line 140: now passed in from `BootReact`

src/web/components/ReaderRevenueLinks.tsx
```diff
@@ -117,13 +133,190 @@ const subMessageStyles = css`
...
+interface SupportHeaderData {
+	module: {
+		url: string;
+		name: string;
+		props: SupportHeaderProps;
+	};
+	meta: TestMeta;
+}
+
+const ReaderRevenueLinksRemote: React.FC<{
```
tomrf1, line 158: This is the component used by the variant in the AB test. It renders the remotely-fetched module

src/web/components/ReaderRevenueLinks.tsx
```diff
@@ -117,13 +133,190 @@ const subMessageStyles = css`
...
+							sendOphanComponentEvent(
+								'INSERT',
+								meta,
+								ophanRecord,
+							);
+						},
+					);
+			})
+			.catch((error) => {
+				// TODO - sentry
```
tomrf1, line 242: we have a ticket to add sentry logs for epic/banner failures as well, so we'll do them together

src/web/components/ReaderRevenueLinks.tsx
```diff
@@ -117,13 +133,190 @@ const subMessageStyles = css`
...
+				<SupportHeader {...supportHeaderResponse.module.props} />
+				{/* eslint-enable react/jsx-props-no-spreading */}
+			</div>
+		);
+	}
+
+	return null;
+};
+
+export const ReaderRevenueLinksNative: React.FC<Props> = ({
```
tomrf1, line 262: This is the control in the AB test. It's essentially the same as before, though I've re-ordered the two links for the UK edition, and improved the tracking for the AB test

shtukas (approved):

PR #2839, opened 2021-04-12
Remove unused nanoevents package

liywjl:
## What does this change?
We do not use `nanoevents` therefore we should remove it


shtukas (approved):

liywjl:
noticed this package thanks to dependabot https://github.com/guardian/dotcom-rendering/pull/2837

PR #2840, opened 2021-04-12
Better images for everyone

oliverlloyd:
## What?
This PR introduces the ability to enhance all images in ways that were previously only available to photo essays

### The enhancements
The following conventions have always existed for photo essays and *some* will now also work for other articles

#### Side by side images (`all`)
If two `halfWidth` images are inserted in sequence then they will be positioned side by side using the `MultiImageBlockComponent`

#### Caption override (`all`)
If the text element immediately following an image is marked as an unordered list (`<ul><li>New caption</li></ul>`) and if any of the prior images were placed side by side, then this text value will replace the caption used for the previous image.

#### Caption removal (`all`)
Side by side images put into a `MultiImageBlockComponent` will have their captions removed. Note. For photo essays, we _always_ remove image captions as there is a convention to expect caption overrides.

Eg.:

```
text
text
image (immersive) <--- Does NOT have caption removed
text
image (halfWidth)  <--- Has caption removed
image (halfWidth)  <--- Has caption removed
ul/li caption text     <--- New caption!
text
```

#### Image Title (`photo essays only`)
If the text element immediately following an image is marked as a subheading (`<h2>Image Title</h2>`) and if the preceding image is not `halfWidth`, then this subheading will be set as the `title` prop for the `ImageBlockElement`, causing it to be overlaid on the image.

Support for this convention on other article types outside of photo essays is *not* included as part of this PR

## Pending work in Composer
In order to enable these features on articles outside of photo essays a change is required in Composer to allow the `halfWidth` role to be set on images if the article `displayHint` is not `PhotoEssay` 🙏 

## Removing the conventions
Ideally, we would not need to rely on brittle, hard to learn conventions for communicating desires on how images should be rendered. It would be better if Composer were to support a native version of the `MultiImageBlockElement` 🙏 

## Why?
Screenshot 2021-04-12 at 13 07 46


shtukas (approved):

oliverlloyd:
[PhotoEssays for all](https://trello.com/c/Om6fNHKA/2805-photoessays-for-all)

gtrufitt:
cc @JamieB-gu @joecowton1

liywjl (approved):
A lot is going on and a lot of logic that makes it hard to follow. But granted this because the problem it is solving is a complex one

src/model/enhance-images.ts
```diff
@@ -0,0 +1,333 @@
...
+							caption,
+						});
+					} else {
+						processed.push({
+							_type:
+								'model.dotcomrendering.pageElements.MultiImageBlockElement',
+							elementId: prevHalfWidth.elementId,
+							images: [prevHalfWidth, image],
+						});
+					}
```
gtrufitt, line 142: Not tested, but wonder if the difference between the returned objects might be better communicated with something like:

```suggestion

						const maybeCaption = endOfBuffer? {caption} : {};
						processed.push({{
							_type:
								'model.dotcomrendering.pageElements.MultiImageBlockElement',
							elementId: prevHalfWidth.elementId,
							images: [prevHalfWidth, image],
						}, ...maybeCaption});
```

src/model/enhance-images.ts
```diff
@@ -0,0 +1,333 @@
...
+			case 'inline':
+			case 'immersive':
+			case 'showcase':
+			case 'supporting':
+			case 'thumbnail':
+			default:
+				if (endOfBuffer) {
+					processed.push(enhanceImage({ image, caption, title }));
+					// Mop up any dangling halfWidth images that never had a sibling
+					if (prevHalfWidth)
```
gtrufitt, line 155: Might be worth thinking on the name of `prevHalfWidth` because I stumbled a couple of times, something like `prevHalfWidthImageBlockElement` (for brevity obvs...).

src/model/enhance-images.ts
```diff
@@ -0,0 +1,333 @@
...
+								image: prevHalfWidth,
+								caption,
+								title,
+							}),
+						);
+				} else {
+					processed.push(enhanceImage({ image }));
+					// Mop up any dangling halfWidth images that never had a sibling
+					if (prevHalfWidth)
+						processed.push(enhanceImage({ image: prevHalfWidth }));
```
gtrufitt, line 167: This comment *may* not make sense, because we're not at the end of the buffer and the *following* sibling might be a halfWidth image?

src/model/enhance-images.ts
```diff
@@ -0,0 +1,333 @@
...
+				const caption = decideCaption({
+					element,
+					isPhotoEssay,
+					imageBuffer,
+				});
+
+				if (caption && imageBuffer.length > 0) {
+					const nextSubheading = elements[
+						i + 1
+					] as SubheadingBlockElement;
```
gtrufitt, line 231: Are we *sure* it's a `subheadingBlockElement`? Is a type guard useful?

src/model/enhance-images.test.ts
```diff
@@ -1,17 +1,10 @@
 import { Pillar } from '@guardian/types';
```
gtrufitt, line 1: A `ul` with more than one `li` (ie, not a Caption?) could be an edge case?

gtrufitt (approved):
You've managed to make a super complex piece of logic simple to follow here, kudos!

src/model/enhance-images.ts
```diff
@@ -0,0 +1,333 @@
...
+	// halfWidth images in the buffer
+	const countOfHalfWidthImages = imageBuffer.filter(
+		(image) => image.role === 'halfWidth',
+	).length;
+	if (!isPhotoEssay && countOfHalfWidthImages < 2) return '';
+	// Extract the caption
+	const frag = JSDOM.fragment(element.html);
+	if (!frag || !frag.firstChild || !frag.firstElementChild) return '';
+	const hasULwrapper = frag.firstChild.nodeName === 'UL';
+	const containsLItags = frag.firstElementChild.outerHTML.includes('<li>');
```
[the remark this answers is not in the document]
oliverlloyd: Belive it or not, [captions with multiple lines](https://www.theguardian.com/politics/2019/dec/10/on-the-road-with-boris-johnson-a-photo-essay?dcr=true) is a thing 🤷 . It's often used to separate the text out, one line for each image. I suspect actually this might be part of the reasoning for using ul/li as the 'convention' here?

Screenshot 2021-04-12 at 13 34 25

src/model/enhance-images.ts
```diff
@@ -0,0 +1,333 @@
...
+
+	elements.forEach((element, i) => {
+		switch (element._type) {
+			case 'model.dotcomrendering.pageElements.ImageBlockElement':
+				// Buffer image in an array for processing later
+				imageBuffer.push(element);
+				break;
+			case 'model.dotcomrendering.pageElements.SubheadingBlockElement':
+				if (imageBuffer.length > 0) {
+					const nextTextBlock = elements[i + 1] as TextBlockElement;
```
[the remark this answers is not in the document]
oliverlloyd: Yeah, you're right, this feels a bit off. I'll play with this to see if I can improve it

src/model/enhance-images.ts
oliverlloyd: I think the destructuring is a bit hard to parse (at least I had to squint for a bit to get it!) but I defo agree this `if` statement could/should be removed. I'll see if I can find a way to do it with less squinting otherwise I'll fallback to this 👍

src/model/enhance-images.ts
oliverlloyd: If you had to write this comment then my comment isn't good enough

src/model/enhance-images.test.ts
oliverlloyd: As per my other comment - they're a thing. I think this, more than anything else, underscores why these conventions have problems and need to be replaced with more resilient solutions soon

src/model/enhance-images.ts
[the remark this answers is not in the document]
gtrufitt: oof, oh right!

src/model/enhance-images.ts
gtrufitt: Yeah, totally not wedded to the destructuring!

src/model/enhance-images.ts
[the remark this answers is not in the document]
paperboyo: Hmm. When we do this properly in Composer, this is a great example: _Above_, _Right_, _Far right_ don’t work on mobile! 🤦‍♂️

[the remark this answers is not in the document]
oliverlloyd: @paperboyo That's a great point! Examples like this are so much more powerful than developers complaining about apis

src/model/enhance-images.ts
```diff
@@ -0,0 +1,333 @@
...
+						...processBuffer({
+							imageBuffer,
+							caption: nextCaption,
+							title,
+							isPhotoEssay,
+						}),
+					);
+					imageBuffer = [];
+					// If this subheading block wasn't a title, pass it through
+					if (!title) enhanced.push(element);
```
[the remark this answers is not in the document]
oliverlloyd: This is one of the more complex areas of this code. So here, when checking the subheadings, it's either going to be an image title or just a normal subheading.

If it's an image title (decided by `decideTitle`) then it gets set as a prop on the image. If it's not, then we just pass it through. 

We're doing the pass though in two places here though, once when there's a buffer and once when there's not. I played around with trying to get this code clearer but it's kind of like playing whack-a-mole, as soon as you clean up one thing another piece of state pops up needing repetition 🐇 ⚫

src/model/enhance-images.ts
oliverlloyd: You were right, this was a very subtle edge case bug - great spot!

oliverlloyd: Oh, actually, scratch that. We do need this. Let me refactor this so it makes more sense

oliverlloyd: Okay, may afternoon brain is fried and I have meetings now. I'll come back to this in the morning when I expect it will all seem a lot easier.

nicl (commented):
I wonder if we could improve the readability of the main `enhance` logic by separating the identifying-candidate-multi-blocks from the transformation of them? E.g. something like:

```
interface MultiImageBlockWindow {
	startIndex: number;
	endIndex: number;
	blockElements: CAPIElement[];
}

const asMultiImageBlockElement = (
	elements: CAPIElement[],
): MultiImageBlockElement | null => {
	return ???;
};

const findMultiImageBlocks = (
	elements: CAPIElement[],
): MultiImageBlockWindow[] => {
	return ???;
};

const enhanceMultiblockElements = (
	elements: CAPIElement[],
	isPhotoEssay: boolean,
): CAPIElement[] => {
	let updatedElements = elements

	const windows = findMultiImageBlocks(elements);
	windows.forEach(w => {
		const multi = asMultiImageBlockElement(w)
		if (multi) {
			updatedElements.splice(w.startIndex, w.endIndex - w.startIndex, multi)
		}
	});

	return updatedElements;
};
```

At the least, I think you could pull out some of the parts of the main `switch` of the `enhance` function into helper functions with naming - as it's hard to get a sense of the overall behaviour here at the moment because it's all inlined and without descriptors. (The tests help here a lot but I think the code could also do some of this work too.)

oliverlloyd:
> separating the identifying-candidate-multi-blocks from the transformation of them

I see where you're going with that and I think I did originally have something like this but the fun part is you need to process the array of elements in it's entirety because the relative context matters. That is, you need to be able to look behind and ahead in the array (sometimes two places, not just one!). Because of this the array needs to be entire.

I agree though, it's complex. I could look at trying to break out some more code out into named functions but I'm not sure there's much left. You've got things like `decideTitle` and `decideCaption` and then `processBuffer` obviously. After that there's really just a bunch of if statements left 🤷. Agreed though, this bunch of i_nested_ f statements is complex and hard to reason about. I'm starting to wonder though if this is just what the business logic is and that sometimes complex problems result in complex code? I could also just be too deep into the problen to see the obvious answers though - quite possible!

> keep this behaviour restricted to Photo Essays for now

Actually, this is how things were before. This PR is all about moving some of the photo essay only logic and using it for multiple articles.

src/model/enhance-images.ts
```diff
@@ -0,0 +1,333 @@
...
+	isPhotoEssay,
+	imageBuffer,
+}: {
+	element?: TextBlockElement;
+	isPhotoEssay: boolean;
+	imageBuffer: ImageBlockElement[];
+}): string => {
+	//  the convention: <ul><li><Caption text</li></ul>
+	if (!element) return '';
+	// We only set a special caption on non photo essay articles if there are more than two
```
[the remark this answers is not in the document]
oliverlloyd: I didn't really address this point in the PR description. So there's a few reasons but before listing them it's also worth addressing the fact that this change doesn't come without cost. It's more code, adds more maintenance and opens up the surface area for bugs - and by a non trivial amount.

So why do it?

1. The obvious reason is because we can. This type of change would have been very hard in Frontend but is reasonably possible in DCR. Sure, this block of if statements is complex but it doesn't touch any other files 😉 . With this PR + the change in Composer to allow th `halfWidth` role we open up a rich new set of possibilities for how images can be laid out.

2. We remove the needs for workarounds like the one in [this article](https://www.theguardian.com/us-news/2020/dec/11/california-covid-19-coronavirus-surging). At the moment, if editors want to position images side by side then they use this embed. But these workarounds are very broken (wrong font, colours, poor browser support, wrong spacing, etc., etc.) and should have been removed a long time ago. They weren't because there's been no alternative.

3. The original idea came out of a *-rendering meeting where the different platforms were discussing photo essays. Both AR and Editions are yet to add support for photo essays and they are both aware of the large anount of work that is needed to get them to work. Instead, we all agreed that a better solution would be to remove them. As in, to take all the features that make a photo essay a photo essay, and make them universal. Hence this PR.

There's still a ways to go to make that happen and there are still photo essay specific things this PR does not remove but this is a good step toward that larger goal.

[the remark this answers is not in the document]
paperboyo: @nicl @oliverlloyd I think it’s safe to have it supported outside Photo Essays as without Composer changes, this won’t ever appear there. And those will need to be well-thought out before introduction. Having DCR (and AR!) full support everywhere will put a nice pressure on Composer to add support ;-)

oliverlloyd:
> though, this bunch of i_nested_ f statements is complex

Hand on heart the way I wrote `i_nested_f` there was a mistake but I love how it turned out!

nicl:
@oliverlloyd yeah it's definitely complex domain logic/somewhat unavoidable. I do think you can separate some concerns to avoid the gigantic switch and make things a bit easier to reason about and test:

1. separate out the lexing from the parsing as it were

This is essentially a grammar being applied. So break up the stages. Some data types would help here. E.g.

```
interface HalfWidthPair {
	images: ImageBlockElement[];
	titleOverride?: string;
	captionOverride?: string;
}

interface MetaOverride {
	image: ImageBlockElement;
	titleOverride?: string;
	captionOverride?: string;
}
```

(May have misunderstood all the cases so take the examples with a pinch of salt.)

2. Separate out the cases

(The different intermediary datatypes, like above, will help with this). It can help to separate out the various cases of enhancement here. Yes it might involve looping through things twice, but that doesn't really matter here. And it will make the specific patterns a lot clearer in the code.

I've got some example code I'll try and share tomorrow though!

> Actually, this is how things were before. This PR is all about moving some of the photo essay only logic and using it for multiple articles.

Doh okay. Then I would probably say I'm unsure about the PR motivation itself! But in any case, I think there are ways we can make the code a bit easier to reason about.

oliverlloyd:
> (The different intermediary datatypes, like above, will help with this). It can help to separate out the various cases of enhancement here. Yes it might involve looping through things twice, but that doesn't really matter here. And it will make the specific patterns a lot clearer in the code.

Ah, interesting. So take multiple passes through the array to slowly build up to where we want to be? That's a really clever idea!

nicl:
@oliverlloyd ha yep! But also, we should both go to bed!!! Will chat in the morning :) Have a good rest.

src/model/enhance-images.ts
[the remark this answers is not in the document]
gtrufitt: FWIW, this was also originally mentioned in the [Interactives - Page Modification Contracts](https://docs.google.com/document/d/1qYpabx_PlTWci0V7mPpY2EIvh7xY-e1Wj44m8oDkuEQ/edit#heading=h.f17refnodigk) review as a way to remove one of the ways in which interactives are used to modify the page.

oliverlloyd:
@nicl You were so right with your suggestion to take multiple passes over the array, taking a stage by stage approach instead of tryinig to do everything at the same time. I've refactored the code to adopt this pattern and it's substantially better!

shtukas (approved):

liywjl (approved):
@nicl great recommendation!! this code is a lot easier to follow!! 👍  @oliverlloyd

src/model/enhance-images.ts
```diff
@@ -0,0 +1,337 @@
...
+					...thisImage.data,
+					caption: extractCaption(nextElement),
+				},
+			} as ImageBlockElement);
+			// Remove the next element
+			elements.splice(i + 1, 1);
+		} else if (
+			isImage(thisElement) &&
+			isTitle(nextElement) &&
+			isCaption(subsequentElement)
```
nicl, line 183: Nice - really like this bit.

src/model/enhance-images.ts
```diff
@@ -0,0 +1,337 @@
...
+	}
+}
+
+const enhance = (
+	elements: CAPIElement[],
+	isPhotoEssay: boolean,
+): CAPIElement[] => {
+	if (isPhotoEssay) {
+		return (
+			new Enhancer(elements)
```
nicl, line 297: The Enhancer class is probably the only bit here I'm less keen on. It adds a load of boilerplate to wrap the functions for little value. Controversial proposal, as know `reduce` can be hard to read, but what about, e.g.:

```
const elems = [1, 2, 3];
const double = (ns: number[]): number[] => ns.map((n) => n * 2);
const half = (ns: number[]): number[] => ns.map((n) => n / 2);
[double, half].reduce((acc, fn) => fn(acc), elems);
```

(I used `number[]` here for simplicity but imagine `CAPIElement[]` instead.)

nicl (approved):
Nice! Funnily enough, this is slightly different from what I was imagining but works out really well.

PR #2845, opened 2021-04-12
Don't use `isLiveBlog`

oliverlloyd:
## What does this change?
We have a legacy boolean `isLiveBlog` on the CAPI model that has been used in the past to denote articles that are liveblogs.  This PR replaces use of this boolean with `format` instead

## Why?
Because `format` is the place to be but also because often `isLiveBLog` doesn't distinguish between live and dead and this causes bugs (we're resurecting dead blogs)

### DIFF
--- a/fixtures/manual/most-read-geo.ts
+++ b/fixtures/manual/most-read-geo.ts
@@ -240,7 +240,7 @@ export const mostReadGeo = {
 			},
 			isLiveBlog: true,
 			format: {
-				design: 'ArticleDesign',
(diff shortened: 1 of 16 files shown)


shtukas (approved):

oliverlloyd:
[Refactor `isLiveBlog`](https://trello.com/c/5TGQZSKS/2807-refactor-isliveblog)

liywjl (approved):
LGTM 👍

PR #2846, opened 2021-04-12
Use LocalMessageCache instead of InMemoryCache

rhystmills:
## What does this change?
We will start using the `LocalMessageCache` for Braze messages. This was required because the Braze Epic will often not be seen by users (as it's at the end of the article) and we will need a mechanism to persist the message beyond a single page/session.

I've tested the delivery of messages locally.

### Before
Braze messages uses the `InMemoryCache`

### After
Braze messages use the local storage-based `LocalMessageCache`.

## Why?
To support the Braze Epic and also to show Braze banners on a subsequent page if they did not display before the time limit specified by the platform.

Fronted PR is [here](https://github.com/guardian/frontend/pull/23692).

### DIFF
--- a/src/web/lib/braze/buildBrazeMessages.ts
+++ b/src/web/lib/braze/buildBrazeMessages.ts
@@ -8,7 +8,6 @@ import { record } from '@root/src/web/browser/ophan/ophan';
 import {
 	BrazeMessages,
 	BrazeMessagesInterface,
-	InMemoryCache,
 	LocalMessageCache,
 	NullBrazeMessages,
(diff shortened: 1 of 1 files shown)


tjmw (commented):
Hey I think this change looks good. I think one of the things we talked about adding at the same time was to be able to track/log (probably via Sentry) when certain events were happening. For example if a message is discarded because the cache is already at max size or if a message already in the cache is discarded because it has expired. What do you think?

shtukas (approved):

PR #2776, opened 2021-04-01
Use getZIndex to make searching easier

oliverlloyd:
## What does this change?
We lost the z-index on the header links at some point so this PR reintroduces it so that the guardian logo doesn't shadow the Search link, hijacking the click event

PR #2852, opened 2021-04-13
Fix account dropdown z-index issue

nicl:
## What does this change?

Fixes a z-index issue where the account dropdown to be partially obscured by the nav. Only affects users in the 'anchor' variant of the sticky-nav test who are also signed in. (%-wise there are 5/3 % of users in the test, so that is the upper bound, but in practice only a small number of these are signed in.)

### Before

image

### After

Screenshot 2021-04-13 at 12 11 41

## Why?

To fix things!

### DIFF
--- a/src/web/components/Links.tsx
+++ b/src/web/components/Links.tsx
@@ -13,6 +13,7 @@ import { DropdownLinkType, Dropdown } from '@root/src/web/components/Dropdown';
 import ProfileIcon from '@frontend/static/icons/profile.svg';
 import { createAuthenticationEventParams } from '@root/src/lib/identity-component-event';
 import { useOnce } from '@frontend/web/lib/useOnce';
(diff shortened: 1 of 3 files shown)


shtukas (approved):

liywjl (approved):
Great work!! 👍

oliverlloyd (changes requested):
I  [added this z index value](https://github.com/guardian/dotcom-rendering/pull/2776) to fix a problem where the Guardian logo was hijaking clicks on the Search link. If we remove it it will regress that fix

nicl:
@oliverlloyd you're right thanks. I've updated to special case the Search link. Special casing is sensible here I think as there are lots of concerns around z-index and nav links so limiting the scope to just where required feels sensible.

shtukas (approved):

oliverlloyd (approved):
Nice approach

PR #2853, opened 2021-04-13
Remove flaky test

oliverlloyd:
## What does this change?
Deletes a test that is causing slow builds and false negatives

### Before
😢 

### After
😄 

## Why?
It would be nice to have this test worknig but over the months we've spent quite a bit of time on this test and I think we've reached the point where it's not worth investing any more effort into this one.


shtukas (approved):

oliverlloyd:
[Less flake](https://trello.com/c/0eaKqkQa/2809-less-flake)

PR #2859, opened 2021-04-14
Fix discussion tests

oliverlloyd:
## What does this change?
Removes a `cy.contains` regex check that seemed to be upsetting things and causing tests to hang. I replaced it with a simple string check, merging the assertion with another test to reduce visits.


oliverlloyd:
[Fix Cypress](https://trello.com/c/I9dzAnwZ/2811-fix-cypress)

liywjl (approved):
LGTM 👍  thought CI seems to be failing (unrelated it seems though)

PR #2861, opened 2021-04-14
Change the size of middots

nitro-marky:
## What does this change?

Changes the standfirst middot so it's the same size as the text.

### Before

Screenshot 2021-04-14 at 15 52 50

### After

Screenshot 2021-04-14 at 15 52 28

### DIFF
--- a/cypress/plugins/index.js
+++ b/cypress/plugins/index.js
@@ -9,6 +9,12 @@
 // ***********************************************************
 
 module.exports = (on, config) => {
+	on('task', {
+		log(message) {
+			console.log(message);
+			return null;
+		},
+	});
 	config.env = { ...config.env, ...process.env };
 	return config;
 };
--- a/cypress/plugins/index.js
+++ b/cypress/plugins/index.js
@@ -9,12 +9,6 @@
 // ***********************************************************
 
 module.exports = (on, config) => {
-	on('task', {
-		log(message) {
-			console.log(message);
-			return null;
-		},
-	});
 	config.env = { ...config.env, ...process.env };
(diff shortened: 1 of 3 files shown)


nitro-marky:
https://trello.com/c/FzfTeb1E/2680-make-grey-standfirst-dot-the-same-height-as-the-capital-letter

nitro-marky:
[Make Grey Standfirst dot the same height as the capital letter](https://trello.com/c/FzfTeb1E/2680-make-grey-standfirst-dot-the-same-height-as-the-capital-letter)

shtukas (approved):

liywjl (commented):
Maybe use `px` instead? I cannot remember if we favour one over the other?

nitro-marky:
> Maybe use `px` instead? I cannot remember if we favour one over the other?

Same here, I'll change to px and see if anyone gets enraged 😄

paperboyo:
> I cannot remember if we favour one over the other?

No idea myself, but if one (or the other!) would mean that platform/browser/whatever font size setting would scale the type but not the dot or vice versa, I would always favour the one that would scale them both unison. Disregard if I’m spewing nonsense!

oliverlloyd:
You should write your code in pixels because it's easier to read. But we want rems in the browser because that gets resized better, so we transform it at build time using this babel plugin https://github.com/guardian/babel-plugin-px-to-rem

Is that plugin working? Maybe not, I don't recall actually seeing rems in the browser but if that's so we should still maintain consistentsy by using pixels. We can look at fixing the plugin if needed.

PR #2868, opened 2021-04-15
Add a TEMP in-memory cache to fetch the anniversary interactive atom

gtrufitt:
*WARNING*

Do not copy this implementation for long-term implementations in products. In-memory caches are not tested, supported or planned for in DCR. This is a temp 10-day solution so that the removal of the implementation is quick and simple.

## What does this change?

Adds an in-memory cache for the Anniversary interactive-atom that will be shown on article pages.

It:

- Gets the CAPI key for the atom from param store
- Fetches the atom from CAPI every 60s
- Enhances the DCRCAPI JSON with the atom at the top level (it's not a block type)

### DIFF
--- a/src/model/enhance-AnniversaryInteractiveAtom.ts
+++ b/src/model/enhance-AnniversaryInteractiveAtom.ts
@@ -1,0 +1,11 @@
+import { getAnniversaryAtomCache } from '@root/src/web/server/cacheForAnniversaryInteractiveAtom';
+
+export const enhanceAnniversaryAtom = (data: CAPIType): CAPIType => {
+	// TODO Add switch
+	// const {  } = data.config.switches;
+	if ('addAProperSwitchHere') {
+		data.anniversaryInteractiveAtom = getAnniversaryAtomCache();
+	}
+
+	return data;
+};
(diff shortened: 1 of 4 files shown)


src/model/enhance-AnniversaryInteractiveAtom.ts
```diff
@@ -0,0 +1,11 @@
+import { getAnniversaryAtomCache } from '@root/src/web/server/cacheForAnniversaryInteractiveAtom';
+
+export const enhanceAnniversaryAtom = (data: CAPIType): CAPIType => {
+	// TODO Add switch
+	// const {  } = data.config.switches;
+	if ('addAProperSwitchHere') {
```
gtrufitt, line 6: TODO

src/web/server/cacheForAnniversaryInteractiveAtom.ts
```diff
@@ -0,0 +1,85 @@
...
+		fetchAnniversaryAtom()
+			.then(refreshAnniversaryAtom)
+			.catch((e) => console.log(`fetchAnniversaryAtom - error: ${e}`));
+	}, oneMinute);
+};
+
+getGuardianConfiguration('prod')
+	.then((config: GuardianConfiguration) => {
+		const key = config.getParameter('anniversaryAtomCAPIKey');
+		if (key) {
```
nicl, line 79: Is it better to throw an error here if key is not set rather than just continue and handle this case at the top of `fetchAnniversaryAtom`?

shtukas:
Nice and simple.

nicl (approved):

liywjl (approved):
LGTM

PR #2870, opened 2021-04-15
Pagination component

oliverlloyd:
## What does this change?
Adds a `Pagination` component.

```ts
type Props = {
	format: Format;
	currentPage: number;
	totalPages: number;
	newest?: string;
	newer?: string;
	oldest?: string;
	older?: string;
};
```

Screenshot 2021-04-15 at 15 15 34

## Why?
The real question is why another one? We already have a `Pagination` component for amp. This PR builds upon that but it is using Source for most things whereas the older amp version was not.

### Where are the double chevrons like in AMP?
Source doesn't currently offer a double chevron and rather than try to shoehorn the DCR version in here I've raised an issue [to update Source](https://github.com/guardian/source/issues/802), taking a more 'Source First' approach

### DIFF
--- a/src/web/components/Pagination.tsx
+++ b/src/web/components/Pagination.tsx
@@ -1,0 +1,153 @@
+import React from 'react';
+import { css } from 'emotion';
+
+import { space } from '@guardian/src-foundations';
+import { LinkButton } from '@guardian/src-button';
+import { textSans } from '@guardian/src-foundations/typography';
+import {
+	SvgChevronLeftSingle,
+	SvgChevronRightSingle,
+} from '@guardian/src-icons';
+import { until } from '@guardian/src-foundations/mq';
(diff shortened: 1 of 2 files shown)


oliverlloyd:
[Pagination](https://trello.com/c/AyMDIq3E/2817-pagination)

src/web/components/Pagination.tsx
```diff
@@ -0,0 +1,153 @@
...
+						Previous
+					</Hide>
+				</LinkButton>
+			</Section>
+			<Section>
+				<Position>
+					<Bold>{currentPage}</Bold>
+					<Of />
+					<Bold>{totalPages}</Bold>
+				</Position>
```
liywjl, line 124: Interesting abstraction

liywjl (approved):
LGTM

PR #2872, opened 2021-04-16
Position Interactive MainMedia Better

oliverlloyd:
## What?
This PR ensures that the `isMainMedia` prop is always passed into `Figure`. It explicitly passes it in the two places where it wasn't being set and then makes the prop required so that we can't forget to set it in future

## Why?
[This piece](https://www.theguardian.com/cities/2017/feb/03/skyscrapers-vanity-height-graphics-numbers) was putting an interactive element in main media but because we were not setting `isMainMedia` it defaulted to `false` causing `Figure` to style the element as though it were part of the article body and because the element `role` was `showcase` that meant it was given a negative left margin.

## Before
Screenshot 2021-04-16 at 09 23 12

## After
Screenshot 2021-04-16 at 09 22 22

### DIFF
--- a/src/web/components/ClickToView.stories.tsx
+++ b/src/web/components/ClickToView.stories.tsx
@@ -66,7 +66,7 @@ const RoleStory = ({
 					level. Truffaut street art edison bulb, banh mi cliche
 					post-ironic mixtape
 				</p>
-				<Figure role={role}>
+				<Figure isMainMedia={false} role={role}>
 					<ClickToView
 						role={role}
 						isTracking={true}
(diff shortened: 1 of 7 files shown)


shtukas (approved):

oliverlloyd:
[Showcase interactive main media flows into left and right columns](https://trello.com/c/G8moiLea/2794-showcase-interactive-main-media-flows-into-left-and-right-columns)

jfsoul:
Nice, I think we can probably enable main media interactive elements in frontend after this.

PR #2879, opened 2021-04-19
Support Braze Epic ophanComponentId parameter

rhystmills:
## What does this change?
The Braze Epic now required Marketing to specify an `ophanComponentId`, which is used as a specific campaign identifier in our logs to Ophan.

## Why?
To allow Marketing to analyse the performance of specific variants of the Braze Epic.

[Related PR for @guardian/braze-components](https://github.com/guardian/braze-components/pull/142).

### DIFF
--- a/src/web/lib/braze/parseBrazeEpicParams.test.ts
+++ b/src/web/lib/braze/parseBrazeEpicParams.test.ts
@@ -14,6 +14,7 @@ describe('parseBrazeEpicParams', () => {
 			highlightedText: 'Example highlighted text',
 			buttonText: 'Button',
 			buttonUrl: 'https://www.example.com',
+			ophanComponentId: 'epic_123',
 		};
 
 		const expected = {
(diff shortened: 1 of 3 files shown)


src/web/lib/braze/parseBrazeEpicParams.test.ts
```diff
@@ -127,6 +136,22 @@ describe('parseBrazeEpicParams', () => {
...
+			buttonText: 'Button',
+			ophanComponentId: 'epic_123',
+		};
+
+		const got = parseBrazeEpicParams(dataFromBraze);
+
+		expect(got).toBeNull();
+	});
+
+	it('returns null when the trackingId is missing', () => {
```
adamnfish, line 154: It looks like this test has two missing fields (`buttonUrl` and `ophanComponentId`). Should this be two separate tests?

adamnfish (approved):
Question about the tests, but looks good otherwise.

liywjl (approved):
LGTM 👍

src/web/lib/braze/parseBrazeEpicParams.test.ts
rhystmills: Good spot, sorted now.

PR #2880, opened 2021-04-19
Replace `dateTime` lib with `timeAgo`

oliverlloyd:
## What does this change?
`dateTime `is a local DCR library to take an absolute date and return a relative string. This PR replaces this with `timeAgo` [from libs](https://github.com/guardian/libs/blob/main/src/timeAgo.ts)

## Why?
Libs is where it's at


shtukas (approved):

liywjl (approved):
LGTM 👍

PR #2882, opened 2021-04-20
Fix `SignedInAs`

oliverlloyd:
## What does this change?
I've updated the tests around `SignedInAs` because they started to fail. In one case we were checking for the wrong case and in another we were testing based on an assumption that was no longer true.


liywjl (approved):
👍

PR #2883, opened 2021-04-20
Mobile Nav Background

liywjl:
## What does this change?
Add background to Navbar

### Before
Screenshot 2021-04-20 at 09 34 33

### After
Screenshot 2021-04-20 at 09 34 02

## Why?
Prevents scroll of the article when on Nav

### DIFF
--- a/src/web/components/Nav/ExpandedMenu/ExpandedMenu.tsx
+++ b/src/web/components/Nav/ExpandedMenu/ExpandedMenu.tsx
@@ -56,7 +56,6 @@ const wrapperMainMenuStyles = css`
 		bottom: 0;
 		height: 100%;
 		overflow: auto;
-		/* padding-top: 6px; */
 		position: fixed;
 		right: 0;
 		will-change: transform;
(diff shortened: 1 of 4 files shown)


src/web/components/Nav/ExpandedMenu/ExpandedMenu.tsx
```diff
@@ -75,11 +56,40 @@ const mainMenuStyles = css`
 		bottom: 0;
 		height: 100%;
 		overflow: auto;
-		padding-top: 6px;
+		/* padding-top: 6px; */
```
oliverlloyd: Removeable?

shtukas (approved):

src/web/lib/getZIndex.test.tsx
```diff
@@ -2,9 +2,10 @@ import { getZIndex } from './getZIndex';
 
 describe('getZIndex', () => {
 	it('gets the correct zindex for group and sibling', () => {
-		expect(getZIndex('banner')).toBe('z-index: 13;');
-		expect(getZIndex('dropdown')).toBe('z-index: 12;');
-		expect(getZIndex('burger')).toBe('z-index: 11;');
+		expect(getZIndex('banner')).toBe('z-index: 14;');
```
oliverlloyd, line 5: Aside (not relevant to this PR): Are these tests adding much value? They need updating each time a new index is added which screams implementation details to me.

Ideally, we'd have a test that renders a component that uses this function under the hood and then asserts based on visibility?

oliverlloyd (approved):

src/web/lib/getZIndex.test.tsx
liywjl: Good point! It might be worth adding this as a health week ticket

PR #2889, opened 2021-04-21
chore: `api` > `useApi`

oliverlloyd:
## What?
Renames the file `api.ts` to `useApi.ts`

## Why?
Because this is the convention we're using. A file (generally, but not religiously) exports one thing and is named the same as the export


shtukas (approved):

oliverlloyd:
[Rename `api.ts` > `useApi.ts`](https://trello.com/c/ALllIUzN/2865-rename-apits-useapits)

liywjl (approved):
👍

PR #2890, opened 2021-04-21
Remove sticky nav test code for now

nicl:
## What does this change?

Removes sticky nav test-related code.

## Why?

The test is complete and we can now evaluate. If successful we can re-introduce some of this code, but scoped to just the successful variant (anchor or backscroll) and with improvements to performance and accessibility.


shtukas (approved):

liywjl (approved):
LGTM

PR #2891, opened 2021-04-21
Enhance NumberedLists

oliverlloyd:
## What?
Adds an enhancer for `Display.NumberedList` and uses it to inline all images

I also included a new fixture, `NumberedList` to use with tests and a series of stories.

## Why?
Because we're using the enhancer pattern to bring support for numbered lists into DCR


oliverlloyd:
[Convert thumbnail images to inline](https://trello.com/c/RGRYUAXY/155-convert-thumbnail-images-to-inline)

shtukas (approved):

nicl:
@oliverlloyd I wonder if this is better done in Frontend/upstream? In general we've tried to 'clean' data in Frontend - although I realise there are a few exceptions to this now. But the motivation is keep DCR focused on rendering rather than this kind of special logic, which is really a (nice) hack to handle historic CAPI oddities.

I don't think this has to be a blocker here - especially to get something out of the door quickly, but I'd be keen we stick to this principle going forward.

oliverlloyd:
> @oliverlloyd I wonder if this is better done in Frontend/upstream? In general we've tried to 'clean' data in Frontend - although I realise there are a few exceptions to this now. But the motivation is keep DCR focused on rendering rather than this kind of special logic, which is really a (nice) hack to tidy up historic CAPI oddities.
> 
> I don't think this has to be a blocker here - especially to get something out of the door quickly, but I'd be keen we stick to this principle going forward.

Totally agree. I think though that it's a staged process to do this.

Stage 1: Is this.
We define the enhancer locally in DCR, work out the components we will need, their props etc. This is much easier locally where changes are fast and easy (easy as in the majority of the team can do them in Typecript and they only impact us)

Stage 2: We hoist this logic into the CAPI client
We did this with `format`. It's non trivial though as it edits the CAPI object and affects other platforms. It's good to have built a clear idea of the changes needed first (stage 1)

Stage 3: We remove the cleaners completely
Is we move up _past_ CAPI, and make these changes inside Composer itself. Once we get to this sunny eldorado then there will ne no need for any cleaners at all!

nicl:
@oliverlloyd ah okay yep that sounds good. I think you're right that CAPI client is a good place for this - I was thinking Frontend but that is even better!

PR #2894, opened 2021-04-21
Faux H3s

oliverlloyd:
## What?
Adds an enhancer step to replace `<p><strong>My Faux H3</strong></p>` with an actual `h3` element.

We area also prefixing this element with a `DividerBlockElement`

## Why?
To match the designs

## Why is the divider so short?
We only have one type of divider in DCR at the moment.We could look to add a full width one if needed, @HarryFischer ?

### DIFF
--- a/src/model/enhance-numbered-lists.test.ts
+++ b/src/model/enhance-numbered-lists.test.ts
@@ -26,7 +26,6 @@ describe('Enhance Numbered Lists', () => {
 				},
 			],
 		};
-
 		const expectedOutput: CAPIType = {
 			...NumberedList,
 			blocks: [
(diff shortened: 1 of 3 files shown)


oliverlloyd:
[Handle H3 elements](https://trello.com/c/KWHvL9cY/158-handle-h3-elements)

shtukas (approved):

HarryFischer:
Yes please for divider - paragraph width would be best

liywjl (approved):
great work!! 👍

oliverlloyd:
> Yes please for divider - paragraph width would be best

Righto, I've had this on my todo list for a while. I'll spin up a new PRand merge that one into this

PR #2896, opened 2021-04-21
Full width divider

oliverlloyd:
## What?
Adds a new style of divider, `full` which stretches the full width of the article body

## Why?
Because we want to use these in NumberedLists


oliverlloyd:
[Long division](https://trello.com/c/iWUg8gi2/167-long-division)

shtukas (approved):

PR #2898, opened 2021-04-22
★★★☆☆ > `StarRatingBlockElement`

oliverlloyd:
## What does this change?
Adds the convention whereby if the enhancers sees `<p>★★★☆☆</p>` then it replaces it with a `StarRatingBlockElement`

This is restricted to `Display.NumberedList` but maybe we want this done on a wider basis,  thoughts?

Screenshot 2021-04-22 at 10 26 28

## Why?
Because this is a react app and so we want to add styles and features using components, not classnames and global css. `StarRatingBlockElement` offers consistent optimised svgs with a nice yellow background that can be resized based on props.

## Shouldn't those stars sit above the image?
Yes, this will be a later piece of work. But sometimes there isn't an image so we still need to support inline stars

### DIFF
--- a/src/model/enhance-numbered-lists.test.ts
+++ b/src/model/enhance-numbered-lists.test.ts
@@ -188,4 +188,190 @@ describe('Enhance Numbered Lists', () => {
 		};
 		expect(enhanceNumberedLists(input)).toEqual(expectedOutput);
 	});
+
+	it('replaces ★★★★☆ with the StarRating component', () => {
+		const input: CAPIType = {
+			...NumberedList,
+			blocks: [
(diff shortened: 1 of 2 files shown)


shtukas (approved):

oliverlloyd:
[Ascii stars (★★★★☆) are converted to `StarRating` components](https://trello.com/c/77Iqcvqm/156-ascii-stars-%E2%98%85%E2%98%85%E2%98%85%E2%98%85%E2%98%86-are-converted-to-starrating-components)

liywjl (approved):
Great work 👍

PR #2899, opened 2021-04-22
full review link component and enhancement

liywjl:
## What does this change?
Adds support for full review link in numbered lists

- Enhance `TextBlockElement` to `FullReviewLink`
- create FullReview component
- add component to `ElementRenderer`

### Example article:
https://www.theguardian.com/technology/2019/dec/17/best-smartphone-2019-iphone-oneplus-samsung-and-huawei-compared-and-ranked

### Before
Screenshot 2021-04-22 at 12 50 00

### After
Screenshot 2021-04-22 at 18 02 53

## Why?
Support for full review links


src/lib/content.d.ts
```diff
@@ -151,6 +151,13 @@ interface ExplainerAtomBlockElement {
 	role?: RoleType;
 }
 
+interface FullReviewElement {
```
oliverlloyd, line 154: The convention here is `FullReviewBlockElement`

src/model/enhance-numbered-lists.ts
```diff
@@ -167,6 +167,65 @@ const addH3s = (elements: CAPIElement[]): CAPIElement[] => {
...
+		element._type !== 'model.dotcomrendering.pageElements.TextBlockElement'
+	) {
+		return false;
+	}
+	const frag = JSDOM.fragment(element.html);
+	if (!frag || !frag.firstElementChild) return false;
+
+	const hasULWrapper = frag.firstElementChild.nodeName === 'UL';
+	const hasLINestedWrapper =
+		frag.firstElementChild.firstElementChild &&
```
oliverlloyd, line 183: Optional chaining?

src/model/enhance-numbered-lists.ts
```diff
@@ -167,6 +167,65 @@ const addH3s = (elements: CAPIElement[]): CAPIElement[] => {
...
+
+	const hasPWrapper = wrapper && wrapper.nodeName === 'P';
+	const containsStrongTags =
+		wrapper && wrapper.outerHTML.includes('<strong>');
+	const containsLinks = wrapper && wrapper.outerHTML.includes('<a ');
+
+	const hasFullReviewText =
+		wrapper &&
+		wrapper.firstElementChild &&
+		wrapper.firstElementChild.textContent === 'Full review:';
```
oliverlloyd, line 198: Is it certain that this text is always present?

src/model/enhance-numbered-lists.ts
```diff
@@ -167,6 +167,65 @@ const addH3s = (elements: CAPIElement[]): CAPIElement[] => {
...
+
+	return (
+		!!hasULWrapper &&
+		!!hasLINestedWrapper &&
+		!!hasPWrapper &&
+		!!containsStrongTags &&
+		!!containsLinks &&
+		!!hasFullReviewText
```
oliverlloyd, line 206: Do you need the double bangs?

src/web/components/elements/FullReview.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { css } from 'emotion';
+
+import { headline, textSans } from '@guardian/src-foundations/typography';
+import { space } from '@guardian/src-foundations';
+
+import { pillarPalette } from '@root/src/lib/pillars';
```
oliverlloyd, line 7: We should be using `palette`. Never `pillarPalette`.

src/web/components/elements/FullReview.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+type Props = {
+	href: string;
+	title: string;
+	pillar: Pillar;
+};
+export const FullReview = ({ href, title, pillar }: Props) => {
+	return (
+		<>
+			<hr />
```
oliverlloyd, line 17: We could replace this higher up with a Divider component?

src/web/lib/ElementRenderer.tsx
```diff
@@ -286,6 +287,17 @@ export const ElementRenderer = ({
...
 			);
+		case 'model.dotcomrendering.pageElements.FullReviewLink':
+			return (
+				<>
+					<FullReview
+						href={element.href}
+						title={element.title}
+						pillar={format.theme}
+					/>
+					<DividerBlockComponent size="full" />
```
oliverlloyd, line 298: As per the divider comment above, you could maybe try using two `DividerBlockComponent`s here? Not sure if that works. I think we should avooid `hr`s if we can

And actually, this html should only return `FullReivew`, the Dividers should be inserted by the enhancer - there's examples of this being done in other enhancers

oliverlloyd (commented):
Just a few things. I wonder though if this text 'Full review' is dependable enoough to write everything on top of? That seems like the contract we're creating here is that the journalist _must_ use these exact words?

src/model/enhance-numbered-lists.ts
liywjl: good point, no https://www.theguardian.com/world/2020/mar/25/coronavirus-key-questions-answers-everything-you-need-to-know-covid-19

Full review no longer makes sense as a component name

src/model/enhance-numbered-lists.ts
liywjl: because every value can be `boolean`, `undefined` or `null`

src/web/components/elements/FullReview.tsx
liywjl: as in `import { news } from '@guardian/src-foundations/palette';`?

Do we not want the text colour to be linked to the pillar?

src/model/enhance-numbered-lists.ts
oliverlloyd: I think you could remove them and it would still work? I find it's easier to read without them

src/web/components/elements/FullReview.tsx
oliverlloyd: In DCR, we use `palette` for colours. As in [decidePalette](https://github.com/guardian/dotcom-rendering/blob/ea111975f28140ca6fabd9c357ce5428bb1c32bb/src/web/lib/decidePalette.ts#L720).

Because we have our _own_ `palette` we never use the Source one (except in that _one_ file). If you search DCR for `pillarPalette` (outside of amp) you will only find it being used in 3 files. Two of those files are not used anywhere and are dead, the other one is `decidePalette`

src/model/enhance-numbered-lists.ts
liywjl: TS isnt happy
Screenshot 2021-04-23 at 09 31 47

src/web/components/elements/FullReview.tsx
liywjl: I have removed `pillarPalette`, instead letting the global scope set the colour of `a` tags

src/model/enhance-numbered-lists.ts
oliverlloyd: Sorry to harp on herebut I'm inclined to dig into this more as it potentially signals there might be a logic issue. Seems like the issue is with `wrapper`, that is potentially `undefined` which then carries through. So what should happen if `wrapper` is undefined?

But perhaps we don't care? Maybe this logic can be simplified? Looking at FE, I see they use this

```
      // Styled link/section end
      document.select("ul > li:only-child").asScala.foreach { li =>
        val ul = li.parent();
        ul.addClass("article-link")
      }
```

So the convention there is, has UL tag with only one LI tag as a child. So maybe we don't need to examine the third child like:

```
const wrapper = frag.firstElementChild?.firstElementChild?.firstElementChild;
```

One option would be to return true when we see `<ul><li>{someHtml}</li></ul>`and then dangerously set that html? That would seem to respect the rules laid out in FE and also give us maximum flexibility. This also gets us around the problem where we're expeccting 'Full review: {someLink}'

src/lib/content.d.ts
```diff
@@ -151,6 +152,13 @@ interface ExplainerAtomBlockElement {
 	role?: RoleType;
 }
 
+interface FullReviewBlockElement {
+	_type: 'model.dotcomrendering.pageElements.ItemListLink';
```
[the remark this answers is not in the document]
oliverlloyd: Also, also (sorry) I worry `ItemListLink` is a bit vague. This is the external api so there's no knowledge here that this is explicit to numbered lists. Maybe `NumberedFooter`?

I think sematically it makes sense to think of it as a footer. It's the last element for each item and contains additional info about the main item content. What do you think?

liywjl:
superseded by https://github.com/guardian/dotcom-rendering/pull/2912

PR #2900, opened 2021-04-22
Overlay stars

oliverlloyd:
## What?
Adds an enhancer where if a numbered list element is preceded by stars, then we take that rating count and add it to the image (via the `starRating` prop)

## Why?
Because we already use this prop on `ImageComponent` so it makes sense to adapt to use it here as well.

## Before
Screenshot 2021-04-22 at 14 23 48

## After
Screenshot 2021-04-22 at 14 23 18

### DIFF
--- a/src/lib/content.d.ts
+++ b/src/lib/content.d.ts
@@ -211,6 +211,7 @@ interface ImageBlockElement {
 	displayCredit?: boolean;
 	role: RoleType;
 	title?: string;
+	starRating?: number;
 }
 
 interface InstagramBlockElement extends ThirdPartyEmbeddedContent {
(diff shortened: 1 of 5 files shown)


src/lib/content.d.ts
```diff
@@ -211,6 +211,7 @@ interface ImageBlockElement {
 	displayCredit?: boolean;
 	role: RoleType;
 	title?: string;
+	starRating?: number;
```
oliverlloyd, line 214: We don't get this from CAPI yet, but we _could_

oliverlloyd:
[Stars are overlaid onto images if present](https://trello.com/c/GaX7rJb0/157-stars-are-overlaid-onto-images-if-present)

shtukas (approved):

liywjl (approved):
👍

PR #2901, opened 2021-04-22
Round thumbnails

oliverlloyd:
## What does this change?
This reverts an earlier change where we were making all images inline. Instead, we now respect `role` and also mark `thumbnail` as round.

We use a new prop `isAvatar` to denote roundness.

### Before
Screenshot 2021-04-22 at 15 19 26

### After
Screenshot 2021-04-22 at 15 14 33

## Why not use `Avatar`?
Although the name is spot on we actually want to use the `ImageComponent` because that is using the Picture tag

### DIFF
--- a/src/lib/content.d.ts
+++ b/src/lib/content.d.ts
@@ -212,6 +212,7 @@ interface ImageBlockElement {
 	role: RoleType;
 	title?: string;
 	starRating?: number;
+	isAvatar?: boolean;
 }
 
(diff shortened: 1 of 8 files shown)


oliverlloyd:
[Circular thumbnails](https://trello.com/c/tOuelJ8k/168-circular-thumbnails)

shtukas (approved):

liywjl (approved):
👍

PR #2902, opened 2021-04-22
`pillarPalette_DO_NOT_USE`

oliverlloyd:
## What?
This PR deprecates the function `pillarPalette` by renaming it to `pillarPalette_DO_NOT_USE`

This function should only be used in `decidePalette` (I did find two other places in the code using this function but they were both dead so I deleted them as a drive by here).

## Why?
Because, as of today, this function is no longer used except in one place, in `decidePalette` (excepting in some amp code). But because it has such strong code muscle memory we still see attempts to re-introduce it into code which we want to discourage. By using `palette` we centralise theming logic allowing it to be better understood and also hoisted and shared.


shtukas (approved):

liywjl (approved):
Great idea! 👍

oliverlloyd:
[Don't use `pillarPalette`](https://trello.com/c/4eLst0GZ/176-dont-use-pillarpalette)

PR #2911, opened 2021-04-23
Create ItemLinkBlockElement

liywjl:
## What does this change?
Create `ItemLinkBlockElement`

### After
Screenshot 2021-04-23 at 15 18 14

## Why?
Prepare to support numbered list item links

### DIFF
--- a/src/web/components/elements/ItemLinkBlockElement.tsx
+++ b/src/web/components/elements/ItemLinkBlockElement.tsx
@@ -9,12 +9,10 @@ type Props = {
 
 const itemLinkStyles = css`
 	${textSans.medium()}
-	font-weight: 400;
 	a {
 		${textSans.medium({ fontWeight: 'bold' })}
 	}
 	strong {
-		font-weight: 400; /* Need to override globalStrongStyles */
 		display: block;
 	}
 `;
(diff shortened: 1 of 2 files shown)


src/web/components/elements/ItemLinkBlockElement.tsx
```diff
@@ -0,0 +1,29 @@
...
+};
+
+const itemLinkStyles = css`
+	${textSans.medium()}
+	font-weight: 400;
+	a {
+		${textSans.medium({ fontWeight: 'bold' })}
+	}
+	strong {
+		font-weight: 400; /* Need to override globalStrongStyles */
```
oliverlloyd, line 17: I think we can (should?) just use the global strong styles

src/web/components/elements/ItemLinkBlockElement.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { textSans } from '@guardian/src-foundations/typography';
+
+type Props = {
+	html: string;
+};
+
+const itemLinkStyles = css`
+	${textSans.medium()}
+	font-weight: 400;
```
oliverlloyd, line 12: You can use `fontWeight` in `textSans` here

oliverlloyd (approved):

src/web/components/elements/ItemLinkBlockElement.tsx
liywjl: it was unusual that without `font-weight: 400;` the styles from global would take priority (although it is defined higher up in the DOM tree).

PR #2913, opened 2021-04-23
Report missing Braze epic params to Sentry

tjmw:
## What does this change?

If the Epic message is misconfigured in Braze (i.e. missing props) instead of just logging to the browser console, report this to Sentry so we've got better visibility of any issues so we can fix them in Braze.

### Before

Messages were only logged to the console so we wouldn't have been able to tell if there was a misconfiguration in prod.

### After

We'll know from Sentry if an epic has been misconfigured.

### DIFF
--- a/src/web/components/SlotBodyEnd/BrazeEpic.tsx
+++ b/src/web/components/SlotBodyEnd/BrazeEpic.tsx
@@ -14,6 +14,7 @@ import { joinUrl } from '@root/src/lib/joinUrl';
 import { useHasBeenSeen } from '@root/src/web/lib/useHasBeenSeen';
 import { submitComponentEvent } from '@root/src/web/browser/ophan/ophan';
 import { setAutomat } from '@root/src/web/lib/setAutomat';
+import { either } from '@guardian/types';
 
 const wrapperMargins = css`
 	margin: 18px 0;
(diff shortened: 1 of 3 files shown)


src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -81,7 +81,14 @@ const buildEpicProps = (
 	countryCode: string,
 ): EpicProps | null => {
 	const variant = parseBrazeEpicParams(dataFromBraze);
-	if (variant === null) return null;
+	if (variant === null) {
+		const msg = 'Not rendering Braze epic - there are props missing';
```
tjmw, line 85: I'm deliberately not sending `dataFromBraze` to Sentry as the message _could_ be personalised for the user (this isn't the case for our first epic message, but that could change down the line).

adamnfish: The names of the missing props might help us debug any sentry alerts that get generated? It's true that we can no longer distinguish between a missing field and a message with too few paragraphs. If this info was useful before, then it may still be worth preserving?

You're quite right to omit the values though, nice one.

adamnfish: In Scala we'd usually use a Sum Type to model this, Either is a common choice. This means we can return the correct value *or* a datastructure that represents failure for our program.

We're already returning a union type from the parsing function `Variant | null`, which is similar to to the above approach. Perhaps we can replace `null` with `[string]` to carry information about the failed fields, or just `string` if we want to generate a description that can go to sentry from the parser? Union types are not a great substitute for sum types in general, but it should be fine here.

(The trouble with union types is that if the two sides share a compatible type it becomes impossible to know which side you have. An obvious example is that `string | string` doesn't make sense, but you can also have more subtle problems with `any`, `object` or types from within your own program that happen to share a similar structure. Another common name for a "sum type" is a "tagged union". It's this tag that lets us distinguish the sides.)

adamnfish: This is:
* a minor point
* a diversionary ramble

So no worries either way!

tjmw: Thanks for the detailed feedback @adamnfish, these are good points. I'll tweak what I've got to improve the visibility of missing fields.

tjmw: Hi @adamnfish I've updated `parseBrazeEpicParams` this to return either a `Variant` or an error string containing the missing fields. I've used the `Result` type from `@guardian/libs` which feels like a good fit here. Interested in your thoughts on how this looks now.

adamnfish (approved):
Great!

tjmw:
Thanks @adamnfish! 🎉

PR #2915, opened 2021-04-26
Adding support for surveys on DCR

ashishpuliyel:
Surveys were supported by DCP but are not in DCR. The ad-slot is not present on page.
This change adds support for the survey ad-slot on page
### After
Screenshot 2021-04-26 at 10 44 21
### Before
Imagine the image the above, without the survey floating over the page.

## Why?
Since DCR went to 100% the team complained they weren't getting survey responses! This aims to fix that.

### DIFF
--- a/docs/commercial/overview.md
+++ b/docs/commercial/overview.md
@@ -31,7 +31,7 @@ config.get('isDotcomRendering', false)
 
 Dotcom rendering has components for the following ad types
 
--   Static ads (top-above, nav, top-right, most-viewed)
+-   Static ads (top-above, nav, top-right, most-viewed, survey)
 -   Dynamic ads (article in-body)
 -   Mechandising slots (such as masterclass below-content onward component) [in progress...]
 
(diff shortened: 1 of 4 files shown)


liywjl (approved):
LGTM 👍  would be great to have a way to test this locally

ashishpuliyel:
@liywjl I setup an adtest so I could test it. You can see it too if you add &adtest=surveyComDevTest to an article url.
Prod DCR nothing happens:
https://www.theguardian.com/politics/2021/apr/25/labour-calls-for-electoral-commission-inquiry-pm-flat-refurbishment-downning-street?adtest=surveyComDevTest

Prod DCP you see the survey:
https://www.theguardian.com/politics/2021/apr/25/labour-calls-for-electoral-commission-inquiry-pm-flat-refurbishment-downning-street?adtest=surveyComDevTest&dcr=false

Local DCR with this branch you'll see the survey:
http://localhost:3030/Article?url=http://localhost:9000/culture/2021/apr/13/edinburgh-international-festival-plans-outdoor-return-for-2021&adtest=surveyComDevTest

liywjl:
Awesome! Might it be worth adding a Cypress test?

PR #2918, opened 2021-04-26
Use author name in title for Opinion pieces

shtukas:
## What does this change?

As per CP's request, this uses the author's name (byline) instead of the section label on opinion pieces.

SITUATION:

situation

BEFORE:

Before1

AFTER:

after

### DIFF
--- a/src/web/server/document.tsx
+++ b/src/web/server/document.tsx
@@ -42,8 +42,8 @@ interface Props {
 	data: DCRServerDocumentData;
 }
 
-const makeTitle = (CAPI: CAPIType): string => {
-	if (CAPI.pillar === 'opinion' && CAPI.author.byline) {
+const decideTitle = (CAPI: CAPIType): string => {
+	if (format.theme === Pillar.Opinion && CAPI.author.byline) {
 		return `${CAPI.headline} | ${CAPI.author.byline} | The Guardian`;
 	}
 	return `${CAPI.headline} | ${CAPI.sectionLabel} | The Guardian`;
--- a/src/web/server/document.tsx
+++ b/src/web/server/document.tsx
@@ -42,9 +42,16 @@ interface Props {
 	data: DCRServerDocumentData;
 }
 
+const makeTitle = (CAPI: CAPIType): string => {
+	if (CAPI.pillar === 'opinion' && CAPI.author.byline) {
+		return `${CAPI.headline} | ${CAPI.author.byline} | The Guardian`;
+	}
+	return `${CAPI.headline} | ${CAPI.sectionLabel} | The Guardian`;
+};
+
 export const document = ({ data }: Props): string => {
 	const { CAPI, NAV, linkedData } = data;
-	const title = `${CAPI.headline} | ${CAPI.sectionLabel} | The Guardian`;
+	const title = makeTitle(CAPI);
 	const { html, css, ids: cssIDs }: RenderToStringResult = extractCritical(
 		renderToString(
 			// TODO: CacheProvider can be removed when we've moved over to using @emotion/core
--- a/src/web/server/document.tsx
+++ b/src/web/server/document.tsx
@@ -13,8 +13,10 @@ import {
 
 import { makeWindowGuardian } from '@root/src/model/window-guardian';
 import { ChunkExtractor } from '@loadable/server';
+import { Pillar } from '@guardian/types';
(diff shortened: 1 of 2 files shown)


src/web/server/document.tsx
```diff
@@ -42,9 +42,16 @@ interface Props {
 	data: DCRServerDocumentData;
 }
 
+const makeTitle = (CAPI: CAPIType): string => {
+	if (CAPI.pillar === 'opinion' && CAPI.author.byline) {
+		return `${CAPI.headline} | ${CAPI.author.byline} | The Guardian`;
+	}
+	return `${CAPI.headline} | ${CAPI.sectionLabel} | The Guardian`;
+};
```
gtrufitt, line 50: We have a convention of `decideSomething` and rather than direction on CAPI.pillar, work with the Format type: 

```suggestion
const decideTitle = (CAPI: CAPIType): string => {
	if (format.theme === Pillar.Opinion && CAPI.author.byline) {
		return `${CAPI.headline} | ${CAPI.author.byline} | The Guardian`;
	}
	return `${CAPI.headline} | ${CAPI.sectionLabel} | The Guardian`;
};
```

src/web/server/htmlTemplate.ts
```diff
@@ -110,32 +110,32 @@ export const htmlTemplate = ({
...
+            GGGGGGGG            GGGGGGGGGGGGG           GGGGGGGGGGGGG
+             GGGGGGGG            GGGGGGGGGGGG           GGGGGGGGGGGG
+              GGGGGGGGG           GGGGGGGGGGG           GGGGGGGGGGG
+                GGGGGGGGGG         GGGGGGGGGG           GGGGGGGGG
+                  GGGGGGGGGGG        GGGGGGGG        GGGGGGGGGG
+                    GGGGGGGGGGGGGG      GGGGG  GGGGGGGGGGGGGG
+                       GGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGG
+                            GGGGGGGGGGGGGGGGGGGGGGGGG
+                                    GGGGGGGGG
+
```
gtrufitt, line 138: What's going on here?

oliverlloyd (commented):
I saw this at the weekend and took a mini peek at FEs code. Looks like [they're using](https://github.com/guardian/frontend/blob/5b970cd7308175cfc1bcae2d4fb8c06ee13c5fa0/common/app/views/support/Title.scala#L14) `webTitle`, which is available as `CAPI.webTitle`, but there's also some additional logic for other use cases such as pagination or hosted content.

Also. might be better having this code higher up?


(sorry, I'm lurking)

src/web/server/htmlTemplate.ts
```diff
@@ -110,32 +110,32 @@ export const htmlTemplate = ({
...
-            GGGGGGGG            GGGGGGGGGGGGG           GGGGGGGGGGGGG           
-             GGGGGGGG            GGGGGGGGGGGG           GGGGGGGGGGGG            
-              GGGGGGGGG           GGGGGGGGGGG           GGGGGGGGGGG             
-                GGGGGGGGGG         GGGGGGGGGG           GGGGGGGGG               
-                  GGGGGGGGGGG        GGGGGGGG        GGGGGGGGGG                 
-                    GGGGGGGGGGGGGG      GGGGG  GGGGGGGGGGGGGG                   
-                       GGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGG                      
-                            GGGGGGGGGGGGGGGGGGGGGGGGG                           
-                                    GGGGGGGGG                                   
-
```
buck06191, line 138: Is this the linter or did you get annoyed at empty space?

gtrufitt (approved):
Just a small update to work with the Format type rather than directly on pillar (we should probably deprecate passing pillar through)

buck06191 (approved):
LGTM

src/web/server/document.tsx
gtrufitt: I also have a feeling you'll want to do this in the metadata too for social sharing?

src/web/server/htmlTemplate.ts
shtukas: I have no idea. Linting did that 🤷‍♀️

src/web/server/htmlTemplate.ts
shtukas: Linter :)

PR #2925, opened 2021-04-27
Send INSERT event to Ophan for Braze epic

tjmw:
## What does this change?

Sends a component `INSERT` event to Ophan when the Braze epic is rendered. Distinct from the `VIEW` event which we're already sending when the reader actually _sees_ the epic.

## Why?

Recommended by data design for consistency with other components on the page.


tjmw:
Thanks @liywjl!

PR #2927, opened 2021-04-27
Relax Braze epic param validation

tjmw:
~~*Note: this PR depends on another branch so I'm opening it in draft mode. Once the base branch/PR (#2913) has been merged I'll update it to `main` and take out of draft.*~~

## What does this change?

The `heading` and `highlightedText` fields should not be required. The component renders fine without them, so we should not enforce that marketing specify these fields (and in fact the first campaign epic doesn't have a heading).

Example epic without a heading:

Screenshot 2021-04-27 at 14 44 16

### Before

If either of `heading` or `highlightedText` were missing the epic message wouldn't render.

### After

The message renders without `heading` or `highlightedText`.

### DIFF
--- a/src/web/lib/braze/parseBrazeEpicParams.test.ts
+++ b/src/web/lib/braze/parseBrazeEpicParams.test.ts
@@ -56,22 +56,27 @@ describe('parseBrazeEpicParams', () => {
 		expect(got).toEqual(expected);
 	});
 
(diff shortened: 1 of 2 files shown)


arausuy (approved):

tjmw:
Thanks @arausuy! 👍

PR #2929, opened 2021-04-27
change amp version for re-consent

MarSavar:
## What does this change?
This change updates the `<amp-consent>` component to allow re-consent for TCFv2. Re-consent for CCPA will be updated as soon as we have an update from SourcePoint.

### Before
`checkConsentHref: .../amp`
### After
`checkConsentHref: .../amp-v2`

## Why?
The inability to repermission on AMP is causing commercial, operational and potentially compliance risk. SourcePoint have released an update to enable reconsent.


mxdvl (approved):
LGTM ⚡

PR #2935, opened 2021-04-28
Dynamically set spaceAbove of DividerBlockElement for h3

liywjl:
## What does this change?
Set `spaceAbove` on `DividerBlockElement` to  `loose` on dividers that sit between Links and h3

### Before
Screenshot 2021-04-28 at 16 20 29

### After
Screenshot 2021-04-28 at 16 08 45

## Why?
We need to have larger spacing between links and h3 blocks, without applying spacing to all h3 blocks

### DIFF
--- a/src/model/enhance-numbered-lists.ts
+++ b/src/model/enhance-numbered-lists.ts
@@ -177,45 +177,6 @@ const makeThumbnailsRound = (elements: CAPIElement[]): CAPIElement[] => {
 	return inlined;
 };
 
-const addH3s = (elements: CAPIElement[]): CAPIElement[] => {
-	/**
-	 * Why not just add H3s in Composer?
-	 * Truth is, you can't. So to get around this there's a convention that says if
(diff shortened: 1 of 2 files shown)


src/model/enhance-numbered-lists.ts
```diff
@@ -177,6 +177,25 @@ const makeThumbnailsRound = (elements: CAPIElement[]): CAPIElement[] => {
...
+	if (!frag || !frag.firstElementChild) return false;
+
+	const hasULWrapper = frag.firstElementChild.nodeName === 'UL';
+	const hasOnlyOneChild = frag.firstElementChild.childElementCount === 1;
+	const hasLINestedWrapper =
+		frag.firstElementChild?.firstElementChild?.nodeName === 'LI';
+
+	return hasULWrapper && hasOnlyOneChild && hasLINestedWrapper;
+};
+
```
liywjl, line 198: nothing changes here, `isItemLink` just needs to be defined before it is being used (it is used in `addH3s`)

shtukas (approved):

PR #2936, opened 2021-04-28
Increase margin top divider component

liywjl:
## What does this change?
Set the top margin of Divider to be at least 20px

### Before
Screenshot 2021-04-28 at 17 25 15

### After
Screenshot 2021-04-28 at 17 24 54

## Why?
Apply design guidlines


nicl (approved):
Looks good. Just wondering is there a relevant spacing in Source or is this a one-off/custom thing?

PR #2939, opened 2021-04-29
Anniversary Atom: Remove tests switches and switch a/b test over to correct one

gtrufitt:
## What does this change?

- Remove the testing 0% test from the checks whether to show the atom
- Adds a cookie for 0% a/b test D with a 2 day TTL (so that it drops on Wednesday, and expires by Saturday)


shtukas (approved):

shtukas (approved):

liywjl (approved):
LGTM

PR #2941, opened 2021-04-29
Add editionId check for AU to anniversary atom

gtrufitt:
## What does this change?

Adds a check to ensure we don't show the article atom on the AU edition.


liywjl (approved):
LGMT

PR #2942, opened 2021-04-29
Anniversary Atom: Inline the atom instead of iframing

gtrufitt:
## What does this change?

This breaks the atom out of the iframe so that:

- We match what is happening with the atom on Fronts (they're not iframed)
- We can target the atom in the CSS with the data attribute

There should be no visual difference between the two as there are no global styles targeting anything outside of the article container.


liywjl (approved):
👍

PR #2943, opened 2021-04-29
Anniversary atom: Sensitivity checks

gtrufitt:
## What does this change?
Checks to make sure that the page is not sensitive, and that adverts haven't been blocked before showing the anniversary atom.


shtukas (approved):

liywjl (approved):
👍

PR #2944, opened 2021-04-29
Right spacing in discussion regardless of advert

nitro-marky:
## What does this change?

In some cases discussion would take up the whole width. This adds right hand padding.

### Before

image

### DIFF
--- a/src/web/components/Discussion.tsx
+++ b/src/web/components/Discussion.tsx
@@ -173,13 +173,13 @@ export const Discussion = ({
 			>
 				<Flex>
 					<div>
-						className={css`
+						className=
+						{css`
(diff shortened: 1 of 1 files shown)


src/web/components/Discussion.tsx
```diff
@@ -172,14 +172,14 @@ export const Discussion = ({
 				}
 			>
 				<Flex>
-					<div
-						className={css`
+					<div>
+						className=
+						{css`
```
liywjl, line 177: lint issue

liywjl (changes requested):
Might need to run prettier and the linter again, also would be great to get an image of what it would look like after the change (you can fake login state to get it to render)

PR #2953, opened 2021-04-30
Remove global image styles

gtrufitt:
## What does this change?

Removes the global image styles:

- Global styles are not ideal and we should aim to always style within components directly
- These global styles were interfering with Interactive Atoms that already existed.
- There does not seem to be anywhere in the platform that required the global styles as far as chromatic is concerned


liywjl (approved):
👍

rcrphillips:
[Remove global image styles](https://trello.com/c/pyH2W7Da/2876-remove-global-image-styles)

PR #2957, opened 2021-05-04
Identity | Sign In Gate | Remove sign in gate from isPaidContent

coldlink:
It was brought to our attention by @jfsoul that the Sign In Gate would appear on paid/sponsored articles. We came to the conclusion that it should not appear on these pages. On frontend there is already a check to see if an article is paid/sponsored, but this was not ported over to DCR.

This PR adds a check for this by looking for the `isPaidContent` value in the the CAPI response, and uses this to determine if the gate should show or not.


liywjl (approved):
LGTM 👍

PR #2958, opened 2021-05-04
Set braze epic timeout to 5 seconds.

michaelclapham:
### Before
If there was a slot body end candidate after the Braze Epic, it would never show if Braze didn't respond.

### After
After a 5 seconds of trying to request a message to show from Braze, slot body end will give up and show next Epic candidate.

## Why?
This is necessary requirement before switching the priority to show the Braze epic before the RR epic.

### DIFF
--- a/src/web/components/SlotBodyEnd/SlotBodyEnd.tsx
+++ b/src/web/components/SlotBodyEnd/SlotBodyEnd.tsx
@@ -83,7 +83,7 @@ const buildBrazeEpicConfig = (
 				/>
 			),
 		},
-		timeoutMillis: null,
+		timeoutMillis: 5000,
 	};
 };
 


adamnfish (approved):

tjmw (approved):
👍

gtrufitt:
Hey @michaelclapham , quick question on this one, where does 5 seconds come from? It's quite a long time to hold up overlaying content and potentially pushes out LCP for users - could we assess what the shortest reasonable timeout is and try and reduce it?

This is the timings for LCP, ideally we want to reduce the timeout to less than 2.5 seconds so we can keep the LCP within the Good range.

image

cc/ @jfsoul

tjmw:
Hey @gtrufitt that's a great question. For context previously the picking logic around the epic didn't enforce any time limit so in theory could have taken much longer. We discussed as a team and realised it could be problematic, so wanted to start enforcing some kind of limit. So 5 seconds was chosen slightly arbitrarily as a starting point, with the idea that it was at least better than no limit at all. Being guided by the timings you shared seems like a good call. (This is all theoretical as we haven't had a Braze powered epic enabled yet, but are aiming to very soon).

Other context: the banner picking logic follows frontend's lead and enforces a 2 second timeout across the board (I think for all banners except the CMP). Should that be mirrored here? Our thinking was that from a user's perspective a delay in the epic was more acceptable as it's below the article, as opposed to the banner rendering over the top.

gtrufitt:
Thanks @tjmw! 

I missed that this was for the epic rather than the banner (even though it's in the title), so I am less concerned. I guess maybe your best bet is to record timeouts and work back to a reasonable number from there? I understand there'll be a balance and cost with faster to fallback vs higher value but later banner.

I'll leave it to you, thanks!

jfsoul:
Definitely worth doing some lighthouse runs from your dev tools to double check that the epic you're serving doesn't end up being the largest element on the page (especially on mobile). Anecdotally we've recently seen the contributions banner be driving LCP, but not the subscriptions banner - purely because of the length of the message vs. the main image of the article. Basically, if we can try to keep main-media as largest element then we won't go too wrong with regards to LCP.

tjmw:
Thanks @jfsoul @gtrufitt good suggestions. Really interesting points about LCP! For the first Braze epic message we're using the exact same component (and similar length content) to the RRCP epic messages so I wouldn't expect much to change there.

> Basically, if we can try to keep main-media as largest element then we won't go too wrong with regards to LCP.

This makes total sense. Is it worth capturing guidelines like this somewhere? Especially if we're expanding usage of these slots and the ability of teams to create new messages independently.

gtrufitt:
>  Is it worth capturing guidelines like this somewhere? 

Yes! We've planned to do this for a while but before we put lines in the sand in place for what is allowed / not allowed, we want to work directly with teams to help make them, as well as use the data we're now collecting for core vitals to inform us

PR #2962, opened 2021-05-04
Numbered list h2 bold

liywjl:
## What does this change?
All h2 in numbered list have bold

### Before
normal

### After
bold

## Why?
headers should be bold


liywjl:
We do not want to make the entire title bold

PR #2963, opened 2021-05-05
Hardcode ImmersiveComment as ImmersiveArticle

buck06191:
## What does this change?
We can't currently handle `ImmersiveDisplay` and `CommentDesign`
together so this override `CommentDesign` to be `ArticleDesign` when the
display type is `ImmersiveDisplay`.

### Before
before

### After
after

## Why?
Temporary fix for a broken format combination


liywjl (approved):
LGTM 👍

PR #2965, opened 2021-05-05
Fix FauxH3 title link

liywjl:
## What does this change?
Use `innerHTML` instead of `textContent` as not to loose other HTML tags in text (notably links)

### Before
title link DCR

### After
fix DCR link title

## Why?
We should not loose nested HTML

### DIFF
--- a/src/model/enhance-numbered-lists.ts
+++ b/src/model/enhance-numbered-lists.ts
@@ -37,7 +37,13 @@ const extractH3 = (element: CAPIElement): string => {
 	const textElement = element as TextBlockElement;
 	const frag = JSDOM.fragment(textElement.html);
 	if (isFalseH3(element)) {
-		return frag.firstElementChild?.firstElementChild?.innerHTML || '';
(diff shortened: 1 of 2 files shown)


shtukas (approved):

src/model/enhance-numbered-lists.ts
```diff
@@ -37,7 +37,13 @@ const extractH3 = (element: CAPIElement): string => {
 	const textElement = element as TextBlockElement;
 	const frag = JSDOM.fragment(textElement.html);
 	if (isFalseH3(element)) {
-		return frag.firstElementChild?.textContent || '';
+		return (
+			frag.firstElementChild?.innerHTML
+				.split('<strong>')
+				.join('')
+				.split('</strong>')
+				.join('') || ''
```
liywjl, line 45: not ideal, but its the simplest way of removing all strong tags from the text while still keeping other HTML tags and text in between strong tags

PR #2970, opened 2021-05-07
Blockquote styling improvements

liywjl:
## What does this change?
- Simplify blockquote logic
- try to align blockquotes `small` to text height
- Use flexbox and `align-items: centre;` to align SVG to text
- Reduce margin-bottom
- Add margin-top

### Before
blockquotes before

### After
blockquotes afterPNG

## Why?
Blockquotes SVG did not have the correct viewBox and it had clashing CSS and nested Height/Width. This made adjusting the blockquote painful. We want to align the SVG to the text which can be done a lot simpler if we string down the CSS logic to only use Height and use flexbox on the parent wrapper.


gtrufitt (changes requested):
Looks like this might need adjusting on cards

image

oliverlloyd:
Is this PR related? https://github.com/guardian/source/pull/804 - ideally, we'd use Source more here

src/web/components/elements/BlockquoteBlockComponent.stories.tsx
```diff
@@ -30,6 +30,11 @@ export const Unquoted = () => {
 					display: Display.Standard,
 					theme: Pillar.News,
 				})}
+				format={{
+					design: Design.Article,
+					display: Display.Standard,
+					theme: Pillar.News,
+				}}
```
OllysCoding, line 37: Would be nice if there was a way we could do this without having to duplicate the `palette` prop - but it's not really a huge deal since it's just for stories

OllysCoding (approved):
Looks good!

nitro-marky (approved):
👍

index.d.ts
```diff
@@ -522,7 +522,7 @@ interface BadgeType {
 
 type ImagePositionType = 'left' | 'top' | 'right';
 
-type SmallHeadlineSize = 'tiny' | 'small' | 'medium' | 'large';
+type HeadlineSizeType = 'xsmall' | 'small' | 'medium' | 'large';
```
liywjl, line 525: we should use `HeadlineSizeType` from Source once released

src/web/components/CardHeadline.tsx
```diff
@@ -18,13 +18,13 @@ type Props = {
 	showPulsingDot?: boolean;
 	showSlash?: boolean;
 	showQuotes?: boolean; // Even with design !== Comment, a piece can be opinion
-	size?: SmallHeadlineSize;
+	size?: HeadlineSizeType;
```
liywjl, line 21: should use `HeadlineSizeType` from Source once released

src/web/components/QuoteIcon.tsx
```diff
@@ -1,84 +1,21 @@
...
-const quoteStyles = (colour?: string) => css`
-	height: 17px;
-	width: 9px;
-	margin-right: 12px;
-	transform: translateY(-1px);
-	overflow: visible;
-	fill: ${colour && colour};
-`;
-
-const sizeStyles = (size: SmallHeadlineSize) => {
```
liywjl, line 14: this logic has moved to `@guardian/src-ed-quote-icon`

src/web/components/QuoteIcon.tsx
```diff
@@ -1,84 +1,21 @@
...
-		case 'large':
-			return css`
-				${largeSvg}
-				${until.desktop} {
-					${mediumSvg}
-				}
-			`;
+const quoteBaseStyles = (colour?: string) => css`
+	span > svg {
+		fill: ${colour && colour};
```
liywjl, line 7: this is difficult, in `@guardian/src-ed-quote-icon` we have logic to set QuoteIcon; however, we have a much more complex set of logic for how we choose colours for quote icons (look at `decidePalette.ts`)

Should we extend `QuoteIcon` to allow us to plug in colour? Or should we try to port over our logic to Source? 
Not sure what is best, but this is a current work around to the issue

src/web/components/QuoteIcon.tsx
```diff
@@ -1,84 +1,21 @@
...
-		<svg
-			width="70"
-			height="49"
-			viewBox="0 0 35 25"
-			css={quoteStyles(colour)}
-			data-testid="quote-icon"
-		>
-			<path d="M69.587.9c-1.842 15.556-3.89 31.316-4.708 48.1H37.043c3.07-16.784 8.391-32.544 17.602-48.1h14.942zM32.949.9c-2.047 15.556-4.094 31.316-4.912 48.1H.2C3.066 32.216 8.592 16.456 17.598.9h15.35z" />
-		</svg>
+export const QuoteIcon = ({ format, colour, size }: Props) => (
```
liywjl, line 17: I wanted to remove this component and instead use `QuoteIcon` from Source. However, there were some differences between colour selection logic, making it easier to override

src/web/components/elements/BlockquoteBlockComponent.stories.tsx
liywjl: the difference here is subtle, you can see `palette` uses `decidePalette`. But granted this looks like a lot of unnecessary repetition, although for stories I think that's ok

src/web/components/elements/BlockquoteBlockComponent.tsx
```diff
@@ -1,52 +1,43 @@
...
-	<blockquote
-		css={css`
-			display: flex;
-			flex-direction: row;
-		`}
-	>
-		{children}
-	</blockquote>
-);
-
```
liywjl, line 24: no need for this abstraction, l plus it allows us to just use `css` from `ClassNames` without aliasing the name

src/model/json-schema.json
```diff
@@ -3721,6 +3707,9 @@
                 "isLive": {
                     "type": "boolean"
                 },
+                "isPreview": {
+                    "type": "boolean"
+                },
```
gtrufitt, line 3712: This seems strange that it appears in the schema but is not changed on the TS model?

src/web/components/QuoteIcon.tsx
gtrufitt: Would this not fall back to `fill: false` in the css if there was no colour?

src/web/components/elements/BlockquoteBlockComponent.stories.tsx
gtrufitt: Probably in another PR, but it's worth removing `palette` as a prop and just using `format` to `decidePalette` at the usage point rather than drilling it all the way down.

gtrufitt (approved):
Couple of comments but the code looks good, has @HarryFischer confirmed that the quote icon size change is expected? AFAIK the original design calls for the quote icon to align to the line-height of the text, which cards no longer do

image

I know Harry is away so maybe @paperboyo @akemitakagi have thoughts?

src/model/json-schema.json
liywjl: good point! will revert

HarryFischer:
Can we increase the quote size so it matches the text cap height please - also from my phone it looks like the quote marks sits slightly below the text baseline so if we can nudge it up too @liywjl

paperboyo:
> thoughts?

Yep, wot Harry said: should sit on the baseline and reach cap height. [This screenshot](https://github.com/guardian/dotcom-rendering/pull/2970#pullrequestreview-655540443) looks off.

liywjl:
@HarryFischer @gtrufitt  I will need to as a result overload the icon here in DCR, I was wondering if maybe this is something that should be done in Source to avoid such overloadings?

gtrufitt:
> @HarryFischer @gtrufitt I will need to as a result overload the icon here in DCR, I was wondering if maybe this is something that should be done in Source to avoid such overloadings?

@SiAdcock @sndrs what's the expected usage of SVG icons like this, would you expect the platform to adjust the size of an icon if required?

mxdvl:
One option would be to have a `tight` setting for icons, where the `viewBox` is within 1 unit of the content’s bounding box. Or allowing to override the `viewBox` with a parameter.

I know I had a chat previously with @akemitakagi and @SiAdcock and they were working on a new version of the icons with much less empty space around.

mxdvl:
https://trello.com/c/Oey0mMZ8/198-add-space-around-quote we’ll pick this up later

PR #2971, opened 2021-05-07
Add custom className to figure component

liywjl:
## What does this change?
- Add custom className to Figure component
- Add string to Figure in image block component

### Before
Ad in between starts and image

## Why?
We want to prevent `spacefinder` from inserting an advert in between stars and images

## Related PR
https://github.com/guardian/frontend/pull/23788


liywjl:
closing in favour of https://github.com/guardian/dotcom-rendering/pull/2975

## Why?
We should always look at reducing complexity and making sure that props are as small as possible in scope.

PR #2973, opened 2021-05-07
Introduce a basic interctive endpoint

nicl:
## What does this change?

Adds a new `/Interactive` endpoint that, for now, does the same thing as the vanilla article endpoint. Note that, it requires https://github.com/guardian/frontend/pull/23786 to go live before this will be useful.

## Why?

Part of supporting interactive pages on DCR!

### DIFF
--- a/src/app/server.ts
+++ b/src/app/server.ts
@@ -12,9 +12,10 @@ import {
 	renderPerfTest as renderAMPArticlePerfTest,
 } from '@root/src/amp/server/render';
 import {
-	render as renderArticle,
+	renderArticle,
 	renderPerfTest as renderArticlePerfTest,
 	renderArticleJson,
+	renderInteractive,
 } from '@root/src/web/server/render';
 
 import {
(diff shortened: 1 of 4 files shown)


shtukas (approved):

src/app/server.ts
```diff
@@ -26,12 +27,16 @@ import { logger } from './logging';
...
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+	switch (options.path) {
+		case '/Article':
+			return renderArticle;
+		case '/ArticleJson':
+			return renderArticleJson;
+		case '/AMPArticle':
+			return renderAMPArticle;
+		case '/Interactive':
+			return renderInteractive;
```
gtrufitt, line 39: This will save me about 1 hour whenever I go to add a route 😆

nicl: Yeah, it's still a bit messy as same info duplicated but better than before at least :)

PR #2975, opened 2021-05-10
Add isImageBlockElement to figure component

liywjl:
## What does this change?
- Add prop `isImageBlockElement` to figure component
- pass `isImageBlockElement` true if figure is rendering image block
- if figure is rendering element block, add className 'element--image-block'

### Before
Ad in between starts and image

## Why?
We want to prevent `spacefinder` from inserting an advert in between stars and images

## Related PR
https://github.com/guardian/frontend/pull/23788

### DIFF
--- a/src/web/components/Figure.tsx
+++ b/src/web/components/Figure.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { css } from 'emotion';
+import { cx, css } from 'emotion';
 
 import { from, until } from '@guardian/src-foundations/mq';
 import { space } from '@guardian/src-foundations';
@@ -9,6 +9,7 @@ type Props = {
 	isMainMedia: boolean;
 	role?: RoleType | 'richLink';
 	id?: string;
+	isImageBlockElement?: boolean;
 };
 
 const roleCss = {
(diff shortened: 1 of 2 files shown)


oliverlloyd:
This coupling with Frontend is a little concerning. If we wanted to work on space finder later this might make that task harder?

One question, wouldn't we also have had the same problem with image title? That [gets set the same way](https://github.com/guardian/dotcom-rendering/blob/240af6a84306376f36cfecdc71e610c47b609202/src/web/components/elements/ImageComponent.tsx#L269) as the stars but I've not seen or heard of space finder affecting this. Either this solution also needs to extend to title or is it possible the problem only happens if the stars are standalone and not set using the `starRating` prop on `ImageComponent`? Maybe the cleaner isn't working as expected? (There are two ways to insert stars in numbered lists, on an image directly or as their own stand alone element.)

gtrufitt:
Similar to Oliver, this doesn't feel like it should be required, spacefinder doesn't ignore images on their own, and the only difference as far as I understand here is that the stars overlay an image? In which case, shouldn't spacefinder be able to do its job and the problem is with how the figure is containing the stars and image? i.e: there's a layout issue with the component, not a spacefinder or ad issue?

PR #2977, opened 2021-05-10
Fix negative margin on divider

liywjl:
## What does this change?
We extend the width of the divider to compensate for the negative margin

#### Desktop
desktop-before-fix

#### Tablet
tablet-before-fix

#### Mobile
mobile-before-fix

#### Desktop
desktop-after-fix

#### Tablet
tablet-after-fix

#### Mobile
mobile-after-fix

## Why?
The divider does not match the full width of the content (thus looking unaligned)


HarryFischer (approved):
🤩

PR #2983, opened 2021-05-11
Remove enhance anniversary atom infra

gtrufitt:
## What does this change?

Removes infrastructure around anniversary atom as it is no longer going to be used.


HarryFischer (approved):

gtrufitt:
We've had confirmation that this is no longer required. Will merge.

PR #2985, opened 2021-05-12
extract rounding function

mxdvl:
## What does this change?

Extract the rounding function. May I ask why this was needed?

## Why?

We're looking at implementing a similar thing for commercial metrics.

### DIFF
--- a/src/web/browser/coreVitals/coreVitals.ts
+++ b/src/web/browser/coreVitals/coreVitals.ts
@@ -33,24 +33,28 @@ export const coreVitals = (): void => {
 		name: string;
 		value: number;
 	};
+	
+	const nineDigitPrecision = (number: number) => {
+		// The math functions are to make sure the length of number is <= 9
+		return Math.round(value * 1_000_000) / 1_000_000;
+	}
 
 	const addToJson = ({ name, value }: CoreVitalsArgs): void => {
 		switch (name) {
-			// The math functions are to make sure the length of number is <= 9
 			case 'FCP':
-				jsonData.fcp = Math.round(value * 1000000) / 1000000;
+				jsonData.fcp = nineDigitPrecision(value);
 				break;
 			case 'CLS':
-				jsonData.cls = Math.round(value * 1000000) / 1000000;
+				jsonData.cls = nineDigitPrecision(value);
 				break;
 			case 'LCP':
-				jsonData.lcp = Math.round(value * 1000000) / 1000000;
+				jsonData.lcp = nineDigitPrecision(value);
 				break;
 			case 'FID':
-				jsonData.fid = Math.round(value * 1000000) / 1000000;
+				jsonData.fid = nineDigitPrecision(value);
 				break;
 			case 'TTFB':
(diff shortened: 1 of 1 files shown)


src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -33,24 +33,28 @@ export const coreVitals = (): void => {
 		name: string;
 		value: number;
 	};
+	
+	const nineDigitPrecision = (number: number) => {
+		// The math functions are to make sure the length of number is <= 9
```
mxdvl, line 38: Does this length of `9` includes the decimal point?

gtrufitt:
cc @nitro-marky on the question

src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -33,24 +33,28 @@ export const coreVitals = (): void => {
 		name: string;
 		value: number;
 	};
+	
+	const nineDigitPrecision = (number: number) => {
```
mxdvl, line 37:
```suggestion
	const nineDigitPrecision = (value: number) => {
```

src/web/browser/coreVitals/coreVitals.ts
mxdvl: And why 9 specifically? Is this a requirement from Data Tech?

nitro-marky: We use 9 so it will fit in the body we send to Fastly. We tried different combinations (high numbers would fail) and found this worked and gave us more than enough overhead on the data granularity.

mxdvl: Cool, when we look at this, we're likely to round to the nearest millisecond. I wonder if the sub-millisecond precision is brining a lot of value in your case?

https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp

Also curious what happens if you have values beyond 999ms, because then you'd get 10 digits. Maybe the method could be renamed to `roundBeyondSixDecimal`.

liywjl (approved):
LGTM 👍

src/web/browser/coreVitals/coreVitals.ts
nitro-marky: With core vitals it would be reported as 999, they send different formats for each metric.

src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -33,24 +33,28 @@ export const coreVitals = (): void => {
 		name: string;
 		value: number;
 	};
+	
+	const nineDigitPrecision = (value: number) => {
+		// The math functions are to make sure the length of number is <= 9
+		return Math.round(value * 1_000_000) / 1_000_000;
+	}
```
liywjl, line 40: I was getting some prettier errors that were preventing me from pushing. I found out it was because you are missing `;` here.

It might be worth checking prettier is working on your IDE (I have it run on file save).

mxdvl: (full disclosure: no IDE was used for this PR)

PR #2991, opened 2021-05-13
Add immersive interactive layout

nicl:
## What does this change?

Adds a basic implementation of immersive interactives. Note, this is aiming for a limited kind of visual parity, but nothing else atm. There are bugs and missing features.

Note, I'm not especially happy with the approach to customising/removing role-specific styling for the elements here. Very open to better suggestions.

### Before

Nowt.

### After

E.g. for https://www.theguardian.com/australia-news/ng-interactive/2021/may/11/the-complete-2021-australian-federal-budget-choose-what-matters-to-you.

(DCR endpoint is: http://localhost:3030/Interactive?url=http://localhost:9000/australia-news/ng-interactive/2021/may/11/the-complete-2021-australian-federal-budget-choose-what-matters-to-you.)

Screenshot 2021-05-13 at 14 08 20

## Why?

Another step towards interactives on DCR.

### DIFF
--- a/src/web/components/Figure.tsx
+++ b/src/web/components/Figure.tsx
@@ -4,13 +4,6 @@ import { css } from 'emotion';
 import { from, until } from '@guardian/src-foundations/mq';
 import { space } from '@guardian/src-foundations';
 
-type Props = {
-	children: React.ReactNode;
-	isMainMedia: boolean;
-	role?: RoleType | 'richLink';
-	id?: string;
-};
-
 const roleCss = {
 	inline: css`
 		margin-top: ${space[3]}px;
@@ -129,7 +122,8 @@ const roleCss = {
 	`,
 };
 
-const decidePosition = (role: RoleType | 'richLink') => {
+// Used for vast majority of layouts.
+export const defaultRoleStyles = (role: RoleType | 'richLink') => {
 	switch (role) {
 		case 'inline':
 			return roleCss.inline;
@@ -150,11 +144,23 @@ const decidePosition = (role: RoleType | 'richLink') => {
 	}
 };
 
+type Props = {
+	children: React.ReactNode;
+	isMainMedia: boolean;
+	role?: RoleType | 'richLink';
+	id?: string;
+
+	// Used to style figures based on role type. Parameterised as this varies
+	// (e.g.) on page layout.
+	roleStylesFn?: (role: RoleType | 'richLink') => string;
+};
+
 export const Figure = ({
 	role = 'inline',
 	children,
 	id,
 	isMainMedia,
+	roleStylesFn = defaultRoleStyles,
 }: Props) => {
 	if (isMainMedia) {
 		// Don't add in-body styles for main media elements
@@ -165,7 +171,7 @@ export const Figure = ({
 		return <figure id={id}>{children}</figure>;
 	}
 	return (
-		<figure id={id} className={decidePosition(role)}>
+		<figure id={id} className={roleStylesFn(role)}>
 			{children}
 		</figure>
 	);
--- a/src/web/layouts/DecideLayout.tsx
+++ b/src/web/layouts/DecideLayout.tsx
@@ -15,6 +15,7 @@ import { ShowcaseLayout } from './ShowcaseLayout';
 import { CommentLayout } from './CommentLayout';
 import { ImmersiveLayout } from './ImmersiveLayout';
 import { LiveLayout } from './LiveLayout';
+import { InteractiveImmersiveLayout } from './InteractiveImmersiveLayout';
 
 type Props = {
 	CAPI: CAPIType;
@@ -41,30 +42,28 @@ export const DecideLayout = ({ CAPI, NAV }: Props): JSX.Element => {
 	};
 	const palette = decidePalette(format);
 
+	// TODO we can probably better express this as data
 	switch (display) {
 		case Display.Immersive: {
-			switch (design) {
-				case Design.Comment:
-				case Design.Editorial:
-				case Design.Letter:
-					return (
-						<ImmersiveLayout
-							CAPI={CAPI}
-							NAV={NAV}
-							palette={palette}
-							format={format}
-						/>
-					);
-				default:
-					return (
-						<ImmersiveLayout
-							CAPI={CAPI}
-							NAV={NAV}
-							palette={palette}
-							format={format}
-						/>
-					);
+			if (design === Design.Interactive) {
+				return (
+					<InteractiveImmersiveLayout
+						CAPI={CAPI}
+						NAV={NAV}
+						format={format}
+						palette={palette}
+					/>
+				);
 			}
+
+			return (
+				<ImmersiveLayout
+					CAPI={CAPI}
+					NAV={NAV}
+					palette={palette}
+					format={format}
+				/>
+			);
 		}
 		case Display.NumberedList:
 		case Display.Showcase: {
--- a/src/web/layouts/InteractiveImmersiveLayout.tsx
+++ b/src/web/layouts/InteractiveImmersiveLayout.tsx
@@ -1,0 +1,329 @@
+import React from 'react';
+import { css, cx } from 'emotion';
+
+import {
+	brandBackground,
+	brandBorder,
+	labs,
+	border,
+} from '@guardian/src-foundations/palette';
+import { from } from '@guardian/src-foundations/mq';
+import { Design, Special } from '@guardian/types';
+import type { Format } from '@guardian/types';
+
+import { GuardianLines } from '@root/src/web/components/GuardianLines';
+import { MainMedia } from '@root/src/web/components/MainMedia';
+import { ArticleTitle } from '@root/src/web/components/ArticleTitle';
+import { ArticleHeadline } from '@root/src/web/components/ArticleHeadline';
+import { Footer } from '@root/src/web/components/Footer';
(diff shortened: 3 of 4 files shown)


shtukas (approved):

src/web/components/Figure.tsx
```diff
@@ -150,11 +144,23 @@ const decidePosition = (role: RoleType | 'richLink') => {
 	}
 };
 
+type Props = {
+	children: React.ReactNode;
+	isMainMedia: boolean;
+	role?: RoleType | 'richLink';
```
liywjl, line 150: maybe worth adding 'richLink' to `RoleType`

src/web/layouts/DecideLayout.tsx
```diff
@@ -41,30 +43,28 @@ export const DecideLayout = ({ CAPI, NAV }: Props): JSX.Element => {
...
-						/>
-					);
-				default:
-					return (
-						<ImmersiveLayout
-							CAPI={CAPI}
-							NAV={NAV}
-							palette={palette}
-							format={format}
-						/>
```
liywjl, line 65: great catch didn't notice they both returned the same 👍

nicl: Ha yes was a bit confused by this!

src/web/components/Figure.tsx
```diff
@@ -150,11 +144,23 @@ const decidePosition = (role: RoleType | 'richLink') => {
...
+	// (e.g.) on page layout.
+	roleStyler?: (role: RoleType | 'richLink') => string;
+};
+
 export const Figure = ({
 	role = 'inline',
 	children,
 	id,
 	isMainMedia,
+	roleStyler = standardRoleStyler,
```
liywjl, line 163: I am a little worried by this pattern. Once you allow anyone to overload styles, you can quickly start adding a lot of unnecessary complexity. I was trying to find an example of this overload being using in the PR but cannot find one that doesn't use the default `standardRoleStyler`. I would suggest maybe remove `roleStyler` props and then progressively adding styles as we find examples.

src/web/components/Figure.tsx
nicl: Yeah it's a weird one. `richLink` isn't a genuine CAPI role, but seems to be used here as if it is one (albeit with a quite specific name!). I didn't attempt to change this approach in the PR but agree it is potentially confusing.

src/web/components/Figure.tsx
```diff
@@ -129,7 +122,8 @@ const roleCss = {
 	`,
 };
 
-const decidePosition = (role: RoleType | 'richLink') => {
+// Used for vast majority of layouts.
+export const standardRoleStyler = (role: RoleType | 'richLink') => {
```
liywjl, line 126: granted `decidePosition` wasn't a great name for the prop (it does a lot more than just decide position) but I am not sure about `standardRoleStyler`. How about something like `roleBasedStyles` ?

src/web/components/Figure.tsx
liywjl: just noticed `'richLink'` was here before ^^ still prob worth adding it maybe? (though could be spun out into a new PR)

src/web/components/Figure.tsx
nicl: It is used for the immersive layouts (which is why I introduced it). It is unfortunate, I agree, as it adds messiness/complexity and will probably only get used in this one place.

In general, I think the design mistake here is to couple the layout aspect of elements (the CSS attached to wrapping `Figure`s with the conversion of elements to their React representations. Unfortunately, it seems hard to separate these out easily - partly because the element => JSX.Element step can result in a null (in which case we wouldn't want to wrap in a figure) but it's impossible to identify these cases. Or rather, we could do by adding a load of code.

src/web/components/Figure.tsx
nicl: 👍 perhaps `defaultRoleBasedStyles`?

src/web/components/Figure.tsx
nicl: I'd like to try and improve this though in a separate PR once we know if we need this code/interactives are a bit clearer.

liywjl: I am not sure I follow 100%, but you mentioned that in some cases the JSX returned is `null` (I am assuming you are referring to `children` prop in this case), and that in those cases you do not want to return the `figure` HTML. If that is the case it would make more sense to conditionally determine that in the parent component and not here?

liywjl: if we were to have a new `const` to determine styles that are specific to Immersives, would it not make more sent to define that variable here and just conditionally call it if it is interactive. We could call it `interactiveStyles` or something on those lines. This would remove the need for some dynamic prop to set styles

src/web/layouts/InteractiveImmersiveLayout.tsx
```diff
@@ -0,0 +1,329 @@
+import React from 'react';
```
liywjl, line 1: I would make the case that it might be worth abstracting out differences rather than copying `ImmersiveLayout.tsx`, can we not use `ImmersiveLayout` in the meantime?

liywjl: it just means that we will likely have to double the work if ever we need to make a change, and these files are big to maintain. I would prefer to include Immersive logic in this file

src/web/layouts/InteractiveLayout.tsx
```diff
@@ -0,0 +1,734 @@
+import React from 'react';
```
liywjl, line 1: Ditto https://github.com/guardian/dotcom-rendering/pull/2991/files#r631828742

src/web/components/Figure.tsx
nicl: Yes that is the alternative, but how to conditionally determine that for each element without essentially reproducing the element logic. Which would add a lot of code and most of it duplicated for the element rendering itself.

I.e. at the moment ElementsRenderer returns: `JSX.Element`, where this can be a null element but impossible to tell. If not null, typically wrapped in `<Figure...>` with some layout/other CSS.
```

This PR parameterises the `Figure` CSS so that it can be customised for different layouts etc. (Required for elements in immersive interactives).

As alternatives to separate out the CSS, we could do:

1. don't wrap in `<Figure ...>` and let the calling code handle that

This would require us to possibly change ElementRenderer to accept all the elements and return an array - that way it could strip out nulls and we could guarantee returned elements required the `<Figure ..` wrapping. However, this is also brittle as it would be difficult to wrap with a different Figure/props based on element type - which we currently do. We could get around this with multiple return etc - e.g. return the element ID along with the element. This would be a bit more code but definitely possible and potentially the 'right' solution here.

2. Duplicate the ElementRenderer code purely for the special case (Immersive interactives) and then modify as required.

This would duplicate a lot of core code though, which is not great.

--

I'm sure there are other options too.

liywjl: could we not filter out "empty" elements using a similar pattern that we do for Numbered lists in the `enhancer`?

src/web/layouts/InteractiveImmersiveLayout.tsx
nicl: Definitely, I don't think there is enough overlap with `ImmersiveLayout` relative to say other layouts (other than the name and nav) but I think we could (+ should) refactor all the *Layout files to pull out common things like nav handling and footer and simplify/re-use code a lot more.

(I think that should be a separate PR though as it will be large.)

src/web/components/Figure.tsx
liywjl: in the end, I would strongly vouch against this approach, if you really wanted to overload certain styles because the article is `Immersive` I think it would be better to pass down a boolean prop `isImmersive` and then have a css `const` that would conditionally apple relevant styles depending on the `role` and `isImmersive`. This way we would no longer need to pass down `roleStylerroleStyler` function

src/web/layouts/InteractiveLayout.tsx
nicl: Ah thanks - actually I should remove this from this PR! Originally, I planned to implement both new layouts here, but then decided against to keep the PR smaller.

nicl:
Okay, to make progress I will merge. But will raise some PRs around *Layout and also ElementRenderer to follow.

src/web/components/Figure.tsx
oliverlloyd: For what it's worth, I'm not a huge fan of this approach either. Passing in functions that return styles goes against the principle of components keeping their styles local and isolated. This was a key driver for the DCR project and I think it's something that is worth continuing to strive for.

Why is it worth it?
The benefit of a component having style logic isolated is
1. It's much easier to reason about what to expect. Everything is local to the file so there's no  need to jump around
2. Any changes only impact the thing you're looking at, there's no style bleed
3. The props can be cleanly and tightly typed. The type of this function is `any` basically

Stepping back, I wonder if the need for this type of change is suggestive that there needs to be a new component? Sometimes the best approach is to create a second comoponent instead of trying to overload the first. I think it's perfectly valid to have `Figure` and `InteractiveFigure` (or whatever name makes sense based on the differences). What do you think?

liywjl: There was a thread about this exact topic https://chat.google.com/room/AAAAWwBdSMs/QJ-dfrnymR4

PR #2996, opened 2021-05-14
Add email prop to RR banner

TomPretty:
## What does this change?
Add email prop to RR banner. This is for the new reminder feature we're adding to the contributions banner (https://github.com/guardian/support-dotcom-components/pull/447)

### DIFF
--- a/src/web/layouts/lib/stickiness.tsx
+++ b/src/web/layouts/lib/stickiness.tsx
@@ -32,7 +32,7 @@ const bannerWrapper = css`
 	position: fixed !important;
 	bottom: 0;
 	${getZIndexImportant('banner')}
-	max-height: 80vh;
+	max-height: 100vh;
 	overflow: visible;
 	width: 100% !important;
 	background: none !important;
--- a/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
+++ b/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
@@ -12,6 +12,7 @@ import {
 	withinLocalNoBannerCachePeriod,
 	setLocalNoBannerCachePeriod,
 	MODULES_VERSION,
+	getEmail,
 } from '@root/src/web/lib/contributions';
 import { getCookie } from '@root/src/web/browser/cookie';
 import {
@@ -52,6 +53,7 @@ type CanShowProps = BaseProps & {
 	remoteBannerConfig: boolean;
 	section: string;
 	isPreview: boolean;
+	idApiUrl: string;
 };
 
 type ReaderRevenueComponentType =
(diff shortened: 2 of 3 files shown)


src/web/browser/coreVitals/coreVitals.ts
```diff
@@ -33,11 +33,11 @@ export const coreVitals = (): void => {
 		name: string;
 		value: number;
 	};
-	
+
 	const nineDigitPrecision = (value: number) => {
 		// The math functions are to make sure the length of number is <= 9
 		return Math.round(value * 1_000_000) / 1_000_000;
-	}
+	};
```
TomPretty, line 40: This is unrelated, but I was failing pre-push hooks without running prettier on this file

liywjl: I have a PR to fix it https://github.com/guardian/dotcom-rendering/pull/2995

Right now github actions are down so will merge on monday

liywjl (approved):
LGTM

src/web/browser/coreVitals/coreVitals.ts
TomPretty: Ah okay nice 👍 I can rebase after that's merged

tomrf1 (approved):
👍

src/web/layouts/lib/stickiness.tsx
```diff
@@ -32,8 +32,6 @@ const bannerWrapper = css`
 	position: fixed !important;
 	bottom: 0;
 	${getZIndexImportant('banner')}
-	max-height: 80vh;
-	overflow: visible;
```
TomPretty, line 36: This is to allow the banner to go 'fullscreen' on mobile and to get the right scrolling behaviour

oliverlloyd: Is this comment something to worry about?

`Overflow is visible because puzzles banner needs it to render correctly`

TomPretty: Very much so! I've changed that back!

oliverlloyd: Ah nice, glad we dodged that!

I wonder if this is perhaps symptomatic of a larger issue? It'd be great to have a more centralised, controlled, solution for showing banners (or any overlay). Right now we've got multiple silos operating in the same problem space and I think we've set ourselves up for these kinds of issues. Using code comments as a way to guard against css changes affecting other teams is not a great position to be in!

Just a random thought, not a blocker for this PR

TomPretty: Yeah it's a great thought. The trickiness here was because the puzzles banner can't be dismissed like the others, but instead is minimized which leaves an absolutley positioned bit of the banner in the bottom right corner (hence the need for `overflow: visible` as the banner then has height 0). In this case, we just wanted `overflow: auto` to allow the banner to be scrollable incase a user has increased their font size so much that the banner overflows.

I'm not really sure what the general solution should be. We've definitely got accessibility concerns with the banners so it'd be good to bash that out!

PR #2998, opened 2021-05-14
Prevent Braze web sdk from auto populating fields

michaelclapham:
# What does this change?

## Before
No devicePropertyAllowlist set on braze sdk initialization meaning Braze was infering multiple fields on the first time a user loads the sdk.

These fields are listed here:
https://www.braze.com/docs/developer_guide/platform_integration_guides/web/cookies_and_storage/#device-properties

## After
devicePropertyAllowlist set to empty list on braze sdk initialisation

## Why?
Braze is infering certain fields incorrectly, this change prevents that.

### DIFF
--- a/src/web/browser/coreVitals/coreVitals.ts
+++ b/src/web/browser/coreVitals/coreVitals.ts
@@ -33,11 +33,11 @@ export const coreVitals = (): void => {
 		name: string;
 		value: number;
 	};
-	
+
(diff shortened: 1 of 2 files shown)


adamnfish (approved):
Looks great!

Once it's out let's keep an extra eye out for any issues with the SDK's loading and display of messages, and keep an ear open for Marketing asking why their campaign that secretly depends on these fields is having problems!

Nice one.

PR #2999, opened 2021-05-17
Update definition of CodeBlockElement

shtukas:
This corrects the definition of `CodeBlockElement` to match the backend definition. This will prevent pages with `CodeBlockElement`s in them to fail.

Note: When the DCR definition of `CodeBlockElement` was first made, it made sense to call the attribute `code`, but it's called `html` in CAPI and in the backend PageElement, so I am mirroring that here. 

The `CodeBlockComponent`'s corresponding attribute can still be called `code` since this is what it means in that context.

### DIFF
--- a/src/lib/content.d.ts
+++ b/src/lib/content.d.ts
@@ -86,7 +86,7 @@ interface QuizAtomBlockElement {
 interface CodeBlockElement {
 	_type: 'model.dotcomrendering.pageElements.CodeBlockElement';
 	elementId: string;
-	code: string;
+	html: string;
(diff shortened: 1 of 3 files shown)


shtukas:
The following picture was taken against ( https://github.com/guardian/frontend/pull/23814 )

Screenshot 2021-05-17 at 14 15 29

liywjl (approved):
LGTM 👍

shtukas:
Now working against PROD: 

Screenshot 2021-05-17 at 14 41 27

PR #3000, opened 2021-05-17
readme

shtukas:
When the list of Languages was defined in DCR, they forgot to add `text` which is actually being used in CAPI

Screenshot 2021-05-17 at 14 47 58


nicl (approved):
Nice. Are there any more we should include too?

shtukas:
@nicl I was asking myself the same question. I am making a frontend PR that is going to guard against sending any wrong one

PR #3001, opened 2021-05-17
Refactor element renderer

nicl:
Follow up from https://github.com/guardian/dotcom-rendering/pulls?q=is%3Apr+is%3Aclosed+author%3Anicl to better separate out concerns.

In particular, the aim is to: separate out role-based styling and the `Figure` wrapper (and metadata) from the element conversion itself. This is to enable re-use of the conversion code even when the wrapping is not required (e.g. for Interactives).

To achieve this, the `ok, resp` pattern is used (inspired by Go indeed!). The dual-return/ok flag feels like it should be unnecessary but unfortunately React doesn't provide a clean way to tell if an element is `null` so it needs to be encoded separately.


liywjl (approved):
It would be great to try and explain what you are trying to do here in the description.

I see the pattern you are implementing is the "comma, ok" idiom (very similar to Golang), I personally would have prefered this abstraction to look more like how we use `enhance-numbered-lists.ts` in `render.ts`, by filtering unused CAPI elements.

Although this is more a question of taste, this implementation works 👍

nicl:
@liywjl thanks. I've updated the description as agree this could be a lot clearer. The new functions are commented though, hopefully in a way that explains things but, unfortunately, as I renamed the main file, Github is collapsing these by default so not that obvious from the file view.

PR #3002, opened 2021-05-17
Disable Prism.highlight (for the moment)

shtukas:
Now that Frontend is about to send ( https://github.com/guardian/frontend/pull/23815 ) to `CodeBlockElement`'s `language` to DCR, then  we need to disable Prism.highlight (for the moment), as it is broken.

BEFORE:

Screenshot 2021-05-17 at 15 43 27

AFTER:

Screenshot 2021-05-17 at 15 44 26


liywjl (approved):
LGTM 👍

PR #3005, opened 2021-05-18
Redefine CodeBlockElement's language as string

shtukas:
## What does this change?

This moves away from defining the `language` attribute of the `CodeBlockElement` from a union type to a string. The previous definition is overkill and would pose validation duplication on the backend side.


liywjl (changes requested):
I feel like we need this to be able to style/format the code correctly?

liywjl (approved):
LGTM 👍

PR #3006, opened 2021-05-18
Add amp-sticky-ad component

ealitten:
## What does this change?

Add StickyAd component which wraps Ad to make it show in a sticky footer

https://trello.com/c/WW61ZO0F/718-implement-mobile-sticky-on-amp-in-us

### After

Screenshot 2021-05-19 at 10 03 32

## Why?

An extra option for displaying ads which stay in view on mobile pages

### DIFF
--- a/src/amp/components/Ad.tsx
+++ b/src/amp/components/Ad.tsx
@@ -9,6 +9,9 @@ const sizes = [
 	{ width: 250, height: 250 }, // Square
 ];
 
+// Note: amp-sticky-ad has max height of 100
+const stickySizes = [{ width: 320, height: 50 }]; // Mobile Leaderboard
+
 const adClass = css`
 	display: none;
 `;
@@ -107,6 +110,7 @@ interface CommercialConfig {
 }
 
 export const Ad = ({
+	isSticky,
 	adRegion,
 	edition,
 	section,
@@ -114,6 +118,7 @@ export const Ad = ({
 	config,
 	commercialProperties,
 }: {
+	isSticky?: boolean;
 	adRegion: AdRegion;
 	edition: Edition;
 	section: string;
@@ -121,8 +126,18 @@ export const Ad = ({
 	config: CommercialConfig;
 	commercialProperties: CommercialProperties;
 }) => {
-	const { width, height } = sizes[0]; // Set initial size to MPU
-	const multiSizes = sizes.map((e) => `${e.width}x${e.height}`).join(',');
+	let width;
+	let height;
+	let multiSizes;
+
+	if (isSticky) {
+		[{ width, height }] = stickySizes; // Set initial size to Mobile Leaderboard
+		multiSizes = stickySizes.map((e) => `${e.width}x${e.height}`).join(',');
+	} else {
+		[{ width, height }] = sizes; // Set initial size to MPU
+		multiSizes = sizes.map((e) => `${e.width}x${e.height}`).join(',');
+	}
+
 	return (
 		<amp-ad
 			class={cx(adClass, adRegionClasses[adRegion])}
--- a/src/amp/components/Ad.tsx
+++ b/src/amp/components/Ad.tsx
@@ -109,6 +109,16 @@ interface CommercialConfig {
 	usePermutive: boolean;
 }
 
+export interface AdProps {
+	isSticky?: boolean;
+	adRegion: AdRegion;
+	edition: Edition;
+	section: string;
+	contentType: string;
+	config: CommercialConfig;
+	commercialProperties: CommercialProperties;
+}
+
 export const Ad = ({
 	isSticky,
 	adRegion,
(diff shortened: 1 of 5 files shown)


src/amp/components/Ad.tsx
```diff
@@ -107,22 +110,34 @@ interface CommercialConfig {
...
+	let multiSizes;
+
+	if (isSticky) {
+		[{ width, height }] = stickySizes; // Set initial size to Mobile Leaderboard
+		multiSizes = stickySizes.map((e) => `${e.width}x${e.height}`).join(',');
+	} else {
+		[{ width, height }] = sizes; // Set initial size to MPU
+		multiSizes = sizes.map((e) => `${e.width}x${e.height}`).join(',');
+	}
+
```
liywjl, line 140:
```suggestion
	const { width, height } = isSticky ? stickySizes : sizes[0]; // Set initial size to MPU
	const multiSizes = sizes.map((e) => `${e.width}x${e.height}`).join(',');
```

liywjl (approved):
LGTM

src/amp/components/Ad.tsx
ealitten: Hey, thanks for the suggestion. It doesn't cover using `stickySizes` vs `sizes` for multiSizes though. I was contemplating doing something like this, which has a bit less repetition in it:
```js
const adSizes = isSticky ? stickySizes : sizes;
[{ width, height }] = adSizes; // Set initial size as first element
multiSizes = adSizes.map((e) => `${e.width}x${e.height}`).join(',');
```

src/amp/components/Ad.tsx
```diff
@@ -107,22 +110,26 @@ interface CommercialConfig {
...
 	edition: Edition;
 	section: string;
 	contentType: string;
 	config: CommercialConfig;
 	commercialProperties: CommercialProperties;
 }) => {
-	const { width, height } = sizes[0]; // Set initial size to MPU
-	const multiSizes = sizes.map((e) => `${e.width}x${e.height}`).join(',');
+	const adSizes = isSticky ? stickySizes : inlineSizes;
+	const [{ width, height }] = adSizes; // Set initial size as first element (should be the largest)
```
mxdvl, line 130:
```suggestion
	const [{ width, height }] = adSizes[0]; // Set initial size as first element (should be the largest)
```

mxdvl (changes requested):
I think the implementation details of AMP should be abstracted away as much as possible, and would prefer to use `<Ad isSticky={true}>` directly rather than having to wrap it inside a `<StickyAd>`.

```tsx
<Ad
  isSticky={true}
  /* … */ 
>
```

If the prop is passed, the Ad component can handle wrapping the html on its own.

src/amp/components/Ad.tsx
ealitten: We're already accessing the first element of the array via destructuring.

`const [{ width, height }] = adSizes[0]` is the same as `const { width, height } = adSizes[0]`

mxdvl: Sorry, I didn't spot those [square brackets]!

mxdvl (approved):
👏

ealitten:
Updated to add 'Advertisement' label to the sticky ad as per our requirements. It was only possible to achieve this through global styles, since <amp-sticky-ad> will only allow <amp-ad> as a child and cannot be wrapped.

src/amp/components/StickyAd.tsx
```diff
@@ -2,6 +2,21 @@ import React from 'react';
 import { Ad } from './Ad';
 import type { AdProps } from './Ad';
 
+// This CSS is added to global styles in document.tsx, to add the Advertisment label to the sticky
+export const stickyAdLabel = `
```
mxdvl, line 6: I think this could be a tad more explicit.

```suggestion
export const stickyAdLabelCss = `
```

src/amp/server/document.tsx
```diff
@@ -5,6 +5,7 @@ import { CacheProvider } from '@emotion/core';
 import { cache } from 'emotion';
 import resetCSS from /* preval */ '@root/src/lib/reset-css';
 import { getFontsCss } from '@root/src/lib/fonts-css';
+import { stickyAdLabel } from '@root/src/amp/components/StickyAd';
```
mxdvl, line 8: I don’t think we need to add this until we have a sticky ad on the page.

```suggestion
```

mxdvl (changes requested):
I think this approach works. Make sure to look at other Ad Labels, as they specifically do not use Guardian fonts.

https://github.com/guardian/dotcom-rendering/blob/b1e5706349518810cd44cf665227e4ee0a1bd037/src/amp/components/BodyArticle.tsx#L70-L81

ealitten:
> I think this approach works. Make sure to look at other Ad Labels, as they specifically do not use Guardian fonts.

Yeah this is the same, the font-family line takes care of that and you can see in the screenshot above it's generic sans font

mxdvl (approved):

PR #3007, opened 2021-05-18
Fix spelling mistakes

mxdvl:
### Before

Mispeltt Donut

### After

Is spelt Doughnut

## Why?

We’re in the publishing business!

### DIFF
--- a/src/amp/components/Blocks.tsx
+++ b/src/amp/components/Blocks.tsx
@@ -24,7 +24,7 @@ const adStyle = css`
 		content: 'Advertisement';
 		display: block;
 		${textSans.xsmall()};
-		/* Adverts specifcally don't use the GU font branding. */
+		/* Adverts specifically don't use the GU font branding. */
 		/* stylelint-disable-next-line property-blacklist */
 		font-family: 'Helvetica Neue', Helvetica, Arial, 'Lucida Grande',
 			sans-serif;
--- a/src/amp/components/BodyArticle.tsx
+++ b/src/amp/components/BodyArticle.tsx
@@ -71,7 +71,7 @@ const adStyle = css`
 		content: 'Advertisement';
 		display: block;
 		${textSans.xsmall()};
-		/* Adverts specifcally don't use the GU font branding. */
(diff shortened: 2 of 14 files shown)


buck06191:
Oh sweet, delicious irony. My favourite flavour of doughnut.
image

On a different note, "Donut" is a generally accepted spelling and whilst not a fan of the general Americanisation of English that seems to happen all over software development, it's generally easier to type and many people who learn English as a second language do so with an Americanised variant. So I'm unsure on whether it's worth the effort to change a component name here?

All the rest I'm a fan of though

shtukas (approved):

mxdvl:
I think it’s best to defer to [the style guide](https://www.theguardian.com/guardian-observer-style-guide-d)?

> **doughnuts**
> despite Dunkin’ Donuts

For the record, it was spelled “properly” in `frontend`: https://github.com/guardian/frontend/blob/5b970cd7308175cfc1bcae2d4fb8c06ee13c5fa0/static/src/javascripts/projects/common/modules/charts/doughnut.js

> On a different note, "Donut" is a generally accepted spelling and whilst not a fan of the general Americanisation of English that seems to happen all over software development, it's generally easier to type and many people who learn English as a second language do so with an Americanised variant.

I don’t know, as an English non-native speaker, the “dough” in “doughnut” makes more sense to me. Happy to go either way if a third party weighs in. I, however won’t weigh myself after eating doughnuts.

Further reading: https://time.com/2837756/donut-or-doughnut/

buck06191:
> I don’t know, as an English non-native speaker, the “dough” in “doughnut” makes more sense to me. Happy to go either way if a third party weighs in. I, however won’t weigh myself after eating doughnuts.

Makes more sense to me as well. Happy to defer to the style guide and preserve the more sensical spelling tbh

oliverlloyd:
> however won’t weigh myself after eating doughnuts

😸 

I suspect there was a reason I went with `Donut` but I have utterly no idea what it was and given frontend is more doughy that seems a good reason to change.

It's surprising how many code foods there are on a single page of the guardian

PR #3008, opened 2021-05-18
Fix Labs Immersive caption left col

liywjl:
## What does this change?
In Immersive Labs, the 1st caption (leftColCaption) needs to be updated to use the triangle asset (#121212), the font colour should be #333333 (font colour was fixed in #3010)
### Before
image

### After
image

## Why?
Update to follow new design


OllysCoding:
Currently re-working as some details on the card were missed - do not merge

shtukas (approved):

PR #3009, opened 2021-05-19
Use body-parser for JSON payload [in dev]

shtukas:
This is uses `bodyParser` to parse the JSON DCR data object on local to prevent a "request entity too large" problem we observed with this url

```
http://localhost:3030/Article?url=http://localhost:9000/info/2018/nov/30/bye-bye-mongo-hello-postgres
```

Note: The DCR data object JSON serialization is well typically below 1Mb, so the 10Mb limit should be enough.

BEFORE:

Screenshot 2021-05-19 at 09 50 30

Screenshot 2021-05-19 at 09 50 51

AFTER:

Screenshot 2021-05-19 at 11 02 37


liywjl (approved):
LGTM, though I am a little concerned we need to do this, could we not look at reducing the payload?

shtukas:
@liywjl 
The payload is small. The dev settings were just overly limiting :)

PR #3010, opened 2021-05-19
Fix labs photo essay caption colour

OllysCoding:
## What does this change?

~~In Labs Photo-essays the caption colour needs to be a darker brand colour (brand-400) instead of using the labs theme colour.~~

Update: Based on discussion, PR will now change *all* labs articles to use neutral.20 as the caption colour, including photo essays.

### Before (PROD/frontend)
Non-Photo Essay:
image
Photo Essay: 
image

### Before (DCR)
Non-Photo Essay:
image

Photo Essay:
image

### After (DCR)
Non-Photo Essay:
image

Photo Essay:
image

## Why?
Bring to parity with frontend

### DIFF
--- a/src/web/lib/decidePalette.ts
+++ b/src/web/lib/decidePalette.ts
@@ -124,9 +124,10 @@ const textTwitterHandle = (format: Format): string => {
 
 const textCaption = (format: Format): string => {
 	if (format.theme === Special.SpecialReport) return specialReport[100];
+	if (format.theme === Special.Labs) return neutral[20];
+
 	switch (format.design) {
 		case Design.PhotoEssay:
-			if (format.theme === Special.Labs) return brand[400];
 			return pillarPalette[format.theme].dark;
 		default:
 			return text.supporting;
--- a/src/web/lib/decidePalette.ts
+++ b/src/web/lib/decidePalette.ts
@@ -126,6 +126,7 @@ const textCaption = (format: Format): string => {
 	if (format.theme === Special.SpecialReport) return specialReport[100];
 	switch (format.design) {
 		case Design.PhotoEssay:
+			if (format.theme === Special.Labs) return brand[400]
 			return pillarPalette[format.theme].dark;
 		default:
 			return text.supporting;
--- a/src/web/lib/decidePalette.ts
+++ b/src/web/lib/decidePalette.ts
@@ -126,7 +126,7 @@ const textCaption = (format: Format): string => {
 	if (format.theme === Special.SpecialReport) return specialReport[100];
 	switch (format.design) {
 		case Design.PhotoEssay:
-			if (format.theme === Special.Labs) return brand[400]
+			if (format.theme === Special.Labs) return brand[400];
 			return pillarPalette[format.theme].dark;
 		default:
 			return text.supporting;


shtukas (approved):

jfsoul:
Let's double check this one with design

liywjl (approved):
LGTM! would be good to link the article used in question to show parity with frontend/DCR

dom-whooley:
You can reference the approved design for captions on the Labs article pages here:  https://www.figma.com/file/KyvpghEkqBItTsS4ZjkjJI/Labs-Accessibility?node-id=301%3A1794 the caption style. the colour was changed to Neutral 2.0. IMPORTANT: This is only on the standard templates not the immersive template for now. Any changes to the immersive template needs approval from the Labs team first.

jfsoul:
@dom-whooley I think there's an extra facet to this one, so just want to make sure we're on the same page. It's for "photo essay" captions specifically, which are a specific type of article with special caption styling. Photo essays would normally (or perhaps always) be immersive as well - so this change is going to apply to the immersive template.

 The current proposed change in this PR aligns the caption with the colour used on frontend, but perhaps we'd be better off aligning with your designs for other types of labs articles? (Which is what you suggested, but just making it clear that this is a change to the immersive content type realistically)

dom-whooley:
I see. Do you have a photo essay example page I can reference to better understand?

OllysCoding:
@dom-whooley 

Here are some reference articles I checked against when putting together the change:

https://www.theguardian.com/saving-for-a-sunny-day-with-nsandi/2021/apr/20/its-incredibly-liberating-what-saving-for-a-piano-taught-me-about-my-finances?dcr=false
https://www.theguardian.com/100-teachers/2021/mar/26/i-create-a-safe-environment-for-them-meet-the-hockey-playing-teacher-helping-pupils-shoot-for-maths-success?dcr=false

jlieb10:
FWIW: #65A897 (brand 400/the light green) always fails the accessibility check on a light background (even pure white) - if there are other instances of this text on the site, they should be flagged for removal

dom-whooley:
Forgive me, but it looks like the caption colour change from Labs.400 to brand.400 been pushed to live pages. Is this correct? I'm just aware that there is an agreement with the labs team that any changes would be approved by them first. Brand.400 isn't in their colour palette and judging by the reaction the last time changes were made without their consent. I suggest this is rolled back until they approve it.

oliverlloyd:
> Forgive me, but it looks like the caption colour change from Labs.400 to brand.400 been pushed to live pages. Is this correct? I'm just aware that there is an agreement with the labs team that any changes would be approved by them first. Brand.400 isn't in their colour palette and judging by the reaction the last time changes were made without their consent. I suggest this is rolled back until they approve it.

I don't think the Labs team should ever act in the way they did previousy. Regardless of if any changes are pushed without their consent or not.

They should not be allowed to act like a mafia.

dom-whooley:
> > Forgive me, but it looks like the caption colour change from Labs.400 to brand.400 been pushed to live pages. Is this correct? I'm just aware that there is an agreement with the labs team that any changes would be approved by them first. Brand.400 isn't in their colour palette and judging by the reaction the last time changes were made without their consent. I suggest this is rolled back until they approve it.
> 
> I don't think the Labs team should ever act in the way they did previousy. Regardless of if any changes are pushed without their consent or not.
> 
> They should not be allowed to act like a mafia.

I'm not disagreeing with you. I'm just saying, that it was my impression that that was what was agreed between dev and labs. I'm just in the middle here trying to avoid what happened before.

jfsoul:
I suggest we take the conversation to a meeting/chat as we're ending up with some confusion, but I would say that I don't think the live pages will have been changed. I suspect brand[400] has always been used for photo essay captions.

jfsoul:
So to update on the situation... we think we should aim to align these caption colours with those of other labs articles  (in the new designs). This corresponds to #333 or neutral.20 by the looks of it.  We can make that change absolutely safely here now because these pages are not live on DCR, and the stakeholders will be able to preview for their information/feedback before it goes live anyway.

And I've said it elsewhere, but just for the thread here - if we do suspect any issues with the live site then reach out to the team on chat or email.

jfsoul (approved):

OllysCoding:
Quick update on this PR, description has been updated

TL;DR: DCR labs captions have been updated to be neutral.20 across the board, in line with the designs. This change does not immediately affect prod.

paperboyo:
2012-05-23-Foto-Falcone-AFP-PHOTOGERARD-FOUET_TINIMA20120523_0205_1

PR #3012, opened 2021-05-19
use latest guardian libs & types

sndrs:
## What does this change?

bumps these libs to remove some ugliness that @ioanna0 was forced to do in #2967

### Before

`import type { CountryCode } from '@guardian/libs/dist/esm/types/countries';`

### After

`import type { CountryCode } from '@guardian/types';`

## Why?

that's where it should have lived, and shows why types should be packaged by themselves


liywjl (approved):
👍  LGTM

mxdvl (approved):
👏 🧹

PR #3013, opened 2021-05-19
Introduce basic interactive layout (WIP)

nicl:
Note, this is a WIP and has at least the following issues:

* lots of duplicated code from other layouts
* interactives look pretty broken(!) - though this is expected and not something I'm seeking to fix here

Very open to thoughts on how to improve this. At the moment I think we can pull out some of the shared layout components to get a bit better re-use. But also keen to emphasise that this is exploratory code (to help us better understand backwards compatibility challenges with interactives) and likely to change a lot as we further work on interactive support.

## What does this change?

Adds a basic interactive page implementation. At the moment, this means:

* no right-column
* some redundant code removed (e..g for match reports)

### Before

Nada.

### After

e.g. http://localhost:3030/Article?url=https://www.theguardian.com/australia-news/datablog/ng-interactive/2021/may/14/covid-19-vaccine-rollout-coronavirus-updates-tracker-australia-daily-live-data-stats-update-total-numbers-distribution-progress-schedule-tracking-new-cases-today-statistics-latest-news

Screenshot 2021-05-19 at 17 45 39

## Why?

To push towards DCR interactives.

### DIFF
--- a/src/web/layouts/InteractiveLayout.tsx
+++ b/src/web/layouts/InteractiveLayout.tsx
@@ -1,0 +1,622 @@
+import React from 'react';
+import { css } from 'emotion';
+
+import {
+	neutral,
+	brandAltBackground,
+	brandBackground,
+	brandBorder,
+	brandLine,
+	labs,
+	border,
+} from '@guardian/src-foundations/palette';
+import { from, until } from '@guardian/src-foundations/mq';
+import { Design, Special } from '@guardian/types';
+import type { Format } from '@guardian/types';
+
+import { GuardianLines } from '@root/src/web/components/GuardianLines';
+import { StarRating } from '@root/src/web/components/StarRating/StarRating';
+import { ArticleBody } from '@root/src/web/components/ArticleBody';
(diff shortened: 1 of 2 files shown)


src/web/layouts/InteractiveLayout.tsx
```diff
@@ -0,0 +1,622 @@
...
+}
+
+export const InteractiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
+	const {
+		config: { isPaidContent, host },
+	} = CAPI;
+
+	const adTargeting: AdTargeting = buildAdTargeting(CAPI.config);
+
+	const showBodyEndSlot =
```
jfsoul, line 221: Is the epic on interactives? Can we kill this?

jfsoul (approved):

src/web/layouts/InteractiveLayout.tsx
nicl: Good point, I don't think so. I'll remove for now and we can always re-add if needed. (Though I suspect this code will change a lot anyway.)

PR #3015, opened 2021-05-20
Hide RR banners in preview

tomrf1:
since preview mode is used internally


cypress/integration/parallel-5/discussion.spec.js
```diff
@@ -1,29 +1,29 @@
...
-	});
-
-	it('should automatically expand and scroll to a comment if the reader loads a permalink', function () {
-		cy.visit(`/Article?url=${permalink}`);
-		cy.contains(
-			'In the world of human psychology, change is glacially slow',
-		);
-	});
-});
+// import { disableCMP } from '../../lib/disableCMP';
```
tomrf1, line 1: these tests are hanging forever and preventing the build from finishing

shtukas (approved):

PR #3018, opened 2021-05-20
Use Emotion 11 CSS

liywjl:
## What does this change?
- Use Emotion 11
- remove `cx` and replace with `[...]` pattern

## Why?
Before depricating Emotion 10, we need to move our logic over to Emotion 11


liywjl:
closing, due to large amount of conflicts

PR #3020, opened 2021-05-21
Use JSX and remove React as import

liywjl:
## What does this change?
- set `tsconfig` to use `react-jsx` and `jsxImportSource` to use `@emotion/react`
- Remove `React` as an import
- turn off Eslint requireing `React` as import  

## Why?
Emotion 11 uses JSX, so we need to move over to explicitly use JSX. In doing so, we no longer need to import React at the top of each React component


shtukas (approved):

nicl (approved):
🥳

PR #3023, opened 2021-05-21
Be more explicit around localStorage availability

tjmw:
## What does this change?

Be more explicit around localStorage availability. If `localStorage` isn’t available then bail early from attempting to show
Braze messages. We rely on localStorage to track whether a user has been eligible for messages and the Braze SDK relies on it too, so check early and return a NullBrazeMessages instance if it’s not available.

### Before

Occasional errors in Sentry which point to attempting to use `localStorage` when it's not available.

### After

Fewer Sentry errors, which should make it easier to spot genuine issues.

### DIFF
--- a/src/web/lib/braze/buildBrazeMessages.ts
+++ b/src/web/lib/braze/buildBrazeMessages.ts
@@ -11,6 +11,7 @@ import {
 	LocalMessageCache,
 	NullBrazeMessages,
 } from '@guardian/braze-components/logic';
+import { isLocalStorageAvailable } from '@root/src/web/lib/isLocalStorageAvailable';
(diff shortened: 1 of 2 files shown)


liywjl (approved):
LGTM

sndrs:
can i interest you in this little thing? https://github.com/guardian/libs/blob/main/src/storage.README.md#isavailable

tjmw:
Thanks for the tip @sndrs 👍  I thought of that lib but discounted it as I'd not realised `isAvailable` existed (I just want my code to bail if localstorage isn't available, rather than silently fail). I'll make this switch, either in this or a subsequent PR.

PR #3024, opened 2021-05-21
Remove 'items' property from guide atom

OllysCoding:
## What does this change?

Removes the `items` property from `GuideAtomBlockElement`. The guide atom renders using the `html` field instead.

[Guide Atom](https://github.com/guardian/atoms-rendering/blob/main/src/GuideAtom.tsx) - [Frontend PR](https://github.com/guardian/frontend/pull/23831)

## Why?

Reduces clutter & keeps code clean and up-to-date


OllysCoding:
Apologies in advance to anyone who was to scroll through the sea of fixtures updates - since the JSON endpoint was updated in the frontend PR the fixtures needed to be re-ran.

shtukas (approved):

PR #3025, opened 2021-05-24
Update atoms version to explicitly avoid audio preload

jfsoul:
## What does this change?
This PR bumps the version of atoms rendering we're using to explicitly avoid preloading audio atom content with `preload="none"`. The reason we're doing this is because clients have different default behaviour for `preload` and some may perform preloading in such a way that Acast interprets it as a listen, meaning our numbers will be inflated.

We've checked this with Acast who have strongly advised this change, and it mirrors previous frontend behaviour too.

### Before
`<audio>`

### After
`<audio preload="none">`

## Why?
As above

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
         "@emotion/react": "^11.4.0",
         "@guardian/ab-core": "^2.0.0",
         "@guardian/ab-react": "^2.0.1",
-        "@guardian/atoms-rendering": "^13.0.0",
+        "@guardian/atoms-rendering": "^14.0.0",
         "@guardian/automat-client": "^0.2.17",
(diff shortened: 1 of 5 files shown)


shtukas (approved):

liywjl (approved):
👍  we should spin up a ticket to update `atoms` package to fix forward

jfsoul:
> we should spin up a ticket to update atoms package to fix forward
@liywjl sorry not sure what you mean - atoms-rendering has been updated so I think this is a forward fix if I understand correctly?

liywjl:
my bad I misread, I thought we were going from atoms v14 to v13 😅 ignore me

PR #3026, opened 2021-05-24
Emotion react for layout files

liywjl:
## What does this change?
Use @emotion/react for files in the Layout folder

## Why?
Progressively migrate over to @emotion/react

### DIFF
--- a/src/web/layouts/DecideLayout.tsx
+++ b/src/web/layouts/DecideLayout.tsx
@@ -6,8 +6,6 @@ import { decideDisplay } from '@root/src/web/lib/decideDisplay';
 import { decidePalette } from '@root/src/web/lib/decidePalette';
 import { decideDesign } from '@root/src/web/lib/decideDesign';
 
-import { injectGlobal } from 'emotion';
-import { focusHalo } from '@guardian/src-foundations/accessibility';
 import { StandardLayout } from './StandardLayout';
 import { ShowcaseLayout } from './ShowcaseLayout';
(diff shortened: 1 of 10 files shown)


src/web/layouts/DecideLayout.tsx
```diff
@@ -21,15 +19,6 @@ type Props = {
...
 
-// eslint-disable-next-line @typescript-eslint/no-unused-expressions
-injectGlobal`
-	/* Crude but effective mechanism. Specific components may need to improve on this behaviour. */
-	/* The not(.src...) selector is to work with Source's FocusStyleManager. */
-	*:focus {
-		${focusHalo}
-	}
-`;
-
```
liywjl, line 32: `injectGlobal` does not exist in `@emotion/react`, instead we should use `Global` which is a react component. As it needs to be rendered as a component, it makes sense to lift the rendering to the parent component (`document.tsx`)

src/web/server/document.tsx
```diff
@@ -54,11 +57,24 @@ const decideTitle = (CAPI: CAPIType): string => {
...
 				<StrictMode>
+					<Global
+						styles={css`
+							/* Crude but effective mechanism. Specific components may need to improve on this behaviour. */
+							/* The not(.src...) selector is to work with Source's FocusStyleManager. */
+							*:focus {
+								${focusHalo}
+							}
+						`}
+					/>
```
liywjl, line 77: ditto `injectGlobal`

shtukas (approved):

PR #3029, opened 2021-05-24
Hide sign-in gate on preview

jfsoul:
## What does this change?
This PR hides the sign-in gate from being shown on `preview` (i.e. the viewer that journalists use to preview content in composer before publishing). We've had a few complaints recently about various dismissible components that interrupt editorial workflow and sign-in gate is a part of that.  Since it serves little purpose in preview and as far as I know we don't use `preview` to test it, we may as well just remove it from there. It can also be harder to maintain a signed-in session on preview, which exacerbates the issue.


liywjl (approved):
LGTM

coldlink (approved):
Thanks! LGTM

PR #3030, opened 2021-05-24
Update Emotion 11 react components (batch 2)

liywjl:
## What does this change?
- use `@emotion/react` 
- change `className` to `css`
- remove `cx` as a reference

## Why?
Slowly move over to use Emotion 11

### DIFF
--- a/src/web/components/AdSlot.tsx
+++ b/src/web/components/AdSlot.tsx
@@ -1,4 +1,4 @@
-import { css, cx } from 'emotion';
+import { css } from '@emotion/react';
 
 import { border, neutral, text } from '@guardian/src-foundations/palette';
 import { textSans } from '@guardian/src-foundations/typography';
@@ -147,12 +147,10 @@ const mobileStickyAdStyles = css`
 
 const AdSlotLabelToggled: React.FC = () => (
 	<div
-		className={cx(
-			'ad-slot__label',
-			'ad-slot__label--toggle',
-			'hidden',
-			adSlotLabelStyles,
+		className={['ad-slot__label', 'ad-slot__label--toggle', 'hidden'].join(
+			' ',
 		)}
+		css={adSlotLabelStyles}
 	>
 		Advertisement
 	</div>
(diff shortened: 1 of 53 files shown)


src/web/components/AdSlot.tsx
```diff
@@ -147,12 +147,10 @@ const mobileStickyAdStyles = css`
 
 const AdSlotLabelToggled: React.FC = () => (
 	<div
-		className={cx(
-			'ad-slot__label',
-			'ad-slot__label--toggle',
-			'hidden',
-			adSlotLabelStyles,
+		className={['ad-slot__label', 'ad-slot__label--toggle', 'hidden'].join(
```
liywjl, line 150: string literals cannot exists in `css` and need to be spun out into thier own `className`

src/web/components/Dropdown.tsx
```diff
@@ -244,16 +244,16 @@ export const Dropdown = ({
...
+					<ul id={dropdownID} css={ulStyles}>
 						{links.map((l, index) => (
 							<li key={l.title}>
 								<a
 									href={l.url}
-									className={cx({
-										[linkStyles]: true,
-										[linkActive]: !!l.isActive,
-										[linkFirst]: index === 0,
-									})}
```
liywjl, line 256: we refactor:
```
{ [style]: condition }
```
to
```
condition && style
```
this is easier to read and works with `css` attribute

src/web/components/Links.tsx
```diff
@@ -91,17 +91,19 @@ const seperatorHideStyles = css`
...
 }: {
-	className?: string;
 	children: React.ReactNode;
 	href: string;
 	dataLinkName: string;
 }) => (
-	<a href={href} className={className} data-link-name={dataLinkName}>
+	<a
+		href={href}
+		css={[linkTablet({ showAtTablet: false }), searchLinkStyles]}
```
liywjl, line 104: before `className` was passed down as a prop. However the component `Search` is only ever used in this file once. So it makes sense to move the `css` directly to the component

src/web/components/RichLink.tsx
```diff
@@ -44,19 +42,19 @@ const neutralBackground = css`
...
 
-const textColour: (palette: Palette) => ColourType = (palette) => {
+const textColour = (palette: Palette) => {
 	return css`
 		color: ${palette.text.richLink};
 	`;
 };
 
-const richLinkTopBorder: (palette: Palette) => ColourType = (palette) => {
+const richLinkTopBorder = (palette: Palette) => {
```
liywjl, line 57: each of these styles no longer return string type, instead we let TS infer type

shtukas (approved):

PR #3034, opened 2021-05-25
Implement new epic article count opt out journey

Aracho1:
## What does this change?
- In line with the new epic article count opt out feature on support-dotcom-rendering, this PR links CMP to the contributions epic and checks whether a reader has opted out of article count


install_nvm.sh
```diff
@@ -0,0 +1,425 @@
+#!/usr/bin/env bash
```
tomrf1, line 1: this file can be deleted!

scripts/nginx/Brewfile.lock.json
```diff
@@ -0,0 +1,22 @@
+{
```
tomrf1, line 1: this can also be deleted

src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx
```diff
@@ -175,7 +182,7 @@ export const canShow = async ({
 
 	const response = await getBodyEnd(
 		contributionsPayload,
-		`${contributionsServiceUrl}/epic${queryString}`,
+		`http://localhost:8082/epic${queryString}`,
```
TomPretty, line 185: This is always really fiddly as we need it when running locally, but we'll don't want this change

src/web/lib/contributions.ts
```diff
@@ -151,7 +148,7 @@ export const getArticleCountConsent = (): Promise<boolean> => {
 				if (!hasRequiredConsents) {
 					removeArticleCountsFromLocalStorage();
 				}
-
+				console.log({ hasRequiredConsents });
```
TomPretty, line 151: we can delete this too!

src/web/components/App.tsx
```diff
@@ -227,7 +227,7 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 	// article pages when other pages are supported by DCR.
 	useEffect(() => {
 		const incrementArticleCountsIfConsented = async () => {
-			if (await getArticleCountConsent()) {
+			if (await hasOptedOutOfArticleCount()) {
```
tomrf1, line 230: just noticed this now needs a `!`
```
if (!hasOptedOut) {
```

src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -149,7 +150,7 @@ export const canShowRRBanner: CanShowFunctionType = async ({
 	}
 
 	const countryCode = await asyncCountryCode;
-	const hasConsentedToArticleCounts = await getArticleCountConsent();
+	const hasConsentedToArticleCounts = await hasCmpConsentForArticleCount();
```
tomrf1, line 153: This actually needs to call `hasOptedOutOfArticleCount`.
Also, the existing name `hasConsentedToArticleCounts` is not helpful here, as it later gets assigned to a different name when we build the payload:
`hasOptedOutOfArticleCount: !hasConsentedToArticleCounts`

Let's just rename `hasConsentedToArticleCounts` to `hasOptedOutOfArticleCount` in this file, and we can avoid using the `!` operator so much

src/web/lib/contributions.ts
```diff
@@ -127,19 +127,19 @@ export const shouldHideSupportMessaging = (
...
 };
 
-export const getArticleCountConsent = (): Promise<boolean> => {
-	if (hasOptedOutOfArticleCount()) {
-		return Promise.resolve(false);
-	}
+export const hasCmpConsentForArticleCount = (): Promise<boolean> => {
 	return new Promise((resolve) => {
+		if (getCookie('gu-cmp-disabled')) {
+			resolve(true);
```
tomrf1, line 141: this prevents the cypress tests from breaking, as `onConsentChange` doesn't work

cypress/integration/parallel-3/article.interactivity.spec.js
```diff
@@ -54,7 +54,7 @@ describe('Interactivity', function () {
 				cy.contains('Lifestyle');
 				cy.get('[data-component="most-popular"]').scrollIntoView({
 					duration: 300,
-					offset: { top: -30 },
+					offset: { top: 50 },
```
tomrf1, line 57: this test is flakey - it needs to scroll down enough to trigger the request

tomrf1 (approved):
great!

PR #3035, opened 2021-05-25
Temp disable share count cypress test

nicl:
* disable Cypress share count test
* respect share count switch (`serverShareCounts`)

Why? - because a Facebook API change has broken our share counts for now. We can turn back on once fixed.


liywjl (approved):
LGTM

jlieb10 (approved):
👍

PR #3038, opened 2021-05-25
Remove emotion 10

liywjl:
## What does this change?
- remove `emotion` and `@emotion/core`
- remove `emotion` as automat export

## Why?
Final step to removing Emotion from DCR


src/web/components/Card/Card.tsx
```diff
@@ -102,18 +101,6 @@ const starWrapper = css`
...
-	if (format.theme === Special.Labs) {
-		return css`
-			${textSans.medium({ fontWeight: 'regular' })}
-		`;
-	}
-	return css`
-		${headline.xxxsmall({ fontWeight: 'regular' })}
-		line-height: 20px;
-	`;
-};
```
liywjl, line 115: technically this was never used, but was flaged as part of the update

shtukas (approved):

src/web/lib/setAutomat.ts
```diff
@@ -9,13 +8,9 @@ let hasAutomatBeenSet = false;
 export const setAutomat = () => {
 	if (!hasAutomatBeenSet) {
 		window.guardian.automat = {
-			react: React,
-			preact: React,
```
liywjl, line 13: we use `react` instead of `preact` in SDC

nicl (approved):
🥳

PR #3040, opened 2021-05-25
Fix navbar Z index bug 

OllysCoding:
## What does this change?

Add a custom z-index for the edition dropdown which is below the stick ad banner. This solves the issue without having to change z-indexes for any other elements on the site.

### Before
image
image
### After
image
image

## Why?

Visual Bug


shtukas (approved):

liywjl:
the snapshot `after` still looks broken. It looks like that Ad still covers it? is this because you scrolled down on the page?

OllysCoding:
> It looks like that Ad still covers it?

Yea so the bug was that as you scroll down, parts of the editions dropdown appeared on top of the banner, rather than being covered by it. In the first 'before' screenshot you can see a small vertical grey line - and on the second, when the dropdown is selected you can see the text outline. 

I don't think the dropdown itself (when open) was part of the original ticket, so with the new z-indexes, the dropdown won't open 'on top' of the banner, if we wanted to change that we could add an additional z-index (the current one wraps the entire element) for just the dropdown menu itself.

It's super hard to convey the bug in screenshots, ticket is here: https://trello.com/c/TWcqpWh1/231-weird-bar-nav-above-top-ads

liywjl (approved):
Ahh that makes more sense, yeah I understand what you mean 👍

PR #3041, opened 2021-05-25
Enable sticky ads in US

ealitten:
### After

Add the StickyAd component to BodyArticle, so they start appearing. Unfortunately due to restrictions on the <amp-sticky-ad> AMP component, the only way to add the "Advertisement" label was to include it in the global CSS in document.tsx.

Component added in previous PR: https://github.com/guardian/dotcom-rendering/pull/3006

Note: This PR is labelled `donotmerge` since turning this on is currently being confirmed with US ads team

https://user-images.githubusercontent.com/19875457/119694523-323fe080-be45-11eb-8a1e-aa75881afbf6.mov

## Why?

Additional ad revenue opportunity

### DIFF
--- a/src/amp/components/BodyArticle.tsx
+++ b/src/amp/components/BodyArticle.tsx
@@ -13,6 +13,7 @@ import { TopMeta } from '@root/src/amp/components/topMeta/TopMeta';
 import { SubMeta } from '@root/src/amp/components/SubMeta';
 import { pillarPalette_DO_NOT_USE } from '@root/src/lib/pillars';
 import { Ad } from '@root/src/amp/components/Ad';
+import { StickyAd } from '@root/src/amp/components/StickyAd';
 import { findAdSlots } from '@root/src/amp/lib/find-adslots';
 import { getSharingUrls } from '@root/src/lib/sharing-urls';
 import { buildAdTargeting } from '@root/src/lib/ad-targeting';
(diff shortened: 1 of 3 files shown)


liywjl (approved):
Could you add some screenshots and a more detailed description?

ealitten:
> Could you add some screenshots and a more detailed description?

Ah yes sorry, I meant to link to the other PR which had a more detailed description. I'll add some in here though

gtrufitt:
Can I recommend that we set this up with a client-side or server side A/B test so that you're able to review against the core-vitals data that we're collecting in dotcom rendering? That way we're able to identify how much these advert types might affect user experience and SEO.

ealitten:
> Can I recommend that we set this up with a client-side or server side A/B test so that you're able to review against the core-vitals data that we're collecting in dotcom rendering? That way we're able to identify how much these advert types might affect user experience and SEO.

@gtrufitt This sticky ad is for AMP pages - I believe we're not able to do our regular A/B testing on AMP pages, unless you know another way? Also do we measure core web vitals on AMP pages?

gtrufitt:
> > Can I recommend that we set this up with a client-side or server side A/B test so that you're able to review against the core-vitals data that we're collecting in dotcom rendering? That way we're able to identify how much these advert types might affect user experience and SEO.
> 
> @gtrufitt This sticky ad is for AMP pages - I believe we're not able to do our regular A/B testing on AMP pages, unless you know another way? Also do we measure core web vitals on AMP pages?

Ah yeah, for some reason I thought this was being added for web. That already exists I think though? Maybe one for the commercial backlog to enable us to track the affect of some ad types on CWV more granularly (may already be in the backlog)

Yeah, A/B testing is possible, but I think a real faff (iirc @buck06191 tried when on commercial). We don't currently record core vitals on AMP.

As you were!

buck06191:
> Yeah, A/B testing is possible, but I think a real faff (iirc @buck06191 tried when on commercial). 

If AMP is being de-prioritised it's probably not worth chasing that work up again, but it would be interesting to know what Google's take on A/B testing on AMP is. From memory their problem was that the AMP consent elements wouldn't trigger the A/B test and so no-one could be tested and tracked properly.

PR #2945, opened 2021-04-29
Identity | Sign In Gate | Fake social test

coldlink:
## Test Definition

We believe that if we show email input and/or social buttons on the gate, it will make the sign in process appear quicker and increase sign in conversion. 

To reduce the effort involved in creating the test, we'd like to first understand whether or not this increase click through from the gate through adding buttons that take users to the first page of sign in/registration - rather than sign in from the gate. 

When a users clicks on the social buttons, take the to the first page of the sign in flow. 

Frontend setup PR: https://github.com/guardian/frontend/pull/23758

## Changes
- Add `SignInGateFakeSocial` test with `fake-social-control`, `fake-social-variant-horizontal`, and `fake-social-variant-vertical` variants
  - `fake-social-variant-horizontal` has the social buttons on mobile lined horizontally
  - `fake-social-variant-vertical` has the social buttons on mobile stacked vertically
  - apart from this on desktop they are identitcal
- Fix global link styling using `data-ignore="global-link-styling"` attribute
- Make the grey body separator line on the gate into it's own css `bodySeparator`
- Fix issue with sign in gate overflow at 320px
- Fix link text colours

## TODO

- [x] Confirm audience size
- [x] Confirm mobile design

### Desktop
m code dev-theguardian com_sport_2021_may_03_mark-selby-beats-shaun-murphy-win-fourth-world-snooker-championship-crucible-match-report(Laptop with MDPI screen)

### Tablet
m code dev-theguardian com_sport_2021_may_03_mark-selby-beats-shaun-murphy-win-fourth-world-snooker-championship-crucible-match-report(iPad) (2)

### Mobile 375
m code dev-theguardian com_sport_2021_may_03_mark-selby-beats-shaun-murphy-win-fourth-world-snooker-championship-crucible-match-report

#### Horizontal:
m code dev-theguardian com_sport_2021_may_03_mark-selby-beats-shaun-murphy-win-fourth-world-snooker-championship-crucible-match-report

#### Vertical:
m code dev-theguardian com_technology_2015_may_14_google-right-to-be-forgotten-academics-letter

## Tested

- [x] Locally
- [x] On CODE (optional)

PR #3042, opened 2021-05-26
Identity | Sign In Gate | Switch off fake social test

coldlink:
## What does this change?
Removes the AB test for the sign-in-gate fake social button, which was added here: #2945

We're keeping the designs, in case we want to roll out this design after the results have been analysed.


liywjl (approved):
LGTM

PR #3043, opened 2021-05-26
Bump @guardian/braze-components to 2.1.0

tjmw:
## What does this change?

Upgrades to a version of `@guardian/braze-components` which contains a more robust approach to fetching data from the cache. If a message in the cache isn’t valid, we’ll skip over it (see guardian/braze-components#155).

### Before

Invalid Braze messages would clog the cache. If a message at the top of the queue was invalid, it wouldn't get rendered and would prevent messages after it from showing until the cache entry expires.

### After

And invalid message at the top of the cache will be skipped over.

## Why?

This new approach is more robust.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
         "@guardian/ab-react": "^2.0.1",
         "@guardian/atoms-rendering": "^14.0.0",
         "@guardian/automat-client": "^0.2.17",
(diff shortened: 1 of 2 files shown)


liywjl (approved):
LGTM 👍

PR #3044, opened 2021-05-26
Enable picnic stories ads on AMP pages

ealitten:
## What does this change?

Add 320x480 ad size to the AMP Ad component, which allows [picnic stories](https://www.picnic-media.com/solutions/#stories) to be shown. The component falls back to MPU size gracefully as seen below (due to data-multi-size-validation="false" flag).

### After

https://user-images.githubusercontent.com/19875457/119652765-6c49bc00-be1e-11eb-8b7e-daa4e75402c2.mov

## Why?

More ad revenue

### DIFF
--- a/src/amp/components/Ad.tsx
+++ b/src/amp/components/Ad.tsx
@@ -4,6 +4,7 @@ import { adJson, stringify } from '@root/src/amp/lib/ad-json';
 
 // Largest size first
 const inlineSizes = [
+	{ width: 320, height: 480 }, // Picnic story
 	{ width: 300, height: 250 }, // MPU
 	{ width: 250, height: 250 }, // Square
 ];
@@ -141,6 +142,7 @@ export const Ad = ({
 						width={width}
 						height={height}
 						data-multi-size={multiSizes}
+						data-multi-size-validation="false"
 						data-npa-on-unknown-consent={true}
 						data-loading-strategy="prefer-viewability-over-views"
 						layout="fixed"


liywjl (approved):

src/amp/components/Ad.tsx
```diff
@@ -4,6 +4,7 @@ import { adJson, stringify } from '@root/src/amp/lib/ad-json';
 
 // Largest size first
 const inlineSizes = [
+	{ width: 320, height: 480 }, // Picnic story
 	{ width: 300, height: 250 }, // MPU
```
mxdvl, line 8: Late comment, but I think we could change the order to start with MPU.

ealitten: I have revived this branch since it was reverted, so I can make the change. What's your reasoning behind this?

mxdvl: As we take the first `width` × `height` combination below, I think starting with the safest, most popular size is better. This does mean that we need to start using the `data-override-width` and `data-override-height` attributes, otherwise multi-sized cannot be bigger than the original size.

https://github.com/guardian/dotcom-rendering/blob/b111a393badd668b3ccc32021a7898c880282c07/src/amp/components/Ad.tsx#L100

The documentation is [shockingly poor][1], but maybe that’s the incantation that will solve it?

[1]: https://github.com/ampproject/amphtml/blob/main/extensions/amp-ad-network-doubleclick-impl/multi-size.md

ealitten: Hmm, I had a go with MPU as the first size and using this
```
data-override-width={320}
data-override-height={480}
```
but it has lead to some unexpected results...

Screenshot 2021-06-03 at 12 07 01

PR #3045, opened 2021-05-26
Remove renderToStaticMarkup from timeline block

liywjl:
## What does this change?
Remove `renderToStaticMarkup` from `TimelineBlockComponent`

## How?
We are changing `Expandable` as a component to use `children` instead of `html` prop. This means that we can use normal React components as chidren.

## Why?
We are already using `renderToStaticMarkup` in `document.tsx` in amp. There is no need to duplicate the logic.


shtukas (approved):

gtrufitt (approved):
Cool!

PR #3048, opened 2021-05-27
Identity | Sign In Gate Interactions

coldlink:
## What does this change?
- Abstract selecting the sign in gate into it's own custom hook `useSignInGateSelector`
- Create another hook to determine if a gate should show on the current page `useSignInGateWillShow`
- Move all shared sign in gate types into `src/web/components/SignInGate/types.ts`
  - There were circular dependencies before these were extracting into it's own file
- Move shared sign in gate values into `src/web/components/SignInGate/signInGate.ts`
  - Again there were circular dependencies issues when these were in `src/web/components/SignInGate/SignInGateSelector.tsx` because of the new hooks 

## Why?
- We want other parts of DCR to determine if a sign in gate is showing on the current pageview using the `useSignInGateWillShow` hook. This will allow other components to take action if a gate is or is not visible on the current page.
- ` src/web/components/SignInGate/SignInGateSelector.tsx` now handles only the display of the sign in gate if a gate should be visible, as we now use the `useSignInGateSelector` hook to select a gate


src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -144,6 +144,38 @@ const ShowSignInGate = ({
 	return <></>;
 };
 
+// custom hook can be used by other parts of DCR to determine if the sign in gate will or will not show
+export const useSignInGateWillShow = ({
+	isSignedIn,
+	CAPI,
```
oliverlloyd, line 150: As a rule, we try to avoid passing `CAPI` as a prop because it is 1) enormous and 2) only vaguely typed. It's also easier to reason about what a thing does when it outlines the exact properties it depends on.

That said, it looks like you need to pass it in here because another function depends on it... Hmm, it's hard to know what to do here. It's not part of this PR but this PR highlights the problem

src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -144,6 +144,38 @@ const ShowSignInGate = ({
 	return <></>;
 };
 
+// custom hook can be used by other parts of DCR to determine if the sign in gate will or will not show
+export const useSignInGateWillShow = ({
```
oliverlloyd, line 148: Most custom hooks in DCR are stored in their own files in the /libs folder. There's a kind of vague convention of one file one export plus it maybe helps with discoverability and refactoring later

src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -144,6 +144,38 @@ const ShowSignInGate = ({
...
+	isSignedIn,
+	CAPI,
+}: SignInGateSelectorProps): boolean => {
+	const ab = useAB();
+
+	const test: Runnable | null = ab.firstRunnableTest(tests);
```
oliverlloyd, line 154: Does `tests` need to be passed in?

Sorry, by the way, I know this PR is just a draft but I was passing and thought I'd do a drive by review!

src/web/components/SignInGate/SignInGateSelector.tsx
coldlink: Great, I'll move the hook into it's own file in the web/lib folder

src/web/components/SignInGate/SignInGateSelector.tsx
coldlink: Yup! The `tests` object in this case relates to the currently live sign in gate tests here: https://github.com/guardian/dotcom-rendering/blob/7c613ec787a9708df96799d2a0203f8d6efe52c1/src/web/components/SignInGate/SignInGateSelector.tsx#L65

src/web/lib/useSignInGateSelector.ts
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * @description
+ * A custom hook to make which selects the sign in gate (component and ab test)
+ * to be displayed on the current page
+ * */
+export const useSignInGateSelector = ():
+	| [SignInGateComponent, CurrentSignInGateABTest]
+	| null => {
+	const ab = useAB();
```
QuarpT, line 20: this used to get executed with useEffect with dependency `[ab]` which then updated some state. 

Does `ab` never change?

src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -151,23 +117,7 @@ export const SignInGateSelector = ({
...
 	const [showGate, setShowGate] = useState(true);
-	const [currentTest, setCurrentTest] = useState<CurrentABTest>({
-		name: '',
-		variant: '',
-		id: '',
-	});
-
-	const ab = useAB();
-
-	useEffect(() => {
```
QuarpT, line 162: 👍  I see we don't need this useEffect hook and you've moved this into `useSignInGateSelector`

src/web/lib/useSignInGateWillShow.ts
```diff
@@ -0,0 +1,26 @@
+import { SignInGateSelectorProps } from '@frontend/web/components/SignInGate/types';
+import { useSignInGateSelector } from './useSignInGateSelector';
+
+/**
+ * @description
+ * A custom hook to determine if a sign in gate will show on the current page
+ * @param {Boolean} isSignedIn - Is the user signed in to the guardian
+ * @param {CAPIBrowserType} CAPI - The CAPI object
+ * */
+export const useSignInGateWillShow = ({
```
QuarpT, line 10: are you planning on using this in a subsequent PR?

QuarpT (approved):

PR #3049, opened 2021-05-27
Remove remote header test

tomrf1:
We've been running a test to import the RR header links module from support-dotcom-components.
It has performed well, so this PR removes the test and puts it behind a switch.
The switch doesn't yet exist in frontend, so for now this feature will be disabled.

We can remove the non-remote (aka native) header in a follow-up PR, once it's no longer being used.

For now it renders the native header:
Screen Shot 2021-05-27 at 13 38 47


src/web/components/ReaderRevenueLinks.tsx
```diff
@@ -180,21 +181,6 @@ const ReaderRevenueLinksRemote: React.FC<{
 		setSupportHeader,
 	] = useState<React.FC<SupportHeaderProps> | null>(null);
 
-	const [hasBeenSeen, setNode] = useHasBeenSeen({
```
tomrf1, line 183: INSERT and VIEW event tracking will now be handled within the header module, see https://github.com/guardian/support-dotcom-components/pull/459

PR #3052, opened 2021-05-28
Environment variable for SDC

tomrf1:
[SDC](https://github.com/guardian/support-dotcom-components) serves RR modules, e.g. epic + banner.
When running DCR locally we sometimes need to point it at a specific SDC environment (DEV/CODE/PROD).

If the SDC_URL environment variable is set then DCR uses that.

### DIFF
--- a/src/model/window-guardian.ts
+++ b/src/model/window-guardian.ts
@@ -126,9 +126,8 @@ export const makeGuardianBrowserCAPI = (CAPI: CAPIType): CAPIBrowserType => {
 		return false;
 	};
 
-	const contributionsServiceUrl = process?.env?.SDC_URL
-		? process.env.SDC_URL
-		: CAPI.contributionsServiceUrl;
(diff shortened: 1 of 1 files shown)


src/model/window-guardian.ts
```diff
@@ -126,6 +126,10 @@ export const makeGuardianBrowserCAPI = (CAPI: CAPIType): CAPIBrowserType => {
 		return false;
 	};
 
+	const contributionsServiceUrl = process?.env?.SDC_URL
+		? process.env.SDC_URL
+		: CAPI.contributionsServiceUrl;
+
```
TomPretty, line 132: Could potentially do: 

```ts
proces?.env?.SDC_URL ?? CAPI.contributionsServiceUrl;
```

If we can use nullish coalescing on DCR? Either way - looks good to me!

tomrf1: Nice, that works

PR #3055, opened 2021-06-01
Labs Guardian Lines

liywjl:
## What does this change?
Add special conditions for Guardian Lines in Labs

### Before
Screenshot 2021-06-01 at 14 06 20

### After
Screenshot 2021-06-01 at 14 04 06

### DIFF
--- a/src/web/layouts/StandardLayout.tsx
+++ b/src/web/layouts/StandardLayout.tsx
@@ -12,6 +12,8 @@ import {
 import { from, until } from '@guardian/src-foundations/mq';
 import { Design, Special } from '@guardian/types';
 import type { Format } from '@guardian/types';
+import { Lines } from '@guardian/src-ed-lines';
+import { space } from '@guardian/src-foundations';
 
 import { GuardianLines } from '@root/src/web/components/GuardianLines';
 import { StarRating } from '@root/src/web/components/StarRating/StarRating';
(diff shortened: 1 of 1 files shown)


src/web/layouts/StandardLayout.tsx
```diff
@@ -540,14 +542,26 @@ export const StandardLayout = ({ CAPI, NAV, format, palette }: Props) => {
...
+										<Lines count={1} effect="straight" />
+										<div
+											css={css`
+												height: ${space[2]}px;
+											`}
+										/>
+										<Lines count={1} effect="dotted" />
+									</>
+								) : (
+									<GuardianLines
```
jfsoul, line 556: Does GuardianLines not support Labs now? Also cc: @jlieb10 who I think may have picked up the card related to this.

SiAdcock: I think we introduced a [`color` prop](https://github.com/guardian/source/blob/main/src/editorial/web/components/lines/README.md#color) with Labs in mind. It would be good to know if it's valuable.

jlieb10: @jfsoul Please see updated PR here https://github.com/guardian/dotcom-rendering/pull/3056

GuardianLines would support labs with some tinkering, but we discovered that we don't need GuardianLines in the longer term because, as @SiAdcock is pointing out, 'Lines' does everything that 'GuardianLines' does - so rather than re-write GuardianLines, it would be better to use this as a starting point to start going back to just using 'Lines'

As for the question about actually using the color prop, I am unsure what designs require it.

jfsoul: Sorry I actually got those the wrong way around, I thought `GuardianLines` was from source.

liywjl:
this was just a proof of concept on how to change the Labs lines, will close

PR #3057, opened 2021-06-01
Prevent reader revenue banner clash with sign-in gate

lucymonie:
## What does this change?
Reader revenue banners are appearing on the page over the top of the sign-in gate, and this should not be happening. This prevents the banners from being shown when the sign-in gate is in the page. 

This is a short-term fix to a problem that will eventually need a more holistic solution.

### Why?
The banner (and any other insert) hierarchy used to be managed by the bannerList in frontend. With the shift to almost all articles now being rendered by DCR, we have lost this means of controlling what is shown and when.

### DIFF
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -71,7 +71,7 @@ import { VideoFacebookBlockComponent } from '@root/src/web/components/elements/V
 import { VineBlockComponent } from '@root/src/web/components/elements/VineBlockComponent';
 
 import type { BrazeMessagesInterface } from '@guardian/braze-components/logic';
-import { OphanRecordFunction } from '@root/node_modules/@guardian/ab-core/dist/types';
+import { OphanRecordFunction } from '@guardian/ab-core/dist/types';
 import {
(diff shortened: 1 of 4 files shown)


src/web/components/App.tsx
```diff
@@ -71,7 +71,7 @@ import { VideoFacebookBlockComponent } from '@root/src/web/components/elements/V
 import { VineBlockComponent } from '@root/src/web/components/elements/VineBlockComponent';
 
 import type { BrazeMessagesInterface } from '@guardian/braze-components/logic';
-import { OphanRecordFunction } from '@root/node_modules/@guardian/ab-core/dist/types';
+import { OphanRecordFunction } from '@guardian/ab-core/dist/types';
```
lucymonie, line 74: I'm not sure if this is a necessary change - husky picked this up as an error so I changed this to what seemed correct.

coldlink (approved):
LGTM from Identity's side,👍

i-hardy (approved):
👍 LGTM!

PR #3058, opened 2021-06-02
Pass stage to RR epic

tomrf1:
Currently there's no way for the epic module to know what stage it is (CODE/PROD).
This change passes the stage through and into the epic props.
This will be useful for an upcoming change to the epic to send view events to a new api. We'll have CODE/PROD stacks for that api.


src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx
```diff
@@ -97,7 +98,7 @@ const wrapperMargins = css`
 	margin: 18px 0;
 `;
 
-type Props = {
```
tomrf1, line 100: the term `Props` is tied to react components, whereas this is just a function

liywjl (approved):
LGTM

PR #3059, opened 2021-06-02
Collapse ad slots

arelra:
#### Background

Specific ad slots can be setup in Google Ad Manager (GAM) to be "excluded" for reasons of sensitivity, controversy or commercial agreements.

DCR does not currently collapse excluded ad slots.

#### Change

The class `.ad-slot--collapse` is added to "excluded" ad slots at runtime by the commercial client side code.

This change adds CSS to collapse "excluded" ad slots via the class `.ad-slot--collapse`.

CSS rules have been added at the container level (i.e. `ArticleContainer` and `Section`) to:
- collapse dynamically created ad slots (e.g. `im` and `carrot` slots)
- collapse fixed slots (e.g. `top-above-nav`, `right` etc)

#### Release Dependency

There is a related frontend PR to improve the naming and behaviour of the previous incarnation of `.ad-slot--collapse`:

https://github.com/guardian/frontend/pull/23865

This PR can be released before Frontend as there will be no functional impact.

Once the Frontend PR is released DCR should start collapsing ads.

#### Before

https://www.theguardian.com/commentisfree/2021/may/07/my-kids-wont-even-pick-up-a-book-and-it-is-definitely-all-my-fault

Screenshot 2021-06-09 at 01 05 35

Screenshot 2021-06-09 at 13 12 59

Screenshot 2021-06-09 at 01 06 55

### After

Screenshot 2021-06-09 at 13 18 56

### Tested

- [x] Locally
- [x] On CODE (optional)

### DIFF
--- a/src/web/components/ArticleContainer.tsx
+++ b/src/web/components/ArticleContainer.tsx
@@ -52,6 +52,9 @@ const articleAdStyles = css`
 		@media print {
 			display: none !important;
 		}
+		&.ad-slot--collapse {
+			display: none;
+		}
 	}
 	.ad-slot--mostpop {
 		${from.desktop} {
--- a/src/web/components/Section.tsx
+++ b/src/web/components/Section.tsx
@@ -12,6 +12,12 @@ const padding = css`
 	}
 `;
 
+const adStyles = css`
+	& .ad-slot.ad-slot--collapse {
+		display: none;
+	}
+`;
+
 const sideBorders = (colour: string) => css`
 	${from.tablet} {
 		border-left: 1px solid ${colour};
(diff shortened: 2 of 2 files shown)


src/web/components/ArticleContainer.tsx
```diff
@@ -52,6 +52,9 @@ const articleAdStyles = css`
 		@media print {
 			display: none !important;
 		}
+		&.ad-slot--collapse {
+			display: none;
+		}
```
mxdvl, line 57: So you can use **Sass** inside a css declaration. Fancy.

mxdvl (approved):

mxdvl:
This is very neat, and a much better way of collapsing ad slots. 👏

jfsoul:
Thanks for this! Looks much better

PR #3060, opened 2021-06-02
Carrot ad in DCR

mxdvl:
## What does this change?

Ads style for the carrot ad.

These styles are taking more from the DCR implementation of `RichLink` than the original carrot from frontend, but a mark has been added to find the relevant Sass declarations: `dca5c7dd-dda4-4922-9317-a55a3789fe4c`.

### Before

N/A (No carrot ad)

### After

carrot ad screenshot

## Why?

We are using it for advertising.

### DIFF
--- a/src/web/components/ArticleContainer.tsx
+++ b/src/web/components/ArticleContainer.tsx
@@ -1,4 +1,5 @@
 import { css } from '@emotion/react';
+import { space } from '@guardian/src-foundations';
 import { from, until } from '@guardian/src-foundations/mq';
 
 import { labelStyles } from '@root/src/web/components/AdSlot';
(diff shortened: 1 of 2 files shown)


liywjl (approved):
LGTM

src/web/components/ArticleContainer.tsx
```diff
@@ -17,6 +18,25 @@ const articleContainer = css`
 	z-index: 1;
 `;
 
+const carrotStyles = css`
```
arelra, line 21: Very minor so feel free to ignore.
Would prefer `carrotAdStyles` to be more specific

mxdvl:
```suggestion
const carrotAdStyles = css`
```

PR #3061, opened 2021-06-02
messagePicker type improvements

tomrf1:
The `messagePicker` logic is used for epic and banner slots. There are quite a few instances of `any` in the RR code, which feels a bit dangerous when making changes.

The main change here is to make the `Candidate` take a type parameter. This is for the `meta` field, which is returned by `canShow` and passed into `show`. Previously it had type `any`.
I'm not sure what the braze epic/banner should have for this type so I've left them as `any`.

### DIFF
--- a/src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx
+++ b/src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx
@@ -41,18 +41,12 @@ type PreEpicConfig = {
 	};
 };
 
-type EpicConfig = {
-	meta: TestMeta;
-	module: {
-		url: string;
-		props: EpicProps;
-	};
+export type EpicConfig = PreEpicConfig & {
 	email?: string;
 	hasConsentForArticleCount: boolean;
 };
 
 type EpicProps = {
-	onReminderOpen: () => void;
 	email?: string;
 	submitComponentEvent?: (componentEvent: OphanComponentEvent) => void;
 	openCmp: () => void;
@@ -70,34 +64,11 @@ const checkForErrors = (response: Response) => {
 	return response;
 };
 
-const sendOphanReminderEvent = (componentId: string): void => {
-	const componentEvent = {
-		component: {
-			componentType: 'ACQUISITIONS_OTHER',
-			id: componentId,
-		},
-		action: 'CLICK',
-	};
-
-	window.guardian.ophan.record({ componentEvent });
-};
-
-interface OpenProps {
-	buttonCopyAsString: string;
-}
-
-const sendOphanReminderOpenEvent = ({ buttonCopyAsString }: OpenProps) => {
-	sendOphanReminderEvent('precontribution-reminder-prompt-clicked');
-	sendOphanReminderEvent(
-		`precontribution-reminder-prompt-copy-${buttonCopyAsString}`,
-	);
-};
-
 const wrapperMargins = css`
 	margin: 18px 0;
 `;
 
-type Props = {
+export type CanShowData = {
 	isSignedIn?: boolean;
 	countryCode?: string;
 	contentType: string;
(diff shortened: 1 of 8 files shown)


src/web/lib/messagePicker.ts
```diff
@@ -72,15 +76,20 @@ const timeoutify = (
...
+	candidateConfig: CandidateConfigWithTimeout<T>;
+	canShow: Promise<CanShowResult<T>>;
+}
+
+interface WinningMessage<T> {
+	meta: T;
+	candidate: Candidate<T>;
+}
 
 export const pickMessage = ({
```
tomrf1, line 94: While fixing up `pickMessage` I did a small rewrite to pass through the `Candidate` objects with their results, rather than indexing into the array of candidates as we go.

src/web/lib/messagePicker.ts
```diff
@@ -1,30 +1,34 @@
...
+type ShowMessage<T> = (meta: T) => MaybeFC;
+
+interface ShouldShow<T> {
+	show: true;
+	meta: T;
+}
+interface ShouldNotShow {
+	show: false;
+}
+export type CanShowResult<T> = ShouldShow<T> | ShouldNotShow;
```
tomrf1, line 13: The `meta` field is now mandatory on a successful `CanShowResult`

src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx
```diff
@@ -70,34 +64,11 @@ const checkForErrors = (response: Response) => {
...
-	};
-
-	window.guardian.ophan.record({ componentEvent });
-};
-
-interface OpenProps {
-	buttonCopyAsString: string;
-}
-
-const sendOphanReminderOpenEvent = ({ buttonCopyAsString }: OpenProps) => {
```
tomrf1, line 89: this is unused by the epic

src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx
```diff
@@ -70,34 +64,11 @@ const checkForErrors = (response: Response) => {
...
-	sendOphanReminderEvent(
-		`precontribution-reminder-prompt-copy-${buttonCopyAsString}`,
-	);
-};
-
 const wrapperMargins = css`
 	margin: 18px 0;
 `;
 
-type Props = {
```
tomrf1, line 100: the term `Props` is tied to react components, whereas this is just a function

src/web/lib/messagePicker.ts
```diff
@@ -90,39 +99,43 @@ export const pickMessage = ({
...
-		const winner: Promise<WinningMessage> = results.reduce(
-			async (winningMessageSoFar, canShow, idx) => {
-				const runningIdx = (await winningMessageSoFar).idx;
-				if (runningIdx >= 0) {
-					return winningMessageSoFar;
-				}
+		const winnerResult = results.reduce<
+			Promise<WinningMessage<any> | null>
+		>(async (winningMessageSoFar, { candidateConfig, canShow }) => {
+			const running = await winningMessageSoFar;
```
tjmw, line 112: I wonder if there's a better name here than `running` 🤔  I know this is a hangover from my previous questionable choice of `runningIdx`, which was supposed to be like "running total". It's tricky because really `running` is the awaited/unwrapped `winningMessageSoFar`. I wonder if `winningMessageSoFar` should be `asyncWinningMessageSoFar` and then running could be just `winningMessageSoFar`? What do you reckon?

tjmw (approved):
This is looking really great, nice work! It'd be great to do some integration testing on CODE before this goes live if that sounds alright? We (TX) can follow up in a subsequent PR and try to remove those `any`s from the Braze message config too.

src/web/lib/messagePicker.ts
tomrf1: Good point, and actually, since we don't use `running` later on this can just be:
`if (await winningMessageSoFar)`

tjmw: Ah great, that's a nice solution.

PR #3064, opened 2021-06-02
Ad styles from frontend

mxdvl:
## What does this change?

Imports the fluid ad styles inside `AdSlot.tsx`, rather than injecting them at runtime via the commercial JS bundle.

### Screenshots

fabric

---

fabric video

---

fabric animation

## Why?

The logic was peppered in too many places.

### DIFF
--- a/src/web/components/AdSlot.tsx
+++ b/src/web/components/AdSlot.tsx
@@ -116,8 +116,8 @@ export const carrotAdStyles = css`
 /**
  * For CSS in Frontend, see mark: 9473ae05-a901-4a8d-a51d-1b9c894d6e1f
  */
-export const fluidAdStyles = css`
-	.ad-slot--fluid {
+const fluidAdStyles = css`
+	&.ad-slot--fluid {
 		min-height: 250px;
 		line-height: 10px;
 		padding: 0;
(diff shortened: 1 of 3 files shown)


liywjl (approved):
LGTM

src/web/components/AdSlot.tsx
```diff
@@ -174,7 +214,7 @@ export const AdSlot: React.FC<Props> = ({ position, display }) => {
 								'ad-slot--rendered',
 								'js-sticky-mpu',
 							].join(' ')}
-							css={labelStyles}
+							css={[labelStyles, fluidAdStyles]}
```
arelra, line 217: If we apply `labelStyles` and `fluidStyles` in all cases could we combine them as:

```js
const adStyles = { ...labelStyles, ...fluidStyles }
```

mxdvl: I was thinking about this, hence why I wanted to confirm which ads require 100% width with DAP.

src/web/components/AdSlot.tsx
```diff
@@ -362,7 +379,7 @@ export const AdSlot: React.FC<Props> = ({ position, display }) => {
 							'ad-slot--mpu-banner-ad',
 							'ad-slot--rendered',
 						].join(' ')}
-						css={[labelStyles, fluidAdStyles, adSlotAboveNav]}
+						css={[adSlotAboveNav, adStyles, fluidFullWidthAdStyles]}
```
arelra, line 382: Should `adSlotAboveNav` come after `adStyles` so it overrides it?

mxdvl: Good catch!

arelra (approved):

mxdvl:
This doggo is tripping up Chromatic
dog diff

PR #3068, opened 2021-06-03
Fix Subscribe cta in immersive nav

tomrf1:
Currently nothing happens when clicked. This affects immersive layouts, e.g. https://www.theguardian.com/science/2021/jun/03/the-empty-office-what-we-lose-when-we-work-from-home

I'm not sure why it was using an `onClick` handler


TomPretty (approved):
👍

PR #3070, opened 2021-06-03
Clean interactive atom HTML

nicl:
## What/why

Some atoms have badly formed HTML which breaks their display on DCR. Frontend 'cleans' this HTML so to avoid this, which this
commit replicates.

E.g. for:

https://www.theguardian.com/info/ng-interactive/2021/mar/17/make-sense-of-the-week-with-australia-weekend

### Before
Screenshot 2021-06-03 at 11 32 49

### After
Screenshot 2021-06-03 at 11 27 47


liywjl (approved):
LGTM

OllysCoding (approved):
👍 Glad to see more of these interactives working!

PR #3071, opened 2021-06-03
Upgrade to Cypress 7

gtrufitt:
## What does this change?

- Upgrades to Cypress 7
- Fixes for some [breaking changes with cypress.intercept](https://docs.cypress.io/guides/references/migration-guide#cy-intercept-changes)
- Fixes for some previously unsurfaced errors, handling globally
- Fixes for some real errors

### DIFF
--- a/cypress/integration/parallel-2/braze.spec.js
+++ b/cypress/integration/parallel-2/braze.spec.js
@@ -7,10 +7,11 @@ const handleGuCookieError = () => {
 	cy.on('uncaught:exception', (err, runnable) => {
 		// When we set the `GU_U` cookie this is causing the commercial bundle to try and do
 		// something with the url which is failing in Cypress with a malformed URI error
-		expect(err.message).to.include('URI malformed');
-		// This error is unrelated to the test in question so return  false to prevent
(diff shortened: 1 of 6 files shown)


cypress/integration/parallel-2/braze.spec.js
```diff
@@ -22,7 +23,7 @@ describe('Braze messaging', function () {
 	});
 
 	const url =
-		'https://m.code.dev-theguardian.com/games/2018/aug/23/nier-automata-yoko-taro-interview';
+		'https://theguardian.com/games/2018/aug/23/nier-automata-yoko-taro-interview';
```
liywjl, line 26: whats the reasoning behind this change?

liywjl (approved):
LGTM 👍

cypress/integration/parallel-2/braze.spec.js
gtrufitt: Code is not a stable environment for testing against, and my assumption was this was left in from testing a non-live version of Frontend that was required for the test. As the test passes on PROD, I switched it over to the more realistic environment for the user.

PR #3072, opened 2021-06-03
Update to lab 'paid for by' typography

jlieb10:
### Before
before

### After
after

## Why?
https://trello.com/c/KpNP6kbA/242-labs-paid-for-by-text-to-be-333-and-regular-weight

### DIFF
--- a/src/web/components/Branding.tsx
+++ b/src/web/components/Branding.tsx
@@ -9,8 +9,8 @@ const brandingStyle = css`
 `;
 
 const brandingLabelStyle = css`
-	${textSans.xxsmall({ fontWeight: 'bold' })};
-	color: ${neutral[46]};
+	${textSans.xxsmall()};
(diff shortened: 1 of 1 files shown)


liywjl (approved):
LGTM 👍 

I assume `Branding.tsx` is only used in Labs so we do not need to conditionally change the css here?

jfsoul:
Hey, I was looking at something related to this, so ended up checking this out and noticed the comment:

> I assume Branding.tsx is only used in Labs so we do not need to conditionally change the css here?

This is actually not the case, we do use `Branding.tsx` elsewhere (e.g. [sponsored content that isn't Labs produced](https://www.theguardian.com/global-development/2020/dec/10/rich-countries-leaving-rest-of-the-world-behind-on-covid-vaccines-warns-gates-foundation)). But I double checked with Harry and the change is fine as-is there too!

PR #3074, opened 2021-06-03
Bump web-vitals library with breaking CLS change

gtrufitt:
## What does this change?

Bump the web-vitals library which contains the new [definition of CLS](https://github.com/GoogleChrome/web-vitals/pull/148) described in https://web.dev/evolving-cls/

In Google search tools, we see a reduction in CLS and increase in Good number of pages, we should see this reflected in our data.


nitro-marky (approved):
LGTM

PR #3079, opened 2021-06-04
Slight Caption Refactor + Adding to InteractiveBlockElement

gtrufitt:
## What does this change?

- shouldLimitWidth is not used at the top level so removed it from the model
- Add MainMedia and limited width caption the interactiveBlockElement


shtukas (approved):

liywjl (approved):
👍  LGTM

PR #3080, opened 2021-06-04
Fix storybook build JSDOM error 

OllysCoding:
## What does this change?

Move cleaning of `InteractiveAtomBlockElement` from element rendering to enhancer

## Why?

Inclusion of JSDOM in clientside build would prevent storybook from building


jfsoul (approved):
Nice!

PR #3081, opened 2021-06-04
Load braze sdk for all signed in users.

michaelclapham:
## What does this change?
Requirements for a user to satisfy for the Braze SDK to be loaded (and therefore see Braze components, Banners/Epic).

### Before
Only logged in users who are also supporters would have the Braze SDK loaded.

### After
All logged in users will have the Braze SDK loaded.

## Why?
This is so we can use Braze for marketing to non supporters.


tjmw (approved):
This LGTM! 👍

PR #3084, opened 2021-06-04
Respect interactive nav switch

nicl:
## What does this change?

Respect the `interactiveFullHeaderSwitch` switch, which toggles slim nav on immersive interactives. As part of this I have also:

* removed some unnecessary code in the immersive interactive layout file
* introduced some basic global interactive styles that are present in Frontend and expected by some interactives

e.g. for https://www.theguardian.com/music/ng-interactive/2021/apr/12/the-months-best-albums...

### Before
Screenshot 2021-06-04 at 17 18 58

### After
Screenshot 2021-06-04 at 17 19 10

## Why?

Parity with Frontend.

### DIFF
--- a/src/web/components/Section.tsx
+++ b/src/web/components/Section.tsx
@@ -2,27 +2,7 @@ import { css } from '@emotion/react';
 
 import { border } from '@guardian/src-foundations/palette';
 import { from } from '@guardian/src-foundations/mq';
-
-export const center = css`
-	position: relative;
-	margin: auto;
-
-	${from.tablet} {
-		max-width: 740px;
-	}
-
-	${from.desktop} {
-		max-width: 980px;
-	}
-
-	${from.leftCol} {
-		max-width: 1140px;
-	}
-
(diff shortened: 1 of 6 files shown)


shtukas (approved):

src/web/components/Section.tsx
```diff
@@ -3,7 +3,7 @@ import { css } from '@emotion/react';
 import { border } from '@guardian/src-foundations/palette';
 import { from } from '@guardian/src-foundations/mq';
 
-const center = css`
+export const center = css`
```
OllysCoding, line 6: Is there somewhere for using shared CSS, rather than exporting it from another component?

liywjl (approved):
LGTM

src/web/components/Section.tsx
nicl: I don't think we have anywhere, but one of the `lib` directories is probably a good choice here - I'll move it.

PR #3086, opened 2021-06-07
Pull through branding for cards

jfsoul:
## What does this change?
This passes through the optional `branding` data we get from CAPI trails for use in cards (i.e. for onward journeys). Some logic around this had been implemented in #2751 but it wasn't wired up from the CAPI data, just in storybook stories. In order to do so, I had to pass through the `Branding` type, which is richer than the one previously used and so I've utilised the properties that is provides (image dimensions, variable label text, inverted logo for dark backgrounds).

### Before
image

### After
image

## Why?
We noticed that labs articles onward containers (which usually have other labs content in them) were not pulling through the appropriate logo/branding component for the footer of the cards. It's obviously important to make sure the branding is correct on sponsored/paid-for content! It also matches frontend.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -112,6 +112,7 @@ type Palette = {
 		sectionTitle: Colour;
 		avatar: Colour;
 		card: Colour;
+		cardInvertLogo: boolean;
 		headline: Colour;
 		headlineByline: Colour;
 		bullet: Colour;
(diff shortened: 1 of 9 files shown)


index.d.ts
```diff
@@ -112,6 +112,7 @@ type Palette = {
 		sectionTitle: Colour;
 		avatar: Colour;
 		card: Colour;
+		cardInvertLogo: boolean;
```
jfsoul, line 115: I don't know that this is the right approach, so looking for ideas on how else to handle. The other option is to have some sort of method of determining if the specific background colour is dark enough to  warrant inverting the logo, and put that in the place it's used.

nitro-marky (approved):

OllysCoding (approved):
LGTM!

PR #3087, opened 2021-06-08
Identity | Sign In Gate | Australia Mandatory Test

coldlink:
## What does this change?
Adds a new mandatory gate test similar too #2661. This gate removes the dismiss button.

Additional requirements are:
- Australia only
- On CMP, `personalisedAdvertising` is set to `true` for the `aus` consent state,

## Other changes
- Refactor the sign in gate `canShow` method to be asynchronous
  - This is because the calls to get the country, and the CMP consents are async
-  Refactor the sign in gate selector and related hooks to use `useOnce` hook so that they're only ever fired once

## Screenshot

localhost_3030_Article_url=https___www theguardian com_money_2017_mar_10_ministers-to-criminalise-use-of-ticket-tout-harvesting-software


src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -116,40 +116,42 @@ export const SignInGateSelector = ({
 	isSignedIn,
 	CAPI,
 }: SignInGateSelectorProps) => {
-	const [showGate, setShowGate] = useState(true);
+	const [isGateDismissed, setIsGateDismissed] = useState(false);
+	const [canShowGate, setCanShowGate] = useState(false);
 	const gateSelector = useSignInGateSelector();
 
 	useEffect(() => {
```
oliverlloyd, line 123: Is the use case here that this event only fires once? If so, maybe DCR's custom `useOnce` hook would be useful?

src/web/components/SignInGate/displayRule.ts
```diff
@@ -71,24 +74,79 @@ export const isValidTag = (CAPI: CAPIBrowserType): boolean => {
 export const isPaidContent = (CAPI: CAPIBrowserType): boolean =>
 	CAPI.pageType.isPaidContent;
 
+// hide the sign in gate on internal tools preview
```
oliverlloyd, line 77: 👍 useful comment!

src/web/components/SignInGate/SignInGateSelector.tsx
coldlink: ah interesting! i'll take a look at the `useOnce` hook

oliverlloyd: The thing to know about `useOnce` is it works slightly differently to `useEffect`. In useEffect, if any of the dependencies _change_ then this triggers the code to re-execute. In useOnce the code is _only_ executed once all the dependencies are truthy.

I think in this case you can probably swap it in directly and it should work. You'll also be able to delete the `if (isGateDismissed) {` check as this is implicit

coldlink: Great, I've changed it to use a `useOnce` hook

src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -116,40 +116,42 @@ export const SignInGateSelector = ({
...
-	}, [showGate]);
+	}, [isGateDismissed]);
 
 	if (!gateSelector) {
 		return null;
 	}
 
 	const [gateVariant, currentTest] = gateSelector;
+	// eslint-disable-next-line @typescript-eslint/no-floating-promises
+	gateVariant?.canShow(CAPI, !!isSignedIn, currentTest).then(setCanShowGate);
```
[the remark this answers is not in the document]
coldlink: also using a `useOnce` hook for this now too

src/web/components/SignInGate/SignInGate.stories.tsx
```diff
@@ -25,6 +25,54 @@ export const mainStandalone = () => {
...
+				guUrl="https://theguardian.com"
+				signInUrl="https://profile.theguardian.com/"
+				dismissGate={() => {}}
+				ophanComponentId="test"
+				isComment={true}
+			/>
+		</Section>
+	);
+};
+mainStandaloneComment.story = { name: 'main_standalone_comment' };
```
oliverlloyd, line 41: Not a huge deal but there's no need to use snake case for the `name` here, this can just be any display string. There's a sort of convention to have name like, 'with default` or 'when foobar is true' as then it reads nicely with the main Story name

Eg. 'My component - with defaults'

src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -116,40 +117,64 @@ export const SignInGateSelector = ({
...
+	const [canShowGate, setCanShowGate] = useState(false);
 	const gateSelector = useSignInGateSelector();
 
-	useEffect(() => {
+	useOnce(() => {
 		// this hook will fire when the sign in gate is dismissed
 		// which will happen when the showGate state is set to false
 		// this only happens within the dismissGate method
-		if (showGate === false) {
+		if (isGateDismissed) {
```
oliverlloyd, line 136: You can delete this line as this code will only fire once `isGateDismissed` is truthy

src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -116,40 +117,64 @@ export const SignInGateSelector = ({
...
-	}, [showGate]);
+	}, [isGateDismissed]);
+
+	useOnce(() => {
+		const [gateSelectorVariant, gateSelectorTest] = gateSelector;
+		if (gateSelectorVariant && gateSelectorTest) {
+			setGateVariant(gateSelectorVariant);
+			setCurrentTest(gateSelectorTest);
+		}
+	}, [gateSelector]);
```
oliverlloyd, line 149: Is `gateSelector` an object? Or rather, is it possible for it to be undefined / falsey? If it's always true then there's no need to have it as a dependency

src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -116,40 +117,64 @@ export const SignInGateSelector = ({
...
+
+	useOnce(() => {
+		if (gateVariant && currentTest) {
```
oliverlloyd, line 152: Ditto above, there's no need to have this guard

src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -116,40 +117,64 @@ export const SignInGateSelector = ({
...
+			// eslint-disable-next-line @typescript-eslint/no-floating-promises
+			gateVariant
+				?.canShow(CAPI, !!isSignedIn, currentTest)
+				.then(setCanShowGate);
+		}
+	}, [currentTest, gateVariant]);
 
-	if (!gateSelector) {
+	if (!currentTest || !gateVariant) {
```
oliverlloyd, line 160: Is it worth assigning these to named vars, like `currentTestUndefined` and then having the negative OR ternary?

It's more code but even though it should be a very simple piece of logic my brain trips over when it sees negation and an OR ternary together - I've no idea why but it confuses me every time!

src/web/components/SignInGate/gateDesigns/SignInGateMain.tsx
```diff
@@ -69,20 +70,21 @@ export const SignInGateMain = ({
...
-					priority="subdued"
-					size="small"
-					onClick={() => {
-						dismissGate();
-						trackLink(ophanComponentId, 'not-now', abTest);
-					}}
-				>
-					I’ll do it later
-				</LinkButton>
+				{!isMandatory && (
```
oliverlloyd, line 73: This reads really nice!

oliverlloyd (approved):
I'm approving here but I should caveat this by saying I'm not really able to follow the logic around how the banner should or should not show. In particular where it's moving from synchronous to asynchronous.

Obviously these changes are operating in the constraints of the existing structure though and these changes seem like the right approach given those constraints but, as discussed offline, it'd be great to rethink the logic around how we display this content

PR #3088, opened 2021-06-08
ELB Scheme set to Internal

nitro-marky:
## What does this change?

Add `Scheme: internal` to the Cloudformation yaml. Previously this had not been set and was open to the outside world.


buck06191 (approved):
LGTM

PR #3089, opened 2021-06-08
Better localStorage check

tjmw:
## What does this change?

Changes the way we test for localStorage availability before attempting to load the Braze SDK (follow on from #3023).

### Before

`'localStorage' in window;`

### After

`storage.local.isAvailable()` from `@guardian/libs`.

## Why?

We’re still seeing some errors in Sentry where localStorage isn’t available, even with the `localStorage in window` test. Use the test from @guardian/libs instead as it’s more exhaustive.


michaelclapham (approved):
Logic in the library looks sound!

shtukas (approved):

PR #3096, opened 2021-06-09
Revert "Revert "Identity | Sign In Gate | Australia Mandatory Test""

coldlink:
Fixes Issue with Australia mandatory gate by using `useEffect` instead of `useOnce` hook. See #3087 for the original PR.

### DIFF
--- a/src/web/components/SignInGate/SignInGateSelector.tsx
+++ b/src/web/components/SignInGate/SignInGateSelector.tsx
@@ -148,14 +148,14 @@ export const SignInGateSelector = ({
 		}
 	}, [gateSelector]);
 
-	useOnce(() => {
+	useEffect(() => {
 		if (gateVariant && currentTest) {
 			// eslint-disable-next-line @typescript-eslint/no-floating-promises
 			gateVariant
 				?.canShow(CAPI, !!isSignedIn, currentTest)
 				.then(setCanShowGate);
 		}
-	}, [currentTest, gateVariant]);
+	}, [currentTest, gateVariant, CAPI, isSignedIn]);
 
 	if (!currentTest || !gateVariant) {
 		return null;
--- a/src/web/components/SignInGate/SignInGateSelector.tsx
+++ b/src/web/components/SignInGate/SignInGateSelector.tsx
@@ -8,6 +8,7 @@ import {
 import { getCookie } from '@frontend/web/browser/cookie';
 import { useSignInGateSelector } from '@frontend/web/lib/useSignInGateSelector';
 
+import { useOnce } from '@frontend/web/lib/useOnce';
 import {
 	ComponentEventParams,
 	submitViewEventTracking,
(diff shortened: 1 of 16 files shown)


oliverlloyd:
If we were incorrectly showing the gate for signed in users when we used `useOnce` and if we are able to fix that by using `useEffect`, then does that mean the first render is calling `gateVariant.canShow` and this is incorrectly returning `true`? If so, it looks like `gateVariant.canShow` is only returning `false` on subsequent renders, which suggests we may have some sort of race condition here.

I guess in practice this might be okay but perhaps there's a risk of the gate being briefly rendered and then taken away? Although it could be there are additional guards later on that prevent this.

coldlink:
> If we were incorrectly showing the gate for signed in users when we used `useOnce` and if we are able to fix that by using `useEffect`, then does that mean the first render is calling `gateVariant.canShow` and this is incorrectly returning `true`? If so, it looks like `gateVariant.canShow` is only returning `false` on subsequent renders, which suggests we may have some sort of race condition here.
> 
> I guess in practice this might be okay but perhaps there's a risk of the gate being briefly rendered and then taken away? Although it could be there are additional guards later on that prevent this.

it's actually `false` when using `useEffect` and then goes to true when it should be visible:

vcxsrv_Y4CCotOuD5

coldlink:
vcxsrv_FMQOVO2AGX

actually this is interesting, once logged in, it briefly sets canShow to true before going back to false 🤔

coldlink:
Definitely a race condition going on

oliverlloyd:
> Definitely a race condition going on, but this is in line with what it was doing before.

Yeah, I agree this is an existing issue; I don't think this PR is the right place to fix it!

In general though, my feeling is this could be fixed, in part, with composition. Often with these kinds of state changes you can solve it by breaking out some rendering logic into a different component and then that component using its own rendering lifecycle and state to prevent things happening until some other, higher state, is true. But I also think there's something else at play here as well, which is giving the race condition.

src/web/lib/useSignInGateWillShow.ts
```diff
@@ -10,17 +16,30 @@ import { useSignInGateSelector } from './useSignInGateSelector';
...
-		return false;
-	}
+	useOnce(() => {
+		const [gateSelectorVariant, gateSelectorTest] = gateSelector;
+		setGateVariant(gateSelectorVariant);
+		setCurrentTest(gateSelectorTest);
+	}, [gateSelector]);
 
-	const [gateVariant, currentTest] = gateSelector;
+	useEffect(() => {
```
coldlink, line 35: Changed from useOnce to useEffect

src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -116,40 +117,64 @@ export const SignInGateSelector = ({
...
+
+	useOnce(() => {
+		const [gateSelectorVariant, gateSelectorTest] = gateSelector;
+		if (gateSelectorVariant && gateSelectorTest) {
+			setGateVariant(gateSelectorVariant);
+			setCurrentTest(gateSelectorTest);
+		}
+	}, [gateSelector]);
+
+	useEffect(() => {
```
coldlink, line 151: Changed from useOnce to useEffect

oliverlloyd (approved):
As discussed, this highlights a state race condition that is not ideal but was already existing so shouldn't block this piece of work

PR #3097, opened 2021-06-09
Replace GuardianLines (DCR-specific) with Lines (from Source)

jlieb10:
## What does this change?
This repo used 'GuardianLines' which was an implementation of the standard 'Lines' component, but with the ability to change the Lines color. The standard Lines component was updated so the use of "GuardianLines" became outdated.

NB: The MostViewedRightWrapper also was modified as it no longer needed the "palette" attribute, as this was exclusively being used by GuardianLines as a child component.

### Before
`<GuardianLines />`

### After
`<Lines />`(there should be no visual difference)

## Why?
health: deduplication efforts; reliance on distributed components over repo-specific components
https://trello.com/c/CcUKrWbV/263-remove-all-guardianlines-component-references-from-dcr-replace-with-lines-from-source


jfsoul (approved):
Nice!

oliverlloyd (approved):
❤️ Sooo satisfying!

liywjl (approved):
Awesome work!!

PR #3101, opened 2021-06-10
Speed up storybook: Upgrade webpack to webpack5

OllysCoding:
## What does this change?

Upgrades storybook webpack building to webpack v5. 

Worth noting that in the future storybook may support frameworks like `vite`. This may be useful for dev storybook builds which only need to run on modern browsers (that support esmodules). Read more about this [here](https://storybook.js.org/blog/storybook-for-webpack-5/) 

## Why?

This should provide both performance for build speed and help bring more parity to builds as the rest of the project uses webpack 5.

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,6 +137,8 @@
         "@loadable/babel-plugin": "^5.13.2",
         "@loadable/webpack-plugin": "^5.14.2",
         "@storybook/addon-essentials": "^6.2.8",
+        "@storybook/builder-webpack5": "^6.2.9",
+        "@storybook/manager-webpack5": "^6.3.0-alpha.37",
         "@storybook/react": "^6.2.9",
         "@testing-library/dom": "^7.27.1",
         "@testing-library/react": "^11.2.6",
--- a/scripts/webpack/browser.js
+++ b/scripts/webpack/browser.js
@@ -32,17 +32,6 @@ const scriptPath = (dcrPackage) =>
 			'webpack-hot-middleware/client?name=browser&overlayWarnings=true',
 	].filter(Boolean);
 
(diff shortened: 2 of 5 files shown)


OllysCoding:
Looks like some stories don't render - investigating now

package.json
```diff
@@ -137,6 +137,8 @@
         "@loadable/babel-plugin": "^5.13.2",
         "@loadable/webpack-plugin": "^5.14.2",
         "@storybook/addon-essentials": "^6.2.8",
+        "@storybook/builder-webpack5": "^6.2.9",
+        "@storybook/manager-webpack5": "^6.3.0-alpha.37",
```
gtrufitt, line 141: Do we need to use the alpha release here?

gtrufitt (approved):

package.json
OllysCoding: I don't think we need the manager-webpack5 at all, so I've removed it since there isn't a non-alpha version from what I can see.

scripts/webpack/browser.js
```diff
@@ -127,3 +116,14 @@ module.exports = ({ isLegacyJS }) => ({
...
+module.exports.babelExclude = {
+	and: [/node_modules/],
+	not: [
+		// Include all @guardian modules, except automat-modules
+		/@guardian\/(?!(automat-modules))/,
+
+		// Include the dynamic-import-polyfill
+		/dynamic-import-polyfill/,
+	],
+};
```
OllysCoding, line 129: I had to move this to the bottom of the file, as `module.exports = ...` was overwriting the property and you couldn't import `babelExclude` in other files.

liywjl (approved):

PR #3102, opened 2021-06-10
Use immersive image roles for immersive stories

jfsoul:
## What does this change?
Adds the `immersive` role to main media images in the immersive layout stories, so that they more accurately reflect actual immersive image quality.

## Why?
Chromatic was throwing up some diffs on the images for these stories, but we shouldn't dig too much into questions of quality on the stories unless they actually reflect real life!


OllysCoding (approved):
LGTM!

PR #3104, opened 2021-06-10
Refactor logo logic out of decidePalette

jfsoul:
## What does this change?
Following feedback on #3086 this refactors (out) the logic that was originally included in `decidePalette` for choosing whether to use a logo for a dark background. While this is palette-related, it's not really part of the palette itself - so now lives alone in a new `decideLogo`. 

## Why?
The original implementation felt wrong, this is better.

### DIFF
--- a/src/web/lib/decideLogo.ts
+++ b/src/web/lib/decideLogo.ts
@@ -1,0 +1,33 @@
+import { Design, Special, Pillar } from '@guardian/types';
+
+const shouldUseLogoForDarkBackground = (format: Format): boolean => {
+	if (format.theme === Special.SpecialReport) return true;
+	switch (format.design) {
+		case Design.Media:
+			return true;
+		case Design.LiveBlog:
+			switch (format.theme) {
+				case Special.Labs:
+					return false;
+				case Pillar.News:
+				case Pillar.Sport:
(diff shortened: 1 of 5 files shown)


src/web/lib/decideLogo.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+export const decideLogo = (
+	format: Format,
+	branding: Branding,
+): BrandingLogo => {
+	return shouldUseLogoForDarkBackground(format) &&
+		branding.logoForDarkBackground
+		? branding.logoForDarkBackground
+		: branding.logo;
```
OllysCoding, line 32: Could this be shortened to `... && (branding.logoForDarkBackground || branding.logo)`?

OllysCoding (approved):
LGTM

src/web/lib/decideLogo.ts
jfsoul: I think it could be shortened, but I think the braces would be like `(... && ...) || ...` instead - which probably indicates why we might prefer the ternary still because it is harder to get wrong and easier to grep IMO.

gtrufitt (commented):
tenor-164188447

PR #3105, opened 2021-06-11
Interactive content block

liywjl:
## What does this change?
- find if we have an interactive content block (check for `InteractiveBlockElement` that has script URL `2019/03/20/boot.js`)
- replace block with new `InteractiveContentBlockElement`
- make sure to add all `subheadingLinks` to `InteractiveContentBlockElement` to be able to render index block
- add IDs for each H2 on each `SubheadingBlockElement` block to allow anchor HTML links
- add sticky nav implementation of Content interactive
- Spoof JSDOM for storybook build to support using enhancers in storybook

### Examples:
https://www.theguardian.com/technology/2019/dec/17/best-smartphone-2019-iphone-oneplus-samsung-and-huawei-compared-and-ranked
https://www.theguardian.com/info/2015/may/18/guardian-australia-meet-the-team

### Before
Screenshot 2021-06-11 at 10 55 26

### After
Screenshot 2021-06-17 at 16 33 09
Screenshot 2021-06-17 at 16 33 18
Screenshot 2021-06-17 at 16 33 25
Screenshot 2021-06-17 at 16 33 47

## Why?
This interactive block is used on several articles and the remote script creates an implicit dependency between the block and the rest of the content on the page. We instead SSR this block and create a more direct dependency, this should also boost performance as there is less JS running on the main thread


src/model/enhanceCAPI.ts
```diff
@@ -0,0 +1,84 @@
...
+import { enhanceDots } from '@root/src/model/add-dots';
+import { setIsDev } from '@root/src/model/set-is-dev';
+import { enhanceImages } from '@root/src/model/enhance-images';
+import { enhanceInteractiveContentElements } from '@root/src/model/enhance-interactive-content-elements';
+import { enhanceNumberedLists } from '@root/src/model/enhance-numbered-lists';
+import { enhanceInteractiveAtomElements } from '@root/src/model/enhance-interactive-atom-elements';
+import { enhanceBlockquotes } from '@root/src/model/enhance-blockquotes';
+import { enhanceEmbeds } from '@root/src/model/enhance-embeds';
+
+class CAPIEnhancer {
```
liywjl, line 12: `CAPIEnhancer` is large enought if feels ok to split it out of `src/web/server/render.ts`

src/model/enhance-interactive-content-elements.ts
```diff
@@ -1,16 +1,17 @@
...
-		) {
-			return true;
-		}
-		return false;
-	});
+	const hasInteractiveContentBlockElement = elements.some((element) =>
+		isInteractiveContentBlockElement(element),
+	);
+
+	console.log(`hasInteractiveContentBlockElement: ${  hasInteractiveContentBlockElement}`)
```
nitro-marky: I'm sure someone will mention removing this 😄

nitro-marky (approved):
Looks good

nitro-marky (approved):

src/model/enhance-interactive-content-elements.ts
```diff
@@ -0,0 +1,81 @@
+import { stripHTML } from '@root/src/model/strip-html';
+
+const isInteractiveContentBlockElement = (element: CAPIElement): boolean =>
+	element._type ===
+		'model.dotcomrendering.pageElements.InteractiveBlockElement' &&
+	element?.scriptUrl === 'https://uploads.guim.co.uk/2019/03/20/boot.js';
```
jfsoul, line 6: I think the script element can also be either

- `https://interactive.guim.co.uk/page-enhancers/nav/boot.js` or
- `https://uploads.guim.co.uk/2019/12/11/boot.js` or
- `https://uploads.guim.co.uk/2019/03/20/boot.js`

I'll forward email from CP as an FYI

OllysCoding: 👍 Thank you! I almost created a `scriptUrls` array when creating the function and then was like, nah, there's probably just this one 🤣

OllysCoding:
Currently investigating fixing chromatic / storybook build. This is to do with using the change JSDOM in the enhancer

.storybook/main.js
```diff
@@ -11,6 +11,13 @@ module.exports = {
 		const rules = config.module.rules;
 		const { extensions } = config.resolve;
 
+		// Mock JSDOM for storybook - it relies on native node.js packages
+		// Allows us to use enhancers in stories for better testing of compoenents & full articles
+		config.resolve.alias.jsdom$ = path.resolve(
+			__dirname,
+			'./mocks/jsdom.js',
```
jfsoul, line 18: might just be me, but the tab size looks huge in this file

src/lib/content.d.ts
```diff
@@ -293,6 +293,12 @@ interface NumberedTitleBlockElement {
 	format: CAPIFormat;
 }
 
+interface InteractiveContentBlockElement {
+	_type: 'model.dotcomrendering.pageElements.InteractiveContentBlockElement';
```
jfsoul, line 297: Real nitpick but I would think of it as a `Contents` block element rather than `Content`. Maybe there's a better name that isn't so easily confused with the word "content" though.

src/model/enhance-interactive-content-elements.ts
```diff
@@ -0,0 +1,87 @@
...
+const scriptUrls = [
+	'https://interactive.guim.co.uk/page-enhancers/nav/boot.js',
+	'https://uploads.guim.co.uk/2019/03/20/boot.js',
+	'https://uploads.guim.co.uk/2019/12/11/boot.js',
+];
+
+const isInteractiveContentBlockElement = (element: CAPIElement): boolean =>
+	element._type ===
+		'model.dotcomrendering.pageElements.InteractiveBlockElement' &&
+	scriptUrls.indexOf(element?.scriptUrl || '') !== -1;
```
jfsoul, line 12: I'm not sure we have any interactive elements without script URLs but this line is a little risky because it'd identify any interactive block without a script URL as a contents block element. Maybe safer to do `element.scriptUrl && scriptUrls.indexOf(...`

src/web/components/elements/InteractiveContentBlockElement.tsx
```diff
@@ -0,0 +1,272 @@
...
+		setStickyNavCurrentHeader,
+	] = useState<null | EnhancedSubheadingType>(null);
+	useEffect(() => {
+		const onScroll = libDebounce(() => {
+			const firstElementTop = enhancedSubheadings[
+				enhancedSubheadings.length - 1
+			]?.ref?.getBoundingClientRect().top;
+
+			const articleBodyBottom = document
+				.getElementById('article-body')
```
jfsoul, line 185: Is there another way to have a reference to this element, or even another way to calculate the position at which we should hide the sticky nav? Just feels like this id could be used/abused by other interactives creating implicit contracts.

gtrufitt: We should defo be using intersection observer

src/model/enhance-interactive-content-elements.ts
OllysCoding: `scriptUrls.indexOf('')` would still return -1, but I think your method is still a cleaner solution so I'll get that changed

.storybook/main.js
OllysCoding: Yea it definitely looks massive but renders normally in vscode and has the same tabsize as all the other files (4 spaces) so very confusing...

src/model/enhance-interactive-content-elements.ts
jfsoul: I executed that in my console and it gave me zero - would be interested to know if you've seen different!

OllysCoding: image
Very odd! Perhaps an inconsistency between versions of node / browsers?

jfsoul: Sorry, my bad, I was reading it as `scriptUrl.indexOf('')` as in `String.indexOf()`!

rcrphillips:
[Contents section on DCR](https://trello.com/c/tt4MPIzf/248-contents-section-on-dcr)

jfsoul (approved):
👍

PR #3108, opened 2021-06-14
Health: Resolve console warnings

OllysCoding:
## What does this change?

- Update URL on index.html that was broken
- Change onClick to onChange as `checked` prop requires this.
- Add `key` to figure using element `id` as this is rendered in a list
- Remove `key` prop from Hide component - this is built into react and will always be undefined

## Why?

Improved best practices + cleaner dev experience + easier to find any bugs / errors in rendering when working on the project

### DIFF
--- a/scripts/frontend/landing/index.html
+++ b/scripts/frontend/landing/index.html
@@ -150,7 +150,7 @@
                 {
                     name: 'GenericAtomBlockElement', // AudioAtom
                     article:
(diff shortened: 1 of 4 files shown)


shtukas (approved):

liywjl (approved):
Nice work 👍

gtrufitt:
tenor-145998898

PR #3114, opened 2021-06-15
Fix for documentary interactives

OllysCoding:
## What does this change?

The ID property for elements was incorrect, preventing hydration of any elements on the document, this changes it from `element.id` to `element.elementId` 

### Before

image

### After

image

## Why?

Hydration was broken


shtukas (approved):

src/web/layouts/InteractiveImmersiveLayout.tsx
```diff
@@ -61,7 +61,10 @@ const Renderer: React.FC<{
 		});
 
 		return ok ? (
-			<figure id={'id' in element ? element.id : undefined} key={index}>
+			<figure
+				id={'elementId' in element ? element.elementId : undefined}
```
buck06191, line 65: Can we type the object this is coming from at all to try avoid these things from happening in general, e.g. if we had `element` and all its fields typed in an interface, we'd have caught this sooner

nicl (approved):
Nice find :)

src/web/layouts/InteractiveImmersiveLayout.tsx
nicl: Yeah, we could switch on type. I think I did it this way originally as it would be quite verbose to handle it otherwise but it would be safer.

PR #3120, opened 2021-06-17
Fix social icons on some interactives

OllysCoding:
## What does this change?

Adds some global CSS required by some interacts to show social icons.
This also shouldn't be something we 'support' going forward for new interactives - but it is useful in supporting older ones.

### Before

image

### After

image

## Why?

Visual parity


shtukas (approved):

nicl (approved):

nicl (commented):
Would it be possible to convert the units from rem to px, just to make it consistent with what we do elsewhere?

liywjl (approved):

PR #3124, opened 2021-06-21
Snyk runthrough

nicl:
I ran the new `make snyk` and committed the results. The file is now near empty because:

1) I moved eslint stuff into dev deps (so no longer snyked)
2) I chose 'notify when upgrade when available'  rather than ignore for things

cc @jlieb10 in case this isn't sensible though!


jlieb10:
@nicl Nice!

I'm not strongly opinionated here but the benefit of ignoring in the `.snyk` policy is that, I believe, we can then re-instate the Github action. If we are running `make snyk` every few weeks and select to ignore and re-write the policy at that point, then it will alert us to any upgrades in any case. I agree that 'notify' sounds more pro-active than 'ignore' but I'm not sure that in practice, with the rota, it will make much of a difference for us.

nicl:
(Closing as out of date.)

PR #3126, opened 2021-06-22
Set subnav background colour for immersive interactives

nicl:
## What does this change?

Some interactives override background colours for the body area which can render the subnav unreadable without this change.

### Before
e.g. for https://www.theguardian.com/us-news/ng-interactive/2021/jun/17/tree-rings-america-megadrought-visual

Screenshot 2021-06-22 at 10 34 27

### After
Screenshot 2021-06-22 at 10 34 18

## Why?

Parity.


shtukas (approved):

OllysCoding (approved):
LGTM!

PR #3127, opened 2021-06-22
Only provide legacy styles for legacy interactives

OllysCoding:
## What does this change?

This PR depends on a PR from frontend: https://github.com/guardian/frontend/pull/23923, it should not be merged until after the frontend PR is merged.

Updates interactives so the legacy 'global' styles are only applied if the interactive came out _before_ the switchover date.

## Why?

We don't really want to be supporting these legacy global styles going forward on new interactives.

### DIFF
--- a/src/web/layouts/InteractiveImmersiveLayout.tsx
+++ b/src/web/layouts/InteractiveImmersiveLayout.tsx
@@ -25,7 +25,7 @@ import { getCurrentPillar } from '@root/src/web/lib/layoutHelpers';
 import { renderElement } from '../lib/renderElement';
 import { Header } from '../components/Header';
 import { HeaderAdSlot } from '../components/HeaderAdSlot';
-import { interactiveGlobalStyles } from './lib/interactiveLegacyStyling';
+import { interactiveLegacyGlobalStyles } from './lib/interactiveLegacyStyling';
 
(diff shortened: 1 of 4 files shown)


src/web/layouts/InteractiveImmersiveLayout.tsx
```diff
@@ -196,7 +196,9 @@ export const InteractiveImmersiveLayout = ({
 
 	return (
 		<>
-			<Global styles={interactiveGlobalStyles} />
+			{CAPI.isLegacyInteractive && (
+				<Global styles={interactiveGlobalStyles} />
```
jlieb10, line 200: I wonder if renaming this to `interactiveLegacyStyles` would be helpful, too.

jlieb10 (approved):
Looks good 👍  Just have a think about my rename question

shtukas (approved):

src/web/layouts/InteractiveImmersiveLayout.tsx
OllysCoding: Good shout

OllysCoding: Done :D

OllysCoding:
Closing

PR #3131, opened 2021-06-23
Refactor `Card` stories

oliverlloyd:
## What does this change?
This PR ensures the list of stories used to capture all each `Card` variation based on `Design` is complete.

## Why?
Before we were missing some designs and also had two stories that were testing variation in `Display` which is already covered by another list of stories

By ensuring this list matches the [root definition](https://github.com/guardian/types/blob/main/src/format.ts) in the /types repo we improve discoverability and reduce confusion


shtukas (approved):

OllysCoding (approved):
LGTM!

PR #3132, opened 2021-06-23
Remove Australia mandatory signin gate test

QuarpT:
### Before
 Australia mandatory signin gate test is live

### After

Remove Australia mandatory signin gate test is removed

## Why?

The test reached the required sample size, so we can turn it off


oliverlloyd (approved):
I checked this against https://github.com/guardian/dotcom-rendering/pull/3133 and it marries well

PR #3133, opened 2021-06-23
Mandatory signin gate test for US readers

QuarpT:
### Before

There is no signin gate mandatory test for US readers

### After

There is a signin gate mandatory test for US readers

## Why?

We want to test the signin conversion for US readers when presented with a signin gate without a dismiss button. This is equivalent to the Australia test but for US readers.

The base branch is set to remove-aus-mandatory-gate to make the diff more realistic.  https://github.com/guardian/dotcom-rendering/pull/3132 needs to be merged first. Afterwards, merge this PR against main.


oliverlloyd (approved):
I checked this against https://github.com/guardian/dotcom-rendering/pull/3132 and it marries well

PR #3136, opened 2021-06-23
Retrieve header bidding information from switch

chrislomaxjones:
## What does this change?

Retrieve header bidding information from switches rather than from the config object.

Also involves the regeneration of fixtures to use a new config object that reflects the changes to header bidding properties in the config.

## Why?

To bring DCR in line with the corresponding [frontend PR](https://github.com/guardian/frontend/pull/23924) that exposes header bidding information via switches instead of via the config. This change has been made since the header bidding config properties were populated with information from switches anyway.


index.d.ts
```diff
@@ -672,7 +672,6 @@ interface ConfigType extends CommercialConfigType {
 	googletagUrl: string;
 	stage: string;
 	frontendAssetsFullURL: string;
-	hbImpl: { [key: string]: any } | string;
```
mxdvl, line 675: This is the good stuff!

src/model/json-schema.json
```diff
@@ -4141,4 +4129,4 @@
         }
     },
     "$schema": "http://json-schema.org/draft-07/schema#"
-}
```
mxdvl, line 4144: Do we want to keep this as-is?

src/model/window-guardian.ts
```diff
@@ -54,7 +54,10 @@ const makeWindowGuardianConfig = (
 			adUnit: config.adUnit,
 			showRelatedContent: true,
 			ajaxUrl: config.ajaxUrl,
-			hbImpl: config.hbImpl,
+			hbImpl: {
+				prebid: config.switches.prebidHeaderBidding,
+				a9: config.switches.a9HeaderBidding,
+			},
```
mxdvl, line 60: Do we still need this? I think that it’s only needed in `dfpEnv` in the end.

mxdvl (changes requested):

src/model/json-schema.json
chrislomaxjones: I think we have to remove it from the schema otherwise DCR still expects a "hbImpl" property in the config

mxdvl: I was only referring to the return at the end of the file!

chrislomaxjones: Oh I see!

src/model/window-guardian.ts
chrislomaxjones: Good point I don't think its actually used anywhere in DCR?

mxdvl (approved):

PR #3137, opened 2021-06-24
Add semantic structure to DCR's markup (Convert `Section` component to allow element name)

gtrufitt:
## What does this change?

Adds more semantic structure to the markup in DCR.

We do this by allowing the `Section` element to take the name of an html element. This is because this Component is the highest level component and the one that we want to change to use better semantic meaning. In fact, we remove Section completely as components are either

### Article

> The `<article>` HTML element represents a self-contained composition in a document, page, application, or site, which is intended to be independently distributable or reusable (e.g., in syndication).

### Aside

>The element can be used for typographical effects like pull quotes or sidebars, for advertising, for groups of nav elements, and for other content that is considered separate from the main content of the page.

### Div
Has little semantic meaning, so in the cases where we want to wrap the more semantic elements lower in the structure - `Nav`, `Footer` - we do *not* want to use Section, but the semantic-less Div.

### Before
image

### After

image

### DIFF
--- a/src/web/components/CommentCount.tsx
+++ b/src/web/components/CommentCount.tsx
@@ -76,7 +76,11 @@ export const CommentCount = ({
 	const { short, long } = formatCount(commentCount);
 
 	return (
-		<div css={containerStyles(palette)} data-cy="comment-counts">
+		<data
+			css={containerStyles(palette)}
+			data-cy="comment-counts"
+			value={`${long} comments on this article`}
+		>
 			<a
 				href="#comments"
 				css={linkStyles}
@@ -101,6 +105,6 @@ export const CommentCount = ({
 					{short}
 				</div>
 			</a>
-		</div>
+		</data>
 	);
 };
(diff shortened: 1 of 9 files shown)


nitro-marky:
I think having something like `ElementContainer` would be useful when trying to find components/debugging in the inspector. With the `Section` tag delineation between components is easy to see.

shtukas (approved):

oliverlloyd:
@SiAdcock @sndrs this is an interesting PR. How does this fit in with Source's `Container` component? Does that need to adapt to this use case or is having every page container (or 'slice') set as a `section` still needed? I see fronts lean on this a lot

Screenshot 2021-07-12 at 12 36 54


> Has little semantic meaning, so in the cases where we want to wrap the more semantic elements lower in the structure - Nav, Footer - we do not want to use Section, but the semantic-less Div.

I think this is going to be useful. I've seen use cases elsewhere where I've wanted the children to be the true semantic markers for the content and so having this option to just use a div could be nice. Of course, keeping all container markers at the same level is_nicer_ but is this really needed for accessibility?

PR #3140, opened 2021-06-24
Fix `AUS` privacy banner on AMP

mxdvl:
## What does this change?

Fixes the url pointing to the AUS AMP Privacy banner.

### Before

No banner

### After

A banner

## Why?

It’s our choice!


ashishpuliyel (approved):
Wow. Good find.

PR #3143, opened 2021-06-28
Use Data-Component to send Youtube Embed visibility

nitro-marky:
## What does this change?

Before there was no way to filter which pages had Youtube embeds. This uses the `data-component` attribute to notify Ophan that there is a Youtube embed on that page. This can be tested for metrics etc.

### Before

Screenshot 2021-06-28 at 12 25 46

Screenshot 2021-06-28 at 12 26 17

### After

Screenshot 2021-06-28 at 12 25 35

Screenshot 2021-06-28 at 12 26 06

### DIFF
--- a/src/web/components/elements/YoutubeBlockComponent.tsx
+++ b/src/web/components/elements/YoutubeBlockComponent.tsx
@@ -32,8 +32,6 @@ type Props = {
 	width?: number;
 	duration?: number; // in seconds
 	origin?: string;
-	// eslint-disable-next-line react/no-unused-prop-types
-	embedHandler?: () => void;
 };
 
 const expiredOverlayStyles = (overrideImage: string) => css`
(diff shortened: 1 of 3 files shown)


src/web/components/elements/YoutubeBlockComponent.tsx
```diff
@@ -148,7 +148,7 @@ export const YoutubeBlockComponent = ({
 	};
 
 	return (
-		<div data-chromatic="ignore">
+		<div data-chromatic="ignore" data-component="Youtube_Embed">
```
gtrufitt, line 151: I think the convention is lowercase, hyphenated ie: youtube-embed

src/web/components/elements/YoutubeBlockComponent.tsx
```diff
@@ -148,7 +148,7 @@ export const YoutubeBlockComponent = ({
 	};
 
 	return (
-		<div data-chromatic="ignore">
+		<div data-chromatic="ignore" data-component="youtube_embed">
```
gtrufitt, line 151:
```suggestion
		<div data-chromatic="ignore" data-component="youtube-atom">
```

gtrufitt (approved):
lgtm

PR #3144, opened 2021-06-28
Accessibility 'skip to main content' button

jlieb10:
## What does this change?
Adds a tab-able button for keyboard-only users and screen reader users to navigate to the "main content" portion of an article.

### Before
No such link existed on DCR

### After
skip to main content button

## Why?
(1) Parity with dotcom
(2) So keyboard users can better use the site


shtukas (approved):

src/web/components/MainContentIndicator.tsx
```diff
@@ -0,0 +1,4 @@
+export const MainContentIndicator = () => (
+	// eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+	<div id="maincontent" tabIndex={0} />
```
gtrufitt, line 3: Would it be worth licking through to the 'h1' or something like that on the page? We'd have to ensure that every layout type had an h1 with an ID but then the user would immediately jump to an h1 and if it was a screen reader, it would read it out to them?

gtrufitt (approved):
Not 100% sure on the reason for a seperate main content element to jump to but otherwise looks clean

OllysCoding (approved):

src/web/components/MainContentIndicator.tsx
jlieb10: >Not 100% sure on the reason for a seperate main content element to jump to but otherwise looks clean

I agree -I copied the implementation as it exists on frontend but it possibly could be improved? It's worth point out that 'main content' in this instance means where the body starts, so skipping past the title (H1) and the author and meta info (in addition to the menus and above the fold ads)

gtrufitt: Hm interesting yeah. `ArticleBody` probably makes sense then, yeah?

src/web/components/ArticleBody.tsx
```diff
@@ -91,6 +91,9 @@ export const ArticleBody = ({
 	}
 	return (
 		<div
+			// eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+			tabIndex={0}
```
mxdvl, line 95: It’s unclear why we introduced this rule. There seems to be no benefit in making the main content tabbable.

In most other places, we explicitely use `tabIndex={-1}`, which “ensures that focus continues throught the page and deos not stay on the skip link itself”.

PR #3145, opened 2021-06-28
Update MatchStats to use the codename

shtukas:
A `codename` attribute was added to team metadata here: https://github.com/guardian/frontend/pull/23958 , we are now consuming it to display a better MatchStats widget.

BEFORE:

BEFORE

AFTER:

AFTER


OllysCoding (approved):
LGTM!

PR #3149, opened 2021-06-29
Render zero stars in cards

jfsoul:
## What does this change?
This updates the Card component to use a ternary rather than `&&`, which means that a zero star rating is rendered as a  StarRatingComponent with zero stars, rather than the number zero. I've also added some stories for zero star reviews.

### Before
image

### After
image

## Why?
Zero star reviews are rare but legitimate! We should  render them  correctly  on cards, as we already do on the articles themselves.

### DIFF
--- a/src/web/components/Card/Card.Designs.stories.tsx
+++ b/src/web/components/Card/Card.Designs.stories.tsx
@@ -31,7 +31,7 @@ const ReviewWithStars = Format(
 		design: Design.Review,
 	},
 	'Review',
-	0 // star rating
+	0, // star rating
 );
 
 const Interview = Format(
--- a/src/web/components/Card/Card.Designs.stories.tsx
+++ b/src/web/components/Card/Card.Designs.stories.tsx
@@ -24,6 +24,16 @@ const Review = Format(
 	'Review',
 );
 
+const ReviewWithStars = Format(
+	{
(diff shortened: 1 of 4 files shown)


OllysCoding (approved):
LGTM!

shtukas (approved):

src/web/components/Card/Card.Designs.stories.tsx
```diff
@@ -24,6 +24,16 @@ const Review = Format(
...
+const ReviewWithStars = Format(
+	{
+		display: Display.Standard,
+		theme: Pillar.News,
+		design: Design.Review,
+	},
+	'Review',
+	0, // star rating
+);
+
```
oliverlloyd, line 36: Semantically this story shouldn't live here. I was in a meeting with Source recently and the fact that this list of stories did not align with `Format` was the cause for some confusion so I [updated them](https://github.com/guardian/dotcom-rendering/pull/3131) to fix this - but this PR reimplements the diversion 😭

jfsoul: I see what you mean, I had a similar thought but landed on the slightly more relaxed approach anyway, but then I didn't know that it had previously caused confusion outside the team. I'll rejig things.

PR #3150, opened 2021-06-29
Richlink Image Enhancement

jlieb10:
## What does this change?
Update richlink images to include alttext, width, and height by passing in this information through the endpoint in the Frontend repo.

This update is further to this endpoint enrichment change: https://github.com/guardian/frontend/pull/23967

### Before
Images were being served without width and height (which can cause layout shift)
Images were not being served with alt text (very bad for accessibility)

### After
images have width, height, and alttext attributes

## Why?
To avoid layout shift and increase accessibility (and keep lighthouse happy)
https://trello.com/c/0f2QE0hw/2899-add-image-dimensions-to-rich-links-on-dcr-and-review-others

### DIFF
--- a/src/web/components/elements/RichLinkComponent.tsx
+++ b/src/web/components/elements/RichLinkComponent.tsx
@@ -1,4 +1,4 @@
-import { RichLink } from '@root/src/web/components/RichLink';
+import { RichLink, RichLinkImageData } from '@root/src/web/components/RichLink';
 import { DefaultRichLink } from '@root/src/web/components/DefaultRichLink';
 
 import { useApi } from '@root/src/web/lib/useApi';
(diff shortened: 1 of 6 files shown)


src/web/components/elements/RichLinkComponent.tsx
```diff
@@ -23,6 +23,24 @@ interface CAPIRichLinkType {
...
+	displayCredit: string;
+	source: string;
+	photographer: string;
+	isMaster: string;
+	altText: string;
+	height: string;
+	credit: string;
+	mediaId: string;
+	width: string;
 }
```
OllysCoding, line 44: What do we use these extra fields for (isMaster, source, etc)? Can we remove them from the interface?

jlieb10: They aren't being used, but I thought it might be good to expose what is available in the data in case we do decide to use them in the future

rcrphillips:
[Add image dimensions and alt text to rich links on DCR (and review others)](https://trello.com/c/0f2QE0hw/2899-add-image-dimensions-to-rich-links-on-dcr-and-review-others)

shtukas (approved):

OllysCoding (approved):

PR #3152, opened 2021-06-29
Send commercial metrics

mxdvl:
## What does this change?

This enables sending commercial metrics via DCR, using the `commercial-core` bundle directly, and enables tests to force the capture of commercial metrics for all pageviews in any of its variants.

Frontend: https://github.com/guardian/frontend/pull/23955

- add `@guardian/commercial-core`
- add link to frontend ab-tests file
- add empty commercial metrics test
- send commercial metrics via DCR

### Before

Analytics are not forced for commercial A/B Tests.

### After

Analytics can be forced.

## Why?

We want to be able to compare commercial parters’ performance adequately.


src/web/components/App.tsx
```diff
@@ -329,6 +331,32 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
...
+
+		log('commercial', { shouldForceMetrics, userIsInSamplingGroup });
+
+		const { isDev } = window.guardian.config.page;
+
+		if (isDev || shouldForceMetrics || userIsInSamplingGroup) {
+			// we’re doing this twice
+			const browserId = getCookie('bwid') || undefined;
+
+			// TODO: only send them once, using the return value of sendBeacon
```
MarSavar, line 355: Thanks Max!
Since we're sending metrics twice, would it make sense to add a static field to `EventTimer` (something like `commercialMetricsSent` which could be either `true` or `false`, and set the value to `true` when metrics are first sent? Since `EventTimer` is a singleton and we always retrieve the same instance with `EventTimer.get()` this should solve the problem of sending duplicate metrics.

mxdvl: Yeah, it’s a draft, we can used the same `logged` value, but I need to first figure out how state works in React!

mxdvl: Should be fixed now!

src/web/components/App.tsx
```diff
@@ -327,7 +336,49 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
...
+	const [sentCommercialMetrics, setSentCommercialMetrics] = useState<boolean>(
+		false,
+	);
+	const visibilityState = useDocumentVisibilityState();
+	useEffect(() => {
+		if (!window.guardian.config.switches.commercialMetrics) return;
+		if (!window.guardian.config.ophan) return;
+		log('commercial', 'document.visibilityState:', visibilityState);
+		if (visibilityState !== 'hidden') return;
+		if (sentCommercialMetrics === true) return;
```
chrislomaxjones, line 358: Could this just be?
```suggestion
		if (sentCommercialMetrics) return;
```

src/web/components/App.tsx
```diff
@@ -327,7 +336,49 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
...
+
+		const shouldForceMetrics = ABTestAPI.allRunnableTests(
+			tests,
+		).some((test) =>
+			[commercialPartner].map((t) => t.id).includes(test.id),
```
chrislomaxjones, line 363: Is `[commercialPartner]` the array of AB tests for which we want to force metrics? If so could it maybe be declared as a constant at the top with the other constants?

chrislomaxjones (approved):
Looks great!

src/web/components/App.tsx
mxdvl: I was wondering about that. Let's make it explicit, and then we can also type it to an array of `ABTest`

src/web/components/App.tsx
```diff
@@ -158,6 +163,11 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 	>();
 
 	const pageViewId = window.guardian?.config?.ophan?.pageViewId;
+	const [browserId, setBrowserId] = useState<string>();
+	useEffect(() => {
+		// TODO: can the browserId actually be null?
+		setBrowserId(getCookie('bwid') ?? undefined);
+	}, []);
```
oliverlloyd, line 170: This `useEffect` means this code only runs after rendering has completed, and only once (because of the empty deps array). No idea if this is the case or not but do we need that? Could this code run straight away, before render?

src/web/components/App.tsx
```diff
@@ -327,7 +337,50 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 				pubData,
 			});
 		}
-	}, [countryCode, CAPI.config.switches.consentManagement, pageViewId]);
+	}, [
+		countryCode,
+		CAPI.config.switches.consentManagement,
+		pageViewId,
+		browserId,
```
oliverlloyd, line 344: Adding `browserId` here (where it is state being set in another `useEffect`) may add a delay and/or cause this code to run more often

src/web/lib/useDocumentHidden.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function useDocumentVisibilityState(): VisibilityState {
+	const [visibilityState, setVisibilityState] = useState(
+		document.visibilityState,
+	);
+	const onVisibilityChange = () =>
+		setVisibilityState(document.visibilityState);
+
+	useEffect(() => {
+		document.addEventListener(
```
oliverlloyd, line 11: I am really not sure if this is an issue or not but my PR radar is being pinged here by adding an event listener inside a useEffect. I guess one issue is it could be set and removed multiple times

oliverlloyd (commented):
This is a little outside of the scope of this PR but it really does seem like there's a real problem in DCR with adding this kind of logic. It's just too hard to do. This code may very well work fine, it's from Max so I suspect it does 😉 , but it is worrying that reviewing it makes my brain hurt

Essentially, you want to fire a function once on page load based on some dependencies, right? Why does our App structure make doing this so painful and prone to errors?

Can we do this better? I feel like as a platform DCR should be offering rails (custom hooks?)  for people to do things like this.

src/web/components/App.tsx
mxdvl: Ah, yes, I think that would be better, I’m not super familiar with the React lifecycle here!

src/web/lib/useDocumentHidden.ts
mxdvl: This is a fair point, and we discussed it with @MarSavar, and opted for using an empty array as the second param.

By adding logs we could actually confirm that it was fired up several times.

src/web/components/App.tsx
```diff
@@ -327,7 +337,50 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
...
+
+	// ************************
+	// *      Commercial      *
+	// ************************
+	const [sentCommercialMetrics, setSentCommercialMetrics] = useState<boolean>(
+		false,
+	);
+	const visibilityState = useDocumentVisibilityState();
+	useEffect(() => {
+		if (!window.guardian.config.switches.commercialMetrics) return;
```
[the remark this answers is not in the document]
mxdvl: [Estus Flask seems to think it does](https://stackoverflow.com/a/54615649/1577447).

It’s only if you return a function that it acts as a cleanup.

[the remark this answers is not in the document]
oliverlloyd: Ha, I guess it's just a normal function so of course it works. Happy to learn a new trick!

src/web/components/App.tsx
oliverlloyd: > Any idea how we’d do that? Just use a `const` one level higher?

So yes, using a const would be the alternative. However, the down side to doing _this_ is `document.cookie` is synchronous and has performance issues so this code will slow down hydration. Sorry, I didn't think that one through before making my original comment!

It's a trade off. Some information is critical and worth blocking the lifecycle for because it's required to make decisions but in other situations it's better to bring the information in asynchronously (after the App code has already been run).

### However
There's another factor to consider here. By adding state to `App`, changing that state will cause the whole file to rerender which could have unintended consequences in other areas. This is a structural problem with how `App` has been built, it advertises itself as the place where things like this should go but actually it might not be. To avoid rerenders in other areas you can drop the state down into a kind of wrapper or container component, using composition to control rendering.

src/web/components/App.tsx
```diff
@@ -327,7 +336,49 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
...
+		pageViewId,
+		browserId,
+	]);
+
+	// ************************
+	// *      Commercial      *
+	// ************************
+	const [sentCommercialMetrics, setSentCommercialMetrics] = useState<boolean>(
+		false,
+	);
```
oliverlloyd, line 351: You might find `useOnce` works as an alternative here, rather than having state to ensure something only occurs once.

oliverlloyd: https://github.com/guardian/dotcom-rendering/blob/44bf079a459f5c29a5c3a03d946bbd28dd5df65a/src/web/lib/useOnce.ts

src/web/components/App.tsx
```diff
@@ -73,13 +74,16 @@ import { VineBlockComponent } from '@root/src/web/components/elements/VineBlockC
 import type { BrazeMessagesInterface } from '@guardian/braze-components/logic';
 import { OphanRecordFunction } from '@guardian/ab-core/dist/types';
 import { ConsentState } from '@guardian/consent-management-platform/dist/types';
+import { ABTest } from '@guardian/ab-core';
```
oliverlloyd, line 77: I think we should be using `@guardian/ab-react` instead of `ab-core` here

src/web/components/App.tsx
mxdvl: So, would you suggest leveraging `useOnce`? This way this is only ever set once.

Should this also contain the lookup for checking if a user is signed-in? I’m looking at `!!getCookie('GU_U')` on L200.

src/web/components/App.tsx
mxdvl: I think then maybe `useOnce` is the right concept, here? We need to have a `countryCode`, the switches, and a `browserId` + `pageViewId` pair, but these should never change during the page lifecycle.

mxdvl: Also, `cmp.init` can only run once. If it’s `init`-ed, it won’t init again.

mxdvl:
After discussing with @oliverlloyd, we moved the logic inside a new component, which contains state but does not render anything on the page. This way, we know that there is no need to add more `useEffect` hooks to `App.tsx`.

src/web/components/App.tsx
```diff
@@ -523,6 +534,15 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
...
 		<React.StrictMode>
+			{[
+				CAPI.config.switches.commercialMetrics,
+				window.guardian.config.ophan !== undefined,
+			].every(Boolean) && (
+				<CommercialMetrics
+					browserId={browserId}
+					pageViewId={pageViewId}
+				/>
+			)}
```
oliverlloyd, line 545: Nice!

oliverlloyd (approved):
This is looking great! This is a really good example for how state can be kept out of App and nicely isolated in it's own component.

I've made a few comments but I'm happy to approve and for you to use your best judgement when carrying this forward

src/web/components/App.tsx
```diff
@@ -327,7 +333,12 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 				pubData,
 			});
 		}
-	}, [countryCode, CAPI.config.switches.consentManagement, pageViewId]);
+	}, [
+		countryCode,
```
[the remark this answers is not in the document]
mxdvl: How can we test this? In doubt, I will revert to `useEffect`.

shtukas (approved):

shtukas:
Merging despite un-accepted changes from chromatic because of some confusion between PROD and chromatic over visual differences.

PR #3155, opened 2021-06-29
Add event listener for resetting hydration state between pageviews

gtrufitt:
- When changing between stories in storybook, the hydration state was not reset
- But the 'server render' *was* reset, and so the 'portals' do not end up being rehydrated
- Therefore things like 'comment count' would not show up when clicking between layout files

### DIFF
--- a/src/web/components/HydrateOnce.tsx
+++ b/src/web/components/HydrateOnce.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useState, SetStateAction } from 'react';
 import ReactDOM from 'react-dom';
 
 import { initPerf } from '@root/src/web/browser/initPerf';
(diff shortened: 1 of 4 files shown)


shtukas (approved):

OllysCoding (approved):
Looks great but definitely worth adding some comments to HydrateOnce for context

src/web/components/HydrateOnce.tsx
```diff
@@ -32,5 +32,11 @@ export const HydrateOnce = ({ rootId, children, waitFor = [] }: Props) => {
 		end();
 	});
 	setAlreadyHydrated(true);
+	// @ts-ignore
+	if (window.STORYBOOK_ENV) {
+		window.addEventListener('storybook:reset', () =>
+			setAlreadyHydrated(false),
+		);
+	}
```
OllysCoding, line 40: Can this go in `useEffect`?

gtrufitt: I did it slightly differently in the end, removing the event listeners immediately before adding it again

PR #3157, opened 2021-06-30
Increase us mandatory gate audience

QuarpT:
## What does this change?

Increase US mandatory signin gate audience to 25%. This eats into the main variant audience. The mandatory gate variant is same as main variant, but without the dismiss button.

## Why?

We want more users to see mandatory variant


coldlink (approved):
aprroving, but it'll be interesting to know why the  TC build is failing

PR #3159, opened 2021-06-30
Restore semantic story layout

jfsoul:
## What does this change?
This removes the additional Card story I added in #3149 and just adds stars to the standard `Review` `Card.Design` story. As pointed out by @oliverlloyd we should ideally try and maintain the list of stories within `Design` as a one-to-one mapping with the possible values of `Format.Design` for semantics and  to avoid confusion.


oliverlloyd (approved):
🙏 ❤️ 🙏

PR #3167, opened 2021-07-05
Wait for countryCode before fetching RR header links

tomrf1:
there's currently a race condition where the header request is made before countryCode becomes available


TomPretty (approved):
Nice one - looks good! 👍

PR #3169, opened 2021-07-05
Send labels for epic insert + view events

TomPretty:
## What does this change?
Send labels for epic insert + view events. This is so we can use the labels field to distinguish "super mode" epic views from ordinary views.

In discussing this, we noticed that the `TestMeta` type is only used by us (contributions) and probably doesn't belong along with the other ophan types. In fact, we actually also provide the same data to the modules themselves through the props, so to avoid double sending it and also having to make tweaks across dcr and frontend, we've created a ticket to update the components to send their own events and remove the `meta` object from the platforms. For the time being, this PR will do the job though!

### DIFF
--- a/src/web/browser/ophan/ophan.ts
+++ b/src/web/browser/ophan/ophan.ts
@@ -44,9 +44,13 @@ export const submitComponentEvent = (
 	ophanRecord({ componentEvent });
 };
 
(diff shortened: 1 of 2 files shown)


tomrf1 (approved):
👍 we chatted IRL about removing the TestMeta type later, as a separate piece of work

PR #3171, opened 2021-07-06
Stop Inverse Headings Being Cut Off

nitro-marky:
## What does this change?

Previously, when inverse headlines were viewed in mobile format any hanging letters would be cut off at the bottom. The changes the line height below tablet size to stop this from happening.

### Before

Screenshot 2021-07-06 at 09 39 50

### After

Screenshot 2021-07-06 at 09 54 26

## Why?

Looks better innit.


shtukas (approved):

nitro-marky:
[Interview headlines have linespace that cuts off descenders on mobile](https://trello.com/c/drDtLgLQ/125-interview-headlines-have-linespace-that-cuts-off-descenders-on-mobile)

PR #3172, opened 2021-07-06
Carousel template for onwards upper

jlieb10:
## What does this change?
changes the onwards component from a static grid layout to a dynamic carousel layout

### Before
before

### After
after

## Why?
The small carousel layout was the winning variant for this component when AB tested


jlieb10:
Awaiting any design requests/changes from Harry

nitro-marky (approved):
Looks great!

src/web/components/Onwards/OnwardsUpper.tsx
```diff
@@ -70,6 +71,116 @@ const onwardsWrapper = css`
 	width: 100%;
 `;
 
+const headlinesContainer = (edition: Edition): string => {
```
lucymonie, line 74: These container functions all seem do the same thing - would it be worth abstracting this and passing a single function the type (eg 'headlines', 'sport', 'opinion' etc) and have it make the regional selection accordingly - maybe by accessing an object with the data?

lucymonie (approved):
Just one suggestion - otherwise, it looks good!

src/web/components/Onwards/OnwardsUpper.tsx
jlieb10: great idea - please check now and lmk your thoughts :)

PR #3174, opened 2021-07-06
Activate Sub-Parent For Genre

nitro-marky:
## What does this change?

**Update:** The use of `!important` has been removed. 

Subnavs are not being highlighted for the correct article. Bizarrely this behaviour works for mobile but the CSS gets overridden on larger displays.

### Before

Screenshot 2021-07-06 at 16 39 16

### After

Screenshot 2021-07-06 at 16 40 08

## Why?

Looks better.

### DIFF
--- a/src/web/components/SubNav/SubNav.tsx
+++ b/src/web/components/SubNav/SubNav.tsx
@@ -98,7 +98,7 @@ const linkStyle = css`
 `;
 
 const selected = css`
-	font-weight: 700;
+	font-weight: 700 !important;
 `;
 
 const spaceBetween = css`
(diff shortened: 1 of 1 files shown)


src/web/components/SubNav/SubNav.tsx
```diff
@@ -98,7 +98,7 @@ const linkStyle = css`
 `;
 
 const selected = css`
-	font-weight: 700;
+	font-weight: 700 !important;
```
gtrufitt, line 101: `!important` is almost always a risky solution, I would try and ensure that the cascade is followed instead, so the order in which variables are added I think means they'll be added in the CSS

shtukas (approved):

OllysCoding (approved):
👍

nitro-marky:
[Subnav active parent not bolded](https://trello.com/c/ul35yfOL/120-subnav-active-parent-not-bolded)

PR #3177, opened 2021-07-07
Dynamically load components

jlieb10:
### Before
We were dynamically loading: 
-EditionDropdown
-YoutubeBlock
-RichLinkBlock
-InteractiveBlock

### After
We are dynamically loading:
-EditionDropdown
-YoutubeBlock
-RichLinkBlock
-InteractiveBlock
-CalloutBlock
-QABlock
-TimelineBlock
-DocumentBlock
-EmbedBlock
-InstagramBlock
-MapBlock
-SpotifyBlock
-FacebookVideoBlock
-VineBlock

## Why?
Page speed and core web vitals: Load javascript bundles based on the needs of the article

### DIFF
--- a/src/web/components/App.tsx
+++ b/src/web/components/App.tsx
@@ -4,7 +4,6 @@ import { useAB } from '@guardian/ab-react';
 import { tests } from '@frontend/web/experiments/ab-tests';
 import { ShareCount } from '@frontend/web/components/ShareCount';
 import { MostViewedFooter } from '@frontend/web/components/MostViewed/MostViewedFooter/MostViewedFooter';
-import { CalloutBlockComponent } from '@root/src/web/components/elements/CalloutBlockComponent';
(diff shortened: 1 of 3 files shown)


src/web/components/App.tsx
```diff
@@ -429,6 +420,186 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
...
+				return import(
+					'@frontend/web/components/elements/SpotifyBlockComponent'
+				);
+			}
+			return Promise.reject();
+		},
+		{
+			resolveComponent: (module) => module.SpotifyBlockComponent,
+		},
+	);
```
gtrufitt, line 561: I wish that these loadable function callbacks could be less repetitive, but if I remember correctly from my original implementation it messes with types when doing that.

gtrufitt (approved):
Nice one! One for the morning so we can track to see if we see any difference in Core Vitals

PR #3178, opened 2021-07-07
Allow iframes in interactive sanitisation

OllysCoding:
## What does this change?

Iframes were being removed as part of the interactive sanitisation

### Before

image

### After

image

## Why?

Missing content


gtrufitt (approved):
Brill

PR #3181, opened 2021-07-09
Add alarm if there are more than 100 requests per minute for 5 minutes that 5xx 

davidfurey:
## What does this change?

Creates a new alarm that will trigger a pager duty alert if DCR produces an unexpected number of 5xx errors.  

NB- this cloudformation change will need to be deployed manually in the AWS console since it adds a new parameter without a valid default value (`NotificationAlarmAction`)

## Why?

There was recently a bug in DCR causing quite a few 5xx for particular article types that was picked up by CP.  Ideally we'd notice these issues before CP does.

### DIFF
--- a/cloudformation.yml
+++ b/cloudformation.yml
@@ -26,6 +26,10 @@ Parameters:
   AMI:
     Description: AMI to use for instances
     Type: AWS::EC2::Image::Id
+  NotificationAlarmAction:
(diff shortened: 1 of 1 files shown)


shtukas (approved):

nicl (approved):
Looks good!

PR #3186, opened 2021-07-13
Fix Braze Epic Rendering

tjmw:
## What does this change?

Short term fix for an issue where some [additional validation was added to the `ContributionsEpic` component] which Braze uses to render epic messages. This adds some missing params required by the validation (which I don't think are actually used in our case) to get the epic rendering again.

Slightly longer term this has made really clear the fragility of relying on the support-dotcom-components epic for Braze messages. We're going to migrate our own version of the epic into `@guardian/braze-components` which will make us more robust to issues like this.

### Before

Braze campaigns using the Epic component not rendering.

### After

Braze campaigns using the Epic component rendering as expected.

### DIFF
--- a/src/web/components/SlotBodyEnd/BrazeEpic.tsx
+++ b/src/web/components/SlotBodyEnd/BrazeEpic.tsx
@@ -101,7 +101,7 @@ const buildEpicProps = (
 			return null;
 		},
 		(variant) => {
-			const pageTracking = {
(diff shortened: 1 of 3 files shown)


arausuy (approved):

michaelclapham (approved):
👍

tomrf1 (approved):

PR #3187, opened 2021-07-13
Pass in opts to skip comments in DOMPurify

lucymonie:


lucymonie:
Trying an approach the hopefully works better!

PR #3188, opened 2021-07-13
remove multiple lines on subnav

DavidLawes:
### Before
layouts included 4 horizontal lines in the sub nav, which didn't match the expected design.
incorrect-subnav

### After
layouts have only a single horizontal line in the subnav.
Screenshot 2021-07-13 at 18 08 17

## Why?
fix layouts to match intended designs.

some notes:
- i validated layouts using `make dev` as well as storybook, ensuring all the layout types defined in DecideLayout.tsx were correctly updated (amp layouts not affected by this issue).
- i was uneasy about having to make the same change in multiple locations. we wondered if it is something that could be encapsulated within subnav - @nicl?

### DIFF
--- a/src/web/layouts/CommentLayout.tsx
+++ b/src/web/layouts/CommentLayout.tsx
@@ -659,7 +659,7 @@ export const CommentLayout = ({
 						currentNavLink={NAV.currentNavLink}
 						palette={palette}
 					/>
-					<Lines count={4} effect="straight" />
+					<Lines count={1} effect="straight" />
 				</ElementContainer>
 			)}
 
--- a/src/web/layouts/ImmersiveLayout.tsx
+++ b/src/web/layouts/ImmersiveLayout.tsx
@@ -692,7 +692,7 @@ export const ImmersiveLayout = ({
 						currentNavLink={NAV.currentNavLink}
 						palette={palette}
 					/>
-					<Lines count={4} effect="straight" />
+					<Lines count={1} effect="straight" />
 				</ElementContainer>
 			)}
 
--- a/src/web/layouts/InteractiveImmersiveLayout.tsx
+++ b/src/web/layouts/InteractiveImmersiveLayout.tsx
@@ -258,7 +258,7 @@ export const InteractiveImmersiveLayout = ({
 						currentNavLink={NAV.currentNavLink}
 						palette={palette}
 					/>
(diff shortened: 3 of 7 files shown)


oliverlloyd:
Hello 👋 ! Just dropping in for some random comments

> i was uneasy about having to make the same change in multiple locations

I hear you. I think it's worth noting that there wasn't an explicit intention to have each layout duplicate all the things.

We did see areas of duplication when we were building these, however the abstractions you _think_ you need at the start of a process rarely turn out to be right and this can cause painful refactoring or, worse, overloaded abstractions. To avoid this we took the approach of duplicating layout work with the view to create abstractions at a later stage once we were confident that they were the right thing to do. There's no good rule for when this is the case - it's kind of a feeling. But the wrong abstraction can be very costly so you need to be careful. Sometimes making the same edits in multiple places is a price worth paying for flexibility and protection against complex coupling.

But I am confident that there _are_ places in the layout files that could be broken out - esp. around the end of the article, onwards is one area that, both technically and from a business perspective, seems to stand apart. I just wanted to sound a note of caution because coupling these layout files together in the wrong way could really come back to bite us!

src/web/layouts/CommentLayout.tsx
```diff
@@ -659,7 +659,7 @@ export const CommentLayout = ({
 						currentNavLink={NAV.currentNavLink}
 						palette={palette}
 					/>
-					<Lines count={4} effect="straight" />
+					<Lines count={1} effect="straight" />
```
oliverlloyd, line 662: Not sure about this but does having 1 line mean we have 2 lines alongside each other? One from here and one from the top border of the Pillar nav?

DavidLawes: Hey Oliver - thanks for commenting :) i've checked in with our designer, they'd prefer no line at all, thanks for highlighting this! i'll remove the Lines element completely

nicl:
Re duplication, agree with @oliverlloyd here. There are places we could better share code, but we should take care here to not couple things too tightly. In this case, we could add a `withLines` (or another name) bool flag to the Subnav component, and use that to toggle things. E.g. as subnav at top of page includes lines, whereas at bottom does not.

But very happy to do that as a separate PR (or even leave it for now).

nicl (approved):
🚀

PR #3189, opened 2021-07-14
Add test around checking Braze dependencies when user isn't signed in

tjmw:
## What does this change?

Adds a missing test scenario to `checkBrazeDependencies.test.ts`. I noticed this case was missing while working on something related.

### Before

No test coverage for the scenario where we're checking the dependencies needed to load the Braze web SDK and the user isn't signed in.

### After

Test coverage for the scenario above.

## Why?

Increased confidence when making changes to this code.

### DIFF
--- a/src/web/lib/braze/checkBrazeDependencies.test.ts
+++ b/src/web/lib/braze/checkBrazeDependencies.test.ts
@@ -161,6 +161,38 @@ describe('checkBrazeDependecies', () => {
 		}
 	});
 
+	it('fails when the user is not signed in', async () => {
+		setWindow({
+			guardian: {
+				config: {
+					switches: {
+						brazeSwitch: true,
+					},
+					page: {
+						brazeApiKey: 'fake-api-key',
+						isPaidContent: false,
+					},
+				},
+			},
+		});
+		mockConsentsPromise = Promise.resolve(true);
+
+		const isSignedIn = false;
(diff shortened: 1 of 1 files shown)


src/web/lib/braze/checkBrazeDependencies.test.ts
```diff
@@ -161,6 +161,39 @@ describe('checkBrazeDependecies', () => {
...
+					},
+					page: {
+						brazeApiKey: 'fake-api-key',
+						isPaidContent: false,
+					},
+				},
+			},
+		});
+		mockBrazeUuid = null;
+		mockConsentsPromise = Promise.resolve(true);
```
oliverlloyd, line 179: Are these globals? I don't see them being used. Not anything for this PR but perhaps a seed worth planting around the api for that function - could these values be passed in?

oliverlloyd (approved):
This test reads really well, nice work!

src/web/lib/braze/checkBrazeDependencies.test.ts
tjmw: Hey great question. These are defined at the top of the file and are used by `jest.mock` as we're mocking at the module level. I think you're right, it's probably a sign that the interface to this function isn't quite right. In the short term (this PR) in this test we don't care about the `mockBrazeUuid` so I'll remove that line.

PR #3190, opened 2021-07-15
Allow scripts in interactive atom HTML

nicl:
## What does this change?

Allow scripts in interactive atom HTML

### Before
Screenshot 2021-07-15 at 10 08 06

### After
Screenshot 2021-07-15 at 10 07 48

## Why?

Some require this, e.g.

https://www.theguardian.com/destination-nsw-road-trips/ng-interactive/2021/mar/12/from-vineyards-to-waterfalls-and-sand-dunes-12-road-trips-to-seek-out-in-nsw

### DIFF
--- a/src/model/enhance-interactive-atom-elements.ts
+++ b/src/model/enhance-interactive-atom-elements.ts
@@ -4,19 +4,18 @@ import { sanitiseHTML } from '@root/src/model/clean';
 // preserve (approximate) parity with Frontend we perform some basic
 // cleaning.
 const enhance = (elements: CAPIElement[]): CAPIElement[] => {
-	elements.map((element) => {
+	return elements.map((element) => {
 		if (
 			element._type ===
 			'model.dotcomrendering.pageElements.InteractiveAtomBlockElement'
-		)
+		) {
 			element.html = element.html
-				? // Allow iframes, this is for youtube embeds in interactives, etc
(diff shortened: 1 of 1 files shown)


lucymonie (approved):
LGTM

src/model/enhance-interactive-atom-elements.ts
```diff
@@ -8,12 +8,12 @@ const enhance = (elements: CAPIElement[]): CAPIElement[] => {
...
 			'model.dotcomrendering.pageElements.InteractiveAtomBlockElement'
-		)
+		) {
 			element.html = element.html
-				? // Allow iframes, this is for youtube embeds in interactives, etc
-				  sanitiseHTML(element.html, { ADD_TAGS: ['iframe'] })
+				? sanitiseHTML(element.html, { ADD_TAGS: ['iframe', 'script'] })
 				: element.html;
-		return element;
+			return element;
```
OllysCoding, line 15: Would having the `return element` inside the `if` block mean that all non `InteractiveAtomBlockElement` are set to undefined?

nicl: Great spot! Thanks. It shows why always wrapping an `if` body in braces is a good idea - avoids this kind of mistake.

OllysCoding (approved):
Looks good !

PR #3191, opened 2021-07-15
Allow comments in DOMPurify for amp so we can set mobileHeight

lucymonie:
## What does this change?
This is a small change to the way we use DOMPurify to clean our html for amp. This will allow us to pass an option to DOMPurify that will skip comments, allowing information about the height of the element on mobile to be passed through.

## Before
Screen Shot 2021-07-15 at 13 33 03

Annoyingly, the screenshot of the before scenario looks more correct than the after shot! However, the height that's getting passed through in the comment is 600, which is the height of the element in the after shot.

## After
Screen Shot 2021-07-15 at 13 31 40

## Why?
On the Euro 2020 golden boot interactive, we are not picking up the `MobileHeight` comment that's used to set the initial height of the amp-iframe element.

https://www.theguardian.com/football/ng-interactive/2021/jul/07/euro-2020-golden-boot-top-goalscorers-game-by-game

This was because DOMPurify (which was called in the clean function in `clean.ts` for all AMP elements) was removing HTML comments before we get a chance to parse the comment.


OllysCoding (approved):
Looks good! Nice solution :)

src/model/clean.ts
```diff
@@ -11,18 +12,14 @@ export const bigBullets = (s: string) =>
...
+export const clean = compose(bigBullets, (s: string) =>
+	minify(s, {
+		collapseWhitespace: true,
+		minifyCSS: true,
+		minifyJS: true,
+	}),
 );
 
-export const sanitiseHTML = (html: string, opts?: any): string =>
+export const sanitiseHTML = (html: string | ReactNode, opts?: any): string =>
```
nicl, line 23: I wasn't sure if this would work? `DOMPurify.sanitize` only works on strings I think.

src/model/clean.ts
```diff
@@ -11,18 +12,14 @@ export const bigBullets = (s: string) =>
...
-	bigBullets,
-	(s: string) =>
-		minify(s, {
-			collapseWhitespace: true,
-			minifyCSS: true,
-			minifyJS: true,
-		}),
-	// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-	DOMPurify.sanitize,
+export const clean = compose(bigBullets, (s: string) =>
```
nicl, line 15: We could rename this to something like `minimise` now that it's purpose is a bit more narrow/just minification?

src/model/clean.ts
lucymonie: TS was complaining about it potentially not being a string, but I can take another look.

src/model/clean.ts
lucymonie: Good point - shall do.

src/model/clean.ts
nicl: okay yeah that would be great. I just wasn't sure if DOMPurify would blow up if passed something other than a string. The docs only mention string at least.

lucymonie: Actually yes - I had misunderstood this - the problem was that it was potentially undefined. I have refactored to make sure TS knows it isn't.

nicl (approved):
🎈

src/model/enhance-interactive-atom-elements.ts
```diff
@@ -1,21 +1,29 @@
...
 		if (
 			element._type ===
 			'model.dotcomrendering.pageElements.InteractiveAtomBlockElement'
 		) {
-			element.html = element.html
-				? sanitiseHTML(element.html, {
-						ADD_TAGS: ['iframe', 'script'],
-				  })
-				: element.html;
+			if (isAmp) {
```
nicl, line 15: I should have mentioned sorry - one of our principles with AMP/web is to avoid `isAmp` checks and keep separate code separate. This is to make refactoring things easier (we don't want AMP changes to break web for example), and also for easier testing etc.

Really we should move the 'model' stuff under `web/enhancers` or something as it's entirely web-related atm. I'll raise that as a separate PR.

src/model/clean.ts
```diff
@@ -3,26 +3,30 @@ import createDOMPurify from 'dompurify';
 import { JSDOM } from 'jsdom';
 import { minify } from 'html-minifier';
 
-// We don't represent lists in incopy, so things will just come across with bullet characters.
+// We don't represent lists in in copy, so things will just come across with bullet characters.
```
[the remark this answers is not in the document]
lucymonie: Okay sure, I'll change, thanks Nic

src/model/enhance-interactive-atom-elements.ts
lucymonie: Sure - does this mean it should be in a separate file? Or just a separate helper?

OllysCoding (approved):
This looks really good! Plenty of room to expand enhancers on AMP if we need but right now this fits the bill :)

PR #3194, opened 2021-07-16
Add some interactive legacy figure classes

nicl:
## What does this change? + Why

Adds a wrapper class to interactive block elements on immersive interactive pages, as expected by some of these to work, e.g.

https://www.theguardian.com/climate-academy/ng-interactive/2020/nov/16/regenerative-agriculture-soil-and-mushrooms-help-solve-climate-crisis

Note, in this specific case, it fixes some things but the interactive still has issues on DCR.


lucymonie (approved):
This looks good to me 👍🏻

OllysCoding (approved):

PR #3195, opened 2021-07-16
Update owners + PR template

nicl:
## What does this change?

Update owners + PR template.

## Why?

To reflect current team, and bump the 'Why' section in the PR template.


lucymonie (approved):
Thanks Nic! I agree with moving 'Why' up as well - it's a more natural place.

PR #3196, opened 2021-07-19
Use `palette` for the background on `InteractiveBlockComponent`

oliverlloyd:
## What?
Replaces the hard-coded white background for `InteractiveBlockComponent`s with `palette.background.article`

## Why?
Because not every article has a white background and so when the container is larger than the iframe you get undesired whiteness, such as this bar shown below.

This was raised on the dotcom channel as affecting embeds in some recent [articles on Pegasus](https://www.theguardian.com/news/2021/jul/18/huge-data-leak-shatters-lie-innocent-need-not-fear-surveillance).

### Before
Screenshot 2021-07-19 at 13 22 51

### After
Screenshot 2021-07-19 at 13 23 23


OllysCoding (approved):
LGTM :)

PR #3197, opened 2021-07-19
Remove tags when linking bylines, so there cannot be duplicates

jonathonherbert:
## What does this change?

This PR removes tags that have already been referenced when linking bylines, to ensure that we don't accidentally mis-link when we have two people with the same name, but different tags.

The assumption here is that bylines only ever happen to the same person once. If that's not true, this'll break.

## Why?

We've seen this issue in production. From Charlotte Naughton:

_> Hi there. This hasn't been a big deal, so haven't mentioned it before, but we can't check bylines in preview, just get an error message. Then something like this comes along, and we need to know we've got the Duncans in the right order! Is this problem fixable?
https://viewer.gutools.co.uk/preview/commentisfree/2021/jul/20/proposed-secrecy-law-journalism-spying-home-office-public-interest-whistleblowing_

### Before

Screenshot 2021-07-20 at 14 29 33

### After

Screenshot 2021-07-20 at 14 32 20

### Dev notes

The last commit is optional – it consolidates code that's common to both AMP and Web components to avoid duplication, but the diff is a bit larger as a result.

### DIFF
--- a/src/amp/components/topMeta/Byline.stories.tsx
+++ b/src/amp/components/topMeta/Byline.stories.tsx
@@ -1,0 +1,63 @@
+import { Byline } from './Byline';
+
+const guardianBaseURL = 'https://theguardian.com';
+
+export const SingleByline = () => (
+	<Byline
+		byline="Eva Smith and friends"
+		guardianBaseURL={guardianBaseURL}
+		tags={[
+			{
+				id: 'eva-smith',
+				type: 'Contributor',
+				title: 'Eva Smith',
+			},
+		]}
+	/>
+);
+
+SingleByline.story = { name: 'Byline with single contributor tag' };
+
+export const MultipleByline = () => (
+	<Byline
+		byline="Eva Smith and Duncan Campbell"
+		guardianBaseURL={guardianBaseURL}
+		tags={[
+			{
+				id: 'eva-smith',
+				type: 'Contributor',
+				title: 'Eva Smith',
+			},
+			{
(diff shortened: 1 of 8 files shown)


oliverlloyd:
Does this need replicating in [this interesting function](https://github.com/guardian/dotcom-rendering/blob/2634175d00566f1907701a2e08bad6d8e4f11682/src/web/components/BylineLink.tsx#L58) for web articles as well as amp?

oliverlloyd:
Random aside: I don't know of a specific reason why bylines in Amp are handled so differently to Web - it's possible there is none and we could merge the two?

Bylines are famously tricky so any work to do this would be fiddly (but fun!) so it would need care but the benefits in the long term would be substantial. Ideally we would end up with one canonical Byline that lives in Source and could be used by all platforms.

jonathonherbert:
Thanks @oliverlloyd! You were absolutely right – I'm just testing a web version now. 

Re: complexity – this deduplication step happens after the tokenisation, so my hope is that by avoid that edge-casey code, we should have to solve just the relatively simple task of ensuring that no contributor tag is used twice in a byline 🤞

nitro-marky (approved):
Looks good!

rcrphillips (approved):
LGTM

jonathonherbert:
Thanks for the approval! Assuming it's OK for me to manually approve the UI elements that have been added as a result of additions to storybook – lmk if that's not the case.

PR #3199, opened 2021-07-20
Record Message Picker timings for the Braze Epic

tjmw:
## What does this change?

This PR adds the capability to record time taken to get an answer for the `canShow` method of message picker candidates. The functionality is configurable per candidate. This PR also enables this for the Braze Epic. Data is recorded via the Ophan tracker and is available to query in `pageview.interactions`. The `component` will be `messagePicker-canShow-<candidate ID>`, in this case `messagePicker-canShow-braze-epic`. The value is the time taken in milliseconds.

## Why?

We want to gather data on how quickly we're typically able to know when there's a Braze epic message to show with a view to potentially reducing the timeout we wrap around this functionality (we're planning in some scenarios to move Braze higher in the epic message candidate priority list, and the timeout potentially blocks lower priority candidates from showing when there isn't a Braze epic message).

### DIFF
--- a/src/web/components/SlotBodyEnd/SlotBodyEnd.tsx
+++ b/src/web/components/SlotBodyEnd/SlotBodyEnd.tsx
@@ -89,6 +89,7 @@ const buildBrazeEpicConfig = (
 			),
 		},
 		timeoutMillis: 5000,
+		reportTiming: true,
 	};
 };
 
--- a/src/web/lib/messagePicker.test.tsx
+++ b/src/web/lib/messagePicker.test.tsx
@@ -15,6 +15,7 @@ jest.useFakeTimers();
 const flushPromises = () => new Promise(setImmediate);
 afterEach(async () => {
 	await flushPromises();
+	jest.clearAllMocks();
 });
 
 describe('pickMessage', () => {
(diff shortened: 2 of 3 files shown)


oliverlloyd:
Just a passing comment and not at all a blocker on this particular PR but I think this is relevant to a larger discussion I've been having around how banners are displayed.

We've got an OKR in Identity to look at issues with the SignInGate showing and I suspect others are hitting similar issues, possibly to do with what seem to be race conditions around different bits of state changing. This PR seems to suggest you might also have this or a similar problem?

I don't have any solutions but my feeling is there needs to be some soft of centralised state control mechanism for this? What do you think?

tjmw:
Hi @oliverlloyd thanks for the comment 👍

> possibly to do with what seem to be race conditions around different bits of state changing

This sounds interesting, I'd be keen to learn more. It doesn't sounds quite like the challenges we've faced, but I still think it'd be worth sharing ideas on this.

Two things to mention which are (at least) tangentially related! 1. We've started holding a regular (monthly) session with developers from contributions, subs and TX to chat about rendering messages on the web platforms to share pain points/solutions/interesting things we're working on. Would identity like to be involved? 2. The sign in gate isn't part of the message picker/candidate configuration, it's kind of its own thing on DCR afaict. I wonder if it should be? Would it be useful to prioritize/play off against the other business units wanting to show messages, or is it okay to show something like the gate along side other banner/epic messages?

oliverlloyd:
> is it okay to show something like the gate along side other banner/epic messages?

That's the issue. One of our bugs was that the gate would appear at the same time as a banner. So a workaround was implemented where there's now a sort of sideways communication between the gate and banners using a custom hook - this works, but could be better

> Would identity like to be involved?
Yes and no. We care about the gate showing properly but we're tied up with migration work so we have been talking about pausing this OKR and, really, this feels more like a platform problem? Ideally, DCR would offer some rails for how and when content shows on the page, it would be great to have some kind of top down contract here. @guardian/dotcom-platform what do you think?

tjmw:
> One of our bugs was that the gate would appear at the same time as a banner

The `messagePicker`'s job is to select a single candidate from a prioritised list of candidates all of which are mutually exclusive and vying for a spot on the page. I wonder if it would be possible to integrate the gate with the message picking for other banners?

> Ideally, DCR would offer some rails for how and when content shows on the page, it would be great to have some kind of top down contract here

I think the `messagePicker` interface is the closest thing we have to this right now. I think it works reasonably well when isolated to a single slot, but as you suggested above there isn't any "sideways" communication. Each slot is independent and basically doesn't have knowledge of any others.

tomrf1 (approved):
looks good!

oliverlloyd:
> I think the `messagePicker` interface is the closest thing we have to this right now.

Yeah, this could well be the place to put things. There's a big win in the fact it is already doing this job.

Would it be difficult to migrate the gate into this framework?

tjmw:
Thanks for the ✅  @tomrf1!

PR #3200, opened 2021-07-20
Support sponsorship regionally on AMP

chrislomaxjones:
## What does this change?

This PR fixes a bug where the sponsorship branding that targets regions outside the US does not display on cached AMP pages. This is achieved by including all possible regional sponsorships on the page and using the `amp-iso-country-*` and `amp-geo-group-*` css classes to display the correct sponsorship for the region.

## Why?

Googles build the AMP cache based on the US version of the site, so cached AMP pages always target the US edition. This means that sponsorships that target regions outside the US were not displayed on these cached pages. By using AMP geo classes we can select which regional branding to show.

### DIFF
--- a/src/amp/components/topMeta/Branding.tsx
+++ b/src/amp/components/topMeta/Branding.tsx
@@ -48,7 +48,7 @@ const brandingLogoStyle = css`
 	padding: 10px 0;
 `;
 
-export const Branding: React.FC<{
+export const BrandingItem: React.FC<{
 	branding: Branding;
 	pillar: Theme;
 }> = ({ branding, pillar }) => {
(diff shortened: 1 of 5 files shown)


src/amp/components/topMeta/Branding.tsx
```diff
@@ -61,3 +61,56 @@ export const Branding: React.FC<{
...
+		.amp-geo-no-group & {
+			display: block;
+		}
+	`;
+	const editionStyles: Record<Edition, SerializedStyles> = {
+		UK: ukStyles,
+		US: usStyles,
+		AU: auStyles,
+		INT: intStyles,
+	};
```
mxdvl, line 98: This is pretty generic stuff, and it would be good to see them extracted to a single file, or even a wrapper component that generates all the variations automatically.

Most of these classes (except for `RoW`) are identical to: https://github.com/guardian/dotcom-rendering/blob/bc8c9c75aa3556fdf0523576f37d5035eaaa2c26/src/amp/components/Ad.tsx#L106-L135

mxdvl (approved):
Good stuff! Some comments below, but not a blocker.

It’s quite confusing that the `AdRegion` and `Edition` types don’t fully overlap, otherwise there would have been more overlap with the `Ad.tsx` logic.

OllysCoding (approved):
Looks good!

src/amp/components/topMeta/Branding.tsx
chrislomaxjones: Really good point, if its not a blocker I'll have a think about the best way to extract these classes (or create a wrapper component) and raise a follow-up PR with those changes.

PR #3202, opened 2021-07-21
Fix missing fields on RichLinkComponent

OllysCoding:
## What does this change?
Quick fix for no `fields` on imageAsset, further investigation should be carried out to ensure that we're more forgiving for missing data

## Why?
Client side error was blocking hydration:

image

### Before
image

### After
image


nitro-marky (approved):
Cool, this should be good enough until we can put in a more robust checking solution in place.

nicl (approved):

PR #3204, opened 2021-07-22
remove js as a required property for InteractiveAtomBlockElement

DavidLawes:
## What does this change?
Some legacy interactives of type `model.dotcomrendering.pageElements.InteractiveAtomBlockElement` did not contain any js in the element object returned from capi. The result was a schema error in function validateAsCAPIType and error `TypeError: Unable to validate request body for url...`. This PR relaxes the schema slightly - making property `js` non-mandatory InteractiveAtomBlockElements.

NB: it looks like atoms-rendering already accounts for missing js in its type declaration https://github.com/guardian/atoms-rendering/blob/main/src/InteractiveLayoutAtom.tsx#L13, so i figured this change was ok.

## Why?
Support migration of legacy interactives.

### Before
Page fails to load (500 error)
Screenshot 2021-07-22 at 14 44 55

### After
Interactive loads (there are some styling/layout issues, but i think this is acceptable for now)

Screenshot 2021-07-22 at 14 47 44


OllysCoding:
Is there any other types that need to be updated or does it just work as-is?

DavidLawes:
> Is there any other types that need to be updated or does it just work as-is?

great question - i should update content.d.ts, although no warnings when rendering page locally :)

src/web/components/AnniversaryAtomComponent.tsx
```diff
@@ -48,14 +48,14 @@ export const InlineAnniversaryAtom = ({
 }: {
 	html: string | undefined;
 	atomCss: string | undefined;
-	js: string;
+	js: string | undefined;
```
nicl, line 51: `js?:` is a nice way of doing this.

nicl (approved):
🥇

PR #3206, opened 2021-07-26
Add regional support to paid for branding

chrislomaxjones:
## What does this change?

As a follow-up to [#3200](https://github.com/guardian/dotcom-rendering/pull/3200), add regional support to "Paid For By" branding components, fixing a bug where cached AMP pages only display US branding. This is achieved by pulling out the required logic for adding correct region classes into a `BrandingRegionContainer` that handles this logic of applying the correct region classes.

## Why?

Paid For By branding suffers from the same problem as sponsorship branding, with AMP pages being cached in the US. The logic that is used for applying the classes to each region for sponsorship branding can be used for Paid For By branding, hence a common container component that can wrap both.

### DIFF
--- a/src/amp/components/topMeta/Branding.tsx
+++ b/src/amp/components/topMeta/Branding.tsx
@@ -3,6 +3,7 @@ import { css, SerializedStyles } from '@emotion/react';
 import { textSans } from '@guardian/src-foundations/typography';
 
 import { pillarPalette_DO_NOT_USE, neutralBorder } from '@root/src/lib/pillars';
+import { isEdition } from '@root/src/amp/lib/edition';
 
 const LinkStyle = (pillar: Theme) => css`
 	a {
@@ -98,19 +99,16 @@ export const BrandingRegionContainer: React.FC<{
 	};
 	return (
 		<>
-			{Object.keys(commercialProperties).map((editionId) => {
-				const { branding } = commercialProperties[editionId as Edition];
-				return branding !== undefined ? (
-					<div
-						css={[
-							brandingStyles,
-							editionStyles[editionId as Edition],
-						]}
-					>
-						{children(branding)}
-					</div>
-				) : null;
-			})}
+			{Object.keys(commercialProperties)
+				.filter(isEdition)
+				.map((editionId) => {
+					const { branding } = commercialProperties[editionId];
+					return branding !== undefined ? (
+						<div css={[brandingStyles, editionStyles[editionId]]}>
+							{children(branding)}
+						</div>
+					) : null;
+				})}
 		</>
 	);
 };
--- a/src/amp/components/topMeta/Branding.tsx
+++ b/src/amp/components/topMeta/Branding.tsx
@@ -34,7 +34,7 @@ const brandingLogoStyle = css`
 	padding: 10px 0;
 `;
 
-const BrandingItem: React.FC<{
+export const Branding: React.FC<{
 	branding: Branding;
 	pillar: Theme;
 }> = ({ branding, pillar }) => {
@@ -62,10 +62,10 @@ const BrandingItem: React.FC<{
 	);
 };
 
-export const Branding: React.FC<{
+export const BrandingRegionContainer: React.FC<{
+	children: (branding: Branding) => React.ReactNode;
 	commercialProperties: CommercialProperties;
-	pillar: Theme;
-}> = ({ commercialProperties, pillar }) => {
+}> = ({ children, commercialProperties }) => {
 	const brandingStyles = css`
 		display: none;
 	`;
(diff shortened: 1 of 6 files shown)


nitro-marky:
Looks good, one recommendation from me is to add some Stories (if possible for these changes) to get a visual idea of any changes.

chrislomaxjones:
> Looks good, one recommendation from me is to add some Stories (if possible for these changes) to get a visual idea of any changes (if there are any!). 😄

Thanks for the feedback! In this instance I'm trying to keep the visuals the same as before and fix a bug where the sponsorship isn't displayed in Google AMP Cache, something I assume difficult to capture in Storybook.

mxdvl:
> Thanks for the feedback! In this instance I'm trying to keep the visuals the same as before and fix a bug where the sponsorship isn't displayed in Google AMP Cache, something I assume difficult to capture in Storybook.

You could force a local geo class on the body?

src/amp/components/topMeta/Branding.tsx
```diff
@@ -107,7 +107,7 @@ export const Branding: React.FC<{
 							editionStyles[editionId as Edition],
```
mxdvl, line 107: if you filtered above, you might be able to remove the `as`.

Something like:
```ts
const isEdition(e: string): e is Edition => Edition.includes(e)
Object.keys(commercialProperties).filter(isEdition).map((editionId)
```

src/amp/lib/edition.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Determine if a string is one of the permitted edition strings
+ *
+ * @param s The string to test
+ */
+export function isEdition(s: string): s is Edition {
+	return ['UK', 'US', 'INT', 'AU'].includes(s);
+}
```
chrislomaxjones, line 8: Is a file like this the best place for `isEdition`?

src/amp/lib/edition.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Determine if a string is one of the permitted edition strings
+ *
+ * @param s The string to test
+ */
+export function isEdition(s: string): s is Edition {
+	return ['UK', 'US', 'INT', 'AU'].includes(s);
```
mxdvl, line 7: If we were to use an `Enum` for Edition in [index.d.ts#L160][enum], we could make this [more independent][ts].

```ts
enum Edition {
    UK = 'UK',
    US = 'US',
    INT = 'INT',
    AU = 'AU',
}

/**
 * Determine if a string is one of the permitted edition strings
 *
 * @param s The string to test
 */
export function isEdition(s: string): s is Edition {
	return s in Edition;
}

const validEdition = ['A','B', 'C', 'UK', 'INT'].filter(isEdition);
```

Maybe one for a subsequent PR, though, as it might have an impact on Dotcom.

[ts]: https://www.typescriptlang.org/play?#code/KYOwrgtgBAogJgSwC4IPYigbwLACgoFQCqA0lALxQDkpVANHocQMoXVHP2OECSAcgBU2VfgK75CAQSLDp4gL548AegBUqxqqgARYEmAAnCAhDAoCAGZQAhlADOSAyYDm5u1HRnUVpAAszAA6Gxkj6cFDAiCjo9o4udpqaUAACAdYG1tDuAv6xTiCuSKhQ+g6aynjAAB4BqAZIUBZgIADG0RgIdvDIaCAAFHYAXHkuAJTD7p2wUb1YeACQBnpgBhiTGN3tANx4irh4LegOUABu1gA2CHCbs5QA2lSS9FQAQvTUAMLvNCTfolQAXQAdBYEOd9AY+p0buhRlsgA
[enum]: https://github.com/guardian/dotcom-rendering/blob/5386ee021368b8aaf45bf52d86a6b792125d7000/index.d.ts#L160

mxdvl: Maybe `as const` is better, actually: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions ?

chrislomaxjones: > You could force a local geo class on the body?

That's a good point but it looks like we currently don't support AMP elements in Storybook (we'd probably need to use an [addon like this](https://storybook.js.org/addons/storybook-amp). This feels a bit out of scope for this PR but may be useful to install it in a separate PR.

chrislomaxjones:
> You could force a local geo class on the body?

That's a good point, but it looks like we currently don't support AMP elements in Storybook (we'd probably need to use an [addon like this](https://storybook.js.org/addons/storybook-amp). This feels a bit out of scope for this PR but may be useful to install it in a separate one.

src/amp/lib/edition.ts
chrislomaxjones: Very good point, I just had a play around with changing it to an enum and it looks like enums don't behave exactly as expected when declared in .d.ts files, I think possibly another one for another PR.

mxdvl (approved):
LGTM

PR #3207, opened 2021-07-26
Bump CWV To 2.1.0

nitro-marky:
## What does this change?

Updates CWV to 2.1.0 which purports to have a breaking change, though we do not use this part of the code, therefore it should have no effect on us. Either way the data coming in will be monitored to make sure it's as expected.

## Why?

It changes the way CLS is recorded which should hopefully see some improvements in our data.


nicl (approved):
🦜 nice!

PR #3208, opened 2021-07-26
make skip to main content smaller

HarryFischer:
## What does this change?
Makes Skip to main content button smaller

## Why?
Too big before

### Before
image

### After
image


OllysCoding (approved):
Looks good!

PR #3209, opened 2021-07-26
Use Braze components for epic messages

michaelclapham:
## What does this change?
Changes Braze epic integration in multiple ways. Instead of lazy loading contributions epic when showing an Epic via a braze message, it instead loads a version of the epic from braze-components.

This PR also changes the interface with braze components by using a different component factory for banners and end of article components, so that the props for each can be more relevant.

## Why?
This change enables new Epics to be delivered via Braze components, including upcoming newsletter epics.

### Before
BrazeEpic in DCR would lazy load contributions epic built from here https://github.com/guardian/support-dotcom-components/blob/main/src/components/modules/epics/ContributionsEpic.tsx

All components (which used to be all banners) were initialised with one Braze components factory.

### After
BrazeEpic in DCR now loads Epic's from Braze-components.


tjmw:
Hi @guardian/guardian-frontend-team 👋  I noticed that we've caused a couple of warnings because of the size increase of `guardian-braze-components` in the `compressed-size-action` report above. Does the warning show above a certain % increase? Or an absolute size increase? (Or both?) In the meantime, we're going to try out some ideas to split that bundle and load things in a more granular way.

oliverlloyd:
> Does the warning show above a certain % increase? Or an absolute size increase?

I believe it's percentage, hence the panic icon!

4Kb is still a hefty bump though. Is this lib already lazy loaded or is that something we can look into? (See [this doc](https://github.com/guardian/dotcom-rendering/blob/44bf079a459f5c29a5c3a03d946bbd28dd5df65a/docs/development/loadable-components.md) and other examples in App.tsx)

tjmw:
> Is this lib already lazy loaded or is that something we can look into?

Great question. Yep it's already lazy loaded. I think we can do better and be more granular in _what_ we lazy load in 
 the different contexts we use this particular bundle, we're currently investigating some possibilities.

tjmw:
We've made a change to `@guardian/braze-components` so it's now possible to load _only_ epics or banners via the `/end-of-article` and `/banner` entry points. That's mitigated somewhat the 20% increase to `dist/guardian-braze-components.js` which we were seeing before. That bundle no longer exists and instead there are two new pairs of separate bundles (hidden by default in the `compressed-size-action` report above:

```
dist/guardian-braze-components-banner.js | 15.1 kB | 0 B |  
dist/guardian-braze-components-banner.legacy.js | 15.3 kB | 0 B |  
dist/guardian-braze-components-end-of-article.js | 10.2 kB | 0 B |  
dist/guardian-braze-components-end-of-article.legacy.js | 10.3 kB | 0 B
```

src/web/components/SlotBodyEnd/BrazeEpic.tsx
```diff
@@ -180,48 +92,64 @@ const BrazeEpic = ({
...
+	if (meta.dataFromBraze) {
+		return (
+			<div ref={setNode} css={wrapperMargins}>
+				<div ref={epicRef}>
+					<BrazeComponent
+						componentName={meta.dataFromBraze.componentName}
+						brazeMessageProps={meta.dataFromBraze}
+						// TODO: Add real implementation for subscribeToNewsletter
+						subscribeToNewsletter={() => {
+							return Math.random() > 0.5
```
arausuy, line 111: Do we want to keep this in or just have it return `Promise.resolve()`

tjmw: Ah great question! This is a placeholder until we add the real `subscribeToNewsletter` functionality, which we'll need to do before configuring any of the newsletter epics in prod. I think this was originally just a `Promise.resolve()`, @michaelclapham do you feel strongly about keeping it this way? I think I'm happy to leave as is, or to revert back to simply returning `Promise.resolve()`.

michaelclapham: Oh yes let's change this back to Promise.resolve()

tjmw: 👍  I've gone ahead and reverted to returning `Promise.resolve()`

arausuy (approved):

PR #3210, opened 2021-07-26
Skip To Navigation

oliverlloyd:
## What?
This PR adds a new 'skip to' link at the top of the page. In addition to the existing (and very nicely implemented 👏  @jlieb10 ) `Skip to main content` link, I also added `Skip to navigation`

## Why?
You ever go to the site with the intended purpose of using the navigation? I do, and I think a lot of our readers do too. But right now keyboard users have to fight their way through a lot of link jungle to find the navigation control, this removes that jungle and serves up navigation as a primary destination.

I also think keyboard navigation is for everyone, not just people who might be partially sighted. Anyone who has ever worked as a data entry clerk (me) knows that a mouse is a horribly inefficient tool for interfacing with a computer and a keyboard is a far superior thing. By enhancing how our site responds to keyboard input we make better use of the keyboard.

## What else?
I abstracted things out into `SkipTo`. I'm famously wary of creating abstractions but I felt this one was justified because it - I hope - makes the code easier to understand.

I also refactored some of the area labels so that the navigation text given by VoiceOver was more concise and useful.

## How did I test this?
In addition to using my keyboard to check things, I also turned on [VoiceOver](https://support.apple.com/en-gb/guide/voiceover/welcome/mac) and made sure that the spoken text made sense.

### Skip to Navigation

Props to @jlieb10 and @liywjl for making these areas accessible


package.json
```diff
@@ -56,6 +56,7 @@
     "@guardian/src-ed-lines": "^3.6.0-rc.0",
     "@guardian/src-foundations": "^3.6.0",
     "@guardian/src-icons": "^3.7.0",
+    "@guardian/src-layout": "^3.8.0",
```
oliverlloyd, line 59: I've added the version of `Hide` from Source here. DCR has it's own `Hide` but I've been meaning to add this shared one for _ages_, the api of @SiAdcock version is so much better than mine 😊

src/web/components/SkipToMainContent.tsx
```diff
@@ -0,0 +1,5 @@
+import { SkipTo } from '@root/src/web/components/SkipTo';
+
+export const SkipToMainContent = () => {
+	return <SkipTo id="#maincontent" label="Skip to main content" />;
+};
```
oliverlloyd, line 5: This is really just putting a name on things but I quite like that

SiAdcock:
This is so awesome, thanks for proposing this @oliverlloyd and your implementation is amazing as ever. A couple of questions that shouldn’t block any merge:

1. Is the pillar menu part of the navigation, do you think? What do you think about focussing the whole menu bar?
2. If we go with focussing the “show more” menu, do you think instead this skip link should expand the menu and focus it, rather than focus the button that expands the menu?

oliverlloyd:
> * Is the pillar menu part of the navigation, do you think? What do you think about focussing the whole menu bar?

This is a really good point. In my mind the navigation 'control' is this button but actually the most useful place to put focus is the start of the Pillar nav. I'll look at doing this 👍 

> * If we go with focussing the “show more” menu, do you think instead this skip link should expand the menu and focus it, rather than focus the button that expands the menu?

Moot now, but one downside to auto-opening, is then people traversing the dom might not understand that they are 'inside' a thing and not know they can escape out of it to jump back into the main semantic flow. Not sure how much of an issue this would be though

mchv:
Thanks for doing this @oliverlloyd 💎

oliverlloyd:
> * Is the pillar menu part of the navigation, do you think? What do you think about focussing the whole menu bar?

This is now done. And it works really well, thanks for thinking of this! It also made me question the aria labels we're using. These were pretty good but in some places there was some duplication so I removed those to try to create a better flow

src/web/components/Nav/Nav.tsx
```diff
@@ -166,8 +166,6 @@ export const Nav = ({ format, nav, subscribeUrl, edition }: Props) => {
 					rowStyles,
 					format.display === Display.Immersive && minHeight,
 				]}
-				role="navigation"
-				aria-label="Guardian sections"
```
oliverlloyd, line 170: I'm deleting these because the child components on `nav` (the ones we land on when tabbing) already have their own labels so having these as well means both are spoken

src/web/components/Pillars.tsx
```diff
@@ -263,6 +263,8 @@ export const Pillars: React.FC<{
 							isSelected && forceUnderline,
 							showDivider && pillarDivider,
 						]}
+						id={i === 0 ? 'navigation' : undefined}
+						aria-label={i === 0 ? 'Navigation' : undefined}
```
oliverlloyd, line 267: We tell the reader they are at the start of the navigation section using this label

oliverlloyd:
Ha, on one hand I love that we fail PRs for accessibility, on the other...
Screenshot 2021-07-27 at 09 22 32

I guess this is because it wants the `role` - I'll add that back

oliverlloyd:
> Ha, on one hand I love that we fail PRs for accessibility, on the other...

I take it back. I love this check. It was spot on to fail this PR because I was trying to insert two duplicate `id`s (we put the `Nav` component on the page twice).

We now get a perfect score from Chrome 😊 
Screenshot 2021-07-27 at 10 07 38

nicl (approved):
This is great. Thanks for thinking this through and also @SiAdcock for the helpful comments.

liywjl (approved):
Nice work! 👍

PR #3211, opened 2021-07-27
extend legacy interactive figure classes

DavidLawes:
Some legacy interactive figures require additional classes in order for the page to correctly render, this change adds the known missing classes.

## What does this change?
Adds more classes on legacy interactive figures: `element` and `interactive`

## Why?
Some legacy interactive immersives rely on classes to render. Without these extra classes legacy interactive immersives cannot be rendered via dcr.

### Before
Interactive does not render via dcr:
Screenshot 2021-07-27 at 16 41 16

### After
Interactive renders via dcr:
Screenshot 2021-07-27 at 16 43 07


rcrphillips (approved):

nicl (approved):
Nice. The `Map([...])` way to declare things confused me - and it's my code 🤦 - but the fix looks good.

DavidLawes:
[tone/news](https://trello.com/c/xBvHTSdI/226-tone-news)

PR #3212, opened 2021-07-28
Stop recording timing data for Braze epic canShow

tjmw:
## What does this change?

In #3199 we added the option to record canShow timings to the messagePicker and enabled this option for the Braze epic. This PR turns off the recording.

## Why?

We've got several days worth of data to look at now which is plenty.


michaelclapham (approved):

tjmw:
Thanks @michaelclapham!

PR #3213, opened 2021-07-28
Move AMP regional classes out of components

chrislomaxjones:
## What does this change?

Pull the AMP classes used to render conditionally in different regions out of their definitions within components and share them across those that use them. Currently in AMP these components are branding and ads.

## Why?

Region classes are used across branding and ad components. Therefore it makes sense to pull these out into common constants that can be used by any components that need to be rendered regionally, or reused in the future by new components that need to display conditionally in regions.


mxdvl (approved):
This is a nice simplification!

PR #3215, opened 2021-07-29
ROTA: Broken embeds due to code not compatible with loadable components

jlieb10:
## Why?
Loadable components created a race condition for UnsafeEmbed type embeds on Dotcom, that was causing the height adjustment of the first 'loaded' embed to fail, so we rolled back this change for this type of component only. Would like to investigate further during health week.

### Before
before

### After
AFTER

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -843,6 +843,11 @@ interface VineBlockLoadable extends ComponentNameChunkMap {
     addWhen: VineBlockElement['_type'];
 }
 
+interface InstagramBlockLoadable extends ComponentNameChunkMap {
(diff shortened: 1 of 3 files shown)


rcrphillips (approved):

lucymonie (approved):
This looks fine to me 👍🏻

DavidLawes:
[[ROTA] Broken embeds due to code not compatible with loadable components](https://trello.com/c/jSUK2g7P/261-rota-broken-embeds-due-to-code-not-compatible-with-loadable-components)

PR #3217, opened 2021-08-02
AdBlock detection for all

oliverlloyd:
## What?
This PR takes the function `isAdBlockInUse` which is used as part of the Sentry loader script, and hoists it up to a shared lib so it can be used elsewhere

In addition, I've added two helper utilities to make consuming this feature easier:

1. `useAdBlockInUse` - A custom hook to abstract using React state and useEffect to manage ad block detection
2. `WhenAdBlockInUse` - A component that uses the `useAdBlockInUse` custom hook and allows the use of composition to only render content when an ad blocker is being used

## Why?
Having the ability to know when an ad blocker is in use unlocks some potential new features and allows us to address a parity issue where Frontend shows a contribution banner in the top right ad slot which an ad blocker is detected, which DCR doesn't do


nitro-marky (approved):
Nice!

PR #3218, opened 2021-08-02
Run DCR locally with more memory

oliverlloyd:
## What?
Tells `make dev` to run `node` with 8GB of memory

## Why?
Because I keep getting out of memory errors when running locally

### DIFF
--- a/makefile
+++ b/makefile
@@ -58,7 +58,7 @@ run: stop build start
 
 dev: clear clean-dist install
 	$(call log, "starting frontend DEV server")
-	@NODE_ENV=development nodemon scripts/frontend/dev-server -- --max-old-space-size=7000
+	@NODE_ENV=development nodemon scripts/frontend/dev-server -- --max-old-space-size=7168
 
 # tests #####################################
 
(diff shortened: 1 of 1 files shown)


oliverlloyd:
> Might be worth running at 7GB to leave some memory free for background operations.

Isn't there a law of the universe that memory allocations must be 1024, 2048, 4096 or 8196?

I have _*no*_ idea why this 🤷

nicl (approved):

nitro-marky:
> > Might be worth running at 7GB to leave some memory free for background operations.
> 
> Isn't there a law of the universe that memory allocations must be 1024, 2048, 4096 or 8196?
> 
> I have __no__ idea why this 🤷

It's just that in their docs they suggest for a 2GB machine setting the allocation at 1.5GB, so running at 6 or 7 should be fine I'd have thought.

nitro-marky (approved):
Nice!

nitro-marky (changes requested):
I think for 7gb it should be 7,168

nitro-marky (approved):
Mega!

PR #3219, opened 2021-08-02
Add Identifier enum for `id` attributes

mxdvl:
## What does this change?

Convert the `id` attributes to using an enum. The [IDs should be globally unique on a page][id], so mapping them in an enum  provides type safety and semantically links them all together.

[id]: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

## Why?

I saw #3210 and thought it was brilliant!

### DIFF
--- a/src/web/components/ArticleBody.tsx
+++ b/src/web/components/ArticleBody.tsx
@@ -7,6 +7,7 @@ import { LiveBlogRenderer } from '@root/src/web/lib/LiveBlogRenderer';
 import { Design, Display } from '@guardian/types';
 import type { Format } from '@guardian/types';
 import { space } from '@guardian/src-foundations';
+import { Identifier } from './SkipTo';
 
 type Props = {
 	format: Format;
@@ -93,7 +94,7 @@ export const ArticleBody = ({
 		<div
 			// eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
 			tabIndex={0}
-			id="maincontent"
+			id={Identifier.MainContent}
 			css={[
 				isInteractive ? null : bodyPadding,
 				globalH2Styles(format.display),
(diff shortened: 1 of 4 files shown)


oliverlloyd (approved):
I think using enums to replace strings offers a number of benefits:

- It prevents typos
- It allows you to see in one place all the possible values
- It means code that clicks on links will always connect to that link
- If you change a link it will change in both places at once

But there are also costs to these types of abstractions and we should weigh them here as well

- Future developers reading this code won't immediately see what the code is doing, they need to open another file
- Searches of the codebase won't work as well
- Adding or editing new links is more work
- Its more complicated code

My personal preference is to avoid abstractions where possible. Each individual abstraction makes sense in isolation, esp. when you create it, but to a person who is not familiar with that abstraction, and when there are more and more of them, it creates a large burden and can make it very hard to work on code. I typically give up trying to read code at about 3 levels of abstraction!

But I know that these types of changes and opinions around them are subjective and that there's no right or wrong - just feelings and ideas based on individual experience.

What to do here? 🤷 I think the trick in these situations is to weigh the pros and cons above against how the code will be used:

If there are going to be lots of changes in future having strong types has more benefit. If it's likely to be fairly static then we should favour readability?

All that said, I'm happy to approve the change 😻

mxdvl:
Thanks @oliverlloyd for your comments. Not being a regular contributor on this codebase, and because there are only two unique `id`, I don’t feel like I should force this abstraction yet.

If there are more than half a dozen, it might start to make sense.

For the sake of removing abstractions, SkipTo should not take an `id` which starts with a `#`. Either it adds the hash itself, or takes a straight `{ href, label }`.

## `id`

```ts
type Props = {
	id: string;
	label: string;
};

export const SkipTo = ({ id, label }: Props) => {
	return (
		<a
			href={`#${id}`}
            // […]
```

## `href`

```ts
type Props = {
	href: string;
	label: string;
};

export const SkipTo = ({ href, label }: Props) => {
	return (
		<a
			href={href}`
            // […]
```

oliverlloyd:
> SkipTo should not take an `id` which starts with a `#`. Either it adds the hash itself, or takes a straight `{ href, label }`.

Totally agree, this is a great idea and would really improve things. Requiring the hash defo feels a bit confusing

mxdvl:
Closed in favour of #3221

PR #3220, opened 2021-08-02
Shady Pie in DCR

oliverlloyd:
## What?
Here we show a static image displaying a contribution ask in place of the top right slot when we detect that an ad blocker is being used. Apparently this is called a 'Shady Pie', I have no idea why this is.

### Before
Screenshot 2021-08-02 at 11 07 18

### After
Screenshot 2021-08-02 at 11 06 54

## Why?
Because we implemented this on Frontend [at the start of 2019](https://github.com/guardian/frontend/pull/21138) and there's been [a ticket](https://trello.com/c/a3gtEQEm/281-contribution-ad-shows-when-ad-blocker-enabled) in the DCR backlog to do this for a while

Ultimately, a contribution ask is better than whitespace. Almost anything is better than whitespace.

## How is this slot different to Frontend?
It's sticky. In DCR we stick the top right slot by default and so this contribution ask is stuck as well. We could remove this stickiness but I think keeping it makes more sense


src/web/components/ContributionSlot.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export const ContributionSlot = () => {
+	return (
+		<div
+			css={css`
+				position: sticky;
+				top: 0;
+			`}
+		>
+			<a href="https://support.theguardian.com/uk/subscribe/digital?acquisitionData=%7B%22componentType%22%3A%22ACQUISITIONS_OTHER%22%2C%22source%22%3A%22GUARDIAN_WEB%22%2C%22campaignCode%22%3A%22shady_pie_open_2019%22%2C%22componentId%22%3A%22shady_pie_open_2019%22%7D&INTCMP=shady_pie_open_2019">
```
mxdvl, line 11: It might be worth using [`URLSearchParams`][mdn] here to make this more readable?

[mdn]: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

ashishpuliyel, line 11: Any chance we can change the params here to reflect that this is a conversion from this, new, component?

mxdvl (approved):
Looks good from a commercial perspective. I think maybe @rupertbates, @i-hardy and @marialani might be able to chime in, too!

ashishpuliyel (approved):
Thanks, happy to see this coming back.
Having said that, I wonder how many people actually click on it.

src/web/components/ContributionSlot.tsx
mxdvl:
```js
const params = new URLSearchParams();
params.set(
	'acquisitionData',
	JSON.stringify({
		componentType: 'ACQUISITIONS_OTHER',
		source: 'GUARDIAN_WEB',
		campaignCode: 'shady_pie_open_2019',
		componentId: 'shady_pie_open_2019',
	}),
);
params.set('INTCMP', 'shady_pie_open_2019');

// you can then use:

params.toString()
```

mxdvl: Fixed upstream: https://github.com/guardian/frontend/pull/24050/commits/9dc5851e17fd792f9e825ba888fb7df6d4c5899e

oliverlloyd: Without a doubt, that code is easier to read and understand the url. But it's harder to verify that the url works and maybe harder to change it? When it's simply a string I can just hold down the CMD key on my laptop and click it and, yep, that works...

Something else to consider is if we keep it raw then we make it possible for non devs to change this code - something I think we should invite more of in general but is perhaps useful in this context?

To be honest I'm not sure, `URLSearchParams ` is a far superior way for managing urls... 🤷 this really depends on how this code will be used going forwards.

marialani: Another note on `URLSearchParams` is that it's [not supported in internet explorer](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#browser_compatibility)

rupertbates: If you mean that maybe we should update the tracking params to distinguish this component from the one in frontend then I'm currently checking with the optimisation team whether they would like this or not. Will update here once I get an answer.

mxdvl: If you click this URL in your IDE, you’ll create a fake event, so not sure it’s really an issue anyways? This gets expanded on the page, too.

You can even share a [Typescript Play][ts] example for people who don’t want to open the console.

[ts]: https://www.typescriptlang.org/play?#code/MYewdgzgLgBADgQwE4ILYRgXhmApgdxgFUAlAGQGVdlgALABWTQgAoBKAbgFgAoRFdADoIuKC14BIAOQJgARwCuASwhKoS8ABEEUBFIA0kgFIUA8gDlhUJErABzJQDMAniwDekiaFRxwuMFAAKs5wuABcMFIAggDCAIpEAJIUiYGJFhQA+qaBABIAoiQGnhAgCkjA4ZEA4kRRJJqJUeaZAOr5AELFPBJeaIhKdmAxIAAmVVIQtAijzplwSriZIKFgmQBMAAwAjACc3b3evngBiaMRk9Oz84vLqxs7+4Y9AL5sz5y8-MzCoixSiXMgRiAFl6AZIlMZnMFksVv4HnspJ8eLwAPRomDOMowYAIMAwKC0fwwBQiMK8XigSAgAA2uEEtJAdhY3yEUBAFGsthZbDYQA

ashishpuliyel: I was suggesting changing shady_pie_open_2019 to shady_pie_open_2021 for example, so we can see in the recorded conversions how many people actually clicked though this newly implemented component.

rupertbates: Ok, just got a reply back it would be better to leave it the same as it will then feed into all the existing reports and dashboards. From a business point of view the difference between the platforms is not really relevant.

mxdvl: > Another note on `URLSearchParams` is that it's [not supported in internet explorer](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#browser_compatibility)

Are there Ad Blockers in Internet Explorer?

marialani: > Are there Ad Blockers in Internet Explorer? Can we [polyfill](https://github.com/ungap/url-search-params) it?

Believe there are a few like AdblockPlus though catering for IE is probably an edge case here

oliverlloyd: > catering for IE is probably an edge case here

I once suggested we think about not supporting the top right slot on IE because, well, it's just IE... I was surprised to learn that this one little slot just on IE makes quite a lot of money each month!

mxdvl: Is this not rendered server-side anyways? If not, surely we could!

oliverlloyd: > Is this not rendered server-side anyways?

It has to be client side because the ad detection script does some dom manipulation to decide if the browser is using an ad blocker or not.

We _could_ generate the url server side and pass it down via `window` but this would add complexity which we probably don't want?

mxdvl: I checked and polyfill.io has it: https://polyfill.io/v3/polyfill.min.js?features=URLSearchParams

Is this enough to carry on and just implement this in a readable manner?

oliverlloyd: I've made a commit so that this url is now constructed using `URLSearchParams`. I'm still generally more in favour of simpler code but I also see the benefit to this approach so I'm happy to follow the Reviewer Wins philosophy here 😺

PR #3221, opened 2021-08-02
Simplify SkipTo props

mxdvl:
## What does this change?

Clarify identifiers and auto-complete them. Automatically add “#“ in `href`.

## Why?

There’s a potential confusion between `id` and `href`, so this fixes that. Enums have been explored, but rejected in #3219.


oliverlloyd (approved):
🚀

PR #3222, opened 2021-08-02
Update braze components to version 3.3.0

michaelclapham:
## What does this change?
Upgrade braze components to version 3.3.0

## Why?
This fixes some bugs and reduces the bundle size loaded by braze components.

### Before
Braze components was version 3.2.0

### After
Braze components is version 3.3.0


DavidLawes (approved):
👍

PR #3223, opened 2021-08-02
Move webPublicationDate to free up space for new format

nicl:
**Note: Frontend PR needs to go live first. (And the build will break until it does.)**

## What + why?

Replaces `webPublicationDate` with `webPublicationDateDeprecated` so that we can reformat the original field as part of wider DCR date refactor.

See: https://docs.google.com/document/d/1Emc0OivrVhH5-svVOiE2QLBER_wREdIxpe9G1XqMy9g/edit#. 

And part-1 here: https://github.com/guardian/frontend/pull/24054.

### DIFF
--- a/fixtures/generated/articles/Analysis.ts
+++ b/fixtures/generated/articles/Analysis.ts
@@ -1796,7 +1796,7 @@ export const Analysis: CAPIType = {
 			elementId: '2e19fdc4-4c9a-4826-bc94-3cceb9f583cb',
 		},
 	],
-	webPublicationDate: '2020-02-10T12:31:25.000Z',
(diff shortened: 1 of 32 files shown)


DavidLawes (approved):
I'm assuming there are corresponding test files/data that need to be updated? I could've misinterpreted the github actions though

nicl:
> I'm assuming there are corresponding test files/data that need to be updated? I could've misinterpreted the github actions though

@DavidLawes yeah. Unfortunately, some of our tests fetch their data dynamically from PROD (https://github.com/guardian/dotcom-rendering/blob/main/src/model/validate.test.ts) and these failed because the Frontend change isn't live yet. I will seek to modify these tests in a separate PR to not rely on PROD as this feels like a mistake to me.

PR #3224, opened 2021-08-02
Upgrade Storybook to 6.3.6

nitro-marky:
## What does this change?

This updates Storybook to the latest version, as well as the database of browsers. It also adds a Storybook extension called 'turbo-build' which speeds things up a bit.

On your first run the results seem to typically be 40-50 seconds for the manager to build and just over a minute for the preview. Once the manager has been cached the next time storybook is run the average results appear to be around 40 seconds (depending on your system of course).

## Why?

Cos it was slow 🐌 

### After
If you're lucky!

Screenshot 2021-08-02 at 17 04 12


nicl (approved):
Would be great to know how much this speeds things up on your laptop - say averaged over 3 runs or something.

PR #3226, opened 2021-08-03
Enable the TS return types rule for dotcom rendering

lucymonie:
## What does this change?
This enables the `"noImplicitReturns": true` rule in the tsconfig.json, which reduces ambiguity in return types.

## Why?
This is something we've always planed to have but needed to spend some time working on the resulting errors when enabling this feature of Typescript.


src/lib/age-warning.ts
```diff
@@ -40,4 +40,5 @@ export const getAgeWarning = (
 			return message;
```
nicl, line 40: Can this be deleted now?

nicl (commented):
Looks good. What are thougths on commenting the @ts-ignore lines to explain why we are ignoring things (as it's a bit of a sledgehammer)?  Ideally, it would be great if we could avoid @ts-ignore here - for the tests could we add an explicit return type - `(): void => ...` - to help here?

src/web/components/elements/InteractiveContentsBlockElement.tsx
```diff
@@ -175,6 +175,11 @@ export const InteractiveContentsBlockElement = ({
 	// -> If we stop intersecting with the current element, and its in view > show the preview element
 	// -> If we're intersecting with an element set that to the active one
 	// -> If the end element is in view, but the last element is not > show the last element
+	// Because useEffect is used for performing side-effects with an optional cleanup function
+	// it doesn't deal with returning values, so no types are necessary. For more info see:
+	// https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/hooks/#useeffect--uselayouteffect
+	// Given this, it seems reasoable to ask typescript to ignore this.
+	// @ts-ignore
 	useEffect(() => {
```
nicl, line 183: Another option is to type the return type as `any`, which retains Typescript checks but also expresses that the function is a hot mess/returns a range of different things.

src/amp/lib/ad-json.test.ts
```diff
@@ -86,7 +89,9 @@ describe('ampadslots', () => {
 		if (p === undefined) {
 			return fail();
 		}
-		expect(p.value).toBe('4,5,1,2,3');
+		// These expect statements are returned to keep typescript happy
+		// Otherwise it complains that not all code paths return a value
+		return expect(p.value).toBe('4,5,1,2,3');
```
nicl, line 94: Thinking about these cases more, what about just removing the return from the `fail()` calls and using if/else. E.g

```
		if (p === undefined) {
			fail();
		} else {
			expect(p.value).toBe('amp');
		}
```

It would avoid the need for returns (which are not doing anything here as no consumer/jest works by side-effects). And also we could then remove the comments.

```
		if (p === undefined) {
			fail();
		}
		
		// imagine some code here
		expect(p.value).toBe('amp');
```

also works but in the `// imagine...` bit stuff would actually get executed even on a fail, which could be confusing/slow things down. So the if/else feels better here.

nicl (commented):
Made some more suggestions, but let me know what you think!

src/amp/lib/ad-json.test.ts
lucymonie: Sounds good to me 👍🏻

src/web/components/elements/InteractiveContentsBlockElement.tsx
lucymonie: Yes, that works for me, and typescript says 👌🏻

nicl (approved):
🚀

PR #3235, opened 2021-08-03
Fix carousel immersive cards

rcrphillips:
## What does this change?
Fixes the display of immersive carousels, as they were broken on prod. Includes bonus card limit increase on curated containers for @paperboyo

### Before
Screen Shot 2021-08-04 at 00 31 16

### After
Screen Shot 2021-08-04 at 00 37 28

### DIFF
--- a/src/web/components/Card/components/CardAge.tsx
+++ b/src/web/components/Card/components/CardAge.tsx
@@ -1,13 +1,11 @@
 import { css } from '@emotion/react';
 
-import { Design, Display } from '@guardian/types';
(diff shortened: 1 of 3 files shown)


paperboyo:
👏 
Thank you kindly! As I now consider this carousel implementation canonical (looking at you, video container!), it’s great that one can use horizontal scroll. I wonder if for mice without it (and when shift is too far), it would be useful to be able to click and drag the carousel too? Something for UX, maybe?

nicl (approved):

PR #3241, opened 2021-08-04
More Nav, less Sticky

oliverlloyd:
## What does this change?
In some layouts we were keeping the `Nav` visible when sticking the top ad slot, in others we were hiding it behind the slot. This PR brings consistency bu always showing the `Nav`

## Why?
For consistency

## Why decide on showing the `Nav` more instead of sticking more?
Because this is the primary site navigation and by giving it more weight we encourage readers to browse the site more

### DIFF
--- a/src/web/layouts/CommentLayout.tsx
+++ b/src/web/layouts/CommentLayout.tsx
@@ -348,51 +348,49 @@ export const CommentLayout = ({
 							/>
 						</ElementContainer>
 					)}
-
-					<ElementContainer
-						showSideBorders={true}
-						borderColour={brandLine.primary}
-						showTopBorder={false}
-						padded={false}
-						backgroundColour={brandBackground.primary}
-					>
-						<Nav
-							nav={NAV}
-							format={{
-								...format,
-								theme: getCurrentPillar(CAPI),
-							}}
-							subscribeUrl={
-								CAPI.nav.readerRevenueLinks.header.subscribe
-							}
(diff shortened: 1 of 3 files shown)


mxdvl (approved):
me likes

Also: you can put `.mov` files directly in PRs, now! No need for huge GIFs.

Worth noting: I have no more power than @paperboyo to say this is the right behaviour, but it sure _feels_ right

oliverlloyd:
> Also: you can put `.mov` files directly in PRs, now! No need for huge GIFs.

lol - I did _not_ know that 👇 
Screenshot 2021-08-04 at 16 15 53

oliverlloyd:
> Worth noting: I have no more power than @paperboyo to say this is the right behaviour, but it sure _feels_ right

Agreed. This PR is an option if we want to take it. @rcrphillips what say you?

ashishpuliyel:
I'd love to see if we could get an a/b test on this change so see if there's any commercial impact. Especially since I suspect the old frontend behaviour (no sticky nav) is the canonical behaviour.  (Since the ad scrolls away slightly sooner, this affects a metric called 'viewability' which google tracks, and which correlates strongly to our monetisation)

oliverlloyd:
As mentioned in chat, shall we merge this PR as is so that we achieve consistent behaviour over all the different layouts and then, once that is merged and the monorepo work is complete, we can have a second PR that we use for adding an AB test to quantify the impact of sticking above or below the Nav?

PR #3246, opened 2021-08-05
Remove the open Load Balancer Security Group

nitro-marky:
## What does this change?

As per the title it removes the old LoadBalancerSecurityGroup which was open and posed a security risk. The new configuration configures frontend with an internal IP address.

## Why?

🔒


nicl (approved):
🎂

PR #3250, opened 2021-08-05
Stick `MostViewed` when no ads are showing

oliverlloyd:
## What does this change?
This PR is a proposal for making use of the white space that exists in the right column when we are not showing any ads on the page (an ad blocker is in use or the reader is signed in).

Instead of positioning the `MostViewed` content in a static position and allowing it to scroll of the screen, this PR sticks it to the top and keeps it in view

## Why?
Because if we're not showing ads then we are free to use this space

https://user-images.githubusercontent.com/1336821/128374878-fc748914-f27e-48c5-b3a3-7b452a5a6e82.mp4


nitro-marky (approved):
Looks great 👍 . Don't know if @rcrphillips wants to take a look as well.

src/web/components/MostViewed/MostViewedRight/MostViewedRight.tsx
```diff
@@ -13,16 +14,26 @@ const wrapperStyles = css`
 	margin-bottom: 24px;
 `;
 
+// We only stick most viewed when an ads are not showing. We do this
```
oliverlloyd, line 17:
```suggestion
// We only stick most viewed when ads are not showing. We do this
```

oliverlloyd:
> Looks great 👍 . Don't know if @rcrphillips wants to take a look as well.

Rob already checked it over and gave the nod (in chat) the other week

PR #754, opened 2019-09-10
Ad styling

nicl:
## What does this change?

Add styling for ads. Note, there are almost certainly variants of the dynamic in-body ads that need different styling, which aren't yet handled.

## Why?

As part of static ad support.

## Link to supporting Trello card

https://trello.com/c/Ry9am7hl

sticky-ads-dcr

Screenshot 2019-09-10 at 14 37 10

PR #3251, opened 2021-08-06
Stuck sans borders

oliverlloyd:
## What does this change?
On every article page we're showing a 1 pixel grey border under any content that we 'stick' - like the `Header`. Sometimes this slides through without being obvious and other times it's a motorway running across the page

## Why is it there?
It was originally added to the code (for perfectly good reasons) over 2 years ago in [this PR here](https://github.com/guardian/dotcom-rendering/pull/754). But those reasons have - I believe - long passed . Yet this line of css has persisted through a series of refactors and copy pastes (lots of them by me!) - each one faithfully preserving the original code but some point along the way the need for this border was lost and it ended up an extra line of grey pixels taking up space

### Before
Screenshot 2021-08-06 at 09 18 02

### After
Screenshot 2021-08-06 at 09 18 13

### DIFF
--- a/src/web/layouts/lib/stickiness.tsx
+++ b/src/web/layouts/lib/stickiness.tsx
@@ -6,7 +6,8 @@ type Props = {
 	children?: React.ReactNode;
 };
 
-type StuckProps = Props & {
+type StuckProps = {
+	children?: React.ReactNode;
 	zIndex?: string;
 };
 
--- a/src/web/layouts/lib/stickiness.tsx
+++ b/src/web/layouts/lib/stickiness.tsx
@@ -10,16 +10,17 @@ type StuckProps = Props & {
 	zIndex?: string;
 };
 
-// The advert is stuck to the top of the container as we scroll
-// until we hit the bottom of the wrapper that contains
-// the top banner and the header/navigation
-// We apply sticky positioning and z-indexes, the stickAdWrapper and headerWrapper
-// classes are tightly coupled.
-const stickyAdWrapper = (zIndex = 'stickyAdWrapper') => css`
-	background-color: white;
+const stickyStyles = css`
 	position: sticky;
 	top: 0;
+`;
+
+const addZindex = (zIndex = 'stickyAdWrapper') => css`
 	${getZIndex(zIndex)}
+`;
+
+const whiteBackground = css`
+	background-color: white;
 `;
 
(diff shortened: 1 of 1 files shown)


src/web/layouts/lib/stickiness.tsx
```diff
@@ -1,30 +1,29 @@
...
+type StuckProps = {
+	children?: React.ReactNode;
 	zIndex?: string;
 };
 
-// The advert is stuck to the top of the container as we scroll
-// until we hit the bottom of the wrapper that contains
-// the top banner and the header/navigation
-// We apply sticky positioning and z-indexes, the stickAdWrapper and headerWrapper
-// classes are tightly coupled.
```
oliverlloyd, line 19: I deleted these comments because they stopped being true at some point and it's better to have no comments than incorrect ones

src/web/layouts/lib/stickiness.tsx
```diff
@@ -1,30 +1,29 @@
...
-const stickyAdWrapper = (zIndex = 'stickyAdWrapper') => css`
```
oliverlloyd, line 20: This css is used for more than just ads so this name is wrong

oliverlloyd: It's also weird that it's the same as 'stickyAdWrapper', the zIndex name

src/web/layouts/lib/stickiness.tsx
```diff
@@ -1,30 +1,29 @@
...
-	background-color: white;
-	border-bottom: 0.0625rem solid ${border.secondary};
```
oliverlloyd, line 22: _This_ is the phantom border!

src/web/layouts/lib/stickiness.tsx
```diff
@@ -1,30 +1,29 @@
...
 	top: 0;
+`;
+
+const addZindex = (zIndex = 'stickyAdWrapper') => css`
 	${getZIndex(zIndex)}
 `;
 
+const whiteBackground = css`
+	background-color: white;
+`;
```
oliverlloyd, line 25: By breaking unrelated css out into separate chunks we make the code more scalable for future use cases - this kind of approach might have prevented the issue this PR deals with where unrelated css got carried through when it wasn't needed.

rcrphillips (approved):

PR #3252, opened 2021-08-06
Standalone Commercial Bundle

mxdvl:
## What does this change?

Conditionally load a standalone commercial bundle which is identical between `frontend` and `dcr`. 

Required steps before this can go live:
- merge https://github.com/guardian/frontend/pull/24058
- regenerate fixtures

Also narrows `abTests` type and fix two typos.

## Why?

This could be the future.


mxdvl:
This won’t be needed, but the typos and narrow types are available in: #3276

PR #3253, opened 2021-08-06
remove list and listtype from youtube block component

DavidLawes:
## What does this change?
Youtube videos in amp were not playing because of invalid values set against query params in the youtube iframe src. The list and listtype params were incorrect, and are only needed to identify the video if we don't have the
videoId. We believe it's safe to remove the data attributes as we have the videoId (data-videoId).

Refs:
https://developers.google.com/youtube/player_parameters#list
https://amp.dev/documentation/components/amp-youtube/

## Why?
To enable youtube videos to be played in amp.

### Before
Screenshot 2021-08-06 at 10 01 14

### After
Screenshot 2021-08-06 at 10 01 06

### DIFF
--- a/src/amp/components/elements/YoutubeBlockComponent.tsx
+++ b/src/amp/components/elements/YoutubeBlockComponent.tsx
@@ -51,8 +51,6 @@ export const YoutubeBlockComponent: React.FC<{
 	if (element.channelId) {
 		// related videos metadata
 		attributes['data-param-rel'] = '0';
-		attributes['data-param-listType'] = 'playlist';
-		attributes['data-param-list'] = element.channelId;
 	}
 
 	return (


DavidLawes:
[[ROTA] youtube videos not showing in amp](https://trello.com/c/W3VNA5QB/266-rota-youtube-videos-not-showing-in-amp)

src/amp/components/elements/YoutubeBlockComponent.tsx
```diff
@@ -51,8 +51,6 @@ export const YoutubeBlockComponent: React.FC<{
 	if (element.channelId) {
```
DavidLawes, line 51: question whether the related videos data attribute still needs to be inside this if statement, but thought i'd change only what was necessary....

nitro-marky (approved):
Have you tried removing each at a time to see if that makes a difference? Otherwise 👍

DavidLawes:
> Have you tried removing each attribute at a time to see if that makes a difference? Otherwise 👍

Yeah, great shout - the data-param-list seems to be the attribute causing the problem. If that attribute is included the video doesn't play, so I definitely need to remove that. Retaining just data-param-listType seems to have no impact, but when I read the docs (https://developers.google.com/youtube/player_parameters#list) I interpreted it like these two params (list + listType) are used in conjunction. If I remove one of them I think it makes sense to remove the other...

PR #3254, opened 2021-08-06
Add GHGuardianHeadline-BoldItalic font for standfirsts, H2s etc.

paperboyo:
## What does this change?
This is similar to https://github.com/guardian/dotcom-rendering/pull/2636 and the rationale is the same: if fonts are loaded **only** when they are needed, we should always be providing all the fonts page design may use. As GHGuardianHeadline-Bold is used in eg. standfirsts and H2s, and those can contain italic, not defining GHGuardianHeadline-BoldItalic leads to ugly faux rendering. Death to faux.

## Why?
We cannot have pages using fonts we do not provide!

## Before (font missing, faux style)

image

## After (font loaded)

image

cc @sndrs @jamie-lynch

### DIFF
--- a/src/lib/fonts-css.ts
+++ b/src/lib/fonts-css.ts
@@ -131,6 +131,28 @@ const fontList: FontDisplay[] = [
 		weight: 700,
 		style: 'normal',
 	},
+	{
+		family: 'GH Guardian Headline',
+		woff2:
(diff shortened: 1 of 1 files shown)


HarryFischer:
Nice!

HarryFischer (approved):

PR #3260, opened 2021-08-09
Use automat v2

tomrf1:
The old automat-client library will be archived.
Frontend uses automat v2 for Supporter Revenue messages. DCR should do too.


nicl (approved):
Nice cleanup!

PR #3265, opened 2021-08-09
Make opinion articles accessible (and update font size)

HarryFischer:
## What does this change?
Changes the opinion.main from .400 to .300 so it is now accessible
Changes the font size of the byline until mobileMedium to match the headline size

## Why?
Wasn't accessible & font size was wrong

### Before
image

### After
image


nitro-marky (approved):
👍

PR #3266, opened 2021-08-09
Add extra classes for interactives

DavidLawes:
## What does this change?

Adding extra class names to support Visuals in customising interactives. These class names represent a contract that we should try to maintain.

## Why?

To support the customisation of interactives and thereby reducing the number of dcr opt-outs for new interactives.

### Before vs After

N/A - no UI changes


nicl (approved):
Not necessarily part of this PR, but I wonder if we should add some tests for the presence of these classes in interactive pages going forward?

docs/contracts/003-extra-classes-for-interactives.md
```diff
@@ -0,0 +1,20 @@
+# Extra Classes For Interactives
```
jlieb10, line 1: Thank you for the documentation!

src/web/components/ArticleMeta.tsx
```diff
@@ -264,7 +265,10 @@ export const ArticleMeta = ({
 
 	const showAvatar = onlyOneContributor && shouldShowAvatar(format);
 	return (
-		<div css={metaContainer(format)}>
+		<div
+			className={interactiveLegacyClasses.metaContainer}
```
jlieb10, line 269: I'm not sure how className and CSS work together when both specified - as the css compiles down into a classname itself.

jlieb10 (approved):
Nice work!

src/web/components/ArticleMeta.tsx
DavidLawes: ah, when I tested the static class name was included as well as the auto-generated one, eg:

Screenshot 2021-08-16 at 14 08 58

does that answer your question?

src/model/json-schema.json
```diff
@@ -3628,10 +3628,7 @@
                     }
                 },
                 "abTests": {
-                    "type": "object",
-                    "additionalProperties": {
-                        "type": "string"
-                    }
+                    "$ref": "#/definitions/Record<string,\"control\"|\"variant\">"
```
DavidLawes, line 3631: I think this is ok? Looks like recently the type for this property was update https://github.com/guardian/dotcom-rendering/pull/3276/files

src/web/components/ArticleMeta.tsx
jlieb10: yes - thanks!

src/model/json-schema.json
jlieb10: i'm not sure I understand these updates

jlieb10:
@DavidLawes the new tests look great and can go ahead - but I just lmk about the schema file when you get a chance. Also, I think you might need to merge `main` in since the code freeze because of the conflicts

src/model/json-schema.json
DavidLawes: changes to this file are now ignored :)

PR #3269, opened 2021-08-10
Use caption field in email embed

rcrphillips:
## What does this change?
Uses the caption field from CAPI instead of the alt field 
## Why?
The caption field is the one that we should really use, instead of alt.
### Before
Screen Shot 2021-08-10 at 10 53 30

### After
Screen Shot 2021-08-10 at 10 51 31

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,6 +7,10 @@
     "files.watcherExclude": {
         "**/target": true
     },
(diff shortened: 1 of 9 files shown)


rcrphillips:
[[ROTA] Email embeds not showing captions](https://trello.com/c/HElhlKKS/206-rota-email-embeds-not-showing-captions)

PR #3271, opened 2021-08-10
Update automat client for smarter article count

tomrf1:
- Doesn't count repeat views of the same url
- Takes local storage as a parameter

For details see automat PR: https://github.com/guardian/automat-client-v2/pull/5


src/web/components/App.tsx
```diff
@@ -232,7 +233,7 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 			const hasOptedOut = await hasOptedOutOfArticleCount();
 			if (!hasOptedOut) {
 				incrementDailyArticleCount();
-				incrementWeeklyArticleCount();
+				incrementWeeklyArticleCount(storage.local, CAPI.pageId);
```
TomPretty, line 236: I guess we should add `CAPI.pageId` to the list of dependencies for this `useEffect`?

tomrf1: I suspect it wouldn't make a difference (CAPI.pageId shouldn't change), but I think `[]` makes it clear that this effect must only run once

PR #3273, opened 2021-08-10
Add bash script for team city build

nitro-marky:
## What does this change?

This adds a CI bash script for TeamCity for DCR builds. It contains all of the TeamCity steps in a single script.

This has been tested in teamcity. Existing teamcity steps will be ignored if this script exists and instead a new created step `Run CI Bash Script` would run. TeamCity is setup so it can run with or without this script meaning there will be no down time by switching from existing setup to use the new bash script, however, after a grace period of a week the old TeamCity steps will be disabled.

## Why?

Allows build configuration to be version controlled.

## After

The second step doesn't run because the ci script exists:

Screenshot 2021-08-11 at 11 28 31

Paired on this: @OllysCoding @marjisound @nitro-marky

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+source ~/.nvm/nvm.sh
+nvm install
+nvm use
+#Code Validation
+echo bundlesize token $BUNDLESIZE_GITHUB_TOKEN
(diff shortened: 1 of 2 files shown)


JamieB-gu (approved):
LGTM

PR #3275, opened 2021-08-11
Disable validation on PROD for perf reasons

nicl:
(Will deploy and see perf impact before any merge.)

## What does this change?

Disables validation of CAPI model on PROD.

## Why?

Validation is expensive, especially for larger JS objects.

### DIFF
--- a/src/model/validate.ts
+++ b/src/model/validate.ts
@@ -16,6 +16,11 @@ addFormats(ajv);
 const validate = ajv.compile(schema);
 
 export const validateAsCAPIType = (data: { [key: string]: any }): CAPIType => {
+	// validation is slow so don't perform on PROD.
+	if (process.env.NODE_ENV === 'production') {
+		return data as CAPIType;
+	}
+
 	const isValid = validate(data);
 
 	if (!isValid) {


DavidLawes (approved):
This make sense, we own the contract between DCR and Frontend so would be aware of any changes to the schema?

mchv:
@idrise @nicl any reason we should not merge this and monitor improvements?

nicl:
@mchv @DavidLawes no reason - I think it was just forgotten about. @DavidLawes might be worth deploying for a short period (say an hour) and monitoring for latency/cpu impact. That would then steer whether to merge or close.

nicl:
(It will need a good rebase first though!)

idrise:
How big a problem was the performance of the validation? Imagine a bug in Capi where the data comes through in an unexpected format would you prefer a log message detailing the validation issue or a random crash on the page?

mchv:
> How big a problem was the performance of the validation? Imagine a bug in Capi where the data comes through in an unexpected format would you prefer a log message detailing the validation issue or a random crash on the page?

I think @nicl will be able to provide a more nuanced view, but I think I can provide some interesting points:

- The data from Content API will not come in an unexpected format because the interface between Content API and its consumer is [already strongly typed](https://github.com/guardian/content-api-models/blob/master/models/src/main/thrift/content/v1.thrift). The `thrift` IDL enables both data validation and using other data serialisation format at no runtime cost.  The Content API Scala client will use by default a [more efficient data serialisation than `JSON`](https://github.com/guardian/content-api-scala-client/blob/main/client/src/main/scala/com.gu.contentapi.client/ContentApiClient.scala#L42).   
  
- The reason there is data validation in `DCR` is because `DCR` got data pushed from `Frontend` in `Json` and because `Frontend` modify and augment the data it retrieves from the Content API. There are several approaches for doing this validation:
   - At `runtime` for every request — the current approach 
   - At `runtime` but only on non `PROD` environment — this PR approach
   - At `compile time` — To do this we would need to change how we serialise data in `Frontend` for `DCR`, for instance using an `IDL` so we know the `JSON` we build in `Frontend` is valid.
   - At `test time` — To do this we will simply need ensure that all the modification from `Frontend` of the Content API response is covered by a test.

The current approach is the most primitive and simple one, it trades off runtime performance for data safety and implementation simplicity (which is important!). I think @nicl PR approach is keeping implementation simplicity, but reduce data safety only _lightly_ to not degrade runtime performances. Lightly because the cases where data validation will found an issue are limited, notably because Content API source data is safe, and extremely likely to be found on `CODE` or `DEVINFRA`    


- Our monitoring approach. From a reader perspective, there is almost no difference between `a log message detailing the validation issue` and  `a random crash on the page`. In both cases we should detect the issue (through `Sentry` in this case) and fix the problem. If we think there is value if knowing if the data retrieved from Frontend was valid, we could run the data validation just before sending the error  to `Sentry`.

nicl:
@mchv thanks for the useful summary.

@idrise I can't remember the performance impact unfortunately - the suggestion was to test this on PROD for a brief window and look at latency and CPU impact on PROD and then make a decision.

A better approach would be to have some kind of perf test built into CI.

oliverlloyd:
This validation (in my experience) leads to a poor reader experience. If DCR gets a CAPI object with an unexpected type (usually a required type that isn't provided) then it means the reader sees a white screen with some error text on it 😨 . But actually - in every case I've ever seen - DCR was perfectly capable of rendering the article.

I'd be very much in favour of giving DCR a better chance to deal with unexpected data and showing how resilient it can be 💪

oliverlloyd:
The contra to removing it is you lose visibility of any issues but I think that doesn't justify the impact having it on Prod

Aside: I'm working on pushing the `CAPI` object up and out of the DCR stack to the point where we can have a defined list (type) of the _exact_ data needed to render an article. This might offer a way to look at sharing types better?

PR #3277, opened 2021-08-12
Support POST /Interactive

nicl:
## What does this change?

Adds POST /Interactive support for the dev-server (to run with Frontend). Also fixes the dev-server error handling - express requires the 4-arity version here.

## Why?

Fixes things.


DavidLawes (approved):
👍

PR #3278, opened 2021-08-12
Remove atom cleaner enhance step

nicl:
For performance reasons, we'll do this in Frontend.

See https://github.com/guardian/frontend/pull/24088 for more motivation.


DavidLawes (approved):
👍

PR #3280, opened 2021-08-12
Common Code Structure

JamieB-gu:
## Why?

We're soon going to need a place in this repo to store code that's common to both DCR and AR. This is a first pass at the directory structure for this, and some examples of what that common code could look like. I'm opening this as a draft to get some feedback from the team. This PR shouldn't be affected by the migration work, so I'll keep it open and work on some of the other stuff we need:

- [ ] Making sure we can import from here into DCR and AR
- [ ] Making sure storybook works with this new directory
- [ ] Making sure Jest works with this new directory
- [ ] Making sure VSCode and other editors can understand this directory

## What does this change?

- Added `common`, `common/src` and `common/src/components`
- Added README for `common/src`
- Added example TS file to `src`
- Added example component


lucymonie (commented):
I'm happy with the folder structure that you've proposed @JamieB-gu

PR #3282, opened 2021-08-13
Add position to atom - required for certain legacy interactives

nicl:
## Why

Some legacy interactives require this to style their main image. e.g. https://www.theguardian.com/environment/ng-interactive/2019/jul/01/its-getting-warmer-wetter-wilder-the-arctic-town-heating-faster-than-anywhere

### Before

Screenshot 2021-08-13 at 10 20 37

### After

Screenshot 2021-08-13 at 10 20 29


OllysCoding (approved):
Nice one!

DavidLawes (approved):
nice!

PR #3284, opened 2021-08-13
updating box sizing for legacy interactive immersives

DavidLawes:
## What does this change?
It was found that some legacy interactive immersives did not display correctly because of the default box-sizing. This change overrides the default.

Example: https://www.theguardian.com/us-news/ng-interactive/2017/dec/08/donald-trump-russia-investigation-key-questions-latest-news-collusion-timeline?dcr

## Why?
To support migration of all interactive immersives to DCR.

### Before
Screenshot 2021-08-13 at 14 09 41

### After
Screenshot 2021-08-13 at 14 09 56


nicl (approved):
Hard to judge impact though!

OllysCoding (approved):

PR #3286, opened 2021-08-13
Add interactive alt attr to top-level figure

nicl:
It needs to be on the same element as has the legacy classes for
some interactives to work. E.g.

https://www.theguardian.com/film/ng-interactive/2015/dec/04/best-films-of-2015-us-america

Note, these interactives (like documentaries) still don't work on DCR locally for some reason. Though they should work on PROD.

### DIFF
--- a/src/web/lib/renderElement.tsx
+++ b/src/web/lib/renderElement.tsx
@@ -721,7 +721,8 @@ export const renderElement = ({
 };
 
 // bareElements is the set of element types that don't get wrapped in a Figure
-// for most article types.
+// for most article types, either because they don't need it or because they
+// add the figure themselves.
 const bareElements = new Set([
 	'model.dotcomrendering.pageElements.BlockquoteBlockElement',
 	'model.dotcomrendering.pageElements.CaptionBlockElement',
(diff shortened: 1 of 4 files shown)


src/web/lib/renderElement.tsx
```diff
@@ -733,6 +733,7 @@ const bareElements = new Set([
 	'model.dotcomrendering.pageElements.SubheadingBlockElement',
 	'model.dotcomrendering.pageElements.TextBlockElement',
 	'model.dotcomrendering.pageElements.InteractiveContentsBlockElement',
+	'model.dotcomrendering.pageElements.InteractiveBlockElement',
```
OllysCoding, line 736: What does adding this here mean? The comment "bareElements is the set of element types that don't get wrapped in a Figure for most article types." seems odd since it _is_ now wrapped in a figure?

nicl: @OllysCoding yeah, elements which wrap themselves in a figure already don't need one added here. I've pushed an update to improve this comment for clarity.

OllysCoding (approved):

PR #3288, opened 2021-08-13
Add missing styling for interactive immersives

DavidLawes:
## Why?
Some interactive immersives assume default styling exists (primarily font family for `p` elements). We should add the default styling as part of the interactive immersive migration to DCR.

Example: https://www.theguardian.com/environment/ng-interactive/2019/oct/17/stripped-bare-australias-hidden-climate-crisis?dcr

The missing default `p` styling was noted during conversations with Visuals - it's something they mentioned having to add.

### Before
Screenshot 2021-08-13 at 17 37 34

### After
Screenshot 2021-08-13 at 17 37 38


rcrphillips (approved):
Much improved

PR #3290, opened 2021-08-16
Exhaustive quote matching for drop caps

AWare:
Having some issues with DropCaps not picking up quotes. I think the open single and open double are the only ones needed to match, but thought it would probably be useful to be exhaustive.

(I also made this all in codespaces so maybe none of the plugins etc have run against this branch)


rcrphillips (approved):

OllysCoding (approved):
This looks good 👍  thanks for doing this :)

PR #3291, opened 2021-08-16
Best website header logo

HarryFischer:
## What does this change?
Added new lock up for best website of year award

## Why?
New design

### Before
image

### After
image


OllysCoding (approved):
LGTM

oliverlloyd:
Ha, I was just about to merge a PR to [add the anniversary stuff to Gateway](https://github.com/guardian/gateway/pull/835) and bam, there's a new version! You guys move too fast for me 😉 

I love this new look though. I'll copy it over.

I'm then planning on hoisting this `Logo` work into Source so perhaps you can start importing it from there for DCR? That way we all have the option to show this new version if we want to.

HarryFischer:
@oliverlloyd What is the Gateway? The like sign in pages?
I'm not sure we would want this new logo on those pages. We need to change this PR later so that only the UK edition shows this "best news website" version and the other will revert to the "For 200 years" version. I can send you an updated SVG too as the previous one had been squashed

oliverlloyd:
> @oliverlloyd What is the Gateway? The like sign in pages?
> I'm not sure we would want this new logo on those pages. We need to change this PR later so that only the UK edition shows this "best news website" version and the other will revert to the "For 200 years" version. I can send you an updated SVG too as the previous one had been squashed

Yeah, Gateway is the new sign in pages. We don't show the anniversary version there but I added it to the PR because I plan on hoisting the code up into Source, where it is useful to have the option to show an anniversary version.

So are there in fact _three_ versions of the Logo now, Standard, Anniversary and BestWebsite?

PR #3296, opened 2021-08-17
Remove legacy ci.sh script

OllysCoding:
## What does this change?

Removes a legacy ci.sh script in the dotcom-rendering subdirectory

## Why?

Leftover from the migration - the current ci script is in `scripts/ci.sh`


lucymonie (approved):
LGTM - do we need to take any action on the 'Unchanged files with check annotations'?

marjisound (approved):

PR #3311, opened 2021-08-17
Only show 'best website' logo on UK edition

rcrphillips:
## What does this change?
Updates the recent logo change to only display it on the UK edition.

## Why?
Because we only won the award in the UK.....

### UK edition
Screen Shot 2021-08-17 at 17 52 48

### US edition
Screen Shot 2021-08-17 at 17 52 14


nitro-marky (approved):
Nice 🎊

PR #3313, opened 2021-08-18
Bug fixes on tone/news interactives

jlieb10:
## What does this change?
This covers the remaining blockers for tone/news articles going live via DCR. There were 4 blockers:
- drop cap on the published date
- first line of article text hidden by image
- share icons cover the headline
- images in body wrong size

## Why?
To enable tone/news articles to be rendered via DCR to an acceptable standard.

### Drop cap on date
Before:
Screenshot 2021-08-18 at 17 10 30

After:
Screenshot 2021-08-18 at 17 11 18

### First line of article hidden by image
Before:
Screenshot 2021-08-18 at 17 10 35

After:
Screenshot 2021-08-18 at 17 11 23

### Share icons cover the headline
Before:
Screenshot 2021-08-18 at 17 11 41

After:
Screenshot 2021-08-18 at 17 12 02

### Images in body are the wrong size
Before:
Screenshot 2021-08-18 at 17 12 38

After:
Screenshot 2021-08-18 at 17 12 55


dotcom-rendering/src/web/layouts/InteractiveImmersiveLayout.tsx
```diff
@@ -67,6 +68,47 @@ const Renderer: React.FC<{
 				case 'model.dotcomrendering.pageElements.TextBlockElement':
 					return el;
 
+				// We map image classes to their role name for compatibility with
+				// legacy interactive immersives
+				case 'model.dotcomrendering.pageElements.ImageBlockElement':
```
DavidLawes, line 73: we should do this modification for all interactive immersives, or only legacy?

nicl (approved):
Great (and painstaking) work!

Share David's question about scoping the picture classes and default role to legacy interactives.

jlieb10:
@nicl @DavidLawes I think these changes should only be in place for legacy as I assume the non-legacy interactive will have up to date markup. I do not know what all of the differences between legacy and non-legacy interactives are, though and it is a good question.

OllysCoding (approved):
This looks good!

DavidLawes:
i've added quite a lot of logic, so was wondering if i should add tests, but unsure where to add these tests (for the layout file itself, extract the logic into separate lib file and test those functions directly, something else)?

DavidLawes:
We've agreed with Visuals to press legacy interactive immersives so there is no need to try and render all via DCR. Any particular interactive immersives the Visuals team identify for migration will be dealt with on a case-by-case basis (hopefully no global changes to dotcom-rendering).

Closing this PR

PR #3315, opened 2021-08-18
Limit legacy interactive styling

nicl:
## What does this change?

Limts the global legacy styles:

* to scope to `body article`
* and only for 'legacy' interactives

## Why?

To avoid unfortunate side effects like introducing side scrolling on the footer (due to the change in box-model). E.g.

Screenshot 2021-08-18 at 14 35 49

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -472,6 +472,11 @@ interface CAPIType {
 	matchUrl?: string;
 	isSpecialReport: boolean;
 
+	// Interactives made on Frontend rather than DCR require special handling.
+	// The logic is date-driven. See:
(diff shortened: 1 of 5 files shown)


OllysCoding:
There's a lot of re-formatting going on here, is this something you were expecting, or is it a result of inconsistencies from the migration?

jlieb10 (approved):
@nicl the logic for loading and scoping the styles looks good to me, and makes a lot of sense. I am confused about all of the linting updates to `dotcom-rendering/index.d.ts` in this PR, though, and am curious if those are even the linting styles we normally use? They look a little bit odd, but I could be mistaken

DavidLawes (approved):
Thanks :)

nicl:
@jlieb10 hey yeah, I *think* the formatting is related to the repo change yesterday. Will confirm before merging though as yeah did seem weird to me too. Thanks!

PR #3322, opened 2021-08-19
Fix label and double rendering of ads in interactive articles

MarSavar:
Co-authored-by: Ravi <7014230+arelra@users.noreply.github.com>

## Why?
Interactives are now rendered by DCR, but the ad styles were referencing classes that only exist in `frontend`. 
This caused two issues:

1) Ad labels were incorrectly styled and showing a cross button. This PR addresses this by importing `labelStyles` from the `AdSlot` component for fixed ads that already exists in DCR.

2) The `.mobile-only` and `.hide-until-tablet` classes were not present in DCR, resulting in two ads being displayed. These classes have now been added to the ad styles CSS.

Example: https://www.theguardian.com/tv-and-radio/ng-interactive/2021/aug/04/whats-on-netflix-and-amazon-this-month-august

### Before
image

### After
image

### DIFF
--- a/dotcom-rendering/src/web/components/ElementContainer.tsx
+++ b/dotcom-rendering/src/web/components/ElementContainer.tsx
@@ -1,11 +1,12 @@
 import { ClassNames, css as emoCss } from '@emotion/react';
 
 import { border } from '@guardian/src-foundations/palette';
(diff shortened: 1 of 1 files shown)


nitro-marky (approved):
LGTM

PR #3325, opened 2021-08-20
Ad targeting should set signed in parameter si

arelra:
## What does this change?

Ad targeting object should set the `si` parameter to reflect whether the user is signed in or not.


nitro-marky (approved):
👍

dotcom-rendering/src/lib/ad-targeting.ts
```diff
@@ -10,9 +12,10 @@ export const buildAdTargeting = (
 		};
 	}
 	const { config } = CAPI;
+	const isSignedIn = (typeof window !== "undefined") && getCookie('GU_U') ? 't' : 'f';
```
oliverlloyd, line 15: Will this work on the server? I think `getCookie` reads `document`

oliverlloyd: Also, as far as I can tell, this adds a blocking synchronous cookie read to `App.tsx`. Does this then slow hydration for everything?

oliverlloyd: We already read this cookie in `App.tsx` but inside a useEffect, like this:

```	useEffect(() => {
		setIsSignedIn(!!getCookie('GU_U'));
	}, []);```
	
Is there perhaps a way to use that state value asynchronously instead?

oliverlloyd: Nitpick: But `isSignedIn` already exists 86 times in the codebase but as a _boolean_. This is creating a new variable as a string but with the same name. So whilst the name does feel right and reads well, there's a risk it might get accidentally caught up in a find and replace in future!

oliverlloyd: One last point: I just wanted to add this link to a note about why reading cookies on the main thread is a problem: https://developers.google.com/web/updates/2018/09/asynchronous-access-to-http-cookies. This page is discussing a future solution to this but is far from ready so in the mean time it's best avoided

Sorry, I'll stop adding comments after this one!

arelra: > Does this then slow hydration for everything?

Yes thanks good catch it would...  a _tiny_ bit.

But tiny bits count so I will try another method.

> Is there perhaps a way to use that state value asynchronously instead?

I did look at that hook and thought it would require a bunch of prop drilling to get the value through all the Layout components to the helper function.

But on balance to unblock the main thread it's probably the most straightforward way so I'll amend to pass it through.

> there's a risk it might get accidentally caught up in a find and replace in future!

There is a unit test which I hope would catch an incorrect refactor of behaviour.

That leaves semantics, but it's not a big deal so I can rename it.

> eliminating_jank.

When I was looking at this I actually thought a 'better' solution might be to modify `render.ts` to extract the cookies already available in the request server-side.

So you'd have something like `userState` passed down to the top level components along with `CAPI` and `Nav`.

As long as we don't have any magical XHR login between page loads then that should always reflect the user signed in state and we don't need any hooks or async re-reading of cookies client side?

I spoke to @OllysCoding and we weren't sure if the user cookies were forwarded from Frontend to DCR servers in production. I'll check locally though.

oliverlloyd: > modify render.ts to extract the cookies already available in the request server-side.

This is a really interesting idea. My initial instinct is to say that it's not worth it because it will split the cache. We'd have to modify the cache key or add the `GU_U` cookie value to the Vary header both of which would reduce our hit ratio.

But then I got to thinking of what we could do if we _did_ do this and it could be a lot. So maybe splitting the cache is worthwhile? Or at least worth thinking about.

Things like `SignInGate` and `Discussion` (+ probably a few others) could potentially be rendered server side. There would no longer be a flash as things like  the `MyAccount` dropdown in the header change after reading the cookie.

Furthermore, if we were to also split on `countryCode` then a pretty substantial chunk of additional client side logic could also happen on the server, mostly around banners.

### So should we split the cache on `isSignedIn` and maybe `countryCode`?
There's a lot of reasons not to do this. Testing would be harder, we'd get more origin traffic, have to pay for more servers, have to pay our CDN provider more and generally have more complex caching logic to manage - all scary stuff.

But the wins would be big. Having so much more content rendered on the server would greatly improve the reader experience along with the page performance. This would lead to much better core vital scores, potentially leading to better search rankings. We'd also be able to really simplify the code in DCR, reducing client side code a lot. All good stuff.

In my opinion, at the moment we shouldn't do this, but I'd love to think it through - maybe there's something here?

oliverlloyd: @arelra just to bring this back to the PR in question:

>> Is there perhaps a way to use that state value asynchronously instead?
>>
> I did look at that hook and thought it would require a bunch of prop drilling to get the value through all the Layout components to the helper function.
>
> But on balance to unblock the main thread it's probably the most straightforward way so I'll amend to pass it through.

If we're only using `adTargeting` in the Your Tube block component then prop drilling `isSignedIn` won't have any impact (other than more props) and should achieve your goals here

The wider discussion of using `isSignedIn` is outside the scope of this PR and doesn't need to be a blocker. I do think though that if we expect this use case to expand (that we might want to know the signed in status of a reader for more ad requests in the future) then that adds more weight to the argument for splitting the cache on this value.

@nicl it'd also be great to hear your thoughts

arelra: Thanks @oliverlloyd .

I have an idea but not sure what 'cache splitting' means in the context of the DCR production architecture and request lifecycle.

Seems like a topic for a CSIT meeting?

PR #3326, opened 2021-08-20
Hf interactive sub meta width

HarryFischer:
## What does this change?
Adjusts the sub meta component to match body text width and removed the "reuse this content" button on interactives

## Why?
Looks better! and I we never had the reuse content before(don't think it's needed?)

### Before
image

### After
image

### DIFF
--- a/dotcom-rendering/src/web/layouts/InteractiveLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/InteractiveLayout.tsx
@@ -93,6 +93,7 @@ const InteractiveGrid = ({ children }: { children: React.ReactNode }) => (
 						'.      border  lines'
 						'.      border  meta'
 						'body   body    body'
+						'bottomlines  bottomlines   bottomlines'
 						'.      .       .';
 				}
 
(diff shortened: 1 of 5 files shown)


rcrphillips (approved):

dotcom-rendering/src/web/layouts/InteractiveLayout.tsx
```diff
@@ -93,6 +93,7 @@ const InteractiveGrid = ({ children }: { children: React.ReactNode }) => (
 						'.      border  lines'
 						'.      border  meta'
 						'body   body    body'
+						'bottomlines  bottomlines   bottomlines'
```
DavidLawes, line 96: we might want to remove this? (i think it was just as a test?)

HarryFischer: Ah yes!

DavidLawes (approved):
looks good! when we can remove some of our legacy styling hacks, I think it'll look even better!

PR #3328, opened 2021-08-23
Put Storybook In Root

JamieB-gu:
## Why?

A draft PR for putting storybook in the root of the project rather than in the `dotcom-rendering` sub-directory. Both the `storybook` and `build-storybook` scripts work.

This PR is open for two reasons:

1. To allow the various checks, Chromatic in particular, to run.
2. To gather feedback from the team.

## Babel Config

I've removed the custom babel config to get this working and it seems fine without. These leaves two questions:

- [ ] Is it _actually_ fine without? Will some stories now be broken even though it builds fine? The Chromatic diff may tell us.
- [ ] Is it problematic to diverge from the PROD environment (which _does_ use the babel config) in this way?

## Chromatic

I'm hoping Chromatic will recognise the same stories, even though the configuration is slightly different, and not present us with an enormous diff. We'll see.

## Changes

- Moved `.storybook` directory up to root and fixed the paths and aliases
- Created a root `package.json` with storybook scripts and dependencies required for building storybook
- Removed babel config (possibly temporarily)


dotcom-rendering/package.json
```diff
@@ -290,4 +287,4 @@
     "src/lib/content.d.ts": "make gen-schema && make gen-fixtures",
     "src/lib/index.d.ts": "make gen-schema && make gen-fixtures"
```
oliverlloyd, line 288: Outside this PR's scope but do these file paths need updating too?

oliverlloyd (commented):
🎉 amazing work!

Really cool to see this coming together (literally and figuratively!)

.storybook/main.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
 	core: {
 		builder: 'webpack5',
 	},
-	stories: ['../src/**/*.stories.@(tsx)'],
+	stories: [ '../dotcom-rendering/src/**/*.stories.@(tsx)' ],
```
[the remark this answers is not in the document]
JamieB-gu: > Is the end goal to have one storybook with both DCR & AR in it?

Yes, and `common`. Each project will have an entry here. Alternatively we could just specify the root directory of the repo and let it walk the entire tree.

.github/workflows/chromatic.yml
```diff
@@ -12,17 +12,15 @@ jobs:
             - uses: actions/checkout@v1
             - uses: guardian/actions-setup-node@main
 
+            # Cache npm dependencies using https://github.com/bahmutov/npm-install
+            - uses: bahmutov/npm-install@v1
```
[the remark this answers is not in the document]
JamieB-gu: I think you may have been looking at an old commit? It's only there once now.

dotcom-rendering/package.json
JamieB-gu: I'm not sure, which tool uses these fields?

oliverlloyd: `make gen-schema` runs `TJS.generateSchema` based on CAPIType and updates `src/model/json-schema.json`. There's a dependency on this command being run whenever the types are changed. Previously this dependency was often forgotten about which could cause the site to crash - we use [run-if-changed](https://www.npmjs.com/package/run-if-changed) to automate this.

oliverlloyd: `make gen-fixtures` refreshes our fixture data for a similar reason

PR #3329, opened 2021-08-23
Support variables for interactive atoms on interactive pages

nicl:
Syntax is, e.g.:

{{ byline }}

## What does this change?

Introduce support for some basic placeholders in interactive atoms. (Limited to interactive pages only for now.)

E.g. include  byline by adding this to the HTML part of your atom  definition:

    <div>My byline: {{ byline }}</div>

## Why?

To enable easier creation of interactives by making it easy to include key metadata. This is particularly relevant to immersive interactives (which provide no meta/header section but require the  interactive to do this). By using placeholders, we can leverage Composer data/changes without requiring changes to the core interactive  code.

cc @HarryFischer

### DIFF
--- a/dotcom-rendering/src/model/enhance-placeholders.ts
+++ b/dotcom-rendering/src/model/enhance-placeholders.ts
@@ -1,0 +1,83 @@
+import { Design } from '@guardian/types';
+import { decideDesign } from '@root/src/web/lib/decideDesign';
+
+export const replacePlaceholders = (
+	elem: InteractiveAtomBlockElement,
+	variables: Map<string, string>,
+): InteractiveAtomBlockElement => {
+	if (!elem.html) {
+		return elem;
+	}
+
+	const replacer = (substring: string, name: string): string => {
+		return variables.get(name) || substring;
+	};
+
+	// Note: use str.replace with 'g' (global) flag as str.replaceAll not
+	// supported in current node version (14.6).
+	const re = /{{\s*(\w+)\s*}}/g;
+	elem.html = elem.html.replace(re, replacer);
+
+	return elem;
+};
+
+/**
+ * generateVariables defines the templated variables we replace. As such, it
+ * represents an ongoing contract with interactive atoms; fields should not be
+ * removed from it once added so be cautious when thinking about adding new
+ * ones.
+ */
+const buildPlaceholders = (data: CAPIType): Map<string, string> => {
+	return new Map(
+		Object.entries({
+			webPublicationDate: data.webPublicationDateDisplay || '',
+			byline: data.author.byline || '',
+			webTitle: data.webTitle,
+			trailText: data.trailText,
+			sectionLabel: data.sectionLabel,
+			// shareLinks: TODO,
+		}),
+	);
+};
+
+/**
+ * enhancePlaceholders replaces templated fields in interactive atom elements
+ * for interactive pages (format.design === Design.Interactive) only.
+ *
+ * Supported variables will be replaced within the HTML of any interactive atom
+ * elements. Note, this will therefore not work on dynamically loaded or
+ * generated markup.
+ *
+ * Variables are wrapped in curlies, e.g. {{ webPublicationDate }}. Surrounding
+ * whitespace is ignored. E.g. {{webPublicationDate    }} is fine.
+ */
+export const enhancePlaceholders = (data: CAPIType): CAPIType => {
+	const design: Design = decideDesign(data.format);
+
+	// Only immersive interactives are modified.
+	if (design !== Design.Interactive) {
+		return data;
+	}
+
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/model/enhance-placeholders.ts
```diff
@@ -0,0 +1,83 @@
...
+ */
+const buildPlaceholders = (data: CAPIType): Map<string, string> => {
+	return new Map(
+		Object.entries({
+			webPublicationDate: data.webPublicationDateDisplay || '',
+			byline: data.author.byline || '',
+			webTitle: data.webTitle,
+			trailText: data.trailText,
+			sectionLabel: data.sectionLabel,
+			// shareLinks: TODO,
```
nicl, line 38: cc @HarryFischer anything else we could include to start here?

dotcom-rendering/src/model/enhance-placeholders.ts
```diff
@@ -0,0 +1,84 @@
...
+};
+
+/**
+ * generateVariables defines the templated variables we replace. As such, it
+ * represents an ongoing contract with interactive atoms; fields should not be
+ * removed from it once added so be cautious when thinking about adding new
+ * ones.
+ */
+const buildPlaceholders = (data: CAPIType): Map<string, string> => {
+	return new Map(
```
DavidLawes, line 31: out of curiosity, why do we chose a Map here (rather than just a plain Object)? I have tended just to use Objects in similar situations so would be glad to understand why Map is better :)

DavidLawes (approved):
This looks good - I worry without examples or explanations then this feature may go unused, maybe when we talk to Visuals it'll be more obvious where it's most useful to put this kind of documentation

dotcom-rendering/src/model/enhance-placeholders.ts
DavidLawes: standfirst?

dotcom-rendering/src/model/enhance-placeholders.ts
```diff
@@ -0,0 +1,84 @@
...
+export const replacePlaceholders = (
+	elem: InteractiveAtomBlockElement,
+	variables: Map<string, string>,
+): InteractiveAtomBlockElement => {
+	if (!elem.html) {
+		return elem;
+	}
+
+	const replacer = (substring: string, name: string): string => {
+		return variables.get(name) || substring;
```
OllysCoding, line 13: In the case where someone enters something unsupported, e.g "{{ test }}", this is then replaced with "test". Is this preferable over leaving it as "{{ test }}"? 

My concern being that there may be other places where this `{{ }}` syntax is used, and we're not being friendly to other libraries / code by overwriting all instances of it in an article. Especially since we're targeting the whole doc, so this could be in a HTML data parameter, an inline script, etc. 

I don't think it has a major impact on the end user, but leaving it as {{ test }} might also help someone creating an interactive notice that they made a typo, etc.

OllysCoding: Resolving as I misunderstood the code.

dotcom-rendering/src/model/enhance-placeholders.ts
OllysCoding: I'm also curious about this!

dotcom-rendering/src/model/enhance-placeholders.ts
nicl: @OllysCoding in fact, if we don't have the variable defined as supported (in the passed in variables map) then we don't modify the string at all. One of the tests aims to capture this.

dotcom-rendering/src/model/enhance-placeholders.ts
jlieb10: Good question

dotcom-rendering/src/model/enhance-placeholders.ts
jlieb10: Why don't they all have fallbacks? Are some of them required elsewhere, in Composer?

jlieb10: Also wondering if headline should exist here? I think there is sometimes a distinction between title and headline

jlieb10 (approved):
Nice work - and the test file is very helpful for understanding the code.

dotcom-rendering/src/model/enhanceCAPI.ts
```diff
@@ -59,6 +60,11 @@ class CAPIEnhancer {
 		this.capi = setIsDev(this.capi);
 		return this;
 	}
+
+	enhancePlaceholders() {
```
jlieb10, line 64: Is anything calling this code at this point?

PR #3330, opened 2021-08-23
change margin so borders align

HarryFischer:
## What does this change?
change margin-left at wide of submeta so borders align

## Why?
was a mistake before

### Before
Screenshot 2021-08-23 at 16 58 08

### After
Screenshot 2021-08-23 at 16 58 15

### DIFF
--- a/dotcom-rendering/src/web/components/SubMeta.tsx
+++ b/dotcom-rendering/src/web/components/SubMeta.tsx
@@ -42,7 +42,7 @@ const setMetaWidth = (palette: Palette) => css`
 		border-left: 1px solid ${palette.border.article};
 	}
 	${from.wide} {
-		margin-left: 230px;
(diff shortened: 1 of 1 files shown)


DavidLawes:
I'm probably getting confused, but if we remove the content-box styling (which we'll do as part of some clean up), would we still need this change? https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/layouts/lib/interactiveLegacyStyling.ts#L37

HarryFischer:
@DavidLawes Good point... I'm not sure to be honest. Let me see if I can test that

DavidLawes (approved):
this isn't impacted by box-sizing, it looks like it's needed because of how the grid layout is defined from wide.

given we don't need to worry about this change when we clean up legacy styling, looks all good to me :)

PR #3333, opened 2021-08-25
update ci script to skip build if only AR files were changed

marjisound:
## What does this change?
This PR updates ci script by adding checks for branch name and file changes. 

- If branch name is main, it always runs the build steps.
- If branch name is not main, it would also skip the build step if all file changes were from apps-rendering sub directory.  
- Otherwise, it would run the build step.

## Why?
This change is a pre-migration enabler. With this, we can remove the filters on teamcity trigger and the teamcity build always runs no matter if changes were in DCR or AR. As a result the DCR build status check in github would always get an update from teamcity (even when the file changes were only for AR)

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -7,7 +7,7 @@ files="$(git diff --name-only origin/main)"
 echo "files: $files"
 
 # files that are not within apps-rendering sub directory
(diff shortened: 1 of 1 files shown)


nitro-marky (approved):

OllysCoding (approved):
This is looking really cool!

PR #3062, opened 2021-06-02
Revert "Enable picnic stories ads on AMP pages"

ealitten:
Reverts guardian/dotcom-rendering#3044

Undo adding the 480x320 size to <Ad> component in AMP pages due to unexpected padding issues (intermittent)

image

PR #3338, opened 2021-08-26
Add picnic ad size to amp ads

chrislomaxjones:
## What does this change?

A follow up to [previous work ](https://github.com/guardian/dotcom-rendering/pull/3069) to add an additional 320x480 ad size to AMP pages.

Changes include:
- Add a new 320x480 ad size as the primary ad size for AMP ads
- Disable mutli size validation, since the secondary ads are less than 2/3rds smaller than this new primary ad size (and if we don't disable it the secondary ads aren't displayed).
- Some clarifying comments.

Note this change was [introduced](https://github.com/guardian/dotcom-rendering/pull/3044) and subsequently [reverted](https://github.com/guardian/dotcom-rendering/pull/3062) because of a problem with secondary ads being rendered in containers the size of the primary ad. This appears to be a feature of the AMP runtime only resizing the containers around ads to the smaller secondary sizes if they do not cause layout shift on the visible page - this means that smaller ads higher up the viewport may not be resized and are rendered in the larger containers.

Since this is the intended AMP behaviour it seems acceptable to launch this new ad size on AMP and look at the sizes of ads returned to get a better handle on how often this occurs.

## Why?

This change should provide the opportunity to increase ad revenue as vendors can place larger ads on AMP pages. It will also allow us to assess the impact of displaying secondary sized ads in primary sized containers.


chrislomaxjones:
Marked as DONOTMERGE until rollout of this new ad size is approved.

mxdvl (approved):
LGTM! 🖼️

nitro-marky (approved):

PR #3339, opened 2021-08-26
Update article count timing

TomPretty:
## What does this change?
Update article count timing to wait for the article count to be incremented before requesting epics & banners. This ensures the article count that is displayed is correct, and also ensures that it is the same in both epics & banners (It was reliably showing me the updated count in the banner and the old count in the epic when running locally).

[**Trello card**](https://trello.com/c/4BeEimqf/2765-dcr-increment-article-count-before-fetching-epic-banner)

### DIFF
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -178,6 +178,10 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 		submitComponentEvent(componentEvent, ophanRecord);
 	};
 
+	const [articleCountIsReady, setArticleCountIsReady] = useState<boolean>(
+		false,
+	);
+
 	// *******************************
 	// ** Setup AB Test Tracking *****
 	// *******************************
@@ -235,6 +239,7 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 				incrementDailyArticleCount();
 				incrementWeeklyArticleCount(storage.local, CAPI.pageId);
 			}
+			setArticleCountIsReady(true);
 		};
 		incrementArticleCountsIfConsented().catch((e) =>
 			console.error(`incrementArticleCountsIfConsented - error: ${e}`),
@@ -1166,22 +1171,24 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 					</Lazy>
 				</Portal>
 			)}
-			<Portal rootId="slot-body-end">
-				<SlotBodyEnd
-					isSignedIn={isSignedIn}
-					countryCode={countryCode}
-					contentType={CAPI.contentType}
-					sectionName={CAPI.sectionName}
-					shouldHideReaderRevenue={CAPI.shouldHideReaderRevenue}
-					isMinuteArticle={CAPI.pageType.isMinuteArticle}
-					isPaidContent={CAPI.pageType.isPaidContent}
-					tags={CAPI.tags}
-					contributionsServiceUrl={CAPI.contributionsServiceUrl}
-					brazeMessages={brazeMessages}
-					idApiUrl={CAPI.config.idApiUrl}
-					stage={CAPI.stage}
-				/>
-			</Portal>
+			{articleCountIsReady && (
+				<Portal rootId="slot-body-end">
+					<SlotBodyEnd
+						isSignedIn={isSignedIn}
+						countryCode={countryCode}
+						contentType={CAPI.contentType}
+						sectionName={CAPI.sectionName}
+						shouldHideReaderRevenue={CAPI.shouldHideReaderRevenue}
+						isMinuteArticle={CAPI.pageType.isMinuteArticle}
+						isPaidContent={CAPI.pageType.isPaidContent}
+						tags={CAPI.tags}
+						contributionsServiceUrl={CAPI.contributionsServiceUrl}
+						brazeMessages={brazeMessages}
+						idApiUrl={CAPI.config.idApiUrl}
+						stage={CAPI.stage}
+					/>
+				</Portal>
+			)}
 			<Portal
 				rootId={
 					isSignedIn
@@ -1273,15 +1280,17 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 					/>
 				</Lazy>
 			</Portal>
-			<Portal rootId="bottom-banner">
-				<StickyBottomBanner
-					isSignedIn={isSignedIn}
-					asyncCountryCode={asyncCountryCode}
-					CAPI={CAPI}
-					brazeMessages={brazeMessages}
-					isPreview={!!CAPI.isPreview}
-				/>
-			</Portal>
+			{articleCountIsReady && (
+				<Portal rootId="bottom-banner">
+					<StickyBottomBanner
+						isSignedIn={isSignedIn}
+						asyncCountryCode={asyncCountryCode}
+						CAPI={CAPI}
+						brazeMessages={brazeMessages}
+						isPreview={!!CAPI.isPreview}
+					/>
+				</Portal>
+			)}
 		</React.StrictMode>
 	);
 };
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -178,9 +178,9 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 		submitComponentEvent(componentEvent, ophanRecord);
 	};
 
-	const [articleCountIsReady, setArticleCountIsReady] = useState<boolean>(
-		false,
-	);
+	const [articleCountIsReady, setArticleCountIsReady] = useState<
+		Promise<true>
+	>();
 
 	// *******************************
 	// ** Setup AB Test Tracking *****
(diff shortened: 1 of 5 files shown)


tomrf1 (approved):
👍 looks good

dotcom-rendering/src/web/components/App.tsx
```diff
@@ -1167,20 +1172,22 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
...
-					shouldHideReaderRevenue={CAPI.shouldHideReaderRevenue}
-					isMinuteArticle={CAPI.pageType.isMinuteArticle}
-					isPaidContent={CAPI.pageType.isPaidContent}
-					tags={CAPI.tags}
-					contributionsServiceUrl={CAPI.contributionsServiceUrl}
-					brazeMessages={brazeMessages}
-					idApiUrl={CAPI.config.idApiUrl}
-					stage={CAPI.stage}
-				/>
+				{articleCountIsReady && (
```
oliverlloyd, line 1175: Checking the rest of `App` it looks like the other places where we're conditionally rendering `Portal`s the check is outside the `Portal`, eg:

```
			{CAPI.matchUrl && (
				<Portal rootId="match-nav">
					<GetMatchNav matchUrl={CAPI.matchUrl} />
				</Portal>
			)}
```

I don't _think_ it will matter too much if we render an empty portal but having one less call to `ReactDOM.createPortal` is nice

oliverlloyd (approved):
Nice! It's a lot better having this logic contained in App like this - thanks for doing this!

Just a small suggestion for a tiny improvement but otherwise great work 👍

dotcom-rendering/src/web/components/App.tsx
TomPretty: Yeah looks good I'll push up that change now!

dotcom-rendering/src/web/components/App.tsx
```diff
@@ -1273,15 +1280,17 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
...
-			<Portal rootId="bottom-banner">
-				<StickyBottomBanner
-					isSignedIn={isSignedIn}
-					asyncCountryCode={asyncCountryCode}
-					CAPI={CAPI}
-					brazeMessages={brazeMessages}
-					isPreview={!!CAPI.isPreview}
-				/>
-			</Portal>
+			{articleCountIsReady && (
```
tjmw, line 1283: I wonder should the check for `articleCountIsReady` be pushed down to the underlying components which require it, rather than blocking everything downstream of `StickyBottomBanner`? It's probably not a big deal performance wise but I think right now this will block both the Braze Banner and the CMP banner from rendering? It might also signal the intent more accurately? What do you reckon?

tjmw (commented):
This is looking good! I left one question about where the check for `articleCountIsReady` should live.

Would be interesting to know what impact (if any) delaying the rendering of `SlotBodyEnd`/`StickyBottomBanner` might have on core metrics like LCP/CLS 🤔

dotcom-rendering/src/web/components/App.tsx
tomrf1: The issue is that articleCount is used to decide _if_ a banner should be displayed. I'm not sure the message picker can handle this

tjmw: I chatted to @tomrf1 offline, but I'd started replying to this comment so leaving it here for posterity!

Hey sorry for the delay in replying I've been out for a few days. Ah yep I see what you mean, because the `canShow` of the message picker only runs once. If you did want to work around that I think the way we've handled that elsewhere is to create a `Promise` which represents the eventual completion of `articleCountIsReady` and then await that in the `canShow` of the relevant banner. I think that's how `asyncCountryCode` in `App.tsx` is used. Let me know if I can help at all with this, or if any of this doesn't make sense!

oliverlloyd: If you pass a promise as a prop and then conditionally render based on that doesn't this kind of step outside the declarative mindset of react?

I guess what we'd be doing is using plain javascript to manage state, which is fine, but when we have both react state _plus_ js state don't we then have the problem where we need to manage re-renders manually, potentially creating race conditions? Eg. If I pass in a promise then every component using it will need to setup a useEffect and create some state so that it can trigger a re-render for when the promise resolves. But if you pass in normal react state then that's all built in for free. The component will re-render when the state changes and then you use composition to control things.

So yes, it's definitely true that the `{articleCountIsReady && <MyThing />}` approach blocks _everything_. But you can also `<MyThing articleCountIsReady={articleCountIsReady} />` and then conditionally block locally so that only the think you want to be affected gets re-rendered. This is what React is good at.

tjmw: > If you pass a promise as a prop and then conditionally render based on that doesn't this kind of step outside the declarative mindset of react?

Hmm good question, I'm not sure. Passing a promise directly to a component doesn't feel very different than passing in a callback/handler function which returns a promise which is a fairly common pattern. For example a button with an `onClick` which is provided by the parent component which does something async (a fetch say) and then updates local component state based on the outcome.

> If I pass in a promise then every component using it will need to setup a useEffect and create some state so that it can trigger a re-render for when the promise resolves. But if you pass in normal react state then that's all built in for free.

Also a good point! But eventually you need to deal with that async-ness somewhere in your component hierarchy don't you? I.e. you can't ignore it completely.

In general I agree though that keeping things simple using core React principles is a good thing to aim for. Perhaps the message picker abstraction isn't quite right currently. I think part of the thing here is that we only ever want to show one message per page, i.e. once a banner has been selected to be shown on the current page view, it shouldn't then get superseded by another banner at a later point in time.

oliverlloyd: It's a tricky one indeed!

_Ideally_, all the variables used to decide if a thing should appear would be pure props and then the rendering logic would be simple if statements. I appreciate though that there's more going on with banners, esp. around the decisions on which one to show.

tomrf1: Yes, it's not very react-y. But the message picker needs to kick off several concurrent `canShow` Promises, and we don't want to tie each message to article count.

An alternative approach would be to run the message picker logic in a useEffect in App.tsx, and pass the chosen message into the StickyBottomBanner component for rendering (and the same for SlotBodyEnd).

oliverlloyd: > An alternative approach would be to run the message picker logic in a useEffect in App.tsx, and pass the chosen message into the StickyBottomBanner component for rendering (and the same for SlotBodyEnd).

That's an interesting idea. I've no idea how feasible this would be (I think @tjmw would know this better than I) but I like the idea of hoisting the logic around what to display up - I feel like this would help understand things better. At some point this decision logic became relevant to other parts of the page (the SignInGate specifically) so having it happen at a more centralised location is the logical solution to how it could be shared and then this opens up a route for having simpler, more react-y components later.

These are all just thoughts and I don't want to block this PR but I do think its great to have this discussion nonetheless!

tjmw: That approach is basically what happens within `StickyBottomBanner` currently (except we use `useOnce` instead of `useEffect`, as we only want to trigger the picking logic/show a banner message once). So I'm sure what Oliver describes is possible.

Definitely agree that this is all outside of the scope of this PR though! Since it sounds like we've all got ideas on how this should work, probably coming at it from slightly different angles, I'd be up for mobbing on this together at some point if others are keen, to explore/sketch out ideas.

dotcom-rendering/src/web/components/App.tsx
```diff
@@ -236,8 +240,9 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 				incrementWeeklyArticleCount(storage.local, CAPI.pageId);
 			}
 		};
-		incrementArticleCountsIfConsented().catch((e) =>
-			console.error(`incrementArticleCountsIfConsented - error: ${e}`),
+
+		setArticleCountIsReady(
+			incrementArticleCountsIfConsented().then(() => true),
```
TomPretty, line 245: I don't love the logic here of chaining on `incrementArticleCountIfConsented` but I couldn't come up with a better option. The rendering of the banners/epics is blocked as long as `articleCountIsReady` is `undefined`, so we don't want to wait before calling `setArticleCountIsReady` e.g by doing 

```ts
useEffect(() => {
		const incrementArticleCountsIfConsented = async () => {
			if (!hasOptedOut) {
				incrementDailyArticleCount();
				incrementWeeklyArticleCount(storage.local, CAPI.pageId);
			}
            setArticleCountIsReady(Promise.resolve(true));
		};
```

As the promise isn't getting us anything here. We can't do it in a separate hook e.g

```ts
useEffect(() => {
  setArticleCountIsReady(hasOptedOutOfArticleCount().then(() => true));
}, [])
```
As I think we have a race condition with actually incrementing the article counts. By chaining on `incrementArticleCountIfConsented` we can guarantee that we'll access the article counts after we update them

dotcom-rendering/src/web/components/App.tsx
TomPretty: I've update the PR to use promises. It might be 'good enough' for merging for now, but I think it would be great to mob at some point to maybe figure out a neater solution

dotcom-rendering/src/web/components/App.tsx
```diff
@@ -178,6 +182,10 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 		submitComponentEvent(componentEvent, ophanRecord);
 	};
 
+	const [asyncArticleCount, setAsyncArticleCount] = useState<
+		Promise<WeeklyArticleHistory | undefined>
```
tomrf1, line 186: I like this. It's still a bit weird to pass a Promise into the components, but at least it contains a dependency (WeeklyArticleHistory).

We chatted about making the epic + banner slot selection into custom hooks, instead of components. App.tsx would render the results of these hooks (the winning epic/banner components). But this is a separate refactor.

PR #3340, opened 2021-08-26
Change `top-above-nav` to 250px high minimum

mxdvl:
## What does this change?

Most of the `top-above-nav` ads are 250px high, resulting in a lot of cumulative layout shift (CLS).

Related to https://github.com/guardian/frontend/pull/24095

## Why?

We ran some tests in `frontend` that showed a marked improvement in layout shift. Thus, we are looking to now roll this change out to all users and continue monitoring.

### Before

image

### After

image


mxdvl:
Here’s the breakdown of number: https://docs.google.com/document/d/1n5JQGJOm8mMVbL_OQw9QohAwHWBYueyqoTqn2frYLd4/edit

PR #3342, opened 2021-08-27
`common-rendering` With Workspaces

JamieB-gu:
## Why?

We're soon going to need a place in this repo to store code that's common to both DCR and AR. This introduces `common-rendering` as that new place, and converts the repo over to using [yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces/) to manage this.

I've testing importing code into DCR via the `@guardian/common-rendering/..` path, and VSCode seems happy enough with the way things are structured. Let's see how CI handles things.

### To Follow

There's no mechanism currently for testing or linting the shared code. I think it would be best to handle this in a follow-up, either by creating test and lint setups specifically for the new project, or by lifting up testing and linting to the top level, as we've done with storybook.

This brings together work in #3280 and #3328.

## Changes

- Added a new `common-rendering` project directory
- Added yarn workspaces to manage multiple projects
- Removed DCR's `yarn.lock` in favour of the top-level workspaces one
- Added some example files for the common project
- Added tsconfig to keep editors happy
- Updated DCR's `install` script to use top-level yarn

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -5,7 +5,10 @@ module.exports = {
 	core: {
 		builder: 'webpack5',
 	},
-	stories: [ '../dotcom-rendering/src/**/*.stories.@(tsx)' ],
+	stories: [
+		'../dotcom-rendering/src/**/*.stories.@(tsx)',
+		'../common-rendering/src/**/*.stories.@(tsx)',
+	],
 	addons: [
 		'@storybook/addon-essentials',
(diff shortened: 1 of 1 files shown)


marjisound (commented):
For Snyk there's this `--file=dotcom-rendering/yarn.lock` in both monitor and test that would need to be updated

JamieB-gu:
This has now been achieved in other PRs.

PR #3355, opened 2021-08-31
`common-rendering` Directory

JamieB-gu:
## Why?

We need a place in this repo to store code that's common to both DCR and AR. This introduces `common-rendering` as that new place, and uses [yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces/) for just that project.

This will allow us to start writing new components and modules against storybook.

**Note:** For now I've included the `common-rendering` directory in storybook's webpack config for AR. Hopefully we can get to the point where all three projects share a webpack config for storybook, otherwise this may cause confusion down the line.

### To Follow

There's no mechanism currently for testing or linting the shared code. I think it would be best to handle this in a follow-up, either by creating test and lint setups specifically for the new project, or by lifting up testing and linting to the top level, as we've done with storybook.

A replacement to #3342 for now since it's proving difficult to get DCR to work with those changes.

## Changes

- Added workspace to root `package.json`
- Included `common-rendering` stories in storybook config
- Created `package.json` for `common-rendering`
- Created `tsconfig.json` for `common-rendering`
- Created example modules

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -8,6 +8,7 @@ module.exports = {
 	stories: [
 		"../apps-rendering/src/**/*.stories.@(js|mdx|ts|tsx)",
 		"../dotcom-rendering/src/**/*.stories.@(tsx)",
+		"../common-rendering/src/**/*.stories.@(tsx)",
 	],
 	addons: [
 		"@storybook/addon-essentials",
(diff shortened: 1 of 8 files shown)


marjisound (approved):
Looks good 👍

PR #3356, opened 2021-08-31
LIVE-2953: add football score parsing

joecowton1:
## Why are you doing this?

We're missing football scores in Editions currently. This PR adds PA based automated football scorers for Editions match reports, bringing it into line with Apps-Rendering and Dotcom.

Paired with @JamieB-gu & @marjisound on the parsing section.

## Changes

- Add football scores to Editions-Rendering endpoint
- fix score number positioning
- add `fromNullable` to `teamScorers` component
- add `optionMap3` and `resultToNullable` to `lib`
- refactor `footballContent.ts` to use `fieldParser` based approach

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/apps-rendering/src/server/footballContent.ts
+++ b/apps-rendering/src/server/footballContent.ts
@@ -8,26 +8,22 @@ import {
 	fromNullable,
 	map2,
 	none,
-	OptionKind,
 	map as optMap,
 	resultAndThen,
 	some,
-	withDefault,
 } from '@guardian/types';
 import type { Option, Result } from '@guardian/types';
 import { padZero } from 'date';
-import { pipe } from 'lib';
+import { fold, pipe } from 'lib';
 import fetch from 'node-fetch';
 import type { Response } from 'node-fetch';
 import type { Parser } from 'parser';
 import {
 	arrayParser,
 	fieldParser,
-	map,
 	map3,
 	map6,
 	map7,
-	maybe,
 	numberParser,
 	oneOf,
 	parse,
@@ -51,15 +47,6 @@ const getFootballSelector = (date: Date, [teamA, teamB]: Teams): string => {
 
 const getFootballEndpoint = (selectorId: string): string =>
 	`https://mobile.guardianapis.com/sport/football/matches?selector=${selectorId}`;
-
-const optionToUndefined = <A>(optA: Option<A>): A | undefined =>
-	withDefault<A | undefined>(undefined)(optA);
-
-const stringOrUndefinedParser: Parser<string | undefined> = pipe(
-	stringParser,
-	maybe,
-	map(optionToUndefined),
-);
 
 const makeScorer = (
 	player: string,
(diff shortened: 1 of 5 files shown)


apps-rendering/src/server/footballContent.ts
```diff
@@ -1,66 +1,162 @@
...
+	homeTeam,
+	awayTeam,
+	venue,
+});
+
+const scorerParser: Parser<Scorer> = map3(makeScorer)(
+	fieldParser('player', stringParser),
+	fieldParser('timeInMinutes', numberParser),
+	oneOf([
+		fieldParser('additionalInfo', stringOrUndefinedParser),
```
davidfurey, line 112: Perhaps I've misunderstood, but to me this reads as:
`{"additionalInfo": "test"}` parses as `"test"`
`{"additionalInfo": 1}` parses as `undefined`
`{}` parses as `""`

Is this what is intended?

apps-rendering/src/server/footballContent.ts
```diff
@@ -90,44 +186,32 @@ const teamsFromTags = (tags: Tag[]): Option<Teams> => {
...
+		fromNullable,
+		optMap(makeDate),
+	);
 
-			const selectorId = getFootballSelector(date, teamA, teamB);
+	const selectorId = map2(getFootballSelector)(date)(teams);
 
-			if (selectorId.kind === OptionKind.Some) {
-				const footballEndpoint = getFootballEndpoint(selectorId);
+	if (selectorId.kind === OptionKind.Some) {
```
davidfurey, line 202: Seems a bit strange to be comparing to `OptionKind.Some` outside of `lib.ts`.  Can you not fold over an option?

apps-rendering/src/server/footballContent.ts
joecowton1: No, you're right here. That one slipped through the net, should be the same as the other `oneOf`s

apps-rendering/src/server/footballContent.ts
joecowton1: Struggling a bit with this one, I haven't really used `fold` before but if i do the following is that correct?
```
	return fold(async (selectorIdValue: string) => {
		const footballEndpoint = getFootballEndpoint(selectorIdValue);

		const response = await fetch(footballEndpoint);

		const footballContent = await parseFootballResponse(
			response,
			selectorIdValue,
		);

		return footballContent;
	}, err('Could not get selectorId'))(selectorId);
	```

Comment (joecowton1) [reply apps-rendering/src/server/footballContent.ts:202]: I'm getting an error for the return type but i'm not sure why this is different now.
```
Argument of type 'Err<string>' is not assignable to parameter of type 'Promise<Result<string, FootballContent>>'.

davidfurey: The second argument of `fold` has to be of the same type as the return type of the first argument, so in this case `Promise<Result<string, FootballContent>>`.  

I think you should be able to change  `err('Could not get selectorId')` to `Promise.resolve(err('Could not get selectorId'))`

joecowton1: Great, thanks. That's updated now.

apps-rendering/src/server/footballContent.ts
joecowton1: Updated.

davidfurey (approved):

PR #3358, opened 2021-08-31
LIVE-3004: fix xl margin widths

joecowton1:
## Why are you doing this?

Fixes margins on wide tablet breakpoints.

## Changes

- Add wide margin width
- update `immersive` and `interview` header styles to fit

## Screenshots

| Before | After |
| --- | --- |
|  |  |


marjisound (approved):
LGTM 👍

PR #3361, opened 2021-09-01
remove storybook from apps-rendering directory

marjisound:
## What does this change?
This PR deletes .storybook from apps-rendering sub directory because storybook is not located in the root and shared between apps-rendering and dotcom-rendering
## Why?
apps-rendering/.storybook is not used


OllysCoding (approved):

OllysCoding (approved):
LGTM!

PR #3362, opened 2021-09-01
LIVE-2953: add football scores when video header media

joecowton1:
## Why are you doing this?

The earlier football scores PR did not render the scores component if the header media was a video.

## Changes
- add football scores component to header media video
- minor styling fixes

## Screenshots

| Before | After |
| --- | --- |
|  |  |


marjisound (approved):
LGTM

PR #3366, opened 2021-09-02
Live blog key event component

marjisound:
## What does this change?

This PR creates KeyEvent component for live blogs in common-rendering. dark mode will be done in a separate PR

Paired on this: @joecowton1 & @marjisound 

### After
The screenshots are from storybook

### Mobile
image

### tablet
image

### desktop
image


common-rendering/src/components/keyEvents.tsx
```diff
@@ -0,0 +1,162 @@
...
+interface ListItemProps {
+	keyEvent: KeyEvent;
+	theme: Theme;
+}
+
+const ListItem: FC<ListItemProps> = ({ keyEvent, theme }) => {
+	return (
+		<li css={listItemStyles}>
+			<div css={timeTextWrapperStyles}>
+				<div css={timeStyles}>{keyEvent.time}</div>
```
oliverlloyd, line 122: 👋 hello

You could make this a [time element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time), if you wanted to feed my semantic html monster 😉

common-rendering/src/components/keyEvents.tsx
```diff
@@ -0,0 +1,162 @@
...
+};
+
+const KeyEvents: FC<KeyEventsProps> = ({ keyEvents, theme }) => {
+	const [showList, setShowList] = useState(true);
+
+	return (
+		<div css={keyEventWrapperStyles}>
+			<div css={titleRowStyles}>
+				<div css={titleStyle}>Key Events</div>
+				<button
```
oliverlloyd, line 138: If we use [Button from source](https://guardian.github.io/source/?path=/story/source-src-button-button--playground) we get lots of really nice accessibility wins

common-rendering/src/keyEvent.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type paletteId = 300 | 400;
+
+export interface KeyEvent {
+	time: string;
+	text: string;
+	url: string;
+}
+
+export const getColor = (pillar: Theme, paletteId: paletteId) => {
```
oliverlloyd, line 12: You could use [decidePalette](https://github.com/guardian/dotcom-rendering/blob/a42df26f3e33fd7340bdea059a3f67bf91b678db/dotcom-rendering/src/web/lib/decidePalette.ts) here. This file is used in DCR to decide colours and has proven to be super useful!

OllysCoding: Do we need to do a separate investigation on how to use palette colours between AR & DCR?

oliverlloyd: Yeah, I think that would be a good idea.

I should point out that the reason `decidePalette` works so well is it is very common for the colour logic to change (design tweaks / new article types / etc.) and it's much easier to change things in one location - where you can see the scope of what you're working on very easily - vs. having to search the codebase for potential impacts/areas that need updating which is hard making it easy to miss things. This pattern coupled with visual regression gave us win after win after win.

oliverlloyd: Oh, and another reason `decidePalette` is my favourite thing is because it makes it really easy for non engineers to edit colours. So a designer can spot a contrast problem, say the Kicker text for Labs articles looks too grey inside a Media Card, and they can read the code themselves and fix it. They can even write additional logic if needed. The regression tests then take these changes and make sure nothing else is being affected. This is so, so great.

marjisound: Great thanks @oliverlloyd , I had seen the decidePalette file and thought that'd be a good fit for here. But as Olly said, we can update the code after we decide on how to share the decidePalette between the AR & dotcom

JamieB-gu (commented):
This is great, that was fast!

Nice job dealing with the potentially fiddly border/triangle icons btw 📐

I've added a handful of small comments/suggestions/questions.

common-rendering/src/components/keyEvents.tsx
marjisound: Thanks @oliverlloyd  When I use the source button, it's adding some hover styles, Is it possible/do we need to remove this ?

marjisound: I just added screenshots for when it's hovered

common-rendering/src/keyEvent.ts
oliverlloyd: I'm not sure we should link the use of `decidePalette` with the decision on how to share it with AR. Even if we decide to not share it with AR, we should still continue to use it in DCR because this is the convention.

We could decide not to use this convention - that's certainly an option - but until then, keeping things consistent makes the most sense I think - just in terms of maintainability and ease of understanding it's always nice if there are consistent ways for doing things.

oliverlloyd: Oh, I _just_ saw that this code is inside `common-rendering`. I _see_ now what you're talking about. Sorry, I missed that!

Hmm, maybe it would be a good idea to think about these problems - how to share code - sooner rather than later? It's great having a monorepo but if we don't implement methods for sharing code then we're only going to be co-locating code and not getting any of the actual benefits.

common-rendering/src/components/keyEvents.stories.tsx
```diff
@@ -0,0 +1,89 @@
...
+const selectPillar = (initial: Pillar): Theme => {
+	return select("Pillar", pillarOptions, initial);
+};
+
+const events: KeyEvent[] = [
+	{
+		time: "1m ago",
+		text: "Gold for Uganda",
+		url:
+			"https://www.theguardian.com/environment/2021/sep/01/opec-member-urges-oil-producers-to-focus-more-on-renewable-energy",
```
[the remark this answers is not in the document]
marjisound: Yes it is just a placeholder

common-rendering/src/keyEvent.ts
oliverlloyd: Umm, I guess the solution here is simply to hoist the utils function `decidePalette` up into a shared folder in the root. Maybe `/shared/decidePalette.ts` or `/shared/utils/decidePalette.ts`?

You _could_ make this folder an npm package but I don't think you need to right away, not unless we decide we need versioning. But this is something that be added later on as required.

I'm happy to have a bash at this if no-one has any objections? It feeds into some work I've got coming up

common-rendering/src/components/keyEvents.tsx
oliverlloyd: I really like @JamieB-gu 's idea to use [details](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details). 

But to your question about hover. In the past, [I've also had issues](https://github.com/guardian/gateway/blob/7c89002de1c8fc2cb5d06672c72f3b31b6f1081a/src/client/components/Footer.tsx#L30) with Source's hover styles and I used `cssOverrides` to remove them. I find it's good to start with too much  styling (by using Source) and remove or override it as needed, hopefully this way we're less likely to forget to implement the accessibility features that get baked in to Source.

common-rendering/package.json
```diff
@@ -5,6 +5,9 @@
     "dependencies": {
         "@emotion/react": "^11.4.1",
         "@guardian/src-foundations": "^3.9.0",
+        "@guardian/src-icons": "^3.8.0",
+        "@guardian/types": "^8.0.0",
+        "@storybook/addon-knobs": "^6.3.0",
```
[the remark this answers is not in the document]
marjisound: Thanks Jamie, I changed the story implementation and used controls instead

common-rendering/src/keyEvent.ts
marjisound: Thanks Oliver for giving great detailed info. We had a chat with everyone and the team is still unsure on the colour logic. Right now AR and dotcom are having different approaches on how to get the colour and we're trying to find what'd be the best for this common-rendering project that'd be used by both AR & DCR. 

So for now we decided to leave the colour logic within the components until we come up with a utility (either using decidePalette or a new util file) to get the colours. I think @OllysCoding is going to chat with you about this too :)

common-rendering/src/components/keyEvents.tsx
marjisound: Thanks both. I'll work on using summary/details

common-rendering/src/keyEvent.ts
oliverlloyd: Hey @marjisound sounds good 👍 and sorry for hijacking your PR!

marjisound: Not at all, I'd appreciate your time and your feedback :)

common-rendering/src/components/keyEvents.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+const titleRowStyles: SerializedStyles = css`
+	position: relative;
+	margin: ${remSpace[1]} ${remSpace[1]} 0 ${remSpace[3]};
+	display: flex;
+	justify-content: space-between;
+	align-items: center;
+
+	&:focus {
+		outline: none;
```
JamieB-gu, line 73: Would this cause problems for people using keyboard navigation?

common-rendering/src/components/keyEvents.tsx
```diff
@@ -0,0 +1,187 @@
...
+		</li>
+	);
+};
+
+const KeyEvents: FC<KeyEventsProps> = ({ keyEvents, theme }) => {
+	return (
+		<div css={keyEventWrapperStyles}>
+			<details open css={detailsStyles}>
+				<summary css={titleRowStyles}>
+					<div css={titleStyle}>Key Events</div>
```
JamieB-gu, line 166: Semantically (and therefore to the benefit of screen-readers) I think this could be a heading? Given the context I think probably an `h2`?

```suggestion
					<h2 css={titleStyle}>Key Events</h2>
```

common-rendering/src/components/keyEvents.tsx
```diff
@@ -0,0 +1,187 @@
...
+				</ul>
+			</details>
+		</div>
+	);
+};
+
+// ----- Exports ----- //
+
+export default KeyEvents;
+export { KeyEvent };
```
JamieB-gu, line 187: This can be a type export if you want:

```suggestion
export type { KeyEvent };
```

JamieB-gu (approved):
Looks great, a couple minor suggestions but otherwise 👍

OllysCoding (approved):
Looks good!!

common-rendering/src/components/keyEvents.tsx
OllysCoding: +1

common-rendering/src/components/keyEvents.stories.tsx
```diff
@@ -0,0 +1,93 @@
...
+	},
+	{
+		time: "35m ago",
+		text: "Gold for Uganda",
+		url:
+			"https://www.theguardian.com/environment/2021/sep/01/opec-member-urges-oil-producers-to-focus-more-on-renewable-energy",
+	},
+];
+
+const Default = (args: { theme: Pillar }) => (
```
oliverlloyd, line 71:
```suggestion
const Default = (args: { theme: Theme }) => (
```

common-rendering/src/components/keyEvents.stories.tsx
```diff
@@ -0,0 +1,93 @@
+// ----- Imports ----- //
+
+import { KeyEvent } from "./keyEvents";
+import KeyEvents from "./keyEvents";
+import { Pillar, Special } from "@guardian/types";
```
oliverlloyd, line 5:
```suggestion
import { Pillar, Theme, Special } from "@guardian/types";
```

common-rendering/src/components/keyEvents.tsx
```diff
@@ -0,0 +1,187 @@
+// ----- Imports ----- //
+
+import { css } from "@emotion/react";
+import type { SerializedStyles } from "@emotion/react";
+import { FC } from "react";
+import { textSans, headline } from "@guardian/src-foundations/typography";
+import { palette, remSpace } from "@guardian/src-foundations";
```
oliverlloyd, line 7: You don't need to import the entire `palette`, you can just import the objects you need, like `sport` and `news` or `text`. We did a big mob with Simon last year going through removing all these palette imports - I think the intention is to remove this export at some point to prevent this.

common-rendering/src/components/keyEvents.tsx
oliverlloyd: +1 I think even if it looks clunky we should preserve focus highlighting for accessibility (and also just for semantics)

common-rendering/src/components/keyEvents.tsx
```diff
@@ -0,0 +1,187 @@
...
+		margin: ${remSpace[3]} ${remSpace[5]} 0;
+	}
+
+	${from.desktop} {
+		margin: ${remSpace[1]} 0 0;
+	}
+
+	li::before {
+		content: "";
+		border-color: transparent ${palette.neutral[7]};
```
oliverlloyd, line 111: Could we use the `border` export from `palette` here? Just to be more aligned with the token system in Source?

common-rendering/src/components/keyEvents.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+const textStyles = (theme: Theme): SerializedStyles => css`
+	${headline.xxxsmall({ fontWeight: "regular", lineHeight: "regular" })};
+	color: ${getColor(theme, 300)};
+	text-decoration: none;
+`;
+
+const timeStyles: SerializedStyles = css`
+	${textSans.xxsmall({ fontWeight: "bold", lineHeight: "tight" })};
+	color: ${palette.neutral[7]};
```
oliverlloyd, line 144: There's `text.anchorSupporting` that might be better here?

https://theguardian.design/2a1e5182b/p/1377a6-tokens/t/85d3b0

common-rendering/src/components/keyEvents.tsx
```diff
@@ -0,0 +1,187 @@
...
+import { palette, remSpace } from "@guardian/src-foundations";
+import { SvgChevronUpSingle, SvgChevronDownSingle } from "@guardian/src-icons";
+import { Pillar, Theme } from "@guardian/types";
+import { from } from "@guardian/src-foundations/mq";
+
+// ----- Component ----- //
+type paletteId = 300 | 400;
+
+interface KeyEvent {
+	time: string;
```
oliverlloyd, line 16: Is there more information on what format this data is sent as - is it an ISO date?

common-rendering/src/components/keyEvents.tsx
```diff
@@ -0,0 +1,187 @@
...
+	${textSans.xxsmall({ fontWeight: "bold", lineHeight: "tight" })};
+	color: ${palette.neutral[7]};
+	display: block;
+`;
+
+const ListItem: FC<ListItemProps> = ({ keyEvent, theme }) => {
+	return (
+		<li css={listItemStyles}>
+			<div css={timeTextWrapperStyles}>
+				<time css={timeStyles}>{keyEvent.time}</time>
```
oliverlloyd, line 152:
```suggestion
				<time dateTime={keyEvent.time} css={timeStyles}>{keyEvent.time}</time>
```

The `time` tag requires the `datetime` attribute to be a valid, machine readable, element. The format of `keyEvent.time` must be a valid time string as defined here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time

common-rendering/src/components/keyEvents.tsx
```diff
@@ -0,0 +1,187 @@
...
+	text-decoration: none;
+`;
+
+const timeStyles: SerializedStyles = css`
+	${textSans.xxsmall({ fontWeight: "bold", lineHeight: "tight" })};
+	color: ${palette.neutral[7]};
+	display: block;
+`;
+
+const ListItem: FC<ListItemProps> = ({ keyEvent, theme }) => {
```
oliverlloyd, line 148: I realise this is subjective but the problem with using `React.FC` to type components is you're actually adding a `children` prop to this component, which isn't being used and is kind of weird. I personally prefer to be more explicit when typing things.

More opinionated [ramblings on this here](https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript).

common-rendering/src/components/keyEvents.tsx
```diff
@@ -0,0 +1,187 @@
...
+					</span>
+					<span className="is-on">
+						<SvgChevronUpSingle />
+					</span>
+				</summary>
+				<ul css={listStyles}>
+					{keyEvents.slice(0, 7).map((event) => (
+						<ListItem keyEvent={event} theme={theme} />
+					))}
+				</ul>
```
oliverlloyd, line 178: This reads really well! 👍

common-rendering/src/components/keyEvents.tsx
```diff
@@ -0,0 +1,187 @@
...
+`;
+
+const ListItem: FC<ListItemProps> = ({ keyEvent, theme }) => {
+	return (
+		<li css={listItemStyles}>
+			<div css={timeTextWrapperStyles}>
+				<time css={timeStyles}>{keyEvent.time}</time>
+				<a css={textStyles(theme)} href={keyEvent.url}>
+					{keyEvent.text}
+				</a>
```
oliverlloyd, line 155: Can we use the `Link` component from Source here? That has styles built in for hat we're trying to achieve

lucymonie (commented):
This looks great Marjan (and Joe). I'm watching the code and the conversations with interest. It seems as though we might need to broaden the conversation about decidePalette.

common-rendering/src/components/keyEvents.tsx
marjisound: I see, thanks for spotting this, do I need to do the import directly from @guardian/src-foundations such as this:
import { remSpace, culture, news, opinion, sport } from "@guardian/src-foundations";

marjisound: Oh OK thanks for spotting 👍

common-rendering/src/components/keyEvents.tsx
marjisound: @oliverlloyd This border is not really a border. It's used to create a triangle bullet point for the list items (shown below). It's colour in figma is #121212 which I couldn't find in the pallet border colours. Do you think it's OK then?

image

common-rendering/src/components/keyEvents.tsx
JamieB-gu: Key event timestamps are always relative, i.e. will always be in the format `2m ago`/`2h ago`/`2d ago`. I guess you could argue that's a presentational concern, so we could pass in an absolute `Date` and run them through a function that produces a relative timestamp within the component? Actually that's probably sensible because we can include the absolute date in a `data` attribute on the element, and then we can update the relative timestamp over time on the client?

All that said I believe @marjisound was planning to handle the time stuff, and looking into the `timeAgo` lib, in a subsequent piece of work? So perhaps we ship this as a working piece of UI and update to provide the more complex time stuff a future PR?

common-rendering/src/components/keyEvents.tsx
JamieB-gu: Interesting point 🤔. What would you use instead?

common-rendering/src/keyEvent.ts
```diff
@@ -0,0 +1,25 @@
+import { palette } from "@guardian/src-foundations";
+import { Pillar, Theme } from "@guardian/types";
+
+export type paletteId = 300 | 400;
```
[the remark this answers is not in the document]
oliverlloyd: Outside the scope of this PR but DCR mostly exports inline. This is very much one of 'those things' where people have opinions but perhaps it intersects with the one component per file rule DCR has?

If you only have one component then typically you only have one export and it becomes moot. The benefit to one component is a clear distinction between scope. The file only contains the things needed for that component which makes understanding and testing much easier.

But like I say, this is totally outside this PR - just thought I'd add a comment as it seemed mildly relevant.

common-rendering/src/components/keyEvents.tsx
oliverlloyd: Ah, I see - in this case then yes, I think you're right, this isn't a 'border' so neutral makes more sense 👍

common-rendering/src/components/keyEvents.tsx
marjisound: Yes if that's OK the plan is to handle the time in an upcoming PR by possibly using the time-ago library. 
It makes sense to make the time element format to be an absolute datetime and produce the relative timestamp within the component

oliverlloyd: You mean the `time` string here is not an absolute time but is a string like '2m ago'?

If so, that sounds like it will need to be changed to be absolute otherwise it won't be possible to keep the content on the page live.

The `timeAgo` lib should probably be implemented using the [useTimeAgo](https://github.com/guardian/dotcom-rendering/pull/2708). Except.. This file has been deleted! I've no idea why that was done 😢 it used to be live on the page.

The `useTimeAgo` custom hook (when it isn't deleted) depends on the contract where `datetime` and `data-relativeformat` are present on a `time` element. `datetime` is already the semantic html contract for these elements, and `data-relativeformat` is a contract created in Frontend many years ago and what was used when `timeAgo` was built, so there's a lot of work gone into respecting this past work when adding this hook

OllysCoding: Thanks for the info! We have a separate ticket currently to investigate the timeAgo library & any accompanying implementations to use for liveblogs, but for the sake of building components now, we're going to go back and implement it slightly further on in the dev process!

common-rendering/src/components/keyEvents.tsx
oliverlloyd: A react component is just a javascript function so I don't think you need to do anything special with it. Most of the DCR components are simply:

```
const MyThing = ({foo, bar}: Props) => { return (<p>My Thing</p>)}; 
```

oliverlloyd: Or even

```
function MyThing({foo, bar}: Props) { return (<p>My Thing</p>)}; 
```

Equally valid!

common-rendering/src/components/keyEvents.tsx
marjisound: anchorSupporting doesn't seem to exist for text. The color I need is #121212 which exists in `text.primary`, `text. anchorSecondary`, `text.userInput`, `text.inputLabel` & `text.groupLabel`

@oliverlloyd Do you think any of these be used for the `time` of the key events?

common-rendering/src/components/keyEvents.tsx
marjisound: Thanks for the detailed explanation Oliver. I will add these details to the timeAgo ticket as guidance 👍

common-rendering/src/components/keyEvents.tsx
marjisound: Thanks @oliverlloyd 
So time tag would need a `datetime` attribute although we want to show the time like `2m ago`.

If the `keyEvent.time` is a valid date and then using timeAgo library we get the `<number>m ago`  from a function, could we use the `keyEvent.time` for `datetime` attribute and use the result of our function as a string as the content of `<time>{content}</time>`?

In any case, since we are having a seperate ticket to work on the timeAgo, is that OK to leave this as it is and update it in that PR?

common-rendering/src/components/keyEvents.tsx
marjisound: I know you mentioned in DCR the `focusHalo` is added as a style in Global. But since we don't yet have this in apps-rendering, I used it within the keyevent component for the focus effect:
```
&:focus {
		${focusHalo}
	}
```

What do you think?

common-rendering/src/components/keyEvents.tsx
oliverlloyd: So my understanding here is that there are, typically, three things you can do with the return type for a React component:

#### Use `React.FC`
This is very wide. It always adds `children` which is usually wrong.

#### Specify your own type
This is just hard. You can pick from `JSX.Element`, `React.ReactNode`, `string`, `JSX.Element[]`, `null`, `undefined`, `React.ReactElement`, etc. or combinations of them but there's no one solution fits all and it's very easy to get this wrong triggering errors later on for people consuming your component.

#### Nothing, and let Typescript imply the type
In most cases, this works. Typescript is clever enough to know what the type should be using inference and in those situations where the component is more complex, you can choose to specify a type.

I, personally, like the last option - doing nothing - because it's not only easier but it means that when a type is specified it shows that the developer really wants to communicate something that I should pay attention to. But I also don't think it's that big of a deal or in any way a blocker for this PR - it's basically just a fun, interesting discussion.

common-rendering/src/components/keyEvents.tsx
oliverlloyd: So yes, everything that you just said is spot on!


```
const displayString = timeAgo(new Date(myISOdate).getTime()); // timeAgo accepts epoch time

<time dateTime={myISOdate}>{displayString}</time>
```

common-rendering/src/components/keyEvents.tsx
oliverlloyd: Looks great 👍

common-rendering/src/components/keyEvents.tsx
oliverlloyd: Oh no, the docs are maybe a bit off. Hopefully this will be resolved when everything moves to Storybook.

To be honest, I find it difficult to marry all the various use cases on a page with the list of token names available in Source. This was one of the reasons we created that `decidePalette` thing, it lets us use our own names!

I'm happy if you want to choose what you feel is best here. Be that one of the tokens or to use `neutral` directly. In general, I don't think we need to be dogmatic about using named tokens, it's just nice to think it through

common-rendering/package.json
```diff
@@ -7,7 +7,11 @@
     },
     "dependencies": {
         "@emotion/react": "^11.4.1",
+        "@guardian/src-button": "^3.8.0",
         "@guardian/src-foundations": "^3.9.0",
+        "@guardian/src-icons": "^3.8.0",
+        "@guardian/src-link": "^3.8.0",
+        "@guardian/types": "^8.0.0",
```
oliverlloyd, line 14: it's worth noting that the [types library](https://github.com/guardian/types/blob/main/README.md) has been deprecated. It got replaced by `@guardian/libs`

oliverlloyd (approved):
Really great work! 🏅 

Thanks loads for all your patience with all my questions - I learnt a lot on this PR!

common-rendering/package.json
marjisound: Ah OK, I'll take a note to change it and use guardian/libs in the coming coming PRs. Thanks for letting us know

PR #3370, opened 2021-09-03
LIVE-2953: swap undefined for empty string

joecowton1:
## Why are you doing this?

If we return `succeed(undefined)` in the case of an empty `venue` field the component does not  render, switching it for `''` means the `footballScores` component still renders.
```
oneOf([fieldParser('venue', stringParser), succeed('')]),
```

## Changes
- swap `undefined` for `''` if no `venue` or `additionalInfo` field

## Screenshots

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/apps-rendering/src/server/footballContent.ts
+++ b/apps-rendering/src/server/footballContent.ts
@@ -95,7 +95,7 @@ const makeFootballContent = (
 const scorerParser: Parser<Scorer> = map3(makeScorer)(
 	fieldParser('player', stringParser),
 	fieldParser('timeInMinutes', numberParser),
-	oneOf([fieldParser('additionalInfo', stringParser), succeed(undefined)]),
+	oneOf([fieldParser('additionalInfo', stringParser), succeed('')]),
 );
 
 const footballTeamParser: Parser<FootballTeam> = map6(makeFootballTeam)(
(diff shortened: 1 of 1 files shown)


davidfurey (approved):
It might be better for the component to decide how to render an `undefined` venue or additionalInfo, rather than this being swallowed by the parser.  But I'd need to checkout the project and getting it running locally to work out how to do that.

I don't want to block you fixing the issue, what you've done works so 👍

JamieB-gu:
> It might be better for the component to decide how to render an `undefined` venue or additionalInfo, rather than this being swallowed by the parser.

I think I agree with this. If it's valid for the venue to be missing then it should explicitly be an `Option` in the `MatchScores` type, and the component can handle that visually.

joecowton1:
> > It might be better for the component to decide how to render an `undefined` venue or additionalInfo, rather than this being swallowed by the parser.
> 
> I think I agree with this. If it's valid for the venue to be missing then it should explicitly be an `Option` in the `MatchScores` type, and the component can handle that visually.

Ok thanks both. I'm not going to merge this now given the time and the fact I'm not working and haven't had chance to test it properly. But I agree we should fix it properly.

@abeddow91 maybe this is another one you can have a look into properly if you've time next week? Otherwise I'll look into into when I'm back the week after :)

PR #3371, opened 2021-09-03
Submit consent events to Ophan

shtukas:
This is the first of few PRs all aimed at submitting consent related events to Ophan, notably the initial capture (if any) and user driven update of user consent.  At the moment we are only registering the consentState.tcfv2 case. The other cases will be introduced shortly. 

This PR comes after the Ophan PR to update the component event model with a new action and a new component type: https://github.com/guardian/ophan/pull/4257

### DIFF
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -324,6 +324,17 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 				}
 			});
 
+			onConsentChange((consentState) => {
+				const consentUUID = getCookie('consentUUID')
+				console.log("browserId", browserId);
+				console.log("consentUUID", consentUUID);
+				if (consentState.tcfv2) {
+					console.log("[tcfv2] consent string", consentState?.tcfv2?.tcString);
+				} else {
+					console.log("Not tcfv2")
+				}
+			});
+
 			cmp.init({
 				country: countryCode,
 				pubData,
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -330,7 +330,7 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 					if (consentState.tcfv2) {
 						return consentState.tcfv2?.tcString;
 					}
-					return "undefined";
+					return "";
 				}
 				const consentUUID = getCookie('consentUUID')
 				const consentString = decideConsentString();
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -70,6 +70,7 @@ import { OphanRecordFunction } from '@guardian/ab-core/dist/types';
 import { ConsentState } from '@guardian/consent-management-platform/dist/types';
 import { storage } from '@guardian/libs';
 import {
+	getOphanRecordFunction,
 	submitComponentEvent,
 	OphanComponentEvent,
 } from '../browser/ophan/ophan';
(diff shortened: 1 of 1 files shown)


oliverlloyd:
This is interesting... I had a few thoughts I was going to post as a review here but then decided to spend a bit more time on it. Watch this space...

dotcom-rendering/src/web/components/App.tsx
```diff
@@ -324,6 +325,27 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 				}
 			});
 
+			onConsentChange((consentState) => {
```
nicl, line 328: Subjective, but I think this could be made more readable through some added whitespace and also inlining and renaming. E.g.

```
onConsentChange((consentState) => {
	const consentUUID = getCookie('consentUUID')
	const consentString = consentState?.tcfv2?.tcString || "undefined";

	const event = {
		component: {
		  componentType: 'CONSENT',
		  products: [],
		  labels:  [consentUUID, consentString],
		},
		action: 'CONSENT',
	}

	const record = getOphanRecordFunction()
	record(event);
});
```

Caveat - `consentString` is slightly different than before in behaviour - you may not want this!

nicl (approved):

dotcom-rendering/src/web/components/App.tsx
```diff
@@ -324,6 +325,27 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
...
+				const decideConsentString = () => {
+					if (consentState.tcfv2) {
+						return consentState?.tcfv2?.tcString;
```
nicl, line 331: Note, you have a typo here - t`c`String -> t`o`String.

oliverlloyd (commented):
I left a couple of small comments on the code but nothing major.

One thing that does concern me here is the conflict between react's declarative pattern and the imperative nature for how we integrate cmp.

The `useEffect` that we're using here currently runs twice. The first time `countryCode` is undefined and then it gets set so the effect is rerun. This is fine right now because we check if this var is defined before adding the listener and initialising cmp. But this feels very loose and open to change. It happens to work now, but changes in future could change this in subtle ways.

I realise this is outside the immediate scope of this PR but can we replace the `useEffect` with a `useOnce` function instead? Like this the behaviour that we have now by chance becomes intentional, which makes the code easier to read and understand (assuming you know and understand how useOnce works (it's [pretty simple](https://github.com/guardian/dotcom-rendering/blob/a42df26f3e33fd7340bdea059a3f67bf91b678db/dotcom-rendering/src/web/lib/useOnce.ts#L8)).

The change from `useEffect` to `useOnce` should be straight forward, you just need to be sure to move the deps array from the useEffect into the `waitFor` param for useOnce.

As I say, this is outside the scope of this change but maybe now that we're adding this additional logic and complexity it is worth this refactor?

oliverlloyd:
Another thought. Is true to say that what we're doing here is adding functionality that means that when a reader interacts with the page in a certain way - namely, setting their consent status via cmp, then we now expect an api call to be made logging this?

If so, can we add a Cypress test for this? That should be fairly simple to do using `cy.intercept()`. There are I believe existing examples of tests that interface with CMP.

dotcom-rendering/src/web/components/App.tsx
shtukas: @nicl 

Actually it's `tcString`

Screenshot 2021-09-03 at 16 02 43

dotcom-rendering/src/web/components/App.tsx
```diff
@@ -324,6 +325,27 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
...
+					}
+					return "undefined";
```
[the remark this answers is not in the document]
shtukas: @oliverlloyd 
Oh Yeah. That's a good point. It's actually a work in progress for me to...., well..., define the spec of the data sent to Ophan. Something I will clarify soon (not needed for the moment) when I start sending the full spectrum of data from other legislations, but you are right, I can easily not use "undefined".

dotcom-rendering/src/web/components/App.tsx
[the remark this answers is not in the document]
shtukas: I was under the impression that we can register more than one callback against `onConsentChange`. I didn't want to touch the existing one.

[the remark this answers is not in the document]
oliverlloyd: Oh sure, you definately _can_, and we do, it's already used in one other place. I was just wondering if we _should_. I was just thinking about if there were performance or readability benefits to centralising things 🤷

dotcom-rendering/src/web/components/App.tsx
nicl: wat. Okay thanks!

mxdvl: As in https://iabtcf.com/api/core/classes/tcstring.html

PR #3372, opened 2021-09-03
Fix Bonzai TrueSkin page skin

arelra:
## Why?

Hides the top-above-nav ad-slot container when a Bonzai TruSkin (Australian 3rd Party page skin) is shown.

The commercial team appreciates this fix is a blunt instrument but we are losing revenue until the vendor applies a fix on their side.

This is a **temporary fix** and we will revert asap.

Frontend PR: https://github.com/guardian/frontend/pull/24146

## Before

image


OllysCoding (approved):
+1 Do we have a date when this should be removed by? sometimes worth adding that to the comment in case it's forgotten, though I assume in this case it won't be

arelra:
> +1 Do we have a date when this should be removed by? sometimes worth adding that to the comment in case it's forgotten, though I assume in this case it won't be

Thanks. We don't have a fixed date atm but will update as soon as we have one.

PR #3373, opened 2021-09-03
Storybook Tidy: Prefix AR storybook

OllysCoding:
## What does this change?

Prefix all AR stories with `AR/` to move them into their own subfolder

A second PR could be raised to move all of the DCR stories into a `DCR/` subdirectory - this would be a much larger process

## Why?

There was no way to distinguish between AR and DCR stories

### Before

image

### After

image


marjisound (approved):
Great stuff 👍

lucymonie (approved):
👍🏻

PR #3378, opened 2021-09-06
AR Layouts

JamieB-gu:
## Why?

DCR has the different article layouts consolidated in a single `layouts` directory, whereas AR has them scattered across several directories for historical reasons. This updates AR to bring it more in line with the conventions in DCR, i.e. a `layout` directory to house all the different layouts.

This has two benefits: it makes AR simpler, and it means AR and DCR have more in common. Hopefully this will make the codebase easier to understand.

## Changes

- Brings AR conventions in line with DCR
- Moves the disparate article layouts under the `layout` component
- Moves `labsLogo` and `cutout` to the top level of components


abeddow91 (approved):
LGTM!

PR #3381, opened 2021-09-06
Enable TCF API on AMP

chrislomaxjones:
## What does this change?

Enable access to the TCF API on AMP pages.

## Why?

This will allow Teads to access the TCF API and TC string on the page (being in a SafeFrame).


mxdvl (approved):
Looking good :)

This should also expose CCPA.

PR #3384, opened 2021-09-06
Enable prebid on mobile sticky ads

chrislomaxjones:
## What does this change?

Enable prebid on mobile sticky ads.

## Why?

Allow SSPs to bid on mobile sticky slots on AMP, hopefully leading to a revenue uplift.

### DIFF
--- a/dotcom-rendering/src/amp/components/Ad.tsx
+++ b/dotcom-rendering/src/amp/components/Ad.tsx
@@ -26,9 +26,24 @@ const ampData = (section: string, contentType: string): string => {
 	return `/${dfpAccountId}/${dfpAdUnitRoot}/amp`;
 };
 
+/**
+ * Determine the Placement ID that is used to look up a given stored bid request
+ *
+ * Stored bid requests are stored by the prebid server instance and each is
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/amp/components/Ad.tsx
```diff
@@ -30,10 +30,26 @@ const ampData = (section: string, contentType: string): string => {
...
+ * @returns The placement id for an ad, depending on its ad region and whether
+ * it is sticky
+ */
+const getPlacementId = (isSticky: boolean, adRegion: AdRegion): number => {
 	switch (adRegion) {
-		case 'US':
-			return 7;
+		case 'US': {
+			// In the US use different placement IDs depending on whether ad is sticky
+			return isSticky ? 22138171 : 7;
```
chrislomaxjones, line 51: 👆 note we know longer have a mapping directly from ad region to placement id, we now also take into account whether an ad is sticky.

mxdvl (approved):
LGTM!

Interesting to ponder whether we'll be able to do sticky regional ads…

PR #3385, opened 2021-09-06
Refactor ad region logic

chrislomaxjones:
## What does this change?

Move the logic that renders ads based on regions into the `<Ad />` component. This component renders a `<RegionalAd />` component for each ad region (currently US, AU and ROW) which has the amp-geo classes applied to ensure only the component in the user's region is displayed.

## Why?

Previously the responsibility for ensuring that ads for each region were included fell upon the component that rendered them. This change means the renderer is required to simply include one component regardless of region and leaves the responsibility of rendering the correct regional ad to `<Ad />`. This should also make it simpler to add/modify regions.


nitro-marky (approved):
LGTM

PR #3387, opened 2021-09-07
`Page`

oliverlloyd:
## What does this change?
Adds the `Page` component. An abstraction broken out from `document.tsx`.

```ts
/**
 * @description
 * Page is a high level wrapper for pages on Dotcom. Sets strict mode, some globals
 * and the Emotion cache
 *
 * @param {ReactNode} children - What gets rendered on the page
 * @param {EmotionCache} cache - Provides the Emotion cache
 * */
```

## Why?
By breaking out the logic for what defines a page in DCR we have the ability to consider other use cases in the future. We will probably always want strict mode on, centralised global css and an emotion cache set but perhaps we want to pass in other content, such as a Newsletters page? This supports that use case.


OllysCoding (approved):
👍 I think this makes sense

PR #3388, opened 2021-09-07
Use Jest For Testing `common-rendering`

JamieB-gu:
## Why?

We want to be able to test our common code. This will also pave the way for bringing `image-rendering` into this repo.

## Changes

- Added `jest`, `@types/jest` and `ts-jest` as dependencies
- Added test script for common-rendering
- Configured Jest to use ts-jest
- Added example tests


marjisound (approved):
this is great 👍

PR #3391, opened 2021-09-07
Migrate `image-rendering` Into `common-rendering`

JamieB-gu:
## Why?

There are a few reasons behind this:

1. This project was only a package because we planned to share it between AR and DCR. Now that they're in the same repo it can exist here too.
2. We're looking to migrate this entire repo over to Yarn, and Yarn has some issues with the `prepare` script `image-rendering` relies on.
3. We like to share the way images are rendered on liveblogs, and `image-rendering` will need updating to handle this. That will be easier to do now it's in the same repo.

FYI @paperboyo 

## Changes

- Added testing dependencies and `@guardian/types`
- Allowed Jest transformation of guardian packages
- Tweaked `tsconfig` for import of guardian packages JS and test deps
- Added the `image-rendering` files, components, stories and tests

### DIFF
--- a/common-rendering/package.json
+++ b/common-rendering/package.json
@@ -8,14 +8,20 @@
     "dependencies": {
         "@emotion/react": "^11.4.1",
         "@guardian/src-foundations": "^3.9.0",
-        "react": "^17.0.2"
+        "react": "^17.0.2",
+        "@guardian/types": "^8.0.0"
     },
     "devDependencies": {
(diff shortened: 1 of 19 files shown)


marjisound (approved):
The PR seemed to be big at first glance but then I realized you have copied the files from image-rendering into common-rendering. Other than that is  storyshots, package.json and ts.config which all look fine to me 👍

OllysCoding (approved):
Looks good!

PR #3376, opened 2021-09-06
[AR] Migrate from `@guardian/types` to `@guardian/libs`

jamie-lynch:
## What does this change?

This PR migrates from the [`@guardian/types`](https://github.com/guardian/types) library to [`@guardian/libs`](https://github.com/guardian/libs) as [`@guardian/types` is now deprecated](https://github.com/guardian/types/pull/140).

## How to test

Run all validation and ensure that everything still runs as expected.

## TODO

- [x] Update the version of `@guardian/discussion-rendering` after https://github.com/guardian/discussion-rendering/pull/520 is released
- [x] Update the version of `@guardian/atoms-rendering` after https://github.com/guardian/atoms-rendering/pull/280 is released
- [x] Update the version of `@guardian/image-rendering` after https://github.com/guardian/image-rendering/pull/322 is released

PR #3375, opened 2021-09-06
[DCR] Migrate from `@guardian/types` to `@guardian/libs`

jamie-lynch:
## What does this change?

This PR migrates from the [`@guardian/types`](https://github.com/guardian/types) library to [`@guardian/libs`](https://github.com/guardian/libs) as [`@guardian/types` is now deprecated](https://github.com/guardian/types/pull/140).

## How to test

Run all validation and ensure that everything still runs as expected.

## TODO

- [ ] Update the version of `@guardian/discussion-rendering` after https://github.com/guardian/discussion-rendering/pull/520 is released
- [ ] Update the version of `@guardian/braze-components` after https://github.com/guardian/braze-components/pull/224 is released
- [ ] Update the version of `@guardian/atoms-rendering` after https://github.com/guardian/atoms-rendering/pull/280 is released

PR #3400, opened 2021-09-08
Use Local Image-Rendering

JamieB-gu:
## Why?

Following up on #3391, now we have `image-rendering` in the mono-repo we can start using the local files instead of the package.

**Note:** There's currently an issue with the tests. I _think_ it might be because there are two instances of `React` (one from `common-rendering` and one from AR) and that's confusing one of the hooks implementations.

## Changes

- Removed `@guardian/image-rendering` dep
- Migrated AR imports over to local image files
- Switched image components over to default exports
- Made `Dpr` a normal `enum` because DCR doesn't support `const enum`s

### DIFF
--- a/.github/workflows/ar-nodejs.yml
+++ b/.github/workflows/ar-nodejs.yml
@@ -21,6 +21,10 @@ jobs:
             - uses: actions/checkout@v1
             - uses: guardian/actions-setup-node@main
 
+            # Cache npm dependencies using https://github.com/bahmutov/npm-install
+            # Root dependencies
+            - uses: bahmutov/npm-install@v1
+
             - uses: actions/cache@v2
               with:
                   path: ~/.npm
--- a/apps-rendering/package-lock.json
+++ b/apps-rendering/package-lock.json
@@ -6026,10 +6026,6 @@
         }
       }
     },
-    "@guardian/image-rendering": {
-      "version": "github:guardian/image-rendering#1167f8bc1f1b7f65ba883e4678e10190abbabb05",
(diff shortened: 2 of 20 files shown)


JamieB-gu:
FYI @jamie-lynch

jamie-lynch:
> FYI @jamie-lynch

Nice! 🎉 

In this case maybe it's best to close #3376 and #3375 and make all the changes in one PR?

JamieB-gu:
> In this case maybe it's best to close #3376 and #3375 and make all the changes in one PR?

I think that might be easier 🤔

> Also, are there plans to move any of the other *-rendering projects into here?

That might be possible, yeah. AFAIK `atoms-rendering` and `discussion-rendering` are only used by DCR and AR, so if both projects can use them without going via the publish/release/update cycle presumably that will be easier? Also it may make it easier to keep their shared dependencies in sync, which could make it simpler to do things like the Emotion upgrade.

jamie-lynch:
> In this case maybe it's best to close #3376 and #3375 and make all the changes in one PR?

@OllysCoding and I are going to try and get these over the line today/tomorrow so we'll see how it goes. Give us a shout if you fancy doing some mobbing on it.

> The other option is to fold some of the Editorial types like Role into this repo

We're also using some of these in Source at the minute, albeit in Editorial components that may also move into this repo in time. For now, I'd prefer to get this work over the line and then we can review whether libs is the best place for some of these things in future.

JamieB-gu:
Closing in favour of #3472.

PR #3402, opened 2021-09-09
live blog KeyEvent component dark mode

marjisound:
## What does this change?
This PR adds dark mode styles for the KeyEvent component. The caller of the component can decide if dark mode is supported or not. If dark mode is supported, and if user device is set to dark mode, styles are applied and.

This PR also changes the way themes are shown in stories so instead of using controls, all the themes are shown in one storybook. 

Paired on this: @abeddow91 & @marjisound & @OllysCoding 

### Light mode desktop
image

### Dark mode desktop
image


OllysCoding (approved):

oliverlloyd (commented):
Can we add stories showing the different theme/dark mode combinations? Normally when I'm adding stories for a component I try to make sure I include examples for all the props.

Storybook is useful as an aid to development but it really comes into its own as a way to document all the visual variations that a component can have. Add onto that the regression testing we get with Chromatic and it's a real force multiplier in terms of benefit

Note. I don't think you _have to_ have a separate story for each theme or each dark mode. You can group things inside a single story. This pattern is used quite a lot in other projects and I think would help here

oliverlloyd (commented):
Here's an example of how you can use Storybook + Chromatic to visually document a component in all it's possible combinations

https://www.chromatic.com/component?appId=5dfcbf3012392c0020e7140b&name=Components%2FCard%2FDesigns&buildNumber=7884&componentInspectorKey=613a14d75a5665003ae91a92-1200-interactive-true&historyLengthAtIndex=8&distanceToMoveBack=-1

This has been really useful as a way to showcase work not only in the dept. but within the editorial team as well. It also catches a lot of bugs!

lucymonie (approved):
LGTM 👍🏻

common-rendering/src/components/keyEvents.stories.tsx
```diff
@@ -68,26 +59,43 @@ const events: KeyEvent[] = [
 	},
 ];
 
-const Default = (args: { theme: Theme }) => (
-	<KeyEvents keyEvents={events} {...args} />
+const keyEventComp = (dark: boolean, theme: Theme, title: string) => (
```
oliverlloyd, line 62: Don't React components have to start with a capital letter? Or is that just in my head?

common-rendering/src/components/keyEvents.stories.tsx
```diff
@@ -68,26 +59,43 @@ const events: KeyEvent[] = [
...
+			justify-content: space-between;
+			flex-wrap: wrap;
+			-ms-flex-positive: 1;
+		`}
+	>
+		{keyEventComp(dark, Pillar.News, "News")}
+		{keyEventComp(dark, Pillar.Culture, "Culture")}
+		{keyEventComp(dark, Pillar.Lifestyle, "Lifestyle")}
+		{keyEventComp(dark, Pillar.Opinion, "Opinion")}
+		{keyEventComp(dark, Pillar.Sport, "Sport")}
```
oliverlloyd, line 87: I know we're not including `Labs` and `Special` themes here because the designs haven't been done for them but  we should probably support them being passed in because they are both valid values for `theme`. Maybe they can default to "News" while we wait for the colours to be worked out?

common-rendering/src/components/keyEvents.stories.tsx
```diff
@@ -68,26 +59,43 @@ const events: KeyEvent[] = [
...
+`;
+
+const keyEventWithTheme = (dark: boolean) => () => (
+	<div
+		css={css`
+			display: flex;
+			flex-direction: row;
+			justify-content: space-between;
+			flex-wrap: wrap;
+			-ms-flex-positive: 1;
```
oliverlloyd, line 80: Is this needed? I think these browser specific properties get added in the build

oliverlloyd (approved):
Looks great!

Just a few small things 👍

common-rendering/src/components/keyEvents.stories.tsx
OllysCoding: +1 I think line 80 can be removed :)

common-rendering/src/components/keyEvents.stories.tsx
OllysCoding: They do, apparently

common-rendering/src/components/keyEvents.stories.tsx
marjisound: They are actually defaulted to News within the `KeyEvents` component. So I just added the them to the storybook options.

oliverlloyd: Ah, I see. Cool, cool, cool

PR #3403, opened 2021-09-09
increase h3 sizing on numbered lists

HarryFischer:
## What does this change?
Increases size of H3 in Numbered List articles

## Why?
Was too small

### Before
Screenshot 2021-09-09 at 12 42 02

### After
image


nitro-marky (approved):
👍

PR #3404, opened 2021-09-09
Bump Atoms to 18.0.2

nitro-marky:
## What does this change?

Adds the comment back after the score, the hypothesis now is that it could have been a user related issue rather than a bug.


DavidLawes (approved):
👍

PR #3406, opened 2021-09-09
make h3 a little smaller

HarryFischer:
## What does this change?
Makes h3 use xsmall font size rather than small
## Why?
Change it but was a bit too big in some situations
### Before
image

### After
image


OllysCoding (approved):
LGTM!

PR #3407, opened 2021-09-09
No global h2 styles when `Design.NumberedList`

oliverlloyd:
## What does this change?
Ensures that the global h2 css rules that are defined in `ArticleBody` do not get applied for Numbered Lists

## How?
We're employing the `data-ignore="global-h2-styling"` contract which was established to allow use cases such as this one where we might want to opt out of the global css rules.

## Why?
Because these article types have an established contract using `h2` tags that should result in unique styles but this was being affected by the global h2 styles that tried to do different things.

### Before
Screenshot 2021-09-09 at 13 57 21

### After
Screenshot 2021-09-09 at 13 54 48

### DIFF
--- a/dotcom-rendering/src/model/enhance-numbered-lists.ts
+++ b/dotcom-rendering/src/model/enhance-numbered-lists.ts
@@ -198,6 +198,37 @@ const isItemLink = (element: CAPIElement): boolean => {
 	return hasULWrapper && hasOnlyOneChild && hasLINestedWrapper;
 };
 
+const removeGlobalH2Styles = (elements: CAPIElement[]): CAPIElement[] => {
+	/**
+	 * Article pages come with some global style rules, one of which affects h2
+	 * tags. But for numbered lists we don't want these styles because we use
+	 * these html elements to represents our special titles. Rather than start a
+	 * css war, this enhancer uses the `data-ignore` attribute which is a contract
+	 * established to allow global styles to be ignored.
+	 *
+	 * All h2 tags inside an article of Design: NumberedList have this attirbute
+	 * set.
+	 */
+	const withH2StylesIgnored: CAPIElement[] = [];
+	elements.forEach((thisElement) => {
+		if (
+			thisElement._type ===
+				'model.dotcomrendering.pageElements.SubheadingBlockElement'
+		) {
+			withH2StylesIgnored.push(
+				{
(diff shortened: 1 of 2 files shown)


OllysCoding (approved):

dotcom-rendering/src/model/enhance-numbered-lists.ts
```diff
@@ -198,6 +198,37 @@ const isItemLink = (element: CAPIElement): boolean => {
...
+					html: thisElement.html.replace('<h2>', '<h2 data-ignore="global-h2-styling">'),
+				},
+			);
+		} else {
+			// Pass through
+			withH2StylesIgnored.push(thisElement);
+		}
+	});
+	return withH2StylesIgnored;
+};
```
OllysCoding, line 230: Can you give some examples of articles this might affect?

oliverlloyd: Here you go: https://www.theguardian.com/lifeandstyle/2021/mar/24/march-design-news-ceramics-uk-lgbt-retirement-homes-and-stan-smiths-forever

As per the before/afters, these H2 tags were getting these global styles applied and preventing the original intent of the designs from appearing properly. If you [look at Storybook](https://www.chromatic.com/component?appId=5dfcbf3012392c0020e7140b&name=Components%2FNumberedTitleBlockComponent&buildNumber=7897&componentInspectorKey=613b37e5faa5b8003a19cdff-1200-interactive-true&historyLengthAtIndex=4&distanceToMoveBack=-1) you will see what the expected outcome should be but this wasn't happening in the wild.

OllysCoding: 👍 If there is a discrepancy in storybook, should we be looking to solve this?

oliverlloyd: You're right, we should.

One option might be to set them higher up, in the `Page` component along with the other global styles. If we did this then it would be simple enuogh to wrap stories in this component as a sort of template. But there's a risk that this will expand their scope causing unintended styling diffs. I guess we could try that in a PR and see what Chromatic brings back?

Alternatively, we could look to selectively apply these global styles on a story by story basis but that's not a great solution as it would need ongoing attention and eventually would end up being missed

PR #3410, opened 2021-09-10
Update Atoms Rendering to 18.0.3

nitro-marky:
## What does this change?

This reworks the logic for knowledge quizzes. Previously batched scores had not been accounted for i.e. scoring between 1 and 5. This will account for a response to every question and batched questions.

## Images

Screenshot 2021-09-09 at 16 57 56

Screenshot 2021-09-09 at 16 58 42

Screenshot 2021-09-09 at 16 59 24


DavidLawes (approved):
👍

PR #3412, opened 2021-09-10
Copy the root .gitignore the into the dotcom-rendering sub-directory

shtukas:
This will prevent *lots* of files to show up in a PR when running Cypress tests on local 😬

### DIFF
--- a/dotcom-rendering/.gitignore
+++ b/dotcom-rendering/.gitignore
@@ -1,0 +1,48 @@
+### Node template
+# Logs
+logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# Dependency directories
+node_modules
+
+# Yarn Integrity file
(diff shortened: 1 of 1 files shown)


dotcom-rendering/.gitignore
```diff
@@ -0,0 +1,48 @@
...
+
+# .git-hooks
+# Ignore auto generate husky hooks
+scripts/git-hooks/*
+!scripts/git-hooks/post-merge
+!scripts/git-hooks/install.js
+
+# Cypress result files
+/cypress/screenshots/*
+/cypress/videos/*
```
OllysCoding, line 42: Instead of copying the full file, could we update these lines in the root gitignore to be prefixed `/dotcom-rendering`? 

This would avoid duplication and hopefully be a bit cleaner

shtukas: Done.

PR #3413, opened 2021-09-10
Refactor CMP Initialisation

oliverlloyd:
## What does this change?
Refactors how we initialise CMP to be a bit more react-y (eg. More declarative)

I also added the `bwid` cookie to some Cypress tests around CMP that were managing to pass before but only because of a quirk, one which is now removed by this PR. The `browserId` state wasn't a strict dependency for `cmp.init` beforehand so not having it (you get it from the `bwid` cookie) was working. But now that we have refactored things to ensure we only initialise cmp once, and have added a strict dependency on this cookie, it had to be added to Cypress.

### DIFF
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -721,11 +721,11 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 				CAPI.config.switches.commercialMetrics,
 				window.guardian.config?.ophan !== undefined,
 			].every(Boolean) && (
-			<CommercialMetrics
-				browserId={browserId}
-				pageViewId={pageViewId}
-					/>
-				)}
+				<CommercialMetrics
+					browserId={browserId}
+					pageViewId={pageViewId}
+				/>
+			)}
 			<Portal rootId="reader-revenue-links-header">
 				<ReaderRevenueLinks
 					urls={CAPI.nav.readerRevenueLinks.header}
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/App.tsx
```diff
@@ -291,72 +295,73 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
...
+				.then((willShow) => {
+					trackPerformance(
+						'consent',
+						'acquired',
+						willShow ? 'new' : 'existing',
+					);
+				})
+				.catch((e) =>
+					console.error(`CMP willShowPrivacyMessage - error: ${e}`),
+				);
```
mxdvl, line 342: I think this logic should be part of the `useOnce`, as it captures how long it takes for the initial state to be available.

mxdvl (changes requested):
Generally I think this is much easier to reason about and makes the CMP declarative.

There’s a small change needed around initialisation performance 👇.

dotcom-rendering/src/web/components/App.tsx
oliverlloyd: Done. I moved the `cmp.willShowPrivacyMessage` promise into the `useOnce`. Note. I also stripped the use of the `recordedConsentTime` boolean as this code by design only runs once now.

mxdvl (approved):

shtukas (approved):

shtukas:
Thanks @oliverlloyd

PR #3414, opened 2021-09-10
Add Header component for Liveblogs

lucymonie:
## What does this change?
This adds a WIP Header component for Liveblogs. It is a work in progress that can been incorporated into the codebase as we're not using it yet. It includes a basic live colour palette. 

Yet to come are:
- post-live or 'dead' blogs colour palette
- dark mode (for apps)
- timeAgo functionality for the UPDATED element (for dotcom)

## Why?
We are building Liveblogs components in a location where they can be shared by apps-rendering and dotcom-rendering.

### Screenshots
mobileMedium
Screen Shot 2021-09-10 at 15 15 15

phablet
Screen Shot 2021-09-10 at 15 15 30

tablet
Screen Shot 2021-09-10 at 15 15 42

desktop
Screen Shot 2021-09-10 at 15 15 55

wide
Screen Shot 2021-09-10 at 15 16 14


common-rendering/src/components/header.tsx
```diff
@@ -0,0 +1,244 @@
...
+`;
+
+const updatedStyles: SerializedStyles = css`
+	color: ${text.ctaPrimary};
+	${textSans.xxsmall()};
+`;
+
+// The following pulsing dot implementation is a placeholder for the PulsingDot component, which
+// we will ideally lift up from dcr to common-rendering
+const allowsAnimation = "@media (prefers-reduced-motion: no-preference)";
```
OllysCoding, line 142: would be nice to export this from somewhere at some point

lucymonie: Yes, definitely. In the note above, I mentioned that this code will be replaced with the PulsingDot component once we can share code between common-rendering and the other projects. :-)

nitro-marky: Yeah, I think for now putting everything you need in the same file is fine. There's going to be an element of fixing forward regardless.

common-rendering/src/components/header.tsx
```diff
@@ -0,0 +1,244 @@
...
+// End of temporary pulsing dot implementation
+
+const Header: FC<HeaderProps> = ({
+	theme,
+	title,
+	credit,
+	standfirst,
+	byline,
+}) => {
+	const [themeColours] = useState<Colours>(() => getThemeColours(theme));
```
marjisound, line 179: This is just a question as I'm not super experienced in react. But I'm just wondeing why we need a state for themeColors? can it not be a function that would get the colour when the component is rendered?

common-rendering/src/components/header.tsx
```diff
@@ -0,0 +1,244 @@
...
+	// TODO: implment timeAgo in the 'LIVE Updated' section, and replace pulsing dot
+	return (
+		<>
```
marjisound, line 182: Would it not be better to use an html `header` instead of `<>`?

PR #3417, opened 2021-09-13
Quick rota fix to remove Journalists Avatar When Author has no byline

nitro-marky:
## What does this change?

This adds a bool check which tells the ArticleMeta whether it should show an avatar or not. It's mainly meant for times where there is no byline in the author data but still has a contributor listed. [See Here](https://www.theguardian.com/music/2021/sep/10/rewrite-rule-britannia-what-would-you-do-with-the-last-night-of-the-proms)

## Before
Screenshot 2021-09-14 at 10 11 14

## After
Screenshot 2021-09-14 at 10 11 49

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleMeta.tsx
+++ b/dotcom-rendering/src/web/components/ArticleMeta.tsx
@@ -272,7 +272,7 @@ export const ArticleMeta = ({
 	console.log('byline' in author);
 
 	const showAvatarFromAuthor = () => {
-		if ('byline' in author) {
+		if ('byline' in author && format.design !== Design.Comment) {
 			return true;
 		}
 		return false;
--- a/dotcom-rendering/src/web/components/ArticleMeta.tsx
+++ b/dotcom-rendering/src/web/components/ArticleMeta.tsx
@@ -23,6 +23,7 @@ type Props = {
 	primaryDateline: string;
 	secondaryDateline: string;
 	branding?: Branding;
+	hasByline?: boolean;
 };
 
 const meta = css`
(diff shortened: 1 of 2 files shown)


dotcom-rendering/src/web/components/ArticleMeta.tsx
```diff
@@ -23,6 +23,7 @@ type Props = {
 	primaryDateline: string;
 	secondaryDateline: string;
 	branding?: Branding;
+	hasByline?: boolean;
```
OllysCoding, line 26: can this line be removed?

dotcom-rendering/src/web/layouts/StandardLayout.tsx
```diff
@@ -562,6 +562,7 @@ export const StandardLayout = ({ CAPI, NAV, format, palette }: Props) => {
 								secondaryDateline={
 									CAPI.webPublicationSecondaryDateDisplay
 								}
+								hasByline={!!CAPI.author.byline}
```
OllysCoding, line 565: and this could also be removed I think?

dotcom-rendering/src/web/components/ArticleMeta.tsx
OllysCoding: Nevermind seems like github has just bugged out!

nitro-marky: This is part of the temporary fix.

nitro-marky (commented):
Yeah, the tests seems to be freaking out here

dotcom-rendering/src/web/components/ArticleMeta.tsx
```diff
@@ -260,10 +260,18 @@ export const ArticleMeta = ({
 	const bylineImageUrl = getBylineImageUrl(tags);
 	const authorName = getAuthorName(tags);
 
+	const showAvatarFromAuthor = () => {
+		if (!('byline' in author) || author.byline === '') {
```
jlieb10, line 264: is this the same as `(!author.byline || author.byline === '')` - i find this consistency of checking for attrs a bit better/more readable if it evaluates to the same thing

jlieb10 (approved):
non-blocking thought/suggestion - nice work

PR #3418, opened 2021-09-13
1% server-side A/B test: remove sticky behaviour from navigation bar to measure ad viewability

MarSavar:
## What does this change?
This PR introduces a change in the server-side rendering of `StandardLayout` that removes the stickiness of the navigation bar. 
This is achieved by moving the navigation bar, the sub-navigation bar and the line divider within the same div as the `Stuck` component for users in the experiment.
In order to achieve visual consistency with `frontend`, an optional `showBottomBorder` prop has been added to the `ElementContainer` component. This can be removed after the experiment is over, if no longer needed.

`frontend` PR introducing the experiment: https://github.com/guardian/frontend/pull/24165

## Why?
The hypothesis is that making the navigation bar non-sticky, there will be an improvement in the [viewability](https://support.google.com/google-ads/answer/7029393?hl=en-GB) score of the `top-above-nav` ad slot.
This might have a slight impact on user engagement as the top navigation bar is the primary form of interaction in article pages (see https://github.com/guardian/dotcom-rendering/pull/3241).
To mitigate this, the A/B test will only run for a small percentage of users (currently 1%), and will only affect  `StandardLayout`. 

The following layouts are **not** affected by this A/B test, and users will continue seeing the sticky behaviour on those pages:

- `CommentLayout`
- `LiveLayout` (currently not in use in DCR)
- `ShowcaseLayout`

Guardian Labs pages are also unaffected.

### Before (control group, i.e. current behaviour)
https://user-images.githubusercontent.com/57295823/133105391-f060acd9-3f88-4d7e-a367-be0dc97aa54d.mov

### After (variant group, i.e. 1% experiment)

https://user-images.githubusercontent.com/57295823/133105630-6f7ca632-d939-4318-8d02-4f1f21d99b10.mov


dotcom-rendering/src/web/components/ElementContainer.tsx
```diff
@@ -55,6 +59,7 @@ type Props = {
 	sectionId?: string;
 	showSideBorders?: boolean;
 	showTopBorder?: boolean;
+	showBottomBorder?: boolean;
```
oliverlloyd, line 62: Is there a way to ensure this prop doesn't remain here? Maybe we can create the PR to revert the test code now and keep it as a draft while the test runs?

The reason a bottom border is a problem is 1. because we want this component to follow (and perhaps eventually be replaced with) the `Container` from Source and also 2. By allowing a top + a bottom border you create the conditions for double borders - which are a chronic problem on the old site. By only allowing one border we help prevent this.

oliverlloyd (approved):
Ah, ternaries in jsx... I can see how making this work is tricky because the code you want to show hide depending on the AB bucket is at different levels, so you can't just do a simple conditional.

Might be worth looking to create the revert PR for this change now so that we have an easy way to back this out later on and then it'll be easier to move forward with whichever solution we decide upon.

dotcom-rendering/src/web/components/ElementContainer.tsx
MarSavar: I sort of agonised over adding this prop, I really didn't want to touch `ElementContainer` but it was the only way I could think of to achieve visual consistency with frontend - the top slot looked *very* weird without the bottom border when scrolling past the nav bar! 
I knew there must be a reason why that prop wasn't there in the first place, I just didn't know what it was, so thanks for explaining! 
Good shout about the revert PR, I'll work on that now. I have also carded this up in our comm dev Trello to make sure the prop is removed when the A/B test is over.

MarSavar:
Here is the PR to revert these changes (to be merged when the experiment is over): https://github.com/guardian/dotcom-rendering/pull/3424

PR #3423, opened 2021-09-14
Update the checks on the if statement for article avatars

nitro-marky:
## What does this change?

Changes the logic of whether an avatar should be shown


paperboyo:
Thanks, Mark! For context, this was to fix a situation where a contributor with an avatar pic was added to Additional contributors, but to the byline field. So in the cases where such a contributor doesn’t feature in the byline field (or when the byline field is empty), we don’t want the pic to show up.

PR #3426, opened 2021-09-14
add useTimeAgo hook to CR

joecowton1:
## What does this change?

This PR imports the `useTimeAgo` hook and tests created by @oliverlloyd to `common-rendering`. I've also added some additional Jest config (`jest-setup.ts` etc) to allow us to use the `jest-dom` testing library.

The original PR is here: https://github.com/guardian/dotcom-rendering/pull/2708/files

We're then implementing this hook in the `KeyEvents` component.

## Original implementation
This code was first added here https://github.com/guardian/dotcom-rendering/pull/2708 which I'm including here for reference and to help any future code archeologists

## Why?

As part of the `KeyEvents` component we need to display accurate 'time ago' data next to each `LiveBlog` post.

See discussion here: https://github.com/guardian/dotcom-rendering/pull/3366#discussion_r704336707

## Before

Screenshot 2021-09-15 at 11 33 54

## After

Screenshot 2021-09-15 at 11 26 54

https://user-images.githubusercontent.com/77005274/134028353-f8b736d0-87a5-4ced-99ae-451974bb5d17.mov


common-rendering/src/components/keyEvents.stories.tsx
```diff
@@ -1,58 +1,72 @@
...
+}
+
+const currentTime = addMinutes(now, 0);
+const thirtySeconds = addMinutes(now, 0.5);
+const fiveMinutes = addMinutes(now, 5);
+const fifteenMinutes = addMinutes(now, 15);
+const oneHour = addMinutes(now, 60);
+const oneDay = addMinutes(now, 60 * 24);
+const oneWeek = addMinutes(now, 60 * 24 * 7);
+const oneYear = addMinutes(now, 60 * 24 * 7 * 52);
```
OllysCoding, line 22: I like that we're including all the different variants here as part of the key events !

common-rendering/src/components/keyEvents.tsx
```diff
@@ -201,10 +203,19 @@ const timeStyles = (supportsDarkMode: boolean): SerializedStyles => css`
 `;
 
 const ListItem = ({ keyEvent, theme, supportsDarkMode }: ListItemProps) => {
+	useTimeAgo();
```
OllysCoding, line 206: Should we only be calling `useTimeAgo()` once per layout? if so is it better to lift it to `KeyEvents` for now, and add a TODO that it should be lifted to the overall layout at some point in the future?

joecowton1: This is an interesting point - do we need to consider at what level we're going to hydrate the component/s too? In AR previously we've tried to keep hydration as minimal as possible, esp as Emotion styling doesn't work client-side.

oliverlloyd: Olly's right, `useTimeAgo` sets up a `setInterval` on the whole page so needs to be called quite high up, probably in `App.tsx`

common-rendering/src/hooks/useTimeAgo.tsx
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Sets up an interval in which the timeAgo function is run against every time element on the
+ * page which has the data-relativeformat attribute set
+ * @param {Number} interval - The length of time in-between time being updated
+ * */
+export const useTimeAgo = (options?: Options) => {
+	useInterval(() => {
+		updateTimeElements();
+	}, options?.interval || DEFAULT_INTERVAL);
+};
```
oliverlloyd, line 50: I wonder if it's worth putting this code (`useTimeAgo.tsx` + `useInterval`) directly into `@guardian/libs`

@SiAdcock @sndrs  I know these are React focussed functions but maybe they're useful in other react codebases too?

oliverlloyd: Especially, the `updateTimeElements` function as that is continuing the contract that exists in Frontend around the `relativeformat` data attribute which I think is important.

common-rendering/src/components/keyEvents.stories.tsx
oliverlloyd: Yeah, this is really nice!

common-rendering/src/components/keyEvents.tsx
oliverlloyd: > do we need to consider at what level we're going to hydrate the component/s too?

Yeah, completely this. Hydration is a _big_ deal and really needs to be thought through. This is a little off topic (for this PR) but I was chatting with @lucymonie about this. I don't really have strong opinions on solutions but we do need to keep the risk of the TTI value increasing on the page if we hydrate the whole article body (which might be needed if we use React state to keep the blocks list live)

joecowton1: Maybe this is one to discuss in planning later? For now I think lifting it to `KeyEvents` and adding a `TODO` makes sense.

common-rendering/src/components/keyEvents.tsx
```diff
@@ -201,10 +203,19 @@ const timeStyles = (supportsDarkMode: boolean): SerializedStyles => css`
...
+	const time = timeAgo(new Date(keyEvent.time).getTime());
+
 	return (
 		<li css={listItemStyles(supportsDarkMode)}>
 			<div css={timeTextWrapperStyles}>
-				<time css={timeStyles(supportsDarkMode)}>{keyEvent.time}</time>
+				<time
+					dateTime={keyEvent.time}
```
marjisound, line 213: If keyEvent.time is an invalid date string for example `2021-15-09` the result would be `NaN NaN` 
Is that something we need to consider? Or we can just be sure that the consumer would always give up a correct date string?

marjisound (dismissed):

common-rendering/src/components/keyEvents.stories.tsx
marjisound: Nice from me too 🙌

common-rendering/src/components/keyEvents.tsx
oliverlloyd: Even with a TODO statement we're setting ourselves up for problems later - `useTimeAgo` is very much a page level execute once action. If it is run twice on the page we might see strange, unintended behaviour.

It's true though that one benefit to adding it here now is that the values are 'live' in storybook. This helps testing. But if that were a desired outcome then maybe we could call it from inside `preview.js`?

joecowton1: That's an excellent idea, given it's the `preview.js` for the entire repo is there any risk of it effecting any of the DCR components?

joecowton1: Ha, `hooks can only be called in the body of a function component`

common-rendering/src/components/keyEvents.tsx
joecowton1: I think we'd have to assume that the date would always be in the correct format, I'm happy to listen to suggestions though if anyone thinks otherwise.

common-rendering/src/components/keyEvents.tsx
```diff
@@ -201,10 +203,18 @@ const timeStyles = (supportsDarkMode: boolean): SerializedStyles => css`
...
 const ListItem = ({ keyEvent, theme, supportsDarkMode }: ListItemProps) => {
+	const time = timeAgo(new Date(keyEvent.time).getTime());
+
 	return (
 		<li css={listItemStyles(supportsDarkMode)}>
 			<div css={timeTextWrapperStyles}>
-				<time css={timeStyles(supportsDarkMode)}>{keyEvent.time}</time>
+				<time
+					dateTime={keyEvent.time}
+					data-relativeformat="short"
```
lucymonie, line 213: Is this part of the `<time>` interface? I've not seen it before.

joecowton1: It's just what the `updateTimeElements` function uses to determine the verbosity of the `timeAgo` string, I don't think it's specific to `time` though.

common-rendering/src/components/keyEvents.tsx
joecowton1: Added to the story.

oliverlloyd: I _think_ you can run the hook globally using a [global decorator](https://storybook.js.org/docs/react/writing-stories/decorators#global-decorators). That should set up the function for all stories, so any time anyone has a `time` element that respects the contract, they will get updated!

common-rendering/src/components/keyEvents.tsx
oliverlloyd: `data-relativeformat` is a convention first establish at the Guardian at least **_9_** years ago in [PR #38](https://github.com/guardian/frontend/pull/38), today we're on number 24179 - so as Guardian code conventions go, it's probably the oldest! (But I could be wrong - who knows what's out there...) It's also still in use today [on Frontend](https://github.com/guardian/frontend/search?q=data-relativeformat). 

See: https://github.com/guardian/frontend/blob/main/static/src/javascripts/projects/common/modules/ui/relativedates.js

The `timeAgo` lib was built to carry this convention on into DCR and to other repos in the department that need relative time. We actually decided to modify the api slightly, ending up on the `verbose` boolean you see. This was carefully worked out in [this discussion](https://github.com/guardian/libs/pull/134) leading the transformation (you see in this PR)

```ts
				switch (relativeFormat) {
					case "short":
					case "med":
						newTime = timeAgo(new Date(absoluteTime).getTime(), {
							verbose: false,
						});
						break;
					case "long":
						newTime = timeAgo(new Date(absoluteTime).getTime(), {
							verbose: true,
						});
						break;
				}
```

common-rendering/src/components/keyEvents.tsx
joecowton1: Updated to use a global decorator 👍

oliverlloyd (approved):
Looks great!

One nice to have: if the PR description has a gif or movie showing how time moves that would be 😻

joecowton1:
> Looks great!
> 
> One nice to have: if the PR description has a gif or movie showing how time moves that would be 😻

Thanks Oliver! Have added a clip to the description.

oliverlloyd:
We added this feature via [this PR](https://github.com/guardian/dotcom-rendering/pull/3704) in the end

PR #3428, opened 2021-09-15
Meta Data for Liveblogs

nitro-marky:
## What does this change?

This is an initial part of the WIP for the MetaData component for liveblogs. There are still areas that need to be decided on which has blocked how we put the logic together for multiple names. So the the current version only works with single names (you'll see a lot of commented out code which has been taken from DCR components but we may end up using those components anyway so 🤷‍♂️ ).

There is also things like dark mode to add. The decision was made to put the count and social icons in a separate component but they could be added here. 

Todo:
- Multiple names
- Dark Mode

### Breakpoints

Desktop:
Screenshot 2021-09-15 at 12 15 28

Tablet:
Screenshot 2021-09-15 at 12 15 54

Phablet:
Screenshot 2021-09-15 at 12 17 11

Mobile:
Screenshot 2021-09-15 at 12 17 29

Mobile with secondary date:
Screenshot 2021-09-15 at 12 17 41

### Themes
News:
Screenshot 2021-09-15 at 12 17 29

Opinion
Screenshot 2021-09-15 at 12 26 10

Sport:
Screenshot 2021-09-15 at 12 26 18

Culture:
Screenshot 2021-09-15 at 12 26 25


common-rendering/src/components/MetaData.tsx
```diff
@@ -0,0 +1,363 @@
...
+	<a
+		rel="author"
+		data-link-name="auto tag link"
+		href={`//www.theguardian.com/${contributorTagId}`}
+	>
+		{contributor}
+	</a>
+);
+
+// Shamelessly borrowed from Lucy! (This should be lifted out as it's very useful!)
```
joecowton1, line 209: Can we add a `styles.ts` or similar to `common-rendering` containing this and other shared styling functions so we're not repeating at component level?

common-rendering/src/components/MetaData.tsx
```diff
@@ -0,0 +1,363 @@
...
+} from "@guardian/src-foundations/palette";
+import { space } from "@guardian/src-foundations";
+
+export interface Authors {
+	id: string;
+	name: string;
+	contributor: boolean;
+}
+
+type BylineComponent = { token: string; tag: TagType } | string;
```
joecowton1, line 24: Could we extract `{ token: string; tag: TagType }` to a new interface and name it in such a way as to make it slightly clearer what's happening here?

common-rendering/src/components/MetaData.stories.tsx
```diff
@@ -0,0 +1,64 @@
+import { palette } from "@guardian/src-foundations";
+import type { FC } from "react";
+import { MetaData } from "./MetaData";
+import { Pillar, Theme, Special } from "@guardian/types";
+import { select } from "@storybook/addon-knobs";
+
+const pillarOptions = {
```
joecowton1, line 7: Feels like this and the `selectPillar` function are being repeated a lot now, could we maybe put them somewhere shared?

common-rendering/src/components/MetaData.tsx
```diff
@@ -0,0 +1,363 @@
...
+interface TagType {
+	id: string;
+	type: string;
+	title: string;
+	twitterHandle?: string;
+	paidContentType?: string;
+	bylineImageUrl?: string;
+}
+
+type Props = {
```
joecowton1, line 35: Any reason this is a type when we're using `interface` in the other instances?

common-rendering/src/components/MetaData.tsx
```diff
@@ -0,0 +1,363 @@
...
+	if (byline.includes("now")) {
+		byline.replace("now", "(now)");
+	}
+	if (byline.includes("earlier")) {
+		byline.replace("earlier", "(earlier)");
+	}
+}*/
+
+const ContributorLink: React.FC<{
+	contributor: string;
```
joecowton1, line 197: Could we extract these types (`contributor` & `contributorTag`) to a `ContributorLinkProps` definition or similar?

common-rendering/src/components/MetaData.tsx
```diff
@@ -0,0 +1,363 @@
...
+	:hover {
+		cursor: pointer;
+	}
+
+	${from.desktop} {
+		color: ${palette.neutral[20]};
+	}
+`;
+
+// I would normally use the Lines component
```
joecowton1, line 276: This comment doesn't really explain _why_  you've chosen to not use the `lines` component - an explanation of that would make a more useful comment.

common-rendering/src/components/MetaData.tsx
```diff
@@ -0,0 +1,363 @@
...
+export const MetaData = ({
+	theme,
+	palette,
+	byline,
+	tags,
+	authors,
+	primaryDateline,
+	secondaryDateline,
+}: Props) => {
+	// Remember how to get array of contributors
```
joecowton1, line 332: Again, maybe think about how this comment could explain better what's happening here - is this a `TODO` and if so why are we not implementing the line below currently?

joecowton1 (changes requested):

common-rendering/src/components/MetaData.tsx
nitro-marky: We can do, this is mainly a draft until we decide what we are doing with DCR components going forward. There is a meeting later today (15/09/21)

common-rendering/src/components/MetaData.stories.tsx
nitro-marky: Yeah, it can do. Again, until there's a clearer picture about this component I'll know more.

common-rendering/src/components/MetaData.tsx
nitro-marky: This line is not being used at the moment, commented out.

common-rendering/src/components/MetaData.tsx
nitro-marky: Out of curiosity what would be the benefit of that? I copied this from DCR as is.

nitro-marky: Out of curiosity what would be the benefit for such a small component? It was re-used from DCR as is.

joecowton1: Just convention really - it's how we've always done it on AR. I think beyond that it just makes it slightly easier to read: the component is split over less lines but tbh it's pretty minor.

nitro-marky: Done

common-rendering/src/components/MetaData.tsx
nitro-marky: I guess it also depends on what the person needs. So I've modified Lucy's as I don't need the same data that she is using.

common-rendering/src/components/MetaData.stories.tsx
nitro-marky: https://github.com/guardian/dotcom-rendering/pull/3430

PR #3429, opened 2021-09-15
Move webfont class to separate style tag

chrislomaxjones:
## What does this change?

Currently the CSS that is used for font styling, reset and all other page styles are concatenated and provided in a single `<style class="webfont'>...</style>` tag in the head of the document. This PR proposes splitting these styles out into two separate style tags: one for styling of fonts with the class `webfont` and another tag for the Reset CSS and other styles.

## Why?

Native ads are displayed on the page in iframes and [retrieve font styling via a call to `get-styles`](https://github.com/guardian/commercial-templates/blob/378955e971f11f6dfce21f37ba3d027a28d03625/src/_shared/js/messages.js#L68) on the surrounding page, with the selector `.webfont`. This selector currently retrieves **all** of the styles from the page - including the reset. This is applied after styling in the ad which causes an issue where font's aren't properly styled.

### Before

Screenshot 2021-09-15 at 12 46 50

### After

(Note the bold elements on the native ad)

Screenshot 2021-09-15 at 12 43 51

### DIFF
--- a/dotcom-rendering/src/web/server/htmlTemplate.ts
+++ b/dotcom-rendering/src/web/server/htmlTemplate.ts
@@ -284,7 +284,8 @@ export const htmlTemplate = ({
                 </noscript>
                 ${loadableConfigScripts.join('\n')}
                 ${priorityScriptTags.join('\n')}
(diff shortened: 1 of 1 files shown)


DavidLawes (approved):
👍

PR #3430, opened 2021-09-15
Select Pillar is now its own importable function

nitro-marky:
## What does this change?

Extracts `selectPillar` from stories to an importable function.

Include this `import { SelectPillar } from "../Fixtures/SelectStoryPillar";` and `SelectPillar(Pillar.News)` can be used as normal.

## Why?

Suggestion from @joecowton1

### DIFF
--- a/common-rendering/src/fixtures/SelectPillar.ts
+++ b/common-rendering/src/fixtures/SelectPillar.ts
@@ -1,0 +1,16 @@
+import { Pillar, Theme, Special } from "@guardian/types";
+import { select } from "@storybook/addon-knobs";
+
+const pillarOptions = {
+	News: Pillar.News,
+	Opinion: Pillar.Opinion,
+	Sport: Pillar.Sport,
+	Culture: Pillar.Culture,
+	Lifestyle: Pillar.Lifestyle,
+	Labs: Special.Labs,
+	SpecialReport: Special.SpecialReport,
+};
+
+export const SelectPillar = (initial: Pillar): Theme => {
+	return select("Pillar", pillarOptions, initial);
+};


common-rendering/src/fixtures/SelectPillar.ts
```diff
@@ -0,0 +1,16 @@
+import { Pillar, Theme, Special } from "@guardian/types";
+import { select } from "@storybook/addon-knobs";
+
+const pillarOptions = {
```
oliverlloyd, line 4:
```suggestion
const themeOptions = {
```

common-rendering/src/fixtures/SelectPillar.ts
```diff
@@ -0,0 +1,16 @@
...
+	News: Pillar.News,
+	Opinion: Pillar.Opinion,
+	Sport: Pillar.Sport,
+	Culture: Pillar.Culture,
+	Lifestyle: Pillar.Lifestyle,
+	Labs: Special.Labs,
+	SpecialReport: Special.SpecialReport,
+};
+
+export const SelectPillar = (initial: Pillar): Theme => {
```
oliverlloyd, line 14:
```suggestion
export const SelectTheme = (initial: Theme): Theme => {
```

common-rendering/src/fixtures/SelectPillar.ts
```diff
@@ -0,0 +1,16 @@
...
+	return select("Pillar", pillarOptions, initial);
```
oliverlloyd, line 15:
```suggestion
	return select("Theme", themeOptions, initial);
```

oliverlloyd (commented):
I dropped some pillar > theme suggestions. `Pillar` is just a subset of `Theme`. Theme is the real deal

Also, I think filenames for non react functions should be camelCase, instead of PascalCase. I believe the convention (or maybe even requirement) is to start react component names  using  a capital letter and so it's nice to keep that exclusive so you know when a thing is a component or not

PR #3433, opened 2021-09-17
DCR as a page rendering service, not an _article_ rendering service

oliverlloyd:
## What does this change?
This refactors how DCR builds the html for a page to remove the dependency on the `CAPI` object. Now both `document` and `htmlTemplate` will work for rendering simpler, non article pages.

### Before
```ts
// document.tsx
interface Props {
	data: DCRServerDocumentData;
}
```

```ts
// htmlTemplate.ts
type Props = {
	title?: string;
	description: string;
	linkedData: { [key: string]: any };
	loadableConfigScripts: string[];
	priorityScriptTags: string[];
	lowPriorityScriptTags: string[];
	css: string;
	html: string;
	fontFiles?: string[];
	windowGuardian: string;
	gaPath: { modern: string; legacy: string };
	ampLink?: string;
	openGraphData: { [key: string]: string };
	twitterData: { [key: string]: string };
	keywords: string;
	skipToMainContent: string;
	skipToNavigation: string;
}
```

### After
```ts
// document.tsx
interface Props {
	extractedCss: string;
	html: string;
	description: string;
	title?: string;
	CAPI?: CAPIType;
	linkedData?: { [key: string]: any };
	priorityScriptTags?: string[];
	lowPriorityScriptTags?: string[];
	gaPath?: { modern: string; legacy: string };
	loadableConfigScripts?: string[];
	windowGuardian?: string;
}
```

```ts
// htmlTemplate.ts
type Props = {
	title?: string;
	description: string;
	linkedData?: { [key: string]: any };
	loadableConfigScripts: string[];
	priorityScriptTags: string[];
	lowPriorityScriptTags: string[];
	css: string;
	html: string;
	windowGuardian?: string;
	gaPath?: { modern: string; legacy: string };
	ampLink?: string;
	openGraphData?: { [key: string]: string };
	twitterData?: { [key: string]: string };
	keywords: string;
	skipToMainContent: string;
	skipToNavigation: string;
}
```

## Why?
Because we want DCR to serve other pages, not just articles

## Do we?
Well, I think so. I want to add pages such as https://www.theguardian.com/email-newsletters but if I want to do that at the moment I need to build a whole react app from scratch, which isn't great 😐 

## Why is this a draft PR?
Personally, I think this PR is a good idea. We should be aligning this kind of high level page structure (and all the build and test logic as well) for all pages on dotcom. But rather than try to push it through I wanted to canvas opinion on this and maybe this PR is a good place to do that?

**I am not going to merge this PR** If there's agreement that this approach is something that we all want then I will go back and make these changes bit by bit in separate PRs so that we can check them more thoroughly.

### DIFF
--- a/dotcom-rendering/src/web/server/render.ts
+++ b/dotcom-rendering/src/web/server/render.ts
@@ -1,7 +1,7 @@
 import express from 'express';
 import { extractNAV } from '@root/src/model/extract-nav';
 
-import { document } from '@root/src/web/server/document';
+import { generateArticleHtml } from '@root/src/web/server/generateArticleHtml';
 import { enhanceCAPI } from '@root/src/model/enhanceCAPI';
 import { extract as extractGA } from '@root/src/model/extract-ga';
 import { Article as ExampleArticle } from '@root/fixtures/generated/articles/Article';
@@ -12,7 +12,7 @@ export const renderArticle = (
 ): void => {
 	try {
 		const CAPI = enhanceCAPI(body);
-		const resp = document({
+		const html = generateArticleHtml({
 			data: {
 				CAPI,
 				site: 'frontend',
@@ -23,8 +23,8 @@ export const renderArticle = (
 			},
 		});
 
-		res.status(200).send(resp);
-	} catch (e) {
+		res.status(200).send(html);
+	} catch (e: any) {
 		// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
 		res.status(500).send(`<pre>${e.stack}</pre>`);
 	}
@@ -48,7 +48,7 @@ export const renderArticleJson = (
 		};
 
 		res.status(200).send(resp);
-	} catch (e) {
+	} catch (e: any) {
 		// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
 		res.status(500).send(`<pre>${e.stack}</pre>`);
 	}
@@ -68,7 +68,7 @@ export const renderInteractive = (
 ): void => {
 	try {
 		const CAPI = enhanceCAPI(body);
-		const resp = document({
+		const html = generateArticleHtml({
 			data: {
 				CAPI,
 				site: 'frontend',
@@ -79,8 +79,8 @@ export const renderInteractive = (
 			},
 		});
 
-		res.status(200).send(resp);
-	} catch (e) {
+		res.status(200).send(html);
+	} catch (e: any) {
 		// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
 		res.status(500).send(`<pre>${e.stack}</pre>`);
 	}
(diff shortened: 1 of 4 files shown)


oliverlloyd:
[[Spike] Investigate moving /email-newsletters to DCR](https://trello.com/c/nqTeVh4E/315-spike-investigate-moving-email-newsletters-to-dcr)

dotcom-rendering/src/web/server/render.ts
```diff
@@ -48,7 +48,7 @@ export const renderArticleJson = (
 		};
 
 		res.status(200).send(resp);
-	} catch (e) {
+	} catch (e: any) {
```
davidfurey, line 51: Sorry, completely unrelated to the substance of your change, but I'm curious as to why you've added the explicit `any` here?

oliverlloyd: I just added that to get rid of the red squiggly line I had in my code editor! Not proposing we move forward with this (but some form of typing needs to be added)

davidfurey: :-) Fair enough.

`unknown` seems most accurate since things can throw anything.  Although in this example I think we'd only have a problem if something throws `undefined` or `null`.

nicl:
I like the idea, but think it shows that we need to refactor things here. These interfaces:

```
// document.tsx
interface Props {
	extractedCss: string;
	html: string;
	description: string;
	title?: string;
	CAPI?: CAPIType;
	linkedData?: { [key: string]: any };
	priorityScriptTags?: string[];
	lowPriorityScriptTags?: string[];
	gaPath?: { modern: string; legacy: string };
	loadableConfigScripts?: string[];
	windowGuardian?: string;
}

// htmlTemplate.ts
type Props = {
	title?: string;
	description: string;
	linkedData?: { [key: string]: any };
	loadableConfigScripts: string[];
	priorityScriptTags: string[];
	lowPriorityScriptTags: string[];
	css: string;
	html: string;
	windowGuardian?: string;
	gaPath?: { modern: string; legacy: string };
	ampLink?: string;
	openGraphData?: { [key: string]: string };
	twitterData?: { [key: string]: string };
	keywords: string;
	skipToMainContent: string;
	skipToNavigation: string;
}
```

are way too broad. (I realise a lot of this predates this PR.) Part of this/a clue is that we have a lot of optional parameters, including the entire CAPIType itself for `document.tsx`. To me this indicates we should be creating multiple entry points/components and sharing code through composition here rather than trying to overload document/htmlTemplate for everything, which will just make things brittle and unmaintainable in the longer term.

oliverlloyd:
> creating multiple entry points/components and sharing code through composition

So rather than having one `htmlTemplate` you're thinking we have one per page and that each one pulls in blocks of logic from a shared pool?

That's an interesting idea. In my head (and I didn't really think this through much) I envisioned making the split at the route level, so instead of only having `renderArticle` we'd have `renderArticle` and `renderPage`. But I think you're right, these props have grown over time and this structure could be better.

I guess we want something flexible that will scale over time in a maintainable way 🤔 

One potential problem to importing blocks is it's harder to specify required logic, like a 'title' or certain script tags. How could we enforce these things?

PR #3435, opened 2021-09-20
LIVE-3071: add min height to headlines

joecowton1:
## What does this change?
Adds `min-height` to `interview` and `immersive` article template `headline` components.

## Why?
Previously we were seeing gaps when we had a one line headline.

### Before
Screenshot 2021-09-20 at 12 08 22

### After

Screenshot 2021-09-20 at 12 08 01


groakland (approved):
LGTM

PR #3438, opened 2021-09-20
LIVE-3075: render links as text in overlay

joecowton1:
## What does this change?
Handles `a` tags in header image overlay.

## Why?
Previously we weren't rendering these so any links were rendered as an empty string.

### Before
Screenshot 2021-09-20 at 12 39 20

### Afte
Screenshot 2021-09-20 at 14 38 35
r

### DIFF
--- a/apps-rendering/src/components/editions/headerImageCaption.tsx
+++ b/apps-rendering/src/components/editions/headerImageCaption.tsx
@@ -86,7 +86,7 @@ const toReact = (node: Node, key: number): ReactNode => {
 		case 'EM':
 			return <em>{node.textContent}</em>;
 		case 'A':
-			return <span>{node.textContent ?? ''}</span>;
+			return <span>{node.textContent}</span>;
 		case '#text':
 			return node.textContent;
 	}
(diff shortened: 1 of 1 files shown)


apps-rendering/src/components/editions/headerImageCaption.tsx
```diff
@@ -85,6 +85,8 @@ const toReact = (node: Node, key: number): ReactNode => {
 	switch (node.nodeName) {
 		case 'EM':
 			return <em>{node.textContent}</em>;
+		case 'A':
+			return <span>{node.textContent ?? ''}</span>;
```
abeddow91, line 89: What do we get from the empty string here? If its needed here, should we add it to the EM switch as well?

joecowton1: Not much tbh, was just copying from another implementation but it works fine without.

abeddow91 (approved):
Looks good!

groakland (approved):
LGTM

PR #3439, opened 2021-09-20
Style inline merchandising ad slot

chrislomaxjones:
## What does this change?

- Move dynamic ad styles from the `ElementContainer` component level to the `ArticleRenderer` component level. 
- Add styling to DCR for the `ad-slot--im` inline merchandising class so that the styles match those of frontend. These are a transformation of the [frontend styles](https://github.com/guardian/frontend/blob/07305ca708926ef75db9b4489644c1174585cd84/static/src/stylesheets/module/_adslot.scss#L379-L400) such that node-sass syntax is removed and measurements are set in pixels to match the rest of the styling in `ArticleContainer`.

## Why?

- The styling of dynamic ads is moved up to the `ArticleRenderer` level so that these styles are not replicated across each individual `ElementContainer`. `ArticleRenderer` is the component to which a class is added that scopes where dynamic ads can be placed in the article by spacefinder.
- So the styling between frontend and DCR for inline merchandising ads matches and specifically that the DCR variant doesn't take up the full width of the article container.

### Before

Screenshot 2021-09-21 at 15 05 55

### After

Screenshot 2021-09-21 at 15 05 35

### DIFF
--- a/dotcom-rendering/src/web/components/ElementContainer.tsx
+++ b/dotcom-rendering/src/web/components/ElementContainer.tsx
@@ -1,12 +1,11 @@
 import { ClassNames, css as emoCss } from '@emotion/react';
 
 import { border } from '@guardian/src-foundations/palette';
-import { from, until } from '@guardian/src-foundations/mq';
+import { from } from '@guardian/src-foundations/mq';
 import { center } from '@root/src/web/lib/center';
 // @ts-ignore-start
 import { jsx as _jsx } from 'react/jsx-runtime';
 // @ts-ignore-end
-import { labelStyles as adLabelStyles } from './AdSlot';
 
 const padding = emoCss`
 	padding: 0 10px;
@@ -14,26 +13,6 @@ const padding = emoCss`
 	${from.mobileLandscape} {
 		padding: 0 20px;
 	}
-`;
-
-const adStylesDynamic = emoCss`
-	& .ad-slot.ad-slot--collapse {
-		display: none;
-	}
-
-	${from.tablet} {
-		.mobile-only .ad-slot {
-			display: none;
-		}
-	}
-
-	${until.tablet} {
-		.hide-until-tablet .ad-slot {
-			display: none;
-		}
-
-	}
-
 `;
 
 const sideBorders = (colour: string) => emoCss`
@@ -96,11 +75,7 @@ export const ElementContainer = ({
 					{children && children}
 				</div>
 			);
-			// Apply ad styles to dynamic ad slots (i.e. slots that are not fixed), e.g. ads inserted
-			// by spacefinder and ads in interactive articles
 			const style = css`
-				${adStylesDynamic}
-				${adLabelStyles}
 				${backgroundColour && setBackgroundColour(backgroundColour)}
 			`;
 			// Create a react element from the tagName passed in OR
--- a/dotcom-rendering/src/web/lib/ArticleRenderer.tsx
+++ b/dotcom-rendering/src/web/lib/ArticleRenderer.tsx
@@ -3,11 +3,38 @@ import { css } from '@emotion/react';
 import { renderArticleElement } from '@root/src/web/lib/renderElement';
 import { withSignInGateSlot } from '@root/src/web/lib/withSignInGateSlot';
 import { ArticleDesign, ArticleFormat } from '@guardian/libs';
+import { from, until } from '@guardian/src-foundations/mq';
 import { interactiveLegacyClasses } from '../layouts/lib/interactiveLegacyStyling';
+import { labelStyles as adLabelStyles } from '../components/AdSlot';
 
 // This is required for spacefinder to work!
 const commercialPosition = css`
 	position: relative;
+`;
+
+// These styles are applied to dynamic ad slots (i.e. slots that are not fixed), e.g. ads inserted
+// by spacefinder across all layout types (articles, comments, etc)
+//
+// spacefinder is scoped to placing elements in spaces within the .article-body-commercial-selector
+// hence we scope the styles at the same level
+const adStylesDynamic = css`
+	${adLabelStyles}
+
+	& .ad-slot.ad-slot--collapse {
+		display: none;
+	}
+
+	${from.tablet} {
+		.mobile-only .ad-slot {
+			display: none;
+		}
+	}
+
+	${until.tablet} {
+		.hide-until-tablet .ad-slot {
+			display: none;
+		}
+	}
 `;
 
 export const ArticleRenderer: React.FC<{
@@ -50,7 +77,7 @@ export const ArticleRenderer: React.FC<{
 					? interactiveLegacyClasses.contentMainColumn
 					: '',
 			].join(' ')}
-			css={commercialPosition}
+			css={[adStylesDynamic, commercialPosition]}
 		>
 			{/* Insert the placeholder for the sign in gate on the 2nd article element */}
 			{withSignInGateSlot(output)}
--- a/dotcom-rendering/src/web/lib/ArticleRenderer.tsx
+++ b/dotcom-rendering/src/web/lib/ArticleRenderer.tsx
@@ -35,6 +35,27 @@ const adStylesDynamic = css`
 			display: none;
 		}
 	}
+
+	.ad-slot--im {
+		float: left;
+		width: 130px;
+		${from.mobileLandscape} {
+			width: 220px;
+		}
+
+		&:not(.ad-slot--rendered) {
+			width: 0;
+			height: 0;
+		}
+
+		&.ad-slot--rendered {
+			margin: 5px 10px 6px 0;
+			${from.mobileLandscape} {
+				margin-bottom: 12px;
+				margin-right: 20px;
+			}
+		}
+	}
 `;
 
 export const ArticleRenderer: React.FC<{


dotcom-rendering/src/web/components/ArticleContainer.tsx
```diff
@@ -79,6 +79,26 @@ const articleAdStyles = css`
 	.ad-slot--fluid {
 		width: 100%;
 	}
+	.ad-slot--im {
```
chrislomaxjones, line 82: Should we keep a record of where these styles came from in frontend? None of the other ad slot styles defined here do

mxdvl: I don’t think it’s essential. What matters most is that it works in DCR, doesn’t it?

For posterity, [here’s the search for `.ad-slot--im`](https://github.com/guardian/frontend/search?q=.ad-slot--im).

chrislomaxjones: I agree, it would be too much to mark every style that was used on both.

arelra:
Something to be aware of...

We have a 'problem' with styling dynamic ads in DCR which are inserted at runtime by JS so don't fit into the component model but still require styling.

Styles added to `ArticleContainer` will only apply to the main article body.

So for other non-article types of pages like immersives and interactives these styles won't apply.

For that we had to resort to `ElementContainer` which was used in lieu of a higher level component that applies to all pages.

There is now a new `Page` component where we could add dynamic ad styles but we would need approval from the DCR folks.

chrislomaxjones:
> Something to be aware of...
> 
> We have a 'problem' with styling dynamic ads in DCR which are inserted at runtime by JS so don't fit into the component model but still require styling.
> 
> Styles added to `ArticleContainer` will only apply to the main article body.
> 
> So for other non-article types of pages like immersives and interactives these styles won't apply.
> 
> For that we had to resort to `ElementContainer` which was used in lieu of a higher level component that applies to all pages.
> 
> There is now a new `Page` component where we could add dynamic ad styles but we should talk to the DCR folks.

I see, perhaps for now I should change these styles to be scoped to `ElementContainer` and we can look at adding them to `Page` in the future.

dotcom-rendering/src/web/components/ArticleContainer.tsx
```diff
@@ -76,7 +76,7 @@ const articleAdStyles = css`
 			}
 		}
 	}
-	.ad-slot--fluid {
+	.ad-slot--fluid:not(.ad-slot--im) {
```
chrislomaxjones, line 79: The `:not(.ad-slot--im)` is included to ensure the inline merch slot doesn't take the full width of the container.

The same functionality is implemented in frontend along with [more rules for other ads](https://github.com/guardian/frontend/blob/07305ca708926ef75db9b4489644c1174585cd84/static/src/stylesheets/module/_adslot.scss#L443-L445). I've neglected to copy it verbatim as there seems little value in doing that without considering whether we really want those rules here.

dotcom-rendering/src/web/components/ElementContainer.tsx
```diff
@@ -34,6 +34,26 @@ const adStylesDynamic = emoCss`
 
 	}
 
+	.ad-slot--im {
```
chrislomaxjones, line 37: These styles now live with the other dynamic ad styles in `ElementContainer` (at least for now...)

chrislomaxjones:
> There is now a new `Page` component where we could add dynamic ad styles but we should talk to the DCR folks.

PR updated to move the dynamic styles to the `Page` level.

oliverlloyd (commented):
This css will affect everything DCR renders but part of the work I'm doing is to try and make DCR more flexible, something that can be used for more than just articles (see [this PR](https://github.com/guardian/dotcom-rendering/pull/3433)). This change would cause problems with that approach.

Can we go the other way and move the css down? Ideally we want these kinds of selectors to be as targeted as possible. Global css should be - with _very few_ exceptions - avoided in DCR

chrislomaxjones:
@oliverlloyd That's a really good point actually. I think the idea was that dynamic ads _could_ be placed on a page outside of articles and therefore the styling would be applied at the Page level. But I do agree that this makes things less flexible and global CSS should be avoided as much as possible.

What would be a better alternative? Previously dynamic ads styles were applied at the `ElementContainer` level - is it safe to assume we'll always place dynamic ads within these components and is it clear that is the case?

oliverlloyd:
> What would be a better alternative?

Fair question. I guess there's two things here

1. Do these styles need to be global?
Ideally, there would be a way to only apply them where and when required. I notice that the reason they were moved out of `ArticleContainer` was because of interactives? Eg:

> So for other non-article types of pages like immersives and interactives these styles won't apply

So maybe another approach would be to keep them in `ArticleContainer`, which feels like a right place for them to live as they are then scoped to the body of the article, and to then have an additional wrapper that is specific to immersives/interactives. You could apply that in their respective layout files, eg: [ImmersiveLayout](https://github.com/guardian/dotcom-rendering/blob/9abeb3cd3d1341b42d3d89156247eddaa3e1f31a/src/web/layouts/ImmersiveLayout.tsx)

Like that it's clear what the intended scope of that css is and you can also pick and, if possible, you can choose the specific styles needed in each use case

2. Making `Page` less article centric
If you _did_ go down the global css route (and I would entreat you to try and avoid this at all costs - the [reason DCR exists](https://www.theguardian.com/info/2019/apr/04/revisiting-the-rendering-tier) was to fight the problem of global css) but if you did, you could duplicate `Page` and call it `Article` and then compose these styles, like:

```
export const Article = ({ children, cache }: Props) => {
	return (
		<CacheProvider value={cache}>
			<StrictMode>
				<FocusHaloStyles>
					<DynamicAdStyles>
						{children}
					</DynamicAdStyles>
				</FocusHaloStyles>
			</StrictMode>
		</CacheProvider>
	);
};
```

At least then you avoid having this blob of global css that acts like an invitation to people to add to

dotcom-rendering/src/web/components/Page.tsx
```diff
@@ -2,12 +2,55 @@ import React, { StrictMode } from 'react';
...
 import { focusHalo } from '@guardian/src-foundations/accessibility';
+import { from, until } from '@guardian/src-foundations/mq';
+import { labelStyles as adLabelStyles } from './AdSlot';
 
 type Props = {
 	children: React.ReactNode;
 	cache: EmotionCache;
 };
 
+const adStylesDynamic = css`
```
mxdvl, line 13: @oliverlloyd Would it help a little if we clearly identified this as an anti-pattern?

```suggestion
// This way of injecting CSS is against the DCR principle and should
// be avoided. It is a temporary solution. Do not expand this list.
// TODO: Real AdSlot components
const adStylesDynamic = css`
```

mxdvl (commented):
Maybe there’s a way we could inject those styles dynamically?

For example, is `.mobile-only` ever used in DCR?

dotcom-rendering/src/web/components/Page.tsx
oliverlloyd: In my experience, when you add TODO comments saying something will be fixed in future, it is almost always never is. There are _six pages_ of [TODO comments in DCR](https://github.com/guardian/dotcom-rendering/search?q=TODO), most of them will continue to live there for years 😞

oliverlloyd:
> Maybe there’s a way we could inject those styles dynamically?

This! If we can remove (or at least reduce) the need to have a painful global contract between DCR and the Commercial code that would be much better. Less coupling 🎉

chrislomaxjones:
Thanks for the feedback @oliverlloyd @mxdvl.

I had a think about this and it guess my main issue was the questions of _where_ dynamic ads can be placed in DCR (i.e. what layouts). Having done some further digging it seems like:

- Ads are inserted by spacefinder into articles based on a set of rules. One of the rules specifies the top level selector of the element into which dynamic ads can be placed, called the `bodySelector`. On DCR this is the `.article-body-commercial-selector` selector.
- On DCR this [class is associated with the `ArticleRenderer` component](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/docs/contracts/001-commercial-selectors.md). This component is rendered within the layouts we intend to support styling of dynamic ads: **Standard Articles**, **Comments** and **Showcases**. 
- `ArticleRenderer` is not used in the process of rendering **Liveblogs** and **Interactives**. However, spacefinder uses a different selector for liveblogs `.js-liveblog-body` which is not currently supported by DCR liveblogs (and production liveblogs are still served by frontend) and interactives do not support the dynamic insertions of ads via spacefinder at all.

Therefore it seems reasonable to me to scope the styling of dynamic ads at the level to which the `.article-body-commercial-selector` is attached. At this component level we are therefore opting into allowing dynamic ad placement via spacefinder by including the class and also providing the styling necessary for these ads at the same point. 

I've updated the PR with this change, would appreciate some feedback on whether this seems like a reasonable approach.

> Maybe there’s a way we could inject those styles dynamically?

This does also seems like an interesting idea but I'm not sure how I'd approach this - would it mean injecting the styles from within the commercial bundle at the point at which spacefinder inserts dynamic ads?

oliverlloyd (approved):
Wow, impressive work @chrislomaxjones 👏 

> * `ArticleRenderer` is not used in the process of rendering **Liveblogs** and **Interactives**. However, spacefinder uses a different selector for liveblogs `.js-liveblog-body` which is not currently supported by DCR liveblogs (and production liveblogs are still served by frontend) and interactives do not support the dynamic insertions of ads via spacefinder at all.

That's really interesting and potentially useful for the LiveBlog team (@lucymonie ☝️ ). Maybe this selector needs to get added to [LiveBlogRenderer](https://github.com/guardian/dotcom-rendering/blob/a42df26f3e33fd7340bdea059a3f67bf91b678db/dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx)? Or at least this looks like a superb place to start when investigating how to ad adds to LiveBlogs on DCR.

> would it mean injecting the styles from within the commercial bundle at the point at which spacefinder inserts dynamic ads?

I _guess_? But I'm not sure how much of this styling is relevant to the page structure and how much of it can simply be scoped to the advert itself. If it's just affecting the ad, then really, it should be local. That would very much be a larger task though and certainly not part of this PR.

mxdvl (approved):
Great digging!

PR #3444, opened 2021-09-21
Use ad sizes defined in `commercial-core`

simonbyford:
Co-Authored-By: Chris Jones <chrislomaxjones96@gmail.com>

## What does this change?

This change pulls `adSizes` from `@guardian/commercial-core` which is the new owner of this information.

## Why?

Part of a change to centralise where we store information about ad sizes. Previously, they were stored separately in `frontend` and `dotcom-rendering` which risked the ad size definitions diverging over time if not managed carefully.

Related PRs:

https://github.com/guardian/commercial-core/pull/353
https://github.com/guardian/frontend/pull/24184

### DIFF
--- a/dotcom-rendering/src/web/components/AdSlot.tsx
+++ b/dotcom-rendering/src/web/components/AdSlot.tsx
@@ -5,45 +5,12 @@ import { textSans } from '@guardian/src-foundations/typography';
 import { from, until } from '@guardian/src-foundations/mq';
 import { Display } from '@guardian/types';
 import { space } from '@guardian/src-foundations';
+import { adSizes } from '@guardian/commercial-core';
 
 type Props = {
 	display: Display;
 	position: AdSlotType;
 };
-
-/*
-    This file's values are meant to mirror the values used by frontend.
-    Look for marks:
-        432b3a46-90c1-4573-90d3-2400b51af8d0
-        1b109a4a-791c-4214-acd2-2720d7d9f96f
-    ... in the frontend code
- */
-
-enum Size {
-	// standard ad sizes
-	billboard = '970,250',
-	leaderboard = '728,90',
-	mpu = '300,250',
-	halfPage = '300,600',
-	portrait = '300,1050',
-	skyscraper = '160,600',
-	mobilesticky = '320,50',
-	// dfp proprietary ad sizes
-	fluid = 'fluid',
-	outOfPage = '1,1',
-	googleCard = '300,274',
-	// guardian proprietary ad sizes
-	video = '620,1',
-	outstreamDesktop = '620,350',
-	outstreamGoogleDesktop = '550,310',
-	outstreamMobile = '300,197',
-	merchandisingHighAdFeature = '88,89',
-	merchandisingHigh = '88,87',
-	merchandising = '88,88',
-	inlineMerchandising = '88,85',
-	fabric = '88,71',
-	empty = '2,2',
-}
 
 export const labelHeight = 24;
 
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/AdSlot.tsx
```diff
@@ -388,21 +355,21 @@ export const AdSlot: React.FC<Props> = ({ position, display }) => {
...
 						].join('|')}
 						data-desktop={[
-							`${Size.outOfPage}`,
-							`${Size.empty}`,
-							`${Size.leaderboard}`,
+							`${adSizes.outOfPage.toString()}`,
+							`${adSizes.empty.toString()}`,
+							`${adSizes.leaderboard.toString()}`,
 							`940,230`,
 							`900,250`,
```
simonbyford, line 369: Possibly outside the scope of this PR, but it would be cool to understand why these values are hardcoded here instead of being given proper names and referenced like the others. Do we need to support these sizes? If so, should we move them to [ad-sizes](https://github.com/guardian/commercial-core/blob/55cdb59619dfce1a61aba3c7b12913f6c5d78501/src/ad-sizes.ts) in `commercial-core`?

dotcom-rendering/src/web/components/AdSlot.tsx
```diff
@@ -325,35 +292,35 @@ export const AdSlot: React.FC<Props> = ({ position, display }) => {
...
-							`${Size.outstreamGoogleDesktop}`,
-							`${Size.fluid}`,
-							`${Size.halfPage}`,
-							`${Size.skyscraper}`,
+							`${adSizes.outOfPage.toString()}`,
+							`${adSizes.empty.toString()}`,
+							`${adSizes.mpu.toString()}`,
+							`${adSizes.googleCard.toString()}`,
+							`${adSizes.video.toString()}`,
+							`${adSizes.outstreamMobile.toString()}`,
```
mxdvl, line 309: Is this meant to be mobile-sized on desktop?

dotcom-rendering/src/web/components/AdSlot.tsx
oliverlloyd: Very good question 🤔 

I dug into this some and ended up on [this PR](https://github.com/guardian/dotcom-rendering/pull/2258) which seems to be where the `Size` abstraction was introduced. That abstraction wasn't arbitrary though, it was built based on [this file](https://github.com/guardian/frontend/blob/main/static/src/javascripts/projects/commercial/modules/ad-sizes.js) in Frontend but as you can see, that file has no abstraction for these two sizes. So rather than change then or create a divergence from the legacy code, they were left hard coded.

That still doesn't explain why they are there though. So I dug further to find the original place where they were added and found they were sourced from [this code ](https://github.com/guardian/frontend/blob/9d2ee56372de4bb730146100527f83b2efd41c35/common/app/views/support/Commercial.scala#L142)in Frontend.

So the issue is/was that the named abstraction originally defined in Frontend, that we reverently copied over to DCR, was not complete and so when we tried to match the two up, we had this size left over.

Should it be added?
Yes. I think so. DCR is now rendering these slots 100% so there's not a problem with divergence anymore.

dotcom-rendering/src/web/components/AdSlot.tsx
```diff
@@ -541,7 +508,7 @@ export const AdSlot: React.FC<Props> = ({ position, display }) => {
 					data-label="false"
 					data-refresh="false"
 					data-out-of-page="true"
-					data-desktop={[`${Size.outOfPage}`].join('|')}
+					data-desktop={[`${adSizes.outOfPage.toString()}`].join('|')}
```
oliverlloyd, line 511:
```suggestion
					data-desktop={[adSizes.outOfPage].map(size => size.toString()).join('|')}
```

oliverlloyd: Obvs., this makes more sense in other contexts where there are more items but mayyybe this is a nicer way to do this? Just a thought, I am very happy to go with whatever you prefer

oliverlloyd (approved):
Nice improvement! I really like these small positive steps that keep happening with this commercial code

dotcom-rendering/src/web/components/AdSlot.tsx
simonbyford: Woopsy, this was a typo. Good spot. Fixing now..

dotcom-rendering/src/web/components/AdSlot.tsx
simonbyford: This is miles better @oliverlloyd. I'll make the change, thanks!

shtukas (approved):

PR #3445, opened 2021-09-21
Allow `null` browserId

mxdvl:
## What does this change?

Allows the `browserId` to be `null`, if there’s no `bwid` cookie set.

## Why?

This allows for the CMP to run in local and CODE environments.

### DIFF
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -340,7 +340,7 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 			country: countryCode,
 			pubData: {
 				platform: 'next-gen',
-				browserId: browserId ?? undefined,
+				browserId: browserId ?? undefined, // if `undefined`, the resulting consent signal cannot be joined to a page view.
 				pageViewId,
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/components/App.tsx
```diff
@@ -338,7 +337,7 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 			country: countryCode,
 			pubData: {
 				platform: 'next-gen',
-				browserId,
+				browserId: browserId ?? undefined,
```
oliverlloyd, line 340: Ah, so here you're transforming back to `undefined` so we're certain there are no down stream impacts? Clever

Might be worth adding a comment on this so no-one comes along later and cleans this up to just pass down `browserId` - I can see how that could be tempting.

mxdvl: In the future, this way of sending `pubData` will likely be deprecated. I’ll let @shtukas share his latest on the matter.

```suggestion
				browserId: browserId ?? undefined, // if `undefined`, the resulting consent signal cannot be joined to a page view.
```

PR #3446, opened 2021-09-21
CommercialMetrics includes ad blocker use

zekehuntergreen:
## What does this change?
- Upgrade to newest version of commercial-core, which includes a change to `sendCommercialMetrics` so that it accepts an optional parameter `adBlockerInUse `
- Pass the new parameter to `sendCommercialMetrics`

## Why?
We'd like to collect data on our readers' use of ad blockers

### `frontend` &rarr; https://github.com/guardian/frontend/pull/24191


dotcom-rendering/src/web/components/CommercialMetrics.tsx
```diff
@@ -16,7 +17,16 @@ export const CommercialMetrics: React.FC<{
 		false,
 	);
 
+	const [adBlockerInUse, setAdBlockerInUse] = useState<boolean>();
+
+	isAdBlockInUse().then((value) => {
```
zekehuntergreen, line 22: @mxdvl I think we have to do this here rather than use `useAdBlockInUse`. That function only sets the value of `isInUse` [here](https://github.com/guardian/dotcom-rendering/blob/150fc2d81e6a66d9c3336185e874fc8cd0288546/dotcom-rendering/src/web/lib/useAdBlockInUse.ts#L17) if `isAdBlockInUse` returns true, but won't set it to false.

mxdvl: Is there ever a case where someone would turn off their ad blocker without refreshing the page and we’d want to re-render? I think `useAdBlockInUse` is fine.

zekehuntergreen: The case that `useAdBlockInUse` doesn't work for is when the user doesn't have an ad blocker. `isAdBlockInUse` returns `false`, `isInUse` is not updated, the callback in `useEffect` isn't executed (because `adBlockerInUse` remains `undefined`)

mxdvl: Gotcha, makes sense.

oliverlloyd (changes requested):
I think it would be better to replace this promise with `useAdBlockInUse`.

Having async calls which set state at the root of react components is likely to cause race conditions because you're fighting with React's re-rendering lifecycle rather than _using_ it

dotcom-rendering/src/web/components/CommercialMetrics.tsx
```diff
@@ -16,7 +17,16 @@ export const CommercialMetrics: React.FC<{
...
+		setAdBlockerInUse(value);
+	}).catch((error) => {
+		console.log(error);
+	});
+
 	useEffect(() => {
```
oliverlloyd, line 28: This is very much outside this PR but this logic could be greatly simplified using `useOnce`. This type of use case where you only want to send a metric one time but only when a series of dependencies are defined (or just ready) is pretty much what that function was built for 😻

dotcom-rendering/src/web/components/CommercialMetrics.tsx
```diff
@@ -33,14 +43,20 @@ export const CommercialMetrics: React.FC<{
 
 		if (isDev || shouldForceMetrics || userIsInSamplingGroup)
 			setSentCommercialMetrics(
-				sendCommercialMetrics(pageViewId, browserId, Boolean(isDev)),
+				sendCommercialMetrics(
+					pageViewId,
+					browserId,
```
oliverlloyd, line 48: Another, non this PR related but perhaps useful, comment: @mxdvl it looks like we're not guarding against `browserId` being undefined here but we _could_, and maybe should?

We now know that if this prop is `undefined` then it means the cookie read hasn't completed and I think we want that to happen before sending these metrics. It's possible that this is always happening _by chance_ but it should be encoded, I think.

One simple, declarative, way to do this would be to just not render this component until `browserId` is defined using a conditional in `App.tsx`? That way we can make the prop required and the logic gets simplified

oliverlloyd: (I'm pinging Max because we were talking about this earlier in another thread)

zekehuntergreen:
Thanks for the feedback, @oliverlloyd !
With @mxdvl we discussed using `useAdBlockInUse` but agreed that we can't use the function as it is now since it doesn't set the state of `isInUse` to `false`. In order to use `useAdBlockInUse` I think we'd need to change [this line](https://github.com/guardian/dotcom-rendering/blob/150fc2d81e6a66d9c3336185e874fc8cd0288546/dotcom-rendering/src/web/lib/useAdBlockInUse.ts#L17) to  `setIsInUse(blockerDetected)`. The comment just above that line suggested that we rely on the current behavior but looking at its usage, I _think_ we might be able to make the change without breaking anything. 
An alternative might be to add a parameter like `onlySetTrue` (or something) to `useAdBlockInUse` so that we could toggle between the current behavior (only set the state of `isInUse` if it's `true`) and new behavior (always set the value). This seemed less clean to me than a direct call to `isAdBlockInUse` in `CommercialMetrics`.

oliverlloyd:
> doesn't set the state of isInUse to false

Ah, I see. Yes, that's true, it doesn't handle the use case where an action is required when _no_ ad blocker is in use

> I _think_ we might be able to make the change without breaking anything

I agree. It currently only changes state in the positive case purely as a micro optimisation based on the assumption we would only ever want to use this logic to decide to do things when an ad blocker _is_ detected. This is illustraighted with the [WhenAdBlockInUse](https://github.com/guardian/dotcom-rendering/blob/a42df26f3e33fd7340bdea059a3f67bf91b678db/dotcom-rendering/src/web/components/WhenAdBlockInUse.tsx) component. But it seems that assumption has been proven wrong!

dotcom-rendering/src/web/components/CommercialMetrics.tsx
zekehuntergreen: It looks like roughly 10% of the rows in `fastly_commercial` have `browser_id is null`. 
I've added a TODO in the code.

dotcom-rendering/src/web/components/CommercialMetrics.tsx
```diff
@@ -31,16 +37,21 @@ export const CommercialMetrics: React.FC<{
...
 			window.location.hostname.includes('localhost');
 
-		if (isDev || shouldForceMetrics || userIsInSamplingGroup)
+		if (isDev || shouldForceMetrics || userIsInSamplingGroup) {
+			const args: [string, string | undefined, boolean, boolean?] = adBlockerInUse === undefined
+				? [pageViewId, browserId, Boolean(isDev)]
+				: [pageViewId, browserId, Boolean(isDev), adBlockerInUse];
 			setSentCommercialMetrics(
-				sendCommercialMetrics(pageViewId, browserId, Boolean(isDev)),
+				sendCommercialMetrics(...args),
```
oliverlloyd, line 45: I think that, depending on the order of things, this code might end up being run more than once. The first time this code is run `adBlockerInUse` will be undefined and then - if an ad blocker is found - it will get set to `true`, triggering this code to rerun.

I wonder if moving over to `useOnce` now might be easiest here? It's actually pretty straight forward refactor and that way it guarantees we only send metrics once.  Happy to pair on this if it helps?

oliverlloyd (approved):
Scratch my comment about needing `useOnce`. @zekehuntergreen rightly pointed out the `sentCommercialMetrics` state prevents double submission so the use of `useOnce` is only needed for the nice here, not as an absolute.

dotcom-rendering/src/web/components/CommercialMetrics.tsx
```diff
@@ -15,43 +15,31 @@ export const CommercialMetrics: React.FC<{
...
-	);
-
-	const adBlockerInUse = useAdBlockInUse()
-
-	// TODO replace useEffect with useOnce, simplifying the below logic
-	useEffect(() => {
-		if (visibilityState !== 'hidden') return;
-		if (sentCommercialMetrics) return;
+	const adBlockerInUse = useAdBlockInUse();
+	const isHidden = visibilityState === 'hidden';
```
mxdvl, line 19: Maybe something like this could make it clearer?

```suggestion
	const isHidden = visibilityState === 'hidden' ?? undefined;
```

dotcom-rendering/src/web/components/CommercialMetrics.tsx
```diff
@@ -1,46 +1,45 @@
...
 			window.guardian.config.page.isDev ||
 			window.location.hostname.includes('localhost');
 
-		if (isDev || shouldForceMetrics || userIsInSamplingGroup)
-			setSentCommercialMetrics(
-				sendCommercialMetrics(pageViewId, browserId, Boolean(isDev)),
-			);
+		if (isDev || shouldForceMetrics || userIsInSamplingGroup) {
+			const args: [string, string | undefined, boolean, boolean?] =
+				adBlockerInUse === undefined
```
oliverlloyd, line 33: The way `useOnce` works means `adBlockerInUse` can never be `undefined` here, so I think you can just pass it in as a simple parameter and we no longer need this check

zekehuntergreen: Good call - that's much simpler

PR #3447, opened 2021-09-21
Update ar dependancies

marjisound:
## What does this change?

Update all AR dependencies & dev dependencies in line with the allowances of the package.json.

Fixes related to package updates:

**1.**  ~~set @typescript-eslint/no-unsafe-call to off in eslintrc
  for lint error `Unsafe call of an any typed value  @typescript-eslint/no-unsafe-call` in server.ts on `express.static(path.resolve(__dirname, '../assets'))`~~

**2.** code update for lint error `Component definition is missing display name  react/display-name` on anonymous function
```
const toReact =
	(format: Format) =>
	(node: Node, index: number): ReactNode => {...}
```
changed to 
```
const toReact = (format: Format) => {
	return function getReactNode(node: Node, index: number): ReactNode {...)
```
https://github.com/guardian/dotcom-rendering/pull/3447/commits/53dd24a4df27791212dffec97a44c48b68bd2edc

**3.** code update for lint error `Unnecessary optional chain on a non-nullish value  @typescript-eslint/no-unnecessary-condition` on the following lines (req is non-nullish):
```
const path = res.req?.path;
const isPreview = res.req?.query.isPreview === 'true';
```
https://github.com/guardian/dotcom-rendering/pull/3447/commits/53dd24a4df27791212dffec97a44c48b68bd2edc

**4.** run lint fix for prettier errors. 37 files were updated as a result of the fix. relevant commit for these fixes is https://github.com/guardian/dotcom-rendering/pull/3447/commits/4e58f704eb14b7b494297957c37b448a293d7342

**5.** ~~Due to compilation failure on new version of typescript for the e type, `// @ts-expect-error` was added:~~
```
catch (e) {
                 // @ts-expect-error because of compilation error
		logger.error(`Unable to load asset mapping`, e);
		throw e;
	}
```
https://github.com/guardian/dotcom-rendering/pull/3447/commits/02ee5d603c50b919da5b8a309f0c64e39fe7293c

As recommended by @JamieB-gu the logger interface was updated to cater for the unknown type https://github.com/guardian/dotcom-rendering/pull/3447/commits/3bc7de0d130e4aaa579085db71d1d2d5108319bf

**6.** update root dependancy `typescript` to match the version in dcr and ar `4.4.3` because otherwise storybook build was failing. https://github.com/guardian/dotcom-rendering/pull/3447/commits/ecfa4ce9df26f63df0c0b3c0dc8418877134ed87

## Why?
When we pull DCR & AR into yarn workspaces in the root of the monorepo, our lock files will be lost. By upgrading our packages to the most recent minor/patch versions available, when a new lock file is created, we'll have already fixed any associated bugs.


JamieB-gu:
> **1.** set @typescript-eslint/no-unsafe-call to off in eslintrc
> for lint error `Unsafe call of an any typed value @typescript-eslint/no-unsafe-call` in server.ts on `express.static(path.resolve(__dirname, '../assets'))`

I believe this is down to our dependency on `@types/serve-static` being locked to an outdated version (more details in this issue: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/49595). Running this should fix the issue and allow you to keep the lint rule intact:

```sh
npm update @types/serve-static --depth=2
```

JamieB-gu:
> **5.** Due to compilation failure on new version of typescript for the e type, `// @ts-expect-error` was added:

Do you think it would be possible to update the logger to deal with `unknown` errors? I think it should be a small change to a few type definitions, plus a couple of type guard checks to extract the stack traces in cases where `e` _is_ an instance of `Error`.

For context, [according to the docs on this TS change](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables), it was introduced because JavaScript allows anything to be thrown, not just `Error` objects.

apps-rendering/config/.eslintrc.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
 		project: 'tsconfig.json',
 	},
 	rules: {
+		'@typescript-eslint/no-unsafe-call': 'off',
```
OllysCoding, line 25: Would be nice to find a way to not have to do this if possible!

marjisound:
Thanks @JamieB-gu for your feedback and suggesting to update the logger interface. I made some updates in this commit https://github.com/guardian/dotcom-rendering/pull/3447/commits/14d70b7c06dba9ce5598f696272c86dc1b9279a0 for adding the unknown for the error type.

apps-rendering/config/.eslintrc.js
marjisound: Thanks Olly. Jamie's suggestion to update @types/serve-static worked so I deleted the `'@typescript-eslint/no-unsafe-call': 'off',` rule

JamieB-gu (approved):

apps-rendering/src/client/interactives.ts
```diff
@@ -12,22 +12,21 @@ function handleMessage(interactive: HTMLIFrameElement, message: string): void {
 			}
 		}
 	} catch (e) {
-		logger.error(e);
+		logger.error('Unable to handle the message', e);
```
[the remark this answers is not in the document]
OllysCoding: Maybe could be re-worded to 'Unable to parse interactive message' ?

OllysCoding (approved):

PR #3455, opened 2021-09-22
Do not constrain container width of `HeaderAdSlot` in Showcase and Comment layouts

MarSavar:
## Why?
There are a few advertising campaigns coming up in October that will use the fabric video template.

Fabric videos take up a max width of 1920px, but the div containing those ads has the `shouldCenter` prop set to `true` by default in certain layouts, which constrains the max width to 1300px on desktop (and lower breakpoints for users on smaller devices). 

This is a problem because advertisers are paying for a 1920px ad that is effectively cropped to 1300px on desktop.

This PR brings consistency with the other layouts so `shouldCenter` is always set to `false` for the div containing the header ad slot. It also has the added benefit of showing the light grey background introduced in https://github.com/guardian/dotcom-rendering/pull/3452 

### Before
image

### After
image

### DIFF
--- a/dotcom-rendering/src/web/layouts/CommentLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/CommentLayout.tsx
@@ -322,6 +322,7 @@ export const CommentLayout = ({
 						showTopBorder={false}
 						showSideBorders={false}
 						padded={false}
+						shouldCenter={false}
 					>
 						<HeaderAdSlot
 							isAdFreeUser={CAPI.isAdFreeUser}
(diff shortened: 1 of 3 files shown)


chrislomaxjones (approved):
👍

arelra (approved):
I'm assuming even though `shouldCenter` is false that ads will still be centred by the ad slot container?

MarSavar:
> I'm assuming even though `shouldCenter` is false that ads will still be centred by the ad slot container?

That's correct, `display: flex` and other alignment rules take care of centering the ads. This is already the case on frontend and other layouts in DCR.

dotcom-rendering/src/web/layouts/CommentLayout.tsx
```diff
@@ -322,6 +322,7 @@ export const CommentLayout = ({
 						showTopBorder={false}
 						showSideBorders={false}
 						padded={false}
+						shouldCenter={false}
```
chrislomaxjones, line 325:
```suggestion
```

Just a thought - since `shouldCenter` is an optional prop could we delete this here? It might avoid any confusion about the fact the ad ends up centred in the slot.

chrislomaxjones: Ignore that it defaults to true doesn't it!

PR #3460, opened 2021-09-23
Storybook : Add AR Live Prototype story

abeddow91:
## What does this change?
Adds a story for an AR live layout component.

## Why?
Provides a visual example of how far we can currently build a live layout with existing AR platform components.


joecowton1 (approved):
Looks great!

PR #3463, opened 2021-09-24
Add `performance.now` to `polyfill.io` bundle

simonbyford:
## What does this change?

Adds `performance.now` to the `polyfill.io` bundle.

## Why?

[Some older browsers don't support the User Timing API](https://caniuse.com/user-timing). Mostly we guard against this case by checking if `window.performance` is `undefined` before calling methods on it, but we discovered one instance where this wasn't happening and we believe this to be the cause of a number of Sentry errors. Adding a polyfill for older browsers feels like the right approach so we don't always need to worry about guards.

Related PR: https://github.com/guardian/frontend/pull/24210

### DIFF
--- a/dotcom-rendering/src/web/server/document.tsx
+++ b/dotcom-rendering/src/web/server/document.tsx
@@ -228,7 +228,7 @@ export const document = ({ data }: Props): string => {
 	];
 
 	const polyfillIO =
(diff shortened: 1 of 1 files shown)


davidfurey:
Ophan currently track how many browsers don't support `performance.now`.  Merging this PR will break that tracking.  It is probably fine for us to stop tracking this feature but @rtyley will have a more informed opinion.

https://github.com/guardian/ophan/pull/4065

simonbyford:
Thanks for this @davidfurey, this isn't something I'd considered (or know anything about!). Keen to hear your thoughts @rtyley 🙂

rtyley (approved):
Many thanks for the heads-up @davidfurey . I would say there's no blocker from the Ophan side - we have the stats we need!

https://github.com/guardian/ophan/pull/4065#issuecomment-928997588

PR #3464, opened 2021-09-24
Remove `emotion/core` Dependency

JamieB-gu:
## Why?

This dependency isn't used explicitly, but somehow it's being included in the AR client-side script and causing errors. Just removing the existence of the dependency seems to fix this, which is odd.

## Changes

- Removed `@emotion/core` dependency
- Update `discussion-rendering` to latest (to stop it relying on `emotion/core`)

### DIFF
--- a/apps-rendering/package-lock.json
+++ b/apps-rendering/package-lock.json
@@ -3113,11 +3113,6 @@
         }
       }
     },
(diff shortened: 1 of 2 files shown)


frankie297 (approved):
Great work on the speedy fix!

PR #3461, opened 2021-09-23
Migrate Ophan types from @guardian/types to @guardian/libs and bump version

shtukas:
[ Replaces: #3453 . Submitting it again to save time ]

Migrate Ophan types from @guardian/types to @guardian/libs and bump version. This is in preparation to solving an ongoing consent submission problem.

PR #3465, opened 2021-09-24
Repair the submission of consent component events

shtukas:
The (currently experimental) consent related component events that are sent from DCR are not reaching their destination in the lake. Careful investigation revealed that it might be caused by an incorrect query parameters (see below)

Before

... which doesn't match the expected name in Ophan ( https://github.com/guardian/ophan/blob/0c1181bb273655af8ae523cc903ca17baeea7a08/the-slab/app/extractors/ComponentEventExtractor.scala#L73 ). This change essentially fix that bug by calling the right Ophan Tracker-JS submission function

Correct URL: 

After

Two previous PRs were needed to make this happen
1. https://github.com/guardian/dotcom-rendering/pull/3461 
2. https://github.com/guardian/libs/pull/286

### DIFF
--- a/dotcom-rendering/package.json
+++ b/dotcom-rendering/package.json
@@ -46,7 +46,7 @@
     "@guardian/commercial-core": "0.19.2",
     "@guardian/consent-management-platform": "~6.11.5",
     "@guardian/discussion-rendering": "^7.0.0",
-    "@guardian/libs": "^3.0.0",
+    "@guardian/libs": "^3.3.0",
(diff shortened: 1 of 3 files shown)


DavidLawes (approved):
I'm assuming that the change to query params is a result of using function submitComponentEvent, and the bump in guardian/libs is to enable us to use predefined ophan types

shtukas:
@DavidLawes 
Yes, I updated the Guardian types here : https://github.com/guardian/libs/pull/286 :)
Cheers,

PR #3468, opened 2021-09-27
LIVE-3100: update logic to include a tags

joecowton1:
## Why?

Gallery captions weren't rendering all `textContent` nodes.

### Before

Screenshot 2021-09-27 at 11 53 30

### After
Screenshot 2021-09-27 at 11 48 12

### DIFF
--- a/apps-rendering/src/components/editions/galleryImage/index.tsx
+++ b/apps-rendering/src/components/editions/galleryImage/index.tsx
@@ -6,7 +6,7 @@ import { neutral, remSpace } from '@guardian/src-foundations';
 import { from } from '@guardian/src-foundations/mq';
 import { textSans } from '@guardian/src-foundations/typography';
 import type { Format, Option } from '@guardian/types';
-import { map, none, some, withDefault } from '@guardian/types';
+import { map, none, OptionKind, some, withDefault } from '@guardian/types';
 import type { Image } from 'bodyElement';
 import { maybeRender, pipe } from 'lib';
 import type { FC } from 'react';
@@ -26,7 +26,7 @@ type CaptionProps = {
 
 type CaptionDetails = {
 	location: Option<string>;
-	description: Option<string>;
+	description: Option<string[]>;
 };
 
 const sizes: Sizes = {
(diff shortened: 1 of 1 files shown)


apps-rendering/src/components/editions/galleryImage/index.tsx
```diff
@@ -49,7 +49,18 @@ const styles = css`
 const getCaptionDetails = (oDoc: Option<DocumentFragment>): CaptionDetails => {
 	const details: CaptionDetails = {
 		location: none,
-		description: none,
+		description: some([]),
```
marjisound, line 52: Just asking so that I can learn more typescript :) Should we not indicate what is the array type? Like is it string[] or number[] or etc?

joecowton1: The `CaptionDetails` type on line 50 contains the `description` type: `Option<string>` :)

marjisound: Oh I see, so here we are instantiating the object by giving the description an empty array. Thanks Joe

apps-rendering/src/components/editions/galleryImage/index.tsx
```diff
@@ -59,9 +70,9 @@ const getCaptionDetails = (oDoc: Option<DocumentFragment>): CaptionDetails => {
 					? some(node.textContent)
 					: none;
 			} else if (node.nodeName === '#text') {
-				details.description = node.textContent
-					? some(node.textContent)
-					: none;
+				details = pushToDescription(details, node);
+			} else if (node.nodeName === 'A') {
+				details = pushToDescription(details, node);
```
marjisound, line 75: I'm wondering if both `if (node.nodeName === '#text')` and `if (node.nodeName === 'A')` are going to have the same implementation as `details = pushToDescription(details, node);` what's the reason to have them seperately?

joecowton1: That's a very good point. Fixed now.

marjisound (approved):
LGTM

abeddow91 (approved):
Nice work!

PR #3471, opened 2021-09-27
Fix JS assets path issues in `frontend`

mxdvl:
## What does this change?

Explicitly serve assets from `localhost:3030/` instead of the root path `/`. Send a CORS header to any request from `localhost`.

Also fix a small typo in British spelling of “favour”.

## Why?

Currently, when an article is loaded by `frontend` locally, the static assets are pointing to `/`, which is incorrect, as the assets are actually served by `dotcom-rendering`’s `:3030` server, not `frontend`’s `:9000`.

The CORS request is needed for the cross-origin request, as different ports count as different origins.

This allows developers working between the two projects to easily navigate between fronts and articles, such as @guardian/commercial-dev.

### Before

failing assets requests

### After

fixed assets requests


nitro-marky (approved):
LGTM

PR #3472, opened 2021-09-27
Local Image-Rendering

JamieB-gu:
## Why?

Now we have image-rendering in the mono-repo we can start using the local files instead of the package.

**Note:** A re-creation of #3400 due to merge conflicts.

## Changes

- Removed `@guardian/image-rendering` dep
- Migrated AR imports over to local image files
- Switched image components over to default exports
- Made `Dpr` a normal `enum` because DCR doesn't support `const enum`s
- Explicitly specified `react` location for Jest, so it doesn't import two different versions


joecowton1 (approved):
Looks good!

marjisound (approved):

marjisound (approved):

PR #3484, opened 2021-09-28
Local Image Rendering - Alias React

JamieB-gu:
## Why?

As seen in #3482, using code from `common-rendering` came with a problem. Somewhere along the way two different instances of react were being resolved and loaded, most likely because `apps-rendering` and `common-rendering` both directly depend on it. This is a common enough problem that React has [some documentation](https://reactjs.org/warnings/invalid-hook-call-warning.html#duplicate-react) on it.

I don't _believe_ two copies of the code were being included in the bundle, but the issue still occurred and caused a runtime crash.

There are two solutions to this that I can think of:

1. Explicitly tell all bundlers how to resolve references to React. In practice that means adding entries to Jest's `moduleNameMapper` and Webpack's `alias`es. This is the approach taken in this PR.
2. Stop depending on `react` directly in AR, `common-rendering` and DCR. Instead depend on it only in the project's root `package.json`. This is a step on the road towards having a single dependency tree for all parts of this project. This is the approach taken in #3485.

I've opened both PRs as drafts to get feedback, and will close one and promote the other once we settle on a preference.

## Changes

- All the `image-rendering` changes from #3472
- Add an alias to AR's `react` in the webpack config

### DIFF
--- a/apps-rendering/config/jest.config.js
+++ b/apps-rendering/config/jest.config.js
@@ -50,4 +50,7 @@ module.exports = {
 	moduleDirectories: ['node_modules', 'src'],
 	snapshotSerializers: ['@emotion/jest/serializer'],
 	transformIgnorePatterns: ['node_modules/(?!@guardian)'],
+	moduleNameMapper: {
+		'^react$': '<rootDir>/node_modules/react'
+	}
 };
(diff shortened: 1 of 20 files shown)


JamieB-gu:
Closing in favour of #3485.

PR #2717, opened 2021-03-18
Update Ophan Tracker-JS for Attention-Time unskew

rtyley:
This upgrade of Ophan's Tracker JS contains two updates:

* https://github.com/guardian/ophan/pull/4063 - work towards removing a possible source of Attention Time metric bias, that is likely to have caused [Attention Times of zero seconds to have been relatively over-reported for dotcom-rendering](https://trello.com/c/Vomnrhwe/39-pageview-underreporting-of-attention-time-on-dcr) (DCR).
  
* https://github.com/guardian/ophan/pull/4065 - gather stats on browser support for the `performance.now()` API, as a potential way to avoid _negative_ attention times being reported by browsers undergoing clock-skew.

See https://github.com/guardian/frontend/pull/23612 for the corresponding update there.

`yarn install` initially failed for this project on my dev laptop, but then I reinstalled XCode and it passed!

PR #3486, opened 2021-09-29
Remove Ophan tracking levels of `performance.now` browser support

rtyley:
This updated version of the Ophan Tracker-JS library no longer tracks `performance.now` browser support:

https://www.npmjs.com/package/ophan-tracker-js/v/1.3.28

The tracking has been in place for about half a year (introduced to DCR with https://github.com/guardian/dotcom-rendering/pull/2717) and has allowed us on the Ophan team to establish that only 0.007% of pageviews are from browsers that _don't_ support the `performance.now` API - having got that, we don't need Ophan's Tracker-JS to track this anymore. Also, DCR have recently introduced a `performance.now` polyfill (https://github.com/guardian/dotcom-rendering/pull/3463) that would means we would no longer be accurately tracking _native_ browser support.

See also:

* https://github.com/guardian/ophan/pull/4065
* https://github.com/guardian/ophan/pull/4277

### DIFF
--- a/dotcom-rendering/package.json
+++ b/dotcom-rendering/package.json
@@ -102,7 +102,7 @@
     "lodash.get": "^4.4.2",
     "log4js": "6.3.0",
     "minify-css-string": "^1.0.0",
-    "ophan-tracker-js": "^1.3.27",
+    "ophan-tracker-js": "^1.3.28",
     "pm2": "5.0.0",
     "preact": "^10.5.14",
     "preact-render-to-string": "^5.1.19",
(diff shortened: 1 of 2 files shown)


nitro-marky (approved):
LGTM

rtyley:
Thanks for the review @nitro-marky! What does _"1 change must be accepted as baseline."_ mean? How do I do that?
 

image

nitro-marky:
It means Chromatic has picked up an inconstancy (which is most likely a third party problem). I'll check them now.

PR #3490, opened 2021-09-30
Move LiveBlockContainer to common-rendering and adapt towards matching new designs

lucymonie:
## What does this change?
This extracts the container for a LiveBlog block (aka a LiveBlock) into a separate and shareable component in common-rendering called LiveBlockContainer. The changes here also encompass changes to the design, including:
- margins and padding have changed a little
- the position of the 'time since publication and time published' info (this used to be in the left margin of the block and is now at the top on breakpoints from tablet upwards
- images are no longer stretched, which means that several individual containers that we used to provide padding can now be removed, and the padding can be provided by the LiveBlock container
- ~the caption icon has changed from a theme-coloured triangle to a grey camera and the caption font size has changed from 12px to 14px (NB: once these changes are merged in, this will change the styles of all DCR captions across articles. I have checked this with @HarryFischer and this is an intentional change to caption styling across all DCR articles).~
- There are also some changes that are introduced by prettier, which hasn't been running due to husky being turned off.

There are some pieces of further work that are out of scope and that will be handled separately:
- Move the timestamp into common-rendering so it can be shared
- Strip out element stretching as this is no longer required
- Move the block heading into common-rendering so it can be shared
- Move the date helpers into a common-rendering so they can be shared

## Why?
The LiveBlockContainer is a component that can accept a borderColour (based on the Pillar), an id and children, so it is a shareable component for Liveblogs.

## Before
Screen Shot 2021-10-15 at 13 57 59

## After
Screen Shot 2021-10-15 at 13 58 58


nitro-marky:
So if I want to add just an image to a block I need to create an image child component?

oliverlloyd:
It looks like the time alignment is a bit off on the stories. It sits too far to the right.

I notice that in the PR description the image shows more information in this header, which is aligned well. So I guess the question is, is just having time like this a valid scenario and if not do we need to adjust the props to prevent it?

Screenshot 2021-10-14 at 12 45 37

lucymonie:
@JamieB-gu to respond to your point about the timestamp:

> 2. Am I right in thinking that the shared container will require the platforms to supply their own published/updated timestamp UI as part of `children`? Do you think we'd want to get to a place where the container can take these datetime strings as simple props, and have it provide the UI for displaying them?

I think that would be a great way to iterate on the LiveBlockContainer component and improve its shared usefulness. Perhaps we could create a card to add this functionality once this piece of work is complete.

dotcom-rendering/src/web/components/Caption.tsx
```diff
@@ -3,9 +3,10 @@ import { css } from '@emotion/react';
 import { from, until } from '@guardian/src-foundations/mq';
 import { textSans } from '@guardian/src-foundations/typography';
 import { space } from '@guardian/src-foundations';
+import { neutral } from '@guardian/src-foundations/palette';
 import { ArticleDisplay, ArticleDesign, ArticleSpecial } from '@guardian/libs';
 
-import TriangleIcon from '@frontend/static/icons/triangle.svg';
+import CameraIcon from '@frontend/static/icons/camera.svg';
```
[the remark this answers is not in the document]
lucymonie: Thanks for flagging this issue. The icon and caption updates are now in a separate PR: https://github.com/guardian/dotcom-rendering/pull/3532

dotcom-rendering/src/web/components/LiveBlock.stories.tsx
```diff
@@ -39,7 +40,10 @@ const Wrapper = ({ children }: { children: React.ReactNode }) => {
 		<div
 			css={css`
 				padding: 20px;
-				max-width: 620px;
+				max-width: 740px;
+				${from.tablet} {
+					width: 740px;
+				}
```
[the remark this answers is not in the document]
lucymonie: Thanks for explaining this. I have updated this to `${breakpoints.tablet}px` as suggested.

common-rendering/src/components/liveBlockContainer.tsx
```diff
@@ -0,0 +1,40 @@
+import { css } from '@emotion/react';
+
+import { from } from '@guardian/src-foundations/mq';
+import { neutral, space } from '@guardian/src-foundations';
+
+const LEFT_MARGIN_DESKTOP = 60;
+const SIDE_MARGIN = 20;
+const SIDE_MARGIN_MOBILE = 10;
```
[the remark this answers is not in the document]
lucymonie: Given this, I will leave this at 10 until further notice.

dotcom-rendering/src/web/components/Caption.tsx
```diff
@@ -107,9 +108,14 @@ const hideIconBelowLeftCol = css`
...
-	fill: ${palette.fill.captionTriangle};
-	padding-right: 3px;
+const iconStyle = css`
+	fill: ${neutral[46]};
+	margin-right: ${space[1] /2}px;
+	display: inline-block;
+	vertical-align: middle;
+	svg {
+		width: 85%;
+	}
```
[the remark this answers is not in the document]
lucymonie: Thanks for flagging this issue. The icon and caption updates are now in a separate PR: https://github.com/guardian/dotcom-rendering/pull/3532

.storybook/main.js
```diff
@@ -142,7 +142,7 @@ const arWebpack = (config) => {
 	});
 
 	config.resolve.modules = [
-		...(config?.resolve?.modules || []),
+		...(config && config.resolve && config.resolve.modules || []),
```
OllysCoding, line 145: Is this is needed? I'm under the impression that the version of node in the `.nvmrc` supports this syntax

dotcom-rendering/src/web/components/LiveBlock.tsx
```diff
@@ -1,6 +1,5 @@
 import { css } from '@emotion/react';
```
OllysCoding, line 1: I noticed a lot of components that are part of the LiveBlock live here (in the dotcom-rendering subdirectory), is the intention to share them, or is it easier to keep them DCR specific?

JamieB-gu: I suspect there'll be a couple more iterations on live blocks:

1. To lift more functionality into the container component, as @lucymonie mentioned above. The timestamp functionality is a good candidate for this, but I think probably the title and header can go that way too.
2. To strip out the logic for stretching content full width (`headerElement`, `typesWeStretch` etc.), since the designs no longer call for this.

Again, as Lucy mentioned above, it may be best to create follow-up tickets to handle this in small chunks?

common-rendering/src/components/liveBlockContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+				background: ${neutral[100]};
+				border-top: 1px solid ${borderColour};
+				${from.tablet} {
+					padding: ${space[2]}px ${SIDE_MARGIN}px;
+					padding-left: ${LEFT_MARGIN_DESKTOP}px;
+				}
+			`}
+		>
+			<main>{children}</main>
+		</div>
```
JamieB-gu, line 35: Just thinking about semantics, should this be?

```tsx
<article id="..." css={...}>
  {children}
</article>
```

I don't think you're allowed to have more than one `<main>` in a document: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main

However, multiple `<article>`s is fine, and I think is semantically exactly what we want here: each block is basically a standalone "article". I think we may also be able to avoid the wrapping `div` if we can style on the `<article>` tag directly?

dotcom-rendering/src/web/components/LiveBlock.tsx
```diff
@@ -149,6 +70,8 @@ const LastUpdated = ({
 	);
 };
 
+const formatTime = (time: number) => (time < 10 ? `0${time}` : time);
```
JamieB-gu, line 73: There's a function in AR's `date` module called `padZero` that does almost exactly this 😄. Perhaps we should consider lifting and sharing? If you think that's outside the scope of this PR I can create a ticket.

JamieB-gu (commented):
Couple final comments (sorry! 🙈).

dotcom-rendering/src/web/components/LiveBlock.tsx
oliverlloyd: There's one being used in Source too `const pad = (n: number): number | string => n.toString().padStart(2, '0');` 😸 

There's probably a lot of this code in the internets.

.storybook/main.js
lucymonie: Hey Olly, good question. This was breaking when I ran `yarn chromatic` and I think @nitro-marky was having an issue with this the other day also.

dotcom-rendering/src/web/components/LiveBlock.tsx
lucymonie: Yes, I think there are more pieces we could share, but I wouldn't think to make them part of the LiveBlockContainer, but potentially separate compose-able elements that could be children of it. Either way, I agree that these can be done as follow-up work.

common-rendering/src/components/liveBlockContainer.tsx
lucymonie: Yeah, good shout. Thanks Jamie 😄

dotcom-rendering/src/web/components/LiveBlock.tsx
lucymonie: Sounds great, yes please! I'll add a TODO here 👍🏻

dotcom-rendering/src/web/components/LiveBlock.tsx
lucymonie: On the point about `typesWeStretch`, I _think_ we'll still need this for the pinned images / videos at the top of the blocks. I could be misunderstanding though.

dotcom-rendering/src/web/components/LiveBlock.tsx
lucymonie: Isn't time always going to be < 24 || < 60 ? In these instances, at least?

JamieB-gu (approved):

dotcom-rendering/src/web/components/LiveBlock.tsx
oliverlloyd: @mxdvl you're right! That `number` type is a hangover from when we added [padStart](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart)

dotcom-rendering/src/web/components/LiveBlock.tsx
```diff
@@ -245,80 +170,46 @@ export const LiveBlock = ({
...
-								index,
-								pageId,
-								webTitle,
-							})}
-						</BlockText>
-					);
-				})}
-			</main>
-			<Footer>
+			{/* For each element, we decide what margins to set depending on the type */}
```
oliverlloyd, line 198: Is this comment still true?

dotcom-rendering/src/web/components/LiveBlock.tsx
```diff
@@ -245,80 +170,46 @@ export const LiveBlock = ({
...
-						})}
-				</span>
+				{block.blockFirstPublished && (
+					<FirstPublished
+						firstPublished={block.blockFirstPublished}
+						blockLink={blockLink}
+					/>
+				)}
+				{block.title && <BlockTitle title={block.title} />}
+				{headerElement &&
```
oliverlloyd, line 185: I'm wondering, do we need this separation (headerElement/mainElements) now? It was first added ( _think_) to accommodate the fact the first element can't stretch, but now that's moot

dotcom-rendering/src/web/components/LiveBlock.tsx
oliverlloyd: Also (and sorry to add to this thread but the best PRs are the ones with lots of comments on them!) maybe we can name this function `padString` (or similar)? `formatTime` suggests more is happening.

dotcom-rendering/src/web/components/LiveBlock.tsx
```diff
@@ -245,80 +170,46 @@ export const LiveBlock = ({
...
+						palette,
+						element,
+						isMainMedia: false,
+						index,
+						pageId,
+						webTitle,
+					})}
+				</p>
+			))}
+			<div>
```
oliverlloyd, line 212:
```suggestion
			<footer>
```

dotcom-rendering/src/web/components/LiveBlock.tsx
```diff
@@ -345,7 +236,7 @@ export const LiveBlock = ({
 							)}
 					</div>
 				</Hide>
-			</Footer>
-		</Container>
+			</div>
```
oliverlloyd, line 239:
```suggestion
			</footer>
```

dotcom-rendering/src/web/components/LiveBlock.tsx
```diff
@@ -245,80 +170,46 @@ export const LiveBlock = ({
...
-								webTitle,
-							})}
-						</BlockText>
-					);
-				})}
-			</main>
-			<Footer>
+			{/* For each element, we decide what margins to set depending on the type */}
+			{mainElements.map((element, index) => (
+				<p key={`${element._type}-${index}`}>
```
oliverlloyd, line 200: Using a `p` tag here feels wrong.  In fact, having a wrapper element at all feels wrong. Looking at frontend, I can see block elements are not wrapped at all - the same as with articles.

dotcom-rendering/src/web/components/LiveBlock.tsx
```diff
@@ -10,9 +9,7 @@ import { decidePalette } from '@root/src/web/lib/decidePalette';
 import { Hide } from '@root/src/web/components/Hide';
 import { ShareIcons } from '@root/src/web/components/ShareIcons';
 import { headline, textSans } from '@guardian/src-foundations/typography';
-
-const ASIDE_WIDTH = 58;
-const GUTTER = space[3];
+import LiveBlockContainer from '@guardian/common-rendering/src/components/liveBlockContainer';
```
oliverlloyd, line 12: Outside this PR but `LiveBlockContainer` is a React component so must be named using a leading capital letter. So the filename should follow this?

dotcom-rendering/src/web/components/LiveBlock.tsx
```diff
@@ -149,6 +70,9 @@ const LastUpdated = ({
 	);
 };
 
+// TODO: update this code to use shared version when it is available
```
oliverlloyd, line 73: Random aside: I have about a 20% hit rate on actually going back and really doing those things I write todos for - I have all the right intentions, it's just other stuff comes along. The rest just sit in the code and linger. There are currently 97 TODOs in DCR, I'm sure many of them are mine.

I wonder if rather than writing them as notes to self, if maybe we should write them as messages to future developers, saying that there are some improvements they can help with. With that mindset we would add hints and tips to help these future versions of us and the task would stand a much better chance of being done

oliverlloyd (commented):
My favourite PRs at the Guardian and the ones that have the most comments 😸 

I wonder if this PR's title and description need updating to reflect the changes from the latest changes?

dotcom-rendering/src/web/components/LiveBlock.tsx
JamieB-gu: I think we're going to deal with this in a follow-up ticket/PR. See https://github.com/guardian/dotcom-rendering/pull/3490#discussion_r741325836

lucymonie:
> My favourite PRs at the Guardian and the ones that have the most comments 😸
> 
> I wonder if this PR's title and description need updating to reflect the changes from the latest changes?

Yes, I thought about that fleetingly the other day. I will change it 👍🏻

dotcom-rendering/src/web/components/LiveBlock.tsx
lucymonie: Yes, sure - that seems reasonable

dotcom-rendering/src/web/components/LiveBlock.tsx
lucymonie: I guess not, given the container is handling the padding now 👍🏻

dotcom-rendering/src/web/components/LiveBlock.tsx
lucymonie: Yes, I think it's best to keep this separate as we're aiming to lift-and-share the header

dotcom-rendering/src/web/components/LiveBlock.tsx
lucymonie: Hmm. How would you propose to handle the key?

dotcom-rendering/src/web/components/LiveBlock.tsx
lucymonie: The convention that's (so far) been established in common-rendering is that the file names of the shared components are lowercased (and in apps-rendering too by the looks of things). Happy to revisit it if this an important question.

dotcom-rendering/src/web/components/LiveBlock.tsx
lucymonie: Yes, perhaps we should. There is a trello card to get this work done, so perhaps that is a more reliable way of ensuring that this particular TODO doesn't end up being another piece of code flotsam.

dotcom-rendering/src/web/components/LiveBlock.tsx
oliverlloyd: We need to remove the existence of the tag completely. Like this:

```js
			{mainElements.map((element, index) => (
				renderArticleElement({
					format,
					palette,
					element,
					isMainMedia: false,
					index,
					pageId,
					webTitle,
				})
			))}
```

dotcom-rendering/src/web/components/LiveBlock.tsx
oliverlloyd: 🤔 I'm not entirely sure that this header element concept relates to sharing the 'header'. I _think_ we're confusing two things with 'header' in the name (but equally I could just be confused myself!).

But I'm happy to revisit this in another PR when there's less noise 👍

dotcom-rendering/src/web/components/LiveBlock.tsx
oliverlloyd: Shall we review where we are with all these areas that we diverge on? I think it would be great to get them out in the open and make some decisions

lucymonie: For sure. Let's talk about it at the next retro 👍🏻

dotcom-rendering/src/web/components/LiveBlock.tsx
oliverlloyd: This is [how we render article elements](https://github.com/guardian/dotcom-rendering/blob/7cc306b44ad6944b2d6c03791b3d31ebeafe7504/dotcom-rendering/src/web/lib/ArticleRenderer.tsx#L58) 🤷

oliverlloyd: For articles, we actually [add considerable complexity](https://github.com/guardian/dotcom-rendering/blob/a42df26f3e33fd7340bdea059a3f67bf91b678db/dotcom-rendering/src/web/components/elements/RewrappedComponent.tsx#L7) for the purpose of avoiding wrapping elements

That use case ^^ is not as relevant here as spacefinder will work differently but I think the principle of maintaining semantic html is still a strong motivator.

mxdvl: the `index` prop will turn into a key, most of the time, when passed to `renderArticleElement`: [`renderElement.tsx` L117](https://github.com/guardian/dotcom-rendering/blob/7cc306b44ad6944b2d6c03791b3d31ebeafe7504/dotcom-rendering/src/web/lib/renderElement.tsx#L117)

JamieB-gu: A couple of things about keys that may or may not be relevant here, based on the docs: https://reactjs.org/docs/reconciliation.html#recursing-on-children

- They recommend _not_ using array index as the key, because it can throw up issues when re-ordering elements.
- I don't think keys are that important for server-side rendering. My understanding is that they exist mainly to improve the performance of client-side re-renders. This may become more relevant if we decide to use React to handle client-side updates however.

dotcom-rendering/src/web/components/LiveBlock.stories.tsx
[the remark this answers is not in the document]
mxdvl: I know you’ve moved away from Source, as these values were not quite right, but here’s the Source PR nonetheless: https://github.com/guardian/source/pull/1144

oliverlloyd (approved):
Amazing work!

PR #3491, opened 2021-09-30
Add legacy element-[role] class to elements in interactive articles

nicl:
## What

Adds role-specific classes to elements (when role is set) for interactives.

Note, this improves things for some articles but doesn't entirely fix image and atom layout as DCRs surround markup/CSS is still quite different. It does seem to remove the horiztonal scrollbar in many cases and also ensure the left-padding is present.

## Why?

Many legacy interactives style elements using this.

### Before

Screenshot 2021-09-30 at 17 37 28

### After

Screenshot 2021-09-30 at 17 37 00

### Before

Screenshot 2021-09-30 at 17 39 11

### After

Screenshot 2021-09-30 at 17 38 39

### DIFF
--- a/dotcom-rendering/src/web/components/Figure.tsx
+++ b/dotcom-rendering/src/web/components/Figure.tsx
@@ -149,6 +149,7 @@ type Props = {
 	role?: RoleType | 'richLink';
 	id?: string;
 	isNumberedListTitle?: boolean;
+	className?: string;
 };
 
 const mainMediaFigureStyles = css`
@@ -161,6 +162,7 @@ export const Figure = ({
 	id,
 	isMainMedia,
 	isNumberedListTitle = false,
+	className = '',
 }: Props) => {
 	if (isMainMedia) {
 		// Don't add in-body styles for main media elements
@@ -181,6 +183,7 @@ export const Figure = ({
 			data-spacefinder-ignore={
 				isNumberedListTitle ? 'numbered-list-title' : null
 			}
+			className={className}
 		>
 			{children}
 		</figure>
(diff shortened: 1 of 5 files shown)


DavidLawes (approved):
Agree that this will help improve some things. Regarding weighting of graphics, do we have a feel for what will be 'acceptable' by the Visuals team?

When I was looking at https://www.theguardian.com/world/ng-interactive/2021/jan/08/which-countries-have-reported-new-uk-covid-variant, to achieve parity with Frontend we needed a slightly different solution (at least for how InteractiveBlockComponents were displaying):
- Unlike Frontend, DCR is wrapping the interactive block component figures inside a figure
- Applying the element--{role} class to the outer figure helps us achieve some of the necessary styling (is only doing this 'acceptable'?)
- To achieve parity I needed to apply a class element--inline to the outer Figure, and apply the element--{role} class to the inner figure (the InteractiveBlockComponent)

^^ Maybe that is taking the fix too far, I'm not sure. What do you think?

Overall, I don't have a feel for:
- what UI is acceptable, if an interactive is to be migrated
- how long we should spend on each fix before considering pressing
- how each fix affects other interactives (eg https://github.com/guardian/dotcom-rendering/pull/3489 could regress other migrated interactives)

nicl:
Thanks @DavidLawes . Re approach, this does feel a bit hacky I agree. I'll spend a bit more time on investigating a fuller fix.

Re:

* what UI is acceptable, if an interactive is to be migrated
I don't think Visuals have an exact answer for this either unfortunately. It would be nice to codify it somehow though.

* how long we should spend on each fix before considering pressing
My own view at least is that we should not be spending any time ourselves now on fixes beyond this week really.

* how each fix affects other interactives (eg #3489 could regress other migrated interactives)
I agree without some sophisticated automation here (e.g. automatic diffs and reporting) this is not possible, which is another argument for us not to invest too much time in these fixes.

nicl:
Update: @DavidLawes I've tidied this a bit and added some tests. It's not complete parity with Frontend now, but pretty close and I don't think we could achieve parity without larger refactors (to do with what gets added to figures in atoms-rendering vs dotcom-rendering, e.g. around role-based classes), which I'm reluctant to embark on.

PR #3492, opened 2021-09-30
Commercial Cypress Tests - Increase Ad Slots To At Least 14

JamieB-gu:
## Why?

The Cypress commercial tests have recently started failing. They expect the number of ad slots on a particular long read piece to be 15, and for some reason the number has consistently started coming back as 17. We're not sure what the reason is yet, but given that the test appears to otherwise work, and it's failing on all new PRs, we'd like to get it back into a working state.

**Update:** The number is proving to be non-deterministic, so we've switched to `≥ 14`.

We can follow up with a health ticket to figure out why this number seems to have changed. Possibly the commercial team will want to look into this @mxdvl?

## Changes

- Increased the number of ad slots expected to 17

### DIFF
--- a/dotcom-rendering/cypress/integration/parallel-6/commercial.spec.js
+++ b/dotcom-rendering/cypress/integration/parallel-6/commercial.spec.js
@@ -22,7 +22,7 @@ describe('Commercial E2E tests', function () {
 		// cases this check will pass much faster
 		cy.get('.js-ad-slot:not([data-name="survey"]', {
 			timeout: 30000,
-		}).should('have.length', 15);
+		}).should('have.length', 17);
 
 		Array(10)
 			.fill()
--- a/dotcom-rendering/cypress/integration/parallel-6/commercial.spec.js
+++ b/dotcom-rendering/cypress/integration/parallel-6/commercial.spec.js
@@ -22,7 +22,7 @@ describe('Commercial E2E tests', function () {
 		// cases this check will pass much faster
 		cy.get('.js-ad-slot:not([data-name="survey"]', {
 			timeout: 30000,
-		}).should('have.length', 17);
+		}).should('have.length.of.at.least', 14)
 
 		Array(10)
 			.fill()


DavidLawes (approved):
nice, thanks! following the chat in the dotcom channel, this makes sense

mxdvl (approved):
Happy in principle, but it seems to have gone back to 15? 🤷

@guardian/commercial-dev one for us to investigate further

mxdvl:
Looking at the original PR #2030, I think that it would be better to check for the presence of certain IDs in the following query:

```js
Array.from(document.querySelectorAll('.js-ad-slot:not([data-name="survey"]')).map(s => s.id)
```

For 15, this is the list I get:
1. "dfp-ad--inline1"
2. "dfp-ad--inline2"
3. "dfp-ad--inline3"
4. "dfp-ad--inline4"
5. "dfp-ad--inline5"
6. "dfp-ad--inline6"
7. "dfp-ad--inline7"
8. "dfp-ad--inline8"
9. "dfp-ad--inline9"
0. "dfp-ad--inline10"
0. "dfp-ad--inline11"
0. "dfp-ad--right"
0. "dfp-ad--merchandising-high"
0. "dfp-ad--mostpop"
0. "dfp-ad--merchandising"

dotcom-rendering/cypress/integration/parallel-6/commercial.spec.js
```diff
@@ -22,7 +22,7 @@ describe('Commercial E2E tests', function () {
 		// cases this check will pass much faster
 		cy.get('.js-ad-slot:not([data-name="survey"]', {
 			timeout: 30000,
-		}).should('have.length', 15);
+		}).should('have.length', 17);
```
mxdvl, line 25:
```suggestion
		}).should('have.length.of.at.least', 14)
```

mxdvl (approved):
Maybe it has to do with the subscription ask?

In any case, this should make it happy. cc @tomrf1

PR #3504, opened 2021-10-05
Stop onwards carousel buttons obscuring banner

tjmw:
## What does this change?

Before this the carousel buttons had a higher z-index so they were sitting on top of banners in the sticky bottom banner slot.

## Why?

The buttons are the only part of the carousel which sit above the banner, so it looks confusing.

### Before

Screenshot 2021-10-05 at 10 58 26

### After

Screenshot 2021-10-05 at 10 57 34


jlieb10 (approved):
Nice one 👍

PR #3569, opened 2021-10-27
Husky (again)

oliverlloyd:
## What does this change?
Adds Husky to the root of the repo.

I strongly advise you to review commit by commit 👇 for your sanity.
Screenshot 2021-10-27 at 17 36 56

## Why?
Because life without prettier is cruel

## Why on push?
Up for debate. It could be on commit. Pros and cons. It should be a team decision.

## Before
anarchy

## After
jon-hamm-explosion


marjisound (approved):

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+    lint:
+        name: DCR Prettier
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+            - uses: guardian/actions-setup-node@main
+
+            # Cache npm dependencies using https://github.com/bahmutov/npm-install
+            - uses: bahmutov/npm-install@v1
```
JamieB-gu, line 16: FYI, I think `guardian/actions-setup-node` is able to handle this directly now? @sndrs pulled in the upstream changes that add support for caching.

.husky/pre-push
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+yarn workspace @guardian/dotcom-rendering prettier:check
+yarn workspace @guardian/dotcom-rendering tsc
+yarn workspace @guardian/dotcom-rendering lint-staged
```
sndrs, line 6: I don't think this will do anything in `prepush` to any of the changed you've already committed. it will only lint and updated staged changed, so I _think_ this will only lint staged, uncommitted changes that are not part of what you are pushing but you happened to have already staged (if that makes sense?)

I think `lint-staged` really needs to be `precommit`

oliverlloyd: Oh interesting. Thanks for resurrecting and highlighting this! I'll create a `pre-commit` file too

oliverlloyd: Done: https://github.com/guardian/dotcom-rendering/pull/3614

PR #3571, opened 2021-10-28
Use Core Web Vitals from `@guardian/libs`

mxdvl:
## What does this change?

Use `@guardian/lib`’s `initCoreWebVitals`. Linked to https://github.com/guardian/frontend/pull/24321

This has been tested locally.

## Why?

Shared code from a shared lib.


dotcom-rendering/src/web/browser/coreVitals/init.ts
```diff
@@ -1,15 +1,31 @@
...
-
-const init = async (): Promise<void> => {
-	// Sample 1 out of 100
-	const inSample = Math.floor(Math.random() * 100);
-	const inLocal = window.location.hostname === 'localhost';
-	if (inSample === 1 || inLocal) {
-		coreVitals();
-	}
+
+const coreVitals = (): Promise<void> => {
```
oliverlloyd, line 8:
```suggestion
const initCoreVitals = (): Promise<void> => {
```

Nitpick but most of these other init scripts are called 'init' - I think it's nice to maintain the convention that we're initialising something here?

Ideally, this code we use to initialised things like this would have a little more structure now that its grown some

dotcom-rendering/src/web/components/CommercialMetrics.tsx
```diff
@@ -31,14 +31,10 @@ export const CommercialMetrics: React.FC<{
 			window.location.hostname.includes('localhost');
 
 		if (isDev || shouldForceMetrics || userIsInSamplingGroup) {
-			sendCommercialMetrics(pageViewId, browserId, Boolean(isDev), adBlockerInUse);
+			sendCommercialMetrics(pageViewId, browserId, isDev, adBlockerInUse);
+			// TODO: force CWV, too
```
oliverlloyd, line 35: Why do we want to force core web vitals here? Might be worth extending this comment to give more context?

oliverlloyd (approved):
👍

dotcom-rendering/src/web/browser/coreVitals/init.ts
mxdvl: Sorry, I looked at the first two (`atomIframe` and `embedIframe`), and they didn’t follow this pattern. Happy to oblige

dotcom-rendering/src/web/components/CommercialMetrics.tsx
mxdvl: Whenever we capture commercial metrics, we also capture CWV for that page, as we want to ensure that better commercial performance doesn’t come at the expense of user experience.

It’s outside the scope of this PR to address this now, but I’ll expand on the comment.

PR #3573, opened 2021-10-28
Wire Up The Live Layout In AR

JamieB-gu:
## Why?

We'd like to begin displaying liveblog/deadblog layouts in AR, using the components we've built so far. This PR wires up the `Live` layout and adds the shared `KeyEvents` component to that layout.

**Note:** Liveblogs and deadblogs are still filtered out for production users, this is only visible for us.

## Changes

- Make casing on `LiveBlog` consistent
- Added `DeadBlog` item
- Stopped limiting number of blocks parsed for liveblogs
- Added support for deadblogs in `LiveblogHeader`
- Wired up `Live` layout
- Added `KeyEvents` to `Live` layout

### DIFF
--- a/apps-rendering/src/fixtures/live.ts
+++ b/apps-rendering/src/fixtures/live.ts
@@ -14,7 +14,7 @@ import type { Option } from '@guardian/types';
 import { parse } from 'client/parser';
 import type { MainMedia } from 'headerMedia';
 import { MainMediaKind } from 'headerMedia';
-import type { Liveblog } from 'item';
+import type { LiveBlog } from 'item';
 import { pipe } from 'lib';
 
 const parser = new DOMParser();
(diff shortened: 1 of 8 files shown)


apps-rendering/src/fixtures/live.ts
```diff
@@ -260,7 +260,7 @@ const fields = {
 	webUrl: '',
 };
 
-const live: Liveblog = {
+const live: LiveBlog = {
 	design: Design.LiveBlog,
 	...fields,
 	blocks: [],
```
marjisound, line 266: Is it worth adding blocks data here, so that we can see the key events in the storybook for live blog?

marjisound (approved):

apps-rendering/src/fixtures/live.ts
JamieB-gu: Yes, good idea. I'll do that in a follow-up.

PR #3574, opened 2021-10-28
Send section and tag ids to banner service

tomrf1:
We're testing targeting banners by content/tag, see https://github.com/guardian/support-dotcom-components/pull/548
`sectionName` was previously being passed into `ReaderRevenueBanner` but wasn't actually used. I've replaced it with `sectionId` which is preferable.

Request now includes `sectionId` and `tagIds`:
Screen Shot 2021-10-28 at 16 13 25


shtukas (approved):

KaliedaRik (approved):
All seems fine to me

PR #3576, opened 2021-10-29
Apply AMP regional classes to containers instead of `amp-` elements

chrislomaxjones:
## What does this change?

Move the application of regional classes to div containers that surround the AMP element they were previously applied to.

## Why?

This fixes an issue with the mobile sticky ad. The AMP runtime applies some additional styling when a sticky ad is filled that cannot be overridden, which caused the sticky ad to be made visible regardless of the regional classes implied. This had the effect of delivering an impression for an ad regardless of region.

With this fix the styling is not overridden as it applies to the container.


mxdvl (approved):
Nice! 💅

PR #3577, opened 2021-11-01
Rename `canShow` functions

oliverlloyd:
## What does this change?
This PR gives the various `canShow` functions more explicit names, indicating what they are relating to.

```
SignInGate/displayRule/canShow > canShowSignInGate
BrazeEpic/canShow > canShowBrazeEpic
ReaderRevenueEpic/canShow > canShowReaderRevenueEpic
BrazeBanner/canShow > canShowBrazeBanner
```

## Why?
Because it will make it easier to refactor the props for some of these functions later on. We we would like to remove use of the `CAPI` object as a prop and this is an easier task when each function has its own name that we can search the codebase for.

### DIFF
--- a/dotcom-rendering/src/web/components/SignInGate/displayRule.ts
+++ b/dotcom-rendering/src/web/components/SignInGate/displayRule.ts
@@ -114,7 +114,7 @@ export const hasRequiredConsents = (): Promise<boolean> => {
 	});
 };
 
-export const canShow = (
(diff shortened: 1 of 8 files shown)


tjmw (approved):
LGTM 👍

Might be worth getting 👀 from someone on the reader revenue team too?

TomPretty (approved):
Yeah looks good to me too 👍

PR #3579, opened 2021-11-01
Stop calling IDAPI GET /user/me on every article page

QuarpT:
## What does this change?

Stop calling IDAPI GET /user/me on every article page

- The Braze ID is retrieved from a new optimised endpoint on IDAPI.
- The email address required for Reader Revenue epic and banner reminder service is provided lazily. It is returned with a timeout because the result is used to autofill the email address in the reminder form. If the result is not returned in one second, the user will be shown the form to fill the email address manually

## Why?

We want to reduce requests to IDAPI GET /user/me in preparation to migrating Identity user data to OKTA. OKTA's API has rate limit thresholds.

https://github.com/guardian/support-dotcom-components/pull/551 needs to be released first
This has been merged ^

### DIFF
--- a/dotcom-rendering/cypress/integration/parallel-2/braze.spec.js
+++ b/dotcom-rendering/cypress/integration/parallel-2/braze.spec.js
@@ -1,7 +1,8 @@
 import { setLocalBaseUrl } from '../../lib/setLocalBaseUrl.js';
 import { overrideGeo } from '../../lib/overrideGeo';
 import { hasCurrentBrazeUser } from '../../lib/hasCurrentBrazeUser';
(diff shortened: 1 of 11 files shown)


oliverlloyd:
I know this is still in draft but I wanted to highlight that this is an area that is currently being discussed and could change.

We use the `user` object in several places, including discussion, and so it makes sense to think about `user` as common, global state for DCR and having a strategy for global state is critical for the Dotcom platform. However, in this case it's made more complex because a lot of the code using this data isn't 'Reacty'. Which is to say, it's doesn't follow the declarative pattern that react components have and works more like traditional javascript.

Neither approach is wrong but having both is problematic and prone to causing race conditions. There's an ongoing discussion around how to deal with this (do we lean more or less into React?) but in the short term we're thinking about ways to simplify things to help create a common approach and that work could impact here.

QuarpT:
> I know this is still in draft but I wanted to highlight that this is an area that is currently being discussed and could change.
> 
> We use the `user` object in several places, including discussion, and so it makes sense to think about `user` as common, global state for DCR and having a strategy for global state is critical for the Dotcom platform. However, in this case it's made more complex because a lot of the code using this data isn't 'Reacty'. Which is to say, it's doesn't follow the declarative pattern that react components have and works more like traditional javascript.
> 
> Neither approach is wrong but having both is problematic and prone to causing race conditions. There's an ongoing discussion around how to deal with this (do we lean more or less into React?) but in the short term we're thinking about ways to simplify things to help create a common approach and that work could impact here.

👍  it sounds good to move to a pattern more idiomatic to React. If the changes you mentioned are merged first, we can refactor this PR. Otherwise would you be happy for this to go in as is, given it follows the existing pattern?

On Identity team our goal is to reduce calls to IDAPI GET /user/me, since we need to get within OKTA's rate limits. Our quotas are not high enough to call GET /user/me from article pages. For articles with comments we should get user discussion profile information from DAPI instead of IDAPI.

oliverlloyd:
> would you be happy for this to go in as is

Happy to unblock you, I know the rate limiting has been a concern for a while now. I'll try to find some time tomorrow to review this properly

Aside from that, I'd be keen to go through these changes in more detail if you've got some time this week? I'd love to get a better understanding of the motivations and what the differences are between the different endpoints so that when we do refactor this code we have a better idea of the use cases

QuarpT:
> Aside from that, I'd be keen to go through these changes in more detail if you've got some time this week? I'd love to get a better understanding of the motivations and what the differences are between the different endpoints so that when we do refactor this code we have a better idea of the use cases

sure, let me know when 😄

oliverlloyd (approved):
Thanks for spending the extra time on the tests here!

I think, in general, it would be preferable to have a more idiomatic, Reacty, way of doing this, where we hoist the shared state up into a common parent and manage asynchronous side effects using React's lifecycle methods. But this wouldn't be simple here and is very much outside the scope of this PR so I'm happy to move forward with this approach 👍

PR #3582, opened 2021-11-01
More logging in `App.tsx`

oliverlloyd:
## What does this change?
This takes advantage of the `log` function from `@guardian/libs` to record more detail on the render flow in `App.tsx`

## Usage
To show/hide these logs in your console

```js
window.guardian.logger.subscribeTo('dotcom');
window.guardian.logger.unsubscribeFrom('dotcom');
```

See: https://github.com/guardian/libs/blob/main/src/logger.README.md

## Why?
Having progress logged like this will offer developers information to help them reason about the rendering flow. It also provides timing data, helping us learn we're spending time during renders

### Output
Screenshot 2021-11-01 at 19 20 01


DavidLawes (approved):
nice!

PR #3583, opened 2021-11-02
Fix Instagram Chromatic diffs

oliverlloyd:
## What does this change?
Disables the Chromatic story for the AR Instagram ClickToView story

## Why?
Because every time the Guardian got a new follower on Instagram it meant we got false negatives on our PRs

### DIFF
--- a/apps-rendering/src/components/embedWrapper.stories.tsx
+++ b/apps-rendering/src/components/embedWrapper.stories.tsx
@@ -113,6 +113,11 @@ const Spotify: FC = () => (
 	</div>
 );
 
+// I'm disabling the requirement for a return type because that doesnt make sense here 🤷. It
+// popped up after I removed the FC type which was causing problems when I tried to set parameters
+// on an individual story. Ideally, we'd use the type from Storybook directly
+// but https://github.com/storybookjs/storybook/issues/13486
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type -- because 👆
 const Instagram = () => (
 	<div>
 		<p>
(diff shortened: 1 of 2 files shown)


lucymonie (commented):
This looks fine to me. @OllysCoding and @marjisound bumping the prettier version won't cause knock on issues for the sub-projects, will it?

OllysCoding:
Is there a way we can avoid the warning it github about the missing function return type?

OllysCoding (approved):
+1

OllysCoding:
> bumping the prettier version won't cause knock on issues for the sub-projects, will it?
@lucymonie In this case I think the version bump should be fine :)

oliverlloyd:
> Is there a way we can avoid the warning it github about the missing function return type?

In DCR, we don't require return types. Forcing a return type can actually lead to problems.

This is [explained better here](https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript). Specifically:

> Ok, so what about the return value? Well, we could type it as React.ReactElement or even wider as a JSX.Element. But honestly, I side with my friend Nick McCurdy when he says that mistakes can easily be made causing the return type to be too wide. So even outside a react context, I default to not specifying the return type (rely on inference) unless necessary. And that's the case here.

Case in point. This warning is happening because I removed the existing return type on this story because that type didn't let me use `Instagram.story` which should be a valid action but the type being used (React.FC`) didn't allow it

lucymonie:
> > Is there a way we can avoid the warning it github about the missing function return type?
> 
> In DCR, we don't require return types. Forcing a return type can actually lead to problems.

We recently updated the tsconfig with `"noImplicitReturns": true`, which is perhaps the cause of the issue? Is the problem that GH doesn't recognise the disabling of this rule?

oliverlloyd:
> We recently updated the tsconfig with `"noImplicitReturns": true`, which is perhaps the cause of the issue? Is the problem that GH doesn't recognise the disabling of this rule?

We only set that for dotcom-rendering but this error is being thrown by apps-rendering

PR #3585, opened 2021-11-02
Add Blocks To Live Fixtures

JamieB-gu:
## Why?

Gives us a better representation of a liveblog layout in storybook. Suggestion from @marjisound here: https://github.com/guardian/dotcom-rendering/pull/3573#discussion_r740229783

You should be able to see the difference in storybook/chromatic.

## Changes

- Added some blocks to the live fixtures


marjisound (approved):
Nice one 👍

PR #3586, opened 2021-11-02
Remove Onward Styles

JamieB-gu:
## Why?

A few reasons:

1. All related content sections had a double-nested `<section>` tag
2. Almost every usage of these was related content
3. We'd like to have styles live alongside components as much as possible

**Note:** This does change the styles of the comments section, but the styles of that section are already majorly broken and need an overhaul at some point.

## Changes

- Removed `onwardStyles`
- Removed nested `<section>` tags


marjisound (approved):
Looks good 👍

PR #3597, opened 2021-11-05
Add video icon for video captions

lucymonie:
## What does this change?
This adds the video icon to video captions rather than the camera option. This is an additional feature off the back of the web caption style update in this PR: https://github.com/guardian/dotcom-rendering/pull/3532

## Why?
It's another visual cue that this is a piece of video content.

### Before
Screen Shot 2021-11-05 at 15 45 11

### After
Screen Shot 2021-11-05 at 15 45 39

#### Android
Screen Shot 2021-11-05 at 16 25 56

#### iPhone
Screen Shot 2021-11-05 at 16 27 39


rcrphillips (approved):

dotcom-rendering/src/web/components/Caption.tsx
```diff
@@ -141,6 +149,7 @@ export const Caption = ({
 	shouldLimitWidth = false,
 	isOverlayed,
 	isLeftCol,
+	isVideo = false,
```
oliverlloyd, line 152: Could we change this to `mediaType`? CAPI already has this a type and it's defined as a global in DCR's index.d.ts file:

```ts
type MediaType = 'Video' | 'Audio' | 'Gallery';
```

dotcom-rendering/src/web/components/Caption.tsx
```diff
@@ -163,9 +172,10 @@ export const Caption = ({
 					iconStyle(palette),
 					format.display === ArticleDisplay.Immersive &&
 						hideIconBelowLeftCol,
+					isVideo && videoIconStyle,
```
oliverlloyd, line 175: Maybe it makes more sense to have `CameraIcon` and `VideoIcon` as their own components here? It will read nicer and offers up a much nicer path for future expansion if we want to have other icons types or different colours, etc. in the future

oliverlloyd (commented):
So, fun fact, there's also the possibility to have a video element in a Witness block element but the numbers are low so I'd only worry about this if it's not a lot of work

dotcom-rendering/src/web/components/Caption.tsx
lucymonie: Good suggestion, thanks Oliver

dotcom-rendering/src/web/components/Caption.tsx
lucymonie: Sure, that would be cleaner 👍🏻

lucymonie:
> So, fun fact, there's also the possibility to have a video element in a Witness block element but the numbers are low so I'd only worry about this if it's not a lot of work

The `WitnessImageBlockComponent` doesn't seem to use the `Caption` component so I have left it be.

oliverlloyd (approved):
🎉

PR #3598, opened 2021-11-07
LIVE-3028: Renders Ordered Lists in AR

umguardian:
## What does this change?
Adds the ability for AR to render ordered list `<ol>` elements.

## Why?
Currently, only unordered list `<ul>` elements are renderable in AR. The implementation for rendering `<ol>` elements is missing in AR, which this pull request addresses.

### Before

LIVE-3028_before_change

### After

LIVE-3028_after_change

### DIFF
--- a/apps-rendering/src/components/orderedList.tsx
+++ b/apps-rendering/src/components/orderedList.tsx
@@ -3,7 +3,6 @@
 import { css } from '@emotion/react';
 import type { SerializedStyles } from '@emotion/react';
 import { remSpace } from '@guardian/src-foundations';
-import { neutral } from '@guardian/src-foundations/palette';
 import type { FC, ReactNode } from 'react';
 
 // ----- Component ----- //
@@ -14,16 +13,19 @@ const styles: SerializedStyles = css`
 	padding-left: 0;
 	clear: both;
 
+	counter-reset: li;
 	> li::before {
-		display: inline-block;
-		content: '';
-		border-radius: 0.5rem;
-		height: 1rem;
-		width: 1rem;
-		margin-right: ${remSpace[2]};
-		background-color: ${neutral[86]};
-		margin-left: -${remSpace[6]};
-		vertical-align: middle;
+		content: counter(li);
+		counter-increment: li;
+		display: block;
+	}
+
+	> li {
+		padding-left: 0;
+	}
+
+	> li p:first-of-type {
+		display: block;
 	}
 `;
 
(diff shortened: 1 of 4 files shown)


apps-rendering/src/components/orderedList.tsx
```diff
@@ -0,0 +1,42 @@
...
+// ----- Component ----- //
+
+const styles: SerializedStyles = css`
+	list-style: none;
+	margin: ${remSpace[3]} 0;
+	padding-left: 0;
+	clear: both;
+
+	counter-reset: li;
+	> li:before {
```
JamieB-gu, line 17: The distinction is subtle, but this is a pseudo-element so it should be `li::before` - see the "Note" here: https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements#syntax

apps-rendering/src/components/orderedList.tsx
```diff
@@ -0,0 +1,42 @@
...
+		display: block;
+	}
+
+	> li {
+		padding-left: 0;
+	}
+
+	> li p:first-of-type {
+		display: block;
+	}
```
JamieB-gu, line 29: Are you sure this rule is needed?

JamieB-gu (commented):
Looks great! Couple of small comments.

apps-rendering/src/components/orderedList.tsx
umguardian: Without this rule, the elements in the ordered list appear with a very less gap between them as shown in the picture below. Note in `listItem`, the `p` has `display: inline`:
```
> p:first-of-type {
	display: inline;
	padding: 0;
}
```

image

JamieB-gu (approved):

marjisound (commented):
Great work @umguardian 
The unordered list styling seems to have changed. I noticed it in the UI test. It used to be like this: 
image

Now it's like this: 
image

Was this intentional?

umguardian:
> Great work @umguardian The unordered list styling seems to have changed. I noticed it in the UI test. It used to be like this: image
> 
> Now it's like this: image
> 
> Was this intentional?

Until this PR, the style definition for `<li>` elements was present at the `<li>` level itself. i.e. the styling would be the same irrespective of whether the `<li>` elements were to be rendered inside `<ul>` or `<ol>` parents.

The change in this PR moves the style definition of `<li>`elements into the parent element. So, `<li>` elements will be rendered differently depending on which element is the parent: `<ul>` or `<ol>`. @marjisound

marjisound (approved):

PR #3607, opened 2021-11-08
Adding break-word property in order to avoid page display issues

rcrphillips:
## What does this change?
Brings in a new property for paragraphs in order to avoid long words breaking the layout on mobile.

https://trello.com/c/gZ7P7psh/374-page-overrun-on-ios-safari

### Before

Screen Shot 2021-11-08 at 17 21 52

### After
Screen Shot 2021-11-08 at 17 25 44


ioannakok (approved):

ioannakok:
+1

PR #3610, opened 2021-11-09
Fix epic when there is a floating element above

tomrf1:
Before:
Screen Shot 2021-11-09 at 08 29 12

After:
Screen Shot 2021-11-09 at 08 29 18


MarSavar:
LGTM but I'll let @guardian/dotcom-platform give the final green light! 

`clear: left` would also work in this case (since `ad-slot--im`, the inline merchandising ad class, has a `float: left` property) but I think it's probably better to use  `clear: both`, which prevents the overlap from happening if there's a floating element on the right as well.

johnduffell (approved):
thanks Tom!

PR #3620, opened 2021-11-10
AR Live Layout Refactor, Part One: Top Half Of Header

JamieB-gu:
## Why?

Let's bring AR's live layout into line with the designs.

I'm breaking the overall refactor of this layout down into chunks - doing it all at once would make a _massive_ PR. This bit is concerned with just the top half of the header.

**Note:** I've started using Source's `Container` and `Columns` components for this. They work extremely well, I think they're fantastic @SiAdcock @oliverlloyd 🎉!

## Changes

- Refactored part of the header (headline, series, standfirst)
- Refactored series styles for clarity
- Introduced updated timestamp (this is still a TODO, we need to parse the updated timestamp and pass it through)

## Screenshots

| | Before | After |
| - | - | - |
| Mobile | live-layout-before-mobile | live-layout-mobile |
| Tablet | live-layout-before-tablet | live-layout-tablet |
| Desktop | live-layout-before-desktop | live-layout-desktop |
| Wide | live-layout-before-wide | live-layout-wide |


apps-rendering/src/components/series.tsx
```diff
@@ -56,85 +107,28 @@ const immersiveStyles = (
...
-		return immersiveStyles(getThemeStyles(format.theme), isLabs);
-	}
-
-	if (
-		format.design === Design.LiveBlog ||
-		format.design === Design.DeadBlog
-	) {
-		return css(liveblogWidthStyles, liveblogPhabletSidePadding);
+	if (design === Design.LiveBlog || design === Design.DeadBlog) {
+		return css();
```
marjisound, line 121: Is this css empty because now that we are using container/column we don't need to have side paddings any more?

marjisound (approved):
Looks good. Just left a small question :)

apps-rendering/src/components/series.tsx
JamieB-gu: Yes 👍

HarryFischer:
Hard to tell from screenshots but I think this looks correct. Is the left column 220px at Desktop?

JamieB-gu:
> Hard to tell from screenshots but I think this looks correct. Is the left column 220px at Desktop?

Yeah fair point. When the whole page is looking better I'll start putting these onto CODE so you can inspect them.

Yup it's 220 🙂. I've started using the Source `Columns` component which uses the grid rules to set its sizes. In the code you can see `span={3}`, which means three columns width. With a gutter of 20px and a column size of 60px that's `(60 x 3) + (20 x 2) = 220`.

The other column is `span={8}`, so `(60 x 8) + (20 x 7) = 620`.

PR #3622, opened 2021-11-11
Remove @guardian/src-svgs as it is unused

jamie-lynch:
## What does this change?

Remove the `@guardian/src-svgs` package from the `package.json`...

## Why?

...because it's not used anywhere.


oliverlloyd (approved):
LGTM!

PR #3623, opened 2021-11-11
Send sectionId for epic targeting

tomrf1:
Currently DCR sends the `sectionName` in the epic targeting payload to SDC.
This is inconsistent because for tag-based targeting we use `tagId`.
With this PR we send `sectionId` instead.

For a time SDC will support both fields:
https://github.com/guardian/support-dotcom-components/pull/559


oliverlloyd (approved):
👍

oliverlloyd:
It's a shame that DCR has this property on its type as `section` when it probably should be `sectionId`.

I think it's worth looking at adding `CAPI.sectionId` to the model and using that instead of `CAPI.config.section`. Not a blocker for this work, more a note to @guardian/dotcom-platform

PR #3624, opened 2021-11-11
Remove `palette` from `App.tsx`

oliverlloyd:
## What?
This deletes all usage of `palette` from `App.tsx`. We no longer have any components that are hydrated or inserted as portals that have this object as a prop.

## Why?
Because `palette` is quite large now and if we ever want to move away from centralised state to a more autonomous model where each component serialised it's own props in the dom, then we want those props to be as explicit and small as possible.

### DIFF
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -84,7 +84,6 @@ import {
 	OphanComponentEvent,
 } from '../browser/ophan/ophan';
 import { trackPerformance } from '../browser/ga/ga';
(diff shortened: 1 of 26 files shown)


DavidLawes (approved):
This change makes sense to me, we're passing in a smaller prop (`format`) to our components instead of `palette` 👍

oliverlloyd:
[Review App.tsx and suggest 'Recommendations' and guardrails](https://trello.com/c/A2Jz7EYm/395-review-apptsx-and-suggest-recommendations-and-guardrails)

PR #3625, opened 2021-11-11
Remove Source's palette object from dotcom-rendering

SiAdcock:
## Why?

The `palette` object is deprecated in Source and will be removed in v4. This change prepares for the upgrade to v4 and might shave off a few bytes, since the palette object is not tree-shakeable


OllysCoding (approved):
LGTM!

PR #3626, opened 2021-11-11
fix: Issue where width wasn't being set well for these articles

oliverlloyd:
## What does this change?
Removes the width for these articles to fix a production issue

### Before
Screenshot 2021-11-11 at 15 15 28

### After
Screenshot 2021-11-11 at 15 13 38


DavidLawes (approved):
I've chatted with Harry, we think this will solve the problem for interactives

PR #3627, opened 2021-11-11
Pass styling to Editions hydrated share icon 

abeddow91:
## What does this change?
Removes existing emotion styling and adds a parent div that passes a classname style down to the share icon. It also moves the existing span wrapper inside the share icon component file.

## Why?
The share icon for Editions gets hydrated by the client. Hydrated components can't use Emotion styling so the share icon is appearing un-styled. Instead we have to pass styles through with a classname on a parent div. 
### After
 

co-authored with @joecowton1

### DIFF
--- a/apps-rendering/src/components/editions/shareIcon/index.tsx
+++ b/apps-rendering/src/components/editions/shareIcon/index.tsx
@@ -66,13 +66,12 @@ const AndroidShareIcon = (): ReactElement => (
 	</svg>
 );
 
-const buttonStyles = css`
-	background: none;
-	border: none;
-	padding: 0;
-	height: 2.5rem;
-
-	svg {
+const hydratedButtonStyles = css`
+	.share-button {
+		background: none;
+		border: none;
+		padding: 0;
+		height: 2.5rem;
 		box-sizing: content-box;
 	}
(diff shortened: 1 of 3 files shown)


apps-rendering/src/components/editions/shareIcon/index.tsx
```diff
@@ -86,18 +85,22 @@ const ShareIcon: FC = () => {
...
-			}
-		>
-			{platform === Platform.IOS ? (
-				<IOSShareIcon />
-			) : (
-				<AndroidShareIcon />
-			)}
-		</button>
+		<div css={hydratedButtonStyles}>
+			<span className="js-share-button" role="button">
```
JamieB-gu, line 89: Do you need both the `div` and the `span`? Can the styles, class and role all go onto the `div`?

JamieB-gu (approved):

apps-rendering/src/components/editions/shareIcon/index.tsx
abeddow91: Good point, we can! Amended as per.

groakland (approved):
LGTM

PR #893, opened 2019-10-29
Partial rendering and hydration of react on the client

oliverlloyd:
## What does this change?
This replaces our current approach of fully hydrating the client app with partial hydration of separate react nodes, only where required.

## React Islands
Currently we render content on the server and then hydrate the entire app on the client. This technique only hydrates the specific areas of the app that require react, leaving the rest as static html.

The components using this approach are:

`Nav (the main pillar menu in the Header)`
`EditionDropdown`
`MostViewed`
`ReaderRevenueLinks`
`RichLinks`
`ShareCount`

The user experience is that the page will load and whatever content that exited the server render will be present on the page. React is then added to each dynamic part of the page with the following impacts:

### `Nav (the main pillar menu in the Header)`
The main pillar menu is immediately present on page load. The More button only starts working after react is loaded

### `EditionDropdown`
The dropdown control is visible on the page but does not become active until react is added

### `MostViewed`
The entire component is missing on initial load and only appears after react is loaded and the api call to get the content returns

### `ReaderRevenueLinks`
This component is missing on initial render and only appears after react is loaded

### `RichLinks`
These are hidden on initial load and pop in once react is present and the api call returns

### `ShareCount`
This is hidden on initial load and appears once react is present and the api call returns

## Next steps
This code tests well and is functional. It is logical that it should be more performant but the improvement is difficult to measure so discussion around an approach for implementation is required.

With this PR in place it would be logical to review the data we pass to the client using `window.guardian`. This could likely be substantially reduced.

## Why?
Performance

## Link to supporting Trello card
https://trello.com/c/cA7ymUYg/781-spike-portals

PR #3629, opened 2021-11-12
🏝 React Islands

oliverlloyd:
## What?
React islands are a form of [partial hydration](https://addyosmani.com/blog/rehydration/) where instead of hydrating the entire server side rendered page, we only hydrate the parts of the page that contain interactivity.

This PR takes the existing [partial hydration solution](https://github.com/guardian/dotcom-rendering/pull/893) in DCR, extends it to be more aligned with the [original island hydration work](https://github.com/guardian/dotcom-rendering/pull/445) by @AWare and @walaura and then adds some new features [inspired by AstroJS](https://github.com/snowpackjs/astro/tree/74372a7d269882fe8bbcbf02a85a79296d58668f/packages/astro/src/frontend/hydrate).

## Why?
The primary goal for this work was to provide a better developer experience for other teams working on DCR but we're also adding two features here that were [raised on the original PR](https://github.com/guardian/dotcom-rendering/pull/893#issuecomment-549850965) by @gtrufitt that we never got around to implementing.

### Developer experience
Although technically effective, the old pattern was complex and led to errors. To hydrate a component required edits in five different parts of the codebase as well as having to reason through the `App.tsx` lifecycle methods.

This PR presents a simpler api that removes the need for new developers to learn non standard, complex patterns and also, because we're no longer executing in a React context, we remove the problem of re-renders. The complexity that does still remain is now abstracted behind a developer friendly api.

### Only send the data required
By serialising the props for each component alongside its declaration and not in a global window object, we ensure we only send what is required.

### Lazy hydration
We now have an api to only hydrate components when visible, deferring content below the fold until scrolled into view. We can also tell the browser to hydrate less important content when it is idle, preventing the thread from being blocked.

## Usage
```jsx
// Server
<Hydrate when="idle">
    <EditionDropdown
        edition={CAPI.editionId}
    />
</Hydrate>

```

To hydrate a new component on DCR, all developers have to do is wrap it with `<Hydrate when="visible">` on the server and then make sure that the component filename follows the naming convention of `[MyThing].importable.tsx`.

## What about global state?
One of the reasons the api offered here is so simple is the lack of any global state. Some hydrated components use shared state managed by `App.tsx` and for these to migrate to this api they would first need to become more independent. [This document](https://docs.google.com/document/d/1gHOzvFZxrAO2mvNPKg4UmMAYZm7Jk1NutZPVLY5UxPY/edit#heading=h.1y29fcc1ygqg) talks about this concept more.

## What about Portals?
Portals are where no content is server side rendered and we're simply inserting new html into the dom. This use case could follow a very similar pattern to the one presented here. It could even share the same utilities like `whenVisible`.


oliverlloyd:
[Review App.tsx and suggest 'Recommendations' and guardrails](https://trello.com/c/A2Jz7EYm/395-review-apptsx-and-suggest-recommendations-and-guardrails)

dotcom-rendering/src/web/browser/hydration/getName.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * getName takes the given html element and returns its name attibute
+ *
+ * We expect the element to always be a `gu-*` custom element
```
SiAdcock, line 4: For future reference (and for reviewers in a hurry!) what is the role and significance of the `gu-*` custom element?

oliverlloyd: They're [cooler](https://github.com/guardian/dotcom-rendering/pull/445#pullrequestreview-209189589) 😉 

Maybe they also provide a nice separation from standard elements as a way to highlight these markers as not being part of the semantic dom structure?

SiAdcock: Okay thanks for clarifying that, I agree it flags up the hydratable markup better than a class. I'm wondering if the naming scales well with some possible future in which many of our components are custom elements. Do you think we should choose a name that implies these elements need to be hydrated on the client?

oliverlloyd: > I'm wondering if the naming scales well with some possible future in which many of our components are custom elements

Good point. It's worth pointing out I had in mind using `gu-hydrate` + `gu-portal` at some point soon, hence the asterisk

dotcom-rendering/src/web/browser/hydration/doHydration.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @param data The deserialised props we want to use for hydration
+ * @param marker The location on the DOM where the component to hydrate exists
+ */
+export const doHydration = (name: string, data: any, marker: HTMLElement) => {
+	const { start, end } = initPerf(`hydrate-${name}`);
+	start();
+	window
+		.guardianPolyfilledImport(
+			/* webpackInclude: /\.importable\.(tsx|jsx)$/ */
+			`../../components/${name}.importable`,
```
OllysCoding, line 19: I'd still like to find an improved way to do this step that doesn't depend on a static path to import components

dotcom-rendering/src/web/browser/hydration/whenIdle.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * whenIdle exectures the given callback when the browser is 'idle'
+ *
+ * @param callback Fired when requestIdleCallback runs or after 300ms
+ */
+export const whenIdle = (callback: () => void) => {
+	if ('requestIdleCallback' in window) {
+		window.requestIdleCallback(callback);
```
arelra, line 8: 1. You can pass a `timeout` option to `requestIdleCallback`
2. Does `requestIdleCallback` get poly-filled as it's not supported by Safari or IE.

https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback

oliverlloyd: > You can pass a timeout option to requestIdleCallback

Do we want to timeout? It's a bit hand wavy but isn't waiting for idle the contract here? Adding a timeout feels like we'd be overriding the intention.


> Does requestIdleCallback get poly-filled as it's not supported by Safari or IE.

I guess the 'polyfill' here is we fallback to 300 milliseconds. I can investigate if there's an actual polyfill 👍

dotcom-rendering/src/web/browser/hydration/doHydration.ts
oliverlloyd: I'd _love_ to find a route around this 🙏

dotcom-rendering/src/web/browser/hydration/doHydration.ts
```diff
@@ -0,0 +1,33 @@
...
+		)
+		.then((module) => {
+			hydrate(h(module[name], data), marker);
```
arelra, line 22: This is all a very nice step forwards!

I assume preact will take care of the component lifecycle and any hooks defined server side would run as normal?

dotcom-rendering/src/web/browser/hydration/getName.ts
arelra: _Very slight_ niggle that this is overloading custom elements syntax when they are not _really_ custom elements but rehydrated Preact components (iiuc).

But it is cooler.

dotcom-rendering/src/web/browser/hydration/whenIdle.ts
arelra: > Do we want to timeout?

The docs recommend a timeout:

> A timeout option is strongly recommended for required work, as otherwise it's possible multiple seconds will elapse before the callback is fired.

You're right though, the contract is in the name :)

I just wonder if consumer expectations will be sub-second when it can in theory take an unknown amount of time for the callback to hit the event loop.

dotcom-rendering/src/web/browser/hydration/whenIdle.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * whenIdle exectures the given callback when the browser is 'idle'
+ *
+ * @param callback Fired when requestIdleCallback runs or after 300ms
```
arelra, line 4:
```suggestion
 * @param callback Fired when requestIdleCallback runs. If requestIdleCallback is not available after 300ms
```

dotcom-rendering/src/web/browser/hydration/whenIdle.ts
oliverlloyd: Interesting. I think the key is in the word 'required'. We can't know for sure but we probably _do_ want to think of all hydration as required, even if the original intent was to delay it in favour of other, more important, work.

So yes, let's add a timeout. How long do you think?

dotcom-rendering/src/web/browser/hydration/doHydration.ts
oliverlloyd: I need to check this. I am drawn to using preact directly like this because I know that DCR alias react > preact. But it's worth considering what happens in the event that were no longer the case.

dotcom-rendering/src/web/browser/hydration/whenIdle.ts
arelra: There are so many factors (device, network etc) that could influence a suitable value for this.

Keeping it simple and not wanting to block a busy main thread too early:

Going by Speedcurve [synthetics](https://app.speedcurve.com/guardian/guardian/vitals-syn/?b=chrome&cs=md&d=30&de=1&ds=1&label=article&s=82463) (TBT caveats aside) and [RAIL](https://web.dev/rail/)... 500ms or thereabouts to start with 🤔 ?

arelra: There's this (gold-plated) pattern too: [idle-until-urgent](https://philipwalton.com/articles/idle-until-urgent/)

oliverlloyd: Thanks for that link, this was a really interesting read! I'd love to spend more time on this to see if there are more places we can use this pattern. It'd be amazing to extend the `whenIdle` abstraction to return an override and then we could use the same thing for libs like Ophan or CMP where we want to defer initialisation but trigger it if there's a sudden requirement. I guess essentially, you'd be setting up a situation where you're deferring initialising things until the point where they're needed.

arelra: The really interesting part of that article was the splitting up of actions _within_ components (e.g. initialisation, creating event listeners, state creation) into atomic 'tasks' that could put on a queue yet still be resolved when required.

It feels partly similar to what React Concurrent mode is attempting (albeit via internals) i.e. interruptible rendering that can be sliced up so that the main thread stays responsive.

Back in our world we have two goals:

- lazy loading of the component
- trigger component render when required

Can we achieve most of this by using `Suspense`?

aiui Suspense works by catching throws or a return value from what they call a `Resource` object:
- a thrown promise denotes the children are still 'suspended'
- a thrown error should be handled by an `ErrorBoundary`
- a returned value denotes successful completion and the component is un-suspended.

But we could in theory use any 'event' (Observables even!) to resolve the promise and trigger completion of the enclosing `Suspense`.

So you could so something like:

```js
<Suspense fallback={<Overlay />}>
  <YouTubePlayer />
</Suspense>
```

But the important bit is the `Resource` which does an async `suspendUntilConsent.read()` in `YouTubePlayer` e.g.:

```js
const YouTubePlayer = () => {
  const consentState = suspendUntilConsent.read();
  return (</>)
}
```

I've created a poc to demonstrate:

https://codesandbox.io/s/zen-tereshkova-c1nsh

The main point of interest is the player which has the suspended loader:

https://codesandbox.io/s/zen-tereshkova-c1nsh?file=/src/YouTubePlayer.jsx

There's a probably a nice HOC that could be written to provide an API like:

```js
<SuspendUntil fallback={<Overlay />} until={onConsentChangePromise()}>
  <YouTubePlayer />
</SuspensdUntil>
```

Another pattern to add to the mix!

mxdvl: [Here’s the polyfill](https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/requestIdleCallback/polyfill.js), by the way.

oliverlloyd: > [Here’s the polyfill](https://github.com/Financial-Times/polyfill-library/blob/master/polyfills/requestIdleCallback/polyfill.js), by the way.

That's a lot of polyfill. Is it worth it? Genuine question. 

Polyfill cons:
- more code
- extra complexity (risk of issues)

Polyfill pros:
- The callback will fire sooner if the thread is idle before 300ms
- This function becomes simpler

OllysCoding (approved):

gtrufitt:
👏

jamesgorrie:
Reference #3784

mxdvl:
Published on the Engineering Blog: https://www.theguardian.com/info/2022/mar/25/react-islands-on-theguardiancom

PR #3630, opened 2021-11-12
Refactor props for `LeftColumn`

oliverlloyd:
## What does this change?
Here we tidy up the props for `LeftColumn`, ensure that we're setting `size` correctly in each use case and then use this component in all the places where we are building a left column

## Why?
We had more than one way to decide the width of the left column and not all of them were taking into account the fact blogs now have a wider column.

### Before
Screenshot 2021-11-12 at 17 18 00

### After
Screenshot 2021-11-12 at 17 06 22


oliverlloyd:
[LeftColumn supports wider widths](https://trello.com/c/p5ZDzsCP/379-leftcolumn-supports-wider-widths)

dotcom-rendering/src/web/components/LeftColumn.tsx
```diff
@@ -64,41 +64,34 @@ const partialRightBorder = (colour: string) => css`
...
-	showRightBorder?: boolean;
-	showPartialRightBorder?: boolean;
+	borderType?: 'full' | 'partial';
 	borderColour?: string;
 	size?: LeftColSize;
 };
 
 export const LeftColumn = ({
 	children,
-	showRightBorder = true,
```
DavidLawes, line 81: I noticed there were a number of chromatic diffs where the right border is now missing. We might want to consider setting a default value for `borderType` too? E.g. ArticleHeadline uses left column like:

```
			<Flex>
				<LeftColumn>
					<></>
				</LeftColumn>
```

Previously, if neither `showRightBorder` or `showPartialRightBorder` were defined then I think we defaulted to showing the full right border:

```
If props not defined:
showRightBorder = true
showPartialRightBorder = false
shouldShowPartialBoard = false

--> border styling applied to <section>
```

At the moment, if `borderType` is not defined, I think we get no right border at all, so the default behaviour has changed (which is why we're seeing diffs for things like ArticleHeadline)? Or I'm confusing myself (or this change is intended)!

DavidLawes (commented):
I believe this PR will help solve a related issue: https://trello.com/c/y29ArEvq

dotcom-rendering/src/web/components/LeftColumn.tsx
oliverlloyd: Hmm, it was confusing to understand the default before because there were two booleans... but you're absolutely right, we should keep to that api otherwise we risk subtle regressions. I'll have a look at this

DavidLawes:
Note that I think we need an additional card raised to deal with:
- LiveLayout
- Related stories container
- Unintentional additional padding

oliverlloyd (commented):
If I were able to approve my own PR, I would

rcrphillips (approved):
LGTM

PR #3631, opened 2021-11-12
Fixing the display of the contents block at narrow breakpoints

rcrphillips:
## What does this change?
Updates the dropdown menu for the contents block when there are a lot of menu items. Previously, the menu did not scroll properly and it was not possible to see all of the items in the menu

## Why?
So that we can enable more contents block articles to be rendered in DCR

### Before

Nov-12-2021 22-10-26
### After
Nov-12-2021 22-10-41


rcrphillips:
@HarryFischer Take a look at this and see if it looks ok

DavidLawes (approved):
👍

PR #3632, opened 2021-11-15
`buildAdTargeting` no longer accepts `CAPI`

oliverlloyd:
## What does this change?
This refactors the props for `buildAdTargeting` so that it no longer accepts `CAPI` and now accepts the specific props it needs

## Why?
Because we want to remove dependencies on CAPI, its a large object and having it as a contract is expensive.

### Before
```ts
type Props = {
    CAPI:
		| CAPIType
		| CAPIBrowserType
		| Pick<CAPIType, 'isAdFreeUser' | 'config'>,
}
```

### After
```ts
type Props = {
	isAdFreeUser: boolean;
	isSensitive: boolean;
	edition: string;
	section: string;
	sharedAdTargeting: Record<string, unknown>;
	adUnit: string;
	videoDuration?: number;
}
```

### DIFF
--- a/dotcom-rendering/src/lib/ad-targeting.test.ts
+++ b/dotcom-rendering/src/lib/ad-targeting.test.ts
@@ -1,63 +1,42 @@
-import { Article } from '@root/fixtures/generated/articles/Article';
 import { buildAdTargeting } from './ad-targeting';
 
-const CAPI = {
-	...Article,
-	config: {
-		...Article.config,
-		adUnit: '/59666047/theguardian.com/money/article/ng',
-		edition: 'UK',
-		section: 'money',
-		sharedAdTargeting: {
-			ct: 'article',
-			co: ['rob-davies'],
-			url: '/money/2017/mar/10/ministers-to-criminalise-use-of-ticket-tout-harvesting-software',
-			su: ['0'],
-			edition: 'uk',
-			tn: ['news'],
-			p: 'ng',
-			k: [
-				'ticket-prices',
-				'consumer-affairs',
-				'internet',
-				'viagogo',
-				'money',
-			],
-			sh: 'https://theguardian.com/p/64ak8',
-		},
-	},
+const sharedAdTargeting = {
+	co: ['rob-davies'],
+	ct: 'article',
+	edition: 'uk',
+	k: ['ticket-prices', 'consumer-affairs', 'internet', 'viagogo', 'money'],
+	p: 'ng',
+	sh: 'https://theguardian.com/p/64ak8',
+	su: ['0'],
+	tn: ['news'],
+	url: '/money/2017/mar/10/ministers-to-criminalise-use-of-ticket-tout-harvesting-software',
 };
 
 describe('buildAdTargeting', () => {
 	const expectedAdTargeting = {
 		adUnit: '/59666047/theguardian.com/money/article/ng',
 		customParams: {
-			cc: 'UK',
-			co: ['rob-davies'],
-			ct: 'article',
-			edition: 'uk',
-			inskin: 'f',
-			k: [
-				'ticket-prices',
-				'consumer-affairs',
-				'internet',
-				'viagogo',
-				'money',
-			],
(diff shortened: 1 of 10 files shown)


DavidLawes (approved):
Nice, I think it's easier to read now we're passing in only the required props.

It's not a functional change but should we float it past comm dev (I'm assuming they're the owners/authors of the function)?

oliverlloyd:
Hi @guardian/commercial-dev this should be a straight forward change but pinging you so you're aware

dotcom-rendering/src/lib/ad-targeting.ts
```diff
@@ -1,27 +1,37 @@
...
-		vl: config.videoDuration || 0,
-		cc: config.edition,
-		s: config.section,
+		vl: videoDuration || 0,
+		cc: edition,
+		s: section,
 		inskin: 'f',
-		...config.sharedAdTargeting,
+		...sharedAdTargeting,
 		pa: 'f',
```
mxdvl, line 31: Looking it this makes me raise a couple of eyebrows, but I appreciate that this PR doesn’t really change the functionality of it all.

Note to self: Are we saying the user is never signed in (`si`)? No personalised advertising (`pa`) ever? Video length (`vl`) should always ever be a multiple of 30… and a string.

dotcom-rendering/src/lib/ad-targeting.test.ts
```diff
@@ -1,63 +1,42 @@
...
-			pa: 'f',
-			s: 'money',
-			sh: 'https://theguardian.com/p/64ak8',
-			su: ['0'],
-			tn: ['news'],
-			url: '/money/2017/mar/10/ministers-to-criminalise-use-of-ticket-tout-harvesting-software',
 			sens: 'f',
 			si: 'f',
 			vl: 0,
+			cc: 'UK',
```
mxdvl, line 22: 'UK' is not a valid country code. 'GB' is.

mxdvl (commented):
It’s really good to see a more _targeted_ approach to targeting.

We’re also adding much stricter types to these key-value pairs ourselves: https://github.com/guardian/commercial-core/pull/421/files

dotcom-rendering/src/lib/ad-targeting.ts
arelra: Signed in is false (and I assume other defaults) because of SSR.
We can't determine that server-side without cache-splitting.
For `YoutubeAtom` `si` gets overridden client-side.

dotcom-rendering/src/lib/ad-targeting.test.ts
zekehuntergreen: It looks from the [CAPI documentation](https://open-platform.theguardian.com/documentation/edition) like `edition`s / edition `id`s are not ISO country codes.
The response from the [/editions endpoint](https://content.guardianapis.com//editions?api-key=test) suggests that any of ('UK', 'US', 'AU', 'International') would be valid values.

mxdvl: Oh wow… [`cc` in AdManager][cc] does not have predefined value (dynamic) and isn’t used in reporting, so I guess we’re fine? Ideally, we’d pass in the real country code. Worth checking whether it’s not a deprecated targeting, too.

[cc]: https://admanager.google.com/59666047#inventory/custom_targeting/detail/custom_key_id=11703293

zekehuntergreen: Good call, I hadn't seen the lack of targeting values in GAM! It does look like `cc` may no longer be in use.

oliverlloyd:
@mxdvl @arelra I presume you're okay to merge this? Really interesting points though and worth following up on as part of the wider work to type commercial data more

mxdvl (approved):
@oliverlloyd here’s your green tick from us 😉 

Thanks for pinging. It should make our lives easier when implementing the upcoming ad targeting changes.

PR #3635, opened 2021-11-15
More explicit props for the SignIn Gate

oliverlloyd:
**👉 Please review commit by commit for a a more enjoyable experience 👈**

## What does this change?
Removes the use of `CAPI` as a prop from `SignInGateSelector`, `useSignInGateWillShow` and `StickyBottomBanner`, replaces it with explicit properties

## Why?
`CAPI` is 30Kb and having it as a prop creates a contract the the platform needs to support. We'd like to refactor `App.tsx` but we need to better understand our contracts first

### DIFF
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -1295,6 +1295,7 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 					isPaidContent={CAPI.pageType.isPaidContent}
 					isPreview={!!CAPI.isPreview}
 					host={CAPI.config.host}
+					pageId={CAPI.pageId}
 				/>
 			</Portal>
 			<HydrateOnce rootId="comments" waitFor={[user]}>
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -1290,7 +1290,7 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 					isSignedIn={isSignedIn}
 					format={format}
 					contentType={CAPI.contentType}
-					sectionName={CAPI.sectionName || ''}
+					sectionName={CAPI.sectionName}
 					tags={CAPI.tags}
 					isPaidContent={CAPI.pageType.isPaidContent}
 					isPreview={!!CAPI.isPreview}
@@ -1349,7 +1349,7 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 					brazeMessages={brazeMessages}
 					asyncArticleCount={asyncArticleCount}
 					contentType={CAPI.contentType}
-					sectionName={CAPI.sectionName || ''}
+					sectionName={CAPI.sectionName}
 					tags={CAPI.tags}
 					isPaidContent={CAPI.pageType.isPaidContent}
 					isPreview={!!CAPI.isPreview}
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -1297,6 +1297,7 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 					host={CAPI.config.host}
 					pageId={CAPI.pageId}
 					idUrl={CAPI.config.idUrl}
+					pageViewId={pageViewId}
 				/>
 			</Portal>
 			<HydrateOnce rootId="comments" waitFor={[user]}>
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -1293,6 +1293,7 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 					sectionName={CAPI.sectionName}
 					tags={CAPI.tags}
 					isPaidContent={CAPI.pageType.isPaidContent}
+					isPreview={!!CAPI.isPreview}
 				/>
 			</Portal>
 			<HydrateOnce rootId="comments" waitFor={[user]}>
@@ -1343,12 +1344,12 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 					asyncCountryCode={asyncCountryCode}
 					CAPI={CAPI}
 					brazeMessages={brazeMessages}
-					isPreview={!!CAPI.isPreview}
 					asyncArticleCount={asyncArticleCount}
 					contentType={CAPI.contentType}
 					sectionName={CAPI.sectionName}
 					tags={CAPI.tags}
 					isPaidContent={CAPI.pageType.isPaidContent}
+					isPreview={!!CAPI.isPreview}
 				/>
 			</Portal>
 		</React.StrictMode>
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -1292,6 +1292,7 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 					contentType={CAPI.contentType}
 					sectionName={CAPI.sectionName}
 					tags={CAPI.tags}
+					isPaidContent={CAPI.pageType.isPaidContent}
 				/>
 			</Portal>
 			<HydrateOnce rootId="comments" waitFor={[user]}>
@@ -1347,6 +1348,7 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 					contentType={CAPI.contentType}
 					sectionName={CAPI.sectionName}
 					tags={CAPI.tags}
+					isPaidContent={CAPI.pageType.isPaidContent}
 				/>
 			</Portal>
 		</React.StrictMode>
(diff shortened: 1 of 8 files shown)


dotcom-rendering/src/web/components/App.tsx
```diff
@@ -1286,7 +1286,18 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 				</Lazy>
 			</Portal>
 			<Portal rootId="sign-in-gate">
-				<SignInGateSelector isSignedIn={isSignedIn} CAPI={CAPI} />
+				<SignInGateSelector
+					isSignedIn={isSignedIn}
+					format={format}
+					contentType={CAPI.contentType}
+					sectionName={CAPI.sectionName || ''}
```
tjmw, line 1293: Elsewhere in this PR (for `host`) we've switched to specifying the default _inside_ the component, rather than at the call site. Do we want to do the same here for consistency?

dotcom-rendering/src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -152,27 +170,51 @@ export const SignInGateSelector = ({
 		if (gateVariant && currentTest) {
 			// eslint-disable-next-line @typescript-eslint/no-floating-promises
 			gateVariant
-				?.canShow(CAPI, !!isSignedIn, currentTest)
+				?.canShow({
+					isSignedIn: !!isSignedIn,
```
tjmw, line 174: Probably out of scope for this PR, but seeing this made me wonder - do we prefer `!!` vs `Boolean()`? I kind of like how explicit and obvious `Boolean()` is. Maybe something for the JS/TS recommendations.

dotcom-rendering/src/web/components/SignInGate/SignInGateSelector.tsx
```diff
@@ -152,27 +170,51 @@ export const SignInGateSelector = ({
...
 
 	if (!currentTest || !gateVariant) {
 		return null;
 	}
 
-	const signInUrl = generateSignInUrl(CAPI, currentTest);
+	const signInUrl = generateSignInUrl({
+		pageId,
+		host,
+		pageViewId: window.guardian.config.ophan.pageViewId,
```
tomrf1, line 202: is it worth passing this in from App.tsx as well? It's already defined there

dotcom-rendering/src/web/components/SignInGate/displayRule.ts
```diff
@@ -127,26 +121,44 @@ export const canShowSignInGate = (
...
-			isValidSection(CAPI) &&
-			isValidTag(CAPI) &&
-			!isPaidContent(CAPI) &&
-			!isPreview(CAPI) &&
+			isValidContentType(contentType) &&
+			isValidSection(sectionName) &&
+			isValidTag(tags) &&
+			// hide the sign in gate on isPaidContent
+			!isPaidContent &&
+			// hide the sign in gate on internal tools preview &&
```
tjmw, line 129: Is the trailing `&&` in this comment meant to be there? `</nitpick>`

dotcom-rendering/src/web/components/SignInGate/displayRule.ts
```diff
@@ -33,18 +33,16 @@ export const isIOS9 = (): boolean => {
...
 
-	return validTypes.some(
-		(type: string): boolean => CAPI.contentType === type,
-	);
+	return validTypes.some((type: string): boolean => contentType === type);
 };
 
 // hide the sign in gate on certain sections of the site, e.g info, about, help etc.
-export const isValidSection = (CAPI: CAPIBrowserType): boolean => {
+export const isValidSection = (sectionName?: string): boolean => {
```
tomrf1, line 45: I appreciate this is existing logic, but I'm pretty sure we should be using `section` not `sectionName`.
E.g. 
Screen Shot 2021-11-16 at 09 38 13

tjmw (approved):
👍  This looks good to me! I left a few really minor comments/suggestions to take or leave.

This definitely makes it clearer which attributes of the `CAPI` object are used where. I wonder if after the process of expanding out the attributes of the `CAPI` object like this in a few places patterns will emerge and we'll end up grouping attributes back together in some way 🤔

dotcom-rendering/src/web/components/SignInGate/displayRule.ts
tomrf1: I've been migrating to `section` for epics - https://github.com/guardian/dotcom-rendering/pull/3623
This was clearly a mistake

oliverlloyd: Happy to change this! Always keen to leave things in a better state

dotcom-rendering/src/web/components/SignInGate/SignInGateSelector.tsx
oliverlloyd: Good idea! I'll do this 👍

dotcom-rendering/src/web/components/App.tsx
oliverlloyd: Great spot! Yes, I think it makes more sense to have all the defaults in one place

dotcom-rendering/src/web/components/SignInGate/SignInGateSelector.tsx
oliverlloyd: Oh interesting. I personally didn't know what `Boolean()` was, or at least not enough to be confident using it. Clearly though if you put them side by side `Boolean` makes more sense that two exclamation marks!!

I doubt there's one good option here. The thing that makes most sense to you is based on your developer history which is different for everyone. I agree though that this should be centralised and a dept. level decision made so that we can move on from stuff like this. I love how Prettier removed these issues for formatting, so maybe we can setup something similar for this?

dotcom-rendering/src/web/components/SignInGate/displayRule.ts
oliverlloyd: 👀 Ha, that's odd. I have no idea how that slipped in!

oliverlloyd:
[Remove CAPI prop from sign in gate code](https://trello.com/c/1jZZBEiv/413-remove-capi-prop-from-sign-in-gate-code)

dotcom-rendering/src/web/components/SignInGate/displayRule.ts
oliverlloyd: Hmm, so this isn't as simple as you'd think. We don't have `section` as a typed property in DCR, only `sectionName`. We can add it to the type but that needs investigation and possibly some work further up the stack so might be better as a separate PR.

Note. It _does_ exist on the config object but it's untyped. It's sort of there by chance but can't be depended on.

oliverlloyd: Ignore my last, it _is_ typed on `config`. I've also made a note to look at improving our type to include `sectionId`

oliverlloyd: lol, so it is _typed_ but isn't actually there. @tjmw pointed out to me that `section` is actually available on `window.guardian.config.page.section`

Screenshot 2021-11-16 at 12 32 01

So. In short, we need to revisit this type and work out how we're going to do things. Which I think should happen in a separate PR.

I suspect this also affect https://github.com/guardian/dotcom-rendering/pull/3623? @tomrf1

tomrf1: Well it worked when I tested it! Not sure how it gets into `CAPI.config.section`, but it's there

oliverlloyd: That's a worry. It could be it is being set on some articles and not others? Or it might be it's available on the CAPI server object but isn't passed into the client object?

tomrf1: It's apparently coming from here: https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/index.d.ts#L745
How does this get constructed?

tomrf1: The data in `window.guardian.config.page` seems to correspond to that model, `ConfigTypeBrowser`, so I'm not sure what the issue is?

oliverlloyd: I had a dig into this and I think I see what's going on 🕵️ 

So the value we need, `section` (or `sectionId`) is - [according to our types](https://github.com/guardian/dotcom-rendering/blob/150fc2d81e6a66d9c3336185e874fc8cd0288546/dotcom-rendering/src/model/window-guardian.ts#L161) - available on `window.guardian.config.section`

But in reality is actually stored on `window.guardian.config.page.section` which is not typed anywhere.

Screenshot 2021-11-16 at 14 51 24

The reason this data is appearing on `window.guardian.config` without being typed is this code

Screenshot 2021-11-16 at 14 48 46

We extend the object sent to us from CAPI, without typing it 😐 

Why? Because we have these contracts with the commercial code that expects certain fields on the `window.config` object. We're not sure exactly what these contracts are so we pass everything through.

The commercial team are working on this. Once we have this contract better defined we can dramatically reduce the size of `window.config`. But until then the `section` value is out in the wild west!

PR #3636, opened 2021-11-15
Lift Editorial Palette

JamieB-gu:
## Why?

Basic colour rules should be the same between DCR and AR because the designs are the same. In order to capture these rules in a single place we've discussed moving the palette decision code up into `common-rendering`. This PR begins the process by lifting AR's `editorialPalette` up. I've also simplified the naming conventions to be more in line with DCR (i.e. removed "primary" and "inverse").

Currently this is based on #3634 to avoid conflicts with the palette changes there.

**Note:** There was a version of `editorialPalette` already in `common-rendering` from when `image-rendering` was migrated across. This PR merges the two together (they were structurally the same already).

## Changes

- Moved AR's `editorialPalette` to `common-rendering`
- Merged with `common-rendering`'s `editorialPalette`
- Removed terms "primary" and "inverse"
- Renamed `border.primary` to `border.articleLink` to match DCR
- Updated imports


marjisound (approved):
It look good Jamie, I didn't go through all of the colors  in `editorialPalette` as it would take ages but when I diffed it with the removed `editorialPalette` the changes look all fine

PR #3638, opened 2021-11-15
Make sure strong elements return a bold font weight

frankie297:
## Why?
A side effect of the recent CSS reset introduced has resulted in <strong> tags not being styled with font-weight: bold. This means that Racing Tips, Sports articles, Letters etc have lost their “boldness”. This PR fixes that.

### DIFF
--- a/apps-rendering/src/renderer.ts
+++ b/apps-rendering/src/renderer.ts
@@ -178,7 +178,11 @@ const plainTextElement = (node: Node, key: number): ReactNode => {
 		case 'BLOCKQUOTE':
 			return h('blockquote', { key }, children);
 		case 'STRONG':
-			return h('strong', { key }, children);
+			return styledH(
+				'strong',
+				{ css: { fontWeight: 'bold' }, key },
+				children,
+			);
 		case 'EM':
 			return h('em', { key }, children);
 		case 'BR':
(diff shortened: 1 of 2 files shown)


groakland (approved):
Lint nit aside, this looks good to go. Editions Production team can test this in code.preview and code.proofed and will verify all is in order

groakland (commented):
Editions Production team have validated the fix and confirm that the text formatting is correct. This is the vital piece of the fix so would like this shipped ASAP

apps-rendering/src/renderer.ts
```diff
@@ -223,7 +225,7 @@ const textElement =
 						key,
 						isEditions,
 					},
-					transform(text, format),
+					children,
```
marjisound, line 228: I'm trying to understand the reason the text of the anchor doesn't need to go through transform any more? If the content has `•` or `* * *` do we not need the styling of the Bullet & HorizontalRule?

frankie297: @JamieB-gu  can probably explain it better than me, but we can get rid of the transform as I understand it, transform is evaluated in the `#text` element further up: https://github.com/guardian/dotcom-rendering/blob/df0be1b184dc51315b7146bdfbf4a6da34682927/apps-rendering/src/renderer.ts#L215-L216 so any styles added there will be inherited via the children in anchor.

marjisound: Oh I see, the content of anchor would go through recursion and end up in `#text` case which is then applied to transform function. Makes sense, thanks for explaining it :)

marjisound (approved):

apps-rendering/src/renderer.ts
marjisound: Oh I see, the content of anchor would go through recursion and end up in #text case which is then applied to transform function. Makes sense, thanks for explaining it :)

frankie297: Correct yes :)

PR #3641, opened 2021-11-15
Picture/Image Optimisations - The DPR Problem (Part 1/?)

OllysCoding:
## Summary

On DCR, there is a bug with how images are rendered which means high resolution displays can often fetch excessively-large variants of images. This means increased cost in bandwidth for both our users and our fastly bill.

This PR replicates the functionality of this [Frontend PR](https://github.com/guardian/frontend/pull/11400) to make use per-breakpoint source elements with media attributes to prevent the browser from selecting larger image sources than we intend.

Some related & interesting background PR's & Articles:
1. [This PR's Frontend Equivalent](https://github.com/guardian/frontend/pull/11400)
2. [Whatwg html spec discussion](https://github.com/whatwg/html/issues/4421) (And [the comment](https://github.com/whatwg/html/issues/4421#issuecomment-788140858) which mentions our solution!)
3. [Twitter engineering blog post](https://blog.twitter.com/engineering/en_us/topics/infrastructure/2019/capping-image-fidelity-on-ultra-high-resolution-devices) about a similar change they made!

## Background Info

[Documentation on how picture tags work & more](https://github.com/guardian/dotcom-rendering/blob/olly/picture-optimisations/dotcom-rendering/docs/architecture/027-pictures.md) - A great place to start if this is new to you!  (Added in this PR)

### What does DCR currently do for images?

DCR renders images in `<picture>` tags, with each picture element containing two `<source>` elements. These source elements differentiate between high & low DPR/DPI (device pixel ratios).

We use the following media query on the first of our two `<source>` elements to tell the browser, "Hey, if you're a 'high DPR device', use this source element":
`<source media"(-webkit-min-device-pixel-ratio: 1.25), (min-resolution: 120dpi)" >`

This lets us do something different for these "High DPR" devices, often on these super high resolution screens, getting an equivalently high resolution image is unnecessary - the difference is barely perceptible to the human eye.

For non-high-dpr image sources we provide the quality attribute to fastly as `?quality=85`, however for these high-dpr image sources we have, we tell fastly `?quality=45&dpr=2`. 

These high-dpr image sources have, as we learned earlier, double the resolution of the regular image sources, however by providing a lower quality number, they're more compressed. At some point in time someone at The Guardian sat and looked at images side-by-side and decided this config was the best for high resolution devices.

All in all this leads to a DCR picture element looking something like:

```html
<picture>
    <source media="(-webkit-min-device-pixel-ratio: 1.25), (min-resolution: 120dpi)" sizes="(min-width: 660px) 620px, 100vw" srcset="https://fastly.link/path?width=300&quality=45&dpr=2 600w, https://fastly.link/path?width=600&quality=45&dpr=2 1200w" >
    <source sizes="(min-width: 660px) 620px, 100vw" srcset="https://fastly.link/path?width=300&quality=85 300w, https://fastly.link/path?width=600&quality=85 600w" >
    <img src="https://fastly.link/to/fallback/for/unsupported/browsers" > 
</picture>
```

## The DPR Problem

Everything up until now is context & background information to make sure anyone who reads this PR can understand what the changes made actually do. Now let's talk about the problem, why it happens & how we have solved it!

This problem comes from how the browser tries to compensate for high DPR displays when choosing an image source. As described above, we provide a set of sources for high DPR displays, and target them with a media query to ensure it's picked.

Unfortunately the browser itself tries to compensate for high DPR as well, but in a less efficient way. Once the browser has figured out what size source it wants from the `sizes` attribute on our source element, it's then multiplied by the DPR of the display to get the desired width it will look for in our `srcset`:

```
# desiredWidth is the width of an image that the browser will look for in `srcset`
# size is the chosen size based off our queries in the `sizes` attribute
# DPR is the ration between CSS Pixels & device resolution (e.g 2, 3, 4)
desiredWidth = size * DPR
```

This poses a problem for us, because if we tell the browser "Hey choose a 620px image", and the browser has a DPR of say 3, it will actually look for an image source for 1860px, far higher resolution than is needed for the user to have a good experience.

### How do we solve this?

Lucky for us, this problem has already been solved on Frontend, and its existence is currently a regression on DCR.

So what does Frontend do? As shown in [this PR](https://github.com/guardian/frontend/pull/11400), rather than having just 2 source elements, we instead have 2 source elements per breakpoint (one for high DPR, one for regular displays), and provide just one size & source set for that source element.

Lets look at a simplified example (with only 1 source per breakpoint):
```html
<picture>
    <source media="min-width: 980px" srcset="https://xxx.png?width=620px 620w" />
    <source media="min-width: 660px" srcset="https://xxx.png?width=620px 620w" />
    <source media="min-width: 480px" srcset="https://xxx.png?width=480px 480w" />
    <source media="min-width: 375px" srcset="https://xxx.png?width=420px 420w" />
</picture>
```

In this example, we have the logic that usually would have been in our sizes attribute (`sizes="(min-width: 660px) 620px, 100vw"`) extrapolated into individual source elements. All our breakpoint which are 660px or larger offer only 1 choice, the 620px source. The lower breakpoints have looked for sources which are closest to their own size, as a replacement for `100vw`.

This solves our DPR problem because, the `media` attribute uses CSS pixels, and because we only offer 1 image source for each of these elements, once the browser has picked its source element, we basically strong arm it into which source to use. Even if it multiplies the size it gets from `sizes`, it still has to use the source we give as it's the only available option.

### Implementation details

The primary change for this PR is the (previously called) `getSizes` function. This function takes various information about the image we're trying to render (role, layout, etc), and returns the contents for the `sizes` attribute. By refactoring this code, we can instead have it also accept a breakpoint, and have it perform the same logic as the queries it was producing to instead return an expected image width for that breakpoint:
```js
// Before
getSizes(...) -> "(min-width: 660px) 620px, 100vw"
// after
getDesiredWidthForBreakpoint(660, ...) -> 620
getDesiredWidthForBreakpoint(480, ...) -> 480
```

I've also introduced an optimisation function called `optimiseBreakpointSizes` which will look for redundant source's and remove them (for example if all your sources for breakpoint 660px and above return 620px, you only need 1). This should help avoid us overfilling the DOM with source elements.

### Final before and after
Before:
```html
<picture itemprop="contentUrl">
    <source srcset="https://i.guim.co.uk/img/media/file.jpg?width=620&amp;quality=45&amp;auto=format&amp;fit=max&amp;dpr=2&amp;s=xxx 1240w,https://i.guim.co.uk/img/media/file.jpg?width=700&amp;quality=45&amp;auto=format&amp;fit=max&amp;dpr=2&amp;s=xxx 1400w,https://i.guim.co.uk/img/media/file.jpg?width=620&amp;quality=45&amp;auto=format&amp;fit=max&amp;dpr=2&amp;s=xxx 1240w,https://i.guim.co.uk/img/media/file.jpg?width=645&amp;quality=45&amp;auto=format&amp;fit=max&amp;dpr=2&amp;s=xxx 1290w,https://i.guim.co.uk/img/media/file.jpg?width=465&amp;quality=45&amp;auto=format&amp;fit=max&amp;dpr=2&amp;s=xxx 930w" sizes="(min-width: 660px) 620px, 100vw" media="(-webkit-min-device-pixel-ratio: 1.25), (min-resolution: 120dpi)">
    <source srcset="https://i.guim.co.uk/img/media/file.jpg?width=620&amp;quality=85&amp;auto=format&amp;fit=max&amp;s=xxx 620w,https://i.guim.co.uk/img/media/file.jpg?width=700&amp;quality=85&amp;auto=format&amp;fit=max&amp;s=xxx 700w,https://i.guim.co.uk/img/media/file.jpg?width=620&amp;quality=85&amp;auto=format&amp;fit=max&amp;s=xxx 620w,https://i.guim.co.uk/img/media/file.jpg?width=645&amp;quality=85&amp;auto=format&amp;fit=max&amp;s=xxx 645w,https://i.guim.co.uk/img/media/file.jpg?width=465&amp;quality=85&amp;auto=format&amp;fit=max&amp;s=xxx 465w" sizes="(min-width: 660px) 620px, 100vw">
    <img alt="The Palace Theatre, London, showing Harry Potter and the Cursed Child" src="https://i.guim.co.uk/img/media/file.jpg?width=465&amp;quality=45&amp;auto=format&amp;fit=max&amp;dpr=2&amp;s=xxx" height="1200" width="2000" class="dcr-1989ovb">
</picture>
```

After:
```html
<picture>
    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=620&quality=45&auto=format&fit=max&dpr=2& 1240w" media="(min-width: 980px) and (-webkit-min-device-pixel-ratio: 1.25), (min-width: 980px) and (min-resolution: 120dpi)">
    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=620&quality=85&auto=format&fit=max& 620w" media="(min-width: 980px)">
    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=700&quality=45&auto=format&fit=max&dpr=2& 1400w" media="(min-width: 740px) and (-webkit-min-device-pixel-ratio: 1.25), (min-width: 740px) and (min-resolution: 120dpi)">
    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=700&quality=85&auto=format&fit=max& 700w" media="(min-width: 740px)">
    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=620&quality=45&auto=format&fit=max&dpr=2& 1240w" media="(min-width: 660px) and (-webkit-min-device-pixel-ratio: 1.25), (min-width: 660px) and (min-resolution: 120dpi)">
    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=620&quality=85&auto=format&fit=max& 620w" media="(min-width: 660px)">
    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=620&quality=45&auto=format&fit=max&dpr=2& 1240w" media="(min-width: 480px) and (-webkit-min-device-pixel-ratio: 1.25), (min-width: 480px) and (min-resolution: 120dpi)">
    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=620&quality=85&auto=format&fit=max& 620w" media="(min-width: 480px)">
    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=465&quality=45&auto=format&fit=max&dpr=2& 930w" media="(min-width: 375px) and (-webkit-min-device-pixel-ratio: 1.25), (min-width: 375px) and (min-resolution: 120dpi)">
    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=465&quality=85&auto=format&fit=max& 465w" media="(min-width: 375px)">
    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=465&quality=45&auto=format&fit=max&dpr=2& 930w" media="(min-width: 320px) and (-webkit-min-device-pixel-ratio: 1.25), (min-width: 320px) and (min-resolution: 120dpi)">
    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=465&quality=85&auto=format&fit=max& 465w" media="(min-width: 320px)">
    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=465&quality=45&auto=format&fit=max&dpr=2& 930w" media="(min-width: 0px) and (-webkit-min-device-pixel-ratio: 1.25), (min-width: 0px) and (min-resolution: 120dpi)">
    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=465&quality=85&auto=format&fit=max& 465w" media="(min-width: 0px)">
    <img alt="The Palace Theatre, London, showing Harry Potter and the Cursed Child" src="https://i.guim.co.uk/img/media/picture.jpg?width=465&quality=45&auto=format&fit=max&dpr=2&s=0492ab78e73c5167d8b4b841e601fbd4" height="1200" width="2000" class="dcr-b5pnrc-css">
</picture>
```

### Future Improvements

There are a few outstanding issues that can be solved with this solution, 

1. More effective immersive main media images by looping through sources individually rather than matching sources to breakpoints
2. Solve the DPR issue for immersive main media portrait as well (This also remains an issue in Frontend)
3. Investigate if it's more effective to generate sources within DCR rather than in Frontend

Once these PR's are raised they'll be linked to here.


dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -16,12 +16,13 @@ type Props = {
...
 type ResolutionType = 'hdpi' | 'mdpi';
 
-const getClosestSetForWidth = (
+export const getClosestSetForWidth = (
 	desiredWidth: number,
 	inlineSrcSets: SrcSetItem[],
 ): SrcSetItem => {
 	// For a desired width, find the SrcSetItem which is the closest match
-	const sorted = inlineSrcSets.sort((a, b) => b.width - a.width);
+	// A match greated than the desired width will always be picked when one is available
```
OllysCoding, line 24: greater*

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -16,12 +16,13 @@ type Props = {
...
+	const sorted = inlineSrcSets.slice().sort((a, b) => b.width - a.width);
```
OllysCoding, line 25: Adding `.slice()` to avoid sideeffects as `.sort` modifies the original array

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -46,63 +47,48 @@ const getSourcesForRoleAndResolution = (
...
-		// on mobile devices even though the viewport width is small, we'll need
-		// a larger image to maintain quality. To solve this problem we're using
-		// the viewport height (vh) to calculate width. The value of 167vh
-		// relates to an assumed image ratio of 5:3 which is equal to
-		// 167 (viewport height)  : 100 (viewport width).
-		return `(orientation: portrait) 167vh, 100vw`;
-	}
+): number => {
+	if (format.display === ArticleDisplay.Immersive && isMainMedia)
+		return breakpoint; // 100vh
```
OllysCoding, line 91: 100vw*

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -111,25 +97,71 @@ const getSizes = (
...
 	}
 
 	switch (role) {
 		case 'inline':
-			return `(min-width: ${breakpoints.phablet}px) 620px, 100vw`;
+			if (
+				breakpoint >= breakpoints.tablet &&
+				breakpoint < breakpoints.desktop
+			)
+				return 680;
```
OllysCoding, line 113: Looking at DCR, there is a breakpoint where the image width increases to aprx. 680px. This is brings DCR in parity with Frontend, which honoured this.

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -111,25 +97,71 @@ const getSizes = (
...
-			return '140px';
+			return 140;
 		case 'immersive':
-			return `(min-width: ${breakpoints.wide}px) 1300px, 100vw`;
+			if (breakpoint >= breakpoints.wide) return 1300;
+			return breakpoint; // 100vw
 		case 'supporting':
-			return `(min-width: ${breakpoints.wide}px) 380px, 300px`;
+			if (breakpoint >= breakpoints.wide) return 380;
+			return breakpoint; // 100vw
```
OllysCoding, line 126: I need to update this one to have 100vw, 300 & 380 variants

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -140,21 +172,67 @@ export const Picture = ({
...
+				<>
+					{portraitImmersiveSource(hdpiSourceSets, true)}
+					{portraitImmersiveSource(mdpiSourcesSets, false)}
+				</>
+			) : (
+				''
+			)}
+
+			{
+				// Loop through the breakpoints from highest to lowest, adding '0' as the last option
```
OllysCoding, line 209: I need to move the 'adding 0 as the last option' explanation to line 187.

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -111,25 +97,72 @@ const getSizes = (
...
+			return 140;
 		case 'immersive':
-			return `(min-width: ${breakpoints.wide}px) 1300px, 100vw`;
+			if (breakpoint >= breakpoints.wide) return 1300;
+			return breakpoint; // 100vw
 		case 'supporting':
-			return `(min-width: ${breakpoints.wide}px) 380px, 300px`;
+			if (breakpoint >= breakpoints.wide) return 380;
+			if (breakpoint >= breakpoints.tablet) return 300;
+			return breakpoint; // 100vw
```
OllysCoding, line 127: Once under the tablet breakpoint, this is 100vw instead of 300px as it was before

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -46,63 +47,48 @@ const getSourcesForRoleAndResolution = (
...
-	imageSources: ImageSource[],
-): string => {
-	// Get the sources for this role and resolution
-	const sources: SrcSetItem[] = getSourcesForRoleAndResolution(
-		imageSources,
-		role,
-		resolution,
-	);
+// The image sources we're provided use double-widths for HDPI images
+// We should therefor multiple the width based on if it's HDPI or not
```
oliverlloyd, line 58:
```suggestion
// We should therefor multiply the width based on if it's HDPI or not
```

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -140,21 +173,71 @@ export const Picture = ({
...
+							srcSet={getSourceSetForSize(
+								size,
+								hdpiSourceSets,
+								true,
+							)}
+							sizes={`${breakpoint || size}px`}
+							media={`(min-width: ${breakpoint}px) and (-webkit-min-device-pixel-ratio: 1.25), (min-width: ${breakpoint}px) and (min-resolution: 120dpi)`}
+						/>
+					) : (
+						''
```
oliverlloyd, line 226: `''`? Not sure if this is an issue or not but normally in this kind of situation you see `<></>`?

OllysCoding: Ah I'll definitely use that instead, I was looking for a similar situation to figure out what the best value to provide there was!

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -140,21 +173,71 @@ export const Picture = ({
...
+				<>
+					{/* HDPI Source (DPR2) - images in this srcset have `dpr=2&quality=45` in the url */}
+					{hdpiSourceSets.length ? (
+						<source
+							srcSet={getSourceSetForSize(
+								size,
+								hdpiSourceSets,
+								true,
+							)}
+							sizes={`${breakpoint || size}px`}
```
oliverlloyd, line 222: Is there a way to lift this conditional logic up so that `size` is always what you want? It feels very late to have a choice here.

I appreciate this is very 'feely' though and I've yet to fully build a model for the code

OllysCoding: I found this part a bit confusing when trying to replicate the functionality of Frontend.

On all regular breakpoints, the breakpoint is set as the 'sizes' attribute, however on the `0` breakpoint, Frontend would pass the lowest image size it had instead.

I don't actually think there's any functional difference, whatever the number is the browser is forced to pick an image, though I'd need to test whether it would be okay with receiving '0' as a size or not.

dotcom-rendering/src/web/components/Picture.tsx
idrise: such awareness of the evil of mutation...

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -46,63 +47,48 @@ const getSourcesForRoleAndResolution = (
...
+	breakpointSizes: [number, number][],
+): [number, number][] =>
+	breakpointSizes.filter(([, size], index) => {
+		// If the NEXT element in the array has the SAME size as this one, we can trust that this current breakpoint is redundant
+		// e.g
+		// [[980, 620], [660, 620]] > [[660, 620]]
+		// Because we use min-width, any source larger than 660 will get the 620px image, even without the min-width: 980 source element
+		if (
+			index + 1 < breakpointSizes.length &&
+			breakpointSizes[index + 1][1] === size
```
idrise, line 77: magic numbers?

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -111,25 +97,72 @@ const getSizes = (
 	) {
 		// Showcase main media images (which includes numbered list articles) appear
 		// larger than in body showcase images so we use a different set of image sizes
-		return `(min-width: ${breakpoints.wide}px) 1020px, (min-width: ${breakpoints.leftCol}px) 940px, (min-width: ${breakpoints.tablet}px) 700px, (min-width: ${breakpoints.phablet}px) 660px, 100vw`;
+		if (breakpoint >= breakpoints.wide) return 1020;
+		if (breakpoint >= breakpoints.leftCol) return 940;
+		if (breakpoint >= breakpoints.tablet) return 700;
```
idrise, line 102: magic numbers could these be named consts ?

idrise (approved):

dotcom-rendering/src/web/components/Picture.tsx
OllysCoding: Just example numbers really, they could be anything

dotcom-rendering/src/web/components/Picture.tsx
OllysCoding: I thought about this function a lot, on it's own it kind of acts as a source-of-truth for what sizes images should be based on their role, breakpoint, etc, which means the numbers mostly bare similarities with each other only by coincidence. I don't know if moving them into consts or another format would bring much benefit, but perhaps the comments on this function could be updated to help describe what the numbers actually mean?

dotcom-rendering/src/web/components/Picture.tsx
OllysCoding: Will update this comment for better clarity

mchv:
Excellent work writing that PR @OllysCoding  💯 🏅 

> These high-dpr image sources have, as we learned earlier, double the resolution of the regular image sources, however by providing a lower quality number, they're more compressed. At some point in time someone at The Guardian sat and looked at images side-by-side and decided this config was the best for high resolution devices.

Unless I am wrong that person is @paperboyo and when he is back may be able to provide some meaningful comments.

mchv:
@OllysCoding I have appended a commit adding a type alias for `Pixel` as I was unclear what unit return `getSizeForBreakpoint`, but the way it used at the end of the function seems to always been `px`.

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -46,63 +48,48 @@ const getSourcesForRoleAndResolution = (
...
-	const sources: SrcSetItem[] = getSourcesForRoleAndResolution(
-		imageSources,
-		role,
-		resolution,
-	);
+// The image sources we're provided use double-widths for HDPI images
+// We should therefor multiply the width based on if it's HDPI or not
+// e.g { url: '... ?width=500&dpr=2 ...', width: '1000' }
+export const getDesiredWidth = (width: number, hdpi: boolean) =>
+	hdpi ? width * 2 : width;
```
oliverlloyd, line 62: Personally, I'd inline this but author's choice.

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -46,63 +48,48 @@ const getSourcesForRoleAndResolution = (
...
 
-	return sources.map((srcSet) => `${srcSet.src} ${srcSet.width}w`).join(',');
-};
+// Turn SrsSets into a string which can be passed to the source element
+export const getSourcesFromSrcSets = (sources: SrcSetItem[]) =>
```
oliverlloyd, line 65: I'd either inline or rename this. Maybe something with fewer 's's

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -46,63 +48,48 @@ const getSourcesForRoleAndResolution = (
...
- *       mobileLandscape: 480
- *       phablet: 660
- *       tablet: 740
- *       desktop: 980
- *       leftCol: 1140
- *       wide: 1300
- */
-
-const getSizes = (
+export const optimiseBreakpointSizes = (
```
oliverlloyd, line 68: As discussed, it seems like this function is removing duplicates so perhaps the name can reflect that?

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -46,63 +48,48 @@ const getSourcesForRoleAndResolution = (
...
+		if (
+			index + 1 < breakpointSizes.length &&
+			breakpointSizes[index + 1][1] === size
+		)
+			return false;
+		return true;
+	});
+
+// Returns the size image desired at a particular breakpoint based on the image role, format, and if it's main media.
+const getSizeForBreakpoint = (
```
oliverlloyd, line 85: As discussed, there's a lot of use of 'size' here but maybe what we're working with is 'width'?

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -140,21 +174,71 @@ export const Picture = ({
...
+		role,
+		'mdpi',
+	);
+	const fallbackSrc = getFallback(
+		hdpiSourceSets.length ? hdpiSourceSets : mdpiSourceSets,
+	);
+
+	// Create an array of breakpoints going from highest to lowest, with 0 as the final option
+	const breakpointSizes: [number, number][] = [
+		...Object.values(breakpoints).reverse(),
```
oliverlloyd, line 193: We could replace this with an explicit array with each breakpoint declared inside. More verbose but the breakpoints are static and it will be easier to read

oliverlloyd (commented):
Awesome stuff!

Comments are more for my own notes as we're pairing on this

OllysCoding:
Thanks for your feedback @mchv! Myself and @oliverlloyd went back over the code to work on readability & maintainability, which I hope is now much better (there were far too many sizes, widths, sources and srcsets all over the place!).

JamieB-gu:
Excellent job in the PR description, very clear summary of a complex problem 😬!

Brief question - I don't claim to know the answer but I'm curious what you think. Currently I believe you're proposing to have a `<source>` tag for each breakpoint and each supported DPR? So number of `<source>` tags, `s`, is number of breakpoints supported, `b`, multiplied by number of DPRs supported, `d`:

```
s = b x d
```

In our case we have 8 breakpoints and 2 supported DPRs, so number of `<source>` tags is 16.

What do you think of the second option proposed in https://github.com/guardian/image-rendering/issues/239? In that scenario we provide a `<source>` tag per breakpoint, and use `srcset` to manage the DPR variations. That would reduce the number of `<source>` tags and drop the need to worry about the webkit (Safari) version of `min-resolution` as @mxdvl mentioned.

Again, I don't know if that's better, as you can see in that issue it's only one of two possible solutions (the other being the approach taken in this PR). But given that you've done a lot of research on this recently I'd be curious what you think 🙂.

Also I'm looking forward to getting all the excellent work in this PR available in common-rendering so AR can benefit too! 🎉

paperboyo:
> Excellent job in the PR description, very clear summary of a complex problem 😬!

Can’t agree enough! Great job, @OllysCoding!

I’m very interested in opinions on different approaches, as Jamie says, but don’t have an opinion myself. I didn’t know our frontend logic uses media queries not supported by Safari! I would only say that I think it’s beneficial for the URLs to explicitly spell out `dpr` argument instead of us just changing `width`. We may introduce more steps in the future (or only for some, eg. immersive/fluid, imagery) and I think this will help then and anyhow seeing `dpr` in URL is just clearer.

OllysCoding:
Thanks for the feedback @JamieB-gu!

> In our case we have 8 breakpoints and 2 supported DPRs, so number of <source> tags is 16.

I was aware while writing this code that it would result in a lot more <source> elements, which is why I added `removeRedundantWidths` (a feature the frontend implementation does not have), to weed out any source tags which aren't needed (e.g, if all image widths > desktop breakpoint are 620px wide for example, then we don't need the <source> elements for the larger breakpoints).

Worth noting that `removeRedundantWidths` would work with either the current implementation, or the one suggested in guardian/image-rendering#239

> What do you think of the second option proposed in guardian/image-rendering#239?

I think ultimately the differences between the two options are very slight.

The key tradeoff between the two is really, how declarative do you want to be at the cost of a larger DOM? The former (as implemented in this PR) gives us the choice of when to switch between high and low DPR variants, which could be beneficial if we want to look into if our choice of 1.25 is the most optimal. The second option entrusts the browser with this responsibility, and if there is room for optimisation there, we lose it - but we do get a slightly smaller DOM structure.

For me it makes sense to stick with the implementation which we use on Frontend, since it's known to work well, and I don't think there's a huge cost to the safari workaround we use. 

I do also want to add further optimisations to this solution in future PRs, so I'd be interested to revisit the two solutions once the full context of the work is complete, and see if there's a more obvious choice!

paperboyo:
> gives us the choice of when to switch between high and low DPR variants, which could be beneficial if we want to look into if our choice of 1.25 is the most optimal. The second option entrusts the browser with this responsibility

Oh, I do have an opinion now, then :-). I’m in favour of switching at 1.25, as this means my usually-zoomed-in-to-133% desktop browser gets HiDPI images as is always have, haha (in all seriousness, whether what I like is _most optimal_ indeed, I don’t know). We also have slightly different cut-off point for cutouts, but this was purely because PNGs were too weighty which is now moot point as everyone gets WebP (and, hopefully, soonish – AVIF/JXL).

dotcom-rendering/docs/architecture/027-pictures.md
```diff
@@ -0,0 +1,133 @@
...
+
+The browser will choose and stick with the first matching source element it finds, so it's important to ensure they're in the DOM in the right order.
+
+Another important media queries we use is (`(orientation: portrait)`) to check if it's a portrait device (e.g a smartphone).
+
+### Sizes (`sizes=`)
+
+The sizes html attribute acts as the translation layer between the size of the viewport and the size of the image source you'd like to pick. A good way to think about this is that, beyond a certain width, main media (inline) images never go beyond `620px` wide, and this html attribute gives us a way to communicate that to the browser.
+
+For example `<source sizes="(min-width: 660px) 620px, 100vw">` tells the browser, "Hey, if your viewport is 660px or wider, always look for an image source which is 620px wide. If not, default to an image which is the same width as 100% of the viewport width". We can provide as many sizes & queries as we want, with the `(query) px, (query) px, ... , px)` syntax, where the last argument without a query is the default if none others match.
```
mxdvl, line 35: Worth noting: it only makes sense to have multiple `sizes` if you have multiple `srcset`.

dotcom-rendering/docs/architecture/027-pictures.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Source Set (`srcset=`)
+
+Source sets work as our final piece of the puzzle. Our browser has already picked a `source` element, and has used `sizes` to figure out what size (in width only) image it is looking for. Our source set allows us to provide a list of URLs & the width of the image for each one, which the browser will then use to look for the best fitting image from.
+
+Source sets are formatted like: `<source srcset="https://url.to/image 300w, https://url.to/larger/image 600w">` Our comma separated list of sources specifies first the URL to a given image source, then the width (where `w` is, from what I can tell just `px`) for that image. In our case where we use fastly image optimiser, the only thing changing between these image urls is in the query parameters, e.g `?width=300` and `?width=600`.
```
mxdvl, line 41:
```suggestion
Source sets are formatted like: `<source srcset="https://url.to/image 300w, https://url.to/larger/image 600w">` Our comma separated list of sources specifies first the URL to a given image source, then the real pixel width for that image. The unit `w` is used, to distinguish pixels inside the image from CCS `px` on the screen–there are many pixels per CSS `px` on high DPI screens. In our case where we use Fastly image optimiser, the only thing changing between these image urls is in the query parameters, e.g `?dpr=2`, or `?width=300` and `?width=600`.
```

dotcom-rendering/docs/architecture/027-pictures.md
```diff
@@ -0,0 +1,133 @@
...
+        media="min-width: 480px"
+        sizes="480px"
+        srcset="https://xxx.png?width=480px 480w"
+    />
+    <source
+        media="min-width: 375px"
+        sizes="420px"
+        srcset="https://xxx.png?width=420px 420w"
+    />
+</picture>
```
mxdvl, line 96: Having a `sizes` attribute if you have only one element in your `scrset` is redundant, and maybe a little confusing.

```suggestion
<picture>
    <source
        media="min-width: 980px"
        srcset="https://xxx.png?width=620px 620w"
    />
    <source
        media="min-width: 660px"
        srcset="https://xxx.png?width=620px 620w"
    />
    <source
        media="min-width: 480px"
        srcset="https://xxx.png?width=480px 480w"
    />
    <source
        media="min-width: 375px"
        srcset="https://xxx.png?width=420px 420w"
    />
</picture>
```

dotcom-rendering/docs/architecture/027-pictures.md
```diff
@@ -0,0 +1,133 @@
...
+    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=465&;quality=85&;auto=format&;fit=max&; 465w" sizes="375px" media="(min-width: 375px)">
+    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=465&;quality=45&;auto=format&;fit=max&;dpr=2&; 930w" sizes="320px" media="(min-width: 320px) and (-webkit-min-device-pixel-ratio: 1.25), (min-width: 320px) and (min-resolution: 120dpi)">
+    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=465&;quality=85&;auto=format&;fit=max&; 465w" sizes="320px" media="(min-width: 320px)">
+    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=465&;quality=45&;auto=format&;fit=max&;dpr=2&; 930w" sizes="0px" media="(min-width: 0px) and (-webkit-min-device-pixel-ratio: 1.25), (min-width: 0px) and (min-resolution: 120dpi)">
+    <source srcset="https://i.guim.co.uk/img/media/picture.jpg?width=465&;quality=85&;auto=format&;fit=max&; 465w" sizes="0px" media="(min-width: 0px)">
```
mxdvl, line 130: In these instances, we have a `size` value that’s ignored, and often not a ratio of the `w` present in the `srcset`. It’s also worth noting that sizes refer to the [_intrinsic size_][] of the image, not its CSS pixel size.

[_intrinsic size_]: https://developer.mozilla.org/en-US/docs/Glossary/Intrinsic_Size

mxdvl (commented):
It’s great to see people caring about images 🖼️ 🎉 

> What do you think of the second option proposed in https://github.com/guardian/image-rendering/issues/239? In that scenario we provide a `<source>` tag per breakpoint, and use `srcset` to manage the DPR variations. That would reduce the number of `<source>` tags and drop the need to worry about the webkit (Safari) version of `min-resolution` as @mxdvl mentioned.

I think, like @JamieB-gu mentioned, that we should investigate Option two, which is the only real reason you’d want to use the `size` attribute. This final output would be something like:

```html 
<picture>
    <source
        media="(min-width: 375px)">
        sizes="(-webkit-min-device-pixel-ratio: 1.25),(min-resolution: 120dpi)  640px, 320px"
        srcset="https://xxx.jpg?width=320px 320w, https://xxx.jpg?width=320px&dpr=2 640w"
    />
    <source
        media="(min-width: 740px)">
        sizes="(-webkit-min-device-pixel-ratio: 1.25),(min-resolution: 120dpi)  1200px, 600px"
        srcset="https://xxx.jpg?width=600px 600w, https://xxx.jpg?width=600px&dpr=2 1200w"
    />
    <source
        media="(min-width: 1200px)">
        sizes="(-webkit-min-device-pixel-ratio: 1.25),(min-resolution: 120dpi)  1800px, 900px"
        srcset="https://xxx.jpg?width=600px 900w, https://xxx.jpg?width=900px&dpr=2 1800w"
    />
</picture>
```

- `media` relates to the size of the viewport and select the right `source`
- `sizes` relates to the pixel density and picks the right `srcset`
- `srcset` shows the possible versions of this image for this specific viewport/`media` attribute.

dotcom-rendering/docs/architecture/027-pictures.md
OllysCoding: Thanks! I have sorted this now :) Sizes begone!

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -15,13 +19,27 @@ type Props = {
 };
 
 type ResolutionType = 'hdpi' | 'mdpi';
+type Pixel = number;
+type Breakpoint = number;
+
+export type DesiredWidth = {
+	breakpoint: Breakpoint;
+	width: Pixel;
+};
```
mxdvl, line 28: I think you can make this an inferred  and narrower type with the following:

```suggestion
type Breakpoint = typeof breakpoint[keyof typeof breakpoints];

export type DesiredWidth = {
	breakpoint: Breakpoint;
	width: Pixel;
};
```

dotcom-rendering/src/web/components/Picture.tsx
mxdvl: One thing you can do if you want a bit of type safety is extracting all the possible options, [like so][]:

```ts
const imageWidths = [
    1020,
    940,
    700,
    680,
    620,
    320,
    300,
    /* … */
] as const;

type Breakpoints = typeof breakpoints[keyof typeof breakpoints]
type Pixel = Breakpoints | typeof imageWidths[number];
```

[like so]: https://www.typescriptlang.org/play?ssl=20&ssc=55&pln=8&pc=1#code/MYewdgzgLgBARgJwKYEMDWAHEBLMUIwC8MA3gFAyUwC2Ic2ANkgFwwBsADBwDQVVQo4TKKwDsAFh59KAEyQQ0UEBlYBOTryowA7tjmsAjAGYumqgHoAVDEBkBDEvmyAXxgoCoSFADcZD9BjY1CgA5kgA6npQABYExADa0jAGHABMUlqqkmaUoqaJbAAc6VRsadkwRmWJJsWUVrb2jgC6ru7g0D5kUACeGEgwAELI6Fi4+EQwPX0gAGbww5g4eBBxaEjds5O9SJuIqItjEE1d2zAACtgAHkgME0P7o8swAD5b03OBIeGRMXFgAK7UOBIBBNLxAA

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -111,25 +127,83 @@ const getSizes = (
...
+				? '(orientation: portrait) and (-webkit-min-device-pixel-ratio: 1.25), (orientation: portrait) and (min-resolution: 120dpi)'
+				: '(orientation: portrait)'
+		}
+		// Immersive MainMedia elements fill the height of the viewport, meaning
+		// on mobile devices even though the viewport width is small, we'll need
+		// a larger image to maintain quality. To solve this problem we're using
+		// the viewport height (vh) to calculate width. The value of 167vh
+		// relates to an assumed image ratio of 5:3 which is equal to
+		// 167 (viewport height)  : 100 (viewport width).
+		sizes="167vh"
```
mxdvl, line 200: This is clever, but ideally we would have a portrait source. Otherwise the majority of the image downloaded isn’t even used.

To investigate in another PR, though.

mxdvl (commented):
Thinking about images again this morning, so wrote down a few ideas and comments.

I wanted to point out that the `85` and `45` quality values for images have not been tested extensively enough, and a associated piece of work would be to review these values and A/B test the performance with our Core Web Vitals metrics.

dotcom-rendering/src/web/components/Picture.tsx
oliverlloyd: It's worth keeping in mind that I don't know what this new code does. Sure, I could go away and learn about this feature but there will be more people like me coming along.

It's also true that this might be a small thing to learn but I am a big fan of simple code with as few requirements to check the docs as possible. Each time we expect people to know about more advanced features of Typescript we raise the bar for access to DCR

paperboyo:
> I wanted to point out that the 85 and 45 quality values for images have not been tested extensively enough, and a associated piece of work would be to review these values and A/B test the performance with our Core Web Vitals metrics.

Thanks for having a look, Max! ❤️ 
I agree they haven’t been tested extensively enough. But they were tested enough to notice that even as-is some WebPs are already weightier than JPEGs, so we couldn’t increase their quality relative to the JPEGs much more if at all. Also, both formats, although WebP more severely, are definitely on the performance side, to put it mildly. Most WebPs are smeary, while some JPEGs exhibit posterisation. Hence, I don’t think measuring against metrics would gain us much. We need to measure against discerning humans caring about image quality more like.

Now, doing it before we will use AVIF makes little sense in my mind. Changing those values would decache all our imagery (even if we would earlier devise a way to test them less globally). I would argue, let’s have a look at quality values, dpr steps and the rest when we will be working on AVIF. Sadly, our image resizer is the slowest with offering support compared to (all?) competitors…

dotcom-rendering/src/web/components/Picture.tsx
paperboyo: Indeed we could use resizer to cut off sides in this scenario. It could mean, browsers will have to download another image when orientation or window ratio changes, though. Worth investigating!

Also worth noting, the assumption of 5:3 crop, while practically often true at the Guardian, isn’t necessary as we carry real ratio in CAPI.

oliverlloyd: > as we carry real ratio in CAPI

Oh, so we could replace `167` with a value calculated based on `ratio`? That seems like an easy win.

paperboyo: Waaaait… Why I can see `aspectRatio` only for the trail (always 5:3) image, but not for Main media which eg. [here](https://www.theguardian.com/artanddesign/2021/nov/23/tom-stoddart-obituary) is a different ratio?! We could calculate from `width` and `height`, but I will ask CAPI why `aspectRatio` isn’t offered for all imagery.

dotcom-rendering/src/web/components/Picture.tsx
mxdvl: This code narrows the type: `number` &rarr; `320 | 640 | 960`.

I don’t find the syntax particularly elegant, and obviously you can create [a wrapper][] like `Values<T>`, but this will still fall within what you call advanced features of TypeScript. I don’t have a good answer for this, and it’s really up to the team to decide whether they prefer knowing which number is returned by a function. If you do, then the compiler can tell you if you’ve missed certain branches, which I like, but may not be everyone’s cup of tea.

[a wrapper]: https://www.typescriptlang.org/play?#code/MYewdgzgLgBARgJwKYEMDWAHEBLMUIwC8MA3gFAyUwC2Ic2ANkgFwwBsADBwDQVVQo4TKKwDsAFh59KAEyQQ0UEBlYBOTryowA7tjmsAjAGYumqgHoAVDEBkBDEvmyAXxgoCoSFADcZMlACeGEgwAGooDACu8gA8ACowSAAeUEhgMgQASkigCDLR0Ai4AObcMBFgaGAg2mAAfLVEMLEA2mhI-iAAZk0Aun6BwQBCyOhYuPiNYZExAUFd8COYOHgQtUA

paperboyo:
@OllysCoding wins [Jimmy’s Award](https://github.com/guardian/frontend/pull/11055#issuecomment-153785228) for Best Described and Most Informative PRs 2021!

Congratulations on behalf of The Committee 🥇

PR #3642, opened 2021-11-16
move firstPublished and lastUpdated to common

marjisound:
## What does this change?
This PR moved the `FirstPublished` & `LastUpdated` components to common-rendering. 
## Why?
In order to share these components between AR & DCR

### DIFF
--- a/common-rendering/src/components/FirstPublished.tsx
+++ b/common-rendering/src/components/FirstPublished.tsx
@@ -18,7 +18,6 @@ const FirstPublished = ({
 		<a
 			href={blockLink}
 			data-ignore="global-link-styling"
-			// title={publishedDate.toLocaleString()}
 			css={css`
 				${textSans.xxsmall({ fontWeight: "bold" })}
 				margin-bottom: ${space[1]}px;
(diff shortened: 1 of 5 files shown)


common-rendering/src/components/FirstPublished.tsx
```diff
@@ -0,0 +1,58 @@
...
+}: {
+	firstPublished: number;
+	blockLink: string;
+}) => {
+	const publishedDate = new Date(firstPublished);
+	return (
+		<a
+			href={blockLink}
+			data-ignore="global-link-styling"
+			// title={publishedDate.toLocaleString()}
```
DavidLawes, line 21: tiny q: I can see we had this same comment in liveblock too, do we know if we need to retain it?

DavidLawes (approved):
👍

common-rendering/src/components/FirstPublished.tsx
marjisound: Good question. @lucymonie do you know why this line was commented out?

lucymonie: I don't know @marjisound, but I think it can come out as it's not in use.

PR #3644, opened 2021-11-16
Make key events sticky and share dropdown functionality for live blocks

lucymonie:
## What does this change?
This makes the key events element stick to the view on desktop and up. It is not sticky on smaller breakpoints.

At smaller breakpoints (below desktop), the key events element is in a dropdown container, which is also needed for the live blocks on these breakpoints. This refactors the dropdown functionality from key events into a separate component so it can be shared between key events and live blocks.

## Why?
Making key events sticky makes it easier for users to navigate to key blocks in our liveblogs. The other changes bring the key events and liveblocks closer to matching the new designs on smaller breakpoints.

### Stickiness
Screen Shot 2021-11-16 at 13 54 14

Screen Shot 2021-11-16 at 13 54 29

### Dropdown
Screen Shot 2021-11-16 at 13 55 11

Screen Shot 2021-11-16 at 13 55 26

### DIFF
--- a/common-rendering/src/components/dropDown.tsx
+++ b/common-rendering/src/components/dropDown.tsx
@@ -16,6 +16,8 @@ interface DropDownProps {
 	children: React.ReactNode;
 	supportsDarkMode: boolean;
 	dropDownTitle: string;
+	backgroundBody?: SerializedStyles;
+	backgroundTitle?: SerializedStyles;
 }
 
 const detailsStyles: SerializedStyles = css`
@@ -31,7 +33,6 @@ const detailsStyles: SerializedStyles = css`
 
 const titleRowStyles = (supportsDarkMode: boolean): SerializedStyles => css`
 	position: relative;
-	margin: 0 ${remSpace[1]} 0 ${remSpace[3]};
 	display: flex;
 	justify-content: space-between;
 	align-items: center;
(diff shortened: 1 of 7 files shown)


oliverlloyd (commented):
There's a lot of different things happening here. Can we break these changes out into smaller PRs? 🙏 

It's much easier to review and manage PRs with small focussed changes in them which results in better reviews that are easier to manage and merge. It can sound strange, but very often the best PRs are where the description is longer than the code changes!

In addition, smaller PRs provide a much more useful document for people coming back to code in the future to help them understand the motivations for changes. We only feel the effort of writing code once but its impact (all code has impact) is felt over and over for years to come.

But just to give some early feedback (without starting confusing threads on this PR), I notice that we're adding a new Dropdown component to common. Did you know that [one already exists](https://github.com/guardian/dotcom-rendering/blob/a42df26f3e33fd7340bdea059a3f67bf91b678db/dotcom-rendering/src/web/components/Dropdown.tsx) in DCR?

A dropdown component is very generic and should perhaps live in Source? Maybe a good first step here would be to look at raising the existing one up to the react kitchen? Then from there, both common and DCR (and whoever else might want to) can import it. Note, this existing version has a lot of great stuff built into it (like accessibility, keyboard usage, it works without javascript!, etc.) so whilst it's probably going to be _more_ effort now to hoist that code I think the long term benefits are worth it.

Also, less important but worth being aware of, there are some styling changes being made to other components here like `MainMedia` and `ArticleMeta`, these seem to be impacting non liveblog articles and the changes to the left column sizing here are already included in [this PR](https://github.com/guardian/dotcom-rendering/pull/3630)

PR #3645, opened 2021-11-16
Increase caption bottom padding if display is Showcase and design is Review

ioannakok:
## What does this change?
Increases padding-bottom in caption if display is Showcase and design is Review.

## Why?
Star rating was obscuring the photo caption in fillm review in mobile

### Before
image

### After
image

### DIFF
--- a/dotcom-rendering/src/web/components/Caption.stories.tsx
+++ b/dotcom-rendering/src/web/components/Caption.stories.tsx
@@ -1,11 +1,13 @@
 import { ElementContainer } from '@frontend/web/components/ElementContainer';
 import { Caption } from '@frontend/web/components/Caption';
+import { StarRating } from '@root/src/web/components/StarRating/StarRating';
 import {
 	ArticleDisplay,
 	ArticleDesign,
 	ArticlePillar,
 	ArticleSpecial,
 } from '@guardian/libs';
+import { css } from '@emotion/react';
 import { decidePalette } from '../lib/decidePalette';
 
 export default {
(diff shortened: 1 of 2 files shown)


DavidLawes (approved):
Nice :) 

Out of curiosity, is it only under these conditions:

```
		format.display === ArticleDisplay.Showcase &&
		format.design === ArticleDesign.Review
```

that we need to be aware of the overlayed padding-bottom?

oliverlloyd:
This diff looks a bit strange. Why would this code change cause the text colour to change?

https://www.chromatic.com/test?appId=5dfcbf3012392c0020e7140b&id=6193e311f0559c003a5a309e

oliverlloyd:
Shall we add a story for this scenario to `Caption.stories.tsx`?

ioannakok:
> This diff looks a bit strange. Why would this code change cause the text colour to change?
> 
> https://www.chromatic.com/test?appId=5dfcbf3012392c0020e7140b&id=6193e311f0559c003a5a309e

My bad! At a previous commit I had accepted this change: https://www.chromatic.com/test?appId=5dfcbf3012392c0020e7140b&id=6193be617b9bc0003aa38eda

ioannakok:
> Shall we add a story for this scenario to `Caption.stories.tsx`?

I had a look into this and it doesn't seem to be very straightforward. It does not seem possible to test `isOverlayed` in `Caption.stories.tsx`: https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/Caption.stories.tsx#L25

I had a look whether we could test this scenario in another stories file but there are some restrictions:
e.g.`ImageBlockComponent.stories.tsx`  --> does not accept `isOverlayed`
`CaptionBlockComponent.stories.tsx` --> It's just the caption without an image

ioannakok:
> Nice :)
> 
> Out of curiosity, is it only under these conditions:
> 
> ```
> 		format.display === ArticleDisplay.Showcase &&
> 		format.design === ArticleDesign.Review
> ```
> 
> that we need to be aware of the overlayed padding-bottom?

Yes, I believe it's the only use case that has come up so far :-)

oliverlloyd:
> doesn't seem to be very straightforward

You're right, it wasn't as simple as just saying `isOverlayed` true! I had a play though and was able to 'construct' the things needed to show how the caption looks when overlayed - I made a commit. It's not perfect, there's a lot of custom test code here, but it does the job.

oliverlloyd (approved):

ioannakok:
> >  I had a play though and was able to 'construct' the things needed to show how the caption looks when overlayed - I made a commit. It's not perfect, there's a lot of custom test code here, but it does the job.

Nice!

PR #3646, opened 2021-11-16
LIVE-3078: Updates sbt-scrooge-typescript plugin version for resolving from non-bintray repo

umguardian:
## What does this change?
Updates version of `sbt-scrooge-typescript` SBT plugin so it resolves corrrectly.

## Why?
With the move away from Bintray for hosting Guardian artifacts, the existing plugin version `1.2.5` does not get resolved. This pull request upgrades the version of the plugin so it can be correctly [resolved](https://repo1.maven.org/maven2/com/gu/sbt-scrooge-typescript_2.12_1.0/).


frankie297 (approved):
👍

PR #3647, opened 2021-11-16
update liveblog & deadblog background colour

DavidLawes:
## What does this change?
The background colour didn't match the design. 

NB: I think there could be further work on the background colours for Liveblogs (particularly Opinion Liveblogs) but certainly for Deadblogs this meets expectations (I believe).

## Why?
This change ensures parity with the design.

### Before
Screenshot 2021-11-16 at 15 34 37

### After
Screenshot 2021-11-16 at 15 34 51


ioannakok (approved):
+1

PR #3653, opened 2021-11-17
Use `@guardian/libs` for cookie functions

oliverlloyd:
## What does this change?
This deletes the functions we were using from the (two 🤷 ) `cookie.ts` files we had in our code and replaces them with functions from `@guardian/libs`

Basically we're swapping

`getCookie('my_thing')`

for

`getCookie({name: 'my_thing'})`

## Why?
Because we want to standardise code like this and get the benefits that the extra love given to this code in libs bring (specifically, `shouldMemoize` ❤️ )

## Isn't this sweeping change scary?
The code in `@guardian/libs` is based on these functions so outside of the api changes (using an object pattern) it should be a direct swap


OllysCoding:
What does the `shouldMemoize` property do?

oliverlloyd:
> What does the `shouldMemoize` property do?

It's a simple in-memory cache for previously read cookies

https://github.com/guardian/libs/blob/main/src/cookies.ts#L153

tomrf1 (approved):
looks good!

lucymonie:
This looks good 👍🏻  - I noticed there are a few cases where you `getCookie` but don't use `shouldMemoize`. Is this because these items shouldn't be cached?

oliverlloyd:
> Is this because these items shouldn't be cached?

Good spot! 

It's because I wasn't _sure_ that they could be cached. Caching is a nice bonus but the non cached call still gives us parity

PR #3660, opened 2021-11-18
Add another contents block

rcrphillips:
## What does this change?
Adds yet another contents block for DCR to render

https://github.com/guardian/frontend/pull/24411


DavidLawes (approved):
👍 

out of curiosity, how are we fixing these? are you updating the boot.js file in some way?

PR #3661, opened 2021-11-18
`bootCmp`

oliverlloyd:
## What does this change?
This PR moves the initialisation of `cmp` out of `App.tsx` and into it's own boot script.

Along with moving the initialisation, I also moved down some code from `App.tsx` that got the `consentState` and passed it down to the `YouTubeBlockComponent`. In doing this, I made it a dynamic import.

## Why?
This is part of a wider refactor to remove logic from `App.tsx`

## Things still under consideration
- Should we use a dynamic import in `YouTubeBlockComponent`
- Can we load the `bootCmp` script later, as a lower priority script tag?


dotcom-rendering/src/web/browser/bootCmp/init.ts
```diff
@@ -0,0 +1,107 @@
...
+		.then((willShow) => {
+			trackPerformance(
+				'consent',
+				'acquired',
+				willShow ? 'new' : 'existing',
+			);
+		})
+		.catch((e) =>
+			console.error(`CMP willShowPrivacyMessage - error: ${e}`),
+		);
```
mxdvl, line 56: FWIW I don’t think anyone is looking at this, sadly. We’ve got better measurement tools anyways, now, but the slowness of CMP is not something we can do anything about at this stage.

dotcom-rendering/src/web/components/elements/YoutubeBlockComponent.tsx
```diff
@@ -83,13 +84,37 @@ export const YoutubeBlockComponent = ({
...
 }: Props): JSX.Element => {
+	const [consentState, setConsentState] = useState<ConsentState | undefined>(
+		undefined,
+	);
+
+	useOnce(() => {
+		import(
+			/* webpackChunkName: "cmp" */ '@guardian/consent-management-platform'
+		)
+			.then((module: { onConsentChange: any }) => {
```
mxdvl, line 100:
```suggestion
			.then((module: { onConsentChange: (callback: Callback) => void }) => {
```

mxdvl: You’ll need to import this type!

dotcom-rendering/src/web/components/App.tsx
```diff
@@ -320,96 +304,6 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 		}
 	}, [CAPI.shouldHideReaderRevenue]);
 
-	// kick off the CMP...
-	useOnce(() => {
-		if (!CAPI.config.switches.consentManagement) return; // CMP turned off!
```
OllysCoding, line 325: Do we replicate this functionality with the new implementation? (If needed???? Should we be able to turn CMP off????)

mxdvl:
As CMP is _comparatively_ slow, I would be tempted to keep it very high up the priority list.

We need it to complete  in order to do anything that requires consent, such as:
- running ads
- embedding YouTube `iframe`s
- accessing local storage

oliverlloyd:
> We need it to complete in order to do anything that requires consent

Noted 👍

dotcom-rendering/src/web/components/App.tsx
oliverlloyd: If the switch exists, we should support it. Whether it should exist is another question.

@ashishpuliyel ?

dotcom-rendering/src/web/browser/bootCmp/init.ts
oliverlloyd: I'll refactor it to use the new `log` command so it's at least following a standard for messaging

dotcom-rendering/src/web/components/App.tsx
mchv: I confirm we should be able to turn the CMP off, for instance if there is an issue with Sourcepoint

oliverlloyd:
Just making a note here that @OllysCoding spoke to me about the bundle size increase of 4Kb we're seeing here and suggested we look at dependency caching with webpack. We're going to pair on this tomorrow

dotcom-rendering/src/web/components/elements/YoutubeBlockComponent.tsx
```diff
@@ -82,13 +86,39 @@ export const YoutubeBlockComponent = ({
...
 	height = 259,
 	width = 460,
 	duration,
 	origin,
 }: Props): JSX.Element => {
+	const [consentState, setConsentState] = useState<ConsentState | undefined>(
+		undefined,
+	);
+
+	useOnce(() => {
```
arelra, line 99: This is just a hunch... although this will initiate CMP and set consent state on a second pass it won't block the render of `YoutubeAtom` on the first pass without consent.

`YoutubeAtom` will render its first and second pass (after its internal useEffect) without consent and therefore default to disabled ads.

Previously the `HydrateOnce` for `YoutubeBlockComponent` would `waitFor={[consentState]}`

I _think_...

oliverlloyd:
[Break out CMP](https://trello.com/c/lqb0XjYk/426-break-out-cmp)

dotcom-rendering/src/web/components/elements/YoutubeBlockComponent.tsx
mxdvl: Is is simply that consentState should be added to the dependency array on L120?

oliverlloyd: Thanks @arelra. I think you're right. Previously we waited for `consentState` in `App.tsx` using a `waitFor` but we're not doing that here. I think I saw it suggested that we could fix this by making `consentState` a dependency for `useOnce` to run, which I think it true. This would essentially replicate the logic we had previously. Ahead of that though, I'm investigating if there are some optimisations that can be made in the `YouTubeAtom` itself. I'll update back here when I've done this.

kenoir:
The T&C docs will need updating to point at this new location I think: https://github.com/guardian/transparency-consent-docs/blob/main/docs/in-the-browser.md

dotcom-rendering/src/web/components/elements/YoutubeBlockComponent.tsx
arelra: You might be already considering this but I think _completely_ separating the overlay from the player will help reduce complexity of `YoutubeAtom` quite a bit and allow blocking of only whats required - i.e. instantiation of the player.

It will just need a method to trigger the play from the overlay to the player.

oliverlloyd: Based on this discussion, I looked deeper at how we render You Tube videos and discovered that we can improve how we do this.

@mchv pointed out [a great lib](https://github.com/justinribeiro/lite-youtube) to me earlier which defers the loading of the you tube javascript until the user clicks the poster overlay. My first response was to say that we already do this but when I looked more I realised we don't! So I've added this feature to our `YouTubeAtom`

https://github.com/guardian/atoms-rendering/pull/303

Whilst doing this I also added a defer to wait for `consentState` which, if we're happy with that approach` would address the issues in this thread, without the need to manage waiting for state locally.

I'm also happy to put the `waitFor` in and not have `YouTubeAtom` manage this (we could make it required) or some combination of the both. Whichever api makes the best sense to the group is fine by me.

oliverlloyd:
> The T&C docs will need updating to point at this new location

@kenoir I've created [a PR](https://github.com/guardian/transparency-consent-docs/pull/17) to do this.

shtukas (approved):

dotcom-rendering/src/web/components/elements/YoutubeBlockComponent.tsx
```diff
@@ -80,13 +84,39 @@ export const YoutubeBlockComponent = ({
...
+					module.onConsentChange((newConsent: ConsentState) => {
+						console.log('consent changed, new', newConsent);
+						setConsentState(newConsent);
+					});
+				},
+			)
+			.catch((error) => {
+				const msg = `Error: ${error}`;
+				// eslint-disable-next-line no-console
+				console.log(msg);
```
arelra, line 112: 👀

dotcom-rendering/src/web/components/elements/YoutubeBlockComponent.tsx
arelra: Is `useOnce` required if the dependency array is `[]` ?
If not you could just have a useEffect with `[]` ?

dotcom-rendering/src/web/components/elements/YoutubeBlockComponent.tsx
```diff
@@ -80,13 +84,39 @@ export const YoutubeBlockComponent = ({
...
+	);
+
+	useOnce(() => {
+		import(
+			/* webpackChunkName: "cmp" */ '@guardian/consent-management-platform'
+		)
+			.then(
+				(module: { onConsentChange: (callback: Callback) => void }) => {
+					module.onConsentChange((newConsent: ConsentState) => {
+						console.log('consent changed, new', newConsent);
```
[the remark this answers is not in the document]
oliverlloyd: !

PR #3662, opened 2021-11-19
Liveblogs: make key events sticky on desktop

lucymonie:
## What does this change?
This makes the key events component stick to view on breakpoints from desktop and above. It also fixes an issue with the key events grid setup where the key events was pushing the body down and creating a gap between main media and body. There's also a margin at the top of key events, requested by @HarryFischer.

## Why?
Making key events sticky makes it easier for users to navigate to key blocks in our liveblogs. 

UPDATE
I noticed that the sticky key events on frontend keeps the top border at the top when it sticks, and I also noticed that we're missing the top border line on the new key events (it's in the design). So I've added this border at the top of key events, and adjusted the margin and padding for this accordingly. I've also updated the screenshots to reflect this change.

### Before
Screen Shot 2021-11-19 at 11 44 00

### After - key events at the top
Screen Shot 2021-11-22 at 12 22 55

### After - key events sticky
Screen Shot 2021-11-22 at 12 23 14


oliverlloyd (approved):
This looks great!

Just one thing, what do you think about moving that wrapper div and css into the `KeyEventsContainer` component itself? It seems like this code is specific to that scope.

lucymonie:
Thanks. I did think about that, but my philosophy about these kids of components is that they shouldn't have opinions outside their boundaries. It should be the layout's job to determine how much space there is between key events and something else, and also whether it's sticky (not sure if you meant the sticky aspect in this case though).

OllysCoding (approved):
LGTM!

oliverlloyd:
> Thanks. I did think about that, but my philosophy about these kids of components is that they shouldn't have opinions outside their boundaries. It should be the layout's job to determine how much space there is between key events and something else, and also whether it's sticky (not sure if you meant the sticky aspect in this case though).

Yeah, I was thinking more the sticky css. I think there's lots of ways of doing things, especially with css, but it's perhaps good to be consistent and in other places where we have wrapping css, list sticking the right column, we put that css inside the components to keep it local to the things it is affecting.

I agree with margins though it's really nice to keep that sort of thing away from the actual component. For example, when rendering elements, we add vertical spacing using `Figure`, not the element. And when using `grid` we space content using `grid-column-gap: 10px;`. Maybe we should also be using `grid-row-cap` here!?

lucymonie:
> I agree with margins though it's really nice to keep that sort of thing away from the actual component. For example, when rendering elements, we add vertical spacing using `Figure`, not the element. And when using `grid` we space content using `grid-column-gap: 10px;`. Maybe we should also be using `grid-row-cap` here!?

This is a good point. I think it could be a good pattern for us to use, but it is probably outside the scope of a PR of this size. Definitely something for us to think about. I'll have a play around with sticky in key events 👍🏻

PR #3666, opened 2021-11-19
Move accordion into separate component for sharing

lucymonie:
## What does this change?
This moves the accordion functionality that was in the KeyEvents component into a separate component.

## Why?
This creates a component that can be shared between key events and the live feed.

There is already an [accordion component](https://theguardian.design/2a1e5182b/p/38c5aa-accordion) in Source, and there is an argument to be made that we should use it here. However, this particular accordion is set up to work with the design for these elements. The accordion shows at breakpoints tablet and smaller, and doesn't show from desktop and above. The component in Source wouldn't presently allow us to hide the accordion without also hiding its children, so for the moment, reuse of the existing code from key events is preferable.

Below are some screenshots from storybook.

### Tablet: open
Screen Shot 2021-11-19 at 14 59 31

### Tablet: closed
Screen Shot 2021-11-19 at 15 00 10

### Mobile: open
Screen Shot 2021-11-19 at 14 59 48

### Mobile: closed
Screen Shot 2021-11-19 at 14 59 59

### Tablet: open
Screen Shot 2021-11-23 at 14 47 57

### Tablet: closed
Screen Shot 2021-11-23 at 14 49 14

### Mobile: open
Screen Shot 2021-11-23 at 14 48 57

### Mobile: closed
Screen Shot 2021-11-23 at 14 48 32


common-rendering/src/components/accordion.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { breakpoints, space } from '@guardian/src-foundations';
+import { from } from '@guardian/src-foundations/mq';
+import { body } from '@guardian/src-foundations/typography';
+import type { FC } from "react";
+
+// ----- Stories ----- //
+
+const textStyle: SerializedStyles = css`
+	${body.medium()};
+	line-height: 150%;
```
JamieB-gu, line 15: Two questions on this:

- Is this the same as `body.medium({ lineHeight: 'loose' })`?
- Is "loose" (i.e. 1.5, 150%) the default for `body` already?

common-rendering/src/components/accordion.tsx
```diff
@@ -0,0 +1,121 @@
+// ----- Imports ----- //
+
+import { css } from "@emotion/react";
+import type { SerializedStyles } from "@emotion/react";
+import { neutral, line, background } from "@guardian/src-foundations/palette";
+import { headline } from "@guardian/src-foundations/typography";
+import { remSpace } from "@guardian/src-foundations";
+import { focusHalo } from "@guardian/src-foundations/accessibility";
```
JamieB-gu, line 8: Had no idea this was a thing! Amazing, thanks for teaching me about this 😊

common-rendering/src/components/accordion.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { SvgChevronUpSingle, SvgChevronDownSingle } from "@guardian/src-icons";
+import { from } from "@guardian/src-foundations/mq";
+import { darkModeCss } from "../lib";
+
+// ----- Component ----- //
+
+interface AccordionProps {
+	children: React.ReactNode;
+	supportsDarkMode: boolean;
```
JamieB-gu, line 17: 👍 🎉

common-rendering/src/components/accordion.tsx
```diff
@@ -0,0 +1,121 @@
...
+const Accordion = ({
+	children,
+	supportsDarkMode,
+	accordionTitle,
+	backgroundBody,
+}: AccordionProps) => {
+	return (
+		<details open css={detailsStyles}>
+			<summary css={titleRowStyles(supportsDarkMode)}>
+				<h2 css={titleStyle(supportsDarkMode)}>{accordionTitle}</h2>
```
JamieB-gu, line 106: If this component is generic and can be used in multiple places, will we always know this is an `h2` in the context of the page?

common-rendering/src/components/accordion.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+const titleStyle = (supportsDarkMode: boolean): SerializedStyles => css`
+	${headline.xxsmall({ fontWeight: "bold", lineHeight: "tight" })};
+	color: ${neutral[7]};
+	${darkModeCss(supportsDarkMode)`
+		color: ${neutral[86]};
+	`}
+`;
+
+const backgroundColour = (backgroundBody: 'white' | 'grey'): SerializedStyles => {
```
JamieB-gu, line 70: Does the background colour change in dark mode?

JamieB-gu (commented):
This is great!

A few small comments.

common-rendering/src/components/accordion.stories.tsx
lucymonie: I thought loose was 135% but I could be wrong!

lucymonie: Nope, you're right - that is loose 👍🏻

common-rendering/src/components/accordion.tsx
lucymonie: Ah - thank Marjan! I just moved it from her code in key events

common-rendering/src/components/accordion.tsx
oliverlloyd: What's the plan for colours? I feel it would be great to have a co-ordinated plan for this kind of thing rather than allowing different approaches to spread

common-rendering/src/components/accordion.tsx
lucymonie: I think in the use cases that we have at the moment, this is the right tag. Do you think it should be handled differently?

common-rendering/src/components/accordion.tsx
lucymonie: Good question - I'm guessing this would need to work differently for dark mode.

common-rendering/src/components/accordion.tsx
JamieB-gu: I don't know that's why I asked 😄. If all the known use cases call for an `h2` then it's probably fine for now 👍, no need to complicate the API before time.

common-rendering/src/components/accordion.tsx
JamieB-gu: The way I see it, if a component takes `supportsDarkMode` then that component already knows what it should look like in dark mode. It just includes the necessary styles if that's set to `true` (AR), and doesn't include them if it's set to `false` (DCR, Editions).

You can see this pattern further up this file in the title styles.

oliverlloyd: Off topic for this PR but I haven't really got my head around how `supportsDarkMode` works (which is not to say it will ot won't!). Do you think it will play well DCR if we want to use css variables to implement dark mode?

oliverlloyd: Also, this code (and I know it's just copied over but it's a good trigger for this discussion) is setting colours directly, not using the palette pattern. Is that intentional or is the expectation that we will migrate such usages to use `decidePalette` at some point?

common-rendering/src/components/keyEvents.tsx
```diff
@@ -220,16 +155,11 @@ const ListItem = ({ keyEvent, theme, supportsDarkMode }: ListItemProps) => {
...
-						<SvgChevronDownSingle />
-					</span>
-					<span className="is-on">
-						<SvgChevronUpSingle />
-					</span>
-				</summary>
+			<Accordion
+				supportsDarkMode={supportsDarkMode}
+				accordionTitle="Key events"
+				backgroundBody="white"
```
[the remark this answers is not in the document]
lucymonie: The background of the accordion body does change depending on its use. For the key events component, this background is white, and for the live feed, it's grey. Are you saying that you'd prefer to pass in css here?

common-rendering/src/components/accordion.tsx
lucymonie: 😄

common-rendering/src/components/accordion.tsx
lucymonie: @marjisound and I have just done some work to make sure that dark mode is working correctly for the accordion.

common-rendering/src/components/keyEvents.tsx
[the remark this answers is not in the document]
oliverlloyd: The down side to setting colour (or any css value) using a prop is that it's javascript, so to change it you'd need a runtime change, like a rerender. But it sounds like this is the case here and it is set once via a prop based on the context it's in (KeyEvents or Feed) and won't need to change?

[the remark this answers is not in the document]
lucymonie: No, it won't change once it's rendered.

common-rendering/src/components/accordion.tsx
marjisound: @oliverlloyd `darkModeCss` implementation in common is that the function only returns dark mode `SerializedStyles` css if `supportsDarkMode` is set to `true` and returns css`` if `supportsDarkMode` is set to `false`. Are you thinking of using css variables in DCR and not make use of the `darkModeCss` function?

marjisound: @oliverlloyd  we're using ` (backgroundBody: 'white' | 'grey')` since in the same liveblog page, we have 2 accordion that need different  background-color (keyevent & live block). This is also a similar issue I was chatting with you today for ShareIcons. Do you think the approach Lucy has taken could be improved?

oliverlloyd: > Are you thinking of using css variables in DCR and not make use of the `darkModeCss` function?

Basically, yes. I think for DCR, to avoid having to rerender the page, we would want to use css-variables. The bit where I'm unsure about is if this approach could live alongside this one?

> we have 2 accordion that need different background-color (keyevent & live block)

Yes, this sounds very similar to the issue we discussion earlier. There's a component that needs to be rendered in multiple contexts and it behaves differently in each. We can give it `format` and use a media query but there's still a missing piece of information to make the choice about which colour to use (i.e. the context for where the component is, like in KeyEvents). So, yeah, you can either:

- Pass in this context information as a prop - `isInKeyevnets`
- Or pass in the colour itself as a prop

I don't think there's a right or wrong. It's about trade offs

marjisound (approved):
Looks great Lucy 👍

PR #3669, opened 2021-11-22
Remove `CAPI` from `StickyBottomBanner`

oliverlloyd:
## What does this change?
This completes the work to remove the usage of `CAPI` from any component further down the React tree after `App.tsx`. We now pass explicit props to `StickyBottomBanner`

## Why?
Creates contracts with each component that needs hydrating on the client that are easier to meet


tjmw (approved):
This LGTM 👍 Let me know if I can be of any help testing on CODE etc

oliverlloyd:
> This LGTM 👍 Let me know if I can be of any help testing on CODE etc

Thanks. These might be famous last words but I'm confident these changes are not introducing any functional impact. I went through step by step when making them, running `yarn tsc` at each point, so I'm not expecting problems

tomrf1 (commented):
nice!

PR #3670, opened 2021-11-22
Update live layout grid to fix right column

DavidLawes:
## What does this change?
Previously the right column width wasn't quite right meaning the ads didn't display correctly. This PR alters the width and padding to match the intended design.

I reviewed the grid layout with @HarryFischer and made the following changes:
- Remove the top padding from the ad slot
- Ensure the body is fluid only up until tablet
- Ensure the body has a fixed width when greater than tablet
- Remove the border grid item and increase the grid-column-gap to achieve consistency in spacing between columns.

## Why?
To achieve parity with design. The screenshots below are at wide breakpoint, but see the chromatic build for how this impacts the live layout at different breakpoints (https://www.chromatic.com/build?appId=5dfcbf3012392c0020e7140b&number=8941).

### Before
Ad slot extends beyond right col
Screenshot 2021-11-22 at 17 30 36

### After
Ad slot sits within right col
Screenshot 2021-11-23 at 16 36 01

### DIFF
--- a/dotcom-rendering/src/web/layouts/LiveLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/LiveLayout.tsx
@@ -583,7 +583,6 @@ export const LiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
 							<GridItem area="right-column">
 								<div
 									css={css`
-										padding-top: 6px;
 										height: 100%;
 										${from.desktop} {
 											/* above 980 */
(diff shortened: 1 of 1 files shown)


DavidLawes:
[Right column and ad not displaying correctly above 1300px](https://trello.com/c/y29ArEvq/409-right-column-and-ad-not-displaying-correctly-above-1300px)

OllysCoding (approved):
+1 This makes sense to me!

oliverlloyd (approved):
Coders sans borders ✊

PR #3672, opened 2021-11-23
Ma/matchtabs in standard layout

nitro-marky:
## What does this change?

Matchtabs were not showing on test pages in DCR and this is potentially because they weren't included in the StandardLayout file. This adds them to it.

### After

Screenshot 2021-11-23 at 09 35 46

Tested locally.

### DIFF
--- a/dotcom-rendering/src/web/layouts/StandardLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/StandardLayout.tsx
@@ -483,14 +483,6 @@ export const StandardLayout = ({ CAPI, NAV, format, palette }: Props) => {
 											height={230}
 										/>
 									)}
-								{format.design === ArticleDesign.DeadBlog ||
-									(ArticleDesign.LiveBlog &&
-										CAPI.matchUrl && (
-											<Placeholder
-												rootId="match-tabs"
-												height={40}
-											/>
-										))}
 							</div>
 						</GridItem>
 						<GridItem area="matchtabs" element="aside">
--- a/dotcom-rendering/src/web/layouts/StandardLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/StandardLayout.tsx
@@ -492,14 +492,12 @@ export const StandardLayout = ({ CAPI, NAV, format, palette }: Props) => {
 						</GridItem>
 						<GridItem area="matchtabs" element="aside">
 							<div css={maxWidth}>
-								{format.design === ArticleDesign.DeadBlog ||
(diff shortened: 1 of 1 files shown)


rcrphillips (approved):
LGTM

lucymonie (changes requested):
The change looks reasonable and straightforward but there seem to be some knock-on effects to some parts of the liveblog, which you can see in the chromatic snapshots. It would be worth digging into these changes.

oliverlloyd (commented):
🤔 it's quirky that this placeholder div is appearing in the dom when the `matchtabs` grid area item isn't added to any grid

I think semantically we would want it to be in the grid css

oliverlloyd:
> The change looks reasonable and straightforward but there seem to be some knock-on effects to some parts of the liveblog, which you can see in the chromatic snapshots. It would be worth digging into these changes.

I think a lot of these diffs are from the leftcol PR but _this one_ isn't and is a problem with this code
Screenshot 2021-11-23 at 10 22 16

This is where the GridItem is being dropped at the bottom of the page because it hasn't been told where to go

nitro-marky:
So what would you recommend?

oliverlloyd (approved):

OllysCoding (approved):
Looks good!

dotcom-rendering/src/web/layouts/StandardLayout.tsx
```diff
@@ -485,6 +490,18 @@ export const StandardLayout = ({ CAPI, NAV, format, palette }: Props) => {
...
+								{format.design === ArticleDesign.DeadBlog ||
+									(format.design === ArticleDesign.LiveBlog &&
+										CAPI.matchUrl && (
+											<Placeholder
+												rootId="match-tabs"
+												height={40}
+											/>
+										))}
+							</div>
+						</GridItem>
```
oliverlloyd, line 504:
```suggestion
						<GridItem area="matchtabs" element="aside">
							<div css={maxWidth}>
								{CAPI.matchUrl && (
									<Placeholder
										rootId="match-tabs"
										height={40}
									/>
								)}
							</div>
						</GridItem>
```

oliverlloyd: I missed this but we don't need this format check here as we're never going to be a live blog at this point, not in StandardLayout

nitro-marky:
[Fix match tabs to show in articles again](https://trello.com/c/hY6MVUef/433-fix-match-tabs-to-show-in-articles-again)

lucymonie (approved):
👍🏻

PR #3673, opened 2021-11-23
Adds webPublicationDate in RelatedItem

umguardian:
## What does this change?

Updates thrift definition to add `webPublicationDate` in `RelatedItem`.

## Why?
This updates `apps-rendering-api-models` Scala and Typescript definitions which can be used to fix LIVE-3078.

NOTE: Corresponding Scala and Typescript packages have been published to Sonatype and NPM respectively.


marjisound (approved):
Great job Unmesh 👍

PR #3676, opened 2021-11-23
AR: Handle related content opinion card styles

frankie297:
## What does this change?
This changes the styles of opinion related content cards as per the figma file: https://www.figma.com/file/SL8IPGVP1FqFBkE7lMTI5l/apps-rendering-details?node-id=126%3A107

## Why?
The byline cards looked different to regular cards, and was pushing the content to be longer in height than necessary. The font was too big, and the picture was too big. It ended up looking like this: 

|

This PR changes that behaviour, and also makes some UI changes such as changing the background colour to be white instead of opinion orange.

| Before (phone) | After (phone) |
| --- | --- |
|||

| Tablet (portrait) | Tablet (landscape) |
| --- | --- |
|||


apps-rendering/src/relatedContent.ts
```diff
@@ -68,6 +69,14 @@ const parseHeaderImage = (content: Content): Image | undefined => {
 	}
 };
 
+const getBylineImage = (content: Content): string | undefined => {
+	const contributors = content.tags.filter(tag => tag.type === TagType.CONTRIBUTOR);
+
+	if (contributors.length > 0) {
+		return contributors[0].bylineLargeImageUrl
+	}
+};
```
JamieB-gu, line 78: There are a couple of functions already available for retrieving contributor information from `Content` objects that may be useful:

- `articleContributors` in `capi.ts`
- `parseContributors` in `contributors.ts`

JamieB-gu (commented):
Looks good.

One comment about existing functionality. Also it looks like the linter would like you to run `lint:fix` 🙂

apps-rendering/src/relatedContent.ts
```diff
@@ -3,7 +3,14 @@ import type { RelatedContent } from '@guardian/apps-rendering-api-models/related
 import { RelatedItemType } from '@guardian/apps-rendering-api-models/relatedItemType';
 import type { Content } from '@guardian/content-api-models/v1/content';
 import { andThen, fromNullable, map, OptionKind } from '@guardian/types';
-import { articleMainImage, isAnalysis, isFeature, isReview } from 'capi';
+import { withDefault } from '@guardian/types/dist';
```
JamieB-gu, line 6:
```suggestion
import { withDefault } from '@guardian/types';
```

JamieB-gu (approved):

apps-rendering/src/relatedContent.ts
frankie297: 👍

PR #3682, opened 2021-11-24
Remove blank space on deadblogs

nitro-marky:
## What does this change?

Deadblogs on DCR were having problems with the right column, which contained the ads, causing a huge space to appear. This modifies that column to hide at certain break points, removing the gap.

## Why?

Parity

### Before

Screenshot 2021-11-24 at 16 26 44

### After
Screenshot 2021-11-24 at 16 27 45


DavidLawes (approved):
👍 

also to note that we reviewed this with Harry too :)

nitro-marky:
[Fix big space on deadblogs](https://trello.com/c/uYSzJpOP/446-fix-big-space-on-deadblogs)

PR #3684, opened 2021-11-24
use non strict sanctions check when getting the discussion profile

pvighi:
## What does this change?
I'm not super familiar with DCR so I would appreciate if someone could take a look to make sure this change makes sense.
Currently, when loading pages with comments we are making a request to /profile/me in `discussion-api`.
This PR adds a new parameter to perform a more lightweight call at the cost of a slightly less strict check for sanctions.  In most cases this should return the same thing, but in some corner cases it would return that the user is allowed to comment even though they are not.

## Why?
we want to reduce the load on identity. In most cases nothing should change, but in some corner cases it would not notify the user in advance that they are banned before they try to comment. For more details look at the [discussion PR that added the new parameter](https://github.com/guardian/discussion-api/pull/695)

### Before (corner case)
image

### After(corner case)
The "commenting disabled" message is not displayed but the user still cannot comment if they try (this screenshot is after attempting to post a comment)
image

### DIFF
--- a/dotcom-rendering/cypress/integration/parallel-4/signedin.spec.js
+++ b/dotcom-rendering/cypress/integration/parallel-4/signedin.spec.js
@@ -28,7 +28,11 @@ describe('Signed in readers', function () {
 			return false;
 		});
 		// Mock call to 'profile/me'
-		cy.intercept('GET', '**/profile/me', profileResponse);
+		cy.intercept(
+			'GET',
+			'**/profile/me?strict_sanctions_check=false',
+			profileResponse,
+		);
 		cy.visit(`Article?url=${articleUrl}`);
 		// This text is shown in the header for signed in users
 		cy.contains('My account');
--- a/dotcom-rendering/src/web/lib/getUser.ts
+++ b/dotcom-rendering/src/web/lib/getUser.ts
@@ -19,7 +19,7 @@ const callApi = (url: string) => {
 };
 
 export const getUser = async (ajaxUrl: string): Promise<UserProfile | void> => {
-	const url = joinUrl([ajaxUrl, 'profile/me']);
+	const url = joinUrl([ajaxUrl, 'profile/me?strict_sanctions_check=false']);
 	return callApi(url)
 		.catch((error) => {
 			window.guardian.modules.sentry.reportError(error, 'get-user');


dotcom-rendering/src/web/lib/getUser.ts
```diff
@@ -19,7 +19,7 @@ const callApi = (url: string) => {
 };
 
 export const getUser = async (ajaxUrl: string): Promise<UserProfile | void> => {
-	const url = joinUrl([ajaxUrl, 'profile/me']);
+	const url = joinUrl([ajaxUrl, 'profile/me?strict_sanctions_check=false']);
```
pvighi, line 22: Ideally we should only perform the less strict sanctions check when we load the page, would it be correct to hard code it here ? We could also add a boolean parameter to this getUser function that defaults to true so that we are sure to only perform the less strict check when we are sure it's ok to do so

dotcom-rendering/cypress/integration/parallel-4/signedin.spec.js
```diff
@@ -28,7 +28,11 @@ describe('Signed in readers', function () {
 			return false;
 		});
 		// Mock call to 'profile/me'
-		cy.intercept('GET', '**/profile/me', profileResponse);
+		cy.intercept(
+			'GET',
+			'**/profile/me?strict_sanctions_check=false',
+			profileResponse,
+		);
```
oliverlloyd, line 35: This is a take it or leave it comment, so I'll leave it up to you if you think it's better or not, but it may be more in keeping with the intentions of the test to use a wildcard here? I.e.:

```suggestion
		cy.intercept(
			'GET',
			'**/profile/me**',
			profileResponse,
		);
```

The downside is it will also match on `profile/me/special` but I don't expect these endpoints to appear.

dotcom-rendering/src/web/lib/getUser.ts
oliverlloyd: There's a [separate thread of work](https://docs.google.com/document/d/1kSrZZrJ8hSw6bek5hGa9onE8QDz5TRHFqHV52EOFjts/edit#heading=h.gsod33a0pc47) looking at how we can refactor the way we load data into the page and this information is really useful for that, thanks for pointing it out!

I think in this case using a hard coded url is fine. It also means we won't split the cache by having different url params. If we need more control at a later stage, we can look at bringing in a different api for `getUser`

oliverlloyd (approved):
This is an awesome PR, it's a really well researched, tested and documented piece of work. Thank you!

dotcom-rendering/src/web/lib/getUser.ts
pvighi: If I understand correctly this api call happens on every article load, even if there's no discussion enabled.  I think a more complete fix to reduce the usage of the api would be to also stop making this discussion api call if we are not going to use the response

PR #3685, opened 2021-11-24
show 'wide' LeftColumn after leftCol breakpoint for live layouts

DavidLawes:
## What does this change?
For live layouts, until the leftCol breakpoint, we don't want the end of page containers to have a left column. From the leftCol breakpoint we want the end of page containers to include the left column. After the last left column change we discussed with design and thought the onwards/most popular content didn't look quite right between desktop and leftCol - this PR changes the behaviour to match Design's expectations.

## Why?
To meet parity with the design.

### Before
Screenshot 2021-11-24 at 17 39 11

### After
Screenshot 2021-11-24 at 17 45 38


idrise (approved):

OllysCoding (approved):
+1

PR #3689, opened 2021-11-25
fix dcr dead/live blog styling for metadata V2

marjisound:
## What does this change?
This PR fixes the folloing:

1. deadblog three Lines Colours in smaller breakpoints 
2. CommentCount colours for liveblog smaller breakpoints
3. ShareCount colours for liveblog smaller breakpoints
4. Fix colour for the vertical line between shareCount and CommentCount in smaller breakpoints 
5. Fix colour for horizontal line in meta between date and share icons
4. fix standfirst layout issue in tablet (740 width)

### Any advice would be very helpful for the following 2 questions:
1. The colour of the horizontal line (border-top of `ShareIcons` section and `CommentCount/ShareCount` section) in the middle of the Meta has an issue in mobile breakpoint. Because the `ShareIcons` container div and `CommentCount/ShareCount` container div are overlapping. Do you know how this can be fixed?
2. The share icons are news[300] in meta but news[400] in blocks. Is this intentional? @HarryFischer 

## Why?

| | Before | After |
| - | - | - |
| Mobile Live | image | image |
| Tablet Live | image | image |
| Mobile Dead | image | image |
| Mobile Dead | image| image |


JamieB-gu (approved):
LGTM! A few small questions/suggestions.

dotcom-rendering/src/web/components/ArticleMeta.tsx
```diff
@@ -71,6 +71,12 @@ const stretchLines = css`
 	}
 `;
 
+const metaBorderTopStandfirstStyling = css`
+	${until.desktop} {
+		border-top: 1px solid rgba(255, 255, 255, 0.4);
```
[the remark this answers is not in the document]
marjisound: This colour is rgba with the need of opacity. If I use opacity outside of colour like below, it'll affect everything within it's container (not just border-top). Hence I have to use rgba which I couldn't find any source equivalent.
```
border-top: 1px solid ${neutral[100]};
opacity: 0.4;
```

dotcom-rendering/src/web/components/CommentCount.stories.tsx
```diff
@@ -30,11 +49,8 @@ export const SpecialReportStory = () => {
 				isCommentable={true}
 				setIsExpanded={() => {}}
 				commentCount={306}
-				palette={decidePalette({
-					theme: ArticleSpecial.SpecialReport,
-					display: ArticleDisplay.Standard,
-					design: ArticleDesign.Standard,
-				})}
+				palette={decidePalette(specialReportFormat)}
+				format={specialReportFormat}
```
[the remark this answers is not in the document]
marjisound: I remember a while ago I was having this chat with oliver and he was also suggesting that we can only pass format instead of both format and palette and then create the palette within the component. 

@oliverlloyd Can this approach be done by gradually refactoring the existing components to only pass the format? OR as Jamie said this was avoided for performance reasons?

dotcom-rendering/src/web/components/Counts.tsx
```diff
@@ -20,8 +21,21 @@ const containerStyles = css`
 	}
 `;
 
-export const Counts = ({ children }: Props) => (
-	<div css={containerStyles}>
+const standfirstColouring = css`
+	${until.desktop} {
+		.meta-number:not(:empty) + .meta-number:not(:empty) {
+			border-left: 1px solid rgba(255, 255, 255, 0.4);
```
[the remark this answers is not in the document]
marjisound: This is again for the same reason I explained above :)

dotcom-rendering/src/web/components/ShareCount.tsx
```diff
@@ -1,16 +1,25 @@
...
+	until,
+} from '@guardian/source-foundations';
 
 import ShareIcon from '@frontend/static/icons/share.svg';
 
 import { useApi } from '@root/src/web/lib/useApi';
 import { formatCount } from '@root/src/web/lib/formatCount';
 import { joinUrl } from '@root/src/lib/joinUrl';
+import { ArticleDesign } from '@guardian/libs';
+import { decidePalette } from '../lib/decidePalette';
```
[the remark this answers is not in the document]
marjisound: Good question. I think here we could also do this `import { decidePalette } from '@root/src/web/lib/decidePalette';` but I'm not sure what's the preferred way as I can find both ways in dotcom. @oliverlloyd what do you think?

dotcom-rendering/index.d.ts
```diff
@@ -141,6 +141,7 @@ type Palette = {
 	fill: {
 		commentCount: Colour;
 		shareIcon: Colour;
+		shareIconDark: Colour;
```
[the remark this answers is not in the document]
marjisound: You're right. I was doubtful what to call it. I changed it to `shareIconGrayBackground` but still wasn't sure if this is the best name. Happy to change if you can come up with a better name :)

lucymonie:
> The colour of the horizontal line (border-top of ShareIcons section and CommentCount/ShareCount section) in the middle of the Meta has an issue in mobile breakpoint. Because the ShareIcons container div and CommentCount/ShareCount container div are overlapping. Do you know how this can be fixed?

Hey @marjisound this looks great. I've been digging into the issue you raised with the overlapping borders, and as far as I can see, this is caused by the negative margins to the right of the left-hand div and the negative margins to the left of the right-hand div. These are on lines 87, 89, 94 and 96 in `metaExtras`, and on lines 115, 117, 122, and 124 in `metaNumbers` inn `Article.tsx`. My understanding is that these negative margins are to make full-width elements overlap their container padding so the lines go right to the border. 

I tried adding some override styles that I applied similarly to the way you conditionally added `metaBorderTopStandfirstStyling` to get the border colour right. The styles I added are below. 

```
const metaExtrasLiveBlog = css`
	${until.phablet} {
		margin-right: 0;
		padding-right: 0;
	}

	${until.mobileLandscape} {
		margin-right: 0;
		padding-right: 0;
	}
`;

const metaNumbersLiveBlog = css`
	${until.phablet} {
		margin-left: 0;
		padding-left: 0;
	}

	${until.mobileLandscape} {
		margin-left: 0;
		padding-left: 0;
	}
`;
```

I then added these to the style array for metaEtras:
```
css={[
	metaExtras(palette),
	format.design === ArticleDesign.LiveBlog &&
		metaBorderTopStandfirstStyling,
	format.design === ArticleDesign.LiveBlog && metaExtrasLiveBlog,
]}
```

And for metaNumbers:
```
css={[
	metaNumbers(palette),
	format.design === ArticleDesign.LiveBlog &&
		metaBorderTopStandfirstStyling,
	format.design === ArticleDesign.LiveBlog && metaNumbersLiveBlog,
]}
```

It's a bit clunky but it seemed to solve the overlap issue.

dotcom-rendering/src/web/components/CommentCount.stories.tsx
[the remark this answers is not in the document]
oliverlloyd: Gradually refactoring is fine here. This is definitely a case where moving things around until the feel right is a sound approach.

Re performance. I don't think the generation of `palette` takes a hugely long time, I would guess we're in microseconds territory here, but at some point as this object grows we can certainly think about breaking the logic apart and exporting individual components 👍

dotcom-rendering/src/web/components/CommentCount.stories.tsx
```diff
@@ -17,6 +18,24 @@ export default {
...
+	display: ArticleDisplay.Standard,
+	design: ArticleDesign.Standard,
+};
+
+const newsFormat: ArticleFormat = {
+	theme: ArticlePillar.News,
+	display: ArticleDisplay.Standard,
+	design: ArticleDesign.Standard,
+};
+
```
oliverlloyd, line 38: You could inline these

dotcom-rendering/src/web/components/ArticleMeta.tsx
```diff
@@ -71,6 +71,12 @@ const stretchLines = css`
 	}
 `;
 
+const metaBorderTopStandfirstStyling = css`
```
oliverlloyd, line 74:
```suggestion
const borderColourWhenBackgroundRed = css`
```

This is a take it or leave it comment. This a hard thing to name!

dotcom-rendering/src/web/components/CommentCount.test.tsx
```diff
@@ -6,16 +6,17 @@ import { CommentCount } from './CommentCount';
 import { decidePalette } from '../lib/decidePalette';
 
 describe('CommentCount', () => {
+	const format = {
+		theme: ArticlePillar.News,
+		design: ArticleDesign.Standard,
+		display: ArticleDisplay.Standard,
+	};
```
oliverlloyd, line 13: I think this abstraction is fair although personally I would still keep this code inline. I guess the thing to consider is what do you need from this code? What is more important:

1) Fewer lines of code
2) Lack of repetition
3) Easy to change in future

My personal view is that we should be writing code that prioritises ease of maintenance in the future. One of our key goals with DCR is around building something that will endure.

All that said, this particular abstraction is fine and I'm totally happy to keep this code. I just thought this was a nice opportunity to raise this discussion as I think it's an important concept

dotcom-rendering/src/web/components/CommentCount.tsx
```diff
@@ -20,6 +20,10 @@ const containerStyles = (palette: Palette) => css`
 	font-weight: bold;
 	color: ${palette.fill.commentCount};
 	padding-top: 5px;
+
+	${until.desktop} {
+		color: ${palette.fill.commentCountUntilDesktop};
+	}
```
oliverlloyd, line 26: Nice 👍 this pattern makes sense!

dotcom-rendering/src/web/components/ShareCount.test.tsx
```diff
@@ -15,6 +16,12 @@ describe('ShareCount', () => {
 	const pageId =
 		'/environment/2020/jan/25/court-probe-carrie-symonds-influence-boris-johnson-badger-cull';
 
+	const defaultFormat = {
+		theme: ArticlePillar.Opinion,
+		display: ArticleDisplay.Standard,
+		design: ArticleDesign.Standard,
+	};
```
oliverlloyd, line 23: I think this is perhaps better inlined. The name here is implying something special with this format - that it is the default one - which isn't true and could be confusing.

dotcom-rendering/src/web/components/ShareCount.tsx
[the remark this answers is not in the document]
oliverlloyd: I know that there was initially a preference for using aliases but then VSCode auto imports came along and stomped all over that.

Personally, I've noticed that the aliases have been the source of a lot of config work but I also appreciate that they make moving files easier

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -118,14 +119,12 @@ const StandFirstGrid = ({ children }: { children: React.ReactNode }) => (
 						'standfirst'
 						'lines'
 						'meta';
+					grid-column-gap: 0px;
```
oliverlloyd, line 122: What's happening with this change?

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -531,19 +536,41 @@ const backgroundSpeechBubble = (format: ArticleFormat): string => {
...
 	if (format.theme === ArticleSpecial.SpecialReport)
 		return specialReport[300];
 	return pillarPalette[format.theme].main;
 };
 
+const fillCommentCountUntilDesktop = (format: ArticleFormat): string => {
+	if (format.design === ArticleDesign.LiveBlog) return WHITE;
+
+	return fillCommentCount(format);
+};
```
oliverlloyd, line 554: 💯

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -531,19 +536,41 @@ const backgroundSpeechBubble = (format: ArticleFormat): string => {
...
+
 const fillShareIcon = (format: ArticleFormat): string => {
 	if (format.theme === ArticleSpecial.Labs) return BLACK;
 	if (format.theme === ArticleSpecial.SpecialReport)
 		return specialReport[300];
 	return pillarPalette[format.theme].main;
 };
 
+const fillShareCountIcon = neutral[46];
```
oliverlloyd, line 563:
```suggestion
const fillShareCountIcon = (): string => {
	return neutral[46];
};
```

What do you think about maintaining the function pattern? I think this will make the code easier to reason about and scale if everything is consistent and it should make moving to importing direct functions easier as well. If I'm importing functions _and_ variables then it will be hard to know when I need to execute them or not.

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -860,6 +895,8 @@ export const decidePalette = (format: ArticleFormat): Palette => {
 			numberedPosition: textNumberedPosition(),
 			overlayedCaption: textOverlayed(),
 			pagination: textPagination(format),
+			shareCount: textShareCount,
```
oliverlloyd, line 898: As above, shall we keep the same pattern for all properties?

dotcom-rendering/src/web/lib/layoutHelpers.ts
```diff
@@ -25,6 +25,15 @@ export const decideLineCount = (design?: ArticleDesign): 8 | 4 => {
...
 
+export const decideLineColour = (
+	design?: ArticleDesign,
+): string | undefined => {
+	if (design === ArticleDesign.LiveBlog) {
+		return 'rgba(255, 255, 255, 0.4)';
+	}
+	return undefined;
+};
+
```
oliverlloyd, line 36: We can inline this abstraction. It's not used anywhere else and it's good to keep code as close to possible to where it is being used so that people reading it in future don't have to open multiple files to see what is happening unless they really need to

oliverlloyd (commented):
Looking great!

The vast majority of my comments have the words 'inline' and 'abstraction' in them. Very happy to discuss this more but a lovely article on the core concept motivating me is [this one](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction). This isn't to say abstractions are wrong, just that we should be cautious with them and that duplication is okay.

dotcom-rendering/src/web/components/ArticleMeta.tsx
lucymonie: It's not always going to be red, is it? Isn't this for any liveblog?

oliverlloyd: Good point!

oliverlloyd: `borderColourWhenBackgroundDark`?

dotcom-rendering/src/web/components/CommentCount.test.tsx
marjisound: I get your point, thanks for sharing. My reason was that the the same format was being repeated several times and thought to abstract it to reduce repetition. I think here maybe I can improve the ease of code reading by giving it a better name such as `standardNewsFormat` rather than just `format`

dotcom-rendering/src/web/components/ShareCount.test.tsx
marjisound: Now it's making more sense to me, thanks :)

dotcom-rendering/src/web/layouts/LiveLayout.tsx
marjisound: I couldn't understand why this is happening, but right now in prod for below desktop 20px gap is causing this:

image

And setting it to 0 would change the number of grid areas

image

dotcom-rendering/src/web/lib/decidePalette.ts
marjisound: Great idea Oliver, thanks. I also think it'd be easier for the consumer if everything is function so they know it always needs to be executed 👍

dotcom-rendering/src/web/lib/layoutHelpers.ts
marjisound: I was also doubtful about this. Made it inline within where it was needed

dotcom-rendering/src/web/layouts/LiveLayout.tsx
oliverlloyd: I seee... Nice spot!

oliverlloyd (approved):
LGTM 🚀

marjisound:
@lucymonie thanks a lot for your wonderful suggestion. I followed your approach and it did fix the issue. This is the commit for that change https://github.com/guardian/dotcom-rendering/pull/3689/commits/c4e34d228fb432691b52cfc1deb78b3490a024bf

PR #3692, opened 2021-11-26
Add accordion to liveblog feed on smaller breakpoints

lucymonie:
## What does this change?
This adds the shared accordion component to the live/dead blog live feed on smaller breakpoints (tablet and below).

~To match the designs, the background of the lower part of the page is now a darker grey. Note that above the main media, the colour remains the lighter grey to provide adequate contrast for the subnav. The background colour change is only for the smaller breakpoints.~

There was an issue with flexbox `justify-content: space-between` for the top of the accordion that I discovered while testing on browserstack. See image:

Screen Shot 2021-11-29 at 22 08 07

I replaced this with `position: absolute` for the up and down arrow icons, which works on iOS and other operating systems that I've since tested.

## Why?
This brings the live feed in line with the designs for the new liveblog on breakpoints tablet and below.

### Both accordions closed
Screen Shot 2021-12-08 at 13 05 20

### Live feed accordion open
Screen Shot 2021-12-08 at 13 05 06

### Key events accordion open
Screen Shot 2021-12-08 at 13 05 43


common-rendering/src/components/accordion.tsx
```diff
@@ -31,12 +31,11 @@ const detailsStyles: SerializedStyles = css`
 
 const titleRowStyles = (supportsDarkMode: boolean): SerializedStyles => css`
 	position: relative;
-	display: flex;
-	justify-content: space-between;
```
lucymonie, line 35: I had to remove this as it was breaking on iOS 😢

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -529,6 +520,10 @@ export const LiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
 						showTopBorder={false}
 						backgroundColour={palette.background.article}
 						borderColour={palette.border.article}
+						backgroundColourMobile={
+							palette.background.liveBlogMobile
+						}
+						padded={false}
```
lucymonie, line 526: I've made this false as we need to go full width on the smaller breakpoints.

DavidLawes (approved):
Looks a lot closer to the design :)

dotcom-rendering/src/web/components/MainMedia.tsx
```diff
@@ -70,29 +70,34 @@ export const MainMedia: React.FC<{
...
+		format.design === ArticleDesign.LiveBlog ||
+		format.design === ArticleDesign.DeadBlog;
+	return (
+		<div
+			css={[
+				mainMedia,
+				isImmersive && immersiveWrapper,
+				!isImmersive && !isLiveOrDeadBlog && noGutters,
+			]}
+		>
```
oliverlloyd, line 85: This code looks like its reached the point where it could benefit from a refactor to make it clearer?

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -180,16 +181,6 @@ const LiveGrid = ({ children }: { children: React.ReactNode }) => (
...
-					grid-template-columns: 700px; /* Main content */
-					grid-template-areas:
-						'media'
-						'lines'
-						'meta'
-						'keyevents'
-						'body';
-				}
-				/* fluid until tablet */
-				${until.tablet} {
```
oliverlloyd, line 192: Yes! This was redundant!

Just one thing, does this mean `700px` can be `1fr` now?

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -218,20 +209,22 @@ const sticky = css`
...
 `;
 
-const keyEventsTopMarginDesktop = css`
+const keyEventsTopMargin = css`
 	${from.desktop} {
 		margin-top: ${space[1]}px;
 	}
 `;
 
-const stretchLines = css`
```
oliverlloyd, line 227: How did you get around the need for negative margins here?

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -531,6 +531,10 @@ const backgroundSpeechBubble = (format: ArticleFormat): string => {
 	return pillarPalette[format.theme].main;
 };
 
+const backgroundLiveblogMobile = (): string => {
```
oliverlloyd, line 534: This feels a bit off. We pass `format` to `decidePalette` and ask it to choose the colour for a context based on the `format` of the article. But in this case we're saying `backgroundLiveblogMobile`. So the fact it is a liveblog is baked into the property instead of `format.Design === ArticleDesign.LiveBlog`, you see?

I'm also wondering if we can't use a media query here? It sounds like we want a particular colour for `mobile`?

dotcom-rendering/src/web/components/ElementContainer.tsx
```diff
@@ -79,7 +87,9 @@ export const ElementContainer = ({
 				</div>
 			);
 			const style = css`
-				${backgroundColour && setBackgroundColour(backgroundColour)}
+				${backgroundColour && setBackgroundColour(backgroundColour)};
+				${backgroundColourMobile &&
+				setBackgroundColourMobile(backgroundColourMobile)};
```
oliverlloyd, line 92: What happens when both `backgroundColour` and `backgroundColourMobile` are passed?

dotcom-rendering/src/web/components/MainMedia.tsx
```diff
@@ -70,29 +70,34 @@ export const MainMedia: React.FC<{
...
+	const isImmersive = format.display === ArticleDisplay.Immersive;
+	const isLiveOrDeadBlog =
+		format.design === ArticleDesign.LiveBlog ||
+		format.design === ArticleDesign.DeadBlog;
+	return (
+		<div
+			css={[
+				mainMedia,
+				isImmersive && immersiveWrapper,
+				!isImmersive && !isLiveOrDeadBlog && noGutters,
```
oliverlloyd, line 83: I think this code has reached a point where it could benefit from a refactor. You could argue the previous addition of logic was already the start but at the time I suspect they (we, I?) decided it was okay, but I think we're at a tipping point now and given this is core code I think it makes sense to review this some more.

It seems like we're saying the margins are different in different scenarios? Standard, Immersive and Liveblog?

oliverlloyd (commented):
Thanks for removing that redundant grid css, that's been bugging me for ages!

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -531,6 +531,10 @@ const backgroundSpeechBubble = (format: ArticleFormat): string => {
...
+	return '#DCDCDC';
```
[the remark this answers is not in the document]
lucymonie: Yes, you're absolutely right. Thanks David.

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -157,7 +158,7 @@ const LiveGrid = ({ children }: { children: React.ReactNode }) => (
 				/* from desktop define fixed body width */
 				${from.desktop} {
 					grid-column-gap: 20px;
-					grid-template-columns: 220px 700px;
+					grid-template-columns: 240px 700px;
```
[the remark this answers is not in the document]
lucymonie: Yes, I had to take out the built-in side padding for mobile but I couldn't put a grid column on the outside, hence this.

dotcom-rendering/src/web/layouts/LiveLayout.tsx
lucymonie: Yes - I suppose it can as the content is full width on breakpoints up to desktop. Good point Oliver!

dotcom-rendering/src/web/layouts/LiveLayout.tsx
lucymonie: We no longer need to stretch the lines as they no longer go full width below desktop or above:

Here's a screenshot of the design on desktop
Screen Shot 2021-12-01 at 19 15 03

And here it is tablet and below
Screen Shot 2021-12-01 at 19 15 33

dotcom-rendering/src/web/lib/decidePalette.ts
lucymonie: Yeah, I have to say this is a sort of preliminary offering as I wanted to talk to you about how to do this in keeping with the decidePalette pattern. Would you have time this morning to have a quick chat about it?

dotcom-rendering/src/web/components/MainMedia.tsx
lucymonie: Yes, it's getting a bit difficult to reason about. Would you have time to chat about a better solution?

dotcom-rendering/src/web/components/ElementContainer.tsx
lucymonie: So if both are passed in you get a main style and a style wrapped in a media query, like this:
```
background-color: ${backgroundColour};
${until.desktop} {
	background-color: ${backgroundColourMobile};
}
```

dotcom-rendering/src/web/components/MainMedia.tsx
```diff
@@ -50,6 +50,24 @@ const immersiveWrapper = css`
...
+					return '';
+				default:
+					return noGutters;
+			}
+		}
+		default:
+			return noGutters;
+	}
+};
+
```
lucymonie, line 70: Hey @oliverlloyd here is my refactor as discussed.

oliverlloyd (commented):
I don't think we can make changes to `ElementContainer` for this use case. That's such a key shared component, one that ideally we will simplify and then hoist up at some point.

So is the issue here that we have this divider and that below a certain breakpoint we want the background colour to change? But we still want to use `ElementContainer` because that gives us the spacing and borders?

Could we do something like this?

```ts
<Hide above="mobile"><ElementContainer backgroundColour="darkGrey" /></Hide>
<Hide below="mobile"><ElementContainer backgroundColour="lightGrey" /></Hide>
```
?

We'd be duplicating that code but it would remove the need to change the props on `ElementContainer`

dotcom-rendering/src/web/components/ElementContainer.tsx
```diff
@@ -29,12 +29,19 @@ const setBackgroundColour = (colour: string) => emoCss`
...
+	}
+`;
+
 type Props = {
 	sectionId?: string;
 	showSideBorders?: boolean;
 	showTopBorder?: boolean;
 	padded?: boolean;
 	backgroundColour?: string;
+	backgroundColourMobile?: string;
```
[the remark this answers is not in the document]
lucymonie: Thanks for this. I can see your point, but I'm equally concerned about the code duplication this represents. It's an additional 200+ lines of code in `LiveLayout` vs an additional optional prop in `ElementContainer`. It seems like a fairly extreme stance to take in order not to alter the interface of this component. Very happy to discuss though.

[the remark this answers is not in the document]
oliverlloyd: In my head the duplication would look a bit like this

```ts
<Hide above="leftCol">
	<ElementContainer
		backgroundColour={palette.background.article}
	>
		<SomeVerticalSpace />
	</ElementContainer>
</Hide>
<Hide below="leftCol">
	<ElementContainer
		backgroundColour={palette.background.someOtherProperty}
	>
		<SomeVerticalSpace />
	</ElementContainer>
</Hide>
```

The divider space is just empty so it could just be a plain div.

Is this possible? I'm very aware I don't have a complete picture of the use cases here!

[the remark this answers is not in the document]
marjisound: Just to understand your point here,
are you suggesting instead of having 2 props `backgroundColour` & `backgroundColourMobile` we can only have `backgroundColour` and the consumer decides the breakpoint colour by hiding and showing element in a specific breakpoint?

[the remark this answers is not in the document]
oliverlloyd: Yes, exactly as you describe.

The issue with adding `backgroundColourMobile` to the api is it makes it seem like having a different background colour at mobile is a 'thing' that the Guardian does. The intention for this component is to extend its usage, perhaps hoisting it into Source, so the simpler and easier to understand the props are the better

oliverlloyd (approved):
🚀

PR #3693, opened 2021-11-26
`SkipToKeyEvents`

oliverlloyd:
## What does this change?
This recognises the Key Events component as a navigation aid, wrapping it in a `nav` tag (with a descriptive aria label) and then adds an additional skip to link to aid discoverability.

## Why?
We've been adding more focus on Key Events recently and as part of that we want to make this navigation aid as accessible as possible


OllysCoding (approved):
lgtm

DavidLawes (approved):
👍

PR #3694, opened 2021-11-26
Add `maincontent` id for liveblogs

oliverlloyd:
## What does this change?
We didn't have a `maincontent` id for the Skip To Main Content accessibility link to use for live blogs, we do now

## Why?
So keyboard users can quickly access the main content of the page


ioannakok (approved):
👍

DavidLawes (approved):
nice!

PR #3697, opened 2021-11-29
fix: Only render MatchNav for MatchReports

oliverlloyd:
## What does this change?
Adds an additional check for when deciding if we should render `MatchNav`. In addition to looking for `CAPI.matchUrl` we now also check if `Design.MatchReport`

## Why?
An issue was raised in an email where some [articles](https://www.theguardian.com/football/2021/nov/28/chelsea-1-1-manchester-united-player-ratings-from-stamford-bridge) have a `matchUrl` value but are not match reports and because we try to add `MatchNav` when it isn't expected by the grid this causes the page layout to break

### Before
Screenshot 2021-11-29 at 08 16 51

### After
Screenshot 2021-11-29 at 08 15 24

### DIFF
--- a/dotcom-rendering/src/web/layouts/StandardLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/StandardLayout.tsx
@@ -97,7 +97,7 @@ const StandardGrid = ({
 									'title  border  matchNav     right-column'
 									'title  border  matchtabs    right-column'
 									'.      border  headline     right-column'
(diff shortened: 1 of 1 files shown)


rcrphillips (approved):

oliverlloyd:
[Football match-tabs displaying when they shouldn't](https://trello.com/c/uWYbLMrv/456-football-match-tabs-displaying-when-they-shouldnt)

PR #3698, opened 2021-11-29
Use PROD build for CI

oliverlloyd:
## What does this change?
This PR means we now run our CI steps for Cypress and `ampValidation` using the production version of our build, not the develop version.

To achieve this we had to fix the DCR `make build` and `make start` commands when running locally, which were not serving assets correctly.

`make build`
Bundles the code into a distributable form. Creates the `dist` folder from where the site can be served

`make start`
Starts a production version of the site on http://localhost:9000 serving the files in `dist`

## Why?
Mainly so that we're testing the thing that matters. But also because we saw some memory issues with the dev build and this should hopefully address those.

## What else?
As well as changes that directly relate to the things mentioned above, I also refactored some of our Cypress tests. Both the `braze` and `consent` spec files contained a similar pattern which was proving to be problematic.

After some fun investigation, it turned out that there was a problem with how we scripted interactions with the privacy banner. This code was causing issues:

```js
cy.get(theBanner).click('Accept All');
cy.reload();
```

When you accept on the privacy banner it makes an api call to get the cookie string that needs to be set to store preferences. But because we reloaded the page straight away this call and subsequent cookie write was not always completing. This problem was more pronounced with the production build of the code.

To address this we added a `cy.intercept` check for the api call and put a `cy.wait` in for that.


.github/workflows/cypress.yml
```diff
@@ -1,32 +1,34 @@
...
-                  browser: chrome
-                  spec: cypress/integration/parallel-${{ matrix.group }}/*.js
+      - name: Generate production build
+        run: make build
+        working-directory: dotcom-rendering
+
+      - name: Cypress run
+        uses: cypress-io/github-action@v2
+        with:
+          start: make start-ci
```
oliverlloyd, line 28: `make start-ci` is the same as `make start` but with logging turned off

.github/workflows/cypress.yml
```diff
@@ -1,32 +1,34 @@
...
+          working-directory: dotcom-rendering
+          wait-on: "http://localhost:9000"
+          wait-on-timeout: 240
```
oliverlloyd, line 31: We can probably reduce this now

dotcom-rendering/cypress/integration/parallel-1/article.elements.spec.js
```diff
@@ -148,6 +148,14 @@ describe('Elements', function () {
 		});
 
 		it('should render the interactive using a non-boot.js', function () {
+			cy.on('uncaught:exception', (err) => {
+				if (err.message.includes('window.require')) {
+					// This error is unrelated to the test in question so return  false to prevent
+					// this error from failing this test
+					return false;
+				}
+			});
```
oliverlloyd, line 157: This error seemed to only start appearing in prod mode? I suspect this interactive wasn't loading before

dotcom-rendering/cypress/integration/parallel-2/braze.spec.js
```diff
@@ -1,6 +1,5 @@
+import { storage } from '@guardian/libs';
```
oliverlloyd, line 1: I inlined the logic for this test to help with debugging

dotcom-rendering/cypress/integration/parallel-3/article.metaAndOphan.e2e.spec.js
```diff
@@ -103,11 +103,4 @@ describe('The web document renders with the correct meta and analytics elements
 		getExists(`[data-link-name="footer"]`);
 		getExists(`[data-link-name="footer : primary : Opinion"]`);
 	});
-
-	it('Sample of script tags have the correct attributes', function () {
-		getExists(`script[type="module"][src$="/assets/react.js"]`);
-		getExists(`script[defer][nomodule][src$="/assets/react.legacy.js"]`);
-		getExists(`script[type="module"][src$="/assets/ophan.js"]`);
-		getExists(`script[defer][nomodule][src$="/assets/ophan.legacy.js"]`);
-	});
```
oliverlloyd, line 112: I removed these tests because the script names are now 'react.SOME_ID.js' and if these scripts don't load we will catch lots of errors elsewhere in the tests

dotcom-rendering/cypress/integration/parallel-6/consent.spec.js
```diff
@@ -7,25 +7,14 @@ import { storage } from '@guardian/libs';
 const firstPage =
 	'https://www.theguardian.com/environment/2020/aug/01/plan-to-curb-englands-most-polluted-spot-divides-residents';
 
-const secondPage =
-	'https://www.theguardian.com/environment/2020/nov/19/blue-whale-sightings-off-south-georgia-raise-hopes-of-recovery';
-
```
oliverlloyd, line 12: Replaced with cy.reload()`

dotcom-rendering/cypress/integration/parallel-6/consent.spec.js
```diff
@@ -7,25 +7,14 @@ import { storage } from '@guardian/libs';
...
 describe('Consent tests', function () {
-	const waitForAnalyticsToInit = () => {
-		// Waiting is sad but we need to ensure the init script has executed which occurs
-		// after the pageLoadEvent has fired
-		// eslint-disable-next-line cypress/no-unnecessary-waiting
-		cy.wait(300);
-	};
```
oliverlloyd, line 19: Removed as redundant?

dotcom-rendering/cypress/integration/parallel-6/consent.spec.js
```diff
@@ -49,36 +38,20 @@ describe('Consent tests', function () {
...
-		cmpIframe().contains("It's your choice");
-		cmpIframe().find("[title='Manage my cookies']").click();
-		// Reject tracking cookies
-		privacySettingsIframe().contains('Privacy settings');
-		privacySettingsIframe().find("[title='Accept all']").click();
-		// Make a second page load now that we have the CMP cookies set to reject tracking and check
-		// to see if the ga property was set by Google on the window object
-		cy.visit(`Article?url=${secondPage}`);
-		waitForAnalyticsToInit();
-		cy.window().its('ga').should('exist');
```
oliverlloyd, line 82: This test is a duplicate of the first

dotcom-rendering/docs/contributing/detailed-setup-guide.md
```diff
@@ -89,16 +89,13 @@ You may want local identity cookies to be available in `dotcom-rendering`. To en
 
 -   `make build` creates production-ready bundles.
 -   `make start` starts the production server.
--   `make stop` stops the production server.
 
 You may need to run these with `sudo`
 
-Production environment uses `pm2`. More scripts can be found in the `makefile` [scripts](https://github.com/guardian/dotcom-rendering/blob/e2c020f7e0ed24751ea729eec93f1271d37e3b50/makefile#L31)
```
SiAdcock, line 96: For future travellers, is `pm2` no longer useful? I'd be wary of making changes to PROD as part of this PR, but maybe this one is small enough to sneak in?

dotcom-rendering/makefile
```diff
@@ -64,13 +50,13 @@ dev: clear clean-dist install
 
 percy: clear clean-dist install
 
-cypress: clear clean-dist install
-	$(call log, "starting frontend DEV server for Cypress")
-	@NODE_ENV=development start-server-and-test 'node scripts/frontend/dev-server' 3030 'cypress run --spec "cypress/integration/**/*"'
+cypress: clear clean-dist install build
```
SiAdcock, line 53: To be clear, is this just for testing locally? It looks like the Github Action runs Cypress in CI

dotcom-rendering/docs/contributing/detailed-setup-guide.md
oliverlloyd: I was a little too keen removing this. I see we are still using pm2

dotcom-rendering/makefile
oliverlloyd: Yes, `make cypress` is just for local execution as it doesn't allow for matrix execution

oliverlloyd:
[Move cypress builds to use production build rather than dev build](https://trello.com/c/I4l74qO1/457-move-cypress-builds-to-use-production-build-rather-than-dev-build)

dotcom-rendering/cypress.json
```diff
@@ -1,9 +1,21 @@
...
+		"*permutive.com",
+		"*adnxs.com",
+		"*adsystem.com",
+		"*casalemedia.com",
+		"*pubmatic.com",
+		"*360yield.com",
+		"*omnitagjs.com",
+		"*the-ozone-project.com",
+		"*openx.net"
+	],
```
oliverlloyd, line 19: I added these when debugging but maybe there's benefit to keeping them?

dotcom-rendering/makefile
oliverlloyd: Actually, it's also used by Team City in [ci-dcr.sh](https://github.com/guardian/dotcom-rendering/blob/356acb5ad77fb417202a8f9e7febdc6a854ccebf/scripts/ci-dcr.sh)

dotcom-rendering/scripts/webpack/browser.js
```diff
@@ -94,6 +94,7 @@ module.exports = ({ isLegacyJS }) => ({
 											},
 									  ],
 							],
+							compact: true,
```
oliverlloyd, line 97: This silences some warnings we saw in the console. See: https://stackoverflow.com/questions/29576341/what-does-the-code-generator-has-deoptimised-the-styling-of-some-file-as-it-e

dotcom-rendering/scripts/webpack/server.js
```diff
@@ -55,6 +53,7 @@ module.exports = () => ({
 									},
 								],
 							],
+							compact: true,
```
oliverlloyd, line 56: See: https://stackoverflow.com/questions/29576341/what-does-the-code-generator-has-deoptimised-the-styling-of-some-file-as-it-e

oliverlloyd:
Tested in `code` 👍

dotcom-rendering/cypress.json
OllysCoding: I think it makes sense if it could contribute to removing the variability of some cypress tests!

dotcom-rendering/makefile
```diff
@@ -35,24 +35,10 @@ build: clean-dist install
...
 
 start: install
-	@make stop
-	$(call log, "starting PROD server...")
-	@echo '' # just a spacer
-	@NODE_ENV=production pm2 start dist/frontend.server.js
-	@echo '' # just a spacer
-	$(call log, "PROD server is running")
-	@pm2 logs
+	@NODE_ENV=production node dist/frontend.server.js
```
OllysCoding, line 38: What happens in this situation now if the node server unexpectedly crashes? Under pm2 it would restart automatically. Presumably the ASG / LB will detect the instance is unhealthy & kill it, but it would be nice to be sure.

oliverlloyd: Don't we start the (actual, live) [server here](https://github.com/guardian/dotcom-rendering/blob/oliver/relative-time/dotcom-rendering/cloudformation.yml#L267)

That's still using pm2

oliverlloyd: In any case, perhaps there's a case for restoring this usage of pm2 locally? I'm happy to do this

PR #3700, opened 2021-11-29
Remove Prebid Timeout AB test

chrislomaxjones:
## What does this change?

Remove the Prebid timeout AB test that was introduced in #3566.

See the corresponding PR in frontend: [Frontend #24445](https://github.com/guardian/frontend/pull/24445)

## Why?

The test has now concluded so it is no longer required.


simonbyford (approved):
🧹

PR #3704, opened 2021-11-30
Relative time

oliverlloyd:
## What does this change?
This PR adds support for relative time on DCR.

There has been previous work in this space [here where we added](https://github.com/guardian/libs/pull/134) timeAgo to @guardian/libs and also [here where the first proposal](https://github.com/guardian/dotcom-rendering/pull/3426) was made to introduce the use of this lib in DCR.

https://user-images.githubusercontent.com/1336821/144039588-8d34eaf3-2199-408c-bd88-08de966af174.mov

(Note. The interval in the video above is artificially fast to keep you entertained)

## Usage
Any `time` element with [a valid datetime](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) (`dateTime` in React) and also a `data-relativeformat` attribute will be kept up to date.

`datetime`: Must be a [valid datetime value](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time#valid_datetime_values)
`data-relativeformat`: 'short' | 'med' | 'long'*

*short and med are the same but we respect 'short' to honour older contracts.

## Why?
A convention was started at the Guardian 9 years ago in the [38th PR on Frontend](https://github.com/guardian/frontend/pull/38) saying that `time` elements should be able to show 'relative' time. That contract today states that if a `time` element has both `datetime` and `data-relativetime` attributes then the page should respect the contract this establishes and ensure this element shows a relative time string, such as `5m ago`.

## What about `useTimeAgo`?
Previous work in this area leaned on React to make time elements relative but in this PR we just use vanilla javascript because the action of setting an interval to read and mutate the dom is a lot simpler to achieve outside of react.

### DIFF
--- a/common-rendering/src/components/FirstPublished.tsx
+++ b/common-rendering/src/components/FirstPublished.tsx
@@ -32,6 +32,7 @@ const FirstPublished = ({
 		>
 			<time
 				dateTime={publishedDate.toISOString()}
+				data-relativeformat="med"
 				css={css`
 					color: ${neutral[46]};
 					font-weight: bold;
--- a/dotcom-rendering/scripts/test/build-check.js
+++ b/dotcom-rendering/scripts/test/build-check.js
@@ -51,6 +51,7 @@ const fileExists = async (glob) => {
 		'atomIframe',
 		'embedIframe',
 		'newsletterEmbedIframe',
+		'relativeTime',
 	].map((name) => {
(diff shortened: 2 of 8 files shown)


oliverlloyd:
[Relative Time](https://trello.com/c/DaBEUEpH/459-relative-time)

OllysCoding (approved):
LGTM

PR #3705, opened 2021-11-30
AR Live Layout With Grid

JamieB-gu:
## Why?

We want the liveblog/deadblog layout to match the designs - this brings us closer to that.

### Why Grid?

The designs change quite a lot between different breakpoints: elements change their order and position within the page. The only way I could get this to work without lots of `display: none`s was to use Grid.

### What's left?

So far the colours are only correct for news liveblogs - in a follow-up we need to extend support for other themes and for deadblogs and dark mode.

Also, I've added a stripped-down version of DCR's `GridItem` to AR. Perhaps we should consider lifting DCR's component to common-rendering in a follow-up? cc @oliverlloyd 

## Changes

- Added `GridItem` component
- Simplified `HeaderImage` styles - relied on `Sizes`
- Pulled metadata and main media out of header
- Fixed up metadata width and padding across breakpoints
- Added semantic HTML to live layout (`article`, `main`)
- Used CSS grid to layout live pages
- Added a TODO for live blocks
- Fixed accordion padding (should jump up at `mobileLandscape`)

## Screenshots

| | Before | After |
| - | - | - |
| Mobile | liveblog-before-mobile | liveblog-mobile |
| Mobile Landscape | liveblog-before-mobilelandscape | liveblog-mobilelandscape |
| Tablet | liveblog-before-tablet | liveblog-tablet |
| Desktop | liveblog-before-desktop | liveblog-desktop |


marjisound (approved):
Looks great 👍

PR #3708, opened 2021-11-30
Create a new WazuhSecurityGroup with another name

waisingyiu:
## What does this change?

We found that there was name clash in the resource "WazuhSecurityGroupNew" between the GuCDK and the one Apps-rendering is currently using.  We cannot change the name of resources in GuCDK so we are changing the name of WazuhSecurityGroup currently used by Apps-rendering in two steps:

1) Create a new WazuhSecurityGroup with a new name "WazuhSecurityGroupNew"

2) After new instances are created with the new WazuhSecurityGroupNew, we remove the WazuhSecurityGroup (the one with the old name)

This PR is the step 1 

## Why?

To resolve the name clash

### Before

The WazuhSecurityGroup resource has the name WazuhSecurityGroup

### After

A new resource "WazuhSecurityGroupNew" is created and it is used in the launcher script instead of the "WazuhSecurityGroup".

### DIFF
--- a/apps-rendering/config/cloudformation.yaml
+++ b/apps-rendering/config/cloudformation.yaml
@@ -70,7 +70,7 @@ Parameters:
     Description: A security group id for editions backend preview lambda to access AR
   EditionsBackendPublishedLambdaSecurityGroupId:
     Type: String
-    Description: A security group id for editions backend published lambda to access AR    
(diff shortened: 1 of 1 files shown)


frankie297 (approved):
LGTM 👍

PR #3709, opened 2021-11-30
Show AMP mobile sticky ad slot in all regions

zekehuntergreen:
## What does this change?
Screenshot 2021-11-30 at 16 25 47

- Shows AMP mobile sticky ad slot in all regions.
- Refactors `Ad`, `RegionalAd`, `StickyAd` to make them simpler and their intent more clear given the change in requirements.
   - Switch the names of `Ad` and `RegionalAd` so that Ad is the base components. `RegionalAd` calls `Ad`, as does `StickyAd`.
   - Put the regional ad component and associated code into its own file, `RegionalAd.tsx`
   - `Ad` takes an `rtcConfig` parameter which is calculated by `RegionalAd` and `StickyAd` separately.

### Before
Currently the ad slot is only shown in the US. It was set up this way due to a constraint imposed by AMP; a page with more than one `amp-sticky-ad` fails AMP validation. This prevents us from regionalizing the sticky ad as we do for other ad slots, by including US, UK, and ROW slots, then using [css](https://github.com/guardian/dotcom-rendering/blob/150fc2d81e6a66d9c3336185e874fc8cd0288546/dotcom-rendering/src/amp/lib/region-classes.ts#L1) to hide all but the relevant one.

### After
The same `amp-sticky-ad` is shown, regardless of the user's region.

## Why?
We'd like to increase ad revenue by showing ads in the ad slot to users outside of the US.


dotcom-rendering/src/amp/components/Ad.tsx
```diff
@@ -50,12 +50,11 @@ const amazonConfig = {
  * @returns The placement id for an ad, depending on its ad region and whether
  * it is sticky
  */
-const getPlacementId = (isSticky: boolean, adRegion: AdRegion): number => {
+const getPlacementId = (isSticky: boolean, adRegion?: AdRegion): number => {
+	if (isSticky) return 9;
```
zekehuntergreen, line 54: This is the placement ID of the new AMP config created on prebid server.

dotcom-rendering/src/amp/components/Ad.tsx
```diff
@@ -121,7 +120,7 @@ export interface AdProps {
 }
 
 export interface RegionalAdProps extends AdProps {
-	adRegion: AdRegion;
+	adRegion?: AdRegion;
 }
 
 export const RegionalAd = ({
```
chrislomaxjones, line 126: Given `adRegion` is now an optional prop we could change the name of the `RegionalAd` component (and `RegionalAdProps`), as at the moment it implies it probably always has a region? `Ad` would probably suffice but it'd require the existing `Ad` component be named something more specific?

chrislomaxjones (approved):
Awesome! I had one small comment about component naming but I don't think its a blocker.

I also wonder if there's somewhere we do/should commit the new stored bid request on the Prebid server?

dotcom-rendering/src/amp/components/Ad.tsx
arelra: Snap :) @zekehuntergreen and I discussed along similar lines.

We could swap the hierarchy so there is now a base 'Ad' Component which the `RegionalAd` will loop over, whereas `StickyAd` will call the base `Ad` directly.

chrislomaxjones: I like that!

dotcom-rendering/src/amp/components/StickyAd.tsx
```diff
@@ -1,9 +1,7 @@
-import { ClassNames } from '@emotion/react';
 import { text, textSans } from '@guardian/source-foundations';
-import { regionClasses } from '../lib/region-classes';
-import { RegionalAdProps, RegionalAd } from './Ad';
+import { realTimeConfig, RegionalAdProps as StickyAdProps, Ad } from './Ad';
```
zekehuntergreen, line 2: I'm not sure I'm convinced that this is the best thing to do. I'm trying to avoid just using `RegionalAdProps` for `StickyAd `. It might be clearer to think of a new name for `RegionalAdProps` that encompasses both the regional and sticky use case. 🤔

arelra: Yeah feels like we could narrow this a bit. Is there a base props that `RegionalAdProps` and `StickyAdProps` can both extend from?

zekehuntergreen: How about [this](https://github.com/guardian/dotcom-rendering/pull/3709/commits/1d4fab7e7cfd3d9ea09b7a75ae4467f048b06385#diff-e161fa864f5d2b3e59703455fad31b8932f8cfc0dde517adc47ee45413ba2b2aR123)?

dotcom-rendering/src/amp/components/Ad.tsx
```diff
@@ -111,28 +104,34 @@ interface CommercialConfig {
...
 }
 
-export const RegionalAd = ({
+export interface RegionalAdProps extends BaseAdProps {
+	config: CommercialConfig;
+}
+
+export interface StickyAdProps extends BaseAdProps {
+	config: CommercialConfig;
+}
```
chrislomaxjones, line 125: This is a nitpick but can we move this definition to the StickyAd file since that's where it's used?

dotcom-rendering/src/amp/components/StickyAd.tsx
chrislomaxjones: I think that makes a lot of sense 👍

dotcom-rendering/src/amp/components/Ad.tsx
zekehuntergreen: Good call!

arelra (approved):

dotcom-rendering/src/amp/components/StickyAd.tsx
```diff
@@ -17,37 +15,34 @@ amp-sticky-ad:before {
 	font-size: 0.75rem;
 }`;
 
+const mobileStickyPlacementId = 9;
+
+export interface StickyAdProps extends BaseAdProps {
```
arelra, line 20: Does this need to be exported now?

zekehuntergreen: Nope, good call

PR #3716, opened 2021-12-02
Apps rendering liveblocks skeleton

DavidLawes:
## What does this change?
This PR is the first part of implementing live blocks for AR. In this PR we add the basic skeleton of liveblocks. We use only pre-existing shared components and do not attempt to extract possibly-sharable components into common-rendering. Extracting and using shared logic will be dealt with under a separate task.

While this PR adds the skeleton for liveblocks we envisage the following changes:
- Format lastUpdatedDisplay according to the apps-rendering standard (we will need a new date format function in date.ts)
- Add pagination
- Wrap liveblocks in the accordion component
- Styling changes to achieve parity with the design

https://trello.com/c/3PIfKTMB

## Why?
This is the first step to supporting deadblogs in apps-rendering.

### Before
Apps-rendering deadblogs had no liveblocks:
Screenshot 2021-12-02 at 09 26 57

### After
Apps-rendering deadblogs have liveblocks:
Screenshot 2021-12-02 at 09 29 18


apps-rendering/src/components/liveBlocks.tsx
```diff
@@ -0,0 +1,97 @@
...
+						>
+							{block.firstPublished.kind === OptionKind.Some && (
+								<FirstPublished
+									firstPublished={Number(
+										block.firstPublished.value,
+									)}
+									blockLink={blockLink}
+								/>
+							)}
+							{block.title && <BlockTitle title={block.title} />}
```
JamieB-gu, line 65: Is this a guard against empty strings? This is subjective because I know the `&&` is a common JS pattern, but can I suggest making this check explicit for clarity? The `&&` causes a couple of implicit conversions that might be considered harder to reason about 🤔 

```suggestion
							{block.title !== '' ? <BlockTitle title={block.title} /> : null}
```

apps-rendering/src/components/liveBlocks.tsx
```diff
@@ -0,0 +1,97 @@
...
+							`}
+						>
+							{block.firstPublished.kind === OptionKind.Some && (
+								<FirstPublished
+									firstPublished={Number(
+										block.firstPublished.value,
+									)}
+									blockLink={blockLink}
+								/>
+							)}
```
JamieB-gu, line 64: There's a helper function in `lib.ts` that you may find useful here (_may_, it depends what you find more readable):

```suggestion
							{maybeRender(block.firstPublished, firstPublished =>
								<FirstPublished
									firstPublished={Number(firstPublished)}
									blockLink={blockLink}
								/>
							)}
```

Couple more thoughts on this:

1. Do you think it's valid to render a block if `firstPublished` is missing? Should we actually consider invalidating the whole block if we don't know when it was published? If so, this is probably something we'd want to do a step earlier during parsing 🤔 
2. Would it make sense for the `FirstPublished` component to take a `Date` instead? I've noticed that internally it's converting the `number` into a `Date` anyway. Also it's a component named after a datetime so semantically I think it makes sense?

apps-rendering/src/components/richLink.tsx
```diff
@@ -171,7 +179,12 @@ const RichLink = (props: {
 		: {};
 
 	const attributes = {
-		css: richLinkStyles(format),
+		css: [
+			richLinkStyles(format),
+			(format.design === ArticleDesign.LiveBlog ||
+				format.design === ArticleDesign.DeadBlog) &&
+				liveBlogRichLinkStyles,
+		],
```
JamieB-gu, line 187: A possible alternative to this is a common pattern we've used in AR when swapping styles between different `Format`s:

```ts
const styles = (format: Format): SerializedStyles => {
  switch (format.design) {
    case ArticleDesign.LiveBlog:
    case ArticleDesign.DeadBlog:
      return css(richLinkStyles(format), liveBlogRichLinkStyles);
    default:
      return richLinkStyles(format);
  }
}
```

It's relatively simple to extend if we want to drop in extra styles for other `Design`s.

JamieB-gu (commented):
This is great! 🎉

A few comments about platform conventions and a couple of suggestions.

apps-rendering/src/components/liveBlocks.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { headline, space } from '@guardian/source-foundations';
+import { OptionKind, partition } from '@guardian/types';
+import type { DeadBlog, LiveBlog } from 'item';
+import type { FC } from 'react';
+import { renderAll } from 'renderer';
+
+// ----- Component ----- //
+interface LiveBlocksProps {
+	item: DeadBlog | LiveBlog;
+}
```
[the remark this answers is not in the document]
DavidLawes: Nice - will update :)

apps-rendering/src/components/liveBlocks.tsx
```diff
@@ -0,0 +1,97 @@
...
+interface BlockTitleProps {
+	title: string;
+}
+
+const BlockTitle: FC<BlockTitleProps> = ({ title }) => {
+	return (
+		<h2
+			css={css`
+				${headline.xxsmall({ fontWeight: 'bold' })}
+				margin-bottom: ${space[2]}px;
```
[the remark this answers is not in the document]
DavidLawes: 👍 

We wondered whether `BlockTitle` would become a shared component, if it does then shared components use remSpace or space?

apps-rendering/src/components/liveBlocks.tsx
```diff
@@ -0,0 +1,97 @@
...
+			`}
+		>
+			{title}
+		</h2>
+	);
+};
+
+const LiveBlocks: FC<LiveBlocksProps> = ({ item }) => {
+	return (
+		<div>
```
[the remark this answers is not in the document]
DavidLawes: It could (I think). However, we made this a `div` because we think we may add some additional attributes to this root container for all liveblocks. The attributes (considering what's been done for dcr) could be: tabIndex (if we consider accessibility in the same way for mobile as we do for web?), an id of "maincontent" and potentially some global styles

Given the rationale above, do you think we could leave as a div for now (at least until we've considered the styling of liveblocks)?

apps-rendering/src/components/liveBlocks.tsx
```diff
@@ -0,0 +1,97 @@
...
+			{/* Pagination? */}
+			{item.blocks.map((block) => {
+				// TODO: get page number
+				const blockLink = `${1}#block-${block.id}`;
+
+				return (
+					<LiveBlockContainer
+						key={block.id}
+						id={block.id}
+						borderColour={'black'}
```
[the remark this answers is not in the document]
DavidLawes: I'll change this, thanks. Just to point out that this definitely isn't the intended borderColour. The design needs to vary depending on the pillar, which we skipped over as part of this skeleton design :)

apps-rendering/src/components/liveBlocks.tsx
```diff
@@ -0,0 +1,97 @@
...
+					>
+						<header
+							css={css`
+								padding-right: ${space[3]}px;
```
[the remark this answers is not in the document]
DavidLawes: 👍

apps-rendering/src/components/liveBlocks.tsx
DavidLawes: Please always make a suggestion :D In this case I'm not sure, possibly the `&&` pattern is more in-keeping with react? https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

Will think about this one a little more :)

JamieB-gu: I don't think React is opinionated about this. That part of the docs talks more broadly about the fact there are a few different ways to inline conditionals - the heading below the one you linked talks about using ternary for example. It also finishes with this phrase 🙂:

> Just like in JavaScript, it is up to you to choose an appropriate style based on what you and your team consider more readable

It's up to us! 😅 

Actually the last example in the `&&` section demonstrates one of the issues I'm talking about. Doing `a && b` implicitly converts `a` to a boolean, and returns `b` if that boolean is `true`. But if that boolean is `false` it returns `a`. In our code above, if `block.title` is an empty string then you're _rendering an empty string_, you're not rendering nothing.

In modern React I _think_ the visual result is the same in this case, i.e. nothing is inserted into the DOM. However, the recommended way to render nothing is `null` as described by that same docs page. Furthermore, in other circumstances the result may _not_ be nothing inserted into the DOM - as in the example, if this were a number then `0` would get rendered, which may not be what you want to happen.

apps-rendering/src/components/liveBlocks.tsx
DavidLawes: I really like `maybeRender` - thanks for highlighting this :)

1) Maybe I've missed your point, are we suggesting not showing a blog post to a reader if we don't know the firstPublished date? My initial feeling is: show as much content as possible, as long as the content itself can be successfully rendered. However, maybe there are legal requirements around displaying the published dateTime that I'm not aware of (or some other consideration entirely).

I think in dcr we perform equivalent logic and so we've, perhaps lazily, replicated equivalent functionality here in the hope that we'd be able to share both the Header and its children (https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/LiveBlock.tsx#L74)

2) About the props for `FirstPublished` - totally agree, when we were pairing we found it frustrating to convert the prop to `Number`, only for it to be converted back to a `Date` inside the component. 

For this skeleton PR I'm tempted to leave as it is - we have a specific checklist item to review shared components and I've included an extra note to review `FirstPublished` props as part of this (https://trello.com/c/3PIfKTMB). Given we have a task to review the existing shared components and their props do you think it's acceptable to leave as-is for now? I think it might be 'cleaner' to raise a separate PR to tidy up these props as it would impact dcr too. Let me know what you think about this approach :)

apps-rendering/src/components/richLink.tsx
DavidLawes: Nice - will update :)

apps-rendering/src/components/liveBlocks.tsx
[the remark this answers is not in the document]
JamieB-gu: I think they'd need to use `remSpace` because AR expects sizes to be in REMs for scaling purposes.

apps-rendering/src/components/liveBlocks.tsx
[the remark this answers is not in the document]
JamieB-gu: > if we consider accessibility in the same way for mobile as we do for web?
I think we should, yes.

> Given the rationale above, do you think we could leave as a div for now
In general I'd be inclined to leave it as a fragment until you're doing the work to add attributes, and change it to a `div` when you know you need it. Just in case you don't end up doing that work and it ends up staying as a `div` with nothing attached. But if you're confident that it will change then I don't think it matters too much 🙂

apps-rendering/src/components/liveBlocks.tsx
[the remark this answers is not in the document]
DavidLawes: Great - thanks! Will make sure we use `remSpace` in common too

apps-rendering/src/components/liveBlocks.tsx
DavidLawes: Just to follow up on this - I understand your point and looking through apps-rendering I think this pattern is more widely used. I decided to change to your suggestion :)

apps-rendering/src/components/liveBlocks.tsx
JamieB-gu: > are we suggesting not showing a blog post to a reader if we don't know the firstPublished date?

That's what I think we should determine, yes. Perhaps we should have a chat about this?

> Given we have a task to review the existing shared components and their props do you think it's acceptable to leave as-is for now?

Yup sounds reasonable.

apps-rendering/src/components/liveBlocks.tsx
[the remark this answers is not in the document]
DavidLawes: Cool - have changed to a fragment for now!

apps-rendering/src/components/liveBlocks.tsx
DavidLawes: We discussed this problem and I think we agreed:
- it would be simpler to only parse valid blocks (e.g. in liveBlock.ts) - rejecting any blocks that do not have a firstPublished or lastModified date. Then in the layout file our rendering logic could be simplified because we'd only have valid blocks (this means we could remove some of the maybeRender functions).
- I have added this task as a new checklist item into the AR liveblocks card so we can address this under a separate PR (https://trello.com/c/3PIfKTMB)

If the above sounds reasonable @JamieB-gu are you happy to approve?

JamieB-gu (approved):

PR #3719, opened 2021-12-03
Fix styling for header media in liveblogs / deadblogs

ioannakok:
…t breakpoints

### DIFF
--- a/common-rendering/src/components/imageDetails.tsx
+++ b/common-rendering/src/components/imageDetails.tsx
@@ -2,7 +2,7 @@
 
 import type { SerializedStyles } from "@emotion/react";
 import { css } from "@emotion/react";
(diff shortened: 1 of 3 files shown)


common-rendering/src/components/imageDetails.tsx
```diff
@@ -58,6 +58,10 @@ const iconStyles = (supportsDarkMode: boolean): SerializedStyles => css`
 	${darkModeCss(supportsDarkMode)`
 		background-color: ${neutral[93]};
 	`}
+
+	${between.mobileLandscape.and.tablet} {
+		right: 20px;
```
lucymonie, line 63: I think this can be `right: ${remSpace[5]};`

lucymonie (approved):
One small suggestion, looks good 👍🏻

PR #3721, opened 2021-12-06
YouTube Lite using `@guardian/atoms-rendering@22.1.1`

oliverlloyd:
## What does this change?
This PR adds version 22.1.1 of @guardian/atoms-rendering, bringing with it the new 'lite' version of `YouTubeAtom`. YouTube elements will now only download the youtube.com files upon interaction.

See: https://github.com/guardian/atoms-rendering/pull/303

To accommodate this new version of the YouTube atom, I've refactored the Cypress tests to be more resilient. In particular, I've changed `HydrateOnce` to add a marker attribute (`data-cy-hydrated`) to the root element when its contents have been hydrated. This allows Cypress to know when it should expect content to be ready to interact with.


nitro-marky (approved):

dotcom-rendering/cypress/integration/parallel-5/atom.video.spec.js
```diff
@@ -1,98 +1,183 @@
...
+	it('plays main media videos', function () {
+		cy.visit(
+			'/Article?url=https://www.theguardian.com/uk-news/2020/dec/04/edinburgh-hit-by-thundersnow-as-sonic-boom-wakes-residents',
+		);
 		cmpIframe().contains("It's your choice");
 		cmpIframe().find("[title='Manage my cookies']").click();
 		privacySettingsIframe().contains('Privacy settings');
-		privacySettingsIframe().find("[title='Accept all']").click();
+		privacySettingsIframe()
+			.find("[title='Accept all']", { timeout: 12000 })
```
mxdvl, line 36: Nifty little formatting syntax which makes large numbers easier to parse:

```suggestion
			.find("[title='Accept all']", { timeout: 12_000 })
```

dotcom-rendering/src/web/components/HydrateOnce.tsx
```diff
@@ -51,6 +51,7 @@ export const HydrateOnce = ({ rootId, children, waitFor = [] }: Props) => {
 	if (!element) return null;
 	start();
 	ReactDOM.hydrate(children, element, () => {
+		element.setAttribute('data-cy-hydrated', 'true');
```
mxdvl, line 54: Is this too Cypress-specific? If we add this to all elements that are hydrated, maybe we want to make it more Guardian-specific, like `data-gu-hydrated`?

mxdvl (approved):
Great improvement! It would be great to compare the CWV impact of this change.

Very minor thoughts below.

dotcom-rendering/src/web/components/HydrateOnce.tsx
oliverlloyd: Ha. Yeah, I went back and forth on this. In the end I realised I can't think of a use case for where this would be used outside of Cypress so it felt right to use the `data-cy-*` convention. But I'm still on the fence. One side of my brain is telling me, but you might want to use this elsewhere in future! And the other half is telling me to only make code changes based on what you need right now 🤷 

All that said, I think that you make a valid point which I haven't considered which is that we're _always_ adding this attribute, not just for you tube elements, so lets go with `data-gu-hydrated` 👍

mxdvl: I feel the struggle, but I for one would find it helpful to be able to tell from bringing up the inspector whether an element has been hydrated. Seeing `gu` makes me feel at home and that it is something _we intended_. Seeing `cy` would not give me the same _feeling_, or make me think it’s done by Cypress.

oliverlloyd: I think you're right. As an aside, the reason we need to use a marker like this is because we're implementing 'islands' of interactivity, instead of hydrating the whole page. If we just did the classic `React.hydrate(TheWholePage)` then you'd be able to tell when hydration has completed with a simple `useEffect` / checking for the UI changes. But with islands things complete at different points and this was causing flakiness in the tests.

dotcom-rendering/cypress/integration/parallel-5/atom.video.spec.js
oliverlloyd: I implemented this as it's a nice feature but I had to revert as it was throwing lint errors 😞

PR #3722, opened 2021-12-06
Refactor `document`

oliverlloyd:
## What does this change?
This is a PR to refactor the way we server side render the page. This is mostly tidying up and moving things around.

- `Page` no longer takes `children`
- `Page` is responsible for `SkipTo` links
- Removed unused `ccsIDs`
- Refactored SSR using Emotion to follow recommendations

## Why?
We were using [an older method](https://emotion.sh/docs/ssr#when-using-emotioncss) for SSR with Emotion based on @emotion/css but now that we're using @emotion/react we can use the [recommended approach](https://emotion.sh/docs/ssr#when-using-emotionreact).

This allows us to remove the unused `cssIDs` property, reducing the bundle size, and also simplifies the code by aligning us with the standards for Emotion.

### DIFF
--- a/dotcom-rendering/src/web/components/Page.tsx
+++ b/dotcom-rendering/src/web/components/Page.tsx
@@ -1,47 +1,41 @@
 import React, { StrictMode } from 'react';
-import { EmotionCache } from '@emotion/cache';
-import { CacheProvider, Global, css } from '@emotion/react';
+import { Global, css } from '@emotion/react';
 import { focusHalo } from '@guardian/source-foundations';
 import { SkipTo } from '../components/SkipTo';
 import { ArticleDesign } from '@guardian/libs';
 
 type Props = {
 	children: React.ReactNode;
-	cache: EmotionCache;
 	format: ArticleFormat;
 };
 
 /**
  * @description
- * Page is a high level wrapper for pages on Dotcom. Sets strict mode, some globals
- * and the Emotion cache
+ * Page is a high level wrapper for pages on Dotcom. Sets strict mode and some globals
  *
  * @param {ReactNode} children - What gets rendered on the page
- * @param {EmotionCache} cache - Provides the Emotion cache
  * @param {ArticleFormat} format - The format model for the article
  * */
-export const Page = ({ children, cache, format }: Props) => {
+export const Page = ({ children, format }: Props) => {
 	return (
-		<CacheProvider value={cache}>
-			<StrictMode>
-				<Global
-					styles={css`
-						/* Crude but effective mechanism. Specific components may need to improve on this behaviour. */
-						/* The not(.src...) selector is to work with Source's FocusStyleManager. */
-						*:focus {
-							${focusHalo}
-						}
-					`}
-				/>
-				<SkipTo id="maincontent" label="Skip to main content" />
-				<SkipTo id="navigation" label="Skip to navigation" />
-				{format.design === ArticleDesign.LiveBlog ||
-					(format.design === ArticleDesign.DeadBlog && (
-						<SkipTo id="keyevents" label="Skip to key events" />
-					))}
-				<div id="react-root"></div>
-				{children}
-			</StrictMode>
-		</CacheProvider>
+		<StrictMode>
+			<Global
+				styles={css`
+					/* Crude but effective mechanism. Specific components may need to improve on this behaviour. */
+					/* The not(.src...) selector is to work with Source's FocusStyleManager. */
+					*:focus {
+						${focusHalo}
+					}
+				`}
+			/>
+			<SkipTo id="maincontent" label="Skip to main content" />
+			<SkipTo id="navigation" label="Skip to navigation" />
+			{format.design === ArticleDesign.LiveBlog ||
+				(format.design === ArticleDesign.DeadBlog && (
+					<SkipTo id="keyevents" label="Skip to key events" />
+				))}
+			<div id="react-root"></div>
+			{children}
+		</StrictMode>
 	);
 };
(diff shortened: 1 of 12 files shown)


dotcom-rendering/src/web/server/htmlTemplate.ts
```diff
@@ -286,15 +280,12 @@ https://workforus.theguardian.com/careers/product-engineering/
                 ${loadableConfigScripts.join('\n')}
                 ${priorityScriptTags.join('\n')}
                 <style class="webfont">${getFontsCss()}</style>
-                <style>${resets.resetCSS}${css}</style>
+                <style>${resets.resetCSS}</style>
+				${css}
```
oliverlloyd, line 284: The return for `css` is now a style tag, not a css string.

dotcom-rendering/src/web/components/Page.tsx
```diff
@@ -1,36 +1,44 @@
...
+						${focusHalo}
+					}
+				`}
+			/>
+			<SkipTo id="maincontent" label="Skip to main content" />
+			<SkipTo id="navigation" label="Skip to navigation" />
+			{format.design === ArticleDesign.LiveBlog ||
+				(format.design === ArticleDesign.DeadBlog && (
+					<SkipTo id="keyevents" label="Skip to key events" />
+				))}
```
ioannakok, line 39: I'm not super familiar with the JS syntax or the requirement here, just double checking: Is this equivalent to the previous syntax?

`if (
		CAPI.format.design === 'LiveBlogDesign' ||
		CAPI.format.design === 'DeadBlogDesign'
	) {
		skipToKeyEvents = renderToString(
			<SkipTo id="keyevents" label="Skip to key events" />,
		);
	}`

oliverlloyd: Basically, yes. The conditional is easier to read in classic javascript vs. JSX and it doesn't help that prettier has added those brackets in such a weird way 🤔

ioannakok (approved):
👍

dotcom-rendering/src/web/components/Page.tsx
```diff
@@ -1,36 +1,44 @@
...
+			<Global
+				styles={css`
+					/* Crude but effective mechanism. Specific components may need to improve on this behaviour. */
+					/* The not(.src...) selector is to work with Source's FocusStyleManager. */
+					*:focus {
+						${focusHalo}
+					}
+				`}
+			/>
+			<SkipTo id="maincontent" label="Skip to main content" />
```
OllysCoding, line 34: Is there a reason that prior to this each of these items were rendered out on their own outside of the main (p)react scope? Just want to make sure this doesn't have any unintentional accessibility side-effects.

dotcom-rendering/src/web/layouts/InteractiveImmersive.stories.tsx
```diff
@@ -40,6 +45,11 @@ const convertToInteractiveImmersive = (CAPI: CAPIType) => {
 const HydratedLayout = ({ ServerCAPI }: { ServerCAPI: CAPIType }) => {
 	fireAndResetHydrationState();
 	const NAV = extractNAV(ServerCAPI.nav);
+	const format: ArticleFormat = {
+		display: decideDisplay(ServerCAPI.format),
+		design: decideDesign(ServerCAPI.format),
+		theme: decideTheme(ServerCAPI.format),
+	};
```
OllysCoding, line 52: It really feels like there should be a single `decideFormat` function that takes care of this?

dotcom-rendering/src/web/components/Page.tsx
oliverlloyd: I don't think there was a specific reason. They just needed to be rendered at a very high level and the `Page` component didn't exist at that point.

They still work in `code`

dotcom-rendering/src/web/server/htmlTemplate.ts
```diff
@@ -286,15 +280,12 @@ https://workforus.theguardian.com/careers/product-engineering/
...
 				<link rel="stylesheet" media="print" href="${CDN}static/frontend/css/print.css">
 			</head>
 
 			<body>
-				${skipToMainContent}
-				${skipToNavigation}
-				${skipToKeyEvents || ''}
-                <div id="react-root"></div>
```
OllysCoding, line 297: What was this div doing before? We weren't putting the `html` string into it, so it would be nice to know why it can now be removed (or if it was used beforehand)

dotcom-rendering/src/web/layouts/InteractiveImmersive.stories.tsx
oliverlloyd: I very much had the same thought when I was writing this code!

dotcom-rendering/src/web/server/htmlTemplate.ts
oliverlloyd: This div has moved into `Page`. It's used as the place where the client side React app we run is hooked into - in BootReact

OllysCoding: Ah yes thank you, I missed that one :)

OllysCoding (approved):
+1 :)

dotcom-rendering/src/web/components/Page.tsx
oliverlloyd: > They still work in code

I lied. I checked again and _dead_ blogs were working but not liveblogs. Thanks for raising this code as suspicious 🙏 🙏 🙏 

I needed to move the brackets around manually to prevent prettier adjusting the way the logic worked

PR #3724, opened 2021-12-07
AR - show liveblocks last updated time

DavidLawes:
## What does this change?
When we implemented the skeleton for liveblocks on AR (https://github.com/guardian/dotcom-rendering/pull/3716) we omitted the ability to display the last updated time of a liveblock. This PR renders the correct format of the liveblock lastUpdated time.

Thanks to @OllysCoding this PR includes the client side bundle for live and dead blogs. Therefore, on client side, lastUpdated time is updated according to the reader's local time zone.

## Why?
The last updated time of a liveblock is a requirement of the design.

### Before
We'd hardcoded the last updated time:
Screenshot 2021-12-07 at 12 03 07

### After
Now we render the last updated time on server side and it gets updated on client side:
Screenshot 2021-12-07 at 12 04 00

This includes reformatting for a user's local timezone:
Screenshot 2021-12-14 at 11 33 47


common-rendering/src/components/LastUpdated.tsx
```diff
@@ -17,8 +17,11 @@ const LastUpdated = ({
 				color: ${neutral[46]};
 			`}
 		>
-			<time dateTime={new Date(lastUpdated).toISOString()}>
-				{`Updated at ${lastUpdatedDisplay}`}
+			<time
+				data-last-updated={lastUpdated.toISOString()}
+				dateTime={lastUpdated.toISOString()}
+			>
+				{`Updated: ${lastUpdatedDisplay}`}
```
ioannakok, line 24: Looks good to me! I only have one observation / question: 
I was checking the Storybook diffs and since this is a common component, this change affects how DCR liveblocks look but (as expected) these don't have the new last updated time formatting, e.g.: https://www.chromatic.com/test?appId=5dfcbf3012392c0020e7140b&id=61b207a3b26af2003a53193f

1. Are we okay with this?
2. Do we want to introduce the new last updated time format in DCR liveblocks as a separate task later?

DavidLawes: Ah, great point :) I think we would want to update how this displays on DCR too. Currently it's formatted in Frontend (https://github.com/guardian/frontend/blob/main/common/app/model/dotcomrendering/DotcomRenderingSupportTypes.scala#L84). I'll create a card to cover this requirement before this PR is merged :)

DavidLawes: NB: I've created a card in the backlog to cover this: https://trello.com/c/4GgN73e9

dotcom-rendering/src/web/components/LiveBlock.tsx
```diff
@@ -111,7 +111,7 @@ export const LiveBlock = ({
 					block.blockLastUpdated &&
 					block.blockLastUpdatedDisplay && (
 						<LastUpdated
-							lastUpdated={block.blockLastUpdated}
+							lastUpdated={new Date(block.blockLastUpdated)}
 							lastUpdatedDisplay={block.blockLastUpdatedDisplay}
```
oliverlloyd, line 115: Are we saying that we will always replace this string with a new 'local' value on the client? I wonder if it's better to just not render anything on the server at all. That way we avoid the flash of incorrect content which could be very distracting when trying to read a post (it could well look as though the post has _just_ updated when in fact we just changed the time format)

apps-rendering/src/client/liveblog.ts
```diff
@@ -0,0 +1,23 @@
+import { ResultKind } from '@guardian/types';
+import { formatLocalTimeDateTz } from 'date';
+import { dateParser, parse } from 'parser';
+import { logger } from '../logger';
+
+function lastUpdatedDates(): void {
+	Array.from(document.getElementsByClassName('js-last-updated-time')).forEach(
```
oliverlloyd, line 7: Are we selecting using `js-` based on backwards compatibility reasons? This is a legacy method for selecting elements.

apps-rendering/src/date.test.ts
```diff
@@ -91,6 +92,20 @@ describe('formatLocal', () => {
 	});
 });
 
+describe('formatLocalTimeDateTz', () => {
+	test('returns correct local time zone (local timezone is Europe/London set in TZ global variable)', () => {
+		expect(formatLocalTimeDateTz(new Date('2020-03-11T17:25:00'))).toBe(
+			'5:25pm 11 Mar 2020 GMT',
+		);
```
oliverlloyd, line 99: Just a random thought, what causes this code to decide the local timezone is `GMT`? Could it make a different choice if it's running on a CI server based in a different timezone?

common-rendering/src/components/LastUpdated.tsx
```diff
@@ -6,19 +6,23 @@ const LastUpdated = ({
...
 				${textSans.xxsmall()};
 				color: ${neutral[46]};
-			`}
+				`}
 		>
-			<time dateTime={new Date(lastUpdated).toISOString()}>
-				{`Updated at ${lastUpdatedDisplay}`}
+			<time
+				className="js-last-updated-time"
+				data-last-updated={lastUpdated.toISOString()}
```
oliverlloyd, line 22: Do we need this? Semantically, `datetime` is the value which is what we need to format thing, so we could just read that?

oliverlloyd (commented):
> This includes reformatting for a user's local timezone
>

Is this PR saying that we should now show dates on articles in local time, not server time? Or is there a specific use case for the last updated string that needs it to be more accurate?

If the latter, perhaps we can solve that use case with relative time? Eg.: `Updated 2 mins ago`

dotcom-rendering/src/web/components/LiveBlock.tsx
DavidLawes: The change to this component is only for the prop type because we thought this could simplify things. Inside LastUpdated we convert the `number` to a `Date`. In apps-rendering the block lastUpdated property is a `Date`, so we had to convert to a `number` only for it to be converted back to a `Date` again.

apps-rendering/src/date.test.ts
DavidLawes: The timezone is defined in the setup (https://github.com/guardian/dotcom-rendering/blob/main/apps-rendering/config/jestglobalSetup.js#L5), so will always be consistent

apps-rendering/src/client/liveblog.ts
DavidLawes: This was a suggestion from Jamie to follow similar patterns already implemented. We apply a className to the component with the prefix `js-` to show that it is used as part of the javascript function. It's used as an indicator for future changes that this className is important.

Previously I was doing a query selector to get all elements by data-attribute, but we decided to change. Are you suggesting I revert back to selecting by data attribute?

common-rendering/src/components/LastUpdated.tsx
DavidLawes: Sorry, maybe I've misread your comment - are you suggesting I don't use the `toISOString` function? In this PR the lastUpdated prop is a `Date`, so I thought converting it to a string would be best?

apps-rendering/src/client/liveblog.ts
oliverlloyd: I can't speak to how AR selects elements but from a DCR perspective, the `js-` thing is a Frontend pattern. It is present in a few isolated locations in DCR but only where we have contracts with external libs, like commercial.

In DCR we use data attributes for selecting elements. See `data-gu-*` or `data-ignore-*` but even then this is done rarely. Because we're using React and Emotion we rarely need global selectors.

Because this use case is proposing a global action I think this should be broken out into its own PR and an approach considered there. You are not going to just want to change this one date value to be local, you will inevitably want to change all the dates on the page (if you don't you will have dates showing in different timezones, which is bad).

[This PR](https://github.com/guardian/dotcom-rendering/pull/3704) shows how we broke out a script to manage relative time, which could, perhaps, offer a suggestion for a solution to local time.

When choosing any contract we should be mindful of the history and try to locate any decisions that were taken in the past. I'm sure this has been discussed before!

common-rendering/src/components/LastUpdated.tsx
oliverlloyd: Ah, no sorry, I meant that you don't need the `data-last-updated` attribute at all. You already have the `datetime` attribute which is more semantic and has the required value

dotcom-rendering/src/web/components/LiveBlock.tsx
oliverlloyd: I was more thinking about the fact that elsewhere in this PR we are adding a client side action to mutate the value of this dom entry. This means the html that is created on the server and sent to the client will change. That is, it will get rendered on the page as one thing, say,'15th Dec 10:00 GMT', and then, after a short delay, will suddenly show as '15th Dec 05:00 EST'.

To avoid that flash of incorrect content, it might be better to not render the value on the server?

DavidLawes: We discussed this with Jamie - we expect the overall date time string to vary by max 1 character (timezones can be either 3 or 4 chars in length). As such, the possible CLS resulting from this might be quite small. Also the client side bundle has been known to break and, given the possible legal requirements to show last updated time, it was decided to render at least the server side time.

apps-rendering/src/client/liveblog.ts
DavidLawes: We spoke about this with Jamie - the pattern for prefixing the class name with `js-` serves as an indicator of this contract and is used elsewhere in apps-rendering and The Guardian. We decided to keep as it is for this case.

common-rendering/src/components/LastUpdated.tsx
DavidLawes: Thanks - will update :)

JamieB-gu (approved):

common-rendering/src/components/LastUpdated.tsx
```diff
@@ -6,19 +6,22 @@ const LastUpdated = ({
...
 				align-items: flex-end;
 				${textSans.xxsmall()};
 				color: ${neutral[46]};
-			`}
+				`}
 		>
-			<time dateTime={new Date(lastUpdated).toISOString()}>
-				{`Updated at ${lastUpdatedDisplay}`}
+			<time
+				className="js-last-updated-time"
```
oliverlloyd, line 21: I know you and Jamie are happy to go forward with the `js-*` prefix for AR but this code is in shared and will end up in DCR and in that case this is not a good idea. DCR took the line that we would not use this pattern.

The reason we made that call is to avoid confusion. We know that Frontend widely makes use of 'js-' selectors but DCR has none, except where we need to support legacy code.

In fact, in DCR we don't use selectors hardly at all. We use React and select content using the React scope. In those situations where we do need to select elements outside of React we create contracts, which, so far, follow the pattern of using either data attributes like `data-relativeformat` or pseudo custom elements, like `gu-hydrate`

If we merge this code then we breach that line and open up the possibility for a very loose contract for how elements are selected which I really do think is a bad idea for DCR.

oliverlloyd (changes requested):
When we discussed the `js-*` business I had it framed in the context of AR code only but it seen like it is now being put into DCR and I don't think that is a pattern we should encourage.

common-rendering/src/components/LastUpdated.tsx
oliverlloyd: Why don't we extend the _existing_ `time` contract that is already in place by having `data-absoluteformat: 'local' | 'edition'`? That way it works for both AR and DCR, which both treat time differently.

We already have `data-relativeformat: 'med' | 'long'` and it feels like keeping these `time` related contracts consistent is a good thing

DavidLawes:
[Support liveblocks in AR - show lastUpdated date](https://trello.com/c/qpV4MjFB/470-support-liveblocks-in-ar-show-lastupdated-date)

common-rendering/src/components/LastUpdated.tsx
DavidLawes: Thanks @oliverlloyd - I can see how this could work within DCR. I'm hoping @JamieB-gu might be able to comment from an AR-perspective as this approach would be a new pattern within that codebase (i.e. selecting using `data-relativeformat` instead of className `js-last-updated-time`).

I'm wondering whether, if we can't reach an agreement, this component shouldn't be shared (maybe that's too extreme)?

oliverlloyd: > I'm wondering whether, if we can't reach an agreement, this component shouldn't be shared (maybe that's too extreme)?

I don't think that's an extreme take. Trying to make shared code work when the use cases are different leads to complex, hard to maintain abstractions. Perhaps if we can't find a pattern that works cross platform here then it makes more sense to prioritise the benefits of consistent patterns and future maintainability vs. the benefits of having shared code in this case?

JamieB-gu: > open up the possibility for a very loose contract for how elements are selected which I really do think is a bad idea for DCR.

My suggestion would be to use one or the other, _not_ to use both. I agree that would be unhelpful 👍

> DCR took the line that we would not use this pattern.
> 
> The reason we made that call is to avoid confusion. We know that Frontend widely makes use of 'js-' selectors but DCR has none, except where we need to support legacy code.

Does this mean that `data-*` attributes were chosen over `js-*` classes primarily to distinguish DCR code from frontend code? Are there other reasons for preferring them?

Just to make sure I understand: we're talking about two different conventions for how elements that are rendered on the server-side can be selected (looked up) on the client-side? One approach is to select based on `data-*` attributes, the other is to select based on `js-*` classes? Both ultimately achieve the same goal, so it's a question of pros/cons of each?

Assuming I have understood correctly (shout if I haven't), I'd argue that `js-*` classes are preferable because this is the only thing they're used for. If you see a `js-*` class you _know_ it's being used to select code on the client-side somewhere. Data attributes, by contrast, are part of a broader web API, and therefore may be used for other purposes as well. They're also designed to be key-value pairs, so that you can transfer information from the HTML context to the JS context. It seems strange to use them just to look up the element, without caring about the data in the attribute, but it may well be that I've not understood the context correctly.

Anyway, all that said, I think the benefit of consistency and being able to share a convention outweighs the above advantages. As mentioned, I think we pick one and run with it. My preference would be to shift DCR over to `js-*` classes, but if you think that would be undesirable/problematic then we can shift AR over to `data-*` attributes instead and go with that.

oliverlloyd: We made the call to only use classnames for styling and use attributes for building contracts in DCR. There's not really any kind of agreement on which way to go and both are valid but my feeling is data attributes are more scalable. Being able to have a structured key pair arrangement is a lot easier to scale than a single string.

Some of the oldest contracts in place at the Guardian use data attributes (`data-component`, `data-link-name`, `data-relativeformat`, etc.) and I - personally - think these might not have endured so long had they been classnames. 

This is [the discussion](https://github.com/guardian/dotcom-rendering/pull/2669) introducing the `data-ignore` contract. There are a lot of others including contracts with other teams (`data-spacefinder`) I worry trying to change all these would be a lot of work and add a lot of risk.

apps-rendering/src/components/liveBlocks.tsx
```diff
@@ -44,18 +45,18 @@ const LiveBlocks: FC<LiveBlocksProps> = ({ blocks, format }) => {
...
 							`}
 						>
 							{block.lastModified.kind === OptionKind.Some &&
 								block.firstPublished.kind === OptionKind.Some &&
 								block.lastModified.value >
 									block.firstPublished.value && (
 									<LastUpdated
-										lastUpdated={Number(
+										lastUpdated={block.lastModified.value}
+										lastUpdatedDisplay={formatUTCTimeDateTz(
```
DavidLawes, line 57: It was agreed with Jamie B and David F that for apps-rendering we'd render UTC on the server

DavidLawes:
@JamieB-gu @oliverlloyd 
We had a chat with CSTI today and it gave me confidence to separate out the lastUpdated component. I think we were reaching a bit of a stalemate and probably how we select elements isn't something we can agree on in this PR. 

This PR now:
- moves `LastUpdated` component from common-rendering into dotcom-rendering. Note that for dotcom-rendering the time is rendered server-side and there is no update on the client. The `LastUpdated` component in dotcom therefore doesn't have any additional selector.
- adds a `lastUpdated` component into apps-rendering. This component includes a `js-*` className which is used by liveblog.ts to grab all the components that need updating to local time on the client.
- for apps-rendering only, renders UTC time on the server. There were was a chat with David F and Jamie B about the date format if the client side bundle breaks, which is how we reached this decision.

oliverlloyd:
> We had a chat with CSTI today and it gave me confidence to separate out the lastUpdated component. I think we were reaching a bit of a stalemate and probably how we select elements isn't something we can agree on in this PR.

Fair.

Thanks for keeping this PR going despite so much back and forth. Really well done!

DavidLawes:
Also, just so we don't forget, I appreciate that the dotcom-rendering last updated time doesn't quite match the design specified by Harry. I created a ticket to cover this piece of work: https://trello.com/c/4GgN73e9

common-rendering/src/components/LastUpdated.tsx
JamieB-gu: Fair enough 👍. In which case I think it's reasonable to switch to `data-*` attributes in AR @DavidLawes. There's clearly been some thought behind this in DCR, and I think we get a lot of benefit from sharing this convention and the resulting code 📈

JamieB-gu (approved):

PR #3726, opened 2021-12-07
Fix match nav spacing

nitro-marky:
## Why?

There's probably a neater way of doing this but I want to get it done! This introduces another version of the the grid which is for sport blogs. This removes the spacing.

### Before

Screenshot 2021-12-06 at 13 03 05

### After

Screenshot 2021-12-07 at 15 36 34

Screenshot 2021-12-07 at 15 38 17


rcrphillips (approved):
LGTM

PR #3727, opened 2021-12-07
Merge cdk into parent package

nicl:
## What + why

No longer treats apps-rendering/cdk as a separate package, but combines it with the parent. This allows us to remove a lot of TS/config files and reduces confusion as nested packages are difficult to reason about.

Previously to do stuff, script/* files within the `/cdk` folder were used. Now we use helpers in the parent package.json. These have been named for consistency with the existing commands. The significant ones are:

    build:cdk
    watch:cdk
    synth
    diff

test, lint, etc. are just the parent existing commands extended to include `cdk/` too.


JamieB-gu (approved):
Looks great! Couple of merge conflicts I think from more recent CDK changes, but otherwise 👍

PR #3730, opened 2021-12-08
Fixed width for comment ad element

frederickobrien:
This addresses one half of an issue documented [here](https://trello.com/c/LgxmPnuH/1665-large-cls-jumps-raised-by-user-help) in which refreshed ads cause [cumulative layout shift](https://web.dev/cls/). Because the ad slot beside comments did not have a fixed height, if the new ad had different dimensions to the previous one that would change the width of the whole slot, which in turn changed the width of the comment section.

As shown below, giving the `dfp-ad--comments` element a fixed width of 300px, its size remains consistent regardless of the dimensions of the ad inside it. 

image

image

image

### DIFF
--- a/dotcom-rendering/src/web/components/AdSlot.tsx
+++ b/dotcom-rendering/src/web/components/AdSlot.tsx
@@ -301,6 +301,7 @@ export const AdSlot: React.FC<Props> = ({ position, display }) => {
 							css`
 								position: sticky;
 								top: 0;
+								width: 300px;
 							`,
 							adStyles,
 						]}


oliverlloyd:
Does this change mean the Discussion will no longer stretch to full width for signed in users?

Screenshot 2021-12-09 at 13 43 10

frederickobrien:
> Does this change mean the Discussion will no longer stretch to full width for signed in users?
> 
> Screenshot 2021-12-09 at 13 43 10

Great question, @oliverlloyd! Thank you, it hadn't occurred to me to check. Running in CODE it looks like Discussion is unaffected when users are signed in:

image

oliverlloyd:
> 
> Great question, @oliverlloyd! Thank you, it hadn't occurred to me to check. Running in CODE it looks like Discussion is unaffected when users are signed in:
> 
Amazing, thanks for checking!

oliverlloyd (approved):

PR #3732, opened 2021-12-08
Refactor email embed event listener

georgeblahblah:
## What does this change?

* Refactors email embed listeners in DCRAR to validate message format and message origin
* Initiates that listener in DCR & AR
* "Unwraps" email embeds in `apps-rendering` from their parent iframes, so they can send the resize message to the parent
  * parse the email signup embed from CAPI

## Why?

There is an incoming change in `frontend` which will add reCAPTCHA to email embed iframes (https://github.com/guardian/frontend/pull/24451).

When displaying a captcha, the iframe needs a height of 500px. This change enables the email embeds to emit a message to resize themselves to accommodate the CAPTCHA display.

For more information, see [reCAPTCHA Implementation Overview](https://docs.google.com/document/d/1O2-aS6o3Lj1lKeFfTB-Y0aBEEE6Xq5gE-q-BWa89EBQ/edit?usp=sharing).

### DIFF
--- a/apps-rendering/src/client/article.ts
+++ b/apps-rendering/src/client/article.ts
@@ -479,15 +479,11 @@ function hydrateClickToView(): void {
 }
 
 function resizeEmailSignups(): void {
-	const isIframe = (elem: Element): elem is HTMLIFrameElement =>
-		elem.tagName === 'IFRAME';
 	const emailSignupIframes = document.querySelectorAll(
 		'.js-email-signup > iframe',
 	);
 	Array.from(emailSignupIframes).forEach((emailSignupIframe) => {
-		if (isIframe(emailSignupIframe)) {
-			emailSignupIframe.style.width = '100%';
-		}
+		(emailSignupIframe as HTMLIFrameElement).style.width = '100%';
 	});
 }
 
--- a/apps-rendering/src/client/article.ts
+++ b/apps-rendering/src/client/article.ts
@@ -479,12 +479,16 @@ function hydrateClickToView(): void {
 }
 
 function resizeEmailSignups(): void {
+	const isIframe = (elem: Element): elem is HTMLIFrameElement =>
+		elem.tagName === 'IFRAME';
 	const emailSignupIframes = document.querySelectorAll(
 		'.js-email-signup > iframe',
 	);
-	Array.from(emailSignupIframes).forEach((emailSignupIframe) => {
-		(emailSignupIframe as HTMLIFrameElement).style.width = '100%';
-	});
+	Array.from(emailSignupIframes)
+		.filter(isIframe)
+		.forEach((emailSignupIframe) => {
+			emailSignupIframe.style.width = '100%';
+		});
 }
 
 setup();
--- a/apps-rendering/src/client/article.ts
+++ b/apps-rendering/src/client/article.ts
@@ -481,7 +481,9 @@ function hydrateClickToView(): void {
 function resizeEmailSignups(): void {
 	const isIframe = (elem: Element): elem is HTMLIFrameElement =>
 		elem.tagName === 'IFRAME';
-	const emailSignupIframes = document.querySelectorAll('.js-email-signup > iframe');
+	const emailSignupIframes = document.querySelectorAll(
+		'.js-email-signup > iframe',
+	);
 	Array.from(emailSignupIframes).forEach((emailSignupIframe) => {
 		if (isIframe(emailSignupIframe)) {
 			emailSignupIframe.style.width = '100%';
--- a/apps-rendering/src/client/article.ts
+++ b/apps-rendering/src/client/article.ts
@@ -481,14 +481,10 @@ function hydrateClickToView(): void {
 function resizeEmailSignups(): void {
 	const isIframe = (elem: Element): elem is HTMLIFrameElement =>
 		elem.tagName === 'IFRAME';
-	const emailSignupIframes = document.querySelectorAll('.js-email-signup');
+	const emailSignupIframes = document.querySelectorAll('.js-email-signup > iframe');
 	Array.from(emailSignupIframes).forEach((emailSignupIframe) => {
 		if (isIframe(emailSignupIframe)) {
-			const innerIframe =
-				emailSignupIframe.contentDocument?.querySelector('iframe');
-			if (innerIframe) {
-				innerIframe.style.width = '100%';
-			}
+			emailSignupIframe.style.width = '100%';
 		}
 	});
 }
--- a/apps-rendering/src/client/article.ts
+++ b/apps-rendering/src/client/article.ts
@@ -478,12 +478,11 @@ function hydrateClickToView(): void {
 		);
 }
 
+const isIframe = (elem: Element): elem is HTMLIFrameElement =>
+	elem.tagName === 'IFRAME';
+
 function resizeEmailSignups(): void {
-	const isIframe = (elem: Element): elem is HTMLIFrameElement =>
-		elem.tagName === 'IFRAME';
-	const emailSignupIframes = document.querySelectorAll(
-		'.js-email-signup > iframe',
-	);
+	const emailSignupIframes = document.querySelectorAll('.js-email-signup');
 	Array.from(emailSignupIframes)
 		.filter(isIframe)
 		.forEach((emailSignupIframe) => {
--- a/apps-rendering/src/client/newsletterEmbeds.ts
+++ b/apps-rendering/src/client/newsletterEmbeds.ts
@@ -22,7 +22,7 @@ export default (): Promise<void> => {
 
 		const iframes: HTMLIFrameElement[] = allIframes.filter((i) => {
 			try {
-				if (i.contentWindow && event.source) {
+				if (i.contentWindow !== null && event.source !== null) {
 					return i.contentWindow === event.source;
 				}
 			} catch (e) {
--- a/apps-rendering/src/client/newsletterEmbeds.ts
+++ b/apps-rendering/src/client/newsletterEmbeds.ts
@@ -1,4 +1,8 @@
-type NewsletterHeightEventType = { source: object; origin: string };
+type NewsletterHeightEventType = { source: MessageEventSource; origin: string };
+type MessageType = {
+	type: string;
+	value: string;
+};
 
 const allowedOrigins = ['https://www.theguardian.com'];
 export default (): Promise<void> => {
@@ -10,8 +14,7 @@ export default (): Promise<void> => {
 	// Otherwise, earlier resize events might be missed
 	// So we don't have to load this script as a priority on each load
 	allIframes.forEach((iframe) => {
-		if (iframe && iframe.contentWindow)
-			iframe.contentWindow.postMessage('resize', '*');
+		iframe.contentWindow?.postMessage('resize', '*');
 	});
 
 	window.addEventListener('message', (event) => {
@@ -34,11 +37,11 @@ export default (): Promise<void> => {
 		});
 		if (iframes.length !== 0) {
 			try {
-				const message: { [key: string]: string } = JSON.parse(
+				const message: MessageType = JSON.parse(
 					event.data,
-				);
+				) as MessageType;
 				switch (message.type) {
-					case 'set-height':
+					case 'set-height': {
 						const value = parseInt(message.value);
 						if (!Number.isInteger(value)) return;
 
@@ -46,9 +49,10 @@ export default (): Promise<void> => {
 							iframe.height = `${value}`;
 						});
 						break;
+					}
 					default:
 				}
-				// eslint-disable-next-line no-empty
+				// eslint-disable-next-line no-empty -- No action required
 			} catch (e) {}
 		}
 	});
--- a/apps-rendering/src/client/newsletterEmbeds.ts
+++ b/apps-rendering/src/client/newsletterEmbeds.ts
@@ -1,8 +1,3 @@
-type MessageType = {
-	type: string;
-	value: string;
-};
-
 const allowedOrigins = ['https://www.theguardian.com'];
 export default (): Promise<void> => {
 	const allIframes: HTMLIFrameElement[] = [].slice.call(
@@ -31,9 +26,6 @@ export default (): Promise<void> => {
 		});
 		if (iframes.length !== 0) {
 			try {
-				// const message: MessageType = JSON.parse(
-				// 	event.data,
-				// ) as MessageType;
 				const message = JSON.parse(event.data);
 				if (!message) return;
 				if (typeof message.type !== 'string') return;
@@ -41,7 +33,7 @@ export default (): Promise<void> => {
 				switch (message.type) {
 					case 'set-height': {
 						if (
-							typeof message.value !== 'number' ||
+							typeof message.value !== 'number' &&
 							typeof message.value !== 'string'
 						)
 							return;
--- a/apps-rendering/src/client/newsletterEmbeds.ts
+++ b/apps-rendering/src/client/newsletterEmbeds.ts
@@ -1,22 +1,23 @@
 type NewsletterHeightEventType = { source: object; origin: string };
 
-const allowedOrigins = ["https://www.theguardian.com"];
-export default (selector: string) => (): Promise<void> => {
+const allowedOrigins = ['https://www.theguardian.com'];
+export default (): Promise<void> => {
 	const allIframes: HTMLIFrameElement[] = [].slice.call(
-		document.querySelectorAll(selector)
+		document.querySelectorAll('.email-sub__iframe'),
 	);
 
 	// Tell the iframes to resize once this script is loaded
 	// Otherwise, earlier resize events might be missed
+	// So we don't have to load this script as a priority on each load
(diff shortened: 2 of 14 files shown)


common-rendering/src/newsletterEmbedIframe.ts
```diff
@@ -0,0 +1,56 @@
+type NewsletterHeightEventType = { source: object; origin: string };
```
oliverlloyd, line 1: Can we perhaps keep this code local to DCR and simply duplicate it in AR, rather than using common-rendering?

I don't think this sort of code belongs here. It was conceived for sharing components but it's unfortunately now very inviting as a place to put things but perhaps we shouldn't be so quick to share and having the sorts of checks and balances you get with things like `@guardian/libs` is arguably a good thing

For example, I am already aware that we have at least two, possibly three, different ways of resizing iframes in DCR so there's clearly a need to review this and consider an abstraction. But that kind of refactoring work is harder when you've already got coupled code.

oliverlloyd (commented):
This looks great and I know how much work has gone into this so well done!

common-rendering/src/newsletterEmbedIframe.ts
georgeblahblah: I'm inclined to agree here, especially in the context of what `common-rendering` sets out to achieve. I will move this utility back into AR/DCR - which is essentially how things were.

apps-rendering/src/components/emailSignup.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+interface Props {
+	embed: EmailSignup;
+}
+
+const styles = css`
+	margin: ${remSpace[4]} 0;
+
+	${darkModeCss`
+        background: ${background.inverse};
```
georgeblahblah, line 21: I'm a bit unsure about this. For `GenericEmbed`, this is set to `white` (https://github.com/guardian/dotcom-rendering/pull/3732/files#diff-b59bf599937aca257cb6d5e80b102b71a4dc7dabef64bf46b47ec2e5acfc170eR17)

DavidLawes (approved):
From a dcr perspective this makes sense to me (and thanks for linking to the frontend pr too - that answered a lot of my questions). I think it would be good to get a review with someone more confident with the AR code before merging :)

I'm also suspicious of the dark mode background colour - maybe that's a question for @JamieB-gu?

I didn't have any familiarity with email embeds so I'll copy some info you shared with me offline that helped me understand the flow better:
* `frontend` is responsible for rendering the email signup embeds (e.g https://www.theguardian.com/email/form/plaintone/inside-saturday). This is where the captcha will get shown and why we only need to add recaptcha in one place (i.e. no corresponding change needed in MAPI).
* anywhere that renders articles/interactives (frontend, DCR & AR) now needs to respond to the resize messages, so that the captcha is visible.

apps-rendering/src/client/article.ts
```diff
@@ -468,17 +468,11 @@ function hydrateClickToView(): void {
...
+	);
 	Array.from(emailSignupIframes).forEach((emailSignupIframe) => {
-		if (isIframe(emailSignupIframe)) {
-			const innerIframe =
-				emailSignupIframe.contentDocument?.querySelector('iframe');
-			if (innerIframe) {
-				innerIframe.style.width = '100%';
-			}
-		}
+		(emailSignupIframe as HTMLIFrameElement).style.width = '100%';
```
JamieB-gu, line 475: I think if you keep the `isIframe` type predicate you can use a `filter(isIframe)` before your `forEach`. I believe TypeScript is clever enough to then narrow the type of `emailSignupIframe` to `HTMLIFrameElement` for you (which is awesome!), which means you no longer need the typecast.

apps-rendering/src/client/newsletterEmbeds.ts
```diff
@@ -0,0 +1,61 @@
...
+export default (): Promise<void> => {
+	const allIframes: HTMLIFrameElement[] = [].slice.call(
+		document.querySelectorAll('.email-sub__iframe'),
+	);
+
+	// Tell the iframes to resize once this script is loaded
+	// Otherwise, earlier resize events might be missed
+	// So we don't have to load this script as a priority on each load
+	allIframes.forEach((iframe) => {
+		iframe.contentWindow?.postMessage('resize', '*');
```
JamieB-gu, line 17: Is there anything that prevents us using `https://www.theguardian.com` as the `targetOrigin` here instead of `*`?

apps-rendering/src/client/newsletterEmbeds.ts
```diff
@@ -0,0 +1,61 @@
...
+	});
+
+	window.addEventListener('message', (event) => {
+		if (
+			!allowedOrigins.includes(
+				(event as NewsletterHeightEventType).origin,
```
JamieB-gu, line 23: In the scenario where you call `window.addEventListener` with `message` as the first argument, is TypeScript automatically able to narrow the type of `event` to `MessageEvent`? If so, does that interface already have `origin` and `source` defined as properties? In which case are the typecasts here and on line 32 needed?

apps-rendering/src/client/newsletterEmbeds.ts
```diff
@@ -0,0 +1,61 @@
...
+			)
+		)
+			return;
+
+		const iframes: HTMLIFrameElement[] = allIframes.filter((i) => {
+			try {
+				return (
+					i.contentWindow ===
+					(event as NewsletterHeightEventType).source
```
JamieB-gu, line 32: TypeScript says that it's possible for both `contentWindow` and `source` to be `null`. In that scenario I think this check will return `true` and include the `iframe` in the filter, which I'm guessing isn't what you intend here? Would it be best to include a check against these properties being `null`?

I believe this check is also being used in the `mobile-apps-article-templates` implementation, so it may also be a good idea to include the `null` check there too?

apps-rendering/src/client/newsletterEmbeds.ts
```diff
@@ -0,0 +1,61 @@
...
+				);
+			} catch (e) {
+				return false;
+			}
+		});
+		if (iframes.length !== 0) {
+			try {
+				const message: MessageType = JSON.parse(
+					event.data,
+				) as MessageType;
```
JamieB-gu, line 42: Can we verify the structure of this data to guard against possible runtime exceptions? Perhaps `typeof message?.type === 'string'` and `typeof message?.value === 'string'`? I think that would sidestep the need to typecast?

apps-rendering/src/client/newsletterEmbeds.ts
georgeblahblah: Oh, yes! This message should only be sent to frames on `https://www.theguardian.com`. It makes testing on `CODE` a little harder but I think the tradeoff is worth it

apps-rendering/src/client/newsletterEmbeds.ts
georgeblahblah: Yes, typescript automatically narrows `event` to `MessageEvent`, so we can remove the typecasts & the `NewsletterHeightEventType`!

apps-rendering/src/client/newsletterEmbeds.ts
georgeblahblah: That's a good idea. I think `.type` should always be a string,` and `.value` could reasonably be `string` or `number`

georgeblahblah: I found a similar implementation in https://github.com/guardian/dotcom-rendering/blob/aeb7f68d9849d3489d6cdbf78894e65c22506fcc/apps-rendering/src/client/interactives.ts, which has been useful here

apps-rendering/src/components/emailSignup.tsx
georgeblahblah: I'll revert this, so we can spend the time looking at it in the context of design & accessibility requirements

dotcom-rendering/src/web/browser/newsletterEmbedIframe/init.ts
```diff
@@ -1,33 +1,56 @@
...
 				return false;
 			}
+			return false;
 		});
 		if (iframes.length !== 0) {
 			try {
-				const message: { [key: string]: string } = JSON.parse(
-					event.data,
-				);
+				const message: MessageType = JSON.parse(event.data);
```
JamieB-gu, line 42: Is there any reason not to perform the same `isObject` and `typeof` message checks here? I think your decision to annotate the result of `JSON.parse` as `unknown` in the AR version was very sensible.

Whenever you annotate a value that was previously of type `any`, that annotation is equivalent to a type cast (`as`). For example:

```ts
const a = JSON.parse('a'); // This is of type `any`
const b: number = a; // This compiles fine
```

In the above example, `b` can be passed anywhere a `number` is expected, which will likely cause runtime errors. One way to solve this problem, as seen in your AR version, is to annotate with `unknown`. That's still a type cast, but it's saying that we known absolutely nothing about that value, so it can't really be used anywhere without causing compiler errors. In order to use it TS will force us to make checks, like `typeof` and `isObject`, to figure out the correct type.

**Note:** If you choose to make this change, in DCR I think `isObject` is available from `@guardian/libs`. The reason it's in the local `lib` in AR is because I believe its existence there predates the introduction to `@guardian/libs`.

George, I wouldn't worry about this for your PR, but I'm concerned that these two implementations (DCR and AR) have already diverged. I believe the two modules are meant to be operating on the same embeds and providing the same functionality? @oliverlloyd I'm aware it was your suggestion to keep them separate and not share the implementation, but to my mind capturing these kind of rules in one place and keeping them in sync is important 🤔

That said, I'm not suggesting you have to refactor things again for now George! I think it's something for the team to consider more broadly.

apps-rendering/src/client/newsletterEmbeds.ts
```diff
@@ -0,0 +1,62 @@
...
+			'https://www.theguardian.com',
+		);
+	});
+
+	window.addEventListener('message', (event) => {
+		if (!allowedOrigins.includes(event.origin)) return;
+
+		const iframes: HTMLIFrameElement[] = allIframes.filter((i) => {
+			try {
+				if (i.contentWindow && event.source) {
```
JamieB-gu, line 25: Small suggestion: can I recommend explicitly checking both against `null` here? E.g. `i.contentWindow !== null`. I think it's easier to understand at a glance what the values can be and what you're checking when reading the code, and implicit conversions can cause unintended consequences at times.

apps-rendering/src/client/newsletterEmbeds.ts
```diff
@@ -0,0 +1,62 @@
...
+					return i.contentWindow === event.source;
+				}
+			} catch (e) {
+				return false;
+			}
+			return false;
+		});
+		if (iframes.length !== 0) {
+			try {
+				const message: unknown = JSON.parse(event.data);
```
JamieB-gu, line 35: Nice!

dotcom-rendering/src/web/browser/newsletterEmbedIframe/init.ts
oliverlloyd: > I'm aware it was your suggestion to keep them separate and not share the implementation, but to my mind capturing these kind of rules in one place and keeping them in sync is important

Sorry, I didn't mean to say I'm against sharing this code. In principle, that makes sense, esp. as you say, we're dealing with the same embeds here!

However, I don't think `common-rendering` is the right place to do it. It's not setup for sharing lib code and once you start to do something more will follow - lib folders have a tendency to grow messy fast!

Rather than create a quick abstraction here, I'd rather we take an approach where we stick with duplication for now and then abstract slower using @guardian/libs 

A slower abstraction allows time to consider what is needed and what, if any, shared areas there are that could be abstracted and what the best api for that abstraction might be. Other benefits to guardian/libs is it is a place that allows wider contributions and tends to have a higher bar to entry, which, in my opinion, is a very good thing when creating abstractions

georgeblahblah: > Is there any reason not to perform the same isObject and typeof message checks here?

No, it was mostly resistance on my part because of `isObject` living in AR. I had not realised it is in `@guardian/libs`, so I've implemented the same checks in DCR as well now.

apps-rendering/src/embed.ts
```diff
@@ -68,8 +70,12 @@ interface TikTok extends GenericFields {
 	kind: EmbedKind.TikTok;
 }
 
-interface EmailSignup extends GenericFields {
+interface EmailSignup {
 	kind: EmbedKind.EmailSignup;
+	alt: Option<string>;
+	caption: Option<string>;
```
georgeblahblah, line 76: @JamieB-gu I added `caption` here as well as `alt`. I've used `alt` to provide the required `title` attribute on the `iframe`

apps-rendering/src/components/embedWrapper.tsx
```diff
@@ -292,18 +300,17 @@ const divElementPropsToEmbedComponentProps = (
...
-							resultMap(
-								(
-									genericFields: GenericFields,
-								): EmailSignup => ({
-									kind: EmbedKind.EmailSignup,
-									...genericFields,
-								}),
-							),
-						);
+						return resultMap<string, EmailSignup>(
```
georgeblahblah, line 303: @JamieB-gu we didn't pair on this, but I also filled in the implementation here in `embedWrapper.tsx`. Could you take a look here (and lines 89-99) please?

davidfurey (approved):

apps-rendering/src/components/embedWrapper.tsx
```diff
@@ -88,7 +86,18 @@ const embedToDivProps = (embed: Embed): Record<string, string> => {
 		case EmbedKind.EmailSignup: {
 			return {
 				kind: EmbedKind.EmailSignup,
-				...genericDivProps(embed),
+				...withDefault({})(
```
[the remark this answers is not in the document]
davidfurey: Seems like this is because `embed.tracking` is an enum, but `alt` is an option

apps-rendering/src/components/embedWrapper.tsx
```diff
@@ -356,11 +364,15 @@ const getSourceDetailsForEmbed = (embed: Embed): SourceDetails => {
...
 		case EmbedKind.TikTok:
-		case EmbedKind.EmailSignup:
 			return {
 				source: embed.source,
 				sourceDomain: embed.sourceDomain,
 			};
+		case EmbedKind.EmailSignup:
+			return {
+				source: some('Guardian'),
+				sourceDomain: some('www.theguardian.com'),
```
JamieB-gu, line 374: Do you think it would be worth storing these two fields in a standalone file and also have `newsletterEmbeds.ts` pull its usage of `www.theguardian.com` from the same variable in that file? An email embeds config file perhaps?

It's not a problem for now, but I'm thinking of a future scenario where we change the way email embeds work (perhaps hosting them on a different domain). Because these lines aren't coupled to the rest of the embed code, whoever makes the change may not realise this needs changing too. If that happens then the embed wrapper will display the wrong information.

The other option is to include these two fields in the main `EmailSignup` type, and parse them from CAPI.

JamieB-gu (commented):
Looks great! One final comment, curious what your thoughts are 🤔

apps-rendering/src/components/embedWrapper.tsx
georgeblahblah: I think my preferred option here is to include `source` and `sourceDomain` in the `EmailSignup` type and parse them from CAPI. Although it would be nice to have them in a separate field and use them from `newsletterEmbeds.ts`, we need them in a slightly different format there (as an `origin` rather than just `domain`) - unless we refactor the origin checking too.
What do you think?

apps-rendering/src/client/newsletterEmbeds.ts
```diff
@@ -0,0 +1,62 @@
...
+
+	window.addEventListener('message', (event) => {
+		if (!allowedOrigins.includes(event.origin)) return;
+
+		const iframes: HTMLIFrameElement[] = allIframes.filter((i) => {
+			try {
+				if (i.contentWindow !== null && event.source !== null) {
+					return i.contentWindow === event.source;
+				}
+			} catch (e) {
```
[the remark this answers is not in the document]
georgeblahblah: no, strangely

apps-rendering/src/components/embedWrapper.tsx
georgeblahblah: We decided to parse  `source` and `sourceDomain` from CAPI

apps-rendering/src/components/embedWrapper.tsx
```diff
@@ -356,11 +378,15 @@ const getSourceDetailsForEmbed = (embed: Embed): SourceDetails => {
...
```
JamieB-gu, line 388: I'm guessing you probably want to read these from `embed.source` and `embed.sourceDomain` now, as with the generic case above? They're currently still hardcoded.

georgeblahblah: D'oh, yes definitely

JamieB-gu (approved):

PR #3733, opened 2021-12-08
Add border to bottom of Container

nitro-marky:
## What does this change?

Adds a border to the bottom of a deadblog container to maintain parity.

### Before

Screenshot 2021-12-08 at 12 36 03

### After

Screenshot 2021-12-09 at 09 40 31

### DIFF
--- a/common-rendering/src/components/liveBlockContainer.tsx
+++ b/common-rendering/src/components/liveBlockContainer.tsx
@@ -29,7 +29,7 @@ const LiveBlockContainer = ({
 					padding: ${space[2]}px ${SIDE_MARGIN}px;
 					padding-left: ${LEFT_MARGIN_DESKTOP}px;
 				}
-				border-bottom: inset;
+				border-bottom: 1px solid ${neutral[86]};
 			`}
 		>
 			{children}
--- a/common-rendering/src/components/liveBlockContainer.tsx
+++ b/common-rendering/src/components/liveBlockContainer.tsx
@@ -29,6 +29,7 @@ const LiveBlockContainer = ({
 					padding: ${space[2]}px ${SIDE_MARGIN}px;
 					padding-left: ${LEFT_MARGIN_DESKTOP}px;
 				}
+				border-bottom: inset;
 			`}
 		>
 			{children}


common-rendering/src/components/liveBlockContainer.tsx
```diff
@@ -29,6 +29,7 @@ const LiveBlockContainer = ({
 					padding: ${space[2]}px ${SIDE_MARGIN}px;
 					padding-left: ${LEFT_MARGIN_DESKTOP}px;
 				}
+				border-bottom: inset;
```
lucymonie, line 32: Hey @nitro-marky I'm not aware of `inset` as a value for border-bottom. How does this work?

lucymonie (commented):
Just a question for you about inset.

common-rendering/src/components/liveBlockContainer.tsx
nitro-marky: According to https://developer.mozilla.org/en-US/docs/Web/CSS/border-style
`inset`
Displays a border that makes the element appear embedded. It is the opposite of outset. When applied to a table cell with border-collapse set to collapsed, this value behaves like groove.

nitro-marky: I personally think this looks better, like it is embedded on the page but I can change it to a solid line if you want.

lucymonie: Interesting! TIL... personally I think it's better to stick with `border-bottom: 1px solid ${neutral[86]};` (or whatever the colour of the line should be. I think this gives us more fine grained control over the border so we can match the design.

nitro-marky:
[deadblog cards missing 1px border bottom](https://trello.com/c/51xNtd8G/481-deadblog-cards-missing-1px-border-bottom)

nitro-marky:
[deadblog cards missing 1px border bottom](https://trello.com/c/51xNtd8G/481-deadblog-cards-missing-1px-border-bottom)

rcrphillips (approved):

lucymonie (approved):

PR #3734, opened 2021-12-08
Add highlighting of underline on hover

nitro-marky:
## What does this change?

This adds colour to the standfirst links when hovered to maintain parity.

### Before

Screenshot 2021-12-08 at 13 13 31

### After

Screenshot 2021-12-08 at 13 13 40

### DIFF
--- a/dotcom-rendering/src/web/components/Standfirst.tsx
+++ b/dotcom-rendering/src/web/components/Standfirst.tsx
@@ -146,11 +146,6 @@ const standfirstStyles = (format: ArticleFormat, palette: Palette) => {
 						margin-bottom: ${space[3]}px;
 						max-width: 540px;
 						color: ${palette.text.standfirst};
-
-						a:hover {
-							border-bottom: solid 1px
-								${palette.hover.standfirstLink};
-						}
 					`;
 				default:
 					switch (format.theme) {
@@ -163,9 +158,6 @@ const standfirstStyles = (format: ArticleFormat, palette: Palette) => {
 								a {
 									color: ${neutral[7]};
 									border-bottom: 1px solid ${neutral[60]};
-								}
-								a:hover {
-									border-bottom: 1px solid ${neutral[7]};
 								}
 							`;
 						default:
@@ -184,13 +176,25 @@ const standfirstStyles = (format: ArticleFormat, palette: Palette) => {
 	}
 };
 
+const hoverStyles = (palette: Palette) => {
+	return css`
+		a:hover {
+			border-bottom: solid 1px ${palette.hover.standfirstLink};
+		}
+	`;
+};
+
 export const Standfirst = ({ format, standfirst }: Props) => {
 	const palette = decidePalette(format);
 
 	return (
 		<div
 			data-print-layout="hide"
-			css={[nestedStyles(palette), standfirstStyles(format, palette)]}
+			css={[
+				nestedStyles(palette),
+				standfirstStyles(format, palette),
+				hoverStyles(palette),
+			]}
 			className={
 				format.design === ArticleDesign.Interactive
 					? interactiveLegacyClasses.standFirst
--- a/dotcom-rendering/src/web/components/Standfirst.tsx
+++ b/dotcom-rendering/src/web/components/Standfirst.tsx
@@ -148,8 +148,8 @@ const standfirstStyles = (format: ArticleFormat, palette: Palette) => {
 						color: ${palette.text.standfirst};
 
 						a:hover {
-							border-bottom: solid 0.0625rem
-								${palette.hover.liveBlogStandfirst};
+							border-bottom: solid 1px
+								${palette.hover.standfirstLink};
 						}
 					`;
 				default:
--- a/dotcom-rendering/src/web/components/Standfirst.tsx
+++ b/dotcom-rendering/src/web/components/Standfirst.tsx
@@ -146,6 +146,11 @@ const standfirstStyles = (format: ArticleFormat, palette: Palette) => {
 						margin-bottom: ${space[3]}px;
 						max-width: 540px;
 						color: ${palette.text.standfirst};
+
+						a:hover {
+							border-bottom: solid 0.0625rem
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -686,6 +686,13 @@ const textRichLink: (format: ArticleFormat) => string = (format) => {
 	return pillarPalette[ArticlePillar.News][400];
 };
 
+const hoverLiveblog = (format: ArticleFormat): string => {
+	if (format) {
```
oliverlloyd, line 690: Did you mean to check a property of `format` here?

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -918,6 +925,7 @@ export const decidePalette = (format: ArticleFormat): Palette => {
 		hover: {
 			headlineByline: hoverHeadlineByline(format),
 			pagination: hoverPagination(format),
+			liveBlogStandfirst: hoverLiveblog(format),
```
oliverlloyd, line 928: Maybe `standfirstLink` is a better name here? Including 'live blog' in the name makes it specific to blogs but we want to use `format` to make that choice.

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -686,6 +686,13 @@ const textRichLink: (format: ArticleFormat) => string = (format) => {
 	return pillarPalette[ArticlePillar.News][400];
 };
 
+const hoverLiveblog = (format: ArticleFormat): string => {
+	if (format) {
+		return pillarPalette[format.theme].main;
+	}
+	return pillarPalette[ArticlePillar.News][400];
```
oliverlloyd, line 693: Is news the right default? It might be better to have no default at all and require `theme`

dotcom-rendering/src/web/components/Standfirst.tsx
```diff
@@ -146,6 +146,11 @@ const standfirstStyles = (format: ArticleFormat, palette: Palette) => {
 						margin-bottom: ${space[3]}px;
 						max-width: 540px;
 						color: ${palette.text.standfirst};
+
+						a:hover {
+							border-bottom: solid 0.0625rem
+								${palette.hover.liveBlogStandfirst};
```
oliverlloyd, line 152: Can we go with `1px` to be consistent. There are a handful of cases where rems are used in DCR but for the most part it's using pixels

dotcom-rendering/src/web/lib/decidePalette.ts
nitro-marky: I assumed this would need to be checked in order to return the right colour for different types of blog. I can take it out if you want.

dotcom-rendering/src/web/components/Standfirst.tsx
nitro-marky: This was just to maintain parity with frontend but I'll change it to 1px

nitro-marky:
[Deadblog standfirst missing hover state](https://trello.com/c/ZUStL9OB/475-deadblog-standfirst-missing-hover-state)

nitro-marky:
[Deadblog standfirst missing hover state](https://trello.com/c/ZUStL9OB/475-deadblog-standfirst-missing-hover-state)

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -918,6 +922,7 @@ export const decidePalette = (format: ArticleFormat): Palette => {
 		hover: {
 			headlineByline: hoverHeadlineByline(format),
 			pagination: hoverPagination(format),
+			standfirstLink: hoverLiveblog(format),
```
oliverlloyd, line 925:
```suggestion
			standfirstLink: hoverStandfirstLink(format),
```

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -686,6 +686,10 @@ const textRichLink: (format: ArticleFormat) => string = (format) => {
 	return pillarPalette[ArticlePillar.News][400];
 };
 
+const hoverLiveblog = (format: ArticleFormat): string => {
```
oliverlloyd, line 689:
```suggestion
const hoverStandfirstLink = (format: ArticleFormat): string => {
```

dotcom-rendering/src/web/components/Standfirst.tsx
```diff
@@ -146,6 +146,11 @@ const standfirstStyles = (format: ArticleFormat, palette: Palette) => {
 						margin-bottom: ${space[3]}px;
 						max-width: 540px;
 						color: ${palette.text.standfirst};
+
+						a:hover {
+							border-bottom: solid 1px
+								${palette.hover.standfirstLink};
+						}
```
oliverlloyd, line 153: We could maybe move this outside this switch statement for `standfirstStyles` and have it in its own, hoverStyles style?

That way, in decidePalette, we can put the complete logic and it will perhaps make more sense? So, instead of having a check on design here and selectively using hover colour. You can put the check on Design inside `hoverStandfirstlink` and only return when Design === blog.

That make sense? Shout if you want to pair on this!

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -686,6 +686,17 @@ const textRichLink: (format: ArticleFormat) => string = (format) => {
 	return pillarPalette[ArticlePillar.News][400];
 };
 
+const hoverStandfirstLink = (format: ArticleFormat): string => {
+	if (format.design === ArticleDesign.DeadBlog)
+		return pillarPalette[format.theme].main;
+	if (format.design === ArticleDesign.LiveBlog) {
+		return pillarPalette[format.theme].dark;
```
nitro-marky, line 693: Liveblogs actually need to use the dark theme rather than main.

oliverlloyd (approved):
👍 👍

PR #3735, opened 2021-12-09
Reduce `App.tsx` state

oliverlloyd:
## What does this change?
Here we're moving the cookie reads for `isSignedIn` and `browserId` out of `App.tsx` and down into the consuming components

## Why?
Because we want to move logic away from `App.tsx` so that we can [remove it in future](https://docs.google.com/document/d/1kSrZZrJ8hSw6bek5hGa9onE8QDz5TRHFqHV52EOFjts/edit#heading=h.gsod33a0pc47)

## Why are we still reading `isSignedIn` then?
We still need to know this value inside `App.tsx` for now as it is a dependency for some other logic.

## Why are we inlining the cookie reads instead of using `useEffect`?
We now block hydration on the cookie read but from testing it became apparent that the time take to read the cookie was negligible


dotcom-rendering/src/web/components/App.tsx
```diff
@@ -1121,7 +1099,6 @@ export const App = ({ CAPI, NAV, ophanRecord }: Props) => {
 			)}
 			<Portal rootId="slot-body-end">
 				<SlotBodyEnd
-					isSignedIn={isSignedIn}
```
tomrf1, line 1124: `isSignedIn` is still available above - isn't it preferable to still pass it in?
Also, `StickyBottomBanner` also takes `isSignedIn` as a prop

oliverlloyd: Ah, good spot, I have also pushed the cookie read down into that component as well.

We do still have `isSignedIn` defined in `App.tsx` so technically we could pass it in but the goal is to remove it from App completely. I'm held back from doing that though until we refactor some other areas first. This is step 1.

oliverlloyd: Just for reference (and I also updated the PR description to include this link) [here is the document](https://docs.google.com/document/d/1kSrZZrJ8hSw6bek5hGa9onE8QDz5TRHFqHV52EOFjts/edit#heading=h.gsod33a0pc47) talking about the goal of removing App.tsx which is what drives this PR

oliverlloyd:

OllysCoding (approved):
This looks good to me!

PR #3736, opened 2021-12-09
Refactor how we decide the `CDN` url (also call it `ASSET_URL`)

oliverlloyd:
## What does this change?
Changes the name of the `CDN` variable to `ASSET_URL` and refactors the way we decide it

## Why are we removing the protocol relative url pattern?
[This commit](https://github.com/guardian/dotcom-rendering/pull/3736/commits/f0f75c744c3cc8e032ed74337b83fddd330f21f6). This is now [listed as an anti-pattern](https://www.paulirish.com/2010/the-protocol-relative-url/)


dotcom-rendering/src/lib/assets.ts
```diff
@@ -15,16 +15,24 @@ try {
 	// do nothing
 }
 
-// TODO: this should be removed in favour of `frontendAssetsFullURL` defined in CAPI
-// GU_STAGE is set in cloudformation.yml, so will be undefined locally
-const stage =
-	typeof process.env.GU_STAGE === 'string'
-		? process.env.GU_STAGE.toUpperCase()
```
DavidLawes, line 22: Do we always receive this env var in capitals or is it worth including `toUpperCase()` in the new flow?

dotcom-rendering/src/lib/assets.ts
```diff
@@ -15,16 +15,24 @@ try {
 	// do nothing
 }
 
-// TODO: this should be removed in favour of `frontendAssetsFullURL` defined in CAPI
```
DavidLawes, line 18: I'm guessing this this TODO no longer relevant?

DavidLawes (approved):
It's quicker to see what the actual asset url is for each environment, so this change makes sense to me, only 2 small questions :)

dotcom-rendering/src/lib/assets.ts
oliverlloyd: We could use the CAPI value but we still need to do something different when running locally so in a way it's easier to keep all the logic here. It also means we don't need to pass CAPI data into this code, keeping things simple.

dotcom-rendering/src/lib/assets.ts
oliverlloyd: That's actually a really good point!

🤔 I don't believe it will be different but if that transform was there before it means someone thought that it could so we should respect that. I've updated the code for this.

PR #3738, opened 2021-12-09
feat: Add another legacy interactive class for the count icons container

oliverlloyd:
## What does this change?
Adds `interactiveLegacyClasses.countIcons`

## Why?
So the css that lives inside the legacy interactive templates can access the count icons inside `ArticleMeta` and applying styles to them.

Yes, I agree. This is creating a brittle code coupling between two very separate codebases and is likely to cause issues if anyone wants to refactor the layout of these components in the future. If that person is you, and if you are struggling to maintain this contract, then consider pressing the interactive content or asking the Visuals team to revisit their strategy.

Example: https://www.theguardian.com/books/ng-interactive/2021/dec/08/the-best-books-of-2021

### DIFF
--- a/dotcom-rendering/docs/contracts/003-extra-classes-for-interactives.md
+++ b/dotcom-rendering/docs/contracts/003-extra-classes-for-interactives.md
@@ -4,10 +4,11 @@
 
 We have added extra classes to the standfirst and main meta sections of the interactives layout:
 
-- Standfirst: class class `content__standfirst` applied to the div that contains the standfirst paragraph.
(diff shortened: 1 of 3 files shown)


DavidLawes (approved):
Agree this probably isn't the best approach, but I'd been keen to be involved in discussions about a better alternative. I think if we're updating the contract it could possibly be worth updating the contract doc too: https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/docs/contracts/003-extra-classes-for-interactives.md

PR #3741, opened 2021-12-09
Reintroduce server-side tests in config

mxdvl:
## What does this change?

Reintroduce server side tests location on the config object. Adds a narrower type, which ensures correct usage.

## Why?

We need parity between `frontend` and DCR. It was removed in df58cd0, without fanfare.

### Before

`window.guardian.config.tests` is `undefined`

### After

`window.guardian.config.tests` is `{}`

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -688,6 +688,11 @@ interface CommercialConfigType {
 	ampIframeUrl: string;
 }
 
+type ServerSideTests = {
+    [k: `${string}Variant`]: "variant";
+    [k: `${string}Control`]: "control";
+};
+
 /**
  * the config model will contain useful app/site
  * level data. Although currently derived from the config model
@@ -700,7 +705,7 @@ interface ConfigType extends CommercialConfigType {
 	sentryHost: string;
 	dcrSentryDsn: string;
 	switches: { [key: string]: boolean };
-	abTests: Record<string, 'control' | 'variant'>;
+	abTests: ServerSideTests;
 	dfpAccountId: string;
 	commercialBundleUrl: string;
 	revisionNumber: string;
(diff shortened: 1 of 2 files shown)


dotcom-rendering/index.d.ts
```diff
@@ -688,6 +688,11 @@ interface CommercialConfigType {
 	ampIframeUrl: string;
 }
 
+type ServerSideTests = {
+    [k: `${string}Variant`]: "variant";
+    [k: `${string}Control`]: "control";
```
oliverlloyd, line 693: Where does `string` come from here?

dotcom-rendering/src/model/window-guardian.ts
```diff
@@ -23,7 +23,7 @@ export interface WindowGuardianConfig {
 		googletag: string;
 	};
 	switches: { [key: string]: boolean };
-	tests?: { [key: string]: string };
+	tests: ServerSideTests;
```
oliverlloyd, line 26: This was optional before, are we now sure it will always be provided?

mxdvl: Yes, from the look of https://www.theguardian.com/uk-news/2021/oct/26/woman-cleared-obstructing-met-police-partner-shot-stun-gun.json?dcr (assuming you’re not in any test)

abTest empty object

oliverlloyd: Nice 👍

dotcom-rendering/index.d.ts
mxdvl: `${string}Variant` is “any string ending in `Variant`”. As a RegEx is would be `/.*Variant/`.

https://github.com/guardian/frontend/blob/0657f594e2fd51125d28df9048c48d6eede26bb6/common/app/experiments/ExperimentsDefinition.scala#L25

And then it’s camelCased a few lines below.

oliverlloyd: Gotcha. Thanks for the detailed explanation!

oliverlloyd (approved):
🚀

PR #3743, opened 2021-12-10
Refactor `countryCode` in `App.tsx`

oliverlloyd:
**This PR is branched off of #3735**

## What does this change?
This PR removes `countryCode` from `App.tsx`, pushing the logic for fetching this value down into local components

## Why?
This is part of a wide effort to remove global state from `App.tsx`. See [this document](https://docs.google.com/document/d/1kSrZZrJ8hSw6bek5hGa9onE8QDz5TRHFqHV52EOFjts/edit#heading=h.gsod33a0pc47) for more details on this but the short answer is we want to make components work in isolation, removing the need for global state management for things like `countryCode` that can be memoized and don't need to be centralised.


oliverlloyd:

mxdvl (commented):
`@guardian/libs` has `getLocale`. Maybe you’ll want to use that instead?

dotcom-rendering/src/web/components/ReaderRevenueLinks.tsx
```diff
@@ -392,6 +402,21 @@ export const ReaderRevenueLinks: React.FC<Props> = ({
 	ophanRecord,
 	pageViewId = '',
 }: Props) => {
+	const [countryCode, setCountryCode] = useState<string>();
+
+	useEffect(() => {
```
tomrf1, line 407: I want to suggest a custom hook for this, but with only 2 uses it's maybe not worth it!

tomrf1 (approved):
👍

oliverlloyd:
> `@guardian/libs` has `getLocale`. Maybe you’ll want to use that instead?

So we already use `getLocale` in our own abstraction [getCountryCode](https://github.com/guardian/dotcom-rendering/blob/oliver/refactor-countrycode/dotcom-rendering/src/web/lib/getCountryCode.ts#L39). But from what I can tell, all the features we're adding with this wrapper are already present in the original lib (caching, fallbacks, etc.) so it could well be the case that we can remove this completely and just call `getLocale`, which would be nice!

I don't want to introduce too many changes here though so I'll have one PR to move _where_ we call things and another to refactor _how_ we call them

dotcom-rendering/src/web/components/ReaderRevenueLinks.tsx
oliverlloyd: I have exactly the same instinct. Perhaps the real benefit to having a custom hook is in terms of dev ex? It might be nice to have a suite of hooks or functions that provide components with the state that they need to achieve their aims.

I'm going to draw up a list of these for review.

PR #3746, opened 2021-12-10
Add sticky key events for apps rendering live layout

OllysCoding:
## What does this change?

Make key events sticky for key events on AR live layout

## Why?

Parity with designs & dcr

### Before
image

### After
image


marjisound (approved):
LGTM

PR #3747, opened 2021-12-10
Refactor how we fetch the user profile

oliverlloyd:
## What does this change?
This PR moves the logic for how we fetch the reader's profile out of `App.tsx` and down into the local components that use it

## Why?
This is part of a wider effort to remove global state from `App.tsx`. See [this document](https://docs.google.com/document/d/1kSrZZrJ8hSw6bek5hGa9onE8QDz5TRHFqHV52EOFjts/edit#heading=h.gsod33a0pc47) for more details on this.

## Doesn't this mean we will fetch the profile twice?
No. Because we're using SWR, we will cache the first call and any subsequent fetch requests to the same endpoint will be served from this cache.

## Why `DiscussionContainer`?
This PR introduces `DiscussionContainer` and `DiscussionWhenSignedIn`. This is required because you [cannot conditionally call a react hook](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level). To get around this - and arguably to separate code concerns better - we have these additional components that manage deciding if the reader is signed in and another to fetch the profile.


oliverlloyd:

oliverlloyd:
Closing now that we've completed this work via separate [PR](https://github.com/guardian/dotcom-rendering/pull/3839)s

PR #3748, opened 2021-12-10
Upgrade @guardian/braze-components to v5.1.0

tjmw:
## What does this change?

Upgrades `@guardian/braze-components` to 5.1.0, which is now built with Node 14.

Have tested this on `CODE`.


OllysCoding (approved):
LGTM

PR #3749, opened 2021-12-10
Bump Source v4 release candidate

SiAdcock:
## Why?

We have published release candidates of the Source packages to test a new code structure in the Source repo. The API has not changed, just the internal code organisation.

We would like to validate this change in dotcom-rendering before publishing a stable v4 release.


oliverlloyd (approved):
All checks have passed 👍

PR #3751, opened 2021-12-13
Move liveblock header into common-rendering

lucymonie:
## What does this change?
This moves the header that was in the dotcom-rendering LiveBlock component into the common-rendering LiveBlockContainer component, so the header can be shared between AR and DCR.

## Why?
So we can share components between the two projects and reduce the maintenance cost.

[Trello card](https://trello.com/c/hwXjOTe0/387-lp-000%C6%92%CB%99-or-the-artist-formerly-known-as-move-liveblock-header-into-common-rendering)

### DIFF
--- a/apps-rendering/src/components/liveBlocks.tsx
+++ b/apps-rendering/src/components/liveBlocks.tsx
@@ -1,13 +1,10 @@
 // ----- Imports ----- //
 
 import { css } from '@emotion/react';
-import { FirstPublished } from '@guardian/common-rendering/src/components/FirstPublished';
 import { LastUpdated } from '@guardian/common-rendering/src/components/LastUpdated';
 import LiveBlockContainer from '@guardian/common-rendering/src/components/liveBlockContainer';
 import type { ArticleFormat } from '@guardian/libs';
-import { headline, remSpace } from '@guardian/source-foundations';
 import { OptionKind, partition } from '@guardian/types';
-import { maybeRender } from 'lib';
 import type { LiveBlock } from 'liveBlock';
 import type { FC } from 'react';
 import { renderAll } from 'renderer';
@@ -18,23 +15,6 @@ interface LiveBlocksProps {
 	format: ArticleFormat;
 }
 
-interface BlockTitleProps {
-	title: string;
-}
-
-const BlockTitle: FC<BlockTitleProps> = ({ title }) => {
-	return (
-		<h2
-			css={css`
-				${headline.xxsmall({ fontWeight: 'bold' })}
-				margin-bottom: ${remSpace[2]}px;
-			`}
-		>
-			{title}
-		</h2>
-	);
-};
-
 const LiveBlocks: FC<LiveBlocksProps> = ({ blocks, format }) => {
 	return (
 		<>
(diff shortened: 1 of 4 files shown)


apps-rendering/src/components/liveBlocks.tsx
```diff
@@ -43,33 +23,19 @@ const LiveBlocks: FC<LiveBlocksProps> = ({ blocks, format }) => {
 			{blocks.map((block) => {
 				// TODO: get page number
 				const blockLink = `${1}#block-${block.id}`;
+				const blockFirstPublished = block.firstPublished.kind === OptionKind.Some
```
JamieB-gu, line 26: This may be unnecessary if this changes to a `Date` (see below), but an alternative way to write this is:

```ts
pipe(
    block.firstPublished,
    map(Number),
    withDefault<number | undefined>(undefined),
)
```

Either are fine, I think it's up to you which you prefer 🙂. If this does change to a `Date`, the conversion from `Option` is simpler (and we should probably add it to `lib` if we're going to be doing it more often 🤔):

```ts
withDefault<Date | undefined>(undefined)(block.firstPublished)
```

In fact maybe we should anyway:

```ts
const toNullable = <A>(opt: Option<A>): A | undefined =>
	withDefault<A | undefined>(undefined)(opt)
```

So the original `pipe` becomes:

```ts
pipe(
    block.firstPublished,
    map(Number),
    toNullable,
)
```

JamieB-gu (commented):
This looks great!

A couple of things occurred to me while reading through so I've added them as comments, but I don't feel too strongly about either or feel they require any changes. They're more of an FYI/what do you think about this?

marjisound (approved):
It looks good Lucy. I also agree with the suggestion Jamie had and also the question about the type of `blockFirstPublished`.  There's also a few lint prettier  issues you can fix. To fix them just cd into apps-rendering. And then run `npm run lint:fix`

apps-rendering/src/components/liveBlocks.tsx
lucymonie: Thanks Jamie, I have had a play around with this. It certainly seems neater and more in keeping with AR to write it as above, so I have changed it. I can see your point about date but I'm less clear about how you intended that to be implemented.

OllysCoding (approved):
+1 LGTM :)

common-rendering/src/components/liveBlockContainer.tsx
```diff
@@ -1,20 +1,52 @@
...
+		</header>
+	);
+};
+
+const BlockTitle = ({ title }: { title: string }) => {
+	return (
+		<h2
+			css={css`
+				${headline.xxsmall({ fontWeight: 'bold' })}
+				margin-bottom: ${space[2]}px;
```
DavidLawes, line 28: I can't remember - what's the consensus for space vs remSpace in common-rendering?

common-rendering/src/components/liveBlockContainer.tsx
```diff
@@ -1,36 +1,77 @@
...
 	borderColour,
+	blockTitle,
+	blockFirstPublished,
+	blockLink,
 }: {
 	id: string;
 	children: React.ReactNode;
 	borderColour: string;
+	blockTitle?: string;
+	blockFirstPublished?: number;
```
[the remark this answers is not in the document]
DavidLawes: I believe I'm covering this change (`number` -> `date`) under a different PR: https://github.com/guardian/dotcom-rendering/pull/3724

I think whoever merges second will deal with the conflicts (probably that will be me :) )

DavidLawes (approved):
👍

ioannakok (approved):
+1 from me as well!

common-rendering/src/components/liveBlockContainer.tsx
lucymonie: Yes, I tried switching it over and then things started changing in unpredictable ways in Chromatic so I switched it back to space.

PR #3752, opened 2021-12-13
Create the mobile and mobile-preview stacks in CDK.  Map DNS to the new stacks.  CDK migration stage 2

waisingyiu:
## What does this change?

We defined the infrastructure using CDK in addition to the old cloudformation yaml. We include the definition of DNS record set in CDK and remove the corresponding definition from cloudformation yaml.

## Why?

We are migrating from cloudformation to CDK.

### Before

We defined our infrastructure using cloudformation yaml.

### After

We define our infrastructure using CDK. We still retain the old infrastructure but have the DNS mapping to the **new** load balancer so it is the **new** infrastructure that will be servicing requests.
It is the step 3 of CDK migration and we aim to set up parallel run of infrastructure so that we can verify the infra that is created by CDK.


groakland (approved):
Not to much to add here apart from the inconsistent whitespace in one of the describe() tests

PR #3754, opened 2021-12-13
AR pagination logic

marjisound:
Paired on this PR: @DavidLawes & @marjisound . Also valuable guidance from @JamieB-gu 
## What does this change?
This PR introduces pagination logic by adding the `pagedBlocks` to the live/dead blog models in AR which will be needed when the pagination component is introduced (in upcoming PRs). 

In regards to implementing pagination on apps-rendering deadblogs this is step 1 of 2. Step 1 introduces the pagination logic only. Step 2 will integrate the pagination component into the live layout (as well as considering how the pagination component can be shared in common-rendering).

We've designed the model for the pagination data to match (as closely as we can) the data structure used by dcr. This should improve the ease of sharing the pagination component with dcr.

Locally we validated the data model against liveblogs rendered via frontend and dcr: checking the page count, current page blocks and the block ids specified in the pagination object (i.e. newest, newer, older, oldest). The page count calculation and current page blocks always matched (for the testing carried out). The primary difference noticed was that for our change the blockId we navigate to (i.e. the string defined in newer, older or oldest) is always the first block of the page, whereas in frontend this is not always the case (i.e. the navigation buttons sometimes specify a block id that could the 2nd or 3rd block id in a given page). I'm not sure if this difference is significant.

Note that this change goes hand in hand with the associated MAPI change: https://github.com/guardian/mobile-apps-api/pull/1785 (in this change we allow the `page` query parameter to be passed through to apps-rendering)

### What does this not do?

- We haven't yet optimised the pagination query to get liveblocks if the reader lands on the first page. There is some more logic to add (to both MAPI and AR) in order to support this. Given this affects both repos I think it's worth separating the optimisation into a different PR - this current PR already contains a lot of logic.

- We haven't yet added the pagination navigation component to the live layout - it's intended to be delivered as part of a separate change (including deciding how we share the Pagination component with DCR).

Outstanding tasks related to pagination logic have been captured on trello: https://trello.com/c/3PIfKTMB

## Changes:
1. Handle `page` query parameter in request
2. Update live/dead blog models in AR
3. Create pagination.ts to return pagination detail and current page detail based on the block id 

## Why?
This is the first step towards enabling pagination for live/dead blogs in AR (currently the AR-rendered liveblogs show every liveblock and has no concept of 'pages').

### Before/After
There are no page elements added/changed/deleted as a result of this PR - this change is only responsible for manipulating the data.


DavidLawes:
[Support liveblocks in AR - pagination logic](https://trello.com/c/4eK7raSA/469-support-liveblocks-in-ar-pagination-logic)

apps-rendering/src/item.ts
```diff
@@ -323,24 +327,33 @@ const isPicture = hasTag('type/picture');
...
 
+		const parsedBlocks = parseLiveBlocks(body)(context);
+		const pagedBlocks = getPagedBlocks(pageSize, parsedBlocks, blockId);
 		return {
 			design:
 				content.fields?.liveBloggingNow === true
 					? ArticleDesign.LiveBlog
 					: ArticleDesign.DeadBlog,
-			blocks: parseLiveBlocks(body)(context),
+			blocks: parsedBlocks,
```
DavidLawes, line 347: I _think_, when we come to implementing the Pagination component in apps-rendering, we'll find that we don't need both `parsedBlocks` and `pagedBlocks`. Possibly we'll change to `blocks: pagedBlocks` because we only want to render the page of blocks, not all the blocks.

apps-rendering/src/item.ts
```diff
@@ -323,24 +327,33 @@ const isPicture = hasTag('type/picture');
...
 const fromCapiLiveBlog =
 	(context: Context) =>
-	(request: RenderingRequest): LiveBlog | DeadBlog => {
+	(
+		request: RenderingRequest,
+		blockId: Option<string>,
+	): LiveBlog | DeadBlog => {
 		const { content } = request;
 		const body = content.blocks?.body ?? [];
+		const pageSize = content.tags.map((c) => c.id).includes('sport/sport')
```
DavidLawes, line 336: We've taken the page size definition from frontend: https://github.com/guardian/frontend/blob/main/article/app/model/LiveBlogHelpers.scala#L45

apps-rendering/src/pagination.ts
```diff
@@ -0,0 +1,146 @@
...
+	older: Option<string>;
+	numberOfPages: number;
+};
+
+export type LiveBlogPagedBlocks = {
+	currentPage: PageReference;
+	pagination: Pagination;
+};
+
+const navContainerId = 'liveblog-navigation';
```
DavidLawes, line 25: I am tempted to remove this from pagination.ts and possibly include it in the to-be-implemented `Pagination` component. All we're trying to achieve here is scrolling to the pagination navigation container when going to any page (other than the first page). I wonder if it seems strange to reference a container id here, when possibly it's more related to the `Pagination` component itself (which will include the navigation container and therefore define this id).

apps-rendering/src/pagination.ts
```diff
@@ -0,0 +1,146 @@
+import { map, none, some, withDefault } from '@guardian/types';
+import type { Option } from '@guardian/types';
+import { pipe } from 'lib';
+import type { LiveBlock } from 'liveBlock';
+
+export type PageReference = {
```
DavidLawes, line 6: Here we can see a small divergence the `PageReference` interface compared to what's been defined in frontend (https://github.com/guardian/frontend/blob/main/common/app/model/LiveBlogCurrentPage.scala#L180) - here we don't define the `isArchivePage` boolean. I decided to remove this from the interface:
- In frontend, I think isArchivePage was used for appending the liveblog navigation container id to the navigation href links (https://github.com/guardian/frontend/blob/main/article/app/views/fragments/liveBlogNavigation.scala.html#L32)
- If I've understood this correctly, I feel like `isArchivePage` doesn't really describe why we're defining this property (I personally didn't find the name intuitive, even after understanding the purpose)
- In ar/dcr I think we can implement this same functionality (i.e. scrolling to a particular container id) in a slightly easier way (possibly inside the `Pagination` component itself) without the need for this extra property.

apps-rendering/src/pagination.ts
```diff
@@ -0,0 +1,146 @@
...
+	pageNumber: number,
+): Option<string> => {
+	if (pageNumber > 1) {
+		return some(`?`);
+	}
+
+	return none;
+};
+
+const getFirstPageLength = (pageSize: number, blocks: LiveBlock[]): number => {
```
DavidLawes, line 76: We want the final page of a liveblog to always equal the designated page size (either 10 or 30 blocks). For example, if we have a page size of 10 and 21 blocks in the liveblog, we don't want 3 pages (10 blocks, 10 blocks, 1 block) we only want 2 pages (11 blocks, 10 blocks).


As such, we need the length of blocks in the first page to flex.

apps-rendering/src/pagination.ts
```diff
@@ -0,0 +1,146 @@
...
+	const remainder = blocks.length % pageSize;
+	return pageSize + remainder;
+};
+
+const getPages = (
```
DavidLawes, line 81: This function does a lot of the groundwork for us: it's returning us an array of liveblock arrays that represent the pages of the liveblock. When we looked at the CAPI data model we were confident that the liveblocks are returned in order (i.e. newest first) so there wasn't any need to sort the liveblocks before creating pages.

apps-rendering/src/server/server.ts
```diff
@@ -240,6 +251,13 @@ async function serveRichLinkDetails(
 	res.status(200).json(response);
 }
 
+export const getPageId = (query: qs.ParsedQs): Option<string> => {
```
DavidLawes, line 254: I'm wondering whether a function name of `getBlockId` would be more descriptive here (and subsequent variable names of `blockId` too)?

apps-rendering/src/pagination.ts
```diff
@@ -0,0 +1,146 @@
...
+};
+
+const getCurrentPage = (
+	pages: LiveBlock[][],
+	blockId: Option<string>,
+): PageReference => {
+	const firstPage = {
+		blocks: pages[0],
+		pageNumber: 1,
+		suffix: '',
```
DavidLawes, line 104: I think for now (i.e. before Journalism's team on filtered key events) the suffix for the first page will always be an empty string

apps-rendering/src/pagination.ts
```diff
@@ -0,0 +1,146 @@
...
+
+	return pipe(
+		blockId,
+		map((id) => {
+			for (let i = 0; i < pages.length; i += 1) {
+				if (pages[i].some((x) => x.id === id)) {
+					return {
+						blocks: pages[i],
+						pageNumber: i + 1,
+						suffix: '',
```
DavidLawes, line 115: To check: will suffix always be an empty string if not the first page? I think possibly we might use suffix to pass through any existing query params (i.e. page with blockId?)

apps-rendering/src/pagination.ts
DavidLawes: Actually, good spot from @marjisound - we possibly do need to consider ordering the blocks (https://github.com/guardian/frontend/blob/2996c4465a09ed108c64e7d12c0b35d042dbcce7/common/app/model/liveblog/BodyBlock.scala#L17)

DavidLawes: I spoke with the CAPI team about ordering of blocks. The CAPI team does not apply any ordering on the blocks, they send in the same order that Composer provides them (I suppose this is where we could possibly introduce out-of-order blocks). 

For now, I'm not going to introduce sorting of blocks - instead I'm going to try and evaluate liveblog blocks to see if I can find examples of any out-of-order blocks. I'll also check with the developers who originally introduced sorting into frontend (https://github.com/guardian/frontend/pull/13566).

apps-rendering/src/pagination.ts
DavidLawes: This has been updated: the suffix now returns a string representation of all query params for the current page.

apps-rendering/src/pagination.ts
DavidLawes: I removed this as it didn't 'feel' right - given the 'scroll to pagination navigation' feature isn't yet working in DCR, it might be that we can support this in the Pagination component itself (providing support for both AR+DCR at the same time)

apps-rendering/src/pagination.ts
DavidLawes: We have ~28k articles that have tag `tone/minutebyminute` (`https://content.guardianapis.com/search?tag=(tone/minutebyminute)&to-date=2022-01-12&show-blocks=body`). I sampled 1000 of these articles that have this tag:
- I performed a simple sort on the liveblog blocks (order descending by firstPublishedDate)
- I compared the ordered blocks to the blocks we receive from CAPI

All 1000 blogs I sampled had the expected ordered. I'll expand the test to see if I can find any examples where the blocks are not ordered as expected.

JamieB-gu (approved):
This is excellent! As previously mentioned, I think the `pagination` module is really well thought out: defining the problem domain as a series of types and then gluing it all together with functions that operate on those types 👍 🚀 

I have written some comments, but there's nothing I think is a blocker, so 👍 🙂

apps-rendering/src/pagination.ts
```diff
@@ -0,0 +1,168 @@
+import { andThen, map, none, some, withDefault } from '@guardian/types';
+import type { Option } from '@guardian/types';
+import { compose, index, pipe } from 'lib';
+import type { LiveBlock } from 'liveBlock';
+
+export type PageReference = {
```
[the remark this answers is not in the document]
DavidLawes: Nice idea - I've added jsdocs for PageReference, Pagination and LiveBlogPagedBlocks :)

apps-rendering/src/pagination.ts
```diff
@@ -0,0 +1,168 @@
...
+	blocks: LiveBlock[];
+	pageNumber: number;
+	suffix: string;
+};
+
+export type Pagination = {
```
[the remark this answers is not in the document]
DavidLawes: Ah - this is a great suggestion :D Our end-state is to have a shared `Pagination` component so we initially created our data types as close as we could to frontend/dcr. I think this kind of abstraction (sum type) could indeed make sense, however I'd like to defer this decision until we've thought more about how we render `Pagination`. I've included a link to this comment on the pagination implementation checklist so we remember to consider this when we render the pagination navigation (https://trello.com/c/3PIfKTMB)... I hope that's ok!

apps-rendering/src/pagination.ts
```diff
@@ -0,0 +1,168 @@
...
+};
+
+const maybePageRef = (
+	pageIndex: number,
+	pages: LiveBlock[][],
+): Option<string> => {
+	const maybeBlocks = (pageIndex: number): Option<LiveBlock[]> =>
+		index(pageIndex)(pages);
+	const maybeFirstBlock = (blocks: LiveBlock[]): Option<LiveBlock> =>
+		index(0)(blocks);
```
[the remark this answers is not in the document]
DavidLawes: Nice, thanks!

apps-rendering/src/pagination.ts
```diff
@@ -0,0 +1,168 @@
...
+				if (blocks.some((x) => x.id === id)) {
+					return {
+						blocks,
+						pageNumber: i + 1,
+						suffix: currentPageRef(pages, i),
+					};
+				}
+			}
+			return firstPage;
+		}),
```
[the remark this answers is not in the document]
DavidLawes: Yes, this is neater - thanks :)

apps-rendering/src/pagination.ts
```diff
@@ -0,0 +1,168 @@
...
+	const maybeFirstBlock = (blocks: LiveBlock[]): Option<LiveBlock> =>
+		index(0)(blocks);
+	const pageRef = (block: LiveBlock): string =>
+		`?page=with:block-${block.id}`;
+
+	const firstBlock = compose(
+		andThen(maybeFirstBlock),
+		maybeBlocks,
+	)(pageIndex);
+	return map(pageRef)(firstBlock);
```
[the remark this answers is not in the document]
DavidLawes: This is great, thanks - I'd initially started using `pipe` but needed to break things out into smaller chunks first to wrap my head around things. Overall I think `pipe` is easier to read so I've updated my function to use `pipe` (I've retained my named functions).

PR #3756, opened 2021-12-14
feat: Reduce the length of the relative time string for key events

oliverlloyd:
## What does this change?
Replaces `long` with `med` for the `time` element used on key events. This means the relative string shown will be formatted as '2ms ago' instead of '2 minutes ago'

### Before
Screenshot 2021-12-14 at 15 03 57

### After
Screenshot 2021-12-14 at 15 01 38


DavidLawes (approved):
👍

ioannakok (approved):
👍

PR #3757, opened 2021-12-14
Fix colours for Series and Standfirst in Deadblogs in apps-rendering

ioannakok:
## What does this change?
Fixes colours for Series and Standfirst in Deadblogs in AR for News, Culture and Lifestyle.

## Why?
To follow the design requirements

### Changes
| Before | After |
| ----------- | ----------- |
| image | image |
| image | image |
| image | image |

#### On hover
image

image

image

### DIFF
--- a/apps-rendering/src/components/series.tsx
+++ b/apps-rendering/src/components/series.tsx
@@ -3,7 +3,7 @@
 import type { SerializedStyles } from '@emotion/react';
 import { css } from '@emotion/react';
 import { text } from '@guardian/common-rendering/src/editorialPalette';
-import type { ArticleFormat } from '@guardian/libs';
+import type { ArticleFormat, ArticleTheme } from '@guardian/libs';
 import { ArticleDesign, ArticleDisplay, ArticleSpecial } from '@guardian/libs';
 import {
 	from,
@@ -26,9 +26,8 @@ interface Props {
 	item: Item;
 }
 
-const standardLinkStyles = (format: ArticleFormat): SerializedStyles => {
-	const kicker = text.kicker(format);
-	const inverted = text.inverted(format);
+const standardLinkStyles = (theme: ArticleTheme): SerializedStyles => {
+	const { kicker, inverted } = getThemeStyles(theme);
 
 	return css`
 		${headline.xxxsmall({ lineHeight: 'loose', fontWeight: 'bold' })}
@@ -41,13 +40,13 @@ const standardLinkStyles = (format: ArticleFormat): SerializedStyles => {
 	`;
 };
 
-const labsLinkStyles = (format: ArticleFormat): SerializedStyles => css`
+const labsLinkStyles = (theme: ArticleTheme): SerializedStyles => css`
 	${textSans.medium({ lineHeight: 'loose', fontWeight: 'bold' })}
 	color: ${labs[300]};
 	text-decoration: none;
 
 	${darkModeCss`
-		color: ${text.inverted(format)};
+		color: ${getThemeStyles(theme).inverted};
 	`}
 `;
 
@@ -84,11 +83,7 @@ const getLinkStyles = ({
 	}
 
 	if (theme === ArticleSpecial.Labs) {
-		return labsLinkStyles({
-			design,
-			display,
-			theme,
-		});
+		return labsLinkStyles(theme);
 	}
 
 	if (
@@ -98,18 +93,14 @@ const getLinkStyles = ({
 		return blogLinkStyles({ design, display, theme });
 	}
 
-	return standardLinkStyles({
-		design,
-		display,
-		theme,
-	});
+	return standardLinkStyles(theme);
 };
 
-const immersiveStyles = (format: ArticleFormat): SerializedStyles => css`
+const immersiveStyles = (theme: ArticleTheme): SerializedStyles => css`
 	padding: ${remSpace[1]} ${remSpace[3]};
-	background-color: ${format.theme === ArticleSpecial.Labs
+	background-color: ${theme === ArticleSpecial.Labs
 		? labs[300]
-		: text.kicker(format)};
+		: getThemeStyles(theme).kicker};
 	position: absolute;
 	left: 0;
 	transform: translateY(-100%);
@@ -138,13 +129,7 @@ const getStyles = ({
 	theme,
 }: ArticleFormat): SerializedStyles => {
 	if (display === ArticleDisplay.Immersive) {
-		return css(
-			immersiveStyles({
-				design,
-				display,
-				theme,
-			}),
-		);
+		return css(immersiveStyles(theme));
 	}
 
 	if (
--- a/apps-rendering/src/components/series.tsx
+++ b/apps-rendering/src/components/series.tsx
@@ -61,15 +61,11 @@ const immersiveLabsLinkStyles = css`
 	${textSans.medium({ lineHeight: 'loose', fontWeight: 'bold' })}
 `;
 
-const liveLinkStyles = (theme: ArticleTheme): SerializedStyles => css`
+const blogLinkStyles = (format: ArticleFormat): SerializedStyles => css`
 	${headline.xxxsmall({ lineHeight: 'tight', fontWeight: 'bold' })}
-	color: ${getThemeStyles(theme).liveblogKicker};
-	text-decoration: none;
-`;
-
-const deadBlogLinkStyles = (format: ArticleFormat): SerializedStyles => css`
-	${headline.xxxsmall({ lineHeight: 'tight', fontWeight: 'bold' })}
-	color: ${text.series(format)};
+	color: ${format.design === ArticleDesign.LiveBlog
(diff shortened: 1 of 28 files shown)


ioannakok:
[AR - fix colours for Deadblog](https://trello.com/c/KCbwcbVz/440-ar-fix-colours-for-deadblog)

apps-rendering/src/components/series.tsx
```diff
@@ -66,6 +66,12 @@ const liveLinkStyles = (theme: ArticleTheme): SerializedStyles => css`
 	text-decoration: none;
 `;
 
+const deadBlogLinkStyles = (theme: ArticleTheme): SerializedStyles => css`
+	${headline.xxxsmall({ lineHeight: 'tight', fontWeight: 'bold' })}
+	color: ${getThemeStyles(theme).deadblogKicker};
```
DavidLawes, line 71: Is there an abstraction where we have a function that accepts both`Design` and `Theme` (or is that part of an ongoing wider discussion)? Agree that here it makes sense to duplicate the linkStyles function for deadBlogs, but wondering how this might scale for other variations between `Design`.

DavidLawes (approved):
Nice - might want to get a review from Jamie/Marjan as well in case I've missed something AR-specific :)

apps-rendering/src/components/standfirst.stories.tsx
```diff
@@ -1,10 +1,15 @@
...
+import { ArticleDesign, ArticleDisplay, ArticlePillar } from '@guardian/libs';
 import { boolean, withKnobs } from '@storybook/addon-knobs';
 import { article, comment, feature, review } from 'fixtures/item';
+import { deadBlog } from 'fixtures/live';
 import type { ReactElement } from 'react';
 import { selectPillar } from 'storybookHelpers';
+import {
+	getAllThemes,
+	getThemeNameAsString,
+} from '../../../common-rendering/src/fixtures/article';
```
JamieB-gu, line 12: I think you can import this with the `guardian` prefix if you want? See `headerImage.tsx` for an example.

apps-rendering/src/themeStyles.ts
```diff
@@ -19,6 +19,8 @@ interface ThemeStyles {
 	liveblogKicker: string;
 	liveblogBackground: string;
 	liveblogDarkBackground: string;
+	deadblogKicker: string;
+	deadblogStandfirstLink: string;
```
JamieB-gu, line 23: I'd recommend adding these to `editorialPalette` rather than `themeStyles`.

- `themeStyles` is a legacy holdover from the old templates that is unfortunately still scattered throughout our codebase. It only understands switching colour based on `Theme` (previously `Pillar`).

- `editorialPalette` is a much more comprehensive way of selecting colours based on the full `Format` object, and mirrors the functionality of `decidePalette` (we'll hopefully be able to fold them together at some point).

Thinking about where they'd fit into `editorialPalette`:

- `deadblogStandfirstLink`: I think we follow the convention already laid down by `decidePalette`, and call this `text.standfirstLink`.
- `deadblogKicker`: I'm less clear here because there appears to be no `Series` component in DCR 🤔. I'm not sure whether the `Kicker` component and `text.linkKicker` are approximate equivalents. Regardless I'd be inclined to call this `text.series`.

JamieB-gu (commented):
Nice! 👍

One recommendation about the best place to capture palette colours.

apps-rendering/src/components/standfirst.tsx
```diff
@@ -93,6 +94,25 @@ const liveblogStyles: SerializedStyles = css`
...
 
+const deadblogStyles = (format: ArticleFormat): SerializedStyles => css`
+	${headline.xxxsmall({ fontWeight: 'bold' })};
+
+	a {
+		text-decoration: none;
+	}
+
+	a:link {
+		color: ${getThemeStyles(format.theme).deadblogStandfirstLink};
```
lucymonie, line 105: Very minor suggestion - you could save this as a variable and reuse it in the three cases here to save repetition.

lucymonie (approved):
Looks good!

common-rendering/src/editorialPalette.ts
```diff
@@ -108,6 +112,47 @@ const textStandfirstDark = ({ design }: ArticleFormat): Colour => {
...
+	switch(format.design) {
+		case ArticleDesign.LiveBlog:
+			return neutral[100];
+		case ArticleDesign.DeadBlog:
+			return blogsGrayBackgroundPalette(format);
+		default: {
+			if(format.theme == ArticleSpecial.Labs)
+				return labs[300];
+
+			const { kicker, inverted } = getThemeStyles(format.theme);
```
JamieB-gu, line 145: Can we specify these colour rules in this file instead? Unfortunately there are a couple of reasons we might want to avoid using `getThemeStyles`:

- We'd like to avoid importing from DCR and AR into common-rendering; the dependency tree should be the other way around, i.e. DCR and AR import _from_ common-rendering
- I think it would be best to replace and eventually remove the theme styles logic, and we'd like to replace it with the logic in _this_ file - so taking logic from there and putting it here would be 👍

JamieB-gu (commented):
Some really nice refactors in this PR, particularly in the renderer! 🚀

One final comment.

common-rendering/src/editorialPalette.ts
```diff
@@ -87,6 +92,44 @@ const textHeadlineDark = (format: ArticleFormat): Colour => {
...
+		case ArticlePillar.Sport:
+			return sport[500];
+		case ArticlePillar.Opinion:
+			return opinion[500];
+		case ArticleSpecial.Labs:
+			return specialReport[500];
+		case ArticleSpecial.SpecialReport:
+			return specialReport[500];
+	}
+}
```
JamieB-gu, line 131: Will people know what `kicker` and `inverted` mean? Are they holdover terms from the old way of doing things? Can the switch statements in `textKicker` and `textInverted` be inlined into the `textStandfirstLink` function instead?

I think one reason for using `decidePalette` and `editorialPalette` is that it's very clear what a colour is and where it's used. That way it's easier for people, Design included, to come in and tweak a colour in isolation, without worrying about the change affecting multiple things.

Besides, I think there's already some discrepancy: `textStandfirstLink` says labs pieces should use `labs[300]` but `textKicker` says it should be `labs[400]` 🤔.

common-rendering/src/editorialPalette.ts
```diff
@@ -87,6 +92,44 @@ const textHeadlineDark = (format: ArticleFormat): Colour => {
...
+		case ArticlePillar.Culture:
+			return culture[500];
+		case ArticlePillar.Lifestyle:
+			return lifestyle[500];
+		case ArticlePillar.Sport:
+			return sport[500];
+		case ArticlePillar.Opinion:
+			return opinion[500];
+		case ArticleSpecial.Labs:
+			return specialReport[500];
```
JamieB-gu, line 127: Labs pieces use colours from the `specialReport` palette? I'm aware this comes from the old `themeStyles` module, I'm just surprised... 🤔 

@HarryFischer this will be the colour of links in the standfirst on Guardian labs media pieces. Is this correct? Should they be using something from the labs palette instead?

common-rendering/src/editorialPalette.ts
```diff
@@ -108,6 +151,48 @@ const textStandfirstDark = ({ design }: ArticleFormat): Colour => {
...
+const textStandfirstLink = (format: ArticleFormat): Colour => {
+
+	switch(format.design) {
+		case ArticleDesign.LiveBlog:
+			return neutral[100];
+		case ArticleDesign.DeadBlog:
+			return blogsGrayBackgroundPalette(format);
+		default: {
+			if(format.theme == ArticleSpecial.Labs)
+				return labs[300];
```
JamieB-gu, line 182:
```suggestion
			if (format.theme === ArticleSpecial.Labs) {
				return labs[300];
		    }
```

JamieB-gu (commented):
Nice! Couple of questions/suggestions about the structure of `editorialPalette` (sorry 😬!).

common-rendering/src/editorialPalette.ts
ioannakok: If we want to have `labs[500]`, we will have to make a change in `source`: https://github.com/guardian/source/blob/main/packages/@guardian/source-foundations/src/palette.ts#L260-L263.

common-rendering/src/editorialPalette.ts
```diff
@@ -108,6 +111,84 @@ const textStandfirstDark = ({ design }: ArticleFormat): Colour => {
...
+				case ArticleSpecial.Labs:
+					return labs[400];
+				case ArticleSpecial.SpecialReport:
+					return specialReport[400];
+			}
+		}
+	}
+}
+
+const textSeriesTitle = (format: ArticleFormat): Colour => {
```
ioannakok, line 173: Renaming this from `textSeries` to `textSeriesTitle` to match `decidePalette`.

PR #3761, opened 2021-12-15
fix: Don't hide standfirst in print view

oliverlloyd:
## What does this change?
We [removed the standfirst from the print view previously](https://github.com/guardian/dotcom-rendering/pull/2234) but it's part of the article and should be included so this PR ensures that it appears

| BEFORE | AFTER |
|--------|-------|
| Screenshot 2021-12-15 at 13 20 58   | Screenshot 2021-12-15 at 13 25 52  |

### DIFF
--- a/dotcom-rendering/src/web/components/Standfirst.tsx
+++ b/dotcom-rendering/src/web/components/Standfirst.tsx
@@ -189,7 +189,6 @@ export const Standfirst = ({ format, standfirst }: Props) => {
 
 	return (
 		<div
-			data-print-layout="hide"
 			css={[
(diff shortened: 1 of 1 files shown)


ioannakok (approved):
👍

lucymonie (approved):
Is there any explanation for why it was hidden?

oliverlloyd:
> Is there any explanation for why it was hidden?

Looking at the original PR I think it got added by mistake, or maybe there was a discussion and it was decided to remove it? It's hard to say, but from feedback it does seem fair to say the standfirst is a valid part of the particle that should be shown in print view.

PR #3763, opened 2021-12-16
Fix banner geolocation targeting

tomrf1:
This was broken by https://github.com/guardian/dotcom-rendering/pull/3743
It's currently passing through a function rather than a Promise containing the geolocation

### Before:
Screen Shot 2021-12-16 at 08 45 32

### After:
Screen Shot 2021-12-16 at 08 42 52

We should really have some tests in DCR to prevent this type of regression


oliverlloyd (approved):
👍 

Aside: Ideally, we would not pass this in at all. The receiving function could import and call `getLocaleCode` itself

PR #3766, opened 2021-12-16
Documentation For Common-Rendering `lib` Functions

JamieB-gu:
## Why?

@OllysCoding mentioned that some of the `lib` functions could do with documentation to describe their use. This PR adds said documentation to some of the functions in `common-rendering`'s `lib.ts`.

## Changes

- Added documentation for `pipe` and `maybeRender`


marjisound (approved):
Very useful, thanks Jamie 👍

PR #3767, opened 2021-12-17
🏖 `Portal`

oliverlloyd:
## What does this change?
This PR introduces a new way to add portals to a page. A 'Portal' is the name given to any content added to the page on the client. This is similar concept to [an island](https://github.com/guardian/dotcom-rendering/pull/3629), but with the difference that the content is not server side rendered.

## Why?
The primary goal for this work is to improve the developer experience of teams working on DCR. Adding content client side previously required an understanding of the wider app architecture and edits to several files. This is now reduce to one api in one location.

In addition, we're making it easier to move away from the `App.tsx` pattern as [discussed here](https://docs.google.com/document/d/1gHOzvFZxrAO2mvNPKg4UmMAYZm7Jk1NutZPVLY5UxPY/edit#heading=h.1y29fcc1ygqg).

### Only send the data required
By serialising the props for each component alongside its declaration on the server and not in a global window object, we ensure we only send the data that is needed.

### Lazy insertion
Using the same [api that was introduced for lazy hydration](https://github.com/guardian/dotcom-rendering/pull/3629), we can now download and render components when visible, deferring content below the fold until scrolled into view. We can also tell the browser to render less important content when it is idle, preventing the thread from being blocked.

## `LegacyPortal`
I wanted to use the `Portal` namespace and also didn't want two things with similar names confusing people so I've sort of deprecated the old component by calling it `LegacyPortal` and stolen it's name

## Usage
```jsx
// Server
<Portal
    componentName="Onwards"
    when="visible"
    props={{
        onwardsUrl: CAPI.onwardUrl,
        format,
    }}
    placeholderHeight={400}
/>

```

This is the only code developers need to write to have their content inserted on the client*

*You also need to rename the component being used to include 'importable'. For example, `Onwards.importable.tsx`.


dotcom-rendering/src/web/components/Portal.tsx
```diff
@@ -1,26 +1,31 @@
...
+import { Placeholder } from './Placeholder';
 
-type Props = {
-	rootId: string;
-	children: React.ReactNode;
-};
+interface Props {
+	componentName: string;
+	when?: 'immediate' | 'idle' | 'visible';
+	props?: any;
```
oliverlloyd, line 6: Is there a way I can use `componentName` to get the expected type and use that here? This seems pretty wild but would really help improve the dev ex!

oliverlloyd: @JamieB-gu @mxdvl ?

dotcom-rendering/index.d.ts
```diff
@@ -690,8 +690,8 @@ interface CommercialConfigType {
 }
 
 type ServerSideTests = {
-    [k: `${string}Variant`]: "variant";
-    [k: `${string}Control`]: "control";
+	[k: `${string}Variant`]: 'variant';
+	[k: `${string}Control`]: 'control';
```
oliverlloyd, line 694: Prettier 🤷

dotcom-rendering/index.d.ts
```diff
@@ -1071,5 +1071,11 @@ declare namespace JSX {
 			props: any;
 			children: React.ReactNode;
 		};
+		'gu-portal': {
+			name: string;
+			when?: 'immediate' | 'idle' | 'visible';
+			props: any;
+			children: React.ReactNode;
```
oliverlloyd, line 1078: `children` is for when we use a `Placeholder`

dotcom-rendering/src/web/browser/getName.ts
```diff
@@ -3,11 +3,11 @@
  *
  * We expect the element to always be a `gu-*` custom element
  *
- * @param marker : The html element that we want to read the name attribute from;
+ * @param element : The html element that we want to read the name attribute from;
```
oliverlloyd, line 6: Respecting a comment made by @sndrs [over in libs](https://github.com/guardian/libs/pull/297#discussion_r755031903) for consistency

dotcom-rendering/src/web/components/Portal.tsx
mxdvl: You mean read it from the existing files? Yes, but you’d have to have some sort of process that generates a declaration file from reading the filesystem, I think.

That’s what the interview tool does in [`build.ts`](https://github.com/guardian/interview-questions/blob/main/build.ts)

dotcom-rendering/index.d.ts
mxdvl: Is it, though? 😆

dotcom-rendering/src/web/browser/hydration/init.ts
```diff
@@ -2,11 +2,11 @@ import '../webpackPublicPath';
...
 import { log } from '@guardian/libs';
-import { whenVisible } from './whenVisible';
-import { whenIdle } from './whenIdle';
+import { whenVisible } from '../whenVisible';
+import { whenIdle } from '../whenIdle';
 import { doHydration } from './doHydration';
-import { getName } from './getName';
-import { getProps } from './getProps';
+import { getName } from '../getName';
+import { getProps } from '../getProps';
```
oliverlloyd, line 9: I moved these to the common root folder so they can be shared. We can consider a different folder structure, this just worked for now

dotcom-rendering/src/web/components/LegacyPortal.tsx
```diff
@@ -0,0 +1,26 @@
...
+	rootId: string;
+	children: React.ReactNode;
+};
+
+// See the preact portal code for details on how this feature works
+// In particular, we can see that portals can safely be recreated and
+// that will not cause the child component being passed in to be mounted
+// again. Instead it is simply rerendered
+// https://github.com/preactjs/preact/blob/df748d106fb78fbd46d14563b4712f921ccf0300/compat/src/portals.js
+export const LegacyPortal = ({ rootId, children }: Props) => {
```
oliverlloyd, line 13: This file was renamed without edits

dotcom-rendering/src/web/components/Portal.tsx
mxdvl: I don’t think it’s a lot of work, and that file has another added benefits: It’s a live document containing all the current portals. I’m happy to set half-an-hour aside to have a crack at this.

The question that remains is how the generation steps gets called, but I’m sure there are precendents… `make portals`?

oliverlloyd: 🤔 I'm not 100% following your intention but would what you're thinking of solve the dev ex problem? That is, I want to insert a new Portal and type

<Portal componentName="MyThing" ...

And then when I go to enter the `props` they are already typed as needed?

dotcom-rendering/src/web/components/Portal.tsx
```diff
@@ -1,26 +1,31 @@
...
+ * @param when - When insertion should take place.
+ * 		- immediate - Insert without delay
+ * 		- idle - Insert when browser idle
+ * 		- visible - Insert when the gu-portal element appears in viewport
+ * @param componentName - The name of the component to insert
+ * @param props - An object containing the props for the component
+ * @param placeholderHeight - Height in pixels. If provided, a Placeholder is server rendered
+ *
+ */
+export const Portal = ({
```
SiAdcock, line 22: The _Naming Things is Hard_  question: is it confusing to repurpose the name Portal, which is a [first-class concept in (P)react](https://reactjs.org/docs/portals.html) that does something similar-ish but not really?

oliverlloyd: I guess what this component is doing is marking a place in the server side rendered content and saying, put some html _here_ on the client using these attributes.

I quite like `RenderOnClient` although it is perhaps a bit verbose? What about `Render`? That nicely aligns with the `ReactDOM.hydrate` vs `ReactDOM.render`.

SiAdcock: Cool, thanks @oliverlloyd, I found this helpful in clarifying the purpose of this component 🙂

You're right that `Render` does align nicely with that React method, assuming people realise that this component name is a reference to that. Personally, I see the word `Render` appearing more and more, and hence meaning less and less! The naming of React's `render` method hasn't scaled well, which is why we see more verbose names like `renderToString` and `renderToStaticMarkup` crop up later. This is why I prefer the more verbose name. 

I'm not sure if that's a _me_ thing, and I would be interested to hear other opinions.

oliverlloyd: Yeah, totally. I'm like you, I have my preference but that's based on my own feelings. `Render` is aligned and concise but it could could also be confusing.

A better way to choose a name should be with wider input and a focus on what will make sense to new people coming to the code with less context.

Aside. It's perhaps interesting to think about how Astro [solved this problem](https://docs.astro.build/core-concepts/component-hydration/index.html#mycomponent-clientonly-). They continued to use their `client` api and incorporated this scenario into that. That wouldn't work for as but maybe there are other examples for similar abstractions we can align with?

oliverlloyd: Thinking this api through some more, I'm starting to think a better approach might be to fold this logic into `Hydrate` and not have `Portal` at all. We simply could ask Hydrate to not always server render it's children, which essentially has the same effect as what this PR is proposing.

The api for `Hydrate` might look something like

```ts
<Hydrate when="immediate" where="client">
  <MyClientOnlyThing />
</Hydrate>
```

or

```ts
<Hydrate when="immediate" ssr={false}>
  <MyClientOnlyThing />
</Hydrate>
```

oliverlloyd:
I'm marking this PR as draft while we consider if we want to change the Hydrate api instead

oliverlloyd:
Closing in favour of https://github.com/guardian/dotcom-rendering/pull/3784

PR #3768, opened 2021-12-17
fix: Disable Chromatic for these third party embeds

oliverlloyd:
## What does this change?
Turns off Chromatic for these stories

## Why?
They keep causing diffs on PRs but it's just because the third party content they are loading changes, not our content


ioannakok (approved):
👍

PR #3770, opened 2021-12-17
feat: Added make cmd for running without logging

oliverlloyd:
## What does this change?
Adds `make run-ci`

## Why?
This runs the same commands as `make run` (Starts a production server) but it does so without the logging, preventing error


ioannakok (approved):
👍

PR #3772, opened 2021-12-17
Update match tabs logic to display at the correct time

rcrphillips:
## What does this change?
This adds the logic to only show the match-tabs component at the right time. We have a production issue whereby the component is showing on match report articles when there is no accompanying minute by minute. This will also mean that we end up displaying the tab correctly when we have a LIVE liveblog that does not yet have a match report associated to it. 

It also ensures that we make the correct tab have the 'active' state.

### Before - incorrect active tab
Screen Shot 2021-12-17 at 16 53 21

### After - incorrect active tab
Screen Shot 2021-12-17 at 16 52 49

### DIFF
--- a/dotcom-rendering/src/web/layouts/LiveLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/LiveLayout.tsx
@@ -689,7 +689,9 @@ export const LiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
 								</GridItem>
 								<GridItem area="meta" element="aside">
 									<Hide until="desktop">
-										<div css={[maxWidth, sidePaddingDesktop]}>
+										<div
+											css={[maxWidth, sidePaddingDesktop]}
+										>
 											<ArticleMeta
 												branding={branding}
 												format={format}
--- a/dotcom-rendering/src/web/layouts/LiveLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/LiveLayout.tsx
@@ -689,7 +689,7 @@ export const LiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
 								</GridItem>
 								<GridItem area="meta" element="aside">
 									<Hide until="desktop">
-										<div css={maxWidth}>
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -689,7 +689,9 @@ export const LiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
 								</GridItem>
 								<GridItem area="meta" element="aside">
 									<Hide until="desktop">
-										<div css={maxWidth}>
+										<div
+											css={[maxWidth, sidePaddingDesktop]}
```
oliverlloyd, line 693: What does this change do?

ioannakok (approved):
👍

dotcom-rendering/src/web/layouts/LiveLayout.tsx
oliverlloyd: Ideally, css relating to the ArticleMeta component would exist inside the component. We have these `maxWidth` divs as a sort of exception but like all exceptions they invite expansion!

oliverlloyd (commented):
Looks great, just one question about some css that's getting added

dotcom-rendering/src/web/layouts/LiveLayout.tsx
rcrphillips: Ah yes. That change accidentally crept in to the branch. Small one to edit the positioning on the meta
Screen Shot 2021-12-17 at 17 41 10
Screen Shot 2021-12-17 at 17 41 22

rcrphillips: @oliverlloyd 🆙

oliverlloyd: Hmm, it seem there's already quite a lot of padding and margin css in this layout file now. I guess this could be reviewed at a later stage but now isn't the time

oliverlloyd (approved):
⚽

PR #3773, opened 2021-12-17
Run Spacefinder on deadblogs

chrislomaxjones:
## What does this change?

Add support for running Spacefinder on DCR deadblogs. The code in the commercial bundle already runs on these pages, this PR adds the hints necessary to insert inline ad slots between liveblocks.

These hints take the form of two added classnames:
- `js-liveblog-body` applied to `ArticleBody` when rendering a liveblog/deadblog. This represents the container in which Spacefinder will attempt to insert inline ads.
- `block` applied to each of the `Liveblock` components. Spacefinder will select for these as the candidates before which inline ads can be inserted. Candidates are the direct descendants of the container above.

Additionally, the adverts that are inserted into liveblogs by Spacefinder require some additional styling. These styles are added to the `ArticleContainer` along with other dynamic ad styles, and scoped to target the classes of slots added by Spacefinder.

### Before

Screenshot 2022-01-07 at 09 26 38

### After

Screenshot 2022-01-07 at 09 26 11

## Why?

As migration to DCR continues it's necessary to also migrate the ability to include inline ads on these new pages.

### DIFF
--- a/common-rendering/src/components/liveBlockContainer.tsx
+++ b/common-rendering/src/components/liveBlockContainer.tsx
@@ -52,6 +52,7 @@ const LiveBlockContainer = ({
 		<article
 			id={`block-${id}`}
 			key={id}
+			// This classname is used by Spacefinder as a possible candidate before which it can insert an inline ad
 			className="block"
 			css={css`
 				padding: ${space[2]}px ${SIDE_MARGIN_MOBILE}px;
--- a/common-rendering/src/components/liveBlockContainer.tsx
+++ b/common-rendering/src/components/liveBlockContainer.tsx
@@ -1,6 +1,6 @@
 import { css } from "@emotion/react";
-import { neutral, from, space, headline } from '@guardian/source-foundations';
-import { FirstPublished } from './FirstPublished';
+import { neutral, from, space, headline } from "@guardian/source-foundations";
+import { FirstPublished } from "./FirstPublished";
 
 const LEFT_MARGIN_DESKTOP = 60;
 const SIDE_MARGIN = space[5];
@@ -24,7 +24,7 @@ const BlockTitle = ({ title }: { title: string }) => {
 	return (
 		<h2
 			css={css`
-				${headline.xxsmall({ fontWeight: 'bold' })}
+				${headline.xxsmall({ fontWeight: "bold" })}
 				margin-bottom: ${space[2]}px;
 			`}
 		>
@@ -52,6 +52,7 @@ const LiveBlockContainer = ({
 		<article
 			id={`block-${id}`}
 			key={id}
+			className="block"
 			css={css`
 				padding: ${space[2]}px ${SIDE_MARGIN_MOBILE}px;
 				margin-bottom: ${space[3]}px;
--- a/dotcom-rendering/docs/contracts/001-commercial-selectors.md
+++ b/dotcom-rendering/docs/contracts/001-commercial-selectors.md
@@ -2,7 +2,10 @@
 
 ## What is the contract?
 
-We place the `article-body-commercial-selector` class on the container element of the article body. (TODO: are there others?)
+We place the following classes on the container element of the article body:
+
+-   `article-body-commercial-selector` on the ArticleRenderer
+-   `js-liveblog-body` on the ArticleBody when rendering a Liveblog/Deadblog
 
 ## Where is it relied upon?
 
--- a/dotcom-rendering/src/web/components/ArticleContainer.tsx
+++ b/dotcom-rendering/src/web/components/ArticleContainer.tsx
@@ -1,6 +1,6 @@
 import { css } from '@emotion/react';
 import { ArticleDesign } from '@guardian/libs';
-import { from, until } from '@guardian/source-foundations';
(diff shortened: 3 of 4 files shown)


common-rendering/src/components/liveBlockContainer.tsx
```diff
@@ -1,6 +1,6 @@
 import { css } from "@emotion/react";
-import { neutral, from, space, headline } from '@guardian/source-foundations';
-import { FirstPublished } from './FirstPublished';
+import { neutral, from, space, headline } from "@guardian/source-foundations";
+import { FirstPublished } from "./FirstPublished";
```
OllysCoding, line 3: I think we lack any code-style checks in common-rendering which is why we're seeing changes like this (which isn't ideal) - perhaps something we need to look at outside this team

oliverlloyd (commented):
This change looks good in general. Is there any help you need progressing this to Ready for Review?

The only risk I can think of here is that the scope of the css might bleed but it looks to me like these changes are limited to specific classnames

common-rendering/src/components/liveBlockContainer.tsx
chrislomaxjones: I see, I think this is auto-formatting that I hadn't noticed then. If it's simpler I can remove the formatting changes from this PR and possibly address them separately (run prettier in common-rendering in a separate PR maybe?)

chrislomaxjones:
> This change looks good in general. Is there any help you need progressing this to Ready for Review?

I'm currently still nailing down the styles for outstream ads, hopefully that won't be too much of a problem though!

> The only risk I can think of here is that the scope of the css might bleed but it looks to me like these changes are limited to specific classnames

Once I've got the styling figured out I think the selectors should all just be classnames, with the elements matching those classnames being inserted by Spacefinder.

common-rendering/src/components/liveBlockContainer.tsx
oliverlloyd: I think it's fine to keep them here. They're small and self contained. We're working on addressing the linting issues for common in a separate thread

chrislomaxjones: Alright, sounds good!

dotcom-rendering/docs/contracts/001-commercial-selectors.md
```diff
@@ -2,7 +2,10 @@
 
 ## What is the contract?
 
-We place the `article-body-commercial-selector` class on the container element of the article body. (TODO: are there others?)
+We place the following classes on the container element of the article body:
+
+-   `article-body-commercial-selector` on the ArticleRenderer
+-   `js-liveblog-body` on the ArticleBody when rendering a Liveblog/Deadblog
```
oliverlloyd, line 8: Thank you for including this change! 🙏

dotcom-rendering/src/web/components/ArticleContainer.tsx
```diff
@@ -109,6 +109,42 @@ const articleAdStyles = css`
...
+		margin: 0 auto ${space[3]}px;
+
+		.ad-slot__label {
+			color: ${neutral[46]};
+			border-top-color: ${neutral[86]};
+		}
+
+		&.ad-slot--outstream {
+			${from.tablet} {
+				width: 620px;
```
oliverlloyd, line 122: I'm not certain what this width is doing but liveblogs have a width of `700px`, not `620px` - does this need to be accounted for here as well?

See: https://github.com/guardian/dotcom-rendering/blob/oliver/instagram-island/dotcom-rendering/src/web/layouts/LiveLayout.tsx#L175

common-rendering/src/components/liveBlockContainer.tsx
```diff
@@ -52,6 +52,7 @@ const LiveBlockContainer = ({
 		<article
 			id={`block-${id}`}
 			key={id}
+			className="block"
```
oliverlloyd, line 55: The comment included in `ArticleBody` was great, can we do something similar here as well?

oliverlloyd (approved):
This looks really great.

I've made a comment about a reference to 620 pixels I saw in the code but liveblogs are 700 pixels wide which might need checking and a teeny ask for a comment but I'm happy to approve this work and leave it up to you to work out the best approach for these comments

dotcom-rendering/src/web/components/ArticleContainer.tsx
chrislomaxjones: This width is to account for [outstream ads](https://github.com/guardian/commercial-core/blob/main/src/ad-sizes.ts#L70) which have a width of `620px`.

oliverlloyd: > This width is to account for [outstream ads](https://github.com/guardian/commercial-core/blob/main/src/ad-sizes.ts#L70) which have a width of `620px`.

Ah, I see. Thanks for checking!

common-rendering/src/components/liveBlockContainer.tsx
chrislomaxjones: Sure, I've added a comment!

PR #3774, opened 2021-12-20
Upgrade to SBT version 1.6.0-RC2

umguardian:
## What does this change?
Upgrading version of SBT.

## Why?
Addressed identified [security vulnerability](https://logging.apache.org/log4j/2.x/changes-report.html#a2.17.0) in log4j, which is used internally in SBT

### Before
`sbt.version = 1.5.7`

### After
`sbt.version = 1.6.0-RC2`


groakland (approved):
LGTM

PR #3784, opened 2021-12-22
🏖 `Island`

oliverlloyd:
## What does this change?
This PR adds the ability to insert an island of interactivity into the client without having to first server side render it. Previously  these client side only islands were referred to as 'Portal's but the concept shares so much with 'Hydrate' that it offers a better developer experience to bring them together.

This means [this Portal PR](https://github.com/guardian/dotcom-rendering/pull/3767) can now be closed.

## Why?
Some content is only ever rendered on the client, never on the server. Examples include most read, onwards, etc. In these cases we can't server side render and hydrate these components but we still want to benefit from the deferred dynamic import that the old `Hydrate` component brings.

## Why rename to `Island`?
`Hydrate` made sense when the purpose of the abstraction was to offer hydration, but now we're not just hydrating but rendering content directly instead so we need a name that encompasses both these use cases. `Island` is a term widely used in the industry to represent this pattern and benefits from not having any other Guardian related connotations that could be confusing. We also considered `Render`, `Interactive` and `DeferRender`.

Happy to hear feedback if others have thoughts on this name?

## Usage
The api for `Island` is conditional based on the props passed in

1) If `clientOnly` is set to `true`, then this triggers the 'client only portal' flow and `placeholderHeight` is required
2) If `clientOnly` is set to `false`, then this triggers the 'server render + hydrate' flow and `placeholderHeight cannot be passed

```ts

// Props when using server side rendering
interface HydrateProps {
	deferUntil?: When;
	clientOnly?: false;
	placeholderHeight?: undefined;
	children: JSX.Element;
}

// Props when server side rendering is disabled
interface PortalProps {
	deferUntil?: When;
	clientOnly: true;
	placeholderHeight: number;
	children: JSX.Element;
}
```

```jsx
  // Client only
  <Island deferUntil="visible" clientOnly={true} placeholderHeight={440}>
    <Onwards />
  </Island>

  // With server side rendering
  <Island deferUntil="idle">
    <SubNav />
  </Island>
```


dotcom-rendering/index.d.ts
```diff
@@ -690,8 +690,8 @@ interface CommercialConfigType {
 }
 
 type ServerSideTests = {
-    [k: `${string}Variant`]: "variant";
-    [k: `${string}Control`]: "control";
+	[k: `${string}Variant`]: 'variant';
+	[k: `${string}Control`]: 'control';
```
oliverlloyd, line 694: Prettier 🤷

OllysCoding (approved):
+1 Looks good to me :) Would be nice to have an E2E test for subnav!

SiAdcock:
I'm guessing this:
```jsx
  // Client only
  <Island defer="visible" ssr={true} placeholderHeight={440}>
    <Onwards />
  </Island>
```
Ought to be this:
```jsx
  // Client only
  <Island defer="visible" ssr={false} placeholderHeight={440}>
    <Onwards />
  </Island>
```
?

I find the `ssr` thing a bit hard to grok. The mental model is that you have to remember to turn off SSR if you don't need it.  I dunno why but props that default to `true` always feel a bit confusing to me. 

What do you think about the following:
```jsx
  // Client only
  <Island defer="visible" clientOnly={true} placeholderHeight={440}>
    <Onwards />
  </Island>
```

dotcom-rendering/index.d.ts
```diff
@@ -1065,9 +1065,9 @@ interface PerformanceEntry {
 
 declare namespace JSX {
 	interface IntrinsicElements {
-		'gu-hydrate': {
+		'gu-island': {
 			name: string;
-			when?: 'immediate' | 'idle' | 'visible';
+			defer?: 'idle' | 'visible';
```
SiAdcock, line 1070: How about `deferUntil`? It seems to read better.

dotcom-rendering/src/web/components/Island.tsx
```diff
@@ -0,0 +1,53 @@
...
+type Defer = 'idle' | 'visible';
+
+interface HydrateProps {
+	defer?: Defer;
+	ssr?: true;
+	placeholderHeight?: undefined;
+	children: JSX.Element;
+}
+
+interface PortalProps {
```
SiAdcock, line 14: If we're moving away from the `Portal` nomenclature, should we try something like `ClientOnlyProps`?

dotcom-rendering/src/web/components/Island.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+/**
+ * Props
+ *
+ * We use a union type here to support conditional typing. This means if the ssr
+ * flag is set to false, then you must supply placeholderHeight. If ssr is true or
+ * not supplied, then placeholderHeight must not be given
+ */
+type Props = HydrateProps | PortalProps;
```
SiAdcock, line 28: Clever!

dotcom-rendering/index.d.ts
oliverlloyd: Done

dotcom-rendering/src/web/components/Island.tsx
oliverlloyd: Agreed, done

dotcom-rendering/src/web/components/Island.tsx
```diff
@@ -0,0 +1,67 @@
...
+	deferUntil?: When;
+	clientOnly?: false;
+	placeholderHeight?: undefined;
+	children: JSX.Element;
+}
+
+interface ClientOnlyProps {
+	deferUntil?: When;
+	clientOnly: true;
+	placeholderHeight: number;
```
tjmw, line 17: In the case of something like a fixed position banner (e.g. `StickyBottomBanner`) would the right thing to do be pass 0 for the placeholder height? Or are there perhaps cases (like this) where a placeholder height doesn't make sense? (Does this maybe imply it should be optional here too?)

oliverlloyd: 🤔 that's a good point. You should be able to specify `clientOnly` as true but just not have a placeholder. Needing to set it as zero is cumbersome.

tjmw:
Excited to see this progressing, nice work! 🎉

dotcom-rendering/src/web/components/Island.tsx
oliverlloyd: I've marked this prop as optional 👍

jamesgorrie:
Reference https://github.com/guardian/dotcom-rendering/pull/3629

PR #3786, opened 2021-12-23
Filter key events

joecowton1:
## What does this change?

Adds key event filtering toggle and timeline link toggle state persistence to DCR.

## Why?

We need to bring DCR into line with Frontend which already has this functionality.

## What does this change?

- adds `filterKeyEvents: boolean` to `CAPI` type
- updates fixtures files 
- adds `FilterKeyEventsToggle` component which updates query params on click
- passes `filterKeyEvents` to `keyEventsContainer` to allow links to persist filter state
- adds the `filterKeyEventsToggle` to `LiveLayout`
- fixes small bug in AR `metadata` component created by removing margin from the `toggleSwitch` in Source.

[This PR](https://github.com/guardian/frontend/pull/24607) in Frontend adds `filterKeyEvents` to the `DotcomRenderingModel`. This is then passed to DCR as `CAPI` data. 

### Before and After

Screenshot 2022-02-04 at 14 44 23

Screenshot 2022-02-04 at 14 44 16


dotcom-rendering/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "editor.formatOnSave": true
```
joecowton1, line 2: This needs removing.

dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
```diff
@@ -0,0 +1,44 @@
...
+	const [checked, setChecked] = useState(false);
+
+	const handleClick = () => {
+		setChecked(!checked);
+
+		const urlParams = new URLSearchParams(window.location.search);
+		urlParams.set('filterKeyEvents', checked ? 'false' : 'true');
+
+		window.location.search = urlParams.toString();
+		window.location.hash = 'maincontent';
```
alinaboghiu, line 29: may be that we're already doing this but we should make sure to keep the toggle in view at the top of the screen when navigating to the main content after a click - would be weird if the button disappears after you click it

dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
```diff
@@ -0,0 +1,44 @@
...
+	};
+
+	useEffect(() => {
+		setChecked(window.location.search.includes('filterKeyEvents=true'));
+	}, [setChecked]);
+
+	return (
+		<ToggleSwitch
```
alinaboghiu, line 37: putting these here so they don't get forgotten about
- [ ] label should be clickable 
- [ ] toggle should be tabbable and keyboard accessible 
- [ ] the animation is a bit off (green tick appears a bit too soon) and the toggle is not vertically aligned 
- [ ] in mobile view toggle should appear below the live feed header/separator rather than above it

dotcom-rendering/.vscode/settings.json
joecowton1: Removed

oliverlloyd:
Just a drive by comment. Can you remove 'WIP' and instead convert this PR to a draft? There are two benefits to using draft PRs:

1) Reviewers are not emails requesting a review if you open a draft PR
2) It's easier to see/filter which PRs are pending review

Screenshot 2022-02-02 at 14 53 12

Once you are ready to ask for reviews you can click 'Ready for Review'

Thanks!

joecowton1:
> Just a drive by comment. Can you remove 'WIP' and instead convert this PR to a draft? 

Done!

dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
```diff
@@ -0,0 +1,49 @@
...
+		window.location.search = urlParams.toString();
+		window.location.hash = 'maincontent';
+	};
+
+	useEffect(() => {
+		setChecked(window.location.search.includes('filterKeyEvents=true'));
+	}, [setChecked]);
+	// return null;
+	return (
+		<div id="maincontent">
```
oliverlloyd, line 40: Can we keep this around article body for semantics

dotcom-rendering/src/web/components/KeyEventsContainer.importable.tsx
```diff
@@ -1,21 +1,30 @@
...
 
 type Props = {
 	keyEvents: Block[];
 	format: ArticleFormat;
 };
 
 export const KeyEventsContainer = ({ keyEvents, format }: Props) => {
+	const [filterKeyEvents, setFilterKeyEvents] = useState(false);
+
+	useEffect(() => {
```
joecowton1, line 14: Do we need this?

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -938,13 +958,14 @@ export const LiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
...
 											keyEventsTopMargin,
 											sidePaddingDesktop,
-											accordionBottomMargin,
 										]}
 									>
-										<KeyEventsContainer
-											format={format}
-											keyEvents={CAPI.keyEvents}
-										/>
+										<Island deferUntil="idle">
```
joecowton1, line 963: If we're not using the useEffect within this component the Island can be removed.

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -35,7 +34,7 @@ import { GridItem } from '@root/src/web/components/GridItem';
 import { AgeWarning } from '@root/src/web/components/AgeWarning';
 import { DiscussionContainer } from '@root/src/web/components/DiscussionContainer.importable';
 import { Pagination } from '@frontend/web/components/Pagination';
-import { KeyEventsContainer } from '@frontend/web/components/KeyEventsContainer';
+import { KeyEventsContainer } from '@root/src/web/components/KeyEventsContainer.importable';
```
joecowton1, line 37: This can be renamed if we're not hyrdating.

dotcom-rendering/src/web/components/KeyEventsContainer.test.tsx
```diff
@@ -1,7 +1,7 @@
 import { render } from '@testing-library/react';
 import { ArticleDisplay, ArticleDesign, ArticlePillar } from '@guardian/libs';
 
-import { KeyEventsContainer } from './KeyEventsContainer';
+import { KeyEventsContainer } from './KeyEventsContainer.importable';
```
joecowton1, line 4: This can be renamed if we're not hydrating.

dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
joecowton1: Reverted.

dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
```diff
@@ -0,0 +1,54 @@
...
+	const [checked, setChecked] = useState(false);
+
+	const handleClick = () => {
+		setChecked(!checked);
+
+		const urlParams = new URLSearchParams(window.location.search);
+		urlParams.set('filterKeyEvents', checked ? 'false' : 'true');
+
+		window.location.search = urlParams.toString();
+		window.location.hash = 'filter-toggle';
```
joecowton1, line 36: This is not persisting in the URL currently - why not? Is something else resetting it?

joecowton1: Resolved by swapping the order.

dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
joecowton1: Think these are all resolved now

dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
joecowton1: Resolved

apps-rendering/src/components/metadata.tsx
```diff
@@ -97,6 +97,11 @@ const liveBlogPadding = css`
 const toggleOverrideStyles = css`
 	padding-right: ${remSpace[1]};
 
+	button {
```
joecowton1, line 100: This unrelated change fixes a styling regression in the AR `metadata` component which was introduced by removing the margin in the Source `ToggleSwitch` component here: https://github.com/guardian/source/pull/1255

dotcom-rendering/src/web/components/KeyEventsContainer.tsx
```diff
@@ -5,17 +5,24 @@ import KeyEvents, {
...
 	const transformedKeyEvents: KeyEvent[] = keyEvents
 		.filter((keyEvent) => {
 			return keyEvent.title && keyEvent.blockFirstPublished;
 		})
 		.map((keyEvent) => {
 			return {
 				text: keyEvent.title || '', // We fallback to '' here purely to keep ts happy
-				url: `?page=with:block-${keyEvent.id}#block-${keyEvent.id}`,
+				url: `?filterKeyEvents=${
+					filterKeyEvents ? 'true' : 'false'
```
abeddow91, line 24: as the original boolean is being interpolated, can we just use `filterKeyEvents` here rather than conditionally returning 'true' or 'false'?

dotcom-rendering/index.d.ts
```diff
@@ -498,6 +498,7 @@ interface CAPIType {
 	// The logic is date-driven. See:
 	// https://github.com/guardian/frontend/blob/main/common/app/model/dotcomrendering/InteractiveSwitchOver.scala#L7.
 	isLegacyInteractive?: boolean;
+	filterKeyEvents: boolean;
```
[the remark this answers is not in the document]
joecowton1: Think you're right, good spot! Have updated.

dotcom-rendering/src/web/components/KeyEventsContainer.tsx
joecowton1: Yeah makes it a bit easier to read too. Updated.

dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
```diff
@@ -0,0 +1,43 @@
...
+	padding-bottom: ${remSpace[3]};
+
+	${from.desktop} {
+		padding: ${remSpace[3]} 0;
+		border-top: 1px solid ${neutral[86]};
+	}
+`;
+
+export const FilterKeyEventsToggle = () => {
+	const [checked, setChecked] = useState(false);
```
oliverlloyd, line 16: Is there a way to get this value from the CAPI object?

oliverlloyd: I know this is outside this PR but I would expect `CAPI.filterKeyEvents` to be `true` here, no?
Screenshot 2022-02-07 at 17 12 27

[Example url](https://www.theguardian.com/world/live/2022/feb/07/ukraine-russia-crisis-macron-heads-for-talks-with-putin-while-scholz-and-biden-meet-in-dc-live-coverage.json?dcr&filterKeyEvents=true)

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -969,6 +994,13 @@ export const LiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
 											accordionTitle="Live feed"
 											context="liveFeed"
 										>
+											<GridItem area="filter">
+												<Hide above="desktop">
+													<Island deferUntil="idle">
```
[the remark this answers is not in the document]
joecowton1: Sounds very reasonable to me.

dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
joecowton1: That's works nicely on CODE, thanks. I'm not sure why behaviour is sometimes different when running locally but I imagine it's something to do with query params being stripped out somewhere along the way.

Regarding the other point above, I've [made a PR ](https://github.com/guardian/frontend/pull/24615/files)to add the `filterKeyEvents` parameter to the DCR path in `renderJson` so if we decided it's needed that's ready to go.

oliverlloyd: To be honest I don't have a clear understand on the renderArticle vs. renderJson functions but I can say that the 'CAPI' object here (in this context) is the blob that Frontend sends to DCR, and this is what you are passing into `KeyEventsContainer` (via `CAPI.filterKeyEvents`) and now also into `KeyEventsToggle`.

So it sounds like this data is already being sent by Frontend to DCR and perhaps the extra PR isn't needed? But the names of all these function confuses me so I could be misunderstanding.

oliverlloyd (approved):
Great stuff

oliverlloyd:
Can I just check, we're happy that the value of `filterKeyEvents` on this json object is not following the query param in the url?

https://www.theguardian.com/world/live/2022/feb/07/ukraine-russia-crisis-macron-heads-for-talks-with-putin-while-scholz-and-biden-meet-in-dc-live-coverage.json?dcr&filterKeyEvents=true

If so, what does this property represent? I thought that [this PR](https://github.com/guardian/frontend/pull/24607) introduced it as a reflection of the url but I didn't follow that too closely.

joecowton1:
> Can I just check, we're happy that the value of `filterKeyEvents` on this json object is not following the query param in the url?
> 
> https://www.theguardian.com/world/live/2022/feb/07/ukraine-russia-crisis-macron-heads-for-talks-with-putin-while-scholz-and-biden-meet-in-dc-live-coverage.json?dcr&filterKeyEvents=true
> 
> If so, what does this property represent? I thought that [this PR](https://github.com/guardian/frontend/pull/24607) introduced it as a reflection of the url but I didn't follow that too closely.

The above url (containing a `.json` path) hits the following article [route](https://github.com/guardian/frontend/blob/ccd037b53012836b3dae5b42b9698a02ede8e01e/article/conf/routes#L17) which in turn calls [LiveBlogController.renderJson](https://github.com/guardian/frontend/blob/7f9c7eaf8d4374c85a7638cb809f262fd9064774/article/app/controllers/LiveBlogController.scala#L73) passing in the `filterKeyEvents` parameter.  

This route is called to render live blog updates, _not_ when live blogs first render,  and at the moment the `filterKeyEvents` parameter is passed through to determine that the correct updates are loaded in [renderNewerUpdatesJson](https://github.com/guardian/frontend/blob/7f9c7eaf8d4374c85a7638cb809f262fd9064774/article/app/controllers/LiveBlogController.scala#L205) which provides the new live blog blocks.

The reason it appears as `false` currently is that we're hardcoding that value in [DotcomRenderingDataModel.forLiveblog](https://github.com/guardian/frontend/blob/7f9c7eaf8d4374c85a7638cb809f262fd9064774/article/app/controllers/LiveBlogController.scala#L240). `DotcomRenderingDataModel.forLiveblog` is called by the [DotcomRenderingService](https://github.com/guardian/frontend/blob/7f9c7eaf8d4374c85a7638cb809f262fd9064774/common/app/renderers/DotcomRenderingService.scala#L32) and creates the data object that we call `CAPI` in DCR. 

This is what the [PR](https://github.com/guardian/frontend/pull/24615) I created earlier amends.

The distinction between `renderJson` and `renderArticle` is that where `renderJson` is called for live blog updates, `renderArticle` is called when the page first loads hitting this route [here](https://github.com/guardian/frontend/blob/ccd037b53012836b3dae5b42b9698a02ede8e01e/article/conf/routes#L22). This takes us down a different rendering flow where we _do_ use the `filterKeyEvents` parameter: renderArticle -> [renderWithRange](https://github.com/guardian/frontend/blob/7f9c7eaf8d4374c85a7638cb809f262fd9064774/article/app/controllers/LiveBlogController.scala#L97) -> [DotcomRenderingService.getArticle](https://github.com/guardian/frontend/blob/7f9c7eaf8d4374c85a7638cb809f262fd9064774/common/app/renderers/DotcomRenderingService.scala#L122).

dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
joecowton1: I've updated the Frontend PR to explain a bit more context, we can resolve discussions on there hopefully 🙂

oliverlloyd:
> The `.json` route in the above url is used for live blog block updates only

I don't think this is a problem but just for the purposes of promoting understanding of DCR: this statement isn't _semantically_ true. DCR uses the `.json` endpoint to fetch the data for an article when running locally and it is a long held contract that when you add `json` to an article you are given the same json object that Frontend posts to DCR. I actually always assumed that the route to generate this json was running same code as Frontend uses when posting it to DCR, if this assumption is false and these paths have diverged then this should be fixed to restore the contract

joecowton1:
> > The `.json` route in the above url is used for live blog block updates only
> 
> I don't think this is a problem but just for the purposes of promoting understanding of DCR: this statement isn't _semantically_ true. DCR uses the `.json` endpoint to fetch the data for an article when running locally and it is a long held contract that when you add `json` to an article you are given the same json object that Frontend posts to DCR. I actually always assumed that the route to generate this json was running same code as Frontend uses when posting it to DCR, if this assumption is false and these paths have diverged then this should be fixed to restore the contract

This is very useful to know, and very probably explains why I was seeing different behaviour locally. Our (my and Alina's) incorrect assumption had been that DCR always hit the _renderArticle_ route. 

I think with this is mind we do need to include the correct `filterKeyEvents` value in the `.json` endpoint blob.

joecowton1:
The above suggestion is implemented here: https://github.com/guardian/frontend/pull/24615

alinaboghiu:
This .json endpoint is a bit confusing for liveblogs because it's used for multiple purposes I think. Frontend currently uses the .json path for liveblogs to either:
- return a json version of the liveblog article (sensibly)
- return a json payload containing the newest blocks when provided with a "lastUpdated" parameter (weirdly)
- return a json version of the liveblog for use on fronts when provided with a "rendered=false" parameter
- return a json version of the minute page (whatever that is - I'm not sure)

DCR currently used the .json path for _I think_:
- requesting a json version of the liveblog article for local/testing purposes 

And soon DCR will also use this path to:
- return a json payload containing the newest blocks when provided with a "lastUpdated" parameter (in line with how Frontend uses this endpoint atm)

I hope this summary makes sense, if it does not we should talk! 

As to whether we should be hardcoding filterKeyEvents to false in the .json response to DCR - I think the answer is no, if it can be avoided (which is what you eventually did, hurray), for consistency with the production behaviour. This is definitely my bad for not thinking this through when I suggested to Joe that we hardcode it to false, so apologies for that!

oliverlloyd:
> * minute page (whatever that is - I'm not sure)

The term 'minute' has deep roots going back to the very first days of liveblogs when they were referred to as 'minute by minute'. See: https://www.theguardian.com/tone/minutebyminute

oliverlloyd:
> I hope this summary makes sense

It does. It's actually great! This really does highlight the way this `.json` api has been overloaded and could be reviewed. I think there could be value in setting up a cleaner api in terms of making it easier to extend in future (which I presume is something that could be desired if we're extending liveness and adding new features)

PR #3788, opened 2021-12-23
Display football match stats on liveblog pages

rcrphillips:
## What does this change?
Adds the match stats component to the live layout.
## Why?
So that we can enable football deadblogs

### above desktop(sitting in left column)
Screen Shot 2022-01-26 at 15 56 06
### mobileLandscape to desktop
Screen Shot 2022-01-26 at 15 13 28
### mobile
Screen Shot 2022-01-26 at 15 12 59


rcrphillips:
[Football - Match stats component](https://trello.com/c/wsFrPSCW/506-football-match-stats-component)

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -748,6 +749,20 @@ export const LiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
 										/>
 									</div>
 								</GridItem>
+								<GridItem area="matchstats">
+									{CAPI.matchUrl && (
```
DavidLawes, line 753: I really like how we have variants of the grid type, it helped me understand how the match stats are separated from the standard grid.

Could matchUrl ever be an empty string? In StandardLayout I think we account for this by `!!CAPI.matchUrl` - it might be something we consider here

dotcom-rendering/src/web/components/Lineup.tsx
```diff
@@ -76,31 +76,19 @@ export const Lineup = ({ players }: Props) => {
...
-										`}
-									>
-										<Event
-											type={event.eventType}
-											time={event.eventTime}
-										/>
-									</div>
-								))}
-							</Row>
+						<div>
```
DavidLawes, line 84: This feels simpler, was there a change in design that allowed us to achieve this or is the result equivalent?

dotcom-rendering/src/web/components/MatchStats.tsx
```diff
@@ -173,19 +256,95 @@ const H4 = ({ children }: { children: React.ReactNode }) => (
...
+			value: home.possession,
+			label: home.codename,
+			color: home.colours,
+		},
+		{
+			value: away.possession,
+			label: away.codename,
+			color: away.colours,
+		},
+	].reverse();
```
DavidLawes, line 279:
```suggestion
	const sections = [
		{
			value: away.possession,
			label: away.codename,
			color: away.colours,
		},
		{
			value: home.possession,
			label: home.codename,
			color: home.colours,
		},
	];
```

Could we change the default ordering so we don't have to `reverse()`? Could be going mad so feel free to ignore!

dotcom-rendering/src/web/layouts/LiveLayout.tsx
DavidLawes, line 753: Out of curiosity, in StandardLayout we check both `matchUrl` and the design format (`ArticleDesign.MatchReport`) before rendering any match stats related stuff. For live/deadblogs should we display match stats for any blog that specifies a matchUrl, regardless of the article format?

DavidLawes (approved):
Nice!

I noticed there weren't any chromatic diffs and I wonder if it might be good to include some stories for match stats on a live/deadblog layout? I'm not sure where this naturally fits in our current setup of stories though - e.g. standard layout has its own set of stories, but liveblog stories are included as part the stories for StandardLayout, ImmersiveLayout, etc.

Maybe I've misunderstood the current setup of stories (in which case ignore), or maybe this is a bigger question about live/deadblog stories that might be a separate change, what do you think?

dotcom-rendering/src/web/components/Lineup.tsx
rcrphillips: We needed to ensure it fit better within the context, and the old way was causing overflow issues.

dotcom-rendering/src/web/components/MatchStats.tsx
rcrphillips: I don't fully understand why we have to do that in the first place.....

dotcom-rendering/src/web/layouts/LiveLayout.tsx
rcrphillips: As liveblogs currently only have one layout expected. i.e. there's no showcase/immersive main media or other elements, we don't need to worry about the article format. That's my understanding at least.

rcrphillips:
> 

I think that the wider lack of storybook entries for live layouts is a separate piece to this. But I do agree it would make sense to ensure we get some in with the different breakpoints and contexts.

dotcom-rendering/src/web/layouts/StandardLayout.tsx
```diff
@@ -626,17 +626,19 @@ export const StandardLayout = ({ CAPI, NAV, format, palette }: Props) => {
...
-										placeholderHeight={800}
-										deferUntil="visible"
-									>
-										<GetMatchStats
-											matchUrl={CAPI.matchUrl}
-										/>
-									</Island>
-								)}
+								{format.design === ArticleDesign.MatchReport &&
+									!!CAPI.matchUrl && (
```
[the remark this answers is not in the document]
rcrphillips: I've updated the others to use `isMatchReport`, but for some reason I can't quite discern, it throws an error around the matchUrl prop on `<GetMatchSats>` when I do.

PR #3793, opened 2022-01-05
Update riff-raff aws-s3 publicReadAcl

ioannakok:
## What does this change?
Explicitly sets `publicReadAcl: true` in `riff-raff.yaml` for `aws-s3`.

## Why?
Previously the parameter was optional. After this change: https://github.com/guardian/riff-raff/pull/665, it is required. The default value is `true`.

### DIFF
--- a/apps-rendering/riff-raff.yaml
+++ b/apps-rendering/riff-raff.yaml
@@ -33,3 +33,4 @@ deployments:
       bucket: mobile-apps-rendering-assets
       cacheControl: public, max-age=315360000
       prefixStack: false
+      publicReadAcl: true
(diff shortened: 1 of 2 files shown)


DavidLawes (approved):
In both of these cloudformations (judging by the name `*-assets`) I'm assuming we deploy stuff that then needs to be publicly accessible, meaning the default public read acl of `true` is needed

akash1810 (approved):

ioannakok:
[Update riff-raff aws-s3 publicReadAcl](https://trello.com/c/QTb5KepQ/521-update-riff-raff-aws-s3-publicreadacl)

PR #3794, opened 2022-01-05
Better Follow icon for apps-rendering

mxdvl:
## What does this change?

Unify the logic to generate the follow / following icons in Apps-rendering. Use knockout icons in both cases, instead of stacked layers of symbols. This means you can see _through_ the icons in both cases.

## Why?

The icons also share 75% of the shapes (head, torso, circle), so the code now reflects this fact. Only the **plus** or **check** signs changes.

### Screenshot

light mode

dark mode


MarSavar (approved):
👍 Went over this together with @JamieB-gu the other day, LGTM

PR #3795, opened 2022-01-05
Remove the need for global state when expanding comments

oliverlloyd:
## What does this change?
This PR adds `@guardian/discussion-rendering@9.1.0` which will expand comments if it sees `#comments` in the url. This removes the need to manually call `setIsExpanded` and allows the `CommentCount` component to work in isolation.

## Why?
By making `CommentCount` independent we can break it out into it's own [Island](https://github.com/guardian/dotcom-rendering/pull/3784), removing the need for global state and furthering the wider goal of [deleting App.tsx](https://docs.google.com/document/d/1kSrZZrJ8hSw6bek5hGa9onE8QDz5TRHFqHV52EOFjts/edit#heading=h.gsod33a0pc47)

## What next?
To achieve the stated aim of removing the count of comments as a piece of global state, we will also need to drop down the api call to get details for a discussion. This will result in the same api call being made twice but by using [SWR](https://swr.vercel.app/) we expect to prevent any increase in origin traffic

### DIFF
--- a/dotcom-rendering/package.json
+++ b/dotcom-rendering/package.json
@@ -49,7 +49,7 @@
     "@guardian/braze-components": "^5.1.0",
     "@guardian/commercial-core": "^0.32.0",
     "@guardian/consent-management-platform": "~6.11.5",
-    "@guardian/discussion-rendering": "^9.0.0",
+    "@guardian/discussion-rendering": "^9.1.0",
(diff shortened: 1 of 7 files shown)


ioannakok (approved):
+1

OllysCoding (approved):

PR #3798, opened 2022-01-05
Isolate `CommentCount`

oliverlloyd:
## What does this change?
Here we refactor `CommentCount` to work in isolation, removing the coupling that previously existed between it and `Discussion`. The actual dependency that caused the coupling was removed in a [previous PR](https://github.com/guardian/dotcom-rendering/pull/3795). This PR adds the final refactor that was needed which was to drop the api call to GET /discussion down into `CommentCount` itself (rather than passing it in)

## Why?
This coupling added complexity and prevented us from extracting this code out of `App.tsx` into an `Island`. By isolating this component it is easier to reason about and maintain

## But doesn't this mean we make the same api call twice?
Yes, it does! We now call the same endpoint twice, in two different places. The `Discussion` component was created in large part to prevent this. So what's going on? 

The reason we can get away with making multiple calls to the same endpoint on page load is because we're using [SWR](https://swr.vercel.app/). This library enhances how we fetch data in a number of ways, one of which is deduplication. If the same call is made in succession within a defined time period then this library will ensure that the response to the first call is served to the second, preventing multiple expensive network calls. This allows us the benefits of isolated, uncoupled code without the costs of duplicated fetch requests

The caveat to this is when the same request is being made over the two different hydration strategies listed here. In that case the de-deuplication provided by SWR does not work because the global SWR cache is not shared between these two processes. Because of this, this PR is actually _introducing_ duplication of this api call. This will be fixed in a follow up PR when `Discussion` is itself brought over to the Island pattern.


dotcom-rendering/src/web/components/ArticleMeta.tsx
```diff
@@ -28,6 +30,9 @@ type Props = {
 	primaryDateline: string;
 	secondaryDateline: string;
 	branding?: Branding;
+	discussionApiUrl: string;
+	shortUrlId: string;
+	isCommentable: boolean;
```
oliverlloyd, line 35: Here we're prop drilling these values down so we can serialise them in the Island

dotcom-rendering/src/web/components/ArticleMeta.tsx
```diff
@@ -407,10 +415,17 @@ export const ArticleMeta = ({
...
+								{isCommentable && (
+									<Island clientOnly={true}>
+										<CommentCount
+											discussionApiUrl={discussionApiUrl}
+											shortUrlId={shortUrlId}
+											format={format}
+										/>
+									</Island>
+								)}
+							</div>
```
oliverlloyd, line 428: This is the main change in this PR. I've removed the code in App.tsx (well, actually I removed it from `Discussion`) which was adding this component and it now lives in its own Island

dotcom-rendering/src/web/components/CommentCount.importable.tsx
```diff
@@ -1,14 +1,17 @@
...
+import { decidePalette } from '../lib/decidePalette';
+import { useDiscussion } from '../lib/useDiscussion';
 
 type Props = {
-	palette: Palette;
-	isCommentable: boolean;
-	commentCount?: number;
+	format: ArticleFormat;
+	discussionApiUrl: string;
+	shortUrlId: string;
```
oliverlloyd, line 14: Previously we passed in the values to display, now we pass in the things the component needs to make the api call to fetch the values itself

dotcom-rendering/src/web/components/Counts.stories.tsx
```diff
@@ -39,12 +42,26 @@ export const Both = () => {
...
+			{
+				status: 200,
+				body: {
+					discussion: {
+						commentCount: 239,
+						isClosedForComments: false,
+					},
+				},
+			},
+			{ overwriteRoutes: false },
```
oliverlloyd, line 64: Now that `CommentCount` makes its own api call, we need to mock this here

dotcom-rendering/src/web/components/Counts.stories.tsx
```diff
@@ -79,36 +96,152 @@ export const ShareOnly = () => {
...
+					/>
+				</div>
+				<div className="meta-number">
+					<CommentCount
+						discussionApiUrl="https://discussion.theguardian.com"
+						shortUrlId="p/3bdii8"
+						format={{
+							...format,
+							theme: ArticleSpecial.SpecialReport,
+						}}
```
oliverlloyd, line 237: I've deleted `CommentCount.stories` and replaced that coverage here in `Counts.stories`

dotcom-rendering/src/web/components/Discussion.tsx
```diff
@@ -141,16 +137,6 @@ export const Discussion = ({
...
-			{commentCount !== undefined && beingHydrated && (
-				<Portal rootId="comment-count-root">
-					<CommentCount
-						isCommentable={isCommentable}
-						commentCount={commentCount}
-						palette={palette}
-					/>
-				</Portal>
-			)}
-
```
oliverlloyd, line 153: This is where we remove the old `Portal` - it is replaced with an `Island` instead

dotcom-rendering/src/web/components/SignedInAs.tsx
```diff
@@ -83,7 +83,7 @@ const rowUntilDesktop = css`
 	}
 `;
 
-const CommentCount = ({ count }: { count?: number }) => {
+const Heading = ({ count }: { count?: number }) => {
```
oliverlloyd, line 86: This was just a drive by refactor because this unrelated component with the same name kept coming up in my searches

dotcom-rendering/src/web/lib/formatCount.test.ts
```diff
@@ -0,0 +1,16 @@
...
+	it('formats very long numbers', () => {
+		expect(formatCount(92878)).toEqual({ short: '93k', long: '92,878' });
+	});
+	it('returns zero for zero', () => {
+		expect(formatCount(0)).toEqual({ short: '0', long: '0' });
+	});
+	it('returns an elipsis for undefined', () => {
+		expect(formatCount()).toEqual({ short: '…', long: '…' });
+	});
+});
```
oliverlloyd, line 16: I deleted `ComentCount.tests` and replaced that coverage with a more focussed set of unit tests on the actual function. This is a better use of unit tests (rather than trying to write unit tests to verify rendering logic which we can much better achieve using Storybook + Chromatic

dotcom-rendering/src/web/lib/useDiscussion.ts
```diff
@@ -51,7 +51,11 @@ type CommentType = {
 };
 
 export const useDiscussion = (url: string) => {
-	const { data } = useApi<DiscussionResponse>(url);
+	const { data } = useApi<DiscussionResponse>(url, {
+		// The default for dedupingInterval is 2 seconds but we want to wait longer here because the cache time
+		// for a discussion is at least 15 seconds
+		dedupingInterval: 8000,
+	});
```
oliverlloyd, line 58: This is the magic. The reason this PR is possible is because the SWR library dedupes multiple api calls to the same endpoint. So, yes, now that we're breaking `CommentCount` out into its own Island and now that it is making its own api call to get the discussion, this means **we are making the same api call twice**. But this is fine because there is actually only one _network_ request made.

### Why 8 seconds?
The default is 2 but because we're already caching the discussion endpoint calls we may as well extend the time we dedupe them to reduce network round trips.

OllysCoding: Just to check, do we know that the SWR library dedupes/caches calls from different react renders/instances? I'm guessing this was checked in the dev tools? :)

.storybook/main.js
```diff
@@ -18,6 +18,17 @@ module.exports = {
...
+				// We're explicitly setting the minification options below because
+				// we want to turn off `minifyIdentifiers`. Why? Because it breaks
+				// Islands hydration. When you minify the component filenames
+				// the dynamic imports fail to find them.
+				esbuildMinifyOptions: {
+					target: "es2015",
+					minify: false,
+					minifyWhitespace: true,
+					minifyIdentifiers: false,
+					minifySyntax: true,
```
jamesgorrie, line 30: Why can't we just send over 
```TS
{ minifyIdentifiers: false }
```

as the comment above suggests?

dotcom-rendering/src/web/lib/useDiscussion.ts
oliverlloyd: This is a very good question. I spent some time checking this and found that:

1. It will _not_ dedupe between an `Island` and `App.tsx`
2. But yes, it will dedupe over multiple `Islands`

SWR uses an in memory cache which gets reinitialised inside `BootReact` but the separate island instances share one. I did look into if there was a way to create one shared global cache but it was not possible

dotcom-rendering/src/web/components/App.tsx
```diff
@@ -1136,7 +1136,6 @@ export const App = ({ CAPI, ophanRecord }: Props) => {
 					format={format}
 					discussionApiUrl={CAPI.config.discussionApiUrl}
```
jamesgorrie, line 1137: More for clarity, the reason we can't remove these from `App`, but have added them to `ArticleMeta`  is they are still depended on elsewhere in the code?

dotcom-rendering/src/web/components/ArticleMeta.stories.tsx
```diff
@@ -124,6 +127,9 @@ export const BrandingStory = () => {
 				tags={tagsWithBylineImage}
 				primaryDateline="Sun 12 Jan 2020 18.00 GMT"
 				secondaryDateline="Sun 12 Jan 2020 21.00 GMT"
+				isCommentable={false}
+				discussionApiUrl=""
+				shortUrlId=""
```
jamesgorrie, line 132: General style question - when do we use `empty string` vs `undefined` or other none-able values?

dotcom-rendering/src/web/components/Counts.stories.tsx
```diff
@@ -79,36 +96,152 @@ export const ShareOnly = () => {
...
 					/>
 				</div>
 			</Counts>
 		</Container>
 	);
 };
-ShareOnly.story = { name: 'with comments disabled' };
+Themes.story = { name: 'with different themes' };
```
jamesgorrie, line 244: Have we lost the `ShareOnly` part of these stories now?

dotcom-rendering/src/web/lib/formatCount.test.ts
jamesgorrie: Maybe a little bit of pedantry here - But does count also mean any number we display? Is this a general content style rule and if so, do we group those anywhere?

dotcom-rendering/src/web/lib/useDiscussion.ts
jamesgorrie: Potential bikeshed: Is there a test to check this?

jamesgorrie (commented):
The change itself seems to make a load of sense, and is clear!

Sorry to not approve, but I feel I'll need to have more confidence in my knowledge around the codebase first.

dotcom-rendering/src/web/components/App.tsx
oliverlloyd: Yes, exactly. They're still used by `Discussion`

dotcom-rendering/src/web/components/ArticleMeta.stories.tsx
oliverlloyd: I don't have strong opinions on this personally but I think typescript might. The type of `discussionApiUrl` is `string` and it is required so by passing `""` here we're kind of keeping TS happy. You could argue this is cheating but the api call is mocked in this case so maybe it's better not passing it so it's very clear it isn't being used?

dotcom-rendering/src/web/components/Counts.stories.tsx
oliverlloyd: I took that story out because it was moot because the `CommentCount` still shows when it is zero but after thinking about it I realised we don't have a story for when the article does not have commenting enabled so I've included that now. This is effectively the same as what `ShareOnly` was

.storybook/main.js
oliverlloyd: The impression I get from the docs on this is that once you turn off minify you need to be explicit about these sub options

https://github.com/privatenumber/esbuild-loader#minify

I added some links to the comment to make this clearer

dotcom-rendering/src/web/lib/formatCount.test.ts
oliverlloyd: You could defo argue that `formatCount` could be a more generic function available to other parts of the code. Maybe we can think about hoisting it [into @guardian/libs](https://github.com/guardian/libs)?

dotcom-rendering/src/web/lib/useDiscussion.ts
oliverlloyd: I do want to test this but the only way we can really achieve that is with a Cypress test and for that I need a better use case. I've got a future PR coming up that will add the second api for this same endpoint to an Island as well so that would be a good moment to add this test

dotcom-rendering/src/web/lib/formatCount.test.ts
jamesgorrie: Interesting as the closest library I can see is [`timeAgo`](https://github.com/guardian/libs#timeago). Do we store content styleguide type libs anywhere? Or is @guardian/libs it?

dotcom-rendering/src/web/lib/useDiscussion.ts
jamesgorrie: Ah, great. I was thinking of a test that ensured we have SWR worked as expected with islands, but that could be a little close to testing the SWR functionality, but would give us confidence in the question above. It would also then define that as our line in the sand and be able to swap out SWR with confidence at a later date.

dotcom-rendering/src/web/components/ArticleMeta.stories.tsx
jamesgorrie: Ah, OK, so it is an expected string outside of testing? In which case this makes perfect sense.

dotcom-rendering/src/web/lib/formatCount.test.ts
oliverlloyd: There's the [Guardian style guide](https://www.theguardian.com/guardian-observer-style-guide-n) which says

> numbers
> Spell out from one to nine; numerals from 10 to 999,999; thereafter use m, bn or tn for sums of money, quantities or inanimate objects in copy, eg 5m tonnes of coal, 30bn doses of vaccine, £50tn; but million or billion for people or animals, eg 1 million people, 25 million rabbits, the world population is 7 billion, etc; in headlines always use m, bn or tn

Which seems like the right reference for any wider number formatting solution.

Code wise, I think libs is it. Source has a lot of centralised styling but is not where we put utility functions like this. Although, there really does seem to be some crossover here between style and code.

@SiAdcock @sndrs what are your thoughts?

dotcom-rendering/src/web/lib/useDiscussion.ts
oliverlloyd: My aim is to load a page that has comments on and then count the number of api requests that get made to this endpoint. The test will expect there to only be one. This test will pass now because we centralise the call and share state and the same test will pass with SWR because that lib solves the same problem another way (using a cache).

Ideally here we would add the test in _before_ this PR so we could show deduplication working before and after but because of the point I made earlier where SWR can't share a cache between App and the Islands, that test would actually fail with this PR and only start working again with the next one

DavidLawes (commented):
These changes make sense to me - thanks for explaining. I wasn't sure if the outstanding question (about where `formatCount` should live) is a blocker or not

DavidLawes (approved):

dotcom-rendering/src/web/lib/formatCount.test.ts
sndrs: yeah this feels like a libs to me too. i think of source at least needing to be doing _something_ will CSS, although maybe that's only in my head.

sndrs: i guess it would be something like `formatNumberAsWords`? we could group styling ones like this `formatXasX` etc

PR #3801, opened 2022-01-06
Hydrate Storybook Islands

oliverlloyd:
## What does this change?
Adds support for hydration of `Islands` within our stories

## Why?
We want the full page stories that we use to test layout code to be as fully featured as possible. To achieve this we need to hydrate the content.

But our storybook pages are not created using the same build process as normal pages so we need a custom solution for hydration - we kind of have to simulate the SSR/Browser flow, except in Storybook it's all happening in the browser.

## Why are we not using the same `doHydration` function as we use on production?
We already have a function that checks the page for `gu-island` tags and hydrates those based on the given props so why not use this? There are two reasons:

1) We don't want to defer hydration because this could cause the snapshots we create in Chromatic to be incomplete.
2) Because we don't have the server/browser separation we started to see errors complaining about different versions of React. By having a custom `doStorybookHydration` function we can work around this.

### DIFF
--- a/dotcom-rendering/src/web/browser/islands/doStorybookHydration.js
+++ b/dotcom-rendering/src/web/browser/islands/doStorybookHydration.js
@@ -1,8 +1,8 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 
-import { getName } from '@root/src/web/browser/islands/getName';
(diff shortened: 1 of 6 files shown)


OllysCoding (approved):
+1 All makes sense to me!

PR #3804, opened 2022-01-07
Pin the `yarn` binary version to v1.22.17

mxdvl:
## What does this change?

Bumps the yarn policy.

## Why?

When running commands, we currently have two versions of `yarn 1`: the latest `1.22` and the previously pinned `1.15`.

screenshot of conflicting yarn versions

Initially introduced in #1322


OllysCoding:
Since yarn runs in the root directory now that we use yarn workspaces, should we move the .yarnrc to the root directory as well?

mxdvl:
@OllysCoding any outstanding issue with merging this?

OllysCoding (approved):
+1 Go for it!

mxdvl:
Thanks @JamieB-gu for bringing attention to this!

PR #3808, opened 2022-01-10
SPIKE: using Esbuild loader to reduce build times

joecowton1:
## What does this change?

This is the beginnings of a spike to look into using `esbuild-loader` in Webpack to reduce dev-server build times. 

## What's good about this

Using `esbuild-loader` reduces build times significantly. Using the [speed-measure-webpack-plugin
](https://www.npmjs.com/package/speed-measure-webpack-plugin) we can observe quite how big a difference this makes:

| Before | After |
| --- | --- |
|  |  |

## What's less good about this

It doesn't actually work currently. In DCR we use [Loadable-Components](https://loadable-components.com/) to chunk our components, this relies on Webpack populating a file called `loadable-manifest-browser.json` (and `loadable-manifest-browser.legacy.json`)  in the `dist` folder with  the correct js chunk names for each loadable component.  This is not happening, and as such when a page loads, the correct js chunks cannot be found and the page crashes.

@oliverlloyd has suggested moving away from `Loadable-Components` - maybe this is a solution?

## Workarounds

In order to get Webpack to run using `esbuild-loader` we've had to employ a number of workarounds, I've commented these in the file changes but in summary:
- use of `@roots` requires `TsconfigPathsPlugin` to resolve aliased paths
- files that use JSX but don't explicitly import React need the `webpack.ProvidePlugin`
- this spike does not account for `ie11`

## Caveats when using Esbuild with Typescript

https://esbuild.github.io/content-types/#typescript-caveats

It may be worth using [this](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin)


dotcom-rendering/scripts/webpack/frontend.js
```diff
@@ -22,16 +23,20 @@ const commonConfigs = ({ platform }) => ({
 			: 'eval-cheap-module-source-map',
 	resolve: {
 		alias: {
-			'@root': path.resolve(__dirname, '.'),
-			'@frontend': path.resolve(__dirname, 'src'),
+			// '@root': path.resolve(__dirname, '.'),
```
joecowton1, line 26: Esbuild was struggling to resolve these aliases.  The `TsconfigPathsPlugin` used below resolves this issue in the majority of cases.

dotcom-rendering/scripts/webpack/frontend.js
```diff
@@ -22,16 +23,20 @@ const commonConfigs = ({ platform }) => ({
...
 			react: 'preact/compat',
 			'react-dom/test-utils': 'preact/test-utils',
 			'react-dom': 'preact/compat',
 		},
+		plugins: [new TsconfigPathsPlugin()],
 		extensions: ['.js', '.ts', '.tsx', '.jsx'],
 		symlinks: false,
 	},
 	plugins: [
+		new webpack.ProvidePlugin({
```
joecowton1, line 37: `esbuild-loader` was struggling with files using `jsx` where React was not imported, the `webpack.ProvidePlugin` resolves that issue.

dotcom-rendering/tsconfig.json
```diff
@@ -65,10 +65,13 @@
...
         "forceConsistentCasingInFileNames": true,
         "resolveJsonModule": true,
-        "preserveConstEnums": true
+        "preserveConstEnums": true,
         /* Experimental Options */
         // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
         // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
-    },
-    "exclude": ["cypress"]
+		"skipLibCheck": true
```
joecowton1, line 72: The `esbuild-loader` package contains `ts` errors which don't pass the commit hook, using `skipLibCheck` allows us to push the branch though would need a better long term solution.

dotcom-rendering/src/app/server.ts
```diff
@@ -4,17 +4,17 @@ import responseTime from 'response-time';
 
 import compression from 'compression';
 
-import { log, warn } from '@root/scripts/env/log';
+import { log, warn } from '../../scripts/env/log';
```
joecowton1, line 7: This file alone still struggles with the `@root` alias.

oliverlloyd:
This is really great! I checked this out locally and things _feel_ substantially better. There's still that delay on the first page loading even though the CLI claims the dev server is ready but it's a lot smaller now

> use of @roots requires TsconfigPathsPlugin to resolve aliased paths

Maybe we don't need these aliases? If they're adding complexity then it might be worth considering adopting relative paths. We already use relative paths in most files now.

> files that use JSX but don't explicitly import React need the webpack.ProvidePlugin

Could we simply add this import where needed or is that not possible?

> this spike does not account for ie11

There's not been an official decision on this but if we were to make it so that IE11 was not served javascript (and readers using that browser saw the no js experience) then this would remove this issue

oliverlloyd:
Regarding `loadable`, as I've mentioned elsewhere, we're on a path to migrate away from this lib towards the islands pattern. I am going to see about accelerating this migration as a means of freeing up this PR

joecowton1:
> Regarding `loadable`, as I've mentioned elsewhere, we're on a path to migrate away from this lib towards the islands pattern. I am going to see about accelerating this migration as a means of freeing up this PR

Great - please say if there's anything I can do as part of this process.

joecowton1:
> Maybe we don't need these aliases? If they're adding complexity then it might be worth considering adopting relative paths. We already use relative paths in most files now.

I looked at this and I think there are around 900 references to `@root` and maybe 300ish for `@frontend` - we could switch to relative paths, and that would remove the need for the additional plugin, but that is quite a few changes to make.

joecowton1:
> Could we simply add this import where needed or is that not possible?
> 
> >

Same with this one, we could just add the import but might involve a lot of changes. Is there a reason to avoid using more plugins?

OllysCoding:
> There's still that delay on the first page loading even though the CLI claims the dev server is ready but it's a lot smaller now

joecowton1:
> @oliverlloyd I believe this is due to the fact that the CLI feedback from the dev-server is reporting only on the _server_ build, but not on either of the client builds, which appear to take longer.

Maybe there's a task here to add reporting for the client too?

oliverlloyd:
> around 900 references to `@root`
>

Just one or two then!

I honestly don't have strong opinions either way. If we can make aliases work then it's true they do bring benefits. Removing them would be a lot of edits (I think you could only do this with a script of some sort) but could help in terms of simplifying tooling, I've definitely seen alias cause snags in various changes we've done

Lets see how we go with this plugin but keep the nuclear option open - it might not be as hard to migrate as we think. We could _defo_ remove `@frontend` as this seems to achieve nothing and could be replaced with `@root`

oliverlloyd:
> > Could we simply add this import where needed or is that not possible?
> > >
> 
> Same with this one, we could just add the import but might involve a lot of changes. Is there a reason to avoid using more plugins?

Fewer plugins just means less maintenance in future. I'm also wary of using webpack to add code imports, I'd much prefer we be explicit about what a file needs and not create couplings to the build system as far as possible

joecowton1:
> > > Could we simply add this import where needed or is that not possible?
> > > >
> > 
> > 
> > Same with this one, we could just add the import but might involve a lot of changes. Is there a reason to avoid using more plugins?
> 
> Fewer plugins just means less maintenance in future. I'm also wary of using webpack to add code imports, I'd much prefer we be explicit about what a file needs and not create couplings to the build system as far as possible

Ok, makes sense!

dotcom-rendering/scripts/webpack/browser.js
```diff
@@ -66,48 +72,30 @@ module.exports = ({ isLegacyJS }) => ({
 		rules: [
 			{
 				test: /\.[jt]sx?|mjs$/,
+				loader: 'esbuild-loader',
```
arelra, line 75: Minor but I don't think `esbuild-loader` has been added to the dependencies?
It must be coming in through another dependency - storybook I think.
It's a non-issue for this spike but it might cause some head-scratching if you're expecting features or perf benefits from the latest version.

dotcom-rendering/scripts/webpack/server.js
```diff
@@ -32,38 +32,31 @@ module.exports = () => ({
 	module: {
 		rules: [
 			{
-				test: /(\.tsx|\.js|\.ts)$/,
-				exclude: {
-					and: [/node_modules/],
-					not: [/@guardian/, /dynamic-import-polyfill/],
+				test: /\.js?$/,
+				loader: 'esbuild-loader',
+				exclude: module.exports.babelExclude,
```
arelra, line 37: I don't think `module.exports.babelExclude` is available in this module?

Might get a small perf boost if the previous excludes are added back.

joecowton1:
Closing for now @OllysCoding

PR #3812, opened 2022-01-10
`MostViewedRight` as an Island

oliverlloyd:
## What does this change?
Moves the `MostViewedRight` component away from the `Portal` pattern and uses the `Island` one instead

## Why?
This is part of a wider piece of work to [migrate away](https://docs.google.com/document/d/1kSrZZrJ8hSw6bek5hGa9onE8QDz5TRHFqHV52EOFjts/edit#heading=h.gsod33a0pc47) from `App.tsx`

## What happened to `MostViewRightIsland`?
I also took this opportunity to refactor away the `MostViewRightIsland` component. I think this was needed previously as part of how we hydrated things in two places but now that we're only calling the component in one place we can merged the wrapper css into the main `MostViewedWrapper` component. Note. The reason for the wrapper, and indeed the reason for hydration in the first place, is so that this component can read the height of it's container and use that to decide if it should appear or not.

## Why do we have a flatter file structure?
To make the dynamic import for for the island I had to move these components out of their subfolder and down into the root of `/components`


dotcom-rendering/src/web/components/MostViewedRightWrapper.importable.tsx
```diff
@@ -11,8 +11,14 @@ type Props = {
...
-const flexGrow = css`
-	flex-grow: 1;
+const MOSTVIEWED_STICKY_HEIGHT = 1059;
+
+// Styling the data island root so it stretches to cover the full height available in the container.
+// Requires us to subtract the height of its sibling in the container (StickyAd).
+const stretchWrapperHeight = css`
+	display: flex;
+	flex-direction: column;
+	height: ${`calc(100% - ${MOSTVIEWED_STICKY_HEIGHT}px)`};
```
oliverlloyd, line 21: This css was previously being added by `MostViewedRightIsland` but is now included here

dotcom-rendering/src/web/components/MostViewedRightWrapper.importable.tsx
```diff
@@ -11,8 +11,12 @@ type Props = {
...
 const HEIGHT_REQUIRED = 482 + 24 + 24;
 
-const flexGrow = css`
-	flex-grow: 1;
+const MOSTVIEWED_STICKY_HEIGHT = 1059;
+
+// Styling the data island root so it stretches to cover the full height available in the container.
+// Requires us to subtract the height of its sibling in the container (StickyAd).
+const stretchWrapperHeight = css`
+	height: ${`calc(100% - ${MOSTVIEWED_STICKY_HEIGHT}px)`};
```
oliverlloyd, line 19: We previously achieved this in `MostViewedRightIsland` but I've now merged that file into this one given we only set this component in one place now. This also negates the need for the `flex` styles.

I wonder if those flex styles were ever needed 🤔

dotcom-rendering/src/web/components/MostViewedRight.test.tsx
```diff
@@ -1,14 +1,14 @@
 import { render } from '@testing-library/react';
 
-import { useApi as useApi_ } from '@root/src/web/lib/useApi';
+import { useApi as useApi_ } from '../lib/useApi';
```
OllysCoding, line 3: Is there a reason to switch from absolute to relative imports?

OllysCoding (approved):
+1  👏 Kill 👏 Off 👏 Loadable

dotcom-rendering/src/web/components/MostViewedRight.test.tsx
oliverlloyd: There are various benefits to relative paths as well as cons. It's a whole trade offs thing. But I think in most cases it's just easier to allow VSCode to automatically add them. I'm pretty sure this is the reason everyone keeps doing this

jamesgorrie:

PR #3813, opened 2022-01-11
Remove `elements` subdirectory

oliverlloyd:
## What?
This PR removes the `components/elements` directory, moving all the element components to the root of `components`.

## Why?
The flat vs. subdirectories argument has raged for years but this PR is not about that. We need this change if we want dynamic imports to work when we use the Island pattern to import these files.

WebPack needs to be told what files we might want to dynamically import so that it can prepare and bundle the code appropriately at build time, code splitting basically. We do this in `doHydration` using:

```js
	import(
		/* webpackInclude: /\.importable\.(tsx|jsx)$/ */
		`../../components/${name}.importable`
	)...
```

This is saying, prepare bundles for any file with `.importable` in the filename that lives in the root of the `/components` directory. It fails if the file you want to dynamically import is in a subdirectory.


jamesgorrie (commented):
You're getting 500s in Cypress tests, assuming some imports are broken?

Is it worth updating [the error message here](https://github.com/guardian/dotcom-rendering/blob/843b561012680eb399f46ad3389eaaffa4d78cb5/dotcom-rendering/src/web/browser/islands/doHydration.ts#L31) to explain the restrictions on the naming conventions we have enforced?

A little test on `doHydration`  could also be a nice place to encapsulate the nice description on this PR.

oliverlloyd:
> You're getting 500s in Cypress tests, assuming some imports are broken?

Yeah, the loadable imports were failing. Hopefully fixed now

> 
> Is it worth updating [the error message here](https://github.com/guardian/dotcom-rendering/blob/843b561012680eb399f46ad3389eaaffa4d78cb5/dotcom-rendering/src/web/browser/islands/doHydration.ts#L31) to explain the restrictions on the naming conventions we have enforced?
> 

Great idea! Done.

> A little test on `doHydration` could also be a nice place to encapsulate the nice description on this PR.

Sorry if I'm misreading this comment but I think what you're saying is can we have tests against `doHydration`? If so, I think what @mxdvl was talking about yesterday applies here. That it's better to test the interfaces to things rather than the internal code. So we do definitely need more tests around hydration but not on the internal code that applies it. Instead it's better to have those tests checking the outputs, not the implementation.

A really good example for why I like this approach is that we do already have Cypress tests that verify that hydration is working and these tests have continued to run and work throughput the process of us completely rebuilding how we execute hydration. These Cypress tests have been useful as a guide for that refactoring work without any need to edit or change them and the reason this was possible was because they are simply loading a page and verifying that things are still interactive, which is what we really care about.

DavidLawes (approved):
I can see why we need to do this and the change makes sense. My only hesitation approving this was that it forces more of a flat hierarchy (and I'm not sure how I feel about a flat structure in general).

oliverlloyd:
> I'm not sure how I feel about a flat structure in general

This is fair. I didn't want this PR to force one opinion or another but - for better or worse - the technical constraints here are leading us towards a flat structure. I have another PR pending for `MostViewed` which has to make a similar refactor in order for the `import` to work

PR #3814, opened 2022-01-11
Revert "Temporarily remove sign up and log in option when discussion is disab…"

ioannakok:
## What does this change?
Reverts guardian/dotcom-rendering#3806

## Why? 
The identity upgrade has completed successfully and this change is no longer needed.


DavidLawes (approved):
👍

PR #3817, opened 2022-01-11
DCR: Remove legacy build from dev server

OllysCoding:
Part of #3884
## What does this change?

When running the DCR dev-server 3 webpack builds kick off:

- Server build (this is the one you see the console output for)
- Client build
- Legacy client build

Given that the majority of the time you're using the dev server, you probably will be using a modern browser (one which supports modules), it doesn't make sense to spend time building & rebuilding the legacy client bundle.

To do this, I modified the webpack config to not include the legacy bundle by default for deb builds and added a second command - `make dev-legacy` - for if you want to run the dev server with legacy bundles included.

To get pages to render with legacy bundles, I had to make some slight tweaks to `document.tsx` & `assets.tsx`, both of which were expecting the legacy versions of files to always exist.

## Why?

Based off my own testing, this reduces the initial build time (as reported by webpack) by up to 40%!


OllysCoding:
[Investigate removing legacy bundle from dev-server build](https://trello.com/c/lC3A0onD/530-investigate-removing-legacy-bundle-from-dev-server-build)

dotcom-rendering/src/lib/assets.ts
```diff
@@ -38,15 +38,24 @@ export const loadableManifestJson = loadableManifest;
 export const getScriptArrayFromFilename = (
 	filename: string,
 ): { src: string; legacy: boolean }[] => {
+	// Get legacy file name if one's available
```
jamesgorrie, line 41: Maybe worth mentioning the `INCLUDE_LEGACY` here so people know where to look for this when coming back to this?

jamesgorrie (approved):
Really nice improvement!

oliverlloyd:
One more thought, is there value in doing the same for the `make run-ci` command? I use that locally a lot now and it we could also get a speed bump for CI so much the better

dotcom-rendering/src/web/server/document.tsx
```diff
@@ -289,7 +289,7 @@ export const document = ({ data }: Props): string => {
 	)[0];
 	const gaPath = {
 		modern: modernScript.src,
-		legacy: legacyScript.src,
+		legacy: legacyScript && legacyScript.src,
```
arelra, line 292: Super minor :)

```suggestion
		legacy: legacyScript?.src,
```

jamesgorrie (approved):
Looks good. Is there any part of the discussion that's worth adding to this PR from the group chat (UX / reporting considerations)?

jamesgorrie (approved):

OllysCoding:
Part of #3884

PR #3824, opened 2022-01-12
refactor: `InstagramBlockComponent` as an Island

oliverlloyd:
## What does this change?
Here we migrate the `InstagramBlockComponent` to the Island pattern.

## Why?
This is part of a wide refactor to remove `App.tsx`

## Why have I dropped `ClickToView` down?
We now wrap the element in a click to view overlay inside the component instead of setting it externally in `renderElement` and `App`. This was done because this approach makes more sense with the islands pattern where we only want to pass a single child to `Island`

Note. I did consider making a separate PR to refactor all usage of `ClickToView` to this new structure prior to moving them over to Islands but that would have meant adding a new prop (`isMainMedia`) only all the element type definitions which would have been unnecessary noise as this extra prop is not needed when using an Island so by making both changes in one place we remove the requirement for it completely.

**Test article for future developers: https://www.theguardian.com/lifeandstyle/2020/apr/27/its-like-a-sexy-story-just-for-me-how-lockdown-has-triggered-a-wave-of-sexting**


dotcom-rendering/src/web/components/ClickToView.stories.tsx
```diff
@@ -1039,23 +1039,12 @@ export const InstagramBlockComponentStory = () => {
...
-							key={1}
-							element={instagramInstramEmbed}
-							index={1}
-						/>
-					</ClickToView>
+					<InstagramBlockComponent
+						key={1}
+						element={instagramInstramEmbed}
+						index={1}
+						isMainMedia={false}
```
oliverlloyd, line 1046: I've added this prop so we can pass it to `ClickToView`

dotcom-rendering/src/web/components/InstagramBlockComponent.importable.tsx
```diff
@@ -0,0 +1,42 @@
...
+		<ClickToView
+			role={element.role}
+			isTracking={element.isThirdPartyTracking}
+			isMainMedia={isMainMedia}
+			source={element.source}
+			sourceDomain={element.sourceDomain}
+			onAccept={() =>
+				updateIframeHeight(`iframe[name="instagram-embed-${index}"]`)
+			}
+		>
```
oliverlloyd, line 29: This is new code

dotcom-rendering/src/web/components/InstagramBlockComponent.importable.tsx
```diff
@@ -0,0 +1,42 @@
+import { css } from '@emotion/react';
+import { updateIframeHeight } from '../browser/updateIframeHeight';
```
oliverlloyd, line 2: New import

OllysCoding:
👍 I like the move of click-to-view into the instagram component itself. Since we shouldn't be using the instagram component without c2v, I think it works to reinforce that c2v is a requirement for the instagram component (and others), rather than just what it happens to be wrapped in most of the time.

It'll be great if we can continue this change with other components as the loadable -> island migration continues!

OllysCoding (approved):

oliverlloyd:
> 👍 I like the move of click-to-view into the instagram component itself. Since we shouldn't be using the instagram component without c2v, I think it works to reinforce that c2v is a requirement for the instagram component (and others), rather than just what it happens to be wrapped in most of the time.
> 
> It'll be great if we can continue this change with other components as the loadable -> island migration continues!

Great point about having ClickToView being local meaning it's intrinsic to the definition of the embed. I like this

jamesgorrie:

PR #3392, opened 2021-09-07
Storybook fixes & cleanup

OllysCoding:
## What does this change?

Update to use source resets, across storybook, AR, and DCR. As AR was not using any resets before, some changes had to be made to support them.

The following fixes were applied

- Fix issues introduced to AR by using a CSS reset
- Remove white background on storybook which breaks some DCR stories
- Fix fonts for AR in storybook (?)

## Why?

Prior to this change, resets were applied to AR in storybook, but not in production, meaning that stories often displayed incorrectly. By moving the resets to source we've introduced a more robust standard for resets across the guardian, where previously they were only used in DCR.

PR #3825, opened 2022-01-12
Remove unused babel plugins `preval` and `prismajs`

arelra:
## What does this change?

Whilst working on the [swc-loader PR](https://github.com/guardian/dotcom-rendering/pull/3820) I noticed the babel plugins `preval` and `prismajs` don't seem to be used - please verify though.


oliverlloyd (approved):
`prismjs` was previously used for syntax highlighting on the `CodeBlockComponent` but this feature [got turned off last year](https://github.com/guardian/dotcom-rendering/pull/3011) due to an issue

This plugin should probably have also been removed at the same time so I think it's fair to remove it here now 👍 

`babel-plugin-preval` also [seems to be unused](https://github.com/guardian/dotcom-rendering/search?q=preval) so 👍 there too

OllysCoding:
Just to confirm that I removed all the usages of `babel-plugin-preval` a while ago in #3392 but apparently forgot to remove the package, so thank you!

PR #3827, opened 2022-01-12
Remove reference to adomik switch

simonbyford:
## What does this change?

Removes the reference to the adomik switch in the test fixtures config

## Why?

The switch was removed in a separate PR: https://github.com/guardian/frontend/pull/24549


chrislomaxjones:
You might need to also run `make gen-fixtures` to re-generate any fixtures that contain `adomik: true`. This will possibly introduce some other changes to the PR though if the fixture articles have changed, but that shouldn't be a problem.

simonbyford:
Lots of unrelated changes as you predicted @chrislomaxjones but hopefully these are harmless. Most of it is just noise from updating `elementId`

chrislomaxjones (approved):
Nice, I think all of the extra changes from `make gen-fixtures` are fine in this instance.

Although these fixture changes will have an impact on the Chromatic UI tests, so those will need checking by someone who knows what they're looking at!

PR #3828, opened 2022-01-12
Use modular AWS SDK for faster build times, send CloudWatch metrics with correct stage

MarSavar:
## Why?
Running [SMP](https://github.com/stephencookdev/speed-measure-webpack-plugin) revealed that the AWS library was taking up a significant chunk of the build time. The version we are currently using has a monolithic architecture, meaning that the entire SDK is imported and Webpack isn't smart enough to treeshake the unused modules.
Luckily, v3 adopts a modular architecture, meaning that we can import just the modules we need, thus reducing bundle size and build time.

This PR proposes replacing the huge `aws-sdk` import with just two modules, the only ones we need for DCR:

- `@aws-sdk/client-cloudwatch`
- `@aws-sdk/client-ssm`

The syntax for these modules is slightly different from the v2 SDK so there is some slight refactoring involved in the TypeScript files that use the AWS library.

Additionally, while working on this, it became apparent that we were logging CloudWatch events in AWS using the `PROD` stage in all cases, meaning that `CODE` logs were being lumped with and recorded as `PROD`. This PR also addresses this issue by using the relevant stage environment variable (thanks @davidfurey for the tip).
We should probably have a discussion re: whether we want to continue logging metrics in a non-production environment.

### Before
image

### After
It varies, but there seems to be a consistent 10-15s improvement.


MarSavar:
Tested this on CODE and CloudWatch metrics are being logged correctly.

image

dotcom-rendering/src/app/aws/metrics-baseline.ts
```diff
@@ -1,16 +1,21 @@
 import os from 'os';
 import { BytesMetric, collectAndSendAWSMetrics } from './aws-metrics';
 
-const maxHeapMemory = BytesMetric('rendering', 'PROD', 'max-heap-memory');
-const usedHeapMemory = BytesMetric('rendering', 'PROD', 'used-heap-memory');
+const stage =
+	typeof process.env.GU_STAGE === 'string'
+		? process.env.GU_STAGE.toUpperCase()
+		: 'DEV';
```
OllysCoding, line 7: 🎉🎉🎉

dotcom-rendering/package.json
```diff
@@ -64,7 +66,7 @@
     "@types/sanitize-html": "^1.27.1",
     "ajv": "^8.1.0",
     "ajv-formats": "^2.0.2",
-    "aws-sdk": "2.814.0",
+    "babel-plugin-prismjs": "^2.0.1",
```
OllysCoding, line 69: I think this being added is a the result of a merge conflict, it probably needs to be removed again!

OllysCoding (approved):
Looks great! Just need the prismjs package re-removed :)

dotcom-rendering/package.json
MarSavar: Yup, done!

arelra: Oh this was recently removed [here](https://github.com/guardian/dotcom-rendering/pull/3825) - a rebase and fixing any merge conflicts should sort it.

MarSavar: Fixed it, thanks Ravi!

oliverlloyd (approved):
Really??!
Screenshot 2022-01-12 at 17 44 35

I don't think this needs to be part of this (really great) PR but I'm just wondering, do you think this env var `DISABLE_LOGGING_AND_METRICS` is or could be used to prevent these imports?

davidfurey (approved):
Great, thanks Mario!

MarSavar:
> I don't think this needs to be part of this (really great) PR but I'm just wondering, do you think this env var `DISABLE_LOGGING_AND_METRICS` is or could be used to prevent these imports?

It would be great if we could avoid importing AWS altogether when working locally - thanks for the suggestion, I'll card it up!

MarSavar:
[Use modular imports for the AWS SDK](https://trello.com/c/4VkVH1IQ/529-use-modular-imports-for-the-aws-sdk)

jamesgorrie:
part of #3884

PR #3830, opened 2022-01-13
`RichLinkComponent` as an island

oliverlloyd:
## What does this change?
This PR brings the `RichLinkComponent` over to the island pattern

## Why?
To continue the work to remove App.tsx

## `ajaxUrl`
This value was previously only used on the client but it is now needed on the server and, in order to make it available to the code which is rendering this element on the server, it needs to be prop drilled down


dotcom-rendering/src/web/components/RichLink.tsx
```diff
@@ -21,7 +21,7 @@ import { Avatar } from '@frontend/web/components/Avatar';
 interface Props {
 	richLinkIndex: number;
 	cardStyle: RichLinkCardType;
-	imageData: RichLinkImageData;
+	imageData?: RichLinkImageData;
```
oliverlloyd, line 24: By making this optional we allow for the variation that is wanted in blogs where the image is not shown

dotcom-rendering/src/web/components/RichLinkComponent.importable.tsx
```diff
@@ -53,19 +53,18 @@ const buildUrl: (element: RichLinkBlockElement, ajaxUrl: string) => string = (
...
 	richLinkIndex,
 }: Props) => {
-	const url = buildUrl(element, ajaxEndpoint);
+	const url = buildUrl(element, ajaxUrl);
 	const { data, error } = useApi<CAPIRichLinkType>(url);
 
 	if (error) {
 		// Send the error to Sentry
-		window.guardian.modules.sentry.reportError(error, 'rich-link');
+		window?.guardian?.modules?.sentry?.reportError(error, 'rich-link');
```
OllysCoding, line 64: +1

OllysCoding (approved):
+1 Lots of prop drilling but unavoidable! Looks good :)

PR #3833, opened 2022-01-14
Default to `SpecialReport` instead of `News` for more greys

oliverlloyd:
## What does this change?
This changes the format used for the `DefaultRichLink` which is the server rendered, basic version of a rich link that gets shown prior to the client side enhancement

## Why?
It's not accurate to, for example, show sport rich links using the reds of news. Arguably `SpecialReport` is equally 'incorrect' but like this we get neutral greys which serve as better defaults

## Why not create custom default colours?
I considered making `format` optional and creating a custom version of the component for when no `format` was known (pre client enhancement) but this would be a lot of extra code and is less declarative that just using `SpecialReport`

## Why are we not enhancing liveblock richlinks?
This is outside the scope of this PR but you'll notice the chromatic diffs are showing the unenhanced default rich links. This is because we still need to implement this feature for this type of article in storybook

### Before
Screenshot 2022-01-14 at 14 29 40

### After
Screenshot 2022-01-14 at 14 27 33

### DIFF
--- a/dotcom-rendering/src/web/components/DefaultRichLink.tsx
+++ b/dotcom-rendering/src/web/components/DefaultRichLink.tsx
@@ -1,4 +1,4 @@
-import { ArticleDesign, ArticleDisplay, ArticlePillar } from '@guardian/libs';
+import { ArticleDesign, ArticleDisplay, ArticleSpecial } from '@guardian/libs';
 
 import { RichLink } from '@root/src/web/components/RichLink';
 
@@ -33,7 +33,7 @@ export const DefaultRichLink: React.FC<DefaultProps> = ({
 			format={{
 				display: ArticleDisplay.Standard,
 				design: ArticleDesign.Standard,
-				theme: ArticlePillar.News,
+				theme: ArticleSpecial.SpecialReport,
 			}}
 			tags={[]}
 			sponsorName=""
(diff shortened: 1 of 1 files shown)


DavidLawes (commented):
Out of curiosity, why is the current default news and did anyone from CP/editorial request this change (or it's something we - dotcom - think could add value)?

DavidLawes:
@HarryFischer - not sure if you want to take a look at this one too? Have default themes for rich links ever been discussed? Maybe this is just something that'd be good for you to be aware of :)

Also @oliverlloyd - we're choosing `SpecialReport` because it gives us a grey, but could the `SpecialReport` theme ever change?

oliverlloyd:
> @HarryFischer - not sure if you want to take a look at this one too? Have default themes for rich links ever been discussed? Maybe this is just something that'd be good for you to be aware of :)
> 
> Also @oliverlloyd - we're choosing `SpecialReport` because it gives us a grey, but could the `SpecialReport` theme ever change?

I agree that picking a specific `theme` as the default has issues but this PR is not introducing that pattern, it's only changing the theme that gets chosen from `News` to `SpecialReport`.

I did consider changing the pattern to use a custom set of colours but it would add a lot of code and my feeling was if we get the result we need by just using SpecialReport then this is an easy win?

In the event `SpecialReport` were to change then we would end up back where we already are right now, with the default not looking quite right

HarryFischer:
Yep I think this makes sense - special report gives us a better default styling

DavidLawes (approved):
👍

PR #3836, opened 2022-01-17
Always return `DefaultRichLink` if there's no client side data 

oliverlloyd:
## What does this change?
This slightly refactors the code used to enhance richlinks such that we're now more explicit about what we're returning.

On the server, we render `DefaultRichLink`
On the client we make an api call and if we got an error or we're still waiting for a response, we render `DefaultRichLink`
Only when the api responds with `data` do we render the enhanced version of a richlink

For context, the different versions are

## Default
_no image, no brand colouring_
Screenshot 2022-01-14 at 14 27 33

## Enhanced
_with image, coloured to the News pillar (brand)_
Screenshot 2022-01-17 at 12 23 28


ioannakok (approved):
👍

PR #3837, opened 2022-01-17
Don't include AWS modules in the development bundle

MarSavar:
## Why?
As part of the ongoing effort to reduce the bundle size/build time of the local development build, this PR excludes AWS modules from the development bundle. They are used for logging metrics in CODE/PROD, so there's no need to bundle them when running DCR locally.

### Before
image
image

### After
image
image

### DIFF
--- a/dotcom-rendering/scripts/webpack/server.js
+++ b/dotcom-rendering/scripts/webpack/server.js
@@ -19,6 +19,12 @@ module.exports = () => ({
 			allowlist: [/^@guardian/],
 		}),
 		({ request }, callback) => {
+			return !process.env.DISABLE_LOGGING_AND_METRICS &&
+				/^(@aws-.+\/).*/i.test(request)
+				? callback(null, `commonjs ${request}`)
+				: callback();
+		},
+		({ request }, callback) => {
 			return request.endsWith('loadable-manifest-browser.json')
 				? callback(null, `commonjs ${request}`)
 				: callback();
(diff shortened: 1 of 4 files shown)


MarSavar:
Tested on CODE, logging still works:

Screenshot 2022-01-17 at 17 31 16

oliverlloyd (approved):

dotcom-rendering/scripts/webpack/server.js
```diff
@@ -18,6 +18,12 @@ module.exports = () => ({
 		require('webpack-node-externals')({
 			allowlist: [/^@guardian/],
 		}),
+		({ request }, callback) => {
+			return process.env.NODE_ENV === 'development' &&
```
jamesgorrie, line 22: Would it be worth leaving a comment here saying it's in regard to speeding up the development build. I know the check indicates this, but it could be nice to say why.

dotcom-rendering/scripts/webpack/server.js
```diff
@@ -18,6 +18,12 @@ module.exports = () => ({
...
+				/^(@aws-.+\/).*/i.test(request)
```
jamesgorrie, line 23: nitpick: Could we use 
```suggestion
				request.startsWith('@aws-sdk')
```

to avoid having to speak RegExp?

jamesgorrie (approved):
A few questions on where this is happening

dotcom-rendering/scripts/webpack/server.js
MarSavar: Not a nitpick at all - it's a very good point! Thanks.

MarSavar:
[Avoid importing AWS modules in local dev build](https://trello.com/c/qiIfZUP6/534-avoid-importing-aws-modules-in-local-dev-build)

jamesgorrie:
part of #3884

PR #3838, opened 2022-01-17
Allow fetch options to be passed to `useApi`

oliverlloyd:
## What does this change?
Adds the ability to pass in a an options object to `useApi` that can be used to customise the `fetch` request

## Why?
We have a use case to use `credentials: 'include'` for a particular call in the future but that's not possible at the moment

## Thanks
Props to @mxdvl and @coldlink for essentially doing all the heavy lifting on this

### DIFF
--- a/dotcom-rendering/src/web/lib/useApi.tsx
+++ b/dotcom-rendering/src/web/lib/useApi.tsx
@@ -10,8 +10,8 @@ function checkForErrors(response: Response) {
 	return response;
 }
 
-const fetcher = (url: string) =>
-	fetch(url)
+const fetcher = (init?: RequestInit) => (url: string) =>
+	fetch(url, init)
 		.then(checkForErrors)
 		.then((res) => res.json());
 
@@ -27,12 +27,14 @@ interface ApiResponse<T> {
  * returns { loading, error, data }
  * @param {String} url - The url to fetch
  * @param {SWRConfiguration} options - The SWR config object - https://swr.vercel.app/docs/options
+ * @param {RequestInit} init - The fetch init object - https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#supplying_request_options
  * */
 export const useApi = <T,>(
 	url: string,
 	options?: SWRConfiguration,
+	init?: RequestInit,
 ): ApiResponse<T> => {
-	const { data, error } = useSWR(url, fetcher, options);
+	const { data, error } = useSWR(url, fetcher(init), options);
 
 	return {
 		data,


dotcom-rendering/src/web/lib/useApi.tsx
```diff
@@ -27,12 +27,14 @@ interface ApiResponse<T> {
  * returns { loading, error, data }
  * @param {String} url - The url to fetch
  * @param {SWRConfiguration} options - The SWR config object - https://swr.vercel.app/docs/options
+ * @param {RequestInit} init - The fetch init object - https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#supplying_request_options
  * */
 export const useApi = <T,>(
```
mxdvl, line 31: That dangly comma, though 👀

ioannakok (approved):
👍

OllysCoding (approved):
+1

dotcom-rendering/src/web/lib/useApi.tsx
```diff
@@ -27,12 +27,14 @@ interface ApiResponse<T> {
...
 	url: string,
 	options?: SWRConfiguration,
+	init?: RequestInit,
 ): ApiResponse<T> => {
```
mxdvl, line 36: Maybe using `Data` instead of `T` this type is more explicit, and `extends unkown` is better than the odd dangling comma to prevent TSX from tripping up?

```suggestion
export const useApi = <Data extends unknown>(
	url: string,
	options?: SWRConfiguration,
	init?: RequestInit,
): ApiResponse<Data> => {
```

dotcom-rendering/src/web/lib/useApi.tsx
```diff
@@ -27,12 +27,14 @@ interface ApiResponse<T> {
...
-	const { data, error } = useSWR(url, fetcher, options);
+	const { data, error } = useSWR(url, fetcher(init), options);
```
mxdvl, line 37: Maybe using `Data` instead of `T` this type is more explicit, and `extends unkown` is better than the odd dangling comma to prevent TSX from tripping up?

You can then also pass these types to the useSWR command, like so:

```suggestion
export const useApi = <Data extends unknown>(
	url: string,
	options?: SWRConfiguration,
	init?: RequestInit,
): ApiResponse<Data> => {
	const { data, error } = useSWR<Data, Error>(url, fetcher(init), options);
```

dotcom-rendering/src/web/lib/useApi.tsx
oliverlloyd: 🤷

mxdvl: I think it’s TS not playing nice with TSX. I’d recommend using using `T extends unknown` instead, which is more informative of what the type of `T` can be. More suggestions below, such as using `Data` instead of `T`.

dotcom-rendering/src/web/lib/useApi.tsx
oliverlloyd: Personally, I'm not especially triggered by the dangly comma. If the linting tooling needs this there then I'm very happy to let it do its thing. But maybe I'm missing something, do you think it could cause an issue?

oliverlloyd: That said, I'm very happy to make this change if it solves the problem!

But do you mind if we branch this off from this PR as this work is blocking another PR which is itself blocking the larger dev build work which I'm keen to progress!

mxdvl: The dangling comma [seems to be what people expect](https://github.com/prettier/prettier/issues/6302). I hold my mind’s breath, but I probably should just let it go… ;)

dotcom-rendering/src/web/lib/useApi.tsx
oliverlloyd: 👍  See: https://github.com/guardian/dotcom-rendering/pull/3840

dotcom-rendering/src/web/lib/useApi.tsx
oliverlloyd: I've created the PR for this change here: https://github.com/guardian/dotcom-rendering/pull/3840 - happy wither way, your call!

PR #3839, opened 2022-01-17
`Discussion` as an island

oliverlloyd:
## What does this change?
This moves `Discussion` over to the island pattern

## Along the way
To make this transition work the following refactors and changes were needed

### Discussion v9.1.1
It turns out `Discussion` wasn't being included in our automated accessibility checks but that when we loaded it onto the page via an island this issue was fixed. However, this meant we started to see new problems in the report failing the build. To get past this I fixed those issues in `discussion-rendering` and brought that new release in here

### No `user` in App.tsx
This PR removes the global state of `user` from App.tsx

### No more `beingHydrated`
This prop was needed to fix an issue with hydration in Storybook. This problem was solved by the islands pattern so this prop could be removed.

### `DiscussionContainer`
This is a wrapper component that is being used to compose the decision about if the user is signed in or not. If they are, we render another wrapper component `DiscussionWhenSigned` which is where we fetch the user profile, and if they are not, we directly render `Discussion` itself.

This level of composition is needed because you [cannot conditionally call a react hook](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level).

## Why?
🏖️ Islands are where we want to be 🏝️


dotcom-rendering/src/web/components/Discussion.importable.tsx
```diff
@@ -25,21 +25,6 @@ type Props = {
...
-	// We use this prop to solve a problem we have with Storybook. If you remove
-	// it then the page will render fine on production but our layout stories
-	// will render two copies of Discussion, side by side. The reason for this
-	// is because we technically server side render these layout stories on the
-	// client, inside the story file itself.
-	//
-	// Yes, it's true, having props purely to solve test implementation problems
-	// is not great. If you feel strongly about this and want to remove this
-	// prop I'm okay with that. If you were able to solve this another way
-	// then thank you!
```
OllysCoding, line 40: 👋

dotcom-rendering/src/web/components/Discussion.importable.tsx
```diff
@@ -74,7 +57,22 @@ export const Discussion = ({
...
+
+	const isSignedIn = !!getCookie({ name: 'GU_U', shouldMemoize: true });
+
+	const { data: user } = useApi<{ userProfile: UserProfile }>(
+		isSignedIn
+			? joinUrl(
+					discussionApiUrl,
+					'profile/me?strict_sanctions_check=false',
+			  )
+			: '',
```
OllysCoding, line 71: How does useApi behave when passed an empty string? Does it throw an error?

dotcom-rendering/src/web/components/Discussion.tsx
```diff
@@ -180,62 +164,29 @@ export const Discussion = ({
...
-								onPermalinkClick={handlePermalink}
-								apiKey="dotcom-rendering"
-								onExpanded={(value) => {
-									handleExpanded(value);
-								}}
-							/>
-						)}
-
-						{beingHydrated && !isExpanded && (
-							<Lazy margin={300} disableFlexStyles={true}>
```
oliverlloyd, line 212: We no longer need this Lazy component because Island provides this deferred loading built in

dotcom-rendering/src/web/components/Discussion.tsx
```diff
@@ -26,20 +25,6 @@ type Props = {
...
-	// will render two copies of Discussion, side by side. The reason for this
-	// is because we technically server side render these layout stories on the
-	// client, inside the story file itself.
-	//
-	// Yes, it's true, having props purely to solve test implementation problems
-	// is not great. If you feel strongly about this and want to remove this
-	// prop I'm okay with that. If you were able to solve this another way
-	// then thank you!
-	beingHydrated?: boolean;
-	// **************************************************************************
```
oliverlloyd, line 42: This prop was needed to fix an issue with hydration in Storybook. This problem was solved by the islands pattern so this prop could be removed.

dotcom-rendering/src/web/components/App.tsx
```diff
@@ -108,7 +106,6 @@ let renderCount = 0;
 export const App = ({ CAPI, ophanRecord }: Props) => {
 	log('dotcom', `App.tsx render #${(renderCount += 1)}`);
 	const isSignedIn = !!getCookie({ name: 'GU_U', shouldMemoize: true });
-	const [user, setUser] = useState<UserProfile | null>();
```
oliverlloyd, line 111: This PR removes the global state of `user` from App.tsx

dotcom-rendering/src/web/layouts/CommentLayout.tsx
```diff
@@ -639,20 +639,23 @@ export const CommentLayout = ({
...
+							<DiscussionContainer
+								discussionApiUrl={CAPI.config.discussionApiUrl}
+								shortUrlId={CAPI.config.shortUrlId}
+								format={format}
+								discussionD2Uid={CAPI.config.discussionD2Uid}
+								discussionApiClientHeader={
+									CAPI.config.discussionApiClientHeader
+								}
+								enableDiscussionSwitch={
+									CAPI.config.switches.enableDiscussionSwitch
```
oliverlloyd, line 653: We now set this property correctly using the actual switch value from CAPI

jamesgorrie:

dotcom-rendering/src/web/components/Discussion.tsx
```diff
@@ -180,62 +164,29 @@ export const Discussion = ({
 								/>
 							</div>
 						</Hide>
-
-						{beingHydrated && isExpanded && (
```
OllysCoding, line 184: Should we test what happens with expanded behaviour

ioannakok: +1 to Olly's suggestion

ioannakok (approved):
+1 I think this is a great refactoring

PR #3832, opened 2022-01-14
Single front door AB test updates

marialani:
## What does this change?
Currently we only show the the "Subscriptions" link in the main navigation to supporters. For the duration of the Single Front Door AB Test we will display a "Print Subscriptions" link which isn't dependant on the user's supporter status and whether they're in the AB test. 

We will also hide the Supporter Revenue links in the footer for all users.

## Changes
- Hide Supporter Revenue links in the footer
- Update "Subscriptions" link in desktop navigation
- Update "Subscriptions" link in mobile navigation

These changes have been reproduced in frontend here: https://github.com/guardian/frontend/pull/24554

## Screenshots
| Before | After |
|-- | -- |
| Screenshot 2022-01-14 at 13 05 20 | Screenshot 2022-01-14 at 12 57 39 |
| Screenshot 2022-01-14 at 13 16 49 | Screenshot 2022-01-14 at 13 16 36 |

## Why
To support the new product proposition work we are going to run an AB test to measure the effect of replacing the current contribute/subscribe split messaging with a single 'support the guardian' call to action.

PR #3841, opened 2022-01-18
Refactor Links.tsx to use react islands

OllysCoding:
## What does this change?

Refactors Links.tsx to use react islands

To achieve this I had to move the previously global state `userId` to within the component. We only want to request the user profile on the client, creating the `MyAccount` component allowed us to ensure it only requests the profile there.

I also generally refactored the component around how it hydrates. We can do this because now that this react component renders as it's own island, it's no longer affecting renders for anything else.

Inspiration for this credited to https://github.com/guardian/dotcom-rendering/pull/3747 (thanks @oliverlloyd)

> Note: This will introduce a duplicate call to the profile endpoint as Discussion also makes use of the App.tsx call we refactored away from. This duplication is only temporary as once discussion is migrated to use Islands, SWR will cache any duplicate calls. You can track progress on discussion migration in https://github.com/guardian/dotcom-rendering/pull/3839

## Why?

This helps with our general goal of migrating away from and deleting App.tsx, but also unblocks https://github.com/guardian/dotcom-rendering/pull/3832, which is having a hydration problem which is solved by islands.

## Testing

I've tested this locally, however since logging in does not work locally, we should also test on CODE to ensure functionality

### DIFF
--- a/dotcom-rendering/src/web/components/Links.importable.tsx
+++ b/dotcom-rendering/src/web/components/Links.importable.tsx
@@ -1,4 +1,3 @@
-import { useEffect, useState } from 'react';
 import { css } from '@emotion/react';
 
 import SearchIcon from '@frontend/static/icons/search.svg';
@@ -15,13 +14,13 @@ import { DropdownLinkType, Dropdown } from '@root/src/web/components/Dropdown';
 
 import ProfileIcon from '@frontend/static/icons/profile.svg';
 import { createAuthenticationEventParams } from '@root/src/lib/identity-component-event';
-import { useOnce } from '@frontend/web/lib/useOnce';
-import { getCookie } from '@guardian/libs';
+import { getCookie, joinUrl } from '@guardian/libs';
+import { useApi } from '@root/src/web/lib/useApi';
 import { getZIndex } from '../lib/getZIndex';
 
 type Props = {
 	supporterCTA: string;
-	userId?: string;
+	discussionApiUrl: string;
 	idUrl?: string;
 	mmaUrl?: string;
 };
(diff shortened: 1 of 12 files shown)


dotcom-rendering/src/web/components/Links.importable.tsx
```diff
@@ -135,34 +134,37 @@ const linksStyles = css`
...
+	const { data } = useApi<{ userProfile: UserProfile }>(
+		joinUrl(discussionApiUrl, 'profile/me?strict_sanctions_check=false'),
+		{},
+		{
+			credentials: 'include',
+		},
+	);
+
+	// If we don't have user data display sign in to the user. SWR will retry in the background if the request failed
+	if (!data?.userProfile?.userId) return <SignIn idUrl={idUrl} />;
```
oliverlloyd, line 167: +1 for the comment

Should we also check for `error` here? This is returned from `useApi`

oliverlloyd: We _could_ use a loading state here. It's not something for this PR but it might be more truthful that showing `SignIn` while we wait for the api call.

Just something to consider, no action needed

oliverlloyd (approved):
A minor comment on what to do if there's an error but not a blocker

dotcom-rendering/src/web/components/Links.importable.tsx
OllysCoding: > Should we also check for error here? This is returned from useApi

We could do, it would be relatively safe to assume that it wouldn't change anything (in the sense that if there's an error, data.userProfile.userId wouldn't be set), but I think it is good practice to check anyway so I will add it!

OllysCoding:
[Support RR & Move Links.tsx to Islands](https://trello.com/c/yN8Wz2Bz/555-support-rr-move-linkstsx-to-islands)

jamesgorrie:

PR #3842, opened 2022-01-18
Update to latest Source versions

joecowton1:
## What does this change?

Updates Source deps to the following:

`"@guardian/source-foundations": "^4.0.0-rc.7",`
`"@guardian/source-react-components": "^4.0.0-rc.5",`
`"@guardian/source-react-components-development-kitchen": "^0.0.25-rc.2",`

In order to maintain the existing styling on the `ToggleSwitch` Source component in AR's `MetaData` component I've added a `platform="ios"` prop. 

## Why?

So we can use the latest Source components in DCAR.

### DIFF
--- a/apps-rendering/src/components/metadata.tsx
+++ b/apps-rendering/src/components/metadata.tsx
@@ -237,6 +237,7 @@ const MetadataWithAlertSwitch: FC<Props> = ({ item }: Props) => {
 				{isLive(design) && (
 					<div css={css(toggleStyles, liveBlogPadding)}>
 						<ToggleSwitch
+							platform="ios"
 							checked={checked}
 							label={'Get alerts on this story'}
 							cssOverrides={toggleOverrideStyles}
--- a/package.json
+++ b/package.json
@@ -29,9 +29,9 @@
 	"devDependencies": {
 		"@babel/core": "^7.15.0",
 		"@guardian/eslint-plugin-source-react-components": "^4.0.0-rc.3",
-		"@guardian/source-foundations": "^4.0.0-rc.5",
-		"@guardian/source-react-components": "^4.0.0-rc.3",
-		"@guardian/source-react-components-development-kitchen": "^1.0.0-rc.1",
+		"@guardian/source-foundations": "^4.0.0-rc.7",
+		"@guardian/source-react-components": "^4.0.0-rc.5",
+		"@guardian/source-react-components-development-kitchen": "^0.0.25-rc.2",
 		"@storybook/addon-essentials": "^6.3.7",
 		"@storybook/addon-knobs": "^6.3.1",
 		"@storybook/builder-webpack5": "^6.3.11",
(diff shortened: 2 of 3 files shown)


apps-rendering/src/components/metadata.tsx
```diff
@@ -237,6 +237,7 @@ const MetadataWithAlertSwitch: FC<Props> = ({ item }: Props) => {
 				{isLive(design) && (
 					<div css={css(toggleStyles, liveBlogPadding)}>
 						<ToggleSwitch
+							platform="ios"
```
oliverlloyd, line 240: Is it worth adding this change to the PR description? Arguably, this is the _real_ change here and the bump to Source is what's needed to get here?

oliverlloyd (approved):
Approving the change but could be revamp the PR name/description to make this PR about the change to the toggle switch? It'd be great to include before and after screenshots in that too 🙏

joecowton1:
> Approving the change but could be revamp the PR name/description to make this PR about the change to the toggle switch? It'd be great to include before and after screenshots in that too 🙏

I have thought about this, but this isn't actually a change, rather we now need to add that prop to maintain the existing styles. We've changed the default `ToggleSwitch` in source from what was previously `ios` to `web` so adding `platform="ios"` is just to ensure we get the correct toggle still. This is just maintaining what previously existed.

I'm bumping the Source versions here so any issues there might be show up before I make the actual `ToggleSwitch` changes for the key events filter. 

So the before and after screenshots would be identical - if that's what you mean and still thing it's worth doing I can do that for sure. I'll certainly add a mention of the `ToggleSwitch` in the description either way :)

apps-rendering/src/components/metadata.tsx
joecowton1: Added 👍

oliverlloyd:
> we now need to add that prop to maintain the existing styles

Gotcha 👍 thanks for explaining. Shall we just add this detail to the description (or even just as a comment in the PR) so it's clear why the prop was changed.

Note. I often go through the diffs on my own PRs adding single comments on things that I think might be confusing or that need more explanation to smooth the flow for reviewers. A surprisingly large amount of time when I do this I find mistakes or things I could have done better!

oliverlloyd:
👍 

Lets get this out first thing tomorrow!

joecowton1:
> > we now need to add that prop to maintain the existing styles
> 
> Gotcha 👍 thanks for explaining. Shall we just add this detail to the description (or even just as a comment in the PR) so it's clear why the prop was changed.
> 
> Note. I often go through the diffs on my own PRs adding single comments on things that I think might be confusing or that need more explanation to smooth the flow for reviewers. A surprisingly large amount of time when I do this I find mistakes or things I could have done better!

Yeah great shout - I did this for a bit but have fallen out the habit, this PR is a good lesson though in making sure to explain things properly!

PR #3844, opened 2022-01-19
fix: Observer tag showing when publication tag is not set to observer

AshCorr:
fix: Observer tag showing when publication tag is not set to observer (Co-Author: email@oliverlloyd.com)

## What does this change?

Updates tagging logic.

## Why?

Previous logic scanned ALL tags included in an article and added a "The Observer" line to the page if it found any tags that had a title of "The Observer".

In reality we should be only checking if there is a tag of type "Publication" with the title "The Observer" before we add the observer line to the article.

See below for a Guardian article that has been incorrectly tagged as an Observer article.

### Before

image

### After

image


OllysCoding (approved):
LGTM

AshCorr:
[Check `type` when looking for The Observer tags](https://trello.com/c/IcUlgVSa/553-check-type-when-looking-for-the-observer-tags)

PR #3846, opened 2022-01-20
Add `urlkw` and `sens` to amp ad targeting

simonbyford:
Co-Authored-By: Chris Jones 

## What does this change?

Adds `urlkw` (URL keywords) and `sens` (sensitive) to amp ad targeting.

This applies to ads on articles and live blogs.

Screenshot 2022-01-20 at 11 58 25

## Why?

Spencer says "Adding this in will help with direct and Teads deal brand safety targeting" (see [ticket](https://trello.com/c/2fGmqIna/1419-sensitive-key-or-urlkw-added-as-value-in-amp))

### DIFF
--- a/dotcom-rendering/src/amp/components/Ad.tsx
+++ b/dotcom-rendering/src/amp/components/Ad.tsx
@@ -28,6 +28,24 @@ const amazonConfig = {
 	aps: { PUB_ID: '3722', PARAMS: { amp: '1' } },
 };
 
+const mapAdTargeting = (adTargeting: AdTargeting): AdTargetParam[] => {
+	const adTargetingMapped: AdTargetParam[] = [];
+
+	if (!adTargeting.disableAds) {
+		adTargetingMapped.push({
+			name: 'sens',
+			value: adTargeting.customParams.sens,
+		});
+
+		adTargetingMapped.push({
+			name: 'urlkw',
+			value: adTargeting.customParams.urlkw,
+		});
+	}
+
+	return adTargetingMapped;
+};
+
 export const realTimeConfig = (
 	usePrebid: boolean,
 	usePermutive: boolean,
@@ -79,6 +97,7 @@ export interface BaseAdProps {
 	section: string;
 	contentType: string;
 	commercialProperties: CommercialProperties;
+	adTargeting: AdTargeting;
 }
 
 interface AdProps extends BaseAdProps {
@@ -93,6 +112,7 @@ export const Ad = ({
 	contentType,
 	commercialProperties,
 	rtcConfig,
+	adTargeting,
 }: AdProps) => {
 	const adSizes = isSticky ? stickySizes : inlineSizes;
 	// Set Primary ad size as first element (should be the largest)
@@ -116,7 +136,12 @@ export const Ad = ({
 			data-loading-strategy="prefer-viewability-over-views"
 			layout="fixed"
 			type="doubleclick"
-			json={stringify(adJson(commercialProperties[edition].adTargeting))}
+			json={stringify(
+				adJson([
+					...commercialProperties[edition].adTargeting,
+					...mapAdTargeting(adTargeting),
+				]),
+			)}
 			data-slot={ampData(section, contentType)}
 			rtc-config={rtcConfig}
 		/>
--- a/dotcom-rendering/src/lib/ad-targeting.ts
+++ b/dotcom-rendering/src/lib/ad-targeting.ts
@@ -1,3 +1,15 @@
+import { isString } from '@guardian/libs';
+
+// TODO: this function already exists in commercial-core, consider exporting it to avoid duplication
+const getUrlKeywords = (url: string): string[] => {
+	const lastSegment = url
+		.split('/')
+		.filter(Boolean) // This handles a trailing slash
+		.slice(-1)[0];
+
+	return isString(lastSegment) ? lastSegment.split('-').filter(Boolean) : [];
+};
+
 export const buildAdTargeting = ({
 	isAdFreeUser,
 	isSensitive,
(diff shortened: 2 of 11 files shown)


dotcom-rendering/src/lib/ad-targeting.ts
```diff
@@ -1,3 +1,14 @@
+import { isString } from '@guardian/libs';
+
+const getUrlKeywords = (url: string): string[] => {
+	const lastSegment = url
+		.split('/')
+		.filter(Boolean) // This handles a trailing slash
+		.slice(-1)[0];
+
+	return isString(lastSegment) ? lastSegment.split('-').filter(Boolean) : [];
+};
```
simonbyford, line 10: This method was copied from [commercial-core](https://github.com/guardian/commercial-core/blob/main/src/targeting/content.ts#L91-L98) because it isn't exported there. This felt okay but I'm open to feedback

dotcom-rendering/src/amp/components/Ad.tsx
```diff
@@ -93,13 +94,27 @@ export const Ad = ({
...
+		adTargetingMapped.push({
+			name: 'sens',
+			value: String(adTargeting.customParams.sens),
+		});
+
+		adTargetingMapped.push({
+			name: 'urlkw',
+			value: adTargeting.customParams.urlkw as string[],
+		});
+	}
```
chrislomaxjones, line 116: Optional suggestion, but since this code doesn't have any side effects it might be nice to move this to its own function outside the component, so that below on L137 you can just call the function like

```typescript
...mapAdTargeting(adTargeting)
```

dotcom-rendering/src/amp/components/Ad.tsx
```diff
@@ -93,13 +94,27 @@ export const Ad = ({
...
+	const adTargetingMapped: AdTargetParam[] = [];
+	if (adTargeting.disableAds !== true) {
+		adTargetingMapped.push({
+			name: 'sens',
+			value: String(adTargeting.customParams.sens),
+		});
+
+		adTargetingMapped.push({
+			name: 'urlkw',
+			value: adTargeting.customParams.urlkw as string[],
```
chrislomaxjones, line 114: It'd be good to avoid the type assertions here since we know what the types are. One option is to update the `CustomParams` type to include more information about the types for the specific keys we know about. Something like:

```typescript
type CustomParams = {
	sens: 't' | 'f';
	urlkw: string[];
	[key: string]: string | string[] | number | number[] | boolean | boolean[];
};
```

So values for `sens` and `urlkw` will have very specific types and any others will have the much looser definition, which should be fine for now. This would avoid the need to call `String(...)` and `as string[]`, but it may require some `as const` assertions in `buildAdTargeting`.

Either way, I think we'll be able to import much more specific types from commercial-core soon!

dotcom-rendering/src/amp/components/Ad.tsx
```diff
@@ -93,13 +94,27 @@ export const Ad = ({
...
+	adTargeting,
 }: AdProps) => {
 	const adSizes = isSticky ? stickySizes : inlineSizes;
 	// Set Primary ad size as first element (should be the largest)
 	const [{ width, height }] = adSizes;
 	// Secondary ad sizes
 	const multiSizes = adSizes.map((e) => `${e.width}x${e.height}`).join(',');
 
+	const adTargetingMapped: AdTargetParam[] = [];
+	if (adTargeting.disableAds !== true) {
```
chrislomaxjones, line 106:
```suggestion
	if (!adTargeting.disableAds) {
```

chrislomaxjones (commented):
This is great, had a couple of suggestions below!

dotcom-rendering/src/amp/components/Ad.tsx
simonbyford: Nice one, thanks for the explanation 👍

dotcom-rendering/src/amp/components/Ad.tsx
simonbyford: Of course

dotcom-rendering/src/amp/components/Ad.tsx
simonbyford: Yeah this is better!

dotcom-rendering/src/amp/components/Ad.tsx
```diff
@@ -28,6 +28,24 @@ const amazonConfig = {
 	aps: { PUB_ID: '3722', PARAMS: { amp: '1' } },
 };
 
+const mapAdTargeting = (adTargeting: AdTargeting): AdTargetParam[] => {
+	const adTargetingMapped: AdTargetParam[] = [];
```
lucymonie, line 32: This is totally a nitpick, but maybe this should be `let` rather than `const` since it's going to have stuff pushed into it?

dotcom-rendering/src/lib/ad-targeting.ts
lucymonie: Maybe we could add a card to export it, and a TODO here to import it and remove the duplicate?

lucymonie (approved):
This looks good, but I am not familiar enough with the code to be 100% on this 😄 

One thing that might help other newbies would be to annotate the code where there are impenetrable params like `sens: 't' | 'f'`. Having a note to explain what this is might reduce confusion.

dotcom-rendering/src/lib/ad-targeting.ts
simonbyford: Good idea, I'll add!

dotcom-rendering/src/amp/components/Ad.tsx
simonbyford: It's not reassigned so I _think_ `const` is appropriate here!

PR #3847, opened 2022-01-20
Hf interactive figure padding

HarryFischer:
## What does this change?
Removes the css for Figure elements in interactive articles

## Why?
Currently the Figure element expects the article grid to add a left column. Interactive articles don't have a left column by default and therefore the margin-left and margin-right are not needed.
Css to position these elements can now be included via a simple enhancer.scss file within the interactive-atom-template-2019 repo

### Before
Screenshot 2022-01-20 at 15 43 24
Screenshot 2022-01-20 at 15 43 33

### After
Screenshot 2022-01-20 at 15 38 05
Screenshot 2022-01-20 at 15 38 12


oliverlloyd (approved):
👍 👍

PR #3848, opened 2022-01-20
Report DCR dev build stats

OllysCoding:
## What does this change?

This PR adds a new custom webpack plugin to DCR dev builds which reports build stats to google big-query, which we'll then be able to use to create a dashboard with.

We send up the following properties
- `project`
- `team`
- `buildName` - `server`, `client` or `legacy-client`
- `buildCount` - how many builds have we done in this 'session' (e.g 1 is the first build, 2, 3, 4 would be HMR). This allows us to track not only the difference between initial & hmr, but also how many rebuilds people are seeing before the dev server runs out of memory
- `gitHash` - current (short) git hash, fetched when the dev server is started
- `gitBranch` - current git branch, fetched when the dev server is started
- `sessionId` - an ID consistent across all 3 builds & all rebuilds which we can use to track a single session of running `make dev`. Useful to link together different datapoints & track how many sessions we are getting
- `cpus` - Number of CPU cores on the machine, as provided by `os.cpus()` in node
- `memoryKb` - Amount of (total - not available) RAM on the machine

& the following metrics

- `buildTime` - the time (in ms) taken for the builds being reported
- `memoryUsageKb` - the amount of memory the node process is using in total (rss)

### What does it look like in Big Query?

image


dotcom-rendering/.eslintrc.js
```diff
@@ -97,4 +97,14 @@ module.exports = {
...
+	overrides: [
+		{
+			files: ['**/**.js'],
+			rules: {
+				'@typescript-eslint/no-var-requires': 'off',
+				'@typescript-eslint/no-unsafe-member-access': 'off',
+				'@typescript-eslint/no-misused-promises': 'off',
+			},
+		},
+	],
```
OllysCoding, line 109: I added these because it was getting too annoying to work on vanilla JS files. I have no idea why typescript-eslint was trying to apply these rules in the first place

dotcom-rendering/scripts/webpack/gu-stats-report-plugin.js
```diff
@@ -0,0 +1,158 @@
...
+							value: 'dotcom',
+						},
+						{
+							name: 'machineUsername',
+							value: this.machineUsername || 'unknown',
+						},
+						{
+							name: 'machineHostname',
+							value: this.machineHostname,
+						},
```
OllysCoding, line 90: Should we be reporting anonymously? These are cool and allow us (among other things) to know how many unique 'uses' DCR has, as well as specifically find and work to diagnose anyone who might be having a bad experience.

However whether it's worth collecting 'identifying' information to achieve this goal should definitely be discussed

dotcom-rendering/scripts/webpack/gu-stats-report-plugin.js
```diff
@@ -0,0 +1,158 @@
...
+		return true;
+	}
+
+	fetchMachineUsername() {
+		try {
+			// if 'username' is undefied os.userInfo() will throw an error
+			// https://nodejs.org/api/os.html#osuserinfooptions
+			this.machineUsername = os.userInfo().username;
+		} catch {
+			console.error('[gu-stats-report] Failed to get machine username');
```
jamesgorrie, line 35: Do we want to return every time? I can't find a definitive way of error management errors in webpack plugins... 

Interesting as `const val = console.error('bloop') // undefined` .

dotcom-rendering/scripts/webpack/gu-stats-report-plugin.js
```diff
@@ -0,0 +1,158 @@
...
+			if (!this.isValidConfig)
+				return console.log(
+					'[gu-stats-report] Unable to report stats - invalid config',
+				);
+
+			const URL = 'https://logs.code.dev-guardianapis.com/log';
+			fetch(URL, {
+				method: 'POST',
+				body: JSON.stringify({
+					label: 'buildstats',
```
jamesgorrie, line 73: Ping @sndrs FYI there was a decision to have this as a generic `label` so that we could potentially have a generic dashboard for other CS build systems.

dotcom-rendering/scripts/webpack/gu-stats-report-plugin.js
```diff
@@ -0,0 +1,158 @@
...
+const os = require('os');
+const { exec } = require('child_process');
+
+class GuStatsReportPlugin {
+	constructor(config) {
+		this.buildName = config?.buildName;
+		this.project = config?.project;
+		this.team = config?.team;
+		this.buildCount = 0;
+		this.gitBranch = null;
```
jamesgorrie, line 11: Any reason to be using `null` and `undefined` (from optional chaining above)?

[Interesting article on this](https://writingjavascript.com/why-you-should-always-use-undefined-and-never-null) - besides the argument of which, I find it best to at least stick to one.

jamesgorrie, line 11: `this.gitBranch = this.fetchGitBranch()` could be easier to test as the fetch function would return the value for testing.

dotcom-rendering/scripts/webpack/gu-stats-report-plugin.js
jamesgorrie: Worth checking in with InfoSec what we would need to do this? Could also be worth checking with people who already use it how they feel.

If they're good - potentially always show the disclaimer and a way of opting out?

jamesgorrie: Might be worth thinking of doing it anonymously first as we probably want to support that and add identification as a second PR?

jamesgorrie (approved):
I'd say this look great so far - for expedience, and to make sure we give the right amount of thought to collecting identifiable info, should we add that separately?

dotcom-rendering/scripts/webpack/gu-stats-report-plugin.js
OllysCoding: The reason I couldn't do this (though I would have liked to) is because the calls for git info are asynchronous, so they cannot be put in a class constructor.

Defining these as either `null`, `undefined` or a default value in a constructor therefor makes it easier to read what properties will be set in the class even if they can't be assigned their value in the constructor (I forget if you can assign properties to a call in JS that haven't be initialised in the constructor, but I still think it's good practice).

There is then an inherent contract that these calls will complete before the initial webpack build, while I wouldn't expect them to take much more than a few hundred milliseconds at the most, it would be worth keeping in mind.

The alternative is we wrap these functions in Promises, and call them as part of the stats report function. However with up to 3 builds running simultaneously, it seems unnecessary to call them for every build & re-build.

dotcom-rendering/scripts/webpack/gu-stats-report-plugin.js
OllysCoding: I've updated this PR for the time being to use uuid (v4 - we could perhaps look at using v1 which is based on time & mac address) to generate 'session ids'. These session IDs are consistent across all 3 builds so do give us the benefit of 'usage' statistics ~ and we can later look at using anonymous ID's (is MAC address a good candidate for this?) or other options.

dotcom-rendering/scripts/webpack/gu-stats-report-plugin.js
```diff
@@ -0,0 +1,144 @@
...
+	constructor(config) {
+		this.buildName = config?.buildName;
+		this.project = config?.project;
+		this.team = config?.team;
+		this.sessionId = config?.sessionId;
+		this.buildCount = 0;
+		this.gitBranch = undefined;
+		this.gitHash = undefined;
+
+		this.fetchGitBranch();
```
jamesgorrie, line 15: Could be worth leaving a note similar to [this](https://github.com/guardian/dotcom-rendering/pull/3848#discussion_r790571082) - although maybe it's obvious enough not to.

jamesgorrie (approved):
Love it.

dotcom-rendering/scripts/webpack/gu-stats-report-plugin.js
```diff
@@ -0,0 +1,144 @@
...
+
+		if (config.displayDisclaimer)
+			console.log(
+				'[gu-stats-report] This project reports build information, stats & basic machine information for internal use only. We will use this information to help us improve developer experience for this project.',
+			);
+	}
+
+	isValidConfig() {
+		if (!this.buildName || !this.project || !this.team) return false;
+		return true;
```
oliverlloyd, line 29:
```suggestion
		if (!this.buildName || !this.project || !this.team) return false;
		return true;
```
```suggestion
		return this.buildName && this.project && this.team;
```

Nit

oliverlloyd (approved):
I love how thorough this PR is

@sndrs @SiAdcock - this looks like a great candidate for a more centralised location. I'm not sure how common webpack is or will be but the general idea of measuring things like this around the developer experience is really great

dotcom-rendering/scripts/webpack/gu-stats-report-plugin.js
```diff
@@ -0,0 +1,146 @@
...
+		const onDone = (stats) => {
+			// Increment the buildCount
+			this.buildCount += 1;
+
+			if (!this.isValidConfig)
+				return console.log(
+					'[gu-stats-report] Unable to report stats - invalid config',
+				);
+
+			const URL = 'https://logs.code.dev-guardianapis.com/log';
```
mxdvl, line 61: You probably want to use the PROD endpoint here.

mxdvl (commented):
This is amazing, sorry I did not see it any earlier!

dotcom-rendering/scripts/webpack/browser.js
```diff
@@ -59,6 +60,13 @@ module.exports = ({ isLegacyJS }) => ({
 	plugins: [
 		DEV && new webpack.HotModuleReplacementPlugin(),
 		DEV && friendlyErrorsWebpackPlugin(),
+		DEV &&
+			new GuStatsReportPlugin({
+				buildName: isLegacyJS ? 'client' : 'legacy-client',
```
arelra, line 65: Hello @OllysCoding sorry for chiming in late. Is this ternary the right way around?

PR #3853, opened 2022-01-21
Live 3586 remove redundant AWS resources from original cloudformation template

waisingyiu:
## What does this change?

This PR is to remove the old infrastructure from the original cloudformation template.  We now have a set of new AWS cloudformation defined using CDK.

This PR also adds the configuration about autoscaling policy and restores the TTL of DNS record from 1 minute to 1 hour.

## Why?

We started CDK migration on apps-rendering and at this moment we have both the old and new cloudformations running in parallel, with DNS record pointing to the new one.  After more than a week of running the service in the new cloudformation, we create this PR to tear down the old infrastructure.

### Before
We have both the old and the new infrastructure running - 
1) the old one was defined using cloudformation template
2) the new one was defined using CDK.  It is currently used to serve requests as DNS record maps the domain name to its load balancer

### After
We have the new infrastructure only - which was defined using CDK.


groakland (approved):
LGTM

PR #3858, opened 2022-01-24
`defer` > `deferuntil`

oliverlloyd:
## What does this change?
This fixes an issue where we were reading the value for the `deferuntil` attribute incorrectly

## Why?
By not reading the right value we are not deferring hydration

### DIFF
--- a/dotcom-rendering/cypress/integration/parallel-3/article.interactivity.spec.js
+++ b/dotcom-rendering/cypress/integration/parallel-3/article.interactivity.spec.js
@@ -52,9 +52,8 @@ describe('Interactivity', function () {
 			it('should change the list of most viewed items when a tab is clicked', function () {
 				cy.visit(`/Article?url=${articleUrl}`);
 				cy.contains('Lifestyle');
-				cy.get('[data-component="most-popular"]').scrollIntoView({
-					offset: { top: 150 },
-				});
+				// Scroll to bottom to trigger hydration
+				cy.scrollTo('bottom', { duration: 300 });
 				cy.wait('@getMostReadGeo');
 				cy.wait('@getMostRead');
 				cy.get('[data-cy=tab-body-0]').should('be.visible');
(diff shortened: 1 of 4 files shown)


dotcom-rendering/cypress/integration/parallel-3/article.interactivity.spec.js
```diff
@@ -52,9 +52,8 @@ describe('Interactivity', function () {
 			it('should change the list of most viewed items when a tab is clicked', function () {
 				cy.visit(`/Article?url=${articleUrl}`);
 				cy.contains('Lifestyle');
-				cy.get('[data-component="most-popular"]').scrollIntoView({
-					offset: { top: 150 },
-				});
+				// Scroll to bottom to trigger hydration
+				cy.scrollTo('bottom', { duration: 300 });
```
OllysCoding, line 56: Was this change required due to tests failing, or was the goal just to generally improve the consistency of this test?

OllysCoding (approved):
+1 Good catch!

dotcom-rendering/cypress/integration/parallel-3/article.interactivity.spec.js
oliverlloyd: I added these changes because the tests failed without them now that we're actually deferring hydration

dotcom-rendering/src/web/components/MostViewed/MostViewedFooter/MostViewedFooterData.tsx
```diff
@@ -72,6 +72,7 @@ export const MostViewedFooterData = ({
 				css={css`
 					width: 100%;
 				`}
+				data-cy="mostviewed-footer"
```
oliverlloyd, line 75: I've added this test attribute so that I can assert that this content is _not_ in the dom prior to scrolling the page

jamesgorrie: Do we use `getByText` with the heading text for selectors like this if we wanted to follow the ["How would users find it?"](https://testing-library.com/docs/queries/about#priority) methodology.

oliverlloyd: For Cypress tests we follow the [recommendations](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements) and use `data-cy` selectors.

Swings and roundabouts but I do think it's nice to separate the hooks you're using for selection away from how you implement the product you're testing. But it's true that selecting by text can feel and read better, when it works.

PR #3864, opened 2022-01-24
Update gu-stats-report-plugin to use prod API

OllysCoding:
## What does this change?

Update gu-stats-report-plugin to use prod API - I accidentally left it as the CODE endpoint


jamesgorrie (approved):
fab

PR #3865, opened 2022-01-24
More tests checking hydration

oliverlloyd:
This PR is branched off of https://github.com/guardian/dotcom-rendering/pull/3858

## What does this change?
This Pr adds some additional Cypress test to verify that the page is being hydrated as we expect

## Why?
We're working on refactoring hydration and want to ensure that we have integration tests that capture the behaviour we need so that we don't introduce regression


dotcom-rendering/cypress/integration/parallel-3/article.interactivity.spec.js
```diff
@@ -36,27 +36,74 @@ describe('Interactivity', function () {
...
+			cy.get('[data-component=discussion]').should('not.exist');
+			// Click the comment count
+			cy.get('[data-cy=comment-counts]').click();
+			cy.get('[data-component=discussion]').should('exist');
+		});
+		it('loads the discussion immediately when you use a url ending in #comments', function () {
+			cy.visit(
+				`/Article?url=https://www.theguardian.com/commentisfree/2022/jan/20/uk-government-yemen-war-saudi-arabia-westminster#comments`,
+			);
+			cy.get('[data-component=discussion]').should('exist');
```
jamesgorrie, line 54: Just trying to wrap my head around the selection. `[data-component=discussion]` is a child of `gu-island[name=DiscussionContainer]`, and also only loaded in async?

dotcom-rendering/cypress/integration/parallel-3/article.interactivity.spec.js
```diff
@@ -36,27 +36,74 @@ describe('Interactivity', function () {
...
+				`/Article?url=https://www.theguardian.com/commentisfree/2022/jan/20/uk-government-yemen-war-saudi-arabia-westminster#comment-154433663`,
+			);
+			cy.get('gu-island[name=DiscussionContainer]').should(
+				'have.attr',
+				'data-gu-hydrated',
+				'true',
+			);
+			cy.get('[data-component=discussion]').should('exist');
+		});
+		it('loads the most viwed list onlt after starting to scroll the page', function () {
```
jamesgorrie, line 69:
```suggestion
		it('loads the most viwed list only after starting to scroll the page', function () {
```

dotcom-rendering/cypress/integration/parallel-3/article.interactivity.spec.js
jamesgorrie: When do we use `data-component`?

dotcom-rendering/cypress/integration/parallel-3/article.interactivity.spec.js
```diff
@@ -36,27 +36,74 @@ describe('Interactivity', function () {
...
+			// The permalink feature is not currently working but once it does we want this test ready to go
+			cy.visit(
+				`/Article?url=https://www.theguardian.com/commentisfree/2022/jan/20/uk-government-yemen-war-saudi-arabia-westminster#comment-154433663`,
+			);
+			cy.get('gu-island[name=DiscussionContainer]').should(
+				'have.attr',
+				'data-gu-hydrated',
+				'true',
+			);
+			cy.get('[data-component=discussion]').should('exist');
```
jamesgorrie, line 67: would `.should('be.visible')` be more user centric?

dotcom-rendering/cypress/integration/parallel-4/signedin.spec.js
```diff
@@ -40,6 +40,17 @@ describe('Signed in readers', function () {
 
 	it('should not display signed in texts when users are not signed in', function () {
 		cy.visit(`Article?url=${articleUrl}`);
+		cy.scrollTo('bottom', { duration: 300 });
+		// We need this second call to fix flakiness where content loads in pushing the page
+		// down and preventing the scroll request to actually reach the bottom. We will fix
+		// this later when we've defined fixed heights for these containers, preventing CLS
+		cy.scrollTo('bottom', { duration: 300 });
+		// Wait for hydration
+		cy.get('gu-island[name=DiscussionContainer]').should(
```
jamesgorrie, line 49: Why do we check hydration in this test?

dotcom-rendering/src/web/browser/islands/init.ts
```diff
@@ -28,8 +28,8 @@ const init = () => {
 			if (!name) return;
 			log('dotcom', `Hydrating ${name}`);
 
-			const defer = element.getAttribute('defer');
-			switch (defer) {
+			const deferUntil = element.getAttribute('deferuntil');
+			switch (deferUntil) {
```
jamesgorrie, line 32: part of another PR.

dotcom-rendering/cypress/integration/parallel-3/article.interactivity.spec.js
oliverlloyd: Yeah, you're right, this better represents what we want to assert on

dotcom-rendering/cypress/integration/parallel-3/article.interactivity.spec.js
oliverlloyd: So we use `data-component` for Ophan. We shouldn't be using that here as it's coupling the tests with the tracking solution, I'll update this

dotcom-rendering/cypress/integration/parallel-4/signedin.spec.js
oliverlloyd: Because when I say 'hydration' here what I really mean is wait for the Island to load the content onto the page. So it's not actually SSR + hydration, it's a pure client side load. I'll update the comment to better reflect this

oliverlloyd: Aside, maybe `data-gu-hydrated` should be `data-gu-ready`?

ioannakok (approved):
👍

dotcom-rendering/cypress/integration/parallel-4/signedin.spec.js
oliverlloyd: This change to `data-gu-ready` has now been done in #3872

PR #3869, opened 2022-01-24
`Liveness`

oliverlloyd:
## What does this change?
This PR implements liveness for blogs in DCR.

In a [previous document](https://docs.google.com/document/d/1MWpk8ddLKp7WSa57MOzPpGhvNBDQMqzknlxffwhzvLY/edit#heading=h.rldsadv154bo) we discussed the general strategy for how we make liveblogs live. This PR takes that strategy and [proposes an implementation](https://github.com/guardian/dotcom-rendering/pull/3869) based upon it.

In particular, we're proposing an approach where we use a react component `Liveness` to connect and execute the code we want to use to keep the content fresh, but it does not use react to maintain the list of blocks.

### What is included?
- Actually adding the liveness code to the page
- Polling the api (Using [SWR](https://swr.vercel.app/))
- React state [`noOfNewPosts`] - Count of new updates pending
- React state [`showToast`] - When to show the Toast button
- React state [`latestBlockId`] - Which block id to use in the api request
- `IntersectionObserver` - For tracking reader scroll position in the blog
- Inserting html onto the page
- Keeping the document title up to date (the count of updates shown in the tab)

### What is not included?
- Revealing and animating the new posts
- Hydrating new content
- The styling for the `Toast` component

## How can we use react but not use react?
A react component is just a function that returns html. But if you tell the function to return `null` then it's just a function. If we insert `Liveness` in the `LiveLayout` file and wrap it in an `Island` then this code will be executed at page load.

With this approach we can choose to use react for some things, like rendering the `Toast`, but use plain javascript for inserting and hydrating the html. **It also allows us to use SWR, [DCR's react data fetching library](https://github.com/guardian/dotcom-rendering/blob/53e002b3aa11153f57eb99ad556c8164a83103d5/dotcom-rendering/src/web/lib/useApi.tsx), rather than having to introduce or create a second fetching lib to the page**.

### Pros to hybrid
- We can use [SWR](https://swr.vercel.app/) - not having multiple solutions for fetching data is a huge win
- We can use react state to show/hide the `Toast` and keep track of pending updates

### Cons to hybrid
- Using a react component as a vehicle to instantiate js functions is non standard

## What alternatives are there to this approach?
Instead of using a react component to trigger this code we could just use a script tag. This is a simple, established pattern for running javascript on the page.

### Pros to purity
- Simple, well understood
- React does not exist so the code follows traditional flows

### Cons to purity
- React (and DCR) was built for a reason, plain javascript does not scale well
- Code is un-scoped, you can do anything!
- We would need to either find a plain javascript lib for fetching data (there are no obvious options) or roll our own (lots of work!)
- We would no longer have a centralised data fetching solution in DCR. Instead we would have multiple ways that we fetch data, needing to maintain both going forwards


dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -0,0 +1,169 @@
...
+		pageId,
+	)}.json?lastUpdate=${latestBlockId}&isLivePage=true&filterKeyEvents=false`;
+
+	// useApi returns { data, loading, error } but we're not using them here
+	useApi<{
+		numNewBlocks: number;
+		html: string;
+	}>(url, {
+		refreshInterval: 10000,
+		refreshWhenHidden: true,
```
DavidLawes, line 118: Is this saying we're polling even when the page isn't in focus? Will we poll with the same interval when not in focus?

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -382,6 +383,15 @@ export const LiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
 	const { branding } = CAPI.commercialProperties[CAPI.editionId];
 	return (
 		<>
+			{format.design === ArticleDesign.LiveBlog && (
```
DavidLawes, line 386: This structure of wrapping Liveness within an Island feels very 'natural' and in-keeping with the model for components requiring hydration. Given you're familiar with dcr's Island pattern, I think it's obvious what we're achieving by doing this.

Could you help me understand: the `deferUntil="idle"` implies we will only do the hydration when the browser is [idle](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback), meaning we're not consuming any compute until the page has fully loaded?

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: Yes, `refreshWhenHidden` allows us to update the `document.title` count on the tab so readers know there's new stuff to see. I assume it polls at the same interval, that would make sense from reading the api.

dotcom-rendering/src/web/layouts/LiveLayout.tsx
oliverlloyd: I think the actual implementation of when idle gets fired might vary browser to browser but in general, yes, it offers a way to defer code execution to a 'lower priority' position. We only really need this code to run after a few seconds so it's much less important than code related to the main rendering thread.

From a technical point of view, the `deferUntil` prop here is wrapping the code in a listener which only gets fired once the browser declares that it is 'idle'

dotcom-rendering/src/web/components/Liveness.importable.tsx
DavidLawes: Thanks :) I was checking in the [frontend equivalent](https://github.com/guardian/frontend/blob/main/static/src/javascripts/projects/common/modules/ui/autoupdate.js#L43), I wondered if they had an increased delay for when the page is hidden? It may have been introduced to reduce load on the api, or improve some kind of device performance metric, or keep user network costs as low as possible if on mobile - it might be something that we consider too? 

PS - I hope I'm looking in the correct 'frontend equivalent', please correct me if i'm wrong :)

oliverlloyd: There's pros and cons to a backoff strategy. Arguably, we are much more focussed on making our content 'live' now than we were back when that code was written. Backing off might reduce background noise but it also means the content stops being kept up to date which sort of breaks the contract we have with the reader (that we will keep them up to date)?

AshCorr (commented):
Love the extensive commenting!

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -0,0 +1,186 @@
...
+function getKey(
+	pageId: string,
+	ajaxUrl: string,
+	filterKeyEvents: boolean,
+	latestBlockId: string,
+): string {
+	// Construct the url to poll
+	const url = new URL(`${pageId}.json`, ajaxUrl);
+	url.searchParams.set('lastUpdate', latestBlockId);
+	url.searchParams.set('isLivePage', 'true');
```
alinaboghiu, line 81: Not sure we need to set this at all - but we should only poll when the liveblog is live (rather than dead)

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -0,0 +1,186 @@
...
+			mostRecentBlockId: string;
+		}) => {
+			if (data && data.numNewBlocks && data.numNewBlocks > 0) {
+				// Always insert the new blocks in the dom (but hidden)
+				insertNewBlocks(data.html);
+				hydrateBlocks();
+
+				if (topOfBlogVisible()) {
+					revealNewBlocks();
+					setNoOfNewPosts(0);
```
alinaboghiu, line 133: if we only set the number of new posts to 0 when we reveal them, won't we end up inserting them all the time? Or is inserting idempotent?

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -0,0 +1,186 @@
...
+				rootMargin: '100px',
+			},
+		);
+
+		observer.observe(topOfBlog);
+	}
+
+	const handleToastClick = () => {
+		setShowToast(false);
+		if (topOfBlog) topOfBlog.scrollIntoView();
```
alinaboghiu, line 176:
```suggestion
		if (!topOfBlog) topOfBlog.scrollIntoView();
```

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * @returns boolean
+ */
+function topOfBlogVisible(): boolean {
+	return !!topOfBlog && !!topOfBlog.classList.contains('in-viewport');
+}
+
+export const Liveness = ({ pageId, webTitle, ajaxUrl }: Props) => {
+	const [showToast, setShowToast] = useState(false);
+	const [noOfNewPosts, setNoOfNewPosts] = useState(0);
```
alinaboghiu, line 101: noOfNewPosts vs numNewBlocks is a bit confusing. Am I correct in thinking that:
noOfNewPosts = cumulative count of new blocks that are not currently visible on the page 
numNewBlocks = latest number of new blocks retrieved by the latest polling request

If so maybe could they be named:
noOfNewPosts -> cumulativeNewBlocksCount
numNewBlocks -> latestNewBlocksCount
Or something like this?

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: I think we can simply not send this query param

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -0,0 +1,186 @@
...
+		// allows us to avoid the problems of imperative code being executed multiple times
+		// inside react's declarative structure (things get re-rendered when any state changes)
+		onSuccess: (data: {
+			numNewBlocks: number;
+			html: string;
+			mostRecentBlockId: string;
+		}) => {
+			if (data && data.numNewBlocks && data.numNewBlocks > 0) {
+				// Always insert the new blocks in the dom (but hidden)
+				insertNewBlocks(data.html);
```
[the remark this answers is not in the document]
oliverlloyd: As discussed, not needed

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: Thanks

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: How about:

`noOfNewPosts ` > `numHiddenBlocks`
Keep `numNewBlocks` as it is

?

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: I'm wrong. We do need to send this param. If [we don't](https://api.nextgen.guardianapps.co.uk/world/live/2022/feb/01/covid-live-news-denmark-lifts-all-covid-restrictions-despite-record-cases-liz-truss-tests-positive.json?lastUpdate=block-61f98fb28f08ed69ff280052&filterKeyEvents=false) then the response no longer contains the html.

This is how Frontend works. Arguably, DCR does not have to follow this pattern but it might be better to keep things consistent between the two apis.

oliverlloyd: To your other point regarding when we poll. The code here in this file is only loaded on live blogs so we would never start polling unless the blog is live on page load.

This does raise the question of how do we _stop_ polling? Frontend uses a backoff strategy but the code here is set to poll for as long as the tab is open. Maybe we can include the blog status in the response and use that to stop polling?

dotcom-rendering/src/web/components/Liveness.importable.tsx
JamieB-gu: I believe some browsers will suspend (e.g. Safari on iOS) or slow down (e.g. Chrome) JS execution in background tabs. So we may not be polling regularly (or at all) in some scenarios.

oliverlloyd: Oh, interesting! That actually sounds ideal. I much prefer the idea of using platform configuration for this kind of thing. It will be interesting to test this but my expectation / hope is that the original polling routine is restored on focus. SWR automatically refetches on focus in any case.

oliverlloyd: This actually tallys with what I've seen from my own testing. I've had a tab open and polling for a couple of days now and I'm not seeing wild volumes of network requests in the console but if I open it (restore focus) and watch it I see requests being made!

Of course, this is all a bit Spooky Effect From a Distance so who knows.

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -0,0 +1,186 @@
...
+	}>(getKey(pageId, ajaxUrl, filterKeyEvents, latestBlockId), {
+		refreshInterval: 15000,
+		refreshWhenHidden: true,
+		// onSuccess runs (once) after every successful api call. This is useful because it
+		// allows us to avoid the problems of imperative code being executed multiple times
+		// inside react's declarative structure (things get re-rendered when any state changes)
+		onSuccess: (data: {
+			numNewBlocks: number;
+			html: string;
+			mostRecentBlockId: string;
```
JamieB-gu, line 124: This may be an issue with the type definition of `useApi` rather than anything to do with this PR, but why is this object different to the type parameter passed to `useApi`? It's got an additional `mostRecentBlockId` field 🤔 

Should `SWRConfiguration` in the `useApi.tsx` module take `T` as a type parameter?

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -0,0 +1,186 @@
...
+					setNoOfNewPosts(0);
+				} else {
+					setShowToast(true);
+					// Increment the count of new posts
+					setNoOfNewPosts(noOfNewPosts + data.numNewBlocks);
+				}
+
+				// Update the block id we use for polling
+				if (data.mostRecentBlockId)
+					setLatestBlockId(data.mostRecentBlockId);
```
JamieB-gu, line 142: Do you have any thoughts on the use of the ESLint "curly" rule? https://eslint.org/docs/rules/curly

```suggestion
				if (data.mostRecentBlockId) {
					setLatestBlockId(data.mostRecentBlockId);
				}
```

dotcom-rendering/src/web/components/Liveness.importable.tsx
JamieB-gu: I think because `topOfBlog` is a mix of `Element`, `null` _and_ `false`, that'll narrow it to `null | false` within the `if` statement, which won't allow you to call `scrollIntoView`.

I've dropped a suggestion in above for tweaking the implementation of `topOfBlog`. If you go with that I think you can make this:

```suggestion
		topOfBlog?.scrollIntoView();
```

Alternatively, if you only want to call this if `topOfBlog` is out of the viewport atm:

```suggestion
		if (!topOfBlogVisible()) { topOfBlog?.scrollIntoView(); }
```

JamieB-gu (commented):
Some really clever stuff in here 🚀

A few comments/questions.

dotcom-rendering/src/web/components/Liveness.importable.tsx
JamieB-gu: I think if browsers _are_ suspending JS they should be able to re-engage it where it was left off when focus returns, otherwise they'd break all sorts of things 😬. I imagine variations of Low Power Mode also factor into this.

Having read through the full PR now, I was thinking this sort of behaviour may affect your ability to show updates in the tab title - but that's probably a small feature and it may not be impacted on all browsers?

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * This allows us to make decisions in javascript based on if the reader
+ * has the top of the blog in view or not
+ *
+ * @returns boolean
+ */
+function topOfBlogVisible(): boolean {
+	return !!topOfBlog && !!topOfBlog.classList.contains('in-viewport');
+}
```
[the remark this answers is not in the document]
oliverlloyd: This is much nicer, thanks!

I've made the changes you suggested locally to avoid prettier errors

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: I think in this example using the braces like this is much better 👍 .

The lint rule though would affect quite a lot of code in DCR where adding braces would - in my opinion - interrupt the flow. Code like

```
if (someError) return <Error />
if (loading) return <Loading />

return (
  ...
)
```

would get messy. I think it's nice that, as it stands, we have the choice to add them if we think they work but they are not enforced unless you have multiple lines.

Now that we have Prettier, these days I'm generally happy to sheep-like follow whatever lint rules a project has. I just write code, save, and let Prettier do it's thing. I really like how this tool removes the need for all those styling discussions we used to have before it existed!

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: Implemented locally 👍

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: In comparison to Frontend, this DCR implementation keeps the tab count more up to date. If a browser does backoff the polling (I have seen this happening) then I think this is a good thing. I can also confirm that the polling resumes when you restore focus after a long time in the background.

We probably want to consider adding a field to the api response with the blog status (`isLive: boolean`) so that we can terminate polling when the blog is dead.

dotcom-rendering/src/web/components/ArticleBody.tsx
```diff
@@ -89,24 +89,29 @@ export const ArticleBody = ({
...
-					pageId={pageId}
-					webTitle={webTitle}
-					ajaxUrl={ajaxUrl}
-				/>
-			</div>
+			<>
+				{format.design === ArticleDesign.LiveBlog && (
+					<span data-gu-marker="top-of-blog" />
+				)}
+				<div
```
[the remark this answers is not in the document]
oliverlloyd: We [refactored the landmark elements](https://github.com/guardian/dotcom-rendering/pull/3599) on an article earlier to try to make the high level structure of the page more semantic - which meant putting `main` higher up than here - but we still want to place that you get scrolled to to be the top of the text.

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -0,0 +1,186 @@
...
+ * @returns string
+ */
+function getKey(
+	pageId: string,
+	ajaxUrl: string,
+	filterKeyEvents: boolean,
+	latestBlockId: string,
+): string {
+	// Construct the url to poll
+	const url = new URL(`${pageId}.json`, ajaxUrl);
```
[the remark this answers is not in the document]
oliverlloyd: My guess here is this is unlikely (maybe impossible?) to ever happen as both pageId and ajaxUrl are required as strings but I've added extra code to handle this just in case

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -0,0 +1,186 @@
...
+	const [showToast, setShowToast] = useState(false);
+	const [noOfNewPosts, setNoOfNewPosts] = useState(0);
+	const [latestBlockId, setLatestBlockId] = useState(
+		// By default we use the first (latest) block id on the page
+		document.querySelector('#maincontent :first-child')?.id || '',
+	);
+
+	// Read current url to get filterKeyEvents
+	const params = new URLSearchParams(window.location.search);
+	const filterKeyEvents = !!params.get('filterKeyEvents');
```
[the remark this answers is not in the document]
oliverlloyd: You are spot on with this! I lifted this code from another branch without checking it (doh) but it's since been fixed there and should be here too.

However, I know that there's been some work to add this property to the CAPI object which should remove the need to read the url completely (better!), I'm going to refactor this code with that in mind.

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: Good spot. They're actually typing two different things. We actually don't need that type which is being passed to `useApi` because we're not using it (we don't reference the return values from `useApi` anywhere)

oliverlloyd: I made a change to remove the unused type

alinaboghiu (approved):
In the interest of not having a huge PR that's open for months we can probably merge this as is and improve as we learn along

PR #3871, opened 2022-01-25
Fix the ASG value for PROD

waisingyiu:
## What does this change?

In the previous PR (#3853), the wrong set of ASG parameters were applied to PROD.  This PR is to fix the bug by using the correct stage-specific parameters on ASG size.

## Why?

A bug in the previous PR (#3853)

### Before

The `mobile` stack on PROD has 1 for minimum instance and 2 for maximum instance.

### After

The `mobile` stack on PROD has 3 for minimum instance and 12 for maximum instance.


groakland (approved):
LGTM

frankie297 (approved):
LGTM 👍

PR #3872, opened 2022-01-25
Only Hydrate Once

oliverlloyd:
## What does this change?
This adds a check to `doHydration` to exit early if the element being passed in has already been hydrated

## Why?
We have some use cases coming up that will need this:

1) Discussion permalinks. We want to be able to force hydrate components early (not when they are deferred to) but if we do this we want to make sure the deferred call will be blocked

2) Liveblogs. When inserting new blocks we want to run a manual hydration command. This command _could_ be us simply repeating the page wide hydration step but this only works if we have this guard in place.

### DIFF
--- a/dotcom-rendering/src/web/browser/islands/doHydration.ts
+++ b/dotcom-rendering/src/web/browser/islands/doHydration.ts
@@ -14,6 +14,11 @@ import { initPerf } from '../initPerf';
  * @param element The location on the DOM where the component to hydrate exists
  */
 export const doHydration = (name: string, data: any, element: HTMLElement) => {
+	// If this function has already been run for an element then don't try to
+	// run it a second time
+	const alreadyHydrated = element.dataset.guReady;
+	if (alreadyHydrated) return;
+
 	const { start, end } = initPerf(`hydrate-${name}`);
 	start();
 	import(
--- a/dotcom-rendering/src/web/browser/islands/doHydration.ts
+++ b/dotcom-rendering/src/web/browser/islands/doHydration.ts
@@ -22,7 +22,7 @@ export const doHydration = (name: string, data: any, element: HTMLElement) => {
 	)
 		.then((module) => {
 			hydrate(h(module[name], data), element);
-			element.setAttribute('data-gu-hydrated', 'true');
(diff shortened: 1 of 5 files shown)


ioannakok (approved):
+1

dotcom-rendering/src/web/browser/islands/doHydration.ts
```diff
@@ -14,6 +14,11 @@ import { initPerf } from '../initPerf';
  * @param element The location on the DOM where the component to hydrate exists
  */
 export const doHydration = (name: string, data: any, element: HTMLElement) => {
+	// If this function has already been run for an element then don't try to
+	// run it a second time
+	const alreadyHydrated = element.dataset.guReady;
```
DavidLawes, line 19: If this property returns a `boolean` how do we feel about naming convention `isAlreadyHydrated`?

DavidLawes (approved):
The guard we're putting in place makes sense.

Can you help me understand the naming of the data attribute a bit better please? We're adding the attribute after the element has been hydrated, so we're saying the element is `ready`?  What is the context for non-ssr'd code? (Sorry, I feel like I'm missing something because the old name `data-gu-hydrated` seems more descriptive to me...)

dotcom-rendering/src/web/browser/islands/doHydration.ts
oliverlloyd: Good question!

It's related to [this discussion](https://github.com/guardian/dotcom-rendering/pull/3865#discussion_r791032527) on another PR. We refer to the act of 'hydration' a lot and it's accurate inasmuch as we are running the _hydrate_ function here, always. But in truth there are two contexts

1) Some content was server side rendered and we're 'hydrating' it (making it interactive)
2) Nothing was server side rendered and we're simply 'rendering' content here

But we use the `React.hydrate` function in both cases because in truth this function will fallback to rendering the content if it can't hydrate it so it's semantically correct to call this hydration but perhaps confusing to say it _has been_ hydrated.

The purpose of the data attribute is so we can know when this function has completed to 'gu-ready` makes more sense

PR #3873, opened 2022-01-25
`initDiscussion` (to fix permalinks)

oliverlloyd:
This PR depends on #3872 

## What does this change?
This adds a new script to the page to force hydration of `DiscussionContainer` in the event that a permalink is present in the url

## Why?
If the reader is using a permalink then we want to directly scroll them to that comment on the page but this can only be done if the discussion is expanded. We needed a way to tell `DiscussionContainer` that we wanted to display the entire list of comments, not just the summary.

## Why is this a script?
We can't interact with the `Discussion` until it has been hydrated. So we can't put the logic to check the url inside it because that code will only be run when that component is hydrated and that only happens when the component is 'visible' in the dom.

We could have changed the `deferUntil` property to be 'idle' or  even 'immediate' but this comes at the cost of performance. So instead we're adding the code to check the url in it's own script and then manually forcing hydration only if required.

## Does it work reliably?
No, it really doesn't. There are two issues preventing permalinks working well:

1) There is other content being loaded on the page shifting things around. This creates a sort of content shift race condition which isn't fun

    The fix: Reserve space for all inline content being loaded onto the page

2) We can't be sure when the discussion has loaded or not

    The fix: Move the code to scroll the page down into `@guardian/discussion-rendering` so that we have a better idea for when to execute it

### `#comment-1234565`
https://user-images.githubusercontent.com/1336821/150977728-0fd7c9d6-5d63-4d95-b4a4-8575f4c95dc1.mov

### `#comments`
https://user-images.githubusercontent.com/1336821/150977743-6d69fb23-f992-4fb3-a5c4-01840ea54454.mov


oliverlloyd:
> My only query is regarding the problems identified in the PR description. I'm assuming we don't think these are blockers to a deploy? Are the suggested issues/fixes recorded anywhere for when we want to follow up on these?

They're not blockers, no. Permalinks don't work at all at the moment and this change makes things a lot better. The two issues listed will improve that but are not required for this PR to add benefit. Basically, this PR means we can now hydrate the code before we scroll, the next steps are to improve the experience when we do.

Point 1 is captured in Trello under

https://trello.com/c/B47pbLeQ/546-set-a-fixed-height-for-onwards-content
&
https://trello.com/c/I9Re6c9O/549-fix-the-height-of-unexpanded-comments

Point 2 isn't ticketed up yet

DavidLawes (approved):

PR #3875, opened 2022-01-25
Make map embed element an island

DavidLawes:
## What does this change?

This change makes the map embed element an island. In this PR we:
- include the click to view wrapper in the map embed element component
- include the importable component in renderElement.tsx (taking account of updated props)

There weren't any stories for the map embed so I added an extra one to ClickToView - I think that makes sense given all map embeds are wrapped by the click to view.

## Why?


oliverlloyd (approved):
I thought this one was a lot simpler but it turns out it was one of the more complicated ones but you knocked it out the park anyway!

PR #3876, opened 2022-01-25
Fix Follow Status

JamieB-gu:
## Why?

**TL;DR** Following journalists isn't currently working in AR; this fixes it.

Unfortunately the changes in #3794 inadvertently removed the `js-follow-status` class from the HTML. This is relied upon by the client-side code to look up the follow status component and apply event listeners/hydrate. This PR reinstates that class. 

**Note:** This is currently live on CODE if anyone would like to test it.

## Changes

- Reinstated the `js-follow-status` class, so the client-side code can look up the follow status component
- Moved the rendering of the contributor name into the `FollowStatus` component, so that it lines up with the "Follow"/"Following" text

### DIFF
--- a/apps-rendering/src/components/followStatus.tsx
+++ b/apps-rendering/src/components/followStatus.tsx
@@ -40,7 +40,9 @@ const FollowIcon: FC<IconProps> = ({ isFollowing }) => {
 const FollowStatus: FC<Props> = ({ isFollowing, contributorName }) => (
 	<>
 		<FollowIcon isFollowing={isFollowing} />
-		<span>{isFollowing ? 'Following' : 'Follow'} {contributorName}</span>
+		<span>
+			{isFollowing ? 'Following' : 'Follow'} {contributorName}
+		</span>
 	</>
 );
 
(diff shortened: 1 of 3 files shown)


apps-rendering/src/components/followStatus.tsx
```diff
@@ -6,9 +6,12 @@ import type { FC } from 'react';
 
 interface Props {
 	isFollowing: boolean;
+	contributorName: string;
 }
 
-type IconProps = Props;
+type IconProps = {
+	isFollowing: boolean;
+};
```
mxdvl, line 14: Maybe you can keep this in sync like so:

```suggestion
type IconProps = Pick<Props, "isFollowing">
```

mxdvl (approved):
LGTM–sorry for introducing this regression 😮‍💨

apps-rendering/src/components/followStatus.tsx
JamieB-gu: I think that's reasonable, and if it were a more complex type I think I'd prefer that. But given how simple it is and the slight overhead to process the `Pick` type for someone reading this I might opt to keep it duplicated if that suits you?

PR #3877, opened 2022-01-25
fix: Disables warnings on previously disabled errors

AshCorr:
## What does this change?

Github released a new feature in its "Files Changed" preview which shows check annotations on all files (not just edited files)

image

These 4 warnings are showing up in almost every single PR.

Hopefully this PR should stop those warnings showing in Github until the underlying issue can be resolved. (Mostly for our sanity)

### DIFF
--- a/apps-rendering/src/components/embedWrapper.stories.tsx
+++ b/apps-rendering/src/components/embedWrapper.stories.tsx
@@ -11,7 +11,7 @@ import { EmbedComponentWrapper } from './embedWrapper';
 // popped up after I removed the FC type which was causing problems when I tried to set parameters
 // on an individual story. Ideally, we'd use the type from Storybook directly
 // but https://github.com/storybookjs/storybook/issues/13486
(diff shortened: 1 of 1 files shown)


apps-rendering/src/components/embedWrapper.stories.tsx
```diff
@@ -11,7 +11,7 @@ import { EmbedComponentWrapper } from './embedWrapper';
 // popped up after I removed the FC type which was causing problems when I tried to set parameters
 // on an individual story. Ideally, we'd use the type from Storybook directly
 // but https://github.com/storybookjs/storybook/issues/13486
-// eslint-disable-next-line @typescript-eslint/explicit-function-return-type -- because 👆
```
AshCorr, line 14: Tried specifically adding the warning here instead of disabling all of the rules but that just ends up triggering the max line length rule.

oliverlloyd (approved):
Thank you!

PR #3881, opened 2022-01-26
`/Blocks`

oliverlloyd:
## What does this change?
This PR adds a new DCR endpoint, `POST /Blocks`. This endpoint accepts blocks in json format and returns the html representation of them that we want to insert into the browser dom.

### Usage
```bash
curl -X POST http://localhost:9000/Blocks -H 'Content-Type: application/json' -d '{
	"format": {
		"design": "LiveBlogDesign",
		"theme": "NewsPillar",
		"display": "StandardDisplay"
	},
	"host": "",
	"pageId": "",
	"webTitle": "",
	"ajaxUrl": "",
	"isAdFreeUser": "",
	"isSensitive": "",
	"videoDuration": "",
	"edition": "",
	"section": "",
	"sharedAdTargeting": {},
	"adUnit": "",
	"host": "",
	"host": "",
	"host": "",
	"host": "",
	"blocks": [{
		"id": "61f1730f8f086b55f0816529",
		"elements": [{
				"html": "<p>From the BBC’s <strong>Peter Saull</strong></p>",
				"_type": "model.dotcomrendering.pageElements.TextBlockElement",
				"elementId": "2ce734b2-181c-41d0-80b2-7309e3acddde"
			},

		],
		"blockCreatedOn": 1643213583000,
		"blockCreatedOnDisplay": "16.13 GMT",
		"blockLastUpdated": 1643213596000,
		"blockLastUpdatedDisplay": "16.13 GMT",
		"blockFirstPublished": 1643213597000,
		"blockFirstPublishedDisplay": "16.13 GMT",
		"primaryDateLine": "Wed 26 Jan 2022 16.13 GMT",
		"secondaryDateLine": "First published on Wed 26 Jan 2022 07.56 GMT"
	}]
}'
```

## Why?
To enable live updates on blogs.

When a blog is live it polls an api asking for updates (new blocks). The request flow is:

1) The browser (running DCR code) makes an api request to get updates
2) Frontend receives this request
3) Frontend calls CAPI asking for any updates
4) CAPI returns the list of new blocks in json format
5) Frontend adapts this json to the DCR model
6) Frontend then makes a call to DCR (`/Blocks`) asking it to render the json blocks into html **this is the api added here**
7) Frontend then returns this html to the browser in response to the original request


dotcom-rendering/src/model/enhance-placeholders.oldtests.ts
```diff
@@ -0,0 +1,54 @@
...
+// 		};
+
+// 		const variables = new Map();
+// 		variables.set('byline', 'Joda');
+// 		const got = replacePlaceholders(el, variables).html;
+// 		const want = el.html;
+
+// 		expect(got).toEqual(want);
+// 	});
+// });
```
mxdvl, line 54: Did you mean to keep this in VCS?

dotcom-rendering/index.d.ts
```diff
@@ -551,6 +551,22 @@ interface TagType {
...
+	pageId: string;
+	webTitle: string;
+	ajaxUrl: string;
+	isAdFreeUser: boolean;
+	isSensitive: boolean;
+	edition: string;
+	section: string;
+	sharedAdTargeting: Record<string, unknown>;
+	adUnit: string;
+	videoDuration?: number | undefined;
```
mxdvl, line 567:
```suggestion
	videoDuration?: number;
```

dotcom-rendering/index.d.ts
```diff
@@ -551,6 +551,22 @@ interface TagType {
 	bylineImageUrl?: string;
 }
 
+interface BlocksRequest {
```
mxdvl, line 554: Where does this come from? Could we have some JSDoc info as part of the interface?

```ts
/**
 * A request from [frontend][]. This type is kept in sync with Scala’s manually.
 *
 * [frontend]: https://github.com/guardian/frontend/blob/…
 */
interface BlocksRequest {
```

dotcom-rendering/src/web/server/blocksToHtml.tsx
```diff
@@ -0,0 +1,71 @@
...
+ * @param webTitle string
+ * @param ajaxUrl string
+ * @param isAdFreeUser boolean
+ * @param isSensitive boolean
+ * @param edition string
+ * @param section string
+ * @param sharedAdTargeting Record<string, unknown>
+ * @param adUnit string
+ * @param videoDuration: number | undefined
+ * @returns string (the html)
```
mxdvl, line 25: This doesn’t add anything if you already use TypeScript. I also think the syntax is slightly off, the type should come first and be inside curly brackets.

It’s worth defining param only if you want to add some information about them.

mxdvl (changes requested):
This is nice!

**EDIT:** I mis-clicked. Meant to comment only.

dotcom-rendering/src/model/enhance-placeholders.oldtests.ts
oliverlloyd: No, I did not! Thanks for pointing this out!

dotcom-rendering/src/web/server/blocksToHtml.tsx
oliverlloyd: You're right. These really don't add anything. Removed

dotcom-rendering/index.d.ts
oliverlloyd: Comment added 👍 

In general though, this is a very long standing issue where we have no way to share types between Frontend and DCR. Personally, I think GraphQL would be a good fit here but that's a _much_ larger discussion than this PR!

dotcom-rendering/index.d.ts
```diff
@@ -551,6 +551,25 @@ interface TagType {
...
+	pageId: string;
+	webTitle: string;
+	ajaxUrl: string;
+	isAdFreeUser: boolean;
+	isSensitive: boolean;
+	edition: string;
+	section: string;
+	sharedAdTargeting: Record<string, unknown>;
+	adUnit: string;
+	videoDuration?: number;
```
mxdvl, line 570: Small but I tend to order these alphabetically. VS Code has a way to “sort lines ascending”.

mxdvl (approved):
🙆

dotcom-rendering/index.d.ts
mxdvl: https://eslint.org/docs/rules/sort-keys

jamesgorrie: That `sort-keys` could make sense, but when combined with TS, and some of the points @oliverlloyd makes (required/optional), with a few other points (shorthand vs explicit `{ a, c, b: thingy }`) - you could get some pretty strange looking code (strange being subjective).

Good way to try is to add the linting rule, auto lint and look at the diff?

PR #3885, opened 2022-01-27
Make document block component an island

DavidLawes:
## What does this change?

This change makes the document block element an island. In this PR we:

- include the click to view wrapper inside the document block component
- include the importable component in renderElement.tsx (taking account of updated props)
- update stories and tests to take account of updated props

## Why?


oliverlloyd (approved):
🚀

mxdvl (commented):
Looking at this island migration is great! 🏝️

PR #3887, opened 2022-01-28
Make youtube block component an island

DavidLawes:
## What does this change?
Makes the youtube block component an island.  In this PR we:

- include the importable component in renderElement.tsx
- rename the amp youtube block component to prevent webpack issues when determining correct importable file
- remove the use of the dynamic import for getting consent state (the previous implementation was not server compatible)
- make use of the latest youtube atom (atoms-rendering 22.3.0) to ensure correct ophan events are recorded

Note: this implementation does not solve the edge case where eager users click play before hydration is complete (in this case the video doesn't autoplay when it is ready, the user has to click again). We removed the `deferUntil` property to force youtube island hydration to happen as soon as possible to reduce the likelihood of this happening.

## Why?


dotcom-rendering/src/web/components/App.tsx
```diff
@@ -127,35 +126,7 @@ export const App = ({ CAPI }: Props) => {
 		theme: pillar,
 	};
 
-	const adTargeting: AdTargeting = buildAdTargeting({
```
DavidLawes, line 130: `adTargeting` can be determined purely on server side. With Islands there is no need to `buildAdTargeting` on the client.

dotcom-rendering/src/web/lib/renderElement.tsx
```diff
@@ -696,24 +696,26 @@ export const renderElement = ({
...
-					assetId={element.assetId}
-					expired={element.expired}
-					overrideImage={element.overrideImage}
-					posterImage={element.posterImage}
-					duration={element.duration}
-					mediaTitle={element.mediaTitle}
-					altText={element.altText}
-					origin={host}
-				/>,
+				<Island>
```
DavidLawes, line 699: There is deliberately no `deferUntil` here in order to force hydration as soon as possible. This will reduce the likelihood of users clicking to play a video before hydration has completed.

DavidLawes:
NB: there are 2 chromatic changes that need approving, but I don't think either of these are a direct result of my changes. Tracing back through the history they've been introduced by previous merges to main (but where the changes were never accepted)

dotcom-rendering/src/amp/components/Elements.tsx
```diff
@@ -212,7 +212,7 @@ export const Elements = (
 				);
 			case 'model.dotcomrendering.pageElements.YoutubeBlockElement':
 				return (
-					<YoutubeBlockComponent
+					<YoutubeBlockComponentAMP
```
OllysCoding, line 215: Could we include a little bit of detail on why this component had to be renamed?

OllysCoding (approved):
+1 Looks great!

dotcom-rendering/src/web/components/YoutubeBlockComponent.importable.tsx
```diff
@@ -94,22 +91,16 @@ export const YoutubeBlockComponent = ({
...
-					});
-				},
-			)
-			.catch((error) => {
-				window.guardian.modules.sentry.reportError(
-					new Error(`Error: ${error}`),
-					'youtube-consent',
-				);
+		try {
+			window.guCmpHotFix.onConsentChange((newConsent: ConsentState) => {
```
oliverlloyd, line 95: @sndrs @ashishpuliyel Where are you at on this pattern? We need some way to introduce cmp code inside a file which gets rendered on the server

oliverlloyd (approved):
Outside the question on the cmp hotfix pattern, this is great. I have looked at this code a lot now so I'm happy to thumbs this up!

dotcom-rendering/src/web/components/App.tsx
arelra: Oh I see its built server side and is available client side - ignore!

dotcom-rendering/src/amp/components/Elements.tsx
DavidLawes: Great question :) I alluded to this in the PR description (perhaps a little too vague!)

> rename the amp youtube block component to prevent webpack issues when determining correct importable file

IIRC, without the file re-naming webpack had issues determining which file to import (the amp one or the web one). The result was an error with the youtube importable on the client (something like 'youtubeblockcomponent.importable does not exist'). It was all very confusing until @oliverlloyd pointed out the necessity for this file name change :) 

I think the component name change probably isn't necessary (in fact, I don't _think_ it is because I just re-tested with the component name unchanged and all seemed ok). However, given the distinction needed between web and amp I quite like the component name updated too though, what do you think?

dotcom-rendering/src/web/components/App.tsx
DavidLawes: Sorry, perhaps my comment wasn't super clear. What I was trying to say here is that we don't need to recreate the `adTargeting` object on the client, it only needs to be created once on the server. `adTargeting` is provided to the youtube atom player (passed down as a prop when server rendered https://github.com/guardian/dotcom-rendering/blob/dlawes/youtube-block-island/dotcom-rendering/src/web/components/YoutubeBlockComponent.importable.tsx#L200) and will be available to youtube atom player on the client

arelra: Thats great yes thanks for explaining 👍

dotcom-rendering/src/web/components/YoutubeBlockComponent.importable.tsx
sndrs: this would be one for @kenoir – we discussed the need, not sure of the priority though. 

it shouldn't be hard to implement though if you want to take a crack @DavidLawes? i think a similar check for `window` around [here](https://github.com/guardian/consent-management-platform/blob/3ca4f9113495f9268ac34b41338850be5604a8af/src/index.ts#L32) might be enough...

kenoir: @oliverlloyd is [this](https://github.com/guardian/consent-management-platform/pull/555) the kind of thing you were thinking of?

sndrs: just to update this – @DavidLawes and i spoke offline, he's going to look into why the import inside `useEffect` failed before (seems like it should work!) and @kenoir is also going to push forward with https://github.com/guardian/consent-management-platform/pull/555 because it should really not fail if it's imported in server-side code anyway

DavidLawes: To update here: I've changed to a dynamic import and now both server and client side rendering appears to work ok. Cmp is dynamically loaded on the client, consent is set correctly and I can play videos.

My only concern / confusion is about the number of times cmp is requested:

page with a single video (cmp requested 3 times):
https://www.theguardian.com/uk-news/2020/jan/09/prince-harry-meghan-decision
Screenshot 2022-02-24 at 12 21 10

page with multiple videos (cmp requested once):
Screenshot 2022-02-24 at 12 21 31


@oliverlloyd suggested this could be a result of being in-between the App.tsx --> Islands migration. 

If the way I've dynamically imported cmp looks ok, maybe it's worth pushing this solution and investigating the multiple imports as a follow-on task? cc @sndrs 

Also @kenoir - I'd attempted to make cmp server safe, your implementation looks much neater & well-informed than mine :)

DavidLawes: ^^ re cmp loading, I think there's something else going on, on `main` this [page](https://www.theguardian.com/uk-news/2020/jan/09/prince-harry-meghan-decision) already loads 2 cmp bundles. I think cmp bundle loading should be investigated separately from this change...

PR #3888, opened 2022-01-28
Track errors in the Supporter Revenue header links

tomrf1:
We can see various errors in sentry relating to the fetching/rendering of this component. It's hard to get a sense of the scale of the problem in sentry, or to see patterns in user agents/devices.
With this PR we send a componentEvent to ophan if the header links component fails to load for any reason. We can then use BigQuery to explore this data.

### Current behaviour if header links component is successfully inserted:
Screen Shot 2022-01-28 at 14 26 41

### New behaviour if header links component insertion fails:
Screen Shot 2022-01-28 at 14 19 19


TomPretty (approved):
Makes sense 👍 I always find sentry so confusing to look at

PR #3823, opened 2022-01-11
Fix deadblog colours for metadata and liveblock

ioannakok:
## What does this change?
Fixes metadata and liveblock colours for deadblogs in AR.

## Why?
To match the designs and follow accessibility standards. 

| Background | White | Gray |
|------------|-------|------|
| News       | 400   | 400  |
| Sport      | 400   | 300  |
| Lifestyle  | 400   | 300  |
| Culture    | 350   | 300  |
| Opinion    | 300   | 200 for dark gray and 300 for light gray  |

| Before      | After |
| ----------- | ----------- |
| image      | image       |
| image   | image        |
| image  | image |
| image  | image      ​|
| image | image       ​|
| image | image       ​|
| image | image       ​|
| image | image  ​|

PR #3889, opened 2022-01-28
Fix deadblog colours for DCR

ioannakok:
## What does this change?
Makes DCR deadblog colours the same as AR as per this PR: https://github.com/guardian/dotcom-rendering/pull/3823

## Why?
To match the designs and follow accessibility standards.

Background | White | Gray
-- | -- | --
News | 400 | 400
Sport | 400 | 300
Lifestyle | 400 | 300
Culture | 350 | 300
Opinion | 300 | 200 for dark gray and 300 for light gray

| Before      | After |
| ----------- | ----------- |
| image      | image       |
| image   | image  |
| image | image |
| image | image  |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |

### DIFF
--- a/dotcom-rendering/src/web/lib/decidePalette.ts
+++ b/dotcom-rendering/src/web/lib/decidePalette.ts
@@ -831,7 +831,7 @@ const textRichLink: (format: ArticleFormat) => string = (format) => {
 			case ArticleSpecial.Labs:
 				return BLACK;
 			case ArticleSpecial.SpecialReport:
-				return specialReport[300];
+				return specialReport[400];
 		}
 	}
 	return pillarPalette[ArticlePillar.News][400];
@@ -864,7 +864,7 @@ const borderRichLink: (format: ArticleFormat) => string = (format) => {
 			case ArticleSpecial.Labs:
 				return BLACK;
 			case ArticleSpecial.SpecialReport:
-				return specialReport[300];
+				return specialReport[400];
 		}
 	}
 	return pillarPalette[ArticlePillar.News][400];
@@ -909,9 +909,9 @@ const fillRichLink: (format: ArticleFormat) => string = (format) => {
 			case ArticlePillar.Opinion:
 				return opinion[300];
 			case ArticleSpecial.Labs:
-				return BLACK;
+				return labs[400];
 			case ArticleSpecial.SpecialReport:
-				return specialReport[300];
+				return specialReport[400];
 		}
 	}
 	return pillarPalette[ArticlePillar.News][400];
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -731,7 +817,22 @@ const hoverHeadlineByline = (format: ArticleFormat): string => {
 
 const textRichLink: (format: ArticleFormat) => string = (format) => {
 	if (format) {
-		return pillarPalette[format.theme].main;
+		switch (format.theme) {
```
DavidLawes, line 820: These changes aren't scoped only to deadblogs (?) - is this ok?

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -204,6 +202,25 @@ const textSyndicationButton = (format: ArticleFormat): string => {
 };
 
 const textArticleLink = (format: ArticleFormat): string => {
+	if (format.design === ArticleDesign.DeadBlog) {
```
DavidLawes, line 205: I like how you've separated out the requirements for deadblogs, I can see how we could refactor (to reduce lines of code) but I prefer your approach :)

DavidLawes (approved):
Thanks for the PR description - the changes look very detailed and specific! Out of curiosity, would we expect any chromatic diffs as a result of this?

ioannakok:
> Thanks for the PR description - the changes look very detailed and specific! Out of curiosity, would we expect any chromatic diffs as a result of this?

Yes! You are right. Here are the diffs I have accepted:
https://www.chromatic.com/build?appId=5dfcbf3012392c0020e7140b&number=9844
https://www.chromatic.com/build?appId=5dfcbf3012392c0020e7140b&number=9846

I did this incrementally, so they don't show in the final version of the PR. Should have accepted them at the end.

mxdvl (approved):
This looks good to me… quite easy to get my head around!

dotcom-rendering/src/web/lib/decidePalette.ts
ioannakok: In RichLink we pass the format of the article we're linking to, e.g. when we're linking to an article, the design will be Standard and not Deadblog. So if we add a case for deadblogs, it won't be reached. I was ambivalent as to what to do but decided to make the change as Culture and Lifestyle richLinks are not accessible atm. Thanks for pointing this out, I will double check with @HarryFischer!

ioannakok: It is okay to make this change 👍

PR #3866, opened 2022-01-24
Modernise related content section

MarSavar:
## Why?
UI modernisation as per request from @ajbreuer

The main changes are:

- Images with rounded corners
- Media cards with rounded corners
- Background colours of cards
- Related content section background colour
- Top border for the related content section
- No more lines on byline cards

These changes reflect what the app currently looks like.

### Light mode wide breakpoint (before)
image
### Light mode wide breakpoint (after)
image
---
### Light mode narrow breakpoint (before)
image
### Light mode narrow breakpoint (after)
image
---
### Light mode byline/comment card (before)
image
### Light mode byline/comment card (after)
image
---
### Dark mode wide breakpoint (before)
image
### Dark-mode wide breakpoint (after)
image
---
### Dark mode narrow breakpoint (before)
image
### Dark mode narrow breakpoint (after)
image
--
### Dark mode byline/comment card (before)
image

### Dark mode byline/comment card (after)
image

PR #3890, opened 2022-01-28
Collapse spacing between article and related content

MarSavar:
## Why?
#3866 introduced new styling for the Related Content section in Apps Rendering, including a top border for the section.
Adding a visibile border revealed a (barely perceptible) gap between the article and the related content section.
This PR removes addresses the issue.

### Before
image

### After
image

### DIFF
--- a/apps-rendering/src/components/shared/relatedContent.tsx
+++ b/apps-rendering/src/components/shared/relatedContent.tsx
@@ -90,7 +90,6 @@ const styles = css`
 	padding-top: ${remSpace[3]};
 	padding-left: ${remSpace[4]};
 	padding-right: ${remSpace[4]};
-	margin-top: -0.125rem;
 
 	${until.wide} {
 		margin-left: -${remSpace[4]};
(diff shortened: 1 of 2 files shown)


MarSavar:
[Remove spacing between article and related content section](https://trello.com/c/eTL8twJ2/610-remove-spacing-between-article-and-related-content-section)

mxdvl:
@MarSavar I would be tempted to investigate why the border is appearing in the first place rather than using a negative margin. I’ve encountered such weirdness with whitespace not collapsing before.

Do you know [which area][] space is made up of? The node’s content `height`, `padding`, `border` or `margin`? 

[which area]: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Introduction_to_the_CSS_box_model

MarSavar:
> @MarSavar I would be tempted to investigate why the border is appearing in the first place rather than using a negative margin. I’ve encountered such weirdness with whitespace not collapsing before.
> 
> Do you know [which area](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Introduction_to_the_CSS_box_model) space is made up of? The node’s content `height`, `padding`, `border` or `margin`?

Good shout - looks like there's 2px padding top coming from `onwardStyles`. Getting rid of it removes the need for using a hacky negative margin in `relatedContent.tsx` :-)

image

PR #3893, opened 2022-01-31
fix: Changes cypress test for soundcloud to not look for specific content in the Soundcloud widget

AshCorr:
## What does this change?

Changes the Cypress test that checks if the soundcloud embed is working to not look for `Cookie policy`

## Why?

- A recent change in soundcloud seems to have removed `Cookie policy` which we were previously scanning for. I've removed the check to look for specific content as we don't really have any control over how and when Soundcloud make changes to their widget so its not really practical for us to be checking their content.

### DIFF
--- a/dotcom-rendering/cypress/integration/parallel-1/article.elements.spec.js
+++ b/dotcom-rendering/cypress/integration/parallel-1/article.elements.spec.js
@@ -168,7 +168,7 @@ describe('Elements', function () {
 				'Article?url=https://www.theguardian.com/music/2020/jan/31/elon-musk-edm-artist-first-track-dont-doubt-ur-vibe',
 			);
 
-			getIframeBody().contains('Cookie policy');
+			getIframeBody();
 		});
 
 		it('should render the football embed', function () {


dotcom-rendering/cypress/integration/parallel-1/article.elements.spec.js
```diff
@@ -168,7 +168,7 @@ describe('Elements', function () {
 				'Article?url=https://www.theguardian.com/music/2020/jan/31/elon-musk-edm-artist-first-track-dont-doubt-ur-vibe',
 			);
 
-			getIframeBody().contains('Cookie policy');
+			getIframeBody();
```
jamesgorrie, line 171: Weird name for this method as it's an `assert`, but it was here before hand.

jamesgorrie (approved):
Makes sense not to rely on third party content.

Could we check the iframe status is 2xx/3xx or something similar? Wondering if we would ever have the `iframe.is.not.empty` but the content errored.

AshCorr:
> Could we check the iframe status is 2xx/3xx or something similar? Wondering if we would ever have the iframe.is.not.empty but the content errored.

Don't think theres really a straight forward way to check the status of an iframe, seems like people are suggesting that you do a text search for error codes in the content of the iframe to do it which doesn't really seem all that robust.

PR #3894, opened 2022-01-31
Fix related content margins

MarSavar:
## Why?
This PR fixes an issue in the related content section (Apps Rendering), which was causing swiping problems on the iOS app.
This is due to negative margins not being correctly rendered by all browsers.

I have also addressed a minor visual bug which was causing an almost invisible 2px gap between the article and the related content section.

I have tested this change in Chrome, Firefox and Safari.

### Margin close-up (Safari)
image

### Margin close-up (Safari)
image

### DIFF
--- a/apps-rendering/src/components/shared/relatedContent.tsx
+++ b/apps-rendering/src/components/shared/relatedContent.tsx
@@ -1,11 +1,11 @@
 import { css } from '@emotion/react';
 import { RelatedItemType } from '@guardian/apps-rendering-api-models/relatedItemType';
 import {
-    from,
-    headline,
-    neutral,
-    remSpace,
-    until,
+	from,
+	headline,
+	neutral,
+	remSpace,
+	until,
 } from '@guardian/source-foundations';
 import type { Option } from '@guardian/types';
 import { map, withDefault } from '@guardian/types';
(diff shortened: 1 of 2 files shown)


apps-rendering/src/components/shared/relatedContent.tsx
```diff
@@ -88,12 +88,10 @@ const listStyles = css`
...
 	border-top: 1px solid ${neutral[46]};
 	padding-top: ${remSpace[3]};
-	padding-left: ${remSpace[4]};
-	padding-right: ${remSpace[4]};
 
 	${until.wide} {
-		margin-left: -${remSpace[4]};
-		margin-right: -${remSpace[4]};
+		padding-left: ${remSpace[4]};
+		padding-right: ${remSpace[4]};
```
mxdvl, line 94: FYI, there's a shorthand for vertical & horizontal:

```suggestion
		padding: 0 ${remSpace[4]};
```

MarSavar: This is intentional as I need `padding-top` to be defined at wider breakpoints. 

`padding-top` is defined outside `${until.wide}`.

image

I only need `padding-left` and `padding-right` to be applied at smaller breakpoints (mobile view) - they are defined within `${until.wide}`
image

Doing `padding: 0 ${remSpace[4]}` would get rid of `top-padding` at smaller breakpoints, but I need to keep it! So I think it makes sense to be explicit and use `padding-left` and `padding-right`, unless there's a more elegant solution...

mxdvl (approved):

MarSavar:
[Remove spacing between article and related content section](https://trello.com/c/eTL8twJ2/610-remove-spacing-between-article-and-related-content-section)

PR #3896, opened 2022-01-31
exclude dependabot branches from chromatic action

jamesgorrie:
part of https://github.com/guardian/dotcom-rendering/issues/3882

## What does this change?
Stops Chromatic being run on dependabot PRs.

## TL;DR

Shall we err on the side of over-testing and run Chromatic on dependabot PRs while monitoring costs in Chromatic, or assume that costs might explode and find a way to run Chromatic tests when certain packages update?

## Why?
Now that `dependabot` can run Chromatic, I would prefer if it didn't. `dependabot` PRs can kick off a lot of workflows through rebasing etc etc, which in turn would cost 💰. 

## But...
We will need to run these on specific package updates especially as we have UI elements that could change in `@guardian/**` modules. I'm not sure if we use `dependabot` to do this for us, or if it's something we do more manually, so would like feedback on whether we think this would be an issue.

Another option would be to keep running the chromaui action and watch costs to see if that's an issue.

While writing this PR, it's made me think erring on the side of testing more, especially as we're add more automation into the CD pipeline, would be better if cost allows.

## Small refactor
`s/CHROMATIC_APP_CODE/CHROMATIC_PROJECT_TOKEN` <= this is to match

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -20,5 +20,6 @@ jobs:
 
       - uses: chromaui/action@v1
         with:
-          projectToken: ${{ secrets.CHROMATIC_APP_CODE }}
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
           token: ${{ secrets.GITHUB_TOKEN }}
+          skip: dependabot/**
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -22,4 +22,3 @@ jobs:
         with:
           projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
           token: ${{ secrets.GITHUB_TOKEN }}
-          skip: dependabot/**


DavidLawes:
I could be wrong, but I thought a release of `guardian/**` modules didn't trigger dependabot? It's up to us to incorporate the latest versions.

If we don't run chromatic on the PR, but we decide to merge the change, at what point do we run chromatic (when latest `main` builds)?

Do we have an idea of how much the cost would be? Over the last year (from 01-01-2021 until now) there's been 285 PRs created by dependabot. Are we saying there's multiple chromatic builds per dependabot PR? And do we know the cost per chromatic build?

Also, how are we intending to utilise dependabot PRs? Are we going to be more proactive in responding to PRs raised (i.e. merging the PRs dependabot raises)? I'd prefer to actually use the dependabot PRs to continually bump module versions, in which case the more testing we have per PR the better?

mxdvl:
I’ve switched to Renovate on some personal projects and turned on auto-merging and found it to be great. You can even skip the creation of PR if all the CI goes green.

As an alternative to auto-merge, Renovate can also group all the the dependencies update into a single PR, which would lead to less rebasing, and less cognitive load for developers to merge these PRs.

Sure, you can achieve both of these things with Dependabot, but it was _way simpler_ with Renovate.

jamesgorrie:
@mxdvl - 100% up for another solution, it kind of stalled in DevX. I'll revive it.

oliverlloyd:
My understanding is that the Chromatic tests do get run automatically when new versions of `@guardian` libs are published. Which I think is correct because these libs are the same as all the other external dependencies that we're importing into DCR and can break the site.

I personally think that we should keep the high level of testing that we have in DCR. It is a joy to use DCR exactly because we have such great tests. That level of confidence is a huge win and reducing it - even if it saves some money - is not the direction I would like to see us go.

It's true that there is a trade off, we can't test everything all the time, but in the case where the tests are automated and have no downsides other than cost then I would prefer that we spend more on testing tools to improve devex and test coverage rather than reduce these things to save money.

jamesgorrie:
Given that dependancy updates can affect UI, we should keep Chromatic running on those updates. Keep the level of testing high!

So:
* Keep chromatic running (I'll leave this PR in as a refactor on the env var)
* If costs become a problem, refine the dependabot process, potentially switch in something like renovate
* Look into renovate anywho

sndrs:
i think you could use [dependabot/fetch-metadata](https://github.com/dependabot/fetch-metadata) to detect if it's a dev or prod dep. there's probably not a lot of point running chromatic on development dep bumps?

JamieB-gu:
> there's probably not a lot of point running chromatic on development dep bumps?

Is there enough of a distinction between dev and prod dependencies to be confident that a dev dependency will never alter production code? For example: TypeScript, Webpack and a number of loaders are in dev dependencies, but will contribute code to the production bundle I believe?

PR #3901, opened 2022-02-01
fix toggle styling regression

joecowton1:
## What does this change?

Updates the `metadata` component in AR so the `cssOverride` prop correctly overrides the label styling on the Source `ToggleSwitch` component.

## Why?

We've recently [updated](https://github.com/guardian/source/pull/1249) the `ToggleSwitch` components' HTML structure to allow users to click the text. This means that the css `label` selector is no longer required.

### Before
Screenshot 2022-02-01 at 11 34 35

### After
Screenshot 2022-02-01 at 11 34 51


ioannakok (approved):
👍

PR #3902, opened 2022-02-01
CAPIElement type safety!

mxdvl:
## What does this change?

Assert that the `type` argument should match the type of block we check for.

## Why?

There’s a type assertion, so we did not have type safety. This resolves that.


oliverlloyd (approved):
Looks good.

It's worth noting that this file is [scheduled to be deleted](https://docs.google.com/document/d/1kSrZZrJ8hSw6bek5hGa9onE8QDz5TRHFqHV52EOFjts/edit#), but this change make sense nonetheless

Issue #3884, opened 2022-01-26
Speed up 🐌 build


jamesgorrie:
The build is currently being reported as slow by us, but also by our colleagues that are clients to our platform 🐌 .

There are a number of areas where we can try to make changes to improve this.

We need to have some visibility into where there might be issues (#3848).

From here we can 
* fix any immediately obvious bottlenecks
* fix any memory leaks
* investigate using different build systems

PR #3906, opened 2022-02-01
Improve dev server logging & Fix webpack externals

OllysCoding:
## What does this change?

- Update logging for each build to reflect the build being completed
- Fix bug with webpack-node-externals which caused a significant number of node_modules to be included in the build
- Add `bufferutil` & `utf-8-validate` to IgnorePlugin to remove unneeded warnings about optional dependencies

This means we'll not get proper warnings from the server build, and our general logging will look more like this:

image

Co-Authored-By: Max Duval <hi@mxdvl.com>

## Why?

It's currently unclear to the dev-server users what's actually happening, we duplicate the same message for client & legacy client build completion, and provide no information at all on the server build completion.

This update therefor provides more more accurate feedback to the user as to when the dev-server is ready, as well as what builds are building & when they complete.

This should also help provide more clarity for the build speed-up work

---

This change bought attention to the fact that some node_modules were being included in the server build, rather than being left as external.

By adding the `additionalModuleDirs` option to `webpack-node-externals`, we can make sure we capture all node_modules and don't include them in the server build. This provides a healthy bump to the speed of the server build, meaning it now even completes before the client build!

---

Part of #3884


dotcom-rendering/scripts/webpack/frontend.js
```diff
@@ -48,7 +48,9 @@ const commonConfigs = ({ platform }) => ({
 		}),
 		// Does not try to require the 'canvas' package,
 		// an optional dependency of jsdom that we aren't using.
-		new webpack.IgnorePlugin({ resourceRegExp: /^canvas$/ }),
+		new webpack.IgnorePlugin({
+			resourceRegExp: /^(canvas|bufferutil|utf-8-validate)$/,
```
OllysCoding, line 52: 'bufferutil' and 'utf-8-validate' are both optional dependencies of 'ws' which are not installed by jsdom. As these are not used, we can filter out their requires, reducing noise on the server build.

dotcom-rendering/scripts/webpack/server.js
```diff
@@ -19,6 +32,8 @@ module.exports = ({ sessionId }) => ({
 	},
 	externals: [
 		'@loadable/component',
+		'express',
+		'log4js',
```
OllysCoding, line 36: Express & log4js both use expressions in `require()`

This is where require takes an expression, e.g
```js
function getModule (name) {
    return require(name)
}```

Which is not suited to webpack ~ as it is unable to know what it is supposed to bundle to meet the requirements of the module. By specifying these are external to the build, they never get bundled. 

I'll be testing this on CODE to make sure everything works as expected.

**Comment by OllysCoding** [comment]
[More clarity on individual build completion](https://trello.com/c/qibqsr9q/539-more-clarity-on-individual-build-completion)

**Review by mxdvl — APPROVED** [review]
We paired on this and it looked sound!

File: dotcom-rendering/scripts/webpack/frontend.js (hunk 1 above)
Comment (jamesgorrie) [reply dotcom-rendering/scripts/webpack/frontend.js:52]: Could we update the code comment to reflect this?

File: dotcom-rendering/scripts/webpack/server.js (hunk 2 above)
Comment (jamesgorrie) [reply dotcom-rendering/scripts/webpack/server.js:36]: It would be really useful to have this in the code for future editors as we don't have a test explaining it.

File: dotcom-rendering/scripts/webpack/server.js [hunk 3]
```diff
@@ -1,7 +1,20 @@
+const FriendlyErrorsWebpackPlugin = require('friendly-errors-webpack-plugin');
+const chalk = require('chalk');
 const GuStatsReportPlugin = require('./gu-stats-report-plugin');
 
 const DEV = process.env.NODE_ENV === 'development';
 
+const friendlyErrorsWebpackPlugin = () =>
```
jamesgorrie, line 7: What's the point of this abstraction? I know we do it in `browser.js` - but I can't see the reason for it existing there either is. Could be bad for maintenance as someone new to the code would wonder the same thing, I assumed it was used elsewhere, but it's not.

jamesgorrie (commented):
Couple of questions

dotcom-rendering/scripts/webpack/webpack.config.server.js
```diff
@@ -19,6 +32,8 @@ module.exports = ({ sessionId }) => ({
 	},
 	externals: [
 		'@loadable/component',
+		'express',
+		'log4js',
 		require('webpack-node-externals')({
```
sndrs, line 37: this _should_ be excluding everything in `node_modules` already – if this change works it suggests it is not? 

[the docs](https://www.npmjs.com/package/webpack-node-externals) (for v3) suggest webpack 5 config should use `externalsPresets` instead of `target`, i wonder if updating `webpack-node-externals` v3 and changing that config setting will get it working again?

OllysCoding: I shall investigate and update here with what I find :)

OllysCoding: Now resolved! PR Description updated with more info!

dotcom-rendering/scripts/webpack/webpack.config.server.js
```diff
@@ -21,6 +24,14 @@ module.exports = ({ sessionId }) => ({
...
 			allowlist: [/^@guardian/],
+			additionalModuleDirs: [
+				// Since we use yarn-workspaces for the monorepo, node_modules will be co-located
+				// both in the '(project-root)/dotcom-rendering/node_modules' directory (default for webpack-node-externals)
+				// but also in project root, and any workspaces we link to (like common-rendering).
+				// We want to make sure all of these are removed from the server build.
+				'../node_modules',
+				'../common-rendering/node_modules',
+			],
```
OllysCoding, line 34: Some useful info on the webpack-node-externals fix!

dotcom-rendering/scripts/webpack/server.js
OllysCoding: Removed!

OllysCoding:
Tested in CODE :)

PR #3911, opened 2022-02-02
declaration-no-important: true

jamesgorrie:
## What does this change?
Adds the `declaration-no-important` to the style linting.

## Why?
We don't want people to use this, because once it's added, it is incredibly hard to remove. Hence the disabling of the rule on the existing uses. The work to remove these is fiddly, so we're choosing to just prevent it from happening in the future.


oliverlloyd (approved):
Love the fast turnaround PR

mxdvl (approved):
Wow—great! Removing the existing one can come later.

PR #3912, opened 2022-02-02
Atoms as islands

oliverlloyd:
## What does this change?
Here we're moving a bunch of atoms into Islands. Specifically,

```
	QandaAtom
	GuideAtom
	ProfileAtom
	TimelineAtom
	ChartAtom
	PersonalityQuizAtom
	KnowledgeQuizAtom
```

We're also add version `22.2.0` of `@guardian/atoms-rendering` which allows us to remove the tracking handlers

## Why?
Islands are cool.

Why are we removing the tracking handlers? Because adding these on the server and then trying to hydrate them was non intuitive and added complexity. By adding a default to the atom itself we can simply not include the handler and use the fallback instead.

## What are these wrapper components?
For code that's imported from an external lib it's not possible for our `Island` abstraction to dynamcally import it because we currently require island components to exist in `/component`. To get around this we use a simple wrapper, as shown here.

### DIFF
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -11,16 +11,6 @@ import { GetMatchNav } from '@frontend/web/components/GetMatchNav';
 import { StickyBottomBanner } from '@root/src/web/components/StickyBottomBanner/StickyBottomBanner';
 import { SignInGateSelector } from '@root/src/web/components/SignInGate/SignInGateSelector';
 
-import {
-	QandaAtom,
-	GuideAtom,
-	ProfileAtom,
-	TimelineAtom,
-	ChartAtom,
-	PersonalityQuizAtom,
-	KnowledgeQuizAtom,
-} from '@guardian/atoms-rendering';
-
 import { AudioAtomWrapper } from '@frontend/web/components/AudioAtomWrapper';
 
 import { Portal } from '@frontend/web/components/Portal';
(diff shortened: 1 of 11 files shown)


dotcom-rendering/src/web/components/App.tsx
```diff
@@ -11,16 +11,6 @@ import { GetMatchNav } from '@frontend/web/components/GetMatchNav';
...
 
-import {
-	QandaAtom,
-	GuideAtom,
-	ProfileAtom,
-	TimelineAtom,
-	ChartAtom,
-	PersonalityQuizAtom,
-	KnowledgeQuizAtom,
-} from '@guardian/atoms-rendering';
```
OllysCoding, line 22: If I'm reading this right, does this mean we were importing all these atoms on the page no matter what?! Glad to see it go!

dotcom-rendering/src/web/components/App.tsx
```diff
@@ -11,16 +11,6 @@ import { GetMatchNav } from '@frontend/web/components/GetMatchNav';
...
-
 import { AudioAtomWrapper } from '@frontend/web/components/AudioAtomWrapper';
```
OllysCoding, line 14: What's happening to the audio atom? Will this be handled by a separate PR?

dotcom-rendering/src/web/components/App.tsx
oliverlloyd: Yes! 😱

dotcom-rendering/src/web/components/App.tsx
oliverlloyd: Good spot. I've added a ticket for this one.

dotcom-rendering/src/web/components/App.tsx
```diff
@@ -525,113 +437,6 @@ export const App = ({ CAPI, ophanRecord }: Props) => {
...
-							'DISLIKE',
-						)}
-						expandCallback={componentEventHandler(
-							'QANDA_ATOM',
-							qandaAtom.id,
-							'EXPAND',
-						)}
-					/>
-				</HydrateOnce>
-			))}
```
OllysCoding, line 554: I see that for Qanda atoms (no idea what these are) we are no longer reporting ophan data. Is this perhaps something we can do in the wrapper to make sure we're not loosing functionality?

oliverlloyd: The tracking requests still get sent. It's just I made these tracking props optional and then added a default inside the atoms themselves. 

https://github.com/guardian/atoms-rendering/pull/325

OllysCoding: 👍 Thanks for clarifying!

OllysCoding (approved):
+1 Looks good!

PR #3915, opened 2022-02-02
`CalloutBlockComponent` as an Island

oliverlloyd:
## What does this change?
Moves `CalloutBlockComponent` over to the island pattern

## Why?
We like islands


DavidLawes (approved):
👍

mxdvl:
🏝️  #3629

PR #3919, opened 2022-02-03
Add `SpacefinderOkr1FilterNearby` test

simonbyford:
## What does this change?

Introduces an A/B test in DCR which can be used to measure the impact of this change: https://github.com/guardian/frontend/pull/24592

This was the brief from Sara:

"In order to detect a statistically significant 1% change in inline programmatic revenue per 1000 pageviews we'd need a sample size of at least 7.4 million across both conditions.

We'd expect to reach these numbers after running a 10% test for 14 days.

This would mean setting up the test so that 5% of desktop web article page views were in the variant condition and 5% were in the control."

## Why?

### DIFF
--- a/dotcom-rendering/src/web/components/CommercialMetrics.tsx
+++ b/dotcom-rendering/src/web/components/CommercialMetrics.tsx
@@ -4,6 +4,7 @@ import { sendCommercialMetrics } from '@guardian/commercial-core';
 import { useOnce } from '@root/src/web/lib/useOnce';
 import { getCookie } from '@guardian/libs';
 import { useAB } from '@guardian/ab-react';
+import { spacefinderOkr1FilterNearby } from '@frontend/web/experiments/tests/spacefinder-okr-1-filter-nearby';
 import { useDocumentVisibilityState } from '../lib/useDocumentHidden';
 import { useAdBlockInUse } from '../lib/useAdBlockInUse';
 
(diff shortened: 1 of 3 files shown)


simonbyford:
For the record, we decided to double the size of the test to 20%

chrislomaxjones:
This looks good, only thing I noticed is if we want to capture commercial metrics for all pageviews in the test you'll need to add the test [here](https://github.com/guardian/dotcom-rendering/blob/573f291bd2644fdb8a6912518760f6adc0126ee0/dotcom-rendering/src/web/components/CommercialMetrics.tsx#L24-L26).

simonbyford:
> This looks good, only thing I noticed is if we want to capture commercial metrics for all pageviews in the test you'll need to add the test [here](https://github.com/guardian/dotcom-rendering/blob/573f291bd2644fdb8a6912518760f6adc0126ee0/dotcom-rendering/src/web/components/CommercialMetrics.tsx#L24-L26).

Ah nice interesting, thanks for raising that @chrislomaxjones! I assume we do? Why are they excluded by default?

chrislomaxjones:
> I assume we do?

I think so! If the experiment involves measuring any of the commercial timers then we need this. I assume in the case of this test we do?

> Why are they excluded by default?

Good Q, all tests are excluded by default as not all of them are commercial, and therefore won't need the timing metrics. It would make sense that all commercial tests are included by default, but there's currently no mechanism to differentiate between a commercial and non-commercial test.

simonbyford:
> > I assume we do?
> 
> I think so! If the experiment involves measuring any of the [commercial timers](https://github.com/guardian/commercial-core/blob/main/src/EventTimer.ts#L21-L32) then we need this. I assume in the case of this test we do?
> 
> > Why are they excluded by default?
> 
> Good Q, all tests are excluded by default as not all of them are commercial, and therefore won't need the timing metrics. It would make sense that all commercial tests are included by default, but there's currently no mechanism to differentiate between a commercial and non-commercial test.

Makes sense, thanks for going through that. I've made the change!

chrislomaxjones (approved):
Looking forward to seeing the results of this one!

ashishpuliyel (approved):
Great work Simon

PR #3920, opened 2022-02-03
Add kickers and fix styling of live blog related cards

MarSavar:
## Why?
Liveblogs were not displaying correctly In the related content section at the bottom of articles in Apps Rendering.

This PR addresses the issue by:

1) Adding padding and rounded corners to liveblog cards
2) Introducing a kicker component to AR in order to display kickers in related cards.

Kickers in related cards obey the following rules:
- If a kicker is passed to a related card component,  display it.
- If a kicker is not passed to a related component, check whether it is a live blog or not.
- If it is a live blog, render the "Live" kicker with a dot (non-pulsing, mirroring DCR).
- If it is not a live blog, do not render the kicker at all.

Kickers are styled according to rules defined in the editorial palette in `common-rendering`.

### Before
image

### After
image
image


ajbreuer:
Looks good @MarSavar

apps-rendering/src/components/shared/kicker.tsx
```diff
@@ -0,0 +1,93 @@
...
+const liveDot = (
+	format: ArticleFormat,
+	colour: string,
+): ReactElement | null => {
+	if (format.design === ArticleDesign.LiveBlog) {
+		return <span css={dotStyles(colour)} />;
+	} else {
+		return null;
+	}
+};
```
JamieB-gu, line 37: This could technically be considered a React component in its own right?

```suggestion
type LiveDotProps = {
  format: ArticleFormat,
  colour: string,
}

const LiveDot: FC<LiveDotProps> = ({ format, colour }) => {
	if (format.design === ArticleDesign.LiveBlog) {
		return <span css={dotStyles(colour)} />;
	} else {
		return null;
	}
};
```

Which is then rendered like this:

```tsx
<LiveDot format={format} colour={kickerColour} />
```

I don't think it's necessary to re-write it btw, just an FYI.

JamieB-gu (commented):
👍 for the use of `editorialPalette`, really nice!

A few comments/suggestions.

MarSavar:
Thanks for the comments @JamieB-gu  - I've decided to do away with the `getKickerText` function altogether and use `maybeRender` which will render the `<Kicker>` component in `card.tsx` if there's a kicker text (or if the card is a liveblog). This simplifies the code even further!

mxdvl (approved):
LGTM!

PR #3923, opened 2022-02-04
`Braze`

oliverlloyd:
## What does this change?
This PR introduces `Braze` a component designed to hold the shared logic between `SlotBodyEnd` and `StickyBottomBanner`

## Why?
Because I think this is the safest refactor in order to make progress towards deleting `App.tsx`. It probably is possible to isolate each of these components and remove the global state shared between them but this logic is complex and I decided to take this approach as a stepping stone towards this goal instead of diving in.

## `slot-body-end`
This id was previously used by the `Portal` component as the marker for where to insert the element. But it has also been picked up as an id that can be used for some css styling. By removing the `Portal` I would break this styling code so I have added an `aside` wrapper around the component.

## `bannerWrapper`
This css previously existed in the `stickiness` lib and was added in the layout files. I've moved it down into the banner component directly as this seemed the more logical location for this code.

## `shouldShowSlotBodyEnd`
Previously we rendered the `SlotBodyEnd` and `StickyBottomBanner` components independently but they did not always appear together. For `Interactive` and `FullPageInteractive` articles `SlotBodyEnd` is not included and we also checked some flags. This function encapsulates this logic.

### DIFF
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -1,14 +1,12 @@
-import React, { useState, useEffect } from 'react';
+import React, { useEffect } from 'react';
 import loadable from '@loadable/component';
 import { useAB } from '@guardian/ab-react';
 import { tests } from '@frontend/web/experiments/ab-tests';
 import { ShareCount } from '@frontend/web/components/ShareCount';
 import { MostViewedFooter } from '@frontend/web/components/MostViewed/MostViewedFooter/MostViewedFooter';
 import { ReaderRevenueLinks } from '@frontend/web/components/ReaderRevenueLinks';
-import { SlotBodyEnd } from '@root/src/web/components/SlotBodyEnd/SlotBodyEnd';
 import { ContributionSlot } from '@frontend/web/components/ContributionSlot';
 import { GetMatchNav } from '@frontend/web/components/GetMatchNav';
-import { StickyBottomBanner } from '@root/src/web/components/StickyBottomBanner/StickyBottomBanner';
 import { SignInGateSelector } from '@root/src/web/components/SignInGate/SignInGateSelector';
 
 import { AudioAtomWrapper } from '@frontend/web/components/AudioAtomWrapper';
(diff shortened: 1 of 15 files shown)


dotcom-rendering/src/web/components/App.tsx
```diff
@@ -1,14 +1,12 @@
-import React, { useState, useEffect } from 'react';
+import React, { useEffect } from 'react';
```
oliverlloyd, line 1: 🎉

dotcom-rendering/src/web/components/Braze.tsx
```diff
@@ -0,0 +1,150 @@
...
+	format: ArticleFormat;
+};
+
+function shouldShowSlotBodyEnd(
+	format: ArticleFormat,
+	switches: Switches,
+	slotMachineFlags?: string,
+) {
+	if (!switches.slotBodyEnd) return false;
+	if (!parse(slotMachineFlags || '').showBodyEnd) return false;
```
oliverlloyd, line 43: Could somebody please check that this code is the same as:

```
	const showBodyEndSlot =
		parse(CAPI.slotMachineFlags || '').showBodyEnd ||
		CAPI.config.switches.slotBodyEnd;
```

This `parse` function is hard to reason about

jamesgorrie: The logic looks the same in each case. Spliting it out as above is nice and clear.
A tiny `slot-machine-flags.test` could be useful. 

What is the source of `CAPI.slotMachineFlags` & `CAPI.config.switches`?

oliverlloyd: > What is the source of CAPI.slotMachineFlags & CAPI.config.switches?

I'm honestly not sure. I think switches maybe come from a tool around Composer?  `slotMachineFlags` I know nothing about @sndrs any ideas?

dotcom-rendering/src/web/components/Braze.tsx
```diff
@@ -0,0 +1,150 @@
...
+	switch (format.design) {
+		case ArticleDesign.Interactive:
+		case ArticleDesign.FullPageInteractive:
+			return false;
+		default:
+			return true;
+	}
+}
+
+export const Braze = ({
```
tjmw, line 54: Pulling this out of `App` looks great. I'm not sure `Braze` is the right name here, as `SlotBodyEnd` and `StickyBottomBanner` are broader than just Braze (they may also show epics/banners from RRCP and also the CMP banner). Not sure what a good name is that ties those two slots together though (maybe this is what `Automat` is 😄 )

oliverlloyd:
Closing now that we have #4126 🎉

PR #3924, opened 2022-02-04
`getCmp`

oliverlloyd:
## What does this change?
This PR adds a small lib called `getCmp` which returns the version of the `cmp` object that gets placed on `window.guCmpHotFix`

## Why?
As we move more code over to the `Island` pattern we are starting to hit issues where the cmp lib is not server safe. If we refactor DCR's access to cmp to use this window object then this allows us to avoid the import, which is where we're getting errors on the server.

This abstraction is a helper utility to make that refactor. The main thing it provides is a check and a call to Sentry in the event that the property is not available. We believe that the property will always be there - we run `bootCmp` very early - but it might not be and we want to be aware if this happens. If we do see these errors, then it suggests we have a race condition and that we will need to implement some form of queue while the code waits for this value to be set.

### DIFF
--- a/dotcom-rendering/src/web/lib/getCmp.ts
+++ b/dotcom-rendering/src/web/lib/getCmp.ts
@@ -1,0 +1,24 @@
+/**
+ * Returns the cmp object so you don't need to import it
+ *
+ * The @guardian/consent-management-platform sets the window.guCmpHotFix
+ * value to equal the `cmp` object.
+ *
+ * If this function is called before this value has been set then an error
+ * is thrown allowing us to gather stats on any race conditions that could
+ * exist with this pattern
+ *
+ * @returns the cmp object
+ */
+export const getCmp = () => {
+	if (typeof window === 'undefined') return undefined;
+	if (!window.guCmpHotFix) {
+		window.guardian.modules.sentry.reportError(
+			new Error(
+				`Tried to access window.guCmpHotFix but it was falsy: ${window.guCmpHotFix}. This is indicative of a race condition`,
+			),
+			'cmp',
+		);
+	}
+	return window.guCmpHotFix;
+};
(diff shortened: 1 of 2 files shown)


oliverlloyd:
@shtukas @kenoir This is perhaps of interest to you based on our discussions the other day?

dotcom-rendering/src/web/lib/getCmp.ts
```diff
@@ -0,0 +1,24 @@
...
+ * value to equal the `cmp` object.
+ *
+ * If this function is called before this value has been set then an error
+ * is thrown allowing us to gather stats on any race conditions that could
+ * exist with this pattern
+ *
+ * @returns the cmp object
+ */
+export const getCmp = () => {
+	if (typeof window === 'undefined') return undefined;
```
mxdvl, line 14: Maybe just:

```suggestion
	if (window === undefined) return undefined;
```

mxdvl (approved):
I think erroring on the race condition should be good enough!

dotcom-rendering/src/web/lib/getCmp.ts
oliverlloyd: I think that will throw an error because the code will attempt to access `window`. You have to use the `typeof` to get around this

mxdvl: Ah, yes, of course!

can’t check `window`

mxdvl:
Did we see any errors crop up from this?

oliverlloyd:
> Did we see any errors crop up from this?

This code doesn't do anything, it's just adding a file. The PR to actually _use_ it is [this one](https://github.com/guardian/dotcom-rendering/pull/3925), which is blocked on the You Tube works

PR #3927, opened 2022-02-04
Tweak css to align bullets correctly in both standfirst and textBlockComponent

HarryFischer:
## What does this change?
Changes height and width of bullets

## Why?
They were a little too small

### Before
Standfirst: 
image

### After
Standfirst:
image

### DIFF
--- a/dotcom-rendering/src/web/components/Standfirst.tsx
+++ b/dotcom-rendering/src/web/components/Standfirst.tsx
@@ -30,9 +30,9 @@ const nestedStyles = (palette: Palette) => css`
 	li:before {
 		display: inline-block;
 		content: '';
-		border-radius: 6px;
-		height: ${space[3]}px;
-		width: ${space[3]}px;
+		border-radius: 50%;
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/Standfirst.tsx
```diff
@@ -30,9 +30,9 @@ const nestedStyles = (palette: Palette) => css`
 	li:before {
 		display: inline-block;
 		content: '';
-		border-radius: 6px;
-		height: ${space[3]}px;
-		width: ${space[3]}px;
+		border-radius: 50%;
+		height: 12px;
+		width: 12px;
```
oliverlloyd, line 35: Space.3 is 12px, I think?

oliverlloyd (approved):
Just a mild comment about maybe using `space` for 12px but, honestly, in this case where there's so many custom numbers I think it's fine.

(I'm working on the Cypress 4 test failing)

PR #3928, opened 2022-02-04
Fix caption svg alignment

HarryFischer:
## What does this change?
Aligns svgs with caption text better

## Why?
Was slightly off - should align with the cap height

### Before
image

### After
image
image

### DIFF
--- a/dotcom-rendering/src/web/components/Caption.tsx
+++ b/dotcom-rendering/src/web/components/Caption.tsx
@@ -159,7 +159,6 @@ const iconStyle = (palette: Palette) => css`
 	::before {
 		content: ' ';
 		display: block;
-		padding-top: 72.222%;
 		padding-top: ${pictureRatio}%;
 	}
 	svg {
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/components/Caption.tsx
```diff
@@ -146,20 +146,37 @@ const hideIconBelowLeftCol = css`
...
+const videoRatio = (23 / 36) * 100;
+
 const iconStyle = (palette: Palette) => css`
 	fill: ${palette.fill.cameraCaptionIcon};
 	margin-right: ${space[1]}px;
 	display: inline-block;
-	vertical-align: middle;
+	position: relative;
+	width: 1em;
+	vertical-align: baseline;
```
mxdvl, line 158: As mentioned by @JamieB-gu, it’s worth checking all browsers respect this in the same way!

mxdvl: There’s [full support for `vertical-align`, according to MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align).

PR #3930, opened 2022-02-07
Email signups in AR should take caption from the alt field instead of the caption field

georgeblahblah:
## What does this change?

* Uses the `alt` field from CAPI to render an email signup's caption, instead of the `caption` field

## Why?

* `caption` is currently unavailable in the CAPI model used by AR
* Previously, `genericEmbed` was also taking the caption content from the `alt` field


georgeblahblah:
Closing this as `caption` is being received, but currently not being decoded in the `RenderingRequestSerde`. This is caused by an outdated version of `@guardian/content-api-models` within `@guardian/apps-rendering-api-models`

alinaboghiu (approved):

PR #3935, opened 2022-02-08
Update .nvmrc to specify patch version, 14.18.3

OllysCoding:
## Should we specify patch versions in our .nvmrc? 

I think this is a worthwhile question to ask. Off the top of my head there are a few potential reasons

1. Helps keep the version of node we use locally in sync with what we use in our productions servers (which are configured separately, in amigo)
2. Ensures everyone working on the project is using the same node version (right now you could be using 14.18.0 up to 14.18.3)
3. Makes sure everyone uses node with the latest security patches

Alternatively, you could argue that since patch versions should be expected to be compatible with one-another, it's unnecessary to specify and spend the time to keep up-to-date a patch version in our nvmrc.

I'd love to hear peoples thoughts!


mxdvl (approved):
I’m a fan of pinned patched version! Most node environment tools like `fnm` or `nvm` will happily keep you on the latest version you have installed if it matches the minor version, so you might never upgrade to the latest patch.

mxdvl:
@jamesgorrie I know Renovate will keep you `.nvmrc` up to date. Don’t think I’ve seen Dependabot do it.

@oliverlloyd – I think you might have misunderstood me: the current setup means you’ll never upgrade locally. This proposal will lead to more updated versions. As we’ve chosen to stick with LTS versions of Node, the even-numbered ones, this should be kept to a minimum. Currently patches are released less than once a month for Gallium (v16 LTS):

[Node 16 releases](https://nodejs.org/download/release/)

sndrs:
👍 to this

> since patch versions should be expected to be compatible with one-another

the "should" here is the key – we've def had issues before where a patch of a dep broke builds (not node but still holds). there's not a lot of downside in pinning – the node version managers dont take long to fetch a new one even if it changed once a day – but pinning could stop things unexpectedly breaking, so why not?

JamieB-gu (approved):

PR #3936, opened 2022-02-08
Remove path aliases from DCR

sndrs:
## What does this change?

- removes the import path aliases from DCR (undoes #281 and follows https://github.com/guardian/frontend/pull/23494)

## Why?

- people find them hard to understand where they're pointing too
- editor cannot manage the paths for you when refactoring
- you can't click through into the original file
- the config has to be maintained in many places


oliverlloyd:
I think there's still some stray refs to the alias in some files (hence the test fails).

Can we add the commit hash for this change to the `.git-blame-ignore-revs` 🙏

jamesgorrie:
You still have some references [here](https://github.com/guardian/dotcom-rendering/blob/29bbdca55a9f348f52a0519e0ce535dde15ce447/.storybook/preview.js#L1-L12).

sndrs:
> Can we add the commit hash for this change to the `.git-blame-ignore-revs`

oooh i've never seen this before, that's great! yeah 100% agree, will do once it goes green

dotcom-rendering/src/amp/components/Caption.tsx
```diff
@@ -3,8 +3,8 @@ import { css } from '@emotion/react';
 
 import { text, textSans } from '@guardian/source-foundations';
 
-import { pillarPalette_DO_NOT_USE } from '@root/src/lib/pillars';
-import TriangleIcon from '@frontend/static/icons/triangle.svg';
+import { pillarPalette_DO_NOT_USE } from '../../lib/pillars';
```
sndrs, line 6: 👀

mxdvl: Could also be marked as `@deprecated`… but it won’t prevent peeps from using it 😉

jamesgorrie (approved):

PR #3937, opened 2022-02-08
`CommercialMetrics` as an island

oliverlloyd:
## What does this change?
Moves the `CommercialMetrics` component over to the island pattern using `clientOnly={true}` to ensure it is not server rendered.

## What's with `Page`?
This is a new pattern. Previously we've put all `Islands` in the various layout files and this has worked well but in this case I felt this component might live better in the high level `Page` component. My assumption is that this code should always run on every page type, _not just articles_ (same for Fronts, same for Galleries, etc) but I'm keen to get this assumption checked.


dotcom-rendering/src/web/components/Page.tsx
```diff
@@ -4,13 +4,17 @@ import { focusHalo } from '@guardian/source-foundations';
...
+import { Island } from './Island';
 
 type Props = {
 	CAPI: CAPIType;
 	NAV: NavType;
 	format: ArticleFormat;
 };
 
+const isServer = typeof window === 'undefined';
+
```
mxdvl, line 17: Do you need this check if you have the Island prop `clientOnly={true}`?

jamesgorrie (commented):
On the current frontend app - are there any exceptions to running commercial metrics?

mxdvl:
@jamesgorrie no, there isn’t. If the switch is on, and the user either randomly sampled or [in an experiment](https://github.com/guardian/frontend/blob/113172a425cc78b3a75e2a714a80a614bf6fa4b3/static/src/javascripts/projects/common/modules/analytics/shouldCaptureMetrics.ts#L18), metrics are captured.

cc @guardian/commercial-dev

dotcom-rendering/src/web/components/CommercialMetrics.tsx
```diff
@@ -8,11 +8,8 @@ import { spacefinderOkr1FilterNearby } from '@frontend/web/experiments/tests/spa
 import { useDocumentVisibilityState } from '../lib/useDocumentHidden';
 import { useAdBlockInUse } from '../lib/useAdBlockInUse';
 
-// TODO disallow undefined browserIds by placing conditional in App.tsx
-// so that we wait to render this component until browserId is defined.
-export const CommercialMetrics: React.FC<{
```
chrislomaxjones, line 13: Does this file `CommercialMetrics.tsx` need to be renamed to `CommercialMetrics.importable.tsx` to be the child of an `Island`?

dotcom-rendering/src/web/components/Page.tsx
```diff
@@ -20,6 +24,10 @@ type Props = {
  * @param {ArticleFormat} format - The format model for the article
  * */
 export const Page = ({ CAPI, NAV, format }: Props) => {
+	const renderCommercialMetrics =
+		!isServer &&
+		window.guardian.config?.ophan !== undefined &&
+		CAPI.config.switches.commercialMetrics;
```
chrislomaxjones, line 30: Please correct me if I've misunderstood, but is the `Page` component SSR only? Which would mean `renderCommercialMetrics` is only evaluated server side, and therefore always `false`, which means that the commercial metrics island isn't rendered?

dotcom-rendering/src/web/components/CommercialMetrics.tsx
oliverlloyd: This is a really good point. I'm going to step back and look at if there are better ways to handle this situation as there could be higher level solutions we want to introduce.

dotcom-rendering/src/web/components/Page.tsx
oliverlloyd: This is a really good point. I'm going to step back and look at if there are better ways to handle this situation as there could be higher level solutions we want to introduce.

dotcom-rendering/src/web/components/CommercialMetrics.tsx
oliverlloyd: Yes, great spot!

chrislomaxjones (approved):

dotcom-rendering/src/web/components/Page.tsx
chrislomaxjones: Looks great, thanks!

mxdvl:
Is `Islands` the right pattern for this?

Maybe moving this to an init script like we did for [coreVitals](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/browser/coreVitals/init.ts) would be more adequate.

oliverlloyd:
> Is `Islands` the right pattern for this?
> 
> Maybe moving this to an init script like we did for [coreVitals](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/browser/coreVitals/init.ts) would be more adequate.

Yeah, I definitely see what you're saying. Are we only using react here because that was how we were able to run javascript but actually we don't need it and could just use plain js? The use of `useOnce` is also suggestive of this.

We would need a way to execute the AB test but that should be doable using [ab-core](https://github.com/guardian/ab-testing/tree/main/packages/ab-core)?

But maybe we want to consider [this approach](https://github.com/guardian/gateway/pull/1513/files#diff-c872bfa40c03f0062727781fff912303b62860942cecd8e1cb7fac615ccaad76R77-R80) that Identity are using? They are [using islands](https://github.com/guardian/gateway/pull/1513) for both React _and_ pure js 😻

oliverlloyd:
Oh, and the other hook that this code uses `useAdBlockInUse` is also just a reactified wrapper around a pure js function (similar to ab-react) so it's not a requirement to use it. Not in the way that SWR is.

mxdvl:
Yes excatly!

PR #3942, opened 2022-02-08
`MostViewedFooter` as an island

oliverlloyd:
## What does this change?
Moves `MostViewedFooter` into an island

## Why?
Less javascript on initial page load

## Why `MostViewedFooterLayout`?
I added this extra component so that we can have something to wrap with `WithABProvider` . The `ABProvider` component must wrap the component that wants to use `useAB` which in this case was happening at the root (in `MostViewedFooter`) so in this case we needed an extra layer of composition

## Do we now need to manually insert `WithABProvider`?
This code introduces a pattern where we're asking developers to wrap their client side code with `WithABProvider` where needed. But maybe we can incorporate this into the Islands abstraction itself? See [this discussion](https://github.com/guardian/dotcom-rendering/discussions/3955)


dotcom-rendering/src/web/components/MostViewedFooter.stories.tsx
```diff
@@ -51,6 +51,9 @@ export const withTwoTabs = () => {
 					}}
 					sectionName="politics"
 					ajaxUrl="https://api.nextgen.guardianapps.co.uk"
+					switches={{}}
+					pageIsSensitive={false}
+					isDev={false}
```
oliverlloyd, line 56: I''m passing these in to be used by `WithABProvider`

dotcom-rendering/src/web/components/MostViewedFooter.test.tsx
```diff
@@ -2,16 +2,16 @@ import { render, fireEvent } from '@testing-library/react';
 
 import { ArticleDesign, ArticleDisplay, ArticlePillar } from '@guardian/libs';
 
-import { useApi as useApi_ } from '../../../lib/useApi';
-import { decidePalette } from '../../../lib/decidePalette';
+import { useApi as useApi_ } from '../lib/useApi';
```
oliverlloyd, line 5: There's a bunch of these import changes which are merge conflict hangovers

dotcom-rendering/src/web/components/SecondTierItem.tsx
```diff
@@ -8,11 +8,11 @@ import {
 	from,
 } from '@guardian/source-foundations';
```
DavidLawes, line 10: With regards to a flat file structure, our hands are tied by the use of the `.importable` pattern. Now we've moved this to the root of components, is it worth considering renaming this file to something like `MostViewedFooterSecondTierItem` so it's clear how this component is used/intended to be used?

dotcom-rendering/src/web/components/MostViewedFooterLayout.tsx
```diff
@@ -130,15 +118,11 @@ export const MostViewedFooter = ({ sectionName, format, ajaxUrl }: Props) => {
 						<Hide when="above" breakpoint="leftCol">
 							<h2 css={headingStyles}>Most popular</h2>
 						</Hide>
-						<Lazy margin={300}>
```
DavidLawes, line 133: Sorry - I can't work out what happens with this lazy margin + suspense, is it included somewhere else or not needed at all?

DavidLawes (approved):

dotcom-rendering/src/web/components/MostViewedFooter.importable.tsx
```diff
@@ -0,0 +1,34 @@
...
+			abTestSwitches={switches}
+			pageIsSensitive={pageIsSensitive}
+			isDev={!!isDev}
+		>
+			<MostViewedFooterLayout
+				sectionName={sectionName}
+				format={format}
+				ajaxUrl={ajaxUrl}
+			/>
+		</WithABProvider>
```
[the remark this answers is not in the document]
jamesgorrie: Is this a pattern we will continue to see then and could we look at adding this to the islands code? e.g. All islands have the same as the Boot context?

(another PR / discussion maybe)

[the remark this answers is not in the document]
oliverlloyd: I considered this approach (I'll update the PR description to include this) but it turned out to make the inclusion of this AB code dynamic (only when needed/wanted) I would have needed to change the code both in the `Island` server side component and also the `doHydration` client side code. It felt like adding more coupling to this for what is a secondary concern wasn't the right thing to do.

The alternative to the dynamic coupling is to just always set it. Which is an option. I guess the discussion is around the trade off between optimisation and dev ex. I think this is a worthwhile discussion to have so I'll raise it 👍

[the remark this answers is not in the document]
oliverlloyd: Discussion: https://github.com/guardian/dotcom-rendering/discussions/3955

dotcom-rendering/src/web/components/SecondTierItem.tsx
oliverlloyd: Thank you for seeing this! I'll do this now

mxdvl:
🏝️  #3629

PR #3943, opened 2022-02-08
Smaller toggle click area

joecowton1:
## What does this change?

Fixes a bug where the toggle label was stretching to full width.

### Before
Screenshot 2022-02-08 at 15 03 25

### After

Screenshot 2022-02-08 at 15 02 45

### DIFF
--- a/dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
+++ b/dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
@@ -12,7 +12,7 @@ const cssOverrides = css`
 	}
 `;
 
-const spanStyles = css`
+const toggleWrapperStyles = css`
 	display: flex;
 
 	${from.desktop} {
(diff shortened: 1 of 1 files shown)


oliverlloyd:
I should also point out the PR description is aces!

dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
```diff
@@ -31,7 +39,7 @@ export const FilterKeyEventsToggle = ({ filterKeyEvents }: Props) => {
 
 	return (
 		<>
-			<span id="filter-toggle" />
+			<span id="filter-toggle" css={spanStyles} />
```
oliverlloyd, line 42: Can we not add styles to this `span`? As I understand it, this element purely exists to enable us to scroll the reader to this position but if we're not adding css to this it creates a coupling we probably don't want.

joecowton1:
> Can I suggest a more declarative PR naming convention here, maybe `Smaller toggle click area` or something like that. These documents form part of a large git history that is read a lot more than you'd think and it's useful if the _intentions_ are clearly expressed.

Yeah good call, done 👍

oliverlloyd (changes requested):
The effect we want is being created here but is there another way to achieve it rather than using this `span`?

PR #3944, opened 2022-02-08
Fetch Full Git History For Chromatic Action

JamieB-gu:
## Why?

Version 2 of the checkout action now only fetches a single commit: https://github.com/actions/checkout/releases/tag/v2.0.0

Chromatic requires more history than that to perform its snapshot comparisons. The suggested solution is to set `fetch-depth` to `0`: https://www.chromatic.com/docs/github-actions#support-for-codeactionscheckoutv2code

## Changes

- Set checkout action fetch depth to 0 for Chromatic workflow


oliverlloyd (approved):
Yup

PR #3945, opened 2022-02-08
`WithABProvider`

oliverlloyd:
## What does this change?
This refactors how we set the `ABProvider` by moving it into a `WithABProvider` wrapper component

## Why?
So that it will be easier to set more of these providers in future as we move over to islands

### DIFF
--- a/dotcom-rendering/src/web/components/WithAB.tsx
+++ b/dotcom-rendering/src/web/components/WithAB.tsx
@@ -1,0 +1,42 @@
+import { CoreAPIConfig, OphanAPIConfig } from '@guardian/ab-core/dist/types';
+import { ABProvider } from '@guardian/ab-react';
+import { getCypressSwitches } from '../experiments/cypress-switches';
+import { getForcedParticipationsFromUrl } from '../lib/getAbUrlHash';
+
+type Props = {
+	arrayOfTestObjects: CoreAPIConfig['arrayOfTestObjects'];
+	abTestSwitches: CoreAPIConfig['abTestSwitches'];
+	pageIsSensitive: CoreAPIConfig['pageIsSensitive'];
+	mvtId: CoreAPIConfig['mvtId'];
+	ophanRecord: OphanAPIConfig['ophanRecord'];
+	children: JSX.Element;
+};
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/WithAB.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+type Props = {
+	arrayOfTestObjects: CoreAPIConfig['arrayOfTestObjects'];
+	abTestSwitches: CoreAPIConfig['abTestSwitches'];
+	pageIsSensitive: CoreAPIConfig['pageIsSensitive'];
+	mvtId: CoreAPIConfig['mvtId'];
+	ophanRecord: OphanAPIConfig['ophanRecord'];
+	children: JSX.Element;
+};
+export const WithAB = ({
```
jamesgorrie, line 14: nit: would this be more explicit as `WithABProvider`?

dotcom-rendering/src/web/components/WithAB.tsx
```diff
@@ -0,0 +1,42 @@
...
+	arrayOfTestObjects,
+	abTestSwitches,
+	pageIsSensitive,
+	mvtId,
+	ophanRecord,
+	children,
+}: Props) => {
+	const windowHash = window && window.location && window.location.hash;
```
jamesgorrie, line 22: Easy refactor?

```suggestion
	const windowHash = window?.location.hash;
```

jamesgorrie (approved):
The code looks good. Is the main reason to encapsulate the logic?
```
const cypressAbSwitches = getCypressSwitches();
const windowHash = window && window.location && window.location.hash;
```

Otherwise it does just _appear_ to be wrapping a Provider in a `FunctionComponent`?

oliverlloyd:
> Is the main reason to encapsulate the logic?

Essentially, yes. Once we start moving components that have AB tests into islands then they will either need to set their own `ABProvider` or we abstract that somehow and I want that experience to be as seamless as possible. I'm not yet 100% sure the best way to offer that api but this seems a good first step.

PR #3706, opened 2021-11-30
feat: Use an absolute Date for KeyEvents

oliverlloyd:
## What does this change?
Replaces `time: string` with `date: Date` on `KeyEvents`

## Why?
So that we can use `timeAgo` to display the relative time and also setup the required attributes on this `time` element to keep it accessible and live. See: #3704 

### Before
Screenshot 2021-11-30 at 13 35 43

### After
Screenshot 2021-11-30 at 13 17 36

PR #3946, opened 2022-02-08
More Recent Chromatic Mock Date

JamieB-gu:
## Why?

We set an artificial "current" datetime in Chromatic to make sure any UI snapshots that rely on it appear static.

When using `timeAgo` (#3706), we attempt to calculate the difference in time between a given timestamp and "now". Unfortunately, when the timestamp is more recent than "now" this causes problems because the difference in time is negative (which isn't possible in this context). This issue has recently started breaking the UI on some of the liveblog snapshots.

There are a few ways to solve this (including updating the fixtures that set the liveblog timestamps). However the simplest is probably just to update the artificial datetime in Chromatic to be more recent, which is what this PR does.

## Changes

- Updated the Chromatic mock date to be more recent

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -17,9 +17,7 @@ Picture.disableLazyLoading = isChromatic();
 
 if (isChromatic()) {
 	// Fix the date to prevent false negatives
-	// And not just any date... 200 years! 🎉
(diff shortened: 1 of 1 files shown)


AshCorr (approved):
LGTM.

I wonder if we can use the `[data-chromatic="ignore"]` attribute to tell Chromatic to ignore the spans containing date fields? https://www.chromatic.com/docs/ignoring-elements

PR #3952, opened 2022-02-09
Add `SpacefinderOkr2ImagesLoaded` test

simonbyford:
## What does this change?

Introduces an A/B test in DCR which can be used to measure the impact of this change: https://github.com/guardian/frontend/pull/24610

## Why?


ashishpuliyel (approved):
Looks good.

PR #3954, opened 2022-02-10
Send today's article count and contentType in banner requests

tomrf1:
SDC PR: https://github.com/guardian/support-dotcom-components/pull/624

This PR makes the client send 2 new fields in the targeting payload to SDC for banners:
- contentType
- articleCountToday (if available)

Screen Shot 2022-02-10 at 12 03 09


dotcom-rendering/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -41,13 +42,13 @@ type BaseProps = {
 	alreadyVisitedCount: number;
 	engagementBannerLastClosedAt?: string;
 	subscriptionBannerLastClosedAt?: string;
-	weeklyArticleHistory?: WeeklyArticleHistory;
```
tomrf1, line 44: this field isn't used

dotcom-rendering/src/web/lib/dailyArticleCount.ts
```diff
@@ -12,11 +12,12 @@ interface DailyArticleCountLocalStorage {
 
 export const DailyArticleCountKey = 'gu.history.dailyArticleCount';
 
-export const getDailyArticleCount = (): DailyArticleCount => {
+// Returns undefined if no daily article count in local storage
+export const getDailyArticleCount = (): DailyArticleCount | undefined => {
```
tomrf1, line 16: I changed the return type here because `undefined` is a more deliberate signal that there is _no_ article count. This is consistent with the weekly article count.

tomrf1:
> Could we move this code out of `App.tsx` and down into the respective components? We're about to delete `App.tsx` and so we don't want to add any more logic to this file.
> ### Wait, what? Deleting `App`, what?!
> 
> There are documents [here](https://docs.google.com/document/d/1kSrZZrJ8hSw6bek5hGa9onE8QDz5TRHFqHV52EOFjts/edit#heading=h.gsod33a0pc47) and [here](https://docs.google.com/document/d/1gHOzvFZxrAO2mvNPKg4UmMAYZm7Jk1NutZPVLY5UxPY/edit) explaining the reasoning and [here](https://github.com/guardian/dotcom-rendering/pull/3629) is a PR introducing Islands, the replacement for `App`.
> ### So what do I do with my React state?
> 
> We moving global state down into local components. This means where you had shared global App state it should be duplicated and moved into the specific components. There are examples of this in [other PRs](https://github.com/guardian/dotcom-rendering/pulls?page=2&q=is%3Apr+is%3Aclosed+island)
> ### I'm still confused
> 
> It's just confusing. Please do reach out to the team; we're extremely happy to pair on this!

I've now merge in main, which means the AC code is no longer in App.tsx.
I've refactored the `getArticleCount` function to return both weekly and daily ACs.

dotcom-rendering/src/web/components/SignInGate/displayRule.ts
```diff
@@ -12,7 +12,7 @@ import { CanShowGateProps } from './types';
 // in our case if this is the n-numbered article or higher the user has viewed then set the gate
 export const isNPageOrHigherPageView = (n: number = 2): boolean => {
 	// get daily read article count array from local storage
-	const [dailyCount = {} as DailyArticle] = getDailyArticleCount();
+	const [dailyCount = {} as DailyArticle] = getDailyArticleCount() || [];
```
tomrf1, line 15: I changed the return type here to `DailyArticleHistory | undefined` because `undefined` is a more deliberate signal that there is _no_ article count. This is consistent with the weekly article count.

dotcom-rendering/src/lib/article-count.ts
```diff
@@ -5,29 +5,50 @@ import {
...
+	if (!window.guardian.dailyArticleCount) {
+		incrementDailyArticleCount();
+		window.guardian.dailyArticleCount = getDailyArticleCount();
 	}
 
-	return window.guardian.articleCount;
+	return {
+		weeklyArticleHistory: window.guardian.weeklyArticleCount,
+		dailyArticleHistory: window.guardian.dailyArticleCount,
+	};
```
jamesgorrie, line 53: Why couple these two and not just use `getWeeklyArticleCount` and `getDailyArticleCount` separately as two different pieces of state?

A bit of my confusion is that I would expect, consent isn't given to return `undefined`, rather than `EmptyArticleCounts`. Decoupling them might help with confusion.

I can see benefit is we used them together, but we don't seem to anywhere in the code, rather we just destructure this type.

dotcom-rendering/src/web/components/SignInGate/displayRule.ts
jamesgorrie: Just to check thinking here:
`undefined = 0`? A thought on this is that `undefined` is also used to indicate that a reader hasn't given consent? Are we happy saying those are the same things?

jamesgorrie (approved):

dotcom-rendering/src/web/components/SignInGate/displayRule.ts
tomrf1: `getDailyArticleCount` returns `undefined` if there is no article count in local storage, which would be because they haven't consented. On the Supporter Revenue side it's good to be clear about this.
While doing this PR I discovered that the SignInGate (this file) is also using article count - so I've kept the logic the same as before here by defaulting to `[]`.

dotcom-rendering/src/lib/article-count.ts
tomrf1: Good question, I spent a while thinking about this!
The banner does use both counts, and so it's quite convenient to pass them through together (from `StickBottomBanner`, down a few function calls) here: https://github.com/guardian/dotcom-rendering/pull/3954/files#diff-9fb4db8893a570211b30acefe9224ff36a020986088db2ae69a32dc36723c6e7R235

Also it also felt natural to keep the initialisation of these 2 counts together in this one function because it all depends on awaiting `hasOptedOutOfArticleCount`.
But - I'm not wedded to this implementation if you think it's preferable to split them?

jamesgorrie: Sorry, I think I might have conflated things. 

I am confused by this method return type as it stands. If someone has not consented, I would expect to get `undefined`. Instead I get an object with `undefined` prop values, which feels like they mean that they are available, just not set. So, in my head the return type of `ArticleCounts | undefined` might make more sense.

Some tests might make this more understandable.

On your point re: async `hasOptedOut` -> this is true, dotcom could potentially have a better way of sharing this state. Given that that is a potential performance issue, we could just leave it as 1.

jamesgorrie (approved):
I think this PR surfaces that how we share async state could use some work. I also think it's outside of the scope of this PR.

There's a question on the return type of `getArticleCounts`.

dotcom-rendering/src/lib/article-count.ts
tomrf1: Ah right I see what you mean, I'll have a go at refactoring to return `ArticleCounts | undefined`

tomrf1: I've just pushed this up

tomrf1:
> `getArticleCounts`

Thanks for the review! I think I have at least partially addressed the question about `getArticleCounts` in the last commit

PR #3958, opened 2022-02-10
`GetMatchNav` as an island

oliverlloyd:
## What does this change?
`GetMatchNav` > 🏖️ 

## Placeholder?
We're using the `placeholderHeight` prop here to ensure something is rendered on the server to reserve the space for the client side component. It's grey and shimmers. There's a ticket for more options on this component (colour, less shimmer, etc.)


mxdvl:
🏝️  #3629

PR #3959, opened 2022-02-10
`ShareCount` as an island

oliverlloyd:
## What does this change?
Moves `ShareCount` over to an island

## Why?
To remove this javascript from the page if it isn't being used

## Is it ever used?
Well, no. Not at the moment and probably not ever again. We will look at removing this code completely (and in all layers) as a separate task


oliverlloyd:
Closing in favour of removing this code

mxdvl (approved):
Yes–but let’s remove shortly 🧹 !

PR #3961, opened 2022-02-10
Move `FocusStylesManager` into an island

oliverlloyd:
## What does this change?
Moves `FocusStylesManager` into it's own component and then into an island

## Why?
We want to move all code out of App.tsx


ioannakok (approved):
👍

PR #3962, opened 2022-02-10
Webpack with TS & ES Modules

mxdvl:
## What does this change?

This is an attempt at adding Typescript types to webpack, and trying to load the configuration with `ts-node`.

## Why?

Easier to know what the shape of functions are meant to be!

## Next steps

Whilst this was a useful learning experience, I think simply adding some TSDoc types will be a sufficient option. The tooling for running TS is still in its infancy.


mxdvl:
Closed in favour of #3967

PR #3963, opened 2022-02-10
`Embeds` as islands

oliverlloyd:
## What does this change?
`Embeds` as islands

## Why?
Reduced js on the page


dotcom-rendering/cypress/integration/parallel-6/commercial.spec.js
```diff
@@ -30,19 +28,13 @@ describe('Commercial E2E tests', function () {
...
-
-	describe('Ad slot Parity between DCR and Frontend for a long read', function () {
-		it(`It should check slots for a long article in DCR`, function () {
-			runLongReadTestFor(`Article?url=${longReadURL}`);
-		});
-
-		it(`It should check slots for a long article in Frontend`, function () {
-			Cypress.config('baseUrl', '');
-			runLongReadTestFor(`${longReadURL}?dcr=false`);
-		});
```
oliverlloyd, line 46: I removed the Frontend test here because we no longer use Frontend!

dotcom-rendering/cypress/integration/parallel-6/commercial.spec.js
```diff
@@ -30,19 +28,13 @@ describe('Commercial E2E tests', function () {
 			});
 
 		cy.get(`[data-name="right"]`).should('have.length', 1);
+		cy.scrollTo('bottom');
 		cy.get(`[data-name="merchandising-high"]`).should('have.length', 1);
+		// Wait for hydration
+		cy.get('gu-island[name=MostViewedFooter]')
```
DavidLawes, line 34: I don't understand (help!) - how do embeds impact these tests?

dotcom-rendering/cypress/integration/parallel-1/article.elements.spec.js
```diff
@@ -106,6 +106,13 @@ describe('Elements', function () {
 				'Article?url=https://www.theguardian.com/sport/blog/2015/dec/02/the-joy-of-six-sports-radio-documentaries',
 			);
 
+			cy.scrollTo(0, 4500);
```
DavidLawes, line 109: We're scrolling to a specific pixel position - will this always yield the next expected action (which I'm assuming scrolls something into view that then gets hydrated)?

DavidLawes (approved):
👍

dotcom-rendering/cypress/integration/parallel-1/article.elements.spec.js
oliverlloyd: This is the expected behaviour, yes. Once our element is in view it becomes hydrated and we can continue the test verifying it does what we expect. The risk is content on this article moves around but that's unlikely (I hope!)

mxdvl:
🏝️  #3629

PR #3967, opened 2022-02-11
Convert Webpack config to Typescript

mxdvl:
## What does this change?

Converts the existing Webpack configuration to Typescript and ES Modules, [which is supported by Webpack][wp].
Convert Storybook config to Typescript, [which is supported by Storybook][sb].

Remove `@types/webpack` as it’s now included in `webpack` itself, since v5
Add some custom declarations for plugins which are failing tests or missing declarations:
- `webpack-messages`
- `webpack-filter-warnings-plugin`
-  `@loadable/webpack-plugin` – [awaiting PR upstream](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/58719)

[wp]: https://webpack.js.org/configuration/configuration-languages/#typescript
[sb]: https://storybook.js.org/docs/react/configure/overview#configure-your-project-with-typescript

## Why?

Get type safety in our configuration. AR does it already.


mxdvl:
Closed in favour of #3967

mxdvl:
Whoops, meant to close #3962

dotcom-rendering/package.json
```diff
@@ -142,15 +142,17 @@
...
     "@types/lodash.clonedeep": "^4.5.4",
     "@types/lodash.get": "^4.4.4",
-    "@types/node": "^14.14.24",
+    "@types/node": "^17.0.17",
     "@types/raven-js": "^3.10.0",
     "@types/react": "17.0.3",
     "@types/react-dom": "17.0.3",
     "@types/react-test-renderer": "17.0.1",
     "@types/response-time": "^2.3.4",
-    "@types/webpack": "^4.41.26",
```
JamieB-gu, line 153: Because the main `webpack` package bundles these now?

dotcom-rendering/scripts/webpack/utils.ts
```diff
@@ -0,0 +1,12 @@
+import type { WebpackPluginInstance, Compiler, Configuration } from 'webpack';
+
+type ValidPlugin =
+	| WebpackPluginInstance
+	| ((this: Compiler, compiler: Compiler) => void);
+
+export const isWebpackPluginInstance = (p: unknown): p is ValidPlugin =>
+	p !== false;
```
JamieB-gu, line 8: Is it worth using `instanceof` here? Or is checking against `false` enough?

dotcom-rendering/scripts/webpack/webpack.config.server.ts
```diff
@@ -35,31 +39,31 @@ module.exports = ({ sessionId }) => ({
...
-			new GuStatsReportPlugin({
+			(new GuStatsReportPlugin({
 				displayDisclaimer: true,
 				buildName: 'server',
 				project: 'dotcom-rendering',
 				team: 'dotcom',
 				sessionId,
-			}),
-	].filter(Boolean),
+			}) as WebpackPluginInstance),
```
JamieB-gu, line 65: `TODO: remove assertion`?

dotcom-rendering/scripts/webpack/webpack.config.ts
```diff
@@ -63,9 +74,10 @@ const commonConfigs = ({ platform }) => ({
 		DEV &&
 			new WebpackMessages({
 				name: platform,
-				logger: (message) => {
+				logger: (message: string) => {
 					// distinguish between initial and subsequent (re)builds in console output
-					if (builds < module.exports.length * 2) {
+					// eslint-disable-next-line @typescript-eslint/no-use-before-define
+					if (builds < configs.length * 2) {
```
JamieB-gu, line 80: Is it `configs` that's being used before definition here?

JamieB-gu (commented):
Nice! Excellent use of type predicates and `filter` 🙂

dotcom-rendering/scripts/webpack/webpack.config.ts
mxdvl: yes, indeed, and it itself relies on `commonConfigs`, so you can’t satisfy both rules here.

dotcom-rendering/tsconfig.json
```diff
@@ -47,6 +47,9 @@
             /* Aliases should also be added to the webpack, babel and jest configurations */
             "*": ["node_modules/@types/*", "*"], // Make sure that yarn linking doesn't confuse things https://github.com/microsoft/TypeScript/issues/11916#issuecomment-257130001
         },
+
+        "skipLibCheck": true,
```
sndrs, line 51: is this ok? this make TS more permissive "the expense of type-system accuracy", as the docs say. do we want that?

dotcom-rendering/scripts/webpack/webpack.config.browser.ts
```diff
@@ -40,18 +58,19 @@ module.exports = ({ isLegacyJS, sessionId }) => ({
...
+			(new GuStatsReportPlugin({
 				buildName: isLegacyJS ? 'legacy-client' : 'client',
 				project: 'dotcom-rendering',
 				team: 'dotcom',
 				sessionId,
-			}),
-	].filter(Boolean),
+				// TODO: remove assertion
+			}) as WebpackPluginInstance),
+	].filter(isWebpackPluginInstance),
```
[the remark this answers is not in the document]
mxdvl: `Boolean` is using the truthiness of objects, which is much less explicit. Type predicates are the correct way to let Typescript know what the type of something will be.

dotcom-rendering/tsconfig.json
mxdvl: I guess I’d like not to…

dotcom-rendering/scripts/webpack/webpack.config.browser.ts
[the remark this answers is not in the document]
sndrs: is it the correct way to let human's know the intention of this code though? 😉 that's what i'm wondering. 

_really_, these are conditionals presented in an already obtuse way:

```js
const config = {
	plugins: [
		DEV && myPlugin
	].filter(Boolean)
}
```

really just means

```js
if (DEV) {
	(config.plugins ||= []).push(myPlugin)
}
```

with this change, someone who is trying to make changes to the config will need to not only understand what is going on with that filter, but also what `isWebpackPluginInstance` is telling them.

personally i'd be tempted to read the code to make sure the name was accurate. 

at which point i discover that the _intention_ here is to respond to `NODE_ENV`, but there's a whole abstracted TS implementation providing a complete definition and test for the validity of a webpack plugin to achieve it.

which is all to say, what the current code does is change based on `NODE_ENV`. what this does still achieves that, but i think at the cost of intelligibility – i'm not sure that's a net gain?

dotcom-rendering/scripts/webpack/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+	| DefinePlugin
+	| FilterWarningsPlugin
+	| LoadablePlugin
+	| IgnorePlugin
+	| BundleAnalyzerPlugin
+	| WebpackMessages;
+
+export const isWebpackPluginInstance = (
+	p: false | ValidPlugin,
+): p is WebpackPluginInstance => p !== false;
```
sndrs, line 23: what if the plugin doesn't have types, is this still true? what does this result in?

```ts
[Math.random()].filter(isWebpackPluginInstance)
```

dotcom-rendering/scripts/webpack/webpack.config.browser.ts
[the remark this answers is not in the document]
mxdvl: Yes! I agree that making these conditionals clearer would be ace.

However, this PR is already overflowing in all sorts of ways, and changing over 50% of a file will break following of git history in Github, so I’m tempted to leave this to a subsequent refactor. I think my favourite approach for conditionals is probably:

```ts
const pluginsDEV: WebpackPluginInstance[] = [ /* … */ ]
const pluginsPROD: WebpackPluginInstance[] = [ /* … */ ]

config = {
  /* … */
  plugins: DEV? pluginsDEV : pluginsPROD,
  /* … */
}
```

dotcom-rendering/scripts/webpack/utils.ts
mxdvl:
```
Type 'number' is not assignable to type 'false | ValidPlugin'
```

dotcom-rendering/scripts/webpack/webpack.config.browser.ts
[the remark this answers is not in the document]
sndrs: so is typescript complaining if you only filter on truthiness?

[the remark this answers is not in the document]
mxdvl: Raised #3993 as one of the key benefits we gained from this investigation

sndrs:
do you what effect this has on the time to run tasks? i.e. how much time is spent transpiling in `ts-node`?

jamesgorrie:
There's a reference to the JS [here](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/package.json#L31)> breaking the build. I tried changing it, but it started to be a long string so stopped.

mxdvl:
> do you know what effect this has on the time to run tasks? i.e. how much time is spent transpiling in `ts-node`?

This is an excellent question, actually. Locally I got similar times, but not sure how we’d go about measuring the change.

sndrs:
you can use `time` – so `time make build` etc

mxdvl:
Seems like it adds a few seconds!

```sh
# Javascript
make build  213.03s user 11.80s system 360% cpu 1:02.34 total
make build  204.18s user 11.99s system 363% cpu 59.465 total
make build  201.53s user 11.85s system 360% cpu 59.253 total

# Typescript (3 tries)
make build  224.59s user 14.81s system 354% cpu 1:07.60 total
make build  220.38s user 13.88s system 347% cpu 1:07.46 total
make build  213.59s user 13.18s system 343% cpu 1:06.04 total
```

sndrs:
ok so assuming an extra 5s on 10 builds a day, this change would consume 3h 20m of a developer's time per year. 

considering this (and i know the pain and effort it took to get this PR together 🙏) what would you say are the benefits of type safety in these files? does the cost of using TS here include known tangible benefits?

oliverlloyd:
> would consume 3h 20m of a developer's time per year

I can't speak to the wider aspects of this PR but It's probably worth considering that `make build` is used in our CI services and is how we test locally with Cypress so any slow down there would impact these things too.

mxdvl:
Alright, you too… Closing in favour of #4015 and #4017.

PR #3968, opened 2022-02-11
`TopRightAdSlot`

oliverlloyd:
## What does this change?
This PR moves shady pie into an island

## What is shady pie?
This is the name given to the contribution slot which gets shown in the top right ad slot when an ad blocker is detected (plus some other criteria).

In this PR we're creating a pure `ShadyPie` component which encapsulates what gets show. We then moved the logic to decide when to show it or not into `TopRightAdSlot`. (I really wanted to call this component `MaybeShady` but in truth what it is actually doing is rendering the top right ad slot and then sometime this slot is the contribution - shady pie - image.)

### But now we have ad slot html in two different files, boo
I agree, it's not ideal that this html exists in multiple locations as this will make refactors harder later. I did try using `children` to keep the `id="dfp-ad--right"` div co-located in `AdSlot.tsx` however this introduced an unwanted side effect. The whole slot html and css was being serialised and inserted into the `gu-island` dom element. This is sub optimal and I felt that the trade off between this performance hit and the fact the code is broken apart was worth it.


mxdvl:
I think we want to _always_ render the ad slot. We conditionally also add the shady pie.

> The whole slot html and css was being serialised and inserted into the `gu-island` DOM element.

Not rendering this ad slot server side may have other consequences. I think this would be preferable to an approach where the slot is rendered at _some unknown time_.

sndrs (approved):

sndrs (approved):

oliverlloyd:
> I think we want to _always_ render the ad slot

We do. The default top right as slot is always rendered server side and only replace with shady pie on the client in the event that we detect an adblocker

mxdvl (commented):
Okay, I get it now: the ad slot is rendered on the server, then hydrated on the client. There’s no risk of it being missing by the time commercial code runs.

mxdvl (approved):

mxdvl:
🏝️  #3629

PR #3972, opened 2022-02-11
Deprecate `themeStyles`

JamieB-gu:
## Why?

We'd like to move over to the more semantic `editorialPalette` module in `common-rendering`. It offers a much larger, more comprehensive palette, with more semantic naming conventions, and the plan is to share it across multiple platforms in the future.

This places deprecation warnings on the `themeStyles` functions, to warn developers not to use them in new code.

## Changes

- Added deprecation warnings to `themeStyles` functions and recommended use of `editorialPalette` instead


ioannakok (approved):
🎉

PR #3974, opened 2022-02-11
Remove `@babel/preset-modules`

arelra:
## Why?

`@babel/preset-modules` was deprecated by an [option](https://babeljs.io/docs/en/babel-preset-env#bugfixes) in `@babel/preset-env` .

It looks like the webpack loaders were updated to reflect this.

However the dependency and outdated comments relating to `preset-modules` are still present.


mxdvl (approved):
yes please

PR #3975, opened 2022-02-11
Move dev dependencies to `devDependencies`

arelra:
## Why?

Mainly to reduce the noise in dependencies and increase the clarity of runtime dependencies.

Probably minimal performance improvement on a fresh production install, but in theory it should be a _tiny_ bit quicker.


mxdvl (approved):
you’re a champ!

arelra:
> you’re a champ!

<3 tis a small thing. I think I have `package.json` OCD.

PR #3976, opened 2022-02-11
Fix the way we remove placeholders post-render

jamesgorrie:
## What does this change?
- moves `data-name=placeholder` to root of `Placeholder`
- refactors `Portal` code to remove the root element's child because of ☝️ 
- allows us to hydrate conditionally in `doHydration` so we can `render` instead of `hydrate` if we're `clientOnly` as we shouldn't be hydrating non SSRed content

## Why?
We're seeing the way placeholders are found and removed to be buggy. This is a little more specific and easy to understand. And it fixes the bug. 

This could probably use a little tidy up, but mostly round the `Portal` code that we're removing anyway.

### Before
Screenshot 2022-02-11 at 15 12 50

### After
Screenshot 2022-02-11 at 15 13 35

### DIFF
--- a/dotcom-rendering/src/web/browser/islands/doHydration.ts
+++ b/dotcom-rendering/src/web/browser/islands/doHydration.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
-import { hydrate, h } from 'preact';
+import { hydrate, render, h } from 'preact';
 import { initPerf } from '../initPerf';
 
 /**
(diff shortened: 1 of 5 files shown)


OllysCoding (approved):
Paired on this, +1

dotcom-rendering/src/web/browser/islands/doHydration.ts
```diff
@@ -26,7 +26,15 @@ export const doHydration = (name: string, data: any, element: HTMLElement) => {
 		`../../components/${name}.importable`
 	)
 		.then((module) => {
-			hydrate(h(module[name], data), element);
+			const clientOnly = element.getAttribute('clientOnly') === 'true';
```
oliverlloyd, line 29: We can do this in another PR but ideally this value would be passed in as a param, from init. Just in terms of following the existing patterns

PR #3978, opened 2022-02-11
Move the `mostpop` ad slot out of its island

chrislomaxjones:
## What does this change?

Move the `mostpop` ad slot out of the island that wraps the `MostViewedFooter`.

Achieve this by:
1. Removing the importable component `MostViewedFooter`.
2. Using the `MostViewedFooterLayout` component directly.
3. Push the island further down to wrap `MostViewedFooterData` and move the cypress AB test attributes and associated logic to this level. This requires we use an `WithABProvider` within `MostViewedFooterData`.

## Why?

The `mostpop` slot is present on pages next to the "most viewed" footer at the bottom of article pages.

We require static ad slots such as `mostpop` be on the page at the point the commercial bundle attempts to fill advert slots. Since it was placed inside an island, hydration is deferred until this section is visible. Therefore it remains possible the slot isn't present at the point we fill advert slots and therefore is never filled.

By moving the `mostpop` slot out of the island, it'll be rendered server side and therefore it's guaranteed to be present on the page at the point the commercial bundle runs.


oliverlloyd (approved):
🥇

dotcom-rendering/src/web/components/MostViewedFooter.stories.tsx
```diff
@@ -3,17 +3,26 @@ import fetchMock from 'fetch-mock';
...
+const HydratedLayout = ({ children }: { children: React.ReactNode }) => {
+	useEffect(() => {
+		doStorybookHydration();
+	});
+	return <>{children}</>;
+};
 
 export default {
-	component: MostViewedFooter,
+	component: MostViewedFooterLayout,
```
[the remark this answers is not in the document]
chrislomaxjones: Good spot!

dotcom-rendering/src/web/components/MostViewedFooterData.importable.tsx
```diff
@@ -2,17 +2,29 @@ import { css } from '@emotion/react';
...
+type WithABProps = {
+	sectionName?: string;
+	palette: Palette;
+	ajaxUrl: string;
+};
 
 type Props = {
 	sectionName?: string;
 	palette: Palette;
 	ajaxUrl: string;
```
lucymonie, line 24: You could use `extends` here to reduce type duplication 🙂

dotcom-rendering/src/web/components/MostViewedFooterLayout.tsx
```diff
@@ -64,26 +63,19 @@ interface Props {
...
+	isDev?: boolean;
 }
 
 export const MostViewedFooterLayout = ({
 	sectionName,
 	format,
+	switches,
+	pageIsSensitive,
+	isDev,
 	ajaxUrl,
```
lucymonie, line 77: This is a total nitpick, but it might be good to have the props listed in the same order as the types for easier comparison.

lucymonie (approved):
This makes sense. Just a couple of tiny suggestions 👍🏻

dotcom-rendering/src/web/components/MostViewedFooterData.importable.tsx
chrislomaxjones: Great idea

PR #3980, opened 2022-02-14
Hydrate blocks

oliverlloyd:
## What does this change?
We now hydrate new live blog blocks as part of inserting them into the dom

## Why?
So that we can have interactive (uses javascript) content dynamically appear in live blogs


AshCorr (approved):
LGTM! Love the commenting!

PR #3981, opened 2022-02-14
`ShareCount` should be named using importable

oliverlloyd:
## What does this change?
This fixes an oversight in [my previous PR](https://github.com/guardian/dotcom-rendering/pull/3959) where I failed to rename the component I was using as an island

## How did this get through?
Because we don't (and won't) use this component anywhere so it's never loaded and doesn't need to be tested

## Why bother?
We're deleting this code soon anyway but we want to do that in it's own PR as part of a wider (careful) code removal exercise and until that point we should still keep things semantically correct to avoid confusion.


mxdvl (approved):
ofc

PR #3982, opened 2022-02-14
Revert single front door test

marialani:
We're now ready to end the Single Front Door test so we can revert these changes.

Reverts guardian/dotcom-rendering#3832


GHaberis (approved):

marialani:
Though we have re-added the footer it doesn't appear in the diff of the chromatic UI tests. Presumably this is the expected behaviour as the component is [lazy loaded](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/Lazy.tsx)

mxdvl:
Interestingly, it started appearing in #4126 🤷

PR #3987, opened 2022-02-14
Don't wait for hydration in Cypress commercial test

chrislomaxjones:
## What does this change?

In the commercial Cypress test, remove the step where we wait for the `MostViewedFooter` island to be hydrated before proceeding.

## Why?

Now that the `mostpop` ad slot is not in an island (#3978), there is no need to wait for this hydration step.

It also should fix the test, as this should really wait for `MostViewedFooterData` (I'm not sure why this didn't get caught in the #3978's branch Cypress run 🤔 )


oliverlloyd (approved):
👍

PR #3988, opened 2022-02-14
feat: Adds Last updated to liveblogs and deadblogs

AshCorr:
## What does this change?

- Adds a "Last updated" timestamp to the left of the standfirst.
- Updates the timestamp every time a new block is loaded.

## Before After

| Before      | After      | Frontend      |
| --------- | -------- | ------------ |
| image| image| image |
| image| image| image|


dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -609,7 +610,16 @@ export const LiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
 								/>
 							</GridItem>
 							<GridItem area="lastupdated">
-								<></>
+								{CAPI.blocks.length &&
```
AshCorr, line 613: Not sure if its actually possible to have an article with 0 blocks?

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -144,6 +149,11 @@ export const Liveness = ({
 				insertNewBlocks(data.html);
 				hydrateBlocks();
 
+				if (lastUpdated) {
+					lastUpdated.setAttribute('dateTime', new Date().toString());
```
AshCorr, line 153: Ideally this date would come from the new blocks being added, but thats not really included in the data we get from frontend. Setting the lastUpdated to Now instead _should_ hopefully be fine since we poll fairly often anyway.

dotcom-rendering/src/web/components/ArticleLastUpdated.tsx
```diff
@@ -0,0 +1,72 @@
...
+	const displayString = timeAgo(lastUpdated);
+	const date = new Date(lastUpdated);
+
+	return (
+		<div css={lastUpdatedStyles(palette)}>
+			{format.design === ArticleDesign.LiveBlog && (
+				<span css={livePulseIconStyles(palette)}>LIVE</span>
+			)}
+			&nbsp;Updated&nbsp;
+			<time
```
AshCorr, line 63: I should make this into a component.

dotcom-rendering/src/web/components/ArticleLastUpdated.tsx
```diff
@@ -0,0 +1,72 @@
...
+	font-size: 0.75rem;
+	line-height: 1rem;
+	font-family: 'Guardian Text Sans Web', 'Helvetica Neue', Helvetica, Arial,
+		'Lucida Grande', sans-serif;
+	padding-bottom: 0.125rem;
+	padding-top: 0.125rem;
+	color: ${palette.text.standfirst};
+`;
+
+const livePulseIconStyles = (palette: Palette) => css`
```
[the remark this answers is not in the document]
AshCorr: Ooh, thanks, didn't realize this existed

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -144,6 +149,11 @@ export const Liveness = ({
...
+					updateTimeElements();
```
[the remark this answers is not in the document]
AshCorr: Yep, originally I left it to be updated automatically. Felt like it looked better if it was updated at the exact same moment that the new blocks get added to the page though.

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: Hmm, that's actually a really good point. Should we display server time or client time?

I wonder what Frontend does here.

One point to consider is that the calculation by `timeAgo` is using `now` so perhaps always using local time is more accurate and less likely to result in strange results (here the reader's local time is wildly off)

dotcom-rendering/src/web/components/ArticleLastUpdated.tsx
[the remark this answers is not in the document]
oliverlloyd: Sorry, my fault. I should have include links in the ticket.

dotcom-rendering/src/web/components/Liveness.importable.tsx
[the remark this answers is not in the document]
oliverlloyd: That's actually a really interesting point. I didn't see this before but the issue is the display string of the element will no longer reflect the `dateTime` attribute value and yeah, it will basically be wrong.

You're right. This string should be updated at this point.

Maybe there's a way to update the string value on the element directly (the same as you're updating the attribute) or maybe there's a way to change the api for `updateTimeElements` to make it more targetted? You could perhaps have `updateTimeElement` (singular) or perhaps pass in an element?

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -610,7 +611,15 @@ export const LiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
...
+									CAPI.blocks[0].blockLastUpdated && (
+										<ArticleLastUpdated
```
ioannakok, line 616: Amazing work, Ash! One thing I noticed: You might want to double check how `ArticleLastUpdated` sits in the grid for 740px breakpoint. I can see a shift happening. Example stories:

Liveblog: https://www.chromatic.com/test?appId=5dfcbf3012392c0020e7140b&id=620bc34e4528a3003a62839b
Deadblog: https://www.chromatic.com/test?appId=5dfcbf3012392c0020e7140b&id=620bc34e4528a3003a62839f

AshCorr: Good catch, thanks Ionna!

dotcom-rendering/src/web/components/ArticleLastUpdated.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const ArticleLastUpdated = ({ format, lastUpdated }: Props) => {
+	const palette = decidePalette(format);
+	const displayString = timeAgo(lastUpdated);
+	const date = new Date(lastUpdated);
+
+	return (
+		<div css={lastUpdatedStyles(palette)}>
+			{format.design === ArticleDesign.LiveBlog && (
+				<span css={livePulseIconStyles}>
+					<Island>
```
oliverlloyd, line 33:
```suggestion
					<Island deferUntil="idle">
```

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -104,6 +105,10 @@ const topOfBlog: Element | null = !isServer
 	? window.document.querySelector('[data-gu-marker=top-of-blog]')
 	: null;
 
+const lastUpdated: Element | null = !isServer
+	? window.document.querySelector('[data-gu-marker=liveblog-last-updated]')
```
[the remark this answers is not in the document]
oliverlloyd: That said, I'm happy with this for now and don't want to block this PR

oliverlloyd (approved):
Really impressive stuff

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -610,7 +611,18 @@ export const LiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
 								/>
 							</GridItem>
 							<GridItem area="lastupdated">
-								<></>
+								<Hide until="desktop">
+									{CAPI.blocks.length &&
+										CAPI.blocks[0].blockLastUpdated && (
```
JamieB-gu, line 616: Is this the same as `content.fields.lastModified`?

AshCorr: When we load new blocks we don't get access to the full CAPI body, just the new blocks, so we won't have access to content.fields.lastModified. I think perhaps we should keep using blockLastUpdated because of that.

PR #3993, opened 2022-02-15
Remove filters from Webpack configs

mxdvl:
## What does this change?

Removes the strange `.filter(Boolean)` used in the Webpack configs for a more idiomatic ternary.

## Why?

Preps the changes coming in #3967, and ensures git history is kept.


jamesgorrie (approved):
✨

PR #3999, opened 2022-02-15
Ignore .env

mxdvl:
## What does this change?

Ignore .env

## Why?

If someone create the file to test chormatic locally, they might share private data by mistake


oliverlloyd (approved):
💯

PR #4001, opened 2022-02-15
Put all npm dependencies in `dependencies`

sndrs:
## What does this change?

collapses each project's npm dependencies into the `dependencies` field

## Why?

- these are application not libraries, the distinction doesn't affect us or our users
- as evidenced by the types of packages in `dependencies` and `devDependencies` of each package, it confuses us
- it makes it hard to know what is installed where, because the alpha sorting is spread across 10s of lines in 2 groups

_Prep work for #3998_


jamesgorrie (approved):
Given we don't use `yarn install --production` anywhere, as I can see, we weren't getting any benefit anywho.

oliverlloyd (approved):
Simpler

sndrs:
> Given we don't use `yarn install --production` anywhere, as I can see, we weren't getting any benefit anywho.

if we _should_ be doing that, we're not going to get only prod-only deps right now anyway. might as well make that explicit (and then fix it that's wrong)

arelra:
Just adding to:

> Given we don't use yarn install --production anywhere

[Yarn](https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-production-true-false) and [npm](https://docs.npmjs.com/cli/v8/commands/npm-install) docs mention:

> Yarn will not install any package listed in devDependencies if the NODE_ENV environment variable is set to production. Use this flag to instruct Yarn to ignore NODE_ENV and take its production-or-not status from this flag instead.

So if `NODE_ENV=production` for the `install` step that's also a route to only get runtime deps.

I can see that it is _not_ set for the `makefile` `install` step but I'm not sure of the setup on TeamCity.

imo separating deps is worth it to:
- make it easier to determine what is used at runtime vs being lost in the noise of devDeps
- reduce our prod build footprint by a couple of neutron stars

PR #4007, opened 2022-02-16
Convert GuStatsPlugin to Typescript

mxdvl:
## What does this change?

Convert GuStatsReportPlugin to Typescript.

## Why?

We’re converting the Webpack config to typescript in #3967.

### DIFF
--- a/dotcom-rendering/scripts/webpack/plugins/gu-stats-report-plugin.ts
+++ b/dotcom-rendering/scripts/webpack/plugins/gu-stats-report-plugin.ts
@@ -1,21 +1,41 @@
-const fetch = require('node-fetch');
-const os = require('os');
-const { exec } = require('child_process');
-const chalk = require('chalk');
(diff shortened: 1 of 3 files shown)


dotcom-rendering/scripts/webpack/plugins/gu-stats-report-plugin.ts
```diff
@@ -1,32 +1,50 @@
...
+	protected logger: Record<
+		'log' | 'warn' | 'error',
+		(...args: unknown[]) => void
+	> = console;
+
+	protected gitBranch: string = 'unknown';
+
+	protected gitHash: string = 'unknown';
+
+	constructor(config?: {
```
OllysCoding, line 28: Should config be an optional param? Feel like it should probably be required

mxdvl:
Made obsolete by #4017

PR #4009, opened 2022-02-16
Inline `setupMiddlewares`

sndrs:
## What does this change?

Inline `setupMiddlewares`

## Why?

Since we pulled the dev server out of the main webpack config, there's less obvious value in pulling these details apart


oliverlloyd (approved):
This is much simpler 😍

PR #4010, opened 2022-02-16
Handle backgrounded liveblogs

oliverlloyd:
## What does this change?
This PR introduces logic to handle when any new blocks are revealed to the reader

## Why?
If a blog is on a page or tab that does not have focus then it will continue to poll but we don't want any updates to be inserted into the top of the blog while the page is backgrounded. Once the page becomes visible again then, at that point, the new blocks are revealed.

### DIFF
--- a/dotcom-rendering/src/web/components/Liveness.importable.tsx
+++ b/dotcom-rendering/src/web/components/Liveness.importable.tsx
@@ -200,17 +200,12 @@ export const Liveness = ({
 			}
 		};
 
-		window.addEventListener(
-			'visibilitychange',
-			handleVisibilityChange,
-			false,
-		);
+		window.addEventListener('visibilitychange', handleVisibilityChange);
 
 		return () => {
 			window.removeEventListener(
 				'visibilitychange',
 				handleVisibilityChange,
-				false,
 			);
 		};
 	}, [numHiddenBlocks]);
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -179,6 +179,42 @@ export const Liveness = ({
...
 	}
 
+	/**
+	 * This useEffect sets up a listener for when the page is backgrounded or restored. We
+	 * do this so that any new blocks that were fetched while the blog was in the
+	 * background are animated in at the point when focus is restored
+	 */
+	useEffect(() => {
+		const handleVisibilityChange = () => {
+			// The blog was either hiiden or has become visible
```
mxdvl, line 189:
```suggestion
			// The blog was either hidden or has become visible
```

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -179,6 +179,42 @@ export const Liveness = ({
...
+				revealNewBlocks();
+				setNumHiddenBlocks(0);
+				setShowToast(false);
+			}
+		};
+
+		window.addEventListener(
+			'visibilitychange',
+			handleVisibilityChange,
+			false,
```
mxdvl, line 206: Isn’t this the default value, anyways? If being explicit is what we’re after, shouldn’t we use the `options` object instead?

mxdvl (approved):
LGTM, with some small notes

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: I think you're right, it's not required so the simplest thing is to just not pass it. Done

PR #4019, opened 2022-02-17
Set a min height for email embed iframes

georgeblahblah:
## What does this change?

* Sets a `min-height` for email embed iframes to `60px`

## Why?

* Prevents the reCAPTCHA message getting clipped

### Before

In some, but not all cases:
Screenshot 2022-02-17 at 12 43 31

### After

Screenshot 2022-02-17 at 12 44 18

### DIFF
--- a/dotcom-rendering/src/web/components/EmbedBlockComponent.importable.tsx
+++ b/dotcom-rendering/src/web/components/EmbedBlockComponent.importable.tsx
@@ -19,11 +19,12 @@ const emailCaptionStyle = css`
 	color: ${text.supporting};
 `;
 
-const embedContainer = css`
+const embedContainerStyles = (isEmailEmbed: boolean) => css`
 	iframe {
 		/* Some embeds can hijack the iframe and calculate an incorrect width, which pushes the body out */
 		/* stylelint-disable-next-line declaration-no-important */
 		width: 100% !important;
+		${isEmailEmbed && `min-height: 60px;`}
 	}
 `;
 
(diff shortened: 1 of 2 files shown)


dblatcher (approved):

dotcom-rendering/src/web/components/EmbedBlockComponent.importable.tsx
```diff
@@ -19,11 +19,12 @@ const emailCaptionStyle = css`
 	color: ${text.supporting};
 `;
 
-const embedContainer = css`
+const embedContainerStyles = (isEmailEmbed: boolean) => css`
 	iframe {
 		/* Some embeds can hijack the iframe and calculate an incorrect width, which pushes the body out */
 		/* stylelint-disable-next-line declaration-no-important */
 		width: 100% !important;
+		${isEmailEmbed && `min-height: 60px;`}
```
oliverlloyd, line 27: What if the embed is shorter than 60px?

The problem here is this component is not specific to the email signup form so semantically I guess you could have smaller content?

georgeblahblah: My understanding is that if `isEmailEmbed` is `false`, the `min-height` won't be set. So although the component is not specific to email signups, this part of the CSS is. Does that match your understanding?

oliverlloyd: Ah, I just saw that you're using `isEmailEmbed`. 🤔 that boolean looks a bit iffy. I see there's a comment saying it was due to be removed but that was a while back now.

Do you mind if we chat this one through?

JamieB-gu: George did some work in AR to parse email embeds out as a separate element (`EmailSignup`) in AR, so there's a standalone component specifically for them (as seen in the other part of this diff). Perhaps a similar approach, with a dedicated `EmailSignup` component, would work for DCR too?

PR #4020, opened 2022-02-17
Move `comments` ad slot out of its island

chrislomaxjones:
Co-authored-by: Lucy Monie Hall [lucy.monie-hall@guardian.co.uk](mailto:lucy.monie-hall@guardian.co.uk)

## What does this change

Move the `comments` ad slot out of the island in which it’s rendered.

To do this, we have introduced a new `DiscussionLayout` component that wraps all of the discussion logic.

In order to ensure the comments slot is SSR, we have to split the current island that contains discussions into two separate islands - one that handles the `SignedInAs` component on the left content one that handles the actual `Discussion` component. This is because of the way `ContentLayout` splits the left versus right, requiring us to hydrate part of the main content, and also the left content. This shouldn’t have a performance impact as SWR will cache equivalent API calls.

Below is an image summarising the change (islands in green):

## Why

Ad slots are not consistently filled when they’re placed in islands - this is because the commercial bundle expects all non-dynamic slots be present on the page at the point the commercial bundle initialises.

We need to move the comments slot out of its island so it’s server-side rendered, and therefore guaranteed to be present on the page when the commercial bundle needs it.

### Before

Screenshot 2022-02-17 at 16 52 12

### After

Screenshot 2022-02-17 at 16 52 02

### DIFF
--- a/dotcom-rendering/src/web/components/Discussion.stories.tsx
+++ b/dotcom-rendering/src/web/components/Discussion.stories.tsx
@@ -11,7 +11,6 @@ const HydratedLayout = ({ children }: { children: React.ReactNode }) => {
 	return <>{children}</>;
 };
 
-
 export default {
 	component: DiscussionLayout,
 	title: 'Components/DiscussionLayout',


oliverlloyd:
I will spend some time reviewing this tomorrow but I just wanted to say that this is one of the best PR descriptions I've seen for a long time, Thank you!

dotcom-rendering/src/web/components/DiscussionLayout.tsx
```diff
@@ -0,0 +1,99 @@
...
+interface DiscussionLayoutProps {
+	format: ArticleFormat;
+	discussionApiUrl: string;
+	shortUrlId: string;
+	discussionD2Uid: string;
+	discussionApiClientHeader: string;
+	enableDiscussionSwitch: boolean;
+	isAdFreeUser: boolean;
+	shouldHideAds: boolean;
+	expanded?: boolean;
```
oliverlloyd, line 20: This prop isn't used anywhere. Not sure if it was alway this way but it can be removed

oliverlloyd: In fact, I think it's use further down (in `Discussion`) can be removed. The internal logic of the discussion itself should decide this

oliverlloyd (approved):
I really like this PR 😻 

It's using existing platform patterns to effectively solve a problem in a well described way and leaves the code in a better state than it was before.

I left a comment about the `expanded` prop but you need to make these changes as part of this PR - you didn't introduce this and it can be addressed as a separate concern.

dotcom-rendering/src/web/components/DiscussionLayout.tsx
chrislomaxjones: Ah yeah, worth bearing in mimd!

PR #4021, opened 2022-02-17
Add LiveDateline component for liveblogs/deadblogs

iainjchambers-guardian:
## What does this change?
Adds a relative time to the header of liveblogs or deadblogs, showing time since last update

### Before
image

### After
image

### DIFF
--- a/apps-rendering/src/components/liveDateline.tsx
+++ b/apps-rendering/src/components/liveDateline.tsx
@@ -1,12 +1,13 @@
 // ----- Imports ----- //
 
 import { css, keyframes } from '@emotion/react';
+import type { ArticleFormat } from '@guardian/libs';
+import { ArticleDesign } from '@guardian/libs';
 import { neutral, pxToRem, textSans } from '@guardian/source-foundations';
-import type { FC } from 'react';
-import { ArticleDesign, ArticleFormat } from '@guardian/libs';
 import type { Option } from '@guardian/types';
-import { maybeRender } from 'lib';
 import { makeRelativeDate } from 'date';
+import { maybeRender } from 'lib';
+import type { FC } from 'react';
 
 // ----- Component ----- //
 
@@ -17,7 +18,7 @@ const livePulse = keyframes`
     100% {opacity: 1;}
 `;
 
-const timestampStyles = (isDeadBlog: boolean = false) => css`
+const timestampStyles = (isDeadBlog = false): ReturnType<typeof css> => css`
 	color: ${isDeadBlog ? neutral[7] : neutral[100]};
 	${textSans.xxsmall({ lineHeight: 'tight' })}
 `;
--- a/apps-rendering/src/components/liveDateline.tsx
+++ b/apps-rendering/src/components/liveDateline.tsx
@@ -1,0 +1,61 @@
+// ----- Imports ----- //
+
+import { css, keyframes } from '@emotion/react';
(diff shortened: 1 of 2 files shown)


apps-rendering/src/components/liveDateline.tsx
```diff
@@ -0,0 +1,62 @@
...
+// ----- Component ----- //
+
+const livePulse = keyframes`
+    0% {opacity: 1;}
+    10% {opacity: .25;}
+    40% {opacity: 1;}
+    100% {opacity: 1;}
+`;
+
+const timestampStyles = (isDeadBlog = false): ReturnType<typeof css> => css`
```
MarSavar, line 21: I think this function can return `SerializedStyles`

MarSavar (approved):
LGTM!  🥳 Just one comment.

Also could you please add before/after screenshots in the PR text? 🙏

AshCorr:
Theres a similar PR open for DCR to add this, I wonder if any of it can be re-used for AR?

https://github.com/guardian/dotcom-rendering/pull/3988

apps-rendering/src/components/liveDateline.tsx
JamieB-gu: What's the concern behind having an additional import? Is it a code quality/style/modularisation thing, or is it a bundle-size thing?

apps-rendering/src/components/liveblogHeader.tsx
```diff
@@ -46,9 +45,12 @@ const LiveblogHeader: FC<Props> = ({ item }) => {
 			>
 				<Columns collapseUntil="desktop">
 					<Column span={3}>
-						<time css={timestampStyles}>
-							TODO: Updated timestamp
-						</time>
+						<Hide below="desktop">
+							<LiveDateline
+								date={item.publishDate}
```
JamieB-gu, line 50: I think we're going to want to use some version of the "last modified" date for this. `publishDate` is when the piece was first published, which is likely to be some hours before it was actually last updated.

We don't currently have this on `Item` so I think we have two options:

1. Duplicate the logic that @AshCorr has used in #3988 to pull last modified from the most recent live block.
2. Parse `content.fields.lastModified` into a new `lastModifiedDate` on `Item` (`item.ts`) and use that.

JamieB-gu (commented):
Looks good, just one point about which date to use.

> Theres a similar PR open for DCR to add this, I wonder if any of it can be re-used for AR?

I agree with you @AshCorr, duplicating this logic/UI is not ideal - I think this is exemplified by the "last modified"/"first published" question below. As discussed offline I think we'll probably merge this as is and then look into a follow-up to share some of what you've done. Perhaps we can pull the side-effect that currently sits in `PulsingDot` into a wrapper?

apps-rendering/src/components/liveblogHeader.tsx
iainjchambers-guardian: I did look at this, but the `publishDate` does get updated when the content does and using it keeps it in sync locally with the live version

JamieB-gu: As discussed offline, I think you're right. `publishDate` is _not_ first published as I originally thought, it's actually the `webPublicationDate`, which is slightly different. For anyone else reading this, we're looking into how exactly that field gets updated, but for now I think we're ok to merge.

PR #4022, opened 2022-02-17
Update Quote icon alignment

mxdvl:
## Why?

It’s not tight enough.

### Before

image

### After

image


HarryFischer:
I've spotted a couple small issues in Chromatic – 10% day so we can pick it up on Monday? @mxdvl

oliverlloyd:
Would it be possible to add an explanation for why this styling change needs the html wrapping logic to be refactored? That's quite complex code and changes to it are hard to follow based on the code alone.

mxdvl:
@oliverlloyd I’ve separated the changes out in their own PR #4033, this should be much more readable.

HarryFischer (approved):
Looks better now - thanks

PR #4024, opened 2022-02-18
update webpack-messages type declaration

mxdvl:
## What does this change?

Will try to bring these changes to DefinitielyTyped,
or the original package.

## Why?

I’ve looked more closely at the package itself.


DavidLawes (approved):
👍 

I'm assuming if there was an error in these updated types it would be obvious as a result of the build?

If you've referenced specific docs/code to derive these types, could you point me in the right direction (purely out of interest)?

mxdvl:
@DavidLawes I think this will be made redundant by https://github.com/lukeed/webpack-messages/pull/7

PR #4025, opened 2022-02-18
Enhance tweets

oliverlloyd:
## What does this change?
We now enhance twitter embeds after they have been inserted.

To be able to target this enhancement request at only those embeds that need it (rather than all twitter embeds on the page) I have also added the `pending` class to mark recently inserted but not yet revealed blocks. We will use the same class to animate in blocks.

## Why?
We want to show readers the enhanced version of Twitter embeds

### Before
image

### After
image


mxdvl (approved):
We paired on this

PR #4027, opened 2022-02-18
Fix Callout Font

oliverlloyd:
## What does this change?
Adds font styling to the callout description text

## Why?
It looks better

### Before
Screenshot 2022-02-18 at 12 51 54

### After
Screenshot 2022-02-18 at 12 50 53


dotcom-rendering/src/web/lib/renderElement.tsx
```diff
@@ -7,7 +7,7 @@ import {
 } from '@guardian/atoms-rendering';
 import { ArticleDesign, ArticleFormat } from '@guardian/libs';
 import { BlockquoteBlockComponent } from '../components/BlockquoteBlockComponent';
-import { CalloutBlockComponent } from '../components/CalloutBlockComponent';
```
AshCorr, line 10: Was this meant to be included?

AshCorr (approved):

dotcom-rendering/src/web/lib/renderElement.tsx
AshCorr: Ah its already in an island

PR #4029, opened 2022-02-18
`Span` Module - Calculating Widths By Guardian Grid Columns

JamieB-gu:
## Why?

We have a series of design grids based on the concept of columns and gutters: https://theguardian.design/2a1e5182b/p/41be19-grids

I thought it might be convenient to have some functions that take a number of columns and return the width in pixels for a given grid size. I know that Source does this under the hood for the `Columns` component, but I couldn't find anywhere that we were exposing the base calculations in an API (I may have missed it though).

This is just a draft as I don't know how useful people might find it. It's just something that occurred to me because I was working on something where it would have been handy. All feedback welcome 🙂, particularly from @SiAdcock @sndrs regarding crossover with Source.

### IntelliSense and Type Safety

For all functions in this API the number of columns allowed for a given grid size is constrained at the type level. Passing a number of columns outside of the accepted range will result in a type error, and TS-based IntelliSense is available.

#### IntelliSense On Hover

intellisense-desktop
intellisense-leftcol

#### IntelliSense For Parameters

intellisense-parameters

#### Type Safety

type-safety-desktop
type-safety-span

#### `Grid`

An `enum` specifying the Grid sizes: Tablet, Desktop, LeftCol and Wide. There's also a Mobile grid size, but it's fluid based on screen width and therefore calculating exact widths isn't possible. It might be interesting to consider percentage-or-viewport-based widths in this case... I haven't thought about it enough.

#### `span`

The core function. Takes a grid size and a number of columns, and returns the width corresponding to that number of columns at that grid size (in pixels). I've exported this directly, but in practice it might be preferable to use the convenience functions in most cases.

#### `spanTablet`, `spanDesktop`, `spanLeftCol` and `spanWide`

Convenience functions that calculate the width for a given grid size. Each of these take a number of columns and return the width in pixels.

## Changes

- Added `span` function to calculate width according to number of columns in the Guardian grid
- Added `spanTablet`, `spanDesktop`, `spanLeftCol` and `spanWide` convenience functions

### DIFF
--- a/common-rendering/src/span.test.ts
+++ b/common-rendering/src/span.test.ts
@@ -1,0 +1,55 @@
+// ----- Imports ----- //
+
+import {
+    Grid,
+    span,
+    spanTablet,
+    spanDesktop,
+    spanLeftCol,
+    spanWide,
+} from "./span";
+
+// ----- Tests ----- //
+
+describe("span", () => {
+    describe("spanTablet", () => {
+        it("calculates the width of 5 columns for a tablet grid", () => {
+            expect(spanTablet(5)).toBe(280);
+        });
+    });
+
+    describe("spanDesktop", () => {
+        it("calculates the width of 6 columns for a desktop grid", () => {
+            expect(spanDesktop(6)).toBe(460);
+        });
+    });
+
+    describe("spanLeftCol", () => {
+        it("calculates the width of 7 columns for a leftCol grid", () => {
+            expect(spanLeftCol(7)).toBe(540);
+        });
+    });
+
+    describe("spanWide", () => {
+        it("calculates the width of 8 columns for a wide grid", () => {
+            expect(spanWide(8)).toBe(620);
+        });
+    });
+
+    describe("span", () => {
+        it("calculates the width of 8 columns for a tablet grid", () => {
+            expect(span(Grid.Tablet)(8)).toBe(460);
+        });
+        it("calculates the width of 7 columns for a desktop grid", () => {
+            expect(span(Grid.Desktop)(7)).toBe(540);
+        });
+        it("calculates the width of 6 columns for a leftCol grid", () => {
+            expect(span(Grid.LeftCol)(6)).toBe(460);
+        });
+        it("calculates the width of 5 columns for a wide grid", () => {
+            expect(span(Grid.Wide)(5)).toBe(380);
+        });
+    });
+});
+
+
--- a/common-rendering/src/span.ts
+++ b/common-rendering/src/span.ts
@@ -1,0 +1,48 @@
+// ----- Definitions ----- //
+
+enum Grid {
+    Tablet,
+    Desktop,
+    LeftCol,
+    Wide,
+}
+
+const tabletColumnSize = 40;
+const desktopColumnSize = 60;
+const gutterSize = 20;
+
+// ----- Functions ----- //
+
+const calcSpan = (columnSize: number) => (numColumns: number): number =>
+    (numColumns * columnSize) + ((numColumns - 1) * gutterSize);
+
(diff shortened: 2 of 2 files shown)


SiAdcock:
Hi @JamieB-gu this looks very interesting. As well as Source's [Columns](https://github.com/guardian/source/blob/main/packages/@guardian/source-react-components/src/columns/Columns.tsx) – which we are revamping this quarter – several other teams have implemented their own grid APIs ([gateway](https://github.com/guardian/gateway/blob/main/src/client/styles/Grid.ts), [manage-frontend](https://github.com/guardian/manage-frontend/blob/main/app/client/styles/grid.ts)). I suppose one more can't hurt 🙈 

>  I was working on something where it would have been handy

This is key to understanding the fragmentation that's occurring. Would you like to elaborate more on this?

JamieB-gu:
> several other teams have implemented their own grid APIs

> Would you like to elaborate more on this?

Oh interesting! Based on a brief scan through that looks fairly similar to what I was experimenting with: using variations of CSS grid to implement the Guardian Grid 🙂. They already match up pretty well conceptually.

There are a couple of approaches that I'm looking at to get the Editorial layout used on DCAR:

1. Using `grid-template-areas` to name regions of the page. Examples: https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/layouts/StandardLayout.tsx, https://github.com/guardian/dotcom-rendering/blob/main/apps-rendering/src/components/layout/live.tsx. These don't really rely on the Guardian columns that much except for a few hardcoded examples in `grid-template-columns`.
2. Using `repeat`, `grid-column` and `grid-row`, possibly with some named lines. This approach allows you to replicate the Guardian grid in CSS grid code directly, which sounds like it may be what `gateway` and `manage-frontend` have done? This is the approach I was looking into last week.

Both of these can result in a bunch of hardcoded width values that correspond to specific Guardian grid widths (e.g. `220px` for the left column), which is what led me to create this helper.

> As well as Source's [Columns](https://github.com/guardian/source/blob/main/packages/@guardian/source-react-components/src/columns/Columns.tsx)

Yup we've used the `Container`/`Columns` components for some things 👍 (https://github.com/guardian/dotcom-rendering/blob/main/apps-rendering/src/components/liveblogHeader.tsx). I think possibly the reason people have reached for CSS grid is that the designs tend to move the ordering and positioning of content across different breakpoints (key events in liveblogs, metadata/bylines on normal articles, ordering of headline/standfirst/main media etc.). For liveblogs, for example, I spent quite a while trying to express that using `Container`/`Columns` but I couldn't get it to work 😔

mxdvl (approved):
What use cases do you envision for this? You’ve mentioned specifying the CSS grid size this way.

For mobile, yeah, you’d have to keep thinking about it. I thought it was as simple as doing something clever like `calc(${span / 4}vw - 40px)` but it’s more complex

JamieB-gu:
> What use cases do you envision for this? You’ve mentioned specifying the CSS grid size this way.

I'd like to try it out for a few things. As you mentioned it's useful for specifying the size of some of our CSS grid layouts:

- Left column: `140px` at leftCol (2 Guardian columns), `220px` at wide (3 Guardian columns)
- Centre column: `620px` (8 Guardian columns)
- Right column: `300px` (4 Guardian columns)
- Gap between centre and right column: `80px` (1 Guardian column + 1 Guardian gutter)

There are also lots of elements throughout the page that follow the Guardian grid rules. For example standfirsts are `540px` on standard, `460px` on immersive - these are equivalent to 7 and 6 columns in Guardian layout terms respectively. Headlines are `620px` (8 columns); showcase images are `860px` (11 Guardian columns). And so on.

PR #4030, opened 2022-02-18
Increase timeout in Cypress click to view embed test

chrislomaxjones:
## What does this change?

Increase the timeout when waiting for the embed element to 30 seconds.

## Why?

This can sometimes take longer than the default 4 seconds, causing the test to fail.

### DIFF
--- a/dotcom-rendering/cypress/integration/parallel-1/article.elements.spec.js
+++ b/dotcom-rendering/cypress/integration/parallel-1/article.elements.spec.js
@@ -109,7 +109,7 @@ describe('Elements', function () {
 			cy.scrollTo(0, 4500);
 
 			// Wait for hydration
-			cy.get('gu-island[name=EmbedBlockComponent]')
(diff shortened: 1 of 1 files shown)


dotcom-rendering/cypress/integration/parallel-1/article.elements.spec.js
```diff
@@ -109,7 +109,7 @@ describe('Elements', function () {
 			cy.scrollTo(0, 4500);
 
 			// Wait for hydration
-			cy.get('gu-island[name=EmbedBlockComponent]')
+			cy.get('gu-island[name=EmbedBlockComponent]', { timeout: 30000 })
```
mxdvl, line 112:
```suggestion
			cy.get('gu-island[name=EmbedBlockComponent]', { timeout: 30_000 })
```

mxdvl (approved):

PR #4031, opened 2022-02-21
Animate new posts

oliverlloyd:
## What does this change?
This adds animation when a new blog post is revealed

## Why?
To make it clear to readers what is happening

## Should we change how we announce new content?
We could, I've implemented the same styling here as we have in Frontend but we can still change this later

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleBody.tsx
+++ b/dotcom-rendering/src/web/components/ArticleBody.tsx
@@ -1,4 +1,4 @@
-import { css } from '@emotion/react';
+import { css, keyframes } from '@emotion/react';
 
 import { headline, body, between, space } from '@guardian/source-foundations';
 import { ArticleDesign, ArticleDisplay } from '@guardian/libs';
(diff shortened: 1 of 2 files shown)


ioannakok (approved):
Looks beautiful 👍

dotcom-rendering/src/web/components/ArticleBody.tsx
```diff
@@ -72,6 +72,20 @@ const globalLinkStyles = (palette: Palette) => css`
...
 `;
 
+const revealStyles = css`
+	/* We're using classnames here because we add and remove these classes
+	   using plain javascript */
+	.reveal {
+		animation: ${keyframes`
+			0% { opacity: 0; }
+			100% { opacity: 1; }
+		`} 4s ease-out;
```
OllysCoding, line 82: 4s is quite a long animation!

oliverlloyd: It's 5s on Frontend! I actually reduced it.

I suspect the reason for the length is because it can actually take a while for the content to be put in front of the reader. If they are half way down the page then the scroll takes a little time or if the content is being rendered that can also take a second or two.

mxdvl:
Do we have any intention of adding a smooth animation for the height, too?

I know animating height from `0` to `auto` is [notoriously difficult](https://css-tricks.com/using-css-transitions-auto-dimensions/). We’ll probably need some JS for this.

PR #4032, opened 2022-02-21
`onFirstPage`

oliverlloyd:
## What does this change?
This PR introduces the `onFirstPage` prop and builds logic using it to vary what we do depending on if the reader is on the first page of the blog or not

## Why?
We don't want to try and insert content at the top of the second page so instead we now use the Toast more and navigate readers to the first page of the blog if they click it on any subsequent page

### But polling is broken outside the first page
I noticed that polling is returning the wrong results if you're not on the first page. This is because it is using the wrong `latestBlockId`. But this bug is unrelated to this PR.

### DIFF
--- a/dotcom-rendering/src/web/components/Liveness.importable.tsx
+++ b/dotcom-rendering/src/web/components/Liveness.importable.tsx
@@ -10,6 +10,8 @@ type Props = {
 	filterKeyEvents: boolean;
 	format: ArticleFormat;
 	switches: Switches;
+	onFirstPage: boolean;
+	webURL: string;
 };
 
 const isServer = typeof window === 'undefined';
@@ -123,6 +125,8 @@ export const Liveness = ({
 	filterKeyEvents,
 	format,
 	switches,
+	onFirstPage,
+	webURL,
 }: Props) => {
 	const [showToast, setShowToast] = useState(false);
 	const [numHiddenBlocks, setNumHiddenBlocks] = useState(0);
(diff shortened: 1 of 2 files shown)


dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -228,16 +233,20 @@ export const Liveness = ({
...
+		if (onFirstPage) {
+			setShowToast(false);
+			document.getElementById('maincontent')?.scrollIntoView({
+				behavior: 'smooth',
+			});
+			window.location.href = '#maincontent';
+			revealPendingBlocks();
+			setNumHiddenBlocks(0);
+		} else {
+			window.location.href = `${webURL}#maincontent`;
```
DavidLawes, line 248: I'm trying to think of the flow:
- if not on first page and we have new blocks, toast will show
- if user clicks on toast they are redirected to first page
- on redirection we request paginated blocks for page 1
--> new paginated blocks for page 1 include latest posts?

DavidLawes (approved):
👍

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: It's more:

- if not on first page and we have new blocks, toast will show
- if user clicks on toast they are redirected to first page

At that point the page already includes the latest blocks so we don't need to do anything. We then start polling as normal. It's essentially as though they just opened the blog for the first time

PR #4033, opened 2022-02-21
Narrower types for `unwrapHtml`

mxdvl:
## What does this change?

Create a new `HTMLTag`, which matches standard HTML Tag names (e.g. `p`, `aside`, `main`, `table`, …)

Use this `HTMLTag` to add type safety to `unwrapHtml` and `RewrappedComponent`.

Add comments to explain more thoroughly the workings of these components using JSDoc.

Simplify some of the inner logic to rely on [`find`][] instead of `filter` then `[0]`.

[`find`]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

## Why?

It’s easier to understand the shape of the function with string literals.

The existing method was notoriously hard to reason about.
This changes makes the public API more narrowly typed.
When specifying a prefix or suffix, the compiler will complain if the
strings do not match opening and closing HTML tag names.

### DIFF
--- a/dotcom-rendering/src/web/components/RewrappedComponent.tsx
+++ b/dotcom-rendering/src/web/components/RewrappedComponent.tsx
@@ -1,21 +1,22 @@
-// @ts-ignore
+// @ts-ignore -- we’re actually using preact
 import { jsx as _jsx } from 'react/jsx-runtime';
 import { ClassNames } from '@emotion/react';
 
 import { unescapeData } from '../../lib/escapeData';
 
-// Ideally we want to want to avoid an unnecessary 'span' wrapper,
-// which also will cause issues with ads (spacefinder rules). React
-// requires a wrapping element for dangerouslySetInnerHTML so we
-// strip the original element markup and rewrap. The fallback case is
-// added for safety but should never happen. If it does happen
-// though it will cause issues with commercial JS which expects
-// paras to be top-level.
-
+/**
+ * React requires a wrapping element for `dangerouslySetInnerHTML` so we
+ * strip the original element markup and rewrap.
+ *
+ * The fallback case (`isUnwrapped == false`) is added for safety,
+ * but should _never_ happen. We want to avoid an unnecessary `<span>` wrapper,
+ * which can cause issues with ads and commercial JS (`spacefinder` rules).
+ * Commercial expects `<p>` tags at the top level of the rewrapped element.
+ */
 export const RewrappedComponent = ({
 	isUnwrapped,
 	html,
-	elCss = '',
+	elCss,
 	tagName,
 }: {
 	isUnwrapped: boolean;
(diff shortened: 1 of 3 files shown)


dotcom-rendering/cypress/integration/parallel-1/article.elements.spec.js
```diff
@@ -109,7 +109,7 @@ describe('Elements', function () {
 			cy.scrollTo(0, 4500);
 
 			// Wait for hydration
-			cy.get('gu-island[name=EmbedBlockComponent]')
+			cy.get('gu-island[name=EmbedBlockComponent]', { timeout: 30000 })
```
OllysCoding, line 112: Guessing this is unrelated but curious why it was added?!

mxdvl: I can’t figure it out… 🤷 

I’ll try rebasing.

dotcom-rendering/src/web/components/RewrappedComponent.tsx
```diff
@@ -2,30 +2,32 @@
...
-
+/**
+ * React requires a wrapping element for `dangerouslySetInnerHTML` so we
+ * strip the original element markup and rewrap.
+ *
+ * The fallback case (`isUnwrapped == false`) is added for safety,
+ * but should _never_ happen. We want to avoid an unnecessary `<span>` wrapper,
+ * which can cause issues with ads and commercial JS (`spacefinder` rules).
+ * Commercial expects `<p>` tags at the top level of the rewrapped element.
+ */
```
mxdvl, line 16: I think that if there case, it should be capture somehow. If we have no visibility on it, how can we hope to fix it?

DavidLawes (approved):
👍 

Out of curiosity, what are the team's thoughts on how strongly typed the codebase is/should be?

At the moment, my feeling on the status quo was:
- AR is quite strongly typed
- DCR is less so

^^ I'm not sure if that is an intentional choice or a side-effect of something else.

oliverlloyd:
> 👍
> 
> Out of curiosity, what are the team's thoughts on how strongly typed the codebase is/should be?
> 
> At the moment, my feeling on the status quo was:
> 
> * AR is quite strongly typed
> * DCR is less so
> 
> ^^ I'm not sure if that is an intentional choice or a side-effect of something else.

It's intentional. The goal with DCR is to have simple, accessible code with new developers needing to know as little as possible to be able to maintain and contribute. These goals were set for DCR because we want this codebase to live for as long as possible, delaying the next (inevitable) rewrite.

These principles are also outlined in the [Guardian's recommendations](https://github.com/guardian/recommendations/blob/main/coding-with-empathy.md).

That said, I think there are situations where adding more complex typing is worth it. In particular in situations where we are exposing apis that we want to be as easy to consume as possible.

Deciding where and when to apply these principles is, of course, subjective but I think we should try to keep the principles above in mind when making these choices.

PR #4034, opened 2022-02-21
feat: Create a PR thread every day and announce all un-reviewed PR's …

AshCorr:
## What does this change?

Adds a new Github action that runs weekday day at 6AM.

This action will create a new Google Chats thread with all of the currently open PR's that are pending review (ignoring any dependabot PR's or Stale PR's)

Action code (Written in Rust!): https://github.com/guardian/google-chats-pr-announcer

## Why?

Although its not much work for someone to create the thread themselves in the morning this bot will also include any un-reviewed PR's from previous days, saving people from having to keep reminding people to review their PR's.

## How does it work

Google provides a very convenient to use Webhook endpoint which creates Google chat messages from any POST request it receives. This bot simply polls the Github API for any un-reviewed PR's and then forwards those results to the Google chat endpoint.

```
curl -X POST -H "Content-Type: application/json" \
    -d '{"text": "Hello World!"}' \
    https://chat.googleapis.com/v1/spaces/AAAAJglfhNo/messages?key=$KEY
```

image

### Screenshots
image

### DIFF
--- a/.github/workflows/prannouncer.yml
+++ b/.github/workflows/prannouncer.yml
@@ -11,4 +11,4 @@ jobs:
     steps:
       - uses: guardian/google-chats-pr-announcer@main
         with:
-          google-webhook-url: ${{ secret.GOOGLE_WEBHOOK_URL }}
+          google-webhook-url: ${{ secrets.GOOGLE_WEBHOOK_URL }}
--- a/.github/workflows/prannouncer.yml
+++ b/.github/workflows/prannouncer.yml
@@ -1,0 +1,14 @@
+name: "Google Chats PR Announcer"
+
+on:
+	workflow_dispatch:
+  schedule:
+    - cron: "0 6 * * 1-5"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: guardian/google-chats-pr-announcer@main
+        with:
+          google-webhook-url: ${{ secret.GOOGLE_WEBHOOK_URL }}
--- a/.github/workflows/prannouncer.yml
+++ b/.github/workflows/prannouncer.yml
@@ -6,7 +6,7 @@ on:
     - cron: "0 6 * * MON-FRI"
 
 jobs:
-  prannouncer:
+  prnouncer:
     runs-on: ubuntu-latest
     steps:
       - uses: guardian/google-chats-pr-announcer@main
--- a/.github/workflows/prannouncer.yml
+++ b/.github/workflows/prannouncer.yml
@@ -3,10 +3,10 @@ name: "Google Chats PR Announcer"
 on:
   workflow_dispatch:
   schedule:
-    - cron: "0 6 * * 1-5"
+    - cron: "0 6 * * MON-FRI"
 
 jobs:
-  stale:
+  prannouncer:
     runs-on: ubuntu-latest
     steps:
       - uses: guardian/google-chats-pr-announcer@main
--- a/.github/workflows/prannouncer.yml
+++ b/.github/workflows/prannouncer.yml
@@ -1,7 +1,7 @@
 name: "Google Chats PR Announcer"
 
 on:
-	workflow_dispatch:
+  workflow_dispatch:
   schedule:
     - cron: "0 6 * * 1-5"
 


.github/workflows/prannouncer.yml
```diff
@@ -0,0 +1,14 @@
+name: "Google Chats PR Announcer"
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 6 * * 1-5"
```
mxdvl, line 6: I think the following syntax is supported:

```suggestion
    - cron: "0 6 * * MON-FRI"
```

.github/workflows/prannouncer.yml
```diff
@@ -0,0 +1,14 @@
...
+
+jobs:
+  stale:
```
mxdvl, line 9: why call it `stale`?

mxdvl (approved):

jamesgorrie:
Ping @guardian/developer-experience for this ✨ new tool!

jamesgorrie (commented):
My thought on this Friday would be that it was useful to separate codeowners / external contributions to help prioritise. 

Not blocking just wondering what you think?

.github/workflows/prannouncer.yml
AshCorr: Didn't know that, Thanks!

.github/workflows/prannouncer.yml
AshCorr: Woops, good catch, I was lazy and copied the action from the stale PR action I previously made!

.github/workflows/prannouncer.yml
```diff
@@ -0,0 +1,14 @@
+name: "Google Chats PR Announcer"
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 6 * * MON-FRI"
+
+jobs:
+  prannouncer:
```
jamesgorrie, line 9: If you wanted to follow guardian naming, you could call it Pronouncer, or even more in line prnouncer.

.github/workflows/prannouncer.yml
```diff
@@ -0,0 +1,14 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: guardian/google-chats-pr-announcer@main
+        with:
+          google-webhook-url: ${{ secret.GOOGLE_WEBHOOK_URL }}
```
jamesgorrie, line 14: Do we need to set the `https://github.com/guardian/google-chats-pr-announcer#github-repositories` and other required values from the action?

jamesgorrie (commented):
A question on whether we have to send more values down to the action.

.github/workflows/prannouncer.yml
AshCorr: Nope, they *should* default to the correct values by themselves.

The only reason you would want to set these yourselves is if you wanted to perhaps scan multiple repositories (which we might want to do) or if you're external to the guardian and you want to ignore different labels / bot users

.github/workflows/prannouncer.yml
AshCorr: Love it.

.github/workflows/prannouncer.yml
AshCorr: By default it just uses the name of the repository the action is created in, but we could set it to
`guardian/dotcom-rendering, guardian/frontend, guardian/discussion-rendering, ... etc` if we wanted to include the PR's from other repositories at the Guardian. I've just left it default for now, but we might want to do something different for the P&E/Dotcom space

AshCorr:
> My thought on this Friday would be that it was useful to separate codeowners / external contributions to help prioritise.

Hm, an interesting thought. I think its possibly, but it is a fair bit of extra work, maybe something to be done next time we have a 10%!

I think for dotcom-rendering most of our PR's are internal, so as far as I know we don't have too much of an issue reviewing any external PR's. I feel like if someone has spent the time to learn and figure out our codebase then the least we can do is treat their PR's fairly and give it the same amount of time that we would give any PR's made by ourselves.

OllysCoding:
> Adds a new Github action that runs weekday day at 6PM.

I might have set this to maybe 8am myself, is there a reason we chose COP the day before?

OllysCoding (approved):
+1 I love it :)

AshCorr:
> > Adds a new Github action that runs weekday day at 6PM.
> 
> I might have set this to maybe 8am myself, is there a reason we chose COP the day before?

Typo in the PR, it runs at 6**am**.

jamesgorrie:
> I think for dotcom-rendering most of our PR's are internal, so as far as I know we don't have too much of an issue reviewing any external PR's.

Most PRs are internal. External PRs are something we should really not let get lost in the slew of internal PRs, hence highlighting them _could_ be useful. Basically prioritising them over ours, which is maybe not a good expectation.

nicl:
This is great! Definitely something that I think other teams would benefit from too.

akash1810:
> This is great! Definitely something that I think other teams would benefit from too.

💯 this is great!

I think this model is better than the one employed by [`PR Exposer`](https://github.com/guardian/google-chat-bots/tree/main/pr-exposer) as the configuration is co-located with the repository, vs `PR Exposer`'s use of S3, which has a higher wall to climb to update it.

PR #4035, opened 2022-02-21
DCR: Add /KeyEvents endpoint to render key events individually

OllysCoding:
## What does this change?

Adds an endpoint to DCR for rendering key-events (for liveblogs) on their own, which can be used as part liveness to update the key-events in situ on DCR liveblogs

This follows the same pattern used to add the `/Blocks` endpoint in #3881

## Why?

Frontend already provides fresh key-events in the live api, however it does not currently render/update them on the page. This brings us closes to parity of functionality

### After
(styling abnormalities in screenshot due to lack of css-reset in individual responses)
image

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -571,6 +571,15 @@ interface BlocksRequest {
 	videoDuration?: number;
 }
 
+/**
+ * KeyEventsRequest is the expected body format for POST requests made to /keyEvents
+ */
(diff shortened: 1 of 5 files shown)


OllysCoding:
[DCR renders /KeyEvents](https://trello.com/c/j7v9aOpn/676-dcr-renders-keyevents)

oliverlloyd (approved):
So simple!

PR #4036, opened 2022-02-21
Log unwrapping errors

mxdvl:
## What does this change?

Logs errors where we failed to unwrap.

## Why?

This is a failure, but we don’t know where or when it’s happening, if ever!


OllysCoding (approved):
+1 Would be interested to see if this gets any calls! Could we add some comments here just quickly explaining why the code is there, given this is the first use of this logging pattern!

DavidLawes (approved):
👍  - nice!

Out of curiosity, now that we're logging errors do we also need to:
- alert on a certain threshold
- ensure the team know how to view these logs
- ensure we check these logs as some kind of regular process

PR #4037, opened 2022-02-21
Split dev url manually

oliverlloyd:
## What does this change?
Refactor the way we get the `url` param value for making the call to Frontend.

### Before
We used the `req.query.url` property

### After
We now take the whole `req.url` string and split it manually

## Why?
Like this we can also pass through any additional query params that might have been passed. This is especially useful for blogs where we use the url for pagination and state like `filterKeyEvents`


AshCorr (approved):
20 lines of comment, 1 line of code! LGTM!

sndrs (approved):

PR #4038, opened 2022-02-21
Poll more frequently

oliverlloyd:
## What?
Decreases the time between polls from 15 seconds to 10

## Why?
More frequent polling has limited impact as these calls are well cached but greatly improves the reader experience


ioannakok (approved):
+1

AshCorr:
Does Fastly not charge based on the number of requests made?

AshCorr (approved):

PR #4050, opened 2022-02-22
Collapse all npm packages into dependencies

tjmw:
## What does this change?

This happened previously in #4001 but the change was subsequently lost. A couple of dependencies have been added in the meantime which is why there are a couple of straggling lines (`@types/twitter-for-web` for example).


mxdvl:
Could we add a little CI test for this? Something like

```ts
import pkg from "./package.json";

if(pkg.devDependecies) throw new Error("no dev deps, please, se PR #4001");
```

mxdvl:
@tjmw + @sndrs &rarr; #4086

mxdvl:
@tjmw + @sndrs &rarr; #4086

PR #4052, opened 2022-02-22
Fix `KeyEvents` width

oliverlloyd:
## What does this change?
Removes a fixed width property from `KeyEvents`

## Why?
It was causing this element to overflow it's container

### Before
Screenshot 2022-02-21 at 21 22 21

### After
Screenshot 2022-02-21 at 21 22 54

### DIFF
--- a/common-rendering/src/components/keyEvents.tsx
+++ b/common-rendering/src/components/keyEvents.tsx
@@ -65,10 +65,6 @@ const keyEventWrapperStyles = (
 `;
 
 const listStyles = (supportsDarkMode: boolean): SerializedStyles => css`
-	${from.desktop} {
-		width: 13.75rem;
-	}
-
 	li::before {
 		content: "";
 		display: block;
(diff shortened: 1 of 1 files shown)


ioannakok (approved):
👍

oliverlloyd:
@JamieB-gu the AR stories look exploded on this but I think that's because there's no longer a width on this element and the stories don't have a container around them. I presume on the page this isn't an issue?

JamieB-gu:
> the AR stories look exploded on this

Which stories are you talking about? I think the only AR-specific stories that contain this component are the liveblog layouts? 🤔

oliverlloyd:
> > the AR stories look exploded on this
> 
> Which stories are you talking about? I think the only AR-specific stories that contain this component are the liveblog layouts? I can't see any diffs for them 🤔

As per our chat offline, from testing on CODE we're not seeing any AR issues with this change and the Chromatic diffs are only happening because the story is not bound by a container

PR #4053, opened 2022-02-22
Infer palette from format

mxdvl:
## What does this change?

If we have `format`, we can forgo `palette`. The consumer can do `decidePalette` – or even the fancier `editorialPalette`.

Added a `decideFormat` method to replace a common pattern:

```ts
// before
const format: ArticleFormat = {
	display: decideDisplay(CAPIformat),
	theme: decideTheme(CAPIformat),
	design: decideDesign(CAPIformat),
}

// after
const format: ArticleFormat = decideFormat(CAPIformat)
```

## Why?

There’s loads of duplicated code. Passing props that contain the same information makes refactors harder.

### DIFF
--- a/dotcom-rendering/src/web/components/App.tsx
+++ b/dotcom-rendering/src/web/components/App.tsx
@@ -1,7 +1,7 @@
 import React, { useState, useEffect } from 'react';
 import loadable from '@loadable/component';
 
-import { ArticleDisplay, ArticleDesign, storage, log } from '@guardian/libs';
+import { storage, log } from '@guardian/libs';
 import type { ArticleFormat } from '@guardian/libs';
 import type { BrazeMessagesInterface } from '@guardian/braze-components/logic';
 import {
@@ -19,8 +19,7 @@ import { AudioAtomWrapper } from './AudioAtomWrapper';
 import { Portal } from './Portal';
 import { HydrateOnce, HydrateInteractiveOnce } from './HydrateOnce';
 import { decideTheme } from '../lib/decideTheme';
-import { decideDisplay } from '../lib/decideDisplay';
-import { decideDesign } from '../lib/decideDesign';
+import { decideFormat } from '../lib/decideFormat';
 import { useOnce } from '../lib/useOnce';
 
 import { incrementAlreadyVisited } from '../lib/alreadyVisited';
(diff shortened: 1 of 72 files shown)


dotcom-rendering/src/web/lib/decideFormat.ts
```diff
@@ -0,0 +1,9 @@
+import { decideDesign } from './decideDesign';
+import { decideDisplay } from './decideDisplay';
+import { decideTheme } from './decideTheme';
+
+export const decideFormat = (format: Partial<CAPIFormat>): ArticleFormat => ({
+	display: decideDisplay(format),
+	theme: decideTheme(format),
+	design: decideDesign(format),
+});
```
mxdvl, line 9: A new method that replaces a common pattern. It’s also more resilient, as it has defaults values if the `CAPIFormat` is partial (any key missing).

mxdvl (commented):
See below for where to focus review attention 👇

dotcom-rendering/src/web/components/App.tsx
```diff
@@ -117,15 +116,9 @@ export const App = ({ CAPI }: Props) => {
 		setBrazeMessages(buildBrazeMessages(CAPI.config.idApiUrl));
 	}, [CAPI.config.idApiUrl]);
 
-	const display: ArticleDisplay = decideDisplay(CAPI.format);
-	const design: ArticleDesign = decideDesign(CAPI.format);
 	const pillar: ArticleTheme = decideTheme(CAPI.format);
```
oliverlloyd, line 119: We can remove this too, I guess?

dotcom-rendering/src/web/components/NumberedTitleBlockComponent.tsx
```diff
@@ -34,12 +32,7 @@ export const NumberedTitleBlockComponent = ({
 	html,
 	format,
 }: Props) => {
-	const dcrFormat = {
-		display: decideDisplay(format),
-		design: decideDesign(format),
-		theme: decideTheme(format),
-	};
-	const palette = decidePalette(dcrFormat);
+	const palette = decidePalette(decideFormat(format));
```
oliverlloyd, line 35: Just checking, are we doing `decideFormat(format)` here because `format` is really `CAPIFormat`?

dotcom-rendering/src/web/lib/decideFormat.ts
oliverlloyd: Should we allow CAPIFormat to be incomplete? That feels wrong to me. If we're missing a property there that sounds like something has gone very wrong and we probably want to throw rather than default (defaulting is likely to result in us rendering the _wrong_ thing - like, calling a sport piece news, which is bad)

oliverlloyd (approved):
Amazing! 😍 

This makes the tests for these components so much simpler

dotcom-rendering/src/web/components/App.tsx
mxdvl: No, not yet. Some component is using it.

dotcom-rendering/src/web/lib/decideFormat.ts
mxdvl: This can be addressed in another PR. This type is only a better reflection of what these methods accept. `decideDisplay`, `decideTheme` and `decideDesign` all have a `default:` in their switch. If we threw, that would be a change in the current behaviour!

In decideTrail, we did this the following:

```ts
const format: ArticleFormat = trail.format
		? {
				display: decideDisplay(trail.format),
				theme: decideTheme(trail.format),
				design: decideDesign(trail.format),
		  }
		: {
				display: ArticleDisplay.Standard,
				theme: ArticlePillar.News,
				design: ArticleDesign.Standard,
		  };
```

Issue #4057, opened 2022-02-23
embed.ly oembeds are returning `url: ""` for Vimeo embeds


jamesgorrie:
It seems [embed.ly](http://embed.ly/) returns `""` as the `url` in the oembed response for vimeo. This is odd, as you can see a more expected result for youtube.

https://embed.ly/docs/explore/oembed?url=https%3A%2F%2Fvimeo.com%2F539597768
https://embed.ly/docs/explore/oembed?url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DW8HVOBzg2pE

Composer uses this response directly, so bubbles down the pipeline to frontend.

We have raised a support ticket with embed.ly.

OllysCoding:
@jamesgorrie to follow up

rhiannareechaye:
@jamesgorrie this looks like it's over a year old so we are putting in the 'to be closed' bucket

PR #4058, opened 2022-02-23
Vimeo element url

jamesgorrie:
part of https://github.com/guardian/dotcom-rendering/issues/4057
## What does this change?
Checks if the `url === ''` then falls back to the `originalUrl`. This seems to be a bug that has been caused by the new way composer / flexible content are no longer omitting the `url` if embed.ly returns an emopty string. Which it does. See the issue for more details.

This is a hack and should be removed once we get to the bottom of that issue, but we're clogging up our errors which is a dangerous place to be.

## Why?
Less of these errors = more visibility into actual errors

### Before
Screenshot 2022-02-23 at 09 46 10

### After
Screenshot 2022-02-23 at 09 46 04


OllysCoding (approved):
+1

PR #4061, opened 2022-02-23
Remove Redundant `max-width` In `Standfirst`

JamieB-gu:
## Why?

It's immediately overwritten by the next `max-width`.

## Changes

- Removed redundant `max-width` in `Standfirst`


MarSavar (approved):
👍

PR #4062, opened 2022-02-23
Add `cardTextByline` to palette

mxdvl:
## What does this change?

Add a new palette key: `cardTextByline`. Follows a similar logic to `cardTextHeadline`.

## Why?

Currently, we use the `textByline` colour for cards, but this breaks for Special Reports, where the background is the same colour as the byline.

### Before

<a href="https://5dfcbf3012392c0020e7140b-qgngtexhes.chromatic.com/?path=/story/components-cardheadline--byline">
image
</a>

### After

image

### DIFF
--- a/dotcom-rendering/src/web/components/Byline.tsx
+++ b/dotcom-rendering/src/web/components/Byline.tsx
@@ -8,6 +8,7 @@ type Props = {
 	text: string;
 	format: ArticleFormat;
 	size: SmallHeadlineSize;
+	isCard?: boolean;
 };
 
 const bylineStyles = (size: SmallHeadlineSize, format: ArticleFormat) => {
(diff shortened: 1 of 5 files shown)


ioannakok (approved):
Looks good to me!

dotcom-rendering/src/web/components/Byline.tsx
```diff
@@ -78,7 +78,7 @@ const bylineStyles = (size: SmallHeadlineSize, format: ArticleFormat) => {
 
 const colourStyles = (palette: Palette) => {
 	return css`
-		color: ${palette.text.byline};
+		color: ${palette.text.cardByline};
```
JamieB-gu, line 81: Is the `Byline` component only used for cards?

JamieB-gu (approved):
Nice! One question.

PR #4063, opened 2022-02-23
Add AR Standard layout to Storybook

MarSavar:
## Why?
Because it doesn't currently exist.

I have also:
- slightly refactored the structure of the layout stories for AR so they match DCR.  

image

- created tag fixtures

### Before
N/A

#### Standard Article
image
#### Review
image
#### Match Report
image

As mentioned above, I have also added tags to better mimic what articles currently look like in AR.

image

### DIFF
--- a/apps-rendering/src/components/layout/standard.stories.tsx
+++ b/apps-rendering/src/components/layout/standard.stories.tsx
@@ -1,0 +1,77 @@
+// ----- Imports ----- //
+import { ArticleDesign, ArticleDisplay, ArticlePillar } from '@guardian/libs';
+import { breakpoints } from '@guardian/source-foundations';
+import { partition } from '@guardian/types';
+import { withKnobs } from '@storybook/addon-knobs';
+import Standard from 'components/layout/standard';
+import { article, matchReport, review } from 'fixtures/item';
+import { renderAll } from 'renderer';
+
+// ----- Stories ----- //
+
+const display = ArticleDisplay.Standard;
+
+export const Article = (): React.ReactNode => {
+	return (
+		<Standard item={article}>
+			{renderAll(
+				{
+					theme: ArticlePillar.News,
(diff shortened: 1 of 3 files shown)


MarSavar:
[More Layout Stories](https://trello.com/c/e9A8cH96/678-more-layout-stories)

apps-rendering/src/components/layout/standard.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+};
+Article.story = { name: 'Article' };
+
+export const Review = (): React.ReactNode => {
+	return (
+		<Standard item={review}>
+			{renderAll(
+				{
+					theme: ArticlePillar.Culture,
+					design: ArticleDesign.Standard,
```
JamieB-gu, line 36: I think this should be `ArticleDesign.Review`?

I suspect this still renders the stars because `review` has `design === ArticleDesign.Review`. Possibly the API for the layouts could be improved to prevent the possibility of this mismatch 🤔

apps-rendering/src/components/layout/standard.stories.tsx
```diff
@@ -0,0 +1,77 @@
+// ----- Imports ----- //
+import { ArticleDesign, ArticleDisplay, ArticlePillar } from '@guardian/libs';
+import { breakpoints } from '@guardian/source-foundations';
+import { partition } from '@guardian/types';
+import { withKnobs } from '@storybook/addon-knobs';
```
JamieB-gu, line 5: I'd recommend against adding new usages of this library; it was deprecated in favour of `addon-controls`.

You could replace one with the other, but I'm not sure any of these stories actually use controls, so perhaps it could be dropped from here entirely?

JamieB-gu (commented):
Nice! Couple of comments.

JamieB-gu (approved):

PR #4064, opened 2022-02-23
Remove The `components/shared` Directory

JamieB-gu:
## Why?

Everything in the main `components` directory is "shared". The `shared` sub-directory is a holdover from an earlier phase of AR. Now it's an extra bit of cognitive load for people currently working on the codebase to think about, so this PR removes it. All components within have been moved to the main `components` directory.

## Changes

- Removed the `components/shared` directory
- Moved all components into main `components` directory


mxdvl (approved):
seems sound!

PR #4065, opened 2022-02-23
GridItem `info`

oliverlloyd:
## What does this change?
This creates the `info` grid area, which contains `lines`, `meta`, `keyevents` & `matchstats`

## Why?
Because this allows us to keep all this content together in the left column without creating space in the main column

### Before
Screenshot 2022-02-23 at 13 04 33

### After
Screenshot 2022-02-23 at 13 04 01


JamieB-gu (approved):

mxdvl (commented):
nice one!

PR #4066, opened 2022-02-23
Update `atom.video.spec` to check consent is sent to YouTube

arelra:
## What does this change?

Updates `atom.video.spec` to check consent is sent to YouTube in both consented and reject-all flows.

## Why?

So we can catch inadvertent bugs that might cause consent not to be sent to YouTube.

### After

Consented:

Screenshot 2022-02-23 at 17 19 26

Reject All:

Screenshot 2022-02-23 at 13 55 06

### DIFF
--- a/dotcom-rendering/cypress/integration/parallel-5/atom.video.spec.js
+++ b/dotcom-rendering/cypress/integration/parallel-5/atom.video.spec.js
@@ -21,20 +21,6 @@ const interceptPlayEvent = (id) => {
 	);
 };
 
-const parseCustParams = (custParamsString) => {
-	const custParams = decodeURIComponent(custParamsString);
-	const custParamsItems = custParams.split('&');
-	return custParamsItems.reduce((newParams, cpi) => {
-		const [key, value] = cpi.split('=');
-		if (value.includes(',')) {
-			newParams[key] = value.split(',').map(decodeURIComponent);
-		} else {
-			newParams[key] = decodeURIComponent(value);
-		}
-		return newParams;
-	}, {});
-};
-
 const interceptYouTubeEmbed = ({ videoId, adUnit, pageUrl, rejectAll }) => {
 	return cy.intercept(
 		{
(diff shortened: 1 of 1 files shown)


dotcom-rendering/cypress/integration/parallel-5/atom.video.spec.js
```diff
@@ -21,6 +22,71 @@ const interceptPlayEvent = (id) => {
 	);
 };
 
+const parseCustParams = (custParamsString) => {
+	const custParams = decodeURIComponent(custParamsString);
```
jamesgorrie, line 26: You can use [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) which has a nicer API (and we could remove this function).

oliverlloyd (approved):
Really great to get more coverage in this area 👍

jamesgorrie (approved):
Really nice to have certainty in this important area!

dotcom-rendering/cypress/integration/parallel-5/atom.video.spec.js
arelra: Thanks for the suggestion! That works nicely for the `cust_params` field.

dotcom-rendering/cypress/integration/parallel-5/atom.video.spec.js
```diff
@@ -21,6 +21,60 @@ const interceptPlayEvent = (id) => {
...
+				decodeURIComponent(adTagParameters.cust_params),
+			);
+			// check consent related properties
+			// cmpGdpr = consentState.tcfv2.gdprApplies
+			expect(adTagParameters.cmpGdpr, 'check GDPR applies').to.equal(1);
+			// cmpVcd = consentState.tcfv2.tcString
+			expect(adTagParameters.cmpVcd, 'check TCFV2 tcString').to.not.be
+				.undefined;
+			if (rejectAll) {
+				// user has NOT consented to all purposes
```
kenoir, line 48:
```suggestion
				// user has not given consent for any purpose
```

kenoir (approved):
Nice ✨

PR #4068, opened 2022-02-23
Repurpose `data-spacefinder-ignore` attribute

simonbyford:
## What does this change?

Rename `data-spacefinder-ignore` to `data-spacefinder-component`. Furthermore, add support for the value `rich-link`.

## Why?

It will allow spacefinder to identify rich links in this PR: https://github.com/guardian/frontend/pull/24690

### DIFF
--- a/dotcom-rendering/src/web/components/Figure.tsx
+++ b/dotcom-rendering/src/web/components/Figure.tsx
@@ -175,13 +175,20 @@ export const Figure = ({
 			</figure>
 		);
 	}
+
+	// See 001-commercial-selectors.md for details on `data-spacefinder-component`
+	let spacefinderComponent: 'rich-link' | 'numbered-list-title' | undefined;
+	if (role === 'richLink') {
+		spacefinderComponent = 'rich-link';
+	} else if (isNumberedListTitle) {
+		spacefinderComponent = 'numbered-list-title';
+	}
+
 	return (
 		<figure
 			id={id}
 			css={defaultRoleStyles(role)}
-			data-spacefinder-ignore={
-				isNumberedListTitle ? 'numbered-list-title' : null
-			}
+			data-spacefinder-component={spacefinderComponent}
 			className={className}
 		>
 			{children}
--- a/dotcom-rendering/docs/contracts/001-commercial-selectors.md
+++ b/dotcom-rendering/docs/contracts/001-commercial-selectors.md
@@ -7,6 +7,17 @@ We place the following classes on the container element of the article body:
 -   `article-body-commercial-selector` on the ArticleRenderer
 -   `js-liveblog-body` on the ArticleBody when rendering a Liveblog/Deadblog
 
+Furthermore, within the article body, we add the following attribute to certain elements:
+
+- `data-spacefinder-component`
+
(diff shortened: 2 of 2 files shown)


chrislomaxjones:
Looks good! One thought, it might be useful to add this attribute to [this documentation](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/docs/contracts/001-commercial-selectors.md)!

dotcom-rendering/src/web/components/Figure.tsx
```diff
@@ -175,13 +175,19 @@ export const Figure = ({
 			</figure>
 		);
 	}
+
+	let spacefinderComponent: 'rich-link' | 'numbered-list-title' | null = null;
```
arelra, line 179: Might be help to have some comments describing how this will be used?

dotcom-rendering/src/web/components/Figure.tsx
```diff
@@ -175,13 +175,19 @@ export const Figure = ({
...
+	if (role === 'richLink') {
+		spacefinderComponent = 'rich-link';
+	} else if (isNumberedListTitle) {
+		spacefinderComponent = 'numbered-list-title';
+	}
```
arelra, line 184: In the `else` condition should we add `data-spacefinder-component` as an attribute?

dotcom-rendering/src/web/components/Figure.tsx
simonbyford: Good idea, I'll add

dotcom-rendering/src/web/components/Figure.tsx
simonbyford: Don't quite follow, can you clarify what you mean?

arelra: Sure - so we currently have `if (x) else if (y)`.

I was wondering what happens (and perhaps what should happen) in the unhandled `else` condition.

But I can see now that the value of `data-spacefinder-component` will be `null` so the attribute won't actually be added to the dom!

Sorry please proceed :)

mxdvl (commented):
Maybe favour `undefined` over `null`?

dotcom-rendering/src/web/components/Figure.tsx
simonbyford: Yeah, that's a fair point, I'll use `undefined` instead

dotcom-rendering/src/web/components/Figure.tsx
simonbyford: Ahhhh gotcha. Agree the intent isn't super clear but hopefully adding to the docs will help :)

simonbyford:
> Looks good! One thought, it might be useful to add this attribute to [this documentation](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/docs/contracts/001-commercial-selectors.md)!

Good shout, I've added!

arelra (approved):

chrislomaxjones (commented):
Thanks for adding the documentation 🙏

PR #4069, opened 2022-02-23
Fix padding on immersive article series section link

DavidLawes:
## What does this change?
There were alignment issues on the badge & series link on immersive layouts (where the immersive article included an immersive header). This change ensures the badge (or just the series link) is always aligned with the headline directly below.

Note that this will result in 2 different types of visual diff:
- At lower breakpoints immersive headers with a badge will have additional left padding. New stories have been added to capture these requirements at lower breakpoints.
- At higher breakers immersive headers without a badge will also have additional left padding. We already have stories to cover this scenario so can expect to see chromatic diffs.

## Why?
To ensure immersive articles with an immersive header more closely meet the intended design.

### Without badge

https://www.theguardian.com//travel/2020/dec/09/my-year-of-roaming-free-in-cornwall-photo-essay-cat-vinton

#### Mobile:

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/45561419/155347344-263f084b-998c-40ff-871d-235ef3dc6170.png
[after]: https://user-images.githubusercontent.com/45561419/155347422-ff32d9be-f8ae-462c-bf07-475f2c7229bf.png

#### Tablet

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: 
Screenshot 2022-02-23 at 15 12 39

[after]:

### DIFF
--- a/dotcom-rendering/src/web/components/SeriesSectionLink.tsx
+++ b/dotcom-rendering/src/web/components/SeriesSectionLink.tsx
@@ -84,13 +84,6 @@ const invertedStyle = css`
 	padding-right: ${space[1]}px;
 	padding-top: ${space[1]}px;
 	padding-bottom: ${space[3]}px;
-	padding-left: ${space[3]}px;
-	${from.mobileLandscape} {
-		padding-left: ${space[5]}px;
-	}
-	${from.tablet} {
-		padding-left: ${space[1]}px;
-	}
 `;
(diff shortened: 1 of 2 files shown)


ioannakok (approved):
Thanks for the great description! 👍

PR #4070, opened 2022-02-23
refactor: Remove dependency on shared article count state in App.tsx

AshCorr:
## What does this change?

Refactors App.tsx to remove the dependency on asyncArticleCount shared state. This has been moved to a library that uses the global window object for managing this state instead.

## Why?

Eventually App.tsx is going to be removed and this allows the SlotBodyEnd and StickyBottomBanner components be moved to their own respective islands


dotcom-rendering/src/lib/article-count.ts
```diff
@@ -0,0 +1,32 @@
...
+import { hasOptedOutOfArticleCount } from '../web/lib/contributions';
+import { incrementDailyArticleCount } from '../web/lib/dailyArticleCount';
+
+// Log an article view using the Slot Machine client lib
+// We should monitor this function call to ensure it only happens within an
+// article pages when other pages are supported by DCR.
+export const getArticleCount = async (pageId: string, keywordIds: string) => {
+	const hasOptedOut = await hasOptedOutOfArticleCount();
+
+	// hasOptedOut could potentially be moved into the scope of the below if condition
```
AshCorr, line 15: Does anybody know enough about how async works in JS that could confirm this?

tomrf1: I think you are correct.
The `window.guardian.articleCount` check must happen after the await.
Otherwise if two invocations await at the same time then both would go onto increment and update `window.guardian.articleCount`

tomrf1 (approved):
Nice!
I just tested locally to confirm:
- AC updates on new articles
- correct AC is sent in epic and banner requests, after it's incremented
- AC opt-out still works

dotcom-rendering/src/lib/article-count.ts
```diff
@@ -0,0 +1,32 @@
+import { storage } from '@guardian/libs';
+import {
+	getWeeklyArticleHistory,
+	incrementWeeklyArticleCount,
+} from '@guardian/support-dotcom-components';
+import { hasOptedOutOfArticleCount } from '../web/lib/contributions';
+import { incrementDailyArticleCount } from '../web/lib/dailyArticleCount';
+
+// Log an article view using the Slot Machine client lib
```
jamesgorrie, line 9: What is Slot Machine? I can't see it referenced in the code.

dotcom-rendering/src/web/components/SlotBodyEnd/SlotBodyEnd.tsx
```diff
@@ -122,6 +131,7 @@ export const SlotBodyEnd = ({
 			contributionsServiceUrl,
 			idApiUrl,
 			stage,
+			// useOnce makes sure that all deps aren't undefined making it safe to cast this.
```
jamesgorrie, line 134: Not sure I understand this - what are `all deps`?

dotcom-rendering/src/lib/article-count.ts
```diff
@@ -0,0 +1,32 @@
...
+// We should monitor this function call to ensure it only happens within an
+// article pages when other pages are supported by DCR.
+export const getArticleCount = async (pageId: string, keywordIds: string) => {
+	const hasOptedOut = await hasOptedOutOfArticleCount();
+
```
jamesgorrie, line 14: We could just `return undefined` and stop execution. e.g.
```suggestion
if (hasOptedOut) { return }
```

which might make the below statement a bit nicer to understand i.e. update and set if not set.

OllysCoding (approved):
+1 This looks good to me.

dotcom-rendering/src/lib/article-count.ts
jamesgorrie: Worth updating this comment to be a statement rather than a question?

jamesgorrie (approved):
Few comments on the comments.

dotcom-rendering/src/lib/article-count.ts
tomrf1: this is an old comment - slot machine was an experiment which didn't really go anywhere.
I suggest deleting the comment

dotcom-rendering/src/lib/article-count.ts
```diff
@@ -0,0 +1,32 @@
...
+	getWeeklyArticleHistory,
+	incrementWeeklyArticleCount,
+} from '@guardian/support-dotcom-components';
+import { hasOptedOutOfArticleCount } from '../web/lib/contributions';
+import { incrementDailyArticleCount } from '../web/lib/dailyArticleCount';
+
+// Log an article view using the Slot Machine client lib
+// We should monitor this function call to ensure it only happens within an
+// article pages when other pages are supported by DCR.
+export const getArticleCount = async (pageId: string, keywordIds: string) => {
```
tomrf1, line 12: it would be good to declare the return type here - `Promise<WeeklyArticleLog[] | undefined>`

mxdvl: Isn’t it `WeeklyArticleHistory`?

dotcom-rendering/src/lib/article-count.ts
mxdvl: Or even:

```ts
if(await hasOptedOutOfArticleCount()) return;
```

dotcom-rendering/src/lib/article-count.ts
tomrf1: True, `type WeeklyArticleHistory = WeeklyArticleLog[]`

`undefined` means no article count (perhaps because they've opted out), but it's not a failure.
I think the promise would only fail if CMP failed.

dotcom-rendering/src/lib/article-count.ts
AshCorr: The existing solution still tries to get the article count even if the user is opted out, I'll have a look to see why and see if its possible to do this.

AshCorr: The existing solution still tries to get the article count even if the user is opted out, I'll have a look to see why and see if its possible to do this.

tomrf1: For context - the local storage item is deleted when a user opts out, so `getWeeklyArticleHistory` should return `undefined` in this case.
But this new function is a nice opportunity to be explicit about returning `undefined` if they've opted out

PR #4073, opened 2022-02-23
Add darkmode styles to Liveblog/Deadblog

iainjchambers-guardian:
## What does this change?
Makes styles available for darkmode for both Liveblogs and Deadblogs

| Before |  After |
| ---- | ---- |
| image | image |

### DIFF
--- a/apps-rendering/src/components/list.tsx
+++ b/apps-rendering/src/components/list.tsx
@@ -2,12 +2,18 @@
 
 import { css } from '@emotion/react';
 import type { SerializedStyles } from '@emotion/react';
+import { background } from '@guardian/common-rendering/src/editorialPalette';
+import type { ArticleFormat } from '@guardian/libs';
 import { neutral, remSpace } from '@guardian/source-foundations';
 import type { FC, ReactNode } from 'react';
+import { darkModeCss } from 'styles';
 
 // ----- Component ----- //
 
-const styles: SerializedStyles = css`
+const styles = (
+	usePillarColour: boolean,
+	format: ArticleFormat,
+): SerializedStyles => css`
 	list-style: none;
 	margin: ${remSpace[3]} 0;
 	padding-left: 0;
@@ -23,14 +29,22 @@ const styles: SerializedStyles = css`
 		background-color: ${neutral[86]};
 		margin-left: -${remSpace[6]};
 		vertical-align: middle;
+
+		${darkModeCss`
+			background-color: ${background.bulletDark(format, usePillarColour)};
+		`}
 	}
 `;
 
 interface Props {
 	children: ReactNode;
+	usePillarColour: boolean;
+	format: ArticleFormat;
 }
 
-const List: FC<Props> = ({ children }) => <ul css={styles}>{children}</ul>;
+const List: FC<Props> = ({ children, usePillarColour, format }) => (
+	<ul css={styles(usePillarColour, format)}>{children}</ul>
+);
 
 // ----- Exports ----- //
 
--- a/apps-rendering/src/components/liveblogHeader.tsx
+++ b/apps-rendering/src/components/liveblogHeader.tsx
@@ -1,6 +1,5 @@
 // ----- Imports ----- //
 
-import { background } from '@guardian/common-rendering/src/editorialPalette';
 import {
 	Column,
 	Columns,
@@ -9,6 +8,10 @@ import {
 } from '@guardian/source-react-components';
 import Headline from 'components/headline';
 import Standfirst from 'components/standfirst';
+import {
+	headlineBackgroundColour,
+	standfirstBackgroundColour,
+} from 'editorialStyles';
 import type { DeadBlog, LiveBlog } from 'item';
 import { getFormat } from 'item';
 import type { FC } from 'react';
@@ -26,10 +29,7 @@ const LiveblogHeader: FC<Props> = ({ item }) => {
 
 	return (
 		<header>
-			<Container
-				element="div"
-				backgroundColor={background.headline(format)}
-			>
+			<Container element="div" css={headlineBackgroundColour(format)}>
 				<Columns collapseUntil="desktop">
 					<Column span={3}>
 						<Series item={item} />
@@ -39,10 +39,7 @@ const LiveblogHeader: FC<Props> = ({ item }) => {
 					</Column>
 				</Columns>
 			</Container>
-			<Container
-				element="div"
-				backgroundColor={background.standfirst(format)}
-			>
+			<Container element="div" css={standfirstBackgroundColour(format)}>
 				<Columns collapseUntil="desktop">
 					<Column span={3}>
 						<Hide below="desktop">
--- a/apps-rendering/src/components/metadata.tsx
+++ b/apps-rendering/src/components/metadata.tsx
@@ -2,6 +2,7 @@
 
 import { css } from '@emotion/react';
 import type { SerializedStyles } from '@emotion/react';
+import { background } from '@guardian/common-rendering/src/editorialPalette';
 import { ArticleDesign, ArticleDisplay } from '@guardian/libs';
 import type { ArticleFormat } from '@guardian/libs';
 import {
@@ -78,6 +79,10 @@ const metaBottomStyles = (isLive: boolean): SerializedStyles => css`
 	}
 
 	${!isLive && `border-top: 1px solid rgba(153, 153, 153, 0.4);`}
+
+	${darkModeCss`
+		border-top-color: ${neutral[20]};
+	`}
 `;
 
 const toggleStyles = css`
@@ -110,6 +115,10 @@ const toggleOverrideStyles = css`
 			background-color: rgba(255, 255, 255, 0.5);
 		}
 	}
+
+	${darkModeCss`
+		color: ${neutral[93]};
+	`}
 `;
 
 const liveBylineStyles = css`
@@ -154,22 +163,22 @@ const tempraryBackgroundStyle = (format: ArticleFormat): SerializedStyles => {
 				${from.desktop} {
 					background-color: ${neutral[97]};
 				}
-				@media (prefers-color-scheme: dark) {
-					background-color: ${neutral[10]};
-				}
+
+				${darkModeCss`
+					background-color: ${background.articleContentDark(format)};
+				`}
 			`;
 		default:
 			return css`
 				background-color: ${themeStyles.liveblogBackground};
+
 				${from.desktop} {
 					background-color: ${neutral[97]};
 				}
-				@media (prefers-color-scheme: dark) {
-					background-color: ${themeStyles.liveblogDarkBackground};
-					${from.desktop} {
-						background-color: ${neutral[10]};
-					}
-				}
+
+				${darkModeCss`
+					background-color: ${background.articleContentDark(format)};
+				`}
 			`;
 	}
 };
--- a/apps-rendering/src/components/series.tsx
+++ b/apps-rendering/src/components/series.tsx
@@ -63,10 +63,12 @@ const immersiveLabsLinkStyles = css`
 
 const blogLinkStyles = (format: ArticleFormat): SerializedStyles => css`
 	${headline.xxxsmall({ lineHeight: 'tight', fontWeight: 'bold' })}
-	color: ${format.design === ArticleDesign.LiveBlog
-		? getThemeStyles(format.theme).liveblogKicker
-		: text.seriesTitle(format)};
+	color: ${text.seriesTitle(format)};
 	text-decoration: none;
+
+	${darkModeCss`
+		color: ${text.seriesTitleDark(format)};
+	`}
 `;
 
 const getLinkStyles = ({
--- a/apps-rendering/src/components/tags.tsx
+++ b/apps-rendering/src/components/tags.tsx
@@ -2,14 +2,10 @@
 
 import type { SerializedStyles } from '@emotion/react';
 import { css } from '@emotion/react';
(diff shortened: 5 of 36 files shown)


MarSavar:
Tested the branch locally and it looks great! 🎉 
Double checked using a football liveblog and it also looks like it's styled correctly, but @HarryFischer will have the final word on that.
The only comment I have is the same as @mxdvl re: formatting, i.e. the diff in `editorialPalette` is a bit noisy.

common-rendering/src/components/liveBlockContainer.tsx
```diff
@@ -135,6 +135,12 @@ const LiveBlockContainer = ({
 					padding-left: ${LEFT_MARGIN_DESKTOP}px;
 				}
 				border-bottom: 1px solid ${neutral[86]};
+
+				@media (prefers-color-scheme: dark) {
+					background-color: ${neutral[10]};
+					color: ${neutral[100]};
+					border-bottom: 0;
+				}
```
JamieB-gu, line 143: This component is also available for DCR, where dark mode isn't supported. We have the `darkModeCss` function in `common-rendering/lib.ts` for this:

```suggestion
				${darkModeCss(supportsDarkMode)`
					background-color: ${neutral[10]};
					color: ${neutral[100]};
					border-bottom: 0;
				`}
```

And `LiveBlockContainer` will take `supportsDarkMode: boolean` as a prop. See the `KeyEvents` component for an example.

**Aside:** I think the CSS vars solution may simplify this. If we use a CSS variable that has the same name in both light and dark mode, while dark mode remains unsupported in dotcom that variable will always be set to the light mode variant.

AshCorr:
> Would it be worth creating Chromatic stories to capture this?

We need to look at how we do darkmode testing in Chromatic. We're relying on `@media (prefers-color-scheme: dark)` to enable darkmode for users that want it and unfortunately it doesn't seem like its possible to simulate a media query in Storybook / Chromatic.

oliverlloyd:
> > Would it be worth creating Chromatic stories to capture this?
> 
> We need to look at how we do darkmode testing in Chromatic. We're relying on `@media (prefers-color-scheme: dark)` to enable darkmode for users that want it and unfortunately it doesn't seem like its possible to simulate a media query in Storybook / Chromatic.

[This document](https://github.com/guardian/source/discussions/1060) contains ideas for setting a page state (using a class or data attribute on the body) which might work better with Storybook

JamieB-gu:
> We need to look at how we do darkmode testing in Chromatic. We're relying on `@media (prefers-color-scheme: dark)` to enable darkmode for users that want it and unfortunately it doesn't seem like its possible to simulate a media query in Storybook / Chromatic.

Unfortunately it's not possible to mutate the browser preference for dark mode using the web APIs. As you say this makes it difficult to replicate in storybook, and is one of the reasons it's proving more difficult to implement dark mode on dotcom than on apps.

I suspect that a CSS vars-based approach, variants of which I've discussed with @oliverlloyd and @iainjchambers-guardian at different times, might also solve this problem 🤔

apps-rendering/src/components/list.tsx
```diff
@@ -23,14 +29,28 @@ const styles: SerializedStyles = css`
 		background-color: ${neutral[86]};
 		margin-left: -${remSpace[6]};
 		vertical-align: middle;
+
+		${darkModeCss`
+		${
+			isStandfirst && format
+				? `background-color: ${background.liveblogStandfirstBulletDark(
```
JamieB-gu, line 36: I think you could just call this `standfirstBulletDark`? The liveblog part of it is taken care of by the presence of the `format` argument?

apps-rendering/src/components/list.tsx
```diff
@@ -23,14 +29,28 @@ const styles: SerializedStyles = css`
...
+				: `background-color: ${neutral[46]};`
+		}
+		`}
 	}
 `;
 
 interface Props {
 	children: ReactNode;
+	isStandfirst?: boolean;
+	format?: ArticleFormat;
```
JamieB-gu, line 48: Can we make these non-optional? I think it's good to be explicit with function/component arguments, and we have `format` and `isStandfirst` available in the other place this is used in `renderer`. I also think it makes the condition above slightly simpler because you don't have to include the check for the existence of `format`.

apps-rendering/src/components/liveblogHeader.tsx
```diff
@@ -26,10 +29,7 @@ const LiveblogHeader: FC<Props> = ({ item }) => {
 
 	return (
 		<header>
-			<Container
-				element="div"
-				backgroundColor={background.headline(format)}
-			>
+			<Container element="div" css={headlineBackgroundColour(format)}>
```
JamieB-gu, line 32: I was thinking we'd have to implement a `backgroundColorDark` prop in the API of this container, but actually this works and it's very simple 🤔

I'm happy to go with this approach, but FYI @SiAdcock I think you'd be interested to know this is how the API is being used to achieve dark mode support?

apps-rendering/src/components/metadata.tsx
```diff
@@ -153,22 +162,22 @@ const tempraryBackgroundStyle = (format: ArticleFormat): SerializedStyles => {
 				${from.desktop} {
 					background-color: ${neutral[97]};
 				}
-				@media (prefers-color-scheme: dark) {
-					background-color: ${neutral[10]};
-				}
+
+				${darkModeCss`
+					background-color: ${background.articleContentDark(format)};
+				`}
```
JamieB-gu, line 168: Just double-checking you noticed the comment at the top of this function? 🙂 I think that was from @marjisound's work on this?

apps-rendering/src/components/tags.tsx
```diff
@@ -38,7 +34,7 @@ const styles = css`
 	${textSans.medium()}
 
 	${darkModeCss`
-		background: ${background.inverse};
+		background-color: ${background.articleContentDark(format)};
```
JamieB-gu, line 37: 👍

apps-rendering/src/components/tags.tsx
```diff
@@ -64,7 +60,7 @@ const anchorStyles = (format: ArticleFormat): SerializedStyles => css`
 	line-height: 1;
 
 	${darkModeCss`
-		color: ${neutral[60]};
+		color: ${neutral[86]};
```
JamieB-gu, line 63: Just making sure we're intentionally changing this for all articles?

common-rendering/src/editorialPalette/background.ts
```diff
@@ -125,13 +124,51 @@ const standfirstDark = ({ design, theme }: ArticleFormat): Colour => {
...
+				case ArticlePillar.Sport:
+					return sport[500];
+				case ArticlePillar.Culture:
+					return culture[500];
+				case ArticlePillar.Lifestyle:
+					return lifestyle[500];
+				case ArticlePillar.News:
+					return news[550];
+				default:
+					return neutral[46];
```
JamieB-gu, line 147: You may want the other themes (labs and special reports) to use the same colour as news. See https://github.com/guardian/dotcom-rendering/pull/4195#discussion_r820642373

dotcom-rendering/src/web/components/LiveBlock.tsx
```diff
@@ -43,7 +43,8 @@ export const LiveBlock = ({
 	return (
 		<LiveBlockContainer
 			id={block.id}
-			borderColour={palette.border.liveBlock}
+			format={format}
+			supportsDarkMode={true}
```
JamieB-gu, line 47: I suspect you want this to be `false`? Dotcom doesn't support dark mode yet.

dotcom-rendering/src/web/components/LiveBlock.tsx
```diff
@@ -62,6 +63,7 @@ export const LiveBlock = ({
 					index,
 					pageId,
 					webTitle,
+					supportsDarkMode: true,
```
JamieB-gu, line 66: I suspect you want this to be `false`? Dotcom doesn't support dark mode yet.

JamieB-gu (commented):
A lot of great stuff here!

I've written quite a few comments, sorry! But most of them are relatively small questions.

dotcom-rendering/src/web/components/LiveBlock.tsx
iainjchambers-guardian: I'm going to add a `supportsDarkMode` prop to this component, default it to false,  and pass it down from AR

dotcom-rendering/src/web/components/LiveBlock.tsx
iainjchambers-guardian: See above

dotcom-rendering/src/web/components/TextBlockComponent.tsx
```diff
@@ -125,6 +128,7 @@ export const TextBlockComponent = ({
 	format,
 	forceDropCap,
 	isFirstParagraph,
+	supportsDarkMode = false,
```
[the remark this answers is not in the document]
iainjchambers-guardian: I've done that because the LiveBlocks used in AR are not a shared component, they're in DCR

common-rendering/src/editorialPalette/background.ts
iainjchambers-guardian: That's an odd one, I see your point and I agree but I think I just copied the switch from the standfirst background so is there a wider issue of missing styles for those special article types.

dotcom-rendering/src/web/components/TextBlockComponent.tsx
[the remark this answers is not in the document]
JamieB-gu: Are you absolutely sure? Isn't this what we're using?

- The `LiveBlocks` component: https://github.com/guardian/dotcom-rendering/blob/main/apps-rendering/src/components/liveBlocks.tsx
- The `LiveBlock` model: https://github.com/guardian/dotcom-rendering/blob/main/apps-rendering/src/liveBlock.ts

apps-rendering/src/components/metadata.tsx
```diff
@@ -153,22 +162,22 @@ const tempraryBackgroundStyle = (format: ArticleFormat): SerializedStyles => {
...
 		default:
 			return css`
 				background-color: ${themeStyles.liveblogBackground};
+
 				${from.desktop} {
 					background-color: ${neutral[97]};
 				}
-				@media (prefers-color-scheme: dark) {
-					background-color: ${themeStyles.liveblogDarkBackground};
-					${from.desktop} {
```
AshCorr, line 168: Don't need this anymore?

apps-rendering/src/components/series.tsx
```diff
@@ -63,10 +63,20 @@ const immersiveLabsLinkStyles = css`
 
 const blogLinkStyles = (format: ArticleFormat): SerializedStyles => css`
 	${headline.xxxsmall({ lineHeight: 'tight', fontWeight: 'bold' })}
-	color: ${format.design === ArticleDesign.LiveBlog
+	color: ${format.design === ArticleDesign.LiveBlog ||
```
AshCorr, line 66: I could be wrong, but I think this function is only ever called in the context of a LiveBlog or a DeadBlog so this condition will always be true?

apps-rendering/src/components/series.tsx
```diff
@@ -63,10 +63,20 @@ const immersiveLabsLinkStyles = css`
...
+	format.design === ArticleDesign.DeadBlog
 		? getThemeStyles(format.theme).liveblogKicker
 		: text.seriesTitle(format)};
 	text-decoration: none;
+
+	${darkModeCss`
+		${
+			format.design === ArticleDesign.LiveBlog ||
```
AshCorr, line 74: > I could be wrong, but I think this function is only ever called in the context of a LiveBlog or a DeadBlog so this condition will always be true?

Same here!

common-rendering/src/editorialPalette/background.ts
```diff
@@ -125,13 +124,61 @@ const standfirstDark = ({ design, theme }: ArticleFormat): Colour => {
...
+		}
+	}
+
+	switch (design) {
+		case ArticleDesign.DeadBlog:
+		case ArticleDesign.LiveBlog:
+			return neutral[46];
+		default:
+			return neutral[46];
+	}
```
JamieB-gu, line 161: Maybe?

```suggestion
	return neutral[46];
```

apps-rendering/src/components/series.tsx
JamieB-gu: I think you're right!

apps-rendering/src/components/series.tsx
JamieB-gu: 👍

apps-rendering/src/components/metadata.tsx
JamieB-gu: Are you double-checking this wasn't an accident?

common-rendering/src/editorialPalette/text.ts
```diff
@@ -447,19 +475,38 @@ const kicker = (format: ArticleFormat): Colour => {
...
+			return sport[600];
 		case ArticlePillar.Culture:
-			return culture[300];
+			return culture[600];
 		case ArticlePillar.Opinion:
-			return opinion[300];
+			return opinion[600];
 		case ArticleSpecial.Labs:
-			return labs[300];
+			return labs[500];
```
JamieB-gu, line 488: I think the build is failing because this colour doesn't exist?

JamieB-gu (approved):
🚀

PR #4077, opened 2022-02-23
Fix interactive embeds

ioannakok:
## What does this change?
This PR consists of three improvements to the way the interactive atoms are rendered on DCR.

1. We stopped Interactive Atoms (which are themselves `<figure>`s) from rendering inside an additional, outer `<figure>` tag. Instead, these are now rendering using the new `<gu-island>` tags which allow them to render when visible rather than on original page load (more efficient), and we no longer have styling or SEO/semantic issues as a result of placing a `<figure>` within a `<figure>`

2. Added `defaultRoleStyles(role)` to the remaining figure class, so that it gets the up-to-date weighting styles using this method in DCR.

3.  We added the correct legacy atom weighting classnames to DCR. Previously, atoms were given weighting using BEM styling, where the weight was the modifier appended with two dashes - eg. `--showcase` or `--inline` - this had not been properly ported over before this PR and caused some styling issues

## Why?
To keep our interactive articles looking great into the future 

### Before
Problem 1:
figure in figure old

Problems 2:
Screenshot 2022-03-08 at 16 53 36

Problems 3:
legacy class name old

### After
Fix 1:
figure in figure new

Fix 2:
Screenshot 2022-03-08 at 16 53 54

Fix 3:
legacy class name new


ioannakok:
[InteractiveBlockComponent](https://trello.com/c/zzkcNzTg/587-interactiveblockcomponent)

OllysCoding (approved):
+1 This looks great!

dotcom-rendering/index.d.ts
```diff
@@ -873,14 +873,7 @@ interface ComponentNameChunkMap {
 	addWhen: BlockElementType;
 }
 
-interface InteractiveBlockLoadable extends ComponentNameChunkMap {
-	chunkName: 'InteractiveBlockComponent';
-	addWhen: InteractiveBlockElement['_type'];
-}
-
 // There are docs on loadable in ./docs/loadable-components.md
```
DavidLawes, line 876: Do we still need the comment if the corresponding type definition has been removed?

DavidLawes (approved):
🎉

dotcom-rendering/index.d.ts
ioannakok: Good catch! Deleting this

PR #4082, opened 2022-02-24
`mockLiveUpdate`

oliverlloyd:
## What does this change?
This adds a function `mockLiveUpdate` on the `window` object. This function is a reference to the `onSuccess` function that gets called after a live update request is made on a live blog

## Why?
We need a way to test liveblogs using Cypress. By exposing this code like this it becomes available to our test scripts without them needing to depend on polling.

## Why `window`?
Using the `window` object is convenient because it is accessible to Cypress using [cy.window()](https://docs.cypress.io/api/commands/window)


dotcom-rendering/window.guardian.d.ts
```diff
@@ -68,6 +68,11 @@ declare global {
 			onConsentChange?: (fn: Callback) => void;
 			getConsentFor?: (fn: Callback) => void;
 		};
+		mockLiveUpdate: (data: {
+			numNewBlocks: number;
+			html: string;
+			mostRecentBlockId: string;
+		}) => void;
```
mxdvl, line 75: If you import onSuccess from that file, you can keep everything in sync.

```suggestion
		mockLiveUpdate: typeof onSuccess
```

mxdvl (approved):
I think it’s best to have our type declaration get its type from functions defined in our code, but otherwise I think it’s good to be able to mock!

dotcom-rendering/window.guardian.d.ts
mxdvl: Sorry, you’ll also need to import the method in this file!

dotcom-rendering/window.guardian.d.ts
```diff
@@ -68,6 +68,7 @@ declare global {
 			onConsentChange?: (fn: Callback) => void;
 			getConsentFor?: (fn: Callback) => void;
 		};
+		mockLiveUpdate: typeof onSuccess
```
jamesgorrie, line 71: Does this need to be imported here?

dotcom-rendering/window.guardian.d.ts
oliverlloyd: I created `LiveUpdateType` here instead. Like this we don't need to export the function, which feels like a win in terms of reducing complexity in the code

dotcom-rendering/window.guardian.d.ts
oliverlloyd: As above, I created `LiveUpdateType` instead

PR #4083, opened 2022-02-24
Change `sign-in-gate` placeholder to `div`

simonbyford:
## What does this change?

Change `sign-in-gate` placeholder to `div`

## Why?

Spacefinder was incorrectly considering it when getting candidates for ad insertion

Tested on code, still works fine:

Screenshot 2022-02-24 at 13 22 14


lucymonie (approved):
🥳

PR #4085, opened 2022-02-24
`useIsInView`

oliverlloyd:
## What?
This PR renames `useHasBeenSeen` to `useIsInView` and adds the `repeat` prop

## Why?
Previously, useHasBeenSeen fired once and was an indication that an element had appeared. But now we've added the `repeat` prop so that we can tell when elements scroll into view, are scrolled away, and appear again.

This is useful for things like sticking videos.

### DIFF
--- a/dotcom-rendering/src/web/lib/useIsInView.ts
+++ b/dotcom-rendering/src/web/lib/useIsInView.ts
@@ -2,7 +2,10 @@ import { useEffect, useState, useRef } from 'react';
 import libDebounce from 'lodash.debounce';
 
 const useIsInView = (
-	options: IntersectionObserverInit & { debounce?: boolean },
+	options: IntersectionObserverInit & {
+		debounce?: boolean;
+		repeat?: boolean;
+	},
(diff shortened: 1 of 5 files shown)


dotcom-rendering/src/web/lib/useIsInView.ts
```diff
@@ -13,7 +16,10 @@ const useHasBeenSeen = (
 	// 200ms before the callback is executed
 	const intersectionFn: IntersectionObserverCallback = ([entry]) => {
 		if (entry.isIntersecting) {
-			setHasBeenSeen(true);
+			setIsInView(true);
+		}
+		if (options.repeat && !entry.isIntersecting) {
+			setIsInView(false);
 		}
```
mxdvl, line 23: Not a suggestion, more of a question: is the following easier to read?

```ts
		if (entry.isIntersecting) {
			setIsInView(true);
		} else if (options.repeat) {
			setIsInView(false);
		}
```

oliverlloyd: Much nicer! Done 👍

PR #4086, opened 2022-02-24
Check devDependencies in package.json

mxdvl:
## What does this change?

Automate checking these deps.

## Why?

We did it manually twice already in #4001 and #4050. Let the robot help us.

### DIFF
--- a/dotcom-rendering/scripts/env/check-deps.js
+++ b/dotcom-rendering/scripts/env/check-deps.js
@@ -1,0 +1,9 @@
+const pkg = require('../../package.json');
+
+if (pkg.devDependencies) {
+	const { warn, log } = require('./log');
+
+	warn('Don’t use devDependencies');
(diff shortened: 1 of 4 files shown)


dotcom-rendering/scripts/env/check-deps.js
```diff
@@ -0,0 +1,9 @@
+const pkg = require('../../package.json');
+
+if (pkg.devDependencies) {
+	const { warn, log } = require('./log');
+
+	warn('Don’t use devDependencies');
+	log('See https://github.com/guardian/dotcom-rendering/pull/4001');
+	process.exit(1);
+}
```
jamesgorrie, line 9: Could we check all `package.json` files?

jamesgorrie (approved):

dotcom-rendering/scripts/env/check-deps.js
mxdvl: Maybe but I didn't feel confident.

PR #4087, opened 2022-02-24
Use correct standfirst background colour for comment design

MarSavar:
## Why?
Standfirsts in AR-rendered pieces using the comment layout default to a white background, as the current `standfirstBackground` logic only supports deadblogs and liveblogs.

This PR fixes fixes the problem by setting the background colour to peach if the design is `ArticleDesign.Comment`.

### Before
image

### After
image

### DIFF
--- a/common-rendering/src/editorialPalette.ts
+++ b/common-rendering/src/editorialPalette.ts
@@ -633,6 +633,10 @@ const backgroundStandfirst = ({
 		}
 	}
 
+	if (design === ArticleDesign.Comment) {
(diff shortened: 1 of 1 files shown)


JamieB-gu (approved):

mxdvl:
Also the comment count appeared!

MarSavar:
[Fix standfirst background colour in comment = des](https://trello.com/c/sh13xRLZ/704-fix-standfirst-background-colour-in-comment-des)

PR #4090, opened 2022-02-24
Refactor AR story structure, add more examples

MarSavar:
## Why?

This PR refactors the way AR stories are shown in Storybook. It also adds more designs to the Standard Display section, specifically 

- Comment
- Letter
- Editorial
- Analysis

The new structure is:

`Layout folder > Display > Design`

which mirrors the structure of DCR stories.

### Before
image

### After
image


apps-rendering/src/components/layout/standard.stories.tsx
```diff
@@ -1,24 +1,47 @@
...
+import { deadBlog, live } from 'fixtures/live';
+import type { Item } from 'item';
+import type { ReactElement } from 'react';
 import { renderAll } from 'renderer';
+import Live from './live';
 
-// ----- Stories ----- //
+// ----- Functions ----- //
+
+const formatFromItem = (
```
MarSavar, line 26: I made this so we can optionally override the `item`'s display type by passing it explicitly. This way we can reuse the same fixture but force different display types (`Showcase`, `Immersive`, etc). 
I think as a long-term solution we should probably have automatically generated fixtures like DCR to showcase more variety but maybe for the time being this is ok?

MarSavar:
[Refactor AR stories and add more examples](https://trello.com/c/uYAZlfM9/705-refactor-ar-stories-and-add-more-examples)

apps-rendering/src/components/layout/standard.stories.tsx
```diff
@@ -1,24 +1,47 @@
...
+	item: Item,
+	forceDisplay: Option<ArticleDisplay>,
+): ArticleFormat => ({
+	theme: item.theme,
+	design: item.design,
+	display:
+		forceDisplay.kind === OptionKind.Some
+			? forceDisplay.value
+			: item.display,
```
JamieB-gu, line 35: Possible alternative, up to you which you prefer.

```suggestion
	display: withDefault(item.display)(forceDisplay),
```

JamieB-gu (approved):
🚀

apps-rendering/src/components/layout/standard.stories.tsx
MarSavar: Oh yes! Forgot about that. Much neater

PR #4092, opened 2022-02-25
Polyfill `Promise.AllSettled`

chrislomaxjones:
## What does this change?

Add `Promise.allSettled` to the polyfill.io bundle.

See corresponding [Frontend PR](https://github.com/guardian/frontend/pull/24700).

## Why?

We have a number of errors being [logged by Sentry](https://sentry.io/organizations/the-guardian/issues/?project=35463&query=is%3Aunresolved+feature%3Acommercial+promise.allSettled&statsPeriod=14d) (non-public URL) that arise from promise.AllSettled not being a function. These arise from use in the commercial bundle and seem related to older browsers that [do not support](https://caniuse.com/mdn-javascript_builtins_promise_allsettled) this function.

To resolve these errors, we can polyfill the function.

### Before

Polyfill.io bundle size: original size: 170.3KB / gzipped: 47.0KB

```
https://assets.guim.co.uk/polyfill.io/v3/polyfill.min.js?rum=0&features=es6,es7,es2017,es2018,es2019,default-3.6,HTMLPictureElement,IntersectionObserver,IntersectionObserverEntry,URLSearchParams,fetch,NodeList.prototype.forEach,navigator.sendBeacon,performance.now&flags=gated&callback=guardianPolyfilled&unknown=polyfill&cacheClear=1
```

### After

Polyfill.io bundle size with Promise.AllSettled:: 171.1KB / gzipped 47.2KB

```
https://assets.guim.co.uk/polyfill.io/v3/polyfill.min.js?rum=0&features=es6,es7,es2017,es2018,es2019,default-3.6,HTMLPictureElement,IntersectionObserver,IntersectionObserverEntry,URLSearchParams,fetch,NodeList.prototype.forEach,navigator.sendBeacon,performance.now,Promise.allSettled&flags=gated&callback=guardianPolyfilled&unknown=polyfill&cacheClear=1```


oliverlloyd:
I made this comment on the original PR but dropping it here as well!

There's an argument to be made against adding more and more polyfills. It's easy to see them as a quick solution but they come with a cost (the output from some polyfills is wild!) and I think we need to have a solid use case and justification for adding more. Ideally, if there is another way to solve your problem (write more basic code?) then that would - I think - be preferable and a better long term solution

Note also, that support for IE11 is in flux so if that is the only motivation we could perhaps revisit this?

OllysCoding (approved):

chrislomaxjones:
> There's an argument to be made against adding more and more polyfills. It's easy to see them as a quick solution but they come with a cost (the output from some polyfills is wild!) and I think we need to have a solid use case and justification for adding more. Ideally, if there is another way to solve your problem (write more basic code?) then that would - I think - be preferable and a better long term solution

That's a good point, I can take a look at the original usage again. My initial reaction was that it could potentially be less costly in terms of size to polyfill the function in older browsers only rather than ship our own equivalent polyfill in the bundle for everyone, but I didn't actually explore this. 

> Note also, that support for IE11 is in flux so if that is the only motivation we could perhaps revisit this?

On this I think quite a lot of the errors we've seen reported in Sentry originate from older browsers that we may continue to support (I don't know what our policy is on older versions of Mobile Safari for example).

mxdvl:
I think 0.6% increase i. The polyfill bundle isn't strong enough justification to prevent this going ahead.

We should rather:
- automate polyfill bundle size comparison
- review our polyfill bundle regularly
- have frontend read DCR's config, or find a way to keep them aligned to leverage client caching

mxdvl (approved):

PR #4097, opened 2022-02-25
`BlockByline`

oliverlloyd:
## What does this change?
Adds the `BlockByline` component 

## Why?
These were missing from the data model before but now that [this PR](https://github.com/guardian/frontend/pull/24707) is merged we're ready to add them


MarSavar:
🎉  This is great!
I wonder why the font for the contributor name doesn't look correct in the screenshot 🤔

oliverlloyd:
> I wonder why the font for the contributor name doesn't look correct in the screenshot 🤔

Probably because it was wrong! I've changed it to use `body` now

MarSavar (approved):

Issue #4080, opened 2022-02-24
GLabs sponsor logo tracking in GA not working


DavidLawes:
Reported by Jack Podmore via the dotcom email:

The commercial data team were asked earlier about the clicks on sponsor logos on our Ocado-supported content. I've looked into it and I think the issue may be that the GA sponsor logo tracking doesn't fire on DCR. I can't see any events being sent to GA when the logo is clicked on article pages.

E.g. On [this article](https://www.theguardian.com/food/2022/feb/23/how-to-make-chicken-cacciatore-recipe-felicity-cloake-masterclass), nothing registers in dev tools when clicking the sponsor logo. However, everything works fine on [this series page:](https://www.theguardian.com/food/series/theguardianfeast)

<img width="483" alt="image (1)" src="https://user-images.githubusercontent.com/45561419/155490979-5f623d3c-9b90-48c1-b9f7-77b67c7ca5be.png">

Everything also seems to be fine on this [sponsored podcast page](https://www.theguardian.com/lifeandstyle/audio/2022/jan/11/s2-e12-guy-garvey-musician).

The article pages on this [series](https://www.theguardian.com/sap-solutions) and this [series](https://www.theguardian.com/rewild-your-life) both show the same behaviour, so I don't think it's specific to any one campaign.

Looking at the long term trend in GA suggests we've gradually been losing sponsor clicks over the past two years:

<img width="538" alt="image" src="https://user-images.githubusercontent.com/45561419/155490895-347c97eb-7f1d-4bce-812d-cac39b063d37.png">

With them fully vanishing from articles around the 25th of June 2021 - when [this pull requests](https://github.com/guardian/frontend/pull/23944) switches paid content over to DCR. 

While this appears to be a long-standing problem and has only recently been noticed, GLabs report this number to clients and it's kind of important that we have an accurate figure.

PR #4098, opened 2022-02-25
Fix GLabs sponsor logo tracking

ioannakok:
## What does this change?
* Turns `Branding` component to an island.
* Adds `trackSponsorLogoLinkClick` method to `ga.ts`. This sends sponsor data to Google Analytics.  
* Passes an `trackSponsorLogoLinkClick` as an `onClick` method to `Branding` island and `CardBranding` in `OnwardsUpper` section. When  user clicks on a sponsor logo this info will be sent to Google Analytics.
* Adds a Cypress `paid.content.spec.ts` test suite.

## Why?
It has been reported by GLabs that GA sponsor logo tracking doesn't fire on DCR. This functionality existed in `frontend` and got lost when paid content got switched over to DCR. This PR adds it back because GLabs need to report the number of clicks to clients.

See more details on the issue: https://github.com/guardian/dotcom-rendering/issues/4080

### After
Screenshot 2022-03-01 at 18 28 38

### Before
The above info was not included in the payload.


.gitignore
```diff
@@ -22,6 +22,7 @@ node_modules
 !.vscode/settings.json
 !.vscode/dcr.code-snippets
 .metals/
+*.iml
```
ioannakok, line 25: Unrelated to the issue

dotcom-rendering/cypress/integration/parallel-6/paid.content.spec.js
```diff
@@ -0,0 +1,65 @@
...
+
+// It is important to use this article for this test because its commercialProperties
+// coming from the CAPI object are the same for all editions. This way we are making sure
+// the Branding island the test expects will be in the DOM. If an article with different
+// commercialProperties across editions were to be used, the test would have different
+// outcome when running locally (UK edition) and when running in CI. For example,
+// it could be US edition if the CI server runs in US. The best way to deal with
+// this would be to control the edition in the test whether by setting the GU_EDITION
+// cookie or by selecting the edition in the UI. Unfortunately, this is not possible
+// at this point of the migration.
```
ioannakok, line 15: I hope this comment makes sense!

dotcom-rendering/cypress/integration/parallel-6/paid.content.spec.js
```diff
@@ -0,0 +1,65 @@
...
+			'data-gu-ready',
+			'true',
+		);
+
+		cy.get('[data-cy=branding-logo]').should('be.visible');
+		cy.get('[data-cy=branding-logo]').click();
+
+		// Make sure the call to Google Analytics contains the info we want
+		cy.wait('@gaRequest').then((interception) => {
+			let requestURL = interception.request.url;
```
ioannakok, line 59: We tried to access this info by doing `interception.request.body` instead of accessing the url but the body is empty.

dotcom-rendering/src/web/browser/ga/ga.ts
```diff
@@ -193,6 +193,22 @@ export const trackNonClickInteraction = (actionName: string): void => {
 	}
 };
 
+export const trackSponsorLogoLinkClick = (sponsorName: string): void => {
+	const { ga } = window;
+
+	if (ga) {
+		ga(send, 'event', 'click', 'sponsor logo', sponsorName, {
```
DavidLawes, line 200: To add some extra visibility, here we're recreating the logic that already exists in frontend (https://github.com/guardian/frontend/blob/main/static/src/javascripts/projects/common/modules/analytics/google.ts#L45). To us, It looked like this logic had never been migrated over to DCR which was why the expected events weren't being sent.

.gitignore
jlieb10: +1

dotcom-rendering/cypress/integration/parallel-6/paid.content.spec.js
jlieb10: Just wondering why we can't use the `GU_EDITION` cookie at this point of the migration

dotcom-rendering/src/web/components/Branding.importable.tsx
```diff
@@ -47,6 +47,12 @@ export const Branding: React.FC<{
 					data-sponsor={branding.sponsorName.toLowerCase()}
 					rel="nofollow"
 					aria-label={`Visit the ${branding.sponsorName} website`}
+					onClick={() =>
+						trackSponsorLogoLinkClick(
+							branding.sponsorName.toLocaleLowerCase(),
```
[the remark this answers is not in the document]
jlieb10: Yeah interesting - never seen `toLocaleLowerCase` before

jlieb10 (approved):
Nice work!

dotcom-rendering/cypress/integration/parallel-6/paid.content.spec.js
ioannakok: I was hoping the part of the sentence where this argument becomes illogical would go unnoticed 🙈  but yes you are right 🙂  We tried setting the cookie in the test but without success and this has nothing to do with the current point of the migration. Happy to give it another go though after we merge this!

dotcom-rendering/src/web/components/Branding.importable.tsx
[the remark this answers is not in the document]
ioannakok: Good catch! Changed it

dotcom-rendering/cypress/integration/parallel-6/paid.content.spec.js
ioannakok: Updated the comment to make it clear!

PR #4099, opened 2022-02-28
Observe `topOfBlog` when not on first page

oliverlloyd:
## What does this change?
Previously, we only set the observer for the topOfBlog when on the first page. In this PR we allow some features of this observer on all pages

## Why?
Because we had a bug where the toast was being removed when it should have been kept on the page but also because this allows us to use this logic in other situations, such as if we wanted to stick the toast and needed a way to dynamically set some state based on scroll position (this is a pending feature)

### DIFF
--- a/dotcom-rendering/src/web/components/Liveness.importable.tsx
+++ b/dotcom-rendering/src/web/components/Liveness.importable.tsx
@@ -121,16 +121,6 @@ const lastUpdated: Element | null = !isServer
 	? window.document.querySelector('[data-gu-marker=liveblog-last-updated]')
 	: null;
 
-/**
- * This allows us to make decisions in javascript based on if the reader
- * has the top of the blog in view or not
- *
- * @returns boolean
- */
-function topOfBlogVisible(): boolean {
-	return topOfBlog ? topOfBlog.classList.contains('in-viewport') : false;
-}
-
 export const Liveness = ({
 	pageId,
 	webTitle,
@@ -143,6 +133,9 @@ export const Liveness = ({
 	mostRecentBlockId,
 }: Props) => {
 	const [showToast, setShowToast] = useState(false);
+	const [topOfBlogVisible, setTopOfBlogVisible] = useState<
+		boolean | undefined
+	>();
 	const [numHiddenBlocks, setNumHiddenBlocks] = useState(0);
 	const [latestBlockId, setLatestBlockId] = useState(mostRecentBlockId);
 
@@ -161,7 +154,7 @@ export const Liveness = ({
 				updateTimeElement(lastUpdated);
 			}
 
-			if (onFirstPage && topOfBlogVisible() && document.hasFocus()) {
+			if (onFirstPage && topOfBlogVisible && document.hasFocus()) {
 				revealPendingBlocks();
 				setNumHiddenBlocks(0);
 			} else {
@@ -207,9 +200,9 @@ export const Liveness = ({
 				const topOfBlogShowing = entry.isIntersecting;
 
 				if (topOfBlogShowing) {
-					entry.target.classList.add('in-viewport');
+					setTopOfBlogVisible(true);
 				} else {
-					entry.target.classList.remove('in-viewport');
+					setTopOfBlogVisible(false);
 				}
 
 				if (topOfBlogShowing && onFirstPage) {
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -199,28 +199,35 @@ export const Liveness = ({
...
-				entry.target.classList.remove('in-viewport');
-			},
-			{
-				root: null,
-				// A margin makes more sense because the element we're
-				// observing (topOfBlog) has no height
-				rootMargin: '100px',
-			},
-		);
+	useEffect(() => {
```
oliverlloyd, line 202: I added a useEffect here as an optimisation to prevent the observer being set multiple times

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -199,28 +199,35 @@ export const Liveness = ({
...
+						entry.target.classList.add('in-viewport');
+					} else {
+						entry.target.classList.remove('in-viewport');
+					}
+
+					if (topOfBlogShowing && onFirstPage) {
+						revealPendingBlocks();
+						setNumHiddenBlocks(0);
+						setShowToast(false);
+					}
```
oliverlloyd, line 218: Previously, this always ran, now it only runs if we're on the first page. This is the change in this PR

dotcom-rendering/src/web/components/Liveness.importable.tsx
mxdvl: Should we make sure that we have a `return` statement that disconnects it, as per hooks specs?

Should we leverage `useOnce`, if we truly want this to only run once? It feels like currently this would run each time the value of `onFirstPage` changes.

mxdvl (commented):
Trying to follow this code, it feels like half is declarative and half imperative. I can approve this PR as it’s not really changing the inner workings, but this feels ripe for a refactor

The fact that `topOfBlogVisible` is not set via useState puzzles me. Should we consider having [a useIntersectionObserver hook](https://usehooks-ts.com/react-hook/use-intersection-observer)?

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: > Should we make sure that we have a `return` statement that disconnects it, as per hooks specs?

Agreed. Great spot. I've added this.

> 
> Should we leverage `useOnce`, if we truly want this to only run once? It feels like currently this would run each time the value of `onFirstPage` changes.

Rerunning if/when `onFirstPage` changes is the intended behaviour. In reality this value is static but the component might rerender for other reasons (these could change in future) and so keeping the logic makes sense. I think `useOnce` is best for when you _really_ only want code to run once, like incrementing a value.

oliverlloyd:
> Trying to follow this code, it feels like half is declarative and half imperative. I can approve this PR as it’s not really changing the inner workings, but this feels ripe for a refactor
> 

The goal with this `Liveness` file is exactly this, to be something half declarative and half imperative. I wrote an [RFC explaining the reasoning](https://github.com/guardian/dotcom-rendering/discussions/3870) behind this approach and also talked about it in [the original PR.](https://github.com/guardian/dotcom-rendering/pull/3869) but in short we are doing exactly what you said, using React's declarative structure for rendering the toast but Javascript imperative approach for manipulating the block list.

### Why don't we just use javascript for everything?
So technically we could change how we render the toast to use a more javascript based approach, adding show hide classes. That would perhaps remove some of the cognitive disconnection but it would also mean we have a hard to maintain coupling between different parts of the codebase.  Right now, `Toast` is 'managed' by `Liveness` using local state but if we show hide it using js we will need to insert that html another way, and also then manage changing the number that gets displayed, and if it is stuck or not. At this point you start to understand why React was created!

### Why don't we use react for everything?
Performance mostly, but also because animating list edits is hard in react

mxdvl:
> in short we are doing exactly what you said, using React's declarative structure for rendering the toast but Javascript imperative approach for manipulating the block list.

I get that distinction, but I’m really puzzled by `topOfBlogVisible`. It’s only used internally by `Liveness`, so why not make it a state? The whole `in-viewport` class seems like an unnecessary indirection. Happy to take it to another PR, though.

oliverlloyd:
> I get that distinction, but I’m really puzzled by `topOfBlogVisible`. It’s only used internally by `Liveness`, so why not make it a state? The whole `in-viewport` class seems like an unnecessary indirection. Happy to take it to another PR, though.

Interesting. So I think you're right, this can live in react state just as well. I wonder if there were maybe legacy reasons this was done this way?

In any case, I've made the change and the code is a lot simpler, thanks for spotting this!

mxdvl (approved):
Happy except for:

> React Hook `useEffect` has a missing dependency: 'topOfBlogVisible'. Either include it or remove the dependency array

PR #4100, opened 2022-02-28
Sticky toast

oliverlloyd:
## What does this change?
This stops the `Toast` being sticky above the top of the blog

## Why?
Because we don't want to have the toast overlaying other content such as the page header

## `#maincontent` is now `#liveblog-body`
I've moved `maincontent` up to `LiveLayout.tsx` where it is now wrapping pagination and key events. I then replaced the old `maincontent` marker with `#liveblog-body

This is useful because now when we scroll to `maincontent` we continue to see navigation elements on the page but we can also use `#liveblog-body` to insert html.


dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -216,7 +220,8 @@ const sidePaddingDesktop = css`
 	}
 `;
 
-const accordionBottomMargin = css`
+const bodyWrapper = css`
```
oliverlloyd, line 223: I slightly refactored this css to make it easier to adapt and to make the names a bit more truthful

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -216,7 +220,8 @@ const sidePaddingDesktop = css`
...
+	position: relative;
```
oliverlloyd, line 224: We use this to contain the absolute toast

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -703,7 +690,35 @@ export const LiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
 								)}
 							</GridItem>
 							<GridItem area="body">
-								<div css={accordionBottomMargin}>
+								<div css={bodyWrapper}>
+									<span data-gu-marker="top-of-blog" />
+									{format.design ===
```
oliverlloyd, line 695: Now the `Toast` and the marker are sat next to each other

AshCorr (approved):
LGTM!

Just a thought, would it be a good idea to show a different icon like a refresh spinner when the toast is at the top of the page? I find it looks a bit odd to have an up arrow when you're already at the top of the liveblog, just my opinion though!

oliverlloyd:
> LGTM!
> 
> Just a thought, would it be a good idea to show a different icon like a refresh spinner when the toast is at the top of the page? I find it looks a bit odd to have an up arrow when you're already at the top of the liveblog, just my opinion though!

You're right, the arrow looks odd. It's actually a stopgap while we wait for the refresh svg to land. I think @mxdvl had this in his sights?

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -195,23 +196,26 @@ export const Liveness = ({
...
-				// A margin makes more sense because the element we're
-				// observing (topOfBlog) has no height
-				rootMargin: '100px',
-			},
-		);
+			} else {
+				setTopOfBlogVisible(false);
+				// If we're scrolling down make the toast sticky
+				if (belowTopOfBlog) {
+					setStickToast(true);
```
AshCorr, line 215: I'm a bit naive on react, but would React not re-render the component every time setStickToast changes? Or is it intelligent enough to realize the value itself hasn't changed?

oliverlloyd: > would React not re-render the component every time setStickToast changes?

This is absolutely what it does. Or at least it will re-render when state _changes_ and it will only rerender the html that needs updating. So in this case, when `stickToast` is changed it updates the css for the element, sticking or unsticking it

dotcom-rendering/src/web/components/Toast.tsx
```diff
@@ -1,27 +1,33 @@
...
-export const Toast = ({ count, onClick, format }: Props) => {
+export const Toast = ({ count, onClick, format, stuck }: Props) => {
 	return (
 		<div
 			css={css`
-				position: fixed;
-				top: 20px;
-				display: flex;
-				${getZIndex('toast')};
+				position: ${stuck ? 'fixed' : 'absolute'};
```
mxdvl, line 21: I’m assuming we’ve tested the css sticky, here?

mxdvl (commented):
Yes, @AshCorr & @oliverlloyd new reload icon here: #4093

dotcom-rendering/src/web/components/Toast.tsx
oliverlloyd: Yeah. But I can't remember why it didn't work... and now when I try it again it works flawlessly.

I think I'm going to just pretend I went with sticky all along and that none of this ever happened.

dotcom-rendering/src/web/components/ArticleBody.tsx
```diff
@@ -104,13 +104,10 @@ export const ArticleBody = ({
...
 		return (
 			<>
-				{format.design === ArticleDesign.LiveBlog && (
-					<span data-gu-marker="top-of-blog" />
-				)}
 				<div
 					// eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
 					tabIndex={0}
-					id="maincontent"
+					id="liveblog-body"
```
alinaboghiu, line 110: Would this affect other things that rely on the "maincontent" id? E.g. accessibility skip-to links?

oliverlloyd: I've moved `#maincontent` to higher up (it now sits just above key events) so things like skip to links and general page navigation will still work. 

There's a change here though, previously we navigated users tight to the top of the blocks but now they arrive further up, and can see additional items.

### Before
Screenshot 2022-03-01 at 11 25 17


### After
Screenshot 2022-03-01 at 11 25 03


What do you think? In my opinion, including this extra content at the top of the page when readers navigate is an improvement but it's perhaps not _semantically_ the actual content.

alinaboghiu: tl;dr I think this is better

I always found it confusing when I got scrolled to the top of the first post because I wasn't sure if I was at the top or in the middle, so always ended up scrolling a bit further up to get my bearings. When we designed the filter key events button UX also wanted us to include the button when scrolling to the top (perhaps a better arguments than my subjective world view).

oliverlloyd: > I always found it confusing when I got scrolled to the top of the first post because I wasn't sure if I was at the top or in the middle, so always ended up scrolling a bit further up to get my bearings

Me too!

> When we designed the filter key events button UX also wanted us to include the button when scrolling to the top

Perfect. This is good to hear

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -44,15 +44,15 @@ function insert(html: string, switches: Switches) {
 	// Shouldn't we sanitise this html?
 	// We're being sent this string by our own backend, not reader input, so we
 	// trust that the tags and attributes it contains are safe and intentional
-	const maincontent = document.querySelector<HTMLElement>('#maincontent');
-	const latestBlock = document.querySelector('#maincontent :first-child');
-	if (!latestBlock || !maincontent) return;
-	maincontent.insertBefore(fragment, latestBlock);
+	const blogBody = document.querySelector<HTMLElement>('#liveblog-body');
+	const latestBlock = document.querySelector('#liveblog-body :first-child');
```
mxdvl, line 48: We’re already doing this below!

```suggestion
	const latestBlock = blogBody.querySelector(':first-child');
```

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -204,10 +204,14 @@ const sticky = css`
...
 
-const keyEventsTopMargin = css`
+const keyEventsMargins = css`
 	${from.desktop} {
 		margin-top: ${space[1]}px;
 	}
+	margin-bottom: ${space[3]}px;
+	${from.desktop} {
+		margin-bottom: 0;
+	}
```
mxdvl, line 214: This one slipped in!

```suggestion
	margin-bottom: ${space[3]}px;
	${from.desktop} {
		margin-top: ${space[1]}px;
		margin-bottom: 0;
	}
```

mxdvl (changes requested):

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: This suggestion causes typescript errors so I've made the change locally

mxdvl (approved):

PR #4101, opened 2022-02-28
Use `useCallback`

oliverlloyd:
## What?
This PR adds the use of the `useCallback` hook to `Liveness`

## Why?
This is a small optimisation to prevent the functions being recreated more than they need to

## Why not?
There's an argument to say this change introduces complexity and that this isn't worth the performance win. I'm on the fence here so if anyone has strong opinions I'm happy to defer

### DIFF
--- a/dotcom-rendering/src/web/components/Liveness.importable.tsx
+++ b/dotcom-rendering/src/web/components/Liveness.importable.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { initHydration } from '../browser/islands/initHydration';
 import { useApi } from '../lib/useApi';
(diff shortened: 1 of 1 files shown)


mxdvl (commented):
Why are we defining these functions _inside_ of the component’s scope in the first place?

oliverlloyd:
> Why are we defining these functions _inside_ of the component’s scope in the first place?

This is normal. Most functions used in React components are defined inline. Each time the component re-renders the function gets recreated based on the latest state and props. This is _intentional_.

From reading around some more though, it seems the problems with this approach start when functions are passed as props (because the reference to the function changes and it is seen as a new prop, causing a re-render) but we don't have that issues here so I actually think using useCallback like I am here as an optimisation is [not a good idea](https://kentcdodds.com/blog/usememo-and-usecallback) so I'm going to close this PR.

PR #4105, opened 2022-02-28
Integration tests for live blogs

oliverlloyd:
## What?
This PR adds some Cypress specs for testing interactivity on liveblogs

### Skipped second page test
This test is present but currently skipped pending [another PR](https://github.com/guardian/dotcom-rendering/pull/4099)

### But isn't this blog going to be dead?
We use `?live=true` to force the blog to be 'live' to the level where our client side polling code will execute.


dotcom-rendering/cypress/integration/parallel-3/liveblog.interactivity.spec.js
```diff
@@ -0,0 +1,208 @@
...
+			.first()
+			.should('have.attr', 'data-gu-ready', 'true');
+		cy.scrollTo('center');
+		cy.get(`[data-cy="toast"]`).should('not.exist');
+		cy.window().then(function (win) {
+			win.mockLiveUpdate({
+				numNewBlocks: 1,
+				html: '<p>New block</p>',
+				mostRecentBlockId: 'abc',
+			});
```
mxdvl, line 55: I think this is the right approach, but highlighting here for others to see.

mxdvl (commented):
Is there a reason we do not use fat arrow functions? Is the scope of `this` important in this case?

EDIT: I checked and this in in line with other Cypress tests

dotcom-rendering/cypress/integration/parallel-3/liveblog.interactivity.spec.js
```diff
@@ -0,0 +1,208 @@
...
+
+const ignoreTwitterError = () => {
+	cy.on('uncaught:exception', (err) => {
+		if (err.message.includes('Illegal invocation')) {
+			// Thrown by the twitter lib. This error is unrelated to the test in question
+			// so return false to prevent this error from failing this test
+			return false;
+		}
+	});
+};
```
mxdvl, line 22: Are we sure it's fine to ignore this error?

Could we extract this into a JSDoc style comment to the const?

mxdvl (approved):
More tests: nice!

dotcom-rendering/cypress/integration/parallel-3/liveblog.interactivity.spec.js
oliverlloyd: It's conceivable that there is an issue but I'd rather solve problems that get surfaced on the real site, not inside the strange Cypress world where globals like `twttr` can have issues. Plus we're testing that the function works so I'm happy to ignore. 

I've moved the comment into JSDoc 👍

PR #4106, opened 2022-02-28
feat(cmp): add a dummy for the server bundle

mxdvl:
## What does this change?

Create a dummy CMP that is used in the server bundle via Webpack’s `resolve`/`alias`. 

## Why?

The CMP lib code is not server-safe, as it relies on `window`.

We need a way to make sure this code does not run on the server,
so this aliases to a dummy CMP that does nothing. This approach can be
pushed upstream so the CMP exports this dummy directly.

### DIFF
--- a/dotcom-rendering/src/server/cmp-dummy.ts
+++ b/dotcom-rendering/src/server/cmp-dummy.ts
@@ -1,0 +1,28 @@
+import type { CMP } from '@guardian/consent-management-platform/dist/types';
+
+const warn = (): void => {
+	console.warn(
+		'This is a dummy version of the @guardian/consent-management-platform',
+		'No consent signals will be received.',
+	);
+};
+
+const warnAndReturn = <T extends unknown>(arg: T): (() => T) => {
+	warn();
+	return () => arg;
+};
+
+export const cmp: CMP = {
+	__disable: warn,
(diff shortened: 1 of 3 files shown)


OllysCoding (approved):
+1 Looks good :)

dotcom-rendering/src/server/cmp-dummy.ts
```diff
@@ -0,0 +1,16 @@
...
+	hasInitialised: () => false,
+	init: () => {},
+	showPrivacyManager: () => {},
+	version: 'n/a',
+	willShowPrivacyMessage: () => Promise.resolve(false),
+	willShowPrivacyMessageSync: () => false,
+};
+
+export const onConsentChange = () => {};
+export const getConsentFor = () => {};
```
OllysCoding, line 16: Would it be worth making it so if any of these functions are called, they warn the user about (or throw an error regarding) not using CMP on the server?

You could also do get() methods for properties too but this might be overkill

mxdvl: Excellent idea. I was wondering if we liked this approach before committing more time to it.

mxdvl:
Whoops, I meant to close and keep as a reference.

PR #4108, opened 2022-02-28
fix initPerf to clear all marked keys and measures when clear is called

marjisound:
## What does this change?

This PR updates the `clear` function in `initPerf`, so that when clear is called, all the marked keys and the measures that was taken as part of the current `initPerf` instance will be cleared from the browser performance entry. 

This way if the consumer wants to measure time multiple times with the same name, they can decide if they want to clear the marked keys and measurements.

This change was also made in this https://github.com/guardian/frontend/pull/24710 in frontend.


mxdvl (approved):
Yes, saw the frontend PR and it makes sense!

mxdvl:
I think you have to rebase or merge `main`. Specifically commit a462f9510d455faec40bc13cdf8672381cf8a7f8.

marjisound:
> I think you have to rebase or merge `main`. Specifically commit [a462f95](https://github.com/guardian/dotcom-rendering/commit/a462f9510d455faec40bc13cdf8672381cf8a7f8).

Thanks @mxdvl for looking. There are 2 UI tests changes that don't related to the change I did. Do you think it's safe to accept them?

Issue #4072, opened 2022-02-23
Small logo adjustment in cards


mxdvl:
**Reported by:** Harry Fischer @HarryFischer 
**On:** Feb 21 or 22
**Via:** Direct chat

The logo should be the same colour as the timestamp. Follow-up on #4046

<img width="142" alt="image" src="https://user-images.githubusercontent.com/76776/155358626-7bd50bc5-a3bd-4e61-a931-bc4966dd2170.png">

mxdvl:
Can be seen on https://www.theguardian.com/news/2022/feb/22/credit-suisse-leak-three-largest-parties-call-for-eu-to-assess-switzerland-risk

PR #4109, opened 2022-02-28
Remove palette.fill.cardIcon in favour of explicit colours

mxdvl:
## What does this change?

Remove `palette.fill.cardIcon`, as we always want to use the current text colour. ~No visual changes.~ The text and icons are made darker to meet WCAG 2.1 standard AA.

Small side notes:
- keep inferring `palette` from `format`, as started in #4053.
- ~use icons from Source in `MediaMeta`.~ moved to own PR: #4119 

## Why?

Smaller bundle, easier to reason about, less decisions required in palette, which can lead to inconsistencies like in #4072

### DIFF
--- a/dotcom-rendering/src/web/lib/decidePalette.ts
+++ b/dotcom-rendering/src/web/lib/decidePalette.ts
@@ -404,7 +404,7 @@ const textCardFooter = (format: ArticleFormat): string => {
 					// https://theguardian.design/2a1e5182b/p/492a30-light-palette
 					return '#ff9941';
 				default:
-					return neutral[60];
+					return neutral[46];
 			}
 		case ArticleDesign.LiveBlog:
 			switch (format.theme) {
--- a/dotcom-rendering/src/web/lib/decidePalette.ts
+++ b/dotcom-rendering/src/web/lib/decidePalette.ts
@@ -676,71 +676,6 @@ const fillCaptionCamera = (format: ArticleFormat): string =>
 const fillBlockquoteIcon = (format: ArticleFormat): string =>
 	pillarPalette[format.theme].main;
 
-const fillCardIcon = (format: ArticleFormat): string => {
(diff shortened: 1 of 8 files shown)


JamieB-gu (commented):
👍 for the `palette -> format` refactor.

Regarding `currentColor`: do you have any concerns that this might make these icons/components harder to reason about in isolation because they're relying on context outside of their control?

mxdvl (commented):
In two of the three instances, the colour is defined in the same file.

Do we have an explicit colour defined every time we use text? How is an icon different?

mxdvl:
Following your input I’ve opted for an explicit approach in 9ef033e2c73c0bd8557169152982af0aa7601457

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -446,7 +446,7 @@ const textCardFooter = (format: ArticleFormat): string => {
 				case ArticleSpecial.SpecialReport:
 					return brandAltBackground.primary;
 				default:
-					return neutral[60];
+					return neutral[46];
```
mxdvl, line 449: This ensures we meet WCAG 2.1 standards AA and AAA for contrast.

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -404,7 +404,7 @@ const textCardFooter = (format: ArticleFormat): string => {
 					// https://theguardian.design/2a1e5182b/p/492a30-light-palette
 					return '#ff9941';
 				default:
-					return neutral[60];
+					return neutral[46];
```
mxdvl, line 407: This ensures we meet WCAG 2.1 standard AA.

mxdvl (commented):
Our beloved designer has spotted many WCAG standard AA violations which are now addressed by making the footer text darker.

ioannakok (approved):
Looks good to me!

PR #4110, opened 2022-02-28
AB testing by content on AMP

chrislomaxjones:
## What does this change?

This  PR adds a simple way to apply different commercial features to different articles on AMP.

Content is bucketed based on taking a fixed prefix from the `SHA256` hash of its `pageId`, which is unique for each article.

Each of these hashes is then assigned to one of a fixed twelve groups: [`0`, `1`, ..., `10`, `11`]. It is currently left as the responsibility of the consumer of this functionality to ensure they construct their groups appropriately and assign their controls and variants to non-overlapping groups (this will be addressed in a subsequent PR).

We employ a similar API to the web [AB testing framework](https://github.com/guardian/ab-testing). A provider component is used with a React context so that components that wish to condition on a test group don't need to have the `shortUrlId` propped drilled to them, and instead can use the `useTestGroup` hook. This has the benefit of allowing us to separate how we bucket content (the point at which we render the `TestProvider`) and how we enable features based on test group (using the `useTestGroup` hook).

We wrap this functionality in a switch which can be used to enable content based testing.

### Example usage

```tsx
const Foo = () => {
  	const { group } = useTestGroup();

	switch (group) {
		case 0: {
			return <p>Group 0</p>;
		}
		case 1: {
			return <p>Group 1</p>;
		}
                // ...
		case 11: {
			return <p>Group 11</p>;
		}
		default: {
			return null;
		}
	}
};
```

Note also that we don't actually record the mapping from article to test group from DCR. This will be done separately at the analysis stage, by deriving the test group for each AMP impression logged in the data lake.

## Why?

It is useful to measure the impact changes to the commercial code have on revenue. On the web we typically do this by running an AB or multivariate test, with percentages of the audience having different commercial features enabled and the revenue uplift measured against a control.

Because of the limitations of AMP, we are unable to run a AB test using our usual framework on the web, as we are unable to vary the experience by partitioning the audience.

An alternative way to achieve something _similar_ is to uniformly bucket articles into different groups, and apply the commercial features to these groups. Whilst this won't result in uniformly assigning pageviews to groups (since pageviews aren't uniformly distributed across articles), we should be able to remove any outlier pages at the analysis stage. This should get us enough of the way to be able to compare the impact of different commercial features.

## Next steps

This PR doesn't provide any support for defining a test and assigning variants/controls to sets of groups, and the ability to switch tests off/on. This will addressed in a subsequent PR.

This PR also doesn't actually implement a test. The first of these tests will concern splitting articles into three groups and using a different Prebid server for adverts on each.


dotcom-rendering/src/amp/components/Footer.tsx
```diff
@@ -183,24 +184,27 @@ const FooterLinks: React.FC<{
...
 
-export const Footer: React.FC<{ nav: NavType }> = ({ nav }) => (
-	<footer css={footer}>
-		<div css={innerContainerStyles}>
-			<div css={footerInner}>
-				<FooterLinks links={footerLinksNew} nav={nav} />
+export const Footer: React.FC<{ nav: NavType }> = ({ nav }) => {
+	const { group } = useContentABTestGroup();
+	return (
+		<footer data-content-test-group={group} css={footer}>
```
chrislomaxjones, line 190: Here we include a `data-content-test-group` attribute so we can inspect a given AMP article to see the assigned group.

chrislomaxjones:
> Would a [MurmurHash](https://en.wikipedia.org/wiki/MurmurHash) be more appropriate than a [Secure Hash Algorithm](https://en.wikipedia.org/wiki/SHA-2), here? It's designed to be well distributed. I was introduced to it via the work we did on `browserId`.

This is interesting, I hadn't heard fo MurmurHash. I wouldn't rule out at a change, and for the moment we're not set on exactly how the grouping will be achieved.

Our main limitation at the moment is that the equivalent grouping function needs to be available in BigQuery. This is why we've gone with a common hashing algorithm like SHA-256.

dotcom-rendering/src/amp/components/ContentABTest.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+		<ContentABTestProvider
+			switches={{ ampContentABTesting: true }}
+			pageId={pageId}
+		>
+			<Component />
+		</ContentABTestProvider>,
+	);
+
+	// Check content has been placed into one of the possible groups
+	expect(container).toHaveTextContent(/0|1|2|3|4|5|6|7|8|9|10|11/);
```
chrislomaxjones, line 62: Once we're settled on the exact grouping function these tests can be enhanced by checking that a known `pageId` maps to the correct group.

zekehuntergreen (approved):
Sounds like we're still confirming how to hash page id but the code looks good 👏

dotcom-rendering/src/amp/components/ContentABTest.tsx
```diff
@@ -0,0 +1,94 @@
...
+	n === 3 ||
+	n === 4 ||
+	n === 5 ||
+	n === 6 ||
+	n === 7 ||
+	n === 8 ||
+	n === 9 ||
+	n === 10 ||
+	n === 11;
+
```
ashishpuliyel, line 25: At the risk of bikeshedding could we simplify the above to:

```typescript
const AB_TEST_GROUPS = [0,1,2,3,4,5,6,7,8,9,10,11] as const;
type ContentABTestGroup = typeof AB_TEST_GROUPS[number];
const NUM_GROUPS = AB_TEST_GROUPS.length;

const isContentABTestGroup = (n: ContentABTestGroup): n is ContentABTestGroup =>  AB_TEST_GROUPS.includes(n);
```

ashishpuliyel: And looking at this isContentABTestGroup looks like it's doing nothing. But the generated js does do the check

ashishpuliyel (approved):
Looks pretty good @chrislomaxjones

dotcom-rendering/src/amp/components/ContentABTest.tsx
chrislomaxjones: Think that makes a lot of sense.

Only thing is we'd probably want

```tsx
const isContentABTestGroup = (n: number): n is ContentABTestGroup => { };
```
since it needs to be able to narrow any `number` to `ContentABTestGroup`. So we'd maybe end up with something like

```tsx
const isContentABTestGroup = (n: number): n is ContentABTestGroup =>
    AB_TEST_GROUPS.includes(n as ContentABTestGroup);
```

do you think that's preferable?

dotcom-rendering/package.json
```diff
@@ -127,6 +128,7 @@
     "compose-function": "^3.0.3",
     "compression": "^1.7.3",
     "cpy": "^8.1.2",
+    "crypto-js": "^4.1.1",
```
arelra, line 131: This package is pretty big and doesn't tree shake according to:
https://bundlephobia.com/package/crypto-js@4.1.1

Or are we less concerned as AMP will SSR and cache - which is fair!

There's also a built-in way to do SHA256 but it doesn't support IE11 - not sure if that is an issue.
https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#basic_example

dotcom-rendering/src/amp/components/ContentABTest.tsx
ashishpuliyel: Fine by me. I _think_ it shakes out the same in the end.

At runtime it's not really narrowing a _number_ to a group id. It's checking if the variable is a valid groupid. Both in my version and this one.

In your version when you're writing it typescript can tell you that only numbers are acceptable to this function. In my version when you're writing it typescript can tell you that only 0-11 are acceptable

In both cases we'll have to tell typescript that the output of hash % 12 is an integer 0-11, and it's a valid ContentABTestGroup (It'll just know it's a number).

ashishpuliyel: They come out (identical)[https://www.typescriptlang.org/play?#code/MYewdgzgLgBAggIQPoBUCiBlFSDiAlAeQFUAFDGAXhgG0AGAGgEZ6AmegZnoBZ6BWegGz0A7PQAc9AJxMGjRgF0YAQwgxQkKAG4AUFACeABwCmMAMLgoRsFEQoj0HACcQAVwOUY+4yABm8ZOhYuISkGNRgLgC2AEZGjvI66tAwAHJEALLBxGQeiKiY2PjZGAB0ADZWAOZQABY62kmwAJYQ5tZWNgh2Ds5uHgAUYABcZhYdtvZQTq4GAJQjYDAto+3WEz0zlAB8MP75QUWhJU1gwGUuACb2g7OJ4MktbZZrXZPTbiwDwzARMXHzPyWqie41eGz6FC22l2uzygUKITIx1O5yuEEGymBYxe3SmvTm9SAA]

dotcom-rendering/package.json
mxdvl: This is all server-side rendered anyways. No JS is passed to the client in AMP.

dotcom-rendering/src/amp/components/ContentABTest.tsx
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
+import sha256 from 'crypto-js/sha256';
+
+const AB_TEST_GROUPS = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11] as const;
```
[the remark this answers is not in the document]
chrislomaxjones: That's a good question, I'm not actually sure what the ad manager groups are used for (partitioning requests to pubads by splitting into groups?), and they're definitely a separate usage to this one.

The reason I've gone zero indexed here is because that matches the output of `(x) => x % NUM_GROUPS`. Since the value of these groups is arbitrary (they could equally be `A`, `B`, `C`, ...), keeping them the same as the output of the grouping function should keep it as simple as possible (both here and from the BigQuery analysis side).

arelra (approved):

dotcom-rendering/package.json
mxdvl: Actuall, [according to the Node 14 LTS docs](https://nodejs.org/docs/latest-v14.x/api/crypto.html#crypto_hash_digest_encoding), `crypto` should cover this use case.

mxdvl: Removed in #10027

PR #4112, opened 2022-02-28
Migrate ReaderRevenueLinks to Islands

AshCorr:


dotcom-rendering/src/web/components/ReaderRevenueLinks.importable.tsx
```diff
@@ -207,6 +204,7 @@ const ReaderRevenueLinksRemote: React.FC<{
 				const { module } = response.data;
 				setSupportHeaderResponse(module);
 
+				setAutomat();
```
AshCorr, line 207: There seems to be some kind of race condition happening here which causes rendering the RR banners to ocasionally fail. It seems like for some reason when setAutomat() is called in the App.tsx context it breaks rendering in the Island context, not too sure why.

Seems like the SWR PR is having a similar issue.

PR #4113, opened 2022-03-01
More information for `spacefinder`

oliverlloyd:
## What does this change?
This PR adds two two attributes for spacefinder to use as it needs

`role: RoleType` - Also know as `weighting`, this is how the element is positioned
`type: CAPIElement._type` - This is the type of element, richlink, image, embed, etc.

## Why?
Because we want to be more declarative, giving situational information to spacefinder to help it best make decisions. By providing the same values that DCR itself uses to make decisions we are also aligning contracts between the platforms (rather than establishing new ones).

### Should we share the types for these things?
If we hoist the type definitions for `role` and `_type` to, say, libs then we can control changes, making sure that the different platforms are not left behind. The con to this is the extra work involved but I think this is justified given these couplings between the codebases are real and need respecting.

### `isNumberedListTitle`
This prop [was added](https://github.com/guardian/dotcom-rendering/pull/2976) to fix a bug where ads were exploding the elements on the page but a [later change](https://github.com/guardian/dotcom-rendering/commit/4b422040e81dd080454175cd5983f97ad0bd8a25) removed the code that implemented it so it is now unused. I've made a note about this and I think if we complete the refactor on the commercial side of things using the `type` value in place of `component` then we can remove this prop entirely.

### DIFF
--- a/dotcom-rendering/cypress/integration/parallel-3/liveblog.interactivity.spec.js
+++ b/dotcom-rendering/cypress/integration/parallel-3/liveblog.interactivity.spec.js
@@ -145,7 +145,7 @@ describe('Liveblogs', function () {
 			},
 			(req) => {
 				expect(req.query.lastUpdate).to.equal(
-					'62148e2d8f081f9e465d1bb7',
+					'block-62148e2d8f081f9e465d1bb7',
 				);
 			},
 		).as('updateCall');
--- a/dotcom-rendering/src/web/components/Figure.tsx
+++ b/dotcom-rendering/src/web/components/Figure.tsx
@@ -178,6 +178,9 @@ export const Figure = ({
 		);
 	}
 
+	// TODO: isNumberedListTitle is not used
+	// TODO: Any usage of data-spacefinder-component can be replaced with data-spacefinder-type
+	//       Once this has been done we can remove this attribute
 	// See 001-commercial-selectors.md for details on `data-spacefinder-component`
 	let spacefinderComponent: 'rich-link' | 'numbered-list-title' | undefined;
 	if (role === 'richLink') {
@@ -191,9 +194,9 @@ export const Figure = ({
 			id={id}
 			css={defaultRoleStyles(role)}
 			data-spacefinder-component={spacefinderComponent}
-			className={className}
 			data-spacefinder-role={role}
 			data-spacefinder-type={type}
+			className={className}
 		>
 			{children}
 		</figure>
--- a/dotcom-rendering/src/web/components/Figure.tsx
+++ b/dotcom-rendering/src/web/components/Figure.tsx
@@ -190,6 +190,7 @@ export const Figure = ({
 			css={defaultRoleStyles(role)}
 			data-spacefinder-component={spacefinderComponent}
 			className={className}
+			data-spacefinder-role={role}
 		>
 			{children}
 		</figure>
--- a/dotcom-rendering/src/web/components/Figure.tsx
+++ b/dotcom-rendering/src/web/components/Figure.tsx
@@ -149,6 +149,7 @@ type Props = {
 	id?: string;
 	isNumberedListTitle?: boolean;
 	className?: string;
+	type?: CAPIElement['_type'];
 };
 
 const mainMediaFigureStyles = css`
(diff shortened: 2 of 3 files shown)


dotcom-rendering/cypress/integration/parallel-3/liveblog.interactivity.spec.js
```diff
@@ -145,7 +145,7 @@ describe('Liveblogs', function () {
 			},
 			(req) => {
 				expect(req.query.lastUpdate).to.equal(
-					'62148e2d8f081f9e465d1bb7',
+					'block-62148e2d8f081f9e465d1bb7',
```
jamesgorrie, line 148: Why did this need updating?

dotcom-rendering/src/web/components/Figure.tsx
```diff
@@ -189,6 +194,8 @@ export const Figure = ({
 			id={id}
 			css={defaultRoleStyles(role)}
 			data-spacefinder-component={spacefinderComponent}
+			data-spacefinder-role={role}
+			data-spacefinder-type={type}
```
jamesgorrie, line 198: Having a test of "a figure has `data-spacefinder` attributes from CAPI" would be useful in not regressing in a place where we have dependencies so monitoring errors is hard - and also start to define the contract.

jamesgorrie: Saying this - is the reason for having both so spacefinder can make more specific matches?

jamesgorrie (approved):

dotcom-rendering/cypress/integration/parallel-3/liveblog.interactivity.spec.js
oliverlloyd: A change on frontend. Previously frontend sent the id without the block prefix but now it sends it with the prefix so any branches with the old expectation need updating.

dotcom-rendering/src/web/components/Figure.tsx
oliverlloyd: These attributes are 100% for spacefinder, yeah

zekehuntergreen: I have the same question - if spacefinder knows an element has `data-spacefinder-role="immersive"` is this enough information or do we need to be more specific?

zekehuntergreen: Sorry, it wasn't very clear! My question is why do we need `data-spacefinder-type` at all? It seems like `data-spacefinder-role` alone is sufficient for the spacefinder rule we need.

jamesgorrie: What @zekehuntergreen said.

oliverlloyd: Gotcha, thanks for clarifying!

So in my head I see `data-spacefinder-type` as a replacement for `data-spacefinder-component`. This latter attribute is - it seems - being used to act differently when the type of element is either a rich link or a numbered list title. Both of these two things are also 'type's so by using 'type' we replace a sort of custom string interface with the actual structured type value of an element.

Note. This type value is only a Frontend/DCR thing and it's not well document or exposed but I think it should have some more limelight

zekehuntergreen: Ok got it - thanks for explaining that!

jamesgorrie: Going to broken record this, but having these use cases encapsulated in a test would be great to explain this to our future selves. Maybe something for the commercial team?

oliverlloyd: All that is to say, I'd rather not block this PR with unit tests and would prefer to have the work to define the commercial contract done separately

jamesgorrie: > not block this PR

:shipit: 

>  If the calling code provides a valid prop then it is set as an attribute

This is the test I think makes sense as it is simple and should always pass and validates the `CAPI => DCR => Commercial` pipeline. 

> if we consider an integration test for this

Might be worth widening the discussion to see if we have other examples where this makes sense?

oliverlloyd: > Might be worth widening the discussion to see if we have other examples where this makes sense?

👍 I'm going to add a ticket to review these areas and create the tests

PR #4114, opened 2022-03-01
refactor(cmp): revert dummy for the server bundle

mxdvl:
## What does this change?

This reverts commit 1b8bfb748c8ba1500320a1f154a77cc75c153e9a.

## Why?

The CMP is server-safe as of https://github.com/guardian/consent-management-platform/pull/555
**EDIT:** issues with automatic release means we can’t consume this yet. This will only be merged once CMP is released.


oliverlloyd (approved):

mxdvl:
Will only merge once the library has actually been released.

OllysCoding:
Updated with main & tested locally ~ You are still able to import CMP on the server without errors as we'd expect from #4145

PR #3957, opened 2022-02-10
improve the DCR dev server

sndrs:
Co-Authored-By: Olly <9575458+OllysCoding@users.noreply.github.com>

PR #4115, opened 2022-03-01
Use `getContentFromURLMiddleware` everywhere

oliverlloyd:
## What?
Replaces the similar but different logic that was being used on the `GET /Article` route for the prod build with the new `getContentFromURLMiddleware` middleware that [was added to](https://github.com/guardian/dotcom-rendering/pull/3957) the same route for the dev build.

## Why?
Because [we improved](https://github.com/guardian/dotcom-rendering/pull/4037) `getContentFromURLMiddleware` to handle query params better and we want that same improved goodness when we run `make run-ci`

### DIFF
--- a/dotcom-rendering/src/server/prod-server.ts
+++ b/dotcom-rendering/src/server/prod-server.ts
@@ -1,6 +1,5 @@
 import compression from 'compression';
 import express, { Request, Response } from 'express';
-import fetch from 'node-fetch';
 import responseTime from 'response-time';
 import {
 	renderArticle,
@@ -21,20 +20,7 @@ import {
 } from './lib/aws/aws-parameters';
 import { recordBaselineCloudWatchMetrics } from './lib/aws/metrics-baseline';
 import { logger } from './lib/logging';
-
-const buildUrlFromQueryParam = (req: Request) => {
-	// Supports urls such as:
-	// http://localhost:9000/Article?url=https://www.theguardian.com/commentisfree/2020/feb/08/hungary-now-for-the-new-right-what-venezuela-once-was-for-the-left
-	// Note. This is the same as how dev-server.js works
-	const DEFAULT_URL =
-		'https://www.theguardian.com/money/2017/mar/10/ministers-to-criminalise-use-of-ticket-tout-harvesting-software';
-	const query = req.query as { url?: string } | undefined;
-	const url = new URL((query && query.url) || DEFAULT_URL);
-	// searchParams will only work for the first set of query params because 'url' is already a query param itself
-	const searchparams = url.searchParams && url.searchParams.toString();
-	// Reconstruct the parsed url adding .json?dcr which we need to force dcr to return json
-	return `${url.origin}${url.pathname}.json?dcr=true&${searchparams}`;
-};
+import { getContentFromURLMiddleware } from './lib/get-content-from-url';
 
 // Middleware to track route performance using 'response-time' lib
 // Usage: app.post('/Article', logRenderTime, renderArticle);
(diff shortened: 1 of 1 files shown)


mxdvl (approved):
Seems sound. Could you point to the original PR implementing `getContentFromURLMiddleware`?

dotcom-rendering/src/server/prod-server.ts
```diff
@@ -87,34 +73,26 @@ export const prodServer = () => {
...
-			req.body = config;
-			return renderArticle(req, res);
-		} catch (error) {
-			// eslint-disable-next-line no-console
-			console.error(error);
-		}
-	});
+	app.get(
+		'/Article',
+		[logRenderTime, getContentFromURLMiddleware],
```
OllysCoding, line 78: Should this be in an array, could it be `('/Article', logRenderTime, getContentFromURLMiddleware, async () ...)`? I don't know if there's a difference, I just haven't seen array middleware syntax in express before

OllysCoding (approved):

dotcom-rendering/src/server/prod-server.ts
oliverlloyd: The [express docs](https://expressjs.com/en/guide/using-middleware.html#middleware.application) suggest it should be an array.

Their example:

```js
const logStuff = [logOriginalUrl, logMethod]
app.get('/user/:id', logStuff, (req, res, next) => {
  res.send('User Info')
})
```

oliverlloyd:
> Seems sound. Could you point to the original PR implementing `getContentFromURLMiddleware`?

Great idea. I've added links to both the original PR and the one where it was improved

dotcom-rendering/src/server/prod-server.ts
jamesgorrie: Taking the line from the docs:

> Middleware can also be declared in an array for reusability.

Interestingly we've not made that abstraction (and I don't think we should), that just makes me think it's not recommended, just useful, as both ways are used in the docs.

Just a wondering though.

jamesgorrie (approved):
Nice and simple. As these methods aren't tested (as I can see) are we _sure_ they're identical as we're touching `prod-server` here?

dotcom-rendering/src/server/prod-server.ts
oliverlloyd: I think you're both right, it's nicer without the array. Done!

oliverlloyd:
> Nice and simple. As these methods aren't tested (as I can see) are we _sure_ they're identical as we're touching `prod-server` here?

_Technically_ this is prod code but really it's dev code. There was actually [a comment in to code](https://github.com/guardian/dotcom-rendering/blob/c27c5b85e24913bb0770488baf75e1bb696616ad/dotcom-rendering/src/server/prod-server.ts#L28) that made reference to how they (dev and prod local endpoints) are doing the same thing

jamesgorrie:
> Technically this is prod code but really it's dev code.

Little confused on this - these endpoints get hit by frontend don't they?

oliverlloyd:
> > Technically this is prod code but really it's dev code.
> 
> Little confused on this - these endpoints get hit by frontend don't they?

No. `GET /Article` is purely for local development and testing. It's a convenience endpoint so that we don't need to make POST requests from our local machines. The actual prod server only expects POST calls from Frontend

jamesgorrie:
> No. `GET /Article` is purely for local development and testing. It's a convenience endpoint so that we don't need to make POST requests from our local machines. The actual prod server only expects POST calls from Frontend

Crumbs - took me a while to work that out. Apologies.

oliverlloyd:
> > No. `GET /Article` is purely for local development and testing. It's a convenience endpoint so that we don't need to make POST requests from our local machines. The actual prod server only expects POST calls from Frontend
> 
> Crumbs - took me a while to work that out. Apologies.

To be fair, I think it took me 4 months to get my head around the DCR isn't a web server thing

PR #3907, opened 2022-02-01
Add ESLint rule: sort-keys

mxdvl:
## What does this change?

Adds [ESLint’s sort-keys rule](https://eslint.org/docs/rules/sort-keys).
Objects with more than 5 keys are expected to be sorted.

## Why?

More consistency for ever-growing objects.

## Known issues

There are a few examples, such as ab-tests, where the ordering is consistent, but not alphabetical. The proposed approach is to ignore this rule for those files.

PR #4120, opened 2022-03-01
Move Pagination component to common-rendering

iainjchambers-guardian:
## What does this change?
Moves `Pagination` component from `dotcom-rendering` to `common-rendering` and adds corresponding colour values to `editorialPalette`.

## Why?
Apps rendering will need to use the component.

**NOTE** this has added a few whitespace changes due to differing lint rules so please hide whitespace changes in github's `Files changed` view using the gear icon at the top to make the reviewing process a lot easier.


jamesgorrie (changes requested):
(Apologies, it's meant to be a comment, not request changes as I've not suggested any changes).

Just so I understand this - we think that this component is usable in it's current state in `app-rendering`?

Could we not just copy the component across to AR to avoid coupling? This coupling means that we'd need to make changes to Pagination and test across both services, which will make making changes harder and cause delays in delivery. An example: We want to start using infinite scroll, but have a pagination fallback for no-js. 

This fits more inline with our share thinking not code across the services.

JamieB-gu:
> Just so I understand this - we think that this component is usable in it's current state in `app-rendering`?

Yup, it's identical visually and functionally. The work already done in AR to model the data in `pagination.ts` was designed so that the input would line up with this component.

> Could we not just copy the component across to AR to avoid coupling? This coupling means that we'd need to make changes to Pagination and test across both services, which will make making changes harder and cause delays in delivery.

I think the coupling is a feature rather than an issue? It's a good thing to make sure we consider both platforms when adding/changing features so that we avoid bugs and confusion for Editorial and users. I think it typically makes changes harder and delays delivery when you've got lots of _duplicated_ code, because you need to go through and make the change in multiple places?

> An example: We want to start using infinite scroll, but have a pagination fallback for no-js.

I think UX would like to do some research into this? From what I remember we discussed running an AB test across both platforms and getting some user feedback? It should be easier to run that test if the implementations are the same.

> This fits more inline with our share thinking not code across the services.

I don't quite understand what you mean here, could you clarify? Sorry! 🙈

common-rendering/src/editorialPalette.ts
```diff
@@ -781,6 +824,7 @@ const text = {
 	standfirstDark: textStandfirstDark,
 	standfirstLink: textStandfirstLink,
 	seriesTitle: textSeriesTitle,
+	pagination: textPagination,
```
MarSavar, line 827: `editorialPalette` isn't showing the diff because it's too large, but I noticed this while we were mobbing - it looks like fields are in alphabetical order so it would be good to maintain the ordering in all objects and move `pagination` up (in `text`, `background`, `palette` etc)! Total nitpick, I know! :-)

JamieB-gu: I think @mxdvl attempted something like this with ESLint?

jamesgorrie: https://eslint.org/docs/rules/sort-keys
https://github.com/guardian/dotcom-rendering/pull/3907

jamesgorrie:
> I think the coupling is a feature rather than an issue

The feature would be help alleviate design creep - has this been happening with the pagination? 

> we consider both platforms when adding/changing features

Making the change in two places is another way of doing this, with few moving parts. I know historically this has been troublesome, but I weighing this against friction and speed of delivery when we want to test new ideas. Pagination is far from a done-and-dusted part of our UI from what I can tell with all the discussion around it.

Duplicated code, especially if used only twice, is very easy to maintain, and [often considered an early abstraction](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) if _only_ used in two places.

> From what I remember we discussed running an AB test across both platforms and getting some user feedback?

Would users ever expect different things from different platforms? I'd like to hear more about how homogenising would help with this. I would hate for us to say "It's the same component so it should be the same across platforms" backseating potential UX gains because of technical decisions. This is hearsay though, as I am very unfamiliar with our UI testing.

>> This fits more inline with our share thinking not code across the services.

> I don't quite understand what you mean here, could you clarify? Sorry! 🙈

Sharing code, in the current state of our mono-repo, is far from perfect. Before leaning more into it, we should fix the general state of the repo before creating more couplings making it hard to change how it works. Duplicating this code is easy, and updating two places is easy. We should maintain sharing development practise and UX/Design research and implementation too.

jamesgorrie:
Would the AR integration with this just be a matter of moving from `apps-rendering/Pagination.tsx` to `common-rendering/Pagination.tsx`?

jamesgorrie (approved):
To capture thoughts discussed outside the PR.

This is a low-risk easy test on `common-rendering`. It's easy to back out if any of the concerns come to fruition, or for any other concerns (revert, copy/paste, move on). 

It saves time on AR having to implement pagination. 

Saying this - I don't feel we should be putting time into trying to bring the codebases together for all the reasons mentioned in other comments.

Apologies @iainjchambers-guardian for pulling this discussion into this relatively simple PR.

common-rendering/src/components/Pagination.stories.tsx
```diff
@@ -1,13 +1,13 @@
-import { ArticleDesign, ArticleDisplay } from '@guardian/libs';
-import { breakpoints } from '@guardian/source-foundations';
-import { Pagination } from './Pagination';
-import { getAllThemes } from '../../../../common-rendering/src/fixtures/article';
+import { ArticleDesign, ArticleDisplay } from "@guardian/libs";
```
sndrs, line 1: why are all these quote marks changing? this kind of thing should never change in a PR (unless the PR is changeing the quote marks etc)

common-rendering/src/components/Pagination.tsx
```diff
@@ -199,3 +199,5 @@ export const Pagination = ({
 		</Container>
 	);
 };
+
+export default Pagination;
```
sndrs, line 203: why is the export changing? our own configs do not allow default exports elsewhere https://github.com/guardian/configs/blob/d42c048b6ab85d1eab6b8f610ff4162aeb63a510/packages/eslint-config/index.js#L46-L47

sndrs (changes requested):

common-rendering/src/components/Pagination.stories.tsx
sndrs: our prettier config will use single quotes https://github.com/guardian/configs/blob/main/packages/prettier/index.js#L9

sndrs (approved):

PR #4125, opened 2022-03-01
Fix `StickyVideo` isSticky conditional

joecowton1:
## What does this change?

- Reverses the `setIsSticky` conditional so that the video is stuck when the video container is **not** intersecting.
- Moves the intersecting threshold so the IO is triggered when the video is 50% out of the viewport
- removes unnecessary height value from the sticky video styles 

## Why?

Previously the video would only be stuck when the video container was intersecting, for obvious reasons this wasn't the behaviour we wanted. See [this PR](https://github.com/guardian/dotcom-rendering/pull/3992) for context.

### Before

https://user-images.githubusercontent.com/77005274/156212950-45c44afd-c73f-4f50-96a0-acda1156a9ac.mov

### After

https://user-images.githubusercontent.com/77005274/156213009-1094c450-a4e8-4f57-8a3d-b63ad02304c3.mov

### DIFF
--- a/dotcom-rendering/src/web/components/StickyVideo.importable.tsx
+++ b/dotcom-rendering/src/web/components/StickyVideo.importable.tsx
@@ -59,7 +59,7 @@ export const StickyVideo = ({ isPlaying, children }: Props) => {
 	});
 
 	useEffect(() => {
-		if (isPlaying) setIsSticky(!isIntersecting);
+		setIsSticky(isPlaying && !isIntersecting);
 	}, [isIntersecting, isPlaying]);
 
 	return (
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/components/StickyVideo.importable.tsx
```diff
@@ -55,12 +54,12 @@ interface Props {
...
 	const [isSticky, setIsSticky] = useState(false);
 	const [isIntersecting, setRef] = useIsInView({
-		threshold: 0.1,
+		threshold: 0.5,
 		repeat: true,
 	});
 
 	useEffect(() => {
-		setIsSticky(isIntersecting && isPlaying);
+		if (isPlaying) setIsSticky(!isIntersecting);
```
oliverlloyd, line 62: This might cause a bug if isPlaying changes to false but we still want to unstick a stuck element

oliverlloyd (approved):
I think we need to revert to the previous style of setting the `isSticky` value but happy to approve pending that change

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
joecowton1: Ah yes, hadn't thought of that..

PR #4126, opened 2022-03-01
Move `SlotBodyEnd`, `StickyBottomBanner` & `ReaderRevenueLinks` to Islands

mxdvl:
## What does this change?

Migrates `StickyBottomBanner `, `SlotBodyEnd` & `ReaderRevenueLinks` into their own respective islands. These three components need moved together as they had shared dependencies that could not run in both Islands and App.tsx.

To support the move to the Islands pattern, we introduce `useBraze`, a hook which makes use of `useSWRImmutable` to ensure we initialise braze only once per page. 

This PR also introduces the `BrazeMessaging` Island, a canonical usage of `useBraze` to ensure that braze is properly initialised on every page.

### Small improvements
- fix: remove getCmp - CMP is now server safe!
- fix: import type explicitly

## Why?

Islands are the new way: #3629 🏝️


tjmw:
Hi @OllysCoding @mxdvl this is on my TODO list to review, looks really interesting! Just wanted to point out that because this affects `StickBottomBanner` and `SlotBodyEnd` this PR affects all systems which are vying for the opportunity to show something in those slots (so RRCP and also the CMP banner I think), not just Braze.

dotcom-rendering/src/web/lib/useSWRPromise.ts
```diff
@@ -0,0 +1,32 @@
...
+import useSWR from 'swr';
+
+/**
+ * We created this as a wrapper to get a promise from SWR.
+ * This was to prevent a complete refactor of the existing braze code,
+ * which expected a promise.
+ *
+ * @deprecated prefer refactoring the code to use SWR natively
+ */
+export const useSWRPromise = <Data extends unknown, Key extends string>(
```
oliverlloyd, line 11: We only use this code in two places so can we just duplicate it and not have the abstraction? If this code is already deprecated and we want to discourage it's use then we would achieve that much better by simply not creating it?

Also, inlining things is going to make it easier to read the code without the need to jump around files, which is always nice

_And_, if we don't need to pass things around so much we will need fewer types, it's all good stuff.

dotcom-rendering/src/web/layouts/StandardLayout.tsx
```diff
@@ -854,7 +886,27 @@ export const StandardLayout = ({ CAPI, NAV, format, palette }: Props) => {
 				/>
 			</ElementContainer>
 
-			<BannerWrapper data-print-layout="hide" />
+			<BannerWrapper data-print-layout="hide">
```
oliverlloyd, line 889: More a question than an assertion but should this live inside the island?

dotcom-rendering/src/web/lib/braze/buildBrazeMessages.ts
```diff
@@ -1,7 +1,7 @@
 import { getCookie, log, storage } from '@guardian/libs';
+import type { BrazeMessagesInterface } from '@guardian/braze-components/logic';
```
oliverlloyd, line 2: Is this import used?

dotcom-rendering/src/web/lib/getCmp.ts
```diff
@@ -1,24 +0,0 @@
...
- * The @guardian/consent-management-platform sets the window.guCmpHotFix
- * value to equal the `cmp` object.
- *
- * If this function is called before this value has been set then an error
- * is thrown allowing us to gather stats on any race conditions that could
- * exist with this pattern
- *
- * @returns the cmp object
- */
-export const getCmp = () => {
```
oliverlloyd, line 13: 👍 Byeee!

dotcom-rendering/src/web/layouts/ImmersiveLayout.tsx
```diff
@@ -380,7 +382,37 @@ export const ImmersiveLayout = ({
 									webTitle={CAPI.webTitle}
 									ajaxUrl={CAPI.config.ajaxUrl}
 								/>
-								{showBodyEndSlot && <div id="slot-body-end" />}
```
oliverlloyd, line 383: We have [some styles](https://github.com/guardian/dotcom-rendering/blob/8f71e12d717f118d095834362f0fed58ea77ee3b/dotcom-rendering/src/web/components/SignInGate/gateDesigns/shared.tsx) that get applied against this id. It's possible that there's another div with the same id but if not then we might need to think of a way to keep those styles

dotcom-rendering/src/web/layouts/lib/stickiness.tsx
```diff
@@ -57,7 +57,5 @@ export const SendToBack = ({ children }: Props) => (
 );
 
 export const BannerWrapper = ({ children }: Props) => (
-	<aside id="bottom-banner" css={bannerWrapper}>
-		{children}
-	</aside>
+	<aside css={bannerWrapper}>{children}</aside>
```
oliverlloyd, line 60: Now that we're not using the id there's no need for this tag to live here so could be dropped down into the component?

oliverlloyd (commented):
Re @tjmw 's comment, shall we rename this PR to `Move Slots to Islands`?

Such great work! I made an attempt to do this and failed but I really like the gentle approach you've taken here - this makes a lot of sense to me.

One thing, and I know I keep harping on about this but I would like to keep some of these more advanced Typescript features out of DCR. I do fully appreciate that there are some things that we could do that _do_ add value but it's also true that, in isolation, they _all_ add value. The problem is the costs of each of them compound and compound until the DCR codebase becomes intimidating and inaccessible which is very much not what we want for our core platform. This is code that is designed to be used for years to come by a wide range of developers with different skills and experience so we should be aiming for the lowest common denominator rather than choosing to write code that requires specific knowledge.

dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
```diff
@@ -88,12 +93,15 @@ export const SlotBodyEnd = ({
 	isPaidContent,
 	tags,
 	contributionsServiceUrl,
-	brazeMessages,
 	idApiUrl,
 	stage,
 	pageId,
 	keywordsId,
-}: Props) => {
+}: Omit<Props, 'isDev' | 'switches' | 'isSensitive'>) => {
```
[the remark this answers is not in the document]
mxdvl: We did try to avoid it, but you then need two `Props` types which are nearly identical. Either that or you have to use `PropsA & PropsB`, which I wouldn't say is any clearer.

Any idea on how we can tackle this?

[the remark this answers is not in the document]
oliverlloyd: To be honest, I can't suggest an alternative because I (genuinely) don't understand the code or what problem it is trying to solve. My guess is though that if DCR has been able to manage this long without this feature there probably is a way to make it work

dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
```diff
@@ -88,12 +93,15 @@ export const SlotBodyEnd = ({
...
+	const brazeMessages = useSWRPromise('braze-messages', () =>
+		buildBrazeMessages(idApiUrl),
```
tjmw, line 102: To make sure I understand correctly - does the use of `useSWRPromise` here effectively memoize the async work which `buildBrazeMessages` does? Under the hood an instance of `BrazeMessages` is returned which wraps the Braze SDK and I think we only ever want to have one of these on the page at a time. This used to happen because the instance was created in `App` and was passed down to the relevant "slots" as a prop.

I think a difference I can see with the new version of this code is that we'll only call `buildBrazeMessages` if one of (or both of) `SlotBodyEnd` and `StickyBottomBanner` exist in the page hierarchy? That's slightly different from before in `App` where we would _always_ call `buildBrazeMessages`. The reason that distinction is important is because even if the slots/islands don't exist on the page, we still want to call `buildBrazeMessage` currently as it does work to clear up after the SDK if the user has logged out or removed permissions for Braze to load. Any thoughts on how we could maintain that with this new approach? It's maybe not _super_ relevant at the moment because `StickBottomBanner` probably exists in every page layout? Even so it'd be nice to settle on an approach which works if that ever isn't true (now or in future).

dotcom-rendering/src/web/layouts/ImmersiveLayout.tsx
mxdvl: As `id` attributes are meant to be unique in valid HTML, I’ll wager that you’re right.

Let me investigate a bit more.

dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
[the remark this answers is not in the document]
jamesgorrie: @mxdvl on intersection types, we could add clarity and maybe have something to share elsewhere.

e.g.
```TS
// AB.ts
export type Props = {
  switches: Switches;
  isSensitive: boolean;
  isDev: boolean;
}

// SlotBodyEnd.tsx
import { ABProps } from './AB.ts';
type Props = {
  contentType: string;
  // ...
}

const SlotBodyEndWithAB = (props: Props & ABProps) = {}
```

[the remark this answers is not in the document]
mxdvl: @oliverlloyd in this case I think we’re coming up against some of the limitations of the Island pattern. The reason DCR managed this long without this feature is because we could confidently use composition. We found ourselves more comfortable with a single child to an Island, opting for a pattern where `SlotBodyEnd` contains its own `ABProvider`:

```tsx
// Layout.tsx
<Island clientOnly={true}>
  <SlotBodyEnd
    propA="A"
    propB="B"
    propC="C"
    propD="D"
    /* … */
    propM="M"
    propN="N"
  />
</Island>

// SlotBodyEnd.tsx
export const SlotBodyEnd = ({ /* … */ }: Props) => {
	return (
		<WithABProvider
			propA={A}
			propF={F}
			propN={N}
		>
			<SlotBodyEndWithAB
				propA="A"
				propB="B"
				propC="C"
				propD="D"
				/* … */
				propM="M"
			/>
		</WithABProvider>
	);
};
```

The alternative to this would be to allow ABProvider to be importable, but that seems odd.

dotcom-rendering/src/web/lib/braze/buildBrazeMessages.ts
mxdvl: Yes, on line 46. It used to be bundled with the `class` imports, so we extracted it explicitely as a `type`.

dotcom-rendering/src/web/lib/useSWRPromise.ts
mxdvl: Yes, the ideal would be to not create it… however this would mean rewriting the slots’s braze logic completely, which we felt would be best achieved in a subsequent PR.

When pairing on it, using generics helped us confirm that the compiler was in agreement with what we were trying to achieve. We definitely did not want to inline this code because while we strongly believe it needs to go, having two copies of a code that does something we don’t want is worse than a single copy.

dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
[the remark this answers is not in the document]
oliverlloyd: Is the problem here that you don't want to have two sets of props for two different components? And want to connect the different types together and create one from the other? Asking because I am still a bit on the edge of understanding the goal with `Omit`.

If my assumption here is right then I can see the benefit. You don't have to create a new type and have instead got a sort of abstraction that creates it for you. Another benefit is you're coupling the types together so any changes to the root type will be carried down into the child version of it.

The counter to that is we're adding complexity which we could avoid by having duplication and that the benefit of coupling is also a constraint.

> limitations of the Island pattern

I don't think this is entirely true. The reason we're using composition is not really related to the islands pattern but is instead caused by the constraint that hooks cannot be called conditionally. The islands pattern really only means we need to add an `ABProvider`, composition is how we would need to add that no matter if we were using an island or not.

In general though, I am happy to adopt these sorts of features if there were team consensus around them. You can have these sorts of discussions endlessly and there's really no right or wrong here.

dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
mxdvl: Yes, `useSWRPromise` memoize the async work of `buildBrazeMessages`. We might be able to use [`useSWRImmutable`][] under the hood to make sure there’s only ever one instance of the BrazeMEssages.

[`useSWRImmutable`]: https://swr.vercel.app/docs/revalidation#disable-automatic-revalidations

When it comes to making sure something is always called on a page, the current approach is having a script in the browser folder, like we do for `bootCmp` #3661, `initDiscussion` #3873 and [`coreVitals`][]

[`coreVitals`]: https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/browser/coreVitals/init.ts

dotcom-rendering/src/web/layouts/lib/stickiness.tsx
oliverlloyd: Down into the component it wraps. This tag was only put in because we needs a wrapper div to attach an id to but now that requirement is redundant

dotcom-rendering/src/web/lib/useSWRPromise.ts
oliverlloyd: If we know we don't want people to use this abstraction then not creating it is probably the best way to stop that happening. Creating an abstraction that we don't want used is metaphorically asking for trouble.

Alternatively, you could rename it to a name that is less inviting? `useSWRPromise_DEPRECATED`? Although I feel it is weird to create something which is immediately deprecated

tjmw: Would it be possible here to use a plain TS function which memoizes the returned `Promise`? I think we have that pattern elsewhere in DCR. (I guess you'd need to combine with `useEffect`/`useState` to ensure this happens at the right time. This also might be relevant to Max's comment below, if we want to ensure `buildBrazeMessages` is always called: 

> When it comes to making sure something is always called on a page, the current approach is having a script in the browser folder, like we do for bootCmp https://github.com/guardian/dotcom-rendering/pull/3661, initDiscussion https://github.com/guardian/dotcom-rendering/pull/3873 and [coreVitals](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/browser/coreVitals/init.ts)

dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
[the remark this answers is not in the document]
mxdvl: > If my assumption here is right then I can see the benefit. You don't have to create a new type and have instead got a sort of abstraction that creates it for you. Another benefit is you're coupling the types together so any changes to the root type will be carried down into the child version of it.
>
> The counter to that is we're adding complexity which we could avoid by having duplication and that the benefit of coupling is also a constraint.

By placing the ABProvider inside this component, we’re fully coupled, so the types might as well reflect it. There might be something I’m not seeing here. I’ll have a go at making `WithABProvider` an importable.

dotcom-rendering/src/web/layouts/lib/stickiness.tsx
mxdvl: It also has some styling applied to it. You’d move all of that out of here?

dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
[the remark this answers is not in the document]
jamesgorrie: I think some of the complexity lies in how we have decided Islands should be implemented if the need access to the `useAB` hook.

For components to work, with AB hooks, in islands, we need to re-include the `WithABProvider`. We have chosen to make this the responsibility of the component rather than the island (can't find the documentation for this).  

This means we have the pattern:
```TS
BootReact
├─ 
├─ WithABProvider
│  ├─ Island
│  │  ├─ ComponentWithABProvider
│  │  │  ├─ WithABProvider
│  │  │  │  ├─ Component
```

That means `ComponentWithABProvider` require both the types for `WithABProvider` and the `Component`.

To this end, `ComponentWithABProvider` would have the type of `WithABProvider & ComponentProps`. This to me is quite easy to understand, or by using the Omit pattern, although for clarity I might use `Omit<Props, ABProps>`.

The parent component would always have to implement the `<Island><ComponentWithABProvider>` if the component uses the `useAB` hook.

Maybe there is something we can do to simplify this?

dotcom-rendering/src/web/lib/getCmp.ts
mxdvl: This deserves its own PR: #4174

dotcom-rendering/src/web/lib/useSWRPromise.ts
OllysCoding: I actually prefer this being broken out in and purposely deprecated, and I'd support oliver's idea of naming it `useSWRPromise_DEPRECATED`

Why?

copy-pasting code is not a rare thing in DCR, so many people work on this codebase it's commonplace to go look for patterns in other places and use them for your own component/function. This is fine, most of this code will be abstracted when and if it's needed, so it's no big deal to go looking for an example of what you're trying to do.

The advantage of abstracting this pattern is that we can mark it globally as something not to use, so if in a years time this pattern still exists for some reason, but we're all gone, if somebody were to copy and paste the way we were using SWR to generate a promise, there's a decent chance nobody would even realise it was a deprecated pattern.

Having a function, clearly named `useSWRPromise_DEPRECATED` would make it explicitly clear for any future developers that this is not a pattern we want to support. There's no chance it accidentally ends up in another place, at least not without the developers explicitly understanding why it was their only option. 

Therefor I feel like it's the _safer_ way to introduce this code, without risking it evolving or being assumed 'okay' in the future.

dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
mxdvl: Worth noting that there's no standard way for code inside the Island to talk to code running through the script tags. If the fact that we have a Slot island on every page is not sufficient and we go the script way, implementing a mediator is a requirement.

One proposal for such a mediator is https://github.com/guardian/commercial-core/discussions/486.

oliverlloyd: > The reason that distinction is important is because even if the slots/islands don't exist on the page, we still want to call `buildBrazeMessage` currently as it does work to clear up after the SDK if the user has logged out or removed permissions for Braze to load. Any thoughts on how we could maintain that with this new approach?

This is a really interesting and useful point!

So the thing that SWR brings to the party for us is request deduplication. It takes any promised based function along with a `key` value and wraps the call to it in some dedupe logic such that any subsequent call will return the same result.

The fact that this SWR takes any function, not just a fetch call, is powerful. It means we can dedupe kind of anything. This is potentially very powerful for our islands architecture. Where we have static global state shared between islands if we wrap the call to get this state using SWR then we effectively share it. Now, I used the words 'static' and 'effectively' there because the limitation to this SWR trick is we don't share dynamic changes to the state. If one island mutates it's state value the other islands are not aware of this.

To solve the issue you raised where we ideally always want to call this function, not just when a slot appears on the page, we could have a _third_ island, one that only makes the call to `buildBrazeMessages`. This will ensure that the call is always made but because of SWR deduplication it will not mean we make any additional actual executions of this function.

tjmw: > To solve the issue you raised where we ideally always want to call this function, not just when a slot appears on the page, we could have a third island, one that only makes the call to `buildBrazeMessages`

Ah cool, yeah that makes sense. If that fits with the way you (as in dotcom) are thinking about islands then I'd be in favour of adding that as part of this PR. That way even if it's true now that `StickBottomBanner` appears on ever page, it guards against that changing in future and going un-noticed. I wonder what a good name is? Would this be the first island where there's no visual/UI element to it?

oliverlloyd: > Would this be the first island where there's no visual/UI element to it?

Not the first, no. [Liveness](https://github.com/guardian/dotcom-rendering/blob/4b1182429a41e65729f968cd10944b8595f773a3/dotcom-rendering/src/web/components/Liveness.importable.tsx) _sometimes_ returns a component but is mostly just a vehicle for code and [CommercialMetrics](https://github.com/guardian/dotcom-rendering/blob/8d0e112b92769bd6266312667c8c09377477c76c/dotcom-rendering/src/web/components/CommercialMetrics.importable.tsx) never returns any html. There might be others.

dotcom-rendering/src/web/lib/useSWRPromise.ts
```diff
@@ -0,0 +1,40 @@
...
+	fetcher: Fetcher<Data, Key>,
+): Promise<Data> => {
+	let resolver: (value: Data | PromiseLike<Data>) => void;
+	let rejecter: (reason?: string) => void;
+	const promise: Promise<Data> = new Promise((resolve, reject) => {
+		resolver = resolve;
+		rejecter = reject;
+	});
+
+	useSWRImmutable(key, fetcher, {
```
oliverlloyd, line 30: Nice. I really like how this communicates intent!

oliverlloyd (approved):
Pending Tom's point about needed to ensure we always clear us (eg. we should always run the `buildBrazeMessages` function) this look's great!

I appreciate there's a lot going on here though and I know you've got some other sub PRs but I'm happy for you to decide what's still needed and when best to merge things.

tjmw:
I'd love to get the chance to see this running on CODE before merging if possible. We've got some test Braze campaigns running in CODE, so a good chance to check everything looks good end to end.

mxdvl:
We’ve introduced a `BrazeMessaging` Island to ensure the code always runs.

We’ve also checked locally that the code never runs more than once:

image

dotcom-rendering/cypress/support/index.js
```diff
@@ -42,5 +42,12 @@ Cypress.on('uncaught:exception', (err, runnable) => {
 		return false;
 	}
 
+	if (
+		err.message.includes("Cannot read properties of null (reading '__H')")
+	) {
+		console.warn(err);
+		return false;
+	}
```
OllysCoding, line 50: Are we sure that all the banners _actually_ work when this error is ignored? When I was working on this, this error generally indicated that something had gone wrong during the preact rendering process, and that the braze messages were not working.

I could usually verify this by going to the articles in the test and seeing that the banner(s) would not show up, but I didn't manage to figure out the root cause.

dotcom-rendering/src/web/components/BrazeMessaging.importable.tsx
```diff
@@ -0,0 +1,22 @@
+import { log } from '@guardian/libs';
+import { useBraze } from '../lib/useBraze';
+
+type Props = {
+	idApiUrl: string;
+};
+
+/**
+ * This component ensure we call buildBrazeMessaging at least once
```
OllysCoding, line 9:
```suggestion
 * This component ensures we call buildBrazeMessaging at least once
```

dotcom-rendering/src/web/components/BrazeMessaging.importable.tsx
```diff
@@ -0,0 +1,22 @@
...
+ * on every page
+ */
+export const BrazeMessaging = ({ idApiUrl }: Props) => {
+	// eslint-disable-next-line no-void -- we don’t use this promise
```
OllysCoding, line 13: I think because of the line below (`if (brazeMessages) ...`) we shouldn't need this eslint-disable?

dotcom-rendering/src/web/components/StickyBottomBanner.importable.tsx
```diff
@@ -207,7 +211,12 @@ export const StickyBottomBanner = ({
 	switches,
 	pageId,
 	keywordsId,
-}: Props) => {
+}: Props & {
+	switches: Switches;
+	isSensitive: boolean;
+}) => {
```
OllysCoding, line 217: It feels like at this point that the two sets of props for `StickyBottomBanner` and `StickyBottomBannerWithAB` are similar enough that they should both use the same props type?

Perhaps it should be defined as 
```
type Props = {
    ...
} & ABProps;
```

Then we can change both `Props & { switches: Switches; isSensitive: boolean; }` and `Props & ABProps` to just `Props`

Then just make sure we use the correct naming scheme (`pageIsSensitive`, `abTestSwitches`) in both components.

dotcom-rendering/src/web/components/StickyBottomBanner.importable.tsx
```diff
@@ -207,7 +211,12 @@ export const StickyBottomBanner = ({
 	switches,
 	pageId,
 	keywordsId,
-}: Props) => {
+}: Props & {
+	switches: Switches;
```
OllysCoding, line 215: Having had a second look at the uses of this switches array, it seems like they're used to look for the keys `puzzlesBanner` and `remoteBanner`, which are not AB switches, but on line 339 we pass in `abTestSwitches`. 

I think if https://github.com/guardian/dotcom-rendering/pull/4190 then the switches required would then never be available

dotcom-rendering/src/web/components/Footer.tsx
```diff
@@ -202,7 +212,16 @@ const FooterLinks: React.FC<{
 
 	const rrLinks = (
 		<div css={readerRevenueLinks}>
-			<div id="reader-revenue-links-footer" />
+			<Island deferUntil="idle" clientOnly={true}>
```
oliverlloyd, line 215:
```suggestion
			<Island deferUntil="visible" clientOnly={true}>
```

Does it work if we use visible for the footer?

dotcom-rendering/src/web/components/Footer.tsx
```diff
@@ -202,7 +212,16 @@ const FooterLinks: React.FC<{
...
+				<ReaderRevenueLinks
+					urls={urls}
+					edition={edition}
+					dataLinkNamePrefix="footer : "
+					inHeader={false}
+					remoteHeader={false}
```
oliverlloyd, line 221: Should this be a switch?

dotcom-rendering/src/web/layouts/InteractiveImmersiveLayout.tsx
```diff
@@ -470,10 +471,33 @@ export const InteractiveImmersiveLayout = ({
...
 					pillars={NAV.pillars}
+					urls={CAPI.nav.readerRevenueLinks.header}
+					edition={CAPI.editionId}
+					contributionsServiceUrl={CAPI.contributionsServiceUrl}
 				/>
 			</ElementContainer>
 
-			<BannerWrapper />
+			<BannerWrapper>
+				<Island clientOnly={true}>
```
oliverlloyd, line 481: When idle?

dotcom-rendering/src/web/layouts/InteractiveLayout.tsx
```diff
@@ -683,10 +689,33 @@ export const InteractiveLayout = ({ CAPI, NAV, format, palette }: Props) => {
...
 					pillars={NAV.pillars}
+					urls={CAPI.nav.readerRevenueLinks.header}
+					edition={CAPI.editionId}
+					contributionsServiceUrl={CAPI.contributionsServiceUrl}
 				/>
 			</ElementContainer>
 
-			<BannerWrapper data-print-layout="hide" />
+			<BannerWrapper data-print-layout="hide">
+				<Island clientOnly={true}>
```
oliverlloyd, line 699: Ditto on idle - etc.

dotcom-rendering/src/web/components/ReaderRevenueLinks.importable.tsx
```diff
@@ -178,8 +177,6 @@ const ReaderRevenueLinksRemote: React.FC<{
 	const [SupportHeader, setSupportHeader] = useState<React.FC | null>(null);
 
 	useOnce((): void => {
-		setAutomat();
```
OllysCoding, line 181: What change does moving 'setAutomat' make here? cc @AshCorr

oliverlloyd (approved):
I'm approving parts of this work but pending eyes from others

AshCorr (approved):
Looks A Ok to me!

dotcom-rendering/cypress/support/index.js
oliverlloyd: This was a great call 👍

oliverlloyd:
@jamesgorrie @mxdvl @tjmw This PR looks like it's ready to merge!

I think [this part](https://github.com/guardian/dotcom-rendering/pull/4126/files#r826211038) of the PR is especially nice

We've all contributed to this work recently and @AshCorr , @OllysCoding and I have been working today to finalise the changes. We're planning on making a final deployment to `code` in the morning and then merging. Let us know if you want to be involved.

tjmw:
Hi @oliverlloyd thanks for the heads up! Yep would love to take a final glance on code before this gets merged. Would you mind pinging me when it's there?

oliverlloyd:
I've deployed this branch to `CODE` for verification

tjmw (approved):

PR #4128, opened 2022-03-01
Remove Babel From AR Jest Config

JamieB-gu:
## Why?

I think Babel existed in our Jest config solely to transpile ES modules for Node. Now that we're on Node 14, which has support for ES modules, this is no longer needed.

## Changes

- Removed the babel-specific parts of our Jest config
- Removed `babel-jest` from our direct dependencies


mxdvl (approved):
There shall be no Babel!

PR #4135, opened 2022-03-02
Add more stories to the Standard Display section in Storybook (AR Layouts)

MarSavar:
## Why?

More stories to capture Standard display options.
DCR stories include `NumberedList`, `Labs` and `SpecialReport` which I didn't include in this PR because `NumberedList` is a Display type separate from Standard, and `Labs` and `SpecialReports` are special themes.

A note on the possibly redundant interfaces that I defined in `item.ts`.
I'm aware that this interface is supposed to capture all the Design types  not included in the `Exclude` type:
```typescript
interface Standard extends Fields {
	design: Exclude<
		ArticleDesign,
		| ArticleDesign.LiveBlog
		| ArticleDesign.DeadBlog
		| ArticleDesign.Review
		| ArticleDesign.Comment
		| ArticleDesign.Letter
		| ArticleDesign.Editorial
	>;
	body: Body;
}
```

**However**, fixtures are defined as being of type `Item`, which is a union type of the following:

```typescript
type Item =
	| LiveBlog
	| DeadBlog
	| Review
	| Comment
	| Standard  // <----- this is the interface above
	| Interactive
	| MatchReport
	| Letter
	| Obituary
	| Editorial
	| Correction;
```

This isn't a problem per se, but the `<Standard>` layout component (which I used to render most of these stories) accepts an `item` of type `Standard | Review | MatchReport`, which throws a bunch of TypeScript errors. For instance, passing the `quiz` fixture to `Standard`  causes the following errors: `Type 'LiveBlog' is not assignable to type 'Review | Standard | MatchReport'`, `Argument of type 'unknown[]' is not assignable to parameter of type 'BodyElement[]'`.

The workaround is to define a `Quiz` interface explicitly instead of falling back to `Standard`. This isn't a new idea and there are more example of this in `item.ts` so I just followed that example. 

### Screenshot

image

### DIFF
--- a/apps-rendering/src/item.ts
+++ b/apps-rendering/src/item.ts
@@ -126,7 +126,33 @@ interface Correction extends Fields {
 	design: ArticleDesign.Correction;
 	body: Body;
 }
+interface Interview extends Fields {
+	design: ArticleDesign.Interview;
+	body: Body;
+}
 
+interface Quiz extends Fields {
+	design: ArticleDesign.Quiz;
+	body: Body;
+}
+interface Recipe extends Fields {
+	design: ArticleDesign.Recipe;
+	body: Body;
+}
+
+interface Feature extends Fields {
+	design: ArticleDesign.Feature;
+	body: Body;
+}
+interface PhotoEssay extends Fields {
+	design: ArticleDesign.PhotoEssay;
+	body: Body;
+}
+
+interface PrintShop extends Fields {
+	design: ArticleDesign.PrintShop;
+	body: Body;
+}
 // Catch-all for other Designs for now. As coverage of Designs increases,
 // this will likely be split out into each ArticleDesign type.
 interface Standard extends Fields {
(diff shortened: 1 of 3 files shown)


apps-rendering/src/item.ts
```diff
@@ -126,7 +126,33 @@ interface Correction extends Fields {
...
+}
+interface PhotoEssay extends Fields {
+	design: ArticleDesign.PhotoEssay;
+	body: Body;
+}
+
+interface PrintShop extends Fields {
+	design: ArticleDesign.PrintShop;
+	body: Body;
+}
```
JamieB-gu, line 155: We may want to a) add these to the `Exclude` statement on `Standard` so that we're not specifying them in two places, or b) consider committing to getting rid of the `Exclude` statement entirely and having an `interface` for each `Design`.

This is perhaps not directly related to storybook layouts so something for a follow-up. Perhaps we should chat about it?

JamieB-gu (approved):

MarSavar:
[Complete Standard display stories](https://trello.com/c/bMrzsSey/721-complete-standard-display-stories)

apps-rendering/src/item.ts
MarSavar: Yes, let me card it up!

PR #4137, opened 2022-03-02
Identity | Sign In Gate | Make dismiss button accessible through keyboard

coldlink:
## Why?
Within the sign in gate, the "I'll do it later" button link was not accessible through keyboard commands. This is frustrating for keyboard only users as they do not have the ability to interact with the link, dismiss the gate, and the sign in gate will remain present.

We solve this by changing the component from a `LinkButton` to a `Button`, which makes it focusable and interactable via keyboard.

### Before
https://user-images.githubusercontent.com/13315440/156404096-ffacd475-d774-4aa2-817d-ca2b81780fa1.mov

### After
https://user-images.githubusercontent.com/13315440/156403145-4fc2ca76-dca4-4045-8eda-78bf39ac2cf7.mov

### DIFF
--- a/dotcom-rendering/src/web/components/SignInGate/gateDesigns/SignInGateFakeSocial.tsx
+++ b/dotcom-rendering/src/web/components/SignInGate/gateDesigns/SignInGateFakeSocial.tsx
@@ -1,6 +1,6 @@
 import { css } from '@emotion/react';
 
-import { LinkButton, Link } from '@guardian/source-react-components';
+import { LinkButton, Link, Button } from '@guardian/source-react-components';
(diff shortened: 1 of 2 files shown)


liamness:
Would it be simpler to make it a `<Button />` instead of a <LinkButton />` as it seems to have no `href` attribute?

coldlink:
> Would it be simpler to make it a `<Button />` instead of a `<LinkButton />` as it seems to have no `href` attribute?

Good shout! I've changed it to button, and that's really simplified it.

SiAdcock (approved):
Lovely job 👏  Thanks @coldlink

PR #4132, opened 2022-03-02
Dcr pinned post

abeddow91:
## Why?

This PR adds a JS implementation for pinned posts to liveblogs on DCR following its roll out on the Frontend platform.

PR #4138, opened 2022-03-02
ServerSideOnly Component 

OllysCoding:
## Adds a new pattern - `<ServerSideOnly>`

This pattern allows the prevention of client-side hydration within the DOM. Instead of attempting to hydrate and compare the preact rendered output to the DOM (as provided by the server), this component will cause preact to skip this comparison and continue with hydration outside of the component.

---
```jsx
<ServerSideOnly>
<p> This tag is not hydrated </p>
</ServerSideOnly>
```
This could be useful if you had a large section of the DOM you knew *wasn't* going to change on the client, it could save client processing power and reduce TTI.

---

Another application of this is avoiding hydration where data & javascript doesn't exist within the island. An example of this is in #4132 where we use this to avoid having to include the `LiveBlock` component in the `PinnedPost.importable.tsx` bundle.

Consider this minimal example:

```jsx
// server.tsx
<Island>
    <MyImportableComponent someProp={stuff}>
         <ComponentNotInClientBundle moreProp={moreStuff} />
    </MyImportableComponent>    
</Island>
```

```jsx
// MyImportableComponent.importable.tsx
<div>
    <div> ... some stuff is hydrated ... </div>
     <ServerSideOnly>{children}</ServerSideOnly>    
</div>
```

In this example, `ComponentNotInClientBundle` gets rendered correctly on the server, but none of the code or data is included on the client, potentially saving us valuable KBs, and, importantly, allowing us to render client-unsafe components 'within' an island.

---
TLDR; being able to avoid hydration = good


apps-rendering/package-lock.json
```diff
@@ -19313,27 +19318,6 @@
...
-          "resolved": "https://registry.npmjs.org/acorn/-/acorn-8.7.0.tgz",
-          "integrity": "sha512-V/LGr1APy+PXIwKebEWrkZPwoeoF+w1jiOBUmuxuiUIaOHtob8Qc9BTrYo7VuI5fR8tqsy+buA2WFooR5olqvQ=="
-        },
-        "acorn-walk": {
-          "version": "8.2.0",
-          "resolved": "https://registry.npmjs.org/acorn-walk/-/acorn-walk-8.2.0.tgz",
-          "integrity": "sha512-k+iyHEuPgSw6SbuDpGQM+06HQUa04DZ3o+F6CSzXMvvI5KMvnaEqXe+YVe555R9nn6GPt404fos4wcgpw12SDA=="
-        }
-      }
-    },
```
OllysCoding, line 19336: Not sure why there's an AR package-lock.json change in here, does this look okay? @JamieB-gu @iainjchambers-guardian @MarSavar

dotcom-rendering/src/web/components/ServerSideOnly.tsx
```diff
@@ -0,0 +1,27 @@
...
+};
+
+/**
+ * ServerSideOnly - Use this if you have some JSX you want to be rendered on the server - but not hydrated on the client
+ *
+ * How does this work? When on the server, this component will render its children within a <div> normally.
+ * 	Once on the client, it will render a <div> with an empty dangerouslySetInnerHTML. This communicates to (p)react that there
+ * 	should be HTML in the <div>, but because it's set with dangerouslySetInnerHTML, (p)react will not parse it as part of hydration.
+ *
+ * When do I need this? DCR will only hyrdrate code that lives within islands -
```
arelra, line 12:
```suggestion
 * When do I need this? DCR will only hydrate code that lives within islands -
```

dotcom-rendering/src/web/components/ServerSideOnly.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * 	you should only need to use this component if you have specific code within an island you do not want to be hydrated.
+ *
+ * Note: Keep in mind this function prevents hydration, but does not prevent code from being included in the webpack bundle.
```
arelra, line 15: Wow :)

Normally when hydrating, the SSR output and the hydrated component [have to match](https://reactjs.org/docs/react-dom.html#hydrate) in order for hydration to correctly 'mount'.

So just for clarification is the `dangerouslySetInnerHTML` a P/React 'trick' to skip hydration for that particular component and therefore the hydration rule around matching output is locally suspended and does not apply?

OllysCoding:
> Maybe we should always only hydrate the Island's child, but never it's grandchildren?

Our option for this would be to prevent the children prop on an <Island> child

AshCorr (approved):
👍 

Is this something we'd want to try and use as much as possible? Also do we have any idea  what the tradeoffs will be? I'm assuming we're getting potentially faster hydration at the cost of a slightly larger bundle

mxdvl:
@AshCorr I think the bundle will be only marginally heavier, as it’s a couple of lines of code. The main benefit is that you can have HTML inside an Island that doesn’t get modified by the Island, and for which you don’t need to bundle to code.

A good example that @OllysCoding gave of “client-unsafe” component is one that includes `renderElement`, which would basically mean you ship all of DCR’s code to the client!

jamesgorrie:
This is a really interesting solution! It would be interesting to see the stats of how this would affect perf. Assuming this _should_ improve Time to Interactive and islands bundle size? 

While playing with it I did wonder if we've exhausted our current implementation? Could we just ensure we never hydrate large components?

Using the `PinnedPost` as an example


```tsx
// With new <ServerSideOnly>
// PinnedPost.importable.tsx
const PinnedPost = ({ children }) => {
  const pinnedPost = useRef()
  const pinnedPostButton = useRef()
  const wrapperRef = useRef()
  
  const [s, setS] = useState('ding')
  useEffect(() => {
    setTimout(() => setS('dong'), 1000)
  }, [])

  useEffect(() => {
    const contentFitsContainer =
		  pinnedPost?.scrollHeight <= pinnedPost?.clientHeight;

    if (contentFitsContainer) {
      pinnedPostButton.style.display = 'none';
    }
  }, [])

  return (
    <div data-gu-marker="pinned-post" ref={pinnedPost}>
      <ServerSideOnly>
        <div id="wrapper" ref={wrapperRef}>
          <h2>Pinned post</h2>
          <p>{s}</p> // This will always remain 'ding'
          {children}
        </div>
      </ServerSideOnly>
      
      <p>{s}</p> // This will become 'dong' after a second.
      <p>{wrapperRef?.id}</p> // I'm not sure this will work

      <button id="pinned-post-button" ref={pinnedPostButton}>Show / hide</button>
    </div>
  )
}

// with our current implementation
// PinnedPost.tsx
const PinnedPost = ({ children }) => {
  return (
    <div data-gu-marker="pinned-post">
      <h2>Pinned post</h2>
      {children}
      <Island><PinnedPostLabel /></Island>
    </div>
  )
}

// PinnedPostLabel.importable.tsx
const PinnedPostLabel = ({ children }) => {
  const [s, setS] = useState('ding')
  useEffect(() => {
    setTimout(() => setS('dong'), 1000)
  }, [])

  useEffect(() => {
    // This is a bit rubbish, but raises an interesting question around sharing
    // client side state between a non-island and an island
    const pinnedPost = window.document.querySelector('[data-gu-marker=pinned-post]')
    pinnedPost.scrollHeight <= pinnedPost.clientHeight
  }, [])

  return (
    <div>
      <div>Show / hide {s}</div> // s will change as this component will be hydrated
    </div>
  )
}
```

A concern with having a component with `<ServerSideOnly>` is you are in a world where you can use client side  information partially (effects from `useEffect`, `DOM` etc), which can lead to seemingly unexpected behaviour e.g. state being different in different parts of the component (see the `setS` example above).

The distinction at the moment of "Island? Client code runs. No Island, no dice" is quite a simple one.

PR #4123, opened 2022-03-01
Bumped `aws-cdk` To 1.147.0

JamieB-gu:
## Why?

We were having issues bumping some transitive dependencies of `aws-cdk` - we couldn't get `npm` to bump them at all despite them using `^` to specify a version range. This PR attempts bumping `aws-cdk` to its latest (in the 1.x series) version instead. This has worked, but it appears to have removed nearly all the packages it was depending on in the lockfile, which seems surprising 🤔

**Note:** It's now specified using `^` notation, which allows bumping across minor versions. Let me know if this is a problem and the version should remain specified exactly as before.

## Changes

- Bumped `aws-cdk` to 1.147.0 (the latest version in the 1.x series)

PR #4139, opened 2022-03-02
Move `vm2` Into `aws-cdk` Transitive Deps

JamieB-gu:
## Why?

Running `npm i` on local machines seems to cause our `package-lock.json` to revert to an older version of `vm2`. It's essentially reverting #4124.

My theory is that `aws-cdk` uses a `npm-shrinkwrap.json` file to specify exact versions of its dependencies. `vm2` is a transitive dependency of `aws-cdk` a few levels deep. It's actually specified using a `^`, so in theory a more recent version should be allowed, but the shrinkwrap is pinned to an older version as mentioned.

Usage of shrinkwrap files in published packages is quite rare (I think this is the only instance of one across all our top-level dependencies). The docs state that (https://docs.npmjs.com/cli/v6/configuring-npm/shrinkwrap-json):

> The recommended use-case for npm-shrinkwrap.json is applications deployed through the publishing process on the registry: for example, daemons and command-line tools intended as global installs or devDependencies. It's strongly discouraged for library authors to publish this file, since that would prevent end users from having control over transitive dependency updates.

This transitive dependency issue is what we're seeing here.

### The Workaround

I believe I've managed to get a more recent version of `vm2` pinned without `npm i` reverting it in this PR. I'm not sure this is intended behaviour; my understanding of shrinkwraps is that they're meant to _prevent_ this from happening. However, it does work on at least two different local machines.

**Note:** I achieved this by running the following commands:

```sh
npm upgrade --depth=5 degenerator
npm i
```

### Longer Term Solution

All that said, I think the long-term solution would be to update `aws-cdk` to a more recent version that has updated dependencies specified in its shrinkwrap file. Or possibly even to a version that doesn't use shrinkwrap any more - I can't see one being used in version `1.147.0` (the latest version in the 1.x series) for example.

As @akash1810 mentioned in https://github.com/guardian/dotcom-rendering/pull/4123#pullrequestreview-896590217 this isn't something we're able to do on our own in this project. We'll need GuCDK to update first before we make the changes here, so that the versions remain in sync.

cc @guardian/devx-operations 

### A Note On Yarn

It's possible that the behaviour will be different in Yarn-based projects, as it seems Yarn doesn't support shrinkwraps: https://classic.yarnpkg.com/lang/en/docs/migrating-from-npm/

## Changes

- Used `npm` to pull `vm2` into `aws-cdk`'s transitive deps in the lockfile

### DIFF
--- a/apps-rendering/package-lock.json
+++ b/apps-rendering/package-lock.json
@@ -7818,6 +7818,16 @@
           "resolved": "https://registry.npmjs.org/@types/node/-/node-10.17.60.tgz#35f3d6213daed95da7f0f73e75bcc6980e90597b",
           "integrity": "sha512-F0KIgDJfy2nA3zMLmWGKxcH2ZVEtCZXHHdOQs2gSaQ27+lNeEfGxzkIw90aXswATX7AZ33tahPbzy6KAfUreVw=="
         },
+        "acorn": {
+          "version": "8.7.0",
+          "resolved": "https://registry.npmjs.org/acorn/-/acorn-8.7.0.tgz",
+          "integrity": "sha512-V/LGr1APy+PXIwKebEWrkZPwoeoF+w1jiOBUmuxuiUIaOHtob8Qc9BTrYo7VuI5fR8tqsy+buA2WFooR5olqvQ=="
+        },
+        "acorn-walk": {
+          "version": "8.2.0",
+          "resolved": "https://registry.npmjs.org/acorn-walk/-/acorn-walk-8.2.0.tgz",
+          "integrity": "sha512-k+iyHEuPgSw6SbuDpGQM+06HQUa04DZ3o+F6CSzXMvvI5KMvnaEqXe+YVe555R9nn6GPt404fos4wcgpw12SDA=="
+        },
         "agent-base": {
           "version": "6.0.2",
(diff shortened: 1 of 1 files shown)


akash1810:
> I can't see one being used in version `1.147.0` (the latest version in the 1.x series) for example.

Should `@guardian/cdk` get a new release to use aws-cdk 1.147.0 and we bump the version of `@guardian/cdk` being used in this repo instead then?

JamieB-gu:
> Should `@guardian/cdk` get a new release to [use aws-cdk 1.147.0](https://github.com/guardian/cdk/pull/1139) and we bump the version of `@guardian/cdk` being used in this repo instead then?

I haven't yet come across a PR or release notes explicitly specifying that they've dropped the shrinkwrap, but when I pulled down a copy locally I couldn't see it any more. That said, even supposing it _is_ still there, a more recent copy may be pinning more recent dependency versions which would solve the problem temporarily.

Presumably in this repo we'll need to bump `@guardian/cdk` and all the direct CDK dependencies in sync? So `aws-cdk`, all the `@aws-cdk/*` packages etc.?

JamieB-gu:
As of #4186 `vm2` is no longer a dependency, so this PR isn't needed any more.

PR #4140, opened 2022-03-02
Implement `StickyVideo` wrapper in `YoutubeBlockComponent` with feature switch

joecowton1:
## What does this change?

This PR connects the `stickyVideos` feature switch to the `YoutubeBlockComponent`, removes the old `StickyVideos` wrapper, and adds tracking of the `STICKY_VIDEO` component events. 

The `stickyVideos` feature flag is passed into the `shouldStick` prop on the `YoutubeAtom`.

### Details
- pass the `switches.stickyVideos` feature switch into the `YoutubeBlockComponent`
- adds component tracking to the `YoutubeBlockComponent` for `STICK` and `CLOSED` events (this was previously in the `StickyVideo` wrapper)
- remove  the `StickyVideo` wrapper as that functionality now lives in [Atoms-Rendering](https://github.com/guardian/atoms-rendering/pull/346)
- updates Atoms-Rendering to the latest version

**Before merging we need to turn off the `stickyVideos` switch on PROD**

### Video

https://user-images.githubusercontent.com/77005274/158564840-26bcd2fd-2588-4ce6-8809-cf4807263456.mov


dotcom-rendering/src/web/components/YoutubeBlockComponent.importable.tsx
```diff
@@ -84,11 +85,16 @@ export const YoutubeBlockComponent = ({
...
 	origin,
+	stickyVideos,
 }: Props): JSX.Element => {
 	const [consentState, setConsentState] = useState<ConsentState | undefined>(
 		undefined,
 	);
 
+	// TODO: connect StickyVideo wrapper once
+	// feature flag is operational
+	console.log('Videos should stick', stickyVideos);
```
joecowton1, line 96: https://trello.com/c/DO8DYFta/251-sticky-video-connect-stickyvideo-wrapper-component-to-youtubeblockcomponent-and-main-media

dotcom-rendering/src/web/components/YoutubeBlockComponent.importable.tsx
```diff
@@ -167,57 +175,69 @@ export const YoutubeBlockComponent = ({
 		});
 	};
 
+	const videoState = (trackingEvent: string) => {
```
marjisound, line 178: I know you mentioned earlier that we the `YoutubeAtom` is not yet emitting `stop` event, but are you probably thinking of also using the stop event when it's implemented?
Also I noticed that the `play` event is only emitted the first time when the video is played from the `YoutubeAtom` but do you think that is sufficient for us?

marjisound (approved):
Looks great 👍

dotcom-rendering/src/web/components/YoutubeBlockComponent.importable.tsx
joecowton1: The `YoutubeAtom` is currently emitting a stop event as `end` - what it's not doing as yet is exposing a way to stop a video from another component. A PR is inflight for that here: https://github.com/guardian/atoms-rendering/pull/337

I think in the case of the sticky video tracking we will want to fire the `close` event when the close button is pressed - this doesn't rely on the video being stopped, but once we've updated the `YoutubeAtom` we should also be able to stop the video when the close button is clicked. Please shout if that doesn't make sense :)

marjisound: Maybe I haven't fully understood the requirement. Please correct me if I'm wrong, From what I see the `end` event seems to be emitting when the video is finished at the end. But when the video is stopped, there's no event emitting from the `YoutubeAtom`. Also when the video is played, only the first time, the `play` event is emitted. But any time after that there's no `play` event emitted.

https://user-images.githubusercontent.com/15894063/157035302-e3e8ec0e-2a74-41bb-9248-2ceb0ee8df63.mov

marjisound: Oh I now see as maybe as part of your change, you have changed how the youtubeAtom is behaving. Let me test it with your version of atoms-rendering :)

joecowton1: I think maybe the thing I'm trying to say is that we don't need to send an `end` event if the video is closed early - that should be a `close` event and that's something we can trigger from the button. Does that make sense?

joecowton1: With regard the play events, I think we're only interested in the fact the user has started the video playing - we don't want to document every time they press play.

joecowton1: Worth remember too that this PR is still a "draft" so don't expect it to work perfectly yet!

joecowton1: This [PR](https://github.com/guardian/atoms-rendering/pull/339) adds event emitters for `resume`, `pause` and `cued` (stopped). Hopefully that's all we need to replicate the sticky video behavior in apps.

abeddow91 (approved):
Nice one, looks good!

PR #4141, opened 2022-03-02
Add tracking to sticky video

marjisound:
## What does this change?
This PR adds tracking events for the sticky video component. The tracking events are as followed:
- STICK => when the video is stuck for the first time in the page or the first time after it's closed
- CLOSE => when the already stuck video is closed so become unstuck

This PR also updates the guardian/libs in all sub projects (apps-rendering, common-rendering & dotcom-rendering).

The tracking events are sent to ophan. The new events were also created:
- in ophan thrift https://github.com/guardian/ophan/pull/4440
- in guardian/libs https://github.com/guardian/libs/pull/315 & https://github.com/guardian/libs/pull/316
- in ophan-data-lake https://github.com/guardian/ophan-data-lake/pull/6620
### After
This video shows when the events are emitted in the console logs. 
 

https://user-images.githubusercontent.com/15894063/157460147-d4b9407d-e4d0-4d23-9af8-e848862ba0d0.mov


dotcom-rendering/src/web/components/StickyVideo.importable.tsx
```diff
@@ -91,27 +92,70 @@ const stickyContainerStyles = (height: number) => css`
...
+			isClosed: false,
+		});
 	}, [isIntersecting, isPlaying]);
 
+	useEffect(() => {
+		const ophanTracking = (trackingEvent: OphanAction) => {
+			if (!id) return;
+			submitComponentEvent({
+				component: {
+					componentType: 'TIMELINE_ATOM',
```
oliverlloyd, line 126: `TIMELINE_ATOM` - I presume this slipped in by mistake?

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
```diff
@@ -91,27 +92,70 @@ const stickyContainerStyles = (height: number) => css`
...
 	useEffect(() => {
-		setIsSticky(isPlaying && !isIntersecting);
+		setStickyState({
+			isSticky: isPlaying && !isIntersecting,
+			isClosed: false,
+		});
 	}, [isIntersecting, isPlaying]);
 
+	useEffect(() => {
+		const ophanTracking = (trackingEvent: OphanAction) => {
```
oliverlloyd, line 122: If we remove the early return for `id` we don't need this function at all, I guess? So then we can just inline the tracking requests (as I'm sure you're aware of by now, I'm a fan of inlining code!)

oliverlloyd (commented):
Sorry, lots of comments!

They are all sent with code 💙 hopefully they are useful

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
```diff
@@ -91,27 +92,70 @@ const stickyContainerStyles = (height: number) => css`
...
+			if (!id) return;
```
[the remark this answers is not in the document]
marjisound: I hope I understood correctly. I think check for the id logic can be done out of the `useEffect` as we don't want to render the sticky video if there's no video id

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
marjisound: Good suggestion. By removing the id check, now I'm calling `submitComponentEvent` in line wherever it needs to emit tracking events.

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
```diff
@@ -91,27 +92,70 @@ const stickyContainerStyles = (height: number) => css`
...
+interface StickyState {
+	isSticky: boolean;
+	isClosed: boolean;
+}
+
+export const StickyVideo = ({ isPlaying, children, id }: Props) => {
+	const [stickyState, setStickyState] = useState<StickyState>({
+		isSticky: false,
+		isClosed: false,
+	});
```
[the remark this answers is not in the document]
marjisound: Thanks for pointing it out. 
I was very unsure about what's the best way to do this. At first I was having some issues with the state changes and thought is there are 2 states that change, that would cause 2 different re-rendering of the component. But now that I tested again, I realised that was not the case. 

Changed the states to be separate. What do you think now?

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
```diff
@@ -91,27 +92,64 @@ const stickyContainerStyles = (height: number) => css`
...
+		if (isSticky) {
+			submitComponentEvent({
+				component: {
+					componentType: 'STICKY_VIDEO',
+					id: videoId,
+				},
+				action: 'STICK',
+			});
+		} else if (isClosed) {
+			submitComponentEvent({
```
joecowton1, line 122: Does this need to live in the useEffect? I think I'd imagined we'd just trigger the event when the button was clicked, maybe I'm missing something though - is there a reason to put it in here?

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
```diff
@@ -91,27 +92,64 @@ const stickyContainerStyles = (height: number) => css`
...
+				component: {
+					componentType: 'STICKY_VIDEO',
+					id: videoId,
+				},
+				action: 'RETURN',
+			});
+		}
+	}, [isSticky, isClosed, videoId]);
+
+	if (!videoId) return null;
```
joecowton1, line 140: Would this be better off at the top of the useEffect?

joecowton1:
One comment, otherwise it looks great! Nice work 👌

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
marjisound: I was at first trying to put it above the useEffect but I git lint error 🤔 
```
React Hook "useEffect" is called conditionally. React Hooks must be called in the exact same order in every component render
```
image

joecowton1: Ah i meant inside the the useEffect!

```javascript
useEffect(() => {
  if(!videoId) return;
  [...]
}

```

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
```diff
@@ -91,27 +92,64 @@ const stickyContainerStyles = (height: number) => css`
...
+		} else {
+			submitComponentEvent({
+				component: {
+					componentType: 'STICKY_VIDEO',
+					id: videoId,
+				},
+				action: 'RETURN',
+			});
+		}
+	}, [isSticky, isClosed, videoId]);
```
jamesgorrie, line 138: Does the `videoId` change in the component? Looking at the code, it's set once, and never changes.

jamesgorrie (commented):
Tiny comment, nothing that should block this.

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
joecowton1: I think eslint would complain if we removed it from the dependencies array?

oliverlloyd: Yeah, it's static but the linter asks for it. I think it's also nice to have it listed in the dependency array in terms of declaring what is being used but maybe this is just me looking for silver linings!

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
oliverlloyd: +1 to @joecowton1 's idea of putting it in the `useEffect` - I think the behaviour is the same either way but this feels closer to the intentions.

But.. is there maybe a way to make this prop required and to not even render the _component_ if it isn't. Then you wouldn't need to have this guard at all?

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
```diff
@@ -91,27 +92,64 @@ const stickyContainerStyles = (height: number) => css`
 interface Props {
 	isPlaying: boolean;
 	children: React.ReactNode;
+	videoId: string;
 }
 
-export const StickyVideo = ({ isPlaying, children }: Props) => {
+export const StickyVideo = ({ isPlaying, children, videoId }: Props) => {
 	const [isSticky, setIsSticky] = useState(false);
+	const [isClosed, setIsClosed] = useState(false);
```
oliverlloyd, line 100: Nitpic but maybe `wasClosed` makes more sense here?

dotcom-rendering/tsconfig.json
```diff
@@ -63,10 +63,11 @@
         // "inlineSources": true,                 /* Emit the source alongside the sourcemaps within a single file; requires '--inlineSourceMap' or '--sourceMap' to be set. */
         "forceConsistentCasingInFileNames": true,
         "resolveJsonModule": true,
-        "preserveConstEnums": true
+        "preserveConstEnums": true,
         /* Experimental Options */
         // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
         // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
+        "skipLibCheck": true, /* temprarily disable lib checks until guardian/libs type error is fixed */
```
oliverlloyd, line 70: Is it worth waiting for [this PR](https://github.com/guardian/libs/pull/319) to merge so we never need to add this

oliverlloyd (approved):
👍 

I've listed a few things but nothing major and at this stage I'm happy for you to manage and make decisions on these areas and decide when to merge yourself. Great work!

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
marjisound: Thanks both for your comments. I removed this check completely as the `videoId` is a required prop. But 1 question I have is that, although the prop is required but the videoId could still end up as an empty string, right?

dotcom-rendering/tsconfig.json
oliverlloyd: Merged!

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
oliverlloyd: It could be an empty string, yes. You could check for this in the parent component but is this possibility, er, possible? Like, are we expecting this to happen?

joecowton1:
Discussed with @marjisound how we handle `sticky` events if the user has scrolled over the video container: in the current setup this would fire `return` and `sticky` events in quick succession, something which we imagine is unhelpful from a data perspective. We're going to discuss this with D&I and consider only firing `sticky` events the first time a video sticks, or if it sticks after a `close` event.

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
jamesgorrie: Ah, it's the exhaustive deps. Definitely has a lot of gain, so nice to have it. Thanks for explaining!

oliverlloyd (commented):
This basically looks fine but on a hunch I wanted to see if there was an even simpler solution -  does my idea make sense? Very happy to chat it though!

dotcom-rendering/tsconfig.json
```diff
@@ -63,7 +63,7 @@
         // "inlineSources": true,                 /* Emit the source alongside the sourcemaps within a single file; requires '--inlineSourceMap' or '--sourceMap' to be set. */
         "forceConsistentCasingInFileNames": true,
         "resolveJsonModule": true,
-        "preserveConstEnums": true
+        "preserveConstEnums": true,
```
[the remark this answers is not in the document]
marjisound: good spot, Done

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
oliverlloyd: Perfect 👍

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
```diff
@@ -104,14 +108,38 @@ export const StickyVideo = ({ isPlaying, children }: Props) => {
 		setIsSticky(isPlaying && !isIntersecting);
 	}, [isIntersecting, isPlaying]);
 
+	useEffect(() => {
+		if (isSticky && wasClosed) {
+			setWasClosed(false);
```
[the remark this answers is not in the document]
oliverlloyd: > just because I like the arrow function more

This is a perfectly fine and valid reason to use arrow functions! I guess I have the habit to use the older pattern for small, locally defined things but it's not for any good reason!

oliverlloyd (approved):
Perfection 💎

dotcom-rendering/src/web/components/StickyVideo.importable.tsx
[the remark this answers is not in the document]
oliverlloyd: 👍

PR #4147, opened 2022-03-03
Simplify `getZIndex`

mxdvl:
## What does this change?

Refactor and simplify `getZIndex`.

Use the `test.each` pattern in Jest, to get more precise errors when something fails.

## Why?

Less code, no array mutation & pure maths!

### Before

before

### After

fater

### DIFF
--- a/dotcom-rendering/src/web/lib/getZIndex.tsx
+++ b/dotcom-rendering/src/web/lib/getZIndex.tsx
@@ -69,16 +69,10 @@ const indices = [
 // Implementation code - you don't need to change this to get a new index
 export type ZIndex = typeof indices[number];
 
-function reverseArray(array: any[]) {
-	return array.map((item, index) => array[array.length - 1 - index]);
-}
-
-const decideIndex = (name: string): number | null => {
-	let decided;
-	reverseArray(indices).forEach((item, index) => {
-		if (item === name) decided = index + 1;
-	});
-	return decided || null;
+const decideIndex = (name: ZIndex): number | null => {
+	const index = indices.indexOf(name);
+	if (index === -1) return null; // indexOf returns -1 if there is no match
+	return indices.length - index; // reverse the indices: last item gets 1
 };
 
 export const getZIndex = (zIndex: ZIndex): string =>
--- a/dotcom-rendering/src/web/lib/getZIndex.tsx
+++ b/dotcom-rendering/src/web/lib/getZIndex.tsx
@@ -64,11 +64,10 @@ const indices = [
 
 	// Main media
 	'mainMedia',
-];
+] as const;
 
-//
-// Implementation code  - you don't need to change this to get a new index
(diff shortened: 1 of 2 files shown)


dotcom-rendering/src/web/lib/getZIndex.tsx
```diff
@@ -64,22 +64,15 @@ const indices = [
 
 	// Main media
 	'mainMedia',
-];
+] as const;
```
mxdvl, line 67: this is where the magic happens!

dotcom-rendering/src/web/lib/getZIndex.test.tsx
```diff
@@ -1,27 +1,30 @@
...
+		['TheGuardian', 7],
+		['headerWrapper', 6],
+		['articleHeadline', 5],
+		['immersiveBlackBox', 4],
+		['bodyArea', 3],
+		['rightColumnArea', 2],
+		['mainMedia', 1],
+	];
+	it.each(tests)('Get correct index for %s : %s', (name, index) => {
+		expect(getZIndex(name)).toBe(`z-index: ${index};`);
```
mxdvl, line 28: [test.each](https://jestjs.io/docs/api#testeachtablename-fn-timeout) fails and succeeds in a more transparent way.

oliverlloyd:
Honestly, I think this does the opposite, I find the code more complex. For me, more verbose code is easier to reason about and shorter, more terse code is harder to understand. 🙃

SiAdcock:
>  more verbose code is easier to reason about

Do you mean the `reverseArray` function? I found it a bit baffling! Using `indexOf` seemed simpler to me 🤔 

Or do you mean the way the tests are defined?

mxdvl:
oh no—another controversial PR 😵

oliverlloyd:
> > more verbose code is easier to reason about
> 
> Do you mean the `reverseArray` function? I found it a bit baffling! Using `indexOf` seemed simpler to me 🤔
> 

Fair point. After spending some time looking at both, I do agree this `indexOf` implementation is better and is actually some really nice work 👍 

> Or do you mean the way the tests are defined?

The refactor of the tests seems to be adding complexity though. We're now using types, tuples and for each arrays when previously it was just a bunch of plain old jest tests with none of these things. Arguably, there's not a great deal of value in these tests but I remember we decided to include them as a form of documentation to help explain the purpose of the code so I think readability would be worth preserving here.

oliverlloyd:
> I’ve found that failing tests are often hard to debug, and a recurring pattern I’d noticed is bundling many checks in a single test, whereas the `test.each` approach gives you a much more verbose and precise output.

I think in this case the test fails with a readable description. I gave it a go by forcing a failure and it makes sense to me.

Screenshot 2022-03-07 at 09 36 57

oliverlloyd (approved):
💙

PR #4174, opened 2022-03-03
fix: remove `getCmp`

mxdvl:
## What does this change?

Remove `getCmp` as it’s not used.

## Why?

#3924 introduced a pattern we decided we do not really need, which is why #3925 is closed.


oliverlloyd (commented):
In principle, I like this change but as mentioned in standup this PR has some diffs from the Slots work.

PR #4175, opened 2022-03-03
Add pagination component to LiveBlogs on AR

iainjchambers-guardian:
## What does this change?
Adds the pagination component in `common-rendering` to Live layout in `apps-rendering` and switches the live blocks to the paginated list. Also fixed bug with links not redirecting to first page (a blank `href` on the link was not removing the query string params).

| **Before** | **After** |
|-----------|-----------|
|image | image |
| image | image |

### DIFF
--- a/apps-rendering/src/components/layout/live.tsx
+++ b/apps-rendering/src/components/layout/live.tsx
@@ -4,6 +4,7 @@ import type { SerializedStyles } from '@emotion/react';
 import { css } from '@emotion/react';
 import type { KeyEvent } from '@guardian/common-rendering/src/components/keyEvents';
 import KeyEvents from '@guardian/common-rendering/src/components/keyEvents';
+import { Pagination } from '@guardian/common-rendering/src/components/Pagination';
 import type { ArticleFormat } from '@guardian/libs';
 import { ArticleDesign } from '@guardian/libs';
 import { from, neutral, news, remSpace } from '@guardian/source-foundations';
@@ -17,6 +18,7 @@ import RelatedContent from 'components/relatedContent';
 import Tags from 'components/tags';
 import HeaderMedia from 'headerMedia';
 import type { DeadBlog, LiveBlog } from 'item';
+import { toNullable } from 'lib';
 import type { LiveBlock } from 'liveBlock';
 import type { FC } from 'react';
 import { articleWidthStyles, onwardStyles } from 'styles';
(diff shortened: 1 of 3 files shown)


apps-rendering/src/pagination.ts
```diff
@@ -98,20 +98,19 @@ const getNewerPage = (
...
 };
 
-const getNewestPage = (pageNumber: number): Option<string> => {
+const getNewestPage = (
+	pages: LiveBlock[][],
+	pageNumber: number,
+): Option<string> => {
 	if (pageNumber > 1) {
-		return some('');
+		return maybePageRef(0, pages);
```
DavidLawes, line 113: Did `some('')` not quite work when testing? Would be curious to know! I think we chose this because, on dcr, when navigating to the newest page we don't need to specify a blockId, which I think results in a more simple request flow.

apps-rendering/src/components/layout/live.tsx
```diff
@@ -102,42 +104,60 @@ interface Props {
...
+						/>
+					</div>
+				</GridItem>
+				<GridItem area="main-media">
+					<HeaderMedia item={item} />
+				</GridItem>
+				<GridItem area="live-blocks">
+					{item.pagedBlocks.currentPage.pageNumber > 1 && pagination}
+					<LiveBlocks
+						blocks={item.pagedBlocks.currentPage.blocks}
```
DavidLawes, line 143: 🎉

apps-rendering/src/components/layout/live.tsx
```diff
@@ -102,42 +104,60 @@ interface Props {
...
-					<Metadata item={item} />
-				</div>
-			</GridItem>
-			<GridItem area="key-events">
-				<div css={keyEventsWrapperStyles}>
-					<KeyEvents
-						keyEvents={keyEvents(item.blocks)}
+const Live: FC<Props> = ({ item }) => {
+	const pagination = (
+		<Pagination
```
DavidLawes, line 109: Looks great! Hope the integration with the shared `Pagination` component wasn't too tricky?

apps-rendering/src/pagination.ts
MarSavar: `some('')` was causing issues when testing locally because it wasn't actually deleting the the query parameter `page` from the URL. :(

DavidLawes: It might be worth looking into this (if there's time!), in this PR we added logic into MAPI for pagination requests: https://github.com/guardian/mobile-apps-api/pull/1785/files. With the query param included in the request we perform some extra logic that isn't necessarily required for the first page. If we could avoid that, I think that'd be good (but not essential). Happy to pair or discuss if I've not been clear :)

JamieB-gu: Interesting, thanks David, we should look into that!

Regarding the `some('')`. Passing an empty string into the pagination component results in an anchor tag with `href=""`. An `href=""` acts as a link to the current page/reloads the current page, rather than deleting the query params and resetting us back to the path-only URL (i.e. the first page).

We suspect the reason this hasn't cropped up on DCR is that it's also appending a `filterKeyEvents` parameter, so it's always setting that string to some variation of `?filterKeyEvents=false` for the first page, rather than using an empty string.

It's likely that AR will have key event filtering at some point (@alinaboghiu ?), so the same thing will apply here. We can decide then whether to remove this solution or keep both params.

**Note:** Having these extra params after navigation back to the first page, for both platforms, likely splits the cache compared to initial hits on the first page (when there are no query params).

apps-rendering/src/components/layout/live.tsx
JamieB-gu: It lined up perfectly 😄 - you and @marjisound did a brilliant job!

JamieB-gu (approved):

PR #4183, opened 2022-03-04
Use the same relative time format everywhere

### LINKED ISSUE #4157: Key events relative time format is different on the server
When we server render the relative time for a key event we're using `long` format but on the client this is being set to `med`. So 'November' is sent down from the server but the client changes this to 'Nov', causing a flash of November.

oliverlloyd:
## What does this change?
Fixes an issue where we were using `verbose` format on the server but not on the client

## Why?
Because it was causing the content to flash with the wrong value on page load

### After

Closes #4157

### DIFF
--- a/common-rendering/src/components/keyEvents.tsx
+++ b/common-rendering/src/components/keyEvents.tsx
@@ -1,7 +1,7 @@
 // ----- Imports ----- //
 
-import { css } from "@emotion/react";
-import type { SerializedStyles } from "@emotion/react";
+import { css } from '@emotion/react';
+import type { SerializedStyles } from '@emotion/react';
 import {
 	textSans,
 	headline,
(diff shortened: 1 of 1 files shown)


mxdvl (approved):
Nice catch!

Do we not have enforced linting rules for `common-rendering`?

oliverlloyd:
> Nice catch!

Well, I shouldn't really take the credit - all these issues are from the liveblog party. But okay, I'll take the credit 😊 
> 
> Do we not have enforced linting rules for `common-rendering`?

It's not so much that common doesn't have its own rules, it's more that AR & DCR have _different_ rules so whenever we touch the same files they get reformatted.

JamieB-gu:
> It's not so much that common doesn't have its own rules, it's more that AR & DCR have _different_ rules so whenever we touch the same files they get reformatted.

Could you elaborate on this? Why would DCR and AR having different rules affect `common-rendering` 🤔? I'm curious because I'm having a bit of a think about the best way to get the Guardian rules into `common-rendering` atm, because:

> Do we not have enforced linting rules for `common-rendering`?

There is now a `.prettierrc` file in `common-rendering` that specifies use of the Guardian prettier config. I think it would be good to upgrade this to a validate/CI check to make sure it's consistent in all environments. Also, depending on when you branched off and how your editor is set up Oliver, it may be this that's now bringing `common-rendering` into line with our standards on this PR?

oliverlloyd:
> A bit of an aside, but ultimately I'd hope that AR and DCR would mostly have the same rules in any case, because these days we have a centralised Guardian config for lint/prettier rules

💯 the joy of prettier is not having to think about these things anymore.

Really not sure why this happens but it seems like something is out of sync somewhere as I often see PRs that touch common files having these kinds of diffs

PR #4187, opened 2022-03-04
Pass through gu headers when fetching content

DavidLawes:
## What does this change?
For CI we may want to override the edition (e.g. force a specific edition of an article to load regardless of server location). We previously couldn't do this because only the url was being extracted from the request. This change extracts and forwards GU headers.

## Why?
We were testing click events in labs articles (https://github.com/guardian/dotcom-rendering/pull/4098) and needed to guarantee a specific article edition. When testing locally the cypress tests passed (edition = UK, so tests passed), but on CI they failed (edition != UK, tests failed). We couldn't find a way to force CI to load articles for a given edition.

To force articles to render for a specific edition you can set a header:
`'x-gu-geolocation': 'ip.dummytext,country:UK'` 

However, this had no effect in CI because only the URL is extracted from the request.

### Before
On `cy.visit(url, { headers :  { 'x-gu-geolocation': 'ip.dummytext,country:UK' } })` the article edition defaults to the server geo location (header is ignored).

### After
On `cy.visit(url, { headers :  { 'x-gu-geolocation': 'ip.dummytext,country:UK' } })` the article edition defaults to the value in the header.

### DIFF
--- a/dotcom-rendering/src/server/lib/get-content-from-url.js
+++ b/dotcom-rendering/src/server/lib/get-content-from-url.js
@@ -1,6 +1,6 @@
 const fetch = require('node-fetch');
 
-async function getContentFromURL(_url) {
+async function getContentFromURL(_url, _headers) {
 	try {
 		if (!_url) {
 			throw new Error('The url query parameter is mandatory');
@@ -14,9 +14,20 @@ async function getContentFromURL(_url) {
 		// Reconstruct the parsed url adding .json?dcr which we need to force dcr to return json
 		const jsonUrl = `${url.origin}${url.pathname}.json?dcr=true&${searchparams}`;
 
-		const { html, ...config } = await fetch(jsonUrl).then((article) =>
-			article.json(),
-		);
+		// Explicitly pass through GU headers - this enables us to override properties such as region in CI
+		let guHeaders = {};
+		if (_headers) {
+			guHeaders = Object.keys(_headers)
+				.filter((key) => key.toLowerCase().startsWith('x-gu-'))
+				.reduce((result, key) => {
+					result[key] = _headers[key];
+					return result;
+				}, {});
+		}
+
+		const { html, ...config } = await fetch(jsonUrl, {
+			headers: guHeaders,
+		}).then((article) => article.json());
 
 		return config;
 	} catch (error) {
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/server/lib/get-content-from-url.js
```diff
@@ -14,9 +14,20 @@ async function getContentFromURL(_url) {
...
+		if (_headers) {
+			guHeaders = Object.keys(_headers)
+				.filter((key) => key.toLowerCase().startsWith('x-gu-'))
+				.reduce((result, key) => {
+					result[key] = _headers[key];
+					return result;
+				}, {});
+		}
+
+		const { html, ...config } = await fetch(jsonUrl, {
```
DavidLawes, line 28: At this point in time, I only saw the need to forward the headers. Maybe we'd also want to forward the cookies too? For now, this might be enough.

dotcom-rendering/src/server/lib/get-content-from-url.js
```diff
@@ -1,6 +1,6 @@
 const fetch = require('node-fetch');
 
-async function getContentFromURL(_url) {
+async function getContentFromURL(_url, _headers) {
```
DavidLawes, line 3: Possibly worth updating the function name `getContentFromURLAndHeaders`?

dotcom-rendering/src/server/lib/get-content-from-url.js
```diff
@@ -14,9 +14,20 @@ async function getContentFromURL(_url) {
...
+		// Explicitly pass through GU headers - this enables us to override properties such as region in CI
+		let guHeaders = {};
+		if (_headers) {
+			guHeaders = Object.keys(_headers)
+				.filter((key) => key.toLowerCase().startsWith('x-gu-'))
+				.reduce((result, key) => {
+					result[key] = _headers[key];
+					return result;
+				}, {});
+		}
```
mxdvl, line 26: I think `Object.fromEntries` here would be _simpler_ than a reducer!

```
const headers = Object.fromEntries(
    Object.entries(_headers)
        .filter(([key]) => key.toLowerCase().startsWith("x-gu-"))
)
```

dotcom-rendering/src/server/lib/get-content-from-url.js
mxdvl: I think it’s best to add JSDoc comments. For example:

```ts
/**
 * Get DCR content from a `theguardian.com` URL.
 * Takes in optional `X-Gu-*` headers to send.
 */
async function getContentFromURL
```

mxdvl (approved):
Nice catch, some thoughts, but not blockers

I know headers names are [case insensitive in the spec](https://fetch.spec.whatwg.org/#terminology-headers), but I’ve found more "X-GU-*" uppercase in our codebase.

PR #4197, opened 2022-03-07
`Dateline` clickability

### LINKED ISSUE #4163: Clicking article timestamp does not reveal more info
DCR

Frontend

oliverlloyd:
## What does this change?
This PR fixes #4163 an issue where the dateline was not expanding when clicked.

## Why?
We use css to expand the secondary date but this css depends on a unique id attribute being given used and because liveblogs have two `ArticleMeta` components on the page this was failing.

### `context`
The solution I implemented to allow unique ids was to have a `context` prop. This prop allows the `Dateline` component to set a different id in different contexts. Having this context available as a prop also meant I could slightly simplify some logic where we set a different border colour depending on a breakpoint (this breakpoint is also the same one used to change context).

### Refactored `Dateline` structure
Along the way, I noticed that we had duplicated divs and css so I made a small refactor to `Dateline` to remove this

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleMeta.stories.tsx
+++ b/dotcom-rendering/src/web/components/ArticleMeta.stories.tsx
@@ -72,7 +72,7 @@ export const ArticleStory = () => {
 				}}
 				tags={tagsWithBylineImage}
 				primaryDateline="Sun 12 Jan 2020 18.00 GMT"
-				secondaryDateline="Sun 12 Jan 2020 21.00 GMT"
+				secondaryDateline="Last modified on Sun 12 Jan 2020 21.00 GMT"
 				isCommentable={false}
 				discussionApiUrl=""
 				shortUrlId=""
@@ -117,7 +117,7 @@ export const BrandingStory = () => {
 				}}
 				tags={tagsWithBylineImage}
 				primaryDateline="Sun 12 Jan 2020 18.00 GMT"
-				secondaryDateline="Sun 12 Jan 2020 21.00 GMT"
+				secondaryDateline="Last modified on Sun 12 Jan 2020 21.00 GMT"
 				isCommentable={false}
 				discussionApiUrl=""
 				shortUrlId=""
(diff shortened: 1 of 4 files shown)


JamieB-gu:
Probably a bit of a tangent and not that helpful for this PR (sorry 🙈!), but would a `details`/`summary` work for this? There's probably a reason why we don't use one, but I thought I'd ask just in case.

mxdvl (approved):
LGTM.

Worth noting we have a `transparentColour` now!

oliverlloyd:
> Probably a bit of a tangent and not that helpful for this PR (sorry 🙈!), but would a `details`/`summary` work for this? There's probably a reason why we don't use one, but I thought I'd ask just in case.

That sounds like an excellent idea. Let me timebox a bit of effort on this and if I can get a win I'll update this PR

MarSavar:
> > Probably a bit of a tangent and not that helpful for this PR (sorry 🙈!), but would a `details`/`summary` work for this? There's probably a reason why we don't use one, but I thought I'd ask just in case.
> 
> That sounds like an excellent idea. Let me timebox a bit of effort on this and if I can get a win I'll update this PR

Just a heads up, support for `<details>` is lacking in IE and `<summary>` is even worse :(
image

oliverlloyd:
Is the lack of support mainly around the [default styles](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary#default_style) for the list item?

If we're only limited by IE, I don't see that as a blocker

JamieB-gu:
> Just a heads up, support for `<details>` is lacking in IE and `<summary>` is even worse :(

> If we're only limited by IE, I don't see that as a blocker

Yeah neither are supported in IE, but I think what's important is what they'll look like there. I'm _guessing_ (we should confirm) that it won't understand either element and fall back to treating them both as something basic like a `div`. All that _should_ mean is that both dates are shown at all times in IE, rather than the second being hidden until click. I'd say that's probably a reasonable fallback.

JamieB-gu (approved):
🍰

PR #4198, opened 2022-03-07
Updates KeyEvents to re-add interactivity and expand link area

### LINKED ISSUE #4161: Key event icon and timestamp are not links

AshCorr:
## What does this change?

 - Wraps the whole KeyEvent in a link so that any part of the KeyEvent can be clicked.
 - Re-adds interactivity to the KeyEven links to make them change colour on hover.

## Why

Hover animations look cool!

Closes #4161 

| Before | After |
| ------- | ----- |
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/21217225/158415023-e4fb048d-0f03-4ec8-8ee0-be92db4ec6e8.gif
[after]: https://user-images.githubusercontent.com/21217225/158414769-94f283b8-675d-43aa-bb30-8e7f65297f67.gif


oliverlloyd (commented):
It might be worth getting these design changes signed off by @HarryFischer . Ideally the link fix and the styling change would be two separate changes but I get that things end up happening together.

common-rendering/src/components/keyEvents.tsx
```diff
@@ -112,7 +90,50 @@ const listItemStyles = (supportsDarkMode: boolean): SerializedStyles => css`
...
+
+	${darkModeCss(supportsDarkMode)`
+		&:hover:before {
+			background-color: ${neutral[86]};
+		}
+
+		&:before {
+			background-color: ${neutral[46]};
+		}
+	`}
```
JamieB-gu, line 124: So in light mode it's a light grey and then gets darker on hover, and in dark mode it's a dark grey and gets lighter on hover?

common-rendering/src/components/keyEvents.tsx
```diff
@@ -89,9 +97,13 @@ const listStyles = (supportsDarkMode: boolean): SerializedStyles => css`
...
 	${darkModeCss(supportsDarkMode)`
-		li::before {
+		&:hover:before {
+			background-color: ${neutral[100]};
+		}
+
+		&:before {
 			border-color: transparent ${neutral[60]};
-			background-color: neutral[60];
+			background-color: ${neutral[60]};
```
JamieB-gu, line 106: Good spot!

oliverlloyd:
I _think_ this PR is ready to roll but we should probably update the title and description to better reflect what is actually (now) happening here

common-rendering/src/components/keyEvents.tsx
```diff
@@ -164,23 +175,23 @@ const timeStyles = (supportsDarkMode: boolean): SerializedStyles => css`
...
+			<Link
+				priority="secondary"
+				css={linkStyles(supportsDarkMode)}
+				href={keyEvent.url}
+			>
+				<div css={timeTextWrapperStyles}>
+					<time
+						dateTime={keyEvent.date.toISOString()}
+						data-relativeformat="med"
+						title={keyEvent.date.toLocaleTimeString()}
```
oliverlloyd, line 187: In [this PR](https://github.com/guardian/dotcom-rendering/pull/4254) I got a bit fancy with the title, maybe we can bring something similar over to here too?

oliverlloyd (approved):
Just a mini one around the title attribute but otherwise 👍

common-rendering/src/components/keyEvents.tsx
```diff
@@ -75,8 +75,16 @@ const keyEventWrapperStyles = (
...
+const linkStyles = (
+	supportsDarkMode: boolean
+): SerializedStyles => css`
+	text-decoration: none;
+
+	&:hover:before {
+		background-color: ${neutral[0]};
+	}
+
+	&:before {
```
mxdvl, line 87: I know this technically works, but the CSS3 syntax is the double-double-colon `::` for pseudo-elements, to distinguish them from states.

https://developer.mozilla.org/en-US/docs/Web/CSS/::before#syntax

mxdvl: Maybe we need a lint rule and it’s actually outside the scope of this PR?

PR #4201, opened 2022-03-07
Compute realtime config within AMP `Ad` component

chrislomaxjones:
## What does this change?

Currently we compute the [Real Time Config](https://github.com/ampproject/amphtml/blob/main/extensions/amp-a4a/rtc-documentation.md) for AMP `Ad` components and pass this in as a prop (either from `StickyAd` or `RegionalAd`).

In this PR, instead we pass the commercial `config` and `placementId` as props, and compute the RTC within the ad component.

## Why?

This leaves us with only one call site where `realTimeConfig` is called. This will simplify subsequent work where we want to run an AB test, with different variants computing different RTC strings.


lucymonie (approved):
LGTM 👍🏻

PR #4203, opened 2022-03-08
Subnav fixes: add vertical lines 

DavidLawes:
## What does this change?
This is the first of a number of fixes to the subnav (https://trello.com/c/KV1Jdgxk)

This change adds vertical lines between pillars at desktop. It also removes the left and right borders of the nav for immersive layouts.

I spent some time trying to add stories for the expanded sub nav but have decided against this for now. Adding stories wasn't quick or easy and I think I'd have had to do a bit of component refactoring to enable this. When discussing about component refactoring I found that there is an upcoming overhaul of the header (it is to be defined in source and consumed by all that need it). At that point in time it may be a more natural point to consider the structure of the associated components. 

For these changes I've deployed to CODE - Harry's tested and given the go-ahead.

## Why?

This brings the nav+subnav UI closer to the intended design (https://www.figma.com/file/V8u2r27TFdjAWNM3p84ypC/%E2%9C%AD-Header-%26-footer?node-id=0%3A1).

### Non-immersive display

From leftCol:

| **Before** | **After** |
|------------|-----------|
| Screenshot 2022-03-08 at 17 19 29 | Screenshot 2022-03-08 at 17 18 50 |

### Immersive display

From leftCol:

| **Before** | **After** |
|------------|-----------|
| Screenshot 2022-03-08 at 17 16 43 | Screenshot 2022-03-08 at 17 16 31 |


dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -111,9 +117,10 @@ const brandExtensionLink = css`
 `;
 
 export const Columns: React.FC<{
+	display: ArticleDisplay;
```
oliverlloyd, line 120: For consistency, can we change this to `format`? I think it's perfectly fine to pass `display` by itself in general, but we just have a convention of always passing `format`. I guess the win is it's more scalable

OllysCoding (approved):
+1

oliverlloyd (approved):
Loving the war on typos

This is really great work. Thank you for keeping this change targeted and easy to reason about 👍

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
DavidLawes: Ah, yes - thanks, I'll update :)

PR #4204, opened 2022-03-08
Refactor `MatchStats`

MarSavar:
## What does this change?
This PR refactors the `MatchStats` component so it now uses `decidePalette` to handle styling.
The diff looks atrocious, sorry, but it's a result of running `yarn prettier:fix` - you can review it with hidden whitespace [here](https://github.com/guardian/dotcom-rendering/pull/4204/files?diff=unified&w=1), which is much better.

The bulk of the PR is around changing the props of `MatchStats`, `StretchBackground`, `GoalAttempts`, `Side` to take `format` as one of the props, which allows the use of `decidePalette`. 

## Why?
This was needed because the component used to take a fixed background colour, which looked incorrect on liveblogs and deadblogs (where it needed to be transparent instead).

## Screenshots
| **Before** | **After** |
|------------|-----------|
| image | image |
| image | image |

The changes introduced by this PR do not introduce a visual difference in standard articles - only in live / deadblogs.

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -141,6 +141,7 @@ type Palette = {
 		mostViewedTab: Colour;
 		matchNav: Colour;
 		analysisUnderline: Colour;
+		matchStats: Colour;
 	};
 	fill: {
 		commentCount: Colour;
--- a/dotcom-rendering/src/web/components/GoalAttempts.stories.tsx
+++ b/dotcom-rendering/src/web/components/GoalAttempts.stories.tsx
@@ -1,4 +1,5 @@
 import { css } from '@emotion/react';
+import { ArticleDesign, ArticleDisplay, ArticlePillar } from '@guardian/libs';
 
 import { GoalAttempts } from './GoalAttempts';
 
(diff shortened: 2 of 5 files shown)


MarSavar:
@HarryFischer

HarryFischer:
@MarSavar Can we shift it across to the right by 10px(i think) so it aligns with the left hand side of the key events dot above?
Basically it should have margin-left: 20px; from mobileLandscape and above

MarSavar:
> @MarSavar Can we shift it across to the right by 10px(i think) so it aligns with the left hand side of the key events dot above? Basically it should have margin-left: 20px; from mobileLandscape and above

I was going to tweak the margin/padding/width issue in a separate PR since it looks broken on DCR anyway - at the moment it looks quite a bit different from frontend (i.e. frontend is a lot wider), see below:

| Frontend | DCR |
|----------|------|
| image | image |

I'd like to keep this PR focused on palette colours if that's ok -  should have specified it in the PR text!

HarryFischer:
@MarSavar Ok cool - look good to me then!

MarSavar:
[MatchStats background is transparent](https://trello.com/c/5iyta0wI/683-matchstats-background-is-transparent)

oliverlloyd (approved):
Oh no! I reviewed this PR the other day but forgot to submit it - sorry!

This is spot on 👍

PR #4211, opened 2022-03-09
NoJS version of Pinned Post

abeddow91:
## What does this change?
Following from [PR 4132](https://github.com/guardian/dotcom-rendering/pull/4132), this PR introduces a new pinned post component that is written using checkbox input and CSS. This allows the core functionality to work without JS. It gets enhanced with JS through the PinnedPostLiveness component similar to [Liveness](https://github.com/guardian/dotcom-rendering/pull/3869). It feels like the existing Liveness component is a natural home for the pinned post enhancements. However, the existing Liveness component is only for liveblogs and pinned posts feature on both live and dead blogs so it has been separated into a new component. 

## Why?
It allows users without JS enabled to get full use of the component. It also allows for full server side rendering and means the core component doesn't need to live in an island. Also, this pattern is preexisting see https://github.com/guardian/dotcom-rendering/pull/1399 and https://github.com/guardian/frontend/pull/24420 

Adding liveness to the component allows us to add some polish to the component that wouldn't be possible without JS. These include

- [ ] - Scroll on collapse
- [ ] - Tracking 
- [ ] - Keyboard accessibility
- [x] - Remove show more button if the content fits within the container's collapsed max-height

This PR serves to demonstrate how we can add liveness so only those checked above have been implemented in this PR, the remainder are to be completed in separate tickets.

### Before
Screenshot 2022-03-09 at 14 07 48

### After
Screenshot 2022-03-09 at 14 07 38


dotcom-rendering/src/static/icons/pin.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="60" height="105" viewBox="0 0 60 105" fill="none" >
```
abeddow91, line 1: This will be replaced by a Source Pinned SVG following release

common-rendering/src/components/FirstPublished.tsx
```diff
@@ -8,9 +8,11 @@ const padString = (time: number) => (time < 10 ? `0${time}` : time);
 const FirstPublished = ({
 	firstPublished,
 	blockLink,
+	showTimeAgo = true,
 }: {
 	firstPublished: number;
 	blockLink: string;
+	showTimeAgo?: boolean;
```
oliverlloyd, line 15: What about changing this to `isPinnedPost`? That would feel more declarative and less imperative (I love using those words in PR reviews). Essentially, it's less tell me what to do and more tell me what's happening and let me decide how to react

I'm a little 🤷  here though and you know this code better than me so if you have strong opinions then I'm happy to defer to you

common-rendering/src/components/liveBlockContainer.tsx
```diff
@@ -129,19 +131,20 @@ const LiveBlockContainer = ({
 				padding: ${space[2]}px ${SIDE_MARGIN_MOBILE}px;
 				margin-bottom: ${space[3]}px;
 				background: ${neutral[100]};
-				border-top: 1px solid ${borderColour};
+				border-top: ${isPinnedPost ? 0 : `1px solid ${borderColour}`};
+				border-bottom: ${isPinnedPost ? 0 : `1px solid ${neutral[86]}`};
```
oliverlloyd, line 135: Does it also work if we make the whole line conditional? So we only set a border if not a pinned post. Of course, this idea falls over if the `0` is needed here

dotcom-rendering/fixtures/manual/liveBlock.ts
```diff
@@ -0,0 +1,24 @@
+export const baseBlock: Block = {
```
oliverlloyd, line 1:
```suggestion
export const liveBlock: Block = {
```

It's always nice if the name of the file matches the name of the export. It makes searching and refactoring easier later on.

dotcom-rendering/index.d.ts
```diff
@@ -434,6 +434,7 @@ interface CAPIType {
 	main: string;
 	keyEvents: Block[];
 	blocks: Block[];
+	pinnedPost?: Block;
```
oliverlloyd, line 437: I think you need to run 'make gen-schema' following any edits to this type

dotcom-rendering/src/web/components/PinnedPost.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+						display: ArticleDisplay.Standard,
+					}}
+					block={block}
+					pageId=""
+					webTitle=""
+					ajaxUrl=""
+					isPinnedPost={true}
+				/>
+			</PinnedPost>
+		</Wrapper>
```
oliverlloyd, line 83: 💯

dotcom-rendering/src/web/components/PinnedPost.tsx
```diff
@@ -0,0 +1,175 @@
...
+	space,
+	textSans,
+	news,
+	from,
+	visuallyHidden,
+	transitions,
+	focusHalo,
+	height,
+} from '@guardian/source-foundations';
+import { Props, SvgMinus, SvgPlus } from '@guardian/source-react-components';
```
oliverlloyd, line 13: What are these `Props` we're importing here? Normally when a lib exports some props they are explicitly named so you know what you're getting so it's strange to see an export called `Props`. I wonder if this was intentional...?

oliverlloyd: @SiAdcock @sndrs - is Source meant to be exporting this?

dotcom-rendering/src/web/components/PinnedPost.tsx
```diff
@@ -0,0 +1,175 @@
...
+		display: block;
+		fill: white;
+		position: relative;
+		width: 24px;
+		height: auto;
+		margin-left: -${space[1]}px;
+	}
+`;
+
+export interface PinnedPostProps extends Props {
```
oliverlloyd, line 129: I _think_ this can just be `type Props = { ... }`. We're not importing it so there's no need for the export and I don't think we need to extend that other type (but I could be wrong on that)

dotcom-rendering/src/web/components/PinnedPost.tsx
```diff
@@ -0,0 +1,175 @@
...
+			</div>
+			<div id="collapsible-body" css={collapsibleBody}>
+				{children}
+			</div>
+			<div id="overlay" css={overlay} />
+			<label
+				css={button}
+				htmlFor="pinned-post-checkbox"
+				id="pinned-post-button"
+			>
```
oliverlloyd, line 163: Would it be possible to use [EditorialButton](https://guardian.github.io/source/?path=/docs/packages-source-react-components-development-kitchen-editorialbutton--playground) from Source here?

It's generally going to be more accessible and it also has the bonus of changing colour based on `theme`.

If you need to add specific attributes I am pretty sure this component extends the standard html button interface so should allow that plus you can also use `cssOverrides` to tweak styles if needed

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
```diff
@@ -0,0 +1,42 @@
...
+ * hide show more button on pinned post
+ */
+function hideShowMore() {
+	const pinnedPostButton = document.querySelector<HTMLElement>(
+		'#pinned-post-button',
+	);
+	if (pinnedPostButton) pinnedPostButton.style.display = 'none';
+}
+
+export const PinnedPostLiveness = ({}) => {
```
oliverlloyd, line 23: Two things:

1) You can delete the empty object and just have `()`. I don't know why you should do this but some linters require this so my brain has been conditioned to call it out
2) How about `EnhancePinnedPost`? Or something along those lines? Just because 'Liveness' in DCR's context signifies polling for updates but this island isn't really that

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
```diff
@@ -0,0 +1,42 @@
...
+	const contentFitsContainer =
+		pinnedPost && pinnedPost.scrollHeight <= pinnedPost.clientHeight;
+	if (contentFitsContainer) hideShowMore();
```
oliverlloyd, line 26: I think this is fine but it made me wonder if we might ever have an issue where the dom isn't ready but we try to perform operations against it?

In react you get around this using a `useEffect` which only runs after things have rendered. But in our islands world that won't work. In pure js you use the [DOMContentLoaded](https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event) event but maybe that's excessive here?

Of course, none of this is relevant to your PR here and is more of a question around how we carry this pattern forward.

dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
```diff
@@ -23,6 +26,21 @@ export const LiveBlogRenderer = ({
 }: Props) => {
 	return (
 		<>
+			{pinnedPost && (
```
oliverlloyd, line 29: Where you've put the `Island` above does feel like a good location but it could also live here alongside the post itself?

dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
```diff
@@ -34,6 +52,7 @@ export const LiveBlogRenderer = ({
 						host={host}
 						ajaxUrl={ajaxUrl}
 						isLiveUpdate={isLiveUpdate}
+						isPinnedPost={true}
```
oliverlloyd, line 55: Should this read `isPinnedPost={false}`?

oliverlloyd (commented):
This is really great work. The description of this PR is so good and it really was a pleasure to review this PR, thanks for the effort you clearly put into this.

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
oliverlloyd: Rereading this comment this is more a note to self for me and you can ignore this

dotcom-rendering/src/static/icons/pin.svg
```diff
@@ -0,0 +1,3 @@
...
+<path d="M60 56.9C56.5 49.6 50.4 43.7 42.9 40.5L40.1 7.9C45.1 6.3 49.5 3.6 53.2 0H39.5H20.5H6.8C10.4 3.6 14.9 6.3 19.9 7.9L17.2 40.5C9.6 43.7 3.5 49.6 0 56.9H15.8H25.3V85.4L30 104.4L34.7 85.4V56.9H44.2H60Z" />
+</svg>
```
mxdvl, line 3: Now https://github.com/guardian/source/pull/1292 is merged, can we remove this file?

dotcom-rendering/src/web/components/PinnedPost.tsx
```diff
@@ -0,0 +1,175 @@
...
+	height,
+} from '@guardian/source-foundations';
+import { Props, SvgMinus, SvgPlus } from '@guardian/source-react-components';
+import { timeAgo } from '@guardian/libs';
+import { css } from '@emotion/react';
+
+// TODO: replace with source pinned icon once this has been merged and released.
+// See PR here: https://github.com/guardian/source/pull/1292
+// See Ticket here: https://trello.com/c/xZr6SP9H/272-pinned-post-replace-local-pinned-svg-with-source
+import PinIcon from '../../static/icons/pin.svg';
```
mxdvl, line 20: Isn’t this already done?

alinaboghiu (approved):

common-rendering/src/components/liveBlockContainer.tsx
```diff
@@ -124,24 +126,25 @@ const LiveBlockContainer = ({
...
 				}
-				border-bottom: 1px solid ${neutral[86]};
 			`}
 		>
 			<Header>
 				{blockFirstPublished && (
 					<FirstPublished
 						firstPublished={blockFirstPublished}
 						blockLink={blockLink}
+						isPinnedPost={isPinnedPost ? false : true}
```
oliverlloyd, line 147: I think this is reversing the logic here. Does it work if we do this?
```suggestion
						isPinnedPost={isPinnedPost}
```

dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
```diff
@@ -76,6 +97,7 @@ export const LiveBlogRenderer = ({
 							switches={switches}
 							isAdFreeUser={isAdFreeUser}
 							isSensitive={isSensitive}
+							isPinnedPost={true}
```
oliverlloyd, line 100: This is the non pinned version so I believe it should be false

```suggestion
							isPinnedPost={false}
```

dotcom-rendering/fixtures/manual/liveBlock.ts
abeddow91: Absolutely, renamed!

dotcom-rendering/src/web/components/PinnedPost.tsx
abeddow91: You're totally right, I've update this.

dotcom-rendering/src/web/components/PinnedPost.tsx
abeddow91: Think  this might have been an accident with VSC auto imports. I've removed in favour of local Props

dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
abeddow91: Yep you're right - amended.

dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
abeddow91: Good call, I've moved the island within the pinnedpost conditional check so its next to the pinned post.

dotcom-rendering/src/web/components/PinnedPost.tsx
```diff
@@ -0,0 +1,173 @@
...
+		content: 'Show more';
+	}
+	#pinned-post-checkbox:checked ~ label::after {
+		content: 'Show less';
+	}
+`;
+
+const pinnedPostRow = css`
+	background: ${news[300]};
+	height: 2rem;
```
oliverlloyd, line 49: Could we change these rems to pixels?

I realise it's somewhat arbitrary and that there are pros and cons to both approaches but in DCR the convention is pixels so it's really just for consistency.

dotcom-rendering/src/web/components/PinnedPost.tsx
```diff
@@ -0,0 +1,173 @@
...
+	#pinned-post-checkbox ~ label::after {
+		content: 'Show more';
+	}
+	#pinned-post-checkbox:checked ~ label::after {
+		content: 'Show less';
+	}
+`;
+
+const pinnedPostRow = css`
+	background: ${news[300]};
```
oliverlloyd, line 48: This will need to use `decidePalette` so that this colour works over different `theme` (Sport blogs for example)

dotcom-rendering/src/web/components/PinnedPost.tsx
```diff
@@ -0,0 +1,173 @@
...
+				{pinnedPost.blockFirstPublished && (
+					<time data-relativeformat="med" css={timeAgoText}>
+						From {timeAgo(pinnedPost.blockFirstPublished)}
+					</time>
+				)}
+			</div>
+			<div id="collapsible-body" css={collapsibleBody}>
+				{children}
+			</div>
+			<div id="overlay" css={overlay} />
```
oliverlloyd, line 156: It might be good to have a more specific name for these ids so that we don't get clashes elsewhere (the css names are fine, Emotion handles this for us)

common-rendering/src/components/FirstPublished.tsx
```diff
@@ -30,25 +32,27 @@ const FirstPublished = ({
...
+					css={css`
+						color: ${neutral[46]};
+						font-weight: bold;
+						margin-right: ${space[2]}px;
+					`}
+				>
+					{timeAgo(firstPublished)}
+				</time>
+			)}
 			<span
```
oliverlloyd, line 48: Not really sure where this change needs to happen (I'm guessing here but could well be wrong!) but it seems there's too much padding in front of the From time
Screenshot 2022-03-17 at 16 27 04

dotcom-rendering/src/web/components/PinnedPost.tsx
```diff
@@ -0,0 +1,173 @@
...
+	${from.tablet} {
+		margin-left: 3.75rem;
+	}
+`;
+
+const collapsibleBody = css`
+	max-height: 40vh;
+	overflow: hidden;
+`;
+
```
oliverlloyd, line 114: I'm just dropping this comment here because I don't know where else it should go but do we need a bit more space above the button here?
Screenshot 2022-03-17 at 16 26 26

oliverlloyd (commented):
This is really starting to look great! Sorry for all the comments. They're all small things though, so hopefully not too much of a drag and if you have any questions or just want to spread the load I'm very happy to pair on this

dotcom-rendering/src/web/components/PinnedPost.tsx
sndrs: i _think_ that's left in from v3 because people had found it deep in the files in the package and had used it for their own components, iirc @SiAdcock? agree it should have a better name, if it should even be public...

common-rendering/src/components/FirstPublished.tsx
```diff
@@ -1,24 +1,26 @@
-import { css } from "@emotion/react";
-import { timeAgo } from "@guardian/libs";
-import { neutral, space, textSans } from "@guardian/source-foundations";
+import { css } from '@emotion/react';
+import { timeAgo } from '@guardian/libs';
+import { neutral, space, textSans } from '@guardian/source-foundations';
```
OllysCoding, line 3: This speaks to the remaining wider problem than anything in common-rendering is not being linted and is subject to the rules of whoever last edited it.

dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const isServer = typeof window === 'undefined';
+
+const pinnedPost: Element | null = !isServer
+	? window.document.querySelector('[data-gu-marker=pinned-post]')
+	: null;
+
+const pinnedPostCheckBox: Element | null = !isServer
+	? window.document.querySelector('input[name=pinned-post-checkbox]')
+	: null;
```
OllysCoding, line 11: Should we be using `undefined` here instead of `null`? cc @mxdvl (I think undefined is generally preferred but I forget why)

oliverlloyd: I think it needs to be `null` as that is what `querySelector` returns when nothing is found

dotcom-rendering/src/web/components/PinnedPost.tsx
oliverlloyd: This is done

dotcom-rendering/src/web/components/PinnedPost.tsx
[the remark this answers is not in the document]
oliverlloyd: This is done

dotcom-rendering/src/web/components/PinnedPost.tsx
oliverlloyd: This is done

common-rendering/src/components/FirstPublished.tsx
oliverlloyd: This is done

dotcom-rendering/src/web/components/PinnedPost.tsx
oliverlloyd: So I think I'm wrong here and that we can't use a classic button component in this situation where we want the css expansion trick

common-rendering/src/components/FirstPublished.tsx
```diff
@@ -30,25 +32,27 @@ const FirstPublished = ({
...
+			)}
 			<span
 				css={css`
 					${textSans.xxsmall()};
 					color: ${neutral[46]};
 				`}
 			>
 				{`${padString(publishedDate.getHours())}:${padString(
-					publishedDate.getMinutes()
+					publishedDate.getMinutes(),
```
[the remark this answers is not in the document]
oliverlloyd: It was prettier so it stays!

common-rendering/src/components/FirstPublished.tsx
oliverlloyd: The Guardian configs repo [recommends single quotes](https://github.com/guardian/configs/blob/main/packages/prettier/index.js#L9)

dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
OllysCoding:
```ts
const pinnedPost: Element | undefined = !isServer && window.document.querySelector('[data-gu-marker=pinned-post]') || undefined;
```
Could work?

oliverlloyd: Why is `undefined` better than `null`? Isn't it that a variable is undefined when it's not been, you know, defined and then in this case one of the possible outcomes is that it becomes `null` which is the value you get when document queries can't find anything.

There used to be a similar thing with `isSignedIn` where `undefined` meant that the check had not completed yet but `false` meant it had completed but the user wasn't signed in. Obviously in this case though the definition is inline so there's never an `undefined` state.

dotcom-rendering/src/web/components/PinnedPost.tsx
oliverlloyd: Done

dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
OllysCoding: I don't think it's a massive deal, though like with everything, there are people with stronger opinions who lean towards or away from using one or the other. 

There were some discussions in Client side around it recently and my interpretation had been that 'undefined' was the generally preferred practice. 

I don't see any material difference between the code that's there and the suggestion I wrote besides whether we have a preferred method as a team/codebase, so it's there if we do!

common-rendering/src/components/FirstPublished.tsx
OllysCoding: Mostly concerned that it was wrong in the first place!

dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
oliverlloyd: To be honest, I find these kinds of chained ternaries like `const foo = !isApple && isGreen || canBeEaten` hurts my brains 😵‍💫

oliverlloyd: Re the null / undefined wars. I'm honestly pretty clueless what it's all about but we do have several examples of this same pattern in the code (I think this code is based on the `Liveness` file) so my preference is to keep things consistent and maybe revisit this separately?

common-rendering/src/components/FirstPublished.tsx
oliverlloyd: Yeah, I see this quote war crop up pretty often and it seems like unnecessary diff noise. I guess the root issue is common not having a config

dotcom-rendering/src/web/components/PinnedPost.tsx
sndrs: deprecated it here https://github.com/guardian/source/pull/1310

dotcom-rendering/src/web/components/PinnedPost.tsx
abeddow91: Yeah, I've been having a look at this this morning and I can't seem to get it working with EditorialButton. If there's a strong desire to use it still, we could ticket this separately so it doesn't block this PR?

oliverlloyd: No, it's fine. I think that your approach is the right one here 👍

dotcom-rendering/cypress/integration/parallel-3/liveblog.interactivity.spec.js
```diff
@@ -201,7 +201,7 @@ describe('Liveblogs', function () {
 			cy.get('#46d194c9-ea50-4cd5-af8b-a51e8b15c65e').should(
 				'not.be.visible',
 			);
-			cy.get('[data-gu-marker=top-of-blog]').scrollIntoView();
+			cy.get('[id=top-of-blog]').scrollIntoView();
```
oliverlloyd, line 204: Using an id here s we can use this element as an anchor tag

dotcom-rendering/src/model/json-schema.json
```diff
@@ -390,6 +393,9 @@
         },
         "filterKeyEvents": {
             "type": "boolean"
+        },
+        "mostRecentBlockId": {
+            "type": "string"
```
oliverlloyd, line 398: This and some other properties crept in here as the schema update routine was not running before

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -14,10 +14,19 @@ type Props = {
...
 const isServer = typeof window === 'undefined';
 
+const topOfBlog: Element | null = !isServer
+	? window.document.getElementById('top-of-blog')
+	: null;
+
+const lastUpdated: Element | null = !isServer
+	? window.document.querySelector('[data-gu-marker=liveblog-last-updated]')
+	: null;
+
```
oliverlloyd, line 29: I moved these up so they were defined earlier

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -45,9 +54,9 @@ function insert(html: string, switches: Switches) {
 	// We're being sent this string by our own backend, not reader input, so we
 	// trust that the tags and attributes it contains are safe and intentional
 	const blogBody = document.querySelector<HTMLElement>('#liveblog-body');
-	const latestBlock = blogBody?.querySelector('#liveblog-body :first-child');
-	if (!latestBlock || !blogBody) return;
-	blogBody.insertBefore(fragment, latestBlock);
+	if (!blogBody || !topOfBlog) return;
+	// nextSibling? See: https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore#example_2
+	blogBody.insertBefore(fragment, topOfBlog.nextSibling);
```
oliverlloyd, line 59: We now insert new blocks immediately after the top-of-blog marker. This marker lives below the pinned post, preventing ordering issues

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -156,7 +158,11 @@ export const Liveness = ({
 				updateTimeElement(lastUpdated);
 			}
 
-			if (onFirstPage && topOfBlogVisible && document.hasFocus()) {
+			if (
+				onFirstPage &&
+				topOfBlogVisible &&
+				document.visibilityState === 'visible'
```
oliverlloyd, line 164: In reviewing this PR I noticed we can improve the way we detect visibility

dotcom-rendering/cypress/integration/parallel-3/liveblog.interactivity.spec.js
jamesgorrie: Not sure if it makes sense as I am uncertain of what `data-gu-marker` is, but if it's a contract between us and others, maybe we could have both `id` and `data-gu-marker` that serve different purposes?

dotcom-rendering/src/model/json-schema.json
jamesgorrie: Should this be automatic?

yarn.lock
```diff
@@ -20844,14 +20826,6 @@ worker-rpc@^0.1.0:
...
     microevent.ts "~0.1.1"
 
-wrap-ansi@^3.0.1:
-  version "3.0.1"
-  resolved "https://registry.yarnpkg.com/wrap-ansi/-/wrap-ansi-3.0.1.tgz#288a04d87eda5c286e060dfe8f135ce8d007f8ba"
-  integrity sha1-KIoE2H7aXChuBg3+jxNc6NAH+Lo=
-  dependencies:
-    string-width "^2.1.1"
-    strip-ansi "^4.0.0"
-
```
jamesgorrie, line 20854: What's the reason for all the `lock` changes?

jamesgorrie (commented):
Code looks good, I have a question on the new pattern of adding JS adjacent to the component it's controlling.

dotcom-rendering/cypress/integration/parallel-3/liveblog.interactivity.spec.js
oliverlloyd: The data attribute is an attempt to move away from using `id`s everywhere but in this case, where we explicitly want an anchor tag, it made more sense to go with that.

I think the data- pattern is emerging. It's nice because you have a more space for description, you can specify both a key and a value, but I don't think there are any hard contracts yet. One contract we _do_ have is to only use classnames for styling. So maybe data attributes can be thought of as alternatives for where classnames used to be used as selectors.

yarn.lock
oliverlloyd: I wish I understood this better. All I know is I'm busily going about my business - not making package changes - and then suddenly I run `yarn` and I get this. So I guess either I've merged in `main` and got some weirdness where the lock wasn't updated from another branch or this is somehow date related?

jamesgorrie (approved):

PR #4214, opened 2022-03-09
Remove Emotion Server From AR

JamieB-gu:
## Why?

As part of the Emotion 11 release, `emotion-server` was renamed to `@emotion/server`, and `create-emotion-server` was pulled into the same package. We're already using those new packages via the repo's top-level dependency tree, so we can remove our dependence on these older versions. This should also remove the need for #4209.

## Changes

- Remove `emotion-server` dependency
- Remove `create-emotion-server` dependency


OllysCoding (approved):
+1

PR #4216, opened 2022-03-10
Show `ContributionsLiveblogEpic` in blogs

### LINKED ISSUE #4180: Support the Guardian
image

The frontend code used to show this slot is [here](https://github.com/guardian/frontend/blob/99bd1cb0ed72eaff2155a78a7f8ee3c844a5e699/static/src/javascripts/projects/common/modules/support/epic.ts).

oliverlloyd:
## What does this change?
This PR closes #4180 by adding the logic to render the `ContributionsLiveblogEpic` in blogs

## Why?
For parity with blogs on Frontend

## Differences to how other epics are shown
The code in this PR follows a lot of the patterns already defined in places such as `ReaderRevenueEpic` but there are some differences

### Asynchronous requests
Here we lean more on idiomatic React patterns for managing asynchronous requests. In effect, rather than using promises we are using custom hooks and composition.

### Data fetching
Another change is we're keeping the `fetch` call inside DCR. By having a centralised solution for data fetching from the platform (using abstractions like `useApi` and libraries like [SWR](https://swr.vercel.app/) we gain a number of benefits. Previously these abstractions were not available because the fetching logic was contained outside of React but with the approach given here it is now possible to adopt these fetching solutions.

### `consent` & `countryCode`
These two values are now passed in from 'outside' using custom hooks. This helps with reasoning through re-renders and also offers the potential opportunity to create shared solutions for any similar use cases in future. This change also means the actual logic for fetching and rendering the Epic is simpler, with less to deal with. It has these values as required props so simply never runs until they exist.


dotcom-rendering/src/web/components/LiveEpic.importable.tsx
```diff
@@ -0,0 +1,288 @@
...
+				setEpic(() => epicModule[name]); // useState requires functions to be wrapped
+			})
+			.catch((err) => {
+				const msg = `Error importing LiveBlog epic: ${err}`;
+				window.guardian.modules.sentry.reportError(
+					new Error(msg),
+					'liveblog-epic',
+				);
+			});
+	}, [url, name]);
```
TomPretty, line 180: could this be factored into a `useEpic` hook?

dotcom-rendering/src/web/components/LiveEpic.importable.tsx
```diff
@@ -0,0 +1,288 @@
...
+		<Render
+			url={data.data.module.url}
+			name={data.data.module.name}
+			props={data.data.module.props}
+		/>
+	);
+};
+
+/**
+ * LiveEpic is the Support the Guardian banner which appears in-between blocks in blogs
```
tomrf1, line 254: Can we call it `LiveblogEpic` to be extra clear about where it appears?
Also, it's best not to say "banner" in this context as the RR epic and banner are very distinct things!

dotcom-rendering/src/web/components/LiveEpic.importable.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+	// Check if we've been asked to force this epic
+	const params = new URLSearchParams(window.location.search);
+	if (params.get('force-liveblog-epic')) {
+		url += '?force=true';
+	}
+
+	// Send the payload to Contributions to request a module
+	const { data } = useApi<{
+		data: {
```
tomrf1, line 214: It's a shame to lose the common [logic and types from SDC](https://github.com/guardian/support-dotcom-components/blob/main/packages/dotcom/src/requests.ts#L10) here.
I think we should have a common way to do these api requests for all RR channels (epic, liveblog epic, banner, header links), and DCR should not define the types.

Also, what are the benefits of `useApi` here? I'm not disputing its use, I just don't know enough about it.

dotcom-rendering/src/web/components/LiveEpic.importable.tsx
oliverlloyd: > Can we call it `LiveblogEpic` to be extra clear about where it appears? 

Good idea, I'll make this change now.

> Also, it's best not to say "banner" in this context as the RR epic and banner are very distinct things!

Thanks. It's actually useful to have this explained, I'll make sure to check any comments so they follow this pattern

dotcom-rendering/src/web/components/LiveEpic.importable.tsx
oliverlloyd: Great idea!

dotcom-rendering/src/web/components/LiveEpic.importable.tsx
tomrf1: Thanks! Additional context -
In RR "banner" always means the popup at the bottom (`StickyBottomBanner`), which is generally not content-aware.
The epic is inserted into the content (either at the end of an article, or as a liveblog block), and is often related to the content.

dotcom-rendering/src/web/components/LiveEpic.importable.tsx
oliverlloyd: > It's a shame to lose the common [logic and types from SDC](https://github.com/guardian/support-dotcom-components/blob/main/packages/dotcom/src/requests.ts#L10) here. 

Agreed. I've made a change to import this type and use that instead

> I think we should have a common way to do these api requests for all RR channels (epic, liveblog epic, banner, header links)

Yeah, this makes a lot of sense. I really like how this PR is highlighting some interesting abstractions. `useEpic`, `usePayload`, etc. These could all - potentially - be hoisted up and used elsewhere. This fetch call could follow a similar route. I think it's good to be cautious though, the wrong abstraction done too early can be costly. But yeah, centralising this logic is a good shout.

> 
> Also, what are the benefits of `useApi` here? I'm not disputing its use, I just don't know enough about it.

We're in the process of moving all of our server state into [SWR](https://swr.vercel.app/). It's a fairly minimal library for storing state using React. Fetching data, storing it, passing it into components, deduplicating requests, caching, etc. is all tricky stuff to do and typically leads to complicated, brittle code. With SWR, you don't really need to think about it, you get automatic solutions for all sorts of common issues like these out the box.

It's obviously React focussed though and will not work at all outside of a React component.

`useApi` itself, is DCR's wrapper on top of SWR. It really just manages passing options and checking for errors.

In more general terms, we're keen to centralise data fetching on the platform. Not because we want to stop any other code making a fetch request but more because we want to make it easier to reason about where and when data is being brought into the stack. This is a medium term goal but one that we try to use to help made decisions. So here, where a fetch call is made which affects the main content on the page, ideally we would have more control over this at the platform level so that if we wanted to, say, introduce a loading state, then that's easier where it's the platform controlling the network call.

tomrf1: Thanks, that all sounds great.

At some point it would be nice to have all of these SDC requests follow the same pattern. The others are:
- [banner](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx#L189)
- [epic](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx#L123)
- [header links](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/ReaderRevenueLinks.tsx#L201)

Currently the `force-liveblog-epic` part comes for free with these existing calls

oliverlloyd: 100% agree. I also appreciate you don't want multiple ways to do the same thing to exist in parallel for a long time

dotcom-rendering/src/web/components/LiveBlogEpic.importable.tsx
```diff
@@ -0,0 +1,289 @@
...
+					'liveblog-epic',
+				);
+			});
+	}, [url, name]);
+
+	return { Epic };
+};
+
+function consentGivenForArticleCounts(consent: ConsentState): boolean {
+	const REQUIRED_CONSENTS_FOR_ARTICLE_COUNT = [1, 3, 7];
```
AshCorr, line 85: What do these numbers mean? Is it worth defining these as constants somewhere?

AshCorr (approved):
👍 

Interesting LiveBlogEpic -> Fetch -> Render pattern!

dotcom-rendering/src/web/components/LiveBlogEpic.importable.tsx
tomrf1: this is actually defined [here](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/lib/contributions.ts#L146) as well.
Though I don't know specifically what these numbers mean, there must be a CMP guide somewhere...

tomrf1: Also, I think we can use [this function](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/lib/contributions.ts#L157) instead here

oliverlloyd: I don't know what they mean (I think it's a SourcePoint thing?) but I've now made a change to import them from `lib/contributions`

oliverlloyd: > Also, I think we can use [this function](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/lib/contributions.ts#L157) instead here

That function returns a promise, which doesn't play nice with React. Also, the pattern here has `consent` being passed in as a prop rather than the function getting it itself. It was for these reasons I recreated it locally.

TomPretty: I remember @mxdvl sent me a link one time to some json where you could see what all of the consents meant. It would be nice to have meaningful labels for these, but I guess that should live in the cmp repo instead of in dcr?

PR #4217, opened 2022-03-10
Don't skip this test anymore

oliverlloyd:
## What does this change?
Enables the test to verify that when clicking on the toast from the second page we are correctly routed

## Why?
Because we merged [this PR](https://github.com/guardian/dotcom-rendering/pull/4099) which is what we were [waiting on](https://github.com/guardian/dotcom-rendering/pull/4105)


DavidLawes (approved):
👍

PR #4218, opened 2022-03-10
Single front door test v2

marialani:
## What does this change?
Currently we only show the the "Subscriptions" link in the main navigation to supporters. For the duration of the Single Front Door AB Test we will display a "Print Subscriptions" link which isn't dependant on the user's supporter status and whether they're in the AB test. 

We will also hide the Supporter Revenue links in the footer for all users.

## Changes
- Hide Supporter Revenue links in the footer
- Update "Subscriptions" link in desktop navigation
- Update "Subscriptions" link in mobile navigation

These changes have been reproduced in frontend here: https://github.com/guardian/frontend/pull/24742

## Screenshots
| Before | After |
|-- | -- |
| Screenshot 2022-01-14 at 13 05 20 | Screenshot 2022-01-14 at 12 57 39 |
| Screenshot 2022-01-14 at 13 16 49 | Screenshot 2022-01-14 at 13 16 36 |

## Why
To support the new product proposition work we are going to run an ABC test to measure the effect of the current contribute/subscribe split messaging vs a single 'support the guardian' call to action. This will be done in combination with bringing forward the existing ‘Supporter Newsletter’ & reduced support asks benefits to contributors on the support site.

### DIFF
--- a/dotcom-rendering/src/web/components/Links.importable.tsx
+++ b/dotcom-rendering/src/web/components/Links.importable.tsx
@@ -232,27 +232,20 @@ export const Links = ({
 
 	const isServer = typeof window === 'undefined';
 
-	const showSupporterCTA =
-		!isServer &&
-		getCookie({
-			name: 'gu_hide_support_messaging',
-			shouldMemoize: true,
-		}) === 'true';
-
 	const isSignedIn =
 		!isServer && !!getCookie({ name: 'GU_U', shouldMemoize: true });
 
 	return (
 		<div data-print-layout="hide" css={linksStyles}>
-			{showSupporterCTA && supporterCTA !== '' && (
+			{supporterCTA !== '' && (
 				<>
 					<div css={seperatorStyles} />
 					<a
-						href={supporterCTA}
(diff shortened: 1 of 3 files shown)


tomrf1 (approved):
👍

dotcom-rendering/src/web/components/Links.importable.tsx
```diff
@@ -251,18 +244,18 @@ export const Links = ({
 			{/* Since 'subscriptions' is only rendered on the client, rendering it within
 				a div is required to prevent DOM elements getting mis-matched during hydration  */}
 			<div css={linkWrapperStyles}>
-				{showSupporterCTA && supporterCTA !== '' && (
+				{supporterCTA !== '' && (
```
GHaberis, line 247: We don't need the conditional `supporterCTA !== ''` check here, as we're not using `supporterCTA` in the link whilst the SFD test is running.

GHaberis (approved):
Looks good 👍  

Feels like we should consider putting these changes behind a feature switch if we're going to be switching further SFD tests on/off in Q1, rather than reverting these changes each time a test ends.

marialani:
> Looks good 👍
> 
> Feels like we should consider putting these changes behind a feature switch if we're going to be switching further SFD tests on/off in Q1, rather than reverting these changes each time a test ends.

Yeah this sounds like a good idea to me especially with the recent migration of a few related components to Islands

PR #4219, opened 2022-03-10
Ignore elements causing false positives

oliverlloyd:
## What does this change?
Adds a couple of `data-chromatic="ignore"` attributes to tell Chromatic to not create diffs for these things

## Why?
Because we've been getting false positives on these for a while now and people are getting antsy

## Why not fix the root issue?
Fair. This is a stopgap to calm the masses but we would ideally look deeper into why these were happening. Will that really happen? Maybe, maybe not. It's all about trades offs in the end.


DavidLawes (approved):
👍

PR #4229, opened 2022-03-11
No RichLink images in blogs

### LINKED ISSUE #4172: Images in rich links
Decide if we should continue to render images in rich links.

Currently this seems tempermental in Frontend where sometimes it renders images and sometimes it doesn't.

Images can sometimes look [out of](https://m.code.dev-theguardian.com/business/live/2022/feb/11/uk-gdp-december-2021-omicron-recovery-stock-markets-ftse-inflation-fed-business-live) [place](https://www.theguardian.com/football/live/2022/feb/09/manchester-city-v-brentford-tottenham-v-southampton-premier-league-live?page=with:block-62043a678f08b0eb1f62bd9a#block-62043a678f08b0eb1f62bd9a), especially in liveblogs.

Let's decide if we want to continue rendering images and what we can do to improve them

oliverlloyd:
## What does this change?
This prevents images from being dynamically inserted into rich links when in live or dead blogs.

## Why?
Richlinks get their image data client side so any image is inserted after the page has already loaded. On normal articles this is not such a problem as the richlinks sit in the left column. But in blogs they are inline with the main content and so dynamically inserting them causes a lot of content layout shift.

Another issue with rich ink images in blogs is their size. In articles the images are small, typically a little larger than a thumbnail, but in blogs the same image is stretched to 700px wide - a size it was not scaled for - and they appear pixelated

### Before
Screenshot 2022-03-11 at 10 32 34

### After
Screenshot 2022-03-11 at 10 32 14

Fixes #4172

### DIFF
--- a/dotcom-rendering/src/web/components/RichLink.tsx
+++ b/dotcom-rendering/src/web/components/RichLink.tsx
@@ -303,7 +303,7 @@ export const RichLink = ({
 								<Avatar
 									imageSrc={contributorImage}
 									imageAlt={mainContributor}
-									palette={decidePalette(linkFormat)}
+									palette={palette}
 								/>
 							</div>
 						)}
--- a/dotcom-rendering/src/web/components/RichLink.tsx
+++ b/dotcom-rendering/src/web/components/RichLink.tsx
@@ -8,7 +8,7 @@ import {
 	textSans,
 	from,
 } from '@guardian/source-foundations';
-import { ArticleFormat, ArticleSpecial } from '@guardian/libs';
+import { ArticleDesign, ArticleFormat, ArticleSpecial } from '@guardian/libs';
 
 import ArrowInCircle from '../../static/icons/arrow-in-circle.svg';
 
@@ -26,6 +26,7 @@ interface Props {
 	contentType: ContentType;
 	url: string;
 	starRating?: number;
+	linkFormat: ArticleFormat;
 	format: ArticleFormat;
 	tags: TagType[];
 	sponsorName: string;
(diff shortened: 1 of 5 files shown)


dotcom-rendering/src/web/components/RichLink.tsx
```diff
@@ -296,7 +303,7 @@ export const RichLink = ({
 								<Avatar
 									imageSrc={contributorImage}
 									imageAlt={mainContributor}
-									palette={decidePalette(format)}
+									palette={decidePalette(linkFormat)}
```
AshCorr, line 306: `palette={pallet}` here?

AshCorr (approved):
Awesome!

JamieB-gu:
> On normal articles this is not such a problem as the richlinks sit in the left column.

Isn't it the case that most pageviews on normal articles are likely to be on smaller breakpoints and therefore the rich links _are_ inline? 🤔

oliverlloyd:
> > On normal articles this is not such a problem as the richlinks sit in the left column.
> 
> Isn't it the case that most pageviews on normal articles are likely to be on smaller breakpoints and therefore the rich links _are_ inline? 🤔

When they're inline on articles they're smaller



But sure, it is still a problem on article, just not as much. There's also the fact that blogs are chock full of richlinks whereas articles have one or two

dotcom-rendering/src/web/components/RichLink.tsx
oliverlloyd: Great spot!

Fixed

PR #4230, opened 2022-03-11
feat: Change cursor to indicate that the acordions are clickable

### LINKED ISSUE #4162: The chevrons on `Accordion` are not semantically marked up
For accessibility and to be able to show the right pointer on hover we should use a `button` or `a` tag here

AshCorr:
## What does this change?

Updates the cursor item to show that the summary and chevron of acordions are clickable.

Fixes #4162 

## Accessibility

Part of #4162 was to address the accessibility concerns of acordions where we wanted to wrap the chevrons in a link or button, however, doing this breaks the collapse and expand behaviour of the accordion. I think the `<summary>` and `<details>` tags are pretty accessible now days and don't need anything extra for accessibility.

I've tested the accordions with NVDA and it seems to announce that our accordions are clickable and announces when they collapse and expand.

### Before

chrome_fIriEHepIp

### After
chrome_PlsreHrj0Z

### DIFF
--- a/common-rendering/src/components/accordion.tsx
+++ b/common-rendering/src/components/accordion.tsx
@@ -37,6 +37,7 @@ const detailsStyles: SerializedStyles = css`
 `;
 
 const titleRowStyles = (supportsDarkMode: boolean): SerializedStyles => css`
+	cursor: pointer;
 	position: relative;
 	display: block;
 	align-items: center;


oliverlloyd (approved):
One line code changes backed with lots of research and thought make the best PRs

PR #4231, opened 2022-03-11
`CustomEvent('liveblog:blocks-updated')`

### LINKED ISSUE #4178: Ads are dynamically inserted in-between new blocks
When a liveblog first loads the commercial script reads the page and chooses which blocks to insert adverts in-between. But as new blocks are added to the top no new ad slots are being created. We need the commercial script to keep dynamically inserting blocks as content is slowly added.

oliverlloyd:
## What does this change?
DCR now sends a custom event when new blocks are revealed on the page

## Why?
We need a way to notify commercial that they should re-run spacefinder

Closes #4178


chrislomaxjones (approved):
Thanks for adding this, looks good from the commercial perspective!

PR #4234, opened 2022-03-11
Coerce https for interactive urls

oliverlloyd:
## What does this change?
Force interactive urls to use `https`

## Why?
This piece was not loading the iframe correctly because https is required but the content is using http. We could update the content but forcing it on the platform is a good safety net.

Is there a better way to do this other than string manipulation? Maybe? I'm open to suggestions.

### DIFF
--- a/dotcom-rendering/src/web/components/InteractiveBlockComponent.importable.tsx
+++ b/dotcom-rendering/src/web/components/InteractiveBlockComponent.importable.tsx
@@ -250,7 +250,7 @@ export const InteractiveBlockComponent = ({
 			iframe.style.width = '100%';
 			iframe.style.border = 'none';
 			iframe.height = decideHeight(role).toString();
-			iframe.src = url;
+			iframe.src = url.replace('http:', 'https:');
 
 			setupWindowListeners(iframe);
 


AshCorr (approved):
Super quick fix!

jamesgorrie (approved):
Anything other than string replace would probably be harder to read. The on;y thing I can think this might break is something like `https://a.com?url=http://bloop.com` but that seems unlikely.

To mitigate without adding complexity you could `.replace(/^http:/)` again, probably overkill, just saying as you asked.

OllysCoding (approved):
+1 Seconding what others are saying. I think a string replace is fine, but another simple solution could be to wrap the replace in a `if (string.startsWith('http:')) { ... }`

JamieB-gu:
> Is there a better way to do this other than string manipulation? Maybe? I'm open to suggestions.

I think the `URL` API is mutable, so this might be an option that may avoid some of the above mentioned pitfalls:

```ts
const a = new URL(url)

if (a.protocol === 'http:') {
  a.protocol = 'https:'
}

iframe.src = a.href
```

PR #4237, opened 2022-03-14
Testing on AMP: Unit tests for `getGroup`

chrislomaxjones:
## What does this change?

Add a set of unit tests for the `getGroup` function.

## Why?

`getGroup` is used to assign a piece of content to one of 12 test groups by hashing its `pageId`.

Since we need to run the exact same function in SQL at the analysis stage, it's important to maintain consistency between the two implementations. These tests will help to catch any regressions introduced by changing the implementation of `getGroup`, as if any of the group assignments change in the future this will be caught in CI.


mxdvl (approved):
Good idea!

PR #4208, opened 2022-03-09
add duration tracking to pinned post

marjisound:

PR #4242, opened 2022-03-14
Add tracking to dcr no-js pinned post 

marjisound:
## What does this change?
This PR adds the following tracking events to pinned pos. PR is blocked until https://github.com/guardian/dotcom-rendering/pull/4211 is merged.
 - view duration tracking 
 - show-less/show-more click tracking
 - ~~INSERT event when the component is inserted (this is because sometimes the component will be loaded but not viewed if it is below the fold.)~~
 - Instead of `INSERT` tracking event, the `pinned-post` is added to the `renderedComponent` list in the page view by adding the `data-component` to the pinned post component. This way whenever a page has a pinned post, in data lake `pageview` table, the pinned-post is added to the `rendered_components` field. 

## Why?
D & I wants to know 
- how long a user stays on the pinned post during their page visit
- when the show-more/show-less button is clicked
- when the pinned post is inserted


marjisound:
Hi @oliverlloyd.  if you remember last week I made a PR https://github.com/guardian/dotcom-rendering/pull/4208 on adding tracking for the pinned post on dcr, and then I realized Anna was re-working to do a No-JS pinned post on dcr. I made this new PR which is a branch based from Anna's new branch (`dcr-nojs-pinned-post`). 

 I couldn't find a way to use `useIsInView` and had to create a new `IntersectionObserver`. I have a feeling that I'm not taking the right approach and will appreciate if you make any suggestions.

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
```diff
@@ -25,9 +27,41 @@ export const PinnedPostLiveness = ({}) => {
...
 
+	if (pinnedPost) {
+		submitComponentEvent({
+			component: {
+				componentType: 'LIVE_BLOG_PINNED_POST',
+				id: pinnedPost?.id,
+			},
+			action: 'INSERT',
+		});
+	}
```
marjisound, line 38: I wanted to emit the `INSERT` event only when pinned post is loaded in the page. Meaning we only want to emit 1 event at the beginning. I'm not sure if it's right to just emit the event like this, and I think maybe, we need to create a isLoaded state or something like that to make sure we only emit the `insert` event when the page loads. 

Right now there's no state in this component `PinnedPostLiveness` so from what I see, there should not be any re-rendering. 

What do you think is the best way to tackle this?

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
```diff
@@ -25,9 +27,41 @@ export const PinnedPostLiveness = ({}) => {
 		pinnedPost && pinnedPost.scrollHeight <= pinnedPost.clientHeight;
 	if (contentFitsContainer) hideShowMore();
 
+	if (pinnedPost) {
```
marjisound, line 30: I wanted to emit the INSERT event only when pinned post is loaded in the page. Meaning we only want to emit 1 event at the beginning. I'm not sure if it's right to just emit the event like this, and I think maybe, we need to create a isLoaded state or something like that to make sure we only emit the insert event when the page loads.

Right now there's no state in this component PinnedPostLiveness so from what I see, there should not be any re-rendering.

What do you think is the best way to tackle this?

oliverlloyd: I'm not sure if I'm reading this right but how does this sound?

If this event represents the pinned post becoming visible (appears in viewport), then I guess you could say it's the same as duration? You will never have a duration value without the post becoming visible and every post that is visible also has a duration. So... do we even need this event?

If D&I need to have a sort of count unique for page views that had a pinned post become visible, then I think this is possible as a query on the data. This would be preferable to the browser having to send more data.

marjisound: Great points. Most probably as you said we may not even need to send this data any more. Let me share this with the team and also with D & I. Will update the decision here later

marjisound: Just before writing this to the D & I, wanted to double check something with you. You mentioned that through query on the data we can get the count unique for page views that had a pinned post? 

But if a page with pinned post never emits the `duration` event meaning the pinned post was never viewed, how does the D & I find out if that page had a pinned post (if we were to stop emitting `INSERT` events)?

oliverlloyd: 🤔 I thought the INSERT event was only fired if the post came into view? In which case the duration would always also be sent

oliverlloyd: Of course, if it is the opposite and we want to track every page view that included a pinned post, then can we derive this information from the plain page view data? Or maybe using a simple `data-component-name` attribute? I'm not super sure how the `tracker-js` script works but maybe there's an easy solution there?

marjisound: Thanks @oliverlloyd for pointing this out. After some digging in with @alinaboghiu and chatting with @rtyley I think the solution could be as simple as adding `data-component="pinned-post"` to the `PinnedPost` component. I just ran and tested it using the ophan tester tool and the following is showing up in the page view data:
image

marjisound: @oliverlloyd what do you think about this change? If you are happy with it, I also prepared the same change in frontend for the pinned post https://github.com/guardian/frontend/pull/24763

oliverlloyd: Amazing! This looks a lot simpler 👍

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
```diff
@@ -38,5 +62,45 @@ export const PinnedPostLiveness = ({}) => {
 			);
 		};
 	}, []);
```
joecowton1, line 64: Does it matter that `pinnedPostCheckbox` is not included as a dependancy here? (in the `[]`)

joecowton1:
@marjisound looks great! Have left a couple of comments which are genuine questions - let me know if you want to discuss anything.

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
```diff
@@ -25,9 +27,31 @@ export const PinnedPostLiveness = ({}) => {
...
-			// TODO: ADD OPHAN CLICK TRACKING
+			if (pinnedPostCheckBox instanceof HTMLInputElement) {
+				if (pinnedPostCheckBox.checked) {
+					submitComponentEvent({
+						component: {
+							componentType: 'LIVE_BLOG_PINNED_POST',
+							id: pinnedPost?.id,
+						},
+						action: 'CLICK',
+						value: 'show-less',
```
joecowton1, line 42: Just checking the `value` field and I wonder if these are coming out the wrong way around? When `checked` is true the click event should have been `show-more`, but it's currently `show-less`  does that sound correct?

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
```diff
@@ -25,9 +27,31 @@ export const PinnedPostLiveness = ({}) => {
 		pinnedPost && pinnedPost.scrollHeight <= pinnedPost.clientHeight;
 	if (contentFitsContainer) hideShowMore();
 
+	const hasBeenSeen = useRef(false);
+
 	useEffect(() => {
 		const handleClickTracking = () => {
```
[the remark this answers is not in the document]
marjisound: Good question, I'm also not sure but I guess maybe because the actual component doesn't live in this file and we also need to do `pinnedPostCheckBox?.addEventListener('change', handleClickTracking);` and `pinnedPostCheckBox?.removeEventListener('change', handleClickTracking);`

But I'm not sure if doing these would be problematic out of the `useEffect` 🤔 

@abeddow91 and @oliverlloyd do you know if an event listener is best to be placed in a useEffect when working with liveness script?

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
marjisound: Yeaa you're right. I'm just a bit confused as why we have it this way in frontend? Maybe that also needs to change? https://github.com/guardian/frontend/blob/c19ce5ce2f0bfe48868e9a5527a2c40f96e459e8/static/src/javascripts/bootstraps/enhanced/liveblog.js#L141

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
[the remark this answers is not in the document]
joecowton1: Sorry I could have been clearer: if you add and remove the event listener within the useEffect, you don't need the `handleClickTracking` function to be inside the useEffect, unless it contains a dependancy that might update, which in this case it doesn't :)

Speaking to @mxdvl about this we think the `useCallback` is unnecessary, but the click handler should live outside the useEffect.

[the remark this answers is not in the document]
oliverlloyd: `useEffect` is soooo confusing!

Did you see @arelra 's post yesterday in the Client Side channel? I found that helpful in building a better mental model for this logic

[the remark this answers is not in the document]
marjisound: Thanks @joecowton1 for the thorough explanation :) I moved the click handler function out of the `useEffect`.

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
joecowton1: Yeah good question - I think @abeddow91 added those to Frontend so she might be the person to ask!

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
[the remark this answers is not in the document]
joecowton1: Great, thanks!

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
```diff
@@ -38,5 +63,45 @@ export const PinnedPostLiveness = ({}) => {
 			);
 		};
 	}, []);
+
+	// calculate duration when user is viewing pinned post
+	// and emit ophan events when the pinned post goes out of view
+	useEffect(() => {
+		if (!pinnedPost) return () => {};
```
mxdvl, line 70: You don’t need to return an empty cleanup function.

```suggestion
		if (!pinnedPost) return;
```

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
```diff
@@ -38,5 +63,45 @@ export const PinnedPostLiveness = ({}) => {
...
+					pinnedPostTiming.clear();
+					pinnedPostTiming.start();
+				} else if (hasBeenSeen.current) {
+					const timeTaken = pinnedPostTiming.end();
+					if (timeTaken) {
+						const timeTakenInSeconds = timeTaken / 1000;
+						submitComponentEvent({
+							component: {
+								componentType: 'LIVE_BLOG_PINNED_POST',
+								id: pinnedPost?.id,
```
mxdvl, line 87: I think you’re checking that `pinnedPost` isn’t undefined above

```suggestion
								id: pinnedPost.id,
```

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
```diff
@@ -25,10 +27,33 @@ export const PinnedPostLiveness = ({}) => {
 		pinnedPost && pinnedPost.scrollHeight <= pinnedPost.clientHeight;
 	if (contentFitsContainer) hideShowMore();
 
+	const hasBeenSeen = useRef(false);
```
mxdvl, line 30: Why are we using a ref here? So it doesn’t reset between renders?

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
abeddow91: Yep this certainly does look to be the wrong way round. I'll pull this branch and have a play. In Frontend, the box is checked before the event fires which is why they seem to be the wrong way round but it would be good to check if this is the same here.

abeddow91: This is the wrong way round here. I've committed the change whilst Marjan's on AL

abeddow91 (approved):
Great work and thanks for the clear description!

dotcom-rendering/src/web/components/PinnedPostLiveness.importable.tsx
mxdvl: This can be fixed forward as far as I’m concerned. Not a blocker. If we can use a state directly, though, we probably should, as this is the main benefit of (P)React!

abeddow91: Given that this only gets set once, could we use a let here instead of useRef?

mxdvl: You’d probably want to `useState`, but I have to say I can’t quite follow timing logic, so if this works, I’m happy.

If we want to keep things as functional as possible, there is this library which provides the observer as a hook: https://github.com/asyarb/use-intersection-observer

```ts
const [hasBeenSeen, setHasBeenSeen] = useState<boolean>(false);

const isInView = useIntersectionObserver({ element: pinnedPost, threshold: 0.1 })

useEffect(() => {
  if(!hasBeenSeen && isInView) {
    setHasBeenSeen(true);
    pinnedPostTiming.clear(); // I think this will no longer be necessary
    pinnedPostTiming.start()
  } else if (hasBeenSeen) {
    pinnedPostTiming.end()
  }
}, [hasBeenSeen, isInView])

```

abeddow91: Thanks Max! That's really interesting. I'll leave it as is for now and merge so as not to block but maybe I'll put a new ticket around this as a possible refactor.
I'll also tag @marjisound here too as I think she'll find this interesting.

oliverlloyd: We already have a custom hook for the intersection observer in DCR [here](https://github.com/guardian/dotcom-rendering/blob/02177fb7f911eb5e2ba50c99ebd4a76342a61bd6/dotcom-rendering/src/web/lib/useIsInView.ts)

marjisound: Hi @mxdvl was just looking at the suggestion you made and had a chat with @oliverlloyd too. I tried using the `useIsInView` hook but it doesn't work in this instance. That's because  `setNode` expects the element it is being used on to be rendered by React. And in our case the element is a dom element we retrieved through query selector `window.document.querySelector('[data-gu-marker=pinned-post]')`

So when I try `setNode(pinnedPost);` I get the following error: 
image

Regarding the other suggestion you had made for using `useState` instead of `useRef` for `hasBeenSeen`: 

I went with the `useRef` approach because I thought it's not necessary to have a re-render when the `hasBeenSeen` value changes. Although, the `hasBeenSeen` is supposed to be changed only the first time intersection happens. And after then the whole time the user stays in the page, the `hasBeenSeen` doesn't change at all. So if we use `useState` we'd only have 1 extra re-render. 

The current work using the `useRef` works fine, but do you think using the `useState` will bring any advantages? I'm happy to change it if it'll give us any benefit :)

mxdvl: The type error you’re getting is only a compiler error, which you can circumvent by telling TS that you’re selecting an `HTMLElement`, not just any `Element`:

```ts
const pinnedPost: HTMLElement | null = !isServer
	? window.document.querySelector('[data-gu-marker=pinned-post]')
	: null;
const pinnedPostCheckBox: HTMLElement | null = !isServer
	? window.document.querySelector('input[name=pinned-post-checkbox]')
	: null;
```

Regarding `useRef` v `useState`, I don’t think we’d have a way to measure the performance impact of choosing one over the other anyways. Let’s keep it as-is.

PR #4246, opened 2022-03-14
Amp: Lines going through byline

OllysCoding:
## What does this change?

Handle edge-case where there is a bylineImage but multiple contributors ~ padding should have been applied to compensate for the image not being rendered.

### Before

image

### After

image


jamesgorrie (approved):

jamesgorrie:
Should cypress have picked up on this?

PR #4247, opened 2022-03-14
Hide gap between main image and header until tablet screen size on Liveblogs

### LINKED ISSUE #4151: Remove space above mainmedia on mobile

AshCorr:
## What does this change?

Hides the gap between the main image and the header if on a display smaller than a tablet.

## Why?

The gap doesn't look very nice on a mobile display.

Fixes #4151 

### Before

image

### After

image


oliverlloyd (approved):
So simple!

PR #4250, opened 2022-03-14
`ArticleCount`

oliverlloyd:
This PR depends on #4126 

## What does this change?
Moves the code to increment the article count out of `App.tsx`

## Why?
We're deleting it

## Another micro island?
Perhaps we can optimise things by not having so many micro islands making http calls from `Page`. This PR expands on this pattern for simplicity but I think we should look at rationalising all of these in a follow up PR.


AshCorr (approved):
👍 

Worth just making App.tsx an island and importing that?

PR #4252, opened 2022-03-15
fix: Force twitter embed to respect container size

### LINKED ISSUE #4158: Resize atoms (Twitter, Youtube, etc) on orientation change.
Currently any embeds on the page won't be resized when the page is resized, this can make things such as Tweets break out of the pages when switching from a wider display (landscape) to a smaller display (portrait)

AshCorr:
## What does this change?

Overrides twitters fixed width on Tweets and instead sets it to be 100% of the containers width. This allows the page to be resized without breaking.

## Why?

Fixes #4158 

### Before

chrome_VJYUSdkiAJ

### After

chrome_yWCui5ODx0

### DIFF
--- a/dotcom-rendering/src/web/components/TweetBlockComponent.tsx
+++ b/dotcom-rendering/src/web/components/TweetBlockComponent.tsx
@@ -13,6 +13,10 @@ const noJSStyling = css`
 		${body.small()};
 	}
 
+	.twitter-tweet iframe {
+		width: 100% !important;
+	}
+
 	.twitter-tweet p {
 		padding-bottom: 10px;
 	}
--- a/dotcom-rendering/src/web/components/TweetBlockComponent.tsx
+++ b/dotcom-rendering/src/web/components/TweetBlockComponent.tsx
@@ -14,6 +14,8 @@ const noJSStyling = css`
 	}
 
 	.twitter-tweet iframe {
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/components/TweetBlockComponent.tsx
```diff
@@ -13,6 +13,10 @@ const noJSStyling = css`
 		${body.small()};
 	}
 
+	.twitter-tweet iframe {
+		width: 100% !important;
```
AshCorr, line 17: Unfortunate that we need to use !important here but I don't think theres any other way we overwrite Twitters CSS.

OllysCoding: If this works I'll take it! Could we add a comment here clarify why we use `!important` maybe?

jamesgorrie: I think you might need to add something like [this](https://github.com/guardian/dotcom-rendering/blob/bd23511a016767ee2d24cce03573fb1e76479710/dotcom-rendering/src/web/components/SignInGate/gateDesigns/shared.tsx#L65) because of our linting.

DavidLawes (approved):
👍

PR #4257, opened 2022-03-15
Better Function Definition For Liveblocks Parser

JamieB-gu:
## Why?

I noticed that this could be improved a little bit.

The intermediate return type was unnecessary; arrow functions allow you to provide a nice, clean way to write curried function definitions.

I've also swapped the order of the arguments. With curried functions it's common to set up the argument order to aid later partial application and composition. In this case I think it's more likely that you'll want to create a partially applied function with the context and then pass the blocks later, possibly as part of a chain of operations.

## Change

- Removed unnecessary intermediate type annotation
- Swapped the arguments to aid composition

### DIFF
--- a/apps-rendering/src/item.ts
+++ b/apps-rendering/src/item.ts
@@ -363,7 +363,7 @@ const fromCapiLiveBlog =
 			? 30
 			: 10;
 
-		const parsedBlocks = parseLiveBlocks(body)(context);
+		const parsedBlocks = parseLiveBlocks(context)(body);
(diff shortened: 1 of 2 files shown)


mxdvl (approved):
neat

JamieB-gu:
Oh! Just looking through the conflicts I think an equivalent change happened in #4347.

I'll close this, but thank you Max for confirming that it was a reasonable thing to do 🙂👍.

PR #4262, opened 2022-03-16
👋  `App.tsx`

oliverlloyd:
## What does this change?
This PR removes `App.tsx` and its associated workings from DCR

## Why?
There are [two](https://docs.google.com/document/d/1gHOzvFZxrAO2mvNPKg4UmMAYZm7Jk1NutZPVLY5UxPY/edit#heading=h.1y29fcc1ygqg) [documents](https://docs.google.com/document/d/1kSrZZrJ8hSw6bek5hGa9onE8QDz5TRHFqHV52EOFjts/edit#) which discuss this but in short: The pattern was hard to work with because of shared state and led to more data being sent to readers than was needed

### Loadable
Is now gone. We used this for lazy loading react components but this is now acheived via the islands pattern

### `CAPIBrowserType`
Is now gone. This was used type as part of the data set at `window.guardian.app` which is now deleted

### `window.guardian.app`
Is now gone. We stored data for here to later use as part of hydration. The islands pattern however serialises each islands data directly so there's no need for a centralised object on `window`

### `webpack-manifest-plugin`
Added to replace functionality that was previously provided by Loadable

### `GA` > `GAData`
Renamed to be clearer about what this property is


dotcom-rendering/src/model/window-guardian.ts
```diff
@@ -66,165 +68,9 @@ const makeWindowGuardianConfig = (
...
-		contributionsServiceUrl,
-		isImmersive: CAPI.isImmersive,
-		isPhotoEssay: CAPI.config.isPhotoEssay || false,
-		isSpecialReport: CAPI.isSpecialReport,
-		isLiveBlog: CAPI.config.isLiveBlog || false,
-		isLive: CAPI.config.isLive || false,
-		isPreview: CAPI.config.isPreview,
-		stage: CAPI.config.stage,
-		matchUrl: CAPI.matchUrl,
-		elementsToHydrate: CAPI.blocks
```
mxdvl, line 207: Have we checked that nothing is expecting this to be present on the page?

I’m thinking mostly of @guardian/commercial-dev

oliverlloyd: I pinged comm dev and they said they are only using `window.guardian.config`. Specifically:

```
Ashish Puliyel, Yesterday 16:58
I'm afraid of commitment. So I'm scared to say we never use it. But I've looked and not seen anywhere where we use window.guardian.app
```

mxdvl: Oh, right, I didn’t see it was kept on L260.

mxdvl (approved):
Begone!

dotcom-rendering/src/lib/assets.ts
```diff
@@ -33,40 +32,30 @@ const decideAssetOrigin = (stage: string | undefined): string => {
...
-	// Builds with the SKIP_LEGACY flag will not include legacy bundles, this aids build performance.
-	// 'ophan.87b473fc83e9ca6250fc.js' -> 'ophan'
-	const chunkName = filename.split('.')[0];
-	const chunks: string[] | undefined =
-		loadableManifestLegacy.assetsByChunkName?.[chunkName];
-	const legacyFilename = chunks && chunks.length > 0 && chunks[0];
+	const filename = isDev ? file : manifest[file];
+	const legacyFilename = isDev
+		? file.replace('.js', '.legacy.js')
+		: legacyManifest[file];
```
oliverlloyd, line 43: So much clearer! 😻

dotcom-rendering/src/web/browser/sentryLoader/sentry.ts
```diff
@@ -50,11 +51,6 @@ Sentry.init({
 	},
 });
 
-Sentry.configureScope((scope) => {
-	scope.setTag('edition', editionLongForm);
-	scope.setTag('contentType', contentType);
-});
```
oliverlloyd, line 56: Removed in favour of simplicity.

It made sense to include this extra data on errors when it was freely available but now we'd need to refactor the window object to also include these properties which feels excessive

oliverlloyd (commented):
I've read through the code and nothing is jumping out to me here. I can't approve because I'm an author but nonetheless I do heartily approve.

OllysCoding (approved):
So much code removed 😍

mxdvl:
🏝️  #3629

Issue #4228, opened 2022-03-10
Image Captions - Replace Triangle With Camera


JamieB-gu:
@HarryFischer recommended that we replace the triangle icon that appears in image captions with a camera icon. This change has already been made on DCR so you can see examples of this on most articles on dotcom. The change can be made here:

https://github.com/guardian/dotcom-rendering/blob/35d5d5b6de53a9a7502c92998e95c3c27cd7302a/common-rendering/src/components/figCaption.tsx#L95-L101

HarryFischer:
Also need to make caption font-size 14px

PR #4265, opened 2022-03-16
Add back captions to interactive block element

OllysCoding:
## What does this change?

Adds back captions to InteractiveBlockElements

## Why?

They went missing in #4077

### Before
image

### After

image

### DIFF
--- a/dotcom-rendering/src/web/lib/renderElement.tsx
+++ b/dotcom-rendering/src/web/lib/renderElement.tsx
@@ -383,6 +383,7 @@ export const renderElement = ({
 						role={element.role}
 						format={format}
 						elementId={element.elementId}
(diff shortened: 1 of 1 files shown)


oliverlloyd (approved):

AshCorr (approved):

paperboyo:
Should they have a camera icon? (just shows that a generic triangle, which itself had replaced the camera, was safer/easier) cc  @HarryFischer

HarryFischer:
I was team triangle but others weren't so in this instance I think camera is correct

JamieB-gu:
> Should they have a camera icon?

> I was team triangle but others weren't so in this instance I think camera is correct

Related: #4228

PR #4270, opened 2022-03-16
feat: use JSDoc types

mxdvl:
## What does this change?

Convert a TSX component to JSX, for demonstration purposes.

[Read Typescript documentation on JSDoc types](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html).

## Why?

This is something we should consider, especially if we’re not using the most intricate features of typescript and want to skip the compilation step.

### DIFF
--- a/dotcom-rendering/src/web/components/MostViewedRightWrapper.importable.jsx
+++ b/dotcom-rendering/src/web/components/MostViewedRightWrapper.importable.jsx
@@ -1,24 +1,29 @@
-import { useRef, useState, useEffect, RefObject } from 'react';
+import { useRef, useState, useEffect } from 'react';
 import { css } from '@emotion/react';
 
 import { MostViewedRight } from './MostViewedRight';
 import { useAdBlockInUse } from '../lib/useAdBlockInUse';
 
-type Props = {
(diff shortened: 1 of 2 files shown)


dotcom-rendering/src/web/components/MostViewedRightWrapper.importable.jsx
```diff
@@ -29,7 +34,8 @@ export const MostViewedRightWrapper = ({ limitItems, isAdFreeUser }: Props) => {
 	`;
 
 	useEffect(() => {
-		const checkHeight = (ref: RefObject<HTMLDivElement>) => {
+		/** @param {import('react').RefObject<HTMLDivElement | null>} ref */
+		const checkHeight = (ref) => {
```
OllysCoding, line 38: I find this particularly awkward to read in JSDOC, is there a nicer way to format it? At least a way to exclude needing `import('react')` from within the comment itself?

mxdvl: I think `@typedef` can be used to import types at the top. I'll have a go.

PR #4272, opened 2022-03-16
`BootPage`

oliverlloyd:
I'm genuinely unsure if this is a good idea or not and am keen to hear thoughts.

My motivation is to reduce the number of micro islands clogging up the http space but am I now re-introducing `App.tsx`? 😱


jamesgorrie (commented):
My first question is why? Specifically what do these components have in common that makes them a single component? Is it because they share similar props?

mxdvl:
If we have an inkling this might improve performance, could we simply measure it?

If there's no perf gain I wouldn't be keen to go down this path.

oliverlloyd:
Closing this for now. We can come back to this if we feel there might be benefit to merging islands

PR #4278, opened 2022-03-16
change padding to margin – geo popular images

HarryFischer:
## What does this change?
Makes most viewed images proper circles
### Before
Screenshot 2022-03-16 at 19 02 02

### After
Screenshot 2022-03-16 at 19 02 09


oliverlloyd (approved):
🟢

PR #4281, opened 2022-03-17
Border colours

### LINKED ISSUE #4152: Border colours for headline and standfirst containers

oliverlloyd:
## What does this change?
This PR fixes #4152 by setting different border colours for different themes

|           | BEFORE | AFTER |
|-----------|--------|-------|
| `News`      |    Screenshot 2022-03-17 at 09 04 24    |   Screenshot 2022-03-17 at 11 27 11  |
| `Opinion`   |   Screenshot 2022-03-17 at 09 09 09  |   Screenshot 2022-03-17 at 11 58 03    |
| `Sport`     |   Screenshot 2022-03-17 at 09 10 03    |    Screenshot 2022-03-17 at 11 55 34   |
| `Lifestyle` |  Screenshot 2022-03-17 at 09 09 39    |   Screenshot 2022-03-17 at 11 54 04   |
| `Culture`   |   Screenshot 2022-03-17 at 09 10 34    |   Screenshot 2022-03-17 at 11 52 11    |

### DIFF
--- a/dotcom-rendering/src/web/lib/decidePalette.ts
+++ b/dotcom-rendering/src/web/lib/decidePalette.ts
@@ -749,45 +749,16 @@ const borderStandfirstLink = (format: ArticleFormat): string => {
 
 const borderHeadline = (format: ArticleFormat): string => {
 	if (format.design === ArticleDesign.LiveBlog) {
-		switch (format.theme) {
-			case ArticlePillar.Culture:
-				return '#716960';
-			case ArticlePillar.Opinion:
-				return '#ae4017';
-			case ArticlePillar.Lifestyle:
-				return '#713b68';
-			case ArticlePillar.Sport:
-				return '#3e657b';
-			case ArticleSpecial.SpecialReport:
-				return specialReport[200];
-			case ArticleSpecial.Labs:
-				return labs[200];
-			case ArticlePillar.News:
-				return '#914b4a';
-		}
+		return 'rgba(255,255,255, 0.2)';
 	}
 	if (format.design === ArticleDesign.DeadBlog) return '#CDCDCD';
 	return border.secondary;
 };
 
 const borderStandfirst = (format: ArticleFormat): string => {
-	if (format.design === ArticleDesign.LiveBlog)
-		switch (format.theme) {
-			case ArticlePillar.Culture:
-				return '#615b53';
-			case ArticlePillar.Opinion:
-				return '#9b3801';
-			case ArticlePillar.Lifestyle:
-				return '#6f3d67';
-			case ArticlePillar.Sport:
-				return '#295873';
-			case ArticleSpecial.SpecialReport:
-				return specialReport[300];
-			case ArticleSpecial.Labs:
-				return labs[300];
-			case ArticlePillar.News:
-				return '#7f3a3a';
-		}
+	if (format.design === ArticleDesign.LiveBlog) {
+		return 'rgba(255,255,255, 0.2)';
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -748,13 +748,46 @@ const borderStandfirstLink = (format: ArticleFormat): string => {
...
+			case ArticlePillar.Lifestyle:
+				return '#713b68';
+			case ArticlePillar.Sport:
+				return '#3e657b';
+			case ArticleSpecial.SpecialReport:
+				return specialReport[200];
+			case ArticleSpecial.Labs:
+				return labs[200];
+			case ArticlePillar.News:
+				return '#914b4a';
```
mxdvl, line 766: Where do these colours come from? Is `decidePalette` the right place to introduce new colour hexes?

mxdvl (approved):
Wow, how did you line up those screenshots so precisely 📏 ?

HarryFischer:
Looks good!

dotcom-rendering/src/web/lib/decidePalette.ts
SiAdcock: I would be cautious adding new colours as it makes colour contrast harder to guarantee at scale. If these borders can be considered decorative, it might not matter. But deciding what is decorative is a non-trivial exercise, and if this question came up every time you make a colour change, it would slow down development.

oliverlloyd: What's the alternative?

These particular colours are tricky to get right as there is this strange effect where the surrounding colour changes how you perceive them so I suspect they are only going to exist in this one place. 

All that said, I'm very happy to adopt a different strategy here but would prefer a staged approach that doesn't block this PR 🙏 as this is going to help us with the liveblog launch (which is coming soon!).

mxdvl: One alternative could be to define a transparent white or transparent black colour, which would introduce less changes, like so:

```ts
		switch (format.theme) {
			case ArticlePillar.Culture:
			case ArticlePillar.Opinion:
			case ArticlePillar.Lifestyle:
				return 'rgba(255,255,255, 0.5)';
			case ArticlePillar.Sport:
			case ArticlePillar.News:
				return 'rgba(0,0,0, 0.5)';
			case ArticleSpecial.SpecialReport:
				return specialReport[200];
			case ArticleSpecial.Labs:
				return labs[200];
```

But I don’t see this as a blocker for this PR.

oliverlloyd: > transparent

Clever! I guess a custom colour is more, well, customisable but maybe we can get what we need using RGB @HarryFischer what do you think?

| Custom hex | RGB 0.5 | RGB 0.3 |
|------------|---------|---------|
|       Screenshot 2022-03-17 at 11 06 25    |   Screenshot 2022-03-17 at 11 06 48      |    Screenshot 2022-03-17 at 11 07 22     |

HarryFischer: @oliverlloyd RGB 0.3 is best for me

oliverlloyd: I refactored everything to use RGB 👍 

@HarryFischer - I went with 0.2 as it seemed to look best over all the different themes but can change it back if you prefer?

### `0.2`
image

PR #4283, opened 2022-03-17
replace more dropdown box with chevron icon from source

DavidLawes:
## What does this change?
Use the chevron icon from Source for the 'more' dropdown link.

## Why?
Fixes a bug raised by Design.

### Before
I think before the 'icon' next to `More` was a rotated box with some borders applied.

Screenshot 2022-03-17 at 11 23 59

### After
Now the icon next to `More` is the chevron icon from Source

Screenshot 2022-03-17 at 11 24 06

### DIFF
--- a/dotcom-rendering/src/web/components/Nav/ExpandedMenu/ShowMoreMenu.tsx
+++ b/dotcom-rendering/src/web/components/Nav/ExpandedMenu/ShowMoreMenu.tsx
@@ -8,6 +8,8 @@ import {
 	brandAlt,
 } from '@guardian/source-foundations';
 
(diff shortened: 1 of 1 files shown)


oliverlloyd (approved):
Code looks good and I also checked this locally without javascript and it's all working for me. Nice 👍

DavidLawes:
> Code looks good and I also checked this locally without javascript and it's all working for me. Nice 👍

Thanks :) Before merging I'm going to deploy to CODE so Harry and I can run through the changes

mxdvl (approved):
Worth noting this code could possibly be simplified with the use of CSS variables.

PR #4286, opened 2022-03-17
Run less tests when not on main

OllysCoding:
## What does this change?

Make PR checks less annoying & inefficient by removing redundant checks that were being run in github actions from our teamcity builds

We keep these checks for `main` branch builds to ensure that in no case a weird merge on githubs part could accidentally publish a broken build to the main site.

## Why?

Faster teamcity builds + less flakyness

### Before

10 minutes 🤬😤😡
image

### After
4 minutes 😇😍🥰
image

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,8 +13,9 @@ jobs:
       - uses: guardian/actions-setup-node@main
         with:
           cache: "yarn"
+      - run: make install
       - name: Lint
-        run: yarn lint
+        run: make lint
         working-directory: dotcom-rendering
       - name: Stylelint
         run: make stylelint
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -1,23 +1,23 @@
...
+        with:
+          cache: "yarn"
+      - run: make install
+        working-directory: dotcom-rendering
+      - name: Lint
+        run: make lint
+        working-directory: dotcom-rendering
+      - name: Stylelint
+        run: make stylelint
+        working-directory: dotcom-rendering
```
OllysCoding, line 23: Adding stylelint here means we can skip it in team city (and makes more sense)

oliverlloyd (approved):
Why did we not do this sooner! Thank you for implementing this!

PR #4288, opened 2022-03-17
feat: Force Core Web Vitals sampling for users in certain AB tests

zekehuntergreen:
## What does this change?
- Moves the initialization of core web vitals sampling into a component inside of an island on `Page.tsx`, following the pattern of [commercial metrics](https://github.com/guardian/dotcom-rendering/blob/d161006ef20c8c70b7f42c7d9709d105937c461d/dotcom-rendering/src/web/components/CommercialMetrics.importable.tsx#L6)
- Defines a currently-empty list of `ABTest`s for which, if the user is a participant, we will bypass sampling and always send core web vitals data to the data lake.

## Why?
- For some AB tests which we expect to effect [core web vitals](https://web.dev/vitals/), we'd like to be able to increase the rate at which we sample these metrics from the default 1% so that significant changes in CWVs can be detected with smaller test samples.
- On frontend, we have the ability to sample core web vitals (and [commercial metrics](https://github.com/guardian/frontend/blob/06165f9ead7a86b565b1683b266ef3d26a869f91/static/src/javascripts/projects/commercial/commercial-metrics.ts#L31)) data [at a rate of to 100%](https://github.com/guardian/frontend/blob/361c3c6a11e05397879107fb13153e1b0df6e225/static/src/javascripts/projects/common/modules/analytics/coreVitals.ts#L22) for users in certain AB tests. This pull request implements that same functionality in DCR.


oliverlloyd:
Will this approach will work given the fact this lib [needs to be loaded prior](https://github.com/GoogleChrome/web-vitals#how-the-polyfill-works) to any content being rendered?

This raises interesting questions about when an island actually runs and when it _should_ run. Maybe the code already executes ahead of the render? I'm not sure this is guaranteed though, not in the way that it is with script tags in the head.

mxdvl:
> Will this approach will work given the fact this lib [needs to be loaded prior](https://github.com/GoogleChrome/web-vitals#how-the-polyfill-works) to any content being rendered?

I think this is only true of the polyfill, and I’m not sure we’re providing it currently.

dotcom-rendering/src/web/components/Page.tsx
```diff
@@ -61,6 +62,13 @@ export const Page = ({ CAPI, NAV, format }: Props) => {
 					isDev={CAPI.config.isDev}
 				/>
 			</Island>
+			<Island clientOnly={true} deferUntil="idle">
```
mxdvl, line 65: Deferring until idle seems 👌

mxdvl (approved):
That’s a nice approach: Islands are the new scripts.

dotcom-rendering/src/web/components/Page.tsx
mxdvl: I think it’s all reading [from the `Performance` API](https://developer.mozilla.org/en-US/docs/Web/API/Performance). Indeed if it’s not available you’d need polyfill.

zekehuntergreen: Thanks both!
Testing in CODE, initialising the listeners slightly later doesn't seem to impact our ability to collect CWV metics.
I'll monitor after deploying to make sure we don't see a dip in the number of pageviews with this data.

PR #4299, opened 2022-03-17
Remove Unnecessary Fragment

JamieB-gu:
## Why?

I don't think it's necessary to wrap this component in a fragment. I think the fragment was introduced in #1740 to put a sign-in gate placeholder alongside (back when this code was in `ArticleRenderer`), but this placeholder was removed in #1819.

## Changes

- Removed enclosing fragment


JamieB-gu:
In the end this was fixed in #4073. I'll close this PR, but hopefully the description will remain useful for future reference.

PR #4287, opened 2022-03-17
feat: Explicit Island chunk names

mxdvl:
## What does this change?

Previously, island chunks generated by Webpack were numeric IDs like `8873`. Now they reflect the actual filenames.

## Why?

Easier to understand the impact of changes are reported by the compress action.

### Before

```sh
ls dist/**.js

dist/1189.9b78234a9be174bcb95f.js
dist/1214.0c3439950f252b3450f3.js
dist/1376.e09de2b0edfb5b7663fb.js
dist/1451.8f29eca0e0395fb89cf0.js
dist/1578.bbbaf25c73420bfce5d8.js
dist/1828.aaabd9914f2832062f6d.js
dist/1924.8dbea87ab294d577cff0.js
dist/1956.af85a4429398e3e6a127.js
```

### After

```sh
ls dist/**.js

dist/AlreadyVisited-importable.dd6db9d38cd8b7efda3f.js
dist/AudioAtomWrapper-importable.898a1bc9bdefdd656c63.js
dist/Branding-importable.6f67ab9266fc97b4d7d3.js
dist/BrazeMessaging-importable.67b10c92553cac70d24b.js
dist/CalloutBlockComponent-importable.6c42df120ac2da56e508.js
dist/ChartAtomWrapper-importable.7a8ea5381c29db6c9088.js
dist/CommentCount-importable.5210ae6d30fe4b42eb46.js
dist/CommercialMetrics-importable.fb68b72be920cdff5150.js
# etc.
```

PR #4305, opened 2022-03-18
Refactor Action: Preact compressed size

mxdvl:
## What does this change?

Set a threshold for compressed size change for a file to qualify as different–100 bytes.
Strip the file hash directly in the action, allowing the same build in GHA as we do in TeamCity.

The action will report a change wrong in this PR, because the file names are changing:

`dist/Branding-importable.js` &rarr; `dist/Branding-importable.********************.js`

## Why?

There can be a lot of noise for minuscule changes with this action, and since #4287, we have gained more visibility on what has actually changed.

Helps pave the way for a full GHA delivery, as the build is identical.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -20,3 +20,4 @@ indent_style = tab
 
 [*.yml]
 indent_size = 2
+indent_style = space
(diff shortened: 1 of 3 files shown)


ioannakok (approved):
Looks good to me 👍 Thanks for the good description and the doc links provided!

PR #4312, opened 2022-03-18
Fix getArticleCountToday

tomrf1:
Currently this function might return e.g. yesterday's count if the browser hasn't yet landed on an article

[frontend change](https://github.com/guardian/frontend/pull/24784)

### DIFF
--- a/dotcom-rendering/src/web/lib/dailyArticleCount.ts
+++ b/dotcom-rendering/src/web/lib/dailyArticleCount.ts
@@ -36,7 +36,7 @@ export const getDailyArticleCount = (): DailyArticleHistory | undefined => {
 	}
 };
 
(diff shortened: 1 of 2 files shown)


dotcom-rendering/src/web/lib/dailyArticleCount.ts
```diff
@@ -36,12 +36,14 @@ export const getDailyArticleCount = (): DailyArticleHistory | undefined => {
 	}
 };
 
+export const getToday = (): number => Math.floor(Date.now() / 86400000);
```
mxdvl, line 39: You can use underscores to make large numbers easier to parse!

```suggestion
export const getToday = (): number => Math.floor(Date.now() / 86_400_000);
```

mxdvl (approved):

PR #4313, opened 2022-03-18
Rename `dcr:page:article:redisplayed`

simonbyford:
## What does this change?

Renames the event `dcr:page:article:redisplayed` to `article:sign-in-gate-dismissed`

## Why?

`dcr:page:article:redisplayed` didn't capture what was actually happening. It was a hangover from the old `mediator` event name


oliverlloyd (approved):

mxdvl:
No one likes a hangover!

PR #4317, opened 2022-03-18
chore: Islands style improvements

mxdvl:
### Fixes
Fix JSDoc syntax usage for Island.tsx, [to match docs](https://jsdoc.app/tags-param.html#parameters-with-properties).

### Refactors
Use idiomatic `never` to omit prop, [as per the docs](https://www.typescriptlang.org/docs/handbook/2/functions.html#never).

~~Make `doHydration` explicitly asynchronous. We now have to ignore these floating promises in `initHydration`, which is more explicit about what we actually doing.~~

## Why?

There’s some non-idiomatic usage that we can fix.

### DIFF
--- a/dotcom-rendering/src/web/browser/islands/doHydration.ts
+++ b/dotcom-rendering/src/web/browser/islands/doHydration.ts
@@ -14,7 +14,11 @@ import { initPerf } from '../initPerf';
  * @param data The deserialised props we want to use for hydration
  * @param element The location on the DOM where the component to hydrate exists
  */
-export const doHydration = (name: string, data: any, element: HTMLElement) => {
+export const doHydration = async (
+	name: string,
+	data: any,
+	element: HTMLElement,
+) => {
 	// If this function has already been run for an element then don't try to
 	// run it a second time
 	const alreadyHydrated = element.dataset.guReady;
@@ -22,7 +26,7 @@ export const doHydration = (name: string, data: any, element: HTMLElement) => {
 
 	const { start, end } = initPerf(`hydrate-${name}`);
 	start();
-	import(
+	return import(
 		/* webpackInclude: /\.importable\.tsx$/ */
 		/* webpackChunkName: "[request]" */
 		`../../components/${name}.importable`
--- a/dotcom-rendering/src/web/browser/islands/doHydration.ts
+++ b/dotcom-rendering/src/web/browser/islands/doHydration.ts
@@ -14,11 +14,7 @@ import { initPerf } from '../initPerf';
  * @param data The deserialised props we want to use for hydration
  * @param element The location on the DOM where the component to hydrate exists
  */
-export const doHydration = async (
-	name: string,
-	data: any,
-	element: HTMLElement,
-) => {
+export const doHydration = (name: string, data: any, element: HTMLElement) => {
 	// If this function has already been run for an element then don't try to
 	// run it a second time
 	const alreadyHydrated = element.dataset.guReady;
@@ -26,7 +22,7 @@ export const doHydration = async (
 
 	const { start, end } = initPerf(`hydrate-${name}`);
 	start();
-	return import(
+	import(
 		/* webpackInclude: /\.importable\.tsx$/ */
 		/* webpackChunkName: "[request]" */
 		`../../components/${name}.importable`
--- a/dotcom-rendering/src/web/components/Island.tsx
+++ b/dotcom-rendering/src/web/components/Island.tsx
@@ -7,7 +7,7 @@ type When = 'idle' | 'visible';
 interface HydrateProps {
 	deferUntil?: When;
 	clientOnly?: false;
-	placeholderHeight?: undefined;
+	placeholderHeight?: never;
 	children: JSX.Element;
 }
 
--- a/dotcom-rendering/src/web/components/Island.tsx
+++ b/dotcom-rendering/src/web/components/Island.tsx
@@ -42,13 +42,13 @@ const decideChildren = (
  * Note. The component passed as children must follow the [MyComponent].importable.tsx
  * namimg convention
  *
- * @param deferUntil - Delay when client code should execute
+ * @param {HydrateProps | ClientOnlyProps} props - JSX Props
+ * @param {When} props.deferUntil - Delay when client code should execute
  * 		- idle - Execute when browser idle
  * 		- visible - Execute when component appears in viewport
- * @param clientOnly - Should the component be server side rendered
- * @param placeholderHeight - The height for the placeholder element
- * @param children - The component being inserted. Must be a single React Element
- *
+ * @param {boolean} props.clientOnly - Should the component be server side rendered
+ * @param {number} props.placeholderHeight - The height for the placeholder element
+ * @param {JSX.Element} props.children - The component being inserted. Must be a single JSX Element
  */
 export const Island = ({
 	deferUntil,
--- a/dotcom-rendering/src/web/browser/initDiscussion/init.ts
+++ b/dotcom-rendering/src/web/browser/initDiscussion/init.ts
@@ -7,9 +7,9 @@ import { getName } from '../islands/getName';
 function forceHydration() {
 	try {
 		// Select the Discussion island element
-		const guElement = document.querySelector(
+		const guElement = document.querySelector<HTMLElement>(
 			'gu-island[name=DiscussionContainer]',
-		) as HTMLElement;
+		);
 		if (!guElement) return;
 
 		// Read the props from where they have been serialised in the dom using an Island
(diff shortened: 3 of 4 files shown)


dotcom-rendering/src/web/browser/islands/doHydration.ts
```diff
@@ -14,40 +14,45 @@ import { initPerf } from '../initPerf';
...
-			if (clientOnly) {
-				element.querySelector('[data-name="placeholder"]')?.remove();
-				log('dotcom', `Rendering island ${name}`);
-				render(h(module[name], data), element);
-			} else {
-				log('dotcom', `Hydrating island ${name}`);
-				hydrate(h(module[name], data), element);
-			}
+	try {
+		const module = await import(
```
oliverlloyd, line 31: We don't use this pattern anywhere else when we import. Is there a particular problem it's solving here or is this just code style preference?

mxdvl: more of a question, really… I’m curious as to why we don’t.

oliverlloyd: I think people just follow idiomatic conventions. I looked at the first 8 results from Google on how to use dynamic imports and none of them suggested `await` 🤷

dotcom-rendering/src/web/components/Island.tsx
```diff
@@ -7,7 +7,7 @@ type When = 'idle' | 'visible';
 interface HydrateProps {
 	deferUntil?: When;
 	clientOnly?: false;
-	placeholderHeight?: undefined;
+	placeholderHeight?: never;
```
jamesgorrie, line 10: Would removing unusable props make more sense?

```ts
interface ServerSideRenderedProps {
	children: JSX.Element;
}

interface ClientOnlyProps {
	clientOnly: true;
	deferUntil?: When;
	placeholderHeight?: number;
	children: JSX.Element;
}

type Props = ServerSideRenderedProps | ClientOnlyProps
```

jamesgorrie (commented):
As we never want to the values from either `forceHydration` or `doHydration`, what is the value of waiting for the Promises to resolve? I think the current way is more explicit that there is nothing you can nor should be waiting for.

oliverlloyd:
> already returning a promise

I don't think it's returning anything at the moment

oliverlloyd:
> we wanted to measure how often and quickly we load all of the page’s Islands

This is already happening inside the function call. We use `initPerf` for this

mxdvl:
> > already returning a promise
> 
> I don't think it's returning anything at the moment

My bad–I meant the `init` method that calls `forceHydration` returns a promise:

```ts
const init = (): Promise<void> => {
  /* does async stuff but ignores it */
  return Promise.resolve()
}
```

oliverlloyd:
> I meant the `init` method

Yeah, that is a pattern we've had a while now. If you look at all the other init functions they all follow the same structure, returning this `return Promise.resolve();`. I think there was a discussion on it ages back

mxdvl:
> > I meant the `init` method
> 
> Yeah, that is a pattern we've had a while now. If you look at all the other init functions they all follow the same structure, returning this `return Promise.resolve();`. I think there was a discussion on it ages back

Some are using it properly, some are not… so currently `initPerf` is only partly useful.

- `bootCmp`: 😀 does it properly, it’s an `async` function, and it only resolves when actually loaded
- `updateIframeHeight`: 😐  also a real async function, but the function it calls returns a fake `Promise.resolve()`. I’d wager you gain nothing from measuring something you’re not waiting for.
- `ophan`: 😀 this only runs synchronous code, so it’s valid to say that the promise can resolve at the end
- `updateTimeElement`: 😢  there’s a synchronous part to this code, but it has side-effect we don’t wait for
- `sentry`: 😃 this is also really async, and we wait for all sorts of stuff.

dotcom-rendering/src/web/browser/islands/doHydration.ts
mxdvl: Counterpoint to this, we have 3 instances in the codebase where we use `await import`:

```
dotcom-rendering/src/web/browser/sentryLoader/init.ts:
  48  			}
  49  
  50  			// Load sentry.ts
  51: 			const { reportError } = await import(
  52  				/* webpackChunkName: "sentry" */ './sentry'
  53  			);
  54  

dotcom-rendering/src/web/components/YoutubeBlockComponent.importable.tsx:
  91  
  92  	useEffect(() => {
  93  		const defineConsentState = async () => {
  94: 			const { onConsentChange } = await import(
  95  				'@guardian/consent-management-platform'
  96  			);
  97  			onConsentChange((newConsent: ConsentState) => {

dotcom-rendering/src/web/lib/braze/initialiseAppboy.ts:
  11  };
  12  
  13  const initialiseAppboy = async (apiKey: string): Promise<typeof appboy> => {
  14: 	const importedAppboy = (await import(
  15  		/* webpackChunkName: "braze-web-sdk-core" */ '@braze/web-sdk-core'
  16  	)) as unknown as typeof appboy;
  17  
```

dotcom-rendering/src/web/components/Island.tsx
mxdvl: This wouldn’t work with Typescript, as we need overlap between the two types to set it as the props of Island’s first de-structured argument.

oliverlloyd: Yeah. I actually think `never` is kind of lovely. I mean, it would be _better_ to just not have the prop but TS doesn't like that so in this case Max's suggestion makes so much more sense than what was there before

oliverlloyd:
> > > I meant the `init` method
> > 
> > 
> > Yeah, that is a pattern we've had a while now. If you look at all the other init functions they all follow the same structure, returning this `return Promise.resolve();`. I think there was a discussion on it ages back
> 
> Some are using it properly, some are not… so currently `initPerf` is only partly useful.
> 
> * 😀 `bootCmp`:  does it properly, it’s an `async` function, and it only resolves when actually loaded
> * 😐 `updateIframeHeight`:   also a real async function, but the function it calls returns a fake `Promise.resolve()`. I’d wager you gain nothing from measuring something you’re not waiting for.
> * 😀 `ophan`:  this only runs synchronous code, so it’s valid to say that the promise can resolve at the end
> * 😢 `updateTimeElement`:   there’s a synchronous part to this code, but it has side-effect we don’t wait for
> * 😃 `sentry`:  this is also really async, and we wait for all sorts of stuff.
> * 😃 `updateIframeHeight`: this does synchronous stuff, looks fine to me
> * 😃 `shimport`: returns an actual promise
> * 😢 `initHydration`: the promise isn’t connected to the underlying logic
> * _and more…_

Love the emoji usage here. And I think you're making a really fair point. There's not an organised approach at all.

It feels like the way we're doing perf metrics here generally could be a lot better. Are we even surfacing this data anywhere? I know we log it locally but that's not very useful. Could this data be rationalised and reported to the perf endpoint?

mxdvl:
Seeing no clear consensus on the benefit of making the async nature of our hydration process more explicit, I will keep this for another PR.

In the meantime, I’ll start looking at how we can improve our performance measurements and start measuring them. No doubt we’ll need to revisit islands then.

jamesgorrie:
Apologies for the last comment, I wasn't sure if this was being done because it was neater, but realise there might be a use case for having returned values from an `init` method.

This feels like one of those areas where we are creating an abstraction, and an abstraction that you would want to be able to tell our future selves.

I'm trying to outline why these init scripts exist, and how they work. I'm going to start with how they work, as I'm not sure why they exist:

These are small fragments of JS. They are wrapped in [a startup](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/browser/startup.ts) script to measure performance.

They're scripts, that get [generated into bundles via webpack](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/scripts/webpack/webpack.config.browser.js).

They're then included in the body as `<script>` tags as either low or high priority. These are defined in [`document.tsx`](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/server/document.tsx#L114-L143) and rendered in `htmlTemplate.tsx`, [high](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/server/htmlTemplate.ts#L286) and [low](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/server/htmlTemplate.ts#L295).

So it seems the problem we're trying to solve is people want script tags on our page.

As a platform, which tries to make this easier for people to work on our platform, it might be worth outlining a few things that stand out with the current setup:

1. We need the developer to remember to add performance tracking to their script
2. If they do add this, I am not 100% certain where I might look for those metrics
3. I need to remember to add it to the high/low priority array in a piece of code that wasn't to intrinsic to find
4. I'm not really sure when I should use this approach over another e.g. Just add a Component

Quickly thinking - if I were a dev, I'd want to:
* read the "How do I do X? Use the script loading thing"
* 'write my script
* add it to low/high priority array

I can't say I understand fully what X is - maybe you could help me figure out when we should be using this script loading?

ioannakok (approved):
LGTM!

PR #4318, opened 2022-03-18
refactor: extract island name as const

mxdvl:
## What does this change?

Explicitly set the name we expect for this Island.

## Why?

We want an island with a specific name, `DiscussionContainer`. There’s no point parsing the `gu-island` element to check if it has this name, as that’s how it’s selected in the first place.

### DIFF
--- a/dotcom-rendering/src/web/browser/initDiscussion/init.ts
+++ b/dotcom-rendering/src/web/browser/initDiscussion/init.ts
@@ -2,22 +2,19 @@ import '../webpackPublicPath';
 import { startup } from '../startup';
 import { getProps } from '../islands/getProps';
 import { doHydration } from '../islands/doHydration';
-import { getName } from '../islands/getName';
 
 function forceHydration() {
 	try {
+		const name = 'DiscussionContainer';
+
 		// Select the Discussion island element
 		const guElement = document.querySelector<HTMLElement>(
-			'gu-island[name=DiscussionContainer]',
+			`gu-island[name=${name}]`,
 		);
 		if (!guElement) return;
 
 		// Read the props from where they have been serialised in the dom using an Island
 		const props = getProps(guElement);
-		const name = getName(guElement);
-
-		// Ensure we have name
-		if (!name) return;
 
 		// Now that we have the props as an object, tell Discussion we want it to expand itself
 		props.expanded = true;


dotcom-rendering/src/web/browser/initDiscussion/init.ts
```diff
@@ -2,22 +2,19 @@ import '../webpackPublicPath';
...
-import { getName } from '../islands/getName';
 
 async function forceHydration() {
 	try {
+		const name = 'DiscussionContainer';
+
 		// Select the Discussion island element
 		const guElement = document.querySelector<HTMLElement>(
-			'gu-island[name=DiscussionContainer]',
+			`gu-island[name=${name}]`,
```
oliverlloyd, line 12:
```suggestion
			`gu-island[name=DiscussionContainer]`,
```

dotcom-rendering/src/web/browser/initDiscussion/init.ts
```diff
@@ -2,22 +2,19 @@ import '../webpackPublicPath';
 import { startup } from '../startup';
 import { getProps } from '../islands/getProps';
 import { doHydration } from '../islands/doHydration';
-import { getName } from '../islands/getName';
 
 async function forceHydration() {
 	try {
+		const name = 'DiscussionContainer';
+
```
oliverlloyd, line 9:
```suggestion
```

dotcom-rendering/src/web/browser/initDiscussion/init.ts
oliverlloyd: Might as well inline it as it's already called 'name'?

oliverlloyd (approved):
Nice!

dotcom-rendering/src/web/browser/initDiscussion/init.ts
oliverlloyd: Oh wait, this is what it already was?! Eh?

mxdvl: We need the `name` variable to `doHydration`.

Previously we said: pick an element that has this name, then find the name of this element. This was quite redundant, required more work and was more obscure about what we’re trying to achieve. It also had an extra check if a name was missing… but this could not happen if the `querySelector` failed.

oliverlloyd: I just meant you can inline it, there's no need for the named var as it's _already_ called 'name='

dotcom-rendering/src/web/browser/initDiscussion/init.ts
mxdvl: We need it for `doHydration` on L23

dotcom-rendering/src/web/browser/initDiscussion/init.ts
oliverlloyd: Okay, we discussed offline and I totally missed there was a second ref to this string so the var makes sense. All good!

PR #4342, opened 2022-03-21
Fix padding of liveblocks on mobile

iainjchambers-guardian:
## What does this change?
Adds a container around the liveblocks on the live layout for AR.

## Why?
The padding on liveblocks was causing an overflow and the pagination was not padded at all so it was flush with the screen edge.

| Before |  After |
| -------|-------|
| image | image |

### DIFF
--- a/apps-rendering/src/components/layout/live.tsx
+++ b/apps-rendering/src/components/layout/live.tsx
@@ -8,7 +8,13 @@ import { Pagination } from '@guardian/common-rendering/src/components/Pagination
 import { background } from '@guardian/common-rendering/src/editorialPalette';
 import type { ArticleFormat } from '@guardian/libs';
 import { ArticleDesign } from '@guardian/libs';
-import { from, neutral, news, remSpace } from '@guardian/source-foundations';
+import {
+	from,
+	neutral,
+	news,
+	pxToRem,
+	remSpace,
+} from '@guardian/source-foundations';
 import { OptionKind } from '@guardian/types';
 import Footer from 'components/footer';
 import GridItem from 'components/gridItem';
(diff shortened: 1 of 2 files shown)


apps-rendering/src/components/layout/live.tsx
```diff
@@ -148,12 +148,23 @@ const Live: FC<Props> = ({ item }) => {
...
-						format={item}
-					/>
-					{pagination}
+					<div
+						css={css`
+							padding: 10px 10px 0 10px;
+
+							${from.mobileLandscape} {
+								padding: 0;
+							}
```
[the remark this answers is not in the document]
iainjchambers-guardian: By that do you mean just change the breakpoint to `${from.tablet}`?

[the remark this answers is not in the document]
JamieB-gu:
```suggestion
							${from.mobileLandscape} {
								padding-left: ${remSpace[5]};
								padding-right: ${remSpace[5]};
							}

							${from.tablet} {
								padding: 0;
							}
```

common-rendering/src/components/liveBlockContainer.tsx
```diff
@@ -141,6 +142,14 @@ const LiveBlockContainer = ({
 				}
 				border-bottom: 1px solid ${neutral[86]};
 
+				&:first-child {
+					margin-top: ${space[3]}px;
```
JamieB-gu, line 146: Heads up: I can see from Chromatic that this has created additional spacing on the DCR live blocks too. Was that intentional?

iainjchambers-guardian: No I was just trying to keep the margins on the liveblocks in the same place but I'll just add padding to the container on AR

apps-rendering/src/components/layout/live.tsx
```diff
@@ -148,12 +155,35 @@ const Live: FC<Props> = ({ item }) => {
...
-					<LiveBlocks
-						blocks={item.pagedBlocks.currentPage.blocks}
-						format={item}
-					/>
-					{pagination}
+					<div
+						css={css`
+							padding-left: ${pxToRem(10)}rem;
+							padding-right: ${pxToRem(10)}rem;
+							padding-top: ${space[3]}px;
```
JamieB-gu, line 162:
```suggestion
							padding-top: ${remSpace[3]};
```

JamieB-gu (approved):
One suggestion, otherwise looks good 👍

PR #4344, opened 2022-03-21
Use browserId for banner targeting

tomrf1:
Send `browserId` to `support-dotcom-components` for banner targeting, if the user has consented. This will mean we can run tests [using propensity models](https://github.com/guardian/support-dotcom-components/pull/658).
We already [send browserId for the epic](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/SlotBodyEnd/ReaderRevenueEpic.tsx#L96).


OllysCoding (approved):
+1 Looks good!!

PR #4347, opened 2022-03-21
Add bylines to blocks

MarSavar:
## Why?
Bylines weren't being rendered in blocks in AR. This PR addresses the issue by adding an array of `contributors` to `LiveBlock`. Figuring this out wasn't straightforward and involved some digging in the `frontend` codebase - thanks @JamieB-gu for pairing on this.

| **Before** | **After** |
|-----------|----------|
| image | image |

## Follow-up work
1) Bylines appear to be larger than `36x36`, despite using the same component as DCR. We need to investigate why this is.
2) Byline images currently do not display a background colour, unlike DCR. This needs to be fixed.

### DIFF
--- a/apps-rendering/src/liveBlock.ts
+++ b/apps-rendering/src/liveBlock.ts
@@ -1,12 +1,13 @@
 // ----- Imports ----- //
 
 import type { Block } from '@guardian/content-api-models/v1/block';
-import { Tag } from '@guardian/content-api-models/v1/tag';
+import type { Tag } from '@guardian/content-api-models/v1/tag';
 import type { Option } from '@guardian/types';
 import type { Body } from 'bodyElement';
 import { parseElements } from 'bodyElement';
 import { maybeCapiDate } from 'capi';
-import { Contributor, tagToContributor } from 'contributor';
+import type { Contributor } from 'contributor';
+import { tagToContributor } from 'contributor';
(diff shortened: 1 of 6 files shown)


apps-rendering/src/liveBlock.ts
```diff
@@ -16,25 +19,42 @@ type LiveBlock = {
...
 // ----- Functions ----- //
 
+const contributorTags = (contributors: string[], tags: Tag[]): Tag[] => {
+	const isTag = (tag: Tag | undefined): tag is Tag => tag !== undefined;
+	return contributors
+		.map((contributor) =>
+			tags.find((tag) => tag.id === `profile/${contributor}`),
+		)
+		.filter(isTag);
+};
```
JamieB-gu, line 34: FYI the `isTag` function could be written in a generic way to make it useful in other circumstances:

```ts
const isDefined = <A>(a: A | undefined): a is A => a !== undefined;
```

I think replacing `isTag` with `isDefined` gives you the same result, but `isDefined` could be moved to `lib.ts` and prove useful in similar scenarios.

Not necessary for this PR, but might make a nice optimisation?

MarSavar: Great suggestion, I'll put it in the backlog so as not to block this PR. 
I'll also card up the 2 follow-up points from the PR description.

PR #3718, opened 2021-12-02
Picture/Image Optimisations - Desktop immersive main-media  (Part 2/?)  

OllysCoding:
image
## TL;DR 📚

The solution we used for solving "The DPR Problem" in #3641 (knowingly) introduced a regression in quality for Immersive Main Media on high resolution devices.

This PR introduces a new way of creating `<source />` elements for immersive main-media pictures to render source elements by their source-sets rather than by breakpoints. This prevents the resolutions of images available being limited to confines of our breakpoint range.

## The "Desktop immersive main-media" Problem  🖥

The "desktop main-media" problem is a regression in the quality of images on larger / high resolution desktop devices, due to how we now render sources.

Imagine we have an array of sources with the following sizes:
```[400px, 700px, 1000px, 1300px, 1900px]```

Previously, we'd place all of these sources within the `srcset` attribute, and allow the browser to pick the most appropriate image source. This was great in that it made all of our image sources available to the browser, but not so good in that it allowed the browser to use [DPR to select the wrong image
](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/docs/architecture/027-pictures.md#the-dpr-problem).

The current solution, which solves the DPR problem, will only select 1 matching source per breakpoint. This works well for fixed size images, but not so well for dynamic images like immersive main media. Imagine our breakpoints look like:
```[320px, 660px, 980px, 1300px]```

When we're matching up breakpoints to sources for an immersive main media we get:
```js 
[
    // breakpoint : source
    320px: 400px,
    700px: 700px,
    980px: 1000px,
    1300px: 1300px
]
```

This works well, but what happened to our 1900px source? Since we only chose sources in the confines of our breakpoints, it gets lost.

The solution to this is then to skip the breakpoints entirely when an image is fully dynamic, and instead just loop over our sources to create the elements on the page:

```js
// before
sourcesMatchedToBreakpoints.map(sourceSet => <source ... >)

// after
ourSorces.map(sourceSet => <source ... >)
```

### Final Before & After
Before:
```html
<picture itemprop="contentUrl">
    /* Portrait immersive sources have been removed */
    <source srcset="https://i.guim.co.uk/img/media/image/4383.jpg?width=1300&amp;quality=45&amp;auto=format&amp;fit=max&amp;dpr=2 2600w" media="(min-width: 1300px) and (-webkit-min-device-pixel-ratio: 1.25), (min-width: 1300px) and (min-resolution: 120dpi)">
[...]

PR #4353, opened 2022-03-22
better quality immersive main media images

oliverlloyd:
## What does this change?
This PR replaces the previous, perfectly valid and actually quite clever, logic to decide which image to use when a `Picture` was being used for immersive main media with a basic map which is forcing larger images to be picked.

## Why?
Ultimately, the source sets provided are not good enough. We're being given images to be used at 1300 which [sometimes](https://www.theguardian.com/society/2021/feb/11/pure-liquid-hope-what-coronavirus-covid-vaccine-means-to-a-gp-doctor) [don't look](https://www.theguardian.com/news/2018/dec/07/china-plan-for-global-media-dominance-propaganda-xi-jinping) [that great](https://www.theguardian.com/news/2021/apr/27/the-clockwork-universe-is-free-will-an-illusion) at this width. The previous logic respected the source sets, expecting the images to look okay at the given widths but after several reports of poor quality images it is now clear that this is not the case.

Frontend was already [doing something similar](https://github.com/guardian/frontend/blob/main/common/app/layout/ContentWidths.scala#L108) so this PR takes that old override logic and implements it in DCR.

### DIFF
--- a/dotcom-rendering/src/web/components/Picture.tsx
+++ b/dotcom-rendering/src/web/components/Picture.tsx
@@ -131,6 +131,27 @@ const getDesiredWidthForBreakpoint = (
 		return breakpoint;
 	}
 
+	if (isMainMedia && format.display === ArticleDisplay.Immersive) {
+		// Immersive main media images are often reported to have poor quality so we
+		// override the default values here and force larger images to be used
+		switch (breakpoint) {
+			case breakpoints.mobile:
+				return 480;
+			case breakpoints.mobileLandscape:
+				return 660;
+			case breakpoints.phablet:
+				return 740;
+			case breakpoints.tablet:
+				return 980;
+			case breakpoints.desktop:
+				return 1140;
+			case breakpoints.leftCol:
+				return 1300;
+			case breakpoints.wide:
+				return 1900;
+		}
+	}
+
 	switch (role) {
 		case 'inline':
 			if (
@@ -177,7 +198,7 @@ const getSourceForDesiredWidth = (
 };
 
 // Create sourcesets for portrait immersive
-// TODO: In a future PR this system will be updated to solve scaling issues with DPR
+// In a future PR this system could be updated to address scaling issues with DPR
 const portraitImmersiveMainMediaSource = (
 	sources: SrcSetItem[],
 	resolution: ResolutionType,
@@ -200,41 +221,6 @@ const portraitImmersiveMainMediaSource = (
 			.join(',')}
 	/>
 );
-
-// Create sourcesets for landscape immersive
-const landscapeImmersiveMainMediaSource = (
-	sources: SrcSetItem[],
-	resolution: ResolutionType,
-) => {
-	const sortedSources = sources.slice().sort((a, b) => b.width - a.width);
-
-	const getMedia = (sourceWidth: number, isLast: boolean): string => {
-		const width = resolution === 'hdpi' ? sourceWidth / 2 : sourceWidth;
-		const breakpoint = isLast ? 0 : width;
-
-		return resolution === 'hdpi'
-			? `(min-width: ${breakpoint}px) and (-webkit-min-device-pixel-ratio: 1.25), (min-width: ${breakpoint}px) and (min-resolution: 120dpi)`
-			: `(min-width: ${breakpoint}px)`;
-	};
-
-	return (
-		<>
-			{/* For immersive main media, images will always take up 100% of the screen width,
-				therefore it is more beneficial to loop through all the image sources, instead of the breakpoints.
-				This means that we are not limited to sources within the boundaries of the breakpoints, and can provide
-				more appropriately sized images where applicable. */}
-			{sortedSources.map((sourceSet, index) => (
-				<source
-					srcSet={`${sourceSet.src} ${sourceSet.width}w`}
-					media={getMedia(
-						sourceSet.width,
-						index === sortedSources.length - 1,
-					)}
-				/>
-			))}
-		</>
-	);
-};
 
 export const Picture = ({
 	imageSources,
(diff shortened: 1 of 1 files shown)


mxdvl:
I find it hard to judge the quality of pictures, but JPEG is notorious for destroying illustrations with big swaths of colours, and especially the blue channel.

If the sizes are picked correctly, could this be more about our `q=45` setting, or ensuring the illustrations are authored in a lossless format like PNG?

oliverlloyd:
> I find it hard to judge the quality of pictures, but JPEG is notorious for destroying illustrations with big swaths of colours, and especially the blue channel.
> 
> If the sizes are picked correctly, could this be more about our `q=45` setting, or ensuring the illustrations are authored in a lossless format like PNG?

Yes, I believe you are correct in that the source sets provided are not good enough. And you may well also be right that the reason for this lies somewhere in how we set `quality`. This PR is explicitly acknowledging that this is the case and making a change to how we decide widths to improve the experience for users based on this knowledge.

mxdvl (commented):
I still don’t understand _how_ that picks larger images. The power of `landscapeImmersiveMainMediaSource` was that it could let the browser pick images in between two defined breakpoints.

Could you show a comparison of resulting generated `picture` tag for comparison?

oliverlloyd:
> Could you show a comparison of resulting generated `picture` tag for comparison?

There's some diffs in Chromatic on this PR which show new images being picked and also the gif but essentially instead of choosing 1300 it picks 1900.

> In what way are they not good enough? Too few? Too many? Just plain wrong?

There's an email chain from the US Long Reads editor saying how the images are of poor quality leading to 💔 . It seems the quality of 1300 is not good enough.

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -131,6 +131,27 @@ const getDesiredWidthForBreakpoint = (
...
+				return 740;
+			case breakpoints.tablet:
+				return 980;
+			case breakpoints.desktop:
+				return 1140;
+			case breakpoints.leftCol:
+				return 1300;
+			case breakpoints.wide:
+				return 1900;
+		}
```
mxdvl, line 152: It wasn’t clear to me that what we were doing here is getting the larger breakpoints.

Instead of semi-magic numbers, could we make it more explicit like:

```suggestion
		switch (breakpoint) {
			case breakpoints.mobile:
				return breakpoints.mobileLandscape;
			case breakpoints.mobileLandscape:
				return breakpoints.phablet;
			case breakpoints.phablet:
				return breakpoints.tablet;
			case breakpoints.tablet:
				return breakpoints.desktop;
			case breakpoints.desktop:
				return breakpoints.leftCol;
			case breakpoints.leftCol:
				return breakpoints.wide;
			case breakpoints.wide:
				return 1900;
		}
```

Or alternatively do an actual review of the perfect values and stray away from breakpoints

```suggestion
		switch (breakpoint) {
			case breakpoints.mobile:
				return 500;
			case breakpoints.mobileLandscape:
				return 700;
			case breakpoints.phablet:
				return 800;
			case breakpoints.tablet:
				return 1000;
			case breakpoints.desktop:
				return 1200;
			case breakpoints.leftCol:
				return 1300;
			case breakpoints.wide:
				return 1900;
		}
```

mxdvl (approved):
I got quite confused by the logic as I didn’t realise the breakpoints were mapping to the next breakpoint up. Have proposed to used less magical numbers below

OllysCoding:
Though I'd be sad to see my code go, this does essentially solve the problem; and do a better job at respecting frontends 'initial' wishes on how image sources should be used (Frontend actually specifies breakpoints in its source-set definitions, but those definitions get lost in the dataset for DCR):

image

---

As an argument for keeping my code though, I did actually anticipate this exact problem, and have a solution available for it - as you can see in this comment here - https://github.com/guardian/dotcom-rendering/pull/3718#issuecomment-985479589

This change would do essentially exactly what this PR is doing (in less code, but also less explicitly) by flipping the `min-width` queries to `max-width` (There is slightly more that needs done to make it work but that's the crux of this). This would essentially give you the same output as you have in this PR; albeit it's achieved through different media queries.

--- 

I'd like to say I'd support this change (or my equivalent one) as a good solution as it solves the immediate problem with minimal side-effects. That is to say, this will _only_ affect immersive main media images ~ ensuring that cost & bandwidth implications are kept to a minimum.

Wider scale change about image quality, sourcesets and formats should be considered of course, but this would require longer term strategy, testing & cost analysis. This PR fixes an immediate issue affecting an upcoming long read

dotcom-rendering/src/web/components/Picture.tsx
OllysCoding: These numbers are actually very specific for a reason: 

image

We have to stay faithful to the sources provided to us by frontend, otherwise we'll select images of an ever higher resolution unintentionally

oliverlloyd: I think magic numbers can be a problem but I think in this case these values are extremely unlikely to change so they are already like constants - you see them in the code a lot - so maybe having them written inline isn't so bad? It's nice to be able to read them straight away without another level of abstraction in-between.

But I also see the case for using a named var in this case where it's not clear what's happening in this logic (it's choosing the next level _up_).

But I think we should address this in a separate PR. I don't want to introduce an inconsistent pattern in this function where some code returns numbers and other code returns named vars.

oliverlloyd:
Thanks all for the comments on this. It's very much the case here that there is an opportunity to do this better and I think we should continue working on this. Maybe a good first step would be to gather the information we have together into one place?

dotcom-rendering/src/web/components/Picture.tsx
mxdvl: > But I also see the case for using a named var in this case where it's not clear what's happening in this logic (it's choosing the next level _up_).

I think this is what tripped me the most. But really these values are not picking larger breakpoints, they’re matching frontend logic, and the fact that these numbers are the same as breakpoints is _somewhat accidental_.

https://github.com/guardian/frontend/blob/6a144f59373929664e7b84148c7699df93dbc63d/common/app/layout/ContentWidths.scala#L104-L116

PR #4354, opened 2022-03-22
Remove `data-spacefinder-component` attribute

chrislomaxjones:
## What does this change?

This PR removes the `data-spacefinder-component` attribute from figures. In doing this we can also remove the `isNumberedList` prop (which is optional and never passed in anywhere anyway).

This PR is dependent on [Frontend #24776](https://github.com/guardian/frontend/pull/24776) being merged first.

## Why?

This has been deprecated in favour of using `data-spacefinder-role` and `data-spacefinder-type`. See #4113 for the basis of this change.


arelra (approved):

lucymonie (approved):
LGTM 👍🏻

Issue #4346, opened 2022-03-21
Show pinned icon on posts that are pinned


oliverlloyd:
If a in body post [is pinned](https://www.theguardian.com/politics/live/2022/mar/21/boris-johnson-ukraine-brexit-sajid-javid-coronavirus-uk-politics-latest-updates?page=with:block-6238a48e8f0832c62ade2e90&dcr=false#block-6238a48e8f0832c62ade2e90) it should have an icon highlighting this

<img width="652" alt="Screenshot 2022-03-21 at 16 44 19" src="https://user-images.githubusercontent.com/1336821/159312386-b3a812f7-835d-44b4-bbb0-7deea977c157.png">

note. This screenshot if from Frontend, the date is positioned differently for DCR

abeddow91:
https://github.com/guardian/dotcom-rendering/pull/4368

PR #4368, opened 2022-03-22
Add pinned icon to original post

abeddow91:
## What does this change?
Adds a pinned icon to the original post that has been pinned. To do this, this PR passes the pinned post id through to liveblock so that it can be compared against each block id. If it matches the pinned post id but isn't itself the pinned post, it should render a pinned post icon next to the time and date stamp. 

This PR also passes the blocks format through so that the background colour of the icon can be derived from its format.

NB - pinned posts currently only exist in DCR not AR so the `isOriginalPinnedPost` prop in liveBlockContainer is optional. Once pinned posts are implemented in AR, this can become required.

## Why?
When a journalist pins a post, the original post stays in its original position in the live blog feed. Adding a pinned icon to this original post signifies it is the same post that is pinned above. 

### Before
Screenshot 2022-03-22 at 16 35 36

### After
Screenshot 2022-03-22 at 15 48 20


oliverlloyd:
We have an issue for this one!

https://github.com/guardian/dotcom-rendering/issues/4346

joecowton1 (approved):
Looks great! Nice work.

common-rendering/src/components/FirstPublished.tsx
```diff
@@ -66,6 +73,23 @@ const FirstPublished = ({
 					publishedDate.getMinutes(),
 				)}`}
 			</span>
+			{isOriginalPinnedPost && (
+				<span
+					css={css`
+						width: 0.875rem;
```
alinaboghiu, line 79: I wonder if these should be px? we seem to be using px elsewhere in this file.

abeddow91: It's a good question, I tend to favour rem where we can for scaling but use Source's space (with px) for spacing (margin, padding etc). @oliverlloyd is there any hard and fast rules for this in DCR?

mxdvl: Here’s a one-button fix, @abeddow91:

```suggestion
						width: 14px;
```

abeddow91: Great, thanks @oliverlloyd and @mxdvl !

PR #4369, opened 2022-03-22
Render byline image in blocks with correct size, use `editorialPalette` to decide the background

MarSavar:
## Why?
Byline images in AR-rendered blocks had two main issues:

1) The resolution was bad and the size was too large.
This happened because the `src` function that was generating the image URL used a width of `64px`, which wasn't enough for the resizer to render an image with high enough resolution. I had a look at the URL generated in DCR and the width in the URL is 300, so I replicated this in AR.

DCR url:

Screenshot 2022-03-22 at 17 04 01

2) No background colour was being rendered.
 The `BlockByLine` component had the `avatarBackgroundColor` prop drilled to it but the parent component also had `format` which seemed redundant (we can generate the colour by using `format` without having to pass an explicit background colour.).
I thought it would be good to get rid of `avatarBackgroundColor` altogether and use `editorialPalette` to decide the colour.

| **Before** | **After** |
|-----------|----------|
| image | image |


common-rendering/src/components/liveBlockContainer.tsx
```diff
@@ -86,9 +86,9 @@ const BlockByline = ({
 						css={css`
 							border-radius: 100%;
 							object-fit: cover;
-							height: 100%;
-							width: 100%;
-							background-color: ${avatarBackgroundColor};
+							height: 36px;
+							width: 36px;
```
MarSavar, line 90: In AR, the byline image wasn't respecting the width and height of the container and was rendered as a 300px wide image. This was puzzling because it is a shared component and the behaviour in DCR was correct. Also, unticking and then ticking `height: 100%; width: 100%;` in the dev tools fixed the issue, which is very mysterious.
I am still not sure why this was happening - maybe some sort of race condition in AR? Anyway, passing an explicit size to the image fixes the problem.

apps-rendering/src/contributor.ts
```diff
@@ -40,7 +40,7 @@ const tagToContributor =
 			fromNullable,
 			map((url) => ({
 				srcset: contributorSrcset(url, salt, Dpr.One),
-				src: src(salt, url, 64, Dpr.One),
+				src: src(salt, url, 300, Dpr.One),
```
OllysCoding, line 43: Is 300px not overkill?

jamesgorrie: Bikeshedding: Do we have any measure of image size similar to [compressed-size-action](https://github.com/preactjs/compressed-size-action)?

If we're going to make changes to image size and quality, it would be good to get some insight into what effect these changes are having. 

Something that looks useful is [lighthouse uses-responsive-images](https://web.dev/uses-responsive-images/) as a guard against adding too large images.

I think speedcurve gives image size, but we don't have that hooked up into CI.

mxdvl: How comes the `srcset` aren’t being picked up? I can see we have them above and below the `src` element.
If the image is rendered at 36px, 72px is probably plenty as a width.

JamieB-gu: > Something that looks useful is [lighthouse uses-responsive-images](https://web.dev/uses-responsive-images/)

> How comes the srcset aren’t being picked up?

I agree. I think using `srcset` and responsive images is preferable here, and AR is already generating them for these contributor images. However, the `LiveBlockContainer` component only takes a single image URL, which it drops into an `img` tag. As far as I'm aware, it does that because DCR only has access to a single URL for these images.

I think we have a couple of options for improvement.

## Responsive Images Everywhere

If we want to upgrade to responsive images (`picture`/`source`/`srcset`) on DCR we need to:

- Make a PR to `frontend` to generate them and send them through (DCR can't do this atm because it requires access to the "salt" for signed URLs)
- Update the DCR model to pass through this new data to the component
- Update the component API to accept the different `srcset`s 

AR already has these `srcset`s, so they can be passed through then too.

## Responsive Images For AR Only

We could also choose to leave the images on DCR as non-responsive and only add them for AR. This would involve adding a couple more optional props to the component to allow it to use `srcset`s if they're available.

The downside is that dotcom doesn't get responsive images here, but with the component API in place there's nothing stopping us doing the work on `frontend` and DCR later to add them if we choose.

MarSavar: > If we're going to make changes to image size and quality, it would be good to get some insight into what effect these changes are having.

This is a really interesting point, thank you. Initially I thought I wasn't really changing anything (and in fact I am not - I am merely doing the same thing as DCR here). But it got me thinking, so prepare for a wall of text.

This PR touches two things, essentially: 

- The apps-rendering backend
- The component (shared by DCR and AR)

The component takes an array of contributors. 
Each contributor is an object with two fields: name and image URL. 

DCR gets contributors from `frontend` ([this snippet of code](https://github.com/guardian/frontend/blob/efa1ec561bf28bab6638f0d644d3d205f8fb81ae/common/app/model/dotcomrendering/DotcomRenderingSupportTypes.scala#L92-L94), specifically):

```scala
    val contributors = block.contributors flatMap { contributorId =>
      tags.find(_.id == s"profile/$contributorId").map(tag => Contributor(tag.title, tag.bylineImageUrl))
    }
```

Essentially, if a block has a tag such as `profile/journalist-name`, the above code returns an array of `Contributor` objects with the journalist's name and byline image URL. 

But where is the URL coming from? Good question! I did a bit of digging, and discovered that frontend and DCR deal with this differently.

## Frontend 
Code [here](https://github.com/guardian/frontend/blob/main/common/app/model/Tag.scala#L202) and [here](https://github.com/guardian/frontend/blob/main/common/app/views/support/ImageProfile.scala#L120
) 
```scala
val contributorImagePath = properties.bylineImageUrl.map(ImgSrc(_, Contributor))
```



```scala
object Contributor extends ImageProfile(width = Some(140), height = Some(140))
```
The default width for contributor images is `140px`. I verified this by looking at a liveblog rendered by frontend ([link to block](https://www.theguardian.com/world/live/2022/mar/21/russia-ukraine-war-latest-news-kyiv-rejects-moscows-deadline-for-mariupol-surrender-biden-to-visit-poland-live?page=with:block-6238e1a48f08ee17b53740ed#block-6238e1a48f08ee17b53740ed))

image

image

Things start getting interesting here. The template generating the byline block image is defined [here](https://github.com/guardian/frontend/blob/main/common/app/views/fragments/meta/bylineLiveBlockImage.scala.html). 
The block byline defaults to 300px width. However, in the screenshot above we can see that the requested image is 140px wide.

```html
<img class="liveblog-block-byline__img" src="@ImgSrc(src, Item300)" alt="@profileTag.name" />
```
 But somehow, it sometimes defaults to 140px in the rendered HTML code. I'm not entirely sure why this is.

## DCR
In the DCR model [Scala code](https://github.com/guardian/frontend/blob/efa1ec561bf28bab6638f0d644d3d205f8fb81ae/common/app/model/dotcomrendering/DotcomRenderingSupportTypes.scala#L28-L40), contributor images *always* default to 300px.

```scala
object Tag {
  implicit val writes = Json.writes[Tag]

  def apply(t: model.Tag): Tag = {
    Tag(
      t.id,
      t.properties.tagType,
      t.properties.webTitle,
      t.properties.twitterHandle,
      t.properties.contributorLargeImagePath.map(src => ImgSrc(src, Item300)),
    )
  }
}
```
We can verify this by looking at the same block above. (I added `dcr` to the URL)
image
image.

## Apps Rendering
AR didn't have a function to convert tags to contributors, which is why byline images weren't being rendered.
This was fixed in this PR -> #4347 
The backend code that was supposed to render bylines in AR was written/last modified 6 months ago but it never worked because the `contributors` array was never being populated and passed as a prop. I'm not sure why `64` was chosen as a width. All I'm doing in this PR is bringing the width in line with DCR (to `300`px).

So the questions I have are.
- If frontend is rendering bylines at 140px, why is DCR rendering them at 300px? Maybe @oliverlloyd knows!
- Should we resize both DCR- and AR-rendered bylines to 140px to bring them in line with frontend?
---
> How comes the srcset aren’t being picked up? I can see we have them above and below the src element.
@mxdvl Because this function in AR's `liveBlock.tsx` is only taking `src` as the `imageUrl`, to have parity with DCR.

```typescript
const contributorToBlockContributor = (
	contributor: Contributor,
): BlockContributor => ({
	name: contributor.name,
	imageUrl: pipe(
		contributor.image,
		map((i) => i.src),
		withDefault<string | undefined>(undefined),
	),
});
```
I think you're right, but for the purpose of going to beta with deadblogs as soon as possible we should probably keep it as it is for now. 100% agree that in the future both DCR and AR should have `scrset`.

MarSavar: The 300px byline img was 40kb. I've changed it to 140px to mirror frontend drops that down to 10px with no noticeable loss of quality.
CC @paperboyo

paperboyo: Thx! I guess, without just following our usual Picture Element syntax with all the different sizes, doing it at twice the CSS pixels width should be fine (and using `quality` for DPR2 (if the CSS pixels dimensions are static between all uses).

common-rendering/src/components/liveBlockContainer.tsx
JamieB-gu: You may want `rem`s instead of px. This is `2.25rem`.

JamieB-gu (approved):
One suggestion, otherwise looks good 👍

PR #4370, opened 2022-03-22
Fix spacing on pinned post button 

abeddow91:
## What does this change?
Adds 4px of space between the icons and the text on the pinned post button in line with Figma designs

### Before
Screenshot 2022-03-22 at 16 53 41

### After
Screenshot 2022-03-22 at 16 53 44


oliverlloyd (approved):
Perfect

PR #4373, opened 2022-03-23
Remove some false negatives from Chromatic

oliverlloyd:
## What does this change?
Disables Chromatic on a couple of stories that keep throwing false negatives


jamesgorrie (commented):
The risk of this is if we break those components visually, we don't have any insight into it?

Could we leave a comment saying why so if this breaks in future and we come back to it we know we just hacked it.

OllysCoding (approved):
+1 Begone!!!

PR #4374, opened 2022-03-23
BUGFIX: Interactive articles using iframes when not required

jlieb10:
## What does this change & why?

co-authored-by: @OllysCoding @mxdvl 

Articles of both design type 'interactive' and design type 'fullpageinteractive' should use the non-iframed interactive renderer code. Because of recent changes in CAPI and DCR, we are now actively using the 'fullpageinteractive' design type and therefore needed to make this change so that interactive articles of this type are rendered correctly

### Before
Screenshot 2022-03-23 at 09 41 39

### After
Screenshot 2022-03-23 at 09 42 10


mxdvl:
Fixes #4375

PR #4379, opened 2022-03-23
refactor: infer palette from format

mxdvl:
## What does this change?

Infer `palette` from `format` in the following layouts:
- CommentLayout
- ImmersiveLayout
- InteractiveLayout
- LiveLayout
- ShowcaseLayout
- StandardLayout

By changing the props of components `ArticleBody`, `Border` and `SubMeta`.

## Why?

Having both is redundant.

Follow-up on #4053


AshCorr (approved):
Awesome!

OllysCoding (approved):
+1

PR #4386, opened 2022-03-23
Allow GetMatchStats and GetMatchNav to fail & hide the placeholder

OllysCoding:
## What does this change?

Max retries on GetMatchStats & GetMatchNav set to 1; This means if for some reason we can't get match stats (For example we don't have the Women's FA Cup 2022), we won't infinitely show the placeholder on the page.

### Before

image

### After

image

(Placeholder is removed after less than 1 second)

### DIFF
--- a/dotcom-rendering/src/web/components/GetMatchNav.importable.tsx
+++ b/dotcom-rendering/src/web/components/GetMatchNav.importable.tsx
@@ -17,12 +17,12 @@ export const GetMatchNav = ({ matchUrl }: Props) => {
 		comments?: string;
 		minByMinUrl?: string;
 		venue?: string;
-	}>(matchUrl);
+	}>(matchUrl, { errorRetryCount: 1 });
 
 	if (loading) return <Loading />;
 	if (error) {
 		// Send the error to Sentry and then prevent the element from rendering
-		window.guardian.modules.sentry.reportError(error, 'match-nav');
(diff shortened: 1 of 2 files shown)


ioannakok (approved):

dotcom-rendering/src/web/components/GetMatchNav.importable.tsx
```diff
@@ -17,12 +17,12 @@ export const GetMatchNav = ({ matchUrl }: Props) => {
 		comments?: string;
 		minByMinUrl?: string;
 		venue?: string;
-	}>(matchUrl);
+	}>(matchUrl, { errorRetryCount: 1 });
```
DavidLawes, line 20: If the match stats exist, but we error randomly once, the match stats will be removed from the page? Would the reader have to refresh the page to then see match stats again (given the response returns what's expected)?

Only flagging this in case we think an error retry count of 1 is too strict, but I have no idea of the error rate in production.

oliverlloyd (approved):

dotcom-rendering/src/web/components/GetMatchNav.importable.tsx
mxdvl: You’d have to error twice in a row (1 attempt + 1 retry). This sounded sane.

mxdvl (approved):

PR #4388, opened 2022-03-23
Ab/fix pinned post scroll

abeddow91:
## What does this change?
Adds a function that scrolls the user to the top of the pinned post when the `see less` button is clicked if the top of the pinned post is out of view. If the full pinned post is within the viewport there is no scroll. 

NB - this is part of the pinned post enhanced JS so users without JS enabled will not be scrolled to the top of the pinned post.

## Why?
This fixes a bug where users were scrolled to the top of the main media instead of the top of the pinned post.

### Before
https://user-images.githubusercontent.com/20416599/159755971-f97ba928-5dc1-430e-aa69-c85a9d5b28e0.mov

### After
https://user-images.githubusercontent.com/20416599/159756328-3973149b-2cdb-4189-b28d-fd010fd0bf28.mov

### DIFF
--- a/dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
+++ b/dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
@@ -23,12 +23,15 @@ function hideShowMore() {
 }
 
 /**
- * Scroll to the top of the main content when the pinned post is collapsed
+ * Scroll to the top of the main content when the pinned post is collapsed if the top of the post is out of view
  */
 function scrollOnCollapse() {
-	document.getElementById('maincontent')?.scrollIntoView({
-		behavior: 'smooth',
-	});
+	const position = pinnedPost?.getBoundingClientRect();
+	if (position && position.top < 0) {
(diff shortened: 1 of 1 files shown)


oliverlloyd (approved):
I love these kinds of small ux improvements. Work like this is what makes a thing great

dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
```diff
@@ -22,6 +22,18 @@ function hideShowMore() {
...
 }
 
+/**
+ * Scroll to the top of the main content when the pinned post is collapsed if the top of the post is out of view
+ */
+function scrollOnCollapse() {
+	const position = pinnedPost?.getBoundingClientRect();
+	if (position && position.top < 0) {
+		pinnedPost?.scrollIntoView({
+			behavior: 'smooth',
```
mxdvl, line 32: Just a note to remind everyone that this is not supported on Safari–including iOS–, where the scroll will be instantaneous.

PR #4389, opened 2022-03-23
Fix key events dark mode at narrow breakpoints

MarSavar:
## Why?
Key events are wrapped in an accordion component at narrow breakpoints (which always has the same background colour regardless of whether it's a liveblog or a deadblog), and we were not considering this previously.

| **Before** | **After** |
|------------|----------|
| image | image |

No changes at wider breakpoints:

image

### DIFF
--- a/common-rendering/src/components/keyEvents.tsx
+++ b/common-rendering/src/components/keyEvents.tsx
@@ -13,6 +13,7 @@ import {
 	neutral,
 	remSpace,
 	from,
+	until,
 } from '@guardian/source-foundations';
 import { Link } from '@guardian/source-react-components';
 import {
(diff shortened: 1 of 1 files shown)


common-rendering/src/components/keyEvents.tsx
```diff
@@ -82,6 +83,10 @@ const listStyles = (
 	supportsDarkMode: boolean,
 ): SerializedStyles => css`
 	${darkModeCss(supportsDarkMode)`
+		${until.desktop} {
+			background-color: ${neutral[10]};
```
JamieB-gu, line 87: Would it be possible to capture this colour in `editorialPalette`?

MarSavar: Do you mean creating a new function where we also pass the breakpoint as an argument?

MarSavar: Got you. 
A hypothetical `keyEventsWide` would be reusing the same background as the article.

For clarity, do you think something like this would be helpful?
```typescript
const keyEvents = (_format: ArticleFormat): Colour => neutral[10];

const keyEventsWide = articleContentDark;
```
That way, the styles in the component would look like the code you wrote in the suggestion above.
Having a function pointing at another function is kind of redundant but would make the component code more readable imho.

PR #4390, opened 2022-03-23
Allow a single Doughnut section

mxdvl:
## What does this change?

Allow a single Doughnut section and refactors the SVG code. Doughnut now only takes a `size` prop instead of `width` and `height`.

Instead of using [`path`][] and drawing arcs, leverage the [`circle`][] element, and use stroke dash to draw the arcs.

[`path`]: https://developer.mozilla.org/en-US/docs/Web/API/SVGPathElement
[`circle`]: https://developer.mozilla.org/en-US/docs/Web/API/SVGCircleElement

## Why?

This was a long-standing missing feature. I took this as an opportunity to refactor some of the doughnut code to simplify it.

## Screenshot

image

### DIFF
--- a/dotcom-rendering/src/web/components/Doughnut.tsx
+++ b/dotcom-rendering/src/web/components/Doughnut.tsx
@@ -16,6 +16,9 @@ type Section = {
 	color: string;
 };
 
+/** three decimal places are plenty */
+const PRECISION = 3;
+
 const unitStyles = css`
 	${headline.medium({ fontWeight: 'bold' })}
 	text-anchor: middle;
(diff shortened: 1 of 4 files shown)


dotcom-rendering/src/web/components/Doughnut.tsx
```diff
@@ -40,33 +39,33 @@ const withoutZeroSections = (sections: Section[]) =>
...
 
-	const halfPI = Math.PI / 2;
-	const doublePI = Math.PI * 2;
+	/** τ = 2π https://en.wikipedia.org/wiki/Turn_(angle)#Tau_proposals */
+	const tau = Math.PI * 2;
 
 	const center = {
-		x: width / 2,
-		y: height / 2,
+		x: size / 2,
```
AshCorr, line 59: Worth refactoring this into a single variable since x and y will be identical now?

AshCorr (approved):
Won't pretend to know any of the maths going on here, seems pretty cool though!

Looks like the doughnut is a bit thin in the chromatic tests

mxdvl:
> Won't pretend to know any of the maths going on here, seems pretty cool though!

Thanks to your comment I’ve pushed this to be even one step simpler, and use SVG `circle`s instead of `path`s. Not only do we send less data 
 
> Looks like the doughnut is a bit thin in the chromatic tests

I did the math wrong 🤓 – fixed now

mxdvl:
Reverted 47c4f920fde822f0c2813da59d0b982e805d9485 in #7816

PR #4393, opened 2022-03-24
`position: sticky` in Safari

### LINKED ISSUE #4372: Sticky toast on Safari
The toast is not sticking on Safari (v14)

oliverlloyd:
## What?
This PR fixes #4372  a bug where when in Safari the `Toast` was not sticking. 

Safari treats position: sticky differently to Chrome et al; it uses the immediate parent when deciding where to stick things but in this case the immediate parent is the `Island` and this element has no height so nothing is stuck.

To get around this I considered various options:

1) Move the Toast out of the Island. If we rendered it by itself we could put it anywhere which would solve this issue but the problem with this approach is the component has props and state which are linked to the code inside the `Liveness` island. Rejected ❌

2) Add css to the `Island`. It would be nice to solve this css problem using css but this neither works (if I give the island height it pushes other content around) and would introduce undesired coupling. Rejected ❌

3) Use a [React Portal](https://reactjs.org/docs/portals.html). Portals in React are a way to render content outside the classic DOM hierarchy that you would normally expect using React. By doing a portal we can continue to benefit from shared state and props but still render the element as a sibling to the Island and not as a child of it. Chosen ✅

### DIFF
--- a/dotcom-rendering/src/web/components/Liveness.importable.tsx
+++ b/dotcom-rendering/src/web/components/Liveness.importable.tsx
@@ -1,4 +1,5 @@
 import { useEffect, useState, useCallback } from 'react';
+import ReactDOM from 'react-dom';
 import { initHydration } from '../browser/islands/initHydration';
 import { useApi } from '../lib/useApi';
 import { updateTimeElement } from '../browser/relativeTime/updateTimeElements';
@@ -25,6 +26,10 @@ const topOfBlog: Element | null = !isServer
 
 const lastUpdated: Element | null = !isServer
 	? window.document.querySelector('[data-gu-marker=liveblog-last-updated]')
+	: null;
+
+const toastRoot: Element | null = !isServer
+	? window.document.getElementById('toast-root')
 	: null;
 
 /**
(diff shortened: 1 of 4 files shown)


dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -284,13 +289,25 @@ export const Liveness = ({
...
+		 * createPortal?
+		 *
+		 * We're using a portal here because of a the way Safari implements position: sticky. If
+		 * the element is rendered directly here as a child of the Island (its parent) then stickiness
+		 * won't work in Safari because this browser positions the element relative to the immediate
+		 * parent, whereas Chrome et al are more forgiving.
+		 *
+		 * By using a Portal we can insert the Toast as a sibling to the Island, which works around
+		 * Safari's behaviour.
+		 */
```
mxdvl, line 303:
```suggestion
		/**
		 * Why `createPortal`?
		 *
		 * We use a Portal because of a the way different browsers implement `position: sticky`.
		 * A stickily positioned element depends on its [containing block][] to determine
		 * when to become stuck.
		 *
		 * In Safari the containing block is set to the immediate parent
		 * whereas Chrome, Firefox and other browser are more forgiving.
		 *
		 * By using a Portal we can insert the Toast as a sibling to the Island, which works around
		 * Safari's behaviour.
		 *
		 * [containing block]: https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block
		 */
```

mxdvl (approved):
Good catch. I think poor old Safari [was only respecting the spec](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block).

I’ve added a suggestion to update the wording slightly.

dotcom-rendering/src/web/components/Liveness.importable.tsx
oliverlloyd: Thanks! 🙏

PR #4394, opened 2022-03-24
Update Blockquote component

iainjchambers-guardian:
## What does this change?
Updates the design of the blockquote component

| Before |  After |
| -------|-------|
| image | image |

### DIFF
--- a/apps-rendering/src/components/blockquote.tsx
+++ b/apps-rendering/src/components/blockquote.tsx
@@ -5,7 +5,6 @@ import { css } from '@emotion/react';
 import { fill } from '@guardian/common-rendering/src/editorialPalette';
 import type { ArticleFormat } from '@guardian/libs';
 import { remSpace } from '@guardian/source-foundations';
-import { SvgQuote } from '@guardian/source-react-components';
 import type { FC, ReactNode } from 'react';
 import { darkModeCss } from 'styles';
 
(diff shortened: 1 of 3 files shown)


apps-rendering/src/components/blockquote.tsx
```diff
@@ -20,24 +19,44 @@ const styles = (format: ArticleFormat): SerializedStyles => css`
...
 	position: relative;
 	margin: ${remSpace[4]} 0 ${remSpace[9]} 0;
-	padding: 0 ${remSpace[6]};
 
-	svg {
-		height: 62.4px;
-		margin-left: -9.2625px;
-		margin-bottom: -12px;
-		margin-top: 4.4px;
+	> svg {
```
AshCorr, line 23: Mostly just nitpickity, feel free to ignore me, I don't think we use the `>` css operator much at all in this codebase

mxdvl: Here, selecting the direct child is exactly what we want to do.

In some ways, if we select children, we should _only_ select direct children, otherwise you can have css cascade issues.

apps-rendering/src/components/blockquote.tsx
```diff
@@ -20,24 +19,44 @@ const styles = (format: ArticleFormat): SerializedStyles => css`
...
+		&:nth-of-type(2) {
+			margin-top: ${remSpace[3]};
+		}
+	}
 `;
 
 const Blockquote: FC<Props> = ({ children, format }: Props) => (
 	<blockquote css={styles(format)}>
-		<SvgQuote />
+		<svg
```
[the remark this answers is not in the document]
mxdvl: It’s has a massive (~8px) blank padding all around. This is a tight version of the icon, which is meant to be used inline, and cannot be authored the same way.

apps-rendering/src/components/blockquote.tsx
```diff
@@ -20,24 +19,44 @@ const styles = (format: ArticleFormat): SerializedStyles => css`
...
-	padding: 0 ${remSpace[6]};
 
-	svg {
-		height: 62.4px;
-		margin-left: -9.2625px;
-		margin-bottom: -12px;
-		margin-top: 4.4px;
+	> svg {
+		width: 18px;
+		height: 12px;
```
mxdvl, line 25: I think ideally you’d have this be defined relatively to the text size?

```suggestion
		width: 1.5em;
		height: 1em;
```

mxdvl (commented):
Would be great to leverage the fact that we did the same thing in #4022.

Maybe [`QuoteIcon`](https://github.com/guardian/dotcom-rendering/blob/9b63598e4ab9e3741dfe8c5580a6e98a6f8a57a0/dotcom-rendering/src/web/components/QuoteIcon.tsx) can be moved to `common-rendering`?

apps-rendering/src/components/blockquote.tsx
JamieB-gu: I think regardless of what you decide on the above point it would be useful to have this defined in relative units? So even if not `em`s, then `rem`s?

JamieB-gu (approved):
Nice, thanks for adding stories 👍

PR #4396, opened 2022-03-24
 Auto accept any chromatic diffs on main builds

AshCorr:
## What does this change?

Sets the automic action to auto approve any diffs on the main branch.

## Why?

Looking at [this build](https://www.chromatic.com/test?appId=5dfcbf3012392c0020e7140b&id=623c46322ac802003a41bc30) I noticed there was some diffs that were completely unrelated to the thing being changed. Checking the "Baseline history" that chromatic provided I could see that there was a [main build](https://www.chromatic.com/test?appId=5dfcbf3012392c0020e7140b&id=623c46322ac802003a41bc30) with the exact same diff that had not been accepted 

I'm not sure how a diff ended up being on main, my guess is that a branch a few commits behind main was merged and caused diffs once it was mixed with the newer main code.

Maybe this'll decrease how many false positives we get?

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -24,3 +24,4 @@ jobs:
         with:
           projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
           token: ${{ secrets.GITHUB_TOKEN }}
+          autoAcceptChanges: main


ioannakok (approved):
👍

.github/workflows/chromatic.yml
```diff
@@ -24,3 +24,4 @@ jobs:
         with:
           projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
           token: ${{ secrets.GITHUB_TOKEN }}
+          autoAcceptChanges: main
```
AshCorr, line 27: Chromatic docs seems to implies you can pass a branch name here instead of `true` or `false` but it's not exactly clear, so this might not work.

https://www.chromatic.com/docs/github-actions

oliverlloyd:
Is there any risk here that a bug which gets introduced as part of the merge into main is then ignored?

We don't require that branches be up to date with main when they are merged as this really slows down merging but the trade of for this is we _sometimes_ get issues on main when two different code changes clash

AshCorr:
> Is there any risk here that a bug which gets introduced as part of the merge into main is then ignored?

Yes, potentially! It's a trade off, do we want to make it easier for people to understand their chromatic diffs or do we want to try and catch edge case scenarios of an old brach being merged.

The other way this could be resolved is if we actually approve main diffs... I think chromatic does send emails when theres a diff on main but these usually get lost in the barrage of automated emails we receive every day. Maybe we can add another step to the chromatic action to send a message to our Google Chats PR thread when a diff is found in main?

oliverlloyd:
> Yes, potentially! It's a trade off, do we want to make it easier for people to understand their chromatic diffs or do we want to try and catch edge case scenarios of an old brach being merged.

I think if we think this will improve the dev ex around this then I'm pro making this change. I've only ever seen two or perhaps three issues crop up like this in the whole lifetime of the project so it is very edge casey.

mxdvl:
I think we want this to go in, @AshCorr ?

AshCorr:
> I think we want this to go in, @AshCorr ?

I think I can improve on this, but yea, this can go in for now!

PR #4398, opened 2022-03-24
Remove blank space in standfirst lists

### LINKED ISSUE #4378: Whitespace Between Headline And Standfirst
On some deadblogs whitespace is appearing between the headline and the standfirst on narrower breakpoints. This appears to be happening when the first item in the standfirst is a list. There seem to be two problems here:

1. When the first item in the standfirst is text, the paragraph padding is being used - the standfirst itself doesn't have top padding
2. When the first item in the standfirst is a list there's no top padding, but there is a margin that's extending beyond the standfirst area and creating the whitespace

We should make sure that the spacing at the top of the standfirst is consistent regardless of the content, and prevent the whitespace from appearing.

| Text | List |
| - | - |
| deadblog-text | deadblog-list |

mxdvl:
## What does this change?

Uses padding instead of margin for the gaps around standfirst’s paragraphs and unordered lists.

## Why?

Margins bleed outside of components so it can be hard to reason about how they will look. This fixes #4378 : a white gap in standfirst with lists.

### Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/159935712-2084c707-61ff-4b64-86d4-8086afca31da.png
[after]: https://user-images.githubusercontent.com/76776/159935710-c57123b4-d5ee-49a4-8180-9a9b0139a293.png

### DIFF
--- a/apps-rendering/src/components/standfirst.tsx
+++ b/apps-rendering/src/components/standfirst.tsx
@@ -52,7 +52,7 @@ const styles = (format: ArticleFormat): SerializedStyles => css`
 
 	p,
 	ul {
-		padding: ${remSpace[3]} 0;
+		padding: ${remSpace[3]} 0 0;
 		margin: 0;
 	}
 
--- a/apps-rendering/src/components/standfirst.tsx
+++ b/apps-rendering/src/components/standfirst.tsx
@@ -52,7 +52,8 @@ const styles = (format: ArticleFormat): SerializedStyles => css`
 
 	p,
 	ul {
-		margin: ${remSpace[3]} 0;
+		padding: ${remSpace[3]} 0;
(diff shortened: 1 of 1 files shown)


MarSavar (approved):
Great, thank you! 👍

mxdvl:
@MarSavar – this has increased spacing on some articles. Not sure if I can accept the Chromatic diff?

MarSavar:
> @MarSavar – this has increased spacing on some stories. Not sure if I can accept the Chromatic diff?

I have tested your branch locally and can't spot any visual differences (I've checked all breakpoints) so go ahead and approve the Chromatic diff!

PR #4404, opened 2022-03-25
Refactor web server

mxdvl:
## What does this change?

Check for error type with `instanceof Error` and leverage that for the `500` response.
Add a method to enhance CAPIType, which is used in three endpoints: `renderArticle`, `renderArticleJson` & `renderInteractive`

## Why?

We currently are doing type assertions and expecting errors. There’s an easy fix which makes the compiler happy and relies on our runtime checks.

`// @ts-expect-error` introduced in https://github.com/guardian/dotcom-rendering/commit/feb22e7ce1dfe92b66bac384296b2b77c60f7c32#diff-8629f7420eec981686c10018f7b4182728efd07fc2646149d231edeb24b251cc


dotcom-rendering/src/web/server/index.ts
```diff
@@ -49,13 +53,7 @@ export const renderArticleJson = (
 	res: express.Response,
 ): void => {
 	try {
-		const data = validateAsCAPIType(body);
-		const CAPI = {
-			...body,
```
mxdvl, line 54: Was it intentional to use `body` instead of `data`, here?

oliverlloyd (approved):

PR #4406, opened 2022-03-25
`Placeholder` can be different colours

### LINKED ISSUE #4383: Ability to colour Placeholder
I[](https://www.theguardian.com/football/2021/nov/06/brentford-norwich-premier-league-match-report)[](https://www.theguardian.com/football/2021/nov/06/brentford-norwich-premier-league-match-report)n DCR, we reserve the space for some elements while the data is fetched. For this, we insert a placeholder element, which is then overwritten once the data is ready. See this match scoe component on a match report - [Teemu Pukki’s penalty beats Brentford to secure Norwich’s first league win](https://www.theguardian.com/football/2021/nov/06/brentford-norwich-premier-league-match-report)

This currently loads in as a grey placeholder, but it would be nice and more visually seamless to have it as the colour of the component that is going to load.

oliverlloyd:
## What does this change?
This PR fixes #4383 by adding the `backgroundColor` prop to the `Placeholder` component

## Why?
We want to use a placeholder when loading in the `MatchNav` component but this sits on a yellow background

### DIFF
--- a/dotcom-rendering/src/web/components/Placeholder.stories.tsx
+++ b/dotcom-rendering/src/web/components/Placeholder.stories.tsx
@@ -101,3 +101,16 @@ export const NoShimmer = () => {
 	);
 };
 NoShimmer.story = { name: 'without shimmer' };
+
+export const Background = () => {
+	return (
+		<Container>
+			<Placeholder
+				height={200}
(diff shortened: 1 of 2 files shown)


mxdvl (approved):
Whoosh

PR #4408, opened 2022-03-25
Remove certain ad sizes from `comments` and `mostpop` slots

simonbyford:
## What does this change?

Remove `video` and `outstream` ad size targeting from `comments` and `mostpop` ad slots

## Why?

Only `inline1` slots need to target `outstream` ad sizes.

The `video` ad size has been deprecated: https://github.com/guardian/commercial-core/pull/527


mxdvl (approved):
less is more

Issue #4264, opened 2022-03-16
Flakey Cypress tests failing on TeamCity


oliverlloyd:
Note from Tom W copied from Dotcom channel:

> I've got a couple of cypress failures on the build for my branch here: https://teamcity.gutools.co.uk/buildConfiguration/dotcom_rendering/752383?showRootCauses=false&pluginCoverage=true&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=752383_6131_494 (specifically parallel-1/article.e2e.spec.js and parallel-3/article.interactivity.spec). They ran fine in GHA. I'm guessing these are just sporadic failures. Would you just re-run the build and hope they resolve themselves?

OllysCoding:
I don't think this is completely resolved, but I'm closing this card for the time being!

PR #4417, opened 2022-03-25
Cypress Consistency - Remove pre-hydration checks for "idle" loading rich links

OllysCoding:
## What does this change?

Removes checks designed to check for the 'pre-hydration' state of rich links.

## Why?

This test if failing consistently in TC main builds. My hope is that this change will improve the consistency on main TC builds - but this can only be tested by merging it in!

Rich-links hydrate on CPU idle, rather than on-scroll; This means that it's essentially a race condition between the cypress pre-hydration check and actual hydration of the rich link to verify the 'pre-hydration' state.

Since we know that without hydration the image in the rich link should not exist, I believe the test will still perform its desired purpose without the pre-hydration check.


oliverlloyd (approved):
Great description with a fair justification for this change

mxdvl:
Fixes #4264

PR #4431, opened 2022-03-28
Refactor `ArticleHeadline` to include age warning

oliverlloyd:
## What does this change?
This refactors the code to move control for when an age warning is shown away from the templates and into the `ArticleHeadline` component

## Why?
Because we always want to show this but if we leave it up to the layout files we risk it being overlooked.

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleHeadline.tsx
+++ b/dotcom-rendering/src/web/components/ArticleHeadline.tsx
@@ -253,7 +253,7 @@ const ageWarningMargins = (format: ArticleFormat) =>
 		  `;
 
 const backgroundStyles = (palette: Palette) => css`
-	background-color: ${palette.background.headline};
+	background-color: ${palette.background.ageWarning};
 `;
 
 const WithAgeWarning = ({
--- a/dotcom-rendering/src/web/components/ArticleHeadline.tsx
+++ b/dotcom-rendering/src/web/components/ArticleHeadline.tsx
@@ -234,6 +234,7 @@ const HeadlineAgeWarning = ({
 	isScreenReader?: boolean;
 	format: ArticleFormat;
 }) => {
+	const palette = decidePalette(format);
 	const ageWarningMargins =
 		format.display === ArticleDisplay.Immersive
 			? css`
@@ -263,6 +264,9 @@ const HeadlineAgeWarning = ({
 						margin-top: 0;
 					}
 			  `;
+	const backgroundStyles = css`
+		background-color: ${palette.background.headline};
+	`;
 
 	const age = getAgeWarning(tags, webPublicationDateDeprecated);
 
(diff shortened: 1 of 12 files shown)


mxdvl (approved):
Seems like a sound refactor!

dotcom-rendering/src/web/components/ArticleHeadline.tsx
```diff
@@ -220,36 +223,100 @@ const zIndex = css`
...
-								maxWidth,
-								immersiveStyles,
-								displayBlock,
-								reducedBottomPadding,
-							]}
-						>
-							{curly(headlineString)}
-						</h1>
+						<>
+							<HeadlineAgeWarning
```
AshCorr, line 283: Nitpicking, worth having a wrapper for this?

AshCorr (approved):
Looks good!

dotcom-rendering/src/web/components/ArticleHeadline.tsx
oliverlloyd: Great idea!

Done 👍

oliverlloyd:
> It doesn’t look great on Immersives

This was because previously they simply never appeared. We got away with this because we only show the age warning for News and Comment pieces, which are rarely, if ever, immersive but I think it's nice to include designs for this here in any case.

dotcom-rendering/src/web/components/ArticleHeadline.tsx
```diff
@@ -223,6 +223,39 @@ const zIndex = css`
...
+
+				${from.leftCol} {
+					margin-left: -10px;
+					margin-top: 0;
+				}
+		  `;
+
+const backgroundStyles = (palette: Palette) => css`
+	background-color: ${palette.background.headline};
+`;
```
mxdvl, line 257: I’m wondering if in cases such as these, we want to make it more explicit which part of the palette we require, maybe like so:

```suggestion
const backgroundStyles = (background: Palette['background']) => css`
	background-color: ${background.headline};
`;
```

oliverlloyd: Does this speak to our discussion around how much we want to lean into typescript vs. keeping things simple? I can see how this might seem like a nicer, more concise solution, and in isolation that is indeed the case, but perhaps the larger impact here is that we're incrementally adding more and more features of Typescript, slowly increasing the breadth of knowledge that we're expecting new developers to have have before they can easily read and understand this code?

I realise the instinct when reading this is to think, but this is easy! When you have Thing['property'] then this means the type being used in only that properties type, not the whole thing. But the issue is when you add up all these small, seemingly small, simple little things you end up with one large, complicated thing which overwhelms people who haven't had the chance to slowly learn them step by step

mxdvl: I was thinking more of the way we structure `palette`. As an example, `common-rendering`’s `editorialPalette` is split into submodules since #4176, so you can call `background.headline(format)`.

PR #4440, opened 2022-03-29
Pillars go yellow and underline on hover

DavidLawes:
## What does this change?
Updates navigation pillar styling - now on hover pillar links will be unlined and yellow.

## Why?
To more closely meet the design

### Before
Screenshot 2022-03-29 at 10 58 17

### After
Screenshot 2022-03-29 at 10 58 28

### DIFF
--- a/dotcom-rendering/src/web/components/Pillars.tsx
+++ b/dotcom-rendering/src/web/components/Pillars.tsx
@@ -6,6 +6,7 @@ import {
 	headline,
 	from,
 	until,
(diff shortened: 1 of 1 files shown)


ioannakok (approved):
Looks good! I can see two chromatic diffs but I don't understand what's different 😆

oliverlloyd (approved):
This is spot on.

I approved those Chromatic diffs as they are unrelated

PR #4441, opened 2022-03-29
Set maximum width on `inline` ads after `inline1`

chrislomaxjones:
## What does this change?

Set a maximum width of `300px` on inline ads `inline2`, `inline3`, etc. These are the inline ads that have creatives of maximum width `300px` across all breakpoints. 

We exclude `inline1` and `top-above-nav` (this is the first inline on mobile) and fluid ads marked with `ad-slot--fluid`.

## Why?

Google performs [ad expansion](https://support.google.com/admanager/answer/9117822?hl=en). If there is demand available, an ad of width greater than those permitted in the size mappings will filled. 

When we don't set a maximum width on the ad container, this expansion can break the layout.

### Before

Image from an an article where an ad has performed expansion.

Screenshot 2022-03-29 at 10 56 20

### After

Image from CODE with the fix applied showing ads from the same company with a creative adhering to our maximum width.

Screenshot 2022-03-29 at 11 48 49

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleContainer.tsx
+++ b/dotcom-rendering/src/web/components/ArticleContainer.tsx
@@ -86,7 +86,7 @@ const articleAdStyles = css`
 			margin-left: 20px;
 		}
 	}
-	.ad-slot--inline:not(.ad-slot--inline1):not(.ad-slot--top-above-nav) {
+	.ad-slot--inline:not(.ad-slot--inline1):not(.ad-slot--top-above-nav):not(.ad-slot--fluid) {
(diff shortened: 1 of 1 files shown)


arelra (approved):
🎉

chrislomaxjones:
### For future readers

Unfortunately this fix didn't manage to fix ad expansion on the slot (screenshot below from @simonbyford). It may be possible that this is in fact a bug in the sizes of ads being returned, since this particular creative doesn't seem like an expanded version of a smaller creative.

Screenshot 2022-03-29 at 15 25 49

PR #4444, opened 2022-03-29
Only stick videos on liveblogs

joecowton1:
## What does this change?

Adds a check to ensure we only stick videos on live blogs and dead blogs, **not** on articles.

### Article Before
Screenshot 2022-03-29 at 14 01 51

### Article After
Screenshot 2022-03-29 at 13 57 42


AshCorr (approved):
Good change!

joecowton1:
[Sticky Video: limit to liveblogs only](https://trello.com/c/6iU05VJK/296-sticky-video-limit-to-liveblogs-only)

PR #4449, opened 2022-03-29
Cypress Consistency - Use mocks on article.e2e AB tests tests

OllysCoding:
## What does this change?

Use mocks on AB Test tests! The AB test test relies on on a `**/most-read/**` completing, by mocking this response we eliminate network conditions that might cause the call to fail!

Note: This is a guess as what might be causing the issue! I will monitor after merging to see if it helps!

## Why?

More consistent cypress tests = Yay


OllysCoding:
Part of #4264

ioannakok (approved):
👍

PR #4452, opened 2022-03-30
`ArticleHeadline` decides padding

oliverlloyd:
## What does this change?
This moves the decision about how to pad the article headline into the `ArticleHeadline` component itself.

## Why?
This makes it easier to move this component around and it also prevents the proliferation of different methods for doing the same thing. Before this refactor there were more than three methods for deciding the bottom padding for headline


dotcom-rendering/src/web/layouts/CommentLayout.tsx
```diff
@@ -427,18 +423,21 @@ export const CommentLayout = ({ CAPI, NAV, format }: Props): JSX.Element => {
...
-											tags={CAPI.tags}
-											byline={CAPI.author.byline}
-											webPublicationDateDeprecated={
-												CAPI.webPublicationDateDeprecated
-											}
-										/>
-									</div>
+									{/* TOP - we position content in groups here using flex */}
+									<ArticleHeadline
+										format={format}
```
DavidLawes, line 428: From the perspective of a layout file, I find this refactored component much easier to reason about - I think it'll be more intuitive to add a correctly padded ArticleHeadline into a new layout file too.

dotcom-rendering/src/web/layouts/headers/ImmersiveHeader.tsx
```diff
@@ -263,6 +263,10 @@ export const ImmersiveHeader = ({ CAPI, NAV, format }: Props): JSX.Element => {
 										webPublicationDateDeprecated={
 											CAPI.webPublicationDateDeprecated
 										}
+										hasStarRating={
```
DavidLawes, line 266: This extra prop is required to allow us to decide the bottom padding of the ArticleHeadline (?)

dotcom-rendering/src/web/components/ArticleHeadline.tsx
```diff
@@ -287,12 +290,85 @@ const WithAgeWarning = ({
...
+	hasStarRating?: boolean;
+	hasAvatar?: boolean;
+}) => {
+	const defaultPadding = css`
+		padding-bottom: ${space[6]}px;
+		${from.tablet} {
+			padding-bottom: ${space[9]}px;
+		}
+	`;
+	switch (format.display) {
```
DavidLawes, line 308: This switch/case is more complicated because previously not all uses of an ArticleHeadline were wrapped with an ArticleHeadlinePadding?

DavidLawes (approved):
I think this is great. I much prefer the refactored layout files and I think encapsulating all the logic for determining padding in a single place is a lot cleaner/consistent. Assuming we have stories for all our cases it looks like this functionality matches what we expect 🎉

dotcom-rendering/src/web/layouts/headers/ImmersiveHeader.tsx
oliverlloyd: Yep. When there are stars present it changes the amount of padding needed because the presence of the stars has already pushed the content down. See below

Screenshot 2022-03-30 at 14 58 29

dotcom-rendering/src/web/components/ArticleHeadline.tsx
oliverlloyd: Exactly. Previously there were something like 4 ways to decide padding, all with their own captured logic. Now we are bringing this together into one location - this switch statement.

PR #4465, opened 2022-03-30
Remove Xandr Prebid Server

chrislomaxjones:
## What does this change?

This removes the Xandr Prebid server from our AMP advert setup. This was the original Prebid server we used on AMP, and represented the control group of our ongoing AMP Prebid Server test (which was setup in #4207).

For more information on AMP Prebid servers, see the [Prebid.org documentation](https://docs.prebid.org/dev-docs/show-prebid-ads-on-amp-pages.html).

The control group is removed from the test and the test groups used by this variant are split amongst the other two variants (see the before and after tables for group allocations). Removing Xandr also allows us to remove the computation of a `placementId`.

## Why?

Our contract with Xandr expires this week. We are currently running an AB test (using the framework introduced in #4110) to evaluate the alternative we'll use. Until we collect more data, we're going to remove the control group and run a test with just two variants instead.

### Before

| Prebid Server    | Variant          | Groups          |
|------------------|------------------|-----------------|
| Xander (current) | `control`        | 0,1,2,3 (33.3%) |
| Relevant Yield   | `relevant-yield` | 4,5,6,7 (33.3%)         |
| Pubmatic         | `pubmatic`       | 8,9,10,11 (33.3%)       |

### After

| Prebid Server    | Variant          | Groups          |
|------------------|------------------|-----------------|
| Relevant Yield   | `relevant-yield` | 0,1,4,5,6,7 (50%)         |
| Pubmatic         | `pubmatic`       | 2,3,8,9,10,11 (50%)       |

### DIFF
--- a/dotcom-rendering/src/amp/components/Ad.tsx
+++ b/dotcom-rendering/src/amp/components/Ad.tsx
@@ -24,24 +24,7 @@ const ampData = (section: string, contentType: string): string => {
 	return `/${dfpAccountId}/${dfpAdUnitRoot}/amp`;
 };
 
-const preBidServerPrefix = 'https://prebid.adnxs.com/pbs/v1/openrtb2/amp';
-
 const relevantYieldURLPrefix = 'https://pbs.relevant-digital.com/openrtb2/amp';
-
-const commonPrebidUrlParams = [
-	'w=ATTR(width)',
-	'h=ATTR(height)',
-	'ow=ATTR(data-override-width)',
-	'oh=ATTR(data-override-height)',
-	'ms=ATTR(data-multi-size)',
-	'slot=ATTR(data-slot)',
-	'targeting=TGT',
-	'curl=CANONICAL_URL',
-	'timeout=TIMEOUT',
-	'adcid=ADCID',
-	'purl=HREF',
-	'gdpr_consent=CONSENT_STRING',
-];
 
 const mapAdTargeting = (adTargeting: AdTargeting): AdTargetParam[] => {
 	const adTargetingMapped: AdTargetParam[] = [];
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/amp/components/Ad.tsx
```diff
@@ -64,30 +47,40 @@ const mapAdTargeting = (adTargeting: AdTargeting): AdTargetParam[] => {
 // Variants for the Prebid server test
 // Assign each variant 4 groups e.g. 33.3% of content types each
 const variants = {
-	control: new Set([0, 1, 2, 3]),
-	'relevant-yield': new Set([4, 5, 6, 7]),
-	pubmatic: new Set([8, 9, 10, 11]),
+	'relevant-yield': new Set([0, 1, 4, 5, 6, 7]),
+	pubmatic: new Set([2, 3, 8, 9, 10, 11]),
```
chrislomaxjones, line 51: Here is where we re-allocate the groups that were originally in the control to the remaining variants.

dotcom-rendering/src/amp/components/Ad.tsx
```diff
@@ -116,18 +109,9 @@ const useRealTimeConfig = (
...
-		// The tag_id in the URL is used to look up the bulk of the request
-		// In this case it corresponds to the placement ID of the bid requests
-		// on the prebid server
-		`${preBidServerPrefix}?tag_id=${placementId}`,
-		...commonPrebidUrlParams,
-	].join('&');
+	// Not in test - dont't run Prebid
 
 	return realTimeConfig({
-		url: usePrebid ? prebidURL : undefined,
```
chrislomaxjones, line 130: If we turn off the test via a switch this will now turn off Prebid entirely. Whilst the test is still running we won't encounter this condition, as pages will be placed into the two variants above.

arelra: > If we turn off the test via a switch this will now turn off Prebid entirely

This would be bad right? ..._right?_  :)

arelra (approved):
Thanks for the really clear PR!

dotcom-rendering/src/amp/components/Ad.tsx
chrislomaxjones: Not as bad as on the web!

But yes, it'd have an impact. Still this feels like the best "safe state" option to go with in the event we needed to stop running the test immediately.

PR #4466, opened 2022-03-30
Add scrolling to pagination

iainjchambers-guardian:
## What does this change?
Adds `id` to live-block container and corresponding anchor link suffix to pagination links 

## Why?
Adds a smooth scroll to the top of the live-blocks whenever pagination has been used to navigate

### DIFF
--- a/apps-rendering/src/pagination.ts
+++ b/apps-rendering/src/pagination.ts
@@ -59,7 +59,7 @@ const maybePageRef = (
 	const maybeFirstBlock: (blocks: LiveBlock[]) => Option<LiveBlock> =
 		index(0);
 	const pageRef = (block: LiveBlock): string =>
-		`?page=with:block-${block.id}`;
+		`?page=with:block-${block.id}#liveblock-container`;
 
 	return pipe(pages, maybeBlocks, andThen(maybeFirstBlock), map(pageRef));
 };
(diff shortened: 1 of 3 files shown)


apps-rendering/src/pagination.ts
```diff
@@ -59,7 +59,7 @@ const maybePageRef = (
 	const maybeFirstBlock: (blocks: LiveBlock[]) => Option<LiveBlock> =
 		index(0);
 	const pageRef = (block: LiveBlock): string =>
-		`?page=with:block-${block.id}`;
+		`?page=with:block-${block.id}#liveblock-container`;
```
DavidLawes, line 62: I think this will work :) When we were looking previously I thought that we only scroll to liveblock-container if not the first page. Is that what's happening here?

For web this was previously defined in `frontend`. I think we decide whether to add the `#liveblock-container` to the pagination url if it's not the first page.
 
Here's where we append to the pagination url: https://github.com/guardian/frontend/blob/main/article/app/views/fragments/liveBlogNavigation.scala.html#L32

Where `isArchivePage` is false if page 1: https://github.com/guardian/frontend/blob/main/common/app/model/LiveBlogCurrentPage.scala#L210
but true otherwise:
https://github.com/guardian/frontend/blob/main/common/app/model/LiveBlogCurrentPage.scala#L216

JamieB-gu: > I thought that we only scroll to liveblock-container if not the first page.

Interesting UX question. Do you happen to know the reason why we didn't want to scroll to the blocks if the user's coming back to the first page via pagination?

cc @HarryFischer

PR #4469, opened 2022-03-30
`GetMatchNav` falls back to headline on error

oliverlloyd:
## What does this change?
If there is an error when making the call to get match nav then we now fallback to rendering the article headline

## Why?
Because sometimes the only way for us to know if match data is available is to make an api call but that call can return a 404 which is a 'valid' response and tells the client that no data was able to be returned. In this scenario we should fallback to rendering the headline

### Before
Screenshot 2022-03-30 at 17 07 41

### After
Screenshot 2022-03-30 at 17 07 59

### DIFF
--- a/dotcom-rendering/src/web/components/GetMatchNav.importable.tsx
+++ b/dotcom-rendering/src/web/components/GetMatchNav.importable.tsx
@@ -1,16 +1,29 @@
+import { css } from '@emotion/react';
+import { from } from '@guardian/source-foundations';
 import { useApi } from '../lib/useApi';
 
 import { Placeholder } from './Placeholder';
 
 import { MatchNav } from './MatchNav';
+import { ArticleHeadline } from './ArticleHeadline';
 
 type Props = {
 	matchUrl: string;
+	headlineString: string;
+	format: ArticleFormat;
+	tags: TagType[];
+	webPublicationDateDeprecated: string;
 };
 
 const Loading = () => <Placeholder height={230} />;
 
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/GetMatchNav.importable.tsx
```diff
@@ -1,16 +1,29 @@
...
 
 import { MatchNav } from './MatchNav';
+import { ArticleHeadline } from './ArticleHeadline';
 
 type Props = {
 	matchUrl: string;
+	headlineString: string;
+	format: ArticleFormat;
+	tags: TagType[];
+	webPublicationDateDeprecated: string;
```
DavidLawes, line 15: Completely tangential, and not related to this PR at all, do why this publication date is deprecated and is there another date property we should be using instead? Maybe this is part of the frontend date migration work that never got completed?

oliverlloyd: I think you're right. There was another piece of work around dates which probably involved renaming this. Not really sure why this was.

It does feel a bit odd to be using it in this PR but I don't want to introduce _changes_ here as that feels apart from what I'm doing.

oliverlloyd:
> was a bug?

Yeah, kind of. I mean, arguably we shouldn't get 'errors' from the api but the whole 'is 404 an error' debate is long... I do think it's fair to fallback to the headline in any case though so DCR not having this feature is a bug, yes.

PR #4473, opened 2022-03-30
Update lazy load margin test

zekehuntergreen:
## What does this change?
Updates duration, audience, offset, and variants of the lazy load margin test

## Why?
See description in https://github.com/guardian/frontend/pull/24839


arelra (approved):
👍

PR #4476, opened 2022-03-31
fix: Render key event bullets in the correct place in Firefox

### LINKED ISSUE #4471: Key Events Bullet Point Misalignment In Firefox
See any live or deadblog rendered by DCR or AR. The bullet points should sit on the line, alongside the timestamps.

Appears to be a Firefox-specific bug. May have been caused by a recent change because I've not seen it before.

key-events-bug

AshCorr:
## What does this change?

Seems like firefox is treating `position: absolute;` differently than other browsers. This `top` styling doesn't seem to actually do anything except break the styling on firefox.

## Why?

Closes #4471

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/21217225/161025129-12560715-c9c8-4ebf-af6d-5c86963c5c42.png
[after]: https://user-images.githubusercontent.com/21217225/161025242-67aa9f09-f26a-4275-8167-59dc2f20a5b3.png

### DIFF
--- a/common-rendering/src/components/keyEvents.tsx
+++ b/common-rendering/src/components/keyEvents.tsx
@@ -97,7 +97,6 @@ const linkStyles = (supportsDarkMode: boolean): SerializedStyles => css`
 		content: '';
 		display: block;
 		position: absolute;
-		top: 0;
 		left: -0.313rem;
 		height: 0.563rem;
 		width: 0.563rem;


OllysCoding (approved):
LGTM

PR #4479, opened 2022-03-31
Carousel margins at tablet

### LINKED ISSUE #4149: Too much margin-left on tablet and desktop breakpoint
Onward journeys tablet and desktop breakpoint have 30px margin-left. Should be 20px

oliverlloyd:
## What does this change?
This PR fixes #4149 by adjusting the margins margins for tablet

The actual change here is to include the tablet css but that then removed the need to have the switch on `format` so I simplified that function

### Before
Screenshot 2022-03-31 at 15 36 48

### After
Screenshot 2022-03-31 at 15 37 12


jlieb10 (approved):
LGTM 👍

PR #4498, opened 2022-04-01
Remove 'stick' and 'close' tracking from `YoutubeBlockComponent`

joecowton1:
## What does this change?

This PR removes the `STICK` and `CLOSE` component event tracking from `YoutubeBlockComponent`.

The companion PR in Atoms-Rendering which brings that tracking into the `YoutubeAtomSticky` component is here:
https://github.com/guardian/atoms-rendering/pull/379/files

## Why?

The benefit of this is that we keep the video player events, and the sticky video events separate, and remove the need for a conditional on the `ophanTracking` function.


marjisound (approved):
Looks great 👍

PR #4499, opened 2022-04-01
feat: Replace editors with HAL 9000

AshCorr:
## What does this change?

Replace 90% of Journalists and Editors with HAL 9000. (We still need a few for images and article layout, but no more writers!)

## Why?

This will save us millions of £ every year in staffing costs.

### After

image

Demo link: https://3030.code.ashleigh.rocks/Article?url=https://www.theguardian.com/world/2022/mar/31/why-is-gchq-saying-putin-has-been-misinformed-about-ukraine-war

### DIFF
--- a/dotcom-rendering/articles.txt
+++ b/dotcom-rendering/articles.txt
@@ -1,0 +1,438 @@
+As civilians are evacuated from the besieged city, three young people tell of their terrifying escape and how their lives have been changed for ever
+Russia-Ukraine war: latest updates
+Headshot of Veronika, wearing thick-rimmed glasses, earrings and nose ring.
+Veronika, a university student and keen ice hockey player, is now living as a refugee in a mountain village inside Ukraine.
+Supported by
+guardian.org
+About this content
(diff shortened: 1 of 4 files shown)


OllysCoding:
This is genius! Am I correct in thinking that we might not need anywhere near as much dev work invested in editorial tools either?!

AshCorr:
> Am I correct in thinking that we might not need anywhere near as much dev work invested in editorial tools either?!

Hadn't even thought of it from that angle! Yes I suppose so!

arelra:
I like that HAL9000 is updating the article on every refresh, it keeps everything up to date!

I have a small suggestion though:

Could we add the article to a blockchain on every render as a replacement for our cache? 🤔

AshCorr:
> I like that HAL9000 is updating the article on every refresh, it keeps everything up to date!

State of the art profiling, every reader sees a article tailored just for them!

AshCorr:
This was an April fools PR 😀

PR #4500, opened 2022-04-01
Hotfix: Respect timezones on DCR liveblocks while maintaining parity with designs

OllysCoding:
## What does this change?

Change DCR to use the date strings provided by frontend which is edition based - as the current solutoin will always use UTC

## Why?

Hotfixing as right now the best way to handle these kinds of date requirements in DCR are not clear

### Before
image

### After
image


AshCorr (approved):

OllysCoding:
Merging & will raise an issue to find a permanent solution to this soon

PR #4506, opened 2022-04-04
Fix `inline1` container size to reduce CLS

arelra:
### Current

The `inline1` slot is a slot where video ad providers (e.g. Teads) have first choice of whether to load an ad.

If the do not wish to fulfil the slot they 'passback' the slot for us to retry fulfilment.

Currently this process suffers from significant CLS due to three separate phases where the container changes size:

1. Initial slot is created at 160x274px and floats right
2. Teads intialises the slot at 620x350px desktop / 300x197px mobile
3. Teads _might_ passback in which case the most likely ad size will be an MPU at 320x250px

### Fix

We now reduce to CLS to potentially just one, or at most two resizes:

1. The container takes the full width and initialises its size to the incoming Teads slot size is (typically 620x350px)
2. _If_ Teads passes back, resize to the new ad height (typically an MPU at 320x250px), shifting only vertically.

### Notes

We investigated fixing the container to the largest possible slot size of 620x350px to avoid the second shift, however in the case of a passback the small ad looked odd surrounded by lots of empty space.

## Why?

To improve the reading experience for our users and reduce CLS.


ashishpuliyel:
The videos on this are great Ravi

mxdvl:
Am I right in thinking this ad is inserted dynamically, hence why there’s a need for at least one layout shift?

arelra:
> Am I right in thinking this ad is inserted dynamically, hence why there’s a need for at least one layout shift?

Yes that's right Max - I should make that clear in the description. Thanks!

PR #4507, opened 2022-04-05
Pinned Post Accessibility

abeddow91:
## What does this change?
Pinned posts should be accessible so that they can be navigated by keyboards and announced by screen readers. 
This PR adds the following accessibility enhancements: 
 - `keydown `event listener on the pinned post button so that the pinned post can be expanded / collapsed on enter.
- moves the button above the collapsible body (and adds the relevant z-index) so that the button is tabbed to first. 
- adds `aria-label` to the pinned post button for screen readers.


dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
```diff
@@ -87,6 +94,13 @@ export const EnhancePinnedPost = () => {
 		};
 	}, []);
 
+	useEffect(() => {
+		pinnedPostButton?.addEventListener('keydown', keyListener);
+
+		return () => {
+			pinnedPostButton?.removeEventListener('keydown', keyListener);
+		};
+	}, []);
```
mxdvl, line 103: We’ve got the `useOnce` pattern for this kind of thing, makes it clearer that this will only execute once.

dotcom-rendering/src/web/components/PinnedPost.tsx
```diff
@@ -110,6 +110,7 @@ const fakeButtonStyles = (palette: Palette) => css`
 	${from.tablet} {
 		margin-left: 60px;
 	}
+	z-index: 1;
```
mxdvl, line 113: Generally, we’ll use the `getZIndex` method to specify indexes globally.

dotcom-rendering/src/web/components/PinnedPost.tsx
```diff
@@ -154,23 +155,16 @@ export const PinnedPost = ({ pinnedPost, children, format }: Props) => {
...
-			</div>
-			<div id="collapsible-body" css={collapsibleBody}>
-				{children}
-			</div>
-			<div id="pinned-post-overlay" css={overlayStyles} />
 			<label
 				css={fakeButtonStyles(palette)}
 				htmlFor="pinned-post-checkbox"
 				id="pinned-post-button"
+				// eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
```
mxdvl, line 165: Can you explain how this is an interactive `tabindex`? Or why the rule is required?

```ts
// eslint-disable-next-line rule-to-disable -- reason for disabling the rule
```

mxdvl (approved):
Generally happy with this, with some small comments for future travellers.

jamesgorrie:
@abeddow91 do we know where we are with this? Just asking as it's been marked as stale (which this comment should undo).

PR #4510, opened 2022-04-05
Poll for latest stats

oliverlloyd:
## What does this change?
This PR adds polling for sport live blogs

## Why?
So that we show the latest match stats without the page needing to be refreshed.


OllysCoding (approved):
+1 Looks good to me!

PR #4511, opened 2022-04-05
Show black text on yellow backgrounds

oliverlloyd:
## What does this change?
Changes the colour of the section title for live sport blogs

## Why?
It was appearing as white on a yellow background which was hard to read

### Before
Screenshot 2022-04-05 at 16 18 12

### After
Screenshot 2022-04-05 at 16 19 16


OllysCoding (approved):

AshCorr (approved):
Simples!

PR #4519, opened 2022-04-05
Content Security Policy: add `data:` scheme to `img-src`

MarSavar:
## Why?
It is currently not possible to render SVGs in Apps Rendering via CSS using `data: svg+xml`.

This is exactly how the squiggly/dotted/dashed variations of the `<Lines>` component are rendered: they use [`mini-svg-data-uri`](https://www.npmjs.com/package/mini-svg-data-uri) under the hood, which converts SVGs into compressed `data:` URIs, resulting in `background-image: url(data:svg+xml,...)`.

AR's current Content Security Policy`img-src` directive only allows 4 origins: 

- `'self'`
- `https://static.theguardian.com`
- `https://*.guim.co.uk`
- Optionally, Twitter (but only if the article contains a Twitter embed)

The `data:` scheme is excluded from the allowed origins, meaning that SVGs rendered in that way will not be displayed in AR. This is a problems because we need to use variations of the `<Lines>` component (the default straight lines used in most articles are rendered with `background-image: repeating-linear-gradient` instead of an SVG, which is why we never noticed the issue before).

@sndrs confirmed that we do not need to be this strict with our CSP and we can allow `data:`.
This PR therefore adds it to both AR and Editions' Content Security Policies.

### DIFF
--- a/apps-rendering/src/server/csp.ts
+++ b/apps-rendering/src/server/csp.ts
@@ -95,7 +95,7 @@ const buildCsp = (
 	`
     default-src 'self';
     style-src ${styleSrc(styles, thirdPartyEmbed.twitter, hasInlineStyles)}
-    img-src 'self' https://static.theguardian.com https://*.guim.co.uk ${
+    img-src 'self' data: https://static.theguardian.com https://*.guim.co.uk ${
 		thirdPartyEmbed.twitter
 			? 'https://platform.twitter.com https://syndication.twitter.com https://pbs.twimg.com data:'
 			: ''
@@ -135,7 +135,7 @@ function buildCspEditions(
 			? 'https://platform.twitter.com https://cdn.syndication.twimg.com'
 			: ''
 	};
-	img-src 'self' https://static.theguardian.com https://*.guim.co.uk ${
+	img-src 'self' data: https://static.theguardian.com https://*.guim.co.uk ${
 		thirdPartyEmbed.twitter
 			? 'https://platform.twitter.com https://syndication.twitter.com https://pbs.twimg.com data:'
 			: ''


mxdvl (dismissed):
Excellent description.

mxdvl:
Incoming fix in Source via https://github.com/guardian/source/pull/1336

sndrs:
this seems like a great idea to me. we need a case to enable this – the engineering effort (and therefore money) to get around it was a lot greater than this PR. given dcr and most of the internet seems happy with permitting this, i don't see why AR can't too (given what it allows us to do)

sndrs (approved):

JamieB-gu:
> given what it allows us to do

I thought #4703 meant that we no longer needed this?

sndrs:
it was an unnecessary work around (this would have sufficed and was far simpler). the restrictions will still apply in any other future instance, therefore to my mind this is still unnecessarily inhibiting the organisation and we should not do that

JamieB-gu:
> it was an unnecessary work around (this would have sufficed and was far simpler). the restrictions will still apply in any other future instance, therefore to my mind this is still unnecessarily inhibiting the organisation and we should not do that

I understand that's your position on this, and that's completely fair, but I don't think it was the only one? Based on the discussion we've had elsewhere I'm not sure we came to a consensus on this, and perhaps we shouldn't push this through without that?

In this case the original requirement for making this change has been solved elsewhere, in #4703 as mentioned. If another requirement comes up that could be met by making this change then I think we should revisit it, but I don't think we should make it without that. I also think the document that Simon wrote discussing a process/framework for these kind of changes will be helpful when we come to make decisions on these situations in the future.

mxdvl:
> this seems like a great idea to me. we need a case to enable this – the engineering effort (and therefore money) to get around it in [guardian/source#1336](https://github.com/guardian/source/pull/1336) was a lot greater than this PR.

> it was an unnecessary work around (this would have sufficed and was far simpler). the restrictions will still apply in any other future instance, therefore to my mind this is still unnecessarily inhibiting the organisation and we should not do that

Speaking in terms of engineering effort, I think that one could argue that the code should have _always_ been written as pure SVG for presentation, rather than relying on background-images. Having an empty `div` instead of an actual SVG was a worse situation to be in, in my opinion, and I’m happy that we were nudged to actually revisit this decision.

In that sense, I’m happy that the `img-src` directive does not include `data:` in apps-rendering.

sndrs:
> Having an empty div instead of an actual SVG was a worse situation to be in, in my opinion, and I’m happy that we were nudged to actually revisit this decision.

that's a really good outcome, but i'd say it was more a happy side effect :)

SiAdcock:
As it stands, the following is true:

- you need to come up with workarounds / alternatives to using the `data` URI
- you spend time debugging / finding workarounds to issues with third party libraries using the `data` URI. Think Source, but also think third party integrations such as atoms and interactives
- you restrict your choices of UI libraries / frameworks to those that don't use the `data` URI, or you have to request changes to those libraries in order to use them

There's a strong case to be made that this a pragmatic change, likely to save developer- and potentially designer-time in the future.

On the other hand, it would be good to understand the danger in making this change. Many websites, including theguardian.com, have taken the step to remove this restriction, and it is not clear that there have been any negative consequences as a result. What kind of attacks are we protecting the apps from by keeping this restriction in place?

SiAdcock:
Thanks for helping me to understand the issues with this directive, @JamieB-gu 🙏 

> I'll preface this by saying that one thing I'm a little unclear on is the differences between using this for img-src vs something like script-src; neither the spec nor MDN seem to discuss this explicitly apart from a brief mention in the CSP sources docs on MDN.

It looks like we found the same sources in our research, and came to the same conclusion. I couldn't find anything to suggest the directive was unsafe when used in the context of `img-src`. My original statement still stands:

> Therefore to me, this looks like we're slowing down developers in order to preserve a default security setting that does nothing to protect us or our users.

You asked how often usages of this come up in practice. It's hard to put a number on this, and you have pointed out that it has come up in Source recently. We had to change Source to make in order to satisfy the requirements of apps-rendering. An imperfect [search of the Guardian org](https://github.com/search?q=org%3Aguardian+data%3Aimage&type=code) reveals a significant number of uses of `data:image`, including in interactives, which I gather apps-rendering will have to support.

One point of pragmatism, as this conversation draws on: it strikes me that if there were an immediate use case for allowing the `data:` scheme within the `img-src` directive, then this PR would be approved. If this change would be approved for any reason, then there's no benefit in not approving it right now. Time spent debugging, requesting changes to shared libraries like Source and potentially interactives in the future, finding alternatives to valid usages within apps-rendering and indeed time spent on conversations like this could all be better spent on delivering value for our users. Lacking a reason to not approve this change, I would say let's just approve it and move on.

> I'm not too sure what you're referring to here, is there a particular browser setting you had in mind?

Apologies for the lack of clarity, I was inarticulately referring to this [Mozilla security blog post](https://blog.mozilla.org/security/2017/11/27/blocking-top-level-navigations-data-urls-firefox-59/):

> To mitigate such issues, top-level navigation to data: URLs is blocked in all modern browsers. See [this blog post from the Mozilla Security Team](https://blog.mozilla.org/security/2017/11/27/blocking-top-level-navigations-data-urls-firefox-59/) for more details.

jfsoul:
Hello! Just to add my 2c:

> I couldn't find anything to suggest the data: scheme was unsafe when used in the context of img-src. My original statement still stands:
> 
> > Therefore to me, this looks like we're slowing down developers in order to preserve a default security setting that does nothing to protect us or our users.

I also can't find any evidence from my searches that this would be unsafe. However, a lack of evidence of vulnerability is certainly not proof of safety! Indeed, even if there are currently no known exploits, that does not mean there will not be one in the future. So while I tend to agree with the sentiment of the whole statement, I would not say the setting does "nothing" to protect us or our users. Defence in depth is a worthy aim and we are right to try and balance the relative risk vs. the benefits of disabling a security feature.

With regards to trying to balance the risk vs. benefit - I don't think I can add much more on the benefit side of things that hasn't already been said. On the risk side, if we felt further due diligence was necessary we could:

- Have a conversation with InfoSec and other stakeholders. I have a meeting with InfoSec later and can bring this up. They will likely not give an absolute decision, but their advice on weighing up these decisions is always useful as ultimately this is a business decision.
- [Threat model](https://martinfowler.com/articles/agile-threat-modelling.html)/assess the potential vulnerabilities:
  - What are the potential vulnerabilities? (XSS has been mentioned?)
  - In what scenarios are those actually exploitable? (e.g. when would a malicious actor be able to control our image sources?)
  - What is the impact of those?
  - What mitigations can we implement if any?

If you believe the above would be useful and you'd like for us to be involved, then let us know. Either way it might be useful to summarise the context, risks and benefits in a document to share outside the people in this PR.

sndrs:
>Time spent debugging, requesting changes to shared libraries like Source and potentially interactives in the future, finding alternatives to valid usages within apps-rendering and indeed time spent on conversations like this could all be better spent on delivering value for our users.

💯 

https://www.theguardian.com/uk permits `data:` for `frame-src`, `img-src`, `media-src` and `font-src`.

Here's how some other organisations approach this:

<table>
<tr>
	<th>Domain
	<th><code>data:</code> scheme
<tr>
	<td>https://www.nytimes.com/
	<td><code>script-src</code>, <code>style-src</code>, <code>img-src</code>, <code>font-src</code>, <code>media-src</code>
<tr>
	<td>https://www.bbc.co.uk/news
	<td>no CSP
<tr>
	<td>https://www.ft.com/
	<td>no CSP
<tr>
	<td>https://www.telegraph.co.uk/
	<td>no CSP
<tr>
	<td>https://twitter.com/home
	<td><code>img-src</code>
<tr>
	<td>https://www.amazon.co.uk
	<td><code>default-src</code>
<tr>
	<td>https://www.facebook.com
	<td><code>default-src</code>, <code>style-src</code>, <code>font-src</code>, <code>img-src</code>, <code>media-src</code>, <code>frame-src</code>, <code>worker-src</code>
</table>

There is a clear case _for_ merging it (rendering our actual journalism in the case of interactives) and I can see none against that wouldn't also apply to dotcom, other major news organisations and the biggest sites on the web.

Without a compelling reason that reading articles in our app needs stricter security constraints than any of the other examples, afaict merging this PR only enables us to work better

JamieB-gu:
Thanks for adding your thoughts @jfsoul! On this:

>  I have a meeting with InfoSec later and can bring this up.

If you have time this would be great, I would find their thoughts on this interesting. As much on the process for dealing with these questions when they come up as on content of this specific issue.

> including in interactives, which I gather apps-rendering will have to support.

@SiAdcock this is a good point, and as I mentioned above I think we may well want to revisit this when we get to interactives. One interesting point there is that content embedded via `iframe[src]` is subject to its own CSP, rather than the CSP of the embedding document, as described in the "Embedded Contexts" section here: https://www.w3.org/TR/CSP2/#which-policy-applies . This allows any interactives that are loaded in this way to use a policy from, for example, `interactive.guim.co.uk`, which can be different from the policy of our main document and written based on the specific requirements of interactives.

> if there were an immediate use case that required allowing the `data:` scheme within the `img-src` directive, then this PR would be merged. If this change would be merged for any reason, then there's no benefit in not merging it right now.

I'm not sure I follow this argument 🤔. Are you suggesting we merge this pre-emptively in case we have a reason to at some point down the line? Would it not make sense to introduce this change _when_ we have a specific reason to do so, for example if it's required to support interactives? Then we have a concrete requirement with which to assess the balance that Jon mentioned:

> balance the relative risk vs. the benefits of disabling a security feature

The "benefit in not merging it right now" that you describe would be keeping a security feature enabled?

SiAdcock:
Thanks so much for your input @jfsoul, it has helped me to better understand the concerns. Defence in depth sounds like something I should familiarise myself with. I'd also be interested in any information about this specific issue from InfoSec, as well as issues like this. As a library maintainer, it is not always clear which features would force a change to a consuming application's CSP, and what the resolution would be. 

@JamieB-gu I think I'm going to bow out of this discussion for now. I'm only interested in helping improve the developer experience of working on apps-rendering, and it doesn't feel like I'll be able to help this time around.

JamieB-gu:
> I think I'm going to bow out of this discussion for now. I'm only interested in helping improve the developer experience of working on apps-rendering, and it doesn't feel like I'll be able to help this time around.

No worries Simon, your thoughts on this were very helpful, thanks for taking the time to discuss! 🙂

PR #4526, opened 2022-04-06
Rename files to be more articlely

oliverlloyd:
## What does this change?
Renames

`document` to `articleToHtml`
`htmlTemplate` to `articleTemplate`
and
`Page` to `Article`

## Why?
To mark these parts of the DCR codebase as article specific

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -744,44 +744,6 @@ interface ConfigType extends CommercialConfigType {
 	isPreview?: boolean;
 }
 
-interface ConfigTypeBrowser {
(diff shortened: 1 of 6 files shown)


dotcom-rendering/index.d.ts
```diff
@@ -744,44 +744,6 @@ interface ConfigType extends CommercialConfigType {
 	isPreview?: boolean;
 }
 
-interface ConfigTypeBrowser {
```
oliverlloyd, line 747: This is a drive by change. Turns out this type is dead. It most likely died during the `App.tsx` wars.

mxdvl (approved):
Clearer!

PR #4527, opened 2022-04-06
Rename `CAPI` to be more articlely

oliverlloyd:
## What does this change?
This PR renames:

`CAPI` > `CAPIArticle`
and
`CAPIType` to `CAPIArticleType`

## Why?
So there's a clearer distinction between article data and other types of data

## But the word `capi` still, really?
I know, I did consider dropping this prefix in this PR but because there are a lot of diffs happening here I thought it safer to try and keep things as simple as possible. After this PR is merged we will have a more specific name `CAPIArticle` which will be a lot easier to change in future (it was hard before because the lexicon of 'CAPI' is written throughout the documentation.

### DIFF
--- a/dotcom-rendering/docs/architecture/025-click-to-view.md
+++ b/dotcom-rendering/docs/architecture/025-click-to-view.md
@@ -57,18 +57,18 @@ rendered, there is no need to hydrate the component.
 Here is a list of elements in the 'frontend' article json model that can contain third party content with details
 of the decision to wrap them in ClickToView or not.
 
-Element | Wrapped in Click to View | Details
---------|--------|-------
-DocumentBlockElement | yes | Can contain both Scribd (tracking) and DocumentCloud (non-tracking)
-EmbedBlockElement | yes | Can contain almost anything including content which has its own specific element type eg youtube, twitter
-InstagramBlockElement | yes | Facebook
-MapBlockElement | yes | maps are provided by google who track
-PullquoteBlockElement | yes | It is possible to include an iframe in a pullquote, its arguable if this should be wrapped in click to view as this rarely if ever happens
-SoundcloudBlockElement | no | Soundcloud claim to not track users
-SpotifyBlockElement | yes | Spotify track users
-TweetBlockElement | no | twitter support a [do-not-track](https://developer.twitter.com/en/docs/twitter-for-websites/privacy) feature
-VideoBlockElement | no | These contain videos from random providers which may track. However we do not currently render these videos 😱
-VideoFacebookBlockElement | yes | Facebook
-VideoVimeoBlockElement | no | Vimeo provide a 'do-not-track' feature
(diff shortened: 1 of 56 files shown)


OllysCoding:
I'd be hesitant to keep `CAPI` as part of the name of these types.

I think `CAPI` has always been inherently misleading ~ our types in DCR do not really reflect that of CAPI, the data-structure we use is heavily modified and created from scratch in Frontend. 

I'd recommend naming it something like `Article` and `ArticleType` or similar.

mxdvl (approved):
Yes, and I think we should drop the `CAPI` prefix.

oliverlloyd:
> I'd be hesitant to keep `CAPI` as part of the name of these types.
> 
> I think `CAPI` has always been inherently misleading ~ our types in DCR do not really reflect that of CAPI, the data-structure we use is heavily modified and created from scratch in Frontend.
> 
> I'd recommend naming it something like `Article` and `ArticleType` or similar.

I do agree on this - CAPI _is_ misleading - but there are three reasons I kept this particular change out of this PR:

1) It's polluted the diff. It's a lot easier visually inspecting `CAPI` > `CAPIArticle`
2) It felt like a different concern. This PR is about making the names article specific. But the change from CAPI is - a bit - different. At least in my head it felt that way.
3) I did try with `Article` as that initially felt logical but this caused a lot of errors. We already use the `Article` namespace in various places and having two things with the same name is undesirable.

oliverlloyd:
> Yes, and I think we should drop the `CAPI` prefix.

So yes. I'm not against this. But I'd like to make that change in a different PR. As per my description, you will find that this name change is now (with this PR) going to be a lot easier to do vs. before when it would have been messy

mxdvl:
Agreed this would be better in a subsequent PR, but how could we settle on a name? I quite like words like `Model`, `Data`, `Blueprint`, `Shape` and `Scheme`.

oliverlloyd:
> Agreed this would be better in a subsequent PR, but how could we settle on a name? I quite like words like `Model`, `Data`, `Blueprint`, `Shape` and `Scheme`.

I think the name should have 'Article' in it somehow...  I considered `ArticleData` but it isn't great...

Ideally we'd use `Article` - that feels right - but this clashes with two other things. Maybe we can first do a PR to rename _those_ things and then make the change?

Or is it just me that feels 'Article'? I guess whatever we pick will end up being normalised but it's perhaps nice to have a thing that's easily understood by people new to the code.

JamieB-gu:
Same question came up in AR recently; there the term `Item` is used for the equivalent data representation of an article. It's named this partly because the MAPI service that handles articles is called `mobile-item`.

However, I'm not entirely happy with `Item`, I think it's a bit too vague a term. `Article` is also the name I'd like to switch to, but similar problem: it's a _lot_ of stuff to re-write.

PR #4531, opened 2022-04-06
feat: colours in Fronts JSON data

mxdvl:
## What does this change?

Add colours and styles to the JSON display.

## Why?

We also get a lot more than just config, so we can investigate in more detail.

image


oliverlloyd (approved):
Fancy

ioannakok (approved):

PR #4532, opened 2022-04-06
`FrontType`

oliverlloyd:
## What does this change?
This PR adds `FrontType`, the type that we use for data sent to use for a front page by Frontend

## Why?
So we can check we get what we expect and also to help developers predict what values will be available to them.

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -513,6 +513,123 @@ interface CAPIArticleType {
 	mostRecentBlockId?: string;
 }
 
+interface FrontType {
+	pressedPage: PressedPageType;
+	nav: CAPINavType;
+	editionId: string;
+	editionLongForm: string;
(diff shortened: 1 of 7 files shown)


ioannakok (approved):
👍 Looks good to me!

dotcom-rendering/index.d.ts
```diff
@@ -513,6 +513,123 @@ interface CAPIArticleType {
...
+	externalEmbedHost: string;
+	ajaxUrl: string;
+	keywords: string;
+	revisionNumber: string;
+	isProd: boolean;
+	switches: Switches;
+	section: string;
+	keywordIds: string;
+	locationapiurl: string;
+	sharedAdTargeting: { [key: string]: any };
```
mxdvl, line 561: I see we’ve used `Record<string, unknown>` above. We might want to keep these consistent throughout?

PR #4543, opened 2022-04-07
A basic Front page

oliverlloyd:
## What does this change?
This PR adds the files and basic contents that we need to render a Front page

## Why?
So that we have the data being passed through and better idea of the gaps we need to fill

Screenshot 2022-04-07 at 15 06 43

### DIFF
--- a/dotcom-rendering/src/model/window-guardian.ts
+++ b/dotcom-rendering/src/model/window-guardian.ts
@@ -169,7 +169,7 @@ const makeFrontWindowGuardianConfig = ({
 };
 
 export const makeFrontWindowGuardian = (
-	Front: FrontType,
+	front: FrontType,
 ): {
 	// The 'config' attribute is derived from DCRServerDocumentData and contains
 	// all the data that, for legacy reasons, for instance compatibility
@@ -185,7 +185,7 @@ export const makeFrontWindowGuardian = (
 	};
 } => {
 	return {
-		config: makeFrontWindowGuardianConfig(Front),
+		config: makeFrontWindowGuardianConfig(front),
 		polyfilled: false,
 		adBlockers: {
 			active: undefined,
--- a/dotcom-rendering/src/model/window-guardian.ts
+++ b/dotcom-rendering/src/model/window-guardian.ts
@@ -1,5 +1,3 @@
-type StageType = 'DEV' | 'CODE' | 'PROD';
-
 export interface WindowGuardianConfig {
 	isDotcomRendering: boolean;
 	isDev: boolean;
@@ -20,6 +18,37 @@ export interface WindowGuardianConfig {
 		ajaxUrl: string;
 		shouldHideReaderRevenue: boolean;
 	} & ConfigType;
+	libs: {
+		googletag: string;
+	};
+	switches: { [key: string]: boolean };
+	tests: ServerSideTests;
+	ophan: {
+		pageViewId: string;
+		browserId: string;
+	};
+}
+
+interface WindowGuardianFrontConfig {
+	isDotcomRendering: boolean;
+	isDev: boolean;
+	stage: StageType;
+	frontendAssetsFullURL: string;
+	page: {
+		dcrCouldRender: boolean;
+		contentType: string;
+		edition: Edition;
+		revisionNumber: string;
+		dcrSentryDsn: string;
+		sentryHost: string;
+		sentryPublicApiKey: string;
+		keywordIds: string;
+		dfpAccountId: string;
+		adUnit: string;
+		showRelatedContent: boolean;
+		ajaxUrl: string;
+		shouldHideReaderRevenue: boolean;
+	};
 	libs: {
 		googletag: string;
 	};
@@ -103,7 +132,7 @@ export const makeWindowGuardian = (
 
 const makeFrontWindowGuardianConfig = (
 	Front: FrontType,
-): WindowGuardianConfig => {
+): WindowGuardianFrontConfig => {
 	const { config } = Front;
 	return {
 		// This indicates to the client side code that we are running a dotcom-rendering rendered page.
@@ -125,6 +154,7 @@ const makeFrontWindowGuardianConfig = (
 			adUnit: config.adUnit,
 			showRelatedContent: true,
 			ajaxUrl: config.ajaxUrl,
+			shouldHideReaderRevenue: false, // TODO Pass this in
 		},
 		libs: {
 			googletag: config.googletagUrl,
@@ -135,7 +165,7 @@ const makeFrontWindowGuardianConfig = (
 			pageViewId: '',
 			browserId: '',
 		},
-	} as WindowGuardianConfig;
+	};
 };
 
 export const makeFrontWindowGuardian = (
@@ -145,7 +175,7 @@ export const makeFrontWindowGuardian = (
 	// all the data that, for legacy reasons, for instance compatibility
 	// with the frontend commercial stack, or other scripts, we want to find
 	// at window.guardian.config
-	config: WindowGuardianConfig;
+	config: WindowGuardianFrontConfig;
 	polyfilled: boolean;
 	adBlockers: any;
 	modules: {
--- a/dotcom-rendering/src/model/window-guardian.ts
+++ b/dotcom-rendering/src/model/window-guardian.ts
@@ -100,3 +100,71 @@ export const makeWindowGuardian = (
 		GAData: dcrDocumentData.GA,
 	};
 };
+
+const makeFrontWindowGuardianConfig = (
+	Front: FrontType,
+): WindowGuardianConfig => {
+	const { config } = Front;
+	return {
+		// This indicates to the client side code that we are running a dotcom-rendering rendered page.
+		isDotcomRendering: true,
+		isDev: process.env.NODE_ENV !== 'production',
+		stage: config.stage,
+		frontendAssetsFullURL: config.frontendAssetsFullURL,
+		page: {
+			dcrCouldRender: true,
+			contentType: 'TODO: Do we need this?',
+			edition: Front.editionId,
+			revisionNumber: config.revisionNumber,
+			dcrSentryDsn:
+				'https://1937ab71c8804b2b8438178dfdd6468f@sentry.io/1377847',
+			sentryPublicApiKey: config.sentryPublicApiKey,
+			sentryHost: config.sentryHost,
+			keywordIds: config.keywordIds,
+			dfpAccountId: config.dfpAccountId,
+			adUnit: config.adUnit,
+			showRelatedContent: true,
+			ajaxUrl: config.ajaxUrl,
+		},
+		libs: {
+			googletag: config.googletagUrl,
+		},
+		switches: config.switches,
+		tests: config.abTests,
+		ophan: {
+			pageViewId: '',
+			browserId: '',
+		},
+	} as WindowGuardianConfig;
+};
+
+export const makeFrontWindowGuardian = (
+	Front: FrontType,
+): {
+	// The 'config' attribute is derived from DCRServerDocumentData and contains
+	// all the data that, for legacy reasons, for instance compatibility
+	// with the frontend commercial stack, or other scripts, we want to find
+	// at window.guardian.config
+	config: WindowGuardianConfig;
+	polyfilled: boolean;
+	adBlockers: any;
+	modules: {
+		sentry: {
+			reportError: (error: Error, feature: string) => void;
+		};
+	};
+} => {
+	return {
+		config: makeFrontWindowGuardianConfig(Front),
+		polyfilled: false,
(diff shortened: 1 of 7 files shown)


dotcom-rendering/src/model/window-guardian.ts
```diff
@@ -100,3 +100,71 @@ export const makeWindowGuardian = (
 		GAData: dcrDocumentData.GA,
 	};
 };
+
+const makeFrontWindowGuardianConfig = (
+	Front: FrontType,
+): WindowGuardianConfig => {
```
mxdvl, line 106: I think we typically keep variables lowercase?

```suggestion
const makeFrontWindowGuardianConfig = (
	front: FrontType,
): WindowGuardianConfig => {
```

But we could also just de-structure the bits we’re using: 

```suggestion
const makeFrontWindowGuardianConfig = ({
	config,
	editionId,
}: FrontType): WindowGuardianConfig => {
```

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -0,0 +1,146 @@
...
+						backgroundColour={brandBackground.primary}
+						element="nav"
+					>
+						<Nav
+							nav={NAV}
+							format={format}
+							subscribeUrl={
+								Front.nav.readerRevenueLinks.header.subscribe
+							}
+							edition={Front.editionId as Edition}
```
mxdvl, line 76: Could we fix it upstream in the Fronts type?

dotcom-rendering/src/web/server/frontTemplate.ts
```diff
@@ -0,0 +1,248 @@
...
+	// "use preconnect when you know for sure that you’re going to be accessing a resource"
+	// - https://www.smashingmagazine.com/2019/04/optimization-performance-resource-hints/
+	// DNS-prefetch should be used for other third party domains that we are likely to connect to but not sure (ads)
+	// Preconnecting to too many URLs can reduce page performance
+	// DNS-prefetch can also be used as a fallback for IE11
+	// More information on preconnecting:
+	// https://css-tricks.com/using-relpreconnect-to-establish-network-connections-early-and-increase-performance/
+	// More information on prefetching:
+	// https://developer.mozilla.org/en-US/docs/Web/Performance/dns-prefetch
+	const staticPreconnectUrls = [
```
mxdvl, line 52: Can we move to JSDoc?

dotcom-rendering/src/web/server/frontTemplate.ts
```diff
@@ -0,0 +1,248 @@
...
+						apiName: 'require'
+					};
+					window.curl = window.curlConfig;
+				</script>
+
+
+                <noscript>
+                    <img src="https://sb.scorecardresearch.com/p?c1=2&c2=6035250&cv=2.0&cj=1&cs_ucfr=0&comscorekw=${encodeURIComponent(
+						keywords,
+					).replace(/%20/g, '+')}" />
```
mxdvl, line 234: Using the `URLSearchParams` constructor would solve the need to replace `%20`:

```suggestion
                    <img src="https://sb.scorecardresearch.com/p?${new URLSearchParams({
						c1:"2",
						c2: "6035250",
						cv: "2.0",
						cj: "1",
						cs_ucfr: "0",
						comscorekw: keywords,
					})" />
```

dotcom-rendering/src/web/server/frontTemplate.ts
```diff
@@ -0,0 +1,248 @@
...
+                ${priorityScriptTags.join('\n')}
+                <style class="webfont">${getFontsCss()}</style>
+                <style>${resets.resetCSS}</style>
+				${css}
+				<link rel="stylesheet" media="print" href="${ASSET_ORIGIN}static/frontend/css/print.css">
+			</head>
+
+			<body>
+                ${html}
+                ${[...lowPriorityScriptTags].join('\n')}
```
mxdvl, line 245: Why are we spreading this array here?

dotcom-rendering/src/web/server/frontToHtml.tsx
```diff
@@ -1,79 +1,145 @@
...
+			default:
+				attrs = 'defer';
+				break;
+		}
+
+		return [
+			...scriptTags,
+			`<script ${attrs} src="${script.src}"></script>`,
+		];
+	}, []);
```
mxdvl, line 41: Isn’t this method entirely duplicated?

dotcom-rendering/src/web/server/frontToHtml.tsx
```diff
@@ -1,79 +1,145 @@
...
+		'https://assets.guim.co.uk/static/frontend/fonts/guardian-textegyptian/noalts-not-hinted/GuardianTextEgyptian-Regular.woff2',
+		// 'https://assets.guim.co.uk/static/frontend/fonts/guardian-textegyptian/noalts-not-hinted/GuardianTextEgyptian-RegularItalic.woff2',
+		'https://assets.guim.co.uk/static/frontend/fonts/guardian-textegyptian/noalts-not-hinted/GuardianTextEgyptian-Bold.woff2',
+		'https://assets.guim.co.uk/static/frontend/fonts/guardian-textsans/noalts-not-hinted/GuardianTextSans-Regular.woff2',
+		// 'http://assets.guim.co.uk/static/frontend/fonts/guardian-textsans/noalts-not-hinted/GuardianTextSans-RegularItalic.woff2',
+		'https://assets.guim.co.uk/static/frontend/fonts/guardian-textsans/noalts-not-hinted/GuardianTextSans-Bold.woff2',
+	];
+
+	const polyfillIO =
+		'https://assets.guim.co.uk/polyfill.io/v3/polyfill.min.js?rum=0&features=es6,es7,es2017,es2018,es2019,default-3.6,HTMLPictureElement,IntersectionObserver,IntersectionObserverEntry,URLSearchParams,fetch,NodeList.prototype.forEach,navigator.sendBeacon,performance.now,Promise.allSettled&flags=gated&callback=guardianPolyfilled&unknown=polyfill&cacheClear=1';
```
mxdvl, line 80: Do we want this do be defined independently of Articles?

dotcom-rendering/src/web/server/frontToHtml.tsx
```diff
@@ -1,79 +1,145 @@
...
+	 * is placed in a script tag on the page
+	 */
+	const windowGuardian = escapeData(
+		JSON.stringify(makeFrontWindowGuardian(Front)),
+	);
+
+	const keywords =
+		typeof Front.config.keywords === 'undefined'
+			? ''
+			: Front.config.keywords;
```
mxdvl, line 132: This can be replaced inline with a nullish coalescing operator:

```suggestion
	const keywords = Front.config.keywords ?? '';
```

dotcom-rendering/src/web/server/index.ts
```diff
@@ -181,14 +181,14 @@ export const renderKeyEvents = (
 };
 
 export const renderFront = (
-	{ body, query }: express.Request,
+	{ body }: express.Request,
 	res: express.Response,
 ): void => {
 	try {
 		const enhanced = enhanceFront(body);
```
mxdvl, line 188: Maybe this could be called `Front`, and maybe we could keep it lowercase?

mxdvl (commented):
This is amazing!

It’s interesting to see the emerging similarities and differences between articles and fronts. I’ve left some comments as I read through the code. There’s a few places where I **do not think** we should duplicate data, such as defining our Polyfill.io bundle. Maybe we can think of these as shared configuration? Exciting questions!

dotcom-rendering/src/model/window-guardian.ts
```diff
@@ -100,3 +100,71 @@ export const makeWindowGuardian = (
...
+		libs: {
+			googletag: config.googletagUrl,
+		},
+		switches: config.switches,
+		tests: config.abTests,
+		ophan: {
+			pageViewId: '',
+			browserId: '',
+		},
+	} as WindowGuardianConfig;
```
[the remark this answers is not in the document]
oliverlloyd: I fixed this by improving the types so now we don't need to cast or ignore errors

dotcom-rendering/src/web/layouts/FrontLayout.tsx
oliverlloyd: Done

dotcom-rendering/src/web/server/frontTemplate.ts
oliverlloyd: Done

dotcom-rendering/src/web/server/frontTemplate.ts
oliverlloyd: Is this one okay to do in a separate PR. This seems like a change that could be done as part of rationalising any duplicated code but it comes with a risk of changing behaviour so would benefit from the tighter focus of a dedicated PR

oliverlloyd: Scratch that. I think your suggestion is solid 👍

dotcom-rendering/src/web/server/frontTemplate.ts
oliverlloyd: This is copied code so is probably outside the scope of this PR but I think perhaps it ended up this way for legacy reasons? That is, maybe an earlier version had more than one item in the array?

dotcom-rendering/src/web/server/frontToHtml.tsx
oliverlloyd: Yeah, I do think things like this should be extracted but I'd like to keep that work separate. I see this duplication work as a way to learn where we both have gaps and also where we have common code. It's good - I think - to be cautious making abstractions as there's potentially cases where we might think we have a common use case but then learn later that there are differences.

dotcom-rendering/src/web/server/index.ts
oliverlloyd: Agreed. Both on dropping 'enhanced' and on using lowercase 👍

dotcom-rendering/src/model/window-guardian.ts
```diff
@@ -31,6 +29,37 @@ export interface WindowGuardianConfig {
 	};
 }
 
+interface WindowGuardianFrontConfig {
```
jamesgorrie, line 32: bikedshed: Do we have a rule of when to use types or interfaces? Just asking as this looks like like a type subjectively.

dotcom-rendering/src/web/server/frontTemplate.ts
jamesgorrie: If it wants tech-debt eradicating probably worth merging this and change it.

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -0,0 +1,146 @@
...
+						</>
+					)}
+				</>
+			</div>
+
+			<main>
+				{front.pressedPage.collections.map((collection, index) => {
+					return (
+						<ContainerLayout
+							title={collection.displayName as string}
```
mxdvl, line 110: Is this still needed?

jamesgorrie (approved):
Not to override any ones opinions, just saying I personally agree with this approach, happy if people want the conversation to continue.

I do think copy/pasting will give us more speed and help us make sure that if we abstract anything, we do it with thought, not just because it looked the same after replicating part of a project to start iterating on it.

It could stop us making optimisations in the name of consistency too, which isn't great. 

Rule of three has many reasons for it, as does not abstracting early.

dotcom-rendering/src/web/layouts/FrontLayout.tsx
oliverlloyd: Good spot, fixed!

dotcom-rendering/src/model/window-guardian.ts
oliverlloyd: They are more or less interchangeable so I tend not to think too much about it. The only difference I'm aware of is an interface can be extended so perhaps feels more... *-struggles for word-* ... systemy?

dotcom-rendering/src/web/server/frontToHtml.tsx
```diff
@@ -1,79 +1,142 @@
...
+import { escapeData } from '../../lib/escapeData';
+import { makeFrontWindowGuardian } from '../../model/window-guardian';
+
+interface Props {
+	front: FrontType;
+	NAV: NavType;
+}
+
+const generateScriptTags = (
+	scripts: Array<{ src: string; legacy?: boolean } | false>,
```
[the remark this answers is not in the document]
oliverlloyd: Yeah, we can probably revisit a lot of these bits when we refactor later. I think how we set script tags especially is an interesting area that could perhaps be abstracted in a nice way

PR #4544, opened 2022-04-07
Refactor Main Media

JamieB-gu:
## Why?

A few reasons:

1. The term used widely across the company and in other parts of this codebase - the field in `Item`; the name of the type - is "main media". Having this component named `HeaderMedia` can cause confusion and make it harder to find what you're looking for in the codebase.
2. The component was taking the entirety of `Item` as a prop, which was unnecessary. This narrows the props to include only what the component needs to render. This makes the component easier to reason about, and can mean it's easier to write tests for.
3. There were a few unnecessary complexities which have been simplified:
    1. Use of fragments (`<></>`) to render nothing; React recommends `null`: https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering
    2. Use of `pipe` for conditional rendering; we have the `maybeRender` function for this
    3. Use of `valueOf`; values can be compared directory
    4. Use of `if` statements; `switch` statements make it easier for TypeScript to do exhaustiveness checking

## Changes

- Renamed `HeaderMedia` to `MainMedia`
- Tidied up component file
- Updated imports across codebase
- Narrowed props to only what's needed


MarSavar (approved):
A lot cleaner!

PR #4546, opened 2022-04-07
replace IntersectionObservor in pinned post importable by useInView hook

marjisound:
## What does this change?
This PR refactors the implementation for using the `IntersectionObservor` on `EnhancePinnedPost.importable` with which it would detect whenever the pinned post is in view and is out of view, hence calculates a duration for tracking purposes. This refactoring had been suggested in this [PR](https://github.com/guardian/dotcom-rendering/pull/4242#discussion_r832086192) by @mxdvl & @oliverlloyd 

Thanks to @mxdvl & @arelra who helped with this PR. 

Note: The functionality has not changed but in the attached video `console logs` you can see that the time calculation is happening correctly every time the pinned post is intersected. 
## Why?
The reason for this refactoring was to try out using the existing `useInView` custom react hook instead of creating a new `IntersectionObservor`. The challenge was that we needed to use `useInView` for an `HtmlElement`  that was retrieved by `window.document.querySelector` rather than within the actual react component. 

### After

https://user-images.githubusercontent.com/15894063/162974485-ff2d23cc-1093-40bf-921e-57cc0eea743b.mov


dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
```diff
@@ -91,40 +94,26 @@ export const EnhancePinnedPost = () => {
 	// and emit ophan events when the pinned post goes out of view
 	useEffect(() => {
 		if (!pinnedPost) return;
-
 		const pinnedPostTiming = initPerf('pinned-post-view-duration');
```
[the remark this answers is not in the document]
mxdvl: I think its best to avoid refs

mxdvl (commented):
I think you're missing a `setHasBeenSeen(false)` for multiple timings.

dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
```diff
@@ -43,14 +44,19 @@ function scrollOnCollapse() {
...
+
 	const contentFitsContainer =
 		pinnedPostContent &&
 		pinnedPostContent.scrollHeight <= pinnedPostContent.clientHeight;
 
 	if (contentFitsContainer) hideShowMore();
 
-	const hasBeenSeen = useRef(false);
-
 	const handleClickTracking = () => {
```
arelra, line 60: This function will get recreated on every render. That's usually okay if it is not too expensive to create, the re-renders are limited and the function is not required to be stable.

iiuc this function doesn't use any dependencies so it could it be moved inside the mount useEffect where it is used?

mxdvl: I actually think this function should be moved outside of the component entirely, next to `scrollOnCollapse` and the like.

arelra: That would also work 👍

marjisound:
> I think you're missing a `setHasBeenSeen(false)` for multiple timings.

@mxdvl the `hasBeenSeen` variable is used to make sure at the beginning of page load we don't try measuring the time (`const timeTaken = pinnedPostTiming.end();`) while the pinned post has not yet being viewed.

If we do `setHasBeenSeen(false);` every time the timing is measured, after when the pinned post goes in view again, there will be 2 re-renders. one for `isInView` state and one for `hasBeenSeen` state as you can see here: 

Although I don't think this would cause any issues, I was still wondering if it'd be better to avoid this extra re-rendering?? What do you think?

https://user-images.githubusercontent.com/15894063/163961462-943c26c0-27fd-4562-87b5-3ff852a2b831.mov

mxdvl (approved):

mxdvl:
> > I think you're missing a `setHasBeenSeen(false)` for multiple timings.
> 
> @mxdvl the `hasBeenSeen` variable is used to make sure at the beginning of page load we don't try measuring the time (`const timeTaken = pinnedPostTiming.end();`) while the pinned post has not yet being viewed.

If you’ve tested it and it works, you’re good to go! No need to set it back to false!

dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
marjisound: Thanks both for pointing this out. I moved the click handler out of the component :)

dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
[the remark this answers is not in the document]
arelra: > [as per the React docs](https://reactjs.org/docs/refs-and-the-dom.html).

Oh those docs are unfortunately out of date and mainly refer to pre-hooks refs and their use as handles to dom elements. The React docs are in dire need of an update!

A more up to date reference is here:
https://reactjs.org/docs/hooks-reference.html#useref

Also the (vastly improved) beta of the new React docs on refs are here:
https://beta.reactjs.org/apis/useref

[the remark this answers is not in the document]
marjisound: Thank you both for your time and reviews. 
We had a session with @arelra about `useRef` usage in react which I briefly explain here:

In react docs [this article](https://reactjs.org/docs/refs-and-the-dom.html) is discouraging the usage of `useRef` but the article seems to be an explanation of useRef only for accessing DOM elements within the class based components. 

In the other 2 docs https://reactjs.org/docs/hooks-reference.html#useref & https://beta.reactjs.org/apis/useref it's explaining the `useRef` as a hook which is how we're doing in our functional components. 

Some interesting points from these docs are: 
- useRef is a React Hook that lets you reference a value that’s not needed for rendering
- By using a ref, you ensure that:
    1. You can store information between re-renders
    2. Changing it does not trigger a re-render

In our case, pinnedPostTiming is not needed for rendering. And we just want to store the object between re-renders. 
If we do it this way: `const pinnedPostTiming = initPerf('pinned-post-view-duration');` , initPerf is called on every re-render. I added a console log within `initPerf` and captured the screen video that shows every re-render is also logging **initPerf is instansiated**:
```
	if (name === 'pinned-post-view-duration') {
		console.log('initPerf is instansiated');
	}
``` 


https://user-images.githubusercontent.com/15894063/164046076-6294b9fc-a1ca-4181-9516-603094bf3dd3.mov

I then used it within a `useRef` as you see below:
```
	const pinnedPostTiming =
		useRef<{ start: () => void; end: () => number; clear: () => void }>();

	useEffect(() => {
		pinnedPostTiming.current = initPerf('pinned-post-view-duration');
	}, []);
``` 

Again I captured the screen video, as you can see below, there's only one call to initPerf this time: 

https://user-images.githubusercontent.com/15894063/164046848-76afd754-1ad4-4c3c-a3bf-590873524e22.mov

So based on this exploration I think it'd be probably better to use `useRef` for `pinnedPostTiming` to avoid redundant calls to `initPerf` but  I'm more than happy to hear any views about this :)

dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
```diff
@@ -42,40 +43,48 @@ function scrollOnCollapse() {
...
 export const EnhancePinnedPost = () => {
+	const [hasBeenSeen, setHasBeenSeen] = useState(false);
+	const [isInView] = useIsInView({
+		threshold: 0.1,
+		repeat: true,
+		node: pinnedPost ?? undefined,
+	});
+
+	const pinnedPostTiming =
+		useRef<{ start: () => void; end: () => number; clear: () => void }>();
```
mxdvl, line 80:
```suggestion
	const pinnedPostTiming =
		useRef<typeof initPerf>();
```

dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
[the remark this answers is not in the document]
mxdvl: This is an excellent explanation. Keep using `useRef`!

mxdvl (approved):
Thorough work here. Well done!

dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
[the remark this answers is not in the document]
arelra: Perfectly explained @marjisound . Great work!

dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
marjisound: It didn't let me use the `useRef<typeof initPerf>();` I'm afraid :( . I think it's because the `initPerf` is a function and `typeof initPerf` is giving the type of the function rather than the return type of the function

mxdvl: Sorry–It’s probably [`ReturnType<typeof initPerf>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#returntypetype)

marjisound: Fantastic, it works now. I learnt so much stuff in this PR 🤩

PR #4549, opened 2022-04-07
Fix series title colours (match & sport non-match)

OllysCoding:
## What does this change?

Differentiate between match & non-match colours for series title

Respects the fix in https://github.com/guardian/dotcom-rendering/pull/4511

## Why?

Didn't look quite right

### Before
image
image

### After
image
image

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -81,6 +81,7 @@ type Palette = {
 	text: {
 		headline: Colour;
 		seriesTitle: Colour;
+		matchSeriesTitle: Colour;
 		sectionTitle: Colour;
 		byline: Colour;
 		twitterHandle: Colour;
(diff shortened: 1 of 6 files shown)


dotcom-rendering/index.d.ts
```diff
@@ -82,6 +82,7 @@ type Palette = {
 		headline: Colour;
 		seriesTitle: Colour;
 		sectionTitle: Colour;
+		matchTitle: Colour;
```
oliverlloyd, line 85: Great name!

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -408,6 +408,7 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 										CAPIArticle.guardianBaseURL
 									}
 									badge={CAPIArticle.badge}
+									isMatch={true}
```
oliverlloyd, line 411: Should this always be true?

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -425,6 +426,7 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
...
```
oliverlloyd, line 429: Ditto

oliverlloyd (approved):
Just one question on a hardcoded `true` but otherwise spot on 👍

PR #4554, opened 2022-04-07
Add issue template for bugs

JamieB-gu:
## Why?

It might be handy to prompt for some useful information when recording a bug, like we do for pull requests. This adds a short issue template for bug reports, and allows the "bug" label to be auto-assigned when the issue is created.

## Changes

- Added an issue template for bugs

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -11,7 +11,8 @@ assignees: ''
 
 A clear and concise description of what the bug is.
 
-Does it occur on any specific browsers or breakpoints?
+- Does it occur on any specific browsers or breakpoints?
+- Should this issue be assigned to one of our GitHub project boards, or to a milestone?
 
 ## Steps To Reproduce
 
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,33 @@
+---
+name: Bug report
+about: Create an issue to track a bug
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+## Description
+
+A clear and concise description of what the bug is.
+
+Does it occur on any specific browsers or breakpoints?
+
+## Steps To Reproduce
+
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+## Screenshots
+
+*Delete if unnecessary.*
+
+| Screenshot One | Screenshot Two |
+| - | - |
+| ![screenshot-one][] | ![screenshot-two][] |
+
+[screenshot-one]: *url here*
+[screenshot-two]: *url here*
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -15,7 +15,7 @@ Does it occur on any specific browsers or breakpoints?
 
 ## Steps To Reproduce
 
-Steps to reproduce the behavior:
+Steps to reproduce the behaviour:
 1. Go to '...'
 2. Click on '....'
 3. Scroll down to '....'
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -12,14 +12,14 @@ assignees: ''
 A clear and concise description of what the bug is.
 
 - Does it occur on any specific browsers or breakpoints?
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+assignees: ''
+
+---
+
+## Description
+
+A clear and concise description of what the bug is.
+
+- Does it occur on any specific browsers or breakpoints?
+- Should this issue be assigned to one of our GitHub project boards, or to a milestone?
```
JamieB-gu, line 15: Should this be:

```suggestion
- Remember to assign this to one of our GitHub project boards and to a milestone if applicable
```

What do you think @VDuczekW?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Steps To Reproduce
+
+Steps to reproduce the behaviour:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
```
mxdvl, line 22: 🙈 

```suggestion
1. Go to '…'
2. Click on '…'
3. Scroll down to '…'
```

mxdvl (approved):
This is a great idea.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+## Description
+
+A clear and concise description of what the bug is.
+
+- Does it occur on any specific browsers or breakpoints?
+- Remember to assign this to one of our GitHub project boards and to a milestone if applicable
+
+## Steps To Reproduce
+
+Steps to reproduce the behaviour:
```
oliverlloyd, line 19: Would it make more sense here to ask for an example article? It's rarely the case that there are any 'steps' to reproduce bugs on Dotcom

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+1. Go to '…'
+2. Click on '…'
+3. Scroll down to '…'
+4. See error
+
+## Screenshots
+
+*Delete if unnecessary.*
+
+| Screenshot One | Screenshot Two |
```
oliverlloyd, line 29: This follow the before and after pattern we use for PRs but I think for bugs there's perhaps less need for this?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+assignees: ''
+
+---
+
+## Description
+
+A clear and concise description of what the bug is.
+
+- Does it occur on any specific browsers or breakpoints?
+- Remember to assign this to one of our GitHub project boards and to a milestone if applicable
```
oliverlloyd, line 15: Is it perhaps better for us to do the triage rather than expect others to know about our projects or milestones?

JamieB-gu: I was actually expecting this template mostly to be used by us as a way of automating our own issue creation. Do you think there's a possibility that others will make regular use of it too?

.github/ISSUE_TEMPLATE/bug_report.md
JamieB-gu: I think asking for an example page is very sensible, yes. Perhaps something like "Please include links to one or more pages where the bug can be found"? I'm saying "page" rather than "article" because it could be e.g. a front.

Regarding steps, maybe there aren't always steps beyond "go to this page and look". But sometimes maybe there's "go here, sign in, click this button", or "go here, in this browser, at this breakpoint, and scroll here". I'm thinking if there aren't specific steps then it's easy to delete the placeholder, but sometimes it's useful as a prompt?

Alternatively we could provide prompts for the environment:

- Page URL
- Browser
- Breakpoint

Or we could, as you say, just ask for an example URL and leave the rest freeform. What do you think?

.github/ISSUE_TEMPLATE/bug_report.md
JamieB-gu: Yeah it's a good point, I wasn't sure on this. I wanted make it easy/prompt people to include screenshots.

I think having a table for screenshots is useful because a) it's nice to have that already there for you rather than typing out all the `| - |` yourself, and b) it constrains the images to prevent them going full width and requiring a lot of scrolling.

That said, the structure of the table is perhaps less consistent from issue to issue as "Before"/"After" for PRs. My thought was I'll just include a very basic structure and people can adapt it to their needs.

What do you think?

PR #4557, opened 2022-04-08
More front types

oliverlloyd:
## What does this change?
This PR builds out `FrontType` to include more typings

## Why?
So we can start to use this data with more confidence


mxdvl (approved):
Seems helpful!

PR #4558, opened 2022-04-08
Refactor constructQuery to use URLSearchParams

georgeblahblah:
## What does this change?

* Refactors `constructQuery` to use `URLSearchParams` instead of the custom implementation

Although the companion test passes, I'm not 100% certain that this is backwards compatible as the `query` parameter has quite a broad type.

## Why?

Less code :)

### DIFF
--- a/dotcom-rendering/src/lib/querystring.ts
+++ b/dotcom-rendering/src/lib/querystring.ts
@@ -1,12 +1,4 @@
 const constructQuery = (query: { [key: string]: any }): string =>
-	Object.keys(query)
-		.map((param: string) => {
-			const value = query[param];
-			const queryValue = Array.isArray(value)
-				? value.map((v) => encodeURIComponent(v)).join(',')
-				: encodeURIComponent(value);
-			return `${param}=${queryValue}`;
-		})
-		.join('&');
+	new URLSearchParams(query).toString();
 
 export { constructQuery };


dotcom-rendering/src/lib/querystring.ts
```diff
@@ -1,12 +1,4 @@
...
-	Object.keys(query)
-		.map((param: string) => {
-			const value = query[param];
-			const queryValue = Array.isArray(value)
-				? value.map((v) => encodeURIComponent(v)).join(',')
-				: encodeURIComponent(value);
-			return `${param}=${queryValue}`;
-		})
-		.join('&');
+	new URLSearchParams(query).toString();
```
georgeblahblah, line 2: on a closer look, I'm not sure why this isn't flagged by typescript. I think the type of `query` and the constructor argument don't match

mxdvl: Don’t you want to mark `constructQuery` as deprecated, if all it does is returning an `URLSearchParams`?

jamesgorrie: Yea, I'd be tempted to remove this abstraction completely in favour of the standard API. A little more work, but fewer abstractions.

georgeblahblah: I agree, deprecating `constructQuery` and possibly refactoring existing usages to `URLSearchParams` is the right thing to do. This turned into something a bit bigger than I'd anticipated, but I'll see if it's possible to refactor.

mxdvl (approved):
Happy for this to go in before a full refactor, by the way.

Issue #4226, opened 2022-03-10
Filter Out Invalid LiveBlocks


JamieB-gu:
Rework parsing of `LiveBlock[]` to only include blocks that are able to be rendered. This includes failing blocks if they do not include firstPublished or lastModified dates. Extracting this logic (into possibly `liveBlock.ts`) would mean we can simplify the layout file because we'd only be rendering valid blocks (e.g. we should be able to remove some of the `maybeRender` functions).

https://github.com/guardian/dotcom-rendering/blob/35d5d5b6de53a9a7502c92998e95c3c27cd7302a/apps-rendering/src/liveBlock.ts#L29-L30

PR #4560, opened 2022-04-08
Filter LiveBlocks by dates

iainjchambers-guardian:
## What does this change?
Filters the liveblocks before they're passed to the container by the dates

## Why?
Expecting this to raise further optimisation discussions about how we can refactor out a lot of the conditional rendering further down the tree

### DIFF
--- a/apps-rendering/src/components/liveBlocks.tsx
+++ b/apps-rendering/src/components/liveBlocks.tsx
@@ -4,11 +4,11 @@ import { css } from '@emotion/react';
 import type { BlockContributor } from '@guardian/common-rendering/src/components/liveBlockContainer';
 import LiveBlockContainer from '@guardian/common-rendering/src/components/liveBlockContainer';
 import type { ArticleFormat } from '@guardian/libs';
-import { map, OptionKind, partition, withDefault } from '@guardian/types';
+import { map, partition, withDefault } from '@guardian/types';
 import { LastUpdated } from 'components/lastUpdated';
 import type { Contributor } from 'contributor';
 import { formatUTCTimeDateTz } from 'date';
-import { pipe, toNullable } from 'lib';
+import { pipe } from 'lib';
 import type { LiveBlock } from 'liveBlock';
 import type { FC } from 'react';
 import { renderAll } from 'renderer';
@@ -42,6 +42,7 @@ const LiveBlocks: FC<LiveBlocksProps> = ({ blocks, format }) => {
 					map(Number),
 					toNullable,
 				);
+				const blockLink = `?page=with:block-${block.id}#block-${block.id}`;
 
 				return (
 					<LiveBlockContainer
@@ -50,7 +51,7 @@ const LiveBlocks: FC<LiveBlocksProps> = ({ blocks, format }) => {
 						format={format}
 						blockTitle={block.title}
 						blockFirstPublished={blockFirstPublished}
-						blockId={block.id}
+						blockLink={blockLink}
 						// TODO pass this value in when available
 						isPinnedPost={false}
 						supportsDarkMode={true}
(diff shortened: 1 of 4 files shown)


mxdvl (dismissed):

mxdvl:
How often do we have blocks without a `firstPublished` or `lastModified` field set?

JamieB-gu:
> How often do we have blocks without neither a `firstPublished` nor a `lastModified` field set?

Maybe never, maybe sometimes? The problem is that those fields are optional in the CAPI model so it's possible for them to be missing... which means we have to handle that possibility.

> Expecting this to raise further optimisation discussions about how we can refactor out a lot of the conditional rendering further down the tree

Since you mention it 🙂. Is this designed to handle #4226? Do you think it would be possible to change the type of `firstPublished` and `lastModified` in the `LiveBlock` type to be non-optional? That way we're only allowed to create `LiveBlock`s if those fields can be populated correctly (the "making impossible states impossible" idea), and the rendering code no longer has to deal with the possibility of them being missing.

Let me know if what I've said makes sense/if I've misunderstood something.

iainjchambers-guardian:
> Since you mention it 🙂. Is this designed to handle #4226? Do you think it would be possible to change the type of `firstPublished` and `lastModified` in the `LiveBlock` type to be non-optional? That way we're only allowed to create `LiveBlock`s if those fields can be populated correctly (the "making impossible states impossible" idea), and the rendering code no longer has to deal with the possibility of them being missing.

Yes indeed, this is exactly the sort of question I thought this would raise. However, the two dates are optional on the `LiveBlock` type which is the type on the prop that is passed down the tree so even when we get to a point that they're guaranteed to be filled, the component that they're passed down to treats them as optional which somewhat negates the benefit of this change. There's a few ways to tackle that so I think it's worth discussing further

JamieB-gu:
> However, the two dates are optional on the `LiveBlock` type which is the type on the prop that is passed down the tree

Yup, so this is the type we're talking about I believe?

https://github.com/guardian/dotcom-rendering/blob/ae7af9d5c00289ed46ccc795cea38f262e9c06db/apps-rendering/src/liveBlock.ts#L15-L23

What if we changed that type to look like this?

```ts
type LiveBlock = { 
 	id: string; 
 	isKeyEvent: boolean; 
 	title: string; 
 	firstPublished: Date; 
 	lastModified: Date; 
 	body: Body; 
 	contributors: Contributor[]; 
 }; 
```

We define and populate the blocks array (`LiveBlock[]`), so we can change it if we like 🙂. When we're building that array, if we're unable to construct certain `LiveBlock` instances because the dates are missing/invalid then we just don't include those blocks. Therefore we make it impossible further down the tree for `LiveBlock`s to have optional dates, so the logic to handle that can be stripped from all our rendering code (components).

apps-rendering/src/liveBlock.ts
```diff
@@ -2,7 +2,7 @@
 
 import type { Block } from '@guardian/content-api-models/v1/block';
 import type { Tag } from '@guardian/content-api-models/v1/tag';
-import type { Option } from '@guardian/types';
+import { err, ok, OptionKind, partition, Result } from '@guardian/types';
```
JamieB-gu, line 5: I think the linter wants you to separate out the import of `Result` into an `import type`.

JamieB-gu (approved):
Nice! Couple of small comments.

apps-rendering/src/components/liveBlocks.tsx
```diff
@@ -38,19 +38,14 @@ const LiveBlocks: FC<LiveBlocksProps> = ({ blocks, format }) => {
...
-				);
 
 				return (
 					<LiveBlockContainer
 						key={block.id}
 						id={block.id}
 						format={format}
 						blockTitle={block.title}
-						blockFirstPublished={blockFirstPublished}
+						blockFirstPublished={Number(block.firstPublished)}
```
mxdvl, line 48: Why is this turned into a number?

mxdvl (approved):
This seems like it provides much better handling of missing data!

apps-rendering/src/components/liveBlocks.tsx
JamieB-gu: Good question. There's some previous discussion on this: https://github.com/guardian/dotcom-rendering/pull/3751#discussion_r767876258

Within `LiveBlockContainer`, this field gets converted to a `Date` here:

https://github.com/guardian/dotcom-rendering/blob/4c1ad4105278ba6c92e5cca5aae98c79b788c103/common-rendering/src/components/FirstPublished.tsx#L31

and used as a plain number here:

https://github.com/guardian/dotcom-rendering/blob/4c1ad4105278ba6c92e5cca5aae98c79b788c103/common-rendering/src/components/FirstPublished.tsx#L63

I think it probably is more accurate to have this field be a `Date`.

Just to be clear though @iainjchambers-guardian, I don't think this is a blocker for this PR, don't worry 🙂

mxdvl: Ok, might be better to use the `getTime()` method so it's clearer what's happening at a glace?

PR #4561, opened 2022-04-08
Hide key events heading on mobile when empty

iainjchambers-guardian:
## What does this change?
Hides the key events header on mobile when empty 

## Why?
When there were no key events the heading would still appear on mobile

### DIFF
--- a/apps-rendering/src/components/layout/live.tsx
+++ b/apps-rendering/src/components/layout/live.tsx
@@ -128,6 +128,7 @@ const Live: FC<Props> = ({ item }) => {
 			older={toNullable(item.pagedBlocks.pagination.older)}
 		/>
 	);
+
 	return (
 		<article
 			className="js-article"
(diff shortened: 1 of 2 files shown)


apps-rendering/src/components/layout/live.tsx
```diff
@@ -141,10 +150,14 @@ const Live: FC<Props> = ({ item }) => {
 						<Metadata item={item} />
 					</div>
 				</GridItem>
+
 				<GridItem area="key-events">
-					<div css={keyEventsWrapperStyles} data-chromatic="ignore">
+					<div
+						css={keyEventsWrapperStyles(keyEventBlocks.length < 1)}
```
mxdvl, line 156:
```suggestion
						css={keyEventsWrapperStyles(keyEventBlocks.length === 0)}
```

mxdvl (approved):
Is this captured in a story?

JamieB-gu:
Is there a way we can do this in the `KeyEvents` component itself? That way we fix this issue for AR and DCR simultaneously I think?

iainjchambers-guardian:
> Is there a way we can do this in the `KeyEvents` component itself? That way we fix this issue for AR and DCR simultaneously I think?

Yeah, absolutely, I was trying to keep the change contained to AR but can make it universal

PR #4565, opened 2022-04-11
Added `enhanceCollections` pattern for Fronts

oliverlloyd:
## What does this change?
This PR adds the enhancer pattern to Fronts. We now have a type for the data before it is enhanced - what gets sent to us from Frontend - and then a type for the version that DCR uses afterwards - the actual data we use for rendering

## Why?
The enhancer pattern allows us to put mutations in one place. This is better for comprehension and also serves to make moving these mutations at a later stage easier if we choose to do so.

In addition, having tighter types for the data that is actually used makes reading and learning the code a lot simpler.

### `DCR...` prefix
Previously, we used the `CAPI...` prefix the denote data that had been sent to use from Frontend and then subsequent types were named normally. Here, I'm adding a prefix to denote that a type is DCR specific. I don't think this needs to be slavishly followed (we don't need to prefix _all_ types with 'DCR') but it is useful to have some form of naming convention that helps explain at which end of the enhancer pattern a type is.

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -528,11 +528,26 @@ interface FrontType {
 	commercialProperties: Record<string, unknown>;
 }
 
+type DCRFrontType = {
+	pressedPage: DCRPressedPageType;
+	nav: CAPINavType;
+	editionId: Edition;
+	webTitle: string;
+	config: FrontConfigType;
+};
+
 type PressedPageType = {
 	id: string;
 	seoData: SeoDataType;
 	frontProperties: FrontPropertiesType;
-	collections: PressedCollectionType[];
+	collections: CollectionType[];
+};
+
+type DCRPressedPageType = {
+	id: string;
+	seoData: SeoDataType;
+	frontProperties: FrontPropertiesType;
+	collections: DCRCollectionType[];
 };
 
 type SeoDataType = {
@@ -547,7 +562,7 @@ type FrontPropertiesType = {
 	commercial: Record<string, unknown>;
 };
 
-type CollectionType =
+type ContainerType =
 	| 'dynamic/fast'
 	| 'dynamic/package'
 	| 'dynamic/slow'
@@ -571,7 +586,7 @@ type CollectionType =
 	| 'nav/media-list'
 	| 'news/most-popular';
 
-type CollectionCard = {
+type FrontCard = {
 	properties: {
 		isBreaking: boolean;
 		showMainVideo: boolean;
@@ -588,7 +603,7 @@ type CollectionCard = {
 							source: string;
 							photographer?: string;
 							isMaster?: string;
-							altText: string;
+							altText?: string;
 							height: string;
 							credit: string;
 							mediaId: string;
@@ -632,6 +647,22 @@ type CollectionCard = {
 		isComment: boolean;
 		isGallery: boolean;
 		isAudio: boolean;
+		kicker?: {
+			type: string;
+			item?: {
+				properties: {
+					kickerText: string;
+				};
+			};
+		};
+		seriesOrBlogKicker?: {
+			properties: {
+				kickerText: string;
+			};
+			name: string;
+			url: string;
+			id: string;
+		};
 		headline: string;
 		url: string;
 		hasMainVideoElement: boolean;
@@ -671,16 +702,26 @@ type CollectionCard = {
 	type: string;
 };
 
-type PressedCollectionType = {
+type DCRFrontCard = {
+	format: ArticleFormat;
+	isLiveBlog: boolean;
+	url: string;
+	headline: string;
+	webPublicationDate?: string;
+	image?: string;
+	kickerText?: string;
+};
+
+type CollectionType = {
 	id: string;
 	displayName: string;
-	curated: CollectionCard[];
-	backfill: CollectionCard[];
-	treats: CollectionCard[];
+	curated: FrontCard[];
+	backfill: FrontCard[];
+	treats: FrontCard[];
 	lastUpdate?: number;
 	href?: string;
 	groups?: string[];
-	collectionType: CollectionType;
+	collectionType: ContainerType;
 	uneditable: boolean;
 	showTags: boolean;
 	showSections: boolean;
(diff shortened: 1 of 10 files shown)


OllysCoding:
I think I would prefer prefixing the pre-enhanced types with something like `FrontendXXX` or `PressedXXX`, then prefixing the now converted DCR types into `DCRXXX`.

My reasoning being that we're now in the DCR codebase, where we have control over our own types, it makes more sense that we acknowledge the difference in types from our data source (frontend/facia-press), rather than give them the more generic type names, and leave us with the un-prefixed ones for use around DCR!

I think this would also be less confusing for developers, especially if the `DCRXXX` prefix isn't used consistently!

mxdvl:
> I think I would prefer prefixing the pre-enhanced types with something like `FrontendXXX` or `PressedXXX`, then prefixing the now converted DCR types into `DCRXXX`.

I quite like that approach. I would opt for `FrontendXXX` as that’s where the types come from.

Also, do people have views on the following:
- organising types into different files, so we get more organisation
- importing types explicitly, so we can know where types are used

oliverlloyd:
> Also, do people have views on the following:
> 
> * organising types into different files, so we get more organisation

I think for location in general it's good to move things around until it feel right.

> * importing types explicitly, so we can know where types are used

For importing vs. globals, I'm happy with either and don't have strong opinions. I think if you forced me to choose I would go explicit but I'm very will to defer to team consensus. There's also the fact that this is where we are now and it's generally easier to follow existing conventions.

> * making sure we format & lint these files, as currently there are `any` that need an `// eslint-ignore`, and some that don’t

Consistent linting rules sound like a great idea 👍

oliverlloyd:
> I think I would prefer prefixing the pre-enhanced types with something like `FrontendXXX` or `PressedXXX`, then prefixing the now converted DCR types into `DCRXXX`.

Sounds good - I'll do that now

dotcom-rendering/index.d.ts
```diff
@@ -515,16 +515,16 @@ interface CAPIArticleType {
 
 type StageType = 'DEV' | 'CODE' | 'PROD';
 
-interface FrontType {
-	pressedPage: PressedPageType;
+interface FEFrontType {
```
mxdvl, line 518: Nice!

Maybe add `/* FE* types are coming from Frontend */`

mxdvl (approved):

dotcom-rendering/index.d.ts
jamesgorrie: Is there a reason not to call it Frontend? Just because it's easier for someone to fathom when approaching the codease. Or maybe a comment pointing to the type in that repo?

jamesgorrie: I know it might be explanatory, but maybe worth describing DCR as well.

dotcom-rendering/src/model/enhanceCollections.ts
```diff
@@ -0,0 +1,40 @@
+import { ArticleDesign } from '@guardian/libs';
+import { decideFormat } from '../web/lib/decideFormat';
+
+const enhanceCards = (collections: FEFrontCard[]): DCRFrontCard[] => {
+	const enhanced: DCRFrontCard[] = [];
+	collections.forEach((faciaCard) => {
```
jamesgorrie, line 6: Is there a reason we don't use map as we do below?

jamesgorrie (approved):
This makes it really clear as to what things are!

dotcom-rendering/src/model/enhanceCollections.ts
oliverlloyd: Because map will return something for _all_ items but in this case we might want to filter things if the data isn't available.

mxdvl: I think it’s best to keep to one style, either mutation with a `for … of` loop, or functional with a `.filter().map()`. 
Sadly, `forEach` is this weird in-between…

oliverlloyd: I did try `filter` but I was still getting typescript errors 😢 

In my mind (with my personal mental models) I would say both `map` and `forEach` are valid options with use cases where both are valid but also use cases where one or the other might be better suited.

oliverlloyd: That said, I'm totally happy to forgo `forEach` in favour of `for ... of` - sorry, I didn't speak to that in my last comment

jamesgorrie: subjectively `for ... of` is really readable.

jamesgorrie: What were the typescript errors?

JamieB-gu: > Because map will return something for _all_ items but in this case we might want to filter things if the data isn't available.

As you say, I think `map` is for situations where you a new array of the same length. If you want to create a new object (which may just mean an array of a different length because you've filtered some stuff, or may be a new data structure entirely) you can go up a level to `reduce`. In this case `reduce` is able to combine `map` and `filter` into one operation:

```ts
const enhanceCards = (collections: FEFrontCard[]): DCRFrontCard[] =>
	collections.reduce((enhanced: DCRFrontCard[], faciaCard: FEFrontCard) => {
		if (faciaCard.format) {
			const format = decideFormat(faciaCard.format);
			return [...enhanced, {
				format,
				isLiveBlog: format.design === ArticleDesign.LiveBlog,
				url: faciaCard.header.url,
				headline: faciaCard.header.headline,
				webPublicationDate: faciaCard.card.webPublicationDateOption
					? new Date(
							faciaCard.card.webPublicationDateOption,
					  ).toISOString()
					: undefined,
				image: faciaCard.properties.maybeContent?.trail.trailPicture
					?.allImages[0].url,
				kickerText:
					faciaCard.header.kicker?.item?.properties.kickerText,
			}];
		}

		return enhanced;
	}, []);
```

I think the code is similar to `forEach` without, as Max mentions, switching to mutation.

mxdvl: Proposed as a simplification in #4582

PR #4569, opened 2022-04-11
`DynamicFast`

oliverlloyd:
## What does this change?
This adds the `DynamicFast` container layout

Screenshot 2022-04-11 at 11 31 02

## Does the encompass all possible variations that might happen for this container?
No. There are still some edge cases which to be supported. Having two cards in the bottom left (third position) for example.

But this PR does support the main use cases and offers a useful introduction to how existing abstractions and tools already built into DCR can be used to create layouts like this.

### DIFF
--- a/dotcom-rendering/src/web/components/DynamicFast.tsx
+++ b/dotcom-rendering/src/web/components/DynamicFast.tsx
@@ -1,0 +1,202 @@
+import { ArticleDesign } from '@guardian/libs';
+
+import { Card } from './Card/Card';
+import { UL } from './Card/components/UL';
+import { LI } from './Card/components/LI';
+
+type Props = {
+	trails: TrailType[];
+};
+
+export const DynamicFast = ({ trails }: Props) => {
+	const primary = trails[0];
+	const secondary = trails[1];
+	const third = trails[2];
+	const groupOne = trails.slice(3, 6);
+	const groupTwo = trails.slice(7, 12);
+
+	return (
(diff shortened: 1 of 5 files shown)


OllysCoding:
My understanding of the `dynamic/*` layouts is that they should adapt to any number of cards in the layout, though I'm not sure this initial incarnation of the component would work that way? 

Would it be better to test out a fixed layout as a start?

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -0,0 +1,202 @@
...
+type Props = {
+	trails: TrailType[];
+};
+
+export const DynamicFast = ({ trails }: Props) => {
+	const primary = trails[0];
+	const secondary = trails[1];
+	const third = trails[2];
+	const groupOne = trails.slice(3, 6);
+	const groupTwo = trails.slice(7, 12);
```
OllysCoding, line 16: Is it worth adding some comments here about what assumptions we make in the current incarnation of this component? That way it's quicker to understand what changes we think need made for the future!

OllysCoding (approved):
+1 This is an awesome starting place for working on layouts :)

dotcom-rendering/src/web/components/DynamicFast.tsx
oliverlloyd: Done 👍

PR #4572, opened 2022-04-11
`FixedLargeSlowXIV`

oliverlloyd:
## What does this change?
This PR adds the rendering component for the `FixedLargeSlowXIV` container type

Screenshot 2022-04-11 at 12 40 21

PR #4581, opened 2022-04-11
Fix margins on `DynamicFast`

oliverlloyd:
## What does this change?
Set the `bottomMargin` prop consistently

## Why?
@ajbreuer [spotted this](https://github.com/guardian/dotcom-rendering/pull/4572#issuecomment-1095070490) when it was implemented 

### Before
Screenshot 2022-04-11 at 15 03 29

### After
Screenshot 2022-04-11 at 15 02 11

### DIFF
--- a/dotcom-rendering/src/web/components/DynamicFast.tsx
+++ b/dotcom-rendering/src/web/components/DynamicFast.tsx
@@ -133,7 +133,7 @@ export const DynamicFast = ({ trails }: Props) => {
 												linkTo={card.url}
 												format={card.format}
 												headlineText={card.headline}
-												headlineSize="medium"
+												headlineSize="small"
 												byline={card.byline}
 												showByline={card.showByline}
 												showQuotes={
--- a/dotcom-rendering/src/web/components/DynamicFast.tsx
+++ b/dotcom-rendering/src/web/components/DynamicFast.tsx
@@ -121,9 +121,9 @@ export const DynamicFast = ({ trails }: Props) => {
 							padSides={true}
 						>
 							<UL direction="column">
-								{group.map((card, groupIndex) => {
+								{group.map((card, cardIndex) => {
 									const isLastCard =
-										groupIndex === group.length - 1;
+										cardIndex === group.length - 1;
 									return (
 										<LI
 											stretch={true}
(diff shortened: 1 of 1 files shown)


ajbreuer:
Awesome! Thanks @oliverlloyd . And apologies for originally posting this in the wrong PR.

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -113,94 +112,61 @@ export const DynamicFast = ({ trails }: Props) => {
...
+						<LI
+							percentage="25%"
+							showDivider={true}
+							showTopMarginWhenStacked={true}
+							padSides={true}
+						>
+							<UL direction="column">
+								{group.map((card, groupIndex) => {
+									const isLastCard =
+										groupIndex === group.length - 1;
```
mxdvl, line 126:
```suggestion
							<UL direction="column">
								{group.map((card, cardIndex) => {
									const isLastCard =
										cardIndex === group.length - 1;
```

mxdvl (commented):
Do we need to have margins if we’re using flex? I think the `gap` property is easier to reason about.

Is this due to [`gap` not being supporter on IE 11](https://caniuse.com/flexbox-gap)?

dotcom-rendering/src/web/components/DynamicFast.tsx
oliverlloyd: Oops, thanks!

oliverlloyd:
> Do we need to have margins if we’re using flex? I think the `gap` property is easier to reason about.

Oh! `gap` is used for flex as well as grid, I didn't know that!

But from reading the docs it seems to only support space _in-between_

> [The gap property](https://css-tricks.com/almanac/properties/g/gap/) explicitly controls the space between flex items. It applies that spacing only between items not on the outer edges.

Which I think won't cover our use cases here

mxdvl:
> > [The gap property](https://css-tricks.com/almanac/properties/g/gap/) explicitly controls the space between flex items. It applies that spacing only between items not on the outer edges.
> 
> Which I think won't cover our use cases here

Can you explain how it doesn’t? I think this is _exactly_ what we want.

image

oliverlloyd:
I guess the real question is: is using `gap` solving a problem?

I fear that introducing this change would be work for no benefit. You could argue that `gap` vs. `margin` is perhaps _nicer_ but even there this is subjective and intrinsically both approaches are valid and work. There's nothing wrong with `margin`, it's been in use for decades. The difference is we already have one in place and the other one would need time and effort to achieve.

mxdvl (approved):
I thought `UL` was a new component. Let’s keep `bottomMargin` for now.

There are some issues with `margin`, but that’s outside the scope of this PR. 😉

PR #4582, opened 2022-04-11
refactor(enhanceCards): use type predicate

mxdvl:
## What does this change?

Use `filter` and `map` rather than `forEach`. Keep TypeScript happy with the use of a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).

## Why?

Keep the code simpler. The alternatives are:
- `for … of` loop
- using `reduce`

### DIFF
--- a/dotcom-rendering/src/model/enhanceCollections.ts
+++ b/dotcom-rendering/src/model/enhanceCollections.ts
@@ -1,12 +1,15 @@
 import { ArticleDesign } from '@guardian/libs';
 import { decideFormat } from '../web/lib/decideFormat';
 
-const enhanceCards = (collections: FEFrontCard[]): DCRFrontCard[] => {
-	const enhanced: DCRFrontCard[] = [];
-	collections.forEach((faciaCard) => {
-		if (faciaCard.format) {
+const enhanceCards = (collections: FEFrontCard[]): DCRFrontCard[] =>
+	collections
+		.filter(
+			(card: FEFrontCard): card is FEFrontCard & { format: CAPIFormat } =>
+				!!card.format,
+		)
+		.map((faciaCard) => {
 			const format = decideFormat(faciaCard.format);
-			enhanced.push({
+			return {
 				format,
 				isLiveBlog: format.design === ArticleDesign.LiveBlog,
 				url: faciaCard.header.url,
@@ -20,11 +23,8 @@ const enhanceCards = (collections: FEFrontCard[]): DCRFrontCard[] => {
 					?.allImages[0].url,
 				kickerText:
 					faciaCard.header.kicker?.item?.properties.kickerText,
-			});
-		}
-	});
-	return enhanced;
-};
+			};
+		});
 
 export const enhanceCollections = (
 	collections: FECollectionType[],


dotcom-rendering/src/model/enhanceCollections.ts
```diff
@@ -1,12 +1,15 @@
...
-const enhanceCards = (collections: FEFrontCard[]): DCRFrontCard[] => {
-	const enhanced: DCRFrontCard[] = [];
-	collections.forEach((faciaCard) => {
-		if (faciaCard.format) {
+const enhanceCards = (collections: FEFrontCard[]): DCRFrontCard[] =>
+	collections
+		.filter(
+			(card: FEFrontCard): card is FEFrontCard & { format: CAPIFormat } =>
+				!!card.format,
+		)
```
mxdvl, line 9: I’ve inlined the [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates), because it’s not used elsewhere.

dotcom-rendering/src/model/enhanceCollections.ts
```diff
@@ -1,12 +1,15 @@
...
 import { decideFormat } from '../web/lib/decideFormat';
 
-const enhanceCards = (collections: FEFrontCard[]): DCRFrontCard[] => {
-	const enhanced: DCRFrontCard[] = [];
-	collections.forEach((faciaCard) => {
-		if (faciaCard.format) {
+const enhanceCards = (collections: FEFrontCard[]): DCRFrontCard[] =>
+	collections
+		.filter(
+			(card: FEFrontCard): card is FEFrontCard & { format: CAPIFormat } =>
```
jamesgorrie, line 7: Really like this "type with non-optional parameter" syntax!

jamesgorrie (approved):

dotcom-rendering/src/model/enhanceCollections.ts
mxdvl: I thought this stroke the right balance of TypeScript for our codebase.

For some generic types, you might enjoy [`SetRequired`](https://github.com/sindresorhus/type-fest/blob/main/source/set-required.d.ts) from `type-fest`.

oliverlloyd: Lone voice of dissent here. I am actually confused by this syntax - sorry!

Is this casting the type?

mxdvl: Type predicates are a way to tell TypeScript that a type matches. It's not dissimilar to a type assertion, but requires a runtime check. In this case, we're saying that if an element has format, then its format key is not undefined.

JamieB-gu: I think there are two features being used in tandem, which might make this confusing:

**Type Predicate**

This _is_ casting the type. It's saying that if the type predicate function returns `true`, then the thing you passed to it is this type from now on. For example, here in DCR:

https://github.com/guardian/dotcom-rendering/blob/e47b1114c94c565cf6f419ca8f82a45ec67910ea/dotcom-rendering/src/amp/lib/real-time-config.ts#L89

If you were to do this:

```ts
const getValue = (): string | undefined => ...
const myValue = getValue()

myValue.split(','); // This is an error! myValue can be undefined

if (notUndefined(myValue)) {
    myValue.split(','); // This is not an error. myValue has been cast to string
}
```

I think, type predicates can be as dangerous as type casts (`as`), because if you were to make a mistake in your predicate, say:

```ts
const notUndefined = <T>(x: T | undefined): x is T => 5 === 5
```

the above example would still compile, but your code would crash at runtime. That said, if you're careful they can be really useful, as Max has shown 🙂

Docs are here: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

**Intersection Type**

I'm less familiar with the second pattern, so @mxdvl correct me if I'm wrong, but I believe it's saying "take `FEFrontCard` and change the type of field `format` to `CAPIFormat`". You're basically creating a new type:

```ts
type FEFrontCardWithRequiredFormat = FEFrontCard & { format: CAPIFormat }
```

This isn't a type cast itself, it's just the definition of a type. It only becomes a cast when used with the type predicate to change the type of the `FEFrontCard`s in this array to `FEFrontCardWithRequiredFormat`. In this case the only difference between the two is that `format` is now a required field.

I don't know if this is helpful, but I think of intersection types as being a bit like object spreading.

```ts
const oldPerson = {
  name: "CP Scott",
  age: 85,
}

const newPerson = {
  ...oldPerson,
   name: "Charles Prestwich Scott",
}
```

The objects later in the spread overwrite the fields of the objects earlier in the spread. Intersection types do the same thing in the type world:

```ts
type Person = {
  name?: string;
  age: number;
}

type NewPerson = Person & { name: string }

const person: NewPerson = { age: 85 } // Type error: missing field "name"
```

Docs here:

- https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals

Sorry this was a bit long-winded @oliverlloyd, it probably went a bit beyond the scope of your comment 🙈. But perhaps it will be useful for others reading this PR and wondering how this works.

PR #4590, opened 2022-04-11
add blinking dot for mobile and update desktop version

marjisound:
## What does this change?
This PR does the following:
- Add blinking dot to liveblogs article title section on below desktop 
- update the size of the blinking dot and the article last updated text on desktop breakpoint

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleLastUpdated.tsx
+++ b/dotcom-rendering/src/web/components/ArticleLastUpdated.tsx
@@ -6,14 +6,14 @@ import { PulsingDot } from './PulsingDot.importable';
 import { Island } from './Island';
 
 const lastUpdatedStyles = (palette: Palette) => css`
-	${textSans.xxsmall()}
+	${textSans.small()}
 	padding-bottom: 0.125rem;
 	padding-top: 0.125rem;
 	color: ${palette.text.standfirst};
 `;
 
 const livePulseIconStyles = css`
-	${textSans.xxsmall({ fontWeight: 'bold' })}
+	${textSans.small({ fontWeight: 'bold' })}
 `;
 
 type Props = {
--- a/dotcom-rendering/src/web/components/ArticleTitle.tsx
+++ b/dotcom-rendering/src/web/components/ArticleTitle.tsx
@@ -1,9 +1,12 @@
 import { css } from '@emotion/react';
 
-import { from, until } from '@guardian/source-foundations';
+import { from, textSans, until } from '@guardian/source-foundations';
 import { ArticleDisplay, ArticleDesign } from '@guardian/libs';
 import { Badge } from './Badge';
 import { SeriesSectionLink } from './SeriesSectionLink';
+import { Island } from './Island';
+import { PulsingDot } from './PulsingDot.importable';
+import { decidePalette } from '../lib/decidePalette';
 
 type Props = {
 	format: ArticleFormat;
@@ -16,7 +19,7 @@ type Props = {
 
 const sectionStyles = css`
 	padding-top: 8px;
-	display: flex;
+	display: inline-flex;
 	flex-direction: row;
 	${from.leftCol} {
 		flex-direction: column;
@@ -56,6 +59,14 @@ const immersiveMargins = css`
 	}
 `;
 
+const livePulseIconStyles = css`
+	${textSans.xxsmall({ fontWeight: 'bold' })}
+	padding-top: 0.25em;
+	${from.desktop} {
+		display: none;
+	}
+`;
+
 export const ArticleTitle = ({
 	format,
 	tags,
@@ -63,29 +74,49 @@ export const ArticleTitle = ({
 	sectionUrl,
 	guardianBaseURL,
 	badge,
-}: Props) => (
-	<div css={[sectionStyles, badge && badgeContainer]}>
-		{badge && format.display !== ArticleDisplay.Immersive && (
-			<div css={titleBadgeWrapper}>
-				<Badge imageUrl={badge.imageUrl} seriesTag={badge.seriesTag} />
+}: Props) => {
+	const palette = decidePalette(format);
+	return (
+		<div css={[sectionStyles, badge && badgeContainer]}>
+			{format.design === ArticleDesign.LiveBlog && (
+				<span
+					css={[
+						livePulseIconStyles,
+						css`
+							color: ${palette.text.seriesTitle};
+						`,
+					]}
+				>
+					<Island deferUntil="idle">
+						<PulsingDot />
+					</Island>
+				</span>
+			)}
+			{badge && format.display !== ArticleDisplay.Immersive && (
+				<div css={titleBadgeWrapper}>
+					<Badge
+						imageUrl={badge.imageUrl}
+						seriesTag={badge.seriesTag}
+					/>
+				</div>
+			)}
+			<div
+				css={[
+					badge && marginTop,
+					format.display === ArticleDisplay.Immersive &&
+						format.design !== ArticleDesign.PrintShop &&
+						immersiveMargins,
+				]}
+			>
+				<SeriesSectionLink
+					format={format}
+					tags={tags}
+					sectionLabel={sectionLabel}
+					sectionUrl={sectionUrl}
+					guardianBaseURL={guardianBaseURL}
+					badge={badge}
+				/>
 			</div>
-		)}
-		<div
-			css={[
-				badge && marginTop,
-				format.display === ArticleDisplay.Immersive &&
-					format.design !== ArticleDesign.PrintShop &&
-					immersiveMargins,
-			]}
-		>
-			<SeriesSectionLink
-				format={format}
-				tags={tags}
-				sectionLabel={sectionLabel}
-				sectionUrl={sectionUrl}
-				guardianBaseURL={guardianBaseURL}
-				badge={badge}
-			/>
 		</div>
-	</div>
-);
+	);
+};
--- a/dotcom-rendering/src/web/components/PulsingDot.importable.tsx
+++ b/dotcom-rendering/src/web/components/PulsingDot.importable.tsx
@@ -1,17 +1,23 @@
 import { css, keyframes } from '@emotion/react';
 
 import { storage } from '@guardian/libs';
+import { until } from '@guardian/source-foundations';
 import { useEffect, useState } from 'react';
 
 const dotStyles = (colour?: string) => css`
 	color: ${colour && colour};
+	vertical-align: middle;
 	:before {
 		border-radius: 62.5rem;
 		display: inline-block;
 		position: relative;
 		background-color: currentColor;
-		width: 0.75em;
-		height: 0.75em;
+		width: 0.917em;
+		height: 0.917em;
+		${until.desktop} {
+			width: 1.084em;
+			height: 1.084em;
+		}
 		content: '';
 		margin-right: 0.1875rem;
 		vertical-align: initial;


dotcom-rendering/src/web/components/ArticleTitle.tsx
```diff
@@ -65,30 +77,43 @@ export const ArticleTitle = ({
...
-			<div css={titleBadgeWrapper}>
-				<Badge imageUrl={badge.imageUrl} seriesTag={badge.seriesTag} />
+}: Props) => {
+	const palette = decidePalette(format);
+	return (
+		<div css={[sectionStyles, badge && badgeContainer]}>
+			{format.design === ArticleDesign.LiveBlog && (
+				<span css={livePulseIconStyles(palette)}>
+					<Island deferUntil="idle">
+						<PulsingDot diameter={1.084} />
```
joecowton1, line 87: If this diameter is always the same, is there a reason to pass it in as a prop?

dotcom-rendering/src/web/components/PulsingDot.importable.tsx
```diff
@@ -3,15 +3,16 @@ import { css, keyframes } from '@emotion/react';
 import { storage } from '@guardian/libs';
 import { useEffect, useState } from 'react';
 
-const dotStyles = (colour?: string) => css`
+const dotStyles = (colour?: string, diameter?: number) => css`
```
joecowton1, line 6: If the diameter needs to be dynamic  - could this be a flag which drives the size of the diameter, rather than the diameter itself?

```
const dotStyles = (colour?: string, large?: boolean) => css..
```


```
width: ${large ? 1 : 0.75}rem
```

dotcom-rendering/src/web/components/ArticleTitle.tsx
```diff
@@ -57,6 +60,15 @@ const immersiveMargins = css`
 	}
 `;
 
+const livePulseIconStyles = (palette: Palette) => css`
```
joecowton1, line 63: Could these styles live inside the `PulsingDot` component?

dotcom-rendering/src/web/components/ArticleLastUpdated.tsx
```diff
@@ -5,15 +5,15 @@ import { decidePalette } from '../lib/decidePalette';
 import { PulsingDot } from './PulsingDot.importable';
 import { Island } from './Island';
 
-const lastUpdatedStyles = (palette: Palette) => css`
-	${textSans.xxsmall()}
+const lastUpdatedStyles = (palette: Palette, design: ArticleDesign) => css`
+	${design === ArticleDesign.LiveBlog ? textSans.small() : textSans.xxsmall()}
```
joecowton1, line 9: Does this need to cover dead blogs too?

dotcom-rendering/src/web/components/ArticleTitle.tsx
```diff
@@ -65,30 +77,43 @@ export const ArticleTitle = ({
...
-	<div css={[sectionStyles, badge && badgeContainer]}>
-		{badge && format.display !== ArticleDisplay.Immersive && (
-			<div css={titleBadgeWrapper}>
-				<Badge imageUrl={badge.imageUrl} seriesTag={badge.seriesTag} />
+}: Props) => {
+	const palette = decidePalette(format);
+	return (
+		<div css={[sectionStyles, badge && badgeContainer]}>
+			{format.design === ArticleDesign.LiveBlog && (
+				<span css={livePulseIconStyles(palette)}>
```
joecowton1, line 85: See earlier comment about these being inside the `PulsingDot` component. Also does it want to be a span or is it better as a div?

dotcom-rendering/src/web/components/PulsingDot.importable.tsx
```diff
@@ -34,9 +35,10 @@ const animate = css`
 
 interface Props {
 	colour?: string;
+	diameter?: number;
```
oliverlloyd, line 38: It feels like the size of the dot is always going to want to be coupled with the size of the text so maybe it makes more sense to have a `size` token here?

dotcom-rendering/src/web/components/ArticleTitle.tsx
oliverlloyd: Yes, please!

dotcom-rendering/src/web/components/ArticleLastUpdated.tsx
oliverlloyd: I _think_ that deadblogs here are the default so this reads right to me but I could be wrong and will leave it to @marjisound to confirm!

dotcom-rendering/src/web/components/ArticleTitle.tsx
marjisound: The `PulsingDot` is used in 3 different places and it's size is different in every one of them. I thought we can do either of these:
1- add a context prop such as  `kicker | ArticleLastUpdated | ArticleTitle` to `PulsingDot` to decide the size based on the context
2- Pass the size of the dot as a prop

But wasn't sure which option to go for 🤔   @oliverlloyd What's your insight about this?

dotcom-rendering/src/web/components/PulsingDot.importable.tsx
marjisound: Do you think adding a size prop with these values `small | medium | large` would be OK?

dotcom-rendering/src/web/components/ArticleTitle.tsx
oliverlloyd: I think those both sound like valid options.

I think though given the dot is always shown alongside a kicker then if we had a `size` prop we could make sure that each size aligns with it's own text `size` value.

Responding to you other comment below:
The design system uses `xxxsmall`, `xxsmall`, etc. which we _could_ use but, yeah, equally we could also just have `small`|`medium`|`large. I'd perhaps be wary of coupling things _too_ close to the design system as that will change and then things will be confusing so maybe `small`, `medium`, etc. is simplest. We use similar apis in other places

dotcom-rendering/src/web/components/PulsingDot.importable.tsx
oliverlloyd: It's very subjective so 🤷 . But I've listed some things I think play into this decision in my other comment above. Your call though!

dotcom-rendering/src/web/components/ArticleTitle.tsx
mxdvl: I don’t think we need to add a `size` prop at all! The icon is already sized relative to current `font-size`, as it uses `em`–not `rem`.

What happens in you remove the `size`/`diameter` prop altogether?

mxdvl: FYI I’ve got it to work by removing the font-size style entirely and the diameter:

image

image

oliverlloyd: > FYI I’ve got it to work by removing the font-size style entirely and the diameter

Woo hoo 🎉

mxdvl: @marjisound and I discussed it and the reason for introducing this was that the Figma design specified a 11px dot for a font-size of 15px, which is just shy of `0.75em`. This is probably not reason enough to introduce a whole diameter parameter, but we could reduce to `0.74em` to get the right rounding of values for specific font-sizes.

marjisound: Having a chat with Ben he was happy with us using the `0.75em` meaning the dot diameter would be 75% of the size of the text beside it. 

I moved the usage of the `PulsingDot` from the `ArticleTitle` to the `SeriesSectionLink` mostly because in SeriesSectionLink, we have the font styles and the colour which is needed by the pulsing dot

dotcom-rendering/src/web/components/ArticleTitle.tsx
marjisound: Sorry, I made some changes and the `PulsingDot` is now being added from within the `SeriesSectionLink`. I removed the span in there :)

dotcom-rendering/src/web/components/ArticleLastUpdated.tsx
marjisound: Thanks for pointing this out, Ben suggested that we increase the LastUpdated text size for deadblogs too 👍

dotcom-rendering/src/web/components/ArticleTitle.tsx
marjisound: I had to again use a span around `Island` in `SeriesSectionLink`. Do you still think a `div` could be more suitable rather than a `span`?

joecowton1: I think a `span` probably makes sense here. The reason I flagged it was because often spans are used just to wrap text, here we're wrapping an Island and a component, though essentially that component is itself just another `span`.

If you wanted to use a `div` just add a `display: inline` to the div styles, I'm not sure how much it matters though really.

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -1084,6 +1105,7 @@ export const decidePalette = (format: ArticleFormat): Palette => {
 			headlineByline: textHeadlineByline(format),
 			standfirst: textStandfirst(format),
 			standfirstLink: textStandfirstLink(format),
+			lastUpdatedStandfirst: textLastUpdatedStandfirst(format),
```
oliverlloyd, line 1108: Is there a better name for this? Having 'standfirst' here is confusing as I don't _think_ this is related to the standfirst text?

oliverlloyd (approved):
Just two points but I'm fine for you to take these away and merge when you're happy

dotcom-rendering/src/web/lib/decidePalette.ts
marjisound: Do you think just `lastUpdated` would be good? or maybe `articleLastUpdated`?

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -150,6 +150,27 @@ const textStandfirst = (format: ArticleFormat): string => {
...
+				return culture[600];
+			case ArticlePillar.Lifestyle:
+				return lifestyle[600];
+			case ArticlePillar.Sport:
+				return sport[600];
+			case ArticlePillar.Opinion:
+				return opinion[600];
+			case ArticleSpecial.Labs:
+			case ArticleSpecial.SpecialReport:
+				return news[600];
```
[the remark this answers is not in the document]
marjisound: Thanks Oliver. I will chat with Ben about this and update it then 👍

dotcom-rendering/src/web/lib/decidePalette.ts
oliverlloyd: 👍 from me

dotcom-rendering/src/web/lib/decidePalette.ts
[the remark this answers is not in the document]
marjisound: I had a chat with Ben and suggested that because we don't yet have a proper design for special report liveblogs, until we have such a design, the fallback in composer is news palette. And so, news[600] would be the colour to use for the pulsating dot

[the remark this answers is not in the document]
oliverlloyd: The problem with this is someone can, and at some point will, launch a special report liveblog and so either DCR supports this theme or we need to rollback support for it

@jamesgorrie In the absence of designs for this theme, should we revert support for DCR and return to having them rendered by Frontend or do we want to move forward on this?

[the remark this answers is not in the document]
oliverlloyd: Note. I think when we added the default to `news` in other places before that was when DCR was only rendering a small percentage of blogs. But now that we're at 100% I'm not sure this is a good approach.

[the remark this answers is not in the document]
mxdvl: We discussed this internally and currently, these types of live blogs are [**not supported by DCR**](https://github.com/guardian/frontend/blob/120da6027da932b3f7036091fc61b5971af6942e/article/app/controllers/LiveBlogController.scala#L348-L355), so it’s not an issue yet.

PR #4593, opened 2022-04-12
Delete immersive `Card` styles

oliverlloyd:
## What does this change?
This PR removes the `isFullCardImage` prop from the codebase and all associated styling and logic. As part of this, we are also removing `carouselImages`.

In addition, this now allows us to remove [a hack](https://github.com/guardian/dotcom-rendering/pull/2765) that had to be setup as a result of this test code.

## Why?
This code was introduced as part of [an AB test](https://github.com/guardian/dotcom-rendering/pull/2539) conducted a year ago. The test was for immersive cards. A decision was made to not go forward with the immersive design but the code remained in the platform. This PR now removes this code.

### Why remove the image sourcesets (`carouselImages`)? Isn't that useful?
Because 1) this code is not used, we only used an image from this array in the immersive case which only happened during the test and 2) because even when we did pull from this array we were not actually using a `picture` tag and the code always manually selected the same image, so the array was moot.

We _should_ consider upgrading from an `img` tag to a `picture` tag for Cards but I think this should be done as its own piece of work without using this `carouselImages` prop.

## Screenshots

| Before      | After      |
|-------------|------------|
| Screenshot 2022-04-12 at 09 02 18 | Screenshot 2022-04-12 at 09 01 01 |


jamesgorrie (approved):
Really nice catch, and great to spot this is an issue with how we run tests, especially unsuccessful ones.

PR #4598, opened 2022-04-12
Remove `isliveblog`

oliverlloyd:
## What does this change?
This removes the `isLiveBlog` property from `TrailType`

## Why?
We don't need this. We can use `format` to determine this status

### DIFF
--- a/dotcom-rendering/src/web/lib/renderElement.tsx
+++ b/dotcom-rendering/src/web/lib/renderElement.tsx
@@ -85,19 +85,19 @@ type Props = {
 // updateRole modifies the role of an element in a way appropriate for most
 // article types.
 const updateRole = (el: CAPIElement, format: ArticleFormat): CAPIElement => {
-	const isLiveBlog =
+	const isBlog =
 		format.design === ArticleDesign.LiveBlog ||
 		format.design === ArticleDesign.DeadBlog;
 
 	switch (el._type) {
 		case 'model.dotcomrendering.pageElements.ImageBlockElement':
-			if (isLiveBlog && el.role !== 'thumbnail') {
+			if (isBlog && el.role !== 'thumbnail') {
 				el.role = 'inline';
 			}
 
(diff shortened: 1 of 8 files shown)


dotcom-rendering/src/web/lib/renderElement.tsx
```diff
@@ -136,7 +136,7 @@ export const renderElement = ({
 }: Props): [boolean, JSX.Element] => {
 	const palette = decidePalette(format);
 
-	const isLiveBlog =
+	const isBlog =
```
jamesgorrie, line 139: On the naming of this, what is a `blog`? Is that a live or dead blog? I thought the language was liveblog with live `status = true|false`? And if the status is false, we sometimes refer to it as deadblog. Never heard of just a plain blog.

jamesgorrie (approved):
Less props = more simple. Nice

dotcom-rendering/src/web/lib/renderElement.tsx
OllysCoding: I guess a liveblog can be dead, but in format, you have `ArticleDesign.LiveBlog` and `ArticleDesign.DeadBlog`, so by extension and live-or-dead-blog could be considered just a blog? Unfortunately I've seen 'blog' refer to other things at the guardian as well!

OllysCoding (approved):
+1 for removing redundant data!

dotcom-rendering/src/web/lib/renderElement.tsx
mxdvl: As this is co-located with the following `format.design` condition, it feels clear enough even for a newcomer that implies either a LiveBlog or a DeadBlog.

oliverlloyd: I think there's also some history to this lexicon with references to 'sports blogs' and 'blogging'.

PR #4600, opened 2022-04-12
Proposal: DecideContainer

OllysCoding:
## What does this change?

Adds 'DecideContainer' as a way of rendering the correct front container

Adds support for backfill containers

## Why?

Step in the direction of rendering more fronts!

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -587,6 +587,8 @@ type FEContainerType =
 	| 'nav/media-list'
 	| 'news/most-popular';
 
+type DCRContainerType = FEContainerType;
+
 type FEFrontCard = {
 	properties: {
 		isBreaking: boolean;
@@ -752,6 +754,7 @@ type FECollectionType = {
 type DCRCollectionType = {
 	id: string;
 	displayName: string;
+	collectionType: DCRContainerType;
 	curated: DCRFrontCard[];
 	backfill: DCRFrontCard[];
 	treats: DCRFrontCard[];
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -587,6 +587,7 @@ type FEContainerType =
 	| 'nav/media-list'
 	| 'news/most-popular';
 
+// TODO: This may need to be declared differently than the front type in the future
 type DCRContainerType = FEContainerType;
 
 type FEFrontCard = {
--- a/dotcom-rendering/src/web/layouts/FrontLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/FrontLayout.tsx
@@ -1,16 +1,16 @@
 import { brandBackground, brandLine } from '@guardian/source-foundations';
 import { ArticleDesign, ArticleDisplay, ArticlePillar } from '@guardian/libs';
-
 import { Lines } from '@guardian/source-react-components-development-kitchen';
+import { renderContainer } from '../lib/renderContainer';
+
 import { SubNav } from '../components/SubNav.importable';
 import { ElementContainer } from '../components/ElementContainer';
 import { Nav } from '../components/Nav/Nav';
-
 import { Island } from '../components/Island';
 import { MostViewedFooterLayout } from '../components/MostViewedFooterLayout';
 import { decidePalette } from '../lib/decidePalette';
 import { Header } from '../components/Header';
-import { renderFrontContainer } from '../lib/renderContainer';
+import { ContainerLayout } from '../components/ContainerLayout';
 
 interface Props {
 	front: DCRFrontType;
@@ -105,10 +105,30 @@ export const FrontLayout = ({ front, NAV }: Props) => {
 
 			<main>
 				{front.pressedPage.collections.map((collection, index) => {
-					return renderFrontContainer({
-						collection,
-						showTopBorder: index > 1,
+					// TODO: We also need to support treats containers
+					// Backfills should be added to the end of any curated content
+					const trails = collection.curated.concat(
+						collection.backfill,
+					);
+					// There are some containers that have zero trails. We don't want to render these
+					if (trails.length === 0) return null;
+
+					const containerElement = renderContainer({
+						trails,
+						containerType: collection.collectionType,
 					});
+
+					return (
+						<ContainerLayout
+							title={collection.displayName}
+							showTopBorder={index > 0}
+							sideBorders={true}
+							padContent={false}
+							centralBorder="partial"
+						>
+							{containerElement}
+						</ContainerLayout>
+					);
 				})}
 
 				{!isPaidContent && (
--- a/dotcom-rendering/src/web/layouts/FrontLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/FrontLayout.tsx
@@ -1,7 +1,7 @@
 import { brandBackground, brandLine } from '@guardian/source-foundations';
 import { ArticleDesign, ArticleDisplay, ArticlePillar } from '@guardian/libs';
 import { Lines } from '@guardian/source-react-components-development-kitchen';
-import { renderContainer } from '../lib/renderContainer';
+import { DecideContainer } from '../lib/DecideContainer';
 
 import { SubNav } from '../components/SubNav.importable';
 import { ElementContainer } from '../components/ElementContainer';
@@ -113,11 +113,6 @@ export const FrontLayout = ({ front, NAV }: Props) => {
 					// There are some containers that have zero trails. We don't want to render these
 					if (trails.length === 0) return null;
 
-					const containerElement = renderContainer({
-						trails,
-						containerType: collection.collectionType,
-					});
-
 					return (
 						<ContainerLayout
 							title={collection.displayName}
@@ -128,7 +123,10 @@ export const FrontLayout = ({ front, NAV }: Props) => {
 							padContent={false}
 							centralBorder="partial"
 						>
-							{containerElement}
+							<DecideContainer
+								trails={trails}
+								containerType={collection.collectionType}
+							/>
 						</ContainerLayout>
 					);
 				})}
--- a/dotcom-rendering/src/web/layouts/FrontLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/FrontLayout.tsx
@@ -121,7 +121,9 @@ export const FrontLayout = ({ front, NAV }: Props) => {
 					return (
 						<ContainerLayout
 							title={collection.displayName}
-							showTopBorder={index > 0}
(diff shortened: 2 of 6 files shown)


dotcom-rendering/src/web/lib/renderContainer.tsx
```diff
@@ -0,0 +1,53 @@
...
+	switch (containerType) {
+		case 'dynamic/fast':
+			return <DynamicFast trails={trails} />;
+		case 'fixed/large/slow-XIV':
+			return <FixedLargeSlowXIV trails={trails} />;
+		default:
+			// TODO: This default allows us to render fronts in-development where we might
+			// not support all the container types, but it should be removed / re-investigated
+			// before fronts are released
+			return <DynamicFast trails={trails} />;
```
mxdvl, line 20: I would be tempted to have a `<Debug trails={trails}>` container as the default, at least in development. This could also tie in some tracking or graph, where we can see the remaining unsupported containers.

dotcom-rendering/src/web/lib/renderContainer.tsx
```diff
@@ -0,0 +1,53 @@
...
+	showTopBorder,
+}: {
+	collection: DCRCollectionType;
+	showTopBorder: boolean;
+}) => {
+	// TODO: We also need to support treats containers
+	// Backfills should be added to the end of any curated content
+	const trails = collection.curated.concat(collection.backfill);
+	// There are some containers that have zero trails. We don't want to render these
+	if (trails.length === 0) return null;
```
mxdvl, line 35: Should we log this as some kind of error?

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -9,9 +9,8 @@ import { Nav } from '../components/Nav/Nav';
 import { Island } from '../components/Island';
 import { MostViewedFooterLayout } from '../components/MostViewedFooterLayout';
 import { decidePalette } from '../lib/decidePalette';
-import { ContainerLayout } from '../components/ContainerLayout';
 import { Header } from '../components/Header';
-import { DynamicFast } from '../components/DynamicFast';
+import { renderFrontContainer } from '../lib/renderContainer';
```
mxdvl, line 13: Can this lib import be grouped with other libs?

dotcom-rendering/index.d.ts
```diff
@@ -587,6 +587,8 @@ type FEContainerType =
 	| 'nav/media-list'
 	| 'news/most-popular';
 
+type DCRContainerType = FEContainerType;
```
mxdvl, line 590: Could we add a `// TODO` here for future us? Alternatively, only introduce the `DCRContainerType` once it’s not identical to `FEContainerType`.

mxdvl (commented):
Exciting!

dotcom-rendering/src/web/lib/renderContainer.tsx
```diff
@@ -0,0 +1,53 @@
...
+			return <FixedLargeSlowXIV trails={trails} />;
+		default:
+			// TODO: This default allows us to render fronts in-development where we might
+			// not support all the container types, but it should be removed / re-investigated
+			// before fronts are released
+			return <DynamicFast trails={trails} />;
+	}
+};
+
+export const renderFrontContainer = ({
```
oliverlloyd, line 24: Is it worth maybe bringing this logic back out into `FrontLayout` and only having the decide container logic here? Our layout files _tend_ to be where these `ContainerLayout` components live

oliverlloyd (approved):
This is fantastic!

A few naming and location nitpicks but this approach is very similar to the mental model I have in my head for this. Great stuff.

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -106,21 +105,10 @@ export const FrontLayout = ({ front, NAV }: Props) => {
...
-							sideBorders={true}
-							padContent={false}
-							centralBorder="partial"
-						>
-							<DynamicFast trails={trails} />
-						</ContainerLayout>
-					);
+					return renderFrontContainer({
+						collection,
+						showTopBorder: index > 1,
```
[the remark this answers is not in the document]
oliverlloyd: Actually, based on my later comment, if we keep the `ContainerLayout` local to `FrontLayout` we don't need this at all

dotcom-rendering/src/web/lib/renderContainer.tsx
OllysCoding: I don't think this would be an error? If it's possible to set these in fronts tool we probably want to treat it as expected behaviour

dotcom-rendering/src/web/lib/renderContainer.tsx
OllysCoding: I think this could be really useful, but I wonder if while we don't support many container types, we would potentially end up with most fronts rendering no cards at all which might not be as useful for development

OllysCoding: In terms of logging container types we don't support, this might be easier done in Frontend?

dotcom-rendering/src/web/layouts/FrontLayout.tsx
OllysCoding: I don't think so, since there's no single export file from the `/lib` directory.

dotcom-rendering/src/web/lib/renderContainer.tsx
```diff
@@ -0,0 +1,53 @@
+import { ContainerLayout } from '../components/ContainerLayout';
+import { DynamicFast } from '../components/DynamicFast';
+import { FixedLargeSlowXIV } from '../components/FixedLargeSlowXIV';
+
+type Props = {
+	trails: DCRFrontCard[];
+	containerType: DCRContainerType;
+};
+
+const renderContainer = ({ trails, containerType }: Props) => {
```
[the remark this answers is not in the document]
OllysCoding: I named it after 'renderElement' but I think 'decideContainer' might make more sense for what we're trying to achieve here, at least for now!

dotcom-rendering/src/web/lib/renderContainer.tsx
OllysCoding: 👍  At first I was concerned that the logic for deciding a `trails` might grow more complex and therefor having it abstracted would be useful, but I think since it's looking relatively simple, keeping it in `FrontLayout` should be fine for now!

dotcom-rendering/src/web/layouts/FrontLayout.tsx
OllysCoding: Nevermind I think I know what you mean haha!

mxdvl: Sorry, I meant as in the line ordering. Probably one for an ESLint config to sort imports alphabetically!

dotcom-rendering/src/web/lib/renderContainer.tsx
oliverlloyd: Maybe we can capture this in ELK?

OllysCoding:
@mxdvl @oliverlloyd Updated the PR would love some feedback :)

dotcom-rendering/src/web/lib/renderContainer.tsx
mxdvl: Yes, this would be entirely appropriate for ELK, but I’m not sure we’ve got the pipeline set up for this yet. Part of what the data bubble will look into.

dotcom-rendering/src/web/lib/renderContainer.tsx
mxdvl: I might be a little unfamiliar with the lingo. What’s the purpose of a container with no trails?

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -106,19 +105,30 @@ export const FrontLayout = ({ front, NAV }: Props) => {
...
-					const trails = collection.curated;
+					// TODO: We also need to support treats containers
+					// Backfills should be added to the end of any curated content
+					const trails = collection.curated.concat(
+						collection.backfill,
+					);
 					// There are some containers that have zero trails. We don't want to render these
 					if (trails.length === 0) return null;
+
+					const containerElement = renderContainer({
```
oliverlloyd, line 116: Can we drop the function pattern here and just make this a component? I think this special pattern only exists in *one* case, `renderElement` which was done as a sort of subjective / for reasons change but ultimately if your function is a react component then it should _be_ a react component and follow the same rules (Pascal case, be rendered using `<Thing />` syntax, etc)

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -106,19 +105,30 @@ export const FrontLayout = ({ front, NAV }: Props) => {
...
+						trails,
+						containerType: collection.collectionType,
+					});
+
 					return (
 						<ContainerLayout
 							title={collection.displayName}
+							// TODO: This logic should be updated, as this relies
+							// on the first container being 'palette styles do not delete'
```
mxdvl, line 125: I think we should extract that as `STARTING_INDEX` and keep the comment there.

oliverlloyd, line 125: Thanks for this!

mxdvl (approved):
A step in the right direction.

dotcom-rendering/src/web/lib/renderContainer.tsx
oliverlloyd: I think there is zero purpose. But the tool and, consequently, the model allow it so we need to have this line. But I don't think we should render anything if there are no trails.

Perhaps historically there have been examples of people using containers as vehicles for weird css but I don't think we should support this pattern.

dotcom-rendering/src/web/lib/renderContainer.tsx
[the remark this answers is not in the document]
oliverlloyd: I think based on changes and my earlier comment, this should be 1) treated like a component and named `DecideContainer` and I guess at that point it should go in /components. I don't think we should allow react components to live in libs but this is me being _feely_ and I bet there are lots of examples to counter my feels so I'm open to push back here

dotcom-rendering/src/web/layouts/FrontLayout.tsx
OllysCoding: Made this change, definitely felt like it cleaned up the code a lot, thanks!

oliverlloyd (approved):
LGTM!

PR #4621, opened 2022-04-13
Remove sticky videos switch from renderElement

joecowton1:
## What does this change?

Removes the sticky videos switch from the `stickyVideos` conditional in `renderElement`. **Once merged the Sticky Videos will be available to all users on PROD.**

The companion PR in Frontend is [here](https://github.com/guardian/frontend/pull/24889).

### DIFF
--- a/dotcom-rendering/src/web/lib/renderElement.tsx
+++ b/dotcom-rendering/src/web/lib/renderElement.tsx
@@ -731,7 +731,7 @@ export const renderElement = ({
 						mediaTitle={element.mediaTitle}
 						altText={element.altText}
 						origin={host}
-						stickyVideos={isBlog && switches.stickyVideos}
+						stickyVideos={isBlog}
 					/>
 				</Island>,
 			];


alinaboghiu (approved):

oliverlloyd:
Do we want to keep a switch for sticky videos?

In Dotcom we maintain a [switchboard](https://frontend.gutools.co.uk/dev/switchboard) of flags that allow us and editorial to turn on and off certain features in the event that there are issues without the need for a deployment. I think sticky videos are a good fit for such a switch.

Sadly, it's a bit confusing because we also use similar screens and even code objects when doing AB testing. We use the switchboard to also turn the _tests_ on and off. This PR is removing a check for that _server_ side AB test but maybe we want to keep this code and convert the switch from a server side test switch to a plain feature switch?

oliverlloyd (changes requested):
Hmm, upon reflection. I think we do need to have a feature switch for this. Sorry for the 'request changes' redness but I think we need to give editorial the ability to turn this feature off out of hours if no engineers are available to create a PR.

joecowton1:
This PR is no longer required, see https://github.com/guardian/frontend/pull/24889/files.

PR #4626, opened 2022-04-13
Defer tweet enhancement

### LINKED ISSUE #4620: Defer tweet enhancement

oliverlloyd:
## What does this change?
Defers enhancement of Twitter tweets until they scroll into view.  Fixes #4620 

This work matches what was [originally](https://github.com/guardian/frontend/pull/6203) [implemented](https://github.com/guardian/frontend/pull/6244) in Frontend

Previously we were using a `script` tag on twitter embeds which [loaded the twitter javascript](https://developer.twitter.com/en/docs/twitter-for-websites/embedded-tweets/overview) for each tweet. When this code was loaded and run it found each tweet on the page and 'enhanced' it.

We now [centralise the loading](https://developer.twitter.com/en/docs/twitter-for-websites/javascript-api/guides/set-up-twitter-for-websites) of this javascript and have [more control](https://developer.twitter.com/en/docs/twitter-for-websites/javascript-api/guides/scripting-loading-and-initialization) over when the enhancement happens.

## Why?
To improve page load experience, esp. for people using permalinks

## How are we deferring the `load` action?
Twitter will automatically run it's enhancement js on all tweets that have the `.twitter-tweet` class. By removing this class at the point we get the raw html from Frontend (using an 'enhancer') we prevent this happening. We later only add this classname back in when the element is visible and then run `twitter.widgets.load` on just that part of the page

### `enhance` and `enhance`?
Sadly, there are two meanings to the word enhance in this PR. The Twitter meaning, which is to take a basic embedded tweet and to add more features, images, tracking, etc. to it and then DCR's meaning, which is where we have a series of 'enhance' functions that take raw Frontend data and transform it into something more useable for DCR.


dotcom-rendering/src/model/enhance-tweets.ts
```diff
@@ -0,0 +1,33 @@
...
+				const withoutTweetClass = element.html.replace(
+					'twitter-tweet',
+					'nojs-tweet',
+				);
+				enhanced.push({
+					...element,
+					...{
+						html: withoutTweetClass,
+					},
+				});
```
jamesgorrie, line 18: not a biggy, but maybe easier to read
```suggestion ts
				enhanced.push({
					...element,
					html: withoutTweetClass,
				});
```

dotcom-rendering/src/web/components/TweetBlockComponent.importable.tsx
```diff
@@ -0,0 +1,79 @@
...
+		/* stylelint-disable-next-line declaration-no-important */
+		width: 100% !important;
+	}
+
+	.nojs-tweet p {
+		padding-bottom: 10px;
+	}
+
+	a {
+		/* stylelint-disable-next-line color-no-hex */
```
jamesgorrie, line 32: ? Is this a twitter colour?

jamesgorrie (approved):
This is a very elegant solution!

jamesgorrie:
Bikeshed re: enhance and enhance - why do we use `enhance` over `transform`? It just seems like we don't add much, rather just change. 

I think I get twitters reason that you have a plain tweet with noJS and a fancy one with JS, thus enhanced.

dotcom-rendering/src/web/components/TweetBlockComponent.importable.tsx
oliverlloyd: No idea 🤷  - this is isn't my code, I renamed the file so it got green'd

dotcom-rendering/src/web/server/articleToHtml.tsx
```diff
@@ -193,5 +219,9 @@ export const articleToHtml = ({ data }: Props): string => {
 		openGraphData,
 		twitterData,
 		keywords,
+		initTwitter:
+			pageHasTweetElements || format.design === ArticleDesign.LiveBlog
```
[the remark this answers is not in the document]
oliverlloyd: Yes, basically. I think there's a similar preference in Frontend to - kind of - expect tweet embeds in blogs.

The alternative is to not load the javascript here and defer it to when (and if) a tweet appears. The downside to this is a slower experience for that tweet block and more complex code.

dotcom-rendering/src/model/enhance-tweets.ts
```diff
@@ -0,0 +1,33 @@
+const removeTweetClass = (elements: CAPIElement[]): CAPIElement[] => {
+	const enhanced: CAPIElement[] = [];
+	elements.forEach((element) => {
```
mxdvl, line 3: If we mutate, I think `for … of` is more explicit

oliverlloyd:
> Bikeshed re: enhance and enhance - why do we use `enhance` over `transform`? It just seems like we don't add much, rather just change.
> 
> I think I get twitters reason that you have a plain tweet with noJS and a fancy one with JS, thus enhanced.

I guess one word makes more sense over another at a point in time and perhaps over time things change? I know in Frontend the same sorts of actions are called 'cleaners' which is an interesting choice. Maybe we wanted to be more positive alongside this?

dotcom-rendering/src/web/components/TweetBlockComponent.importable.tsx
mxdvl: This is somewhat odd… usually git managed to see this as a rename :(

dotcom-rendering/src/web/lib/renderElement.tsx
```diff
@@ -600,6 +600,14 @@ export const renderElement = ({
...
 		case 'model.dotcomrendering.pageElements.TweetBlockElement':
+			if (switches.enhanceTweets) {
+				return [
+					true,
+					<Island deferUntil="visible">
+						<TweetBlockComponent element={element} />
+					</Island>,
+				];
+			}
 			return [true, <TweetBlockComponent element={element} />];
```
mxdvl, line 611: Wouldn’t that get two tweets?

dotcom-rendering/src/web/server/articleToHtml.tsx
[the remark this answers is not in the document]
mxdvl: That’s how I read this

mxdvl (approved):
Nice

dotcom-rendering/src/web/lib/renderElement.tsx
oliverlloyd: It's returning, so you can only get one. My linter insists on this pattern.

jamesgorrie:
> cleaners

I suppose the thinking was it only ever removed stuff.

> Maybe we wanted to be more positive alongside this?

True, but it's nice to try use language others will understand quickly, but, as usual that's hard, especially as things change.

PR #4628, opened 2022-04-13
Fix dev server query url

abeddow91:
## What does this change?
Adds additional url parsing logic to the replace the first `&` in the url with a `?` if we are in the DEV environment.

## Why?
When in DEV environment, we start each query string with `?url=`. This means that when we set additional query params on the client side (such as `filterKeyEvents`), they are added to the existing query string with an `&` when, actually, not query string exists within the url itself. This breaks the build as our url is passed back as `https://www.theguardian.com/my/article&filterKeyEvents=true` instead of `https://www.theguardian.com/my/article?filterKeyEvents=true`. 

## Screenshots

| Before      | After      |
|-------------|------------|
| Screenshot 2022-04-13 at 17 01 41 | Screenshot 2022-04-13 at 16 47 49


mxdvl (commented):
Could we add tests for a double `?`, and one where we don't have special characters?

mxdvl (approved):

oliverlloyd (approved):
Very nice code!

PR #4629, opened 2022-04-14
Disable `LiveBlogEpic` for permalinks

oliverlloyd:
## What does this change?
This prevents the `LiveBlogEpic` from rendering when the reader is using a permalink

## Why?
Because this epic will cause content to shift down unexpectedly after the page has already appeared on the screen. This is a poor experience in general but it is especially problematic for readers that are using permalinks where they justifiably expect the page to show them the block they are requesting (instead of showing it to them and then snatching it away as it gets pushed down the page)

## But... 💲 ?!
There's a balance to be made here. It is true in an absolute sense that if you show more banners then you get more money. But this is simplistic and does not take into account other factors that also impact how readers choose to behave. If the product you are giving people is unpleasant to use and does not work well then ultimately, you will get fewer dollars. I can't measure this and prove it to you though so if you only allow things that can be easily measured to influence decisions then we should not merge this PR and we should cover the site in banners. But if you allow for more subjective reasoning then my argument here is we should reduce the volume of these banners (by a bit) in favour of a giving people a better experience when reading our content

| Before 😠       | After 😃       |
|-------------|------------|
|  |  |

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleBody.tsx
+++ b/dotcom-rendering/src/web/components/ArticleBody.tsx
@@ -146,7 +146,8 @@ export const ArticleBody = ({
 						globalH3Styles(format.display),
 						globalLinkStyles(palette),
 						// revealStyles is used to animate the reveal of new blocks
-						format.design === ArticleDesign.LiveBlog &&
+						(format.design === ArticleDesign.DeadBlog ||
+							format.design === ArticleDesign.LiveBlog) &&
 							revealStyles,
 					]}
 				>
--- a/dotcom-rendering/src/web/components/LiveBlogEpic.importable.tsx
+++ b/dotcom-rendering/src/web/components/LiveBlogEpic.importable.tsx
@@ -276,8 +276,22 @@ export const LiveBlogEpic = ({
 		isPaidContent,
 		tags,
 	});
-
 	if (!payload) return null;
+
+	/**
+	 * Because this epic is loaded client side it will cause CLS (it will shift the
+	 * blocks down after the page has already loaded). This is a poor reader experience
+	 * in general because it moves the content whilst they are trying to read it but
+	 * it is especially problematic where a permalink is being used, so we're disabling
+	 * this element in those cases.
+	 */
+	if (window.location.href.includes('#block-')) {
+		log(
+			'dotcom',
+			'A permalink was detected so the LiveBlogEpic will not show',
+		);
+		return null;
+	}
 
 	// Using the payload, make a fetch request to contributions
 	return (
--- a/dotcom-rendering/src/web/components/LiveBlogEpic.importable.tsx
+++ b/dotcom-rendering/src/web/components/LiveBlogEpic.importable.tsx
@@ -1,3 +1,4 @@
+import { createPortal } from 'react-dom';
 import React, { useEffect, useState } from 'react';
 import { css } from '@emotion/react';
 import { space } from '@guardian/source-foundations';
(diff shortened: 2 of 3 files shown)


jamesgorrie:
Isn't this actually a place where we can use a simplistic measure of "Do we lose conversions by removing the banner?"?

If that is the case, it might push us harder to look for a better technical implementation e.g. rendering the banner server side, or only taking the person to the link after it's loaded (the later sounds rubbish).

oliverlloyd:
> Isn't this actually a place where we can use a simplistic measure of "Do we lose conversions by removing the banner?"?
> 
> If that is the case, it might push us harder to look for a better technical implementation e.g. rendering the banner server side, or only taking the person to the link after it's loaded (the later sounds rubbish).

I think this is what I was speaking to when I said that favouring the things that we can measure is _too_ simplistic and perhaps we want to factor in more subjective things in the decision? I mean, sure, if you show the banner less then you will inevitably get fewer conversions but I don't think that should be a barrier to change. If we only factor in this in the decisions around when to show it then we would delete all the news content and just show one big banner! 🙃

jamesgorrie:
We need a set of metrics to show that we're improving if we're going to impair others. What are the metrics We're proposing to increase?

Assuming these should be around attention time, dropoff, CWV, ...? My worry is that we say that this UX improvement is worth the `n` amount of conversions we lose because we think so, which doesn't sound massively convincing. 

We haven't let the other team know what `n` is to help them decide, and we haven't given ourselves any data to work with.

Saying that if we did this as a test and there was no drop in conversions, this is a moot point, but at least it's a point. 

We know we have tensions between UX and some commercial features, and should try to figure out how to weigh the decisions up. This might also help for decisions like this in the future.

I also still think we could be fixing a more fundamental issue with how we render commercial components with the wrong tool (turn them off, there's no issue).

oliverlloyd:
> My worry is that we say that this UX improvement is worth the `n` amount of conversions we lose because we think so, which doesn't sound massively convincing.

I think my whole point here is that it is not possible to measure this impact but that we should not allow the fact that you can't measure something be a gate to making the change. I think if we enforce a mindset that all changes must be measurable then we risk not seeing the wider picture.

I think that this change should be considered a bug fix. We offer the ability for journalists to offer readers links to their content. These are shared on Twitter and other platforms. CLS _breaks_ this feature, the permalink no longer works as intended. This is a bug that got introduced when the liveblog epic got added to the blog. Arguably, it should have been caught then but we didn't have tests for CLS then so it got through.

Note. Blocking changes based on negative impact to CLS is _explicitly_ something commercial already do. We didn't do this for this change but that should not stop us making this change retrospectively.

In Frontend there are lots of similar PRs to this one where changes are being made in an effort to not break permalinks, I consider this change to be of the same family and think DCR should continue to support a good permalink experience, and yes, even if that impacts conversions.

jamesgorrie:
> I think that this change should be considered a bug fix.

Agreed, but it's a bug fix that just removes the feature, rather than fixing the fact it creates CLS. This is definitely a harder thing to fix.

>  Blocking changes based on negative impact to CLS is explicitly something commercial already do

This is what I am trying to say - we do have a metric this positively changes. Sorry if it's unclear. I am also not saying we block this, but not having an indication of how we might affect metrics with a change feels chancy, and I would say set's a bad president. 

Given [this is a lab field in CWV](https://web.dev/vitals/#lab-tools-to-measure-core-web-vitals) we could have it reported more visibly in the PR - @mxdvl and @jlieb10 are talking about this. Having this reported will also also help us avoid regressions.

On the wider picture - we should be working at having a set of metrics laid out in a way that can give us that wider picture at a glance. 

I'll let the other team decide on whether they would want more insight into how this might affect their metrics.

jamesgorrie (approved):
From a code and UX POV 👍

oliverlloyd:
Closing in favour of a better, more dynamic, solution

PR #4630, opened 2022-04-14
refactor(ophan): use conditional chaining

mxdvl:
## What does this change?

Use conditional chaining for possibly undefined window object for the ophan tracker.

## Why?

Less lines of code, more idiomatic.


ioannakok (approved):
Nice!

PR #4637, opened 2022-04-14
feat: Add margin to liveblog main media captions when on a tablet display

### LINKED ISSUE #4150: Mainmedia caption at tablet needs some margin-left
image

AshCorr:
## What does this change?

Adds left and right padding to main media captions on Liveblogs when on a tablet display.

## Why?

Closes #4150 

## Screenshots

| Before  / After     |
|-------------|
| ![before][] | 
| ![after][] |

[before]:  https://user-images.githubusercontent.com/21217225/163406438-c1df2758-38b9-415e-b95d-685912c746e2.png
[after]: https://user-images.githubusercontent.com/21217225/163405928-e7281644-6641-409e-a069-af99bd6ae2fe.png

### DIFF
--- a/dotcom-rendering/src/web/components/Caption.tsx
+++ b/dotcom-rendering/src/web/components/Caption.tsx
@@ -28,6 +28,7 @@ type Props = {
 	isOverlayed?: boolean;
 	isLeftCol?: boolean;
 	mediaType?: MediaType;
+	isMainMedia?: boolean;
 };
 
 type IconProps = {
@@ -116,6 +117,12 @@ const veryLimitedWidth = css`
 const captionPadding = css`
 	padding-left: 8px;
 	padding-right: 8px;
+`;
+
+const mobileCaptionPadding = css`
+	${until.desktop} {
+		${captionPadding}
+	}
 `;
 
 const videoPadding = css`
@@ -231,6 +238,7 @@ export const Caption = ({
 	isOverlayed,
 	isLeftCol,
 	mediaType = 'Gallery',
+	isMainMedia = false,
 }: Props) => {
 	// Sometimes captions come thorough as a single blank space, so we trim here to ignore those
 	const noCaption = !captionText?.trim();
@@ -239,6 +247,9 @@ export const Caption = ({
 	if (noCaption && (noCredit || hideCredit)) return null;
 
 	const palette = decidePalette(format);
+	const isLiveblog =
+		format.design === ArticleDesign.LiveBlog ||
+		format.design === ArticleDesign.DeadBlog;
 
 	const defaultCaption = (
 		<figcaption
@@ -247,6 +258,7 @@ export const Caption = ({
 				shouldLimitWidth && limitedWidth,
 				!isOverlayed && bottomMargin,
 				isOverlayed && overlayedStyles(palette, format),
+				isMainMedia && isLiveblog && mobileCaptionPadding,
 				padCaption && captionPadding,
 				mediaType === 'Video' && videoPadding,
 			]}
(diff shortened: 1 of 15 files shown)


dotcom-rendering/src/web/components/ImageComponent.tsx
```diff
@@ -368,6 +368,7 @@ export const ImageComponent = ({
 										displayCredit={element.displayCredit}
 										shouldLimitWidth={shouldLimitWidth}
 										isOverlayed={true}
+										isMainMedia={true}
```
oliverlloyd, line 371: Is the prop value of `isMainMedia` always going to be true here? If so, it might be safer to just pass it down. I can imagine a future when things are moved around or copy pasted and this potentially creating hard to find bugs.

dotcom-rendering/src/web/components/Caption.tsx
```diff
@@ -239,6 +247,9 @@ export const Caption = ({
 	if (noCaption && (noCredit || hideCredit)) return null;
 
 	const palette = decidePalette(format);
+	const isLiveblog =
```
mxdvl, line 250: We’ve picked up `isBlog` recently, might be a valid option here.

```suggestion
	const isBlog =
```

dotcom-rendering/src/web/components/ImageComponent.tsx
```diff
@@ -387,6 +388,7 @@ export const ImageComponent = ({
 						credit={element.data.credit}
 						displayCredit={element.displayCredit}
 						shouldLimitWidth={shouldLimitWidth}
+						isMainMedia={true}
```
oliverlloyd, line 391: Ditto on passing through the root var for `isMainMedia`

mxdvl (approved):
🎨

dotcom-rendering/src/web/components/Caption.tsx
```diff
@@ -247,6 +258,7 @@ export const Caption = ({
 				shouldLimitWidth && limitedWidth,
 				!isOverlayed && bottomMargin,
 				isOverlayed && overlayedStyles(palette, format),
+				isMainMedia && isLiveblog && tabletCaptionPadding,
```
oliverlloyd, line 261: What do you think about merging this into `captionPadding`? You could make the original css thing a function and pass in the props it needs to make the choice?

oliverlloyd (approved):
Looks great! 

Some small comments but nothing major.

oliverlloyd:
Chore: I notice Chromatic didn't capture this change so we probably want a story that renders main media at this breakpoint.

dotcom-rendering/src/web/components/Caption.tsx
AshCorr: Thats actually a much better name, thanks!

AshCorr:
> Chore: I notice Chromatic didn't capture this change so we probably want a story that renders main media at this breakpoint.

It did! I just accepted the change already!

dotcom-rendering/src/web/components/Caption.tsx
AshCorr: Hm, maybe, let's give it a try!

OllysCoding (approved):
+1 LGTM!

mxdvl:
For conditional CSS, you might just want to create two helpers, and let the the css helper do its thing, as in https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/CalloutBlockComponent.importable.tsx#L319

```tsx
const a = css`/* … */`
const b = css`/* … */`

<Element css={[a, condition && b]}>
```

AshCorr:
> For conditional CSS, you might just want to create two helpers, and let the the css helper do its thing, as in https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/CalloutBlockComponent.importable.tsx#L319
> 
> ```tsx
> const a = css`/* … */`
> const b = css`/* … */`
> 
> <Element css={[a, condition && b]}>
> ```

Yeaa, thinking I should just stick with the original solution... the idea of moving it all to a single function seemed nice but I think it probably ends up being harder to understand than what was there before

PR #4640, opened 2022-04-14
fix(email): replace + with %20 in email shares

### LINKED ISSUE #4638: Subject line appears with plus signs when sharing article
Issue reported in Dotcom channel:

> A user has reported that the subject line appears with plus signs when sharing this article (https://www.theguardian.com/commentisfree/2022/apr/10/stench-entitlement-now-oozing-from-rishi-sunaks-home-as-well-as-boris-johnsons) from the web. I can reproduce using an iPhone 12 Pro (15.3.1) with Safari and Gmail.

image

> It didn't happen with this liveblog: https://www.theguardian.com/politics/live/2022/apr/13/boris-johnson-uk-politics-live-rishi-sunak-partygate-lockdown but does for other articles.

> I can only reproduce the issue with the Apple mail app. It doesn't when I use Gmail as the default.

mxdvl:
## What does this change?

Replace the + by encoded spaces `%20` so Safari mail creates the right subject lines

## Why?

Fixes #4638

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/19683595/163427203-a7686786-fe46-4438-8ea7-5a99dcaa5113.png
[after]: https://user-images.githubusercontent.com/76776/163432814-f5a0540f-b996-44b3-b7a9-9ac4947dddda.png

### DIFF
--- a/dotcom-rendering/src/web/components/ShareIcons.tsx
+++ b/dotcom-rendering/src/web/components/ShareIcons.tsx
@@ -234,7 +234,9 @@ export const ShareIcons = ({
 								blockId,
 								CMP: 'share_btn_link',
 							}),
-						})}`}
+						})
+							.toString()
+							.replace(/\+/g, '%20')}`}
 						role="button"
 						aria-label="Share via Email"
 						target="_blank"


ioannakok (approved):
👏

PR #4643, opened 2022-04-15
Decide `LiveBlogEpic` position based on if there is a permalink

oliverlloyd:
## What does this change?
This PR moves the decision about where to position the `LiveBlogEpic` from the server to the client. We also now animate in the epic when a permalink is in use to reduce any confusion from new content suddenly appearing on the page

## Why?
This allows us to take into consideration if a permalink is being used and choose to place the epic below the linked to block in that case. This prevents CLS from breaking the permalink experience

This PR replaces https://github.com/guardian/dotcom-rendering/pull/4629. After suggestions from @tomrf1 @TomPretty I was able to implement a solution that fixed the CLS problem with permalinks without reducing the volume of impressions for the epic. Indeed, this could potentially increase conversions as we're now making the epic more prominent for those 8% of readers using permalinks

| Before      | After      |
|-------------|------------|
| before | after |

### DIFF
--- a/dotcom-rendering/src/web/components/LiveBlogEpic.importable.tsx
+++ b/dotcom-rendering/src/web/components/LiveBlogEpic.importable.tsx
@@ -276,8 +276,22 @@ export const LiveBlogEpic = ({
 		isPaidContent,
 		tags,
 	});
-
 	if (!payload) return null;
+
+	/**
+	 * Because this epic is loaded client side it will cause CLS (it will shift the
+	 * blocks down after the page has already loaded). This is a poor reader experience
+	 * in general because it moves the content whilst they are trying to read it but
+	 * it is especially problematic where a permalink is being used, so we're disabling
+	 * this element in those cases.
+	 */
+	if (window.location.href.includes('#block-')) {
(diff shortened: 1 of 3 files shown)


tomrf1 (approved):
Looks good - such a nice solution!

TomPretty:
Yeah looks great to me - glad to see it wasn't too much extra effort and I think this makes for a neat solution!

dotcom-rendering/src/web/components/LiveBlogEpic.importable.tsx
```diff
@@ -201,6 +202,12 @@ const Render = ({
 	);
 };
 
+/**
+ * Using the payload, make a fetch request to contributions service
+ *
+ * @param payload - unknown
+ * @param contributionsServiceUrl - string
```
mxdvl, line 209:
```suggestion
 * @param {unknown} payload
 * @param {string} contributionsServiceUrl
```

mxdvl (commented):
I'm not entirely sure the animation works, but otherwise this is a great approach!!

PR #4644, opened 2022-04-15
Trail text on fronts

### LINKED ISSUE #4613: Cards on front pages show the standfirst
We support `standfirst` as a string but the data model is sending us this as html

oliverlloyd:
## What does this change?
This includes the trailText in cards that appear in front containers. This fixes #4613 

## Why?
For that parity win

## When should the trailText appear?
This is specific to each container. We currently support `DynamicFast` and `FixedLargeSlowXIV` which both show the trailText for the first card only*

*This assertion is based on experimentation using the fronts tool and also from clicking around the site looking at examples but it's possibly there are edge cases outside this rule

| Before      | After      |
|-------------|------------|
| Screenshot 2022-04-15 at 17 10 42 | Screenshot 2022-04-15 at 17 08 59 |


paperboyo:
Just regarding terminology: it’s Trail texts that appear on fronts, not Standfirsts (a separate field in Composer and the only one available in the frontstool).

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -230,7 +230,11 @@ export const Card = ({
 							<div>
 								{standfirst && (
 									<StandfirstWrapper palette={cardPalette}>
-										{standfirst}
+										<div
+											dangerouslySetInnerHTML={{
+												__html: standfirst,
+											}}
```
OllysCoding, line 236: Will this receive the correct styling in terms of font, size, etc, or do we need an emotion global to take care of this like we have on articles?

OllysCoding (approved):
+1 Looks good to me (with 1 question)

mxdvl (commented):
Can we rename the prop to match the convention of “trail text” as per @paperboyo’s suggestion?

dotcom-rendering/src/model/enhanceCollections.ts
```diff
@@ -9,6 +9,7 @@ const enhanceCards = (collections: FEFrontCard[]): DCRFrontCard[] => {
 				format,
 				url: faciaCard.header.url,
 				headline: faciaCard.header.headline,
+				standfirst: faciaCard.card.trailText,
```
mxdvl, line 12: This is a good indication that we probably should not rename.

oliverlloyd:
> Can we rename the prop to match the convention of “trail text” as per @paperboyo’s suggestion?

We absolutely can, and should, so we shall!

dotcom-rendering/src/model/enhanceCollections.ts
oliverlloyd: Yeah, and it's a nicer name too

oliverlloyd: Yeah, and it's a nicer name too

mxdvl (approved):

dotcom-rendering/src/web/components/Card/Card.tsx
oliverlloyd: Looking at today's front, there is a single example of the `trailText` containing html, and it's a `strong` tag.
Screenshot 2022-04-22 at 08 40 36
Looking in the Fronts Tool, it seems like this field is just plain text so I think there's a case to say this field is unlikely to contain html as you'd need to manually write it but it's not impossible.

That said, I recall the reason I decided to dangerously set it was because I had seen a p tag in use here and because I also see one other example I am inclined to keep this approach. But I don't think we need to add any additional styling. We can review this as more examples arrive.

mxdvl: It's quite simple to add some logs for these cases, so we get a sense of how frequently they happen.

PR #4648, opened 2022-04-18
Align cards better

oliverlloyd:
__This PR is branched from #4647__
## What does this change?
This PR fixes an issue where cards in adjacent columns were not aligning with each other. It does this by adding a new prop to `UL` of `wrapCards: boolean;` and then adding a little more control to how we decide when to pad the bottom of a card.

#### `margin-bottom` 👎 , `padding-bottom` 👍  
Essentially, I did this so that I could go from

Screenshot 2022-04-18 at 10 41 05

to

Screenshot 2022-04-18 at 10 40 28

## Why?
Because it's more pleasing to have things aligned

## Why flex and not grid?
We're using `flex-wrap` here but this layout could also be achieved using a css grid. I considered this approach - it was actually how I expected to solve this problem (look at the branch name 😉 ) - but after starting to add the grid code it became clear it would be a lot of work so I gave some time to trying to get `flex-wrap` to work and in the end this turned out easier than I thought.

The advantages to continuing to support the existing `LI` / `UL` approach are:

1) this is how Frontend does it
2) it's more accessible
3) it's simpler
4) it means we continue to use one approach
5) and we don't need to add any new components or abstraction

| Before      | After      |
|-------------|------------|
| Screenshot 2022-04-18 at 10 27 10 | Screenshot 2022-04-18 at 10 26 14|


mxdvl (approved):
Seems sound–a good flex 💪.

mxdvl (commented):
Am I right in thinking that the order is now in columns, rather than rows:

```
1 | 4
2 | 5
3 | 6
```

Whereas before it was:

```
1 | 2
3 | 4
5 | 6
```

oliverlloyd:
> Am I right in thinking that the order is now in columns, rather than rows:
> 
> ```
> 1 | 4
> 2 | 5
> 3 | 6
> ```
> 
> Whereas before it was:
> 
> ```
> 1 | 2
> 3 | 4
> 5 | 6
> ```

Correct. We now align with how Frontend does this

PR #4649, opened 2022-04-18
Added `imagePositionOnMobile`

### LINKED ISSUE #4607: Larger images on mobile
On mobile `Huge` (and/or maybe `boosted`?) images should appear larger, above the text

Screenshot 2022-04-12 at 14 18 39

oliverlloyd:
## What does this change?
This PR fixes #4607 by adding

1) The `imagePositionOnMobile` prop to `Card` and
2) Extending `ImagePositionType` to include `bottom`

Along the way I refactored the code to remove the `alwaysVertical` prop

## Why
We want to be able to render the headline above the image to support some front layouts.

I removed `alwaysVertical` as this was added for the specific use case of carousels where we no longer wanted to default to showing the image on the left at mobile. But now that we have two new use cases where this default no longer works I felt there was justification for a different api so I replaced

```ts
type Props = {
  imagePosition: ImagePositionType;
  alwayVertical: boolean;
  ...
}
```

with

```ts
type Props = {
  imagePosition: ImagePositionType;
  imagePositionOnMobile: ImagePositionType;
  ...
}
```

| Before      | After      |
|-------------|------------|
| Screenshot 2022-04-18 at 10 50 37 | Screenshot 2022-04-18 at 10 49 57 |

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -889,7 +889,7 @@ interface BadgeType {
 	imageUrl: string;
 }
 
-type ImagePositionType = 'left' | 'top' | 'right';
+type ImagePositionType = 'left' | 'top' | 'right' | 'bottom';
 
 type SmallHeadlineSize = 'tiny' | 'small' | 'medium' | 'large';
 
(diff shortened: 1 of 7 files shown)


mxdvl (approved):
simple

Issue #4606, opened 2022-04-12
`FEContainerPalette` is used to style containers


oliverlloyd:
Create the framework for passing [this property](https://docs.google.com/document/d/1nkntFJUunqrwNYszTrc2RBfyYUd1bEAeQh76yVFOLjo/edit#heading=h.n8dqpx8lx9rv) into containers so that we can make styling decisions using it.

This could simply be a prop or perhaps we extend the model more?

PR #4650, opened 2022-04-18
Better stories for `Card`

oliverlloyd:
## What does this change?
This PR refactors the stories for the `Card` component.

### Different themes
Screenshot 2022-04-18 at 20 45 00

### Different media types
Screenshot 2022-04-18 at 20 51 00

### Different headline sizes
Screenshot 2022-04-18 at 20 53 01

### Different image positions
Screenshot 2022-04-18 at 20 46 27

etc.

I've brought the variations that lived in different files into one location, adding more coverage and using loops to iterate `Format` properties so we can be sure to respond to any changes(such as [this one](https://github.com/guardian/libs/pull/332)).

## What else?
I have deleted the stories in the `examples` folder. These were useful as a POC to demonstrate what DCR was capable of but now it is _doing_ these things for real so they're no longer needed!

## Why?
Now that we're working on fronts we want a sound base to start from and capturing variations of a `Card` is one area where we've not been strong. Hopefully this, coupled with adding stories for the [various container types/palettes](https://github.com/guardian/dotcom-rendering/issues/4606), will provide suitable visual regression coverage so that we can be confident making changes in this area of the codebase in future.

This also helps surface what actually happens when we render a card using different combinations of the `Format` property, helping inform designers and others about what possibilities exist

### Why are there changes to some types packages?
Introducing `storiesOf` triggered typescript errors that could [only be resolved by upgrading](https://github.com/storybookjs/storybook/issues/16925#issuecomment-987296116) `@types/react` and `@types/react-dom` (which then triggered some other changes)

### Why is this PR re-introducing the `storiesOf` api?
This api is still valid - it's not deprecated - but it is marked as legacy. However this is the [only way to programmatically add stories](https://github.com/storybookjs/storybook/issues/9828#issuecomment-585101678) and this is a feature we needed here to be able to loop over the properties for `Format`.


dotcom-rendering/src/web/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,382 @@
...
+	ArticleDisplay,
+	ArticlePillar,
+	ArticleSpecial,
+} from '@guardian/libs';
+import { breakpoints, from } from '@guardian/source-foundations';
+import { storiesOf } from '@storybook/react';
+import React from 'react';
+import { Card } from './Card';
+
+const basicCardProps = {
```
mxdvl, line 14: If you exported the `Props` type from `Card.tsx`, you could use set this type to it, and avoid the `SmallHeadlineSize` and `ImagePositionType` assertions

dotcom-rendering/src/web/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,382 @@
...
+	design,
+	title,
+}: {
+	display: ArticleDisplay;
+	design: ArticleDesign;
+	title: string;
+}) => {
+	const cards = [];
+	for (const [pillarName, pillarValue] of Object.entries(ArticlePillar)) {
+		if (Number.isNaN(Number(pillarName))) {
```
mxdvl, line 78: Would be nice to have a comment to explain that this is a [workaround the TS enum gotcha][ts], where the object is basically mirrored between keys and values:

```ts
console.log(Object.entries(ArticlePillar));
/*
[
    ["0", "News"],
    ["1", "Opinion"],
    ["2", "Sport"],
    ["3", "Culture"],
    ["4", "Lifestyle"],
    ["News", 0],
    ["Opinion", 1],
    ["Sport", 2],
    ["Culture", 3],
    ["Lifestyle",4]
] 
*/
```

[ts]: https://www.typescriptlang.org/play?#code/KYOwrgtgBAggTgFwJYGMA2wAKS1oIZxQDeAsAFBSVQBywA7gM5QC8UADADTlVQDyADkhBIA9iBZQAjFwpUAyvxGIJAJhk8AwmDQIwcYBIDM6qgBkkAM2AMEATwwSALDIC+5cijEMRGAHRoRAHMACl4AIwArYBQEX1AEOCRrYPhkdCwcfDgASmyAbnIAegAqcgBtbioygCI2ao4oatpGaoBdE0oayXrGgSFREDaOqBqVHuqFJQQhys7qw3GtHT1gGdk5x3HzKxt7VfbZkab6Bh62A-WjvuExHskLnhrJxB6VB6rqpd19HsN3ue21jsGHqjla5FaUHIxUKQA

dotcom-rendering/src/web/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,382 @@
...
+			cards.push({
+				title: pillarName,
+				format: {
+					display,
+					design,
+					theme: pillarValue as ArticleTheme,
```
mxdvl, line 84: Did you mean `ArticlePillar`?

dotcom-rendering/src/web/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,382 @@
...
+			<CardWrapper>
+				<Card
+					{...basicCardProps}
+					format={{
+						display: ArticleDisplay.Standard,
+						design: ArticleDesign.Media,
+						theme: ArticlePillar.Sport,
+					}}
+					mediaType="Gallery"
+					mediaDuration={360}
```
mxdvl, line 262: Does a gallery have a media duration?

mxdvl (approved):
This is a great refactor.

Some comments, no blockers.

I was a little confused by the use of `storiesOf` API, but I understand that otherwise we would have a lot more individual stories to write.

dotcom-rendering/src/web/components/Card/Card.stories.tsx
oliverlloyd: Sensible. Done

dotcom-rendering/src/web/components/Card/Card.stories.tsx
oliverlloyd: 👍 done - thanks for the playground link!

dotcom-rendering/src/web/components/Card/Card.stories.tsx
oliverlloyd: Good point!

dotcom-rendering/src/web/components/Card/Card.stories.tsx
oliverlloyd: Good catch, thanks

dotcom-rendering/src/web/components/Card/Card.stories.tsx
mxdvl: We could possibly force it to be `never` if the media type is `Gallery`?

oliverlloyd: Probably not worth the complexity if we're [getting rid of this soon](https://github.com/guardian/libs/pull/332) anyway?

PR #4652, opened 2022-04-18
Reveal Onwards

oliverlloyd:
## What does this change?
This PR is an attempt to improve the CLS issues we were seeing with the end of article content that gets dynamically loaded in in the browser using javascript. In this case we're looking at `OnwardsUpper` but this approach could and should be used for other similar sections.

To improve reader experience, we're stealing the reveal animation currently only used for new content in liveblogs to gently announce the new content rather than suddenly showing it

| Before      | After      |
|-------------|------------|
| before | after |


dotcom-rendering/src/web/components/OnwardsData.tsx
```diff
@@ -18,6 +21,24 @@ type OnwardsResponse = {
...
+	.reveal {
+		animation: ${keyframes`
+			0% { opacity: 0; }
+			100% { opacity: 1; }
+		`} 4s ease-out;
+	}
+	.pending {
+		display: none;
+	}
+`;
```
oliverlloyd, line 40: This is currently duplicated here but once this (and other PRs) are merged then we will have a use case for setting this css higher up, perhaps at the page level.

dotcom-rendering/src/web/components/Island.tsx
```diff
@@ -32,7 +32,14 @@ const decideChildren = (
 	placeholderHeight?: number,
 ) => {
 	if (!clientOnly) return children; // Server side rendering
-	if (placeholderHeight) return <Placeholder height={placeholderHeight} />; // Portal using placeholder
+	if (placeholderHeight)
+		return (
+			<Placeholder
+				height={placeholderHeight}
+				shouldShimmer={false}
+				backgroundColor="white"
```
oliverlloyd, line 40: Turn off the grey shimmer for all Islands

dotcom-rendering/src/web/components/OnwardsData.tsx
```diff
@@ -35,18 +56,51 @@ export const OnwardsData = ({
...
+			<Placeholder
+				// 300 is a best guess
+				height={300}
+				shouldShimmer={false}
+				backgroundColor="white"
 			/>
 		);
 	}
 
+	if (data && data.trails) {
```
mxdvl, line 88:
```suggestion
	if (data?.trails) {
```

dotcom-rendering/src/web/components/Island.tsx
mxdvl: Why are we turning this off?

dotcom-rendering/src/web/components/OnwardsData.tsx
```diff
@@ -35,18 +56,51 @@ export const OnwardsData = ({
...
-			<Container
-				heading={data.heading || data.displayname} // Sometimes the api returns heading as 'displayName'
-				trails={buildTrails(data.trails, limit)}
-				description={data.description}
-				ophanComponentName={ophanComponentName}
-				format={format}
-				isCuratedContent={isCuratedContent}
+			<Placeholder
+				// 300 is a best guess
+				height={300}
```
mxdvl, line 81: Looking at a few pages, we tend to be closer to `600`.

Even with a maximum of three lines for cards in on “More on this story” and four for “More from Headlines”, you get to `540`.

mxdvl (commented):
I get point of setting a `min-height` and love the fact that we’ve introduced a placeholder for this kind of content, but I’m not sure why we’ve changed the shimmer look and added a fade-in animation.

Could these two things happen in distinct PRs?

mxdvl:
Also, this `OnwardsUpper` island is the chunkiest, by nearly a 50% margin! Maybe we could tame that a bit. Will create an issue.

dotcom-rendering/src/web/components/Island.tsx
oliverlloyd: Because there are two loading states, the server side loader and then the loader that gets rendered by the client component, and the transition between them can be janky.

I discussed the approach of having a plain white background for both the server and client loader with @HarryFischer . He's happy to start with this and then we can look at building on top of this. For example, maybe we can render white on the server and only have a grey loader on the client?

Note. I think this should be 'transparent' not 'white', I'll fix that.

dotcom-rendering/src/web/components/OnwardsData.tsx
oliverlloyd: There are two sets of onwards cards so it's 300 + 300

OllysCoding:
I'm definitely a big fan of having the minimum reserved height, but I'm not so sure about the animation? Especially since 4 seconds is quite a long fade in time!

oliverlloyd:
> I'm definitely a big fan of having the minimum reserved height, but I'm not so sure about the animation? Especially since 4 seconds is quite a long fade in time!

I think the time for this animation could be reviewed outside of this PR? There's cases for and against this time and it's come up as a discussion a few times. Maybe we can have a ticket to review it and make a decision?

As for if there should be _any_ animation. That sounds like a design question. But I think something is better than nothing, simply not doing anything and having content appear in a flash is pretty jarring in my opinion

oliverlloyd:
> As for if there should be _any_ animation. That sounds like a design question. But I think something is better than nothing, simply not doing anything and having content appear in a flash is pretty jarring in my opinion

Note. I have already discussed this with @HarryFischer before creating these PRs. We agreed that the best approach was to go forward with what we have now and review and rethink how we do loading site wide.

mxdvl (approved):
If this has been discussed previously with designers, I’m happy to go for the 4s animation.

oliverlloyd:
> 4 seconds is quite a long fade in time!

I've reduced this to 2 seconds now and we will review it once we have everything in place

PR #4656, opened 2022-04-19
Fix twitter handle colour in liveblogs below desktop

### LINKED ISSUE #4482: twitter handle needs colour logic
In this scenario (mobile) is should be white, then grey at wider breakpoints when we introduce the leftColumn

image

ioannakok:
Closes #4482 

## What does this change?
Adds colour logic for twitter handle. This fixes the colour in liveblogs for breakpoints below desktop (980px).

## Why?
To match the liveblog designs.

## Screenshots

| Before      | After      |
|-------------|------------|
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |

### DIFF
--- a/dotcom-rendering/src/web/components/Contributor.tsx
+++ b/dotcom-rendering/src/web/components/Contributor.tsx
@@ -6,6 +6,7 @@ import {
 	headline,
 	textSans,
 	until,
+	from,
 } from '@guardian/source-foundations';
 
 import { BylineLink } from './BylineLink';
@@ -15,25 +16,59 @@ import TwitterIcon from '../../static/icons/twitter.svg';
 import { interactiveLegacyClasses } from '../layouts/lib/interactiveLegacyStyling';
 import { decidePalette } from '../lib/decidePalette';
 
-const twitterHandleStyles = (palette: Palette) => css`
-	${textSans.xxsmall()};
-	font-weight: bold;
+const defaultTwitterHandleColour = (palette: Palette) => css`
 	color: ${palette.text.twitterHandle};
 
-	padding-right: 10px;
-	display: inline-block;
-
 	svg {
-		height: 10px;
-		max-width: 12px;
-		margin-right: 0px;
 		fill: ${neutral[46]};
 	}
 
 	a {
 		color: ${palette.text.twitterHandle};
+	}
+`;
+
+const twitterHandleColour = (palette: Palette, format: ArticleFormat) => {
+	switch (format.design) {
+		case ArticleDesign.LiveBlog: {
+			return css`
+				color: ${palette.text.twitterHandleLiveBackground};
+
+				svg {
+					fill: ${neutral[100]};
+				}
+
+				a {
+					color: ${palette.text.twitterHandleLiveBackground};
+				}
+
+				${from.desktop} {
+					${defaultTwitterHandleColour(palette)}
+				}
+			`;
+		}
+		default: {
+			return defaultTwitterHandleColour(palette);
+		}
+	}
+};
+
+const twitterHandleStyles = css`
+	${textSans.xxsmall()};
+	font-weight: bold;
+
+	svg {
+		height: 10px;
+		max-width: 12px;
+		margin-right: 0px;
+	}
+
+	a {
 		text-decoration: none;
 	}
+
+	padding-right: 10px;
+	display: inline-block;
 `;
 
 // for liveblog smaller breakpoints article meta is located in the same
@@ -120,7 +155,12 @@ export const Contributor: React.FC<{
 				</div>
 			)}
 			{onlyOneContributor && author.twitterHandle && (
-				<div css={twitterHandleStyles(palette)}>
+				<div
+					css={[
+						twitterHandleStyles,
+						twitterHandleColour(palette, format),
+					]}
+				>
 					<TwitterIcon />
 					<a
 						href={`https://www.twitter.com/${author.twitterHandle}`}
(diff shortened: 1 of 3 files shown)


HarryFischer (approved):
👍

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -1055,6 +1059,7 @@ export const decidePalette = (format: ArticleFormat): Palette => {
 			matchTitle: textMatchTitle(),
 			byline: textByline(format),
 			twitterHandle: textTwitterHandle(format),
+			twitterHandleLiveBackground: textTwitterHandleLiveBackground(),
```
oliverlloyd, line 1062: Can we 
1) drop 'background' from the name here? It's already implied by the structure of the object
2) drop 'live' as this is implied by the usage of `format`

oliverlloyd: For this to work you will need to pass in `format`

dotcom-rendering/src/web/components/Contributor.tsx
```diff
@@ -15,25 +16,59 @@ import TwitterIcon from '../../static/icons/twitter.svg';
...
+		fill: ${neutral[46]};
+	}
+
+	a {
+		color: ${palette.text.twitterHandle};
+	}
+`;
+
+const twitterHandleColour = (palette: Palette, format: ArticleFormat) => {
+	switch (format.design) {
```
oliverlloyd, line 32: I think - if I'm reading this right - this switch needs to happen in `decidePalette`?

dotcom-rendering/src/web/components/Contributor.tsx
```diff
@@ -15,25 +16,59 @@ import TwitterIcon from '../../static/icons/twitter.svg';
 import { interactiveLegacyClasses } from '../layouts/lib/interactiveLegacyStyling';
 import { decidePalette } from '../lib/decidePalette';
 
-const twitterHandleStyles = (palette: Palette) => css`
-	${textSans.xxsmall()};
-	font-weight: bold;
+const defaultTwitterHandleColour = (palette: Palette) => css`
```
oliverlloyd, line 19: This is subject so take it or leave it but I find nesting css variables can be confusing. I am a fan of keeping things simple and having all the css code used by a variable declaration contained inside itself. And then you put any logic in `<div css={myLogic && simpleCss}>`

dotcom-rendering/src/web/lib/decidePalette.ts
oliverlloyd: Ahh, I think I see what you are trying to do here. You want to vary the colour based something not contained in `format` like context or a breakpoint?

ioannakok: Exactly! I want to vary the colour based on the breakpoint. Do you think `twitterHandleBelowDesktop` would make more sense? And maybe something like:

```suggestion
                         // Call it like that and pass format
			twitterHandleBelowDesktop: textTwitterHandleBelowDesktop(format),
			
			// And in palette 
			const textTwitterHandleBelowDesktop = (format: ArticleFormat): string => {
			        if(format.design === ArticleFormat.Liveblog)
			               return WHITE;
			         
			         return textTwitterHandle(format);      	
                         };
```

What do you think?

ioannakok: I pushed a 2nd version of the code. Let me know what you think and if it makes sense!

dotcom-rendering/src/web/components/Contributor.tsx
```diff
@@ -15,25 +16,46 @@ import TwitterIcon from '../../static/icons/twitter.svg';
 import { interactiveLegacyClasses } from '../layouts/lib/interactiveLegacyStyling';
 import { decidePalette } from '../lib/decidePalette';
 
-const twitterHandleStyles = (palette: Palette) => css`
+const twitterHandleColour = (palette: Palette) => css`
+	color: ${palette.text.twitterHandleBelowDesktop};
+
```
oliverlloyd, line 21: Just an observation (and this is _very_ subjective), maybe the different groups of css properties can be grouped together without spaces in-between? Personally I would find it easier to distinguish them like that but hey, we all feel space differently!

dotcom-rendering/src/web/components/Contributor.tsx
```diff
@@ -15,25 +16,46 @@ import TwitterIcon from '../../static/icons/twitter.svg';
...
+		svg {
+			fill: ${neutral[46]};
+		}
+
+		a {
+			color: ${palette.text.twitterHandle};
+		}
+	}
+`;
```
oliverlloyd, line 41: 🎯 this makes total sense!

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -157,6 +157,12 @@ const textTwitterHandle = (format: ArticleFormat): string => {
 	return text.supporting;
 };
 
+const textTwitterHandleBelowDesktop = (format: ArticleFormat): string => {
+	if (format.design === ArticleDesign.LiveBlog) return WHITE;
+
+	return textTwitterHandle(format);
+};
```
oliverlloyd, line 164: 💯 This is spot on

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -1126,6 +1138,7 @@ export const decidePalette = (format: ArticleFormat): Palette => {
 			richLink: fillRichLink(format),
 			quoteIcon: fillQuoteIcon(format),
 			blockquoteIcon: fillBlockquoteIcon(format),
+			twitterHandleBelowDesktop: fillTwitterHandleBelowDesktop(format),
```
oliverlloyd, line 1141: There's no `fill.twitterHandle` here, is that because the designs don't need it?

oliverlloyd (approved):
This is spot on! 

This is defo a tricky area, `decidePalette` handles most use cases well but occasionally the design requirements are more complex but this feels like a great way to handle that.

Thanks for making the extra effort on this 🙏

PR #4663, opened 2022-04-20
Refactor how we link to pinned posts

oliverlloyd:
## What does this change?
This PR changes how we link to pinned posts such that

- The pinned post element has it's own id `pinned-post` so that we can scroll to the _container_, not the contents
- We now present a different link when readers hover the pinned post icon. This new link will take them to the `pinned-post` container
- We no longer set a block id on posts contained inside the pinned post wrapper. This was causing two blocks on the page to have the same id, which is invalid

## Why?
Permalinks to liveblogs make up 8% of traffic to these pages. EIGHT. So we should care about the reader experience when using them.

Because of this effort to improve this user flow, we [recently made a change](https://github.com/guardian/dotcom-rendering/pull/4643) to how we decide where to place the `LiveBlogEpic` 'Support The Guardian' component which meant that we factor in if the reader is using a permalink or not. However, @Amouzle and Max Walker pointed out that there was a bug in this approach when the permalink was pointing at the pinned post. Because the html structure is different for these posts we were inserting the epic _inside_ the pinned post wrapper, rather than outside it.

A simple solution to get around this bug would have been to ignore permalinks to pinned posts. They are always at the top of the blog so we can be certain that they won't be affected by the epic being created above them (which was the issue the previous PR was solving). But upon closer inspection I saw that the _real_ problem here is that there were two elements on the page with the same id, the `article` inside the pinned post container and the same `article` in the body of the blog.

By removing that second id and then refactoring how we link between the two I was able to solve the bug raised by Alex and also improve the reader experience a bit (permalinks around pinned posts make more sense now).

### The bug raised by Alex looked like this
Screenshot 2022-04-20 at 07 13 40

### DIFF
--- a/dotcom-rendering/src/web/components/PinnedPost.tsx
+++ b/dotcom-rendering/src/web/components/PinnedPost.tsx
@@ -140,6 +140,7 @@ export const PinnedPost = ({ pinnedPost, children, format }: Props) => {
 	const palette = decidePalette(format);
 	return (
 		<div
+			id="pinned-post"
 			css={pinnedPostContainer(palette)}
 			data-gu-marker="pinned-post"
 			data-component="pinned-post"
--- a/apps-rendering/src/components/liveBlocks.tsx
+++ b/apps-rendering/src/components/liveBlocks.tsx
@@ -37,7 +37,6 @@ const LiveBlocks: FC<LiveBlocksProps> = ({ blocks, format }) => {
 		<>
 			{/* Accordion? */}
 			{blocks.map((block) => {
-				const blockLink = `?page=with:block-${block.id}#block-${block.id}`;
 				const blockFirstPublished = pipe(
 					block.firstPublished,
 					map(Number),
@@ -51,7 +50,7 @@ const LiveBlocks: FC<LiveBlocksProps> = ({ blocks, format }) => {
 						format={format}
 						blockTitle={block.title}
 						blockFirstPublished={blockFirstPublished}
-						blockLink={blockLink}
+						blockId={block.id}
 						// TODO pass this value in when available
 						isPinnedPost={false}
 						supportsDarkMode={true}
--- a/common-rendering/src/components/FirstPublished.stories.tsx
+++ b/common-rendering/src/components/FirstPublished.stories.tsx
@@ -11,7 +11,7 @@ const Default: FC = () => (
 	<FirstPublished
 		supportsDarkMode={true}
 		firstPublished={1613763003000}
-		blockLink={'#block-60300f5f8f08ad21ea60071e'}
+		blockId="#block-60300f5f8f08ad21ea60071e"
 		isPinnedPost={false}
 		isOriginalPinnedPost={false}
 		format={{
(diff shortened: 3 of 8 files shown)


dotcom-rendering/src/web/components/PinnedPost.tsx
```diff
@@ -140,6 +140,7 @@ export const PinnedPost = ({ pinnedPost, children, format }: Props) => {
 	const palette = decidePalette(format);
 	return (
 		<div
+			id="pinned-post"
```
alinaboghiu, line 143: Do you know if the sharing links (Facebook and Twitter) still work ok? They take you to the pinned post at the top at the moment though one could argue they should take you to the original post

alinaboghiu (approved):

OllysCoding:
One thing I would note about this change is that it does introduce a change in behaviour:

If I've shared the link to the _pinned post_ on twitter, email, wherever; and then the pinned post changes, a user clicking the link would:
_Previously_: Be taken to the exact block I originally shared
_After this change_: The new pinned post block, which would have different content than what I shared.

---

I'm not sure if this is would be considered a bug or not, but definitely worth taking note of this change in behaviour.

alinaboghiu:
@OllysCoding >Previously: Be taken to the exact block I originally shared
I think if you use the share buttons in the pinned post you get taken to the pinned post at the top of the blog. But I agree this isn't ideal since it changes and that link will not take you to a specific post once it does. Think it's fine if we wither keep the existing behaviour or always link to the original post (that may have to be a Frontend change - having a look)

alinaboghiu:
Actually ignore me I was just looking at the frontend version.

oliverlloyd:
@OllysCoding @alinaboghiu useful comments! Thanks both. I actually spotted a bug with my implementation when checking your queries which I've now fixed

So, firstly, any existing link using a `#block-123` format will continue to work. It will now always take you to the original post, never the pinned post. Before, we had two elements on the page with the same id so it was a roulette where you ended up! A link to a post should be canonical and never change, even if a block is unpinned.

The other thing I changed was to add a _new_ link from the original post that _navigates_ you to the pinned post _container_. Not the contents

Existing permalinks:
https://www.theguardian.com/world/live/2022/apr/20/russia-ukraine-war-latest-ukraine-repels-numerous-donbas-attacks-as-russian-logistics-falter-says-uk-live?page=with:block-625fd2788f08931717fb4910#block-625fd2788f08931717fb4910

The _new pinned post container link
https://www.theguardian.com/world/live/2022/apr/20/russia-ukraine-war-latest-ukraine-repels-numerous-donbas-attacks-as-russian-logistics-falter-says-uk-live#pinned-post

But this new link is *only* on the little svg pin thing that shows on the original block
Screenshot 2022-04-20 at 10 54 18
The time links to that block, same as always, the pin _now_ links to the pinned post container

You could make this pinned container link more prominent if you wanted but readers can also navigate to the first page which has a similar effect

alinaboghiu:
This is great and very sensible. While you're here, would it be easy to make the pinned post timestamp link to the original post? If not then please ignore!

oliverlloyd:
> This is great and very sensible. While you're here, would it be easy to make the pinned post timestamp link to the original post? If not then please ignore! We have had a ticket about this for a while https://trello.com/c/kcSRX24W/212-bug-clicking-on-the-timestamp-of-the-pinned-post-does-not-take-you-to-the-original-post-position

Very easy. This bug should already be fixed by this PR. It wasn't working before because there were two elements with the same id but now the original post is the only one so that is where the link will take readers.

PR #4669, opened 2022-04-20
Reveal `Discussion`

oliverlloyd:
## What does this change?
This PR continues [the work](https://github.com/guardian/dotcom-rendering/pull/4652) to improve CLS when scrolling to the bottom of content.

This PR is using a beta version of discussion and is dependent on https://github.com/guardian/discussion-rendering/pull/568 being released before we can merge this PR.

## Why?
Better CLS improves the reader experience

| Before      | After      |
|-------------|------------|
| comments before | comments after |


dotcom-rendering/src/web/components/Discussion.tsx
```diff
@@ -91,44 +132,83 @@ export const Discussion = ({
...
+		);
+		pendingElements?.forEach((element) => {
+			element.classList.add('reveal');
+			element.classList.remove('pending');
+		});
+	}, []);
+
+	useEffect(() => {
+		// There's no point showing the view more button if there isn't much more to view
+		if (commentCount === 0 || commentCount === 1 || commentCount === 2) {
```
mxdvl, line 147: Is this not simply checking that there are less than 3 comments?

```suggestion
		if (commentCount < 2) {
```

mxdvl (approved):
Happy with this, but obviously it would be better if `discussion-rendering` was not in a pre-release stage.

dotcom-rendering/src/web/components/Discussion.tsx
oliverlloyd: But seriously, I prefer being plain and explicit. What if there's an edge case if `commentCount` is zero or something like that? I find the original code instantly understandable and readable, even if it's a bit verbose.

mxdvl: Okay—just curious as I had not seen this pattern before.

I should have written `< 3` or `<= 2`, but I'm fine with the three conditions if intentional!

PR #4671, opened 2022-04-20
reduce mobile immersive header to 80vh

HarryFischer:
## What does this change?
Reduces height of immersive header until desktop - was 100vh now 80vh
## Why?
@ajbreuer @Bryony-Szekeres request
## Screenshots

| Before      | After      |
|-------------|------------|
|  image |  image |
|  image |  image |

### DIFF
--- a/dotcom-rendering/src/web/components/ImageComponent.tsx
+++ b/dotcom-rendering/src/web/components/ImageComponent.tsx
@@ -259,12 +259,12 @@ export const ImageComponent = ({
 					width: 100%;
 
 					${from.desktop} {
-		height: 100vh;
-		min-height: 31.25rem;
-	}
-	${from.wide} {
-		min-height: 50rem;
-	}
+						height: 100vh;
+						min-height: 31.25rem;
+					}
+					${from.wide} {
+						min-height: 50rem;
+					}
 					img {
 						object-fit: cover;
 					}
--- a/dotcom-rendering/src/web/components/ImageComponent.tsx
+++ b/dotcom-rendering/src/web/components/ImageComponent.tsx
@@ -259,8 +259,12 @@ export const ImageComponent = ({
 					width: 100%;
 
 					${from.desktop} {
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/ImageComponent.tsx
```diff
@@ -273,7 +276,6 @@ export const ImageComponent = ({
 					isLazy={!isMainMedia}
 					isMainMedia={isMainMedia}
 				/>
-				{starRating && <PositionStarRating rating={starRating} />}
```
HarryFischer, line 276: Note: removed review stars from immersive layout. @oliverlloyd has a PR to add them back into the correct place

jamesgorrie: Do you have the link to the PR?

oliverlloyd: I actually ended up putting this work aside. It's not complicated but we don't have a use case for showing stars on immersive pieces at this point so there's no urgency to this

dotcom-rendering/src/web/layouts/headers/ImmersiveHeader.tsx
```diff
@@ -27,13 +27,14 @@ import { Island } from '../../components/Island';
 import { decidePalette } from '../../lib/decidePalette';
 
 const hasMainMediaStyles = css`
-	height: 100vh;
+	height: 80vh;
 	/**
     100vw is normally enough but don't let the content shrink vertically too
```
oliverlloyd, line 32:
```suggestion
    80vw is normally enough but don't let the content shrink vertically too
```

oliverlloyd (approved):
So when are you going to become a full time developer?

oliverlloyd:
I made a small change to increase the number of breakpoints we capture this story at so we have regression for how this code lands at different widths

PR #4681, opened 2022-04-20
Fix ad label max-height

arelra:
## What does this change?

We have seen issues where the ad label has its width and height set to that of its related advert.

Until we track down the cause of this we have decided to add a `max-height` to the ad label.

### Before

Screenshot 2022-04-20 at 16 53 03


chrislomaxjones (approved):

mxdvl:
Could it be a stacking context issue?

PR #4685, opened 2022-04-20
Pass ad slot props down into `<TopRightAdSlot />`

chrislomaxjones:
## What does this change?

Pass the props `shouldHideReaderRevenue` and `isPaidContent` down from the ad slot into the top right ad slot component.

## Why?

These props were previously being ignored, with the values passed hardcoded as `false`. These values should be passed in and used to decide when to render a shady pie.


arelra (approved):
Great spot!

PR #4688, opened 2022-04-21
Apply 300px `max-width` to ad slot containers

simonbyford:
## What does this change?

Applies a `max-width` of 300px to ad slot containers, instead of the ad slots themselves.

We use ad slot containers to wrap `inline2+` ads in an attempt to stop Google performing [ad expansion](https://support.google.com/admanager/answer/9117822?hl=en). 

Ad slot containers are introduced in this PR: https://github.com/guardian/frontend/pull/24917

## Why?

Our [previous attempt at addressing this problem](https://github.com/guardian/dotcom-rendering/pull/4441), which involved applying the `max-size` to the ad slot itself, didn't work. When we asked Google about this, they recommended applying the `max-size` to the parent element instead.

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleContainer.tsx
+++ b/dotcom-rendering/src/web/components/ArticleContainer.tsx
@@ -86,7 +86,7 @@ const articleAdStyles = css`
 			margin-left: 20px;
 		}
 	}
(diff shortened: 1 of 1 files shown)


arelra (approved):
👍

Issue #4692, opened 2022-04-21
Fronts Ophan tag parity (`data-link-name`)


mxdvl:
## Remaining Front components mismatches

Comparing **`data-link-name`** on the international Front between [Frontend][] & [DCR][].

[Frontend]: https://www.theguardian.com/international?dcr=false
[DCR]: https://www.theguardian.com/international?dcr=true

### Missing from DCR (26 / 601)

- [X] **`nav3 : topbar : my account`** &rarr; `<INPUT/>`
- [X] **`nav3 : topbar : account overview`** &rarr; `<A/>`
- [X] **`nav3 : topbar : billing`** &rarr; `<A/>`
- [X] **`nav3 : topbar : profile`** &rarr; `<A/>`
- [X] **`nav3 : topbar : email prefs`** &rarr; `<A/>`
- [X] **`nav3 : topbar : data privacy`** &rarr; `<A/>`
- [X] **`nav3 : topbar : settings`** &rarr; `<A/>`
- [X] **`nav3 : topbar : help`** &rarr; `<A/>`
- [X] **`nav3 : topbar : comment activity`** &rarr; `<A/>`
- [X] **`nav3 : topbar : sign out`** &rarr; `<A/>`
- [X] **`nav3 : topbar : edition-picker: toggle`** &rarr; `<INPUT/>`
- [X] **`nav3 : topbar : edition-picker: toggle`** &rarr; `<LABEL/>`
- [X] **`nav2 : overlay`** &rarr; `<LABEL/>`
- [X] **`nav3 : edition picker`** &rarr; `<BUTTON/>`
- [ ] **`nav2 : subnav-toggle`** &rarr; `<BUTTON/>`
- [X] **`container-0 | palette-styles-new-do-not-delete`** &rarr; `<SECTION/>`
- [ ] **`Labs front container | US | international | container-13 | What's possible? Ask Toronto | University of Toronto | card-1 | How nanotechnology could improve cancer and fertility treatment`** &rarr; `<A/>`
- [ ] **`Labs front container | US | international | container-13 | What's possible? Ask Toronto | University of Toronto | card-2 | How researchers remade ‘the world’s most widely used petrochemical’`** &rarr; `<A/>`
- [ ] **`Labs front container | US | international | container-13 | What's possible? Ask Toronto | University of Toronto | card-3 | Why researchers are adding nutrients to salt and tea`** &rarr; `<A/>`
[...]

PR #4664, opened 2022-04-20
docs(ophan): add Ophan tracking information and logs

mxdvl:
## What does this change?

Adds logs when the Ophan tracker `record` function is called. Improve types based on an analysis of the [`transmit.coffee` source][transmit.coffee] to indicate the presence of this callback.

Adds JSDocs to the `data-component` and `data-link-name` attributes in JSX to gain Intellisense info. Based on [**Tracking components in the Data Lake**] from `frontend`.

[transmit.coffee]: https://github.com/guardian/ophan/blob/ccaa57bcee3f5f3f83ec28973074c9b3f98f1438/tracker-js/assets/coffee/ophan/transmit.coffee#L43-L50
[**Tracking components in the Data Lake**]: https://github.com/guardian/frontend/blob/5b970cd7308175cfc1bcae2d4fb8c06ee13c5fa0/docs/03-dev-howtos/19-tracking-components-in-the-data-lake.md#tracking-components-in-the-data-lake

## Why?

@jlieb10 and I have been discovering a lot of things, and we want to share these findings to make them as visible as possible for all. Part of #4662 

## Intellisense for attributes

| data-link-name      | data-component      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/164194321-03e5c0a4-db1a-424b-aabf-4398043f78b4.png
[after]: https://user-images.githubusercontent.com/76776/164194460-f7313381-547e-49f5-a5ae-01cd4286bd89.png

PR #4693, opened 2022-04-21
Ophan Components parity tracking

mxdvl:
## What does this change?

Introduce a workflow that runs a script every work day morning to check the parity of [Ophan Components][].

The script is written in Deno / Typescript. It checks the [`/international`][] Front on both rendering platforms, and compares the presence of `data-link-name` and `data-component` attributes. The action uses the `GITHUB_TOKEN` and Octokit to update issues with the latest comparison.

As of this writing, there are only 74 / 302 of  `data-link-name`, and 1 / 30 `data-component` parity in DCR.

A neat feature of issues is that updating with identical text will not register as a new edit. This means that we can also track the progress over time without too much noise. I’ve tested this by running the script locally several times–using a Personal Access Token to gain issue editing rights.

## Why?

Parity on Fronts between `frontend` and `dotcom-rendering` is an essential part of launching fronts. It’s easy to forget invisible elements, so running this script regularly ensures there’s visibility on these components.

For the parity progress, see:
- #4692
- #4698

For docs on [Ophan Components][]
- #4664 

[`/international`]: https://www.theguardian.com/international
[Ophan Components]: https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/docs/development/track-user-behaviour.md

## Screenshots

It’s possible to enable [a pure CSS viewer](https://gist.github.com/mxdvl/b17f344678ed34d45683896e208ecdc1) to what components are one a page:

image

### DIFF
--- a/.github/workflows/fronts.yml
+++ b/.github/workflows/fronts.yml
@@ -1,0 +1,58 @@
+name: DCR Fronts Ophan tag status
+on:
+  push:
+    branches: [main]
+
+  # Allows you to run this workflow manually from the Actions tab.
+  workflow_dispatch:
+
+jobs:
+  data-link-name:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup deno
+        uses: denolib/setup-deno@v2
+        with:
+          deno-version: v1.21.0
+
+      - name: Compare components "data-link-name"
+        id: text
+        run: |
+          echo "About to run Deno script"
+          TEXT=$(deno run --allow-net ./scripts/deno/ophan-components.ts data-link-name)
+          echo "::set-output text=$TEXT"
+
+      - name: Update Comment
+        uses: peter-evans/create-or-update-comment@v2.0.0
+        with:
+          comment-id: 1211151898
+          body: "{{ steps.text.outputs }}"
+          edit-mode: replace
+
+  data-component:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup deno
+        uses: denolib/setup-deno@v2
+        with:
+          deno-version: v1.21.0
+
+      - name: Compare components "data-component"
+        id: text
+        run: |
+          echo "About to run Deno script"
+          TEXT=$(deno run --allow-net ./scripts/deno/ophan-components.ts data-component)
+          echo "::set-output name=text::$TEXT"
+
+      - name: Update Comment
+        uses: peter-evans/create-or-update-comment@v2.0.0
+        with:
+          comment-id: 1212191944
+          body: "{{ steps.text.outputs }}"
+          edit-mode: replace
--- a/.github/workflows/fronts.yml
+++ b/.github/workflows/fronts.yml
@@ -1,7 +1,11 @@
-name: DCR Fronts Ophan tag status
+name: DCR Fronts–Ophan Components
 on:
   push:
     branches: [main]
+
+  schedule:
+    # Every work day of the week at 08:08
+    - cron: "8 8 * * MON-FRI"
 
   # Allows you to run this workflow manually from the Actions tab.
   workflow_dispatch:
--- a/.github/workflows/fronts.yml
+++ b/.github/workflows/fronts.yml
@@ -1,8 +1,5 @@
 name: DCR Fronts–Ophan Components
 on:
-  push:
-    branches: [main]
-
   schedule:
     # Every work day of the week at 08:08
     - cron: "8 8 * * MON-FRI"
--- a/.github/workflows/fronts.yml
+++ b/.github/workflows/fronts.yml
@@ -19,18 +19,7 @@ jobs:
           deno-version: v1.21.0
 
       - name: Compare components "data-link-name"
-        id: text
-        run: |
-          echo "About to run Deno script"
-          TEXT=$(deno run --allow-net ./scripts/deno/ophan-components.ts data-link-name)
-          echo "::set-output text=$TEXT"
-
-      - name: Update Comment
-        uses: peter-evans/create-or-update-comment@v2.0.0
-        with:
-          comment-id: 1211151898
-          body: "{{ steps.text.outputs }}"
-          edit-mode: replace
+        run: deno run --allow-env="GITHUB_TOKEN" --no-check=remote --allow-net=www.theguardian.com,api.github.com  scripts/deno/ophan-components.ts data-link-name
 
   data-component:
     runs-on: ubuntu-20.04
@@ -44,15 +33,4 @@ jobs:
           deno-version: v1.21.0
 
       - name: Compare components "data-component"
-        id: text
-        run: |
-          echo "About to run Deno script"
-          TEXT=$(deno run --allow-net ./scripts/deno/ophan-components.ts data-component)
-          echo "::set-output name=text::$TEXT"
-
-      - name: Update Comment
-        uses: peter-evans/create-or-update-comment@v2.0.0
-        with:
-          comment-id: 1212191944
-          body: "{{ steps.text.outputs }}"
-          edit-mode: replace
+        run: run: deno run --allow-env="GITHUB_TOKEN" --no-check=remote --allow-net=www.theguardian.com,api.github.com  scripts/deno/ophan-components.ts data-component
--- a/.github/workflows/fronts.yml
+++ b/.github/workflows/fronts.yml
@@ -20,7 +20,7 @@ jobs:
           deno-version: v1.21.0
 
       - name: Compare components "data-link-name"
-        run: deno run --allow-env="GITHUB_TOKEN" --no-check=remote --allow-net=www.theguardian.com,api.github.com  scripts/deno/ophan-components.ts data-link-name
+        run: scripts/ci-ophan.sh
 
   data-component:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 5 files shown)


.github/workflows/fronts.yml
```diff
@@ -0,0 +1,34 @@
+name: DCR Fronts Ophan tag status
+on:
+  # Triggers the workflow on push and pull request events but only for the main branch
```
jamesgorrie, line 3: What do you mean push _**and pull**_? Is the comment explaining anything that isn't intrinsic to actions?

.github/workflows/fronts.yml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  data-link-name:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Setup deno
+        uses: denolib/setup-deno@v2
+        with:
+          deno-version: v1.21.0
+
+      - name: Run Deno script
```
jamesgorrie, line 19: Could we be more specific, "Compare data link names" or something similar?

.github/workflows/fronts.yml
```diff
@@ -0,0 +1,34 @@
...
+        id: text
```
jamesgorrie, line 20: What's this doing?

.github/workflows/fronts.yml
```diff
@@ -0,0 +1,34 @@
...
+        run: |
+          echo "About to run Deno script"
```
jamesgorrie, line 22: I'm concerned about running scripts outside of this repo. It'd be good to get this in here, I think it's just a matter of copying this script into this PR.

.github/workflows/fronts.yml
```diff
@@ -0,0 +1,34 @@
...
+          TEXT=$(deno run --allow-net https://gist.githubusercontent.com/mxdvl/75007429b33d14ff2838c3cb51b50673/raw/73677246f9ea1652a516e8b16bed4ce842b084d0/data-link-name.ts)
+          echo "::set-output text=$TEXT"
+
+      - name: Create or Update Comment
+        uses: peter-evans/create-or-update-comment@v2.0.0
+        with:
+          comment-id: 557858210
```
jamesgorrie, line 29: I think you might want to use something from this doc:
https://github.com/peter-evans/create-or-update-comment#where-to-find-the-id-of-a-comment

jamesgorrie (commented):
This is really nice idea, just a few comments and questions on the implementation

.github/workflows/fronts.yml
mxdvl: I’m bringing it in!

.github/workflows/fronts.yml
mxdvl: I think these two are right:
- `data-link-name`: https://github.com/guardian/dotcom-rendering/issues/4692#issue-1211151898
- `data-component`: https://github.com/guardian/dotcom-rendering/issues/4698#issue-1212191944

.github/workflows/fronts.yml
mxdvl: My understanding is that it allows us to use the output of the script as a variable. Might be redundant with `"::set-output text=$TEXT`.

.github/workflows/fronts.yml
```diff
@@ -0,0 +1,58 @@
...
+        id: text
+        run: |
+          echo "About to run Deno script"
+          TEXT=$(deno run --allow-net ./scripts/deno/ophan-components.ts data-link-name)
+          echo "::set-output text=$TEXT"
+
+      - name: Update Comment
+        uses: peter-evans/create-or-update-comment@v2.0.0
+        with:
+          comment-id: 1211151898
```
jamesgorrie, line 31: You'll need this:
https://github.com/peter-evans/create-or-update-comment#where-to-find-the-id-of-a-comment

.github/workflows/fronts.yml
```diff
@@ -0,0 +1,58 @@
...
+        id: text
+        run: |
+          echo "About to run Deno script"
+          TEXT=$(deno run --allow-net ./scripts/deno/ophan-components.ts data-component)
+          echo "::set-output name=text::$TEXT"
+
+      - name: Update Comment
+        uses: peter-evans/create-or-update-comment@v2.0.0
+        with:
+          comment-id: 1212191944
```
jamesgorrie, line 56: You'll need this:
https://github.com/peter-evans/create-or-update-comment#where-to-find-the-id-of-a-comment

.github/workflows/fronts.yml
OllysCoding: I think the idea is that this particular comment (which is actually an issue) acts as a source of truth, rather than it adding a comment to each PR?

.github/workflows/fronts.yml
mxdvl: Actually no, but I’ve opted for doing the updating in Deno directly with Octokit.

.github/workflows/fronts.yml
mxdvl: I’ve bailed out of the action and just update via the script itself. Simpler.

.github/workflows/fronts.yml
mxdvl: What I want is actually updating the issue directly! Got it working locally with a personal token. Should work nicely within github

.github/workflows/fronts.yml
```diff
@@ -0,0 +1,40 @@
+name: DCR Fronts–Ophan Components
+on:
+  push:
+    branches: [main]
+
+  schedule:
+    # Every work day of the week at 08:08
+    - cron: "8 8 * * MON-FRI"
```
OllysCoding, line 8: So this is on every push to main and once a day on the work week?

.github/workflows/fronts.yml
```diff
@@ -0,0 +1,40 @@
...
+
+  # Allows you to run this workflow manually from the Actions tab.
+  workflow_dispatch:
```
OllysCoding, line 11: Should the colon be removed? e.g just `workflow_dispatch`?

scripts/deno/ophan-components.ts
```diff
@@ -0,0 +1,165 @@
+import { Octokit } from "https://cdn.skypack.dev/octokit?dts";
+
+const attributes = ["data-link-name", "data-component"] as const;
+type OphanAttribute = typeof attributes[number];
```
OllysCoding, line 4: why is it `attributes[*number*]`, I'm finding this part quite hard to compute!

mxdvl: Ah yes, this is one of the weirdness of Typescript, but is also very convenient for runtime type-checking.

The way you access an array is `array[0]`, and because we’ve declared that `attributes as const`, Typescript is able to determine that any call in form of `attribute[number]` will yield one of its members, it can create the union type `"data-link-name" | "data-component"`. The benefit of using this syntax is that checking if a `string` is an `OphanAttribute` is as simple as doing `attributes.includes(string)`, which is what the `isOphanAttribute` type predicate does. See [Typescript Playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAhgJwXAnjAvDA2gIjjgGhhwCNDjgcBdAbhgHp6ZoEBLMAcyyoFgAoUJFiJkKAIIQAwuGgZsecqUWUq8CDEHQ6jGAgCmcACbgANmlz4iSqyv78oKAA56YAGRAgILzA+cgAZvBIqFhgAK4AtiR6CLQMTCzsHPZOLgByIiAA7nK+egFBohLSQqGR0bHaTAowAD7EZHUUOHYCMrCsEBlI2XIAFBAAXMxQbJwAlMPqnTDdCL3oAHyFqMXtAHTswCZhhnoQA2qzmVnjNK0WigBe1OsRcI6HSxoyICZ66yYgHH0ABjgAEgA3hAAL44GDAzpzbIDcYwAD8xBmYBOMEAvBuAMT2IcMcDMvp4XOjAKJ7OFBv3G434OhqKLRWOp1RuMHxHi8GOJQA).

```ts
const array = ["a", "b", "c"]; // string[]
const arrayAsConst = ["a", "b", "c"] as const; // readonly ["a", "b", "c"]

type Loose = typeof array[number]; // string
type Narrow = typeof arrayAsConst[number]; // "a" | "b" | "c"

const isNarrow = (s: string): s is Narrow => arrayAsConst.includes(s as Narrow);

["a", "z"].map(s => console.log(`"${s}" ${isNarrow(s) ? "is narrow 🙆" : "is loose 🙅"}`))
// "a" is narrow 🙆
// "z" is loose 🙅
```

.github/workflows/fronts.yml
mxdvl: Whoops, good point, if we’re doing once a week we can remove the pushes!

scripts/deno/ophan-components.ts
OllysCoding: 😰😰😰 Definitely a confusing syntax! It feels like there should be a more readable (atleast for typescript non-experts like me), because I agree that having the ability to do `attributes.includes(string)` is great, but when I'd tried to work backwards to see _why_ you were able to do that, it made little sense to me!

Thanks for the great explanation (I'm still not quite sure I understand, but thats on me!) :)

mxdvl: I have managed to get away from this syntax for the time being, so hopefully this is a problem for another day.

I am aware that this file could benefit from being a little more accessible by favouring simplicity. As this is not in the critical path of DCR rendering and are bound by their own set of TS rules, I feel tempted to indulge in slightly more advanced TS features. Happy to try an refactor for simplicity, but maybe in a subsequent PR?

jlieb10 (approved):
Looks great. Non-blocking question: why are we running it on a schedule instead on PRs?

mxdvl:
> Looks great. Non-blocking question: why are we running it on a schedule instead on PRs?

This is looking at PROD urls, so looking at the page on PRs doesn't make much sense. As a possible future improvement we could have a web hook trigger this run.

.github/workflows/fronts.yml
jamesgorrie: This then = `workflow_dispatch: {}`, where Olly's suggestion is `workflow_dispatch: undefined`.

PR #4694, opened 2022-04-21
Don't render drop cap if the first character of a paragraph is "I"

MarSavar:
## Why?
As per title.

I understand regular expressions are not everyone's cup of tea, and the tradeoff is readability.

If people are really against it, I've come up with a more traditional solution here: https://gist.github.com/MarSavar/164b689fb3f37ebcc983e93c68f3e4e9 but personally I don't love it because I feel like manipulating strings is always a bit of a mystery in JS. I actually had to go and check that slicing with an index bigger than the length of the string doesn't throw an out-of-bounds exception or unintended behaviour.
Also, it's several times slower than the regex. And we are performing this check on every single paragraph in every eligible article, so to me it makes sense to optimise this code. (I have deliberately made sure it is the rightmost logical check in order to skip it altogether if the previous checks fail),

I have improved the JS doc of the regex explaining it in the clearest possible terms so hovering over it gives a pretty understandable (I hope) explanation of what is going on.

image

### DIFF
--- a/apps-rendering/src/renderer.ts
+++ b/apps-rendering/src/renderer.ts
@@ -226,12 +226,12 @@ const plainTextElement = (node: Node, key: number): ReactNode => {
  * This regular expression checks that a string begins with a word that is at least
  * three characters long, ignoring the initial quotation mark.
  *
- * - The regex can be broken down as follows:
+ *  The regex can be broken down as follows:
  *
- * - ["'\u2018\u201c]? matches an optional quotation mark, apostrophe, open single quote
+ * - `["'\u2018\u201c]?` matches an optional quotation mark, apostrophe, open single quote
  * or open double quote.
  *
- * - (?!I) is a negative lookahead checking that the first letter is not "I".
+ * - `(?!I)` is a negative lookahead checking that the first letter is not "I".
  *
  * - The rest of the expression matches 3 or more characters in the Latin-1 Unicode block,
  * which includes diacritics (e.g. å, č, Ë, etc.).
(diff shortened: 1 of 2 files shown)


oliverlloyd (approved):
Superb explanation, I'm sold! 👏 

Do you also need to exclude `A`?

apps-rendering/src/renderer.ts
```diff
@@ -224,16 +224,23 @@ const plainTextElement = (node: Node, key: number): ReactNode => {
 
 /**
  * This regular expression checks that a string begins with a word that is at least
- * three characters long.
- * ["'\u2018\u201c]? matches an optional quotation mark, apostrophe, open single quote
+ * three characters long, ignoring the initial quotation mark.
+ *
+ * - The regex can be broken down as follows:
+ *
+ * - ["'\u2018\u201c]? matches an optional quotation mark, apostrophe, open single quote
```
mxdvl, line 231:
```suggestion
 * The regex can be broken down as follows:
 *
 * - `["'\u2018\u201c]?` matches an optional quotation mark, apostrophe, open single quote
```

apps-rendering/src/renderer.ts
```diff
@@ -224,16 +224,23 @@ const plainTextElement = (node: Node, key: number): ReactNode => {
...
  * or open double quote.
- * The rest of the expression matches any character in the Latin-1 Unicode block,
+ *
+ * - (?!I) is a negative lookahead checking that the first letter is not "I".
```
mxdvl, line 234:
```suggestion
 * - `(?!I)` is a negative lookahead checking that the first letter is not "I".
```

mxdvl (approved):
I think this RegEx is very well documented. Well done.

I would add a test to show how it matches.

MarSavar:
Thanks @mxdvl, added a few tests to make sure the logic is correct!

Issue #4191, opened 2022-03-07
Analysis Card underline


mxdvl:
The line-height of Analysis cards does not match the `background-size` of the underline.

<img width="242" alt="image" src="https://user-images.githubusercontent.com/76776/157001964-18fd2b3d-c329-42c8-99de-422ae0615612.png">

Most recent PR touching these is #4048

AshCorr:
This also seems to be an issue with frontend (screenshot from https://www.theguardian.com/tone/analysis)

<img width="1308" alt="image" src="https://user-images.githubusercontent.com/21217225/164474441-f0cc7c01-9a52-4441-8665-c6875b8f7307.png">

AshCorr:
Going to leave Frontend out of this as this issue will be fixed once the current migration to DCR is done.

PR #4695, opened 2022-04-21
fix: Change analysis underlines to use the line heights provided by S…

AshCorr:
## What does this change?

Updates underline styles of analysis cards to have a `until.desktop` media query to set line-height. Theres another media query that is replacing the line height set by the underline CSS which causes the underlines to overlay over the text.

## Why?

Stop underlines from intersecting through text on analysis cards.
Part of #4191 

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/21217225/164747524-92b2499b-472a-45ca-81d5-e470e07588e5.png

[after]: https://user-images.githubusercontent.com/21217225/164747563-f6dbce88-e469-4f43-9f02-416997c21700.png

### DIFF
--- a/dotcom-rendering/src/web/components/CardHeadline.tsx
+++ b/dotcom-rendering/src/web/components/CardHeadline.tsx
@@ -78,7 +78,7 @@ const labTextStyles = (size: SmallHeadlineSize) => {
 };
 
 const underlinedStyles = (size: SmallHeadlineSize, colour: string) => {
-	function generateUnderlinedCss(baseSize: number) {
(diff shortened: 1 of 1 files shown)


oliverlloyd:
You fixed it! This has been a fiddly bug for so long...

mxdvl (approved):
magical, thanks 🧙‍♀️

HarryFischer (approved):
Nice!

PR #4696, opened 2022-04-21
Create new KeyEventsCards for carousel 

joecowton1:
## What does this change?

This PR is a WIP which creates new individual Key Event carousel cards. The styling will need refining as designs are completed but hopefully this gives something to work with.
## Screenshots

Screenshot 2022-04-21 at 21 20 24


dotcom-rendering/src/web/components/KeyEventsCards.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+	return {
+		design: ArticleDesign.Standard,
+		display: ArticleDisplay.Standard,
+		theme,
+	};
+};
+
+const format = getFormat(ArticlePillar.News);
+
+const events: KeyEvent[] = [
```
abeddow91, line 25: Non blocking but maybe we could move this into the fixtures?

abeddow91, line 25: Non blocking but maybe we could move this into the fixtures?

dotcom-rendering/src/web/components/KeyEventsCards.tsx
```diff
@@ -0,0 +1,110 @@
+import { css } from '@emotion/react';
+import { textSans, neutral, space, from } from '@guardian/source-foundations';
+import { Link } from '@guardian/source-react-components';
+import { timeAgo } from '@guardian/libs';
+import { decidePalette } from '../lib/decidePalette';
+
+export interface KeyEvent {
```
abeddow91, line 7: This is a different model than what a key event is - they are just of type Block. I wonder if we should pass in key events as blocks here and move the logic that determines what's a date, text, url etc into this component?

dotcom-rendering/src/web/components/KeyEventsCards.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+);
+
+const MultipleCards = () => (
+	<>
+		<Cards theme={ArticlePillar.News} count={7} />
+		<Cards theme={ArticlePillar.Culture} count={7} />
+		<Cards theme={ArticlePillar.Lifestyle} count={7} />
+		<Cards theme={ArticlePillar.Sport} count={7} />
+		<Cards theme={ArticlePillar.Culture} count={7} />
+		<Cards theme={ArticlePillar.Opinion} count={1} />
```
marjisound, line 117: Was this not supposed to be `count={7}` ?

abeddow91, line 117: Count looks wrong here - guessing it should be 7?

abeddow91, line 117: Count looks wrong here - guessing it should be 7?

abeddow91 (commented):
Nice one! A few (non-blocking) comments but looks great :)

joecowton1:
@abeddow91 @marjisound thanks for the comments! Can one of you please finish this one off in my absence?

dotcom-rendering/src/web/components/KeyEventsCards.tsx
```diff
@@ -0,0 +1,110 @@
...
+	}
+`;
+
+const timeStyles = css`
+	${textSans.xsmall({ fontWeight: 'bold', lineHeight: 'tight' })};
+	color: ${neutral[7]};
+	display: block;
+`;
+
+export const KeyEventsCard = ({ text, date, url, format }: KeyEvent) => {
```
oliverlloyd, line 84: I think if the export is `KeyEventCard` and it's only exporting one thing then the filename should also be `KeyEventCard.tsx`?

oliverlloyd:
I think the hover state needs adjusting? Maybe we can achieve this by lifting the link up?

dotcom-rendering/src/web/components/KeyEventsCards.tsx
abeddow91: This component handles one key event so `KeyEventCard` feels right to me.

abeddow91:
> I think the hover state needs adjusting? Maybe we can achieve this by lifting the link up? 

@oliverlloyd what would you expect instead here? I've taken a quick look at the existing implementation on DCR and I *think* these look the same. Maybe this is one for @benwuersching

oliverlloyd:
> What would you expect instead here?

The hover state and the link itself (where the mouse pointer changes) should be aligned

paperboyo:
Lovely. Is the line height a tad too loose? Toulouse.

ajbreuer:
> Lovely. Is the line height a tad too loose? Toulouse.

@benwuersching @HarryFischer Can you check this? :)

benwuersching:
Regarding line-height: please make it 135% (it's currently more than 160%). @abeddow91 @ajbreuer

oliverlloyd (approved):
Great work!

joecowton1:
@abeddow91 thanks for seeing this through Anna! I've pushed a fix for the line-height so hopefully this is good to go now?

alinaboghiu:
Yep, line height was the last thing as far as I know so good to merge.

PR #4699, opened 2022-04-22
Add headline refactor

iainjchambers-guardian:
## What does this change?
Modifies the file/directory structure of the `Headline` and `Byline` components to separate the variations for different designs into their own files and adopt the new pattern of `Polymorphic Components`. As well as this I have begun the process of moving other components that do not need to adopt this complex structure into their own directories to represent a more consistent structure.

## Why?
This represents a proposal for a larger overhaul to make the components more readable and make code more findable as it will stop us having monolithic components full of logic and styles for different use cases.


JamieB-gu (commented):
I think this is both interesting and promising, thanks for doing this!

I've written a bit more of an essay here than I normally would on a PR, sorry! I've done it mainly because I think if and when this change goes through it'll be setting several conventions for the rest of the codebase, so it would be good to take a bit of time to discuss while we're here, I hope you don't mind! 🙈 

## Non-Headline Components

The description specifies `Headline` but there are a few other components included in this change, like `Anchor`, `BodyImage`, `Audio` and so on. Are they connected to `Headline` and need to be migrated too?

## BodyImage

As far as I can tell we've split up a single `Headline` component into multiple sub-components, but for `BodyImage` it looks like we've gone the other way and merged sub-components into a single, main component? What are your thoughts here, did you find that `BodyImage` wasn't a good candidate for multiple components?

## Component Names And Types

You've standardised on exporting a default, anonymous function for these sub-components. What are the advantages of this over using variables and the `FC` type to define components and then exporting those (you can see this pattern elsewhere in the codebase, I'm not sure why `Headline` never used `FC`)?

```tsx
const MyComponent: FC<Props> = ({ propOne, propTwo }) => (...)
```

It seems the linter is unhappy about this pattern in at least the `AnalysisHeadline`, although I'm not sure why it's not elsewhere 🤔. Also, one caveat to bear in mind is that `ReactElement` is not a fully accurate return type for a React component, because it doesn't include `null`.

## Format vs Item

It looks like most of the components take `Item` as a prop, but actually only need the `Format` type to pass on to style functions? Could we narrow their inputs (the domain) to only require `Format`?

## Style Prop

There seem to be a few different ways to pass styles to a custom component, including `className` (used here and in Source), and `css` used by Emotion for built-in elements and possibly also by us for custom ones. I believe here you've introduced `styles` for the first time. Can we settle on a common convention for this?

iainjchambers-guardian:
@JamieB-gu 

> The description specifies Headline but there are a few other components included in this change, like Anchor, BodyImage, Audio and so on. Are they connected to Headline and need to be migrated too?

I agree, the description on the PR is outdated, while you were away I thought it would be helpful to refactor a few components in one go to highlight if there were any caveats to this proposed new pattern that hadn't been flagged by only doing one. 

Speaking of, in my head I've been referring to this pattern as `Polymorphic Components`

> As far as I can tell we've split up a single Headline component into multiple sub-components, but for BodyImage it looks like we've gone the other way and merged sub-components into a single, main component? What are your thoughts here, did you find that BodyImage wasn't a good candidate for multiple components?

While refactoring a handful of these components, I found that `bodyImage` didn't conform to this pattern of `Polymorphic Components` (I'm going to keep using that until someone comes up with something better). It's not a concept that renders differently for different designs, it's just a component that needed more props to change its rendering. That said, having done this work, I searched for references to update the import path and I couldn't find any. I think it's a dead component that's been replaced by something in common rendering. So...bit of a waste of time.

A few more notes:
- On the component typing; I agree, we should just standardise the type as `React.FC<Props>`
- Happy to use `format` as a consistent prop
- I used `styles` as a prop because `css` is a function exported by `@emotion` so it's tricky to have a prop and an import share a name but I agree this is a little janky, maybe I should name the prop `css` and alias it to maintain consistency?
- I didn't like the idea of having the default headline component be a separate file because it doesn't follow the same pattern as the others and it's essentially just a utility to prevent the duplication of those lines of code. The idea was to let the variants either use the default markup and just pass it extra styles or return a complete override of the markup with styles if it's different. The drawback as you pointed out is circular imports but what's the real issue with that? I like circles. They're neat. Seriously though...lets chat about that.

JamieB-gu:
> pattern of `Polymorphic Components` (I'm going to keep using that until someone comes up with something better)

😄

> I think it's a dead component that's been replaced by something in common rendering. So...bit of a waste of time.

Ah I think you may be right. Do you want to create an issue to delete it?

> I used styles as a prop because css is a function exported by @emotion so it's tricky to have a prop and an import share a name but I agree this is a little janky, maybe I should name the prop css and alias it to maintain consistency?

I think the consistency with the `css` prop for standard elements would be nice. As we've discussed separately I've also asked around the dept to see what people think.

> The idea was to let the variants either use the default markup and just pass it extra styles or return a complete override of the markup with styles if it's different.

This is interesting 👍, let's run with this.

apps-rendering/src/components/BodyImage/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+	remSpace,
+} from '@guardian/source-foundations';
+import Img from 'components/img';
+import type { BodyImageProps } from 'image';
+import type { FC } from 'react';
+import { darkModeCss } from 'styles';
+
+type BodyImageType = 'standard' | 'halfWidth' | 'thumbnail';
+
+const getSizes = (type: BodyImageType): string => {
```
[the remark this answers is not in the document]
paperboyo: If we are changing `role`, then in the tools it’s called `weighting`. But it’s `role` in CAPI, so introducing a third term for the same thing…

apps-rendering/src/components/Byline/Byline.defaults.tsx
```diff
@@ -0,0 +1,87 @@
...
+};
+
+export const renderText = (
+	format: ArticleFormat,
+	byline: DocumentFragment,
+	anchorStyles: SerializedStyles,
+): ReactNode =>
+	Array.from(byline.childNodes).map((node, i) =>
+		toReact(format, anchorStyles)(node, i),
+	);
```
JamieB-gu, line 68: Possible simplification?

```suggestion
	Array.from(byline.childNodes).map(toReact(format, anchorStyles));
```

apps-rendering/src/components/Byline/CommentByline.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { getThemeStyles } from 'themeStyles';
+import { DefaultByline } from './Byline.defaults';
+
+const commentStyles = (kicker: string): SerializedStyles => css`
+	color: ${kicker};
+	width: 75%;
+	${headline.medium({ fontWeight: 'light', fontStyle: 'italic' })}
+`;
+
+const commentAnchorStyles = (
```
JamieB-gu, line 17: Nitpick, but in this file can these just be `styles` and `anchorStyles`? Slightly fewer terms to hold in your head 🤔

apps-rendering/src/components/Byline/DeadBlogByline.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { headline, neutral } from '@guardian/source-foundations';
+import type { Option } from '@guardian/types';
+import { DefaultByline } from './Byline.defaults';
+import { blogColor } from './LiveBlogByline';
+
+const blogStyles = css`
+	${headline.xxxsmall({ lineHeight: 'regular', fontStyle: 'italic' })}
+`;
+
+const blogAnchorStyles = css`
```
JamieB-gu, line 13: Same question as above regarding `styles` and `anchorStyles`.

apps-rendering/src/components/Byline/LabsByline.tsx
```diff
@@ -0,0 +1,44 @@
...
+const labsStyles = css`
+	${textSans.medium({ lineHeight: 'regular', fontStyle: 'italic' })}
+	color: ${labs[300]};
+
+	${darkModeCss`
+        color: ${labs[400]};
+    `}
+`;
+
+const labsAnchorStyles = css`
```
JamieB-gu, line 17: Same question as above regarding `styles` and `anchorStyles`.

apps-rendering/src/components/Byline/LiveBlogByline.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { css } from '@emotion/react';
+import { text } from '@guardian/common-rendering/src/editorialPalette';
+import type { ArticleFormat } from '@guardian/libs';
+import { from, headline, neutral } from '@guardian/source-foundations';
+import type { Option } from '@guardian/types';
+import { darkModeCss } from 'styles';
+import { getThemeStyles } from 'themeStyles';
+import { DefaultByline } from './Byline.defaults';
+
+export const blogColor = (
```
JamieB-gu, line 11: Given the `defaults` pattern you're using, would it make sense for this to live there since it's used in a couple of sub-components?

apps-rendering/src/components/Byline/LiveBlogByline.tsx
```diff
@@ -0,0 +1,70 @@
...
+			color: ${inverted};
+		}
+	`}
+`;
+
+const blogStyles = css`
+	${headline.xxxsmall({ lineHeight: 'regular', fontStyle: 'italic' })}
+`;
+
+const blogAnchorStyles = css`
```
JamieB-gu, line 33: Same question as above regarding `styles` and `anchorStyles`.

apps-rendering/src/components/Headline/Headline.defaults.tsx
```diff
@@ -0,0 +1,46 @@
...
+			${headline.medium()}
+		}
+	`;
+
+	switch (format.design) {
+		case ArticleDesign.Interview:
+			return css`
+				${baseStyles}
+				padding-bottom: ${remSpace[1]};
+			`;
```
JamieB-gu, line 26: Should this be in `InterviewHeadline`?

apps-rendering/src/components/Headline/LabsHeadline.tsx
```diff
@@ -0,0 +1,24 @@
+import { css } from '@emotion/react';
+import { from, textSans } from '@guardian/source-foundations';
+import type { Item } from 'item';
+import { DefaultHeadline, defaultStyles } from './Headline.defaults';
+
+const labsStyles = css`
+	${textSans.xxxlarge({ lineHeight: 'regular' })}
+	${from.desktop} {
+		${textSans.xxxlarge({ lineHeight: 'regular' })}
+	}
```
JamieB-gu, line 10: Possibly not one for this PR, but these are the same 🙂

JamieB-gu (approved):
Nice! A few small comments but they're mostly not too important and unlikely to be blockers 🚀

PR #4708, opened 2022-04-25
Move `EmailEmbed` caption from underneath to above

### LINKED ISSUE #4704: Change email embed caption position
From Rhianna:

> Hey dotcom, I was wondering if you could help us understand something. Our email sign up embeds which sit in articles don’t look great and we’d love to move their caption so that it sits above the sign up field. It’s too complex to do this in Frontend as the captions are done using the same feature as image captions, but it looks like we could do this for DCR-rendered embeds (dotcom-rendering/src/web/components/EmbedBlockComponent.importable.tsx).

oliverlloyd:
## What does this change?
Moves the caption shown under email embeds from below the input to above it

Fixes #4704 

## Why?
So it better describes the content being seen

| Before      | After      |
|-------------|------------|
|Screenshot 2022-04-25 at 16 36 48 | Screenshot 2022-04-25 at 16 36 08 |

### DIFF
--- a/dotcom-rendering/src/web/components/EmbedBlockComponent.importable.tsx
+++ b/dotcom-rendering/src/web/components/EmbedBlockComponent.importable.tsx
@@ -1,5 +1,5 @@
 import { css } from '@emotion/react';
-import { textSans, text } from '@guardian/source-foundations';
+import { textSans, text, space } from '@guardian/source-foundations';
 import { unescapeData } from '../../lib/escapeData';
 import { ClickToView } from './ClickToView';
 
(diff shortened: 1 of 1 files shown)


ioannakok (approved):

OllysCoding:
Where else is EmbedBlockComponent used? Could this result in unforseen design changes for other components across the site?

oliverlloyd:
> Where else is EmbedBlockComponent used? Could this result in unforseen design changes for other components across the site?

There's this check in the code to filter this change to only impact email sign up embeds.

```js
	// TODO: Email embeds are being turned into atoms, so we can remove this hack when that happens
	const isEmailEmbed = html.includes('email/form');
```

This is obviously fairly awful but does mean we are _probably_ only going to impact email embeds 🤞 🍀 🙈 . I did raise this point when discussing this approach and it was felt the best solution while we wait for email embeds to be revisited

PR #4712, opened 2022-04-25
Fix `o` on fronts

oliverlloyd:
## What does this change?
Don't try to render false

## Why?
It ends up as a zero in the html

| Before      | After      |
|-------------|------------|
| Screenshot 2022-04-25 at 17 23 48 | Screenshot 2022-04-25 at 17 23 14|

### DIFF
--- a/dotcom-rendering/src/web/components/Card/Card.tsx
+++ b/dotcom-rendering/src/web/components/Card/Card.tsx
@@ -160,8 +160,9 @@ export const Card = ({
 
 	const cardPalette = decidePalette(format);
 
-	const moreThanTwoSubLinks =
-		supportingContent?.length && supportingContent.length > 2;
+	const moreThanTwoSubLinks: boolean = !!(
+		supportingContent?.length && supportingContent.length > 2
+	);
 
 	const renderFooter = ({
 		renderAge = true,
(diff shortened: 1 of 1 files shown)


mxdvl (commented):
Should we not render `null`?

oliverlloyd:
> Should we not render `null`?

I think they are essentially the same but a fragment  feels more semantic? For example, `<></>` is a valid `React.Element` but `null` is not  so typescript is less likely to complain in some cases where the return type is set by, maybe say, `React.FC`.

I think there are examples of both in the codebase. I'm not especially strong on one way or another but I can imagine the migration from `<></>` to `null` would be a lot easier 🤣

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -317,13 +317,16 @@ export const Card = ({
 								</Hide>
 							)}
 							{/* Show the card footer in the same column as the headline content */}
-							{!moreThanTwoSubLinks &&
-								renderFooter({ forceVertical: true })}
+							{!moreThanTwoSubLinks ? (
```
jamesgorrie, line 320: This is where setting boolean values as booleans, instead of boolean scan be useful to avoid this. Solving the cause of the issues would make the value more predictable in how it behaves

i.e.
```ts
// Currently
const moreThanTwoSubLinks: boolean | 0 | undefined =
		supportingContent?.length && supportingContent.length > 2;

// Predictable
const moreThanTwoSubLinks: boolean =
		supportingContent?.length && supportingContent.length > 2 ? true : false;
```

jamesgorrie (approved):
A thought on how to make it a little simpler, but the `moreThanTwoSubLinks` value doesn't leak so probably not a major issue.

PR #4713, opened 2022-04-25
Show which container types we don't support

oliverlloyd:
## What does this change?
This removes the `DynamicFast` default and replaces it with some debug text
Screenshot 2022-04-25 at 17 58 28

## Why?
It doesn't look as exciting but it more useful for development


AshCorr (approved):
👍

PR #4718, opened 2022-04-26
`DynamicSlow`

oliverlloyd:
## What does this change?
This PR adds support for the `DynamicSlow` container

Screenshot 2022-04-25 at 18 48 20

### DIFF
--- a/dotcom-rendering/src/web/components/DynamicSlow.tsx
+++ b/dotcom-rendering/src/web/components/DynamicSlow.tsx
@@ -1,0 +1,192 @@
+import { ArticleDesign } from '@guardian/libs';
+
+import { Card } from './Card/Card';
+import { UL } from './Card/components/UL';
+import { LI } from './Card/components/LI';
+
+type Props = {
+	trails: TrailType[];
+};
+
+export const DynamicSlow = ({ trails }: Props) => {
+	const primary = trails[0];
+	const secondary = trails[1];
+	const smallCards = trails.slice(2, 6);
+	const bigCards = trails.slice(6, 8);
+
+	return (
+		<>
+			<UL direction="row" padBottom={true}>
+				<LI padSides={true} percentage="75%">
+					<Card
+						linkTo={primary.url}
+						format={primary.format}
+						headlineText={primary.headline}
+						trailText={primary.trailText}
+						headlineSize="large"
+						byline={primary.byline}
+						showByline={primary.showByline}
+						showQuotes={
+							primary.format.design === ArticleDesign.Comment ||
+							primary.format.design === ArticleDesign.Letter
+						}
+						webPublicationDate={primary.webPublicationDate}
+						kickerText={primary.kickerText}
+						showPulsingDot={
+							primary.format.design === ArticleDesign.LiveBlog
+						}
+						showSlash={true}
+						showClock={false}
+						imageUrl={primary.image}
+						imagePosition="right"
+						imagePositionOnMobile="top"
+						imageSize="large"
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/DynamicSlow.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+import { Card } from './Card/Card';
+import { UL } from './Card/components/UL';
+import { LI } from './Card/components/LI';
+
+type Props = {
+	trails: TrailType[];
+};
+
+export const DynamicSlow = ({ trails }: Props) => {
```
mxdvl, line 11: As a general rule, I think that we should always have some kind of documentation for exported methods. Even if it only contains a link to a Figma file.

```js
/**
 * A `Card` container for up to 8 trails.
 * 
 * [Figma][] · [Chromatic][] · []
 * [Figma]: https://www.figma.com/file/sx2vMFHbL7SsUo0LcpsKNe/%E2%AC%A3--Front-container?node-id=123%3A137122
 * [Chromatic]: https://www.chromatic.com/component?appId=5dfcbf3012392c0020e7140b&csfId=components-dynamicslow&buildNumber=12105&k=62678bafdd54f5004a5dcb68-320-interactive-true&h=6&b=-1
 */
 const DynamicSlow = ({ trails }: Props) => { /* … */ };
```

mxdvl (approved):
Nice!

Looking at the Figma file, it seems like there are more possible variations than the one shown here. Are we trying to get each container completely right first, or improving each further along the journey?

dotcom-rendering/src/web/components/DynamicSlow.tsx
OllysCoding: +1

OllysCoding (approved):
+1 I love how approachable the code for these layouts look! If we were to encounter issues where they needed changed and/or fixed, it'd be super easy for even someone new to the project to figure it out :)

oliverlloyd:
> 
> Looking at the Figma file, it seems like there are more possible variations than the one shown here. Are we trying to get each container completely right first, or improving each further along the journey?

This is a good question. I think it would be nice if the PR to add a container was properly feature complete but I don't think this is possible at this point. We have not added support for boosting or for having cards marked as Huge, Very Big, etc; both of these will likely impact layout.

My feeling is that - at this stage - there is benefit in having a non feature complete container in the codebase. It means future work to add support for additional complexity can be tightly scoped with smaller PRs, making it easier to review. But I don't think that we should keep doing this and future PRs to add containers should have stricter requirements. For example, once we support boosting, we should make sure it is included.

dotcom-rendering/src/web/components/DynamicSlow.tsx
oliverlloyd: Good shout. I added a comment with a link using the [see tag](https://jsdoc.app/tags-see.html).

Note. I removed the link to Chromatic as the issue there is the true, canonical source for that view only exists after a component is merged into main and so either we add these retrospectively or we depend on Chromatic's navigation. That said, if there were some kind of code mod to go through and add these at a later stage that would be cool!

ajbreuer:
May be being premature about this, as it still wip, apols if so. but the four small stories stacked on the left are missing their pictures. :)

oliverlloyd:
> May be being premature about this, as it still wip, apols if so. but the four small stories stacked on the left are missing their pictures. :)

Not premature at all. It's true this work is wip but that will be the case for a while and we still want to know where we're going wrong. So thanks! I'll add these images in

PR #4719, opened 2022-04-26
Only render key event toggle if some exist

### LINKED ISSUE #4523: Hide keyevents toggle when no key events

oliverlloyd:
## What does this change?
Hides the key event toggle if there are no key events

This fixes #4523 

## Why?
It's pointless to show it when there are no events

| Before      | After      |
|-------------|------------|
|Screenshot 2022-04-26 at 08 38 24|Screenshot 2022-04-26 at 08 37 34|

### DIFF
--- a/dotcom-rendering/src/web/layouts/Liveblog.stories.tsx
+++ b/dotcom-rendering/src/web/layouts/Liveblog.stories.tsx
@@ -1,0 +1,67 @@
+import { useEffect } from 'react';
+
+import { breakpoints } from '@guardian/source-foundations';
+
+import { decideFormat } from '../lib/decideFormat';
+
+import { Live } from '../../../fixtures/generated/articles/Live';
+
+import { embedIframe } from '../browser/embedIframe/embedIframe';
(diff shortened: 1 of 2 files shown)


mxdvl (approved):
seems obvious once you point it out!

dotcom-rendering/src/web/layouts/Liveblog.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+			viewports: [breakpoints.wide],
+			pauseAnimationAtEnd: true,
+		},
+	},
+};
+
+// HydratedLayout is used here to simulated the hydration that happens after we init react on
+// the client. We need a separate component so that we can make use of useEffect to ensure
+// the hydrate step only runs once the dom has been rendered.
+const HydratedLayout = ({
```
mxdvl, line 32:
```suggestion
/**
 * HydratedLayout is used here to simulated the hydration that happens after we init react on
 * the client. We need a separate component so that we can make use of `useEffect` to ensure
 * the hydrate step only runs once the DOM has been rendered.
 */
const HydratedLayout = ({
```

PR #4742, opened 2022-04-26
Do not show small byline images in articles

ioannakok:
Co-authored-by: James Gorrie <james.gorrie@guardian.co.uk>

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |


dotcom-rendering/index.d.ts
```diff
@@ -572,11 +573,11 @@ type FEFrontPropertiesType = {
 
 type FESupportingContent = {
 	properties: {
-		href: string;
+		href?: string;
```
AshCorr, line 576: Was this intentional?

AshCorr (approved):
LGTM!

dotcom-rendering/index.d.ts
ioannakok: Yes, we had to do it. That's a change not related to this PR but we had to run `gen-schema` because this PR introduces changes in `index.d.ts` and it picked up previous changes as well. The new schema had this as required and it was breaking `http://localhost:3030/Front?url=https://www.theguardian.com/uk`. This in turn was breaking Lighthouse checks - now passing: https://github.com/guardian/dotcom-rendering/runs/6210201717?check_suite_focus=true. I hope this makes sense!

PR #4745, opened 2022-04-27
Access Parameter store config locally via Dotenv

jamesgorrie:
Part of https://github.com/guardian/dotcom-rendering/issues/4823

Generates a `.env` file from parameter store as part of our `make dev` command.

`gen-dotenv` has been added to the `build`,  `build-ci` and `dev` targets as it is only needed when running the server and is a file that is part of the build.

I've added a check to the see if we're running `NODE_ENV=production` to fail if the file cannot be generated, and then set that for the general targets `build` and `build-ci` rather than just when running the node script.

- [x] check env
- [x] gen env
- [x] read generated `.env`

# Dev messaging
Screenshot 2022-05-04 at 21 47 37

Co-authored-by: Olly Namey <olly.namey@guardian.co.uk>
Co-authored-by: Ash Carr <ashleigh.carr@guardian.co.uk>
Co-authored-by: Ioanna Kokkini <ioanna.kokkini@guardian.co.uk>


dotcom-rendering/scripts/env/check-dotenv.ts
```diff
@@ -0,0 +1,16 @@
+import { promises as fs } from 'fs';
+import * as path from 'path'
+import { getGuardianConfiguration } from '../../src/server/lib/aws/aws-parameters';
+
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
+(async () => {
+	const envVars = ['images.signature-salt'];
```
OllysCoding, line 7: Is this where the salt is currently held or is it this a new var?

dotcom-rendering/tsconfig.json
```diff
@@ -68,5 +68,14 @@
...
+    "exclude": ["cypress"],
+    "ts-node": {
+        "esModuleInterop": true,
+        "allowSyntheticDefaultImports": true,
+        "compilerOptions": {
+            "module": "commonjs",
+            "esModuleInterop": true,
+            "allowSyntheticDefaultImports": true,
+        }
+    },
```
OllysCoding, line 80: Did you have much success getting ts-node to work on DCR? I think when @mxdvl tried we couldn't get it to work!

dotcom-rendering/scripts/env/check-dotenv.ts
jamesgorrie: It's what it was previously.

dotcom-rendering/scripts/env/gen-dotenv.js
```diff
@@ -0,0 +1,16 @@
...
+(async () => {
+	const env = process.env.NODE_ENV === 'production' ? 'prod' : 'dev';
+  const configuration = await getGuardianConfiguration(env)
+	const envFile = path.resolve(__dirname, '../../.env');
+
+	for (const secretKey of secrets) {
+		// eslint-disable-next-line no-await-in-loop
+		await fs.appendFile(envFile, `${secretKey}=${configuration.getParameter(secretKey)}`)
+	}
+})()
```
mxdvl, line 16: Is this wrapped in an IIFE only so it can be `async`?

- Could we use top-level `await`?
- Could we use the non-promise `fs` API?
- _🦕_ ? 😉

jamesgorrie: The await was setup to avoid race conditions on writing lines to a file. We've removed the need for this.

dotcom-rendering/scripts/dev/aws-parameters.js
```diff
@@ -0,0 +1,70 @@
+const {
+	GetParametersByPathCommand,
+	SSMClient,
+} = require('@aws-sdk/client-ssm');
+
+process.env.AWS_PROFILE = 'frontend';
```
jamesgorrie, line 6: Should we potentially start a new namespace of dotcom?

dotcom-rendering/scripts/dev/aws-parameters.js
```diff
@@ -0,0 +1,70 @@
...
+const ssm = new SSMClient({ region: 'eu-west-1' });
+const STACK = 'frontend';
```
jamesgorrie, line 8: Should we potentially start a new namespace of dotcom?

dotcom-rendering/docs/values/README.md
```diff
@@ -0,0 +1,13 @@
+§# Values
```
jamesgorrie, line 1:
```suggestion
# Values
```

dotcom-rendering/docs/values/README.md
```diff
@@ -0,0 +1,13 @@
+# Values
+
+## Approachable
+
+The project should have as few barriers as possible for anyone to work on it, including designers, UX, people outside of the Guardian and more.
+
+### Lines in the sand
+
+#### `.env` shouldn't be required
```
OllysCoding, line 9: +1 An important point not to be forgotten

dotcom-rendering/makefile
```diff
@@ -30,13 +30,15 @@ deploy:
...
 	$(call log, "building production bundles")
-	@NODE_ENV=production webpack --config ./scripts/webpack/webpack.config.js --progress
+	webpack --config ./scripts/webpack/webpack.config.js --progress
 
-build-ci: clean-dist install
+build-ci: export NODE_ENV=production
+build-ci: clean-dist install gen-dotenv
 	$(call log, "building production bundles")
-	@NODE_ENV=production SKIP_LEGACY=true webpack --config ./scripts/webpack/webpack.config.js
+	SKIP_LEGACY=true webpack --config ./scripts/webpack/webpack.config.js
```
OllysCoding, line 41: I'm intrigued by this change, could we quickly outline what it does?

I've also noticed all other calls in the makefile start with `@`, but on line 41 and 36 this has been removed?

OllysCoding (commented):
This looks really good!

Should we document somewhere how this all works, and I think an important thing to cover is that right now we only pass our .env variables into the server build - meaning that they're not available on the client.

It's important that if that were to ever change appropriate considerations were taking (like requiring a `CLIENT_` prefix) to avoid any accidental security leaks.

dotcom-rendering/scripts/dotenv/gen-dotenv.js
```diff
@@ -0,0 +1,81 @@
+const { CredentialsProviderError } = require('@aws-sdk/property-provider');
+const path = require('path');
+const { prompt, log, warn } = require('../env/log');
```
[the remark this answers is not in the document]
OllysCoding: I didn't actually know this library existed! Are there some other places we should be using it?

dotcom-rendering/docs/values/README.md
jamesgorrie: Do you think it's worth enforcing in some or other way? That might be how we access to config is the only way I can think of.

dotcom-rendering/makefile
jamesgorrie: It's using [target specific variable values](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html) so it sets the value for the whole step, and thus substeps, not just the command. 

The removal of the `@` sign was a mistake. It doesn't make a difference to the scripts as it [suppresses echoing](https://www.gnu.org/software/make/manual/make.html#Recipe-Echoing). I've added it back - nice spot!

dotcom-rendering/scripts/webpack/@types/dotenv-webpack/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'dotenv-webpack';
```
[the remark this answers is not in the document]
jamesgorrie: It imports `@types/webpack` which then throws a tonne of errors. Happy to do that as another PR as it would really muddy this one.

jamesgorrie:
@mxdvl "floating promises" Could you point this out - if you're talking about in `aws-parameters`, this was a full copy/paste. Happy to refine, and maybe test this script? 

"common-js" <- same it was just copy/paste.

@OllysCoding nice idea, I'll add an ADR.

mxdvl (approved):

dotcom-rendering/docs/architecture/028-dotenv.md
```diff
@@ -0,0 +1,40 @@
+# `.env` File
+
+Any values in the `.env` file [shouldn't be required to run the application locally](../values/README.md#env-shouldnt-be-required) - to keep is at approachable as possible.
```
ioannakok, line 3:
```suggestion
Any values in the `.env` file [shouldn't be required to run the application locally](../values/README.md#env-shouldnt-be-required) - to keep it as approachable as possible.
```

👩‍🏫

PR #4746, opened 2022-04-27
Better layout stories

oliverlloyd:
## What does this change?
This PR refactors how we do stories for layouts. It replaces the series of different files we had before, each trying to address a particular use case, with a single file that loops over the variations contained in `Format`. Any additional layout stories that we need, such as a liveblog with no key events, can be added underneath in the same file.

## Why?
This ensures we have the best coverage, capturing all the possible combinations of the different Format properties that could exist

### But wait, do we really offer an immersive lifestyle liveblog?!
Well, no, not at the moment. But this is purely by _convention_. From a technical point of view, such an article is possible and the platform should support it, even if that support simply means defaulting to something else.


mxdvl (approved):
Yes, full coverage is the only way forward!

🤖

PR #4750, opened 2022-04-27
Track epic events in liveblog

tomrf1:
This was missed in the migration to dcr. The epic component needs a `submitComponentEvent` function in its props to support ophan tracking - https://github.com/guardian/frontend/blob/main/static/src/javascripts/projects/common/modules/support/epic.ts#L100

Tested locally:
Screen Shot 2022-04-27 at 09 53 58

### DIFF
--- a/dotcom-rendering/src/web/components/LiveBlogEpic.importable.tsx
+++ b/dotcom-rendering/src/web/components/LiveBlogEpic.importable.tsx
@@ -250,6 +250,7 @@ const Fetch = ({
 	}
 	log('dotcom', 'LiveBlogEpic has a module');
 
+	// Add submitComponentEvent function to props to enable Ophan tracking in the component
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/components/LiveBlogEpic.importable.tsx
```diff
@@ -249,12 +250,17 @@ const Fetch = ({
 	}
 	log('dotcom', 'LiveBlogEpic has a module');
 
+	const props = {
```
oliverlloyd, line 253:
```suggestion
	// We add submitComponentEvent to the props like this here on the client because
	//  the ophan record function is not available on the server
	const props = {
```

oliverlloyd (approved):
Looks great - thanks for the PR 🙏

PR #4756, opened 2022-04-27
refactor: Move front containers to their own subfolder

AshCorr:
## What does this change?

Move all front containers to a `FrontContainer` sub folder

## Why?

We're going to end up with quite a lot of different front containers, it might be helpful to group all of these in a single folder.

### DIFF
--- a/dotcom-rendering/src/web/components/FrontContainer/DynamicFast.stories.tsx
+++ b/dotcom-rendering/src/web/components/FrontContainer/DynamicFast.stories.tsx
@@ -1,12 +1,12 @@
 import { breakpoints } from '@guardian/source-foundations';
-import { ContainerLayout } from './ContainerLayout';
-import { DynamicSlow } from './DynamicSlow';
+import { ContainerLayout } from '../ContainerLayout';
+import { DynamicFast } from './DynamicFast';
 
-import { trails } from '../../../fixtures/manual/trails';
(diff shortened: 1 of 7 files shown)


OllysCoding (approved):
I love me a good sub-folder but I've heard mixed views so I'd love to see some thoughts from the rest of the team too :)

mxdvl (commented):
I rarely, if ever, navigate using the tree structure[^1], so I’m ambivalent on this one.

[^1]: I mostly open files directly with `cmd+P`

AshCorr:
> I rarely, if ever, navigate using the tree structure[1](#user-content-fn-1-52dea5213713d962cf82906789ae51de), so I’m ambivalent on this one.

In this case I'd probably use it for finding similar front containers for reference... Right now I think you'd do this by looking at the DecideContainer function

oliverlloyd (commented):
The flat structure was originally by design. The thinking as it was explained to me was that people find files using CMD+P or other means so discovery isn't a problem and folder structures can feel nice but they have downsides, like, now I have a non standard way to write imports which can make copy pasting/refactoring less predictable.

But arguably this is a 'type' of component and so it deserves its own home?

🤷‍♂️ There's no right or wrong here. Maybe things are [getting annoying](https://mobile.twitter.com/dan_abramov/status/1027248875072114689) and we need to take a different approach?

AshCorr:
Seems like this is fine for now

PR #4758, opened 2022-04-27
Liveblogs - Cricket Scoreboard

OllysCoding:
## What does this change?

Frontend PRs: 
- https://github.com/guardian/frontend/pull/24883
- https://github.com/guardian/frontend/pull/24928
- https://github.com/guardian/frontend/pull/24959

This PR updates DCR to do 2 things:
1. Consume the new `MatchType` passed from frontend in the LiveLayout
2. Render the live cricket summary scorecard

The cricket scorecard is identical to that in frontend except for a few contrast & font size changes to improve accessibility

### Testing 
1. I've tested this myself with a screenreader, and confirmed it aligns with Frontend & is usable
2. Added storybook story
3. Added unit tests for score processing
4. Added cypress test for live updating

## Why?

Working towards liveblogs to 100%

## Screenshots

| Frontend      | DCR      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/9575458/165790129-899e8a78-e618-4777-9841-af724e96fcb8.png
[after]: https://user-images.githubusercontent.com/9575458/165790049-47a15b24-c1a3-4812-8e39-37b4e5216eff.png

[before2]: https://user-images.githubusercontent.com/9575458/165790442-c6c81716-0feb-4a79-8860-a58b92471be9.png
[after2]: https://user-images.githubusercontent.com/9575458/165790343-238764d3-d5c0-4c85-adb8-3e9ebdac3ec9.png


dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -393,7 +400,7 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 
 			<main>
 				<article>
-					{CAPIArticle.matchUrl ? (
+					{CAPIArticle.matchUrl && isFootballMatch ? (
```
OllysCoding, line 403: You need to keep the `CAPIArticle.matchUrl` check here otherwise TS doesn't consider it to `string` instead of `string | undefined` (a weird limitation of TS in my opinion)

dotcom-rendering/src/web/components/GetCricketScoreboard.importable.tsx
```diff
@@ -0,0 +1,41 @@
...
+export const GetCricketScoreboard = ({ matchUrl, format }: Props) => {
+	// TODO: This is for testing only
+	const url = matchUrl.replace(
+		'http://localhost:9000',
+		'https://api.nextgen.guardianapps.co.uk',
+	);
+
+	const { data, error, loading } = useApi<CricketMatch>(url, {
+		errorRetryCount: 1,
+	});
```
OllysCoding, line 22: TODO: Live refreshing

dotcom-rendering/src/web/components/CricketScoreboard.tsx
```diff
@@ -0,0 +1,184 @@
...
+						<td css={linkPaddingStlye} colSpan={2}>
+							<a
+								css={linkStyle(palette)}
+								href={matchUrl.replace('.json?dcr=true', '')}
+							>
+								View full scorecard
+							</a>
+						</td>
+					</tr>
+				</tfoot>
```
OllysCoding, line 180: TODO: Pass a 'scoreboardUrl' from Frontend

dotcom-rendering/src/web/components/CricketScoreboard.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+const tableStyle = css`
+	width: 100%;
+	${textSans.xsmall()}
+`;
+
+const captionStyle = (palette: Palette) => css`
+	text-align: left;
+	font-weight: bold;
+	border-top: 0.0625rem solid ${palette.border.cricketScoreboardTop};
```
OllysCoding, line 41: Should this be `px` or `rem` in DCR?

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -908,6 +908,18 @@ const backgroundMatchNav = (): string => {
 	return '#FFE500';
 };
 
+const cricketScoreboardTop = (): string => {
+	return sport[300];
+};
+
+const cricketScoreboardLink = (): string => {
+	return sport[300];
+};
```
mxdvl, line 917: Do we need this distinction yet?

mxdvl (approved):
🏏

dotcom-rendering/src/web/lib/decidePalette.ts
JamieB-gu: I think this is an interesting question and has come up a couple of times recently. If two different pieces of UI are nearby, e.g. a link and an adjacent border as in this case, and follow the same colour rules, do you use the same palette function?

I'd be inclined to agree with Olly and say they should be different even if the code looks the same because a) they're semantically different and b) they may diverge in future. But you could also argue that if they diverge you can split them up then, so I'd be interested to hear what people think.

cc @iainjchambers-guardian @MarSavar because the same principle applies to `editorialPalette`.

mxdvl: I completely agree that these should be members of different subsets of `palette`, namely `text` and `border`.

I think that you’re seeing it this way: 

```mermaid
graph LR
  decidePalette --> text & border
  text --> cricketScoreboardLink --> sport300
  border --> cricketScoreboardTop --> sport300
```

And I’d be inclined to do:

```mermaid
graph LR
  decidePalette --> text & border
  text --> cricketScoreboardColour
  border --> cricketScoreboardColour
  cricketScoreboardColour --> sport300
```

OllysCoding: Regardless of of anyones opinions here, I think we can all say with confidence that you went above and beyond with those diagrams 😍😍😍

--- 

I decided to take another look at palette and see what stood out to me in terms of practices for the file.

One thing I noticed was that outside of a few cases, all `palette.text` function were prefixed with `text`, and the same for border, fill, etc.

For example:
```js
{
    text: {
        richLink: textRichLink()
    },
    border: {
         rickLink: borderRickLink()
    }
}
```

I think it makes sense in this case to continue this pattern, so my additions would now look like

```js
{
    text: {
        cricketScoreboardLink: textCricketScoreboardLink()
    },
    border: {
        cricketScoreboardTop: borderCricketScoreboardTop(),
        cricketScoreboardDivider: borderCricketScoreboardDivider()
    }
}
```
While this does result in functions with duplicated responses, it keeps the explicit and ordered pattern which the `decidePalette` file currently uses.

If we wanted to, we could revisit this file (especially as it may be further expanded for fronts?) and see if the patterns used there are still providing the best experience they could!

PR #4763, opened 2022-04-27
Ad placeholder as a React component

MarSavar:
## Why?
Ad slots in AR were defined in `ads.tsx` as a mix of React's `createElement` (aliased as `h`) and JSX. This prevented the file from having a `.ts` extension (which we reserve for logic only).
`ads.tsx` handled both the ad insertion logic and the implementation of the placeholder component. Additionally, CSS styles were being pulled from the `adStyles` function in the `ArticleBody` component, which made the ad placeholder styles dependant on the body of an article.

This PR attempts to simplify how we reason about ads in AR by:

1) Creating a separate `AdSlot` React component.
2) Decoupling the CSS styles from `ArticleBody` and `AdSlots`: instead of being declared in the `ArticleBody` component, all styles now live in the `AdSlot` component and make use of `editorialPalette` to decide the colours.

As a result of 1. and 2., the component can be visually tested in isolation because styles are no longer applied in the article body (I've created a Storybook story to reflect this).

Also, because JSX is no longer used in the ad logic, the file previously known as `ads.tsx` is now `ads.ts`, and the placeholder component lives separately in `adSlot.tsx`, which should hopefully be an easier mental model to follow.

**NOTE**: class names are still needed in the component itself because of how placeholders are injected between paragraphs.

cc @guardian/commercial-dev

## Screenshots
*There is no visual difference between the old way of rendering ad placeholders and the new one*
| Before      | After      |
|-------------|------------|
| ![before1][] | ![after1][] |

[before1]: https://user-images.githubusercontent.com/57295823/165552817-612c6321-ac27-4d18-bebb-257b3b62b869.png
[after1]: https://user-images.githubusercontent.com/57295823/165551589-159a70ef-d8e3-4814-b2e6-6af0c83197f0.png
[before2]: https://user-images.githubusercontent.com/57295823/165553080-b8aab50c-d5be-4963-a896-a6ca039d93ae.png
[after2]: https://user-images.githubusercontent.com/57295823/165551213-496c9bac-4c41-4913-be26-aac599abeab6.png

## Storybook

image

### DIFF
--- a/apps-rendering/src/adSlot.tsx
+++ b/apps-rendering/src/adSlot.tsx
@@ -84,16 +84,6 @@ const styles = (format: ArticleFormat): SerializedStyles => css`
 		margin: 1em -${remSpace[3]};
 	}
 
-	// This class is applied in the nativeCommunication module
-	// to the first ad slot if Teads ads are enabled.
-	&.ad-slot-square {
-		height: 344px;
-		width: 320px;
-		margin-left: auto;
-		margin-right: auto;
-		padding-bottom: 0;
-	}
-
 	// This class is applied if the article has fewer than 15 paragraphs.
 	&.short:nth-of-type(1) {
 		${from.desktop} {
@@ -107,6 +97,16 @@ const adHeight = '258px';
 const adSlotStyles = css`
 	clear: both;
 	padding-bottom: ${adHeight};
+
+	// This class is applied in the nativeCommunication module
+	// to the first ad slot if Teads ads are enabled.
+	&.ad-slot-square {
+		height: 344px;
+		width: 320px;
+		margin-left: auto;
+		margin-right: auto;
+		padding-bottom: 0;
+	}
 `;
 
 const AdSlot: FC<Props> = ({ className, paragraph, format }): ReactElement => (
--- a/apps-rendering/src/adSlot.tsx
+++ b/apps-rendering/src/adSlot.tsx
@@ -1,0 +1,25 @@
+import { ThemeProvider } from '@emotion/react';
+import { Button, buttonThemeBrandAlt } from '@guardian/source-react-components';
+import { FC, ReactElement } from 'react';
+
+type Props = {
+	className: string;
+	paragraph: number;
+};
+
+const AdSlot: FC<Props> = ({ className, paragraph }): ReactElement => (
+	<aside className={className} key={`ad-after-${paragraph}-para`}>
+		<div className="ad-labels">
+			<h1>Advertisement</h1>
+		</div>
+		<div className="ad-slot"></div>
+		<div className="upgrade-banner">
+			<h1>Support the Guardian and enjoy the app ad-free.</h1>
+			<ThemeProvider theme={buttonThemeBrandAlt}>
+				<Button>Support the Guardian</Button>
+			</ThemeProvider>
+		</div>
+	</aside>
+);
+
+export default AdSlot;
--- a/apps-rendering/src/adSlot.tsx
+++ b/apps-rendering/src/adSlot.tsx
@@ -27,7 +27,8 @@ const adLabelsStyles = (format: ArticleFormat): SerializedStyles => css`
 	color: ${text.adLabel(format)};
 	padding: ${remSpace[3]};
 	float: left;
-	width: calc(100% - ${remSpace[6]});
+	// We need to account for padding on both sides
+	width: calc(100% - 2 * ${remSpace[3]});
 
 	h1 {
 		margin: 0;
--- a/apps-rendering/src/adSlot.tsx
+++ b/apps-rendering/src/adSlot.tsx
@@ -1,19 +1,124 @@
-import { ThemeProvider } from '@emotion/react';
+import { css, SerializedStyles, ThemeProvider } from '@emotion/react';
+import {
+	background,
+	text,
+} from '@guardian/common-rendering/src/editorialPalette';
+import { ArticleFormat } from '@guardian/libs';
+import {
+	from,
+	headline,
+	remSpace,
+	textSans,
+	until,
+} from '@guardian/source-foundations';
 import { Button, buttonThemeBrandAlt } from '@guardian/source-react-components';
 import type { FC, ReactElement } from 'react';
+import { darkModeCss, wideContentWidth } from 'styles';
 
 type Props = {
 	className: string;
 	paragraph: number;
+	format: ArticleFormat;
 };
 
-const AdSlot: FC<Props> = ({ className, paragraph }): ReactElement => (
-	<aside className={className} key={`ad-after-${paragraph}-para`}>
-		<div className="ad-labels">
+const adLabelsStyles = (format: ArticleFormat): SerializedStyles => css`
+	${textSans.xsmall()}
+	color: ${text.adLabel(format)};
+	padding: ${remSpace[3]};
+	float: left;
+	width: calc(100% - ${remSpace[3]} - ${remSpace[3]});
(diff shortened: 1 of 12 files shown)


apps-rendering/src/client/nativeCommunication.ts
```diff
@@ -126,7 +126,7 @@ function ads(): void {
 
 			insertAds();
 			Array.from(
-				document.querySelectorAll('.ad-labels, .upgrade-banner button'),
+				document.querySelectorAll('.ad-labels, .support-banner button'),
```
MarSavar, line 129: I've changed the class name to reflect the new ad copy (see #4697)

apps-rendering/src/ads.tsx
```diff
@@ -1,21 +1,24 @@
...
+import type { ReactNode } from 'react';
+import { isValidElement } from 'react';
 
 function getAdIndices(): number[] {
 	const adEveryNParagraphs = 6;
 	const firstAdIndex = 3;
 	const totalAds = 15;
 
-	const indiciesAfterFirstAd = [...Array(totalAds - 1).keys()].map(
+	const indicesAfterFirstAd = [...Array(totalAds - 1).keys()].map(
```
MarSavar, line 12: This is just a spelling correction

apps-rendering/src/adSlot.tsx
```diff
@@ -0,0 +1,131 @@
...
+	paragraph: number;
+	format: ArticleFormat;
+};
+
+const adLabelsStyles = (format: ArticleFormat): SerializedStyles => css`
+	${textSans.xsmall()}
+	color: ${text.adLabel(format)};
+	padding: ${remSpace[3]};
+	float: left;
+	width: calc(100% - ${remSpace[3]} - ${remSpace[3]});
```
mxdvl, line 30:
```suggestion
	width: calc(100% - 2 * ${remSpace[3]});
```

apps-rendering/src/adSlot.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+	// This class is applied if the article has fewer than 15 paragraphs.
+	&.short:nth-of-type(1) {
+		${from.desktop} {
+			top: 0;
+		}
+	}
+`;
+
+const adHeight = '258px';
```
mxdvl, line 105: Interesting to have this as a separate const. This could be an opportunity to explain where this value come from, especially the extra 8px more than an MPU.

apps-rendering/src/adSlot.tsx
```diff
@@ -0,0 +1,131 @@
...
+		key={`ad-after-${paragraph}-para`}
+	>
+		<div css={adLabelsStyles(format)} className="ad-labels">
+			<h1>Advertisement</h1>
+		</div>
+		<div css={adSlotStyles} className="ad-slot"></div>
+		<div css={supportBannerStyles(format)} className="support-banner">
+			<h1>Support the Guardian and enjoy the app ad-free.</h1>
+			<ThemeProvider theme={buttonThemeBrandAlt}>
+				<Button>Support the Guardian</Button>
```
mxdvl, line 125: How does this button know it's hooked to the right action?

apps-rendering/src/adSlot.tsx
```diff
@@ -0,0 +1,131 @@
...
+		margin-bottom: ${remSpace[6]};
+	}
+
+	${until.phablet} {
+		margin: 1em -${remSpace[3]};
+	}
+
+	// This class is applied in the nativeCommunication module
+	// to the first ad slot if Teads ads are enabled.
+	&.ad-slot-square {
```
mxdvl, line 89: This is a bill confusing as the ad slot has two ways of having a defined height. Either on the wrapper, as here, or on the actual slot, as below.

Could this be unified?

mxdvl (commented):
This is a neat refactor!

I've left some small comments and refrained from approving only because of my lack of understanding of how ads work in apps.

apps-rendering/src/adSlot.tsx
MarSavar: There's some code in `nativeCommunication.ts` which uses TypeScript Thrift definitions defined in [Bridget](https://github.com/guardian/bridget), the bridge layer between web views and native code. The code in question attaches `click` event listeners to elements with the class name `support-banner`, which allows the calling of `void acquisitionsClient.launchPurchaseScreen`. You can see the code here: https://github.com/guardian/dotcom-rendering/blob/ms-ar-adslot/apps-rendering/src/client/nativeCommunication.ts#L128-L136

apps-rendering/src/adSlot.tsx
MarSavar: Interesting question - I thought about inlining this value in the CSS but it was a separate const in the previous code so I kept it in the refactor. I'm not sure where the extra 8 pixels come from - perhaps to make sure the ad itself doesn't butt up to the support banner and the ad label? cc @JamieB-gu

apps-rendering/src/adSlot.tsx
JamieB-gu: `remSpace[6]`...?

common-rendering/src/editorialPalette/background.ts
```diff
@@ -22,6 +22,17 @@ import { Colour } from '.';
 
 // ----- Functions ----- //
 
+const adSlot = (format: ArticleFormat): Colour => {
+	if (format.design === ArticleDesign.Comment ||
+	format.design === ArticleDesign.Letter ||
+	format.design === ArticleDesign.Editorial) {
+		return neutral[86];
+	}
+	return  neutral[97];
```
JamieB-gu, line 31: Possible alternative, up to you whether you think it's more/less readable:

```suggestion
	switch (format.design) {
		case ArticleDesign.Comment:
		case ArticleDesign.Letter:
		case ArticleDesign.Editorial:
			return neutral[86];
		default:
			return  neutral[97];
	}
```

JamieB-gu (approved):
Nice! Pending Max's outstanding question, 👍 

This is a big improvement 🚀

apps-rendering/src/adSlot.tsx
MarSavar: Well spotted! This should actually live in the actual ad slot rather than the wrapper. I've moved it :)

apps-rendering/src/adSlot.tsx
```diff
@@ -0,0 +1,131 @@
...
+	paragraph: number;
+	format: ArticleFormat;
+};
+
+const adLabelsStyles = (format: ArticleFormat): SerializedStyles => css`
+	${textSans.xsmall()}
+	color: ${text.adLabel(format)};
+	padding: ${remSpace[3]};
+	float: left;
+	width: calc(100% - ${remSpace[6]});
```
mxdvl, line 30: I don’t know if there’s a contract about spaces growing linearly. If what you’re trying to capture is that there’s padding on both sides, and you’re already using `calc`, it would be safer to be explicit and let the css calculate twice whatever `remSpace[3]` is.

```suggestion
	// We need to account for padding on both sides.
	width: calc(100% - 2 * ${remSpace[3]});
```

mxdvl: Yes, that’s right, I knew that not all values were available–there’s no `space[8]`–so just wanted to make sure. Either is fine in this case!

```ts
export const space = {
	1: 4,
	2: 8,
	3: 12,
	4: 16,
	5: 20,
	6: 24,
	9: 36,
	12: 48,
	24: 96,
} as const;
```

PR #4765, opened 2022-04-27
Refactor reveal styles

oliverlloyd:
## What does this change?
We use similar `reveal` and `pending` classes in three different places in the code. This PR centralises this styling logic as globals

## Why?
This creates a contract that can be re-used

### DIFF
--- a/dotcom-rendering/src/web/components/Article.tsx
+++ b/dotcom-rendering/src/web/components/Article.tsx
@@ -49,7 +49,7 @@ export const Article = ({ CAPIArticle, NAV, format }: Props) => {
 						animation: ${keyframes`
 							0% { opacity: 0; }
 							100% { opacity: 1; }
-						`} 2s ease-out;
+						`} 1s ease-out;
 					}
 					.reveal-slowly {
 						animation: ${keyframes`
--- a/dotcom-rendering/src/web/components/Article.tsx
+++ b/dotcom-rendering/src/web/components/Article.tsx
@@ -1,5 +1,5 @@
 import { StrictMode } from 'react';
-import { Global, css } from '@emotion/react';
+import { Global, css, keyframes } from '@emotion/react';
 import { focusHalo, brandAlt, neutral } from '@guardian/source-foundations';
 import { ArticleDesign } from '@guardian/libs';
 import { filterABTestSwitches } from '../../model/enhance-switches';
@@ -42,6 +42,23 @@ export const Article = ({ CAPIArticle, NAV, format }: Props) => {
 					::selection {
 						background: ${brandAlt[400]};
 						color: ${neutral[7]};
+					}
+					/* We're using classnames here because we add and remove these classes
+	   				   using plain javascript */
+					.reveal {
+						animation: ${keyframes`
+							0% { opacity: 0; }
+							100% { opacity: 1; }
+						`} 2s ease-out;
+					}
+					.reveal-slowly {
+						animation: ${keyframes`
+							0% { opacity: 0; }
+							100% { opacity: 1; }
+						`} 4s ease-out;
+					}
+					.pending {
+						display: none;
 					}
 				`}
 			/>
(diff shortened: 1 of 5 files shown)


dotcom-rendering/src/web/components/Article.tsx
```diff
@@ -43,6 +43,23 @@ export const Article = ({ CAPIArticle, NAV, format }: Props) => {
...
+					}
+					.reveal-slowly {
+						animation: ${keyframes`
+							0% { opacity: 0; }
+							100% { opacity: 1; }
+						`} 4s ease-out;
+					}
+					.pending {
+						display: none;
+					}
```
OllysCoding, line 62: Worth noting this will only work on articles and not on fronts!

dotcom-rendering/src/web/layouts/Layout.stories.tsx
```diff
@@ -163,13 +163,20 @@ for (const [displayName] of Object.entries(ArticleDisplay)) {
...
+				ServerCAPI={{
+					...Fixtures.LiveBlog,
+					keyEvents: [],
+				}}
+			/>
+		);
+	})
+	.addParameters({
+		chromatic: {
+			diffThreshold: 0.2,
```
OllysCoding, line 179: What does diffThreshhold do?

OllysCoding:
I find this interesting but I also wonder if this is the right approach for re-usable styles!

Rather than using the global styling option provided by emotion` could we have a global file which exports emotion css objects? e.g

```js
// file: style-globals.ts
export const revealStyles = css`
    /* We're using classnames here because we add and remove these classes
    using plain javascript */
    .reveal {
	animation: ${keyframes`
		0% { opacity: 0; }
		100% { opacity: 1; }
	`} 1s ease-out;
    }
    .reveal-slowly {
	animation: ${keyframes`
		0% { opacity: 0; }
		100% { opacity: 1; }
	`} 4s ease-out;
    }
    .pending {
          display: none;
    }
`
```

And then we could import or use the styles in places where they're needed? 

The reason I'd suggest this change is because if I'm exploring the code and see `reveal` or `pending` classes in places, it I don't immediately get an idea of _why_ they're there or what they're doing, especially when DCR rarely uses classes.

To me this is also better because it fits better with the purpose of emotion, whereas global styles we mainly use when we want styling applied to HTML which is outside of our control (e.g from CAPI).

oliverlloyd:
> if I'm exploring the code and see `reveal` or `pending` classes in places, it I don't immediately get an idea of _why_ they're there or what they're doing - and I can't find out from clicking around in the file I'm in - especially confusing when DCR rarely uses class names

This makes a lot of sense. Let me refactor things

dotcom-rendering/src/web/layouts/Layout.stories.tsx
JamieB-gu: Here's an example of where this has been useful in the past @OllysCoding https://github.com/guardian/image-rendering/pull/137

OllysCoding: Great, thank you :)

OllysCoding (approved):
Looks good!

mxdvl:
Is there a way to do the following?

```ts
const reveal = css`
  @keyframe() /* etc. */
`

document.querySelector('div').classList.add(reveal)

```

oliverlloyd:
> > Is there a way to do the following?
> > ```ts
> > const reveal = css`
> >   @keyframe() /* etc. */
> > `
> > 
> > document.querySelector('div').classList.add(reveal)
> > ```
> 
> So you're saying instead of adding a classname of 'reveal' we add one called 'dcr-ky8fye' which is assigned to a variable called `reveal`?
> 
> I think this would be possible. I guess the benefit is that you don't have to have the wrapper styles. The downside is that maybe it reads less well? As in, defining the classnames in a more classic way is perhaps easier to reason about but I think if you are used to Emotion then this shouldn't be a big problem

Hmm, I started to do this refactor but ran into a problem when I hit this code

```
		const pendingElements = document?.querySelectorAll<HTMLElement>(
			'.discussion > .pending',
		);
```

Is there a way around this?

My inclination is that this is an indication that having the traditional class names is easier to work with but I could be missing something obvious (happens a lot!)

oliverlloyd:
As per our discussion offline, we will stick with the classic classnames approach for now

PR #4766, opened 2022-04-28
Fix the height of the merchandising containers

oliverlloyd:
## What does this change?
This adds a `fixedHeight` prop to `ElementContainer` allowing the height to be fixed and then uses it for the merchandising ad slots at the bottom of the page.

## Why?
This allows us to hold open a container at a fixed value ahead of any client content that might be loaded into it. For merchandising, this ameliorates the CLS caused by the advert loading in and shifting content on the page. The content that gets rendered can be of different heights so we do still get some layout shift but this is greatly reduced with this change.

## Why not?
This will now hold this space open permanently. If a reader is using an ad blocker they will be left with a grey box on the page.

And potentially, there are other circumstances where an unwanted grey box may remain - this is being investigated.

## 248?
This is the height of the narrowest content I saw but it is possible narrower content will be loaded in in which case the slot will remain at 248 pixels high, with grey space below

## Why not set the height on `AdSlot`?
I think that the commercial code is removing this div prior to rendering the ad which seems to cause a flash and subsequent CLS spike

### Scores
These CLS scores are for the particular area of the page where one of the slots is loaded

| Before      | After      |
|-------------|------------|
| bb | aa |

### DIFF
--- a/dotcom-rendering/src/web/components/AdSlot.tsx
+++ b/dotcom-rendering/src/web/components/AdSlot.tsx
@@ -453,6 +453,7 @@ export const AdSlot: React.FC<Props> = ({
 					css={[
 						css`
 							position: relative;
+							min-height: 332px;
 						`,
 						adStyles,
 						fluidFullWidthAdStyles,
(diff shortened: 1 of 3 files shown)


mxdvl:
Is it `248` or `332`?

oliverlloyd:
> Is it `248` or `332`?

Oops, I need to push changes for them to appear...

PR #4767, opened 2022-04-28
feat: Add support for fixed/small/slow-IV container type

AshCorr:
## What does this change?

Adds the `fixed/small/slow-IV` container type!

## Why?

More fronts!

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[after]: https://user-images.githubusercontent.com/21217225/165709602-af8e93f7-e70d-424f-baad-94bb0e4597aa.png
[before]: https://user-images.githubusercontent.com/21217225/165709711-213c6b78-ccb1-45a9-951e-249241ff086e.png

[before2]: https://user-images.githubusercontent.com/21217225/165711207-2653480c-c00b-45c8-ba89-66704cc9f755.png
[after2]: https://user-images.githubusercontent.com/21217225/165711792-0f04397f-ca2f-42f6-91ce-407b8d0641c7.png

### DIFF
--- a/dotcom-rendering/src/web/components/FixedSmallSlowIV.stories.tsx
+++ b/dotcom-rendering/src/web/components/FixedSmallSlowIV.stories.tsx
@@ -1,0 +1,37 @@
+import { breakpoints } from '@guardian/source-foundations';
+import { ContainerLayout } from './ContainerLayout';
+import { FixedSmallSlowIV } from './FixedSmallSlowIV';
+
+import { trails } from '../../../fixtures/manual/trails';
+
+export default {
+	component: FixedSmallSlowIV,
+	title: 'Components/FixedSmallSlowIV',
+	parameters: {
+		chromatic: {
+			viewports: [
+				breakpoints.mobile,
+				breakpoints.mobileMedium,
+				breakpoints.mobileLandscape,
+				breakpoints.phablet,
+				breakpoints.tablet,
+				breakpoints.desktop,
+				breakpoints.leftCol,
+				breakpoints.wide,
+			],
+		},
+	},
+};
+
+export const Default = () => (
+	<ContainerLayout
+		title="FixedSmallSlowIV"
+		showTopBorder={true}
+		sideBorders={true}
+		padContent={false}
+		centralBorder="partial"
+	>
+		<FixedSmallSlowIV trails={trails} />
+	</ContainerLayout>
+);
+Default.story = { name: 'FixedSmallSlowIV' };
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/FixedSmallSlowIV.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import { trails } from '../../../fixtures/manual/trails';
+
+export default {
+	component: FixedSmallSlowIV,
+	title: 'Components/FixedSmallSlowIV',
+	parameters: {
+		chromatic: {
+			viewports: [
+				breakpoints.mobile,
```
AshCorr, line 13: If we're planning on adding all these breakpoints for every type of container we probably need to find a way to give chromatic some kind of focus so that it only checks files that could have changed.

oliverlloyd: > If we're planning on adding all these breakpoints for every type of container we probably need to find a way to give chromatic some kind of focus so that it only checks files that could have changed.
> 
> This might be some help? https://www.chromatic.com/docs/turbosnap

Interesting. It's hard to say if this is a good option for us or not, I think the write up at the link you gave (thanks) is fairly accurate in being hesitant but if this does work the way we want it to then it could be a big win

AshCorr: Raised https://github.com/guardian/dotcom-rendering/pull/4768 for this, tried to explain the incompatibility the way I understand it in this PR

dotcom-rendering/src/web/components/FixedSmallSlowIV.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export const FixedSmallSlowIV = ({ trails }: Props) => {
+	const slicedTrails = trails.slice(0, 4);
+
+	return (
+		<UL direction="row">
+			{slicedTrails.map((trail, index) => {
+				return (
+					<LI
+						percentage="25%"
```
oliverlloyd, line 18: I think you can remove this line and it should default to evenly distributing the cards

```suggestion
```

oliverlloyd (approved):
Very cool

dotcom-rendering/src/web/components/FixedSmallSlowIV.tsx
AshCorr: Thanks!

OllysCoding (approved):
Looks good!!

PR #4771, opened 2022-04-28
Added `FEContainerPalette` & `DCRContainerPalette`

### LINKED ISSUE #4605: `FEContainerPaletteType`
Create a type that captures the variations of different [container palettes](https://docs.google.com/document/d/1nkntFJUunqrwNYszTrc2RBfyYUd1bEAeQh76yVFOLjo/edit#heading=h.n8dqpx8lx9rv) that are possible.

[Here's a document](https://docs.google.com/spreadsheets/d/1drj8NWFDmNaNGwg8nBAde3Cf5WNyK1SWl2QcUo_dGxE/edit#gid=0) details container usage

Container palettes are applied with tags on the container config

oliverlloyd:
## What does this change?
This adds the typings for the container palette

Closes #4605 

## Why?
So we have any palette associated with a container available to do things with


AshCorr (approved):
👍

dotcom-rendering/src/model/enhanceCollections.ts
```diff
@@ -44,6 +44,12 @@ const enhanceCards = (collections: FEFrontCard[]): DCRFrontCard[] =>
 			};
 		});
 
+const decideContainerPalette = (
+	metadata: { type: FEContainerPalette }[],
+): DCRContainerPalette => {
+	return metadata[0].type;
```
AshCorr, line 50: Would the idea be that we can return certain style properties from this function?

oliverlloyd: The idea is that we can take the array of tags (where there are currently conventions where you sometimes have two tags) and rationalise it into one value by putting all the magic incantations here, in one place

dotcom-rendering/src/model/enhanceCollections.ts
```diff
@@ -44,6 +44,12 @@ const enhanceCards = (collections: FEFrontCard[]): DCRFrontCard[] =>
 			};
 		});
 
+const decideContainerPalette = (
+	metadata?: { type: FEContainerPalette }[],
+): DCRContainerPalette | undefined => {
+	return metadata?.length && metadata[0].type ? metadata[0].type : undefined;
```
mxdvl, line 50: FYI, this also works, but maybe too terse?

```ts
return metadata?.[0]?.type; 
```

PR #4773, opened 2022-04-28
feat: Surface Lighthouse results

### LINKED ISSUE #4584: Surface Github Lighthouse numbers on PRs
- [ ] [Surface budgets](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/lighthouserc.js#L16-L24) on PRs. These are already [breaking](https://github.com/guardian/dotcom-rendering/runs/6177839941?check_suite_focus=true#step:6:47), but we aren't reporting them
- [ ] Surface change from `main` on ~PRs~ [this issue](#issuecomment-1113315505).
- [ ] Choose the pages we should be reporting on
  - We should be testing across as many features as we can
  - We should have a representative list of general content that we produce
  - We should make sure we don't negatively impact build speed too much

mxdvl:
## What does this change?

Surface Lighthouse results in all PRs. On merge to `main`, update the comment on #4584 to reflect the latest Lighthouse scores.

The script updating the comment is powered by Deno and Github’s official `@octokit`’s libraries, as well as types from the `lhci/cli` library.

## Why?

Closes #4584 

## Screenshots

image


mxdvl:
if we’re all happy with this I will squash & merge, because there’s lots of WIP commits in there.

bryophyta:
> if we’re all happy with this I will squash & merge, because there’s lots of WIP commits in there.

that's fine by me 👍

.github/workflows/lighthouse.yml
```diff
@@ -1,25 +1,40 @@
...
+      - uses: bahmutov/npm-install@v1
+      - run: make build
+        working-directory: dotcom-rendering
+      - name: Install and run Lighthouse CI
+        working-directory: dotcom-rendering
+        env:
+          LHCI_GITHUB_TOKEN: ${{ secrets.LHCI_GITHUB_TOKEN }}
+        run: |
+          npm install -g puppeteer-core@2.1.0 @lhci/cli@0.8.2
+          lhci autorun --assert.includePassedAssertions
```
bryophyta, line 30: ah, so before it was only rendering results for the failed budgets? that makes sense!

fwiw it looks as though we can include this flag in the [lighthouserc.js](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/lighthouserc.js) config instead, if we want to (see https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/configuration.md#includepassedassertions). I don't think it makes a huge difference either way but might be good to keep all of the config in one place?

scripts/deno/surface-lighthouse-results.ts
```diff
@@ -0,0 +1,206 @@
...
+ * on which to add or update the Lighthouse report
+ *
+ * - on a "push" event: the original issue to track the latest report:
+ * https://github.com/guardian/dotcom-rendering/issues/4584
+ */
+const issue_number = isPullRequestEvent(payload)
+	? // If PullRequestEvent
+	  payload.pull_request.number
+	: // If PushEvent
+	  4584;
```
jlieb10, line 42: What's this hard-coded number here?

scripts/deno/surface-lighthouse-results.ts
```diff
@@ -0,0 +1,206 @@
...
+
+console.log(`Using issue #${issue_number}`);
+
+console.log({ payload });
```
jlieb10, line 46: Is this for debugging? Or do we want to see this logged out in the GH Action?

scripts/deno/surface-lighthouse-results.ts
```diff
@@ -0,0 +1,206 @@
...
+/* -- Methods -- */
+
+/** If large number, round to 0 decimal, if small, round to 6 decimal points */
+const formatNumber = (expected: number, actual: number): string =>
+	expected > 100 ? Math.ceil(actual).toString() : actual.toFixed(6);
+
+const getStatus = (
+	passed: boolean,
+	level: AssertionResult["level"]
+): "✅" | "⚠️" | "❌" => {
```
AshCorr, line 106: Theres something interesting about a function that has an "emoji" return type, love it!

jlieb10:
> if we’re all happy with this I will squash & merge, because there’s lots of WIP commits in there.

Yes I sometimes do a WIP after pairing if a task hasn't reached its endpoint but I need to sign off, so the branch can be pulled by whoever else is working on it. Big fan of squashing when I merge.

scripts/deno/surface-lighthouse-results.ts
```diff
@@ -0,0 +1,206 @@
...
+
+console.log(`Using issue #${issue_number}`);
+
+console.log({ payload });
+
+/** The Lighthouse results directory  */
+const dir = "dotcom-rendering/.lighthouseci";
+
+const links: Record<string, string> = JSON.parse(
+	Deno.readTextFileSync(`${dir}/links.json`)
```
jlieb10, line 52: great

scripts/deno/surface-lighthouse-results.ts
```diff
@@ -0,0 +1,206 @@
...
+
+/** The string to search for when looking for a comment */
+const REPORT_TITLE = "⚡️ Lighthouse report";
+const GIHUB_PARAMS = {
+	owner: "guardian",
+	repo: "dotcom-rendering",
+	issue_number,
+} as const;
+
+const octokit = new Octokit({ auth: token }) as {
```
jlieb10, line 87: what is happening after the `as` ?

AshCorr (approved):
This is awesome! This will help a lot!

scripts/deno/surface-lighthouse-results.ts
AshCorr: I believe its the original issue # https://github.com/guardian/dotcom-rendering/issues/4584, the action updates the original issue when a PR is pushed to main

AshCorr: Would be interesting if we could update the repo's README and surface the numbers there instead?

scripts/deno/surface-lighthouse-results.ts
mxdvl: Debugging only!

scripts/deno/surface-lighthouse-results.ts
mxdvl: I’ve refactored a bit to make it clearer that we’re building the type ourselves.

scripts/deno/surface-lighthouse-results.ts
jlieb10: We're going to re-visit this (great) idea in a separate Spike

mxdvl:
A bit cheeky, usually this would be for others’s PRs, but I’ll nominate this as a hidden gem 💎 .

https://github.com/guardian/playground/issues/14

PR #4775, opened 2022-04-28
Stop dependabot from rebasing its open PRs every time `main` updates

sndrs:
## why?
following on from the $10-per-build cost @AshCorr identified in #4768, i had a look in chromatic usage stats. for april this year, dotcom accounts for entire excess (i.e. all other repos cover the allowance by themselves)

image

at the time of writing, there are 26 open dependabot PRs open in this repo. every time `main` changes, dependabot will rebase those PRs triggering a full set of chromatic tests. 

this means every merge to `main` currently costs us 26 x $10 = $260!

unless dependabot PRs are about to be merged, those tests don't need to be run right then. this PR will stop dependabot from doing that.

to make this safe, it would probably be a good idea enable [Require branches to be up to date before merging](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-status-checks-before-merging). this should give you the same confidence merging dependabot PRs as before, but without the multiple chromatic tests being run before the PR is ready to merge

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,12 +9,17 @@ updates:
       directory: '/'
       schedule:
           interval: 'daily'
+      # Minimise the number of chromatic runs.
+      # We've enabled 'Require branches to be up to date before merging',
+      # so you won't be able to merge a dependabot PR unless you rebase manually anyway.
+      rebase-strategy: "disabled"
     - package-ecosystem: 'npm'
       directory: '/apps-rendering'
       schedule:
           interval: 'daily'
       labels:
         - "AR Dependency"
+      rebase-strategy: "disabled"
     - package-ecosystem: "npm" 
       directory: "/dotcom-rendering"
       schedule:
@@ -22,7 +27,9 @@ updates:
       labels:
         - "DCR Dependency"
       open-pull-requests-limit: 10
+      rebase-strategy: "disabled"
     - package-ecosystem: 'github-actions'
       directory: '/'
       schedule:
           interval: 'daily'
+      rebase-strategy: "disabled"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,9 +9,7 @@ updates:
       directory: '/'
       schedule:
           interval: 'daily'
-      # Minimise the number of chromatic runs.
-      # We've enabled 'Require branches to be up to date before merging',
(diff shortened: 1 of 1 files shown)


OllysCoding (approved):
+1

AshCorr (approved):
Change is good, but I'm a little bit worried this is going to make merging dependabot PR's even rarer than it already is. Would this be adding another step to merging a dependabot PR by having to manually tell dependabot to rebase and then wait for the tests to complete?

jamesgorrie (commented):
Another knock on effect this would have is that it applies to all PRs to be up to date, so our feature branches would also always have to be up to date, which I know I merge branches that aren't.

@OllysCoding @AshCorr @mxdvl are we happy with this?

sndrs:
> Would this be adding another step to merging a dependabot PR by having to manually tell dependabot to rebase and then wait for the tests to complete?

it would yes - the tests would run against the state of `main` when it first opens but not again until you manually kick them off when you're ready to merge. 

you can do that with the comment instruction (`@dependabot rebase`) or if you enable the requirement to make sure branches are up to date with main before merging, using the button github adds:

image

it's worth considering that that requirement is currently disabled, meaning all non-dependabot PRs only have checks run against the state of `main` at the point they run, and therefore `main` may not look like it did by the time you merge the PR.

if this hasn't been causing issues you may be fine to treat dependabot PRs the same way (i.e. as long as a test passed in the past it's considered safe enough to merge)

mxdvl:
> if this hasn't been causing issues you may be fine to treat dependabot PRs the same way (i.e. as long as a test passed in the past it's considered safe enough to merge)

Yes, I think this is what we would want. The same way that most of our merges are out-of-date with `main`.

ashishpuliyel:
There seems to be consensus we can merge this, without changing the setting that requires branches to be up-to-date.
@sndrs Make it so?

ashishpuliyel (approved):
I took the liberty of doing the thing I said you'd need to do. So the comment no longer references the setting

sndrs:
thanks @ashishpuliyel – @guardian/dotcom-platform do i need to worry about the chromatic changes? wouldn't have expected them?

sndrs:
Thanks @oliverlloyd

PR #4776, opened 2022-04-28
Only run chromatic tests when a review is requested

sndrs:
following on from #4775, this PR will stop chromatic tests from running unless a PR has a review requested.

the codeowners will automatically add a review request for most new PRs. but any subsequent pushing the branch will _also_ trigger chromatic tests, even if the PR is not ready.

this PR will remove the pass check for any previous chromatic runs if new commits are added, meaning that once you're happy the PR is good to go, you can formally (re)request a review and chromatic will kick off again.

this repo already requires an approval to merge, so it shouldn't change the PR workflow too much but could save a good few chromatic snapshots 🙏 

it might be worth enabling 
image 
on the repo too if you decide to merge this

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,6 +1,10 @@
 name: Chromatic 👓
 on: push
-
+  branches:
+    - main
+  pull_request:
+    types: ['review_requested']
+    
 jobs:
   chromatic:
     name: Chromatic


OllysCoding:
> this PR will remove the pass check for any previous chromatic runs if new commits are added, meaning that once you're happy the PR is good to go, you can formally (re)request a review and chromatic will kick off again.

I would love it if there was a better method for this flow. I understand not running chromatic checks until I've requested a review (though this does make it harder for me quickly understand the changes my PR has made) - but it's quite normal for people to push up further commits fixing things even after approvals - having to re-request reviews in order for the checks to pass doesn't really make sense to me. 

It'd be nice if any commits after I've requested reviews automatically re-ran chromatic. The value that it adds to our project is immeasurable and the flow of its use is just as important as what it does.

sndrs:
> I would love it if there was a better method for this flow

that's totally fair enough – if the team feel it doesn't work then please do close it

sndrs:
fwiw, we also disabled this behaviour in https://github.com/guardian/source because no one is in the habit of re-requesting a lot of the time and it became obstructive

PR #4788, opened 2022-05-02
Choose the right cards

oliverlloyd:
## What does this change?
This PR changes the cards that we pick for the different sections of `DynamicSlow`

## Why?
We were using the last two, not the first two

| Before      | After      |
|-------------|------------|
| Screenshot 2022-05-02 at 10 00 33| Screenshot 2022-05-02 at 10 00 08|


AshCorr (approved):
👍

PR #4789, opened 2022-05-02
Hide sublinks on small cards

oliverlloyd:
## What does this change?
Hides siblinks on small cards that don't have enough space to show them

## Why?
This is the convention in place at the moment and we want to respect this, even if the card has `supportingContent` added to it

| Before      | After      |
|-------------|------------|
| Screenshot 2022-05-02 at 10 01 43 | Screenshot 2022-05-02 at 10 02 52|


jamesgorrie (approved):
Great. I do wonder how we say to future devs "This was intentional, not forgotten about because X"... but maybe when someone adds it back and it looks bad, that's enough.

Issue #4795, opened 2022-05-02
`Card` border width changes based on palette


oliverlloyd:
See: https://github.com/guardian/interactive-atom-container-colours/blob/master/shared/css/_variables.scss#L7

VDuczekW:
to be a rota bug ticket @ParisaTork please include when closing previous palette epic

Issue #4794, opened 2022-05-02
Decide hover styling based on palette tags


oliverlloyd:

alinaboghiu:
This is the same issue as: https://github.com/guardian/dotcom-rendering/issues/7951

I think

rhiannareechaye:
Closing as it's the same issue https://github.com/guardian/dotcom-rendering/issues/7951

Issue #4792, opened 2022-05-02
`verticalDivider` should allow specific colours


oliverlloyd:
We need to be able to chose specific colours when the background colour for a container changes in response to a palette tag

AshCorr:
Believe this is done in #6248

Issue #4791, opened 2022-05-02
Support 'Dynamo'


oliverlloyd:

jamesgorrie:
This is supported and work continuing https://github.com/guardian/dotcom-rendering/issues/6381

PR #4796, opened 2022-05-02
`FrontPalette`

### LINKED ISSUE #4606: `FEContainerPalette` is used to style containers

oliverlloyd:
## What does this change?
This PR adds support for using container palettes. Closes #4606 

```ts
type FEContainerPalette =
	| `EventPalette`
	| `SombreAltPalette`
	| `EventAltPalette`
	| `InvestigationPalette`
	| `LongRunningAltPalette`
	| `LongRunningPalette`
	| `SombrePalette`
	| `BreakingPalette`;
```

Specifically, it introduces a pattern for overrinding the colours used by `ElementContainer` and `Card` when a container has been given a [palette tag in the fronts config tool](https://docs.google.com/document/d/1GBooDxJWhxKh5z8nT1qCEqMHuXOwsAmuenLXdCzCyT8/edit)

## What are palettes?
Palettes are a way to change the look of a container on front pages. When a palette tag is applied then both the container itself and the cards within it have their default styles overridden. 

Previously, when [discussing approaches to how we would model this](https://docs.google.com/document/d/1nkntFJUunqrwNYszTrc2RBfyYUd1bEAeQh76yVFOLjo/edit#heading=h.n8dqpx8lx9rv) we considered passing in the `containerType` into our existing `decidePalette` function and having this use that context to make the choices for which colours to use. But this PR takes a slightly different approach. Here we have created a new `decideFrontPalette` function

```ts
export const decideFrontPalette = (
	containerPalette: DCRContainerPalette,
): FrontPalette => {
         return // An object with the overrides values
}
```

The advantage to this approach is we have all the logic in one self-contained location. This nicely reflects [the approach taken in Frontend](https://github.com/guardian/interactive-atom-container-colours/blob/master/shared/css/_variables.scss). New palettes can be created and defined here without needing lots of changes in other areas of the code.

## What's pending?
The following areas are still pending:

- Styling for sub links on cards - https://github.com/guardian/dotcom-rendering/issues/4790
- The Dynamo - https://github.com/guardian/dotcom-rendering/issues/4791
- Setting a specific colour for `UL` & `LI` borders - https://github.com/guardian/dotcom-rendering/issues/4792
- Making the headline bold when some or all palettes are in use - https://github.com/guardian/dotcom-rendering/issues/4793
- Hover styles with a palette - https://github.com/guardian/dotcom-rendering/issues/4794
- The width of a `Card`s border can change based on palette - https://github.com/guardian/dotcom-rendering/issues/4795


dotcom-rendering/index.d.ts
```diff
@@ -620,21 +635,13 @@ type FEContainerType =
...
 	| `LongRunningAltPalette`
 	| `LongRunningPalette`
 	| `SombrePalette`
-	| `Canonical`
-	| `Dynamo`
-	| `Special`
-	| `DynamoLike`
-	| `Special`
-	| `Breaking`
-	| `Podcast`
```
oliverlloyd, line 637: After looking into this area more I see we don't support these other tags as 'palettes' so they don't need to exist here

AshCorr (approved):
Woah this came together quickly, it looks amazing!

dotcom-rendering/index.d.ts
```diff
@@ -620,21 +635,13 @@ type FEContainerType =
 	| 'news/most-popular';
 
 type FEContainerPalette =
-	| `Branded`
```
mxdvl, line 623: I think we definitely need to keep `Branded` for the Labs container!

image

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -19,6 +21,10 @@ export const DynamicFast = ({ trails }: Props) => {
 			<UL direction="row" padBottom={true}>
 				<LI padSides={true} percentage="75%">
 					<Card
+						frontPalette={
+							containerPalette &&
+							decideFrontPalette(containerPalette)
+						}
```
mxdvl, line 27: Worth extracting as a `const` at the top of this component?

mxdvl (commented):
We need to keep `Branded` as a potential `containerPalette`.

dotcom-rendering/index.d.ts
oliverlloyd: I think that there will defo be other 'non palette' tags that we will eventually bring into this pattern. Branded seems like a good example. But right now that isn't included as a, er, classic palette, it's a palette like thing, but doesn't exist in [this file](https://github.com/guardian/interactive-atom-container-colours/blob/3733462f6a475dfa49c28320041633d8b426aaaa/shared/css/_variables.scss).

I think for these types of tags we probably need to do an investigation into each one and include them as separate PRs. Those PRs may or may not end up with us adding a new 'DCR' palette type. Branding looks like it could end up there but there's a chance we learn something unexpected. Eg. That container has a lot of additional use cases that we don't support right now, so maybe we end up creating a new container component entirely? I'm advocating for this but using it as an example for how we want to dig into these other types more before committing them to a pattern

dotcom-rendering/src/web/components/Byline.tsx
```diff
@@ -7,6 +7,7 @@ import { decidePalette } from '../lib/decidePalette';
 type Props = {
 	text: string;
 	format: ArticleFormat;
+	palette?: Palette;
```
OllysCoding, line 10: What's the purpose of passing around Palette here? Is this to do with sublinks styles?

OllysCoding: maybe passing around some kind of `format` and `containerFormat` props, then you can do something like `const palette = decidePalette(containerFormat || format)`

I don't know if I've looked at this PR in enough detail yet, but accepting that we maybe be reaching a point where a card may have to chose from multiple formats/palettes, getting the devx of this right is quite important!

Looking forward to seeing what we can come up with :)

dotcom-rendering/index.d.ts
mxdvl: Interesting… That’s an entirely valid reason to think about this edge case in more details. I think that having a way to style containers will be needed for at `Branded` and `Video`, unless the latter is an altogether different Container type.

In `frontend`, here’s what I found about the supported container types for the `Labs` container: https://cs.github.com/guardian/frontend/blob/09669f922da66350190fff5e44d0dda968db3a81/common/app/views/fragments/commercial/containers/paidContainer.scala.html?q=org%3Aguardian+labs-container#L30-L43

dotcom-rendering/src/web/components/Byline.tsx
oliverlloyd: Well, in the end that change made a lot of sense, thanks for seeing it!

I've set things up so we're passing around `format` and `containerPalette` as the two 'types' which define the context of what is being rendered and then we expect the components to decide which palette to use. This is more declarative and seems to follow the existing patterns we've established.

However, it's a lot of changes being added to this PR but hopefully commit by commit it's manageable?

oliverlloyd: Note. I'm very aware that `containerPalette` as a name doesn't work here. We can change it here in this PR but it might be easier as a follow on

mxdvl (commented):
An alternative to making `decidePalette` take in an optional `containerPalette` parameter is to use define the palette by spreading the `decideFrontPalette` after `decidePalette`, like so:

```ts
// where containerPalette is either `DCRContainerPalette` or `undefined` 

const palette = {
	...decidePalette(format),
	...(containerPalette ? decideFrontPalette(containerPalette) : {}),
};
```

dotcom-rendering/src/web/components/DynamicFast.tsx
oliverlloyd: I've refactored away from this pattern now in favour of passing the type down

dotcom-rendering/index.d.ts
oliverlloyd: It's interesting that the container types would be restricted here, in the rendering layer. I would say these sorts of things should be part of the Fronts config tool. Ideally, the rendering layer would have minimal business logic and just do what is asked of it 🤔

mxdvl:
> It just reads easy and the override logic happens in one place. If we use spreading locally then it's hard to know when a property might be overridden or even which ones

I wasn’t trying to be particularly clever with the object spreading. We know exactly which properties are overridden: it’s all of the ones in `frontPalette`. The logic does not happen in one place, it happens in many different methods. I think `decidePalette` is not the same as `decidePaletteWithOverrides` and that we’re actually making `decidePalette` harder to reason about by having optional parameters passed through. That’s what I thought about as I first approached this code.

I don’t think the override method should spread into internal methods such as `borderLines` and if spreading or a `decidePaletteWithOverrides` is really off the table, I think we can at least restrict it to `decidePalette` like so, keeping the information of which keys are likely to be replaced:

```ts
export const decidePalette = (format: ArticleFormat, frontPalette?: FrontPalette): Palette => {
	return {
		text: {
			cardHeadline: frontPalette?.cardHeadline ?? textCardHeadline(format),
			sectionTitle: textSectionTitle(format),
			/* … */
		},
		/* … */
		hover: {
			headlineByline: hoverHeadlineByline(format),
			standfirstLink: frontPalette?.standfirstLink ?? hoverStandfirstLink(format),
		},
	};
};
```

oliverlloyd:
@mxdvl I spent some more time refactoring this based on your (and @OllysCoding 's) feedback. I've brought how we create the container palette object closer to how things work already, so we're not using a similar structure.

I also brought the construction of the object into `decidePalette` and called it `overrides` as I think this is a better representation for what it is.

I had a look at using spreading but actually it doesn't work that well. The issue is that whilst some of the properties on the overrides object are the same as the article Palette, some are not and so you can't really spread one into the other. I also think it makes more sense to be more explicit here, so that when an override happens it happens alongside the point where the main colour decision is made, rather than later on in a subsequent step. The advantage there is that it is clear when a particular value could be overridden and when it can't.

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -1212,12 +1184,14 @@ export const decidePalette = (
 			syndicationButton: textSyndicationButton(format),
 			articleLink: textArticleLink(format),
 			articleLinkHover: textArticleLinkHover(format),
-			cardHeadline: textCardHeadline(format, overrides),
-			cardByline: textCardByline(format, overrides),
-			cardKicker: textCardKicker(format, overrides),
+			cardHeadline:
```
oliverlloyd, line 1187: Ahh! 😻

dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -1212,12 +1184,14 @@ export const decidePalette = (
...
+				overrides?.text.cardHeadline || textCardHeadline(format),
+			cardByline: overrides?.text.cardByline || textCardByline(format),
```
oliverlloyd, line 1189: Ahh! 😻

dotcom-rendering/index.d.ts
```diff
@@ -199,6 +199,7 @@ type Palette = {
 type ContainerOverrides = {
 	text: {
 		cardHeadline: Colour;
+		cardStandfirst: Colour;
```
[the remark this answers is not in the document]
oliverlloyd: Oh, I see this was already called this - sorry, missed that!

Well, at some point we should rename this but it doesn't have to be here in this PR

oliverlloyd:
Thanks all for some really useful and insightful comments. I think this code is now a lot better thanks to all the back and forth.

PR #4797, opened 2022-05-02
Fix `inline1ContainerSizing` float right CSS

arelra:
## What does this change?

In the previous PR: #4506 we introduced an AB test to _not_ float right inline1 slots and take up the column full width.

However because of a concurrent change to add new ad slot containers, the _variant_ float right behaviour of inline2+ was not applied correctly to the new containers. This PR fixes that.

The switch for the test has been off so there has been no user impact.

ps - some prettier noise so it's easier to look at the commits diffs.

 ## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]:https://user-images.githubusercontent.com/7014230/166289930-4fe6801f-ca57-4a30-83e1-a12d24cebb25.png

[after]:https://user-images.githubusercontent.com/7014230/166289950-6d478d77-48d7-41db-a64b-5ec6d48d0d93.png

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleContainer.tsx
+++ b/dotcom-rendering/src/web/components/ArticleContainer.tsx
@@ -107,6 +107,18 @@ const articleAdStyles = (isInline1ContainerSizingVariant: boolean) => {
 				width: 100%;
 			}
 		}
+		/** ad-slot-offset-right is only added to inline2+ slots */
+		.ad-slot--offset-right {
+			${from.desktop} {
+				float: right;
+				width: auto;
+				margin-right: -318px;
+			}
+
+			${from.wide} {
+				margin-right: -398px;
+			}
+		}
 	`;
 	const controlCSS = css`
 		.ad-slot--inline,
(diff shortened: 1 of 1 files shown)


simonbyford (approved):
Looks good to me 🚀 

As an aside I find this setting really helpful for whitespacey PRs!

Screenshot 2022-05-03 at 09 56 32

dotcom-rendering/src/web/components/ArticleContainer.tsx
```diff
@@ -217,7 +203,8 @@ export const ArticleContainer = ({ children, format, abTests }: Props) => {
 			css={[
 				articleWrapper,
 				articleWidth(format),
-				articleAdStyles(isInline1ContainerSizingVariant),
+				adSharedCSS,
+				isInline1ContainerSizingVariant ? adVariantCSS : adControlCSS,
```
mxdvl, line 207: Nice pattern 😉

mxdvl (approved):

dotcom-rendering/src/web/components/ArticleContainer.tsx
arelra: Thank you for the suggestion!

PR #4798, opened 2022-05-02
Nested links

### LINKED ISSUE #4706: Nested links on Cards
Cards are themselves one big a tag but they also contain sub links that should also be a tags

See: https://css-tricks.com/nested-links/ for suggestions on how to achieve this

oliverlloyd:
## What does this change?
This PR fixes #4706 by adding support for nested links on cards.

We're using a similar approach to that used on Frontend where we have a faux link sitting outside the card which is absolutely positioned across the card surface. The nested sublinks are then brought to the top using z-index.

You can read more about this css pattern in [this great article on CSS Tricks](https://css-tricks.com/nested-links/)

## Why?
Having sublinks inside cards is an established pattern and one that we want to continue to support.


dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -236,78 +236,54 @@ export const Card = ({
...
+								}
+								showSlash={
+									format.design === ArticleDesign.LiveBlog ||
+									showSlash
+								}
+								byline={byline}
+								showByline={showByline}
+							/>
+						</HeadlineWrapper>
+						<>
```
AshCorr, line 284: Is this fragment necessary?

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -317,18 +293,40 @@ export const Card = ({
...
-					</ContentWrapper>
-				</CardLayout>
-				{/* If there are more than two sublinks break footer out of the headline column into a row below */}
-				{moreThanTwoSubLinks ? renderFooter({}) : <></>}
-			</CardWrapper>
-		</CardLink>
+						</>
+					</Flex>
+					<div>
+						{trailText && (
```
AshCorr, line 299: Could be wrong but I think the trailtext should be hidden when there are nested links... I'm not sure if this is decided by the layout or if its the same for all cards regardless of layout.

AshCorr (approved):
Thank you!

dotcom-rendering/src/web/components/Card/Card.tsx
oliverlloyd: You do sometimes get both together
Screenshot 2022-05-03 at 09 50 15
I don't have a complete picture for the rules yet but they do seem to be layout specific. Indeed it looks like specific cards in certain positions have their own rules about what appears in them

AshCorr: Ach, so complicated!

dotcom-rendering/src/web/components/Card/Card.tsx
oliverlloyd: Good spot. Fixing!

dotcom-rendering/src/web/components/Card/components/CardLink.tsx
```diff
@@ -1,99 +1,23 @@
...
-	children,
-	linkTo,
-	format,
-	dataLinkName = 'article',
-}: Props) => (
-	<a href={linkTo} css={linkStyles(format)} data-link-name={dataLinkName}>
-		{children}
+export const CardLink = ({ linkTo, dataLinkName = 'article' }: Props) => (
+	<a href={linkTo} css={fauxLinkStyles} data-link-name={dataLinkName}>
+		{linkTo}
```
mxdvl, line 21: Was it intentional to have the link content here?

dotcom-rendering/src/web/components/Card/components/CardLink.tsx
```diff
@@ -1,99 +1,23 @@
...
-				}
-			`;
-	}
-};
+import { getZIndex } from '../../../lib/getZIndex';
+
+const fauxLinkStyles = css`
+	position: absolute;
+	${getZIndex('card-link')};
+	opacity: 0;
```
mxdvl, line 7: Why make it transparent?

mxdvl (commented):
Two small questions that will inform an upcoming PR.

dotcom-rendering/src/web/components/Card/components/CardLink.tsx
oliverlloyd: It was. I'm taking inspiration from Frontend here but I think a link isn't valid if it has no content. That said, I think this is for accessibility reasons and we still need to review our solution for accessibility.

mxdvl: Lighthouse didn't care so we removed it in #xxx

PR #4801, opened 2022-05-03
Import ad size mappings from comm-core

Jakeii:
## What does this change?

Move ad size mappings to `@guardian/commercial-core` and import from there.

I think next (or instead?) the size mappings could be imported directly into frontend instead of passing via data-attributes but I'm not sure.

## Why?

Consolidate configuration into commercial core.


arelra (approved):
Nice! As long as this has been verified locally or on CODE all good to go 👍

PR #4807, opened 2022-05-03
feat(github): add octokit abstraction

mxdvl:
## What does this change?

Improve the Ophan Components tracker info by creating an `Octokit` abstraction.

## Why?

We learned from the work in #4773


bryophyta (approved):
This looks good to me; helpful to have the octokit instance with typing as something easily importable 👍

PR #4809, opened 2022-05-03
`decideSubLinkFormat`

oliverlloyd:
## What does this change?
This PR adds the function `decideSubLinkFormat` which is where we make the decision about which `format` property to use for sublinks.

## Why?
On fronts, a Card can have sublinks along the bottom. In some situations we want to style these sublinks according to the container that they sit in. A good example of this is where the container is for a liveblog, with the red background. In this case the sublink text should be white, regardless of the format of the article they link to. But if the container format is a simple news article and the sublink points to a lifestyle article then we would not want to style the link using the news tone, that would be inaccurate.

This variation in design logic is encompassed in this function.

| Before      | After      |
|-------------|------------|
| Screenshot 2022-05-03 at 16 14 27 | Screenshot 2022-05-03 at 16 13 53 |

### DIFF
--- a/dotcom-rendering/src/model/enhanceCollections.ts
+++ b/dotcom-rendering/src/model/enhanceCollections.ts
@@ -1,19 +1,65 @@
+import { ArticleSpecial, ArticleDesign } from '@guardian/libs';
 import { decideFormat } from '../web/lib/decideFormat';
+
+/**
+ *
+ * This function makes the decision about when we use the parent's (the
+ * container's) format property to override the styling of a sublink, and
+ * when we allow the sublink to express its own styling.
+ *
+ * Eg. If you had a sublink to a lifestyle article, when should it use pink for
+ * the kicker and when would that not look right
+ *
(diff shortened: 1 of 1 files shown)


oliverlloyd:
@HarryFischer this might be of interest to you?

dotcom-rendering/src/model/enhanceCollections.ts
```diff
@@ -1,19 +1,65 @@
...
+ * the kicker and when would that not look right
+ *
+ * @returns the format property that we will use to style the sublink
+ */
+const decideSubLinkFormat = ({
+	linkFormat,
+	containerFormat,
+	containerPalette,
+}: {
+	linkFormat?: CAPIFormat;
```
OllysCoding, line 20: Might not be necessary for this PR, but should we add a front enhancer which coverts the `card.format` from `CAPIFormat` to `ArticleFormat`? We should avoid having two different format types floating around the rendering layers if possible

PR #4817, opened 2022-05-03
fix: Wait until fonts have loaded before rendering storybooks

AshCorr:
## What does this change?

Force chromatic to wait for fonts to be loaded before taking a screenshot.

## Why?

This is also a bit of a punt and may or may not do anything at all. 

When I was looking at the footer https://5dfcbf3012392c0020e7140b-ubkpogwarw.chromatic.com/?path=/story/layouts-format-variations-immersive-interactive--labs of this storybook I noticed the "Obituary" link would overflow to the next line on first page load but on subsequent loads it would render without overflowing. By clearing the cache I was able to make this happen consistently.

I couldn't really find any difference in CSS or DOM for that particular section of the page between the two page loads, so it seemed like it could potentially be a seperate issue. Furthermore, blocking all fronts from loading seemed to make the footer load consistently (although not correctly).

Chromatic does apparently  have some issues with fronts causing false negatives according to this page https://www.chromatic.com/docs/resource-loading and this is one of their suggested solutions.

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -61,6 +61,13 @@ window.guardian = {
 }
 setCookie({name:'bwid', value: 'mockBrowserId'});
 
+// See https://www.chromatic.com/docs/resource-loading
+// Seems like chromatic isn't super great at loading font's correctly.
+// This might help reduce the false positives we get when running chromatic tests.
+const fontLoader = async () => ({
+	fonts: await document.fonts.ready,
+});
+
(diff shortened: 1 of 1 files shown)


AshCorr:
The change detected by Chromatic on this PR seems consistently broken now compared to the baseline which breaks on reload, not sure if thats better or worse 😅

AshCorr:
Nope, still inconsistent.

PR #4818, opened 2022-05-04
Footer position

oliverlloyd:
## What does this change?
We no longer position the footer in the same column as the headline and trail text if the image is positioned to the left or right

## Why?
The footer should go full width in these circumstances

| Before      | After      |
|-------------|------------|
| Screenshot 2022-05-04 at 09 18 14 | Screenshot 2022-05-04 at 09 15 55 |

### DIFF
--- a/dotcom-rendering/src/web/components/Card/components/CardFooter.tsx
+++ b/dotcom-rendering/src/web/components/Card/components/CardFooter.tsx
@@ -48,7 +48,7 @@ export const CardFooter = ({
 	) {
 		return (
 			<footer>
-				<div>{supportingContent}</div>
+				{supportingContent}
 				<div css={spaceBetween}>
 					{age}
 					<StraightLines
@@ -64,7 +64,7 @@ export const CardFooter = ({
 	if (format.design === ArticleDesign.Media) {
 		return (
 			<footer>
-				<div>{supportingContent}</div>
+				{supportingContent}
 				<div css={spaceBetween}>
 					{mediaMeta}
 					{/* Show age if we have it otherwise try for commentCount */}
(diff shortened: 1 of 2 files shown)


mxdvl (approved):
Perfectly fine.

Is this conditional rendering of the footer a hint that we might want to revisit using a grid layout for `Card`s?

oliverlloyd:
> Perfectly fine.
> 
> Is this conditional rendering of the footer a hint that we might want to revisit using a grid layout for `Card`s?

I'm not sure if the existence of conditional rendering is justification for using grid 🙃 . I think there are lots of pros and cons here but perhaps the biggest pro for flex vs. grid is we are already using flex and it would take a lot of time to refactor to grid.

mxdvl:
> > Is this conditional rendering of the footer a hint that we might want to revisit using a grid layout for `Card`s?
> 
> I'm not sure if the existence of conditional rendering is justification for using grid 🙃 . I think there are lots of pros and cons here but perhaps the biggest pro for flex vs. grid is we are already using flex and it would take a lot of time to refactor to grid.

That’s a _big pro_.

dotcom-rendering/src/web/components/Card/components/CardFooter.tsx
```diff
@@ -48,7 +48,7 @@ export const CardFooter = ({
 	) {
 		return (
 			<footer>
-				<div>{supportingContent}</div>
```
OllysCoding, line 51: Any reason for removing this div?

OllysCoding (approved):
LGTM!

dotcom-rendering/src/web/components/Card/components/CardFooter.tsx
oliverlloyd: Just a drive by. One less div!

Issue #4662, opened 2022-04-19
Data Parity for Fronts


mxdvl:
1. What metrics are we currently sending from fronts?
2. Are these all reproducible for DCR?
3. What else would we like to add to DCR in terms of metrics/observability?

mxdvl:
Tracking of Ophan components:
- #4692 
- #4698

PR #4827, opened 2022-05-04
Add `data-link-name` to Front Cards

mxdvl:
## What does this change?

All trails now have an optional `dataLinkName` key that may contain the Ophan component attribute.

A new helper method `getDataLinkNameCard` helps retrieve a `data-link-name` attribute for a card given its format, group and index.

## Why?

`data-link-name` attributes are missing on Front cards.

Part of #4692 & #4662 

## Screenshots

Using the [Ophan component bookarklet](https://gist.github.com/mxdvl/b17f344678ed34d45683896e208ecdc1).

image


jlieb10 (approved):
👍

Issue #4826, opened 2022-05-04
Support blockquotes in DCR `CommentBlockElement`


OllysCoding:
To support opinion liveblogs, the `CommentBlockElement` in DCR requires styling for blockquotes in the [CommentBlockComponent.tsx](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/CommentBlockComponent.tsx)

The most simple way to do this would be to update the `bodyContentStyles` within this file to include styles for the `blockquote` element!

An example can be found: [on Frontend](https://www.theguardian.com/commentisfree/live/2019/mar/15/webchat-share-questions-for-george-monbiot-and-climate-strike-activists?page=with:block-5c8b9709e4b0cf92e5a57a5f#block-5c8b9709e4b0cf92e5a57a5f) & [on DCR](https://www.theguardian.com/commentisfree/live/2019/mar/15/webchat-share-questions-for-george-monbiot-and-climate-strike-activists?dcr&page=with:block-5c8b9709e4b0cf92e5a57a5f#block-5c8b9709e4b0cf92e5a57a5f)

| Frontend      | DCR      |
|-------------|------------|
| [Example](https://www.theguardian.com/commentisfree/live/2019/mar/15/webchat-share-questions-for-george-monbiot-and-climate-strike-activists?page=with:block-5c8b9709e4b0cf92e5a57a5f#block-5c8b9709e4b0cf92e5a57a5f) | [Example](https://www.theguardian.com/commentisfree/live/2019/mar/15/webchat-share-questions-for-george-monbiot-and-climate-strike-activists?dcr&page=with:block-5c8b9709e4b0cf92e5a57a5f#block-5c8b9709e4b0cf92e5a57a5f) |
| ![opinionFrontend2][] | ![opinionDCR2][] |

[opinionFrontend2]: https://user-images.githubusercontent.com/9575458/166668118-3de62f36-e4d6-4838-b266-1d902b930354.png
[opinionDCR2]: https://user-images.githubusercontent.com/9575458/166668205-ef749262-e50b-414b-b589-f8f19c4033f1.png

> Note: While the DCR comment block does not look identical to that of Frontend, the important change would be to ensure there is a clear visual indicator that text in the `blockquote` element is quoted.

OllysCoding:
@bryophyta This looks really good! As you said it's not used too regularly so I don't think we need to over-invest time in it.

[...]

PR #4838, opened 2022-05-05
Style Opinion comment blockquotes

bryophyta:
## What does this change?

It adds styling to `<blockquote>` elements when they are contained in embedded comments.

## Why?

There was no visual differentiation between the main text of the embedded comment, and the text that was being quoted in that comment (#4826)

## Screenshots

| Before      | After      |
|-------------|------------|
| 'before' screenshot, desktop display | 'after' screenshot, desktop display |
| 'before' screenshot, mobile display | 'after' screenshot, mobile display |

### DIFF
--- a/dotcom-rendering/src/web/components/CommentBlockComponent.stories.tsx
+++ b/dotcom-rendering/src/web/components/CommentBlockComponent.stories.tsx
@@ -26,3 +26,40 @@ export const defaultStory = () => {
 	);
 };
 defaultStory.story = { name: 'default' };
+
+export const embeddedBlockquoteStory = () => {
+	const bodyContent = `
+		<blockquote>
+			<p>
+				Children and youth are striking to demand action. But many of not most are
+				themselves involved in one of the most damaging practices towards earth's
+				climate, I mean consuming massive amounts of meat and diary in their diets.
+			</p>
+			<p>What is their thought on this regard ?</p>
+		</blockquote>
+		<p>
+			Are they? I see young people leading on this. There's been a massive surge
+			towards veganism by young people, which has resulted in a sevenfold increase
+			in the UK within 5 years. Quite remarkable. And it should be recognised.
+		</p>
+	`;
+	return (
+		<div
+			css={css`
+				padding: 20px;
+			`}
+		>
+			<CommentBlockComponent
+				profileURL="https://profile.theguardian.com/user/id/1395989"
+				profileName="George Monbiot"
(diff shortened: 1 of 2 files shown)


dotcom-rendering/src/web/components/CommentBlockComponent.tsx
```diff
@@ -1,5 +1,11 @@
 import { css } from '@emotion/react';
-import { textSans, neutral, space } from '@guardian/source-foundations';
+import {
+	textSans,
+	neutral,
+	space,
+	remSpace,
+	text as textColor,
```
mxdvl, line 7: Why the renaming?

mxdvl (approved):
> This looks great.

It really does!

oliverlloyd (approved):
Nice work! It'd be great to move over to pixels (for consistency) but I'm happy to approve for expediency

dotcom-rendering/src/web/components/CommentBlockComponent.tsx
```diff
@@ -25,6 +31,17 @@ const bodyContentStyles = css`
 		word-break: break-word;
 		margin-bottom: ${space[2]}px;
 	}
+
+	blockquote {
+		font-style: italic;
+		color: ${textColor.supporting};
+		border-left: ${remSpace[1]} ${neutral[86]} solid;
```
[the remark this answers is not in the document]
bryophyta: ah okay great, that's good to know thanks!

I've just pushed an update now which uses `space` instead -- afaict the `space[n]` gives the same computed value as `remSpace[n]` (assuming that we never change the root font size) so I've left the indexes as they were before.

dotcom-rendering/src/web/components/CommentBlockComponent.stories.tsx
```diff
@@ -26,3 +26,40 @@ export const defaultStory = () => {
...
+		<p>
+			Are they? I see young people leading on this. There's been a massive surge
+			towards veganism by young people, which has resulted in a sevenfold increase
+			in the UK within 5 years. Quite remarkable. And it should be recognised.
+		</p>
+	`;
+	return (
+		<div
+			css={css`
+				padding: 20px;
```
oliverlloyd, line 49:
```suggestion
				padding: 20px;
				width: 700px;
```

oliverlloyd: Nitpicking  but it's nice if the story has a bit of context for how the content might look

PR #4840, opened 2022-05-05
Dblatcher/add newsletter signup layout file

dblatcher:
## What does this change?
Adds a layout component to partially implement a dedicated design for newsletter sign-up page. The component is not used, so **this PR has  no user impact**.

## Why?
Newsletter sign-up pages are have often been built individually using customised immersive atoms. Editorial cannot make content changes as this requires manual code changes to the atoms. The plan is to move away from this method to having sign-up pages written and published in composer and rendered in a dedicated page design.

The design has been implemented once as an immersive page:
https://www.theguardian.com/global/ng-interactive/2022/apr/13/first-edition-sign-up-guardian

To avoid having a single PR with a very large diff, this PR is intended to add the outline of the layout so the required functionality can be progressively added with future PR's until the feature is ready for release.

## Screenshots

| Mobile | Desktop |
|-------------|------------|
| Screenshot 2022-06-01 at 10 12 12 | Screenshot 2022-06-01 at 10 12 46 |


dblatcher:
[DCR sign up pages (CAPI)](https://trello.com/c/vKiXte7q/622-dcr-sign-up-pages-capi)

dotcom-rendering/src/web/layouts/lib/newsletterSignupLayoutUsage.tsx
```diff
@@ -0,0 +1,28 @@
+import { ArticleDesign, ArticleDisplay } from '@guardian/libs';
+import { decidePalette } from '../../lib/decidePalette';
+import { NewsletterSignupLayout } from '../NewsletterSignupLayout';
+
+export const hasNewsletterSignUpTag = (CAPIArticle: CAPIArticleType): boolean =>
+	CAPIArticle.tags.some((tag) => tag.id === 'info/newsletter-sign-up');
+
+export const buildNewsletterSignupLayout = (
```
oliverlloyd, line 8: Is this a function used for local testing? If so, can I suggest we replace it with a story? Storybook is a really great tool for this kind of work, I think you will end up loving it if you give it a chance 😉 

Very happy to pair on getting this setup

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -0,0 +1,275 @@
...
+import { ArticleBody } from '../components/ArticleBody';
+
+interface Props {
+	CAPIArticle: CAPIArticleType;
+	NAV: NavType;
+	format: ArticleFormat;
+	palette: Palette;
+}
+
+const newslettersSubNav: SubNavType = {
```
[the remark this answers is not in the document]
dblatcher: Will do - they will want a subnav link to the newsletters page, so I'll follow up separately about how to do that properly.

dotcom-rendering/src/web/layouts/lib/newsletterSignupLayoutUsage.tsx
dblatcher: Thanks, I've added a story with a few variations for the layout and removed this file

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -0,0 +1,275 @@
...
+				</>
+			</div>
+
+			{CAPIArticle.config.switches.surveys && (
+				<AdSlot position="survey" display={format.display} />
+			)}
+
+			<main>
+				<ElementContainer padded={false}>
+					<NewsLetterSignupBanner />
```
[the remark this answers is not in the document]
dblatcher: The logo in the  banner is part of the layout design rather than the `MainMedia` for the article (on the design, the madin media will be placed alongside the content - like the "first Edition"graphic on https://www.theguardian.com/global/ng-interactive/2022/apr/13/first-edition-sign-up-guardian.

I'll see if I can get a smaller version of the file - 32kb is more than it should be.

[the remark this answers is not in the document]
oliverlloyd: Ah, I see now. Thanks for explaining. Sideways main media, cool!

Re the svg. I notice that the Guardian logo is in there. We already have that SVG in the source system so we could just import that from there - it's already being imported in the Header. I also see the yellow Newsletters text gets positioned in different places based on breakpoint so having them separate makes sense. I wonder in that 'Newsletters' could just be...text? It's using the font we already have on the page so unless there's a requirement for it to be an svg I think text would be much smaller and easier to work with here?

dotcom-rendering/src/web/layouts/lib/newsletterSignupLayoutUsage.tsx
oliverlloyd: We should catch up about this. I think if we just add the [version of format](https://github.com/guardian/libs/blob/main/src/format/ArticleDesign.ts) that contains `NewsletterSignup` then you should automatically just get stories (including all the variation with different themes). I will chase up adding this version for you

oliverlloyd: Oh, so actually, we already have the `NewsletterSignup` design type, it was added recently!

If you look at [this Chromatic page](https://www.chromatic.com/component?appId=5dfcbf3012392c0020e7140b&csfId=layouts-format-variations-standard-newslettersignup&buildNumber=12558&k=627e22c3fe2deb004a36404f-1200-interactive-true&h=13&b=-1) you can see the way the page currently looks.

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -0,0 +1,232 @@
...
+import { getContributionsServiceUrl } from '../lib/contributions';
+import { ArticleContainer } from '../components/ArticleContainer';
+import { ArticleBody } from '../components/ArticleBody';
+import { ContainerLayout } from '../components/ContainerLayout';
+
+interface Props {
+	CAPIArticle: CAPIArticleType;
+	NAV: NavType;
+	format: ArticleFormat;
+	palette: Palette;
```
oliverlloyd, line 32: Tiny thing: There was a [recent refactor](https://github.com/guardian/dotcom-rendering/pull/4379) to have `palette` created locally, rather than passed in as a prop, can we do that here too, please?

dblatcher: sure, no problem

dblatcher:
> 
Yes, I think we will also need the equivalent rendering code in apps on a subsequent PR. The plan is that the new sign-up pages will be written in composer as regular articles (but with the info/newsletter-sign-up tag applied) so the I assume the standard App article layout would work, but we'd want to get the newsletter design elements in on the app too.

dotcom-rendering/src/web/components/NewsLetterSignupBanner.tsx
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	${until.desktop} {
+		justify-content: flex-start;
+		width: 100%;
+	}
+`;
+
+export const NewsLetterSignupBanner = () => (
+	<div css={bannerStyle}>
```
oliverlloyd, line 34: I think, now that [this change](https://github.com/guardian/dotcom-rendering/pull/4903) to `ContainerLayout` to enable a background colour just for the content has merged, this code could make use of that component here

dblatcher: I refactored the static SVG for the 'newsletters' badge into a component and separated it from the main Guardian logo, so the `ContainerLayout` does work well for the banner now.

dotcom-rendering/src/web/components/NewsLetterSignupBanner.tsx
```diff
@@ -0,0 +1,76 @@
...
+	${until.leftCol} {
+		justify-content: flex-start;
+		max-width: 200px;
+	}
+
+	${until.desktop} {
+		max-width: 170px;
+	}
+
+	svg:nth-of-type(1) {
```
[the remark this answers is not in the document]
dblatcher: I didn't see any other obvious selectors to distinguish the two svg elements, so I've put wrapper elements around each  SVG component and applied the css directly on those - I think this makes it clearer overall.

dotcom-rendering/src/web/components/NewsLetterSignupBanner.tsx
```diff
@@ -0,0 +1,76 @@
...
+		${until.leftCol} {
+			bottom: 0;
+		}
+	}
+`;
+
+const leftColWrapperStyle = css`
+	display: flex;
+	justify-content: flex-end;
+	margin: ${space[2]}px 0 ${space[9]}px;
```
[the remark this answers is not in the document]
dblatcher: 👍

dotcom-rendering/src/web/components/NewsLetterSignupBanner.tsx
```diff
@@ -0,0 +1,76 @@
...
+		${until.leftCol} {
+			display: block;
+		}
+	}
+
+	svg:nth-of-type(2) {
+		padding-top: ${space[1]}px;
+		width: 100%;
+		position: relative;
+		bottom: -11%;
```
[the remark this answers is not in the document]
dblatcher: The values were to align the baseline of the text within the newsletters badge svg with the guardian logo, rather than the bottom of the badge's frame (design requirement). It works as well with `px` so I've changed to that and added a comment about it.

dotcom-rendering/src/web/components/NewsLetterSignupBanner.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export const NewsLetterSignupBanner = () => (
+	<ContainerLayout
+		backgroundColour={brand[400]}
+		leftContent={
+			<div css={leftColWrapperStyle}>
+				<SvgGuardianLogo textColor={neutral[100]} width={200} />
+			</div>
+		}
+	>
```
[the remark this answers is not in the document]
dblatcher: sure - done!

dotcom-rendering/src/web/components/NewslettersBadge.tsx
```diff
@@ -0,0 +1,65 @@
...
+const textStyle = (textColour: string) => css`
+	${headline.small({ fontWeight: 'bold' })};
+	font-size: 19.5px;
+	letter-spacing: -0.5px;
+	fill: ${textColour};
+`;
+
+interface Props {
+	textColour?: string;
+	borderColour?: string;
```
oliverlloyd, line 13: We're not actually using these props anywhere (I think?) so I guess we could just inline these values?

If I'm wrong and we are using them, ignore me!

dotcom-rendering/src/web/components/NewslettersBadge.tsx
```diff
@@ -0,0 +1,65 @@
+import { css } from '@emotion/react';
+import { brandAlt, headline, neutral } from '@guardian/source-foundations';
+
+const textStyle = (textColour: string) => css`
+	${headline.small({ fontWeight: 'bold' })};
```
oliverlloyd, line 5:
```suggestion
	${headline.xxsmall({ fontWeight: 'bold' })};
```
`xxsmall` is 20px

See: https://guardian.github.io/source/?path=/story/packages-source-foundations-typography--page#headline

dotcom-rendering/src/web/components/NewslettersBadge.tsx
```diff
@@ -0,0 +1,65 @@
...
+	font-size: 19.5px;
```
oliverlloyd, line 6:
```suggestion
```

As above, we can use the `headline` api to set this. Note. I think decimals in px are valid but for consistency I guess best avoided? Consistent font sizing when reading text is relaxing... 🧘

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -0,0 +1,332 @@
...
+			width: 65%;
+			display: block;
+		}
+	}
+`;
+
+// the negative bottom values at the two column layout are to
+// align the baseline of the text in the newsletters badge svg
+// with the guardaina logo in the leftCol, rather than aligning
+// the bottom of the SVG frame (design requirement)
```
oliverlloyd, line 76: 👍

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -0,0 +1,332 @@
...
+const mainColNewsLettersBadgeContainerStyle = css`
+	svg {
+		width: 100%;
+		position: relative;
+		bottom: -10.75px;
```
oliverlloyd, line 81: Genuine question: do the decimal pixels make a difference here? Is this something to do with how expensive my screen is?

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -0,0 +1,332 @@
...
+					</>
+				)}
+			</div>
+
+			{CAPIArticle.config.switches.surveys && (
+				<AdSlot position="survey" display={format.display} />
+			)}
+
+			<main>
+				<ContainerLayout
```
oliverlloyd, line 223: Bit of a design question here but would this look better if there were a border on this container? You can set borders using `borderColour={brandLine.primary}`

oliverlloyd (approved):
Looking great!

I am down to just some small little things (thanks for sticking with me on this!). I'm happy to approve this and leave these changes with you to manage and merge when you're ready 👍 

Nice work

dotcom-rendering/src/web/components/NewslettersBadge.tsx
dblatcher: We aren't using them now, but I've seen other designs in the pipeline using the same logo, but with different colors (eg  dark text on a light background) - I'd tend to keep the option to future use, but not a strong preference.

dotcom-rendering/src/web/components/NewslettersBadge.tsx
dblatcher: I know this is pretty horrible, but because the style is being applied to a`<text>` within an `<svg>`, the size is relative to SVG's coordinate system, and 19.5px is the size I need to make it match the design - it equates to about 35px text in  the document.
I'll put some comments in about that.

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
dblatcher: I think it's only noticeable if you zoom in - I assumed people will to be on the safe side. I'm not sure on this, but it might vary between browsers.

dotcom-rendering/src/web/components/NewslettersBadge.tsx
dblatcher: no problem - I'll inline the values!

dotcom-rendering/src/web/components/NewslettersBadge.tsx
oliverlloyd: I went away and checked this with a few people (because I really wasn't sure myself). The advice I've got is to never use sub pixel values. The reasons for this are:

- Browsers will round up and down inconsistently, so things will look different in different contexts
- You can end up with antialiasing or blurriness on different screen types (non retina)

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
dblatcher: 👍 I've rounded them down to the pixel here.

dotcom-rendering/src/web/components/NewslettersBadge.tsx
dblatcher: It's probably best not to use pixels (let alone sub pixels) for font-size within an SVG as its misleading to have css where "px"≠1px  - I've changed this to a % value for the style applied to the svg `text`

PR #4842, opened 2022-05-06
Unify sanitisation

mxdvl:
## What does this change?

Use type declarations from `dompurify`, `jsdom` and `sanitize-html` to remove type assertions.

Unifiy the two usages of dompurify in the src/model folder as they mostly overlap.

## Why?

If there are type definitions available, we should avoid declaring our own types, or using `any`.


dotcom-rendering/src/amp/lib/sanitise-html.test.ts
```diff
@@ -1,7 +1,7 @@
-import { sanitise } from '../../lib/sanitise-html';
+import sanitise from 'sanitize-html';
 
 describe('sanitise-html', () => {
-	it('Remove rouge attributes', () => {
+	it('Remove rogue attributes', () => {
```
OllysCoding, line 4: +1

bryophyta (approved):
I'm not sure I have enough context to follow this in detail, but can't spot any obvious issues 👍

PR #4844, opened 2022-05-06
Add `mailto` hrefs to in body links

### LINKED ISSUE #4843: Tracking mailto links `in body link`
_From:_ Guy Par
_Date:_ May 5th, 2022
_Via:_ email

> I've just checked and an event is firing when the link is clicked. However, this is just labelled as 'in-body-link', which is the same tracking as for any link embedded within the text of an article, so we'd be unable to split them two apart. I'm just looping in the Dotcom team to see if they might be able to help with the tracking. Dotcom, is there any way we could have a specific component added for this email link (screenshot below)? Or a different link name?  

image

mxdvl:
## What does this change?

Add more tracking for `mailto:` links.

## Why?

Data analysts would like to know when users are going to an email.

Closes #4843


.vscode/settings.json
```diff
@@ -16,6 +16,7 @@
         "titleBar.activeBackground": "#648FFF",
         "titleBar.activeForeground": "#000000"
 	},
+	"jest.jestCommandLine": "fnm use; yarn workspace @guardian/dotcom-rendering test",
```
mxdvl, line 19: This is actually part of #4850

OllysCoding: I'm a bit concerned about seen `fnm` here and all our docs advise `nvm` and I would assume this won't run on my machine since I don't have `fnm` instaled?

mxdvl: Good shout. Have removed from this PR, and updated #4850 to skip the node manager step.

OllysCoding: One for a client side infra meeting?!

OllysCoding: We could always write a wrapper in the makefile which allows us to run 'use' on nvm or fnm (or even asdf if we wanted)

PR #4847, opened 2022-05-06
feat: Snap links first pass

AshCorr:
Part of #4855 

## What does this change?

Add support to render Snap links in DCR

## Why?

 One more step towards 100% fronts! This should let us start implementing some of the more complicated layouts such as `dynamic/package` and `fixed/thrasher`

### Frontend
![before][]

### DCR
![after][]

[before]: https://user-images.githubusercontent.com/21217225/167393882-a55ec21c-ec57-4bcc-b42c-6fbd8f36554c.png

[after]: https://user-images.githubusercontent.com/21217225/167393906-09c47bf4-31e4-4037-82bf-fce15e892f62.png


dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -235,6 +244,23 @@ export const Card = ({
 		);
 	};
 
+	if (type === 'LinkSnap' && enriched?.embedHtml !== undefined) {
+		return (
+			<div css={snapStyles}>
+				<Global styles={enriched?.embedCss} />
```
AshCorr, line 250: Can we / should we try to namespace this CSS so that it can't affect other page elements?

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -235,6 +244,23 @@ export const Card = ({
...
+				<div
+					dangerouslySetInnerHTML={{ __html: enriched?.embedHtml }}
+				/>
+
+				{enriched?.embedJs !== undefined && (
+					<script
```
AshCorr, line 256: This JS should probably be ran asynchronously.

dotcom-rendering/src/web/components/Snap.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+type Props = {
+	enriched?: DCREnrichedContent;
+};
+
+export const Snap = ({ enriched }: Props) => {
+	return (
+		<div css={snapStyles}>
+			{enriched?.embedCss !== undefined && (
+				<Global styles={enriched?.embedCss} />
```
AshCorr, line 16: Currently putting the styles for Snap in the page header along with the rest of the page styles. Should we copy frontend and put the styles for the snap next to the snap itself?

I don't feel too strongly either way, having them local might make it easier to debug.

dotcom-rendering/src/web/components/Snap.tsx
```diff
@@ -0,0 +1,32 @@
...
+			)}
+
+			{enriched?.embedHtml !== undefined && (
+				<div
+					dangerouslySetInnerHTML={{ __html: enriched?.embedHtml }}
+				/>
+			)}
+
+			{enriched?.embedJs !== undefined && (
+				<script
```
AshCorr, line 26: Currently on Frontend the JS is ran at page load, I wonder if we should try and do it differently in DCR and load it asynchronously.

dotcom-rendering/src/model/enhanceCollections.ts
```diff
@@ -62,41 +68,38 @@ const enhanceCards = (
...
-							faciaCard.supportingContent,
-							format,
-							containerPalette,
-					  )
-					: undefined,
-			};
-		});
+	collections.map((faciaCard, index) => {
+		// Cards with snaps won't have a set format.
+		const format = decideFormat(faciaCard.format || {});
```
oliverlloyd, line 73: If we're saying `format` can be undefined and this is valid then I guess it needs to be an optional prop?

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -57,6 +58,8 @@ export type Props = {
 	// Labs
 	branding?: Branding;
 	supportingContent?: DCRSupportingContent[];
+	type?: string;
+	enriched?: DCREnrichedContent;
```
oliverlloyd, line 62: This feels like the wrong place to do this. I'm not super familiar with what a 'snap' is though but is it really a type of `Card`?

dotcom-rendering/src/web/components/Snap.tsx
oliverlloyd: I would say we probably don't want to introduce global css on the page if it isn't _really_  global

dotcom-rendering/src/model/enhanceCollections.ts
```diff
@@ -39,6 +39,12 @@ const decideSubLinkFormat = ({
 	return linkFormat;
 };
 
+// There are certain collections used to add things like global styles to a front page
+// which we don't want to use on DCR as we'd like to try and migrate these styles into DCR itself.
+const blockedCollections = [
+	'ba744122-6c99-417f-8ddd-5a5905eb4928', // Palette styles new do not delete
+];
```
oliverlloyd, line 46: Why do we need to block this? Is the css in here causing issues or is this a drive by?

oliverlloyd (commented):
I'm just wondering. Is there another way to do this? I am super reluctant to start introducing the embed pattern on Fronts. It caused a lot of headaches in articles but we kind of put up with it because it had limited usage but this seems like we'd be running this code considerably more often.

OllysCoding:
Could we investigate placing these snaps in `iframes`? We could do something like what we do for InteractiveAtoms: https://github.com/guardian/atoms-rendering/blob/main/src/InteractiveAtom.tsx

AshCorr:
> Is there another way to do this?

I think the best other way we could do this is to come up with a framework for Visuals to make these containers straight in DCR, theres 2 problems I see with this though:

- Visuals would need to maintain the snap in 2 different places until fronts gets to 100%
- It'll take a lot of work to recreate all these snaps in DCR, if we want to get Fronts out in a reasonable amount of time it could be problematic.

I feel like we should try to get Fronts to 100% first and then build a new framework for visuals to produce content

dotcom-rendering/src/model/enhanceCollections.ts
AshCorr: Yea probably, I think I was just lazy here, I'll see about refactoring it to be optional

jamesgorrie:
> I think the best other way we could do this is to come up with a framework for Visuals to make these containers straight in DCR

This sounds like a very sensible idea. If we take this DCR first perspective, we would require Editorial design to build their snaps in DCR and all the goodness they get from this.

> Visuals would need to maintain the snap in 2 different places until fronts gets to 100%

Currently [this repo](https://github.com/guardian/interactive-atom-template-2019#create-a-new-atom) exposes an endpoint that [looks like this](https://gist.github.com/jamesgorrie/01773b4e5a983d44709143ecd2a5a973) that is used to add interactive atoms (thrashers) to fronts. If `frontend` exposed an endpoint like this rendered from `DCR`, we could support frontend rendered fronts while we decommission them while working on a new solution.

This would require quite a lot of time and process change across multiple teams (Editorial design, CP, us as support).

> It'll take a lot of work to recreate all these snaps in DCR

@HarryFischer reckons there's about 15, but this is from Editorial design. Another team we need to consider is visuals who follow a similar process. 

> It caused a lot of headaches in articles

@oliverlloyd do we have examples of what these were? @OllysCoding's suggestion of using the interactive atom sounds like a good one, but would it address those issues? It also fits in with the fact that, as far as I can tell, thrashers are interactive atoms.

oliverlloyd:
> > It caused a lot of headaches in articles
> 
> @oliverlloyd do we have examples of what these were? @OllysCoding's suggestion of using the interactive atom sounds like a good one, but would it address those issues? It also fits in with the fact that, as far as I can tell, thrashers are interactive atoms

The headaches were related to the fact we had undocumented and unstructured couplings between the html, css and javascript being written inside the interactive atoms and the platform code. These embeds are created based on expectations that things will exist on the page and when they move or change it causes the content to break.

Things like

- Classnames
- elements (like `main` or `article`) in specific positions
- specific hierarchical structure to the dom
- Global styles

Because these types of couplings are _possible_, they are essentially inevitable.

I think there are certain things we could do to mitigate these problems:

1. Ban all and any effects on content outside of the containing element
A snap/thrasher etc. should never be able to affect content outside of itself. This would disallow the Saul Goodman snap on Culture which absolutely positions itself above the container ahead of itself. This is a conceptual ban but could be enforced with process and review. Usage of a tool like Storybook would help encourage this pattern at the point where the content is created (as oppose to developing the content directly on the page locally which invites couplings)

2. Ban javascript
Javascript has been used in the past to manipulate the DOM after the page has loaded. Elements are removed or inserted based on design desires. This is perhaps the most egregious example of bad coupling and by not allowing it we we guard against problems in the future.

3. Force the content into an iframe
@OllysCoding can probably speak more to the benefits here but my understanding is that this makes it harder for the embed to affect content outside of itself. But not impossible, is that true?

dotcom-rendering/src/web/components/Snap.tsx
```diff
@@ -10,23 +10,19 @@ type Props = {
...
-				<script
-					dangerouslySetInnerHTML={{ __html: enriched?.embedJs }}
-				/>
-			)}
-		</div>
+	return (
+		<div
+			css={[
+				snapStyles,
+				css`
```
AshCorr, line 21: I'm not sure if you're meant to be able to do this with emotion but it seems to work for sandboxing CSS!

dotcom-rendering/src/model/enhanceCollections.ts
AshCorr: If we're disabling JS and preventing CSS from breaking out of snaps I don't think this is needed anymore. This was to stop things like the sombre pallet snap that purely exists to change other elements on the page from loading.

mxdvl:
Are we losing the subscript in DCR?


Frontend: CO<sub>2</sub>
DCR: CO2

AshCorr:
> Are we losing the subscript in DCR?
> 
> Frontend: CO2 DCR: CO2

Good catch, I think frontend has some global CSS for subscripts which DCR does not have. Don't think theres anything we can do about it, its something that has to be corrected in the snap itself.

dotcom-rendering/src/web/components/Snap.tsx
oliverlloyd: Well, what do you know, that's actually kinda neat

jamesgorrie:
> its something that has to be corrected in the snap itself

I'd find out where to raise the issue, and raise it there outside of this PR.

dotcom-rendering/src/model/enhanceCollections.ts
```diff
@@ -75,41 +75,38 @@ const enhanceCards = (
...
-					? enhanceSupportingContent(
-							faciaCard.supportingContent,
-							format,
-							containerPalette,
-					  )
-					: undefined,
-			};
-		});
+	collections.map((faciaCard, index) => {
+		// Cards with snaps won't have a set format.
```
oliverlloyd, line 79: This comment is perhaps a bit confusing out of context. What is it referring to?

dotcom-rendering/src/web/lib/decideDesign.ts
```diff
@@ -4,10 +4,10 @@ import { ArticleDesign } from '@guardian/libs';
...
  * i.e. when `CommentDesign` and `ImmersiveDisplay` are used jointly.
  */
-export const decideDesign = ({
-	design,
-	display,
-}: Partial<CAPIFormat>): ArticleDesign => {
+export const decideDesign = (
+	design: CAPIDesign | undefined,
+	display: CAPIDisplay | undefined,
+): ArticleDesign => {
```
oliverlloyd, line 10:
```suggestion
export const decideDesign = ({
	design,
	display,
}:{
	design: CAPIDesign | undefined,
	display: CAPIDisplay | undefined,
}): ArticleDesign => {
```

This is more verbose but passing objects is nice because there's not much chance for confusion!

dotcom-rendering/src/web/lib/decideDisplay.ts
```diff
@@ -1,8 +1,8 @@
 import { ArticleDisplay } from '@guardian/libs';
 
-export const decideDisplay = ({
-	display,
-}: Partial<CAPIFormat>): ArticleDisplay => {
+export const decideDisplay = (
+	display: CAPIDisplay | undefined,
+): ArticleDisplay => {
```
oliverlloyd, line 5: As above

dotcom-rendering/src/web/lib/decideFormat.ts
```diff
@@ -2,8 +2,10 @@ import { decideDesign } from './decideDesign';
...
-export const decideFormat = (format: Partial<CAPIFormat>): ArticleFormat => ({
-	display: decideDisplay(format),
-	theme: decideTheme(format),
-	design: decideDesign(format),
+export const decideFormat = (
+	format: CAPIFormat | undefined,
+): ArticleFormat => ({
+	display: decideDisplay(format?.display),
+	theme: decideTheme(format?.theme),
+	design: decideDesign(format?.design, format?.display),
```
oliverlloyd, line 10: 🤔 why are we accepting an undefined `format` object? This feels... wrong? I guess the reason is the snap might not have `format` defined but that feels like a problem of the snap, rather than something that `decideFormat` should have to deal with.

Ideally we would fix the snap to be more consistent with predictable structured data but in the event this isn't possible I think we should default it to a standard format property and then pass *that* into our decide functions, keeping these props as required - because they really should be required.

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -54,6 +55,8 @@ export type Props = {
 	// Labs
 	branding?: Branding;
 	supportingContent?: DCRSupportingContent[];
+	isSnap?: boolean;
+	snapData?: DCREnrichedContent;
```
oliverlloyd, line 59: Forgive me if I've already asked this and you've patiently explained why I'm wrong and then I forgot and I'm now asking the same thing again, but would it work to just have `snapData` here and then the decision about if a `Card` is a snap or not is based on the existence of this data?

dotcom-rendering/index.d.ts
```diff
@@ -679,6 +679,14 @@ type DCRSupportingContent = {
...
 };
 
+type FEEnrichedContent = {
+	embedHtml?: string;
+	embedCss?: string;
+	embedJs?: string;
+};
+
+type DCREnrichedContent = FEEnrichedContent;
+
```
oliverlloyd, line 689:
```suggestion
type FEEnrichedContent = {
	embedHtml?: string;
	embedCss?: string;
	embedJs?: string;
};

type DCREnrichedContent = {
	embedHtml?: string;
	embedCss?: string;
};

```
This might be a nice way to convey our intentions. What do you think?

oliverlloyd: Also, just a thought, but what are your feelings on landing on just 'snap' as the naming convention and leaving 'enriched' out of DCR? E.g.: Replace `DCREnrichedContent ` with `DCRSnapType`. Up to you 🤷‍♂️

oliverlloyd (commented):
Looking very cool.

I did want to push back on the changes to the decide functions though. These are very core to what DCR does and making those props optional cold create problems in future. Keeping the props required and forcing a default higher up prevents this.

oliverlloyd (approved):
Thanks for sticking with this one, really great work

dotcom-rendering/src/web/components/Snap.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
..._component-button:hover{background-color:#ffd622;text-decoration:none}.us-morning-briefing-post-election-2020__component-button:focus span,.us-morning-briefing-post-election-2020__component-button:hover span{color:#000;text-decoration:none}.us-morning-briefing-post-election-2020__component-button:focus svg,.us-morning-briefing-post-election-2020__component-button:hover svg{fill:#000}@media (min-width:41.25em){.us-morning-briefing-post-election-2020__component-button{margin-top:16px}}.us-morning-briefing-post-election-2020__roundel{top:4px;right:4px;position:absolute;top:3px;right:9px}@media (min-width:30em){.us-morning-briefing-post-election-2020__roundel{right:18px}}@media (min-width:30em){.us-morning-briefing-post-election-2020__roundel{right:10px;top:3px}}@media (min-width:46.25em){.us-morning-briefing-post-election-2020__roundel{right:11px;top:6px}}.interactive-atom{margin:0;overflow:hidden}.android .us-morning-briefing-post-election-2020,.ios .us-morning-briefing-post-election-2020{overflow:hidden;margin:0;padding:0}.android .us-morning-briefing-post-election-2020__component-button span,.ios .us-morning-briefing-post-election-2020__component-button span{letter-spacing:unset}',
```
mxdvl, line 15: Is it typical of snaps to have `@font-face` declarations? [According to MDN, they will be discarded if inserted inside a css selector](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face#notes).

I think we should warn whenever we find such declaration, but this can be done in a subsequent PR.

mxdvl (approved):
📈

dotcom-rendering/src/web/components/Snap.stories.tsx
AshCorr: Unfortunately it is very common for snaps to have `@font-face`, I'm pretty sure that the emotion CSS parser is intelligent enough to not wrap `@font-face` in a selector

PR #4849, opened 2022-05-06
feat(KeyEventsContainer): add type predicate

mxdvl:
## What does this change?

Use a type predicate to figure out if a key event is valid.

An alternative would be to filter the data upstream an never pass Blocks without a `title` and a `blockFirstPublished` key.

## Why?

Work with Typescript, not against it.


oliverlloyd (approved):
Yep!

PR #4851, opened 2022-05-06
Reduce passing around of 'switches' object - Islands

OllysCoding:
## What does this change?

In some components we take the whole 'switches' object as props, and then consume just 1 or 2 keys from it. This a problem especially in Island components because this entire object must be serialised onto the DOM, downloaded, then unserialised by the client!

```ts
interface Switches {
	[key: string]: boolean;
}
```
As seen above, switches is an untyped object where the size and exact contents are inherently unknown, this is not great for dev experience!

Where possible then, it'd likely be preferred to pass around the individual switches, rather than the whole object!

This PR updates `Liveness` and `StickyBottomBanner` (the two island components which consume switches) to use the specific switches required as props rather than the entire object.

### DIFF
--- a/dotcom-rendering/src/web/components/StickyBottomBanner.importable.tsx
+++ b/dotcom-rendering/src/web/components/StickyBottomBanner.importable.tsx
@@ -52,7 +52,7 @@ type RRBannerConfig = {
 	id: string;
 	BannerComponent: React.FC<BannerProps>;
 	canShowFn: CanShowFunctionType<BannerProps>;
-	isEnabled: (switches: CAPIArticleType['config']['switches']) => boolean;
+	isEnabled: boolean;
 };
 
 const getBannerLastClosedAt = (key: string): string | undefined => {
@@ -106,7 +106,6 @@ const buildRRBannerConfigWith = ({
 		tags,
 		contributionsServiceUrl,
 		idApiUrl,
-		switches,
 	}: {
 		isSignedIn: boolean;
 		asyncCountryCode: Promise<string>;
@@ -122,14 +121,13 @@ const buildRRBannerConfigWith = ({
 		tags: TagType[];
 		contributionsServiceUrl: string;
 		idApiUrl: string;
-		switches: Switches;
 	}): CandidateConfig<BannerProps> => {
 		return {
 			candidate: {
 				id,
 				canShow: () =>
 					canShowFn({
-						remoteBannerConfig: isEnabled(switches),
+						remoteBannerConfig: isEnabled,
 						isSignedIn,
 						asyncCountryCode,
 						contentType,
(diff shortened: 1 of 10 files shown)


dotcom-rendering/src/web/components/StickyBottomBanner.importable.tsx
```diff
@@ -169,19 +167,29 @@ const buildRRBannerConfigWith = ({
...
+	remoteBannerSwitch,
+}: {
+	remoteBannerSwitch: boolean;
+}) =>
+	buildRRBannerConfigWith({
+		id: 'reader-revenue-banner',
+		BannerComponent: ReaderRevenueBanner,
+		canShowFn: canShowRRBanner,
+		isEnabled: remoteBannerSwitch,
+	});
```
OllysCoding, line 192: Why do it like this?

previously we had `isEnabled` as a function which took in `switches` and output a boolean. This worked well as it kept the definition of which switch was required in the `buildReaderRevenueBannerConfig` & `buildPuzzlesBannerConfig`, rather than relying on the caller to know to provide the correct switch value.

Doing my best to keep this then, I used currying to to take the name of the switch and the existing pattern used!

---

I don't love this though, so if anyone has any suggestions I'd love to hear them!

bryophyta:
I don't think I've got enough familiarity with the context to review properly, but fwiw this looks like a well-motivated change and the implementation looks good to me 👍

dotcom-rendering/src/web/components/StickyBottomBanner.importable.tsx
mxdvl:
```ts
const buildReaderRevenueBannerConfig = (isEnabled: boolean) =>
	buildRRBannerConfigWith({
		id: 'reader-revenue-banner',
		BannerComponent: ReaderRevenueBanner,
		canShowFn: canShowRRBanner,
		isEnabled,
	});
```

mxdvl (approved):
This is great and explicit.

The alternative would be to have method to pick specific keys from an object. However, this would create an abstraction that hides away which exact switches are being used, so I don’t think we should jump to this abstraction.

dotcom-rendering/src/web/components/StickyBottomBanner.importable.tsx
```diff
@@ -52,7 +52,7 @@ type RRBannerConfig = {
 	id: string;
 	BannerComponent: React.FC<BannerProps>;
 	canShowFn: CanShowFunctionType<BannerProps>;
-	isEnabled: (switches: CAPIArticleType['config']['switches']) => boolean;
+	isEnabled: boolean;
```
oliverlloyd, line 55: Much nicer

oliverlloyd (approved):
This is great, having this more explicit code is both more performant and easier to reason about

PR #4856, opened 2022-05-09
Add `KeyEventCard` click tracking

joecowton1:
## What does this change?

- Adds click tracking to the new `KeyEventCard` component.
- Adds a missing type definition to the stories file.

## Why?

This is a new component and currently there is no tracking in place.


oliverlloyd (commented):
If we only need to capture click events then can we achieve this goal using a basic `data-link-name` attribute? It would be great if we can avoid adding more javascript to the page.

joecowton1:
> If we only need to capture click events then can we achieve this goal using a basic `data-link-name` attribute? It would be great if we can avoid adding more javascript to the page.

100% agree - for my understanding though, what's the distinction between this and say the pinned posts or the sticky video we were also capturing click events but used `submitComponentEvent`?

oliverlloyd:
> > If we only need to capture click events then can we achieve this goal using a basic `data-link-name` attribute? It would be great if we can avoid adding more javascript to the page.
> 
> 100% agree - for my understanding though, what's the distinction between this and say the pinned posts or the sticky video we were also capturing click events but used `submitComponentEvent`?

The current implementation of `data-link-name` has limitations. It will only register event listeners on the initial page load and can only send a fixed string as the payload when a click event happens (note. it's common to use a `|` in that string to allow more data to be sent). It possible these limitations prevented those use cases or it could just be there wasn't an awareness of this option as an alternative, `tracker-js` is not well documented.

joecowton1:
> The current implementation of `data-link-name` has limitations. It will only register event listeners on the initial page load and can only send a fixed string as the payload when a click event happens (note. it's common to use a `|` in that string to allow more data to be sent). It possible these limitations prevented those use cases or it could just be there wasn't an awareness of this option as an alternative, `tracker-js` is not well documented.

Thanks for explaining - that all makes sense.

dotcom-rendering/src/web/components/KeyEventCard.tsx
```diff
@@ -90,7 +90,12 @@ export const KeyEventCard = ({ text, date, url, format }: KeyEvent) => {
 
 	return (
 		<li css={listItemStyles}>
-			<Link priority="secondary" css={linkStyles} href={url}>
+			<Link
+				priority="secondary"
+				css={linkStyles}
+				href={url}
+				data-link-name="key event card"
```
oliverlloyd, line 97: Oh, it's a prop on `Link`, nice!

oliverlloyd (approved):
So much simpler!

dotcom-rendering/src/web/components/KeyEventCard.tsx
joecowton1: `Link` has a wildcard `...props` input which is then spread onto the `a` tag - works nicely for this :)

PR #4858, opened 2022-05-09
Display the toast over a wider area

### LINKED ISSUE #4852: The toast alert should appear over comments
A reader kindly highlighted that we are no longer displaying the new post alert above the discussion and that this makes it difficult to navigate the blog.

We stop the toast (our name for the alert button) from showing below the bottom of the blog posts but if there is a discussion on a blog we should continue to show it all the way to the bottom of that discussion.

oliverlloyd:
## What does this change?
This PR fixes #4852 by displaying the toast over the whole `main` component, not just inside the article space.

I initially thought to conditionally place the toast over a larger area for blogs that have discussion on them only but I think there is benefit here in doing this in all situations. We should be displaying this information about updates over a wider area in all cases.

## Why?
A reader wrote in pointing out that the '5 new updates' button that appears when new content is available was no longer showing when they were reading discussion posts. This made navigating the liveblog difficult. This was a regression introduced as part of the work to migrate to the new DCR platform.

### The button now shows _above_ the article as well as below?
By moving the toast container up to `main` we are now showing it over both the standfirst and the headline. Previously we kept it inside the body of the article so it sat underneath main media. Using `main` as the container is the natural solution if we want to also show this button over content at the bottom of the page but that also means we will show the button further up as well. I think this is a positive change, it is surfacing this information more. @HarryFischer what do you think?

## Top
| Before      | After      |
|-------------|------------|
| top_before | top_after |

## Bottom
| Before      | After      |
|-------------|------------|
| bottom_before | bottom_after |

### DIFF
--- a/dotcom-rendering/src/web/layouts/LiveLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/LiveLayout.tsx
@@ -400,6 +400,39 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 			</div>
 
 			<main>
+				{format.design === ArticleDesign.LiveBlog && (
+					<>
+						{/* The Toast component is inserted into this div using a Portal */}
+						<div
+							id="toast-root"
+							css={css`
+								position: sticky;
+								top: 0;
+								${getZIndex('toast')};
+								display: flex;
+								justify-content: center;
+							`}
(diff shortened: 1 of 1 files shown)


HarryFischer:
Yep looks good @oliverlloyd

PR #4860, opened 2022-05-09
Fix `FilterKeyEventsToggle` scroll issue

silvacb:
## What does this change?
Adds a `breakpoint` prop to FilterKeyEventsToggle to fix a scroll issue caused by two components sharing the same ID.

## Why?
Giving each of the two filters a unique ID ensures that the `window.location.hash` navigates to the correct place on the page.

### Before:
https://user-images.githubusercontent.com/55602675/167612549-0061a67f-e070-4748-b80f-4b3adc8990e9.mov

### After:
https://user-images.githubusercontent.com/55602675/167613337-d85eb16c-b08a-4dd7-bddb-c647194e89cd.mov

### DIFF
--- a/dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
+++ b/dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
@@ -22,12 +22,12 @@ const toggleWrapperStyles = css`
 
 interface Props {
 	filterKeyEvents: boolean;
-	breakpoint: 'mobile' | 'desktop';
+	id: 'filter-toggle-mobile' | 'filter-toggle-desktop';
 }
 
 export const FilterKeyEventsToggle = ({
 	filterKeyEvents,
-	breakpoint,
+	id,
 }: Props) => {
 	const [checked, setChecked] = useState(filterKeyEvents);
 
@@ -38,13 +38,13 @@ export const FilterKeyEventsToggle = ({
 		urlParams.delete('page'); // direct to the first page
 		urlParams.set('filterKeyEvents', checked ? 'false' : 'true');
 
-		window.location.hash = `filter-toggle-${breakpoint}`;
+		window.location.hash = id;
 		window.location.search = urlParams.toString();
 	};
 
 	return (
 		<>
-			<span id={`filter-toggle-${breakpoint}`} />
+			<span id={id} />
 			<div css={toggleWrapperStyles}>
 				<ToggleSwitch
 					label="Show key events only"
--- a/dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
+++ b/dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
@@ -25,10 +25,7 @@ interface Props {
 	id: 'filter-toggle-mobile' | 'filter-toggle-desktop';
 }
 
-export const FilterKeyEventsToggle = ({
-	filterKeyEvents,
-	id,
-}: Props) => {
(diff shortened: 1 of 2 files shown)


dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -805,6 +805,7 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 													filterKeyEvents={
 														CAPIArticle.filterKeyEvents
 													}
+													id="1"
```
joecowton1, line 808: Could this `id` values be more descriptive ie `filter-key-events-toggle-desktop`?

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -823,6 +824,7 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 														filterKeyEvents={
 															CAPIArticle.filterKeyEvents
 														}
+														id="2"
```
joecowton1, line 827: and ``filter-key-events-toggle-mobile`?

dotcom-rendering/src/web/layouts/LiveLayout.tsx
silvacb: Great tip, thanks Joe!

dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
```diff
@@ -22,9 +22,10 @@ const toggleWrapperStyles = css`
 
 interface Props {
 	filterKeyEvents: boolean;
+	id: string;
```
joecowton1, line 25: Thinking about it, maybe `id` isn't the right name for this prop? What it actually signifies is the breakpoint (roughly at least). 

With that in mind, one approach could be to call the prop `breakpoint` and limit the values we can pass to it to `desktop` and `mobile`.

That would then look like this:
```
interface Props {
   filterKeyEvents: boolean;
   breakpoint: "mobile" | "desktop"
}
```

Does that make sense? We'd need to update the other places we've used `id` to be `breakpoint` as well, but again I think it explains what's going on better.

joecowton1: @silvacb I'm free this morning so if you're not hacking I'm very happy to go through all these comments together :) (I'd recommend getting involved in the hackday though!)

alinaboghiu (approved):
This is great, it's been bugging me for ages, thanks for doing it! 
Left one suggestion but very happy with this.

abeddow91 (approved):

dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
```diff
@@ -34,13 +38,13 @@ export const FilterKeyEventsToggle = ({ filterKeyEvents }: Props) => {
...
 		urlParams.set('filterKeyEvents', checked ? 'false' : 'true');
 
-		window.location.hash = 'filter-toggle';
+		window.location.hash = `filter-toggle-${breakpoint}`;
 		window.location.search = urlParams.toString();
 	};
 
 	return (
 		<>
-			<span id="filter-toggle" />
```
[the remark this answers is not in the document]
joecowton1: @silvacb if you've implemented these changes it would be good to reply here and we can wrap this one up!

joecowton1 (approved):

marjisound (approved):

PR #4875, opened 2022-05-11
Fix overflow on header image caption

iainjchambers-guardian:
## What does this change?
Hides overflow on full size image captions on editions
## Why?
The absolute position on the image caption is causing an overflow and adding a scroll
## Screenshots

| Before      | After      |
|-------------|------------|
| image | image |


mxdvl (approved):
LGTM

PR #4878, opened 2022-05-12
Extract Caption From Renderer

JamieB-gu:
## Why?

There are a few advantages:

- It's arguably easier to understand the component in isolation
- This migrates to JSX, which people seem to find easier to read
- It's useful to break JavaScript down into small files to avoid accidental side effects
- It follows a component structure that's common elsewhere in the codebase, and can have stories written for it more easily
- The renderer file has less in it, which may make it easier to reason about

## Changes

- Extracted the caption rendering code into standalone component file
- Updated usages to use new component

### DIFF
--- a/apps-rendering/src/components/caption/index.tsx
+++ b/apps-rendering/src/components/caption/index.tsx
@@ -1,0 +1,3 @@
+import Caption from './caption';
+
+export { Caption };
--- a/apps-rendering/src/components/caption/caption.tsx
+++ b/apps-rendering/src/components/caption/caption.tsx
@@ -36,7 +36,7 @@ const anchorStyles = (format: ArticleFormat): SerializedStyles | undefined =>
 		: undefined;
 
 const textStyles = css`
-	${textSans.xsmall()}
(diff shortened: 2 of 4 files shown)


apps-rendering/src/components/caption/index.tsx
```diff
@@ -0,0 +1,3 @@
+import Caption from './caption';
```
MarSavar, line 1: Is there a reason for not having the content of `caption.tsx` inside `index.tsx`?

MarSavar (approved):
Looks great, thanks for refactoring and making `renderer.ts` smaller!
Glad to see the new folder/index.tsx pattern as well :D

Just a question on the file structure (not a blocker, just curious) - otherwise this looks good to go (it would be nice to have stories at some point for this component).

apps-rendering/src/components/caption/index.tsx
JamieB-gu: Good question. This is actually part of a larger refactor and I'm anticipating having more components in this folder.

PR #4881, opened 2022-05-12
Follow takes format

JamieB-gu:
## Why?

This doesn't change the breadth of inputs (the domain) of the `Follow` component, but it does hopefully make the props a bit easier to understand, and similar to other components in the codebase. It also narrows the amount of information being passed at the call site.

## Changes

- Stop passing the whole `Item` to `Follow`


MarSavar (approved):
👍

PR #4883, opened 2022-05-12
Refactor Card Footer

oliverlloyd:
## What does this change?
This refactors the layout of the Card footer to break the article meta data out from the `Lines` and position it directly under the trailText instead. If you open the before and after images below and squint you will see the comment count and card date are in new positions.

## Why?
We're now adding more meta data to cards (specifically, the article date, maybe) and with this extra information showing we wanted to revisit how things are positioned.

This work was done in collaboration with @HarryFischer but is still a work in progress in terms of design. But by making this change we simplify the code and also make it possible to share how different combinations of cards look using this changed layout using the `?dcr` param.

### Headlines
| Before      | After      |
|-------------|------------|
| headline-before | healines-after |

### Opinion
| Before      | After      |
|-------------|------------|
| opinion-before | opinion-after |

### DIFF
--- a/dotcom-rendering/fixtures/manual/trails.ts
+++ b/dotcom-rendering/fixtures/manual/trails.ts
@@ -15,6 +15,39 @@ export const trails: TrailType[] = [
 			display: ArticleDisplay.Standard,
 		},
 		dataLinkName: 'news | group-0 | card-@1',
+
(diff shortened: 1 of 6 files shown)


mxdvl (approved):
This looks neat!

Issue #4836, opened 2022-05-05
Jump to comment not working persistently


ioannakok:
Reported from userhelp:

1) Chrome

As you can see, normally two comments are pinned to the top by the admins

https://www.theguardian.com/commentisfree/2022/may/04/us-lend-lease-act-ukraine-1941-second-world-war#comment-156234098

<img width="1148" alt="image" src="https://user-images.githubusercontent.com/19683595/166912072-cdcdc16a-b148-4f47-bd91-11d88035374e.png">

[whatsmybrowser.org/b/EWVA5](http://whatsmybrowser.org/b/EWVA5)

- pick any comment
- jump to comment works the first time 
- scroll back up to 'jump to comment' link for that comment
- link does not work
- link does work for the other comment, or if you refresh the page it will load onto the comment that it failed to jump to

(video attached)

2) Firefox - intended function

https://www.whatsmybrowser.org/b/8K738

- pick any comment
- jump to comment works over and over for the same comment

https://user-images.githubusercontent.com/19683595/166912387-5c5af4da-b17b-4a30-ac70-0aade84ea4e9.mp4

jlieb10:
TBD: there is a PR open for this, but it may be closed in favour of doing a better longterm solution. 

https://github.com/guardian/dotcom-rendering/pull/4886

jlieb10:
Related issue: https://github.com/guardian/discussion-rendering/issues/580

jlieb10:
We won't pursue this bug at this time, as this behaviour is not fundamentally breaking the page and the fix is non-trivial - It will involve revisiting the way discussion rendering is architected into DCR

PR #4886, opened 2022-05-12
Fix: Scroll to comment multiple times

bryophyta:
Code co-authored with @mxdvl 

## What does this change?
It adds a call to `scrollToView` in the permalink click handler that DCR passes to discussion-rendering.

## Why?
If a user clicked 'Jump to comment', scrolled away, then clicked it again, then the page wouldn't scroll back to the comment the second time (at least on Chrome). See issue #4836.

Our hypothesis is that previously, the scroll behaviour was triggered by a useEffect hook which watched the selected comment id in the URL. But when the same comment was 'jumped to' twice in a row, React wasn't recognising the updated as a change, and so wasn't triggering the hook.

We looked at ways of triggering a re-render in these circumstances, but we felt as though these added an extra dependency on state which might not be obvious to a reader of the codebase. Our chosen solution (triggering scroll directly from the click handler) also goes somewhat against the implicit contract between DCR and discussion-rendering. However, we felt that it makes some sense insofar as what we need to react to here is an event, rather than a state change as such. It would be good to hear from others, though, whether this reasoning makes sense to them.

## Drawbacks and potential further steps
Our solution leaves open an existing edge case: if you jump to a comment, navigate to another page of comments, then try to jump to the original comment again, this won't work. We thought that this case was fairly niche, and also that it might be better addressed by a more thorough refactoring of the navigation logic for discussion-rendering.

Specifically, the logic governing jump to comments is currently quite complicated, and the division of labour between the discussion-rendering app and DCR is also quite complex. It looks like this is an area of the codebase that has been revisited a few times recently, so we thought it would be good to discuss with previous contributors (e.g. @oliverlloyd ?) before deciding whether a slightly larger refactor would make sense, and if so what it might look like.

### DIFF
--- a/dotcom-rendering/src/web/components/Discussion.tsx
+++ b/dotcom-rendering/src/web/components/Discussion.tsx
@@ -92,7 +92,7 @@ export const Discussion = ({
 		window.location.hash = `#comment-${commentId}`;
 		// Put this comment id into the hashCommentId state which will
 		// trigger an api call to get the comment context and then expand
-		// and reload the discussion based on the resuts
+		// and reload the discussion based on the results
 		setHashCommentId(commentId);
 		return false;
 	};
--- a/dotcom-rendering/src/web/components/Discussion.tsx
+++ b/dotcom-rendering/src/web/components/Discussion.tsx
@@ -89,7 +89,20 @@ export const Discussion = ({
 
 	const handlePermalink = (commentId: number) => {
 		if (typeof window === 'undefined') return false;
-		window.location.hash = `#comment-${commentId}`;
+
+		const newHash = `#comment-${commentId}`
+
+		// If the comment is already on the page, scroll to
+		// it directly. (If comment is not loaded, updating the hashCommentId
+		// should trigger setCommentPage, which will trigger a page load in
+		// discussion rendering.) We're adding scrolling here because React
+		// doesn't register a click as a change of state, if the same link is
+		// clicked twice. (Problem still persists for edge case of user navigating
+		// to a different page of comments without changing the comment id.)
+		document.querySelector(newHash)?.scrollIntoView();
+
+		window.location.hash = newHash;
+
 		// Put this comment id into the hashCommentId state which will
 		// trigger an api call to get the comment context and then expand
 		// and reload the discussion based on the results
--- a/dotcom-rendering/src/web/components/Discussion.tsx
+++ b/dotcom-rendering/src/web/components/Discussion.tsx
@@ -90,7 +90,7 @@ export const Discussion = ({
 	const handlePermalink = (commentId: number) => {
 		if (typeof window === 'undefined') return false;
 
-		const newHash = `#comment-${commentId}`
+		const newHash = `#comment-${commentId}`;
 
 		// If the comment is already on the page, scroll to
 		// it directly. (If comment is not loaded, updating the hashCommentId


dotcom-rendering/src/web/components/Discussion.tsx
```diff
@@ -89,10 +89,23 @@ export const Discussion = ({
...
+		const newHash = `#comment-${commentId}`;
+
+		// If the comment is already on the page, scroll to
+		// it directly. (If comment is not loaded, updating the hashCommentId
+		// should trigger setCommentPage, which will trigger a page load in
+		// discussion rendering.) We're adding scrolling here because React
+		// doesn't register a click as a change of state, if the same link is
+		// clicked twice. (Problem still persists for edge case of user navigating
+		// to a different page of comments without changing the comment id.)
+		document.querySelector(newHash)?.scrollIntoView();
```
bryophyta, line 102: The existing scroll behaviour is controlled by the discussion-rendering app, and is triggered by the change of state triggered here (`setHashCommentId`). However, when a user clicks the same permalink twice, the second time doesn't register as a change of state and so doesn't trigger scrolling. In some ways it feels not-ideal for us to control scrolling from both this event handler and also from within the discussion App. But it seemed preferable for the time being (see PR description). Other views are very welcome, though!

https://github.com/guardian/discussion-rendering/blob/d034c4ab2e32d197cbd6da99c08cc4666550d4f1/src/App.tsx#L327-L334

jamesgorrie: My confusion here is why do we have this coupling in the first place? Could this whole method be handled by `discussion-rendering`? It'd be good to know if this is the right place for this to live before we overload it.

This would have the knock on effect of having to move `getCommentContext` into `discussion-rendering`, but that also feels like a less leaky place for that to be.

If DCR were to decide if it is a permalink e.g. the hash is set to `#comment-{id}`, and just pass that `id` down to `discussion-rendering` and then it's all handled there, would that remove a bit of discussion specific code from DCR and make it easier to handle there?

bryophyta: Yep, so if I remember correctly for longer-term solutions we discussed lifting the state up to discussion-rendering and exporting API utility functions for the parent to be able to construct the desired behaviour. The more I looked into it though, the more state there actually is in `discussion-rendering` itself, so I'm inclined to agree that moving `getCommentContext` down might be better than lifting the rest of the state up.

I guess one of the questions is whether we see `discussion-rendering` being used in any other contexts in the future, and if so what constraints that implies? e.g. allowing discussion to control scroll behaviour and also to set the window url might not be what we'd want in every context, so if it's going to be used outside DCR we'd need to think about how to make these behaviours optional, whilst also containing the logic for them within discussion-rendering?

jamesgorrie: > whether we see `discussion-rendering` being used in any other contexts in the future

Good questions - @oliverlloyd @JamieB-gu @nicl - any information on this, I can't find any information [in the repo](https://github.com/guardian/discussion-rendering) or in the [architecture documents](https://github.com/guardian/dotcom-rendering/tree/main/dotcom-rendering/docs/architecture).

oliverlloyd: My impression is that there's a lot of weight behind moving `getCommentContext` into discussion-rendering, this has come up a few times and would simplify things.

I think the original intention was to use discussion-rendering in apps-rendering which was what drove a lot of the dividing lines but I think it's fair to revisit this, maybe we can have `getCommentContext` in discussion-rendering with no issues even if this code were used elsewhere?

JamieB-gu: > I think the original intention was to use discussion-rendering in apps-rendering
That's still the plan to the best of my knowledge.

bryophyta:
As mentioned above, I've opened a more general issue, encompassing this one, in discussion-rendering [#580](https://github.com/guardian/discussion-rendering/issues/580)

PR #4891, opened 2022-05-13
Run chromatic action on PRs and main branch

jamesgorrie:
## Why?
We run Chromatic on all pushes to any branch. The only places we use and confirm these tests are PRs and merges to main. This should be a no-op in terms of process and hopefully run Chromatic less in places where we never use it.

We've used something similar in Source:
https://github.com/guardian/source/blob/main/.github/workflows/chromatic.yml#L2-L6

Co-authored-by: @sndrs <alex.sanders@theguardian.com>

## Screenshots

| Before      | After      |
|-------------|------------|
| 💰 💰 💰 | 💰 💰  |

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,6 +1,9 @@
 name: Chromatic 👓
-on: push
-
+on:
+  push:
+    branches:
+      - main
+  pull_request:
 jobs:
   chromatic:
     name: Chromatic


AshCorr (approved):
👍

sndrs:
🤞

oliverlloyd:
Chromatic recommend [not doing this](https://www.chromatic.com/docs/github-actions#recommended-configuration-for-build-events).

> The pull-request event requires special consideration. Like other CI systems, GitHub allows workflow execution on either commits pushed to a branch in a pull request. Or for “merge” commits between that branch and the base branch (main).
>
> These specific types of commits (merge) don’t persist in the history of your repository. That can cause Chromatic’s baselines to be lost in certain situations. Hence why we recommend running Chromatic’s step on push.

This isn't specifically clear when the 'certain situations' are but I think we should understand this before merging this PR.

oliverlloyd:
> I think we should understand this before merging this PR

I sent a message to Chromatic asking them if they have any more information on why this might be unadvisable or if this is in fact fine for us to do. Hopefully they will join in here!

jamesgorrie:
> Just to check, once a PR has been opened, any push will trigger the action?

@OllysCoding from a GitHub actions POV, yes. 

> Runs your workflow when activity on a pull request in the workflow's repository occurs. For example, if no activity types are specified, the workflow runs when a pull request is opened or reopened or **when the head branch of the pull request is updated**.

-- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request (My boldness added to the comment)

jamesgorrie:
@oliverlloyd I can't say I understand "That can cause Chromatic’s baselines to be lost in certain situations" any more. This is the way source are doing it, but perhaps they weren't aware of this. Ping @sndrs @SiAdcock.

I think we should either experiment with this or Turbosnap to allow us to start adding more coverage e.g. different breakpoints which we are currently holding back on. Do we have a preference? TurboSnap seems longer term. We can always undo these changes if they are insufficient.

ashishpuliyel (approved):
I spoke to the chromatic rep. To reduce 'wasted' usage he suggested two things:

1. Run chromatic on only PRs, and all subsequent commits to PRs. Exact configuration incoming. But it sounds like PR will do what we want reasonably safely, and reliably, without much in the way of downsides.
2. Enable TurboSnap. He said it's very reliable, but may take a small amount of configuration at first. It errs on the side of caution (running more tests when it can't be certain if things could have changed - for example any change to yarn.lock will re-run everything). Teams get a between 20-80% saving using it (which is such a vague number it's not particularly useful).

He did say we shouldn't do both of these things at the same time, and should do them in sequence to ensure all any kinks are ironed out (I think he was speaking particularly with reference to turbosnap here). 

This is a fairly conclusive confirmation that @jamesgorrie you're entirely on the right track here. And I agree that this seems like an easy-to-try, low risk change. Go for it.

jamesgorrie:
> These specific types of commits (merge) don’t persist in the history of your repository. That can cause Chromatic’s baselines to be lost in certain situations.

@SiAdcock @oliverlloyd I think the [docs for TurboSnap](https://www.chromatic.com/docs/turbosnap#github-pullrequest-triggers) make it more clear why:

> The reason is that `pull_request` workflows run against an ephemeral merge commit, which doesn’t actually exist in your Git history yet, but would if you were to merge the PR at that point.

You can see the code doing that [here](https://github.com/chromaui/chromatic-cli/blob/63edc4ca1d1ed8f7213c081dc13115057eff37ea/action-src/main.ts#L27-L36).

If this doesn't reduce cost to free us up to add more coverage, or breaks the workflow we can always try out Turbosnap. Or in the name of testing run this for a month, and if we have no workflow issues, try Turbosnap after to see what the differences are. Maybe something for DexX/Source to try @SiAdcock?

jamesgorrie:
>  too much impact

@oliverlloyd off the top of your head do you know anywhere it might affect us?

AshCorr:
> If this doesn't reduce cost to free us up to add more coverage, or breaks the workflow we can always try out Turbosnap. Or in > the name of testing run this for a month, and if we have no workflow issues, try Turbosnap after to see what the differences ? > are. Maybe something for DexX/Source to try @SiAdcock?

https://github.com/guardian/dotcom-rendering/pull/4768

Turbosnap seems to work OK with our codebase (seems to handle things like common-rendering and Islands fine) but it does have an issue when Storybook stories are created dynamically using `storiesOf` which we use for generating all of the different layout combinations.

Last I heard the Storybook team were working on something that might solve this for us: https://github.com/storybookjs/storybook/issues/9828

jamesgorrie:
Raising the point here so it doesn't get lost:

> I almost always use it on merge I think

-- @JamieB-gu

PR #4893, opened 2022-05-13
Stop relying on star rating to parse Design.Review

### LINKED ISSUE #4539: Stop Relying On Star Rating To Parse `Design.Review`
AR currently mandates that a piece has to have a star rating to be classed as a review:

https://github.com/guardian/dotcom-rendering/blob/09f166c66d56661adab97473e34e9e9a9a42c990/apps-rendering/src/item.ts#L398

Unfortunately it turns out that some review pieces don't include a star rating, but still pick up the rest of the `Design.Review` design treatment: bold, themed headline, avatar picture, light standfirst etc. See the difference here: [AR](https://mobile.guardianapis.com/rendered-items/books/2022/apr/03/in-brief-the-secret-world-of-weather-song-for-the-missing-espana-book-reviews), [Dotcom](https://www.theguardian.com/books/2022/apr/03/in-brief-the-secret-world-of-weather-song-for-the-missing-espana-book-reviews).

We'll want to remove this check and refactor the definition of `Review` to alter the `starRating` field to `Option<number>`:

https://github.com/guardian/dotcom-rendering/blob/09f166c66d56661adab97473e34e9e9a9a42c990/apps-rendering/src/item.ts#L94-L98

MarSavar:
## Why?
Currently, AR uses star rating to determine whether an article is of type `Review`. However, there are articles with no star rating that should still be rendered with the `Review` design and aren't being captured by the current logic.

This PR changes the way we parse `ArticleDesign.Review` using tags instead of star rating. Closes #4539 

## Screenshots
Here is an example of a review article with no star rating.
| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/57295823/168320066-0251b091-0fae-45c1-8bcc-9664385a556a.png
[after]: https://user-images.githubusercontent.com/57295823/168320120-e24f905a-0de0-417c-8bed-ce3bce508647.png

### DIFF
--- a/apps-rendering/src/capi.ts
+++ b/apps-rendering/src/capi.ts
@@ -48,9 +48,6 @@ const isPhotoEssay = (content: Content): boolean =>
 const isFeature = (content: Content): boolean =>
 	content.tags.some((tag) => tag.id === 'tone/features');
 
-const isReview = (content: Content): boolean =>
-	[0, 1, 2, 3, 4, 5].includes(content.fields?.starRating ?? -1);
-
 const isAnalysis = (content: Content): boolean =>
 	content.tags.some((tag) => tag.id === 'tone/analysis');
 
@@ -227,7 +224,6 @@ export {
 	isImmersive,
 	isInteractive,
 	isFeature,
-	isReview,
 	isAnalysis,
 	articleSeries,
 	articleContributors,
--- a/apps-rendering/src/components/editions/starRating/index.tsx
+++ b/apps-rendering/src/components/editions/starRating/index.tsx
@@ -5,6 +5,7 @@ import { ArticleDesign } from '@guardian/libs';
 import { brandAltBackground, brandAltLine } from '@guardian/source-foundations';
 import { SvgStar } from '@guardian/source-react-components';
 import type { Item } from 'item';
+import { maybeRender } from 'lib';
 import type { FC, ReactNode } from 'react';
 
 // ----- Subcomponents ----- //
@@ -54,9 +55,11 @@ const containerStyles = css`
 `;
 
 const StarRating: FC<Props> = ({ item }) =>
-	item.design === ArticleDesign.Review ? (
-		<div css={containerStyles}>{stars(item.starRating)}</div>
-	) : null;
+	item.design === ArticleDesign.Review
+		? maybeRender(item.starRating, (starRating) => (
+				<div css={containerStyles}>{stars(starRating)}</div>
+		  ))
+		: null;
 
 // ----- Exports ----- //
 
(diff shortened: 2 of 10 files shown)


apps-rendering/src/capi.ts
```diff
@@ -48,9 +48,6 @@ const isPhotoEssay = (content: Content): boolean =>
 const isFeature = (content: Content): boolean =>
 	content.tags.some((tag) => tag.id === 'tone/features');
 
-const isReview = (content: Content): boolean =>
-	[0, 1, 2, 3, 4, 5].includes(content.fields?.starRating ?? -1);
-
```
MarSavar, line 53: This function was only being used to perform a check in `relatedContent.ts`.
A separate function checking for tags  `tone/review`, `tone/livereview`, `tone/albumreview` already exists in `item.ts` so this is now redundant).

apps-rendering/src/relatedContent.ts
```diff
@@ -23,6 +22,7 @@ import {
 	isLabs,
 	isLetter,
 	isLive,
+	isReview,
```
MarSavar, line 25: This is the function mentioned previously that checks for any of these tags: `tone/review`, `tone/livereview`, `tone/albumreview`

apps-rendering/src/components/editions/starRating/index.tsx
```diff
@@ -54,9 +55,11 @@ const containerStyles = css`
...
 
 const StarRating: FC<Props> = ({ item }) =>
-	item.design === ArticleDesign.Review ? (
-		<div css={containerStyles}>{stars(item.starRating)}</div>
-	) : null;
+	item.design === ArticleDesign.Review
+		? maybeRender(item.starRating, (starRating) => (
+				<div css={containerStyles}>{stars(starRating)}</div>
+		  ))
+		: null;
```
MarSavar, line 62: This ternary conditional may be redundant because `maybeRender` already deals with the possibility of `starRating` not being there, but it's there to enforce the rendering of stars only if `Design === Review` (is it even possible to have stars on non-review articles?)

JamieB-gu: > (is it even possible to have stars on non-review articles?)
I hope not.

JamieB-gu: Ah, if you don't perform this check does TypeScript complain that `starRating` may not exist on `Item`?

MarSavar: Good shout - it does complain! So it's not a bad idea to keep the ternary there.

PR #4896, opened 2022-05-16
Add key event carousel 

abeddow91:
## What does this change?
Adds a new carousel component and story for key events on liveblogs. The carousel is scroll only until desktop when buttons are also revealed. ~~The swipe and button nav are kept in sync using an intersection observer and the carousel is fully navigable by keyboard.~~ Following a discussion with UX/Design and Engineering, we have simplified this component as it sits above the fold and we wanted as little JS as possible. The trade off of this is this component does not provide any visual feedback to the user when the carousel reaches either end and the buttons remain clickable. 

This PR adds the component and stories but does not implement the component which will need to be behind a switch before going into production. 

## Why?
As part of the liveblog redesign, the key events have been redesigned into a carousel.  

### Desktop
https://user-images.githubusercontent.com/20416599/168549933-1219475f-a431-4dd2-acad-c8897e657328.mov

### Mobile / Tablet
https://user-images.githubusercontent.com/20416599/168549786-f2b09ca0-31d5-4ed0-abd3-d4e60f6fa53f.mov


dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -807,6 +807,7 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 										) : (
 											<></>
 										)}
+
```
joecowton1, line 810: Nit  - can we remove this?

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -0,0 +1,204 @@
...
+		const carouselElement = document.getElementById('key-event-carousel');
+		const cardsArray = Array.from(
+			document.querySelectorAll('#key-event-card'),
+		);
+		if (carouselElement) setCarousel(carouselElement);
+		if (cardsArray) setCards(cardsArray);
+	}, []);
+
+	useEffect(() => {
+		function handleIntersect(entries: IntersectionObserverEntry[]) {
```
joecowton1, line 123: Does it make sense for this to be an arrow function, inline with the rest of the functions in this component?

joecowton1 (commented):
Really nice work!

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+	const goPrevious = () => {
+		if (carousel && !isFirstCard) carousel.scrollLeft -= cardWidth;
+	};
+
+	const goNext = () => {
+		if (carousel && !isLastCard) carousel.scrollLeft += cardWidth;
+	};
+
+	const transformedKeyEvents = keyEvents
```
abeddow91, line 149: This has been taken from the existing key events implementation on DCR. Should this transformation live within the key event card itself rather than here in the carousel?

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
abeddow91: Yep, good spot, thanks!

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -0,0 +1,204 @@
...
+							date={keyEvent.date}
+							format={format}
+						/>
+					);
+				})}
+			</span>
+			<Hide until="desktop">
+				{keyEvents.length > 6 && (
+					<span css={buttonContainerStyles}>
+						<button
```
oliverlloyd, line 177: I suspect there's a good reason for this - I vaguely recall this was discussed before so sorry if I'm repeating myself - but just to be sure, are we able to use `Button` from source here?

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -0,0 +1,204 @@
...
+							type="button"
+							aria-label="Move key events carousel backwards"
+							onClick={goPrevious}
+						>
+							<SvgChevronLeftSingle />
+						</button>
+						<button
+							css={[
+								buttonStyles,
+								isLastCard && disabledButtonStyles,
```
oliverlloyd, line 191: There's an argument we shouldn't style controls as disabled. See this note on this [in Source](https://theguardian.design/2a1e5182b/p/435225-button/t/78e328)

>  “It is a bad practice to disable buttons. Disabling the button prevents our chance to tell the user what is wrong. The user keeps clicking the button and is totally in the dark why nothing happens. Keep the button enabled. Let the user click the button. Then show the message why it can’t proceed.” - [A Definitive Guide to Sensible Form Validations ](https://a%20definitive%20guide%20to%20sensible%20form%20validations/)

One could argue this is specific to forms, and this could well be the case here. Has there already been some UX input on this or is this something we can get help on?

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -0,0 +1,204 @@
...
+		.map((keyEvent) => {
+			return {
+				text: keyEvent.title || '', // We fallback to '' here purely to keep ts happy
+				url: `?filterKeyEvents=${filterKeyEvents}&page=with:block-${keyEvent.id}#block-${keyEvent.id}`,
+				date: new Date(keyEvent.blockFirstPublished || ''), // We fallback to '' here purely to keep ts happy
+			};
+		});
+	return (
+		<div id="key-event-carousel" css={carouselStyles}>
+			<span css={containerStyles}>
```
oliverlloyd, line 162: If the `KeyEventCard` is an `li` then this element should be a `ul` 🙏

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
abeddow91: Ah good spot, I think we should be fine to use Source here.

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
abeddow91: This is interesting! At the moment, most of the JS comes from needing to keep track of the index to add disabled styling when we reach either end of the carousel. If we were to remove any feedback that we'd reached the beginning and end of the carousel then we could remove most of the JS in this component (as we don't need to track the index) but if we were to show a different type of feedback (such as a message as referenced above) then we would still need some mechanism (and presumably therefore JS) to keep track of the index.

As for UX input, the disabled styling is certainly in the designs which also reflect how the buttons in the existing carousel work, so moving away from this would probably require some input from either @benwuersching or @zeek01

joecowton1: If we weren't disabling the buttons what user feedback would we give? Or is it ever acceptable that an active button just does nothing?

abeddow91: And if we are to provide some kind of feedback, I think we would still need some mechanism (probably state and useeffects?) to keep track of the index.

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -0,0 +1,204 @@
...
+`;
+
+const buttonStyles = css`
+	border: 0 none;
+	border-radius: 100%;
+	height: 34px;
+	width: 34px;
+	cursor: pointer;
+	padding: 0;
+	background-color: ${neutral[0]};
```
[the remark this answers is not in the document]
abeddow91: No this will stay the same across all blogs

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -0,0 +1,204 @@
...
+		})
+		.map((keyEvent) => {
+			return {
+				text: keyEvent.title || '', // We fallback to '' here purely to keep ts happy
+				url: `?filterKeyEvents=${filterKeyEvents}&page=with:block-${keyEvent.id}#block-${keyEvent.id}`,
+				date: new Date(keyEvent.blockFirstPublished || ''), // We fallback to '' here purely to keep ts happy
+			};
+		});
+	return (
+		<div id="key-event-carousel" css={carouselStyles}>
```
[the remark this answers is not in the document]
abeddow91: This isn't possible as far as I can tell, I have tried a few variations but we need two here so that one controls the hidden overflow of blocks and the other allows for overflow scrolling

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
abeddow91: Following a chat with UX and UI, its been decided we can move forward here without visual feedback or disabling and strip out most of the js here as this sits above the fold. We should keep an eye on this though that a) this doesn't become an assumed design pattern moving forward - UX and UI were ok with this in isolation, but not for the whole site - and b) that we monitor any user feedback that we receive to make sure we're not having a negative impact.

oliverlloyd: Thanks for summarising this so well Anna, this makes sense and sounds like a reasonable approach.

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -0,0 +1,204 @@
...
+	}
+`;
+
+const disabledButtonStyles = css`
+	background-color: ${neutral[60]};
+	cursor: default;
+
+	&:hover,
+	&:focus {
+		background-color: ${neutral[60]};
```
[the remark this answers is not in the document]
oliverlloyd: Gotcha 👍

dotcom-rendering/src/web/components/KeyEventCard.tsx
```diff
@@ -52,6 +52,7 @@ const listItemStyles = css`
 	list-style: none;
 	display: block;
 	width: 162px;
+	scroll-snap-align: start;
 
 	${from.desktop} {
 		background-color: ${neutral[93]};
```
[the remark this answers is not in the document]
abeddow91: Thanks Oliver! Myself and @silvacb have move these references into the decide palette file

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -0,0 +1,133 @@
...
+	filterKeyEvents,
+	format,
+}: Props) => {
+	const cardWidth = 200;
+	const goPrevious = () => {
+		console.log(carousel);
+		if (carousel) carousel.scrollLeft -= cardWidth;
+	};
+	const goNext = () => {
+		console.log(carousel);
```
joecowton1, line 79: And this?

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -589,8 +590,22 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 							</Hide>
 						</GridItem>
 					</StandFirstGrid>
-				</ElementContainer>
 
+					<Hide until="desktop">
```
joecowton1, line 594: Do we need both of these?

yarn.lock
```diff
@@ -4918,10 +4918,10 @@
     "@types/minimatch" "*"
     "@types/node" "*"
 
-"@types/google.analytics@^0.0.41":
-  version "0.0.41"
-  resolved "https://registry.yarnpkg.com/@types/google.analytics/-/google.analytics-0.0.41.tgz#7905553890bb8651c93927c1ce4c1c4a26e41505"
-  integrity sha512-LE6AfJFJtxUl1d56oQ1PvQDEOtU+0fgqTVgmOw+T0eaz7FB4DTqKxbI6x8sq+7FDvAeIXOugZyGHJLR4CAxO4A==
+"@types/google.analytics@^0.0.42":
```
joecowton1, line 4921: If we merge in `main` do theses diffs go?

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -0,0 +1,138 @@
...
+`;
+
+const rightButton = css`
+	right: ${space[5]}px;
+`;
+
+const isServer = typeof window === 'undefined';
+const carousel: HTMLElement | null = !isServer
+	? window.document.getElementById('key-event-carousel')
+	: null;
```
OllysCoding, line 72: Can this be achieved with [refs/useRef hook](https://reactjs.org/docs/hooks-reference.html#useref)?

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
OllysCoding: I think these transformations originally existed because key event card was sitting in common-rendering and we needed a shared interface, from what I can tell now the code is in DCR alone, it'd be much nicer to have this logic moved into the card component :)

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -0,0 +1,138 @@
...
+	const goPrevious = () => {
+		if (carousel) carousel.scrollLeft -= cardWidth;
+	};
+	const goNext = () => {
+		if (carousel) carousel.scrollLeft += cardWidth;
+	};
+	const transformedKeyEvents = keyEvents
+		.filter((keyEvent) => {
+			return keyEvent.title && keyEvent.blockFirstPublished;
+		})
```
mxdvl, line 90: You can turn this into a type predicate to prevent having to keep TS happy below.

```suggestion
		.filter((keyEvent): keyEvent is Block & { title: string, blockFirstPublished: number } => {
			return keyEvent.title && keyEvent.blockFirstPublished;
		})
```

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
mxdvl: May I suggest using the type predicate pattern that’s present in `keyEvents`, too?

https://github.com/guardian/dotcom-rendering/blob/68aca01b92dcd71777ca0da980d0ad26c43b2ac1/dotcom-rendering/src/web/components/KeyEventsContainer.tsx#L28

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -0,0 +1,138 @@
...
+		.map((keyEvent) => {
+			return {
+				text: keyEvent.title || '', // We fallback to '' here purely to keep ts happy
+				url: `?filterKeyEvents=${filterKeyEvents}&page=with:block-${keyEvent.id}#block-${keyEvent.id}`,
+				date: new Date(keyEvent.blockFirstPublished || ''), // We fallback to '' here purely to keep ts happy
+				isSummmary: keyEvent.attributes.summary,
+			};
+		});
+	return (
+		<div id="key-event-carousel" css={carouselStyles(palette)}>
```
oliverlloyd, line 100: Huge nitpick but does it make more sense if the id is pluralised, like:

```suggestion
		<div id="key-events-carousel" css={carouselStyles(palette)}>
```

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -0,0 +1,138 @@
...
+	format,
+}: Props) => {
+	const palette = decidePalette(format);
+	const cardWidth = 200;
+	const goPrevious = () => {
+		if (carousel) carousel.scrollLeft -= cardWidth;
+	};
+	const goNext = () => {
+		if (carousel) carousel.scrollLeft += cardWidth;
+	};
```
oliverlloyd, line 86: 😻

dotcom-rendering/src/web/components/KeyEventCard.tsx
```diff
@@ -106,7 +107,7 @@ export const KeyEventCard = ({
 	const palette = decidePalette(format);
 
 	return (
-		<li css={listItemStyles}>
+		<li css={listItemStyles(palette)} id="key-event-card">
```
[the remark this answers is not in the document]
oliverlloyd: 🤔 that said, are we using this id anywhere?

oliverlloyd (approved):
Really nice work on this. There was a lot of different people giving input on this but you managed it all really well.

I've made a few comments and see some others have some questions but I'm happy to leave this with you to deal with and merge when you're ready

dotcom-rendering/src/web/components/KeyEventCard.tsx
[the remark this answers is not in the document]
abeddow91: Thanks! This is not being used so is removed

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
abeddow91: Of course!

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -0,0 +1,138 @@
...
+	scroll-behavior: smooth;
+	overflow-x: auto;
+	overflow-y: hidden;
+	display: flex;
+	flex-direction: column;
+	scrollbar-width: none; /* for Firefox */
+	-ms-overflow-style: none; /* for Internet Explorer, Edge */
+	&::-webkit-scrollbar {
+		display: none; /* for Chrome, Safari, and Opera */
+	}
```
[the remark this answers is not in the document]
abeddow91: Oh brilliant! I didn't realise that. Do you mean I can just use `scrollbar-width: none;` here, remove the rest and deploy to code to test?

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
abeddow91: Excellent, thanks @OllysCoding and @mxdvl !

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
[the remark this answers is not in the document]
oliverlloyd: That would be my approach, yeah. Hopefully it just works

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
oliverlloyd: Refs hurt my brains. @OllysCoding is it true that the advantage of a ref is you can be sure that `carousel` is always pointing at the right instance of the DOM in the event that there are re-renders? If so, if we don't have state here then do we need to use one?

oliverlloyd: Hmm, actually this section of the docs is quite useful

> However, useRef() is useful for more than the ref attribute. It’s [handy for keeping any mutable value around](https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables) similar to how you’d use instance fields in classes.
>
> This works because useRef() creates a plain JavaScript object. The only difference between useRef() and creating a {current: ...} object yourself is that useRef will give you the same ref object on every render.

So the question is. Does this component re-render (possible) and does this cause problems for the code using `carousel`?

PR #4884, opened 2022-05-12
Update DCR Schemas

OllysCoding:
Update the DCR schemas cause they were out of date :(

PR #4897, opened 2022-05-16
Add "schema check" action

OllysCoding:
Adds an action which compares a newly generated schema to the existing one. 

This ensures that a user would be notified on their PR if they'd made a change to any types in `index.d.ts` which would change our json schemas, but hasn't run `make gen-schema` to update them.

Removes run-if-changed as this didn't work anyway and the action will replace it.

### DIFF
--- a/.github/workflows/schema-check.yml
+++ b/.github/workflows/schema-check.yml
@@ -1,0 +1,26 @@
+name: DCR Schema Check
+on:
+  push:
+    paths-ignore:
+      - "apps-rendering/**"
+
+jobs:
+  build_check:
+    name: DCR Schema Check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Install Node
+        uses: guardian/actions-setup-node@main
+
+      # Cache npm dependencies using https://github.com/bahmutov/npm-install
+      - uses: bahmutov/npm-install@v1
+
+      - name: Install
+        run: yarn
+
+      - name: Run check-schema script
+        run: make check-schema
+        working-directory: dotcom-rendering
--- a/dotcom-rendering/scripts/json-schema/check-schema.js
+++ b/dotcom-rendering/scripts/json-schema/check-schema.js
@@ -1,0 +1,26 @@
+const path = require('path');
+const fs = require('fs');
+
+const root = path.resolve(__dirname, '..', '..');
+const { generateArticleSchema, generateFrontSchema } = require('./schema');
+
+const existingArticleSchema = fs.readFileSync(
+	`${root}/src/model/json-schema.json`,
+	{ encoding: 'utf-8' },
+);
+const existingFrontSchema = fs.readFileSync(
+	`${root}/src/model/front-schema.json`,
+	{ encoding: 'utf-8' },
+);
+
+const articleSchema = generateArticleSchema();
+const frontSchema = generateFrontSchema();
+
+if (
+	existingArticleSchema !== articleSchema ||
+	existingFrontSchema !== frontSchema
+) {
+	throw new Error('Schemas do not match ... please run "make gen-schema"');
+} else {
+	console.log('Schemas match!');
+}
--- a/dotcom-rendering/scripts/json-schema/check-schema.js
+++ b/dotcom-rendering/scripts/json-schema/check-schema.js
@@ -2,7 +2,7 @@ const path = require('path');
 const fs = require('fs');
 
 const root = path.resolve(__dirname, '..', '..');
-const { generateArticleSchema, generateFrontSchema } = require('./schema');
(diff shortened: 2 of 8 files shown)


dotcom-rendering/scripts/json-schema/check-schema.js
```diff
@@ -0,0 +1,26 @@
+const path = require('path');
+const fs = require('fs');
+
+const root = path.resolve(__dirname, '..', '..');
+const { generateArticleSchema, generateFrontSchema } = require('./schema');
+
+const existingArticleSchema = fs.readFileSync(
+	`${root}/src/model/json-schema.json`,
```
jamesgorrie, line 8: As part of this refactor, we could rename this to `article-schema`?

.github/workflows/schema-check.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Install Node
+        uses: guardian/actions-setup-node@main
+
+      # Cache npm dependencies using https://github.com/bahmutov/npm-install
+      - uses: bahmutov/npm-install@v1
+
+      - name: Install
+        run: yarn
+
+      - name: Run check-schema script
```
jamesgorrie, line 24: Maybe bikeshedding, but for a more streamlined DevX, what about committing as a bot rather than requiring a human to do the task? Something similar to [this](https://lannonbr.com/blog/2019-12-09-git-commit-in-actions).

So a workflow that follows:
* Check schema
* On pass: no-op
* On fail: gen-schema & commit
* Handle any gen-schema errors

dotcom-rendering/scripts/json-schema/schema.js
```diff
@@ -0,0 +1,33 @@
...
+	],
+	{
+		skipLibCheck: true,
+	},
+);
+
+const settings = { rejectDateType: true, required: true };
+
+module.exports = {
+	generateArticleSchema: () => {
```
jamesgorrie, line 19: Just a naming issue, we seem to be overloading `gen` / `generate`. We use `gen` / `generate` here to refer to creating a string from TS. We then use `gen` to refer to writing the files to disk.

Not a biggy, but would `gen-schema` and  `write-schema`, or `get-schema` and `gen-schema` be more clear>?

jamesgorrie (commented):
A question on naming, and maybe a suggestion on improving the DevX for another PR.

.github/workflows/schema-check.yml
OllysCoding: I'm actually really hesitant to do this - 

The schemas generated are verified when articles are loaded, for example - if a new required property was added to index.d.ts, and it turned out that it wasn't always provided from Frontend, DCR would reject the article, 500ing, as it doesn't the schema. 

Therefor I think making sure that people are being purposeful when updating the schema, and are aware of the impact it can have is really important.

Theoretically our tests will catch these cases, but it would feel weird to have a bot commit changes that would make tests fail.

dotcom-rendering/scripts/json-schema/check-schema.js
OllysCoding: I'd actually be tempted to do this in a second PR, and keep this PR solely about the github action!

jamesgorrie: > solely about the github action

That makes sense, but we have done quite a big refactor on gen-schema already... smaller is better though.

jamesgorrie (approved):

.github/workflows/schema-check.yml
jamesgorrie: Thinking about the workflow of this, wouldn't it start with frontend? e.g.
* update frontend types
* update types in DCR
* gen-schema

I'd say `gen-schema` is just a step we have to do once the thinking has been done, as we should be thinking hard if things are going to break when updating `frontend`.

This also exposes the tight coupling we have between the two apps, and that we're more at risk of breaking this in frontend e.g. removing a required field from the models there. I wonder if there something thinking to be done in exposing FE interfaces from there, especially as it becomes more of an API service.

OllysCoding: For day-to-day dev it starts with Frontend, but as we do migration things with fronts - there are times we start typing things in DCR which already exist in Fronts, and missing schema updates can lead to [mistakes](https://github.com/guardian/dotcom-rendering/pull/4884).

This certainly isn't an end-game in DevX or the relationship between DCR and Frontend, but it will ensure that we always have an up-to-date schema based on our types in DCR :)

OllysCoding: (It would be awesome to take a closer look at the type definitions between frontend and DCR and improve the workflows there!)

jamesgorrie: > (It would be awesome to take a closer look at the type definitions between frontend and DCR and improve the workflows there!)

This is definitely what I was getting at...

PR #4899, opened 2022-05-16
🏝 Add "expediteLoading" pattern to Islands

OllysCoding:
🏝Islands🏝 are an awesome way to add interactivity to our page. However their loading mechanism can add lead time to a piece of code being hydrated: 

```mermaid
flowchart LR
    id1("Page download") --> id2("islands.js") --> id3("Look for gu-island") --> id4("MyIsland-importable.js") --> id5H("hydrate(MyIsland)") 
```  

This works for loads of things, and is especially great for lazy-loading with `deferUntil="visible"`. Sometimes though, we need code run 🏎💨 fast 🏎💨 - so I've added an `expediteLoading` prop which allows this!

Under the hood it runs a regex on the react server rendered HTML, and extracts any applicable islands, adding their scripts to the document head!

This changes the load to look like:
```mermaid
flowchart LR
    id1("Page download") --> id2("islands.js") --> id3("Look for gu-island") --> id5("hydrate(MyIsland)") 
    id1 -->  id4("MyIsland-importable.js") -.-> id3
```  

By including `MyIsland-importable` in the document head, the file is downloaded earlier - speeding up the page download -> hydration lead time!

Because these are fetched at render time; this continues the patten that only islands actually on the page will be downloaded, meaning we don't add any excess javascript with this pattern.

---

At this time, I haven't added `expediteLoading` to any islands, but we are aware as a team that as the number of islands increase; hydration may get slower - and this pattern can be used to supercharge those islands which we need to give priority to!


dotcom-rendering/src/web/server/extractIslands.ts
```diff
@@ -0,0 +1,21 @@
+const regex = new RegExp(
+	// Gets the name of any island where expediteLoading is set to 'true'
+	// e.g <gu-island name="MyComponent" props="..." expediteLoading="true"></gu-island>
+	/<gu-island name="([a-zA-Z]+)"[^>]*expediteLoading="true"[^>]*><\/gu-island>/g,
```
mxdvl, line 4: Do we want to [start introducing `RegExp` for html](https://blog.codinghorror.com/parsing-html-the-cthulhu-way/)? I’m tempted to lean towards a proper parser like `sanitize-html`, which we currently use, or `jsdom`, and simply call a `querySelectorAll(gu-island[defer="critical"])` on it.

dotcom-rendering/src/web/server/extractIslands.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const extractExpeditedIslands = (html: string): string[] => {
+	const matches = html.matchAll(regex);
+
+	return unique(
+		Array.from(matches).map((match) => {
+			// MyComponent.importable becomes MyComponent-importable during the build
+			return `${match[1]}-importable`;
+		}),
+	);
```
mxdvl, line 20: Why the `unique` indirection? I would prefer to spread that `Set` directly.

mxdvl (commented):
I would be tempted to keep this as part of the `deferUntil` prop.

Before releasing this, we should set up an AB test to measure any real impact this may have, and assess whether it’s a pattern that introduces real benefits.

dotcom-rendering/src/web/server/extractIslands.ts
OllysCoding: I don't really mind! It was a little hard to parse with so much happening at once for me:
```js
[...new Set(Array.from(matches).map((match) => {
	// MyComponent.importable becomes MyComponent-importable during the build
	return `${match[1]}-importable`;
}),
```
Calling it `unique` or `deduplicate` makes it more clear what the purpose is

mxdvl: This can happen in a subsequent step. And maybe pick spreading or `Array.from`, but not both?

```ts
const islands = [...matches].map(([,match]) => 
	// MyComponent.importable becomes MyComponent-importable during the build
	`${match}-importable`
);
return [...new Set(islands)];
```

OllysCoding:
> I would be tempted to keep this as part of the `deferUntil` prop.

I put some thought into this but struggled to find a way where you could keep the descriptiveness of the interface: 
`deferUntil="expedited"` - this feels like an oxymoron

other options like `loadPriority: "expedited" | "idle" | "visible"` didn't feel as descriptive or clear as `deferUntil="visible"` for example.
  
----

Having its own prop does include some complexity, right now it could be possible to have both `expediteLoading` and `deferUntil="visible"` set, which while would still do what it said on the tin (the script would be ready-to-go when the element is scrolled into view), could also be seen as contradictory - If we wanted I assume we could prevent this with typescript

OllysCoding:
> Before releasing this, we should set up an AB test to measure any real impact this may have, and assess whether it’s a pattern that introduces real benefits.

Definitely, I'd love to see this tested with some of the header/navbar islands like signed in and the support the guardian messaging, though I think it's real value might come from above-the-fold interactive elements like https://github.com/guardian/dotcom-rendering/pull/4896

dotcom-rendering/src/web/server/extractIslands.ts
mxdvl: That’s a valid point… but then again we’re not really measuring any of this, so it’s hard to make an educated guess.

Running the code below on a recent article, it seems that parsing the DOM takes about 6ms on average. I could easily parse it 5000+ times before getting a crash

```js
const html = await fetch(
  "https://www.theguardian.com/football/2022/may/16/chelsea-have-robert-lewandowski-on-radar-despite-favoured-barcelona-move"
).then((r) => r.text());

const parser = new DOMParser();
const doms = [];
performance.mark("parsing-start");
for (let index = 0; index < 1_000; index++) {
  const dom = parser.parseFromString(html, "text/html");
  doms.push(dom);
}
performance.measure("parsing", "parsing-start");
```

mxdvl:
> > I would be tempted to keep this as part of the `deferUntil` prop.
> 
> I put some thought into this but struggled to find a way where you could keep the descriptiveness of the interface: `deferUntil="expedited"` - this feels like an oxymoron
> 
> other options like `loadPriority: "expedited" | "idle" | "visible"` didn't feel as descriptive or clear as `deferUntil="visible"` for example.
>

I personally don’t mind the `hydrate: "preload" | "when loaded" | "when visible" | "when idle"`.

> Having its own prop does include some complexity, right now it could be possible to have both `expediteLoading` and `deferUntil="visible"` set, which while would still do what it said on the tin (the script would be ready-to-go when the element is scrolled into view), could also be seen as contradictory - If we wanted I assume we could prevent this with typescript

I think that “expedite” is maybe a looser term, and we should favour `preload`? This is more explicit about what we’re doing. We can even avoid the attribute having a value and only check for its presence.

We should definitely track which islands are using the attribute in any case, as we’re at a danger of falling into the trap of “once everybody’s special, no one will be”.

OllysCoding:
> I think that “expedite” is maybe a looser term, and we should favour preload? This is more explicit about what we’re doing. We can even avoid the attribute having a value and only check for its presence.

I'd actually steer away from `preload` as this has a [specific meaning in html](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preload) which is not actually what we're doing!

In this case we're simply including the javascript for the island in the document head, like a usual js script, not providing the `rel="preload"` link which can be used to tell browsers to prioritise a download - I'd say we _don't_ want to do this as we don't use it for any other JS on the page, we only use this for fonts!

What we're actually doing is taking the requirement for the `islands.js` script to download the script after it discovers the element on the DOM, instead the browser is aware of the script much earlier in the loading process making it potentially already available when islands.js tries to import it!

dotcom-rendering/src/web/server/extractIslands.ts
OllysCoding: Much tidier! Thank you :)

dotcom-rendering/src/web/server/extractIslands.ts
OllysCoding: Awesome! I shall switch over to this then :)

ioannakok:
Thanks for the excellent description and conversation. For my understanding, could the fact that we want this prop to be used sparingly be a good reason why we would want to keep it separate and not have it as a possible value of `deferUntil` (or something similar)? That would make sense to me but might not be enough.

mxdvl (approved):
This feels like an interesting pattern. Let's get this merged and trial it for some components in an A/B test (server-side).

OllysCoding:
👋  Bringing this PR back to life !

I've updated this PR to use JSDOM instead of regex for the parsing, however I would note the performance implications - 

On my machine:
The regex method took 0.2ms
JSDOM method took 40-100ms

I'd really like to hear people's thoughts on whether the security of using a parsing engine given the performance implications!

dotcom-rendering/src/web/components/Island.tsx
```diff
@@ -9,13 +9,15 @@ interface HydrateProps {
 	clientOnly?: false;
 	placeholderHeight?: never;
 	children: JSX.Element;
+	expediteLoading?: boolean;
 }
 
 interface ClientOnlyProps {
 	deferUntil?: When;
```
oliverlloyd, line 16: This prop started out life being called `when`. We moved to `deferUntil` because it better represented the options at the time but maybe now 'when' makes sense?

oliverlloyd (commented):
Well this is neat!

Do you think there would be a way to test this using Cypress?

oliverlloyd:
> The regex method took 0.2ms
> JSDOM method took 40-100ms

This is a pretty big performance hit, I am really not a fan of regexs and especially not using them to parse html but 100ms... 🐌

dotcom-rendering/src/web/components/Island.tsx
OllysCoding: I did think about having it in a single interface, e.g 
`when="expedited | loaded | idle | visible"` but I didn't think I could find something as concise as deferUntil - for example the default state (in this example loaded) didn't have an obvious name for me?

The other advantage of having it as its own prop is that it's likely to be used more thoughtfully - adding 'expediteLoading' should be a conscious and purposeful choice because if we use it too much it'll affect overall page speeds!

Would be open to new interface suggestions!!

OllysCoding:
> This is a pretty big performance hit, I am really not a fan of regexs and especially not using them to parse html but 100ms... 🐌

I've now updated the PR to use 'htmlparser2' which allows us to parse the dom and get all the expedited islands in ~3-5ms (with a few outliers as high as 50ms when the server first starts) which is much better!

AshCorr (approved):
Looks good! Can't wait to see what the results of this are!

OllysCoding:
Just moving this over from google chat; some ideas on testing were

- Using some existing above-the-fold compoments like Links.importable.tsx
- Trying on show/hide containers on DCR (https://github.com/guardian/dotcom-rendering/pull/4905) to reduce the 'flash' of hidden contaienrs (@AshCorr)

PR #4901, opened 2022-05-16
Rename "json-schema.json" to "article-schema.json"

OllysCoding:
Since we've introduced `front-schema.json`, it makes more sense to call our article schema `article-schema.json`, instead of `json-schema.json`


ioannakok (approved):
+1

mxdvl (approved):

PR #4903, opened 2022-05-16
`InnerBackgroundColour`

oliverlloyd:
## What does this change?
This PR adds the `innerBackgroundColour` prop on `ContainerLayout`. This prop allows the background of the inner container to be set independently of the main, full width, `backgroundColour`

## Why?
We are seeing designs that call for the ability to set the background colour differently to how `ContainerLayout` currently works so we need more flexibility here. We want to be able to sometimes set just the inner background by itself and also to set both at the same time

### DIFF
--- a/dotcom-rendering/src/web/components/ContainerLayout.stories.tsx
+++ b/dotcom-rendering/src/web/components/ContainerLayout.stories.tsx
@@ -95,19 +95,39 @@ export const BackgroundStory = () => {
 	return (
 		<ContainerLayout
 			title="Background Colour"
-			description="About this content"
+			description="When fullWidthBackground is false, the background colour is contained"
 			fontColour={brandBackground.ctaPrimary}
 			showTopBorder={true}
 			sideBorders={true}
 			centralBorder="full"
 			backgroundColour={brandBackground.primary}
 			borderColour={brandBorder.primary}
+			fullWidthBackground={false}
 		>
 			<Grey />
 		</ContainerLayout>
 	);
 };
 BackgroundStory.story = { name: 'with a blue background' };
+
+export const FullWidthBackgroundStory = () => {
+	return (
+		<ContainerLayout
+			title="Background Colour"
+			description="When fullWidthBackground is true (default), then the background colour stretches all the way to the viewport edges, left and right"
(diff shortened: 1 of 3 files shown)


bryophyta (approved):
This looks good to me. The one thing I was going to ask is whether there would be any way to signal that the prop should be a css colour string, and/or whether it was possible to constrain the values, given that the `setBackgroundColour` function seems not to be used very often. However, if I've searched correctly it looks like the other use of `setBackgroundColour` is also in this component file, in which case using it here doesn't add any extra need for documentation on top of the status quo, so my comment feels like it would be a separate issue?

oliverlloyd:
> whether there would be any way to signal that the prop should be a css colour string, and/or whether it was possible to constrain the values

I think I once saw a typescript function that tried to constrain input to valid hex strings. It's was... I'm going to say, interesting? Ideally, we would not allow plain strings, ideally we would constrain input to only tokens from decidePalette or Source, but I think we need more flexibility than this would give us.

It's a common pattern throughout the codebase though, we pass colour as a string in many places. We do use an alias of `type Colour = string` in at least one place but that is itself a subject for debate.

bryophyta:
> > whether there would be any way to signal that the prop should be a css colour string, and/or whether it was possible to constrain the values
> 
> I think I once saw a typescript function that tried to constrain input to valid hex strings. It's was... I'm going to say, interesting? Ideally, we would not allow plain strings, ideally we would constrain input to only tokens from decidePalette or Source, but I think we need more flexibility than this would give us.
> 
> It's a common pattern throughout the codebase though, we pass colour as a string in many places. We do use an alias of `type Colour = string` in at least one place but that is itself a subject for debate.

that makes sense, thanks for the context!

JamieB-gu:
> I think I once saw a typescript function that tried to constrain input to valid hex strings. It's was... I'm going to say, interesting?

It's hard to do this. TypeScript has an upper limit on the size of union types and you tend to run into it trying this 😬. @MarSavar I think found a way around this but it was... complex 🙂

> We do use an alias of `type Colour = string` in at least one place but that is itself a subject for debate.

Yeah I'd agree with this - it's been argued before that type aliasing primitive types may have limited value.

PR #4904, opened 2022-05-16
Make the top right ad sticky on immersive pages

simonbyford:
## What does this change?

Makes the top right ad sticky on immersive pages.

## Why?

There's no reason they shouldn't behave the same way as standard article pages.

## Screenshots

Oh no, what happened to my gifs. You get the idea...

| Before      | After      |
|-------------|------------|
| before (1) | after (1) |


simonbyford:
Speaking to dotcom about this, the reason for this switch statement was to avoid sticky ads overlapping images with immersive weighting, which are common on immersive pages. See:

https://github.com/guardian/frontend/pull/20933

Closing for now...

PR #4905, opened 2022-05-16
DCR Fronts Show/Hide containers

AshCorr:
## What does this change?

Adds the Show / Hide button to containers on fronts to allow users to toggle seeing containers they don't want to see.

### What this PR does not do

 - Disable the show hide button on Labs containers
 - Disable the show hide button on containers with the weather widget
 - Disable the show hide button on special containers such as Thrasher

### Before
![before][]
### After
![after][]

[before]: https://user-images.githubusercontent.com/21217225/168631807-5b8c5ca6-d0c7-48a8-97b5-ac6d28594232.png
[after]: https://user-images.githubusercontent.com/21217225/168632288-2375feb1-c0f0-49f3-bf32-af98fa317d4c.png

### DIFF
--- a/dotcom-rendering/src/web/components/ContainerLayout.tsx
+++ b/dotcom-rendering/src/web/components/ContainerLayout.tsx
@@ -1,14 +1,16 @@
 import { css } from '@emotion/react';
 import type { ArticleFormat } from '@guardian/libs';
 import { ArticleDesign } from '@guardian/libs';
-import { from, space } from '@guardian/source-foundations';
+import { from, space, until } from '@guardian/source-foundations';
 import type { DCRContainerPalette } from '../../types/front';
 import { decideContainerOverrides } from '../lib/decideContainerOverrides';
+import { hiddenStyles } from '../lib/hiddenStyles';
 import { ContainerTitle } from './ContainerTitle';
 import { ElementContainer } from './ElementContainer';
 import { Flex } from './Flex';
 import { Hide } from './Hide';
 import { LeftColumn } from './LeftColumn';
+import { ShowHideButton } from './ShowHideButton';
 
 type Props = {
 	title?: string;
@@ -32,6 +34,7 @@ type Props = {
 	ophanComponentName?: string;
 	ophanComponentLink?: string;
 	containerPalette?: DCRContainerPalette;
+	toggleable?: boolean;
 	innerBackgroundColour?: string;
 	showDateHeader?: boolean;
 	editionId?: EditionId;
@@ -42,6 +45,15 @@ const containerStyles = css`
 	flex-grow: 1;
 	flex-direction: column;
 	width: 100%;
+`;
+
+const headlineContainerStyles = css`
+	display: flex;
+	justify-content: flex-end;
+
+	${until.leftCol} {
+		justify-content: space-between;
+	}
 `;
 
 const margins = css`
@@ -124,12 +136,14 @@ export const ContainerLayout = ({
 	ophanComponentLink,
 	ophanComponentName,
 	containerPalette,
+	toggleable = false,
 	innerBackgroundColour,
 	showDateHeader,
 	editionId,
 }: Props) => {
 	const overrides =
 		containerPalette && decideContainerOverrides(containerPalette);
+
 	return (
 		<ElementContainer
 			sectionId={sectionId}
@@ -170,18 +184,32 @@ export const ContainerLayout = ({
 					stretchRight={stretchRight}
 					format={format}
 				>
-					<Hide when="above" breakpoint="leftCol">
-						<ContainerTitle
-							title={title}
-							fontColour={fontColour || overrides?.text.container}
-							description={description}
-							url={url}
-							containerPalette={containerPalette}
-							showDateHeader={showDateHeader}
-							editionId={editionId}
-						/>
-					</Hide>
-					{children}
+					<div css={headlineContainerStyles}>
+						<Hide when="above" breakpoint="leftCol">
+							<ContainerTitle
+								title={title}
+								fontColour={
+									fontColour || overrides?.text.container
+								}
+								description={description}
+								url={url}
+								containerPalette={containerPalette}
+								showDateHeader={showDateHeader}
+								editionId={editionId}
+							/>
+						</Hide>
+						{toggleable && sectionId !== undefined && (
+							<ShowHideButton
+								sectionId={sectionId}
+								overrideContainerToggleColour={
+									overrides?.text.containerToggle
+								}
+							/>
+						)}
+					</div>
+					<div css={hiddenStyles} id={`container-${sectionId}`}>
+						{children}
+					</div>
 				</Container>
 			</Flex>
 		</ElementContainer>
--- a/dotcom-rendering/src/web/components/FrontPage.tsx
+++ b/dotcom-rendering/src/web/components/FrontPage.tsx
@@ -8,6 +8,7 @@ import { CoreVitals } from './CoreVitals.importable';
 import { FetchCommentCounts } from './FetchCommentCounts.importable';
 import { FocusStyles } from './FocusStyles.importable';
 import { Island } from './Island';
+import { ShowHideContainers } from './ShowHideContainers.importable';
 import { SkipTo } from './SkipTo';
 
 type Props = {
@@ -52,6 +53,9 @@ export const FrontPage = ({ front, NAV }: Props) => {
 			</Island>
 			<Island clientOnly={true} deferUntil="idle">
 				<FetchCommentCounts repeat={true} />
+			</Island>
+			<Island clientOnly={true} expediteLoading={true}>
+				<ShowHideContainers />
 			</Island>
 			<FrontLayout front={front} NAV={NAV} />
 		</StrictMode>
--- a/dotcom-rendering/src/web/components/ShowHideButton.tsx
+++ b/dotcom-rendering/src/web/components/ShowHideButton.tsx
@@ -1,0 +1,50 @@
+import { css } from '@emotion/react';
+import { from, neutral, space, textSans } from '@guardian/source-foundations';
+import { ButtonLink } from '@guardian/source-react-components';
+
+type Props = {
+	sectionId: string;
+	overrideContainerToggleColour?: string;
+};
+
+const showHideButtonCss = (
+	overrideContainerToggleColour: string | undefined,
+) => css`
+	color: ${overrideContainerToggleColour ?? neutral[46]};
+	${textSans.xsmall()};
+
+	margin-top: ${space[2]}px;
+	margin-right: 10px;
+	margin-bottom: ${space[2]}px;
+	position: relative;
+	align-items: bottom;
+
+	${from.wide} {
+		position: absolute;
+		top: 0;
+		right: 0;
+	}
+`;
+
+/**
+ * This component creates the styled button for showing & hiding a container,
+ * The functionality for this is implemented in a single island 'ShownHideContainers.importable'
+ **/
+export const ShowHideButton = ({
+	sectionId,
+	overrideContainerToggleColour,
+}: Props) => {
+	return (
+		<ButtonLink
+			priority="secondary"
+			subdued={true}
+			cssOverrides={showHideButtonCss(overrideContainerToggleColour)}
+			data-link-name="Hide"
+			data-show-hide-button={sectionId}
+			aria-controls={`container-${sectionId}`}
+			aria-expanded={true}
+		>
+			Hide
+		</ButtonLink>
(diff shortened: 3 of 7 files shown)


dotcom-rendering/src/web/components/ContainerLayout.tsx
```diff
@@ -162,15 +176,29 @@ export const ContainerLayout = ({
...
-					<Hide when="above" breakpoint="leftCol">
-						<ContainerTitle
-							title={title}
-							fontColour={fontColour || overrides?.text.container}
-							description={description}
-							url={url}
-						/>
-					</Hide>
-					{children}
+					<div css={headlineContainerStyles}>
```
oliverlloyd, line 179: I think this is fundamentally good and can be left as is if it's working but we might need to think about what happens when there is more content in the left column. That will push the container down here but we might want to force it closed somehow 🤔 

Anyhoo, unless you have any thoughts or ideas you want to try this can be one for a later PR

dotcom-rendering/src/web/components/ShowHideContainer.importable.tsx
```diff
@@ -0,0 +1,123 @@
...
+import { Dispatch, SetStateAction, useEffect, useState } from 'react';
+
+type Props = {
+	sectionId: string;
+	overrideContainerToggle: string | undefined;
+};
+
+type ContainerStates = {
+	value: { [id: string]: string };
+};
```
oliverlloyd, line 13: 👏  for following the pattern used in Frontend here

dotcom-rendering/src/web/components/ShowHideContainer.importable.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+	${from.wide} {
+		position: absolute;
+		top: 0;
+		right: 0;
+	}
+`;
+
+const getContainerStates = (): ContainerStates => {
+	const item = localStorage.getItem(`gu.prefs.container-states`) as
```
oliverlloyd, line 40: We can use [libs/storage](https://github.com/guardian/libs/tree/main/src/storage) here 😎

oliverlloyd (commented):
This is very cool. Really nice work on this. I've made lots of small comments but nothing earth shattering, this approach is solid 👍

dotcom-rendering/src/web/components/ShowHideContainer.importable.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+const toggleContainer = (
+	containerId: string,
+	closed: boolean,
+	setClosed: Dispatch<SetStateAction<boolean>>,
+) => {
+	const containerStates = getContainerStates();
+
+	if (closed) {
+		containerStates.value[containerId] = 'opened';
```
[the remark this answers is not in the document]
OllysCoding: +1 This is definitely the right way to go imo! We don't wanna be putting all that data on the page

dotcom-rendering/src/web/components/ShowHideContainer.importable.tsx
```diff
@@ -0,0 +1,123 @@
...
+		position: absolute;
+		top: 0;
+		right: 0;
+	}
+`;
+
+const getContainerStates = (): ContainerStates => {
+	const item = localStorage.getItem(`gu.prefs.container-states`) as
+		| undefined
+		| string;
```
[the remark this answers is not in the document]
OllysCoding: Would the [@guardian/libs wrapper](https://github.com/guardian/libs/tree/main/src/storage) provide this to us (and should we be using it anyway?)

dotcom-rendering/src/web/components/FrontPage.tsx
```diff
@@ -35,6 +35,9 @@ export const FrontPage = ({ front, NAV }: Props) => {
 						background: ${brandAlt[400]};
 						color: ${neutral[7]};
 					}
+					.hidden {
+						display: none;
+					}
```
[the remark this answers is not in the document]
OllysCoding: +1 Would love to see this either exported as an apply-able style, or if its only going to be doing something on the one element, could you just put it on the container with something like:

```css
css={css`&.hidden { display: none }`}
```

paperboyo:
Just a question: currently, having any containers hidden causes massive delayed layout shift on navigation (reload, back). Is it possible to eliminate it in beautiful world of DCR?

dotcom-rendering/src/web/components/ShowHideContainer.importable.tsx
```diff
@@ -0,0 +1,99 @@
...
+	sectionId: string;
+	overrideContainerToggleColour?: string;
+};
+
+type ContainerStates = { [id: string]: string };
+
+const showHideButtonCss = (
+	overrideContainerToggleColour: string | undefined,
+) => {
+	console.log(overrideContainerToggleColour);
```
bryophyta, line 17: is this console.log here intentionally?

dotcom-rendering/src/web/components/ShowHideContainer.importable.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+	return item;
+};
+
+export const ShowHideContainer = ({
+	sectionId,
+	overrideContainerToggleColour,
+}: Props) => {
+	const [isClosed, setClosed] = useState(
+		getContainerStates()[sectionId] === 'closed',
```
bryophyta, line 53: This looks great!

On quick question: do you know whether getting/setting localstorage is a blocking operation? If not, is there any risk of race conditions between here and ln.79?

bryophyta: (or does `getContainerStates()` only run the first time the component is rendered?)

dotcom-rendering/src/web/components/ShowHideContainer.importable.tsx
AshCorr: Good catch, thanks!

```suggestion
```

dotcom-rendering/src/web/components/ShowHideContainer.importable.tsx
AshCorr: Good question!

It is blocking, getContainerStates will be called once before the component has rendered, it does get called again on line 69 and modified on line 79 but I don't think there can be a race condition as those lines could only be accessed once the component has finished rendering

dotcom-rendering/src/web/components/ShowHideContainer.importable.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+	${from.wide} {
+		position: absolute;
+		top: 0;
+		right: 0;
+	}
+`;
+
+const getContainerStates = (): ContainerStates => {
+	const item = storage.local.get(`gu.prefs.container-states`);
```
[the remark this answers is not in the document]
AshCorr: Hm yes... I hadn't thought of that... I don't think the overhead from reading from localstorage is all that much, but hydrating it will be... Maybe I can turn this component into something similar to the Liveness component for Liveblogs so we only load it in once and use it to hydrate all of the containers on the page.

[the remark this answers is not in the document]
oliverlloyd: > Maybe I can turn this component into something similar to the Liveness component for Liveblogs so we only load it in once and use it to hydrate all of the containers on the page

💯 I think this is going to be the most performant option

I also think you're right that the read from LS won't necessarily be onerous but that hydration of multiple islands is another matter

dotcom-rendering/src/web/components/ShowHideContainer.importable.tsx
bryophyta: that's really helpful, thanks!

dotcom-rendering/src/web/components/ShowHideContainer.importable.tsx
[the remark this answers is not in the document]
AshCorr: Should all be hydrated from a centralized source now!

dotcom-rendering/src/web/components/ContainerLayout.tsx
```diff
@@ -169,17 +183,31 @@ export const ContainerLayout = ({
...
+								fontColour={
+									fontColour || overrides?.text.container
+								}
+								description={description}
+								url={url}
+								showDateHeader={showDateHeader}
+								editionId={editionId}
+							/>
+						</Hide>
+						{toggleable && sectionId !== undefined && (
```
OllysCoding, line 199: Nice to have but, can we introduce a type that requires a sectionId to be set when toggleable is true?

dotcom-rendering/src/web/components/ShowHideButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+			subdued={true}
+			cssOverrides={showHideButtonCss(overrideContainerToggleColour)}
+			data-link-name="Hide"
+			data-show-hide-button={sectionId}
+			aria-controls={`container-${sectionId}`}
+			aria-expanded={true}
+		>
+			Hide
+		</ButtonLink>
+	);
```
OllysCoding, line 45: Could we add some kind of comment to this highlighting that the button functioanlity is added by `ShowHideContainers.importable`?

Something like
```
/**
 * This component creates the styled button for showing & hiding a container,
 * The functionality for this is implemented in a single island 'ShownHideContainers.importable'
 **/
```

dotcom-rendering/src/web/components/ShowHideContainers.importable.tsx
```diff
@@ -0,0 +1,56 @@
...
+		return {};
+	}
+
+	return item;
+};
+
+export const ShowHideContainers = () => {
+	const containerStates = getContainerStates();
+
+	const toggleContainer = (element: HTMLElement) => {
```
OllysCoding, line 19: Could this be defined outside of the component or use `useMemo` so that the function is re-defined on every render (in reality I think this will only render once but it's probably still best practice :))

dotcom-rendering/src/web/lib/hiddenStyles.tsx
```diff
@@ -0,0 +1,7 @@
+import { css } from '@emotion/react';
+
+export const hiddenStyles = css`
+	&.hidden {
+		display: none;
+	}
+`;
```
OllysCoding, line 7: Is there a reason hiddenStyles is defined in its own file?

OllysCoding (approved):
+1 A few small questions but this is really great!

dotcom-rendering/src/web/lib/hiddenStyles.tsx
AshCorr: This was to follow the same pattern that was used for revealStyles
https://github.com/guardian/dotcom-rendering/blob/f7a9bc828b4959510a24715672ff36805db1d742/dotcom-rendering/src/web/lib/revealStyles.ts

dotcom-rendering/src/web/components/ShowHideContainers.importable.tsx
AshCorr: I can take a look at defining it outside, the reason it was inside is because it relies on the containerStates variable... I'm not entirely sure how an Island handles variables if they're defined outside of a component but I'll give it a go!

dotcom-rendering/src/web/components/FrontPage.tsx
```diff
@@ -52,6 +53,9 @@ export const FrontPage = ({ front, NAV }: Props) => {
 			<Island clientOnly={true} deferUntil="idle">
 				<FetchCommentCounts repeat={true} />
 			</Island>
+			<Island clientOnly={true} deferUntil="idle">
```
OllysCoding, line 56: Will deferUntil="idle" add to CLS? Wasn't sure if you wanted to try expediteLoading={true} instead?

AshCorr: I do! Lets give it a go!

dotcom-rendering/src/web/components/ContainerLayout.tsx
AshCorr: Not sure typescript has a way to do this but I can have a poke about.

AshCorr: I don't think theres a good / easy way to do this, feel free to raise a follow up PR though!

oliverlloyd: I think the `Island` component has something like this in it. But to do this it needed multiple sets of props though, which added quite a lot of complexity.

Not sure if there's a better way but I bet @mxdvl knows 😉

PR #4911, opened 2022-05-17
Fix  `PinnedPost` content height with long embeds

joecowton1:
## What does this change?

~Wraps the check for block content height in a `setTimeout` to allow twitter embeds to load before the height is calculated.~

Wraps the check for block content height in a `MutationObserver` so the height is recalculated when changes are made to the target elements. 

## Why?

Previously the content height was only calculated before the embed was loaded, meaning the `showMore` button would not be shown. This meant if a post contained a long embed the post could not be expanded and some of the post was not visible.

@paperboyo reported this as a bug.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/77005274/168791984-825fe97d-36a3-4652-abbe-008891d8e12c.png
[after]: https://user-images.githubusercontent.com/77005274/168792086-e633970f-9118-4f9d-bb0b-dd9ae5527d7d.png

### DIFF
--- a/dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
+++ b/dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
@@ -17,6 +17,10 @@ const pinnedPostContent: HTMLElement | null = !isServer
 	? window.document.querySelector('#collapsible-body')
 	: null;
 
+const contentFitsContainer = () =>
+	pinnedPostContent &&
+	pinnedPostContent.scrollHeight <= pinnedPostContent.clientHeight;
+
 /**
  * hide show more button and overlay on pinned post
  */
@@ -78,11 +82,16 @@ export const EnhancePinnedPost = () => {
 
 	const pinnedPostTiming = useRef<ReturnType<typeof initPerf>>();
 
-	const contentFitsContainer =
-		pinnedPostContent &&
-		pinnedPostContent.scrollHeight <= pinnedPostContent.clientHeight;
+	/**
+	 * Timeout allows embeds to load before calculating content height
+	 */
+	useEffect(() => {
+		const timeout = setTimeout(() => {
+			if (contentFitsContainer()) hideShowMore();
+		}, 1000);
 
-	if (contentFitsContainer) hideShowMore();
+		return () => clearTimeout(timeout);
+	}, []);
 
 	useEffect(() => {
 		pinnedPostCheckBox?.addEventListener('change', handleClickTracking);
--- a/dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
+++ b/dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
@@ -17,10 +17,6 @@ const pinnedPostContent: HTMLElement | null = !isServer
 	? window.document.querySelector('#collapsible-body')
 	: null;
 
-const contentFitsContainer = () =>
-	pinnedPostContent &&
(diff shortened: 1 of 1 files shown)


abeddow91 (approved):
Nice!

dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
```diff
@@ -78,11 +82,16 @@ export const EnhancePinnedPost = () => {
 
 	const pinnedPostTiming = useRef<ReturnType<typeof initPerf>>();
 
-	const contentFitsContainer =
-		pinnedPostContent &&
-		pinnedPostContent.scrollHeight <= pinnedPostContent.clientHeight;
+	/**
+	 * Timeout allows embeds to load before calculating content height
+	 */
+	useEffect(() => {
```
alinaboghiu, line 88: @oliverlloyd tells me this bug may have been introduced a couple of weeks back when they started to defer enhancing tweets until they're in view. So perhaps it may be better to change that logic and not defer enhancing the tweet if it's inside a pinned post? This way we can avoid a timeout hopefully.

oliverlloyd: This is a fiddly one!

I did initially thought that this could have been related to the work we did to defer tweet enhancement but after looking further I don't think this was the issue. I think we've always had this race condition. Essentially, we need to be able to trigger the height recalculation based on the DOM being updated. Luckily, I had a meeting with @mxdvl earlier where he mentioned [mutation observers](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) and I have a feeling this is what we need here.

I tried the code below locally and it worked for the use case that this PR is addressing but I didn't test it for all other situations

```ts
	const checkContentHeight = () => {
		const contentFitsContainer =
			pinnedPostContent &&
			pinnedPostContent.scrollHeight <= pinnedPostContent.clientHeight;
	};

	useEffect(() => {
		if (!pinnedPost) return;
		const observer = new MutationObserver(checkContentHeight);
		observer.observe(pinnedPost);

		return () => observer.disconnect();
	}, []);
```

joecowton1: Thanks @oliverlloyd ! Much appreciate you looking at it. This works beautifully when a tweet starts above the height cut off for a pinned post.

However when tweet is initially hidden, after adding the above suggestion the tweet no longer enhances. Do you have any idea what might be causing that?

An example live blog / pinned post is here:
http://localhost:3030/Article?url=https://www.theguardian.com/politics/live/2022/may/17/northern-ireland-protocol-brexit-liz-truss-boris-johnson-uk-politics-latest 

Or here:
http://localhost:3030/Article?url=https://www.theguardian.com/business/live/2022/may/13/markets-volatile-ftse-wall-street-oil-bitcoin-crypto-recession-worries-business-live

alinaboghiu (approved):
This is very cool

dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
oliverlloyd: Great spot on the config fix 👍

oliverlloyd (approved):

paperboyo:
Most beautiful team effort, Team, thank you! 😘

PR #4917, opened 2022-05-17
Labs liveblogs in DCR

OllysCoding:
## What does this change?

Makes labs liveblogs in DCR look _okay_

## Why?

There has never been a published liveblog ~ but we want to make sure they would look _okay_ before completing our migration

## Screenshots

These screenshots were taking through storybook ~ as this is the only way to see what a labs liveblog would look like!

| type   | Before      | After      |
|-----------|-------------|------------|
| Live | ![before][] | ![after][] |
| Dead | ![before2][] | ![after2][] |

[before]: https://user-images.githubusercontent.com/9575458/168846027-40b3d513-3ce7-4f31-80da-bdd42f0d3f4b.png
[after]: https://user-images.githubusercontent.com/9575458/168845085-5dbe6316-b332-4554-b7db-1cd524f29983.png
[before2]: https://user-images.githubusercontent.com/9575458/168846158-b75419fe-69ae-48dd-b2de-a43d3db37161.png
[after2]: https://user-images.githubusercontent.com/9575458/168846870-f8bfc9c5-fa1e-41ad-9f60-919d413f848f.png


oliverlloyd (approved):
This is fine

It would be _nice_ if the headline and standfirst colours were labby but maybe not having this is an indication that this isn't complete support and if there were ever a situation where we did have this type of blog then it would need additional effort

jlieb10:
Looks good - also thinking labs colors might be better than red, which does feel very 'breaking news'

jlieb10 (approved):

OllysCoding:
> It would be nice if the headline and standfirst colours were labby

I agree, but for the extra required work and based off of the incredible shortage of labs colours to work with:
image


It didn't seem worth the effort for something that may never be published!

Issue #4869, opened 2022-05-11
Card footer layout


oliverlloyd:
Refactor `Card` to add more support for different variations of the footer based on the type of article, its orientation, meta data, etc.

oliverlloyd:
Fixed by https://github.com/guardian/dotcom-rendering/pull/4883

Issue #4705, opened 2022-04-25
Correct padding on mobile cards on fronts


ajbreuer:
The padding on headlines and the meta text on cards on the mobile front is not correct.

Attached figma file shows the details 

https://www.figma.com/file/qYoVxqhuA97d7xR4bVMUpu/DCR-Fronts-mobile-Corrections?node-id=0%3A1
<img width="626" alt="Screenshot 2022-04-25 at 14 34 27" src="https://user-images.githubusercontent.com/1144827/165100274-86223eb0-ccde-41fc-834b-d59e1ad35887.png">

ioannakok:
Bottom-padding will be fixed as part of: https://github.com/guardian/dotcom-rendering/issues/4869

oliverlloyd:
Dupe of https://github.com/guardian/dotcom-rendering/issues/4723?

jamesgorrie:
![nice](https://user-images.githubusercontent.com/31692/221852354-c9b60653-206f-4fd7-87d1-9a00b4490737.png)

Fixed on DCR fronts

PR #4921, opened 2022-05-17
Fix top padding on cards on fronts

ioannakok:
Fixes part of: #4705
Bottom padding will be fixed as part of the work for https://github.com/guardian/dotcom-rendering/issues/4869

Co-authored-by: Harry Fischer <harry.fischer@guardian.co.uk>

## What does this change?
It modifies padding in `CardWrapper`, `HeadlineWrapper` and `ImageWrapper`.

## Why?
To fix the top padding on cards on fronts and match Figma designs.

## Screenshots

The change is applied to the `CardWrapper` so it affects all breakpoints (hence the 839 chromatic diffs I had to accept!) but it will be mostly visible in mobile breakpoints.

| Before      | After      |
|-------------|------------|
| image | image |
| image | image |

### DIFF
--- a/dotcom-rendering/src/web/components/Card/components/CardWrapper.tsx
+++ b/dotcom-rendering/src/web/components/Card/components/CardWrapper.tsx
@@ -17,7 +17,7 @@ const cardStyles = (format: ArticleFormat, palette?: Palette) => {
 		flex-direction: column;
 		justify-content: space-between;
 		width: 100%;
-		/* We absolutely position both the 1 pixel top bar below and the faux link
+		/* We absolutely position the faux link
 		so this is required here */
 		position: relative;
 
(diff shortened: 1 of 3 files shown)


mxdvl:
Is this not just a case of having a top-padding on the whole container, and not margin for the image or text?

ioannakok:
> Is this not just a case of having a top-padding on the whole container, and not margin for the image or text?

You are absolutely right - description updated

dotcom-rendering/src/web/components/Card/components/CardWrapper.tsx
```diff
@@ -17,18 +17,14 @@ const cardStyles = (format: ArticleFormat, palette?: Palette) => {
 		flex-direction: column;
 		justify-content: space-between;
 		width: 100%;
-		/* We absolutely position bioth the 1 pixel top bar below and the faux link
+		/* We absolutely position both the 1 pixel top bar below and the faux link
```
mxdvl, line 20:
```suggestion
		/* We absolutely position the faux link
```

PR #4931, opened 2022-05-18
Split Up Media Designs - Stage Two: Update `@guardian/libs` To v5

JamieB-gu:
## Why?

We'd like to split up `Design.Media` into `Design.Gallery`, `Design.Audio` and `Design.Video`.

This had to be done over several PRs. This is partly because AR and DCR use the model in `libs` and that had to be updated first. DCR also uses a version of this model maintained in the CAPI client, which needed updating. In addition, frontend passes this version of the model through to DCR, so we had to synchronise the change across that API boundary to make sure DCR temporarily understood both the old and new models that frontend could send. The full list of PRs and the projects they relate to:

- (**AR & DCR**) PR to `libs` to update the version of the model there: https://github.com/guardian/libs/pull/332
- (**DCR**) PR to CAPI client to update the version of the model there: https://github.com/guardian/content-api-scala-client/pull/353
- (**DCR**) PR to temporarily allow DCR to understand the old and new models: #4824 
- (**DCR**) PR to frontend to update to the new model there: https://github.com/guardian/frontend/pull/24986
- (**AR & DCR**) This PR

## Changes

- Bumped `@guardian/libs` to v5 (technically v5.1.0) across AR, DCR and common-rendering
- Refactored all the places that `Design.Media` appeared in AR
- Refactored all the places that `Design.Media` appeared in DCR

### DIFF
--- a/dotcom-rendering/src/web/components/Card/Card.tsx
+++ b/dotcom-rendering/src/web/components/Card/Card.tsx
@@ -230,8 +230,8 @@ export const Card = ({
 							}
 							mediaMeta={
 								(format.design === ArticleDesign.Gallery ||
-								format.design === ArticleDesign.Audio ||
-								format.design === ArticleDesign.Video) &&
+									format.design === ArticleDesign.Audio ||
+									format.design === ArticleDesign.Video) &&
 								mediaType ? (
 									<MediaMeta
 										containerPalette={containerPalette}
--- a/dotcom-rendering/src/web/components/Card/components/CardAge.tsx
+++ b/dotcom-rendering/src/web/components/Card/components/CardAge.tsx
@@ -40,7 +40,8 @@ const ageStyles = (format: ArticleFormat, palette: Palette) => {
 					format.design === ArticleDesign.Gallery ||
 					format.design === ArticleDesign.Audio ||
 					format.design === ArticleDesign.Video
-					? `bold` : `regular`,
+						? `bold`
+						: `regular`,
 			})};
 		}
 	`;


JamieB-gu:
FYI @joecowton1 I think this may include your change from https://github.com/guardian/libs/pull/348 ?

JamieB-gu:
FYI @mxdvl I think this may include your change from https://github.com/guardian/libs/pull/327 ?

apps-rendering/src/item.ts
```diff
@@ -391,9 +384,22 @@ const fromCapi =
 				design: ArticleDesign.Interactive,
 				...itemFieldsWithBody(context, request),
 			};
-		} else if (isMedia(tags)) {
+			// This isn't accurate, picture pieces look different to galleries.
+			// This is to prevent accidentally breaking Editions until we have
+			// a model for pictures.
```
MarSavar, line 389: 👍 Great comment

common-rendering/package.json
```diff
@@ -7,7 +7,7 @@
 	},
 	"dependencies": {
 		"@emotion/react": "^11.4.1",
-		"@guardian/libs": "^3.8.1",
+		"@guardian/libs": "^5.0.0",
```
MarSavar, line 10: Should this be `^5.1.0` like AR?

dotcom-rendering/package.json
```diff
@@ -64,7 +64,7 @@
     "@guardian/commercial-core": "^3.4.0",
     "@guardian/consent-management-platform": "^10.5.0",
     "@guardian/discussion-rendering": "^10.1.1",
-    "@guardian/libs": "^4.0.0",
+    "@guardian/libs": "^5.0.0",
```
MarSavar, line 67: Same question - AR is on `^5.1.0`

MarSavar (approved):
Great work.
Just one question!

common-rendering/package.json
JamieB-gu: This is due to a difference between `npm` and `yarn`. I asked for version `^5.0.0.` from both. `npm` set `5.1.0` in both the `package.json` and the lockfile. `yarn` set `5.0.0` in the `package.json` and `5.1.0` in the lockfile.

Either way, it's always `5.1.0` in the lockfile, so that's the version all three are on.

MarSavar: Makes perfect sense - thanks

PR #4933, opened 2022-05-18
chore(deps-eslint): bump eslint to v8

mxdvl:
Bumps [eslint](https://github.com/eslint/eslint) from 7.32.0 to 8.15.0.

Add ignore to new ESLint errors.

Also bump related packages:

- @typescript-eslint/eslint-plugin
- @typescript-eslint/parser
- eslint-config-airbnb-base@14.2.1
- eslint-config-airbnb-typescript@12.3.1
- eslint-config-prettier@8.1.0
- eslint-import-resolver-babel-module@5.1.2
- eslint-plugin-cypress@2.11.2
- eslint-plugin-import@2.25.4
- eslint-plugin-jsx-a11y@6.4.1
- eslint-plugin-mocha@10.0.3
- eslint-plugin-react@7.23.2
- eslint-plugin-react-hooks@4.2.0

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |


OllysCoding (approved):
Looks good!! 

It's probably too late now but I was thinking since we already have some 'more permanent' eslint ignores across our codebase, I wonder if we wanted to mark which of those came about due to the migration by adding something to the end of the comment?

mxdvl:
> It's probably too late now but I was thinking since we already have some 'more permanent' eslint ignores across our codebase, I wonder if we wanted to mark which of those came about due to the migration by adding something to the end of the comment?

No, that’s an excellent suggestion, and I’ve actually just ignored these rules for now.

PR #4934, opened 2022-05-19
ESLint: Use @guardian-eslint-typescript config

mxdvl:
## What does this change?

Use `@guardian-eslint-typescript` config, but don’t enforce them yet.

The strategy is to treat all of these errors as warnings (`1`) for now. We’ll introduce each rule in individual commits/PRs so they can be reviewed meaningfully.

Note: add ESLint to  `.git-blame-ignore-revs file`

## Why?

We want more uniformity, and Dependabot is eagerly awaiting this.

Related to:
- #4935 
- #4933  

## Errors

```
✖ 2699 problems (0 errors, 2699 warnings)
  0 errors and 2132 warnings potentially fixable with the `--fix` option.
```


OllysCoding (approved):
+1

PR #4942, opened 2022-05-19
Load interactives on page idle

simonbyford:
## What does this change?

Configures interactives to load on page idle instead of when they become visible.

## Why?

Since [10th March](https://github.com/guardian/dotcom-rendering/pull/4077/files#diff-bfe01c729b02a7387fbe09422851114b59f1445c1785129940980a3d0aaebb16R367), interactive elements have been configured to load only when they become visible in the viewport. This makes sense from an efficiency standpoint but presents issues for spacefinder which has to wait for all interactives to load before running. This is because interactives have a variable height which cannot be anticipated in advance, and spacefinder needs to know this to make sensible decisions about where to place ads.

Consequently, any article containing interactives won't show inline ads until either the user scrolls to them, or until spacefinder times out. This happens after 5 seconds for `inline1` ads and after 10 seconds for `inline2+` ads.

It also means the following scenario is possible:

1) Article loads. Interactives do not load. Interactives [occupy a placeholder of 500px in height](https://github.com/guardian/dotcom-rendering/blob/3e89f4fa67bd8671b17b2fdbf79544bb6fcfc54a/dotcom-rendering/src/web/components/InteractiveBlockComponent.importable.tsx#L312) before they are loaded.
2) Spacefinder times out after waiting 10 seconds for interactives to load
3) Spacefinder runs and inserts ads accordingly
4) User scrolls to interactive, which loads. The actual height is less than 500px and the adverts are pulled up, clashing with the "Most viewed" element.

To reproduce: https://www.theguardian.com/politics/2022/apr/08/rishi-sunak-akshata-murty-us-green-cards

## Screenshot

Screenshot 2022-05-19 at 14 54 05

(in this case, the interactive is the "Contact Guardian Politics" block)

## A/B test

To test the impact of this change, we've put the fix behind a server-side A/B test. 

The configuration is here: https://github.com/guardian/frontend/pull/25049

### Step 1

Enable the [switch](https://frontend.code.dev-gutools.co.uk/dev/switchboard#Server-side%20Experiments):

Screenshot 2022-05-26 at 11 54 52

### Step 2

Visit the [CAPI object](https://m.code.dev-theguardian.com/politics/2022/apr/08/rishi-sunak-akshata-murty-us-green-cards.json?dcr) and note:

Screenshot 2022-05-26 at 11 51 29

Visit the [article](https://m.code.dev-theguardian.com/politics/2022/apr/08/rishi-sunak-akshata-murty-us-green-cards) and note:

Screenshot 2022-05-26 at 11 49 40

### Step 3

Opt into the test: 

https://m.code.dev-theguardian.com/opt/in/interactives-idle-loading

> **Note**
> This doesn't work on localhost. Instead use the following "Header Hacker" configuration:
> Screenshot 2022-05-26 at 13 57 34
> Screenshot 2022-05-26 at 13 56 16

### Step 4

Visit the [CAPI object](https://m.code.dev-theguardian.com/politics/2022/apr/08/rishi-sunak-akshata-murty-us-green-cards.json?dcr) and note:

Screenshot 2022-05-26 at 11 47 06

Visit the [article](https://m.code.dev-theguardian.com/politics/2022/apr/08/rishi-sunak-akshata-murty-us-green-cards) and note:

Screenshot 2022-05-26 at 11 50 19

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleBody.tsx
+++ b/dotcom-rendering/src/web/components/ArticleBody.tsx
@@ -30,6 +30,7 @@ type Props = {
 	isSensitive: boolean;
 	isDev: boolean;
 	onFirstPage?: boolean;
+	abTests?: ServerSideTests;
 };
 
 const globalH2Styles = (display: ArticleDisplay) => css`
@@ -110,6 +111,7 @@ export const ArticleBody = ({
 	isSensitive,
 	isDev,
 	onFirstPage,
+	abTests,
 }: Props) => {
 	const isInteractive = format.design === ArticleDesign.Interactive;
 	const palette = decidePalette(format);
@@ -191,6 +193,7 @@ export const ArticleBody = ({
 				isDev={isDev}
 				isAdFreeUser={isAdFreeUser}
 				isSensitive={isSensitive}
+				abTests={abTests}
 			/>
 		</div>
 	);
--- a/dotcom-rendering/src/web/components/CommercialMetrics.importable.tsx
+++ b/dotcom-rendering/src/web/components/CommercialMetrics.importable.tsx
@@ -39,6 +39,8 @@ export const CommercialMetrics = ({ enabled }: Props) => {
 
 		const serverSideTestsToForceMetrics: Array<keyof ServerSideTests> = [
 			/* keep array multi-line */
+			'interactivesIdleLoadingVariant',
+			'interactivesIdleLoadingControl',
 		];
 
 		const userInServerSideTestToForceMetrics =
--- a/dotcom-rendering/src/web/components/CoreVitals.importable.tsx
+++ b/dotcom-rendering/src/web/components/CoreVitals.importable.tsx
@@ -30,6 +30,8 @@ export const CoreVitals = () => {
 
 	const serverSideTestsToForceMetrics: Array<keyof ServerSideTests> = [
 		/* linter, please keep this array multi-line */
+		'interactivesIdleLoadingVariant',
+		'interactivesIdleLoadingControl',
 	];
 
 	const userInServerSideTestToForceMetrics =
--- a/dotcom-rendering/src/web/layouts/CommentLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/CommentLayout.tsx
@@ -584,6 +584,7 @@ export const CommentLayout = ({
 										isPreview={CAPIArticle.config.isPreview}
 										idUrl={CAPIArticle.config.idUrl || ''}
 										isDev={!!CAPIArticle.config.isDev}
+										abTests={CAPIArticle.config.abTests}
 									/>
 									{showBodyEndSlot && (
 										<Island clientOnly={true}>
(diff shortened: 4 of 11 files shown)


lucymonie (approved):
This makes sense as a change - thanks for the detailed explanation. One thing I am wondering about is the Lighthouse report, which says that CLS has been negatively impacted, and I would have expected the opposite. Am I misunderstanding?

mxdvl (commented):
Would it be possible to keep deferring until visible for `adFree` users? I believe this is known server-side.

Also, would it be possible to put his behind an A/B test to compare the impact on web vitals? This could be done in a subsequent step.

mxdvl:
> One thing I am wondering about is the Lighthouse report, which says that CLS has been negatively impacted, and I would have expected the opposite. Am I misunderstanding?

Don't worry about the CLS in that report, @lucymonie : it's been too high for a long time. If you look at the current number for main, you'll see it hasn't changed.

simonbyford:
Thanks for your reviews @lucymonie and @mxdvl!

> Would it be possible to keep deferring until visible for `adFree` users? I believe this is known server-side.

Good idea, I've added this 👍 

> Also, would it be possible to put his behind an A/B test to compare the impact on web vitals? This could be done in a subsequent step.

I've also added this. Server-side test definition here: https://github.com/guardian/frontend/pull/25049

arelra:
> I've also added this. Server-side test definition here: [guardian/frontend#25049](https://github.com/guardian/frontend/pull/25049)

Hey Simon I think we will also need to add this test to the logic that forces 100% sampling for CWV and CommercialMetrics.

Examples here:
CommercialMetrics:
https://github.com/guardian/dotcom-rendering/pull/4506/commits/5c06bc9c4e1f0af881e54294ca395671ba464ae6#diff-6b632a69ca1bb452ecb0f511b62ad18a6d13b421000d9e221b1e4996e7a03a52R42-R46
CWV:
https://github.com/guardian/dotcom-rendering/pull/4506/commits/e109f18d422534ca4ed5abb63fa8eda9a197210f#diff-a2c5b7de377104e4073846caabae488e45a5679902c73df41d67c731152f10d5R26-R30

simonbyford:
> > I've also added this. Server-side test definition here: [guardian/frontend#25049](https://github.com/guardian/frontend/pull/25049)
> 
> Hey Simon I think we will also need to add this test to the logic that forces 100% sampling for CWV and CommercialMetrics.
> 
> Examples here: CommercialMetrics: [5c06bc9#diff-6b632a69ca1bb452ecb0f511b62ad18a6d13b421000d9e221b1e4996e7a03a52R42-R46](https://github.com/guardian/dotcom-rendering/commit/5c06bc9c4e1f0af881e54294ca395671ba464ae6#diff-6b632a69ca1bb452ecb0f511b62ad18a6d13b421000d9e221b1e4996e7a03a52R42-R46) CWV: [e109f18#diff-a2c5b7de377104e4073846caabae488e45a5679902c73df41d67c731152f10d5R26-R30](https://github.com/guardian/dotcom-rendering/commit/e109f18d422534ca4ed5abb63fa8eda9a197210f#diff-a2c5b7de377104e4073846caabae488e45a5679902c73df41d67c731152f10d5R26-R30)

Thanks for this Ravi! I've added them now

chrislomaxjones (approved):
Nice! Looking forward to see the results from this one.

arelra (approved):
👍

PR #4947, opened 2022-05-19
ESLint @guardian: no default export

mxdvl:
## What does this change?

Enable no-default-export


.git-blame-ignore-revs
```diff
@@ -5,3 +5,6 @@
 9394011685c98faf3deab3e5327526941a686630
 e325296f4c0e349e69b0e905e718e8c2f0b9f146
 9523184ba928d242f24dbc91c441a7b9d7635aae
+
+# ESLint big refactor May 2022
+1af2a53b093f0a2e1ae62e817f8cebdc57ae8777
```
oliverlloyd, line 10: 😻

oliverlloyd (approved):
👍 (assuming that vulnerability alert is okay?)

.git-blame-ignore-revs
mxdvl: It’s going to be a chunky one!

PR #4951, opened 2022-05-19
ESLint @guardian: auto-fix problems

mxdvl:
## What does this change?

Fix all auto-fixable problems.

- Turn off rules around object and array types
- Remove unused comments for rules that are off

## Why?

Part of #4935


dotcom-rendering/scripts/jest/setup.ts
```diff
@@ -10,6 +10,10 @@ const windowGuardianConfig = {
 		dcrSentryDsn:
 			'https://1937ab71c8804b2b8438178dfdd6468f@sentry.io/1377847',
 	},
+	ophan: {
+		browserId: 'jest-browser-id',
+		pageViewId: 'jest-page-view-id',
+	},
```
mxdvl, line 16: This is an addition that aligns our test with our type declaration.

dotcom-rendering/src/model/enhance-numbered-lists.ts
```diff
@@ -82,9 +82,9 @@ const extractStarCount = (element: CAPIElement): number => {
 	return starCount;
 };
 
-const isStarableImage = (element: CAPIElement): boolean => {
+const isStarableImage = (element: CAPIElement | undefined): boolean => {
```
mxdvl, line 85: This might be `undefined`.

dotcom-rendering/eslint-guardian.js
```diff
@@ -17,14 +17,14 @@ const guardianRules = {
...
-		WARNING,
+		OFF,
 		{
 			default: 'array-simple',
 		},
 	],
 
 	// use `Record<string, unknown>` instead of `{ [key: string]: unknown }`
-	'@typescript-eslint/consistent-indexed-object-style': [WARNING, 'record'],
+	'@typescript-eslint/consistent-indexed-object-style': [OFF, 'index-signature'],
```
OllysCoding: So just to check, the rule is being turned off, but we are moving to index-signature?

dotcom-rendering/src/model/enhance-numbered-lists.ts
OllysCoding: What change caused this to flag up, or should it have been flagged before?

OllysCoding (approved):
This is a huge PR but everything seems okay at least :)

dotcom-rendering/eslint-guardian.js
mxdvl: We're _keeping_ index-signature. Once we make a decision we can enforce and auto-fix.

PR #4957, opened 2022-05-20
Don't always show card age

### LINKED ISSUE #4955: Restrict when card age is shown
Based on discussions with @HarryFischer and editorial, we only want to show the card age when:

- The card is linking to a liveblog
- The card is inside the Headlines container
- The card is inside the Sports container
- The card is inside a container which has a Palette applied to it

oliverlloyd:
## What does this change?
This fixes #4955 by introducing a function to decide if we should display the card age or not. The rules for when the age should show are:

- The card is linking to a liveblog
- The card is inside the Headlines container
- The card is inside a container which has a Palette applied to it

## String comparison, noooo!
We are using `collection.displayName` to decide if a container is the Headlines container. This isn't great because this logic will break in the event this container get a different name. I think this is unlikely to happen - this name is are pretty static - but it is a risk to this approach.

### Headlines
Screenshot 2022-05-20 at 14 01 38

### Spotlight (no age shown)
Screenshot 2022-05-20 at 14 01 27

### A liveblog inside the Business container
Screenshot 2022-05-20 at 14 00 47


mxdvl (approved):
👍

HarryFischer:
Nice! 👍

HarryFischer:
@oliverlloyd I wonder if we actually don't always display on Sport. If we remove that we'd still get it applied to LiveBlogs which was the main use case

PR #4958, opened 2022-05-20
ESLint reoganise and fix rules

mxdvl:
## What does this change?

- chore(eslint): remove unreported rules
- chore(eslint): fix no-param-reassign
- chore(eslint): correctly map over type
- chore(eslint): enfore no global require
- chore(eslint): enfore no-empty-pattern
- refactor(eslint): reorganise rules in eslint config
- fix(eslint): enforce react/jsx-no-target-blank
- chore(eslint): fix odd error
- chore(eslint): turn one-expression-per-line off

## Why?

End-of-spring cleanup. Part of #4935

### DIFF
--- a/dotcom-rendering/.eslintrc.js
+++ b/dotcom-rendering/.eslintrc.js
@@ -66,34 +66,25 @@ module.exports = {
 
 		'@typescript-eslint/explicit-function-return-type': [0],
 		'@typescript-eslint/no-inferrable-types': [0],
-		// TODO, review these
-		'@typescript-eslint/no-explicit-any': [0],
-		'react/jsx-one-expression-per-line': [0],
-		'react/no-array-index-key': [0],
-		'@typescript-eslint/require-await': [0],
-		'array-callback-return': [0],
-		'consistent-return': [0],
-		'default-case': [0],
-		'global-require': [0],
-		'import/no-extraneous-dependencies': [0],
-		'no-case-declarations': [0],
-		'no-empty-pattern': [0],
-		'no-param-reassign': [0],
-		'no-restricted-syntax': [0],
-		'no-underscore-dangle': [0],
-		'no-useless-escape': [0],
-		'react/button-has-type': [0],
-		'react/jsx-no-target-blank': [0],
-		'react/sort-comp': [0],
-		'react/state-in-constructor': [0],
-		'react/no-danger': [0],
-		'react/jsx-curly-newline': [0],
-		'react-hooks/rules-of-hooks': 'error',
-		'react-hooks/exhaustive-deps': 'warn',
-		'arrow-body-style': [0],
-		'react/require-default-props': [0],
-		'react/jsx-uses-react': 'off',
-		'react/react-in-jsx-scope': 'off',
+
+		// TODO, review these (1314 problems)
+		'import/no-extraneous-dependencies': 'warn', // 683 problems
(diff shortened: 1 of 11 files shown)


dotcom-rendering/.eslintrc.js
```diff
@@ -1,5 +1,34 @@
...
+	'default-case': 'warn', // 50 problems
+	'react/no-danger': 'warn', // 48 problems
+	'no-underscore-dangle': 'warn', // 45 problems
+	'react/no-array-index-key': 'warn', // 34 problems
+	'react/button-has-type': 'warn', // 23 problems
+	'@typescript-eslint/require-await': 'warn', // 22 problems
+	'react/jsx-curly-newline': 'warn', // 8 problems
+	'no-case-declarations': 'warn', // 7 problems
+	'no-useless-escape': 'warn', // 6 problems
+};
```
oliverlloyd, line 16: This is very cool

dotcom-rendering/src/model/enhance-dots.ts
```diff
@@ -2,26 +2,25 @@ import { transformDots } from './transformDots';
 
 const checkForDots = (elements: CAPIElement[]): CAPIElement[] => {
 	// Loop over elements and check if a dot is in the TextBlockElement
-	const enhanced: CAPIElement[] = [];
-	elements.map((element, i) => {
+	const enhanced: CAPIElement[] = elements.map((element, i) => {
```
oliverlloyd, line 5: I think we should keep code changes outside of these kinds of lint PRs. I think for lots of reasons but perhaps mainly as we're going to exclude this commit from the history.

oliverlloyd (approved):
Looks good. Love the organisation of the rules!

I wasn't sure if the enhanced code change was a linting thing or not but that pattern is pretty consistent in all the enhancers so maybe we want to look at that in a separate PR?

dotcom-rendering/.eslintrc.js
mxdvl: https://github.com/guardian/dotcom-rendering/blob/mxdvl/eslint/count-offending-rules/scripts/deno/eslint.ts

dotcom-rendering/src/model/enhance-dots.ts
mxdvl: I actually thing that if we’re changing a handful of elements, it’s okay to include these, because the diff truly is small. This whole PR is only `+56 -60`.

dotcom-rendering/src/model/enhance-dots.ts
```diff
@@ -3,7 +3,7 @@ import { transformDots } from './transformDots';
 const checkForDots = (elements: CAPIElement[]): CAPIElement[] => {
 	// Loop over elements and check if a dot is in the TextBlockElement
 	const enhanced: CAPIElement[] = [];
-	elements.map((element, i) => {
+	elements.forEach((element, i) => {
```
mxdvl, line 6: @oliverlloyd – I’ve kept this in line with other enhancers

dotcom-rendering/src/model/enhance-dots.ts
mxdvl: Yes, as I was saying, I’ve moved to using `forEach` instead.

This is a typical example of doing something functional in a non-functional way and it puzzles me, but I agree that consistency is key. The ESLint rule is very helpful in pointing out us that using `map` without a return statement is **wrong**.

PR #4960, opened 2022-05-20
Implement `inline1ContainerSizing` variant - `inline1` takes full width

arelra:
## What does this change?

Following the completion of the `inline1ContainerSizing` test introduced in https://github.com/guardian/dotcom-rendering/pull/4506, the variant indicates it:

- improves CLS on desktop,  with a range of ~2% - 3% across regions
- increases revenue on desktop, with a range of ~1% - 7% across regions 
- worsens CLS on mobile,  with a range of ~ -2% - -5% across regions

The result for mobile is surprising as in theory it should be the same. We are investigating and will follow up.

`inline1` shifting around on desktop causing the text to reflow is a long standing source of complaints from users and editorial, so on balance we decided to go ahead with the variant.

## Why?

To improve the reading experience for our users and reduce CLS.

## Before

Passback flow - x3 layout shifts:

https://user-images.githubusercontent.com/7014230/162030374-9dcc1f32-bbb5-4918-a280-4590b1d95571.mp4

Teads flow - x2 layout shifts:

https://user-images.githubusercontent.com/7014230/162030531-0b1c7210-2fae-403b-8687-9c10dde489b5.mp4

## After

Passback flow - x2 layout shifts:

https://user-images.githubusercontent.com/7014230/162030708-e7f99e19-77de-4673-b0dd-3834c57e5e0d.mp4

Teads flow - x1 layout shift:

https://user-images.githubusercontent.com/7014230/162030855-2b5750e3-7d3f-449f-a1e0-b3dd47cd6af4.mp4

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleContainer.tsx
+++ b/dotcom-rendering/src/web/components/ArticleContainer.tsx
@@ -48,7 +48,7 @@ const articleWrapper = css`
 	z-index: 1;
 `;
 
-const adSharedCSS = css`
+const adStyles = css`
 	.ad-slot {
 		@media print {
 			/* stylelint-disable-next-line declaration-no-important */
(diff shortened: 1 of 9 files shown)


dotcom-rendering/src/web/components/ArticleContainer.tsx
```diff
@@ -204,16 +161,13 @@ const adVariantCSS = css`
 	}
 `;
 
-export const ArticleContainer = ({ children, format, abTests }: Props) => {
-	const isInline1ContainerSizingVariant =
-		abTests?.inline1ContainerSizingVariant === 'variant';
+export const ArticleContainer = ({ children, format }: Props) => {
```
chrislomaxjones, line 164: Rather than just omitting the optional property `abTests` could we remove it from `Props` altogether? Or is there a reason to have it hanging around?

chrislomaxjones (approved):
This makes sense, just had one minor question.

dotcom-rendering/src/web/components/ArticleContainer.tsx
arelra: Thanks good spot!

PR #4977, opened 2022-05-23
Fix Lighthouse results for Dependabot

mxdvl:
## What does this change?

Fail softly if the comment cannot be updated on a PR in the lighthouse script.

Chore: Format the deno scripts with Prettier

## Why?

All our Dependabot PRs are currently failing because of this, but the actual lighthouse results are fine.

## Screenshots

image


OllysCoding (approved):
LGTM!

jlieb10 (commented):
Excellent, thanks

PR #4978, opened 2022-05-23
ESLint fix import order

mxdvl:
## What does this change?

Auto-fix 56 problems with import orders.

## Why?

Part of #4935

## ESLint status

- 1537 problems (0 errors, 1537 warnings)
- 0 errors and 15 warnings potentially fixable with the `--fix` option.


OllysCoding (approved):
+1

PR #4979, opened 2022-05-23
ESLint: remove DCR import package rules

mxdvl:
## What does this change?

Remove `dcr/only-import-below`.

## Why?

This package is not maintained and there’s no clear value gained from having it.

Part of #4935


OllysCoding (approved):
+1

PR #4985, opened 2022-05-23
Filter out invalid liveblocks

iainjchambers-guardian:
## What does this change?
Filters live blocks by the presence of either `firstPublishedDate` or `lastModifiedDate` to stop them from being passed down the render tree

## Why?
There's a few conditionals down the tree that check for the presence of these dates, this allows us to simplify the code and remove those conditionals by filtering the liveblocks before they're passed down the tree.


apps-rendering/src/components/LiveBlocks/index.tsx
```diff
@@ -38,7 +38,7 @@ const LiveBlocks: FC<LiveBlocksProps> = ({ blocks, format }) => {
 			{/* Accordion? */}
 			{blocks.map((block) => {
 				const blockFirstPublished = pipe(
-					block.firstPublished,
+					some(block.firstPublished),
 					map(Number),
 					toNullable,
 				);
```
JamieB-gu, line 44: I think you might be able to drop this intermediate assignment entirely and pass

```
blockFirstPublished={Number(blockFirstPublished)}
```

directly to the `LiveBlockContainer`?

I think that would also allow flattening this whole block into a single-return arrow function?

```tsx
{blocks.map((block) => (
  <LiveBlockContainer
    key={block.id}
    ...
  >
  ...
  </LiveBlockContainer>
))}
```

JamieB-gu (commented):
Looks great! One suggestion.

JamieB-gu (approved):

PR #4992, opened 2022-05-24
use non-deprecated waitFor method

jamesgorrie:
## What does this change?
Uses the non-deprecated method name. This should help with [the upgrade](https://github.com/guardian/dotcom-rendering/runs/6163291446?check_suite_focus=true).


mxdvl (approved):
Yes!

We should also soon move [`jest-dom`][] to [`@testing-library/jest-dom`][]

[`jest-dom`]: https://www.npmjs.com/package/jest-dom
[`@testing-library/jest-dom`]: https://www.npmjs.com/package/@testing-library/jest-dom

mxdvl:
> @mxdvl I think this upgrade will force that eventually.

Here’s the PR #4994

PR #4999, opened 2022-05-24
Front container date header (& `edition` > `editionId` refactor)

OllysCoding:
First of all apologies in advance for not splitting this PR up into more readable commits ~ it all got a bit muddled!

## What does this change?

Adds the 'date header' to containers on fronts where specified!

We also introduce `lib/edition.ts` - a basic start to what's required for DCR to manage its own dates! As part of this I went through to all the places where `edition` is currently specified in DCR and renamed the uses to `editionId`. I'm hoping this sets a standard of the `edition` object and `editionId` type being easily differentiated :)

## Why?

Part of the fronts migrations

## Screenshots

| Before      | After      |
|-------------|------------|
| N/A | ![after][] |
| N/A | ![after2][] |
| N/A | ![after3][] |

[after]: https://user-images.githubusercontent.com/9575458/170078593-8fbea181-d231-4408-9b74-0c16534ba78f.png
[after2]: https://user-images.githubusercontent.com/9575458/170078769-0af07fd8-06d0-4b9b-9f80-112e3909868f.png
[after3]: https://user-images.githubusercontent.com/9575458/170078874-c407af1e-5e0a-4fe5-af72-0e93330ee256.png

The date takes the kicker colour from the container palette or falls back to `news[400]`


dotcom-rendering/index.d.ts
```diff
@@ -707,7 +714,7 @@ type FEContainerPalette =
 	| 'Breaking'
 	| 'Podcast'
 	| 'Branded'
-	| 'BreakingPalette';
+	| 'Breaking';
```
OllysCoding, line 717: BreakPalette was misnamed so I fixed this - though we need to revisit how palettes work as currently the /uk front has breaking as the first palette and long running alt as the second - and DCR only choses the first one

dotcom-rendering/src/model/extract-ga.ts
```diff
@@ -58,6 +58,6 @@ export const extract = (data: CAPIArticleType): GADataType => ({
 	toneIds: filterTags(data.tags, 'Tone'),
 	seriesId: filterTags(data.tags, 'Series'),
 	isHosted: 'false', // TODO - This is missing from the Frontend Data model
-	edition: data.editionId.toLowerCase() as Edition,
+	edition: data.editionId.toLowerCase() as EditionId,
```
jamesgorrie, line 61: This is odd, as `toLowerCase` would never produce a value that could be an `EditionId`. (I know this is pre-existing, just thought it work pointing out as it's breaking the type system).

dotcom-rendering/src/amp/lib/edition.ts
```diff
@@ -3,6 +3,6 @@
  *
  * @param s The string to test
  */
-export function isEdition(s: string): s is Edition {
+export function isEdition(s: string): s is EditionId {
```
jamesgorrie, line 6:
```suggestion
export function isEditionId(s: string): s is EditionId {
```

jamesgorrie (commented):
To avoid sometimes using `editionId` or `getEditionFromId` in certain components, would it be clearer to set `edition: Edition` at the [client edge (`Layouts`)](https://github.com/guardian/dotcom-rendering/pull/4999/files#diff-4dd02430ab6bd3970cee718aef853c1f5c02af905e06613ef8d0c5671837b5ce) of our application and then just use `edition.id` when we need the raw string?

I think this might actually make this PR way larger, and not sure it adds much more clarity, but wondered what you thought.

dotcom-rendering/src/model/extract-ga.ts
OllysCoding: I've actually just removed the EditionId from this type because as far as I could tell it was both wrong, and unused in 'typed' land anyway.

OllysCoding:
> would it be clearer to set edition: Edition at the [client edge (Layouts)](https://github.com/guardian/dotcom-rendering/pull/4999/files#diff-4dd02430ab6bd3970cee718aef853c1f5c02af905e06613ef8d0c5671837b5ce) of our application and then just use edition.id

I think if we wanted to do this, we might prefer to do it in the [article enhancer](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/server/index.ts#L18) - but currently the enhancer both receives and returns the `CAPIArticleType` - so we couldn't changed editionId to edition on the type.

I think updating the article enhancer to having an incoming type and rendering type (e.g `FrontendArticleType` and `ArticleType` similar to what we're doing on fronts) would be ideal, but beyond the scope of this PR!

For now I may leave it as is, and allow the use of ediitons in Fronts to guide what we might want articles to look like in the future!

dotcom-rendering/src/model/enhanceCollections.ts
```diff
@@ -17,6 +17,9 @@ export const enhanceCollections = (
 			curated: enhanceCards(collection.curated, containerPalette),
 			backfill: enhanceCards(collection.backfill, containerPalette),
 			treats: enhanceCards(collection.treats, containerPalette),
+			config: {
```
jlieb10, line 20: Is it worth expanding this with a note on other configuration options we might anticipate inside this object? My guess is that you do have some in mind, as you didn't surface this on a higher level

dotcom-rendering/src/model/window-guardian.ts
```diff
@@ -6,7 +6,7 @@ export interface WindowGuardianConfig {
 	page: {
 		dcrCouldRender: boolean;
 		contentType: string;
-		edition: Edition;
+		edition: EditionId;
```
jlieb10, line 9: same question as above - why are we leaving the key as edition in this instance?

dotcom-rendering/src/model/window-guardian.ts
```diff
@@ -37,7 +37,7 @@ interface WindowGuardianFrontConfig {
...
```
jlieb10, line 40: idem - why are we leaving the key as edition in this instance?

dotcom-rendering/src/web/components/ContainerTitle.tsx
```diff
@@ -53,18 +66,34 @@ const descriptionStyles = (fontColour?: string) => css`
...
+	containerPalette,
+	showDateHeader,
+	editionId,
+}: Props) => {
 	if (!title) return null;
+
+	const overrides =
+		containerPalette && decideContainerOverrides(containerPalette);
+
+	const now = new Date();
```
jlieb10, line 94: Is there anywhere else we can extract this from, like the publication Date from one of the articles in the container? I imagine it would be... inconvenient, but I'm worried about this going out of sync with the articles in the container. My first thought was because of caching, but another more likely scenario is just that the container won't be updated with all new content at the stroke of (locale-specific) midnight, when this will change

dotcom-rendering/src/web/lib/edition.ts
```diff
@@ -0,0 +1,36 @@
...
+	locale: 'en-au',
+};
+
+const InternationalEdition: Edition = {
+	id: 'INT',
+	displayName: 'International edition',
+	locale: 'en-gb',
+};
+
+export const getEditionFromId = (id: EditionId): Edition => {
```
jlieb10, line 25: Are we anticipating using this file, method, contents, etc only in the one location it is currently used? If not, we could simplify this to something like 

```
export const getEditionFromId = (id: EditionId): Edition => {
switch (id) {
		case 'US':
			return 'en-us';
		case 'AU':
			return 'en-au';
		default:
			return 'en-gb';
	}}
```

jlieb10 (approved):
Olly - amazing work! Non-blocking questions and comments, but clearly well thought through.

dotcom-rendering/src/web/lib/edition.ts
OllysCoding: I would expect that this will be a starting point for migrating date strings from Frontend to DCR, and that we might see this file used more often in the future!

dotcom-rendering/src/model/window-guardian.ts
OllysCoding: As above, where we're staying on window.guardian I'm hesitant to change it!

dotcom-rendering/src/web/components/ContainerTitle.tsx
OllysCoding: > but I'm worried about this going out of sync with the articles in the container

It's been a while since I worked on this exact logic, but I believe that this is the same way it currently works on Frontend, so let me double check!

OllysCoding: Searching through the frontend code I was able to confirm this - https://github.com/guardian/frontend/blob/main/common/app/views/fragments/containers/facia_cards/standardHeaderMeta.scala.html#L35

OllysCoding: There's definitely a margin for error here with the caching though - but I believe this is mostly used on network fronts which are updated very regularly!

PR #5011, opened 2022-05-25
fix: Disable broken Cypress test for Sport liveblogs

AshCorr:
## What does this change?

Disable the Cypress test that checks if a score is updating in a liveblog.

## Why?

We've had this issue for a while now where deadblogs sometimes lose their score component after a certain period of time. I could replace the Deadblog that this test uses with a more recent one but thats just going to break again at some point.

Propose we re-enable this test once the bug with Deadblogs is fixed.

### DIFF
--- a/dotcom-rendering/cypress/integration/parallel-5/liveblog.interactivity.spec.js
+++ b/dotcom-rendering/cypress/integration/parallel-5/liveblog.interactivity.spec.js
@@ -209,45 +209,47 @@ describe('Liveblogs', function () {
 		});
 	});
 
-	it('should render live score updates to a cricketblog', function () {
-		// Get article with ?live to force the article to be 'live'
-		cy.visit(
-			'/Article?url=https://theguardian.com/sport/live/2022/mar/27/west-indies-v-england-third-test-day-four-live?live',
-		);
+	// For some reason Sport deadblogs lose their scores after a certain duration. Not quite sure what causes this just yet
+	// but disabling this test until the issue with Sports blogs is fixed.
+	// it('should render live score updates to a cricketblog', function () {
+	// 	// Get article with ?live to force the article to be 'live'
+	// 	cy.visit(
+	// 		'/Article?url=https://theguardian.com/sport/live/2022/mar/27/west-indies-v-england-third-test-day-four-live?live',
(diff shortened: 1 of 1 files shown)


dotcom-rendering/cypress/integration/parallel-5/liveblog.interactivity.spec.js
```diff
@@ -209,45 +209,47 @@ describe('Liveblogs', function () {
...
+	// 			body: {
+	// 				scorecardUrl: '/test',
+	// 				match: match2,
+	// 			},
+	// 		},
+	// 	);
+
+	// 	// The cricket data is updated every 14 seconds, so we need a longer timeout here
+	// 	// Wait for the 'updated' cricket score to appear on the page correctly!
+	// 	cy.contains('297 & 104 - 3 (25 overs)', { timeout: 20000 });
```
jamesgorrie, line 253: Instead of commenting it out you can use [`it.skip`](https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests#Excluding-and-Including-Tests) - this also reports it as a skipped test which can be useful.

AshCorr: Good call!

dotcom-rendering/cypress/integration/parallel-5/liveblog.interactivity.spec.js
```diff
@@ -209,7 +209,9 @@ describe('Liveblogs', function () {
 		});
 	});
 
-	it('should render live score updates to a cricketblog', function () {
+	// For some reason Sport deadblogs lose their scores after a certain duration. Not quite sure what causes this just yet
+	// but disabling this test until the issue with Sports blogs is fixed.
+	it.skip('should render live score updates to a cricketblog', function () {
```
jamesgorrie, line 214:
```suggestion
	// but disabling this test until the issue with Sports blogs is fixed.
	// see: https://github.com/guardian/frontend/issues/25046
	it.skip('should render live score updates to a cricketblog', function () {
```

jamesgorrie (approved):

PR #5014, opened 2022-05-25
Track scroll depth

mxdvl:
## What does this change?

Add a scroll depth tracker using the `logs.guardianapis.com/log` endpoint. This data is sent once, at the moment the page is left.

## Why?

The Data & Insight team would like to have this data.


oliverlloyd (commented):
Now that we're interested in points on the page to track scroll, should we replace the use of debounced listeners with an intersection observer?

Perhaps with the exception of scroll start, which can perhaps stay as a one time listener, an intersection observer is likely to be more efficient and I suspect make the code simpler. What do you think?

I'm not sure if we need it but if you wanted a more fine grained level of data for exactly where a reader scrolled down to then you could do something like [what commercial did](https://github.com/guardian/commercial-core/blob/bfec510107947bb3a3fb5830d8b76447373cf543/src/track-scroll-depth.ts) and decorate the page with markers at regular intervals and include those in the overall intersection observer (along with the elements on the page with the `data-component` attribute).

PR #5016, opened 2022-05-25
Use Log4js for server-side logging

mxdvl:
## What does this change?

Use the existing server-side logger to send information about SSR failures. When `RewrappedComponent` fails to rewrap a string of html with a new tag, it will send this as a data point ot our elastic stack.

This rarely happens in practice, but it does happen!

## Why?

`RewrappedComponent` failures are currently recorded via the generic logging endpoint, but it is more appropriate to use server-side logging for errors that happen during SSR.


jamesgorrie (approved):
Nice refactor. 1 way to log 'em all!

mxdvl:
Really struggling to get `log4js` not to crash Storybook 😠

PR #5019, opened 2022-05-25
Remove BodyImage component

iainjchambers-guardian:
## What does this change?
Removes BodyImage component
## Why?
Because it's not used anywhere


JamieB-gu (approved):
Good spot!

PR #5020, opened 2022-05-25
Fix PinnedPost content height check

joecowton1:
## What does this change?
Adds a `checkContentHeight` call before the `MutationObserver` is created to ensure the Show More button is rendered correctly. 

## Why?

Currently `checkContentHeight` is only called by the `MutationObserver` - this means that if no DOM mutation occurs, `checkContentHeight` isn't called.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/77005274/170302409-08cc7055-2b8e-47a4-bfaa-4bae7b5d7c8d.png
[after]: https://user-images.githubusercontent.com/77005274/170302512-774e3252-d171-4c11-9b54-38000868dd4f.png


abeddow91 (approved):
Nice fix!

PR #5021, opened 2022-05-25
Capitalise Layout dir

iainjchambers-guardian:
## What does this change?
Capitalises the layout dir and brings it in line structurally with the other components
## Why?
The majority of the directories in the component folder are capitalised now to denote that they are in fact components, layout should be no different, it's a component after all and already half follows the polymorphic structure we've applied to headline and byline. 


JamieB-gu (approved):
This brings this into line with other components so 👍

One thing I have been wondering more generally though is whether we need to repeat the component's directory name in the suffix of every file?

Thinking about directory structure like a URL, GitHub doesn't do `/guardian/dotcom-rendering/pulls/pull-5021` because you already know that `5021` is a PR from the previous path segment. Likewise with `/src/components/Headline/CommentHeadline.tsx` maybe the directory already tells you it's a headline component and we don't need to repeat that in the file name?

So you'd end up with something like this:

```
Headline/
  Comment.tsx
  Standard.tsx
  Immersive.tsx
  index.tsx
```

Thoughts?

Issue #3818, opened 2022-01-11
iOS most pop tabs are blue (Chrome / Safari)


gtrufitt:
![tenor-126691672](https://user-images.githubusercontent.com/638051/148979459-81a8defd-f9fb-4243-8523-64fe248aa677.gif)
![IMG_7746](https://user-images.githubusercontent.com/638051/148979490-12389a0a-1d43-4279-8152-90e53d621874.PNG)

On this article https://www.theguardian.com/lifeandstyle/2022/jan/11/to-all-the-bosses-ive-begged-for-a-job-comic?utm_term=Autofeed&CMP=twt_gu&utm_medium&utm_source=Twitter#Echobox=1641893210

Can't see it on desktop but in the Twitter iOS window and in Chrome iOS they're blue as above.

@OllysCoding @oliverlloyd @rcrphillips

oliverlloyd:
So blue

I was staring at these the other day... Have they always been this way on devices? I feel like this is a new thing

Also, hello!

gtrufitt:
Hi! 👋 

Not that I remember? Tabs have just gone now... no more tabs.

![image](https://user-images.githubusercontent.com/638051/148991352-65ce4852-84d7-40d3-8ed3-af0b91c04b83.png)

oliverlloyd:
We only show tabs when [there is more than one](https://github.com/guardian/dotcom-rendering/blob/2d57e5d80638db1cfef69d441c73c5a54873e9a6/dotcom-rendering/src/web/components/MostViewed/MostViewedFooter/MostViewedFooterGrid.tsx#L128). I don't remember _why_ this is, maybe to save space?

This does not explain why there's only one, nor why it's blue.

mxdvl:
This is due to Webkit defaulting to [`-apple-system-blue`][] for the colour of `button`s.

Most other browser pick a colour that’s close to black, so we never noticed, but it would be best to set this colour explicitely.

[`-apple-system-blue`]: https://github.com/WebKit/WebKit/blob/60a5bb34ca68a176d03301ba5956f64569ced3c9/Source/WebCore/css/CSSValueKeywords.in#L254

PR #5028, opened 2022-05-26
set colour for tabs so it doesn't use browser default

HarryFischer:
## What does this change?
Sets the colour for most popular tabs
## Why?
On iOS Safari they currently appear blue: https://github.com/guardian/dotcom-rendering/issues/3818


oliverlloyd (approved):
Thanks Harry, spot on 👍 

Sorry this took so long @gtrufitt ! I had a great imitator the other day and it was very nice if that at all helps?

gtrufitt:
Your ever lasting support is appreciated @oliverlloyd

mxdvl:
Funny how Safari user-agent is the only one that chooses a non-black colour for `button` in its user-agent stylesheets. For the record, it's `-apple-system-blue`

PR #5040, opened 2022-05-27
feat(enum): add enumEntries method

mxdvl:
## What does this change?

Add a new helper method `enumEntries`, that returns the entries from an enum anyone unfamiliar with the odd workings of TypeScript would expect.

```ts
enumEntries(ArticlePillar);
[
	["News", 0],
	["Opinion", 1],
	["Sport", 2],
	["Culture", 3],
	["Lifestyle",4]
]

// Compare with :
Object.entries(ArticlePillar);
[
	["0", "News"],
	["1", "Opinion"],
	["2", "Sport"],
	["3", "Culture"],
	["4", "Lifestyle"],
	["News", 0],
	["Opinion", 1],
	["Sport", 2],
	["Culture", 3],
	["Lifestyle",4]
];
```

## Why?

This simplifies our loops a lot and removes many type assertions.

### DIFF
--- a/dotcom-rendering/src/web/components/Card/Card.stories.tsx
+++ b/dotcom-rendering/src/web/components/Card/Card.stories.tsx
@@ -1,4 +1,5 @@
 import { css } from '@emotion/react';
+import type { ArticleFormat } from '@guardian/libs';
 import {
 	ArticleDesign,
 	ArticleDisplay,
@@ -8,6 +9,7 @@ import {
 import { breakpoints, from } from '@guardian/source-foundations';
 import { storiesOf } from '@storybook/react';
 import React from 'react';
+import { enumEntries } from '../../layouts/lib/enumEntries';
 import type { Props as CardProps } from './Card';
 import { Card } from './Card';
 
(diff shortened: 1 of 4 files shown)


dotcom-rendering/src/web/components/Card/Card.stories.tsx
```diff
@@ -84,32 +86,26 @@ const CardsWithDifferentThemes = ({
...
-			cards.push({
-				title: pillarName,
-				format: {
-					display,
-					design,
-					theme: pillarValue as ArticlePillar,
-				},
-			});
-		}
+	const cards: Array<{ title: string; format: ArticleFormat }> = [];
```
jamesgorrie, line 89: I really need to scratch my itch and ask - why don't we use `map`?

mxdvl: I am as puzzled as you are, but I think that should live in a subsequent PR.

Maybe the issue with the enum meant that it was hard to filter? Maybe the team is more comfortable with imperative approaches than purely functional ones?

jamesgorrie: We use this pattern all over the codebase. If you search for ` = []` outside of the test files, you'll see what I mean (GitHub search isn't good for this one, so can't share a link).

Screenshot 2022-05-27 at 11 56 13

mxdvl: Yes, I don’t think it’s an ideal pattern. In many of these cases, it would be trivial to have a `map`, especially the enhancers.

[Github CodeSearch gives over 100 matches.](https://cs.github.com/guardian/dotcom-rendering?q=%3D+%5B%5D)

JamieB-gu: There are `map`s used further down this file. It looks like, from the sample size of this one file, that `for...of...push` is being used when there's a need to filter, and `map` is being used when there isn't?

oliverlloyd: Is this maybe a bit? 😉 
Screenshot 2022-05-27 at 13 28 32

dotcom-rendering/src/web/layouts/lib/enumEntries.ts
```diff
@@ -0,0 +1,39 @@
...
+ * 	["Sport", 2],
+ * 	["Culture", 3],
+ * 	["Lifestyle",4]
+ * ];
+ *
+ * @see https://www.typescriptlang.org/play?#code/KYOwrgtgBAggTgFwJYGMA2wAKS1oIZxQDeAsAFBSVQBywA7gM5QC8UADADTlVQDyADkhBIA9iBZQAjFwpUAyvxGIJAJhk8AwmDQIwcYBIDM6qgBkkAM2AMEATwwSALDIC+5cijEMRGAHRoRAHMACl4AIwArYBQEX1AEOCRrYPhkdCwcfDgASmyAbnIAegAqcgBtbioygCI2ao4oatpGaoBdE0oayXrGgSFREDaOqBqVHuqFJQQhys7qw3GtHT1gGdk5x3HzKxt7VfbZkab6Bh62A-WjvuExHskLnhrJxB6VB6rqpd19HsN3ue21jsGHqjla5FaUHIxUKQA
+ */
+export const enumEntries = <T extends Record<string | number, string | number>>(
+	record: T,
+): Array<[keyof T, number]> => Object.entries(record).filter(isEntry);
```
oliverlloyd, line 39: I agree that it is a shame that looping over a Typescript enum gives strange results but I'm not sure it's a good idea to add more complexity to mitigate this. In all honesty, I don't really understand this code. This is a failing on my part because I don't know enough about Typescript, but I don't think I'm alone in lacking this knowledge.

The verbose if statements inside the original for loops are indeed messy and far from pretty but they are the simplest way to solve this problem and don't require people to learn new abstractions or understand special Typescript concepts.

Maybe we improve this code by adding more comments explaining why there is a strange if statement checking for `isNan`?

mxdvl: Yes, an alternative would be to simplify the types altogether to be string unions in `@guardian/libs`, and stop using the extremely confusing no-quite-enums of TypeScript and stick with idiomatic JavaScript.

```ts
// the options array exported if you want to loop over the options
const options = ["a", "b", "c"] as const;

// the type which as no runtime overhead at all
type Option = typeof options[number]; // "a" | "b" | "c"

// the predicate for filtering or other runtime checks
const isOption = (option: string): option is Option => options.include(option)


export { options, isOption }
export type { Option }
```

mxdvl: [Captured in this gist](https://gist.github.com/mxdvl/2e5fa0562b35a4fc9d18dcd5ac791b0f)

OllysCoding: To be honest as a testing util I don't think abstracting some of this complexity away is the worst thing in the world!

Most of the time when you're writing on and working on tests, you don't want to get bogged down in how to deal with typescripts awkward enum implementations, so if you can get a tidy and easy to use abstraction which is verbose and clear in what it outputs then I don't think its a big deal

PR #5042, opened 2022-05-27
chore(chromatic): ignore Obituaries

mxdvl:
## What does this change?

Ignore chromatic diffs for the Obituaries element in the `SubNav`.

## Why?

We are getting too many false positive on changes. Sometimes the element is on the same line, sometimes it’s not, and we can’t get it to be consistent accross runs.


oliverlloyd (approved):
This is the most beautifully done, well implemented, ignore of a test that I've ever seen 😻

mxdvl:
> This is the most beautifully done, well implemented, ignore of a test that I've ever seen 😻

Let’s see if it actually works…

PR #5085, opened 2022-05-27
Refactor `MainMedia`

MarSavar:
## What does this change?
Refactors `mainMedia` to follow the new AR component structure. This was achieved by:

- Renaming `mainMedia.tsx` into `mainMedia.ts`: it now only contains types and no JSX.
- Creating a folder called `MainMedia` with an `index.tsx` file that decides whether to render an image or a video as the main media.
- Renaming `HeaderImage` and `HeaderVideo` into `MainMediaImage` and `MainMediaVideo` and moving them into the `MainMedia` folder as subcomponents.
- Splitting `MainMediaImage` into subcomponents based on display and design.

## Why?
So that it's easier to make changes that apply to specific formats only.

## Screenshots
N/A - there are no visual changes, this is just a structural refactor.

### DIFF
--- a/apps-rendering/src/components/MainMedia/MainMediaImage/MainMediaImage.defaults.tsx
+++ b/apps-rendering/src/components/MainMedia/MainMediaImage/MainMediaImage.defaults.tsx
@@ -56,10 +56,7 @@ const defaultSizes: Sizes = {
 	default: '100vw',
 };
 
-export const imgStyles = (
-	width: number,
-	height: number,
-): SerializedStyles => css`
+const imgStyles = (width: number, height: number): SerializedStyles => css`
 	display: block;
 	width: 100%;
 	height: calc(100vw * ${height / width});
@@ -72,7 +69,9 @@ export const imgStyles = (
 
 interface DefaultProps {
 	image: Image;
-	className?: SerializedStyles;
+	className?: string;
+	css?: SerializedStyles;
+	sizes: Sizes;
 	format: ArticleFormat;
 }
 
@@ -80,11 +79,12 @@ const DefaultMainMediaImage: FC<DefaultProps> = ({
 	className,
 	image,
 	format,
+	sizes,
 }) => (
-	<figure css={[defaultStyles, className]} aria-labelledby={captionId}>
+	<figure css={className} aria-labelledby={captionId}>
 		<Img
 			image={image}
-			sizes={defaultSizes}
+			sizes={sizes}
 			className={some(imgStyles(image.width, image.height))}
 			format={format}
 			supportsDarkMode
@@ -98,4 +98,7 @@ const DefaultMainMediaImage: FC<DefaultProps> = ({
 	</figure>
 );
 
+// ----- Exports ----- //
+
 export default DefaultMainMediaImage;
+export { defaultSizes, defaultStyles, imgStyles };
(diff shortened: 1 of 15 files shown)


apps-rendering/src/mainMedia.ts
```diff
@@ -0,0 +1,15 @@
+// ----- Imports ----- //
```
iainjchambers-guardian, line 1: Should we have a `/types` direcotry in src for files only containing types? I think we've got a few files like this that can be tucked away neatly in there

apps-rendering/src/components/MainMedia/MainMediaImage/MainMediaImage.defaults.tsx
```diff
@@ -0,0 +1,101 @@
...
+	}
+`;
+
+interface DefaultProps {
+	image: Image;
+	className?: SerializedStyles;
+	format: ArticleFormat;
+}
+
+const DefaultMainMediaImage: FC<DefaultProps> = ({
```
iainjchambers-guardian, line 79: You've created this, but I don't see that you've used it anywhere even though the markup returned by your sub components is near enough identical at first glance. Might be something we should talk about

apps-rendering/src/mainMedia.ts
JamieB-gu: I think I'd prefer not to consider types as a separate "thing" to the rest of the code 🤔. With TypeScript we can treat types as no different to any other feature of the language. In my head I can see creating a directory for files that just contain types as analogous to creating a directory called "functions" where we keep files that only have functions in them, or one called "variables" where we keep files that only have variables in them?

I'm aware that what you're suggesting is maybe more geared toward tidiness than treating types as inherently different, but I worry it may look to future developers that "this is the place where we keep our types, and over here is the place where we keep the rest of our code"?

A pattern that we've tried to follow in AR is to organise modules around a type that models a particular feature and the functions that operate on that type. For example, `liveBlock.ts`, `football.ts`, `embed.ts` and so on. I think this may be a common approach found in other typed languages, and could be considered analogous to organising files around classes in Java/Scala. In some cases a type might not have any functions living alongside it at the moment, but I think that's ok?

Let me know what you think.

JamieB-gu (commented):
This looks great!

Some comments covering a couple of themes, but nothing major:

- There are a few places where you've passed a `SerializedStyles` object as a single parameter to the `css` function. I _think_ in this case that function will act like identity and just return you the same thing back so you don't need the call, but if I'm wrong let me know!
- There are a few places where I've suggest renaming `fooStyles` to `styles` when you're already in a file named `fooComponent.tsx`. My thinking is that this reduces the amount of redundant information, so there's less to comprehend for someone reading through the code. However, this is a code style/subjective thing so let me know if you disagree.

apps-rendering/src/components/MainMedia/MainMediaImage/MainMediaImage.defaults.tsx
```diff
@@ -0,0 +1,110 @@
...
+	image,
+	imgCss,
+	format,
+	sizes,
+}) => (
+	<figure css={className} aria-labelledby={captionId}>
+		<Img
+			image={image}
+			sizes={sizes}
+			className={imgCss}
```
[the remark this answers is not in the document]
MarSavar: > This looks great!
> 
> Some comments covering a couple of themes, but nothing major:
> 
> * There are a few places where you've passed a `SerializedStyles` object as a single parameter to the `css` function. I _think_ in this case that function will act like identity and just return you the same thing back so you don't need the call, but if I'm wrong let me know!
> * There are a few places where I've suggest renaming `fooStyles` to `styles` when you're already in a file named `fooComponent.tsx`. My thinking is that this reduces the amount of redundant information, so there's less to comprehend for someone reading through the code. However, this is a code style/subjective thing so let me know if you disagree.

Thanks for the comments! I've standardised the variable names so they're called `styles` and `sizes` - much cleaner.
Re: the `css` function, you're absolutely right:  if I'm passing `SerializedStyles` it doesn't need to be wrapped in `css()` at all. I think it's just a leftover from when I was passing more than one style (in which case you would normally use an array of SerializedStyles or wrap them in the `css()` function) and then reduced it to one single style, but forgot to remove the `css()` wrapper. Thanks for spotting!

JamieB-gu (approved):

PR #5087, opened 2022-05-30
Replace article link in Cypress test

ioannakok:
## What does this change?
Replaces an article link in a Cypress test

## Why?
The previous article was taken down, so we're temporarily using a new article to unblock the build. 2nd PR we will follow to add a more permanent solution as this article might also be taken down in the future.

## Screenshots

Before
image

After

image


jamesgorrie (approved):
Nice clear description on the PR!

PR #5088, opened 2022-05-30
Chromatic: Ignore Obituaries (again)

mxdvl:
## What does this change?

Ignore the whole `ul` instead of only Obituaries.

## Why?

We get too many false positive. #5042 did not succeed.

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -1429,6 +1429,15 @@ declare namespace JSX {
 
 	interface IntrinsicAttributes {
 		/**
+		 * ## Chromatic
+		 * Set this attribute to ignore elements Chromatic should ignore.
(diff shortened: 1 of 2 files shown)


AshCorr:
Is there a way to get eslint to warn where we're using data-chromatic? Would be nice if we had a way to keep track of everything that we're ignoring...

mxdvl:
> Is there a way to get eslint to give us warnings when we're using `data-chromatic=ignore`? Would be nice if we had a way to keep track of everything that we're ignoring...

I’m not sure how much value we’d get, when [a code search](https://cs.github.com/?scopeName=All+repos&scope=&q=repo%3Aguardian%2Fdotcom-rendering+%22data-chromatic%3D%22) shows the only 4 current instances readily.

mxdvl:
UI tests pending… seems like this does not achieve the intended effect :(

PR #5090, opened 2022-05-30
Render email signup embed caption with dangerouslySetInnerHTML

georgeblahblah:
## What does this change?

* Renders captions for email embeds using `dangerouslySetInnerHtml` to match [other caption-rendering implementations](https://github.com/guardian/dotcom-rendering/blob/a401bd0fbb839916af4fade22e8d1a3b2974b373/dotcom-rendering/src/web/components/Caption.tsx#L269)

## Why?

* Captions can contain HTML

## Screenshots

| Before      | After      |
|-------------|------------|
| Screenshot 2022-05-30 at 10 48 36 | Screenshot 2022-05-30 at 10 46 32 |

### DIFF
--- a/dotcom-rendering/src/web/components/EmbedBlockComponent.importable.tsx
+++ b/dotcom-rendering/src/web/components/EmbedBlockComponent.importable.tsx
@@ -50,7 +50,12 @@ export const EmbedBlockComponent = ({
 		>
 			<div data-cy="embed-block" css={embedContainerStyles(isEmailEmbed)}>
 				{isEmailEmbed && caption && (
-					<div css={emailCaptionStyle}>{caption}</div>
+					<div
+						css={emailCaptionStyle}
+						dangerouslySetInnerHTML={{
+							__html: caption,
+						}}
+					/>
 				)}
 				<div dangerouslySetInnerHTML={{ __html: unescapeData(html) }} />
 			</div>


oliverlloyd (approved):

mxdvl (commented):
Is it worth sanitising the HTML at all? We’ve got a handful of packages for this already.

georgeblahblah:
> Is it worth sanitising the HTML at all? We’ve got a handful of packages for this already.

@mxdvl good idea! What do you think about using [`dotcom-rendering/src/model/sanitise.ts#sanitiseHTML`](https://github.com/guardian/dotcom-rendering/blob/a401bd0fbb839916af4fade22e8d1a3b2974b373/dotcom-rendering/src/model/sanitise.ts#L14) for that?

oliverlloyd:
> Is it worth sanitising the HTML at all?

I think for content that we send ourselves (so to speak) we don't sanitise and trust it to be non malicious?

mxdvl:
I think it would be worth having a review. Some stuff is sanitised and some isn’t, and I couldn’t find a clear answer as to how we made that decision.

oliverlloyd:
> I think it would be worth having a review. Some stuff is sanitised and some isn’t, and I couldn’t find a clear answer as to how we made that decision.

Yeah, agree a review would be good. I've had conversations where both approach (sanitise / trust) was advocated for. I think from a security point of view we're safe so long as the content is only ever being sent from Composer (which has solid authorisation in place) but maybe there are other sources and these might be less secure.

There's also the question of if sanitisation creates any issues and if not then maybe it's just better to always do it? Maybe we can call it `SafelySetHtml` 😸

PR #5091, opened 2022-05-30
`FetchCommentCounts`

### LINKED ISSUE #4943: Show comment count on front cards
The `Card` component offers the ability to display the comment count for an article but this is not being passed through from the data model to the props

oliverlloyd:
## What does this change?
This PR changes the way we add the count of comments onto a `Card`, this closes #4943 . Previously, when rendering onwards content, we made the api call to get the comments as part of the rendering flow for the onwards container. This was a blocking call. But now that we also want to decorate cards on fronts with the count of comment this strategy is not optimal.

We now set a marker at server render time showing where we would like to see a comment count inserted. The client then runs an island which looks for these markers and deals with fetching the count from an api endpoint and creating and inserting the html to display it.

### Card props
Instead of passing in the count of comments into a `Card` - which requires us to know it ahead of time - we now pass in the `discussionId` and wait for some clientside javascript to find the marker and update it
| Before      | After      |
|-------------|------------|
| `commentCount?: number; `| `discussionId?: string;` |

### Liveness
An advantage to this approach is that, because we're using [SWR](https://swr.vercel.app/) to make the fetch call, we can upgrade the call to be made on a [smart interval](https://swr.vercel.app/docs/revalidation) with little effort

### Fewer stories?
Now that we are not setting the count imperatively, we cannot support as many stories. If we wanted to capture this presentational variation we would need to break up the `Card` component to make it more testable but this would add complexity to that code which is probably not justified here, esp. as we already have [CardCommentCount.stories.tsx](https://github.com/guardian/dotcom-rendering/blob/3ac24426c459fb8456c81b5148da6c883872189e/dotcom-rendering/src/web/components/CardCommentCount.stories.tsx)

### What about how Frontend does this?
This aligns us to the approach [that Frontend uses](https://github.com/guardian/frontend/blob/e7cd89a41fd1b8cc9d90b3ba29c9f38bb9f93bbb/static/src/javascripts/projects/common/modules/discussion/comment-count.js). One aspect of the Frontend code we are not supporting is the decision not to show a comment count when the discussion is closed and there were zero comments. We don't have the state of the discussion in context to be able to do this. If we were to edit the api response to [the comment count call](https://api.nextgen.guardianapps.co.uk/discussion/comment-counts.json?shortUrls=/p/ygjgj,/p/yg93t,/p/yh5am,/p/yf4x5,/p/yf7jj,/p/yh33k,/p/ygh8b,/p/ygtzt,/p/yf7jj,/p/yf4x5,/p/yg93t,/p/ygjgj,/p/yhxfn,/p/yh33k,/p/ygjxk,/p/yf333,/p/ygzzm,/p/ygjat) to include the discussion status then we could factor this information in when deciding if to render or not


dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -272,12 +270,28 @@ export const Card = ({
...
-										format={format}
-										long={longCount}
-										short={shortCount}
+								discussionId ? (
+									<Link
+										// This a tag is initially rendered empty. It gets populated later
+										// after a fetch call is made to get all the counts for each Card
+										// on the page with a discussion (see FetchCommentCounts.tsx)
+										data-name="comment-count-marker"
+										data-discussionid={discussionId}
```
mxdvl, line 279: If you use `data-discussion-id`, it will [be available as `discussionId` on the `dataset`, with proper `camelCase`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion).

```suggestion
										data-discussion-id={discussionId}
```

dotcom-rendering/src/web/components/FetchCommentCounts.importable.tsx
```diff
@@ -0,0 +1,173 @@
...
+ * Reads the dom looking for markers that contain the information needed to render the
+ * count
+ */
+function extractMarkers() {
+	const markers: { discussionId: string; format: ArticleFormat }[] = [];
+	document
+		.querySelectorAll('[data-name="comment-count-marker"]')
+		.forEach((element: Element) => {
+			if (element instanceof HTMLElement) {
+				const { discussionid: discussionId, format } = element.dataset;
```
mxdvl, line 40: Can be simplified if you set the attribute to `data-discussion-id`.

dotcom-rendering/src/web/components/FetchCommentCounts.importable.tsx
```diff
@@ -0,0 +1,173 @@
...
+		const format = markers.find(
+			(marker) => marker.discussionId === count.id,
+		)?.format as ArticleFormat; // We cast because we're sure we will find it
+		return {
+			id: count.id,
+			long,
+			short,
+			format,
+		};
+	});
```
mxdvl, line 84: If we’re not using the count object itself, we can de-structure here:

```suggestion
	return counts.map(({ count, id }) => {
		const { long, short } = formatCount(count);
		// We don't get format in the api response so look it up in the array
		// of DOM markers
		const format = markers.find(
			(marker) => marker.discussionId === id,
		)?.format as ArticleFormat; // We cast because we're sure we will find it
		return {
			id,
			long,
			short,
			format,
		};
	});
```

dotcom-rendering/src/web/components/FetchCommentCounts.importable.tsx
```diff
@@ -0,0 +1,173 @@
...
+							id: renderedCount.id,
+							html: renderedCount.html,
+							short: renderedCount.short,
+						});
+					});
+				} catch (e) {
+					// Do nothing
+				}
+			}
+			setUrl(buildUrl());
```
mxdvl, line 168: Why is this set again?

dotcom-rendering/src/web/components/FetchCommentCounts.importable.tsx
```diff
@@ -0,0 +1,173 @@
...
+			}
+		});
+	return markers;
+}
+
+/**
+ * @description
+ * Creates the url used to fetch the count data
+ */
+function buildUrl() {
```
mxdvl, line 52: I’d rather this function take in an array of markers and return a `string`, making it purely functional and testable.

I don’t see why we would need to collect the markers inside the body function.

mxdvl (commented):
This is a lovely abstraction.

Obviously, `FetchCommentCounts` cannot be a pure function as it’s modifying elements in the DOM. However, I think that any function that returns a value–as opposed to `void`–would benefit from being pure. I think in this case it’s only a matter of extracting the markers at the top level of `FetchCommentCounts`, and change the signature of:

- `enhanceCounts` 
- `buildUrl`

dotcom-rendering/src/web/components/Card/Card.tsx
oliverlloyd: Thanks, great suggestion - I'll do this now

dotcom-rendering/src/web/components/FetchCommentCounts.importable.tsx
oliverlloyd: Much nicer!

dotcom-rendering/src/web/components/FetchCommentCounts.importable.tsx
oliverlloyd: Good question. This code wasn't clear but basically we have `url` in state and set it anew on each interval to ensure that we're fetching counts for what's on the page (`Cards` could be adding in dynamically - Onwards, Show More, etc.).

I've added a comment in to explain this and also refactored this code a bit to hopefully make things clearer

dotcom-rendering/src/web/components/FetchCommentCounts.importable.tsx
oliverlloyd: Done. I had originally added the check for markers inside here as I wanted it to be 'fresh' but that's the wrong approach. I've changed things now to - hopefully - be clearer. Let me know what you think

mxdvl (approved):

dotcom-rendering/src/web/components/Carousel.tsx
```diff
@@ -617,10 +622,14 @@ export const Carousel: React.FC<OnwardsType> = ({
 							format: trailFormat,
 							image,
 							kickerText,
+							shortUrl,
 						} = trail;
 						// Don't try to render cards that have no publication date. This property is technically optional
 						// but we rarely if ever expect it not to exist
 						if (!webPublicationDate) return null;
+						const discussionId =
+							shortUrl && new URL(shortUrl).pathname;
```
mxdvl, line 631: I looks like this may have broken things… See #5948

PR #5095, opened 2022-05-30
Standfirst restructure

iainjchambers-guardian:
## What does this change?
Restructures the `Standfirst` component

## Why?
Separates out the large component file in line with others like `Headline` and `Byline`

## Screenshots
N/A structural changes only


apps-rendering/src/components/Standfirst/ImmersiveStandfirst.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const ImmersiveStandfirst: React.FC<Props> = ({ item }) => (
+	<DefaultStandfirst
+		item={item}
+		css={css(
+			defaultStyles(getFormat(item)),
+			css`
+				${headline.xsmall({ fontWeight: 'light' })}
+				margin-top: ${remSpace[3]};
+			`,
```
JamieB-gu, line 19: I notice you've inlined styles like this (as opposed to storing them in a variable above) a few times. Are you finding this offers some advantages that may make us want to adopt it as a convention? Are you finding it easier to read the component at a glance like this for example?

apps-rendering/src/components/Standfirst/Standfirst.defaults.tsx
```diff
@@ -0,0 +1,104 @@
...
+import type { ArticleFormat } from '@guardian/libs';
+import { ArticleDesign, ArticleDisplay } from '@guardian/libs';
+import { remSpace } from '@guardian/source-foundations';
+import { map, withDefault } from '@guardian/types';
+import { standfirstBackgroundColour } from 'editorialStyles';
+import type { Item } from 'item';
+import { getFormat } from 'item';
+import { pipe } from 'lib';
+import type { FC, ReactElement, ReactNode } from 'react';
+import { renderStandfirstText } from 'renderer';
```
JamieB-gu, line 17: I think we might want to consider extracting this rendering code from `renderer.ts` and placing it somewhere in the standfirst component folder? Possibly as a `StandfirstText.tsx` component or something along those lines? Again, not a blocker, perhaps a follow-up?

apps-rendering/src/components/Standfirst/index.tsx
```diff
@@ -1,219 +1,77 @@
...
+					css={css(
+						defaultStyles(format),
+						css`
+							${headline.xxxsmall({ fontWeight: 'bold' })}
+							padding: 0;
+						`,
+					)}
+				/>
+			);
 	}
```
JamieB-gu, line 73: Should these be their own components rather than existing in the `index.ts` file?

JamieB-gu (commented):
Looks great! A few comments and questions but mostly not blockers; just curious on your thoughts, or suggestions for some things we may want to do in follow-ups.

apps-rendering/src/components/Standfirst/Standfirst.defaults.tsx
```diff
@@ -0,0 +1,104 @@
...
+import {
+	background,
+	border,
+	text,
+} from '@guardian/common-rendering/src/editorialPalette';
+import type { ArticleFormat } from '@guardian/libs';
+import { ArticleDesign, ArticleDisplay } from '@guardian/libs';
+import { remSpace } from '@guardian/source-foundations';
+import { map, withDefault } from '@guardian/types';
+import { standfirstBackgroundColour } from 'editorialStyles';
```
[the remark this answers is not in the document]
iainjchambers-guardian: Totally agree

apps-rendering/src/components/Standfirst/ImmersiveStandfirst.tsx
iainjchambers-guardian: At the time I was thinking that it added the advantage of not having to check where the styles are being added in the markup, given that they're only a few lines in most cases. However in hindsight, given that the files are so much smaller now this advantage is pretty minimal and the loss of consistency outweighs it so I've reverted to a more conventional pattern

apps-rendering/src/components/Standfirst/index.tsx
iainjchambers-guardian: Yes, as long as we're happy that 4 different layouts use a component that is named for one of them (ie ReviewStandfirst) seeing as the idea was to name our components after layouts not teeny style variations like `<ThinStandfirst />`

apps-rendering/src/components/Standfirst/Standfirst.defaults.tsx
```diff
@@ -0,0 +1,104 @@
...
+import type { Item } from 'item';
+import { getFormat } from 'item';
+import { pipe } from 'lib';
+import type { FC, ReactElement, ReactNode } from 'react';
+import { renderStandfirstText } from 'renderer';
+import { darkModeCss } from 'styles';
+
+const isNotBlog = (format: ArticleFormat): boolean =>
+	format.design !== ArticleDesign.LiveBlog &&
+	format.design !== ArticleDesign.DeadBlog;
```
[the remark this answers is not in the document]
iainjchambers-guardian: Yes, good point, I'll look into that

apps-rendering/src/components/Standfirst/index.tsx
```diff
@@ -1,219 +1,59 @@
...
-		case ArticleDesign.Video:
-			return media(format);
-
+			return <ReviewStandfirst item={item} />;
 		default:
-			return css(styles(format), normalHeadline);
+			return (
+				<DefaultStandfirst
+					item={item}
+					css={css(defaultStyles(format))}
```
JamieB-gu, line 52: Is the `css` function needed here?

```suggestion
					css={defaultStyles(format)}
```

apps-rendering/src/components/Standfirst/index.tsx
JamieB-gu: I think it's fine, although I also think `ThinStandfirst` is reasonable - see the `Metadata` component for examples similar to this. There are always going to be components that look the same across multiple `Designs`/`Displays`, because many of these variations only change some features of the page (e.g. the headline) but not others 🤷.

apps-rendering/src/components/Standfirst/DeadBlogStandfirst.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { css } from '@emotion/react';
+import { border, text } from '@guardian/common-rendering/src/editorialPalette';
+import type { ArticleFormat } from '@guardian/libs';
+import { headline, neutral, remSpace } from '@guardian/source-foundations';
+import type { Item } from 'item';
+import { getFormat } from 'item';
+import { darkModeCss } from 'styles';
+import DefaultStandfirst, { defaultStyles } from './Standfirst.defaults';
+
+const deadblogStyles = (format: ArticleFormat): SerializedStyles => {
```
JamieB-gu, line 11: What are your thoughts on this review comment from @MarSavar 's PR? https://github.com/guardian/dotcom-rendering/pull/5085#pullrequestreview-998077705

> There are a few places where I've suggest renaming `fooStyles` to `styles` when you're already in a file named `fooComponent.tsx`. My thinking is that this reduces the amount of redundant information, so there's less to comprehend for someone reading through the code. However, this is a code style/subjective thing so let me know if you disagree.

I think the same question applies here regarding `deadblogStyles` -> `styles`, `immersiveLabsStyles` -> `styles` etc. Same caveat too though, this is a code style/subjective thing so let me know if you disagree.

apps-rendering/src/components/Standfirst/index.tsx
JamieB-gu, line 52: Is this `css` function needed here?

```suggestion
					css={defaultStyles(format)}
```

JamieB-gu (approved):
Looks great, couple final comments/questions but not blockers.

PR #5096, opened 2022-05-30
Set geolocation header in cypress test 

ioannakok:
## 1st change
* Removes unnecessary comment

## Why?
* After this PR https://github.com/guardian/dotcom-rendering/pull/4187 we can use any article we want and specify the edition to be rendered in the test.

## Screenshots

| Before      | After      |
|-------------|------------|
| US edition (no Ocado logo) | UK edition |
| image | image |

## 2nd change
* Uses two different articles in tests in `dotcom-rendering/cypress/integration/parallel-6/paid.content.spec.js` to make sure we're sending the expected information to Google Analytics: https://github.com/guardian/dotcom-rendering/pull/4098

## Why?
To test two types of articles where we could have this type of content

| Article type |      |
|-------------|-------------|
| Standard with metadata only | image |
| Immersive with Onwards | image |

### DIFF
--- a/dotcom-rendering/cypress/integration/parallel-6/paid.content.spec.js
+++ b/dotcom-rendering/cypress/integration/parallel-6/paid.content.spec.js
@@ -4,18 +4,6 @@ import { cmpIframe } from '../../lib/cmpIframe';
 import { privacySettingsIframe } from '../../lib/privacySettingsIframe';
 import { storage } from '@guardian/libs';
 
-// It is important to use this article for this test because its commercialProperties
-// coming from the CAPI object are the same for all editions. This way we are making sure
-// the Branding island the test expects will be in the DOM. If an article with different
-// commercialProperties across editions were to be used, the test would have different
-// outcome when running locally (UK edition) and when running in CI. For example,
-// it could be US edition if the CI server runs in US. The best way to deal with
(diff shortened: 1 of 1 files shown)


dotcom-rendering/cypress/integration/parallel-6/paid.content.spec.js
```diff
@@ -59,12 +52,16 @@ describe('Paid content tests', function () {
...
 			expect(requestURL).to.include('ea=sponsor%20logo');
-			expect(requestURL).to.include('el=british%20airways%20holidays');
+			expect(requestURL).to.include('el=ocado');
 		});
 	});
 
 	it('should send Google Analytics message on click of sponsor logo in onwards section', function () {
-		cy.visit(`Article?url=${paidContentPage}`);
+		const paidContentPage =
+			'https://www.theguardian.com/feasting-with-ocado/2022/apr/08/rukmini-iyers-recipe-for-slow-cooked-lamb-leg-al-pastor-with-pineapple-salsa';
```
ioannakok, line 61: I could have used this article for the metadata test as well but that has the same `commercialProperties` across all editions but that wouldn't show line `cy.visit(`Article?url=${paidContentPage}`, { headers: { 'x-gu-geolocation': 'ip.dummytext,country:UK' }, });` is working as expected.

PR #5098, opened 2022-05-30
Add key events carousel to layout

abeddow91:
## What does this change?
This PR adds the new key events carousel and removes the old key events component by placing these components behind the [keyEventsCarousel switch](https://github.com/guardian/frontend/pull/24908). 

The PR includes the following changes:
- show carousel nav buttons on desktop only when there are 7+ key events
- Add key events title to carousel until desktop
- hide / change position of the carousel depending on the total number of key events available. As the key events only refreshes on page load, this shouldn't cause content layout shift but this design may need rethinking if we decide to auto refresh key events.
- Move the key events toggle below the carousel on mobile only
- Add a min-width to the Live Grid `grid-template-columns` to [prevent grid blowout](https://css-tricks.com/preventing-a-grid-blowout/). (more details below)

## Why?
As part of the standfirst redesign, key events are being moved into a carousel component. For desktop and above, this component sits below the standfirst. For anything below desktop, the carousel sits within the main section of the live grid, below the pinned post and above the first post in the liveblog. 

#### What's grid blowout and why does this fix work?
Grid blowout is where adding a new element within a css grid causing the layout to spread out of the specified columns and rows. 

We set our grid columns to `1fr` which is equal to one fraction of the left space. By default the minimum width of a grid column is auto (or indefinite space). The assumption here is that the grid would adapt to the parent's width and the 1fr would be calculated in relation to this, making the parent's width the max width. However, this isn't the case. What's actually happening is that a grid-item looks up the minimum space that it’s content would have (in our case the total width of all key event cards) and then expand to 1fr. Other elements within the grid will potentially then stretch to fit 1fr, causing grid blowout.

To fix this, instead of setting `grid-template-columns: 1fr`, we set `grid-template-columns: minmax(0, 1fr)`. By setting a definite minimum width of 0, we prevent the grid from looking to the contents for a minimum width. 

## Screenshots

| Mobile | Tablet |
|-------------|-------------|
| Screenshot 2022-05-31 at 09 23 03 | Screenshot 2022-05-31 at 09 23 19 |

| Desktop (<=4 events) | Desktop (5 - 6 events) | Desktop  (>= 7 events) |
|-------------|-------------|-------------|
| Screenshot 2022-05-31 at 09 45 35 | Screenshot 2022-05-31 at 09 20 53 | Screenshot 2022-05-31 at 09 23 35 |


dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -585,6 +586,21 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 					</StandFirstGrid>
 				</ElementContainer>
 
+				<ElementContainer
```
marjisound, line 589: I may be wrong, was just curious if we also need to use the switch in there too?

abeddow91: You're right! thanks Marjan!

dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
```diff
@@ -70,6 +77,25 @@ export const LiveBlogRenderer = ({
 					</PinnedPost>
 				</>
 			)}
+			{switches.keyEventsCarousel && keyEvents?.length ? (
```
alinaboghiu, line 80: At the risk of distracting, would it be possible to 
- move the Pinned Post out of the LiveBlogRenderer and into the LiveLayout
- move LiveBlogRenderer out of ArticleBody and into LiveLayout and remove ArticleBody from LiveLayout

Feels like it would be simpler to have everything in one place. Now that we've moved where the key events are displayed using article body doesn't make as much sense as it would now contain several non body things like pinned post and key events. But if this is too big a task it's fine, just a suggestion. 

If we do this then the KeyEventsCarousel would also only live in LiveLayout and not inside LiveBlogRenderer.

alinaboghiu:
Not sure this PR is anything to do with it but I just noticed when we only have one key event in the carousel the dot is red instead of grey, e.g. http://localhost:3030/Article?url=https://www.theguardian.com/us-news/live/2022/jun/08/supreme-court-texas-school-shooting-survivor-testify-congress-biden-politics-latest-updates

alinaboghiu (approved):

PR #5111, opened 2022-06-01
Fix email embeds when rendering alongside a rich link

oliverlloyd:
## What does this change?
This ensures the email embed elements stay together around any rich link that is placed inline in the article content.

Example: https://www.theguardian.com/politics/2022/may/30/tory-chris-philp-rules-out-new-inquiry-illicit-gathering-boris-johnson-flat

| Before      | After      |
|-------------|------------|
|  |  |


ioannakok (approved):
Looks good!

Issue #4737, opened 2022-04-26
Standardise the spinning animation on the '+ More Opinion' button to align with other loading buttons styles in Source


HarryFischer:
Design to provide details

HarryFischer:
![image](https://user-images.githubusercontent.com/20658471/172357834-4c85f458-94da-47b5-8455-09d344790974.png)

bryophyta:
<img width="299" alt="image" src="https://user-images.githubusercontent.com/37048459/199234673-a93a8c82-a724-4923-a75b-149b1213f535.png">
<img width="239" alt="image" src="https://user-images.githubusercontent.com/37048459/199234704-4d534071-4506-44c9-8486-f51a4decacd5.png">

@HarryFischer do these look about right to you? I couldn't find the exact colour that Frontend's using in the [current Source palette](https://guardian.github.io/source/?path=/docs/packages-source-foundations-palette--palette#colour-tokens),  so opted for `neutral[7]`.

Also, I wasn't sure if you were  intending there to be a smooth transition from one state to the other? At the moment we're using the `loading` state from the Source Button component to indicate the in-between state, which adds a spinner (hence no smooth transition). But if it's important for the cross icon to actually rotate into position then we should be able to do that (though perhaps in a separate PR from the main showmore one?)

PR #5116, opened 2022-06-03
Add card container endpoint for 'show more' functionality

### LINKED ISSUE #4604: `Show More`
This button, shown at the bottom of containers, will make an api call to fetch in more content that can be appended to the bottom of the container.

The layout for the expanded content is currently the same, simple, one.

### Questions
How is the api url created? Is this passed down or constructed client side?
Do we need a `?dcr` version of these endpoints to allow json to be passed back or should we take the same approach that we did for key events and pass back html?

bryophyta:
## What does this change?

It adds a new endpoint, `Cards/`, and a new interactive importable button `ShowMore.importable`. The cards endpoint can render a list of cards of arbitrary length, using the `ExtraCardsContainer` container. The `show more` button makes a request to Frontend/Facia, which makes a request to the cards endpoint. If all goes well, the DCR cards endpoint will return markup which Frontend will pass to the client, to be inserted into the DOM for the relevant parent container.

## Why?

This largely recreates the basic functionality of 'show more' buttons on Frontend, so it closes #4604. There are a couple of features which this implementation does not replicate:

1. It does not handle ads, because DCR Fronts as a whole does not handle ads yet.
2. Frontend deduplicates cards in the show more container both on the server side and on the client side. While we looked into implementing that here, it was decided that deduplication is not a high priority for this container at the moment, so it is being left for a future PR. (For what it's worth, it may be simplest to d _all_ of the deduplication on the client side, given that [we can't know on the server side](https://github.com/guardian/dotcom-rendering/issues/4604#issuecomment-1140818041) which cards are actually displayed on the client at the time when the extra cards are requested.)

There are a couple of further points about design which have been raised by @HarryFischer ([button styling](https://github.com/guardian/dotcom-rendering/issues/4737) and also whether to have 3 or 4 columns of cards although my recollection is that it was decided that it didn't matter too much for this container?). We have also discussed adding Sentry integration. My feeling is that at this point it makes sense to incorporate these as separate PRs in the near future, because this PR has grown quite large already, and the main work has been to integrate the rendering functionality with Frontend ([Frontend PR](https://github.com/guardian/frontend/pull/25086)).

## Screenshots

| Before (Frontend)     | After (DCR)      |
|-------------|------------|
| Screenshot 2022-07-11 at 16 58 27 | image |
|image|Screenshot 2022-07-11 at 16 57 50|

## Tested
- [x] Locally
- [x]  On CODE

## Dependencies
Full deployment requires merging [this PR](https://github.com/guardian/frontend/pull/25086) on Frontend.


dotcom-rendering/src/web/components/ShowMore.importable.tsx
```diff
@@ -0,0 +1,88 @@
...
+	collectionId,
+	displayName,
+	editionId,
+}: {
+	collectionId: string;
+	displayName: string;
+	editionId: Edition;
+}) => {
+	const url = `https://api.nextgen.guardianapps.co.uk/${
+		editionId === 'INT' ? 'international' : editionId.toLowerCase()
```
ioannakok, line 54: I think instead of editionId this part of the url is coming from `pressedPage.id` so we should pass this as a prop. For example, when we are here https://www.theguardian.com/uk/environment and click show more in Spotlight section:

Screenshot 2022-06-22 at 17 06 10

bryophyta: this issue should be fixed as of now 👍 

https://github.com/guardian/dotcom-rendering/blob/5e8ba23b89667911a8834f7773346532307c2146/dotcom-rendering/src/web/layouts/FrontLayout.tsx#L159-L162

dotcom-rendering/src/web/server/index.ts
```diff
@@ -184,6 +187,36 @@ export const renderKeyEvents = (
...
+		const dcrTrails = enhanceCards(cards, dcrContainerPalette, startIndex);
+
+		const html = cardsToHtml({
+			cards: dcrTrails,
+			containerPalette: dcrContainerPalette,
+		});
+
+		res.status(200).send({ html });
+	} catch (e) {
+		// todo: need to replicate error-handling behaviour of existing component
```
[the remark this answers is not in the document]
bryophyta: I can't really see anything useful that Frontend could pass to us here if something breaks down either within Frontend or within DCR while it's trying to fetch the new cards. Given this, I'm tempted to hard-code a html string in the ShowMore.Importable component which contains a generic error message to put into the DOM if it receives anything other than a 200 response which has a `html` key in the response body. Does anyone have thoughts on this?

dotcom-rendering/src/web/components/ShowMore.importable.tsx
```diff
@@ -0,0 +1,94 @@
...
+	displayName,
+}: {
+	pageId: string;
+	collectionId: string;
+	displayName: string;
+}) => {
+	const [showMore, setShow] = useState(false);
+	// We only pass an actual URL to SWR when 'showMore' is true.
+	// Toggling 'showMore' will trigger a re-render
+	//   see: https://swr.vercel.app/docs/conditional-fetching#conditional
```
mxdvl, line 61: If you want to turn this into a JSDoc:

```suggestion
	// @see https://swr.vercel.app/docs/conditional-fetching#conditional
```

dotcom-rendering/src/web/server/index.ts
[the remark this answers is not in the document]
mxdvl: This is an interesting question!

I think a generic error message would be acceptable. It would be interesting to share with designers and UX to get their thoughts.

dotcom-rendering/src/web/server/index.ts
```diff
@@ -184,6 +187,36 @@ export const renderKeyEvents = (
 	}
 };
 
+/**
+ * Sends markup to Frontend, which forwards it to the client.
+ * Frontend expects that it will receive the markup string as a value of the
+ * key 'html' in a json object, i.e. { "html": "<some markup>" }
+ */
+export const renderCards = (
+	{ body }: { body: CardsRequest },
```
mxdvl, line 196: I tend to think it's best to not make an assumption about the type `body`, and parse it—we use AJV for `enhanceCAPIType` and `enhanceFronts`. We're not very consistent, though.

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -165,11 +166,22 @@ export const FrontLayout = ({ front, NAV }: Props) => {
...
 							/>
+							{collection.hasMore && (
+								<Island deferUntil={'visible'}>
+									<ShowMore
+										pageId={front.pressedPage.id}
+										collectionId={collection.id}
+										displayName={collection.displayName}
+									/>
+								</Island>
+							)}
```
oliverlloyd, line 184: Does this mean that each container on the page will download, evaluate and execute the show more island?

dotcom-rendering/cypress/e2e/parallel-4/show-more-test.cy.js
```diff
@@ -0,0 +1,81 @@
+/* eslint-disable no-undef */
+/* eslint-disable func-names */
+import { disableCMP } from '../../lib/disableCMP';
+import { setLocalBaseUrl } from '../../lib/setLocalBaseUrl.js';
+
+import showMoreRequestBody from '../../fixtures/manual/science-show-more-req.json';
+import scienceFrontJson from '../../fixtures/manual/science-front.json';
```
oliverlloyd, line 7: In DCR we typically don't use default exports; we favour named exports. It helps because it's then easier to trace the name of the file through the code (it doesn't change). We also keep the name of the file the same as the named export (for similar reasons!). That said it is true that there are many examples in the code of this pattern not being used!

dotcom-rendering/cypress/e2e/parallel-4/show-more-test.cy.js
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * This tests both the '/Cards' endpoint, and the client-side 'Show More' i
+ * interaction that fetches and displays extra cards for a container.
+ * It mocks the responses from Frontend in order to keep the contents stable.
+ * This is partly to allow for testing properties like the URLs of the rendered
+ * cards, but also partly with a view to adding visual regression testing in the
+ * near future.
```
oliverlloyd, line 15: This is a change to how we run Cypress. We currently don't mock data in cypress, we read it from production. There are well known pros and cons to this but I think it's important that we either stay consistent or, if we choose to diverge, we do so in a way that makes it clear which tests are mocked and which are not. Maybe this could be pulled out into a separate PR? 🙏

dotcom-rendering/cypress/e2e/parallel-4/show-more-test.cy.js
oliverlloyd: This is importing from the fixtures directory used by Storybook, Cypress has it's own fixtures folder

dotcom-rendering/src/web/server/index.ts
```diff
@@ -184,6 +187,36 @@ export const renderKeyEvents = (
...
+	res: express.Response,
+): void => {
+	try {
+		const { cards, startIndex, config } = body;
```
oliverlloyd, line 200: Are we only using `startIndex` for tracking? This feels like a pretty big api change just for this very small edge case. Maybe we can replace the numeric index value here with the text show-more'.This would simplify this code, prevent a need for a major api edit and arguably be more useful for anyone analysing the data

dotcom-rendering/src/web/components/FixedSmallSlowIV.tsx
```diff
@@ -4,63 +4,72 @@ import { LI } from './Card/components/LI';
 import { UL } from './Card/components/UL';
 
 type Props = {
+	collectionId: string;
 	trails: TrailType[];
 	containerPalette?: DCRContainerPalette;
 	showAge?: boolean;
+	hasMore: boolean;
```
oliverlloyd, line 11: What do you think about this property being optional?

dotcom-rendering/src/web/components/FixedSmallSlowIV.tsx
```diff
@@ -4,63 +4,72 @@ import { LI } from './Card/components/LI';
 import { UL } from './Card/components/UL';
 
 type Props = {
+	collectionId: string;
```
oliverlloyd, line 7: What about `collectionId` being optional?

oliverlloyd: It seems as though `showMore` and `collectionId` are coupled here. Is there a way to combine them into one property? Maybe we could use `showMoreId` and then if that exists we add the placeholder, and we can also use that when constructing the url? That way we no longer need the `showMore` boolean at all 🎉 

It's true that the value of the `showMoreId` would be the collection id, but I think it is valid for collection id to be _used_ as the value of another things id. This feels semantic to me

oliverlloyd (commented):
Loads of great work has gone into this, really well done.

I've made some small comments and suggestions for improvements but I do have one larger concern with how we are setting n different islands, one per container. Is there a way to have one piece of javascript that controls multiple buttons on the same page?

dotcom-rendering/src/web/server/index.ts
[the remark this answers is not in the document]
bryophyta: thanks! I've double-checked the contract with Frontend with James, and have added a hard-coded message on the client side which will be displayed if it receives an error from Frontend. This (more-or-less) replicates the existing Frontend-rendered error-handling behaviour, but I agree that it would be good to revisit the UX and design of this component with others. I'd like to leave this for another PR to avoid more scope creep, or do you think that this is a blocker for this PR?

dotcom-rendering/src/web/server/index.ts
bryophyta: I believe there was some discussion about this before? I agree that it doesn't seem that important in itself, but quite a lot of work was going into setting this in the original Frontend endpoint, and from what I remember of previous discussions it was decided that this is something we should support..

Re: adding 'show-more' to the `data-link-name` attr, I've already done this with an optional prop on the `ExtraCardsContainer.tsx` (ln.40). Does that look like a reasonable pattern to youo?

dotcom-rendering/src/web/server/index.ts
bryophyta: this definitely seems like a good practice to me. The current pattern is (afaict) the same as used by the other endpoints in this file though, so do you think it might make sense to open this as a separate issue, or to roll it into this one?

dotcom-rendering/src/web/components/FixedSmallSlowIV.tsx
bryophyta: I can see the appeal of this, but as far as I'm aware, the `collectionId` is a more general identifier. Even though I'm not sure if/where else we use it at the moment, it seems like the sort of id that might well be used in other contexts. Using its presence as an indicate of whether there are more cards to show would make that kind of re-use into a bigger refactor. Given that combining them would also introduce a new transformation between the Frontend model and the DCR model, I'm inclined to leave it as it is?

bryophyta: I'd be happy to make both properties optional though (though it would be good if we had a way of ensuring that if 'hasMore' is set, then 'collectionId' should also be set.

dotcom-rendering/cypress/e2e/parallel-4/show-more-test.cy.js
bryophyta: > This is importing from the fixtures directory used by Storybook, Cypress has it's own fixtures folder

is it? this is the path from repo root on my computer: `dotcom-rendering/cypress/fixtures/manual/science-front.json`

bryophyta: > In DCR we typically don't use default exports; we favour named exports.

ah okay, my reason for not using a named export here is because it's importing from a `.json` file, but if I understand you correctly then it would make sense to refactor so that it's importing from a `js` file instead?

dotcom-rendering/cypress/e2e/parallel-4/show-more-test.cy.js
bryophyta: okay yep that makes sense. I added mock data here partly because I'm anticipating adding a Percy snapshot at some point down the line, but happy to discuss further how we want to approach live vs. mock here.

When you say pull into a separate PR, do you mean take the test file out of this PR and open a separate PR for it? Or use live data here and then open a new PR to discuss mocking?

dotcom-rendering/cypress/e2e/parallel-4/show-more-test.cy.js
oliverlloyd: > is it? this is the path from repo root on my computer: `dotcom-rendering/cypress/fixtures/manual/science-front.json`

Oh, you're right. Sorry, I completely misread that!

dotcom-rendering/cypress/e2e/parallel-4/show-more-test.cy.js
oliverlloyd: Using live data in our end to end test is nice because it gives us extra security. If the test works on live data then it gives us a higher level of confidence, the scope of what we're testing is larger.

I think in this case we would need a static front though, one that isn't going to change over time. This is a lot easier to achieve using articles but might be possible?

dotcom-rendering/src/web/server/index.ts
oliverlloyd: Maybe we can revisit the motivations around the earlier discussion? It could be there are strong reasons for having this prop that we should learn about here. At the very least we should document them for future engineers landing here and trying to understand our thinking.

My gut feeling is this prop might not be worth it but lets lay both sides out and see where we land!

dotcom-rendering/src/web/components/FixedSmallSlowIV.tsx
oliverlloyd: > might well be used in other contexts

This feels like we'd be 'coding for the future' which is famously impossible to do. You _think_ you know what future us will want but 99% of the time we get it very wrong!

> Using its presence as an indicate of whether there are more cards to show
I do agree that we shouldn't use `collectionId` as the indication for when to use show more. But if we call the prop `showMoreId` (or similar) then that is a lot clearer that this prop is only being used in this context.

Adding props has a cost and adding two props, which are coupled, is an even greater one.

> would also introduce a new transformation
It's true that this pattern would mean we would populate showMoreId using collectionId, but I think this is a valid thing to do. We routinely transform data in this way where the context demands it.

>  it would be good if we had a way of ensuring that if 'hasMore' is set, then 'collectionId' should also be set.
I feel that this is a good indication of why having two props is a problem 😉

dotcom-rendering/src/web/layouts/FrontLayout.tsx
oliverlloyd: As discussed offline, we should look at having a central island, one that only gets downloaded and executed once, which acts as the controller for these buttons.

We could use some plain javascript here that loops all the placeholders and sets an event listener on each to fetch and render the content

oliverlloyd: I think @AshCorr 's [PR had a similar issue](https://github.com/guardian/dotcom-rendering/pull/4905/files) as this which suggests we have a design constraint and probably don't want to ever have an island inside a container.

Your use case is more complex in that you're making a fetch but hopefully there will be useful similarities.

dotcom-rendering/cypress/e2e/parallel-4/show-more-test.cy.js
OllysCoding: > This is a change to how we run Cypress. We currently don't mock data in cypress, we read it from production.

Worth noting that while in cypress we do go to live articles, we do often spoof some API responses with cy.intercept (e.g https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/cypress/lib/mocks.js).

I think the tradeoff to consider here is that while it's relatively easy for us to pick an older 'quiet' article where the chances of it changing in the future is very low - it may be harder to do this with fronts, and relying on specific containers/cards/layouts existing may result in us having to spend a lot more time fixing broken tests.

The advantage of having live data of course being that fixtures can get out-of-date, and tests can end up testing that out-of-date data works, rather than if 'live' data works.

TL;DR: I don't mind too much, but we do need to recognise that testing on fronts could require a different approach than testing on articles!

dotcom-rendering/cypress/e2e/parallel-4/show-more-test.cy.js
```diff
@@ -0,0 +1,81 @@
...
+ * been pressed in PROD yet.
+ */
+
+describe('Show More button', function () {
+	beforeEach(function () {
+		disableCMP();
+		setLocalBaseUrl();
+	});
+
+	it('should insert markup for extra cards into container element', function () {
```
OllysCoding, line 26: I'd love some comments &  info in this test as there is a lot going on

dotcom-rendering/src/web/components/ShowMore.importable.tsx
```diff
@@ -0,0 +1,114 @@
...
+			`[data-show-more-placeholder="${collectionId}"]`,
+		);
+		/**
+		 * it is expected that the value of `html` will be a server-rendered
+		 * instance of the ExtraCardsContainer component
+		 */
+		// @ts-expect-error -- We want to catch this error if the element is missing
+		placeholder.innerHTML = html;
+	} catch (e) {
+		// TODO: pass to Sentry
```
OllysCoding, line 38: `window.guardian.modules.sentry.reportError(e, 'show-more-insert');`

dotcom-rendering/src/web/components/ShowMore.importable.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+function removeHtml(collectionId: string) {
+	try {
+		const placeholder = document.querySelector<HTMLElement>(
+			`[data-show-more-placeholder="${collectionId}"]`,
+		);
+		// @ts-expect-error -- We want to catch this error if the element is missing
+		placeholder.innerHTML = '';
+	} catch (e) {
+		// TODO: pass to sentry
```
OllysCoding, line 50: `window.guardian.modules.sentry.reportError(e, 'show-more-remove');`

dotcom-rendering/cypress/e2e/parallel-4/show-more-test.cy.js
bryophyta: > I don't mind too much, but we do need to recognise that testing on fronts could require a different approach than testing on articles

This seems right to me. Bracketing visual regression testing for now, it seems to me that in order to check the rendering logic of `ShowMore.Importable` here we'd need to write some pretty complex logic into the test, so that we could derive the number of cards that should be shown, and the urls, indices, etc that they should have.

I agree that we miss some things by mocking with static data though. Is it worth considering adding a test with live data in _addition_ to the mocked one?

dotcom-rendering/cypress/e2e/parallel-4/show-more-test.cy.js
```diff
@@ -0,0 +1,81 @@
+/* eslint-disable no-undef */
+/* eslint-disable func-names */
```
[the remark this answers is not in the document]
bryophyta: I don't know to be honest! I used another test file as a template, which had these... Do you think they might cause trouble?

dotcom-rendering/src/web/components/ShowMore.importable.tsx
```diff
@@ -0,0 +1,125 @@
...
+			`[data-show-more-placeholder="${collectionId}"]`,
+		);
+		/**
+		 * it is expected that the value of `html` will be a server-rendered
+		 * instance of the ExtraCardsContainer component
+		 */
+		// @ts-expect-error -- We want to catch this error if the element is missing
+		placeholder.innerHTML = html;
+	} catch (e) {
+		// todo: `window` is throwing a referenceError; find a fix
```
bryophyta, line 38: @OllysCoding I mentioned this in chat but just realised I didn't flag it here. I feel like we could leave this for a separate issue, but if you think otherwise (or if you know how to fix this?) then lmk?

dotcom-rendering/src/web/server/showMoreCardsToHtml.tsx
```diff
@@ -0,0 +1,27 @@
...
+	cards: DCRFrontCard[];
+	containerPalette?: DCRContainerPalette;
+}
+/**
+ * cardsToHtml is used by the /Cards endpoint to render extra fronts cards
+ *
+ * @returns string (the html)
+ */
+
+export const cardsToHtml = ({
```
oliverlloyd, line 14: If the export of the file is `cardsToHtml` the name of the file should be the same

dotcom-rendering/src/web/server/showMoreCardsToHtml.tsx
```diff
@@ -0,0 +1,27 @@
...
+	cards,
+	containerPalette,
+}: DCRShowMoreContainer): string => {
+	const html = renderToString(
+		<ExtraCardsContainer
```
oliverlloyd, line 19: If I call `/Cards` it seems like the html I get in response is exactly the  `ExtraCardsContainer`?

It can be tempting to write code that is adaptable and open to future use cases but I think in this case it might be clearer if we're more explicit here.

Rather than

Frontend calls `/Cards`
Receives `ExtraCardsContainer`
Inserts it in `ShowMore.importable`

We could have

Frontend calls `/ShowMore`
Receives `ShowMoreContainer`
Inserts it in `ShowMore.importable`

or maybe

Frontend calls `/ShowMore`
Receives `ShowMore`
Inserts it in `EnhanceShowMore.importable` (this follows existing patterns, like `EnhancePinnedPost`)

I think this would really help future developer understand dependencies and what code is related to what.

One use case that was discussed that we thought might have used /Cards was onwards. But - as so often happens - this had it's own special requirements and now we have `/Onwards`.

dotcom-rendering/src/web/components/FixedSmallSlowIV.tsx
bryophyta: I've consolidated the two props in the new PR: https://github.com/guardian/dotcom-rendering/pull/5463

dotcom-rendering/src/web/components/FixedSmallSlowIV.tsx
bryophyta: resolved by consolidation into a single optional prop here: https://github.com/guardian/dotcom-rendering/pull/5463

dotcom-rendering/src/web/server/index.ts
bryophyta: happy to discuss this, but as it involves a change to the functionality, as opposed to straightforward migration, I suggest that we discuss this as a separate issue at a later time?

dotcom-rendering/src/web/server/showMoreCardsToHtml.tsx
bryophyta: this is a fair point. I think you're right that the original motivation for keeping it more general was the connection with `Onwards`.

My only concern is that `ShowMoreContainer` never contains `ShowMore.importable`, if that makes sense? Would `MoreCardsContainer` work, do you think? Or does that have the same problem as `ExtraCardsContainer`?

bryophyta:
This PR has been superseded by #5463 #5467 and #5476 

They are largely replicate the material from this branch in slightly smaller chunks, although some changes have been made as well, in response to discussion here.

PR #5118, opened 2022-06-03
Change byline avatar positioning

iainjchambers-guardian:
## What does this change?
Changes the avatar in the byline to just have a negative margin instead of `position: absolute`
## Why?
On some screens the avatar image overlaps the standfirst because it's aligned with the bottom of the byline
## Screenshots

| Before      | After      |
|-------------|------------|
| image | image |


groakland (approved):
LGTM

Issue #5094, opened 2022-05-30
Show byline on `Card`s


oliverlloyd:
The byline should appear for `Card`s when `Design.Comment`

<img width="235" alt="Screenshot 2022-05-30 at 16 43 48" src="https://user-images.githubusercontent.com/1336821/171026541-15b55937-147f-4165-86e1-dcec45d34a95.png">

bryophyta:
closed by #5119

PR #5119, opened 2022-06-03
Add byline to DCFrontCard

bryophyta:
## What does this change?
- Add `byline` and `showByline` fields to DCRFrontCard model.
- Pass values for these fields through from FE cards to DCR cards (when present) in `enhanceCards` function.

## Why?
The initial aim was to add bylines to comment cards to match with how they're rendered by Frontend (as per #5094). 

The final implementation has a slightly more general scope, however. The simplest way to recreate the byline behaviour of Frontend on DCR Fronts was to pass the `showByline` field through from the json for a Front card through to the container renderer. 

The reason for doing this is that it seems sensible to delegate showing bylines on Fronts cards to the logic that's already built out in Frontend by default, unless we want to add any significant deviations. Overrides could be added for special cases, until a time where it makes sense to implement a separate logic in DCR.

### Frontend, unchanged
frontend

### Before (DCR)
DCR before

### After (DCR)
DCR after

### DIFF
--- a/dotcom-rendering/src/model/enhanceCards.ts
+++ b/dotcom-rendering/src/model/enhanceCards.ts
@@ -109,5 +109,11 @@ export const enhanceCards = (
 					  )
 					: undefined,
 				discussionId: faciaCard.discussion.discussionId,
+				// there are two 'byline' properties on FEFrontCard; are they
+				// always the same? If not, which should be preferred?
+				byline:
+					faciaCard.properties.maybeContent?.trail.byline ??
+					undefined,
+				showByline: faciaCard.properties.showByline,
 			};
 		});
--- a/dotcom-rendering/src/model/enhanceCards.ts
+++ b/dotcom-rendering/src/model/enhanceCards.ts
@@ -109,8 +109,8 @@ export const enhanceCards = (
 					  )
 					: undefined,
 				discussionId: faciaCard.discussion.discussionId,
-				// there are two 'byline' properties on FEFrontCard; are they
-				// always the same? If not, which should be preferred?
+				// nb. there is a distinct 'byline' property on FEFrontCard, at
+				// card.properties.byline
 				byline:
 					faciaCard.properties.maybeContent?.trail.byline ??
 					undefined,
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -865,6 +865,8 @@ type DCRFrontCard = {
 	/** @see JSX.IntrinsicAttributes["data-link-name"] */
 	dataLinkName: string;
 	discussionId?: string;
+	byline?: string;
+	showByline?: boolean;
 };
 
 type FECollectionType = {


dotcom-rendering/src/model/enhanceCards.ts
```diff
@@ -109,5 +109,11 @@ export const enhanceCards = (
 					  )
 					: undefined,
 				discussionId: faciaCard.discussion.discussionId,
+				// nb. there is a distinct 'byline' property on FEFrontCard, at
```
bryophyta, line 112: As noted here, there are two `byline` properties on an FEFrontCard. I've preferred this one, even though it's more deeply nested, because I'm not aware of any differences between the two, and because in general it seems as though we get card contents from inside `maybeContent`.

Does anyone know of any scenarios in which the two `byline` values might be different?

oliverlloyd: The second property looks to be optional whereas the first is (currently) required so the first might be cleaner? I think to really understand the differences we'd need to dig into Frontend.

oliverlloyd (approved):
Looks great!

dotcom-rendering/src/model/enhanceCards.ts
bryophyta: Where is it defined as required? I was basing this on the the type definition for `FEFrontCard` here, where it looks like they're both optional, so maybe we need to update that? https://github.com/guardian/dotcom-rendering/blob/2dd43453e2d425dc652443f1ee743315e58887ec/dotcom-rendering/index.d.ts#L768-L791

oliverlloyd: Oh interesting, they're both optional. Well, then in that case, I think you're previous reasoning sounds good - lets favour this part of the object from where we read other properties

oliverlloyd:
> I can just add a couple of stories to test the card bylines specifically, but I wondered if the false negative from Chromatic here suggests that we need a broader change in our storybook for cards/fronts? This feels like the sort of change we'd want to be able to catch if it happened by accident.

👍 💯 this

Thanks for noticing this and raising it here, it would indeed appear that we have a gap in our coverage. Do you think if we update `basicCardProps` in `Card.stories.tsx` that this will give us what we want?

oliverlloyd:
> > Do you think if we update basicCardProps in Card.stories.tsx that this will give us what we want?
> 
> Yes this seems like the important question, although I'm not quite sure of the answer! We've got some cards in that story already in there which have bylines, and that functionality was working before so hopefully shouldn't be changed by this PR.
> 
> The context I would expect the difference to show up in is when cards with bylines appear in one of the fronts containers that's populated via the `DCRFrontsCard` model, because that's where the `byline` data was being filtered out before. But I'm not sure actually whether we aim for storybook to catch these sorts of issues or not? To mimic this properly I guess we'd need to pass mock Fronts data through `enhanceCards`, and then into one of the Fronts containers, which starts becoming a bit integration-y. Do we have a heuristic for what goes into storybook?

I think you're running up against a problem we've had a while now. The architecture for DCR is broadly Raw Data > DCR > `enhance()` > renderToHtml. But our stories are mostly only capturing the renderToHtml step so we miss some coverage for our enhancers and a lot of the other server side steps.

We could write a whole load of jest tests verifying the outputs are what we expect given mocked inputs but it would be painful and could end up becoming a maintenance burden.

One option we have been discussing recently is if we could integrate visual regression with Cypress. Cypress will load an entire page in an identical fashion to a real user (instead of using an iframe like in Storybook) and then if we took a snapshot of the pixels at that point we'd be able to capture these sorts of variations. I think adding Percy [to Cypress](https://docs.percy.io/docs/cypress) would allow this, maybe we can run up a POC?

bryophyta:
> I think adding Percy [to Cypress](https://docs.percy.io/docs/cypress) would allow this, maybe we can run up a POC?

This sounds quite promising -- does anyone have prior experience with Percy? I'd be quite interested in helping out setting up a proof of concept, as I've been wanting to find a chance to look into Cypress a bit more anyway. Is this something that deserves a separate ticket?

oliverlloyd:
> Is this something that deserves a separate ticket?

Yeah, good call.

We were using Percy briefly before moving over to Chromatic. It should be pretty simple to [setup](https://docs.percy.io/docs/cypress), I think it's straight forward to install and then follow the docs. The key function that we would benefit from is `cy.percySnapshot()`, e.g.:

```
describe('Integration test with visual testing', function() {
  it('Loads the homepage', function() {
    // Load the page or perform any other interactions with the app.
    cy.visit(<URL under test>);

    // Take a snapshot for visual diffing
    cy.percySnapshot();
  });
});
```

This could potentially replace the [full page layout stories](https://www.chromatic.com/library?appId=5dfcbf3012392c0020e7140b&groupPrefix=Layouts%2FFormat+variations) we currently have.

bryophyta:
closes #5094

Issue #4752, opened 2022-04-27
Solution for storing secrets in DCR


AshCorr:
In order for us to be able to use the image optimizer or thrashers in DCR we're going to need some way of storing the secrets used for calling the aforementioned API's.

PR #5128, opened 2022-06-06
DCR signs urls ✍️ 

oliverlloyd:
## What?
This PR adds the logic for constructing and signing Fastly image sources to DCR. Previously, this step happened in Frontend and the resulting source urls were sent to DCR inside the data model.

As part of the move to bring the decision logic for image sizes into DCR the values in [this Frontend file](https://github.com/guardian/frontend/blob/oliver/immersives-for-all/common/app/layout/ContentWidths.scala) have been respected

## Why?
Firstly, because we can, or at least we can now following [this PR](https://github.com/guardian/dotcom-rendering/issues/4752) which added the ability for DCR to work with secret values, such as `IMAGE_SALT`.

But also because it's a good idea. Having the logic that decided what size of image is used spread over two codebases was confusing and led to bugs. Now the same logic all exists in one file and can be reasoned about much easier.

### `development`
If a developer is running DCR locally and does not have the `IMAGE_SALT` variable, then we show a default image. This image is designed to be useful in terms of allowing the site to be run without additional config steps (you can still run DCR without access to the VPN) but also a strong enough indication that you need an extra step if you want to see optimised images.

### `production`
If you run DCR in production without the `IMAGE_SALT` environment variable set then the server _should_ fail to start. In the very unlikely event that the code is still executed without this variable then an error should be thrown

## Do we have a fallback in place for if Fastly does not respond in time?
Yes. This is [managed in the VCL code](https://github.com/guardian/fastly-image-service/blob/master/fastly-io_guim_co_uk/src/main/resources/varnish/main.vcl#L122).


davidfurey:
It is great to see this logic moving into DCR.

I think the risk that the default image could bleed into production is real, and that it would be acceptable to either use a Guardian branded placeholder image or a sensible sized image from the CAPI response (ie not the `isMaster: true` image)

dotcom-rendering/src/web/components/ImageComponent.tsx
```diff
@@ -242,6 +242,19 @@ export const ImageComponent = ({
 
 	const palette = decidePalette(format);
 
+	const master = element.media.allImages.find(
+		(image) => image.fields.isMaster,
+	)?.url;
+
+	if (!master) {
```
AshCorr, line 249: I think `main` is becoming more the norm now instead of `master`? Or `source`?

davidfurey: I think consistency with how the Grid refers to this concept is important.  Unilaterally changing to `main` or `source` here would cause confusion.

oliverlloyd: Thanks for raising this. I agree with @davidfurey we probably don't want to diverge just in DCR but I also think it would be good to try and change this so I raised the question over in The Grid channel. Maybe there's already been some thinking around this.

dotcom-rendering/docs/architecture/029-signing-image-urls.md
```diff
@@ -0,0 +1,35 @@
+# DCR now signs image urls
+
+**Refactor Picture.tsx now that DCR can sign image urls itself**
+
+## Context
+
+For more information on how images work in DCR see [this excellent document](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/docs/architecture/027-pictures.md). The changes discussed here build upon that work but, now that DCR is capable of signing urls, it refactors the code to move that step away from Frontend and into DCR.
+
+The image urls used at the Guardian are generated using the [fastly image optimiser [FIO]](https://developer.fastly.com/reference/io/). This service allows us to specify the image widths we want along with other query params affecting the quality of the resulting image.
+For a request to this servive to be valid it needs to be signed and to date that signing process has happened in Frontend. This architectural change moves that into DCR. This was made possible by [this PR](https://github.com/guardian/dotcom-rendering/issues/4752).
```
mxdvl, line 10:
```suggestion
For a request to this service to be valid it needs to be signed. To date, that signing process has happened in Frontend. This architectural change moves signing into DCR. This was made possible by [PR #4752](https://github.com/guardian/dotcom-rendering/issues/4752).
```

mxdvl (commented):
Exciting

oliverlloyd:
> I think the risk that the default image could bleed into production is real, and that it would be acceptable to either use a Guardian branded placeholder image or a sensible sized image from the CAPI response (ie not the `isMaster: true` image)

I've made a change to remove the default that was being used before, replacing it with master.

I did look at choosing one of the smaller images in that `allImages` array but I wasn't able to make that work. I'm not sure why but if I didn't use master then the quality was very erratic even when the nominal width was high.

I think we still need to make the choice between a Guardian branded default and using master.

### Branded
- Very obvious to developers that they need to take steps to set the `.env` file
- If the `IMAGE_SALT` env var were missing this surfaces the problem very quickly

### Master
- The impact on readers will be much higher page sizes but image content will still work
- Local development will be possible for most tasks

oliverlloyd:
> if resizer is down or says no, we return closest-bigger artifact (via VCL?)

This is interesting. I think this is a valid justification for writing more complex code. The chances of the salt not being available seem insignificant (the server is configured to not even start if it doesn't have one) but the chance of an api call failing or the Fastly service having an outage is very real.

I'll spend some time looking into what Frontend is doing and add a fallback mechanism to DCR.

I think the question of what should we use in the event of the salt not being present it still open though. We could return the fallback but maybe there's benefit in terms of dev ex to serving a placeholder with information and a message?

davidfurey:
The fallback mechanism if the resizer fails is in VCL not frontend, so you get it for free.

> the server is configured to not even start if it doesn't have one

Where is this configuration? I could be misunderstanding `gen-dotenv.js`, but it looks like it does not attempt to validate the `.env` file if it is created from SSM Parameters.  So I think the server will still start if the parameter is accidentally deleted.

jamesgorrie:
> Where is this configuration?

We're working on this over [here](https://github.com/guardian/dotcom-rendering/issues/5155).

> What do we do about Storybook? Are we happy to [use default images there](https://www.chromatic.com/test?appId=5dfcbf3012392c0020e7140b&id=629e17d7d3918c004a6eb0d3)?

I think we might be able to look into a better solution while looking at how we generate / read out app config. We'll make sure we consider it [here](https://github.com/guardian/dotcom-rendering/issues/5155#issuecomment-1149606241).

oliverlloyd:
I've updated this PR's description but to summarise:

There is work underway( See issue [#5155](https://github.com/guardian/dotcom-rendering/issues/5155)) to ensure we guard against the server running in production without an `IMAGE_SALT` - this work should be completed before merging this PR. 

No DCR server should run in production without an `IMAGE_SALT`. If it does an error should be thrown to surface the problem.

A DCR server _can_ run in development mode without and `IMAGE_SALT`. If it does a default image should be used fopr improved dev ex.

Failures in Fastly do not need to be handled by DCR. There is already [fallback code in the VCL](https://github.com/guardian/fastly-image-service/blob/master/fastly-io_guim_co_uk/src/main/resources/varnish/main.vcl#L122).

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -1,319 +1,314 @@
...
-	return getBestSourceForDesiredWidth(300, sources).src;
-};
-
-/**
- * Removes redundant widths from an array of DesiredWidths
+ * Based on image role, if it's main media or not and the format of the article
+ * it is in, this function decides the width the image should have at different
+ * breakpoints.
+ *
+ * Previously, this logic was set [in this Frontend file](https://github.com/guardian/frontend/blob/oliver/immersives-for-all/common/app/layout/ContentWidths.scala)
```
OllysCoding, line 31: +1 Great to provide this context!

OllysCoding:
I noticed in this PR in most places functions were being written as `function(...) { ... }` rather than as `const xxx = () => {}` and I was wondering if there was any reason why?

To my knowledge arrow functions are used much more commonly within DCR but I don't believe we have any particular rules around it?

OllysCoding:
> A DCR server can run in development mode without and IMAGE_SALT. If it does a default image should be used fopr improved dev ex.

Is this logic still-to-come in this PR?

OllysCoding:
🎉 This is a really exciting PR and a huge step for DCR to take!! 

I'm quite interested in whether we can preserve the image resolution logic in our 'fallback image' - specifically I'd like that if we knew every possible width DCR might try and render an image at (which we do), then we can generate all of resolutions needed for the fallback image as some sort of fixture - still ensuring that somebody working on DCR without a .env gets an 'authentic image resolution'

The caveats to that of course being, we'd need to upload the fallback image to the grid (not sure if we're allowed to do this?!) and that trying to implement this may muddy the very clean picture API that's been created. 

However if things like chromatic, etc - are using the fallback image (since they might not have access to aws for .env creds), having it use the 'correct' resolutions would allow greater safety in making changes & adjustments to the picture component in the future!

paperboyo:
No idea, just a thought. Why local dev needs a fake image if it could just do what VCL fallback does and use Grid articfacts directly (it never is correct resolution, though, if that’s important)?

oliverlloyd:
> > A DCR server can run in development mode without and IMAGE_SALT. If it does a default image should be used fopr improved dev ex.
> 
> Is this logic still-to-come in this PR?

Yes. I have a the commit pending, I'm just waiting on the image file from @HarryFischer

OllysCoding:
> Why local dev needs a fake image if it could just do what VCL fallback does and use Grid articfacts directly (it never is correct resolution, though, if that’s important)?

I think for me having the correct resolutions during testing is quite important, as this factors into the results of things like our lighthouse & CWV tests & the 'accuracy' of resource loading in e2e as well as affecting things like the loading times in the dev environment (for example https://www.theguardian.com/uk?dcr is painfully heavy at the moment).

That said if for example in the dev environment there was a reasonable fallback we could chose that's preferable over an accurate fallback image (but I'd still argue for testing we need accurate images resolutions over what the image actually looks like), this could be a fair trade-off. Worth remembering that very few DCR developers will have no .env and if they do it's likely only for the first few hours they use it until someone can provide them with a copy if they don't have janus creds.

oliverlloyd:
> No idea, just a thought. Why local dev needs a fake image if it could just do what VCL fallback does and use Grid articfacts directly (it never is correct resolution, though, if that’s important)?

With a test image it's clear to the developer that they are not getting optimised images and a step is still pending. If we show them fallbacks from VCL they could think DCR is broken

oliverlloyd:
> I'm quite interested in whether we can preserve the image resolution logic in our 'fallback image'

This is actually a great idea! We can absolutely upload the test image to the Grid and use the resulting image sources as a fallback. The downside to this is these sources will not be generated by DCR so any changes to that logic will not be reflected in the fixture but I think if we want to update that data we can do it manually and if that proves pesky we can automate it.

I'll work on making this change 👍

oliverlloyd:
> I noticed in this PR in most places functions were being written as `function(...) { ... }` rather than as `const xxx = () => {}` and I was wondering if there was any reason why?

🤔 In my head I have this weird rule where I use `function` to declare simple local functions used inside a component, and then arrow functions for "👋 more complicated 👋 " functions that get exported.

This is 100% my own weird way of thinking and upon reflection not really based on any technical reasoning at all. I guess the only basis for it is as a way to communicate that a function is "private". So.. myeh, I can see how that doesn't work and I'm very happy to dump my weird rule!

mxdvl:
One key thing to note about arrow functions is that they are not hoisted, so you need to make sure that they are defined before being called. I believe we have an ESLint rule for this.

JamieB-gu:
> However if things like chromatic, etc - are using the fallback image (since they might not have access to aws for .env creds)

Is this the case in Chromatic? It's not using the generated image URLs because it doesn't have access to the image salt?

OllysCoding:
> It's not using the generated image URLs because it doesn't have access to the image salt?

It could be that we can add the image salt as a secret in github and update gen-env or the storybook webpack to be able to consume it, but until then - yes it won't be able to generate image URLs

JamieB-gu:
> It could be that we can add the image salt as a secret in github and update gen-env or the storybook webpack to be able to consume it, but until then - yes it won't be able to generate image URLs
> 
> Edit: Because we publish out storybook publicly, if we did add the image salt to the storybook build we would be exposing that salt to the public which we do not want to do

I don't know if this is helpful, but AR/image-rendering solves this problem by separating the URL generation from the components. The components take a set of URLs that have already been built from the image salt, wrapped up in an `Image` object:

https://github.com/guardian/dotcom-rendering/blob/fd05189f022656e63f17aad5469655a1601313ac/common-rendering/src/image.ts#L8-L16

https://github.com/guardian/dotcom-rendering/blob/fd05189f022656e63f17aad5469655a1601313ac/common-rendering/src/components/img.tsx#L36-L37

This way the components never need access to the image salt, and so neither do storybook/Chromatic: you can just put pre-generated URLs in your fixtures.

The components still control the presentational side: they specify the slots that images should populate with `sizes`. This is roughly analogous to the separation the browsers make for responsive images: `srcset` provides you with the data, and the `sizes` attribute specifies the presentation.

Some helpers for sizes available here if you're interested: https://github.com/guardian/dotcom-rendering/blob/fd05189f022656e63f17aad5469655a1601313ac/common-rendering/src/sizes.ts#L10-L18

OllysCoding:
> How does the salt get exposed? Isn't the output from the storybook build html? I could be missing something here but I thought SB publishes the same kind of output that we send to browsers so it should be okay to have this public.

My understanding is that Storybook published a html file which is the entry-point for a react app which is responsible for loading & rendering all of our stories. It's essentially one big SPA - all client-side rendered

What this means is somewhere there will be some webpack-ified script which is downloaded to the client and responsible for rendering our storybook components.

e.g

```
// Source code (Picture.tsx)
console.log(process.env.IMAGE_SALT);

// Webpack output (main.js)
console.log("The actual secret is here")
```

THerefor if anyone wanted to look into the main.js file downloaded onto their client, they would be able to find our image salt hidden in there !

oliverlloyd:
> It's essentially one big SPA - all client-side rendered

I see, I see, I see. Thanks for explaining this (here and in our chat). So yes, env vars cannot be used inside Storybook.

dotcom-rendering/docs/architecture/029-signing-image-urls.md
```diff
@@ -0,0 +1,35 @@
...
+
+-   the role/weighting given to an image
+-   the format for the article the image sits in and
+-   whether the image is being used as the main media or not
+
+There is now no other code anywhere else which influences this decision. This will make future changes easier.
+
+## No need to pick a source based on a desired width
+
+This paradign is no longer needed because DCR is now deciding which width to use at each breakpoint so we no longer need to treat the list of sources Frontend gave us as a range of options; we're now building them ourselves and so we're able to be exact in our decisions.
```
oliverlloyd, line 35:
```suggestion
This paradigm is no longer needed because DCR is now deciding which width to use at each breakpoint so we no longer need to treat the list of sources Frontend gave us as a range of options; we're now building them ourselves and so we're able to be exact in our decisions.
```

dotcom-rendering/src/web/components/ImageComponent.tsx
```diff
@@ -242,6 +242,27 @@ export const ImageComponent = ({
...
+	const master = element.media.allImages.find(
+		(image) => image.fields.isMaster,
+	)?.url;
+
+	// Legacy images do not have a master so we fallback to the largest available
+	const descendingByWidth = (a: Image, b: Image) => {
+		return parseInt(b.fields.width) - parseInt(a.fields.width);
+	};
+	const fallbackImage =
+		element.media.allImages.sort(descendingByWidth)[0].url;
```
mxdvl, line 254: Does `sort` modify the array in-place? Should we `slice` it first?

dotcom-rendering/src/web/components/ImageComponent.tsx
```diff
@@ -242,6 +242,27 @@ export const ImageComponent = ({
...
+
+	const isSupported = (imageUrl: string) => {
+		const supportedImages = ['jpg', 'jpeg', 'png'];
+		const extension = imageUrl.split('.').pop();
+		return extension && supportedImages.includes(extension);
```
mxdvl, line 259: Is there any risk that the image format is uppercase?

FYI, you can use `.slice(-1)` for an immutable version of `pop()`.

dotcom-rendering/src/web/components/ImageComponent.tsx
```diff
@@ -242,6 +242,27 @@ export const ImageComponent = ({
...
 
+	const master = element.media.allImages.find(
+		(image) => image.fields.isMaster,
+	)?.url;
+
+	// Legacy images do not have a master so we fallback to the largest available
+	const descendingByWidth = (a: Image, b: Image) => {
+		return parseInt(b.fields.width) - parseInt(a.fields.width);
+	};
+	const fallbackImage =
```
mxdvl, line 253: Do we want to generate a fallback image if we already have a `master`?This seems like unnecessary work for all modern images.

I would also favour `largestImage` rather than `fallbackImage`, to be more explicit.

```ts
const image = master ?? getLargestImage()
```

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -1,320 +1,320 @@
...
-			isMainMedia,
-			format,
-		),
-	}));
-
-	const desiredWidths: DesiredWidth[] =
-		removeRedundantWidths(allDesiredWidths);
+	const imageWidths = decideImageWidths({ role, format, isMainMedia });
+	const sources = imageWidths
+		.sort(descendingByBreakpoint)
```
mxdvl, line 231: Sort modifies the array in-place, you might want to `.slice()` it first.

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -1,320 +1,320 @@
...
-		return breakpoint;
-	}
+const generateSignedUrl = ({
+	master,
+	imageWidth,
+	quality,
+}: {
+	master: string;
+	imageWidth: number;
+	quality: 'low' | 'high';
```
mxdvl, line 177: I'm not sure the word `quality` is adequate, here. I would prefer `resolution` or `density`, as higher resolution images get a lower quality value.

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -1,320 +1,320 @@
...
-			if (breakpoint >= breakpoints.leftCol) return 780;
-			if (breakpoint >= breakpoints.phablet) return 620;
-			return breakpoint;
-	}
+	// Construct and sign the url
+	const url = new URL(master);
+	const service = url.hostname.split('.')[0];
+	const params = new URLSearchParams({
+		width: imageWidth.toString(),
+		quality: quality === 'high' ? '45' : '85',
```
mxdvl, line 204: The values `45` and `85` are tied to the VCL, so it would be good to hint towards where this information can be found. I can add the link if you reply with a 👍

mxdvl (commented):
It's nice to bring this logic home!

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -1,320 +1,320 @@
...
+				// network connections so we're going to fallback to the smallest image
+				src={sources[0].lowResUrl}
 				height={height}
 				width={width}
 				loading={
 					isLazy && !Picture.disableLazyLoading ? 'lazy' : undefined
 				}
 				// https://stackoverflow.com/questions/10844205/html-5-strange-img-always-adds-3px-margin-at-bottom
-				// why did we add the css `vertical-align: middle;` to the img tag
+				// explains why we added the css `vertical-align: middle;` to the img tag
```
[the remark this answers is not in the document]
OllysCoding: Might this change be better approached in a separate PR?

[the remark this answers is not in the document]
mxdvl: #5225

dotcom-rendering/src/web/components/ImageComponent.tsx
oliverlloyd: Both fixed 👍

dotcom-rendering/src/web/components/Picture.tsx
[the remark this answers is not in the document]
oliverlloyd: This code is unrelated to the changes here. I see that there's [a PR open for this](https://github.com/guardian/dotcom-rendering/pull/5225) (nice!) but that it has visual diffs (😨) so maybe the align middle trick is still needed? Or maybe we have coded on top of it  now and depend on img's not being display block which might be hard to change (😞)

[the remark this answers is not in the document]
oliverlloyd: Ah, I missed those later comments. You're on top of this 👍

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -1,323 +1,332 @@
...
+		// If no IMAGE_SALT env variable is set in production then something has
+		// gone wrong and we want to know about it
+		throw new Error(
+			'No IMAGE_SALT value found when constructing picture source',
+		);
+	} else if (!salt) {
+		// If no IMAGE_SALT is set outside of production it likely means the
+		// developer has yet to set the env file. In this case we continue
+		// using a default image
+		return 'https://i.guim.co.uk/img/media/52f8e9183bbc03c48fcdf4ed9a4e12b3a19a9927/0_0_5000_3000/master/5000.jpg?width=700&quality=45&auto=format&fit=max&dpr=2&s=196c7f25bf2fddcb45512d3ff25e20c0';
```
OllysCoding, line 196: This isn't something for this PR but in the future I'd love to have a bunch of signed urls for all possible image widths and DPRs so this reflects the 'correct' resolutions wherever it's rendered!

OllysCoding (approved):
+1 What an exciting step forward!

PR #5152, opened 2022-06-07
Add role attribute to unordered lists

MarSavar:
## What does this change?
Adds a `role="list"` to `<ul>` tags in AR components.

## Why?
Because Safari won't recognise them as lists if they are styled with `list-style: none`. 
See https://developer.mozilla.org/en-US/docs/Web/CSS/list-style#accessibility_concerns and #5041 for more details.

As described in the MDN docs, adding the role of list is necessary to avoid the issue in Safari.

## Screenshots

N/A

### DIFF
--- a/apps-rendering/config/.eslintrc.js
+++ b/apps-rendering/config/.eslintrc.js
@@ -96,8 +96,14 @@ module.exports = {
 		'import/no-default-export': 'off',
 		// The CSS `list-style: none` causes accessibility issues in Safari when applied to unordered lists (<ul>).
 		// Adding the `role="list"` attribute to the <ul> tag fixes the issue in Safari, but triggers
-		// the `no-redundant-roles` linting error.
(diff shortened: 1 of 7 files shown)


MarSavar:
@mxdvl Would love to hear your thoughts on disabling the linting rule!

mxdvl (commented):
While [the docs](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-redundant-roles.md) are not super clear, you can add an explicit key to the rule for `ul`:

```js
{
  'jsx-a11y/no-redundant-roles': [
    'error',
    {
      nav: ['navigation'],
      ul: ['list'],
    },
  ],
}
```

(unless I've misunderstood)

MarSavar:
Thanks @mxdvl - that's exactly what I needed. Tested locally and it works!

mxdvl (approved):
Happy this worked!

PR #5159, opened 2022-06-08
Explicitly exit deno once Lighthouse script has finished

AshCorr:
## What does this change?

Adds a `Deno.exit(0)` after action has completed.

## Why?

I'm not a deno expert, but I'm assuming that deno isn't exiting because its waiting for a promise to be fullfilled somewhere? This issue in octokit sounds potentially related and might be the root issue https://github.com/octokit/octokit.js/issues/2079 , although I'm not quite sure why it started now.

This PR makes deno explicitly exit with a success code once all tasks have been completed.

### DIFF
--- a/scripts/deno/surface-lighthouse-results.ts
+++ b/scripts/deno/surface-lighthouse-results.ts
@@ -182,7 +182,7 @@ try {
 		} Lighthouse report comment`,
 	);
 	console.log('See:', data.html_url);
-	Deno.exit(0);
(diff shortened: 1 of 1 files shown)


AshCorr:
Potentially related https://github.com/octokit/octokit.js/issues/2079

ioannakok (approved):

bryophyta (approved):

OllysCoding (approved):

PR #5160, opened 2022-06-08
Remove problematic article from fixtures

oliverlloyd:
## What does this change?
Drops the usage of a (potentially) problematic article from our fixtures

I also chased my tail a bit here and added `Gallery`, `Audio` and `Video` fixtures. In the end I don't think I _had_ to do this but it's no bad thing so I left these edits in. Apologies for the noise!

## Why?
I think this article is causing false negatives in Chromatic diffs. I _think_. It's an experiment

### DIFF
--- a/dotcom-rendering/src/web/layouts/Layout.stories.tsx
+++ b/dotcom-rendering/src/web/layouts/Layout.stories.tsx
@@ -104,7 +104,7 @@ for (const [displayName] of Object.entries(ArticleDisplay)) {
 						});
 
 						const fixture =
-							Fixtures[designName] || Fixtures.Article;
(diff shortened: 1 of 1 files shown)


AshCorr (approved):
Looks good, however I am a bit confused, what does adding these new fixtures do?

mxdvl (approved):
Fingers-crossed-driven-development

bryophyta:
> Looks good, hopefully this works! Out of curiosity what does adding these new fixtures do?

Likewise, I'm curious about why some of the `elementId` values change, without the content itself being changed? Are they automatically generated `uuid`-like ids, or do they correspond to something in CAPI/Frontend, etc.?

Issue #5155, opened 2022-06-08
App secrets and config should be injected at start rather than build time


jamesgorrie:
It is common practice at the Guardian to have config and secrets injected at runtime. We should follow this pattern so our application work as expected.

Part of the reason for this is that if config has changed, we wouldn't want to have to wait for another build to bake for us to be able to change it. An example of this is if we cycle a secret for security issues, we'd like to just redeploy and have the config take affect.

jamesgorrie:
The solution to this might take into account that we might want to access secrets in other platforms such as github actions, or chromatic.

jamesgorrie:
We could potentially split the makefile into:
```
- start

-- production
--- gen file
--- validate env vars
--- fail on anything

-- dev
--- check creds
--- if available gen file
--- if not, do nothing and log
```

PR #5162, opened 2022-06-09
`.env` - Extra production protections & move injection from buildtime > runtime

OllysCoding:
Co-author: @jamesgorrie 

## What does this change?

This PR addresses the issues raised in https://github.com/guardian/dotcom-rendering/issues/5155:
1. Prior to this PR, dotenv variables were injected at builld time during webpack, however standard practice within the company is to have this happen at run or deploytime
2. There were not enough safeguards or clear flow to ensure we always have a valid .env file

**To address this we:**

Moved from using `dotenv-webpack` in the webpack config to directly using the`dotenv` package in the server entrypoint. This means that whenever the server is started; it loads in the .env file. 

As part of this I updated the documentation to clarify the limitations of this way of adding dotenv variables, as well as highlighting the existing security risk of leaking our dotenv variables into clientside code.

Broke `gen-dotenv` script into two separate flows, one for dev and one for prod, the new flows being: 

PROD (Failure on any of these steps will result in the server not starting):
```mermaid
flowchart LR
    id1("Prod start") --> id2("generate dotenv") --> id3("validate generated dotenv has all keys set") --> id4("start server")
```  

DEV:
```mermaid
flowchart LR
id1("Dev start") --> id2("Attempt to generate .env")
id2 -- success --> id4("start dev server")
id2 -- fail --> id5("does user have valid existing env?")
id5 -- yes --> id4
id5 -- no -->  id8("warn user they do not have valid env") --> id4
```

This should provide a secure prod experience where the server is unable to start unless it has a newly generated, valid `.env` file, and ensure that the dev experience both with and without credentials is as clear and easy to use as possible.

### DIFF
--- a/dotcom-rendering/docs/architecture/028-dotenv.md
+++ b/dotcom-rendering/docs/architecture/028-dotenv.md
@@ -38,3 +38,27 @@ const secrets = [
     // ...
 ];
 ```
+
+## Security
+
+`.env` variables are only available on the server, you should ensure that you **never** pass an environment variable into the clientside code
+e.g
+
+```jsx
+<Island>
+	<!-- Bad :((( -->
+	<MyComponent myProp={process.env.MY_VAR} />
+</Island>
+```
+
+This has considerable security implementations as this variable would then be available to any consumer of the site.
+
+## Limitations
+
+### Server Only
+
+DCR currently only supports `.env` variables being used on the _server_, but not the client.
+
+This limitation exists as the to maintain common & best practice within the organisation, we want to ensure that a re-deploy of the application would involve fetching a fresh environment variable. This means we cannot [inject environment variables through webpack](https://www.npmjs.com/package/dotenv-webpack) (which is done at build time) - and instead we load the dotenv directly when running the server bundle.
+
+This could be revisited in the future, but when doing so the [security implications](#security) of having `.env` variables available on the client should be considered and documented.
--- a/dotcom-rendering/makefile
+++ b/dotcom-rendering/makefile
@@ -31,23 +31,25 @@ deploy:
 # prod #########################################
 
 build: export NODE_ENV=production
-build: clean-dist install gen-dotenv
+build: clean-dist install
 	$(call log, "building production bundles")
 	@webpack --config ./scripts/webpack/webpack.config.js --progress
 
 build-ci: export NODE_ENV=production
-build-ci: clean-dist install gen-dotenv
+build-ci: clean-dist install
 	$(call log, "building production bundles")
 	@SKIP_LEGACY=true webpack --config ./scripts/webpack/webpack.config.js
 
-start-ci: install
+start-ci: export NODE_ENV=production
+start-ci: install gen-dotenv
 	$(call log, "starting PROD server...")
 	@NODE_ENV=production DISABLE_LOGGING_AND_METRICS=true node dist/frontend.server.js
 
-start: install
+start: export NODE_ENV=production
+start: install gen-dotenv
 	$(call log, "starting PROD server...")
 	@echo '' # just a spacer
-	@NODE_ENV=production pm2 start dist/frontend.server.js
+	pm2 start dist/frontend.server.js
 	@echo '' # just a spacer
 	$(call log, "PROD server is running")
 	@pm2 logs
--- a/dotcom-rendering/scripts/dotenv/gen-dotenv.js
+++ b/dotcom-rendering/scripts/dotenv/gen-dotenv.js
@@ -35,14 +35,21 @@ const genEnv = async (parameters) => {
 };
 
 const prod = async () => {
-	const params = await getAwsSsmParameters('prod');
-	await genEnv(params);
-	const isValid = await checkEnv();
-	if (!isValid) {
-		warn(
-			'[scripts/dotenv] Could not validate that the .env file was generated correctly',
-		);
-		throw new Error('Could not validate .env file was generated correctly');
+	try {
+		const params = await getAwsSsmParameters('prod');
+		await genEnv(params);
+		const isValid = await checkEnv();
+		if (!isValid) {
+			warn(
+				'[scripts/dotenv] Could not validate that the .env file was generated correctly',
+			);
+			throw new Error(
+				'Could not validate .env file was generated correctly',
+			);
+		}
+	} catch (err) {
+		console.error(err);
+		process.exit(1);
 	}
 };
 
--- a/dotcom-rendering/scripts/dotenv/gen-dotenv.js
+++ b/dotcom-rendering/scripts/dotenv/gen-dotenv.js
@@ -1,5 +1,7 @@
-const { CredentialsProviderError } = require('@aws-sdk/property-provider');
 const path = require('path');
+const { CredentialsProviderError } = require('@aws-sdk/property-provider');
+
+console.log(typeof ExpiredTokenException);
 const fs = require('fs').promises;
 const { prompt, log, warn } = require('../env/log');
 const secrets = require('../secrets');
@@ -41,7 +43,7 @@ const genEnv = async () => {
 		const validEnv = await checkEnv();
 		if (!validEnv) {
 			log(
-				'[scripts/dotenv] .env file is missing, attemting to generate it from AWS parameters...',
+				'[scripts/dotenv] .env file is missing or out of date, attemting to generate it from AWS parameters...',
 			);
 
 			await genEnv();
@@ -51,7 +53,13 @@ const genEnv = async () => {
 			log('[scripts/dotenv] valid .env file exists, moving on...');
 		}
 	} catch (err) {
-		if (err instanceof CredentialsProviderError) {
+		console.log(err);
+		console.log(typeof err);
+		if (
+			err instanceof CredentialsProviderError ||
+			// eslint-disable-next-line no-underscore-dangle -- '__type' is the only way to identify this exception type
+			err.__type === 'ExpiredTokenException'
+		) {
 			const PROD = process.env.NODE_ENV === 'production';
 			if (PROD) {
 				warn(
--- a/dotcom-rendering/scripts/dotenv/gen-dotenv.js
+++ b/dotcom-rendering/scripts/dotenv/gen-dotenv.js
@@ -1,9 +1,7 @@
+const fs = require('fs').promises;
 const path = require('path');
 const { CredentialsProviderError } = require('@aws-sdk/property-provider');
-
-console.log(typeof ExpiredTokenException);
-const fs = require('fs').promises;
-const { prompt, log, warn } = require('../env/log');
+const { prompt, warn } = require('../env/log');
 const secrets = require('../secrets');
 const { getAwsSsmParameters } = require('./get-aws-ssm-parameters');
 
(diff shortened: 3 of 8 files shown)


dotcom-rendering/docs/architecture/028-dotenv.md
```diff
@@ -38,3 +38,27 @@ const secrets = [
...
+```jsx
+<Island>
+	<!-- Bad :((( -->
+	<MyComponent myProp={process.env.MY_VAR} />
+</Island>
+```
+
+This has considerable security implementations as this variable would then be available to any consumer of the site.
+
+## Limitations
```
jamesgorrie, line 56: Do we have any here?

dotcom-rendering/docs/architecture/028-dotenv.md
```diff
@@ -38,3 +38,27 @@ const secrets = [
...
+
+`.env` variables are only available on the server, you should ensure that you **never** pass an environment variable into the clientside code
+e.g
+
+```jsx
+<Island>
+	<!-- Bad :((( -->
+	<MyComponent myProp={process.env.MY_VAR} />
+</Island>
+```
```
jamesgorrie, line 52: Maybe we could give a good example here e.g. using the env var to pass a salted URL to the component?

What happens if we read a process.env in an `.importable`?

```suggestion
```jsx
return (
    <Island>
	    <!-- Bad :((( -->
	    <Image imageSalt={process.env.IMAGE_SALT} />
    </Island>
)

const url = getSaltedUrl(process.env.IMAGE_SALT)
return (
    <Island>
	    <!-- Good :))) -->
	    <Image url={url} />
    </Island>
)
```
```

File: dotcom-rendering/docs/architecture/028-dotenv.md [hunk 3]
```diff
@@ -38,3 +38,27 @@ const secrets = [
...
+
+This has considerable security implementations as this variable would then be available to any consumer of the site.
+
+## Limitations
+
+### Server Only
+
+DCR currently only supports `.env` variables being used on the _server_, but not the client.
```
jamesgorrie, line 60: nit: 
```suggestion
DCR currently only supports `.env` variables being used on the _server_, not the client.
```

dotcom-rendering/docs/architecture/028-dotenv.md
```diff
@@ -38,3 +38,27 @@ const secrets = [
...
+
+This limitation exists as the to maintain common & best practice within the organisation, we want to ensure that a re-deploy of the application would involve fetching a fresh environment variable. This means we cannot [inject environment variables through webpack](https://www.npmjs.com/package/dotenv-webpack) (which is done at build time) - and instead we load the dotenv directly when running the server bundle.
```
jamesgorrie, line 62: Does the fact we only make this available on the server ensure that we can make changes to env vars on redeploy? Or is it that we run this on start, not build?

dotcom-rendering/scripts/dotenv/gen-dotenv.js
```diff
@@ -11,77 +11,100 @@ const checkEnv = async () => {
 	try {
 		const env = (await fs.readFile(ENV_PATH)).toString();
 
-		let valid = true;
+		let isValid = true;
 		for (const secret of secrets) {
-			const regex = new RegExp(`^${secret.key.replace('.', '\\.')}=.*`);
-			if (!env.match(regex)) valid = false;
+			const regex = new RegExp(`^${secret.key.replace('.', '\\.')}=.+`);
+			if (!env.match(regex)) isValid = false;
```
jamesgorrie, line 17: We could just return here, but what ever is clearer.

jamesgorrie (commented):
A few comments on the docs

dotcom-rendering/docs/architecture/028-dotenv.md
OllysCoding: > What happens if we read a process.env in an .importable

On the client this will simply be undefined (and perhaps cause an error) and on the server is should work as expected

dotcom-rendering/docs/architecture/028-dotenv.md
OllysCoding: Both are necessary for this to be possible on redeploy. To make env variables available on the client you need to [inject them through webpack](https://github.com/mrsteele/dotenv-webpack#resulting-bundle), which happens at build time - moving to loading them with the regular `dotenv` package enabled us to use gen-env on start rather than build.

dotcom-rendering/docs/architecture/028-dotenv.md
OllysCoding: I'm not sure I understand the question!

dotcom-rendering/scripts/dotenv/gen-dotenv.js
```diff
@@ -11,77 +11,99 @@ const checkEnv = async () => {
...
 
-// eslint-disable-next-line @typescript-eslint/no-floating-promises
-(async () => {
+const prod = async () => {
 	try {
-		const validEnv = await checkEnv();
-		if (!validEnv) {
-			log(
-				'[scripts/dotenv] .env file is missing, attemting to generate it from AWS parameters...',
+		const params = await getAwsSsmParameters('prod');
```
oliverlloyd, line 38: Is this always going to be 'prod'?

dotcom-rendering/scripts/dotenv/gen-dotenv.js
```diff
@@ -11,77 +11,99 @@ const checkEnv = async () => {
...
+			const isValid = await checkEnv();
+			if (isValid) {
 				prompt(
-					`[scripts/dotenv]  * ${secret.key}: ${secret.missingMessage}`,
+					'[scripts/dotenv] .env file is valid but has not been updated',
+				);
+			} else {
+				prompt(
+					"[scripts/dotenv] This won't stop dotcom-rendering from working, it will just vary from PROD by:",
 				);
```
oliverlloyd, line 80: This text seems to depend on another message for context but maybe it could have that context built into it?

dotcom-rendering/scripts/dotenv/gen-dotenv.js
OllysCoding: Because we've separated out the functions which are called, e.g `prod` and `dev`, we can hardcode these values here. 

There may be a future where we want to also have 'code', however for this we'd need to introduce a new environment variable passed that isn't NODE_ENV (e.g GU_ENV) which isn't in the scope of this PR

dotcom-rendering/docs/architecture/028-dotenv.md
jamesgorrie: Apologies, I didn't realise that this is a heading for the next points...

dotcom-rendering/docs/architecture/028-dotenv.md
jamesgorrie: Worth adding the good example so people know how they might do it?

dotcom-rendering/docs/architecture/028-dotenv.md
```diff
@@ -38,3 +38,36 @@ const secrets = [
     // ...
 ];
 ```
+
+## Security
+
+`.env` variables are only available on the server, you should ensure that you **never** pass an environment variable into the clientside code
```
mxdvl, line 44:
```suggestion
`.env` variables are only available on the server, you should ensure that you **never** pass an environment variable into the client-side code.
```

dotcom-rendering/makefile
```diff
@@ -30,24 +30,23 @@ deploy:
...
 	$(call log, "building production bundles")
-	@webpack --config ./scripts/webpack/webpack.config.js --progress
+	@NODE_ENV=production webpack --config ./scripts/webpack/webpack.config.js --progress
 
-build-ci: export NODE_ENV=production
-build-ci: clean-dist install gen-dotenv
+build-ci: clean-dist install
 	$(call log, "building production bundles")
-	@SKIP_LEGACY=true webpack --config ./scripts/webpack/webpack.config.js
+	@NODE_ENV=production SKIP_LEGACY=true webpack --config ./scripts/webpack/webpack.config.js
```
mxdvl, line 39: We’re doing a bit of mix-and-match here. Do we want to settle on one approach?

Either `export NODE_ENV=production` or `@NODE_ENV=production`.

dotcom-rendering/scripts/dotenv/gen-dotenv.js
```diff
@@ -11,77 +11,100 @@ const checkEnv = async () => {
 	try {
 		const env = (await fs.readFile(ENV_PATH)).toString();
 
-		let valid = true;
 		for (const secret of secrets) {
-			const regex = new RegExp(`^${secret.key.replace('.', '\\.')}=.*`);
-			if (!env.match(regex)) valid = false;
+			const regex = new RegExp(`^${secret.key.replace('.', '\\.')}=.+`);
```
mxdvl, line 15: Why the double-backslash in the `replace`?

Maybe outside the scope of this PR, but it would be nice to always have `RegExp` that are documented and testable. We could also circumvent the regex by splitting the string on `\n` and  `=`.

```ts
/*
KEY=value
OTHER_KEY=other_value
*/
const content = s.readFileSync('file.txt').toString().split("\n");
const entries = content.filter(line => line.trim() !== "").map(line => line.split("="));
const env = Object.fromEntries(entries);
/*
{
  "KEY": "value",
  "OTHER_KEY": "other_value"
} 
*/
```

mxdvl (approved):
Nice improvements.

It might be worth having a check for the presence of secrets in the final client bundles. I believe it could be as simple as `grep -rl "dist" -e "secret"` and checking there’s no result beyond `dist/frontend.server.js`.

> #### [`grep` man page](https://linux.die.net/man/1/grep)
> `-R`, `-r`, `--recursive`
> Read all files under each directory, recursively; this is equivalent to the `-d` recurse option.
> 
> `-l`, `--files-with-matches`
> Suppress normal output; instead print the name of each input file from which output would normally have been printed. The scanning will stop on the first match. (`-l` is specified by POSIX .)
> `-e PATTERN`, `--regexp=PATTERN`
> Use `PATTERN` as the pattern. This can be used to specify multiple search patterns, or to protect a pattern beginning with a hyphen (-). (`-e` is specified by POSIX .)

dotcom-rendering/scripts/dotenv/gen-dotenv.js
OllysCoding: > Why the double-backslash in the replace

I believe this is required because of the weird way that the regex is getting added as a string? 

>  could also circumvent the regex by splitting the string on \n and =.

I'd agree this is the nicer solution so I'll give this a shot just now :)

mxdvl: And you can then check for the existence of a key using `'key' in obj`

OllysCoding: > And you can then check for the existence of a key using 'key' in obj

We actually want to check that both the key exists and is not empty (e.g `IMAGE_SALT=` should error)

dotcom-rendering/docs/architecture/028-dotenv.md
jamesgorrie: Ah, OK I see. I suppose I am getting confused as there are a few points:
* We don't want to expose these values to the client for security reasons (all are considered secrets)
* We want to have these values recyclable on redeploy

I wonder if these aren't limitations but what we expect, and should expect going forward?

dotcom-rendering/makefile
jamesgorrie: Running `export`  exports the value to all child processes created from that shell, which is why you would run it at the beginning of the step and it will percolate down.

Having `NODE_ENV other/stuff/here` just sets it for that process.

The `@` suppresses the echoing of the command:
https://stackoverflow.com/questions/9967105/suppress-echo-of-command-invocation-in-makefile

OllysCoding: Yea the specific reason for this is if we had node_env set to production in CI it would always fail

OllysCoding: However I might go back and try and find a tidier way to do this so that it's more clear thats why it works that way!

dotcom-rendering/scripts/dotenv/gen-dotenv.js
oliverlloyd: Gotcha. Thanks @OllysCoding 

Having `CODE` marked up as `prod` is an existing pattern that I'm happy for this PR to continue to support.

Outside of this PR, I do quite like how `CODE` can be considered production like but I also see how this can reduce flexibility and the ability to customise things.

oliverlloyd (approved):
Great improvements 👍

PR #5163, opened 2022-06-09
Update Relevant Yield AMP script

domlander:
## What does this change?

The Relevant Yield AMP script domain has been changed from `pbs.relevant-digital.com` to `guardian-pbs.relevant-digital.com`.

Note that the Relevant Yield cookie sync script that lives in the frontend repo has not been changed.

## Why?

Relevant Yield have asked us to make this change to their domain before our test with them next week (where we switch them on and switch off Pubmatic).

## Screenshots

Shows the ads loading locally with the new script:

screenshot


arelra (approved):

chrislomaxjones (approved):
👍 

I think you can go ahead and approve the Chromatic diffs, as they're unrelated false positives.

Issue #4720, opened 2022-04-26
What Containers might we approach first?


mxdvl:
Most used containers, sorted by usage count.

List based on [this Google Doc][1]. For corresponding designs, [see the Figma file][2].

[1]: https://docs.google.com/document/d/1XTqrTkiSkJwaRuTYscXenUDHbrr37u5eap9EZ9ANjes/edit
[2]: https://www.figma.com/file/sx2vMFHbL7SsUo0LcpsKNe/%E2%AC%A3--Front-container?node-id=0%3A1

- [x] #4767
- [x] #5132
- [x] #5133
- [x] #5134
- [x] #5135
- [x] #5136
- [X]  #4572 
- [x] #5137
- [x] #5138
- [ ] #5139
- [x] #5140
- [x] #5141
- [x] #5142
- [x] #5143
- [X] #4569 
- [x] #5144
- [ ] #5145
- [ ] #5146
- [X] #4718 
- [x] #5147
- [ ] #5148
- [ ] #5149
  
### Other containers
  
- [ ] #5150
- [x] #5151

jlieb10:
@

mxdvl:
@jlieb10 are you referring to the `data-link-name` and `data-component` attributes? These are added automatically to `ContainerLayout` and `Card`.

PR #5165, opened 2022-06-09
Add Fronts container: fixed/small/slow-III

### LINKED ISSUE #5132: "fixed/small/slow-III" (`323`)

bryophyta:
## What does this change?

Adds a container for the `fixed/small/slow-III` Fronts layout.

## Why?

Part of the Fronts migration (see: #4720, closes #5132)  

## Screenshots

| Before (Frontend prod)    | After (DCR local)     |
|-------------|------------|
| Before on Frontend | After on DCR |
| Before, Frontend, tablet | After, DCR, tablet |
| Before, Frontend, mobile | After, DCR, mobile |

Two of the visible differences ('Show More' button and 'Hide' link) are distinct features which will be added separately for all containers, so not part of this PR.

### DIFF
--- a/dotcom-rendering/src/web/components/Card/components/TrailTextWrapper.tsx
+++ b/dotcom-rendering/src/web/components/Card/components/TrailTextWrapper.tsx
@@ -28,7 +28,7 @@ export const TrailTextWrapper = ({
 				padding-right: 5px;
 				padding-bottom: 6px;
 
-				${until.desktop} {
+				${until.tablet} {
 					display: none;
 				}
 			`}
--- a/dotcom-rendering/src/web/components/Card/components/TrailTextWrapper.tsx
+++ b/dotcom-rendering/src/web/components/Card/components/TrailTextWrapper.tsx
@@ -28,7 +28,7 @@ export const TrailTextWrapper = ({
 				padding-right: 5px;
 				padding-bottom: 6px;
 
-				${until.tablet} {
+				${until.desktop} {
 					display: none;
 				}
 			`}
--- a/dotcom-rendering/src/web/components/FixedSmallSlowIII.tsx
+++ b/dotcom-rendering/src/web/components/FixedSmallSlowIII.tsx
@@ -4,76 +4,65 @@ import { LI } from './Card/components/LI';
 import { UL } from './Card/components/UL';
 
 type Props = {
-	collectionId: string;
 	trails: TrailType[];
 	containerPalette?: DCRContainerPalette;
 	showAge?: boolean;
-	hasMore: boolean;
 };
 
 export const FixedSmallSlowIII = ({
-	collectionId,
 	trails,
 	containerPalette,
 	showAge,
-	hasMore,
 }: Props) => {
 	const smallTrails = trails.slice(0, 3);
 
 	return (
-		<>
-			<UL direction="row">
-				{smallTrails.map((trail, index) => {
-					return (
-						<LI
-							padSides={true}
-							showDivider={index > 0}
-							padBottomOnMobile={true}
-							percentage={index === 0 ? '50%' : '25%'}
-						>
-							<Card
-								containerPalette={containerPalette}
-								showAge={showAge}
-								linkTo={trail.url}
-								format={trail.format}
-								headlineText={trail.headline}
-								headlineSize={index === 0 ? 'large' : 'medium'}
-								byline={trail.byline}
-								showByline={trail.showByline}
-								showQuotes={
-									trail.format.design ===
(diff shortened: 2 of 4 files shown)


dotcom-rendering/src/web/components/Card/components/TrailTextWrapper.tsx
```diff
@@ -28,7 +28,7 @@ export const TrailTextWrapper = ({
 				padding-right: 5px;
 				padding-bottom: 6px;
 
-				${until.tablet} {
+				${until.desktop} {
```
bryophyta, line 31: I've added this here because it recreates the behaviour of this container on Frontend. However, it's the sort of change that will have implications for all other cards, too. If there are other requirements for other containers, we'll need to introduce different ways of setting this value for different containers. Any thoughts?

dotcom-rendering/src/web/components/FixedSmallSlowIII.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+	return (
+		<UL direction="row">
+			{smallTrails.map((trail, index) => {
+				return (
+					<LI
+						padSides={true}
+						showDivider={index > 0}
+						padBottomOnMobile={true}
+						percentage={index === 0 ? '50%' : '25%'}
```
bryophyta, line 27: having these properties decided inline is a bit awkward, but the obvious alternative (having two distinct component calls which hard-code all of their relevant values) is a bit awkward too, as it makes it difficult to see what the cards in this container have in common. Given that the decision-rule here is always based on the same criterion ("is it the first card or not?") having the logic inline doesn't seem to bad to me here, but interested to hear people's thoughts!

bryophyta: Thanks Ash! I like your suggestion, do we know roughly how many options there would need to be? My only concern is that the system would need to be designed well to avoid recreating the existing complexity, adding multiple override props, etc. Personally I don't know whether/how that would be possible at the moment as I've not looked into the existing combinations of props in enough detail. But were you saying that Frontend already has an abstraction for this? I'd be interested to check that out if you know where to look?

dotcom-rendering/src/web/components/Card/components/TrailTextWrapper.tsx
mxdvl: I’m a little bit puzzled by this change. This has been introduced in 2019 in [PR #1019](https://github.com/guardian/dotcom-rendering/pull/1019/commits/d5566e2953ba4dbcee290b1b168a3178b84c054f) and I would like to push this to a separate PR if at all possible?

mxdvl (approved):
This is looking good, but I’d prefer the change from `tablet` to `desktop` to be extracted in its own PR.

dotcom-rendering/src/web/components/FixedSmallSlowIII.tsx
```diff
@@ -0,0 +1,68 @@
...
+	showAge,
+}: Props) => {
+	const smallTrails = trails.slice(0, 3);
+
+	return (
+		<UL direction="row">
+			{smallTrails.map((trail, index) => {
+				return (
+					<LI
+						padSides={true}
```
OllysCoding, line 24: `key={trail.url}` - We must suppress all the warnings!

dotcom-rendering/src/web/components/Card/components/TrailTextWrapper.tsx
oliverlloyd: +1 for a separate PR for this. This could have impacts in a wider area than just fronts.

oliverlloyd (approved):
Chip, chip

dotcom-rendering/src/web/components/FixedSmallSlowIII.tsx
AshCorr: Thats right :) Not saying that you should be doing that in this PR, it would be quite a big task to do this.

bryophyta: Okay yep that sounds good, thanks!

dotcom-rendering/src/web/components/Card/components/TrailTextWrapper.tsx
bryophyta: Thanks, I've taken it out of this branch now. Do you think it makes sense to add an issue treating  the current behaviour of the trail text as a bug because it diverges from Frontend fronts? Or should we treat this as the baseline now given that it seems to follow an established convention on DCR?

PR #5168, opened 2022-06-10
Begone, react 'key' warnings 👋

OllysCoding:
Keys, keys, **_KEYS_**! React is always telling us to use keys ... 

Well no more! I've gone through and added keys everywhere I could find that react was complaining about in my terminal.

And the grand reveal.... drum role please 🥁🥁🥁

image
😍😍😍😍😍 I am at peace once again 

--- 

Actual technical notes because PRs can't be all fun and games 😢

In most if not all these places I don't believe adding keys actually impacts our site - since these are all server rendered, they're only rendered once and never re-rendered, therefor keys aren't strictly necessary ... but react will still warn you in console anyway.

I switched the `renderArticleElement` function to being a `RenderArticleElement` component so that I could add a key to it, but in all honesty there was no real reason for it not to be a component anyway so I think this is better, or at least more 'reacty'!

### DIFF
--- a/dotcom-rendering/src/web/components/LiveBlock.tsx
+++ b/dotcom-rendering/src/web/components/LiveBlock.tsx
@@ -1,6 +1,6 @@
 import { css } from '@emotion/react';
 import LiveBlockContainer from '@guardian/common-rendering/src/components/liveBlockContainer';
-import { renderArticleElement } from '../lib/renderElement';
+import { RenderArticleElement } from '../lib/renderElement';
 import { LastUpdated } from './LastUpdated';
 import { ShareIcons } from './ShareIcons';
 
@@ -63,22 +63,23 @@ export const LiveBlock = ({
 			host={host}
 			pageId={pageId}
 		>
-			{block.elements.map((element, index) =>
-				renderArticleElement({
-					format,
-					element,
-					isMainMedia: false,
-					host,
-					adTargeting,
-					ajaxUrl,
-					index,
-					pageId,
-					webTitle,
-					isAdFreeUser,
-					isSensitive,
-					switches,
-				}),
-			)}
+			{block.elements.map((element, index) => (
+				<RenderArticleElement
+					key={'elementId' in element ? element.elementId : index}
+					format={format}
+					element={element}
+					adTargeting={adTargeting}
+					ajaxUrl={ajaxUrl}
+					host={host}
+					index={index}
+					isMainMedia={false}
+					pageId={pageId}
+					webTitle={webTitle}
+					isAdFreeUser={isAdFreeUser}
+					isSensitive={isSensitive}
+					switches={switches}
+				/>
+			))}
 			<footer
 				css={css`
 					display: flex;
--- a/dotcom-rendering/src/web/components/LiveBlock.tsx
+++ b/dotcom-rendering/src/web/components/LiveBlock.tsx
@@ -65,7 +65,8 @@ export const LiveBlock = ({
 		>
 			{block.elements.map((element, index) => (
 				<RenderArticleElement
-					key={'elementId' in element ? element.elementId : index}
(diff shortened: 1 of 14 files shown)


dotcom-rendering/src/web/components/LiveBlock.tsx
```diff
@@ -63,22 +63,23 @@ export const LiveBlock = ({
...
-					pageId,
-					webTitle,
-					isAdFreeUser,
-					isSensitive,
-					switches,
-				}),
-			)}
+			{block.elements.map((element, index) => (
+				<RenderArticleElement
+					key={'elementId' in element ? element.elementId : index}
```
AshCorr, line 68: For simplicity is it worth just using `index` instead of relying on elementId at all?

AshCorr (approved):
Oh mah gawd!! Awesome!

dotcom-rendering/src/web/lib/withSignInGateSlot.tsx
```diff
@@ -34,7 +35,7 @@ export const withSignInGateSlot = ({
 }: Props): React.ReactNode => {
 	return renderedElements.map((element, i) => {
 		return (
-			<>
+			<React.Fragment key={element?.key}>
```
AshCorr, line 38: Same here regarding index

dotcom-rendering/src/web/components/SupportingContent.tsx
```diff
@@ -64,7 +65,8 @@ export const SupportingContent = ({ supportingContent, alignment }: Props) => {
 			{supportingContent.map((subLink: DCRSupportingContent, index) => {
 				// The model has this property as optional but it is very likely
 				// to exist
-				if (!subLink.headline) return <></>;
```
AshCorr, line 67: Would you still need an ID if we returned null here? I can't remember if this has caused us issues before

oliverlloyd (approved):
Thank you!!!

dotcom-rendering/src/web/components/LiveBlock.tsx
mxdvl: Or `element.elementId ?? index` ?

dotcom-rendering/src/web/lib/ArticleRenderer.tsx
```diff
@@ -86,21 +86,24 @@ export const ArticleRenderer: React.FC<{
...
-			pageId,
-			webTitle,
-			isAdFreeUser,
-			isSensitive,
-			switches,
-			abTests,
-		});
+		return (
+			<RenderArticleElement
+				key={'elementId' in element ? element.elementId : index}
```
mxdvl, line 91:
```suggestion
				key={element.elementId ?? index}
```

dotcom-rendering/src/web/lib/withSignInGateSlot.tsx
mxdvl:
```suggestion
			<React.Fragment key={element?.key ?? i}>
```

mxdvl (approved):
So soothing!

Some proposal for handling `undefined` keys with an index fallback, building on @AshCorr’s suggestion.

dotcom-rendering/src/web/components/LiveBlock.tsx
OllysCoding: Unfortunately elementId doesn't exist on the element type due to the way elements are typed in DCR, which is why the funky syntax is used!

I'm going to switch to this:
```jsx
// eslint-disable-next-line react/no-array-index-key -- This is only rendered once so we can safely use index to suppress the warning
```

Worth remembering that `index` is _not recommended_ for use as the key prop and the only reason its safe is because this is being server side rendered! We even have an eslint rule reminding us not to do it!

PR #5172, opened 2022-06-10
SWC compiler

arelra:
## What does this change?

Use `swc-loader` to compile/transpile instead of Babel.

SWC is a Rust based compiler: https://swc.rs/ 

SWC is used by:
- Next.js v12: https://nextjs.org/docs/advanced-features/compiler
- Deno: https://choubey.gitbook.io/internals-of-deno/architecture/core/tsc
- Parcel: https://parceljs.org/blog/beta3/

Implementation of the previous spike https://github.com/guardian/dotcom-rendering/pull/3820 and builds on the great work done in https://github.com/guardian/dotcom-rendering/issues/3884.

## Why?

To go faster

#### Dev

Screenshot 2022-06-10 at 15 18 21

Screenshot 2022-06-10 at 15 21 06

~x8 speedup

#### CI

Screenshot 2022-06-10 at 15 34 46

Screenshot 2022-06-10 at 15 34 37

~x3 speedup

#### PROD

Before

Screenshot 2022-06-14 at 17 38 01

After (with legacy build)

Screenshot 2022-06-14 at 17 27 50

After (without legacy build)

Screenshot 2022-06-14 at 17 37 35

~x4 speedup

#### Rebuild

After (i.e. it works)

Screenshot 2022-06-14 at 17 47 18

## Todo

- verify
    - build output
    - minification
    - sourcemaps
- run Cypress tests against more browsers
- legacy build?


jamesgorrie (commented):
Couple of notes on what's already looking pretty nifty:

- legacy might not be a thing as I _think_ that's there for IE11. @guardian/dotcom-platform - am I right/wrong about this? 
- Do we know this isn't affecting re-build time? I can't see why it would, but might be worth adding to your list of TODOs

dotcom-rendering/scripts/webpack/.swcrc.json
```diff
@@ -0,0 +1,21 @@
...
+    "minify": {
+      "compress": true
+    },
+    "parser": {
+      "syntax": "typescript",
+      "tsx": true,
+      "decorators": false,
+      "dynamicImport": true
+    },
+    "target": "es2020",
```
mxdvl, line 13: Is this the same as `@babel/preset-env`?

dotcom-rendering/scripts/webpack/webpack.config.browser.js
```diff
@@ -62,44 +62,10 @@ module.exports = ({ isLegacyJS, sessionId }) => ({
...
-						loader: 'ts-loader',
-						options: {
-							configFile: 'tsconfig.build.json',
-							transpileOnly: true,
-						},
-					},
-				],
+				use: {
+					loader: 'swc-loader',
+					options: swcConfig,
```
mxdvl, line 67: It would be nice to have a typed config, if at all possible! Some packages have a `defineConfig` method for this purpose.

mxdvl (commented):
Really exiting 🚤

oliverlloyd:
> * legacy might not be a thing as I _think_ that's there for IE11. @guardian/dotcom-platform - am I right/wrong about this?

I came here to say the same thing. I believe we no longer need the legacy build. Dropping that would give us another perf bump 😸

dotcom-rendering/scripts/webpack/.swcrc.json
arelra: Good question... it's related but not the same transform as `@babel/preset-env`.

`target` specifies the output in terms of ES specification i.e. es2020.

https://swc.rs/docs/configuration/compilation#jsctarget

`preset-env` describes target output in terms of supported browsers so you will get the lowest common denominator of ES syntax given your target browsers.

What's interesting is that DCR currently targets browsers that support `esmodules`:

https://github.com/guardian/dotcom-rendering/blob/1c0f164ea7d4e57a8f9e76850ee8ba2a55d5c1ea/dotcom-rendering/scripts/webpack/webpack.config.browser.js#L83-L89

SWC also supports `preset-env` browserslist style targets so I'll update it to use the `esmodules` equivalent.

https://swc.rs/docs/configuration/supported-browsers

However in the interest of not shipping unnecessary transforms (and hence reducing the bundle size) its worth reviewing if this is an appropriate target to carry forward.

arelra:
@jamesgorrie @oliverlloyd 

> * legacy might not be a thing as I _think_ that's there for IE11. @guardian/dotcom-platform - am I right/wrong about this?
> I came here to say the same thing. I believe we no longer need the legacy build. Dropping that would give us another perf bump

I didn't make it super clear but the initial times were actually for dev & CI which doesn't include legacy builds.

I've updated the PR description to include prod build times which shows the result of dropping the legacy build.

> * Do we know this isn't affecting re-build time? I can't see why it would, but might be worth adding to your list of TODOs

Rebuilding seems fine :)

I've added a screenshot to the PR description.

arelra:
> 💎 Great work! @arelra Do you know if there is a way we could have both loaders and enable only one based on a specific environment variable? Did not find [anything in webpack loaders configuration](https://webpack.js.org/concepts/loaders/#configuration).

Thanks @mchv :)

Yes the normal pattern is to check `process.env.NODE_ENV` and then conditionally pass the the appropriate configuration (including loaders) to webpack as required. For example:

https://github.com/guardian/dotcom-rendering/blob/d54ec25b49c5be77aec0743a7d76c6ad3bd8f507/dotcom-rendering/scripts/webpack/webpack.config.browser.js#L5

https://github.com/guardian/dotcom-rendering/blob/d54ec25b49c5be77aec0743a7d76c6ad3bd8f507/dotcom-rendering/scripts/webpack/webpack.config.browser.js#L49-L58

My preference however would be to keep the same compiler for dev and prod for consistency.

We would of course need to be confident that SWC works well in all our supported browsers and platforms.

I am wondering if we would be able to run something like an AB test but with the bundle output. I'll talk to dotcom about this.

mxdvl (changes requested):
Just wanted to make it clear that this should not be merged in before we settle on dropping the legacy bundle in #5206, and this is very much still at RFC stage because of the implactions for our readers on older browsers.

arelra:
@Jakeii's build times on an M1 Pro :fire:

Screenshot 2022-07-14 at 12 23 52 copy

arelra:
SWC could in theory support our existing browser targets as it supports [browserlists](https://swc.rs/docs/configuration/supported-browsers).

But as pre-requisites:
- removing legacy first would make the the testing required for SWC _much_ simpler.
- the introduction of an A/B testing framework for build output would allow us to roll out with confidence

oliverlloyd:
> @Jakeii's build times on an M1 Pro 🔥
> 
> Screenshot 2022-07-14 at 12 23 52 copy

_*Initial*_ 💥

arelra:
Closed in favour of https://github.com/guardian/dotcom-rendering/pull/6578

Issue #5134, opened 2022-06-07
Support `fixed/thrasher` container


bryophyta:
## JS Support

We need to decide how we enable JS for thrashers.

@HarryFischer has nicely put together [a list of thrashers](https://preview.gutools.co.uk/design/thrashers?dcr) where they're not working because of this missing support.

AshCorr:
Depends on #4847

PR #5182, opened 2022-06-13
Adds `fixed/thrasher` support to DCR Fronts

AshCorr:
## What does this change?

Depends on https://github.com/guardian/dotcom-rendering/pull/4847
Closes https://github.com/guardian/dotcom-rendering/issues/5134

Adds the `fixed/thrasher` container layout for rendering custom cards on DCR front pages.

### Why?

This makes it easy to see which thrashers are working on DCR and which ones aren't.

### Screenshots

<details>
<summary>Thrashers working really well on some pages like Sports</summary>

| Frontend | DCR |
| - | - |
| image | image |
</details>

<details>
<summary>Thrashers that aren't working as well</summary>

| Secure Drop: tries to modify the page styles which is no longer allowed by DCR | 
|:--:| 
| *Secure Drop: tries to modify the page styles which is no longer allowed by DCR* |

| Nuggets: we don't let thrashers use JS anymore and are planning on making Nuggets their own component in DCR | 
|:--:| 
| *Nuggets: we don't let thrashers use JS anymore and are planning on making Nuggets their own component in DCR* |

| Nuggets: we don't let thrashers use JS anymore and are planning on making Nuggets their own component in DCR | 
|:--:| 
| *Lots of issues with buttons in thrashers relying on Frontend CSS styles* |

</details>

### DIFF
--- a/dotcom-rendering/src/web/components/ContainerLayout.tsx
+++ b/dotcom-rendering/src/web/components/ContainerLayout.tsx
@@ -34,6 +34,7 @@ type Props = {
 	innerBackgroundColour?: string;
 	showDateHeader?: boolean;
 	editionId?: EditionId;
+	showLeftColumn?: boolean;
 };
 
 const containerStyles = css`
--- a/dotcom-rendering/src/web/components/ContainerLayout.tsx
+++ b/dotcom-rendering/src/web/components/ContainerLayout.tsx
@@ -34,7 +34,6 @@ type Props = {
 	innerBackgroundColour?: string;
 	showDateHeader?: boolean;
 	editionId?: EditionId;
-	showLeftColumn?: boolean;
 };
 
 const containerStyles = css`
--- a/dotcom-rendering/src/web/components/Snap.tsx
+++ b/dotcom-rendering/src/web/components/Snap.tsx
@@ -3,14 +3,15 @@ import { css } from '@emotion/react';
 const snapStyles = css`
 	overflow-y: hidden;
 	position: relative;
+	display: flex;
 `;
 
 type Props = {
-	snapData: DCRSnapType;
+	snapData?: DCRSnapType;
 };
 
 export const Snap = ({ snapData }: Props) => {
-	if (snapData.embedHtml === undefined) {
+	if (snapData?.embedHtml === undefined) {
 		return <></>;
 	}
 
@@ -19,10 +20,10 @@ export const Snap = ({ snapData }: Props) => {
 			css={[
 				snapStyles,
 				css`
-					${snapData?.embedCss}
+					${snapData.embedCss}
 				`,
 			]}
-			dangerouslySetInnerHTML={{ __html: snapData?.embedHtml }}
+			dangerouslySetInnerHTML={{ __html: snapData.embedHtml }}
 		/>
 	);
 };
--- a/dotcom-rendering/src/web/layouts/FrontLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/FrontLayout.tsx
@@ -13,6 +13,7 @@ import { Header } from '../components/Header';
 import { Island } from '../components/Island';
 import { MostViewedFooterLayout } from '../components/MostViewedFooterLayout';
 import { Nav } from '../components/Nav/Nav';
+import { Snap } from '../components/Snap';
 import { SubNav } from '../components/SubNav.importable';
 import { DecideContainer } from '../lib/DecideContainer';
 import { decidePalette } from '../lib/decidePalette';
@@ -126,6 +127,24 @@ export const FrontLayout = ({ front, NAV }: Props) => {
 					if (trails.length === 0) return null;
 
 					const ophanName = ophanComponentId(collection.displayName);
+					const ophanComponentLink = `container-${
+						index + 1
+					} | ${ophanName}`;
+
+					if (collection.collectionType === 'fixed/thrasher') {
+						return (
+							<ElementContainer
+								padded={false}
+								showTopBorder={false}
+								showSideBorders={false}
+								ophanComponentLink={ophanComponentLink}
+								ophanComponentName={ophanName}
+								element="section"
+							>
+								<Snap snapData={trails[0].snapData} />
+							</ElementContainer>
+						);
+					}
 
 					return (
 						<ContainerLayout
@@ -139,10 +158,8 @@ export const FrontLayout = ({ front, NAV }: Props) => {
 							centralBorder="partial"
 							url={collection.href}
 							// same as above re 'palette styles' for index increment
-							ophanComponentLink={`container-${
-								index + 1
-							} | ${ophanName}`}
-							ophanComponentName={`${ophanName}`}
+							ophanComponentLink={ophanComponentLink}
+							ophanComponentName={ophanName}
 							containerPalette={collection.containerPalette}
 							showDateHeader={collection.config.showDateHeader}
 							editionId={front.editionId}
--- a/dotcom-rendering/src/model/enhanceCards.ts
+++ b/dotcom-rendering/src/model/enhanceCards.ts
@@ -1,6 +1,7 @@
 import { ArticleDesign, ArticleSpecial } from '@guardian/libs';
 import { decideFormat } from '../web/lib/decideFormat';
 import { getDataLinkNameCard } from '../web/lib/getDataLinkName';
+import { enhanceSnaps } from './enhanceSnaps';
 
 /**
  *
@@ -115,6 +116,6 @@ export const enhanceCards = (
 			byline:
 				faciaCard.properties.maybeContent?.trail.byline ?? undefined,
 			showByline: faciaCard.properties.showByline,
-			snapData: faciaCard.enriched,
+			snapData: enhanceSnaps(faciaCard.enriched),
 		};
 	});
--- a/dotcom-rendering/src/model/enhanceSnaps.ts
+++ b/dotcom-rendering/src/model/enhanceSnaps.ts
@@ -1,0 +1,24 @@
+/**
+ *
+ * This function takes a FESnapType and turns it into a DCRSnapType mutating
+ * certain properties such as embedCss to work with Emotion.
+ *
+ * @returns the DCR snap with its properties mutated
+ */
+export const enhanceSnaps = (
+	snap: FESnapType | undefined,
+): DCRSnapType | undefined => {
+	const dcrSnap = snap;
+
+	// Emotion doesn't parse this conditional CSS correctly which is used by some thrashers to check
+	// if theres a pageskin. This regex fixes the vast majority of these cases should allow future snaps
+	// to be built with DCR in mind.
+	if (dcrSnap?.embedCss) {
+		dcrSnap.embedCss = dcrSnap.embedCss.replace(
+			/body:not\(\.has-active-pageskin\)(?! &)/g,
+			'body:not(.has-active-pageskin) &',
+		);
+	}
+
+	return dcrSnap;
+};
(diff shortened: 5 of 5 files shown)


dotcom-rendering/src/web/components/Snap.tsx
```diff
@@ -14,6 +15,15 @@ export const Snap = ({ snapData }: Props) => {
 		return <></>;
 	}
 
+	// Emotion doesn't parse this conditional CSS correctly which is used by some thrashers to check
+	// if theres a pageskin.
+	if (snapData.embedCss !== undefined) {
+		snapData.embedCss = snapData.embedCss.replaceAll(
+			'body:not(.has-active-pageskin)',
+			'body:not(.has-active-pageskin) &',
```
oliverlloyd, line 23: I presume this is a commercial classname? Do you think there's a way to break this coupling? It seems very brittle

dotcom-rendering/src/web/components/FixedThrasher.tsx
```diff
@@ -0,0 +1,45 @@
...
+	trails: TrailType[];
+	containerPalette?: DCRContainerPalette;
+	showAge?: boolean;
+};
+
+export const FixedThrasher = ({ trails, containerPalette, showAge }: Props) => {
+	const thrasher = trails[0];
+
+	return (
+		<Card
```
oliverlloyd, line 14: I find this relationship between snaps, cards and thrashers pretty hard to get my head around. I don't see that there's a lot of the `Card` code being used here... so... I wonder if there's a way to break this apart?

dotcom-rendering/src/web/components/ContainerLayout.tsx
```diff
@@ -32,6 +32,7 @@ type Props = {
 	ophanComponentLink?: string;
 	containerPalette?: DCRContainerPalette;
 	innerBackgroundColour?: string;
+	showLeftColumn?: boolean;
```
oliverlloyd, line 35: Rather than extend ContainerLayout to turn off features maybe we could give thashers a simpler layout?

A bit like 

```ts
					if (isThrasher) {
						return (
							<ElementContainer
								key={index}
								padded={false}
								showSideBorders={false}
							>
								<DecideContainer
									trails={trails}
									containerType="fixed/thrasher"
									containerPalette={
										collection.containerPalette
									}
								/>
							</ElementContainer>
						);
					} else {
						return (
							<ContainerLayout
```

Arguably thrashers have too many different use cases mixed into one and we will likely need to deal with edge cases in the thrasher code itself

dotcom-rendering/src/web/components/Snap.tsx
oliverlloyd: Ah, I think I see the use case now (sorry, it was clear before, I just didn't get it). 

So the problem here is that Thrashers are using a style of css that Emotion does not support. I think in this case it might be better to push back and make the request that this change happens at source, in the Thrasher code directly.

I think this is good for several reasons

- It educates those teams about the problem, making it and others like it less likely in future
- It is less code in DCR
- We want to encourage better links to the teams writing this code, this will help us improve Thrashers more
- and perhaps most importantly, it doesn't start a pattern where DCR mutates the css being passed to it to fix/enhance/add things. Those sorts of patterns have a habit of growing!

The downside to this is DCR will break when passed this code but I think that early failure is desirable, esp. now when we're not live and can get away with surfacing things like this using errors

OllysCoding: I would expect pageskins are the full screen takeovers we still occasionally see (though arguably we shouldn't) running on the site.

When this happens all the margins around the edge of the page are given to ad space, and the site feels like it's 'floating' in the middle of an ad - My guess is when this happens they probably don't want the protruding colours covering the ad!

If I find a screenshot I'll send it here for future reference

oliverlloyd (approved):
This is very cool.

Happy to thumbs up and leave it up to you decide how best to approach the comments, you know what you're doing

dotcom-rendering/src/web/components/Snap.tsx
AshCorr: > So the problem here is that Thrashers are using a style of css that Emotion does not support. I think in this case it might be better to push back and make the request that this change happens at source, in the Thrasher code directly.

I'm not quite sure its possible to fix it at source, atleast not while Frontend is still serving Fronts. Designers would have to add the `&` themselves which isn't valid CSS.

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -126,6 +127,28 @@ export const FrontLayout = ({ front, NAV }: Props) => {
...
+					} | ${ophanName}`;
+
+					if (collection.collectionType === 'fixed/thrasher') {
+						if (trails[0].snapData === undefined) {
+							return (
+								<>
+									First item of ${collection.displayName}
+									should be a SnapLink
+								</>
+							);
```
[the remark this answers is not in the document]
AshCorr: I spent a while thinking about this! My worry was that I think the front tool will let you place a normal card in a `fixed/thrasher` container instead of a Snap, I don't think we'd ever want to allow that but at the same time I don't want to return a 500 error because one container is misconfigured. It seemed better to fail gracefully this way, but happy to change if people disagree.

dotcom-rendering/src/web/components/ThrasherContainer.tsx
```diff
@@ -0,0 +1,27 @@
...
+	ophanComponentName,
+}: Props) => {
+	return (
+		<ElementContainer
+			padded={false}
+			showTopBorder={false}
+			showSideBorders={false}
+			ophanComponentLink={ophanComponentLink}
+			ophanComponentName={ophanComponentName}
+			element={'section'}
```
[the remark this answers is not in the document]
AshCorr: Thats odd, even the Github lint action hasn't picked it up

dotcom-rendering/src/web/components/Snap.tsx
```diff
@@ -14,15 +15,25 @@ export const Snap = ({ snapData }: Props) => {
...
 
+	// Emotion doesn't parse this conditional CSS correctly which is used by some thrashers to check
+	// if theres a pageskin. This regex fixes the vast majority of these cases should allow future snaps
+	// to be built with DCR in mind.
+	if (snapData.embedCss !== undefined && snapData.embedCss !== null) {
+		snapData.embedCss = snapData.embedCss.replace(
+			/body:not\(\.has-active-pageskin\)(?! &)/g,
+			'body:not(.has-active-pageskin) &',
+		);
+	}
```
mxdvl, line 26: I would like this to be extracted to its own method, so it's easier to refactor/remove in the future.

I'm puzzled as whether the `embedCss` can be missing. Here we're checking it is defined, but below we removed the optional chaining…

dotcom-rendering/src/web/layouts/FrontLayout.tsx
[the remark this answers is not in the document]
jamesgorrie: > My only concern is this might get left in and appear in production

This is a possibility in the current state, and surfacing debug data to readers is undesirable. [@mxdvl did some logging with unwrapHTML](https://github.com/guardian/dotcom-rendering/pull/4036/files). I'm not sure using that logger is the right place, but having any logs in Kibana would be useful.

Or maybe Sentry? Or though I think our Sentry is a little messy.

jamesgorrie (approved):
Looking good, just a couple of nits and questions.

dotcom-rendering/src/web/components/Snap.tsx
mxdvl: > **Note**
> There’s two kinds of page skins:
> 1. the good ones, in the UK, where we know at server-render time that they will be around
> 2. the not-so-good ones, where anything goes and are entirely client-side, mainly in Australia

dotcom-rendering/src/web/layouts/FrontLayout.tsx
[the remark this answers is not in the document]
mxdvl: For the logs to Kibana, rather than the logs endpoint, [it’s this commit](https://github.com/guardian/dotcom-rendering/commit/a8274881eb6a758a3732f283a9f0e3f0dcfa3f81)!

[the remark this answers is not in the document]
AshCorr: Kibana sounds promising, I'll take a look at doing that.

I think Sentry is just for client side errors unless I'm mistaken

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -126,6 +127,28 @@ export const FrontLayout = ({ front, NAV }: Props) => {
...
+						}
+
+						return (
```
oliverlloyd, line 144: I think we need a `key` somewhere here to keep React (and @OllysCoding ) happy!

dotcom-rendering/src/web/components/Snap.tsx
jamesgorrie: I'd argue it'd be good to keep here so it's in context, and isn't a bit of functionality that we would want to leak out into anywhere else in the code. 

How would making it a function make it easier to work with @mxdvl?

oliverlloyd: We typically make model changes in enhancers. It's hard to know how to classify what css mutations are (because I don't think I've ever seen one happen before) but if we decide it's the same as the other enhancements, and we decide to keep it  (I'm still iffy on if we want this pattern or not) then maybe want to keep all of these types of changes in one place, in the `enhanceCollections` step?

dotcom-rendering/src/web/components/Snap.tsx
oliverlloyd: Also, looking through some thrasher source code it seems clear that there are a number of expectations made around the html structure and what classnames might be available on the page. I can see code expecting `fc-container__inner`, `fc-container--thrasher`, etc. This code is not DCR safe.

This makes me think that we have two possible routes open to us:

1) We take the path that was taken for interactives on DCR, we amend the DCR code to try to accommodate the expectations from the html and css that we are trying to render on the page or
2) We ask the teams creating this content to edit their code to make it work on DCR

I think that our experience with option 1 to date tells us that this was not an ideal solution. Even so, in that case (interactive articles) we had such a long tail of content that this was the best option.

But there is a very limited number of thrashers so I think editing them is a viable option. I think this is the choice we should be making when we find errors with how they render on DCR.

AshCorr: > Do you have more information on what the original error was that meant we need to make this change? I wasn't able to replicate it locally by removing this code

3030 code ashleigh rocks_Front_url=https___www theguardian com_uk (3)
With ampersans added ^

3030 code ashleigh rocks_Front_url=https___www theguardian com_uk (2)
Without ampersans added ^

Without adding the ampersans Thrashers use the pageskin version of their styles and not the standard version.

> Also, looking through some thrasher source code it seems clear that there are a number of expectations made around the html structure and what classnames might be available on the page. I can see code expecting fc-container__inner, fc-container--thrasher, etc. This code is not DCR safe.

Theres no denying that there still is a lot of little visual bugs with a lot of thrashers from where they depend on styles that used to exist in Frontend

> If the extra ampersand isn't valid css I'm curious why DCR needs to add it to keep Emotion happy 🙃

Because its part of the emotion Syntax for doing nested selectors and not part of standard CSS syntax. https://emotion.sh/docs/nested 

> We ask the teams creating this content to edit their code to make it work on DCR

I agree with that, and thats probably how we'll go about resolving all of the other little issues with thrashers, but this one particular issue I don't think theres an easy way for us to fix them at the source

PR #5184, opened 2022-06-14
Add SlotBodyEnd (epic) to interactive layout

tomrf1:
Marketing want to display the RR epic below interactive pages.
This PR adds the `SlotBodyEnd` as an island.
Unlike with the standard article layout, it needs some extra styling to align with the `SubMeta` component. Otherwise it renders on the left of the page.

Screen Shot 2022-06-14 at 09 23 58
Screen Shot 2022-06-14 at 09 24 11
Screen Shot 2022-06-14 at 09 24 30
Screen Shot 2022-06-14 at 09 24 46
Screen Shot 2022-06-14 at 09 25 03


tjmw (approved):

dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
```diff
@@ -162,7 +193,11 @@ export const SlotBodyEnd = ({
 	}, [isSignedIn, countryCode, brazeMessages, asyncArticleCount]);
 
 	if (SelectedEpic) {
-		return <SelectedEpic />;
+		return (
+			<div css={wrapperStyle(contentType)}>
```
oliverlloyd, line 197: This wrapper code can be replaced with `ContainerLayout`, which I suspect does the same thing but in a centralised manner so everything will line up nicely 📐

dotcom-rendering/src/web/layouts/InteractiveLayout.tsx
```diff
@@ -531,6 +535,42 @@ export const InteractiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 										isDev={!!CAPIArticle.config.isDev}
 									/>
 
+									<Island clientOnly={true}>
```
oliverlloyd, line 538: I'm not completely sure of the flow here but I _think_ this is roughly where you want to have `ContainerLayout`. I can't be sure though as it depends on if the slot will always show or only sometimes

oliverlloyd (commented):
I'm not certain here but I think that we can make use of the `ContainerLayout` component to achieve the left margins you need here? The rest of the page/site is using this component so it offers consistency and ease of maintenance to do the same thing here.

dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
tomrf1: I've just tried ContainerLayout but it's too wide -
Screen Shot 2022-07-06 at 09 56 02

I tried playing with some props but it doesn't change anything, e.g.
```
<ContainerLayout
	stretchRight={false}
	leftColSize={'compact'}
>
	<SelectedEpic />
</ContainerLayout>
```

tomrf1: no effect

oliverlloyd: 🤔 I _feel_ like this should be possible but, equally, I might be wrong 🤷 

Could you ping me the steps required to test this? I'm happy to dig around a bit. I've got the branch running but I wasn't sure which url to use and how to ensure the slot shows? Also very happy to pair on this, if you like?

dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
```diff
@@ -59,12 +59,18 @@ const buildBrazeEpicConfig = (
 	brazeMessages: BrazeMessagesInterface,
 	countryCode: string,
 	idApiUrl: string,
+	contentType: string,
```
oliverlloyd, line 62: I can see that it is already in use in this component but it's worth keeping in mind that the preferred method for branching logic like this is to use `format`, properties like `contentType` are still in use and do still work but this could change in the future.

oliverlloyd (approved):
Looks good 👍

PR #5185, opened 2022-06-14
Introduce `CommercialLazyLoadMarginReloaded` test

simonbyford:
## What does this change?

Introduces an A/B test in DCR which can be used to measure the impact of this change: https://github.com/guardian/frontend/pull/25107

## Why?

Screenshot 2022-06-14 at 09 52 50


arelra (commented):
Do we also need to add to `CoreVitals.importable.tsx` ?

simonbyford:
> 

Yes - good shout

simonbyford:
> Do we also need to add to `CoreVitals.importable.tsx` ?


Yes - good shout

PR #5190, opened 2022-06-14
fix: Add a title to InteractiveBlockElement iframes

### LINKED ISSUE #4492: Issue ID: DAC_Frame_Without_Title4 - Feature: David Squires
Report link: https://drive.google.com/file/d/14MWXhnst-l-HAm__v20GUOWl6K9QbE-6/view

Link to full issue list: https://docs.google.com/spreadsheets/d/1vWQtZfD49uq7uNCNELuYN_HcIj-BKiE4t4Hwda3tJaQ/edit#gid=0

AshCorr:
## What does this change?

Sets a title on the iframes used by interactive block elements which is used by screen readers to announce the contents of an iframe.

Previously when focusing an iframe it would get announced as "frame", now it'll be announced as hopefully something a bit more relevant or fall back to "Interactive Content".

Not a perfect solution as the alt-text on some Interactives is ocasionally used as a "parameter" for data https://github.com/guardian/interactive-atom-picture-loop/blob/e2eeb5034ca9405a363b5b37537cf5a81112b23c/src/js/app.js#L30

## Why?

Fixes #4492

### DIFF
--- a/dotcom-rendering/src/web/components/InteractiveBlockComponent.importable.tsx
+++ b/dotcom-rendering/src/web/components/InteractiveBlockComponent.importable.tsx
@@ -252,6 +252,7 @@ export const InteractiveBlockComponent = ({
 			iframe.style.width = '100%';
 			iframe.style.border = 'none';
 			iframe.height = decideHeight(role).toString();
+			iframe.title = caption ?? alt ?? 'Interactive Content';
 			if (url.startsWith('http:')) {
 				iframe.src = url.replace('http:', 'https:');
 			} else {


mxdvl (approved):
Simple but effective

jlieb10:
> > Not seen the `??` ... `??` syntactic sugar before - looks cool
> 
> Nullish coalescing is only choosing the alternative if the value is `null` or `undefined`. This is usually what we want instead of `||` and there's an ESLint rule for it.

Yeah I assumed it was like `||` but with error handling

PR #5191, opened 2022-06-14
Apply default card styles to analysis pieces

silvacb:
## What does this change?
Specific styling for analysis pieces has been removed from cards in the "Related content" section at the bottom of articles, and default styling is applied.

## Why?
It has been decided that the specific styling of analysis pieces should be removed.

## Screenshots

**Before**
Screenshot 2022-06-14 at 15 18 11

**After**
Screenshot 2022-06-14 at 15 18 38


JamieB-gu (approved):
Nice 👍

PR #5192, opened 2022-06-14
ESLint: `explicit-module-boundaries`

mxdvl:
## What does this change?

Enforce explicit-module-boundaries for `*.ts` files, and suggest dropping it for Preact components in `*.tsx` files.

## Why?

Using explicit return types for React results in types that are often too wide, as expressed in Kent C Dodd’s article [How to write a React Component in TypeScript][1]:

> ```ts
> function Calculator({left, operator, right}: CalculatorProps) {
>   // implementation clipped for brevity
> }
> ```
>
> This doesn't have any of the shortcomings of React.FC and it's no more complicated than typing the arguments to a regular function.
> 
> Ok, so what about the return value? Well, we could type it as React.ReactElement or even wider as a JSX.Element. But honestly, I side with my friend Nick McCurdy when [he says](https://twitter.com/nickemccurdy/status/1365384372908621833) that mistakes can easily be made causing the return type to be too wide. So even outside a react context, I default to not specifying the return type (rely on inference) unless necessary. And that's the case here.

[1]: https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript


dotcom-rendering/src/lib/joinUrl.ts
```diff
@@ -1,16 +1,15 @@
-export const joinUrl = (parts: string[]) => {
+export const joinUrl = (parts: string[]): string => {
 	// Remove any leading or trailing slashes from all parts and then join cleanly on
 	// a single slash - prevents malformed urls
 	const trimmed = parts
 		.map((part) => {
 			// Trim left
-			if (part.substr(0, 1) === '/') return part.slice(1);
+			if (part.startsWith('/')) return part.slice(1);
```
oliverlloyd, line 7: 👍 nice

dotcom-rendering/src/lib/joinUrl.ts
```diff
@@ -1,16 +1,15 @@
...
 			return part;
 		})
 		.map((part) => {
 			// Trim right
-			if (part.substr(part.length - 1, 1) === '/')
-				return part.slice(0, -1);
+			if (part.endsWith('/')) return part.slice(0, -1);
```
oliverlloyd, line 12: Yep

dotcom-rendering/src/model/extract-nav.ts
```diff
@@ -80,8 +80,7 @@ const buildRRLinkModel = (data: any): ReaderRevenuePositions => ({
 	ampFooter: buildRRLinkCategories(data, 'ampFooter'),
 });
 
-// TODO refactor now that we know data must be CAPINavType.
```
oliverlloyd, line 83: Thanks

dotcom-rendering/src/server/dev-server.ts
```diff
@@ -9,10 +9,10 @@ import {
 	renderKeyEvents,
 } from '../web/server';
 
-// see https://www.npmjs.com/package/webpack-hot-server-Middleware
+// see https://www.npmjs.com/package/webpack-hot-server-middleware
```
oliverlloyd, line 12: Yep 👍 

I'm enjoying reviewing this PR

dotcom-rendering/src/web/lib/contributions.ts
```diff
@@ -127,7 +127,7 @@ const dateDiffDays = (from: number, to: number): number => {
 
 const AskPauseDays = 90;
 
-export const isRecentOneOffContributor = () => {
+export const isRecentOneOffContributor = (): boolean => {
```
oliverlloyd, line 130: This just makes sense to have this type here. You can infer it from the name but it's nice that it's super clear

oliverlloyd (approved):
Really nice. This was a very fun PR to review!

PR #5195, opened 2022-06-14
Remove 3rd party module declarations

mxdvl:
## What does this change?

Remove the following modules:
- `compose-function` it’s used in only one place
- `minify-css-string` it’s trivial to inline

Improved the typing of `dynamic-import-polyfill` by [reading the documentation](https://github.com/GoogleChromeLabs/dynamic-import-polyfill#configuration-options).

## Why?

There’s a plethora of `any`s in our declaration file, and these ones can easily be removed.


bryophyta (approved):
this looks good to me!

PR #5197, opened 2022-06-15
Lines in the footer

oliverlloyd:
## What does this change?
This refactors the `CardFooter`, reversing a [change made earlier](https://github.com/guardian/dotcom-rendering/pull/4883) to position the `StraightLines` component independently of the footer. We now place the lines in-between the footer meta content

Screenshot 2022-06-15 at 10 02 43

## Why?
We are fine tuning [the designs for cards](https://www.figma.com/file/5CfbWOeZPRi15lXBD5u1rW/Card-layout-system?node-id=0%3A1) and the new designs call for the meta to be inline

## Screenshots

| Before      | After      |
|-------------|------------|
| Screenshot 2022-06-15 at 09 59 40 | Screenshot 2022-06-15 at 09 57 59 |

### DIFF
--- a/dotcom-rendering/src/web/components/Card/Card.tsx
+++ b/dotcom-rendering/src/web/components/Card/Card.tsx
@@ -2,14 +2,13 @@ import { css } from '@emotion/react';
 import { ArticleDesign } from '@guardian/libs';
 import { brandAltBackground } from '@guardian/source-foundations';
 import { Link } from '@guardian/source-react-components';
-import { StraightLines } from '@guardian/source-react-components-development-kitchen';
-import { decidePalette } from '../../lib/decidePalette';
 import { getZIndex } from '../../lib/getZIndex';
 import { Avatar } from '../Avatar';
 import { CardHeadline } from '../CardHeadline';
 import { Flex } from '../Flex';
 import { Hide } from '../Hide';
 import { MediaMeta } from '../MediaMeta';
+import { Snap } from '../Snap';
 import { StarRating } from '../StarRating/StarRating';
 import { SupportingContent } from '../SupportingContent';
 import { AvatarContainer } from './components/AvatarContainer';
@@ -23,7 +22,6 @@ import { ContentWrapper } from './components/ContentWrapper';
 import { HeadlineWrapper } from './components/HeadlineWrapper';
 import { ImageWrapper } from './components/ImageWrapper';
 import { TrailTextWrapper } from './components/TrailTextWrapper';
-import { Snap } from '../Snap';
 
 export type Props = {
 	linkTo: string;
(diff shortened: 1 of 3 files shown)


bryophyta:
In the 'After' images for the Kerry Smith and John Crace articles, there is a gap between the bottom of the photo and the lines in the footer, which doesn't seem to be there in the 'Before' images, and to my mind at least looks a bit awkward.

image

Just from looking at it though, it looks like this is probably caused by the comment count and age being on a separate row from the straight lines. I'm not sure what the scope of the current issue is, so not sure whether this counts as being in scope or not.

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -139,16 +138,90 @@ export const Card = ({
...
-	const noOfSublinks = (supportingContent && supportingContent.length) || 0;
-
+	const noOfSublinks = supportingContent?.length ?? 0;
 	const isOpinion =
 		format.design === ArticleDesign.Comment ||
 		format.design === ArticleDesign.Editorial ||
 		format.design === ArticleDesign.Letter;
 
+	const cardIsHorizontal =
+		imagePosition === 'left' || imagePosition === 'right';
```
bryophyta, line 149: In theory we could have `imagePosition` set to `top`, while `imagePositionOnMobile` was set to `left` or `right`. If I've understood correctly then this card would not count as `cardIsHorizontal`. Is that the behaviour we want (i.e. only triggering `isHorizontal` above mobile)? Or does the decision logic need to take this case into account?

(actually, it looks like we have at least one actual case, although it's in code that I wrote last week so it's possible that I violated an assumption without realising? https://github.com/guardian/dotcom-rendering/blob/d54ec25b49c5be77aec0743a7d76c6ad3bd8f507/dotcom-rendering/src/web/components/FixedSmallSlowIII.tsx#L51-L52 )

bryophyta:
> > In the 'After' images for the Kerry Smith and John Crace articles, there is a gap between the bottom of the photo and the lines in the footer, which doesn't seem to be there in the 'Before' images, and to my mind at least looks a bit awkward.
> 
> Oh whoops, my before and after images were the wrong way around! Sorry about that.
> 
> For spacing, I do also have [this footer PR](https://github.com/guardian/dotcom-rendering/pull/5200) which should address any issues in this area.

Ah I see, that makes sense! I've made a small comment about a possible corner-case around the `isHorizontal` logic, but aside from that this looks good to me 👍

bryophyta (approved):

oliverlloyd:
Closing in favour of https://github.com/guardian/dotcom-rendering/pull/5224

PR #5198, opened 2022-06-15
RFC: Fronts container incorporating another container

### LINKED ISSUE #5133: "fixed/medium/slow-VI" (`266`)

bryophyta:
## What does this change?
Adds a container for `fixed/medium/slow-vi`. This will close #5133.

## The question
In this version I have used another container component (`fixed/small/slow-iv`) to render the second row of cards. Is this a pattern that we are happy with as a team?

## Pros
1. As far as I can tell, when these containers were originally designed they were thought of as being made up of rows of reusable 'slices'. So it's possible to re-use `fixed/small/slow-iv` here without editing it at all, and this seems to be in keeping with the original rationale.
2. If adopted as a more general pattern, it will hopefully make the relationships between the different container/slice types more explicit (it can be hard to tell at a glance what the list of hard-coded props in different rows are doing). And potentially be easier to maintain as there will be less duplicated code where changes will need to be made.

## Cons
1. Introduces coupling between the two components. Regardless of the original intentions behind the container system, this might not be desirable in the future. Hopefully Chromatic would show up any unintended changes in the medium component which were caused by a change in the small component, but nevertheless this coupling potentially violates the principle of least surprise.
2. I know there are some other ongoing discussions about a) whether it's actually desirable to reduce code repetition in these containers, all things considered, and b) how best to do it if we are going to try. Adding the component reuse pattern in this PR might make it harder to adopt a different abstraction later on.

## tl;dr
Overall, I'm very happy to just duplicate the code if people have reservations, but thought it might be worth discussing.

### DIFF
--- a/dotcom-rendering/src/web/components/FixedMediumSlowVI.stories.tsx
+++ b/dotcom-rendering/src/web/components/FixedMediumSlowVI.stories.tsx
@@ -1,0 +1,36 @@
+import { breakpoints } from '@guardian/source-foundations';
+import { trails } from '../../../fixtures/manual/trails';
+import { ContainerLayout } from './ContainerLayout';
+import { FixedMediumSlowVI } from './FixedMediumSlowVI';
+
+export default {
+	component: FixedMediumSlowVI,
+	title: 'Components/FixedMediumSlowVI',
+	parameters: {
+		chromatic: {
+			viewports: [
+				breakpoints.mobile,
+				breakpoints.mobileMedium,
+				breakpoints.mobileLandscape,
+				breakpoints.phablet,
+				breakpoints.tablet,
+				breakpoints.desktop,
+				breakpoints.leftCol,
+				breakpoints.wide,
+			],
+		},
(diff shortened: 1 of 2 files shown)


oliverlloyd (commented):
Thanks for putting the extra time and effort for thinking about the api here 💙 👍 . There's not really a right and wrong in these things. As you've laid out already, it's all about pros and cons and weighing up the choices. I think in thi case it's better to keep thing uncoupled but I do think we should revisit this in the future, maybe once we've got things in place and we can have another look at how it all looks together? It's a lot easier to reason about abstractions when the use cases are implemented already.

On the code, I think we're missing some properties for this container. From looking at the designs it seems there are three types of cards here, primary (the main one), secondary (the one next to the main one) and the remaining ones on the next row.

primary
Has the image positioned to the right (`imagePostition: 'right'`) and is larger (imageSize). It also has a larger image on mobile.

secondary
This has larger headline font

remaining
These are standard

PR #5204, opened 2022-06-16
fix(deno): explicit exit

mxdvl:
## What does this change?

Be explicit about exiting from this script.

## Why?

It seems that [sometimes the action times out after 6h](https://github.com/guardian/dotcom-rendering/runs/6895667299?check_suite_focus=true) and this might be because we’re not exiting.


AshCorr:
Related: #5159

It seems (based on my very minimal research) that this might be an issue with Octokit

PR #5206, opened 2022-06-16
refactor!: remove legacy JS bundle

mxdvl:
## What does this change?

We no longer serve our JavaScript bundle to browsers that do not support modules. Mainly IE11. There are a few exception with inline scripts used for accessibility improvements.

> **Warning**
> Our **Google Analytics** and **Ophan** scripts will no longer function in browsers that do not support ES Modules.

## Why?

We no longer officially support IE11. Let’s improve the experience for no JavaScript instead of outdated JavaScript.

See also https://github.com/guardian/configs/pull/207 for context on current browser usage.

cc. @guardian/client-side-infra

### DIFF
--- a/dotcom-rendering/docs/architecture/029-only-esmodules.md
+++ b/dotcom-rendering/docs/architecture/029-only-esmodules.md
@@ -1,0 +1,40 @@
+# Only serve JavaScript ES Modules
+
+## Context
+
+As of June 2022, [the Guardian’s website no longer supports IE11][recommended-browsers].
+
+[recommended-browsers]: https://www.theguardian.com/help/recommended-browsers
+[cut the mustard]: ../principles/browser-support.md#cutting-the-mustard
+
+DCR currently build two JavaScript bundles:
+
+1. for modern browsers, that [natively support ES modules](https://caniuse.com/es6-module)
+2. for legacy browsers, that don’t support modern features
+
+Resources are better spent improving the experience of readers with
+no JavaScript than readers with outdated JavaScript.
+
+## Decision
+
+Stop serving our legacy JavaScript bundle to readers whose browser do not [cut the mustard][].
+
+Only serve our modern JavaScript bundle to modern browsers that natively support ES Modules:
+
+-   Google Chrome 61+
+-   Mozilla Firefox 60+
+-   Microsoft Edge 16+
+-   Apple Safari 10.1+
+-   Apple iOS Safari 10.3+
+-   Opera 48+
+
+Still serve minimal JavaScript which is necessary to improve accessibility.
+
+By doing so, we can better focus on:
+
+-   optimising our bundle for modern browsers
+-   favour a `noscript` experience over a heavily polyfilled one
+
+## Status
+
+Approved
--- a/dotcom-rendering/scripts/webpack/webpack.config.browser.js
+++ b/dotcom-rendering/scripts/webpack/webpack.config.browser.js
@@ -1,24 +1,15 @@
-/* eslint-disable @typescript-eslint/no-var-requires */
 const { WebpackManifestPlugin } = require('webpack-manifest-plugin');
 const GuStatsReportPlugin = require('./plugins/gu-stats-report-plugin');
 
 const DEV = process.env.NODE_ENV === 'development';
 
-/**
- * @param {boolean} isLegacyJS
- * @returns {string}
- */
-const generateName = (isLegacyJS) => {
-	const legacyString = isLegacyJS ? '.legacy' : '';
-	const chunkhashString = DEV ? '' : '.[chunkhash]';
-	return `[name]${legacyString}${chunkhashString}.js`;
-};
+const outputFilename = DEV ? `[name].js` : `[name].[chunkhash].js`;
 
 /**
- * @param {{ isLegacyJS: boolean, sessionId: string }} options
+ * @param {{ sessionId: string }} options
  * @returns {import('webpack').Configuration}
  */
-module.exports = ({ isLegacyJS, sessionId }) => ({
+module.exports = ({ sessionId }) => ({
 	entry: {
 		sentryLoader: './src/web/browser/sentryLoader/init.ts',
 		bootCmp: './src/web/browser/bootCmp/init.ts',
@@ -34,8 +25,8 @@ module.exports = ({ isLegacyJS, sessionId }) => ({
 		initDiscussion: './src/web/browser/initDiscussion/init.ts',
 	},
 	output: {
-		filename: generateName(isLegacyJS),
-		chunkFilename: generateName(isLegacyJS),
+		filename: outputFilename,
+		chunkFilename: outputFilename,
 		publicPath: '',
 	},
 	// fix for known issue with webpack dynamic imports
@@ -44,12 +35,12 @@ module.exports = ({ isLegacyJS, sessionId }) => ({
 	},
 	plugins: [
 		new WebpackManifestPlugin({
-			fileName: isLegacyJS ? 'manifest.legacy.json' : 'manifest.json',
+			fileName: 'manifest.json',
 		}),
 		...(DEV
 			? [
 					new GuStatsReportPlugin({
-						buildName: isLegacyJS ? 'legacy-client' : 'client',
+						buildName: 'client',
 						project: 'dotcom-rendering',
 						team: 'dotcom',
 						sessionId,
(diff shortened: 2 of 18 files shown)


dotcom-rendering/scripts/webpack/webpack.config.browser.js
```diff
@@ -1,31 +1,27 @@
...
 
 const DEV = process.env.NODE_ENV === 'development';
 
 /**
- * @param {boolean} isLegacyJS
  * @returns {string}
  */
-const generateName = (isLegacyJS) => {
-	const legacyString = isLegacyJS ? '.legacy' : '';
+const generateName = () => {
```
jamesgorrie, line 9: Does it need to be a `function`?

```ts
const chunkhashString = DEV ? '' : '.[chunkhash]';
const outputFilename = `[name]${chunkhashString}.js`;
```

dotcom-rendering/src/web/server/frontToHtml.tsx
```diff
@@ -102,19 +83,13 @@ export const frontToHtml = ({ front, NAV }: Props): string => {
 	 * unlikely.
 	 */
 	const lowPriorityScriptTags = generateScriptTags([
-		...getScriptArrayFromFile('atomIframe.js'),
-		...getScriptArrayFromFile('embedIframe.js'),
-		...getScriptArrayFromFile('newsletterEmbedIframe.js'),
-		...getScriptArrayFromFile('relativeTime.js'),
+		...getScriptFromFile('atomIframe.js'),
```
jamesgorrie, line 86: These can't / shouldn't be `...spread`?

dotcom-rendering/docs/architecture/029-only-esmodules.md
```diff
@@ -0,0 +1,21 @@
...
+## Context
+
+As of June 2022, the Guardian no longer supports IE11.
+
+[cut the mustard]: ./asdf.md
+
+## Decision
+
+Chrome, Edge, Firefox, Opera, Safari all support ES Modules, so we can stop
+serving JavaScript for browsers that don’t [cut the mustard][].
```
ioannakok, line 12: Same as above

dotcom-rendering/src/web/server/articleTemplate.ts
```diff
@@ -196,13 +196,7 @@ https://workforus.theguardian.com/careers/product-engineering/
 
                 <script type="module">
                     window.guardian.mustardCut = true;
-                    window.guardian.gaPath = "${gaPath.modern}";
-                </script>
-
-                <script nomodule>
-                    // Browser fails mustard check
-                    window.guardian.mustardCut = false;
```
sndrs, line 204: is this ok to remove?

dotcom-rendering/src/web/browser/dynamicImport/init.ts
```diff
@@ -10,36 +10,21 @@ const initialiseDynamicImport = () => {
 		window.guardianPolyfilledImport = new Function(
 			'url',
 			`return import(url)`,
-		) as (url: string) => Promise<any>;
+		) as (url: string) => Promise<unknown>;
```
ioannakok, line 13: For my understanding, why are we doing this? Is it for safety?

mxdvl: By typing as `unknown`, you have to do actual type checking to make the compiler happy.

Probably best to keep out for now, it's not directly related.

dotcom-rendering/docs/architecture/029-only-esmodules.md
oliverlloyd: I think it's going to be helpful to flag this change to future devs as best as possible but just to say that these [decisions docs](https://adr.github.io/) are meant as historical records and - I think - they are not meant to be edited.

I've actually violated that principle myself before but after reading up on this maybe I was wrong to do this. As I understand it, the correct process is to create a new document to record the change in thinking

dotcom-rendering/src/lib/assets.ts
```diff
@@ -34,33 +31,15 @@ const decideAssetOrigin = (stage: string | undefined): string => {
...
-	if (legacyFilename) {
-		scripts.push({
-			src: `${ASSET_ORIGIN}assets/${legacyFilename}`,
-			legacy: true,
-		});
-	}
+	if (!filename) throw new Error(`Could not find filename: ${file}`);
 
-	return scripts;
+	return `${ASSET_ORIGIN}assets/${filename}`;
```
oliverlloyd, line 44: 😻 so much clearer!

dotcom-rendering/src/web/server/articleTemplate.ts
oliverlloyd: 👍 We have `window.guardian.mustardCut` set to true above so I think it's fair to say we should also keep the code that sets it to false.

I think the potential use case here is another team depending on this property. I'm not aware of this being the case but we can't be sure and I don't think we lose anything but having this property available

dotcom-rendering/src/web/server/frontTemplate.ts
```diff
@@ -154,13 +154,7 @@ https://workforus.theguardian.com/careers/product-engineering/
 
                 <script type="module">
                     window.guardian.mustardCut = true;
-                    window.guardian.gaPath = "${gaPath.modern}";
-                </script>
-
-                <script nomodule>
-                    // Browser fails mustard check
-                    window.guardian.mustardCut = false;
```
oliverlloyd, line 162: As per @sndrs 's comment above, we should keep this negative value to maintain the contract

dotcom-rendering/window.guardian.d.ts
```diff
@@ -16,7 +16,6 @@ declare global {
 	 */
 	interface Window {
 		guardian: {
-			mustardCut: boolean;
```
oliverlloyd, line 19: We're still setting this value so I think we need the type as well

oliverlloyd (approved):
I paired on this so it'd be good to get some other eyes here but I feel happy with the changes. Actually, more than happy, excited. Elated even 🎉

dotcom-rendering/src/web/browser/dynamicImport/init.ts
jamesgorrie: An extra note here - you can read and write anything to `any`, as it's defined as being able to be anything. It is the most loose type you can get. 

`unknown` is the tightest type you can get, as you can't run most read / write operations on it. e.g:

```typescript
const wat: any = undefined
console.log(wat.things) // valid, but errors on runtime
console.log(wat + 0) // valid, but errors on runtime

const wot: unknown = undefined
console.log(wot.things) // invalid, can't compile
console.log(wot + 0) // invalid, can't compile
```

jamesgorrie (approved):
Loving it - in [sentry](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/browser/sentryLoader/sentry.ts), should we exclude any errors that come from these browsers as we expect them?

sndrs:
> Loving it - in [sentry](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/browser/sentryLoader/sentry.ts), should we exclude any errors that come from these browsers as we expect them?

Would you expect errors? Should you not serve JS to browsers you know will error?

dotcom-rendering/docs/architecture/029-only-esmodules.md
```diff
@@ -0,0 +1,21 @@
+# Only serve JavaScript ES Modules
+
+## Context
+
+As of June 2022, the Guardian no longer supports IE11.
+
+[cut the mustard]: ./asdf.md
```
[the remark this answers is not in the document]
mxdvl: I don’t think we want to update the document.

dotcom-rendering/docs/architecture/029-only-esmodules.md
ioannakok: Thanks both for your explanation. Didn't realise we do not update these docs.

dotcom-rendering/src/web/browser/dynamicImport/init.ts
ioannakok: Makes sense. Thanks both for your explanations!

dotcom-rendering/docs/architecture/029-only-esmodules.md
mxdvl: I’ve reworked the document and split it in its own commit.

There’s a [live document currently circulating in the department](https://docs.google.com/document/d/1Fhu_g4Q9-a18jWftnGt1n6Scn8x88X-7jX2Ygr8LJrk/edit#), too.

dotcom-rendering/src/web/server/frontTemplate.ts
mxdvl: It’s back in!

dotcom-rendering/window.guardian.d.ts
mxdvl: Back in!

oliverlloyd:
Are there any more steps needed to get this one over the line? Can I help?

dotcom-rendering/scripts/webpack/webpack.config.browser.js
```diff
@@ -69,25 +60,15 @@ module.exports = ({ isLegacyJS, sessionId }) => ({
...
-													esmodules: true,
-												},
-											},
-									  ],
+								[
+									'@babel/preset-env',
+									{
+										bugfixes: true,
+										targets: {
+											esmodules: true,
```
sndrs, line 68: how you feel about using https://www.npmjs.com/package/@guardian/browserslist-config here?

PR #5208, opened 2022-06-16
Add topics to types

abeddow91:
## What does this change?
Adds topics data to the CAPI article model and regenerates schema. 
The corresponding frontend PR can be found [here](https://github.com/guardian/frontend/pull/25116) 

## Why?
Journalism Dev is adding automatic filters to liveblogs. This PR updates the CAPI article model to expect a new field containing an optional array of top mention filters

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -260,13 +260,13 @@ type CustomParams = {
 
 type AdTargeting =
 	| {
-			adUnit: string;
-			customParams: CustomParams;
-			disableAds?: false;
-	  }
+		adUnit: string;
+		customParams: CustomParams;
+		disableAds?: false;
+	}
 	| {
-			disableAds: true;
-	  };
+		disableAds: true;
+	};
 
 interface SectionNielsenAPI {
 	name: string;
@@ -604,6 +604,7 @@ interface CAPIArticleType {
 
 	// Included on live and dead blogs. Used when polling
 	mostRecentBlockId?: string;
+	topics?: Topic[];
 }
 
 type StageType = 'DEV' | 'CODE' | 'PROD';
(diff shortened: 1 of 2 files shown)


dotcom-rendering/index.d.ts
```diff
@@ -604,6 +604,7 @@ interface CAPIArticleType {
 
 	// Included on live and dead blogs. Used when polling
 	mostRecentBlockId?: string;
+	topMentionFilters?: TopMentionFilterType[];
```
oliverlloyd, line 607: Should we align our different names for this data to just `topics`?

From my limited understanding, this data is a list of topics that we are offering readers the ability to filter a blog by?

The main [Frontend PR](https://github.com/guardian/frontend/pull/25099) is naming the url param `topics`. So a url would look like

`https://theguardian.com/some/blog?topic=cats,dogs`

I think it would be more semantic to keep to this `topics` name throughput. It is clear and understandable and does not attach any implementation details to the field.

Beyond comprehension, I offer three other reasons for this change:

1) Maybe one day we use this `topics` data for a different reason, not just filtering a blog. By calling them what they are, not what they are used for, we make life easier for future us
2) Right now we decide what the topics should be based on frequency of mention (how often they exist in the text I presume) but maybe we use a different approach for this in future and we want this implementation detail out of the name
3) 'top' - we have a lot of other fields that list things but only the first n of them but these don't refer to this limit in their name

abeddow91: Totally agree and thanks for the detailed write up. I'll migrate this PR over to `topics`

paperboyo: This is very interesting and all good points. `filters` are more a device than a thing it operates on. `topics` are a reader-facing name for `tag/keyword`s. These here are coming from a new API altogether and do not map onto our keyword tags (topics) that live in Tag Manager and travel via Tools and CAPI. But calling them `topics` should be alright, I think.

> So a url would look like
> `https://theguardian.com/some/blog?topic=cats,dogs`

I just absolutely love seeing more URL manipulation devices! URL should be like a most verbose cli 👏

marjisound (approved):
Nice one Anna :)

PR #5219, opened 2022-06-17
RFC: Replace Storybook `storiesOf` with a generator script

AshCorr:
## What does this change?

Replaces all instances of `storiesOf` with a script to generate static CSF (Component story format https://storybook.js.org/docs/react/api/csf) for all format variations.

Paired with @sndrs & @jamesgorrie 

## Why?

`storiesOf` is technically part of Storybooks legacy API and is no longer recommended for use.

Theres 2 benefits from this:
- We might be able to save a lot of money on Chromatic screenshots by enabling Turbosnap. Turbosnap doesn't seem to like `storiesOf` very much.
- Storybook has a new version of Story Store but it doesn't support `storiesOf`. They claim that this new version of storystore can drastically speed up build times and memory efficiency. https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#story-store-v7

image

## Why the `stories` folder?

I felt like generated code should not be mixed with non-generated code, similar to fixtures I've created a seperate folder for it outside of `src`

## Exporting components from Stories?

I wanted to avoid having anything other than CSF in the generated story files, it felt nautral putting wrapper components in their non-generated counterparts, but does anyone have any other ideas?


dotcom-rendering/scripts/gen-stories/gen-stories.js
```diff
@@ -0,0 +1,171 @@
...
+const { writeFile, rm, readFile } = require('fs').promises;
+const path = require('path');
+const { log, success } = require('../env/log');
+
+const STORIES_PATH = path.resolve(__dirname, '..', '..', 'stories');
+const LAYOUT_STORIES_FILE = path.resolve(STORIES_PATH, 'Layout.stories.tsx');
+const CARD_STORIES_FILE = path.resolve(STORIES_PATH, 'Card.stories.tsx');
+
+const CARD_TEMPLATE_HEADER = `
+// File generated using dotcom-rendering/scripts/gen-stories/gen-stories.js
```
jamesgorrie, line 27: Worth leaving a note not to edit? [e.g.](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/fixtures/generated/images.ts#L1-L12)

dotcom-rendering/scripts/gen-stories/gen-stories.js
```diff
@@ -0,0 +1,171 @@
...
+							stories++;
+						}
+					}
+				}
+			}
+		}
+	}
+
+	await writeFile(LAYOUT_STORIES_FILE, template);
+	success(`[scripts/gen-stories] Generated ${stories} stories!`);
```
jamesgorrie, line 142: Worth mentioning it's from `Layouts`?

.storybook/main.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
 	stories: [
 		"../apps-rendering/src/**/*.stories.@(js|mdx|ts|tsx)",
 		"../dotcom-rendering/src/**/*.stories.@(tsx)",
+		"../dotcom-rendering/stories/**/*.stories.@(tsx)",
```
jamesgorrie, line 12: Should this live in a `generated` subfolder similar to fixtures? I know we don't have manual, but it feels consistent. Or a flatter `generated-stories`?

jamesgorrie (approved):
This looks great! Do we want to add a check similar to what @OllysCoding added for [gen-schema](https://github.com/guardian/dotcom-rendering/pull/4897/files#diff-2c51538768f1641842fc5522760d5794809a244ca98cf79ef371bdc239b4ce44) (next PR)?

.storybook/main.js
OllysCoding: I'd agree giving it a naming scheme that indicates it should only be for generated/automatic stories would be a good idea!

oliverlloyd (approved):
Thanks for doing this work, this opens up lots of exciting possibilities

dotcom-rendering/scripts/gen-stories/gen-stories.js
```diff
@@ -0,0 +1,171 @@
+/*
+
+Use: node dotcom-rendering/scripts/gen-stories/gen-stories.js
+
+This script was created as a replacement for storiesOf to generate all of the possible variants
+of our Card and Layout components.
+
+It should be run whenever a new Display, Design, or Theme
```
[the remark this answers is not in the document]
AshCorr: Woops, thanks for catching that, must have been in a bit of a heat daze when I wrote that!

dotcom-rendering/scripts/gen-stories/gen-stories.js
```diff
@@ -0,0 +1,171 @@
+/*
+
+Use: node dotcom-rendering/scripts/gen-stories/gen-stories.js
+
+This script was created as a replacement for storiesOf to generate all of the possible variants
+of our Card and Layout components.
+
+It should be run whenever any of the Display, Design, or Theme `format` properties change
```
sndrs, line 8: should it be run before the `chromatic ` script in the root `package.json`? that way you never need to remember to run it.

sndrs: so something like 

```json
	"chromatic": "node dotcom-rendering/scripts/gen-stories/gen-stories.js && chromatic --build-script-name=build-storybook --exit-zero-on-changes",
```

sndrs: or maybe better in `.storybook/main.js`, just before the config is exported

oliverlloyd: > it might be useful to run this script from .storybook/main.js, just before the config is exported

If this triggers this code to run prior to the build then that sounds ideal. I find commit hooks a bit flakey personally

dotcom-rendering/src/web/components/Card/Card.stories.tsx
```diff
@@ -92,8 +91,8 @@ const CardsWithDifferentThemes = ({
 			cards.push({
 				title: pillarName,
 				format: {
-					display,
-					design,
+					display: display,
```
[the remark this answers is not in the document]
AshCorr: Leftover from when I tried something else, shouldn't be needed anymore!

dotcom-rendering/scripts/gen-stories/gen-stories.js
sndrs: It should recreate the stories every time storybook builds, I *think*

AshCorr: I think I can do this, but I might have to make the script synchronous.

oliverlloyd:
Not sure if this is relevant or not but I think we're quite close to adding [Percy + Cypress](https://github.com/guardian/dotcom-rendering/pull/5256) which I believe takes away some of the use cases for this PR, at least in terms of the Layout stories.

Percy is able to make snapshots from pages rendered with Cypress so we are able to achieve the benefits of being in a real browser and then we only need to mock the page responses whilst looping over types.

This still leaves the Card stories though so I presume this PR still has value

dotcom-rendering/scripts/gen-stories/gen-stories.js
OllysCoding: > anyway we should just not check in the files

I think I'd say we should check them in - only for the purpose of when somebody makes a change to the gen-stories script, its visible in the git diff exactly how the output script has changed.

AshCorr:
> Not sure if this is relevant or not but I think we're quite close to adding [Percy + Cypress](https://github.com/guardian/dotcom-rendering/pull/5256) which I believe takes away some of the use cases for this PR, at least in terms of the Layout stories.

Sounds great, I'll keep the layout stories in as I'd like to get this merged so I can play around with Turbosnap but feel free to remove them in the Percy PR!

dotcom-rendering/src/web/components/Card/Card.stories.tsx
```diff
@@ -20,7 +19,7 @@ const basicCardProps: CardProps = {
 	},
 	headlineText: 'Headline text',
 	trailText:
-		'The 29-year-old source behind the biggest intelligence leak in the NSA’s history explains his motives',
+		'The 29-year-old source behind the biggest intelligence leak in the NSAs history explains his motives',
```
mxdvl, line 22: What’s this diff about?

mxdvl (approved):
Looks great!

dotcom-rendering/scripts/gen-stories/gen-stories.js
jamesgorrie: Probably for outside of this PR

Just thinking about the dev flow for this.
- [`libs`](https://github.com/guardian/libs) updates something in the [Format](https://github.com/guardian/libs/blob/main/src/format/README.md)
- update `dotcom-rendering/package.json` to new version of `libs`
- commit
- push

Would you have expected to run `storybook` because you updated `libs`?

This also means you need to remember to do that, or remember to run `make gen-stories`, which adds a little ambiguity as to when to use which method.

By adding a build step to the start script, we also have potential to lose idempotence across environments. e.g. I don't check in the new auto-generated stories files, when story book is run, it will run with the updated files. This feels quite confusing.

Not sure I have a suggestion just yet, but thought I'd point out where I find this current setup confusing.

PR #5200, opened 2022-06-15
Footer spacing

oliverlloyd:
## What does this change?
Following a [design review](https://www.figma.com/file/5CfbWOeZPRi15lXBD5u1rW/Card-layout-system?node-id=220%3A5279) this PR adjusts the spacing in the card footer

## Why?
We wanted

1) a space below the `Lines` and
2) better alignment for content

| Before      | After      |
|-------------|------------|
| b1 | a1 |
| b2 | a2 |
| b3 | Screenshot 2022-06-15 at 15 00 35 |

PR #5224, opened 2022-06-20
Card footer ❤️

### LINKED ISSUE #5104: Redesign Card Footer
Implement new card footer designs.

Figma file: https://www.figma.com/file/5CfbWOeZPRi15lXBD5u1rW/?node-id=0%3A1

oliverlloyd:
## What does this change?
This PR replaces #5200 and #5197 and fixes #5104 

Following a [design review](https://www.figma.com/file/5CfbWOeZPRi15lXBD5u1rW/Card-layout-system?node-id=220%3A5279) this PR adjusts position and spacing for the card footer

## Why?
There were some edge cases where alignment wasn't great and the meta data was appearing in a sub optimal location.

We now space and align the footer and its metadata better and have rationalised the decision logic for where the footer should be placed

### Opinion cards
Most of the changes were made for opinion cards

The lines that show at the bottom of Opinion cards now always show at the bottom. If the card has any sublinks then the footer metadata is shown _above_ the sublinks, immediately below the trail text.

| Before      | After      |
|-------------|------------|
| abb | aaa |
| bbb | baa |
| cbb | caa |

### Everything else
The footer (the card age if showing and any comment count) always appears below the trail text

| Before      | After      |
|-------------|------------|
| Screenshot 2022-06-20 at 12 24 40 | Screenshot 2022-06-20 at 12 27 17 |
|Screenshot 2022-06-20 at 12 30 47|Screenshot 2022-06-20 at 12 30 04|

### DIFF
--- a/dotcom-rendering/src/web/components/Card/Card.tsx
+++ b/dotcom-rendering/src/web/components/Card/Card.tsx
@@ -142,12 +142,82 @@ export const Card = ({
 	const palette = decidePalette(format, containerPalette);
 
 	const hasSublinks = supportingContent && supportingContent.length > 0;
-	const noOfSublinks = (supportingContent && supportingContent.length) || 0;
+	const noOfSublinks = supportingContent?.length ?? 0;
 
 	const isOpinion =
 		format.design === ArticleDesign.Comment ||
 		format.design === ArticleDesign.Editorial ||
 		format.design === ArticleDesign.Letter;
+
+	const renderFooter = ({
+		renderAge,
+		renderLines,
+	}: {
+		renderAge?: boolean;
+		renderLines?: boolean;
+	}) => {
+		return (
+			<CardFooter
+				format={format}
+				containerPalette={containerPalette}
+				renderLines={renderLines}
+				age={
+					webPublicationDate ? (
+						<CardAge
+							format={format}
+							containerPalette={containerPalette}
+							webPublicationDate={webPublicationDate}
+							showClock={showClock}
+						/>
+					) : undefined
+				}
+				mediaMeta={
+					(format.design === ArticleDesign.Gallery ||
+						format.design === ArticleDesign.Audio ||
+						format.design === ArticleDesign.Video) &&
+					mediaType ? (
+						<MediaMeta
+							containerPalette={containerPalette}
+							format={format}
+							mediaType={mediaType}
+							mediaDuration={mediaDuration}
+						/>
+					) : undefined
+				}
+				commentCount={
+					discussionId ? (
+						<Link
+							// This a tag is initially rendered empty. It gets populated later
+							// after a fetch call is made to get all the counts for each Card
+							// on the page with a discussion (see FetchCommentCounts.tsx)
+							data-name="comment-count-marker"
+							data-discussion-id={discussionId}
+							data-format={JSON.stringify(format)}
+							data-ignore="global-link-styling"
+							data-link-name="Comment count"
+							href={`${linkTo}#comments`}
+							subdued={true}
+							cssOverrides={css`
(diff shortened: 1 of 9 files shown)


dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -142,13 +142,83 @@ export const Card = ({
 	const palette = decidePalette(format, containerPalette);
 
 	const hasSublinks = supportingContent && supportingContent.length > 0;
-	const noOfSublinks = (supportingContent && supportingContent.length) || 0;
+	const noOfSublinks = supportingContent?.length ?? 0;
```
OllysCoding, line 145: +1

oliverlloyd:
> I find the logic of inner and outer footers in the rendering a little hard to follow, I wonder if moving the logic into more purposeful functions might make it more understandable when reading the code?
> 
> ```js
> const renderInnerFooter = () => {
>     const isOpinion = ...
>     if (isOpinion) return hasSubLinks ? renderFooter(...) : null
>     else return renderFooter(...)
> }
> 
> const renderOuterFooter = () => {
>     const isOpinion = ...
>     if (isOpinion) {
>         if (hasSubLinks) return <Lines ... />
>         else return renderFooter(...)
>     }
>     return null
> }
> 
> ...
> 
> {/* INSIDE FOOTER */}
> {renderInnerFooter()}
> 
> ...
> 
> {/* OUTSIDE FOOTER */}
> {renderOuterFooter())
> ```

Yeah, this looks like a sound improvement 👍

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -108,6 +108,63 @@ const decideIfAgeShouldShow = ({
...
+			// without the lines, these sit by themselves at the bottom
+			return renderFooter({
+				displayAge,
+				displayLines: false,
+			});
+		} else {
+			// Opinion cards without sublinks render the whole footer including lines, outside, sitting along the very bottom of the card
+			return null;
+		}
+	} else {
```
bryophyta, line 134: If I'm reading this correctly then two of the branches here are returning the same value, right? In which case presumably we could have instead:

```js
if (isOpinion && !hasSublinks) {
  return null;
}
return renderFooter({
  displayAge,
  displayLines: false,
});
```

Even though it avoids some nested conditionals, I'm slightly reluctant to suggest this because I think that the current structure gives a good way to spell out what the logic is doing in the comments. (I found the comments super helpful btw!) That said, if it's possible to keep the comments as illuminating with a simpler conditional structure, that could be nice.

oliverlloyd: Yeah, I kind of started here but then wrote the more verbose version so that I could spell it out a bit more.

It's super hard to know what's best in these situations. My brain has been full of this for a few days now so it's difficult for me to put myself in the shoes of someone coming to this code from scratch. Happy to take a steer from others on this

HarryFischer:
Hard to tell without seeing in a browser ... but on the whole I think this looks good! Thanks @oliverlloyd

dotcom-rendering/src/web/components/Card/Card.tsx
bryophyta: If we're not anticipating adding other conditions at the moment, then I think I might incline towards the simpler conditional. I think nested conditionals are generally harder to interpret, and it's also the sort of thing I can easily imagine someone deciding to refactor in passing, which would risk untethering the logic from the comments. So I don't have strong feelings on this but fwiw I'd be inclined to go with the simpler logic (although also definitely keeping the helpful comments that explain the rationale!)

bryophyta (approved):
This looks good to me 👍

PR #5225, opened 2022-06-20
Display `img` as block

mxdvl:
## What does this change?

Display `img` and `picture` as `block`, instead of the default `inline`.

## Why?

We were previously using a `vertical-align` property, which yield the same effect, but it’s more explicit to set the display type. The best explanation I found of how this think about this is in [Understanding Layout Algorithms by Josh W. Comeau](https://www.joshwcomeau.com/css/understanding-layout-algorithms/#inline-magic-space):

> Hmm… Why is there a bit of extra space underneath the image?
> […]
> By default, inline elements are “baseline” aligned. This means that the bottom of the image will align with the invisible horizontal line that text sits on. That's why there's some space below the image — that space is for the descenders, like the letters `j` and `p`.
> […]
> There are a number of ways to solve this problem. Perhaps the simplest is to treat this image as a block, within Flow layout:

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/174595563-d54091b2-0ae9-4b5e-9b8a-957479a33e0c.png
[after]: https://user-images.githubusercontent.com/76776/174594929-056b6a8e-96ed-4b23-b2bd-bf90c5d3ddaa.png

### DIFF
--- a/dotcom-rendering/src/web/components/ImageComponent.tsx
+++ b/dotcom-rendering/src/web/components/ImageComponent.tsx
@@ -288,6 +288,9 @@ export const ImageComponent = ({
 					img {
 						object-fit: cover;
 					}
+					picture {
+						height: 100%;
+					}
 				`}
 			>
 				<Picture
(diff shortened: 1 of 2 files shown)


AshCorr (approved):
Super!

PR #5228, opened 2022-06-20
Bypass metric sampling in AB test definition

zekehuntergreen:
## What does this change?
See [corresponding frontend change](https://github.com/guardian/frontend/pull/25121)


mxdvl (approved):
If it works, I think i this is a lot neater!

PR #5229, opened 2022-06-20
Migrate Image Code From `common-rendering` To AR

JamieB-gu:
## Why?

These image modules and components originally existed in [image rendering](https://github.com/guardian/image-rendering). The hope was for this library to power how images, backed by the Fastly image optimiser, would be rendered in AR, DCR and perhaps other projects too. When we started the initiative to share code via common-rendering, we migrated image-rendering across in #3391 and archived the original repo.

Although this code is currently used to render images in AR, two things have changed:

1. We no longer intend to share code via `common-rendering`
2. DCR has come up with a different approach to rendering images in #5128

Therefore this PR folds the image rendering code back into the AR part of this codebase. This should help to simplify it (e.g. we can make assumptions about the presence of a lightbox, which AR always has but DCR doesn't), and make it easier to reason about for new starters.

**Note:** This PR only performs the job of migrating the code back into AR; it doesn't simplify it yet. I plan to do that in a follow-up: removing the lightbox code, merging `Img` and `ImgAlt` etc. I've done it this way to minimise how many changes are being made in one go - hopefully this will reduce the chance of bugs and make the PRs easier to review!

## Changes

- Migrate `BodyImage` and `Img` components to AR
- Migrate image modules to AR: `image`, `srcset`, `sizes`, `lightbox`
- Migrate image fixtures to AR
- Temporarily name `Img` as `ImgAlt` to avoid conflicts
- Update test snapshots for `sizes`


iainjchambers-guardian (approved):
Looks great 👍

Issue #4848, opened 2022-05-06
Remove Wavy Lines From Feature


JamieB-gu:
We'd like to remove the wavy lines that appear as part of the metadata on feature pieces. We'd like to replace them with the straight lines that appear on standard articles. This change should be made for both AR and DCR.

See this for more information: https://github.com/guardian/dotcom-rendering/issues/4812#issuecomment-1119446374

## Examples

Examples of feature pieces can be found here: https://www.theguardian.com/tone/features

HarryFischer:
@ajbreuer There were some questions around this that I'm not sure about:
1. Does this effect both Features and Recipes? Are we replacing with flat lines?
2. Do we then lose the other lines variants across other Pillars? (dotted lines on Sport, 8 lines on Opinion, single line on Labs)

PR #5233, opened 2022-06-21
Remove squiggly lines & Refactor decide lines

OllysCoding:
## What does this change?

Removes squiggle lines from DCR as requested in https://github.com/guardian/dotcom-rendering/issues/4848

As part of this I refactored our `Lines` component logic as the use of `<Lines />` from source-react-kitchen is [deprecated](https://github.com/guardian/source/blob/main/packages/%40guardian/source-react-components-development-kitchen/src/lines/Lines.tsx#L35) - I think this solution ends up being a bit tidier in terms of logic as well :) 

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/9575458/174814033-5382d9f7-0ed6-456a-aa80-4074a9404586.png
[after]: https://user-images.githubusercontent.com/9575458/174814099-54ab9f4b-132d-4b6d-970a-8de48b51fa94.png


dotcom-rendering/src/web/components/DecideLines.tsx
```diff
@@ -0,0 +1,44 @@
...
+	useLabsLines?: boolean;
+};
+
+export const DecideLines = ({ format, color, useLabsLines }: Props) => {
+	const count = format.design === ArticleDesign.Comment ? 8 : 4;
+
+	if (useLabsLines && format.theme === ArticleSpecial.Labs) {
+		return <GuardianLabsLines />;
+	}
+
```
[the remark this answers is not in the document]
jamesgorrie: What is it you don't like about it? That they don't live in `@guardian/source-react-components-development-kitchen`?

dotcom-rendering/src/web/components/DecideLines.tsx
```diff
@@ -0,0 +1,44 @@
...
+} from '@guardian/source-react-components-development-kitchen';
+import { GuardianLabsLines } from './GuardianLabsLines';
+
+type Props = {
+	format: ArticleFormat;
+	color?: string;
+	useLabsLines?: boolean;
+};
+
+export const DecideLines = ({ format, color, useLabsLines }: Props) => {
```
jamesgorrie, line 15: This isn't necessary where it's used [in the ImmersiveLayout](https://github.com/guardian/dotcom-rendering/pull/5233/files#diff-10db6a9a97924d5e0c6e781cacae999645361d4aba55cace3f4bae9ce3d57666R328-R337) and [the InteractiveImmersiveLayout](https://github.com/guardian/dotcom-rendering/pull/5233/files#diff-10db6a9a97924d5e0c6e781cacae999645361d4aba55cace3f4bae9ce3d57666R328-R337) as there is a `format.theme === ArticleSpecial.Labs` guard in those files.

It's only [in the StandardLayout](https://github.com/guardian/dotcom-rendering/pull/5233/files#diff-c9298775646478b23d69b23a9e0b4c5e3b1a5daf67529f11a2b62317b516f662R586-R589) that we need this, creating two ways to get `GuardianLabsLines`, which is potentially confusing.

Could we potentially move all the conditions into `DecideLines`, and allow it to decide to render no lines (`<></>`, meaning it would also needs surrounding HTML<sup>*</sup>), or add the `format.theme === ArticleSpecial.Labs` guard to the `StandardLayout` and remove this override prop?


<sup>*</sup>surrounding HTML
```jsx
<div css={maxWidth}>
  <div css={stretchLines}>
  </div>
</div>

```

ioannakok (approved):
Looks good to me! Nice refactoring 👏

jamesgorrie (commented):
Looks good, just a question about the conditionals in and around `GuardianLabsLines`.

(None of the comments are blocking!)

dotcom-rendering/src/web/components/DecideLines.tsx
OllysCoding: I shall re-investigate this logic! I want to be careful not to over-abstract and turn it into an overly complex component!

OllysCoding: > creating two ways to get GuardianLabsLines, which is potentially confusing.

I'm a tad confused what you mean by this! Are you able to elaborate?

OllysCoding: After going over this with @jamesgorrie we decided that the cleanest implementation kept the `<GuardianLabsLines />` component in the layout files

jamesgorrie (approved):
Nice and tidy, that's easier to understand.

Just a Q that I think I may have asked in person, but why do some layouts not care about `GuardianLabsLines`? Is it just a design choice?

dotcom-rendering/src/web/layouts/InteractiveImmersiveLayout.tsx
```diff
@@ -406,19 +405,9 @@ export const InteractiveImmersiveLayout = ({
...
-													ArticleDesign.Standard,
-													format.theme,
-												)}
-												count={decideLineCount(
-													ArticleDesign.Standard,
-												)}
-											/>
+											<DecideLines format={format} />
 										)}
+										ss
```
mxdvl, line 410: 🤫

dotcom-rendering/src/web/components/DecideLines.tsx
```diff
@@ -0,0 +1,38 @@
...
+					count={count}
+					color={color}
+				/>
+			);
+		default:
+			return (
+				<StraightLines
+					cssOverrides={css`
+						display: block;
+					`}
```
mxdvl, line 32: I believe SVG elements are block by default, so this may be redundant? The way we display these has changed slightly with the new Line components.

mxdvl (commented):
Having one last look as the wiggle departs… 😢 

〰️〰️〰️〰️〰️

dotcom-rendering/src/web/layouts/InteractiveImmersiveLayout.tsx
OllysCoding: On its way out ... my `cmd` key has decided that today it no longer wants to work ...

dotcom-rendering/src/web/components/DecideLines.tsx
OllysCoding: As a test I tried removing this but it did end up having a visual impact, from looking in the chrome styles no `display` property was set by default and adding/removing it affected the spacing around it!

PR #5238, opened 2022-06-21
Dblatcher/email embed js

dblatcher:
## What does this change?
Adds a currently unused `EmailSignup` component (and `SecureSignup` and `SecureSignupIframe` sub-components) applying a new design. 

Based on work started in https://github.com/guardian/dotcom-rendering/tree/oliver/email-embed .

## Why?
The plan is to replace replace the existing email embeds, which are manually inserted into articles using embed blocks and rendered as iframes containing a page served by frontend with data-driven elements that platforms can render without relying on frontend. 

The structure of the Components is to address the following requirements:
 - The email input must be within an iframe to protect it from scripts run by third party ads
   - `SecureSignup` uses `renderToString` and `createEmotionServer` to generate strings of html and css  from server-side rendered source components that can be used to generate an iframe `srcDoc` property.
   - The html and css strings are passed as props to the importable `SecureSignupIframe` within an `Island` since the interactions with the rendered iframe requires client-side code.
 - The form need to have a reCaptcha to mitigate bot attacks, and use the 'invisible' implementation (no captcha element is displayed unless a challenge is invoked)
 - The `EmailSignup` (including the content rendered in the `SecureSignupIframe`) must be responsive
   - `SecureSignupIframe` uses a `resize` event listener to set the iframe height based on the space required for the content.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |


apps-rendering/package-lock.json
```diff
@@ -5915,7 +5915,7 @@
         "find-up": {
           "version": "2.1.0",
           "resolved": "https://registry.npmjs.org/find-up/-/find-up-2.1.0.tgz",
-          "integrity": "sha512-NWzkk0jSJtTt08+FBFMvXoeZnOJD+jTtsRmBYbAIzJdX6l7dLgR7CTubCM5/eDdPUBvLCeVasP1brfVR/9/EZQ==",
+          "integrity": "sha1-RdG35QbHF93UgndaK3eSCjwMV6c=",
```
JamieB-gu, line 5918: Are the changes to `package-lock.json` here needed? They appear to be unrelated to this change?

JamieB-gu: Nice 👍, thanks!

oliverlloyd:
Hi @dblatcher I've not been able to fully review this yet but just some early feedback.

This PR is great! 🙌  - The description is super clear and useful to read and the code is really well laid out 🏅 

One question. I can imagine there are benefits with using the 3rd party npm package for adding the recaptcha but this also comes with down sides. Would you be able to add a section in the PR description talking about this?

dblatcher:
> Hi @dblatcher I've not been able to fully review this yet but just some early feedback.
> 
> This PR is great! 🙌 - The description is super clear and useful to read and the code is really well laid out 🏅
> 
> One question. I can imagine there are benefits with using the 3rd party npm package for adding the recaptcha but this also comes with down sides. Would you be able to add a section in the PR description talking about this?

Thanks - will do!

OllysCoding:
> It seems to be incompatible with storybook (this may have a mistake I made, but stories for the component wouldn't compile)

I don't think you've made a mistake here, neither `renderToString` and `createEmotionServer` are likely to be 'client safe' and will rely on internal node features - storybook builds a SPA which runs in the browser, so I'd expect that this code would not work in this environment.

It's likely this component could be an early one to benefit from some of the work thats being done to add the cypress + percy integration (e.g https://github.com/guardian/dotcom-rendering/pull/5256) as cypress does the 'regular' server build and rendering process rather than compressing components into a SPA.

OllysCoding:
> but this would mean committing blocks of stringified, compiled (and largely unmaintainable) frontend code into src.

I think for this PR, it makes sense to stick with rendering these per-request, but for future I'd be interested if we can use some form of build time generation which could replace the need to do this - as we start to see PRs like https://github.com/guardian/dotcom-rendering/pull/5219 come into play, it'd be interesting to see how viable a pattern like this could be!

dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
```diff
@@ -0,0 +1,368 @@
...
+		const button = iframe?.contentDocument?.querySelector('button');
+		button?.addEventListener('click', handleClickInIFrame);
+		form?.addEventListener('submit', handleSubmitInIFrame);
+		resetIframeHeight();
+	};
+
+	// read siteKey
+	useEffect(() => {
+		setCaptchaSiteKey(
+			window.guardian.config.page.googleRecaptchaSiteKey ?? '',
```
OllysCoding, line 279: Does this need to be in a useeffect, or could you so something like 

`const captchaSiteKey = typeof window !== 'undefined' ?? window.guardian.config.page.googleRecaptchaSiteKey`

OllysCoding:
Posting this for visibility, but (and unfortunately gha does a bad job at surfacing this) it seems that this island adds about ~8kb of javascript to the page (im not sure if this is a gzipped size or not):

image

I think this is absolutely fine for the funcionality it adds, I'm not sure how it compares to the size of content downloaded in the existing iframe (not sure if you know this? @dblatcher) - and of course this is only for pages with the email signup component on it!

dotcom-rendering/src/web/components/EmailSignup.tsx
```diff
@@ -0,0 +1,132 @@
...
+			<StackBelowTablet>
+				<Title name={name} />
+				<Frequency frequency={frequency} />
+			</StackBelowTablet>
+			<Description description={description} />
+			<SecureSignup
+				newsletterId={newsletterId}
+				successText={successText}
+			/>
+		</Container>
```
OllysCoding, line 130: I think this probably comes entirely down to a matter of personal opinion, but a lot of these components created in this file are only used once, and I wonder if it would be more readable to keep all the html in one place; then extract the CSS from being inline to declared in variables, e.g

```jsx
const containerStyles = css`
	border: black 3px dashed;
	border-radius: 12px;
	padding: 10px;
	margin-bottom: ${space[3]}px;
`

const stackBelowTabletStyles = ...

const titleStyles = ...

return (
    <div css={containerStyles}>
		<div css={stackBelowTabletStyles}>
			<div css={titleStyles}>{title}</div>
             ....
		</div>
		...
	</div>
)

```

I think this is more fitting on the 'style' of DCR (e.g [CricketScoreboard.tsx](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/CricketScoreboard.tsx)) - but its definitely subjective!

dotcom-rendering/src/web/components/EmailSignup.tsx
```diff
@@ -0,0 +1,132 @@
...
+			`}
+		>
+			{children}
+		</div>
+	);
+};
+
+const Title = ({ name }: { name: string }) => {
+	return (
+		<div
```
OllysCoding, line 38: Could this be a hX or p tag?

dotcom-rendering/src/web/components/EmailSignup.tsx
```diff
@@ -0,0 +1,132 @@
...
+			>
+				{frequency}
+			</div>
+		</div>
+	);
+};
+
+const Description = ({ description }: { description: string }) => {
+	return (
+		<div
```
OllysCoding, line 86: Could this be a p tag?

dotcom-rendering/src/model/window-guardian.ts
```diff
@@ -17,6 +17,7 @@ export interface WindowGuardianConfig {
 		showRelatedContent: boolean;
 		ajaxUrl: string;
 		shouldHideReaderRevenue: boolean;
+		googleRecaptchaSiteKey?: string;
```
OllysCoding, line 20: Where is this getting set?

dotcom-rendering/src/web/components/SecureSignup.tsx
```diff
@@ -0,0 +1,163 @@
...
+const generateForm = (
+	newsletterId: string,
+): { html: string; styles: string } => {
+	const cache = createCache({ key: 'email-signup-iframe' });
+
+	// eslint-disable-next-line @typescript-eslint/unbound-method -- not a react method, 'this' binding not required?
+	const { extractCriticalToChunks, constructStyleTagsFromChunks } =
+		createEmotionServer(cache);
+
+	const html = renderToString(
```
OllysCoding, line 84: It'd be great to have a few comments here briefly explaining what's happening here and why, perhaps with a link to this PR to provide more information if needed!

dotcom-rendering/src/web/components/SecureSignup.tsx
```diff
@@ -0,0 +1,163 @@
...
+};
+
+export const SecureSignup = ({ newsletterId, successText }: Props) => {
+	const { html, styles } = generateForm(newsletterId);
+
+	return (
+		<>
+			<Island
+				clientOnly={true}
+				deferUntil={'idle'}
```
OllysCoding, line 143: Are we able to use 'visible' here?

dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
```diff
@@ -0,0 +1,368 @@
...
+// https://developers.google.com/recaptcha/docs/faq#id-like-to-hide-the-recaptcha-badge.-what-is-allowed
+
+type Props = {
+	styles: string;
+	html: string;
+	newsletterId: string;
+	successText: string;
+};
+
+const ErrorMessageWithAdvice = (props: { text?: string }) => (
```
OllysCoding, line 35: Does this work? `({ text }: { text?: string})` then `... {text} Please try again or contact ...`

OllysCoding (commented):
This looks really good! Sorry for the many, many comments, hopefully some will prove useful! 

I see there's still a few TODO's in place, are you looking to resolve these outside of this PR?

dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
dblatcher: You're right, this doesn't need to be in useEffect. 
I see a lot of other importable components set an `isServer` variable using `typeof window` so I'll adapt that pattern.

dblatcher:
> I think this is absolutely fine for the functionality it adds, I'm not sure how it compares to the size of content downloaded in the existing iframe (not sure if you know this? @dblatcher) - and of course this is only for pages with the email signup component on it!

The document loaded by the iframe (eg https://www.theguardian.com/email/form/plain/morning-mail ) is about 16kb, so this should be a net positive

dotcom-rendering/src/model/window-guardian.ts
dblatcher: I think it's this module in frontend that sets gets the values from configuration so they can be added to `window.guardian` (and used elsewhere) but I'm not clear on the mechanism. The actual value of the site key is ultimately generated by google. https://github.com/guardian/frontend/blob/bb3a156ac26b38da78cec3a2230b12d5a2095215/common/app/common/configuration.scala

It's already being put on the `window.guardian` article pages - I may have this wrong, but I assume it just handled been included on the interface yet as no components required it before.

dotcom-rendering/src/web/components/SecureSignup.tsx
dblatcher: Will do - it's potentially a function that could be generalised for use elsewhere, so good to outline what it does.

dotcom-rendering/src/web/components/EmailSignup.tsx
dblatcher: good point - I'd better use a p tag - having a hx might mess up the heading hierarchy and make the page less parseable on a screen reader

dotcom-rendering/src/web/components/EmailSignup.tsx
dblatcher: Yes, that works nicer here. I think it would make sense to extract the 'Frequency' component to its own file (its a recurring design element that will be used in other newsletter-y components) and use CSS variables for the rest.

dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
dblatcher: yes - will change

dotcom-rendering/src/web/components/SecureSignup.tsx
OllysCoding: I think given the results suggested in this PR: https://github.com/guardian/dotcom-rendering/pull/5276 it would be better to keep this as on idle as you have it!

OllysCoding (approved):
+1 I'm really excited to see how this new pattern grows in DCR :) What an improvement over a regular embed

dotcom-rendering/src/web/components/SecureSignup.tsx
```diff
@@ -0,0 +1,179 @@
...
+			<strong>Privacy Notice: </strong>
+			Newsletters may contain info about charities, online ads, and
+			content funded by outside parties. For more information see our{' '}
+			<Link
+				href="https://www.theguardian.com/help/privacy-policy"
+				rel="noopener noreferrer"
+			>
+				privacy policy
+			</Link>
+			.&nbsp;
```
mxdvl, line 50: Why the non-breaking space?

dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
```diff
@@ -0,0 +1,373 @@
...
+		requestBodyStrings.push(
+			`${encodeURIComponent(key)}=${encodeURIComponent(
+				value.toString(),
+			)}`,
+		);
+	});
+
+	return fetch(endpoint, {
+		method: 'POST',
+		body: requestBodyStrings.join('&'),
```
mxdvl, line 95: Can't you just pass the `formData` as-is? https://developer.mozilla.org/en-US/docs/Web/API/FormData

You might have to set the encoding type to `multipart/form-data`.

If not, maybe using a `map` instead of pushing fields to an array would keep things more functional.

PR #5240, opened 2022-06-21
Move accordion behind auto filter feature switch

silvacb:
## What does this change?
Moves the Live Feed header accordion behind the auto filter AB test. 
**This test is currently switched off and set to 0%.**

## Why?
As part of the automatic filters redesign, the header accordion is being removed.
NB: When removing the accordion, we lose horizontal padding. In the interest of keeping the PR small, this has been ticketed here (https://trello.com/c/XL3ulWuM/414-auto-filters-fix-padding-in-live-feed) and will be addressed in an upcoming PR.

## Screenshots

**Before:** 
Screenshot 2022-06-22 at 11 34 43

**After:** 
Screenshot 2022-06-22 at 11 34 58

### DIFF
--- a/dotcom-rendering/src/web/layouts/LiveLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/LiveLayout.tsx
@@ -301,6 +301,9 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 	const showKeyEventsCarousel =
 		CAPIArticle.config.abTests.keyEventsCarouselVariant == 'variant';
 
+	const isInFilteringBeta =
+		CAPIArticle.config.abTests.automaticFiltersVariant == 'variant';
+
 	return (
 		<>
 			<div data-print-layout="hide">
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -301,6 +301,9 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 	const showKeyEventsCarousel =
 		CAPIArticle.config.abTests.keyEventsCarouselVariant == 'variant';
 
+	const isInFilteringBeta =
+		CAPIArticle.config.abTests.automaticFiltersVariant == 'variant';
```
marjisound, line 305: Maybe this needs testing, but I think we could have scenarios where `topics` is not provided from frontend meaning we don't want to add filter for this blog, but the `automaticFiltersVariant` could still be `variant`

silvacb: Thanks for the suggestion Marjan, this has been implemented now by adding an additional check for topics.

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -301,6 +301,10 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 	const showKeyEventsCarousel =
 		CAPIArticle.config.abTests.keyEventsCarouselVariant == 'variant';
 
+	const isInFilteringBeta =
+		CAPIArticle.config.abTests.automaticFiltersVariant == 'variant' &&
```
joecowton1, line 305:
```suggestion
		CAPIArticle.config.switches.automaticFilters &&
```

marjisound (approved):
Great work Silvija 👍

PR #5242, opened 2022-06-21
Remove all usages of `getThemeStyles` from Apps Rendering (excluding Editions)

MarSavar:
## Why?
`getThemeStyles` is deprecated. `editorialPalette` should be used instead.

This PR does not touch editions code - usages of `getThemeStyles` in Editions components will be removed in a future PR.

## Screenshots

**NOTE** The only difference is a bug fix I spotted in related byline cards that were all defaulting to orange regardless of the theme. Everything else should be the same (verified with Chromatic)

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/57295823/175302743-ede22bb7-1b9f-4663-8027-64acb118684c.png
[after]: https://user-images.githubusercontent.com/57295823/175302695-8100f48c-8d16-479d-9200-9156cdb24556.png
[before2]: https://user-images.githubusercontent.com/57295823/175303002-6b6a44a0-1c4b-4ff8-9847-c9027873d874.png
[after2]: https://user-images.githubusercontent.com/57295823/175303065-2f66802f-d015-443b-9e92-a7c2bb92c3bc.png


common-rendering/src/editorialPalette/background.ts
```diff
@@ -297,6 +373,107 @@ const headlineTag = (format: ArticleFormat): Colour => {
...
+				case ArticlePillar.Opinion:
+					return opinion[200];
+				case ArticleSpecial.SpecialReport:
+					return specialReport[200];
+			}
+		case ArticleDesign.Audio:
+		case ArticleDesign.Video:
+		case ArticleDesign.Gallery:
+			return neutral[10];
+	}
```
JamieB-gu, line 402: Can I recommend adding a `default` to any inner nested `switch` statements in this file (I'd recommend using news as the default)? The `switch` statement in JS can cause some subtle bugs due to its fall-through behaviour.

I'll try to demonstrate what I mean by using the above code as an example. Within the `LiveBlog` case there's currently no guarantee that every `theme` has been handled. At the moment we can figure out by inspection that they have, because six of the seven are handled here and labs is handled in the `if` statement above. But what if someone deletes that `if` statement or we add an extra type to `theme`? Your expectation might be that if none of the themes match then both switches will be broken out of and we'll hit `return neutral[97]` at the bottom of the function. But what actually happens is that there's no early `return` in the `LiveBlog` case, so JS _falls through_ to the cases below _without checking_ that they match... and we hit `return neutral[10]`.

More details about this behaviour [on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#what_happens_if_i_forgot_a_break). The implication there is that a `break` is the way to solve this, but I'd recommend a `default` return here because that's more consistent with the pattern-matching-style approach we take to `switch` statements in this codebase.

If this code were in the AR part of the codebase I think the linter would have caught this 🤔. @jamesgorrie possible motivation here for splitting up common-rendering: the tooling here isn't as robust as in the AR/DCR projects?

common-rendering/src/editorialPalette/background.ts
```diff
@@ -227,7 +269,22 @@ const bulletDark = (
 						return neutral[46];
 				}
 			default:
-				return neutral[46];
+				switch (theme) {
```
JamieB-gu, line 272: I'd recommend adding a `default` to this.

common-rendering/src/editorialPalette/background.ts
```diff
@@ -297,6 +373,107 @@ const headlineTag = (format: ArticleFormat): Colour => {
...
+			return neutral[10];
+	}
+
+	return neutral[97];
+};
+
+const relatedCardDark = (format: ArticleFormat): Colour => {
+	switch (format.design) {
+		case ArticleDesign.LiveBlog:
+			switch (format.theme) {
```
JamieB-gu, line 410: I'd recommend adding a `default` to this.

common-rendering/src/editorialPalette/text.ts
```diff
@@ -188,22 +323,47 @@ const bylineLeftColumn = (format: ArticleFormat): Colour => {
...
-		case ArticleSpecial.Labs:
-			return labs[300];
-		case ArticleSpecial.SpecialReport:
-			return specialReport[300];
+	switch (format.design) {
+		case ArticleDesign.LiveBlog:
+			return neutral[100];
+		case ArticleDesign.DeadBlog:
+		default:
+			switch (format.theme) {
```
JamieB-gu, line 341: I'd recommend adding a `default` to this.

JamieB-gu (commented):
This is great! 🚀 

I haven't had a chance to fully review, but I've noticed one possibly unexpected bit of JS behaviour that might be worth taking a look at - more details in one of the comments below.

common-rendering/src/editorialPalette/background.ts
```diff
@@ -263,6 +320,25 @@ const avatar = (format: ArticleFormat): string => {
...
+			return sport[400];
+		case ArticlePillar.Culture:
+			return culture[400];
+		case ArticlePillar.Lifestyle:
+			return lifestyle[400];
+		case ArticleSpecial.SpecialReport:
+			return specialReport[500];
+		case ArticlePillar.News:
+		default:
+			return opinion[400];
```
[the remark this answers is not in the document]
MarSavar: This function is used for in `BylineCard` component which is exclusively used for opinion pieces in the "Related content" section of articles.
I don't think labs opinion pieces exist, but I was running into an issue locally where cards from the series "[Nils Pratley on Finance](https://www.theguardian.com/business/nils-pratley-on-finance)" were being rendered with red quotation marks and and a red avatar background. I have never seen that in DCR, so I decided to default to opinion colours.
What is happening is: the CAPI response for articles in that series says that the pillar is "uk/news". 
DCR transforms that data at some point (`format.theme` is `OpinionPillar`) but that information is not available in AR because the cards are of type `RelatedItemType` which only exists in the AR model. 
In short: I have never seen opinion pieces with the news colours so I'm using opinion colours in that situation.

JamieB-gu (approved):
This is fantastic! Thanks for doing this 🎉

iainjchambers-guardian (approved):

PR #5243, opened 2022-06-22
Slightly less padding on the headline to align with image

### LINKED ISSUE #4721: Align card headline with image

oliverlloyd:
## What does this change?
This aligns the headline with the image to close #4721.

There has [already been work done](https://github.com/guardian/dotcom-rendering/pull/4921) in this area which did create the correct alignment but it seems there could have been a regression since then because when I checked we were 1 pixel out.

| Before      | After      |
|-------------|------------|
| Screenshot 2022-06-22 at 05 53 21] | Screenshot 2022-06-22 at 05 56 57 |


ioannakok (approved):

oliverlloyd:
Looking at the diffs on this PR it turns out this isn't as simple as I thought. I think the way we do this now is correct and this can be closed

PR #5246, opened 2022-06-22
Fix dividers and spacing on FixedLargeSlowXIV

bryophyta:
## What does this change?

It does three main things:
1. Puts the bottom two rows of cards in this container into a single `UL`.
2. Fixes a small bug in the way that the spacing was implemented on mobile devices.
3. Fixes a small bug in the way the cards were being sliced, which led to skipping a couple of cards.

## Why?

Change (1) allows us to add bottom padding to the `<li>` elements in the 3rd row of cards, instead of to the `<ul>` which wrapped it before. Because the padding is on the `li`, the divider line now spans the gap created by that padding (because the divider line is set to be the same height as the `li`. This means that the dividers for these two rows are continuous, [as in the figma design](https://www.figma.com/file/sx2vMFHbL7SsUo0LcpsKNe/%E2%AC%A3--Front-container?node-id=0%3A1).

Change 2 also brings the component into line with the figma designs. On single-column layouts in the figma, there is uniform spacing between each card, regardless of which 'slice' they belong to.

## Screenshots

These are screenshots from Storybook, and they focus in on the main visual change. All of the changes are quite subtle, and none of them _should_ affect anything outside the component itself, so to see the other differences it's probably easiest to check out the [chromatic build](https://www.chromatic.com/build?appId=5dfcbf3012392c0020e7140b&number=13319) where the diffs are highlighted.

It's worth bearing in mind that (3) leads to larger diffs on wider views, but this is just because the cards being rendered are different. (I've put an explanatory note in Chromatic for the relevant pages.)

| Before      | After      |
|-------------|------------|
| before (gaps between dividers on different rows)| image |

You can then reference the labels and map them to corresponding links.

[before2]: https://example.com/before2.png
[after2]: https://example.com/after2.png
-->

### DIFF
--- a/dotcom-rendering/src/web/components/FixedLargeSlowXIV.tsx
+++ b/dotcom-rendering/src/web/components/FixedLargeSlowXIV.tsx
@@ -139,7 +139,11 @@ export const FixedLargeSlowXIV = ({
 						<LI
 							padSides={true}
 							percentage="25%"
-							showDivider={cardIndex !== 4 && cardIndex !== 8}
+							showDivider={
+								cardIndex !== 0 &&
+								cardIndex !== 4 &&
+								cardIndex !== 8
+							}
 							padBottom={cardIndex < 4}
 							padBottomOnMobile={
 								cardIndex !== thirdSlice.length - 1


dotcom-rendering/src/web/components/FixedLargeSlowXIV.tsx
```diff
@@ -59,12 +56,7 @@ export const FixedLargeSlowXIV = ({
 						discussionId={primary.discussionId}
 					/>
 				</LI>
-				<LI
-					padSides={true}
-					showDivider={true}
-					showTopMarginWhenStacked={true}
```
bryophyta, line 65: I've replaced top margin here with bottom padding on the element above. The reason for doing so is because it seemed as though padding on `UL`s can only be set for all breakpoints (no 'not on mobile' option). And once one element has padding on the bottom you need to set all of the spacing on the bottom of the elements, to avoid getting double-sized gaps.

So this seemed like the option which leaves shared code intact. But it also feels like `showTopMarginWhenStacked` would be the more _semantic_ way to express the desired gaps here. Might not be a big deal either way, but thoughts are very welcome! I'm thinking @oliverlloyd you might have some insight here?

dotcom-rendering/src/web/components/FixedLargeSlowXIV.tsx
```diff
@@ -95,61 +87,99 @@ export const FixedLargeSlowXIV = ({
...
+			</UL>
+			<UL direction="row" padBottom={true} wrapCards={true}>
+				{thirdSlice.map((card, cardIndex) => {
+					return (
+						<LI
+							padSides={true}
+							percentage="25%"
+							showDivider={cardIndex % 4 !== 0}
+							padBottom={cardIndex < 4}
+							padBottomOnMobile={cardIndex !== 11}
```
oliverlloyd, line 144:
```suggestion
							padBottomOnMobile={cardIndex !== thirdSlice.length - 1}
```

I think the 11 should have been 7 but maybe using length here explains intentions better? I'm not strong on this though so if you prefer the more explicit use of 7 I'm down with that too

dotcom-rendering/src/web/components/FixedLargeSlowXIV.tsx
```diff
@@ -95,61 +87,99 @@ export const FixedLargeSlowXIV = ({
...
+					);
+				})}
+			</UL>
+			<UL direction="row" padBottom={true} wrapCards={true}>
+				{thirdSlice.map((card, cardIndex) => {
+					return (
+						<LI
+							padSides={true}
+							percentage="25%"
+							showDivider={cardIndex % 4 !== 0}
```
oliverlloyd, line 142:
```suggestion
							showDivider={cardIndex !== 4 && cardIndex !== 8}
```

mod is neat but it's niche, it might be more inclusive to be more explicit, albeit more verbose

oliverlloyd (approved):
Really well done here. You got your head around a pretty narly problem and communicated it well in the PR description.

I think your solution is sound. I see in Frontend we use an additional wrapper element in these situations and arguably that makes the code simpler but means we have an extra layer of divs on the page which is a shame.

There is an argument here to say the benefit of that border stretching down might not be worth the complexity needed to get there @HarryFischer ? But in any case this is impressive work.

dotcom-rendering/src/web/components/FixedLargeSlowXIV.tsx
bryophyta: good call, thanks! I partly had my mind still on the show more case, where we need something similar but without knowing how many rows there will be. I think you're right though that spelling the conditions out explicitly here is good, and if I use modulo in the more flexible version then I will wrap it in a descriptive function that makes clear what it's doing 👍

dotcom-rendering/src/web/components/FixedLargeSlowXIV.tsx
bryophyta: > I think the 11 should have been 7 but maybe using length here explains intentions better? I'm not strong on this though so if you prefer the more explicit use of 7 I'm down with that too

Great call, thanks!

PR #5247, opened 2022-06-22
Ab/add filter bank

abeddow91:
## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |


dotcom-rendering/src/web/components/TopicFilterBank.tsx
```diff
@@ -0,0 +1,73 @@
...
+export const TopicFilterBank = ({ topics, format }: Props) => {
+	const palette = decidePalette(format);
+	return (
+		<div css={containerStyles}>
+			<div css={headlineStyles}>
+				Filters{' '}
+				<span css={headlineAccentStyles}>
+					(
+					<span
+						css={css`
```
joecowton1, line 47: For consistency does it make sense for this to be defined outside the component like the other styles?

joecowton1 (commented):
Looks great! Couple of comments, as well as incorporating the `FilterKeyEvents` functionality and click handlers, we'll need to handle which button is currently active at this filter bank level too.

dotcom-rendering/src/web/components/TopicFilterBank.stories.tsx
```diff
@@ -0,0 +1,57 @@
+import { css } from '@emotion/react';
+import { ArticleDesign, ArticleDisplay, ArticlePillar } from '@guardian/libs';
+import { TopicFilterBank } from './TopicFilterBank';
+
+interface Topic {
```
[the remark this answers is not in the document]
abeddow91: good point! reesolved

dotcom-rendering/src/web/components/TopicFilterBank.tsx
abeddow91: Yeah makes sense, I've refactored this.

abeddow91:
You're totally right, Joe,  this is just the UI and the functionality will follow. These are all ticketed on the trello board :)

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -783,6 +784,23 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 										/>
 									</div>
 								)}
+
+								{CAPIArticle.config.switches.automaticFilters &&
```
joecowton1, line 788: Do you think its worth extracting these to a variable and adding a note to explain what's going on? Might make it easier to unpick when the time comes

abeddow91: Good idea, I've updated with a comment for clarity

joecowton1 (approved):

PR #5223, opened 2022-06-20
Immersive Layout - Caption And Metadata

JamieB-gu:
## Why?

Following on from #5175, this finishes off the header section of the immersive layout. I've added the main media caption, the decorative lines and the metadata (datetime, follow button etc.).

**Note:** The follow button doesn't look amazing in the left column layout, we may want to take a look at this.

## Changes

- Re-wrote immersive caption and added to MainMedia
- Added caption, lines and metadata to immersive layout
- Re-wrote immersive metadata
- Removed immersive caption from standard layout

## Screenshots

| Mobile | Tablet | Wide |
| - | - | - |
| ![immersive-mobile][] | ![immersive-tablet][] | ![immersive-wide][] | 

[immersive-tablet]: https://user-images.githubusercontent.com/53781962/174590464-44b69afc-045c-4421-bc1a-0f3f96b6ebfd.jpg
[immersive-mobile]: https://user-images.githubusercontent.com/53781962/174590467-b7680a26-b6b2-4327-8c69-bc50e713b921.jpg
[immersive-wide]: https://user-images.githubusercontent.com/53781962/174590457-50c5f328-53cf-424a-b894-ec72ed760473.jpg

PR #5248, opened 2022-06-22
Immersive Layout - Body

JamieB-gu:
## Why?

Following on from #5223, this adds the article body to the new immersive layout. Most of the changes here are tweaks to padding and grid alignment to position the body correctly across all breakpoints. The biggest change is the addition of a new `LeftCentreBorder` component, the effect of which can be seen in the wide screenshot below.

**Note:** I haven't settled on the API for the border component yet. There are other borders to add for other designs, and there may be some limitations in the way the `rows` prop currently works. The way this component behaves may well have to change in future.

## Changes

- Added body to immersive layout
- Added `LeftCentreBorder` component

## Screenshots

| Mobile | Tablet | Wide |
| - | - | - |
| ![immersive-body-mobile][] | ![immersive-body-tablet][] | ![immersive-body-wide][] |

[immersive-body-wide]: https://user-images.githubusercontent.com/53781962/175032776-a1cfee96-2994-4286-93f8-ba68881cfeb2.jpg
[immersive-body-tablet]: https://user-images.githubusercontent.com/53781962/175032787-76301170-dfb8-41bf-87bf-935022fa1333.jpg
[immersive-body-mobile]: https://user-images.githubusercontent.com/53781962/175032791-1a53ed9e-56ee-4a0a-b1fb-27c151773d31.jpg


apps-rendering/src/grid/LeftCentreBorder.tsx
```diff
@@ -0,0 +1,36 @@
+// ----- Imports ----- //
```
iainjchambers-guardian, line 1: Is this going to be re-used in multiple places? Far as I can tell it's only used in the `ImmersiveLayout` and all you really need is the function that returns styles from here.

JamieB-gu: Yes there are several layouts that include borders along the side (galleries, for example), so the expectation is that it'll be used for each of those. Also, as mentioned in the description, there may be a couple more of these border components for the other borders and the API might change a bit 🤔.

There's also a possibility this grid module might get lifted to Source, so I'm trying to design it as generically as possible to make it easy to lift out if and when that happens.

Issue #5131, opened 2022-06-07
Spike: Percy/Cypress integration


bryophyta:
Certain fairly small code changes can cause significant and systematic changes to the visual rendering of the live site, without being picked up by our Chromatic/Storybook tests.

An example is in PR #5119. This PR means that bylines will now be rendered on Fronts cards, whereas they never were before. But the code change itself is limited to the `enhance` pipeline, rather than the rendering code, so the changes aren't picked up by Chromatic because existing stories don't use this pipeline.

In the discussion under that PR, @oliverlloyd suggested that using the [Percy integration with Cypress](https://docs.percy.io/docs/cypress) might be a good way to introduce visual regression testing for elements which integrate a larger part of the pipeline from request to rendering.

If we haven't tried Percy for DCR before, it might be worthwhile to try a test to see whether it will suit our purposes.

bryophyta:
see also @oliverlloyd 's [comment](https://github.com/guardian/dotcom-rendering/pull/5119#issuecomment-1148385402)

bryophyta:
Sorry @jlieb10  I missed this before somehow.

A quick update from earlier this week: It looks like other teams are interested in implementing Percy/Cypress integration into the DCR codebase. The basic process looks straightforward, but they are currently addressing a couple of webpack issues, and once these are resolved I believe that the next step is to integrate it with CI.

I think that some of our use-cases (esp. testing Fronts) will likely pose distinctive challenges, but it seems like all of these will come after getting the basic CI integration set up. Given that work is being done on the latter independently, there's probably limited value in trying a spike for our use-cases at the moment. I think it makes sense to revisit this either along with the other teams who are looking into Percy, or else after they've finished their initial integration.

So I'll put this in the backlog for now, but if anyone has other thoughts, please feel free to say so!

arelra:
I have experience implementing Percy for the commercial dev team.

There are also cost implications which I can convey to whoever picks this up.

Given how far the project has come without Percy we should also reconsider if this is still a requirement.

rhiannareechaye:
Closing this issue as I don't believe it to be a requirement anymore

PR #5256, opened 2022-06-23
SPIKE: Percy & Cypress

arelra:
Co-authored-by: Pete Faulconbridge <37048459+bryophyta@users.noreply.github.com>
Co-authored-by: James Gorrie <jamesgorrie@users.noreply.github.com>

## What does this change?

Spike for (Browserstack's) Percy platform for visual regression testing.

https://percy.io/

## Why?

Commercial would like to add visual regression tests but we found we were limited by Chromatic's component-centric approach.

DCR have also expressed an interest at looking at Percy: https://github.com/guardian/dotcom-rendering/issues/5131

Integrating Percy with Cypress enables:
- snapshots directly from our existing Cypress tests
- tests runs in a real browser
- out of the box support for Safari, Firefox, Chrome & Edge browsers

With Cypress v10 we also open the possibility of:
- component testing in Cypress and using Percy for visual testing
- a single solution for e2e tests, component testing and visual testing for both

## Findings

- Test project here: https://percy.io/ced09b1a/article-demo
- Straightforward to setup and integrate
- UI is very Chromatic like and simple to use
- Less diffing features (e.g. no strobing) than Chromatic
- Simple to test multiple browsers and viewport widths
- More browsers supported than Chromatic:
    https://docs.percy.io/docs/cross-browser-visual-testing vs
    https://www.chromatic.com/docs/browsers
- Requires an environment variable `PERCY_TOKEN` to point to the relevant Percy account and project
- In theory no special setup required for CI but not tested
    https://docs.percy.io/docs/github-actions
- Each combination of browser and viewport width counts as a distinct screenshot
- pricing plans primarily based on screenshots per month:
    https://www.browserstack.com/accounts/subscriptions?product=percy&ref=percy

### Multiple browser and viewport support

Screenshot 2022-06-20 at 23 29 49

### Snapshot View

percy-snapshot

### Snapshot View - Diff

percy-snapshot-diff


dotcom-rendering/cypress/e2e/parallel-3/article.interactivity.cy.js
```diff
@@ -15,8 +16,12 @@ describe('Interactivity', function () {
 	});
 
 	describe('Verify elements have been hydrated', function () {
-		it('should open the edition dropdown menu when clicked and hide when expected', function () {
+		// eslint-disable-next-line mocha/no-exclusive-tests
+		it.only('should open the edition dropdown menu when clicked and hide when expected', function () {
 			cy.visit(`/Article?url=${articleUrl}`);
+			cy.percySnapshot('Article', {
+				widths: [320, 375, 480, 660, 740, 980, 1140, 1300],
```
oliverlloyd, line 23: Nit: You can get these by importing `breakpoints` from Source

oliverlloyd (commented):
Well, this is very simple!

Shall we start by adding in a simple test that allows us to get CI setup and then we can continue by enhancing the scope of this test to include more article types? I presume eventually we will want to replace `Layouts.stories`?

dotcom-rendering/cypress/e2e/parallel-3/article.interactivity.cy.js
arelra: Yep it was just something rough for demo purposes :)

arelra:
> I presume eventually we will want to replace `Layouts.stories`?

Is this the test you'd most like to see run through Percy?

We can focus on that one or any others you think are good candidates

oliverlloyd:
> > I presume eventually we will want to replace `Layouts.stories`?
> 
> Is this the test you'd most like to see run through Percy?
> 
> We can focus on that one or any others you think are good candidates

For DCR, yes, this set of tests makes the most sense to run in Percy.

After that, I guess anywhere else where we're using `doStorybookHydration`, as this function was created to get around the problem of the code not running in the context of a browser

oliverlloyd (approved):
Shall we?

arelra:
> Shall we?

Yes however I think we have remaining TODOs:

- implement a deterministic method to hydrate components (onIdle and onVisible) in Cypress. I thought we could do this with a helper function to scroll to each island, waiting for it to be ready and then finally scroll to the bottom of the page
- remove the current test and add a new layout test specifically for Percy

I can pick this up

oliverlloyd:
> * implement a deterministic method to hydrate components (onIdle and onVisible) in Cypress. I thought we could do this with a helper function to scroll to each island, waiting for it to be ready and then finally scroll to the bottom of the page

Are you saying that for Percy we want to be sure that all hydration is activated and complete before we take the entire page snapshot to prevent inconsistent results?

We do already do this in a targeted fashion for some Cypress tests; if a particular island needs hydrating we scroll to it and wait. But perhaps if we want _all_ content to be hydrated we could have a higher level switch, maybe a property similar to `disableLazyLoading` or even a query param on the url? We'd still need to wait for hydration to complete but this could be done with loop over all islands.

Either way, it'd be great to remove `doStorybookHydration`!

arelra:
> Are you saying that for Percy we want to be sure that all hydration is activated and complete before we take the entire page snapshot to prevent inconsistent results?

Yes exactly

> we could have a higher level switch, maybe a property similar to `disableLazyLoading` or even a query param on the url? 

That would make testing a lot easier. I am just unsure if we would be diverging too far from user behaviour i.e. would we lose coverage/reassurance if we bypass the normal hydration?

oliverlloyd:
> That would make testing a lot easier. I am just unsure if we would be diverging too far from user behaviour i.e. would we lose coverage/reassurance if we bypass the normal hydration?

For me, the full page layout tests are focussed on verifying that content renders as expected - we need hydration to verify this but the act of hydration is not the thing being verified. We have separate tests that check things like deferred hydration and that hydration occurs as elements are scrolled into view so maybe we can take the view that - in these cases - hydration is a means to an end but not the primary subject under test?

On the other hand, if we could achieve the goal - hydrate and render all elements - without diverging from standard behaviours then this is defo preferable! Maybe we can try this to start with and only fall back to a high level switch if needed?

bryophyta:
The custom hydrate command looks great to me!

A couple of further thoughts on config etc.:

- I tried to test the diff sensitivity, but it said we don't have the right plan for it, do we know much about that?
- Using Percy for Fronts, I've been thinking in terms of having fixtures with Fronts json so that we can get consistent content from one build to another. This won't work in the same way for client-rendered content though (e.g. ads, onwards containers, etc.) Which more dynamic elements do we need to test using this method, and are there suitable ways of mocking those?
- I assume that we don't want to make Percy a condition of a passing build until if/when we're confident that we've got good tests which aren't throwing up too many false alarms. But do we want to think about setting up some kind of webhook/alerting in the meantime to monitor diffs? Or will we just use the dashboard for the time being?

arelra:
> * I tried to test the diff sensitivity, but it said we don't have the right plan for it, do we know much about that?

I don't sorry. We could ask our BrowserStack contacts for clarification though.

arelra:
Hi @jamesgorrie the Snyk job on this branch is repeatedly failing with this error message:

```
Error: Path does not exist: snyk-node.sarif
Error: Path does not exist: snyk-node.sarif
    at getSarifFilePaths (/home/runner/work/_actions/github/codeql-action/v1/lib/upload-lib.js:151:15)
    at Object.uploadFromActions (/home/runner/work/_actions/github/codeql-action/v1/lib/upload-lib.js:13[9](https://github.com/guardian/dotcom-rendering/runs/7305938289?check_suite_focus=true#step:6:10):30)
    at run (/home/runner/work/_actions/github/codeql-action/v1/lib/upload-sarif-action.js:52:47)
    at async runWrapper (/home/runner/work/_actions/github/codeql-action/v1/lib/upload-sarif-action.js:74:9)
```

@bryophyta mentioned that you are the best person to ask for help with Snyk. Any idea whats happening here? 🙏

arelra:
The Synk error has been fixed thanks to @NovemberTang and @jfsoul - I just had to update `yarn.lock` 🤦 

@bryophyta and @oliverlloyd this is ready to be merged once you have re-reviewed

bryophyta (approved):
This looks good to me! 

The on question I had is: it looks to me like the `snapshot-standard-layout.e2e.cy.js` will get run twice with the current config -- once as part of parallel 7 in the Cypress action, and once in the Percy action. Is that the desired behaviour?

dotcom-rendering/cypress/support/commands.js
```diff
@@ -25,3 +25,21 @@ import 'cypress-wait-until';
...
+				.log(`Scrolling to ${el.attr('name')}`)
+				.scrollIntoView({ duration: 100, timeout: 10000 })
+				.should('have.attr', 'data-gu-ready', 'true', {
+					timeout: 30000,
+				});
+		})
+		.then(() => {
+			cy.scrollTo('top');
+			// eslint-disable-next-line cypress/no-unnecessary-waiting
+			cy.wait(5000);
```
oliverlloyd, line 43: I'm guessing you've already thought this through but is there something in the dom that we can check for here rather than having this wait?

dotcom-rendering/cypress/support/commands.js
```diff
@@ -25,3 +25,21 @@ import 'cypress-wait-until';
...
+Cypress.Commands.add('hydrate', () => {
+	return cy
+		.get('gu-island')
+		.each((el) => {
+			cy.wrap(el)
+				.log(`Scrolling to ${el.attr('name')}`)
+				.scrollIntoView({ duration: 100, timeout: 10000 })
+				.should('have.attr', 'data-gu-ready', 'true', {
+					timeout: 30000,
+				});
```
oliverlloyd, line 38: This is very cool 😎

dotcom-rendering/cypress/e2e/parallel-7/snapshot-standard-layout.e2e.cy.js
```diff
@@ -0,0 +1,23 @@
+import { disableCMP } from '../../lib/disableCMP.js';
+import { setLocalBaseUrl } from '../../lib/setLocalBaseUrl.js';
+import { breakpoints } from '@guardian/source-foundations';
+import '@percy/cypress';
+
+describe('Snapshot standard layout', function () {
```
oliverlloyd, line 6:
```suggestion
describe('Visual regression', function () {
```

dotcom-rendering/cypress/e2e/parallel-7/snapshot-standard-layout.e2e.cy.js
```diff
@@ -0,0 +1,23 @@
...
+	beforeEach(function () {
+		disableCMP();
```
oliverlloyd, line 8: Not for this PR, but perhaps we could have a test that allows the banner to show, dismisses it, refreshes the page, and verifies a different banner shows. We already have tests that check this in other places but given how critical this content is an extra level of assurance would be good

dotcom-rendering/cypress/e2e/parallel-7/snapshot-standard-layout.e2e.cy.js
```diff
@@ -0,0 +1,23 @@
...
+		setLocalBaseUrl();
+	});
+
+	it('Snapshot all breakpoints', function () {
+		cy.visit(
+			'/Article?url=https://www.theguardian.com/uk-news/2020/dec/04/edinburgh-hit-by-thundersnow-as-sonic-boom-wakes-residents',
+		);
```
oliverlloyd, line 15: Is the intention here to map the possible `Design` and `Display` variations, snapshotting each one?

dotcom-rendering/cypress/e2e/parallel-7/snapshot-standard-layout.e2e.cy.js
```diff
@@ -0,0 +1,23 @@
...
+
+		cy.hydrate();
+
+		cy.percySnapshot('Standard Layout', {
+			widths: Object.values(breakpoints),
```
oliverlloyd, line 20: Assuming we're going to be mapping all variations, would a subset of breakpoints make more sense here? It's very much a trade off but this could be a good place to reduce coverage in favour of speed, cost, etc.

oliverlloyd (approved):
Exciting

arelra:
> will get run twice with the current

Good catch thanks! I'll update to separate the snapshot tests from the regular Cypress tests

dotcom-rendering/cypress/support/commands.js
arelra: Although we assert on all islands being 'ready' this is additional safety to ensure that the page has stopped shifting and fully loaded post hydration. 

I thought about waiting for x amount of idle time but that feels error prone in itself 

Without adding more state to islands rendering I can't think of a signal we could use to achieve this 🤔

dotcom-rendering/cypress/e2e/parallel-7/snapshot-standard-layout.e2e.cy.js
arelra: I'd suggest we snapshot important variations although overlap with Chromatic will need to considered. 

@bryophyta if you don't mind I think you are better placed to answer this - thanks.

dotcom-rendering/package.json
```diff
@@ -28,6 +28,7 @@
     "cypress:run": "cypress run",
     "cypress:open:e2e": "cypress open --e2e --browser=chrome",
     "cypress:run:e2e": "cypress run --spec 'cypress/e2e/**/*'",
+    "cypress:run:percy": "percy exec -- cypress run --spec 'cypress/e2e/**/snapshot-*'",
```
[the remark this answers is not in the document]
oliverlloyd: `visual` works. I do also see how `snapshot` works as well...

Lets go snapshot, I sense that this could be my own personal perspective tainting things and maybe snapshot is the best option overall.

dotcom-rendering/cypress/e2e/parallel-7/snapshot-standard-layout.e2e.cy.js
oliverlloyd: > overlap with Chromatic

Is the goal to remove our layout tests in Chromatic and replace them with these?

dotcom-rendering/cypress/e2e/parallel-7/snapshot-standard-layout.e2e.cy.js
oliverlloyd: 👍

dotcom-rendering/cypress/e2e/parallel-7/snapshot-standard-layout.e2e.cy.js
bryophyta: It feels like there are still some unknowns around how this might work in practice, but it seems to me that this would be a positive move if we can make it work. Shall we make some time to pair on this next week?

oliverlloyd: Sounds good 👍  

This shouldn't block this, PR by the way, it was just a question

dotcom-rendering/package.json
```diff
@@ -28,6 +28,7 @@
     "cypress:run": "cypress run",
     "cypress:open:e2e": "cypress open --e2e --browser=chrome",
     "cypress:run:e2e": "cypress run --spec 'cypress/e2e/**/*'",
+    "cypress:run:percy": "percy exec -- cypress run --config specPattern='cypress/snapshot/**/*'",
```
arelra, line 31: This overrides the default `specPattern` i.e. `e2e/**/*.cy.*` so we can have tests outside of the default convention.

Otherwise tests would have to be in `e2e/` and an `excludeSpecPattern ` specified for snapshots when running regular e2e tests.

arelra: So for clearer separation we now have:

```
/cypress
  /e2e
  /snapshot
```

dotcom-rendering/package.json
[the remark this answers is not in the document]
arelra: `visual` is growing on me as ye olde Cypress also used to have a `snapshot` plugin.

But I will merge as is and it can be updated once a wider consensus forms.

PR #5257, opened 2022-06-23
Add padding to live feed when header is removed

silvacb:
## What does this change?
Adds padding to live feed element when header is removed (building on #5240)

## Why?
As part of the automatic filters redesign, the header accordion is being removed. Removing the header also removes the padding, so this needs to be added back in.

## Screenshots
**Before**: 
Screenshot 2022-06-22 at 11 34 58
**After**: 
Screenshot 2022-06-23 at 10 43 55

### DIFF
--- a/dotcom-rendering/src/web/layouts/LiveLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/LiveLayout.tsx
@@ -1,4 +1,5 @@
-import { css, SerializedStyles } from '@emotion/react';
+import type { SerializedStyles } from '@emotion/react';
+import { css } from '@emotion/react';
 // eslint-disable-next-line import/no-extraneous-dependencies -- it’s a yarn workspace
 import Accordion from '@guardian/common-rendering/src/components/accordion';
 // eslint-disable-next-line import/no-extraneous-dependencies -- it’s a yarn workspace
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -254,6 +256,16 @@ interface Props {
 	format: ArticleFormat;
 }
 
+const paddingBody: SerializedStyles = css`
+	padding: ${remSpace[3]};
```
abeddow91, line 260: I've checked in with the dotcom team and on DCR, px is preferred over rem. As such, we should replace these remSpace values for space and adopt px. This would mean this line would change to 	`padding: ${space[3]}px;` for example.

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -254,6 +256,16 @@ interface Props {
...
+	${from.mobileLandscape} {
+		padding: ${remSpace[3]} ${remSpace[5]};
```
abeddow91, line 262: Again, we should change this to use space and px to follow DCR convention

abeddow91 (commented):
Great work! Just a small tweak for consistency then it should be good to go.

dotcom-rendering/src/web/layouts/LiveLayout.tsx
silvacb: Thanks so much Anna, this has been fixed!

abeddow91 (approved):
Great work 🥳

PR #5262, opened 2022-06-23
Migrate Data Types To AR

JamieB-gu:
## Why?

This migrates the remaining code from `@guardian/types` into AR (the rest made its way into `@guardian/libs` a while back). Some bug fixes and feature requests have started accumulating for these data types, and they can't be implemented while the [original repo](https://github.com/guardian/types) is archived. AR is the most extensive consumer of this library and will therefore benefit most from these updates.

This doesn't add any new functionality, it copies the code across verbatim (except for a few whitespace changes made by the version of prettier in this repo 🙄).

**Note:** This PR just moves the code across; it doesn't remove the dependency on the original `@guardian/types` library or update any of the imports yet. I'm trying to break up the changes to avoid enormous PRs, so that work will come in a follow-up.

## Changes

- Migrated `Option`
- Migrated `Result`
- Migrated tests

### DIFF
--- a/apps-rendering/src/option.ts
+++ b/apps-rendering/src/option.ts
@@ -1,0 +1,120 @@
+// ----- Types ----- //
+
+enum OptionKind {
+	Some,
+	None,
+}
+
+type Some<A> = {
+	kind: OptionKind.Some;
+	value: A;
+};
+
+type None = {
+	kind: OptionKind.None;
+};
+
+/**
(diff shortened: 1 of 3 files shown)


iainjchambers-guardian (approved):
Far as I can tell this is just a copy from another location and doesn't add anything new. If I'm wrong please let me know, otherwise LGTM 👍

PR #5264, opened 2022-06-23
Add active topic to CAPIArticle model and use in url construction

abeddow91:
## What does this change?
Adds an optional `activeTopic` field to the CAPIArticle model and regenerates the schema based on this. It also threads this new field through the the `liveness` script so that we can use the active topic as a query in the url, if available.

## Why?
As part of the addition of automated topics in liveblogs, we want to pass the active topic to the server via a query param so that we have the correct query parameters when polling.

We also want to pass this data back to DCR via the CAPIArticle model so that we don't need to rely on extracting the topic from the url query param string. This follows the data flow pattern as filterKeyEvents.

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -612,6 +612,7 @@ interface CAPIArticleType {
 	// Included on live and dead blogs. Used when polling
 	mostRecentBlockId?: string;
 	topics?: Topic[];
+	activeTopic?: string;
 }
 
 type StageType = 'DEV' | 'CODE' | 'PROD';
--- a/dotcom-rendering/src/web/components/Liveness.importable.tsx
+++ b/dotcom-rendering/src/web/components/Liveness.importable.tsx
@@ -16,6 +16,7 @@ type Props = {
 	webURL: string;
 	mostRecentBlockId: string;
 	hasPinnedPost: boolean;
+	activeTopic?: string;
 };
 
 const isServer = typeof window === 'undefined';
@@ -106,6 +107,7 @@ function getKey(
 	ajaxUrl: string,
 	latestBlockId: string,
 	filterKeyEvents: boolean,
+	activeTopic?: string,
 ): string | undefined {
 	try {
 		// Construct the url to poll
@@ -117,6 +119,7 @@ function getKey(
 			'filterKeyEvents',
 			filterKeyEvents ? 'true' : 'false',
 		);
+		if (activeTopic) url.searchParams.set('activeTopic', activeTopic);
 		return url.href;
 	} catch {
 		window.guardian.modules.sentry.reportError(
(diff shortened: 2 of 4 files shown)


dotcom-rendering/index.d.ts
```diff
@@ -612,6 +612,7 @@ interface CAPIArticleType {
 	// Included on live and dead blogs. Used when polling
 	mostRecentBlockId?: string;
 	topics?: Topic[];
+	activeTopic?: string;
```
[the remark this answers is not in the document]
marjisound: Maybe you were thinking to pass the whole thing as string?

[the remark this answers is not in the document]
joecowton1: I don't know if there's any advantage to passing the `activeTopic` in two parts(type / value) if we're only ever going to combine them in DCR? Its an interesting one though.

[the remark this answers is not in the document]
marjisound: So you think even for the buttons, we'd just use `org:nhs` as the idendifier of the buttons to show the button is active or not?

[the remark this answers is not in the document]
joecowton1: No, I think for the `topics` we need both so the button label can be just the value, but for the `activeTopic` we only ever need them together

marjisound (approved):
Looks great 🎉

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -117,6 +119,7 @@ function getKey(
 			'filterKeyEvents',
 			filterKeyEvents ? 'true' : 'false',
 		);
+		if (activeTopic) url.searchParams.set('activeTopic', activeTopic);
```
oliverlloyd, line 122: Is this `activeTopic` query param semantically the same as the other `topics` query param? If so, it might be better to be consistent?

alinaboghiu: I think it is. 

There is a **topics** field in the CAPI article to represent all existing topics on that article.
And an **activeTopic** to represent the current selected topic. 

Elsewhere though we've been naming the parameter "topics", so I think it's in those places we might have to rename to "activeTopic". And probably stick to the singular until we allow selecting multiple. 

@marjisound @joecowton1 @abeddow91 what do you think?

alinaboghiu: Maybe better still we should have the model contain "availableTopics" and "selectedTopics", then the parameter could always be "topics", to keep it concise.

alinaboghiu: This way we don't have to rename the parameter again either :)

joecowton1: I think that makes sense

> Maybe better still we should have the model contain "availableTopics" and "selectedTopics", then the parameter could always be "topics", to keep it concise.

I think that makes sense. Anna and Marjan are both of today but I'll update this and the corresponding Frontend PR

Issue #5143, opened 2022-06-07
"dynamic/package" (`61`)


bryophyta:

oliverlloyd:
Fixed by https://github.com/guardian/dotcom-rendering/pull/5270

Issue #5267, opened 2022-06-24
Dynamo floating sublinks


oliverlloyd:
The Dynamo (`dynamo/package`) container has floating sub links that sit above the image

![Image](https://user-images.githubusercontent.com/1336821/175522214-eb0df801-766a-4cec-9e20-b2b1b8124c59.png)

on mobile

![Image](https://user-images.githubusercontent.com/1336821/175532952-ceacc1b8-2dbf-464c-9d72-941423290c68.png)

Issue #5196, opened 2022-06-14
Dynamo liveblog latest links


oliverlloyd:
The latest three liveblog posts are displayed as snippets under the headline on Dynamo containers.
<img width="971" alt="Screenshot 2022-06-14 at 21 11 59" src="https://user-images.githubusercontent.com/1336821/173680075-b6502382-5e93-4a26-b960-a03a6e445379.png">

These links only appear if:

1. The container is `dynamic/package`
2. The switch `Live updates` is on in the fronts tool

This is where the fronts tool switch lives
![image](https://user-images.githubusercontent.com/1336821/188608779-16fd3d46-14d9-4bbc-9edd-2b33ac2f675a.png)

Example request
https://api.nextgen.guardianapps.co.uk/politics/live/2022/sep/05/liz-truss-pm-prime-minister-rishi-sunak-energy-bill-freeze-tory-conservative-leader-uk-politics-live-latest-updates.json?rendered=false

### Consider
- These are currently loaded in client side. Space has to be reserved
- Should we refresh these on a poll or offer the ability to refresh?
- Can we use [css elipsis](https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/) for truncation?
- Do we need a loading state?
- What if there are no results?
- Do we filter by key events or all?
- Can we create an in-memory cache for the latest n posts on each live blog and read in from there so we can server render this content? We can then progressively enhance the container by adding javascript to keep these posts up to date

OllysCoding:
Latest link support is also required on

dynamic/fast 'twoOrMoreBigsFirstBoosted' layout

<img width="1307" alt="image" src="https://user-images.githubusercontent.com/9575458/188897320-2fb5a300-8a51-492b-b0e0-1aa50665894e.png">

rhiannareechaye:
See [Investigate what work is left re Dynamo#6381](https://github.com/guardian/dotcom-rendering/issues/6381)

mxdvl:
> These links only appear if:
> 
>     1. The container is `dynamic/package`
> 
>     2. The switch `Live updates` is on in the fronts tool
> 
>     3. The card is the first in the container
> 
>     4. The first card is in the `standard` group

I’ve based the logic to simply rely on: **The switch `Live updates` is on in the fronts tool**

Issue #4612, opened 2022-04-12
The image slideshow


oliverlloyd:
Some Cards have an image carousel. Those images should have captions

AshCorr:
Closing in favour of #5654

PR #5270, opened 2022-06-24
Basic support for `dynamo/package`

oliverlloyd:
## What does this change?
This adds basic support for the `dynamo/package`. There remain additional features to be added to this container but I wanted to get the layout changes in as one PR and introduce other features one by one.

### Changes to `Card` props

`headlineSize: huge`
I added this prop value to support the extra large font size used on dynamos

`transparent: boolean`
We set background colour using `decidePalette` but the main image in a dynamo uses a transparent background so we need an imperative way to set this

`containerType`
We now pass the type of container down as the use case for the Dynamo is to vary some styles based on the type of container they are in. In this case, the variation is with the height of the top border line which shows for `dynamo/package`. Instead of 2 pixels we use 4.

| Frontend      | DCR      |
|-------------|------------|
| Screenshot 2022-06-24 at 11 10 06 | Screenshot 2022-06-24 at 11 09 44 |

## What's remaining?
[Floating sublinks](https://github.com/guardian/dotcom-rendering/issues/5267)
image

[Latest links](https://github.com/guardian/dotcom-rendering/issues/5196)
image

[Image Carousel](https://github.com/guardian/dotcom-rendering/issues/4612)


HarryFischer:
🤩

dotcom-rendering/src/web/components/CardHeadline.tsx
```diff
@@ -24,36 +25,50 @@ type Props = {
 	linkTo?: string; // If provided, the headline is wrapped in a link
 };
 
-const fontStyles = (size: SmallHeadlineSize) => {
+const fontStyles = (
+	size: SmallHeadlineSize,
+	containerPalette?: DCRContainerPalette,
+) => {
+	const options: FontScaleArgs = {};
+	if (containerPalette) options.fontWeight = 'bold';
```
OllysCoding, line 33: If this logic is just looking for container palette to be defined, would it be cleaner if this function took a parameter like `bold: boolean` and we did `fontSyles(size, !!containerPalette)` when calling it?

OllysCoding (approved):
+1 A really good start on this container :)

dotcom-rendering/src/web/components/CardHeadline.tsx
OllysCoding: Very tidy! Nice 👍

oliverlloyd:
Closes #5143

PR #5272, opened 2022-06-24
Add recipe schema atom

frederickobrien:
## What does this change?

This adds a barebones recipe atom to dotcom-rendering codebase. The desired result is being able to add custom blobs of [recipe schema](https://schema.org/Recipe) to specific articles. This will have no impact on what appears visually but will render under the hood as something web crawlers can understand. 

This ought to make pages with recipe schema atoms eligible for [rich results](https://search.google.com/test/rich-results) in search engines - and machine readable for all sorts other interesting possibilities. 

Used [chart atom](https://github.com/guardian/dotcom-rendering/blob/10f09832a0b2295b69da896bd2ada8d17efda32e/dotcom-rendering/src/web/lib/renderElement.tsx#L199-L205) and [partial hydration documentation](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/docs/architecture/027-better-partial-hydration.md) for reference. I've erred on the side of not wrapping them in islands as the schema needs to be present for web crawlers and the like (though obviously happy to be corrected).

## Why?

This is for an experiment Ophan is hoping to run in which recipe schema can be rendered under the hood for a handful of pages, allowing us to track a before and after picture of SEO performance. Semantic markup for recipes has a storied Guardian history of not quite being implemented. If we can get some schema on the site and a clear SEO performance bump follows, maybe that will help make the case for it to be implemented more broadly.

### DIFF
--- a/dotcom-rendering/src/lib/content.d.ts
+++ b/dotcom-rendering/src/lib/content.d.ts
@@ -333,6 +333,12 @@ interface QABlockElement {
 	role?: RoleType;
 }
 
+interface RecipeSchemaAtomBlockElement {
+	_type: 'model.dotcomrendering.pageElements.RecipeSchemaAtomBlockElement';
+	id: string;
+	json: string;
+}
+
 interface RichLinkBlockElement {
 	_type: 'model.dotcomrendering.pageElements.RichLinkBlockElement';
 	elementId: string;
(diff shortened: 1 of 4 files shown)


frederickobrien:
Closed as it turns out this isn't the right approach for the experiment we want to run.

PR #5273, opened 2022-06-27
Extra cards container

bryophyta:
## What does this change?

- Adds a new container type, `ExtraCardsContainer`, which can display a set of cards of arbitrary length.
- Adds a new prop, `offsetBottomPaddingOnDivider`, to the `LI` component, to allow the length of the card dividers to be adjusted in the new container component.
- Also adds a new vertical divider style object, to implement this adjustment when needed.

## Why?

- The motivation for adding a flexible container is, primarily, in anticipation of adding `Show More` functionality to DCR Fronts.
- This container needs to be flexible on the server side because a 'show more' container can contain any number of cards.
- But the reason for rendering all of the rows of cards in a single `UL` component is that this allows the cards to re-arrange automatically if some of the cards are removed on the client side. (This should fix a bug in the existing Frontend implementation of client-side deduping, which is [documented in this issue](https://github.com/guardian/dotcom-rendering/issues/4604#issuecomment-1140818041).)
- However, there is some doubt about when we will introduce client-side deduping. If we decide not to pursue it then it might be worth revisiting this component, because some of the relatively complicated logic could be avoided if we can follow the established DCR pattern and place each row in a different `UL` 'slice' on the server side.

## Screenshots
Desktop:
Desktop screenshot

Mobile:
Mobile screenshot

### DIFF
--- a/dotcom-rendering/src/web/components/Card/components/LI.tsx
+++ b/dotcom-rendering/src/web/components/Card/components/LI.tsx
@@ -1,9 +1,10 @@
+import type { SerializedStyles } from '@emotion/react';
 import { css } from '@emotion/react';
 import { from, until } from '@guardian/source-foundations';
 import { verticalDivider } from '../../../lib/verticalDivider';
 
 const liStyles = css`
-	/* This position relative is needed to contain the veritcal divider */
+	/* This position relative is needed to contain the vertical divider */
 	position: relative;
 
 	display: flex;
@@ -72,6 +73,7 @@ type Props = {
 	padBottomOnMobile?: boolean; // Should be true if spacing below is desired on mobile devices
 	showTopMarginWhenStacked?: boolean;
 	snapAlignStart?: boolean; // True when snapping card when scrolling e.g. in carousel
+	customStyles?: SerializedStyles;
 };
 
 export const LI = ({
@@ -85,6 +87,7 @@ export const LI = ({
 	padBottomOnMobile,
 	showTopMarginWhenStacked,
 	snapAlignStart = false,
+	customStyles = undefined,
 }: Props) => {
 	// Decide sizing
 	const sizeStyles = decideSize(percentage, stretch);
@@ -100,6 +103,7 @@ export const LI = ({
 				padBottomOnMobile && mobilePaddingBottomStyles,
 				showTopMarginWhenStacked && marginTopStyles,
 				snapAlignStart && snapAlignStartStyles,
+				customStyles && customStyles,
 			]}
 		>
 			{children}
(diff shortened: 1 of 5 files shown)


dotcom-rendering/src/web/lib/verticalDivider.ts
```diff
@@ -12,6 +12,7 @@ export const verticalDivider = css`
 			left: 0;
 			width: 1px;
 			height: 100%;
+			height: calc(100% - var(--card-divider-offset-bottom));
```
OllysCoding, line 15: Is there a situation where `--card-divider-offset-bottom` is not defined and if so how does it behave?

I'm very cautious here that using a CSS variable like this introduces a 'hidden' relationship between components, is there any way we can avoid this? If not we should definitely add a comment explaining exactly what this is doing and where it's set!

dotcom-rendering/src/web/components/Card/components/LI.tsx
```diff
@@ -85,19 +101,22 @@ export const LI = ({
 	padBottomOnMobile,
 	showTopMarginWhenStacked,
 	snapAlignStart = false,
+	offsetBottomPaddingOnDivider = false,
 }: Props) => {
 	// Decide sizing
 	const sizeStyles = decideSize(percentage, stretch);
+	const paddingSize = '10px';
```
OllysCoding, line 108: I think it could worth adding comment here describing _why_ we're defining paddingSize in a central place for this component?

dotcom-rendering/src/web/lib/verticalDividerWithBottomOffset.ts
```diff
@@ -0,0 +1,24 @@
+import { css, SerializedStyles } from '@emotion/react';
+import { border, from } from '@guardian/source-foundations';
+
+export function verticalDividerWithBottomOffset(
```
OllysCoding, line 4: Do we have plans for this to be used in other places? If not would it be better to live in LI.tsx?

OllysCoding (approved):
There's a lot of clever logic going on here to make this work and I'm not 100% sure I have wrapped my head around it fully, but it seems well thought out and reasoned so I'm quite happy :)

bryophyta:
> There's a lot of clever logic going on here to make this work and I'm not 100% sure I have wrapped my head around it fully, but it seems well thought out and reasoned so I'm quite happy :)

Yes I'm a bit concerned about adding some maintenance overhead with the logic here... The initial motivation for doing it like this was to allow for client-side removal of duplicate cards (this is what Frontend does currently, although with some bugs which the current container is aiming to fix). I've just realised that I haven't spelled out that motivation in the main PR, so I'll add that now.

However, after a recent conversation with @jamesgorrie it looks like the client-side deduplication functionality is not something that we should prioritise in the initial rollout of the 'show more' functionality. Perhaps if we decide not to add client-side deduping in the long run then we should revisit this container and simplify the logic?

PR #5274, opened 2022-06-27
Add check for `FilterButton` active state

joecowton1:
## What does this change?
Checks whether a `FilterButton` is currently active by comparing the `selectedTopics` with the button's topic `type` and `value`.

_Note: this approach will work if we expand to allow multiple active filters (not currently supported in Frontend), but will require a conditional implementation for the `FilterKeyEvents` button._

## Why?

Within a filter bank we need to highlight which button is currently active.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/77005274/175944235-5a107b2e-497a-4c8a-9cff-11fad7560b76.png
[after]: https://user-images.githubusercontent.com/77005274/175944273-bd8f3a97-27ea-4307-8ccd-ca7f9ced23d5.png

### DIFF
--- a/dotcom-rendering/src/web/components/FilterButton.importable.tsx
+++ b/dotcom-rendering/src/web/components/FilterButton.importable.tsx
@@ -13,7 +13,7 @@ interface ButtonProps {
 	type?: TopicType;
 	count?: number;
 	format: ArticleFormat;
-	isActive?: boolean;
+	isActive: boolean;
 	onClick: () => void;
 }
 
--- a/dotcom-rendering/src/web/components/TopicFilterBank.tsx
+++ b/dotcom-rendering/src/web/components/TopicFilterBank.tsx
@@ -59,7 +59,9 @@ export const TopicFilterBank = ({
 			<div css={topicStyles}>
 				{availableTopics.slice(0, 5).map((topic: Topic) => {
 					const buttonParams = `${topic.type}:${topic.value}`;
-					const isActive = selectedTopics?.includes(buttonParams);
+					const isActive = selectedTopics
+						? selectedTopics.includes(buttonParams)
+						: false;
 
 					return (
 						<FilterButton


dotcom-rendering/src/web/components/FilterButton.importable.tsx
```diff
@@ -13,7 +13,7 @@ interface ButtonProps {
 	type?: TopicType;
 	count?: number;
 	format: ArticleFormat;
-	isActive: boolean;
+	isActive?: boolean;
```
abeddow91, line 16: Does this need to be optional? Looking below I can't see why it wouldn't ever be true or fasle

joecowton1: If `selectedTopics` is undefined then as things are currently, `isActive` would be undefined.

We could do this instead though which might be preferable:
```js
	const isActive = selectedTopics
						? selectedTopics.includes(buttonParams)
						: false;
						```

File: dotcom-rendering/src/web/components/TopicFilterBank.tsx [hunk 2]
```diff
@@ -52,14 +57,17 @@ export const TopicFilterBank = ({ availableTopics, format }: Props) => {
 			</div>
 
 			<div css={topicStyles}>
-				{availableTopics.slice(0, 5).map((topic) => {
+				{availableTopics.slice(0, 5).map((topic: Topic) => {
+					const buttonParams = `${topic.type}:${topic.value}`;
+					const isActive = selectedTopics?.includes(buttonParams);
```
abeddow91, line 62: Appreciate this is mentioned in the PR but I think this is going to be a problem once we add in key events as we'll need the two button types to be in sync with each other.

joecowton1: I think each button click will set new params: this would override whatever the previous params were so everything would always be in sync

dotcom-rendering/src/web/components/FilterButton.importable.tsx
joecowton1: Ok, updated to the above.

abeddow91 (approved):
Nice!

PR #5275, opened 2022-06-27
Create new click handler

joecowton1:
## What does this change?

Adds a click handler to update the `topics` query param.

As the filter bank will contain both `topics` and a `filterKeyEvents` button, this click handler is only for `topics`, and will reset a `filterKeyEvents` param if one exists, keeping the params in sync with the filter buttons.


dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
```diff
@@ -63,14 +64,30 @@ export const TopicFilterBank = ({
 						? selectedTopics.includes(buttonParams)
 						: false;
 
+					const handleTopicClick = () => {
```
abeddow91, line 67: Could we move this out of the return statement so it doesn't muddy the code?

abeddow91 (approved):

marjisound (approved):
Nice one

PR #5285, opened 2022-06-29
fix: Hydration of components when rendered through Frontend

chrislomaxjones:
## What does this change?

Re-introduce a variant of a fix first used in https://github.com/guardian/dotcom-rendering/pull/3471.

When running DCR locally in development, ensure we use absolute paths for the assets that we serve. 

Do this in two places:
-  `decideAssetOrigin`: This is where we decide the path to give scripts included in the `<head>` e.g. Sentry, the CMP, islands, etc.
- `__webpack_public_path__`: The base of the URL that is used to fetch dynamic imports. We need to ensure this is absolute in development so that `.importable.js` bundles can be fetched when requesting via Frontend.

Without these two changes, DCR articles rendered via Frontend will not be able to fetch the `<head>` scripts, preventing critical services such as CMP and Island hydration from starting. _Incidentally, this is why the commercial bundle hangs, since we're not able to obtain consent._

This will not affect production asset serving, since all assets are served from a separate `assets.guim.co.uk` URL.

This should fix https://github.com/guardian/frontend/issues/24874.

## Why?

The issue arises when running DCR locally and proxying through Frontend. Illustrated in the diagram below, when DCR returns the HTML, it includes asset paths that are relative to DCR's dev server and not Frontend. Therefore, when the browser evaluates the page, it'll go off and fetch assets relative to Frontend, resulting in 404s.

```mermaid
sequenceDiagram
    participant Browser as Browser
    participant Frontend as Frontend :9000
    participant DCR as DCR :3030

    Browser ->> Frontend: GET http://localhost:9000/...

    Frontend ->> DCR: POST DCR data model

    activate DCR

    DCR ->> Frontend: Article HTML

    deactivate DCR
    activate Frontend

    Frontend ->> Browser: Article HTML

    deactivate Frontend

    par
        Browser -x Frontend: GET /assets/islands.js
    and
        Browser -x Frontend: GET /assets/bootCMP.js
    and
        Browser -x Frontend: GET /assets/loadSentry.js
    and
        Browser -x Frontend: GET ...
    end
```

We can fix this by supplying absolute paths:

```mermaid
sequenceDiagram
    participant Browser as Browser
    participant Frontend as Frontend :9000
    participant DCR as DCR :3030

    Browser ->> Frontend: GET http://localhost:9000/...

    Frontend ->> DCR: POST DCR data model

    activate DCR

    DCR ->> Frontend: Article HTML

    deactivate DCR
    activate Frontend

    Frontend ->> Browser: Article HTML

    deactivate Frontend

    par
        Browser ->> DCR: GET http:localhost:3030/assets/islands.js
        activate DCR
        DCR ->> Browser: islands.js
        deactivate DCR
    and
        Browser ->> DCR: GET http:localhost:3030/assets/bootCMP.js
        activate DCR
        DCR ->> Browser: bootCMP.js
        deactivate DCR
    and
        Browser ->> DCR: GET http:localhost:3030/assets/loadSentry.js
        activate DCR
        DCR ->> Browser: loadSentry.js
        deactivate DCR
    and
        Browser ->> DCR: GET ...
    end
```

### DIFF
--- a/dotcom-rendering/src/lib/assets.ts
+++ b/dotcom-rendering/src/lib/assets.ts
@@ -22,7 +22,7 @@ try {
  * @param {'PROD' | 'CODE' | undefined} stage the environment code is executing in
  * @returns {string}
  */
-const decideAssetOrigin = (
+export const decideAssetOrigin = (
 	stage: string | undefined,
 	isDev: boolean,
 ): string => {
--- a/dotcom-rendering/src/lib/assets.ts
+++ b/dotcom-rendering/src/lib/assets.ts
@@ -33,6 +33,8 @@ const decideAssetOrigin = (
 			return 'https://assets-code.guim.co.uk/';
 		default: {
 			if (isDev) {
+				// Use absolute asset paths in development mode
+				// This is so paths are correct when treated as relative to Frontend
 				return 'http://localhost:3030/';
 			} else {
 				return '/';
--- a/dotcom-rendering/src/lib/assets.ts
+++ b/dotcom-rendering/src/lib/assets.ts
@@ -22,25 +22,34 @@ try {
  * @param {'PROD' | 'CODE' | undefined} stage the environment code is executing in
  * @returns {string}
  */
-const decideAssetOrigin = (stage: string | undefined): string => {
+const decideAssetOrigin = (
+	stage: string | undefined,
+	isDev: boolean,
+): string => {
 	switch (stage?.toUpperCase()) {
 		case 'PROD':
 			return 'https://assets.guim.co.uk/';
 		case 'CODE':
 			return 'https://assets-code.guim.co.uk/';
-		default:
-			return '/';
+		default: {
+			if (isDev) {
+				return 'http://localhost:3030/';
+			} else {
+				return '/';
+			}
+		}
 	}
 };
-export const ASSET_ORIGIN = decideAssetOrigin(process.env.GU_STAGE);
+
+const isDev = process.env.NODE_ENV === 'development';
+
+export const ASSET_ORIGIN = decideAssetOrigin(process.env.GU_STAGE, isDev);
 
 export const getScriptArrayFromFile = (
 	file: `${string}.js`,
 ): { src: string; legacy?: boolean }[] => {
 	if (!file.endsWith('.js'))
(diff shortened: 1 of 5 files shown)


dotcom-rendering/src/lib/assets.ts
```diff
@@ -22,26 +22,37 @@ try {
...
 			return 'https://assets.guim.co.uk/';
 		case 'CODE':
 			return 'https://assets-code.guim.co.uk/';
-		default:
-			return '/';
+		default: {
+			if (isDev) {
+				// Use absolute asset paths in development mode
+				// This is so paths are correct when treated as relative to Frontend
+				return 'http://localhost:3030/';
```
chrislomaxjones, line 38: I was unsure if there's a way to avoid hardcoding the dev server port here. Is there a way to obtain it from the webpack dev server config?

arelra, line 38: Could we add a simple unit test for this? Just conscious that it's been inadvertently broken twice. Hopefully a unit test will flag any regressions early.

dotcom-rendering/src/web/browser/webpackPublicPath.ts
```diff
@@ -1,7 +1,26 @@
+/**
+ * Determine the path that webpack should use as base for dynamic imports
+ *
+ * @returns The webpack public path to use
+ */
+const decidePublicPath = () => {
```
arelra, line 6: ditto re unit test

arelra (approved):
Super work ! Great to have this fixed again 👏👏👏 

Oh and thanks for the very clear PR description 👍

dotcom-rendering/src/lib/assets.ts
Jakeii: you could require/import it like the mainifest.json above perhaps?

chrislomaxjones: Very good idea, have added them to this PR.

chrislomaxjones: This is a neat idea but I'm actually going to leave it, for now at least. I think it might be worth just keeping it as simple as possible.

PR #5288, opened 2022-06-29
Add hash to topic bank for scrolling

abeddow91:
## What does this change?
Adds a new element to the key events carousel with an id that we can use to scroll to when the page refreshes after automatic filtering. Because we render the key events carousel, once for mobile and once for desktop, the id is passed in as a prop so that we don't have duplicate ids.

This PR also adds a `urlHash` property to the topic filter bank that takes the id of the key event carousel mentioned above, again varying depending on if it is mobile or desktop.

It also adds a missing `Island` wrapper to the mobile implementation of `TopicFilterBank` to ensure it is hydrated. 

## Why?
By adding the above properties, we can assign a hash to the url on filter click  that means we scroll to a relevant part of the page after reloading with the live blog filter.


dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
```diff
@@ -9,6 +9,7 @@ type Props = {
 	format: ArticleFormat;
 	filterKeyEvents: boolean;
 	keyEvents?: Block[];
+	urlHash: 'key-events-carousel-desktop' | 'key-events-carousel-mobile';
```
joecowton1, line 12: Does it make more sense to just call this `id` everywhere?

marjisound: Yea I agree too

abeddow91: cool, amended!

dotcom-rendering/src/web/components/KeyEventsCarousel.importable.tsx
```diff
@@ -112,6 +114,7 @@ export const KeyEventsCarousel = ({
 	const shortCarousel = carouselLength <= 4;
 	return (
 		<>
+			<div id={id} />
```
abeddow91, line 117:
```suggestion
			<span id={id} />
```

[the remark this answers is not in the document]
abeddow91: Thanks Joe! I've updated

PR #5289, opened 2022-06-30
useApi: return 'loading=false' if url undefined

bryophyta:
[SWR recommends](https://swr.vercel.app/docs/conditional-fetching#conditional) handling conditional fetching for their useSWR hook by passing a falsey value if you don't want to make a
request.

We have a `useApi` hook which wraps `useSWR`, and this wrapper returns `loading=!data & !error`. This means it was returning `loading=true` when a falsey value was passed as the `url`, even though nothing was actually loading.

It seems more intuitive to me that `loading` should be `false` if no request has actually been made, and this would also be useful for an upcoming feature where we want to wait until a user event before we make the API request. As far as I can tell, no components which use `useAPI` currently rely on the existing behaviour, because they all have `url` as a required parameter, so their behaviour shouldn't be affected by the extra condition being added here.

The components which currently use the `loading` value returned by the hook are:

- src/web/components/GetCricketScoreboard.importable.tsx
- src/web/components/GetMatchNav.importable.tsx
- src/web/components/GetMatchStats.importable.tsx
- src/web/components/GetMatchTabs.importable.tsx
- src/web/components/OnwardsData.tsx

I've tested these locally and they appear to be working as they do in prod. See e.g. [this cricket liveblog](https://www.theguardian.com/sport/live/2022/jun/24/england-v-new-zealand-third-test-day-two-live-score-updates-cricket) and [this football match report](https://www.theguardian.com/football/live/2022/may/29/huddersfield-town-v-nottingham-forest-championship-play-off-final-live).

### DIFF
--- a/dotcom-rendering/src/web/lib/useApi.tsx
+++ b/dotcom-rendering/src/web/lib/useApi.tsx
@@ -40,6 +40,6 @@ export const useApi = <T,>(
 	return {
 		data,
 		error,
-		loading: url && !error && !data,
+		loading: !!url && !error && !data,
 	};
 };
--- a/dotcom-rendering/src/web/lib/useApi.tsx
+++ b/dotcom-rendering/src/web/lib/useApi.tsx
@@ -40,6 +40,6 @@ export const useApi = <T,>(
 	return {
 		data,
 		error,
-		loading: !error && !data,
+		loading: typeof url !== 'undefined' && !error && !data,
 	};
 };
--- a/dotcom-rendering/src/web/lib/useApi.tsx
+++ b/dotcom-rendering/src/web/lib/useApi.tsx
@@ -40,6 +40,6 @@ export const useApi = <T,>(
 	return {
 		data,
 		error,
-		loading: url !== undefined && !error && !data,
+		loading: url && !error && !data,
 	};
 };
--- a/dotcom-rendering/src/web/lib/useApi.tsx
+++ b/dotcom-rendering/src/web/lib/useApi.tsx
@@ -40,6 +40,6 @@ export const useApi = <T,>(
 	return {
 		data,
 		error,
-		loading: typeof url !== 'undefined' && !error && !data,
+		loading: url !== undefined && !error && !data,
 	};
 };


dotcom-rendering/src/web/lib/useApi.tsx
```diff
@@ -40,6 +40,6 @@ export const useApi = <T,>(
 	return {
 		data,
 		error,
-		loading: !error && !data,
+		loading: typeof url !== 'undefined' && !error && !data,
```
OllysCoding, line 43: Would `url !== undefined` do the same thing?

OllysCoding (approved):
Sorry meant to approve with my question! good fix :)

dotcom-rendering/src/web/lib/useApi.tsx
bryophyta: oh yes that's right, thanks!

bryophyta: I'd actually quite like to make `url` mandatory, with a type of `string | null`, because it feels like this fits better with the SWR approach, but that feels like a riskier change if we don't have a great overview of how the hook is used, right?

Anyway, I've fixed the current implementation in line with your suggestion, thanks!

dotcom-rendering/src/web/lib/useApi.tsx
```diff
@@ -40,6 +40,6 @@ export const useApi = <T,>(
 	return {
 		data,
 		error,
-		loading: !error && !data,
+		loading: url !== undefined && !error && !data,
```
mxdvl, line 43:
```suggestion
		loading: url && !error && !data,
```

bryophyta: I considered this but I this would return false if `url` was an empty string, right? But I thought maybe we don't want to encourage passing an empty string as a way of preventing an API call from being made here, so it would be good to restrict it to `undefined` (or `null` if we're willing to refactor later).

On the other hand, if someone passes an empty string by mistake, your suggestion would be more forgiving at runtime, I think, so might be preferable.

I don't have strong opinions either way so lmk what you think and I'm happy to change it if so

mxdvl: I think the SWR docs assume that if falsey, no request will be made, so I'd tend to follow that same pattern here!

bryophyta: sounds good, change accepted -- thanks!

PR #5290, opened 2022-06-30
Ensure selected topic is always available in ui

marjisound:
## What does this change?
This PR does the following:

- Updates the `selectedTopics` in `CAPIArticleType` interface to be `Topic[]`
- make `count` field optional in `Topic` model
- Ran `gen-schema` to create the new schema

- Adds logic in `TopicFilterBank` component to make sure the selected topic is always included within the 5 available topics 
## Why?
User loads the liveblog page, which has a list of 5 filter topics (we only allow 5 topics for now due to space limitation). After a few minutes, user chooses one of the topics and page refreshes again. 
But during the time between when user first loaded the page and when user chooses a topic, the available topics list could have updated in the server, so a new list is given to DCR. Due to this, there is a possibility that the selected topic is not within the top 5 of the  available topics. 

This change makes sure that, the selected topic would always show up within the UI topics. 

## Screenshots

| Before      | After      |
|-------------|------------|
| before | after |

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -619,7 +619,7 @@ interface CAPIArticleType {
 	// Included on live and dead blogs. Used when polling
 	mostRecentBlockId?: string;
 	availableTopics?: Topic[];
-	selectedTopics?: string;
+	selectedTopics?: Topic[];
 }
 
 type StageType = 'DEV' | 'CODE' | 'PROD';
@@ -1173,7 +1173,7 @@ type MatchReportType = {
 interface Topic {
 	type: TopicType;
 	value: string;
-	count: number;
+	count?: number;
 }
 
 type TopicType = 'ORG' | 'PRODUCT' | 'PERSON' | 'GPE' | 'WORK_OF_ART' | 'LOC';
--- a/dotcom-rendering/src/web/components/Liveness.importable.tsx
+++ b/dotcom-rendering/src/web/components/Liveness.importable.tsx
@@ -16,7 +16,7 @@ type Props = {
 	webURL: string;
 	mostRecentBlockId: string;
 	hasPinnedPost: boolean;
-	selectedTopics?: string;
+	selectedTopics?: Topic[];
 };
 
 const isServer = typeof window === 'undefined';
@@ -107,7 +107,7 @@ function getKey(
 	ajaxUrl: string,
 	latestBlockId: string,
 	filterKeyEvents: boolean,
-	selectedTopics?: string,
+	selectedTopics?: Topic[],
 ): string | undefined {
 	try {
 		// Construct the url to poll
@@ -119,7 +119,11 @@ function getKey(
 			'filterKeyEvents',
 			filterKeyEvents ? 'true' : 'false',
 		);
-		if (selectedTopics) url.searchParams.set('topics', selectedTopics);
+		if (selectedTopics && selectedTopics.length > 0)
+			url.searchParams.set(
+				'topics',
+				`${selectedTopics[0].type}:${selectedTopics[0].value}`,
+			);
 		return url.href;
 	} catch {
 		window.guardian.modules.sentry.reportError(
--- a/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
+++ b/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
@@ -81,6 +81,7 @@ export const TopicFilterBank = ({
 	const palette = decidePalette(format);
 
 	const selectedTopic = selectedTopics?.[0];
+	const topFiveTopics = availableTopics.slice(0, 5);
 
 	if (selectedTopic) {
 		const selectedIndex = availableTopics.findIndex(
@@ -88,10 +89,7 @@ export const TopicFilterBank = ({
 		);
 
 		if (selectedIndex > 4) {
-			// this removes the topic at it's original index
-			const topicToMove = availableTopics.splice(selectedIndex, 1)[0];
-			// this adds the removed topic in index 4
-			availableTopics.splice(4, 0, topicToMove);
+			topFiveTopics.splice(4, 1, availableTopics[selectedIndex]);
 		}
 	}
 
@@ -115,7 +113,7 @@ export const TopicFilterBank = ({
 					/>
 				)}
 
-				{availableTopics.slice(0, 5).map((topic) => {
+				{topFiveTopics.map((topic) => {
 					const buttonParams = `${topic.type}:${topic.value}`;
 					const isActive =
 						selectedTopic !== undefined &&
--- a/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
+++ b/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
@@ -113,8 +113,7 @@ export const TopicFilterBank = ({
 				{topFiveTopics.map((topic) => {
 					const buttonParams = `${topic.type}:${topic.value}`;
 					const isActive =
-						selectedTopic !== undefined &&
-						isEqual(selectedTopic)(topic);
+						!!selectedTopic && isEqual(selectedTopic)(topic);
 
 					return (
 						<FilterButton
--- a/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
+++ b/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
@@ -5,7 +5,7 @@ import { FilterButton } from './FilterButton.importable';
 
 type Props = {
 	availableTopics: Topic[];
-	selectedTopics?: string;
+	selectedTopics?: Topic[];
 	format: ArticleFormat;
 	filterKeyEvents: boolean;
 	keyEvents?: Block[];
@@ -66,6 +66,11 @@ const handleKeyEventClick = (filterKeyEvents: boolean) => {
 	window.location.search = urlParams.toString();
 };
 
+const availableTopicIsSelected =
+	(selectedTopic: Topic) => (availableTopic: Topic) =>
+		availableTopic.type === selectedTopic.type &&
+		availableTopic.value === selectedTopic.value;
+
 export const TopicFilterBank = ({
 	availableTopics,
 	selectedTopics,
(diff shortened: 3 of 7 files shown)


dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -119,7 +119,11 @@ function getKey(
 			'filterKeyEvents',
 			filterKeyEvents ? 'true' : 'false',
 		);
-		if (selectedTopics) url.searchParams.set('topics', selectedTopics);
+		if (selectedTopics && selectedTopics.length > 0)
```
abeddow91, line 122: could we use `selectedTopics?.length` here?

dotcom-rendering/index.d.ts
```diff
@@ -1173,7 +1173,7 @@ type MatchReportType = {
 interface Topic {
 	type: TopicType;
 	value: string;
-	count: number;
+	count?: number;
```
alinaboghiu, line 1176: sorry if you already explained this and I missed it but how come this now optional? Do we not always know a count?

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
```diff
@@ -74,6 +79,20 @@ export const TopicFilterBank = ({
 	filterKeyEvents = false,
 }: Props) => {
 	const palette = decidePalette(format);
+
+	const selectedTopic = selectedTopics?.[0];
+	const topFiveTopics = availableTopics.slice(0, 5);
+
+	if (selectedTopic) {
+		const selectedIndex = availableTopics.findIndex(
```
alinaboghiu, line 87: could move this inside the if statement below as we only need it if selectedIntex > 4

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
```diff
@@ -66,6 +66,11 @@ const handleKeyEventClick = (filterKeyEvents: boolean) => {
 	window.location.search = urlParams.toString();
 };
 
+const availableTopicIsSelected =
+	(selectedTopic: Topic) => (availableTopic: Topic) =>
```
joecowton1, line 70: I'm not sure there's a need to split the arguments like this? Passing them in a single parenthesis would be more idiomatic to DCR / React.

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
```diff
@@ -66,6 +66,11 @@ const handleKeyEventClick = (filterKeyEvents: boolean) => {
 	window.location.search = urlParams.toString();
 };
 
+const availableTopicIsSelected =
```
alinaboghiu, line 69: I'm tempted to name this isEqual as it's just defining whether we consider two topics to be the same?

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
```diff
@@ -74,6 +79,20 @@ export const TopicFilterBank = ({
...
+	const selectedTopic = selectedTopics?.[0];
+	const topFiveTopics = availableTopics.slice(0, 5);
+
+	if (selectedTopic) {
+		const selectedIndex = availableTopics.findIndex(
+			availableTopicIsSelected(selectedTopic),
+		);
+
+		if (selectedIndex > 4) {
+			topFiveTopics.splice(4, 1, availableTopics[selectedIndex]);
```
alinaboghiu, line 92: neat!

dotcom-rendering/index.d.ts
marjisound: In DCR we are going to use the same type for both `availableTopics` & `selectedTopics` as part of this PR https://github.com/guardian/frontend/pull/25169. I think because for the `selectedTopics` is coming without a `count` 

Though, I think it's possible for us to make the count required. @joecowton1 is now working on that part in frontend to update the DotcomRenderingModel for `selectedTopic` .  So if that's all OK, I'll make the count in here as required too. But please shout if that'd be a problem

alinaboghiu (approved):

dotcom-rendering/index.d.ts
alinaboghiu: Makes sense, was just curious to understand, thanks

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
alinaboghiu: agree, might be simpler to read

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
marjisound: I may not have fully understood what you mean. Here we're getting the `selectedIndex` and in the if statement below, it's using the value of the `selectedIndex` as a condition.

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
```diff
@@ -94,11 +113,11 @@ export const TopicFilterBank = ({
...
 				)}
 
-				{availableTopics.slice(0, 5).map((topic) => {
+				{topFiveTopics.map((topic) => {
 					const buttonParams = `${topic.type}:${topic.value}`;
-					const isActive = selectedTopics
-						? selectedTopics.includes(buttonParams)
-						: false;
+					const isActive =
+						selectedTopic !== undefined &&
```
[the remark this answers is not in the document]
marjisound: The type of the `selectedTopic` is `Topic | undefined` so I don't think it can be anything other than these 2 types, right?

further up in the code I'm doing this: `const selectedTopic = selectedTopics?.[0];` which will return `Topic | undefined`

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
marjisound: The main reason I did it like this was to be able to use it more easily in the array methods like fineIndex. 

`availableTopics.findIndex(availableTopicIsSelected(selectedTopic));`

Changing it to have 2 params would mean we can't use it without providing the second param straight away and therefore this:
```
const selectedIndex = availableTopics.findIndex((availableTopic) =>
	isEqual(selectedTopic, availableTopic),
);
```

I'm still happy to change if you think it'd make the code simpler. Let me know what you think please :)

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
marjisound: great suggestion, do you also think the param names should change? Or `selectedTopic` & `availableTopic` are fine?

abeddow91: +1 to isEqual but I think we can keep the param names as is.

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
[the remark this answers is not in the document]
joecowton1: Ah ok sure, I'd missed that bit

joecowton1 (approved):

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
joecowton1: I think i'd still probably break it down a bit, something like this is quite hard to digest unless you already know what it's doing:
```
availableTopics.findIndex(availableTopicIsSelected(selectedTopic));
```

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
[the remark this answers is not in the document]
marjisound: I just changed it to `!!selectedTopic && isEqual(selectedTopic)(topic);` . This is hopefully nicer to read?

abeddow91 (approved):
Really nice work!!

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
marjisound: Yea I agree, just changed it 👍

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
```diff
@@ -94,11 +110,10 @@ export const TopicFilterBank = ({
...
 				)}
 
-				{availableTopics.slice(0, 5).map((topic) => {
+				{topFiveTopics.map((topic) => {
 					const buttonParams = `${topic.type}:${topic.value}`;
-					const isActive = selectedTopics
-						? selectedTopics.includes(buttonParams)
-						: false;
+					const isActive =
+						!!selectedTopic && isEqual(selectedTopic)(topic);
```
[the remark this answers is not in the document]
marjisound: Thanks Anna for the suggestion, it looks a lot nicer with `!!`

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
joecowton1: Thanks Marjan

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
```diff
@@ -74,6 +78,20 @@ export const TopicFilterBank = ({
...
+
+	if (selectedTopic) {
+		const selectedIndex = availableTopics.findIndex((availableTopic) =>
+			isEqual(selectedTopic, availableTopic),
+		);
+
+		if (selectedIndex > 4) {
+			topFiveTopics.splice(4, 1, availableTopics[selectedIndex]);
+		}
+	}
```
mxdvl, line 93: This code is not following the React lifecycle, and should probably use hooks rather than splicing data!

marjisound: Oh really? thanks for letting us know. I'll create a ticket in our board to look into it 👍

mxdvl: If it works, I guess it’s not a major necessity, but generally, you would use a useEffect hook to compute this value.

However, in this case, I’m wondering if it would be beneficial to simply create the the array in one pass, which would 

```ts
const relevantTopics = availableTopics.filter(({ count }) => (count ?? 0) > 2);
/** take all the selected topics,
 * append the relevant remaining ones
 * and keep the first five
 */
const topFiveTopics = selectedTopics.concat(relevantTopics).slice(0, 5);
```

mxdvl: Here’s the React docs on mutating data: https://react.dev/learn/updating-arrays-in-state#updating-arrays-without-mutation

arelra: Wrt to using hooks I think this looks okay as the state is just derived state. Afaics the state is not React owned state managed by useState or a useEffect dependency with the expectation of updates that will trigger a re-render.

I think the docs are aimed at when arrays are used as React state and are being mutated in place which is 'bad' as mutations then will not trigger re-renders.

Agree however we should not mutate the array in place to avoid any side effects in JS or React if the state is used differently in future.

PR #5291, opened 2022-06-30
Dblatcher/consume newsletter data

dblatcher:
## What does this change?
 - Defines a `CAPIElement` subtype of  `NewsletterSignupBlockElement` 
 - Refactors `EmailSignup` to take a single  `Newsletter` prop
- adds a case to `renderElement` to return an `EmailSignup` instance for the `NewsletterSignupBlockElement`

There is be no custom-facing changes directly from this PR as frontend will no be providing the `NewsletterSignupBlockElement` in requests yet.

## Why?

The plan is to replace replace the existing email embeds, which are manually inserted into articles using embed blocks and rendered as iframes containing a page served by frontend. 

### expected follow-up PR's
 - [frontend] : provide the newsletter data in the request to DCR when the article has an "include sign-up" tag
 - [DCR]: use the newsletter data from frontend (when present) to enhance the article content with a `NewsletterSignupBlockElement`


oliverlloyd (commented):
Would it be possible to break the styling changes in this PR out into their own PR, please? That will make reviewing easier 🙏 

Regarding the remaining changes. If the email sign up embed is content designed to be inserted in the article then perhaps it can be given to DCR as a proper content type? E.g.: it's own block element.

DCR is a rendering tier. It basically gets past an array of [content blocks](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/lib/content.d.ts#L193) and renders each one and then adds furniture to the page around this list of blocks. In most cases, the blocks are text and then sometimes you get other types of embeds. I think this use case is one of these embeds?

So we could add a new block element, defines the types it needs, and then build that data model in Frontend and pass this content to DCR. I think that will make the task of adding support for these elements to AR easier as well

dblatcher:
> Regarding the remaining changes. If the email sign up embed is content designed to be inserted in the article then perhaps it can be given to DCR as a proper content type? E.g.: it's own block element.
> 
> DCR is a rendering tier. It basically gets past an array of [content blocks](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/lib/content.d.ts#L193) and renders each one and then adds furniture to the page around this list of blocks. In most cases, the blocks are text and then sometimes you get other types of embeds. I think this use case is one of these embeds?

A difficulty with that is we've been told that the sign-up blocks are *not* content and should not be part of the article in the same way that the current embeds are - that was why were told not to use atoms. 

Representing them in the data model as content blocks with a predetermined position in the article seems contrary to that. They seem more akin to furniture than content. 

If I'm understanding right, for the sign-up block to be represented within the array of content blocks, frontend would need to take the array of actual content elements from CAPI, and (if a sign-up tag is present) decide at what point in the array to insert the sign up (pseudo) content block, then pass the modified array to DCR which must render an embed-content-block at the point decided by frontend. That would limit DCR's ability to make platform-specific decisions about how and where to render the block.

dblatcher:
I've changed round so the `EmailSignup` can be rendered as a content block, with only the data fields needed  - as discussed, we can have frontend determine the position of the block within the array of CAPIElements.

dotcom-rendering/src/web/components/SecureSignup.tsx
```diff
@@ -54,7 +54,7 @@ const PrivacyTerms = () => {
 
 const RecaptchaTerms = () => (
 	<span css={termsStyle}>
-		We operate Google reCaptcha to protect our website and the Google{' '}
+		We use Google reCaptcha to protect our website and the Google{' '}
```
oliverlloyd, line 57: We show this for legal reasons; do we need to confirm what the text should be? I notice it reads as something else on [other pages](https://profile.theguardian.com/signin) on the site

dotcom-rendering/src/lib/content.d.ts
```diff
@@ -286,6 +286,11 @@ interface MultiImageBlockElement {
 	role?: RoleType;
 }
 
+interface NewsletterSignupBlockElement {
+	_type: 'model.dotcomrendering.pageElements.NewsletterSignupBlockElement';
+	newsletter: NewsletterData;
```
oliverlloyd, line 291: Can we inline these properties here? The other block element types mostly follow this same convention and it would be good to keep this pattern consistent

dotcom-rendering/src/lib/content.d.ts
```diff
@@ -286,6 +286,11 @@ interface MultiImageBlockElement {
 	role?: RoleType;
 }
 
+interface NewsletterSignupBlockElement {
+	_type: 'model.dotcomrendering.pageElements.NewsletterSignupBlockElement';
```
oliverlloyd, line 290: I wonder if we will need an `elementId` here? This is a [property that gets added](https://github.com/guardian/frontend/blob/43fd3eed9641e23211dc8e1b743120eca362040f/common/app/model/dotcomrendering/PageElement-Identifiers.md) by Frontend at some point so that we can have a unique id

oliverlloyd (approved):
This very much feels the right approach for introducing this element 👍 

We now have the option to add the block element in Frontend or we could also use an enhancer to do the same on DCR. I think both at valid although my instinct would be for it to be an enhancer as this allows us more control locally which might work better with personalisation

dotcom-rendering/src/web/components/SecureSignup.tsx
dblatcher: Our product manager got an email from data privacy saying we could replace 'operate' with 'use' here (the original question was to the effect of "can we make this text shorter?").  The google's recommended text is here: https://developers.google.com/recaptcha/docs/faq#id-like-to-hide-the-recaptcha-badge.-what-is-allowed

dotcom-rendering/src/lib/content.d.ts
dblatcher: Sounds good - I'm keen to keep `NewsletterData` as a type so it aligns with the object we are going to generate in Frontend, so I used `interface NewsletterSignupBlockElement extends NewsletterData` - hope that's ok.

oliverlloyd: I actually think in this case it might be better to not try to align. The reason I say this is because we have no solution at the moment for sharing types between Frontend and DCR, this is a known issue and we've discussed various options (GraphQL, Thrift, etc.). This is very much a long term project and until we have a code based solution it is not possible to couple types between the two platforms.

Of course, that doesn't stop us from creating one type in Frontend and the same one in DCR but I would say doing this - without technical constraints - would create more problems. In that scenario it is very easy to edit one or the other type and for the other one to get out of alignment. Once this happens we will have code that is 'meant to be' the same but isn't 😬 

Instead of this, in DCR, we consciously made the choice to have completely different models. This is why all the other element blocks have types that are subsets or constructions formed of other data from Frontend. Like this you are free to make changes at either end without creating confusion.

Note. This was the route we took for DCR. On the apps rendering project they went the _other_ way, they pass in the data from CAPI directly to the rendering tier, so it's not that one method is right or wrong, just that this is the convention DCR follows.

Sorry for the long rambling comment!

dotcom-rendering/src/web/components/SecureSignup.tsx
oliverlloyd: Perfect, thanks for the extra detail!

PR #5294, opened 2022-07-01
Inject test recipe schema tags into specified articles

frederickobrien:
This PR implements the plumbing for an experiment Ophan wants to run in which structured data is injected onto a handful of recipe pages so we can gauge a before and after of SEO performance. Following on from a chat with @shtukas this seemed like a clean, safe way to test the concept without having to implement anything too widespread.

This works by adding a temporary `recipeSchema` object which contains recipe page urls as keys and [recipe schema](https://developers.google.com/search/docs/advanced/structured-data/recipe) as values. `articleToHtml.tsx` checks whether the page it's rendering matches one of the url keys in the `recipeSchema` object. If it does, it injects a script tag containing the corresponding schema. If it doesn't nothing is rendered. Either way nothing changes visually on the page. The schema is under the hood purely for the benefit of web crawlers and the like.

If this approach is approved and shown to work with the example used (using [this testing tool](https://search.google.com/test/rich-results) on [the guinea pig recipe](https://www.theguardian.com/lifeandstyle/wordofmouth/2013/nov/07/how-to-make-perfect-apple-pie)), I would follow up with a PR adding semantic markup to 5-10 more recipes in the test object.

### DIFF
--- a/dotcom-rendering/src/web/experiments/tests/recipe-structured-data.ts
+++ b/dotcom-rendering/src/web/experiments/tests/recipe-structured-data.ts
@@ -1,0 +1,68 @@
+// This is for an experiment testing SEO performance on recipes before
+// and after structured data is added. Please don't delete - it will be
+// removed once the the experiment is finished, we swear!
+
+export type RecipeSchemaObject = {
+	[url: string]: string;
+};
+
+export const recipeSchema: RecipeSchemaObject = {
+	'https://www.theguardian.com/lifeandstyle/wordofmouth/2013/nov/07/how-to-make-perfect-apple-pie': `{
+		"@context":"https://schema.org/",
+		"@type":"Recipe",
+		"name":"The perfect hamburger",
+		"image":[
+		   "https://i.guim.co.uk/img/static/sys-images/Guardian/Pix/pictures/2010/8/3/1280851276363/Perfect-hamburger-006.jpg?width=700&quality=45&auto=format&fit=max&dpr=2&s=32ffe62eaad00a538d50aedc097237d7"
+		],
+		"author":{
+		   "@type":"Person",
+		   "name":"Felicity Cloake"
+		},
+		"datePublished":"2010-08-05",
+		"description":"Burgers may be fast food, but they're also a craft.",
+		"prepTime":"PT80M",
+		"cookTime":"PT100M",
+		"totalTime":"PT180M",
+		"keywords":"burgers, barbecue",
+		"recipeYield":"6",
+		"recipeCategory":"Main",
+		"recipeCuisine":"American",
+		"nutrition":{
+		   "@type":"NutritionInformation",
+		   "calories":"1000 calories"
+		},
+		"recipeIngredient":[
+		   "1 tbsp oil or butter",
+		   "1 large onion, finely chopped",
+		   "1kg roughly minced chuck steak (or any non-lean mince)",
+		   "100ml stout",
+		   "2 tbsp brown breadcrumbs",
+		   "2 tsp chopped herbs (parsley or thyme work well)",
+		   "1 tsp salt",
+		   "Black pepper",
+		   "Garnishes, sauces and rolls, as desired"
+		],
+		"recipeInstructions":[
+		   {
+			  "@type":"HowToStep",
+			  "name":"Brown the onion",
+			  "text":"Heat the oil in a frying pan over a low heat, and cook the onion until soft and slightly browned. Leave to cool."
+		   },
+		   {
+			  "@type":"HowToStep",
+			  "name":"Prepare the meat for cooking",
+			  "text":"Spread the beef out and sprinkle over the onion. Add the stout, breadcrumbs, herbs and seasoning and mix together with a fork, being careful not to overwork it."
+		   },
+		   {
+			  "@type":"HowToStep",
+			  "name":"Divide into burgers",
+			  "text":"Divide the meat into 12 flattish burgers, putting a dimple in the centre of each. Cover and refrigerate for an hour."
+		   },
+		   {
+			  "@type":"HowToStep",
+			  "name":"Cook the burgers!",
+			  "text":"Cook the burgers on a medium to hot barbecue or griddle pan: leave them undisturbed for the first 3 minutes so they build up a good seal on the bottom, then carefully turn them over, adding a slice of cheese on top if desired. Cook for a further 4 minutes for rare, and 7 for well done, and allow to rest for a few minutes before serving. (You can toast buns, cut-side down, on the barbecue at this point.)"
+		   }
+		]
+	 }`,
+};
--- a/dotcom-rendering/src/web/server/articleTemplate.ts
+++ b/dotcom-rendering/src/web/server/articleTemplate.ts
@@ -19,6 +19,7 @@ export const articleTemplate = ({
 	twitterData,
 	keywords,
 	initTwitter,
+	recipeMarkup,
 }: {
 	title?: string;
 	description: string;
@@ -35,6 +36,7 @@ export const articleTemplate = ({
 	twitterData: { [key: string]: string };
 	keywords: string;
 	initTwitter?: string;
+	recipeMarkup?: string;
 }): string => {
 	const favicon =
 		process.env.NODE_ENV === 'production'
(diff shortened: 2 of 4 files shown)


oliverlloyd:
> Would this experiment work better as an AB test? I think to really know if the change your proposing has an impact you need to compare like for like using a control and variant.

Or, if the objective here is to simply verify that the page delivers the correct structure as needed by Google using their test tool then could we perhaps achieve this by using `code`? If so, we would be able to avoid a production deployment. Sorry if I'm miss-understanding what's needed here

dotcom-rendering/src/web/experiments/tests/recipe-structured-data.ts
```diff
@@ -0,0 +1,68 @@
+// This is for an experiment testing SEO performance on recipes before
```
oliverlloyd, line 1: Could we put this file in a different location, please? This `experiments/` folder is currently only used for AB tests and there is [code which uses these files](https://github.com/guardian/dotcom-rendering/blob/a401bd0fbb839916af4fade22e8d1a3b2974b373/dotcom-rendering/src/web/components/SetABTests.importable.tsx) and expects them to be structured in a certain way.

I would suggest placing it closer to where it is being used and perhaps giving it a very explicit name to indicate that it is to be deleted

dotcom-rendering/src/web/experiments/tests/recipe-structured-data.ts
```diff
@@ -0,0 +1,68 @@
...
+// and after structured data is added. Please don't delete - it will be
```
oliverlloyd, line 2: Can we add an explicit date from when this file can be deleted. It is astounding how easy it is for code like this to remain for years even when everyone has the best intentions, it happens. I am guilty of this, it is almost impossible to prevent. It is about 500M times easier to add code than it is to delete it.

dotcom-rendering/src/web/experiments/tests/recipe-structured-data.ts
```diff
@@ -0,0 +1,68 @@
...
+// removed once the the experiment is finished, we swear!
+
+export type RecipeSchemaObject = {
```
oliverlloyd, line 5:
```suggestion type RecipeSchemaObject = {
```

I can't see this type being used elsewhere so I think we're safe not exporting it

dotcom-rendering/src/web/experiments/tests/recipe-structured-data.ts
```diff
@@ -0,0 +1,68 @@
...
+	[url: string]: string;
+};
+
+export const recipeSchema: RecipeSchemaObject = {
```
oliverlloyd, line 9: Thinking some more, do we need this type? I'm wondering if having it has any value 🤔

frederickobrien:
> > Sorry, bit of a marauding answer
> 
> It's an amazing answer 😻 I completely understand now, thanks
> 
> You're right, we typically use AB tests to capture in page behaviour and SEO performance is decided out of the browser so would not be a good fit here.
> 
> Would the Google test tool be a good way to verify that the changes here are valid (I presume the tool only checks the page content and isn't crawling the web)? If so, could we use `code` to do that check and then if it looks good we can deploy to production?

Aw thanks! I think that tool's a safe bet. If it's happy with the markup then that's Google effectively saying, 'this qualifies', which is something we can count on their web crawlers echoing when the time comes.

Now you mention it, yes, code should be good for checking. Either we give it a url to parse or if that's not possible it lets you paste in code, so I could copy and paste the entire rendered HTML from `code`. That would be ideal actually, as our thinking was to check this approach is viable then work with the food desk to write up a good test batch of recipes and add them after. I originally suggested adding those in a followup PR, but maybe we confirm it works in `code`, add the other recipes in another commit, then deploy all under the umbrella of this PR.

What do you reckon?

dotcom-rendering/src/web/server/articleTemplate.ts
```diff
@@ -275,6 +277,7 @@ https://workforus.theguardian.com/careers/product-engineering/
 					${initTwitter}
 				</script>
 
+				${recipeMarkup}
```
oliverlloyd, line 280: Could we move this tag into the body along with the low priority tags? I am assuming that the meta content is read by crawlers which will wait for the page to load?

oliverlloyd, line 280: I think it would be easier to read this file and follow the conventions here if we pass the raw json as the string parameter and set the script tags explicitly here in `articleTemplate`.

dotcom-rendering/src/web/server/articleToHtml.tsx
```diff
@@ -207,6 +208,13 @@ export const articleToHtml = ({ data }: Props): string => {
 		}(document, "script", "twitter-wjs"));
 	`;
 
+	const url = CAPIArticle.webURL;
+
+	const recipeMarkup =
+		url in recipeSchema
+			? `<script type="application/ld+json">${recipeSchema[url]}</script>`
```
oliverlloyd, line 215: Could this live in `articleTemplate`? There's a number of examples of similar code being done inline

oliverlloyd (commented):
Apologies for the large volume of comments here but these two files are core components of the site so it's good to be thorough!

In terms of removing this code. Something I've seem other people do when creating PRs that they know will be removed later is to create the revert PR at the same time. That way, so long as there have been no major changes in this area since then, you can just merge the revert PR to remove the code. What do you think?

oliverlloyd:
> Now you mention it, yes, code should be good for checking. Either we give it a url to parse or if that's not possible it lets you paste in code, so I could copy and paste the entire rendered HTML from `code`. That would be ideal actually, as our thinking was to check this approach is viable then work with the food desk to write up a good test batch of recipes and add them after. I originally suggested adding those in a followup PR, but maybe we confirm it works in `code`, add the other recipes in another commit, then deploy all under the umbrella of this PR.
> 
> What do you reckon?

Using code to check would be ideal. That's a very low risk thing to do and it sounds like it will provide some good information 👍 (You could also ignore all my comments that way too!)

Somewhat of a tangent: How would the workflow look like for content authors adding this metadata to articles? Would this need a change to Composer or something else?

dotcom-rendering/src/web/experiments/tests/recipe-structured-data.ts
frederickobrien: Yeah great shout, my original was too vague. How many 'Will delete later we swear!' comments there must be in the world's code!

dotcom-rendering/src/web/experiments/tests/recipe-structured-data.ts
frederickobrien: TypeScript kicks up a fuss without it I'm afraid

oliverlloyd: Gotcha

oliverlloyd (approved):
Really nice job on this. You handled _lots_ of input from multiple people really well 👏

oliverlloyd:
> I'm a little rusty on deploying Dotcom to `code`. Is it just a case of deploying on Riff Raff and going to a certain url? If so, what's the url?

First here
Screenshot 2022-07-05 at 16 22 44

Then

https://riffraff.gutools.co.uk/deployment/request

and then

https://m.code.dev-theguardian.com/

🚀

frederickobrien:
By god, I think it works:

image

This is the [Rich Results Test](https://search.google.com/test/rich-results/result?id=0mVfwEndJ2uB1ycB9wtslQ) tool reading the rendered HTML of [the example page](https://www.theguardian.com/lifeandstyle/wordofmouth/2013/nov/07/how-to-make-perfect-apple-pie). I think we might be in good shape to add further recipes and see what happens...

frederickobrien:
Hi @oliverlloyd! Just wanted to check in on this again as I've updated the PR with a bunch of recipe schema for the experiment and we're hoping to give it a go. I've checked the affected pages in `CODE` and all seems well. Would we good to merge this tomorrow morning?

oliverlloyd:
> Hi @oliverlloyd! Just wanted to check in on this again as I've updated the PR with a bunch of recipe schema for the experiment and we're hoping to give it a go. I've checked the affected pages in `CODE` and all seems well. Would we good to merge this tomorrow morning?

Yep 🚢

PR #5295, opened 2022-07-04
fix(tsc): add id to TopicFilterBank

jamesgorrie:
## What does this change?
Adds the required `id` prop to `TopicFilterBank`. 

## Why?
To make things type right.


jamesgorrie:
Ping @abeddow91 for some 👀 as I'm not familiar with this piece of the codebase.

abeddow91:
Thanks @jamesgorrie! Looks perfect

PR #5296, opened 2022-07-04
Performance monitoring: Islands

mxdvl:
## What does this change?

Refactor `initPerf` to simplify and shorten the code. Stop logging willy-nilly for data that we **never look at**.

Log islands performance info: how long it took then to render or hydrate, how long to download their JavaScript.

Chore: bump lockfiles.

## Why?

We need more fine-grained information about our Islands hydration. In order to know which ones are performing best and worse, and start enforcing budget, we need information broken down by Island.

## Related

- #3629
- #4661

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/177115119-61656d16-ab97-4971-9649-2bdbd52d2b80.png
[after]: https://user-images.githubusercontent.com/76776/177115291-58911327-41b1-481c-a962-0225f0c36474.png

PR #5298, opened 2022-07-04
Fix Island performance logs

mxdvl:
## What does this change?

Remove useless and innaccurate measurements in favour of the [`duration` property](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry/duration), which is supported in all modern browsers.

Fall back to a invalid duration (`-1`)if it was not found.

## Why?

The measurements in #5296 were wrong! The dowload time started from time origin.

## Screenshots

| When | logs |
|--|--|
| Before | ![before][] |  |
| After | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/177144978-9db7212b-77a0-45c0-b08c-2ccfc6bd4177.png
[after]: https://user-images.githubusercontent.com/76776/177145247-51ac2012-487a-4e5d-89ca-f5a75aa4619e.png

### DIFF
--- a/dotcom-rendering/src/web/browser/islands/doHydration.ts
+++ b/dotcom-rendering/src/web/browser/islands/doHydration.ts
@@ -49,12 +49,10 @@ export const doHydration = (
 		})
 		.then(({ clientOnly, timeTaken }) => {
 			// Log performance info
-			const entry = window.performance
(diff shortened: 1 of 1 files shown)


OllysCoding (approved):
+1

PR #5299, opened 2022-07-04
Fix when performance.measure is undefined

mxdvl:
Many errors have started showing in Sentry since
the island performance changes have been introduced.

Looking for the existence of `measure` in window.performance should
be a better indicator of support than simply `mark`.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |


OllysCoding (approved):
lgtm!

PR #5301, opened 2022-07-04
Fix spacing for pinned post and key events carousel

silvacb:
## What does this change?
Amends spacing within Pinned Post and Key Events Carousel.

## Why?
As part of the automatic filters redesign, the spacing between the Pinned Post and the Key Events Carousel is changing to 24px.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/55602675/177186106-c4d8037c-9995-45ee-9328-19dad59223e9.png

[after]: https://user-images.githubusercontent.com/55602675/177186189-2f4bf16e-40e9-4d59-b111-44cf03d014f6.png

### DIFF
--- a/dotcom-rendering/src/web/components/PinnedPost.tsx
+++ b/dotcom-rendering/src/web/components/PinnedPost.tsx
@@ -20,9 +20,14 @@ import { decidePalette } from '../lib/decidePalette';
 const pinnedPostContainer = (palette: Palette) => css`
 	border: 3px solid ${palette.border.pinnedPost};
 	padding-bottom: ${space[1]}px;
-	margin-bottom: ${space[9]}px;
 	position: relative;
 	background: ${neutral[100]};
+	${from.mobile} {
+		margin-bottom: ${space[6] + space[5] / 2}px;
+	}
+	${from.desktop} {
+		margin-bottom: ${space[9]}px;
+	}
 	#pinned-post-checkbox:checked ~ #collapsible-body {
 		max-height: fit-content;
 		margin-bottom: ${space[1]}px;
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/PinnedPost.tsx
```diff
@@ -20,9 +20,14 @@ import { decidePalette } from '../lib/decidePalette';
 const pinnedPostContainer = (palette: Palette) => css`
 	border: 3px solid ${palette.border.pinnedPost};
 	padding-bottom: ${space[1]}px;
-	margin-bottom: ${space[9]}px;
 	position: relative;
 	background: ${neutral[100]};
+	${from.mobile} {
+		margin-bottom: ${space[6] + space[5] / 2}px;
```
abeddow91, line 26: This feels a little bit unorthodox. `Space` represents the guardian space conventions so we should be able to use them but if not, its preferable to favour the raw px rather than create a calculation

dotcom-rendering/src/web/components/PinnedPost.tsx
```diff
@@ -20,9 +20,14 @@ import { decidePalette } from '../lib/decidePalette';
...
+	}
+	${from.desktop} {
```
abeddow91, line 28: Nice!

dotcom-rendering/src/web/components/PinnedPost.tsx
silvacb: Thanks for the comment Anna, I changed this to raw pixels.

alinaboghiu (approved):
Makes sense!

abeddow91 (approved):
All looks good! Great work 🥳

OllysCoding (approved):

PR #5307, opened 2022-07-05
Dblatcher/footer promotes daily newsletter

dblatcher:
## What does this change?
The link in the footer points to the sign-up pages for the curated daily newsletters (First Edition(UK), Morning Mail(AUS) and First Thing(US)) instead of the guardian headlines newsletters.

Small css change so when the pillar links wrap on (very) narrow screen, the second row will not float to the left of the containers below it.

## Why?

duplicates the change to frontend: https://github.com/guardian/frontend/pull/25182

## Screenshots

| Before      | After      |
|-------------|------------|
| www theguardian com_australia-news_2022_jul_04_cargo-ship-stranded-off-sydney-nsw-coast-portland-bay-bulk-carrier | localhost_3030_Article_url=https___www theguardian com_australia-news_2022_jul_04_cargo-ship-stranded-off-sydney-nsw-coast-portland-bay-bulk-carrier|
| Screenshot 2022-07-05 at 14 17 01 | Screenshot 2022-07-05 at 14 18 39  |

### DIFF
--- a/dotcom-rendering/src/web/components/Footer.tsx
+++ b/dotcom-rendering/src/web/components/Footer.tsx
@@ -60,10 +60,10 @@ const emailSignup = css`
 		width: 247px;
 	}
 	${between.leftCol.and.wide} {
-		width: 300px;
+		width: 325px;
 	}
 	${from.wide} {
-		width: 460px;
+		width: 498px;
 	}
 `;
 
--- a/dotcom-rendering/src/web/components/Footer.tsx
+++ b/dotcom-rendering/src/web/components/Footer.tsx
@@ -239,13 +239,13 @@ const year = new Date().getFullYear();
 const decideSignupLink = (edition: EditionId): string => {
 	switch (edition) {
 		case 'US':
-			return 'https://www.theguardian.com/info/2015/dec/08/daily-email-us';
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/components/Footer.tsx
```diff
@@ -236,16 +237,31 @@ const FooterLinks = ({
 
 const year = new Date().getFullYear();
 
+// Hard coding the values here is not ideal - ideally, they would sourced from the
+// newsletters API by frontend and included in the request.
```
oliverlloyd, line 241:
```suggestion
```

I think it's okay to have these values hard coded here. We do this in a lot of other places so it's an established pattern

oliverlloyd (approved):
👍

dotcom-rendering/src/web/components/Footer.tsx
mxdvl: I agree with Oliver that updating `DotcomRenderingDataModel` in `frontend` to pass these is not a path you need to go down just yet.

I think the comment is helpful, though, as it can help us weigh the number of use-cases for these kind of uses.

PR #5311, opened 2022-07-05
Fix sponsor logo sizes

### LINKED ISSUE #5309: Add max-width to sponsor logo in interactive articles
Currently in an interactive article the meta component is rendered in the main content column (not left column like usual). This means the sponsor logo appears huge.

Can we add a `max-width: 140px;` in this instance

https://www.theguardian.com/football/ng-interactive/2022/jul/05/david-squires-on-euro-2022-coming-to-merry-old-england

mxdvl:
## What does this change?

Fix the sponsor logo size to a maximum of 140.

Reduce CLS by setting a explicit height. If the logo is oversized, do the math to scale it down.

## Why?

Fixes #5309 

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/177353477-457d95c8-f1d0-4312-8333-a1e4d2b80d4d.png
[after]: https://user-images.githubusercontent.com/76776/177353509-a0ee6bce-79db-4c85-acd0-4caae5432a49.png

### DIFF
--- a/dotcom-rendering/src/web/components/Branding.importable.tsx
+++ b/dotcom-rendering/src/web/components/Branding.importable.tsx
@@ -1,5 +1,5 @@
 import { css } from '@emotion/react';
-import { neutral, textSans, until } from '@guardian/source-foundations';
+import { neutral, textSans } from '@guardian/source-foundations';
 import { trackSponsorLogoLinkClick } from '../browser/ga/ga';
 
 const brandingStyle = css`
@@ -12,14 +12,11 @@ const brandingLabelStyle = css`
 `;
 
 const brandingLogoStyle = css`
-	${until.phablet} {
-		max-width: 140px;
-	}
-	max-width: 280px;
-	width: 100%;
 	padding: 10px 0;
+
+	display: block;
 	img {
-		max-width: 100%;
+		display: block;
 	}
 `;
 
@@ -33,11 +30,14 @@ const brandingAboutLink = (palette: Palette) => css`
 	}
 `;
 
-export const Branding: React.FC<{
+type Props = {
 	branding: Branding;
 	palette: Palette;
-}> = ({ branding, palette }) => {
-	if (!branding) return null;
+};
+
+export const Branding = ({ branding, palette }: Props) => {
+	const sponsorId = branding.sponsorName.toLowerCase();
+
 	return (
 		<div css={brandingStyle}>
 			<div css={brandingLabelStyle}>{branding.logo.label}</div>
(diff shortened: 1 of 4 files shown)


dotcom-rendering/src/web/components/Branding.importable.tsx
```diff
@@ -33,11 +30,28 @@ const brandingAboutLink = (palette: Palette) => css`
...
+const getDimensions = ({
+	logo: { dimensions },
+}: Branding): { width: number; height: number } => {
+	if (dimensions.width <= 140) return dimensions;
+	const scaling = 140 / dimensions.width;
+	return {
+		width: 140,
+		height: Math.round(dimensions.height * scaling),
+	};
+};
```
oliverlloyd, line 42: 🤔 Having javascript just for a logo feels wrong to me. I can see that we enabled the execution of javascript on this image so that we can track clicks. I think based on some recent conversations and where our priorities are going we might want to revisit that and maybe there are better ways to do this tracking without the need for another island.

This is in theory irrelevant to your change but I see we are now adding more javascript which would make future simplifications harder to do. Is there a way to fix this width problem using just css?

mxdvl: Oh no, I think it’s absolutely bonkers to load react to track a single click! But that’s for another PR :)

mxdvl: Wholehearteadly agree, I had difficulty looking the other way as this is so wasteful! This is for another PR, however.

I would rather we inlined the function run in `onClick` rather than hydrate this part of the DOM with Preact and run some Emotion. Making this an Island is objectively the wrong approach.

mxdvl: This does not make future refactoring any harder, as this component shold be server rendered. As you can see in the bundle analyser, we’re actually getting a smaller island with this code, because Preact + Emotion + Source foundations are actually a lot larger than my addition!

jamesgorrie: > 🤔 Having javascript just for a logo feels wrong to me.

By javascript - do you mean client-side javascript? From what I understand in this PR is this JS is only run on the server.

If we do keep this method, could we call it something like `limitWidthToMax140` or something describing what it does?

oliverlloyd: An enhancer 😻 !

bryophyta (commented):
I'm not familiar with this part of the codebase, but this looks good to me 👍

dotcom-rendering/src/web/components/Branding.importable.tsx
bryophyta: +1 to inlining (as a separate PR)

bryophyta (approved):

PR #5312, opened 2022-07-05
Updated top right ad size reference

emma-imber:
## What does this change?
Changed the way that the top right ads get their sizes to use a standardised pre-defined array instead of cherry picking sizes.
## Why?
Keeping this in line with the way that other ads fetch their sizing info.


Jakeii (approved):
Nice!

arelra (approved):
👍

PR #5314, opened 2022-07-05
Add tooltip to `FilterKeyEventsToggle`

silvacb:
## What does this change?
Passes a tooltip prop to the FilterKeyEventsToggle.
Sets the tooltip to true in `LiveLayout` and `LiveBlogRenderer`.

## Why?
Some users may have JS disabled - in this case we want to display a tooltip when a user clicks on the filter key events toggle.

### Desktop
Before: 

https://user-images.githubusercontent.com/55602675/177370779-b0dcc3c7-ace5-4270-8269-f07095f1397e.mov

After: 

https://user-images.githubusercontent.com/55602675/177370837-c4cdbb1a-5387-4ba3-8826-4cd4d4212214.mov

### Mobile
Before: 

https://user-images.githubusercontent.com/55602675/177371801-7805e808-c399-41a8-861c-095b319673dc.mov

After: 

https://user-images.githubusercontent.com/55602675/177371872-741f61ad-4f99-476c-bff7-5be72c3c3a56.mov

### DIFF
--- a/dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
+++ b/dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
@@ -23,7 +23,6 @@ const toggleWrapperStyles = css`
 interface Props {
 	filterKeyEvents: boolean;
 	id: 'filter-toggle-mobile' | 'filter-toggle-desktop';
-	// tooltip: boolean;
 }
 
 export const FilterKeyEventsToggle = ({ filterKeyEvents, id }: Props) => {
--- a/dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
+++ b/dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
@@ -23,7 +23,7 @@ const toggleWrapperStyles = css`
 interface Props {
 	filterKeyEvents: boolean;
 	id: 'filter-toggle-mobile' | 'filter-toggle-desktop';
-	tooltip: true;
+	tooltip: boolean;
 }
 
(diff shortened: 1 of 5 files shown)


dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
```diff
@@ -23,6 +23,7 @@ const toggleWrapperStyles = css`
 interface Props {
 	filterKeyEvents: boolean;
 	id: 'filter-toggle-mobile' | 'filter-toggle-desktop';
+	tooltip: boolean;
```
joecowton1, line 26: I don't see how this is going to work: do you not need to pass this prop into the `ToggleSwitch ` component rather the the `FilterKeyEventsToggle` ?

joecowton1: Is it possible that when you were working on this that maybe you were still yarn linked to a version of Source with this hardcoded? 

As it is I think we'd need to update the version of Source development kitchen being used by DCR in the package.json of the parent repo.

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -839,6 +839,7 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 														CAPIArticle.filterKeyEvents
 													}
 													id="filter-toggle-desktop"
+													tooltip={true}
```
joecowton1, line 842: This means you can remove each of these too

silvacb: Done, thanks!

PR #5324, opened 2022-07-06
Fix:  hidden scrollbar on narrow desktop breakpoint for key events carousel

abeddow91:
## What does this change?
Show the horizontal scroll bar on the key events carousel until desktop breakpoint and add `padding-top` to the key events toggle. 

## Why?
Previously we hid the scroll bar as a design choice. However, this means that users without a trackpad who are viewing the key events carousel on desktop on a narrow breakpoint are unable to move left or right. To fix this, we are adding back in the visible scroll bar until desktop breakpoint so users can navigate via this. This PR also adds padding top to the key events toggle until desktop to account for the space now taken up by the scroll bar. 

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[after]: https://user-images.githubusercontent.com/20416599/177518218-bce37990-7773-4181-b24b-4b869a2b02c6.png
[before]: https://user-images.githubusercontent.com/20416599/177518237-b355bbc6-6a3f-40ee-8b2a-333dd2485976.png

### DIFF
--- a/dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
+++ b/dotcom-rendering/src/web/components/FilterKeyEventsToggle.importable.tsx
@@ -1,5 +1,5 @@
 import { css } from '@emotion/react';
-import { from, neutral, remSpace } from '@guardian/source-foundations';
+import { from, neutral, remSpace, until } from '@guardian/source-foundations';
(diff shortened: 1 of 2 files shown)


marjisound (approved):
Looks good to me 👍

PR #5327, opened 2022-07-06
Use DCR's layout components more in #5184

oliverlloyd:
## What does this change?
This PR is proposing change to another PR (https://github.com/guardian/dotcom-rendering/pull/5184), specifically it is proposing the use of DCR's own components, `ContainerLayout` & `ElementContainer` to decide the position of content

## Why?
The specific problem of how much space to have to the left of content at different breakpoints is well known and the solution to this problem has been abstracted and wrappers for this created by the DCR platform. This PR allows these centralised solutions to be used

### Why are we moving content around inside `InteractiveLayout`?
Initially I made comments on @tomrf1 's PR thinking this would be a simple change. But it turned out my simple proposals did not work. The reason was because the grid being used by the interactive layout is different, content is stretched further left and right, so the only way to make this solution work was to break the bottom parts of the article out of the grid and into their own containers.

Screenshot 2022-07-06 at 11 33 41

### DIFF
--- a/dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
+++ b/dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
@@ -24,30 +24,6 @@ import type {
 	CanShowData as RRCanShowData,
 	EpicConfig as RREpicConfig,
 } from './SlotBodyEnd/ReaderRevenueEpic';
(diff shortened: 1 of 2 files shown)


tomrf1 (approved):
thanks!

PR #5328, opened 2022-07-06
Docs: Add instructions for running CI server locally

bryophyta:
## What does this change?

It adds instructions for how to run the CI server locally.

## Why?

The server needs to be running in order to run Cypress tests locally, and currently this process doesn't seem to be documented.

### DIFF
--- a/dotcom-rendering/docs/testing.md
+++ b/dotcom-rendering/docs/testing.md
@@ -76,6 +76,7 @@ By using mocked data and endpoints, we increase the speed that tests execute at 
 ### How to run locally
 
 Running Cypress locally requires having a DCR server running. To run the CI server locally, you can run `make run-ci`. This command runs `make stop`, followed by `make build-ci` and `make start-ci`. You can run these separate commands yourself instead if you want more control over when the site gets re-built.
+The Cypress server should automatically re-load in response to changes in `.spec.` files, but the CI server will need to be re-built every time you want to update the DCR code itself, as opposed to the spec code. You can re-build by running `make ci-build`, or re-running `make ci-run`.
 
 To run Cypress in interactive mode (visually):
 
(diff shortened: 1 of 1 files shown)


dotcom-rendering/docs/testing.md
```diff
@@ -74,6 +74,9 @@ The down side to these types of tests is that they are slower and have a depende
 By using mocked data and endpoints, we increase the speed that tests execute at and have complete certainty in what to expect from our mocked endpoints. The risk though is that an api might have changed or break and we won't be aware of this.
 
 ### How to run locally
+
+Running Cypress locally requires having a DCR server running. To run the CI server locally, you can run `make run-ci`. This command runs `make stop`, followed by `make build-ci` and `make start-ci`. You can run these separate commands yourself instead if you want more control over when the site gets re-built.
```
jamesgorrie, line 78: > This command runs `make stop`, followed by `make build-ci` and `make start-ci`. You can run these separate commands yourself instead if you want more control over when the site gets re-built.

I wonder if adding the extra technical info without a use-case adds a little confusion to the really clear "To do this, do this" sentence before?

Not as part of this PR - If this is a requirement to run the below steps and we already have a `make cypress` - maybe we could have a make `dev-cypress-{open | run | mocked | e2e}`? Saying this I am unsure of where we want to use `yarn` vs `make`, but my understanding is that `make` was introduced to stop the need to remember running multiple things.

bryophyta: that's a good point! I've removed the unnecessary details.

on the second point -- can I just check that we don't already have anything like the single command that you described @jamesgorrie ? I couldn't find one when I was looking, but perhaps I missed something?

mxdvl:
I think that if we stop building the legacy bundle, we can probably go back to a place where we don’t have disctinct CI scripts, which would simplify this, too.

PR #5333, opened 2022-07-07
Turbosnap

AshCorr:
## What does this change?

Enables Turbosnap in our Chromatic build.

## Why?

This should reduce our overall Chromatic spending by only taking screenshots of components that we think have changed.

## How does it work?

https://www.chromatic.com/docs/turbosnap#how-it-works

1. Chromatic considers the Git changes between the current commit an

> d the commit of the [ancestor build](https://www.chromatic.com/docs/branching-and-baselines#calculating-the-ancestor-builds).
> 2. Chromatic then uses Webpack’s dependency graph to track those changes back up to the story files that depend on them.
> 3. Chromatic only tests the stories defined in those story files.
> 
> Stories that have not changed will not be tested (i.e., snapshotted), despite appearing in Chromatic’s UI as if they were. In many cases this will lead to much decreased snapshot usage and faster build times.
> 
> Certain circumstances could potentially affect all stories. To prevent false positives, we re-test everything if any of the following situations apply:
> 
> - Changes to package versions in package.json, yarn.lock, package-lock.json
> - Changes to your Storybook’s configuration
> - Changes in files that are imported by your [preview.js](https://storybook.js.org/docs/react/configure/overview#configure-story-rendering) (as this could affect any story)
> - Changes in your static folder (if specified using --static-dir / -s)
> - Changes to files specific by the --externals option (see below)
> - Re-run of the same build (commit and branch match the parent build)
> - [Infrastructure upgrades](https://www.chromatic.com/docs/infrastructure-upgrades)
> - [UI Test in a new browser](https://www.chromatic.com/docs/browsers)

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,9 +1,6 @@
 name: Chromatic 👓
 on:
   push:
-    branches:
-      - main
-  pull_request:
 jobs:
   chromatic:
     name: Chromatic
@@ -28,3 +25,4 @@ jobs:
           projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
           token: ${{ secrets.GITHUB_TOKEN }}
           autoAcceptChanges: main
+          onlyChanged: true


jamesgorrie:
Part of #5940

PR #5337, opened 2022-07-07
refactor: use Navigation Timing v2

mxdvl:
## What does this change?

Use the Performance Navigation Timing v2.

- https://developer.mozilla.org/en--US/docs/Web/API/PerformanceNavigationTiming

## Why?

The v1 API is deprecated:

- https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming
- https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigation


mxdvl:
was too eager

PR #5351, opened 2022-07-11
Fix: show toggle if there are no available topics

abeddow91:
## What does this change?
when we show automatic filters we hide the key events toggle. However, we only how automatic filters on liveblog but the render logic for the key events toggle is handled by an on/off switch. This PR adds additional logic to check if there are any filters available before hiding the key events toggle so that we don't hide the toggle on other types of liveblogs such as sports and politics. 

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before-small][] | ![after-small][] |

[before-small]: https://user-images.githubusercontent.com/20416599/178255782-35388068-4340-4097-9e15-ff0a53ba78e0.png
[after-small]: https://user-images.githubusercontent.com/20416599/178255631-8e9bef98-a7e1-44d4-abff-03f1a8589bb5.png

| Before      | After      |
|-------------|------------|
| ![before-desktop][] | ![after-desktop][] |

[before-desktop]: https://user-images.githubusercontent.com/20416599/178255986-afe01304-5707-4868-9030-40b6ce4fd725.png
[after-desktop]: https://user-images.githubusercontent.com/20416599/178255638-98750887-9f5a-4e25-bb2d-91f232c523a8.png


marjisound (approved):
good spot 👍

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -320,6 +320,8 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 	const showTopicFilterBank =
 		CAPIArticle.config.switches.automaticFilters && showKeyEventsCarousel;
 
+	const showToggle = !showTopicFilterBank || !CAPIArticle.availableTopics;
```
oliverlloyd, line 323: Would it make things clearer if we centralised these decisions into one place and then had one decision outcome? A bit like

```ts
const decideWhichFilterToShow = ( ...things ): 'topics' | 'keyevents' => {
    ...logic
};

const filterToShow = decideWhichFilterToShow( ...things );
```

and then later

```ts
<div id="maincontent" css={bodyWrapper}>
	{CAPIArticle.keyEvents.length && filterToShow === 'keyevents' ? ( ... )}
</div>
```

oliverlloyd (commented):
Appreciate this is a hotfix for a production issue but maybe there's an opportunity to refactor things a bit here? I'm not fully over this code though so feel free to take this as a suggestion and even if you want to move ahead with it I'm happy for you to decide if you want to do it here or later

dotcom-rendering/src/web/layouts/LiveLayout.tsx
abeddow91: @oliverlloyd - this definitely would but we are actually going to be removing the key events carousel switch which will reduce the complexity around the logic here (PR to follow). As this is a hotfix, for now we will go ahead as is, but I like this pattern for future switch coupling!

PR #5352, opened 2022-07-11
Fix beta text colour for automatic filters

marjisound:
## What does this change?
This PR removes the `BETA` text colour style so that it would stay the same colour as it's sibling text `FILTERS`

## Why?
It was initially styled to be red as it was supposed to be a link that would go to a page to get user feedback. But at this stage we haven't yet worked on the user feedback link so the colour needs to stay black for now. 

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

### DIFF
--- a/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
+++ b/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
@@ -1,6 +1,5 @@
 import { css } from '@emotion/react';
 import { from, headline, space, textSans } from '@guardian/source-foundations';
-import { decidePalette } from '../lib/decidePalette';
 import { FilterButton } from './FilterButton.importable';
 
 type Props = {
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
```diff
@@ -105,7 +98,7 @@ export const TopicFilterBank = ({
 			<div css={headlineStyles}>
 				Filters{' '}
 				<span css={headlineAccentStyles}>
-					(<span css={betaTextStyles(palette)}>BETA</span>):
+					(<span>BETA</span>):
```
oliverlloyd:
```suggestion
					(BETA)
```

oliverlloyd: No need for the span? (Will need formatting)

oliverlloyd (approved):
Well, I made a comment about removing that extra span and then it magically disappeared in front of my eyes. You're too fast for me!

alinaboghiu (approved):
Neat!

PR #5358, opened 2022-07-12
Card headlines can now be `ginormous`

oliverlloyd:
## What does this change?
This PR adds a new `ginormous` headline size.

## Why?
When the dynamo card (the primary card in the `dynamic/package` container) is 'boosted' then the headline is 50 pixels

### Why 'ginormous'?
We already have `huge` so we needed a lexicon clearly larger than that, which I think ginormous is.

I considered adopting a similar pattern to what Source does and using `xlarge`, `xxlarge`, but the problems there is that not only would this need to be done over the whole codebase consistently it would also mean we would be using the same size denominations as Source but they would often mean something else (by design, Source and the actual font sizes in use on theguardian.com are not aligned because Source is about more than just dotcom).

Sure, I could have used 'enormous' or 'gigantic' but my gut feeling was to go with 'ginormous'.

Will there be yet another, even larger size after this one? I mean, maybe? But I seriously doubt this. You can't go much beyond 50 pixels without things getting daft.

Screenshot 2022-07-12 at 09 53 04


mxdvl (approved):
A VERY BIG +1

PR #5360, opened 2022-07-12
move colon to after the BETA styling is applied

marjisound:
## What does this change?
Move the colon out of the span where `BETA` styling exists
## Screenshots

| Before      | After      |
|-------------|------------|
| image | image |


AshCorr (approved):
Sorry accidentally approved earlier without saving my comments ;)

PR #5364, opened 2022-07-12
refactor: Add a `FrontCard` component that provides sensible defaults for Cards being used on Fronts

AshCorr:
## What does this change?

- Adds a `FrontCard` wrapper class around the regular `Card` class that provides reasonable defaults used by most Front cards (These defaults can be overriden easily)

## Why?

1. We have a lot of copy-pasted behaviour that will always be the same regardless of Container such as deciding if we should show the Live icon or if we should show the Opinion squiggly lines.
2. Card has a **lot** of props, to the point that just defining a single card in a Container takes up almost an entire viewport which I think has 3 problems:
   1. Hard to understand what the differences are between Cards just by looking at the code.
   2. Easy to make copy paste errors.
   3. Hard to add new features/props to Card without needing to do a lot of prop-fu in all of the Containers (we only have 7 containers at the moment so its not so bad but once we support all of the containers we'll have to edit dozens and dozens of files if we want to add new props)

### DIFF
--- a/dotcom-rendering/src/web/components/DynamicFast.tsx
+++ b/dotcom-rendering/src/web/components/DynamicFast.tsx
@@ -1,7 +1,6 @@
-import { ArticleDesign } from '@guardian/libs';
-import { Card } from './Card/Card';
 import { LI } from './Card/components/LI';
 import { UL } from './Card/components/UL';
+import { FrontCard } from './FrontCard';
 
 type Props = {
 	trails: TrailType[];
@@ -19,39 +18,15 @@ export const DynamicFast = ({ trails, containerPalette, showAge }: Props) => {
 		<>
 			<UL direction="row" padBottom={true}>
 				<LI padSides={true} percentage="75%">
-					<Card
+					<FrontCard
+						trail={primary}
 						containerPalette={containerPalette}
 						showAge={showAge}
-						linkTo={primary.url}
-						format={primary.format}
-						headlineText={primary.headline}
-						trailText={primary.trailText}
+						//Overrides
 						headlineSize="large"
-						byline={primary.byline}
-						showByline={primary.showByline}
-						showQuotes={
-							primary.format.design === ArticleDesign.Comment ||
-							primary.format.design === ArticleDesign.Letter
-						}
-						webPublicationDate={primary.webPublicationDate}
-						kickerText={primary.kickerText}
-						showPulsingDot={
-							primary.format.design === ArticleDesign.LiveBlog
-						}
-						showSlash={true}
-						showClock={false}
-						imageUrl={primary.image}
 						imagePosition="right"
 						imagePositionOnMobile="top"
 						imageSize="large"
-						mediaType={primary.mediaType}
-						mediaDuration={primary.mediaDuration}
-						starRating={primary.starRating}
-						branding={primary.branding}
-						supportingContent={primary.supportingContent}
-						dataLinkName={primary.dataLinkName}
-						snapData={primary.snapData}
-						discussionId={primary.discussionId}
 					/>
 				</LI>
 				<LI
@@ -60,35 +35,10 @@ export const DynamicFast = ({ trails, containerPalette, showAge }: Props) => {
 					showTopMarginWhenStacked={true}
 					percentage="25%"
 				>
-					<Card
+					<FrontCard
+						trail={secondary}
 						containerPalette={containerPalette}
 						showAge={showAge}
-						linkTo={secondary.url}
-						format={secondary.format}
-						headlineText={secondary.headline}
-						headlineSize="medium"
-						byline={secondary.byline}
-						showByline={secondary.showByline}
-						showQuotes={
-							secondary.format.design === ArticleDesign.Comment ||
-							secondary.format.design === ArticleDesign.Letter
-						}
-						webPublicationDate={secondary.webPublicationDate}
-						kickerText={secondary.kickerText}
-						showPulsingDot={
-							secondary.format.design === ArticleDesign.LiveBlog
-						}
-						showSlash={true}
-						showClock={false}
-						imageUrl={secondary.image}
-						mediaType={secondary.mediaType}
-						mediaDuration={secondary.mediaDuration}
-						starRating={secondary.starRating}
-						branding={secondary.branding}
-						supportingContent={secondary.supportingContent}
-						dataLinkName={secondary.dataLinkName}
-						snapData={secondary.snapData}
-						discussionId={secondary.discussionId}
 					/>
 				</LI>
 			</UL>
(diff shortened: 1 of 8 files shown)


OllysCoding:
👀 I'd really like to see this!!!

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -19,39 +18,16 @@ export const DynamicFast = ({ trails, containerPalette, showAge }: Props) => {
...
-					<Card
+					<FrontCard
+						trail={primary}
 						containerPalette={containerPalette}
 						showAge={showAge}
-						linkTo={primary.url}
-						format={primary.format}
-						headlineText={primary.headline}
-						trailText={primary.trailText}
+						//Overrides
```
oliverlloyd, line 25:
```suggestion
```

I tend to find comments like this that apply to multiple props below it often don't age well. Other changes come in later that mean the comment no longer applies or worse, is inaccurate.

Maybe we can do something here with JSDoc?

dotcom-rendering/src/web/components/FrontCard.tsx
```diff
@@ -0,0 +1,50 @@
...
+	const { trail } = props;
+	const defaultProps: CardProps = {
+		linkTo: trail.url,
+		format: trail.format,
+		headlineText: trail.headline,
+		byline: trail.byline,
+		showByline: trail.showByline,
+		showQuotes:
+			trail.format.design === ArticleDesign.Comment ||
+			trail.format.design === ArticleDesign.Letter,
```
oliverlloyd, line 33: Not something that should be done here but this PR makes it clear that this logic could and should live inside the `Card` component itself

dotcom-rendering/src/web/components/FrontCard.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * A wrapper around the normal Card component providing sensible defaults for Cards on front containers.
+ *
+ * Any prop used by the original Card can be used in FrontCard to override the defaults.
+ *
+ * Note: Below parameters are not an exhaustive list of params used by FrontCard, rather they are a list of
+ * commonly overridden params to make it easier for a developer to know if they actually need to override these values.
+ *
+ * @param {SmallHeadlineSize} [headlineSize="medium"] - Defaults to "medium"
```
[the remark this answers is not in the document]
oliverlloyd: I think you don't need to add the type information in JSDoc, we already have TS for that

And lo, I justify my statement with a random blog post: https://dev.to/kamranayub/why-you-don-t-need-types-in-jsdoc-when-documenting-typescript-1pb0

dotcom-rendering/src/web/components/FrontCard.tsx
```diff
@@ -0,0 +1,50 @@
+import { ArticleDesign } from '@guardian/libs';
+import type { Props as CardProps } from './Card/Card';
+import { Card } from './Card/Card';
+
+type Props = {
+	trail: TrailType;
+} & Partial<CardProps>;
```
oliverlloyd, line 7: I wonder. So `trail` is a required prop here because you'd never use `Card` without it. And `trail` itself has some required params which I guess you'd never want to use this component and override. So is the list of override props more of a subset rather than all `Card` props?

E.g.: 

```ts
	linkTo: string;
	format: ArticleFormat;
	headlineText: string;
```

Should never be overridden

```ts
	showQuotes?: boolean;
	byline?: string;
	showByline?: boolean;
	webPublicationDate?: string;
	imageUrl?: string;
	imagePosition?: ImagePositionType;
	imagePositionOnMobile?: ImagePositionType;
	imageSize?: ImageSizeType;
	trailText?: string;
	avatar?: AvatarType;
	showClock?: boolean;
	mediaType?: MediaType;
	mediaDuration?: number;
	kickerText?: string;
	showPulsingDot?: boolean;
	showSlash?: boolean;
	starRating?: number;
	minWidthInPixels?: number;
	dataLinkName?: string;
	branding?: Branding;
	supportingContent?: DCRSupportingContent[];
	snapData?: DCRSnapType;
	containerType?: DCRContainerType;
	showAge?: boolean;
	discussionId?: string;
	isDynamo?: true;
```

_Could_. But maybe only some of these should_ be overridable? Like:

```ts
	showQuotes?: boolean;
	showByline?: boolean;
	imageUrl?: string;
	imagePosition?: ImagePositionType;
	imagePositionOnMobile?: ImagePositionType;
	imageSize?: ImageSizeType;
	trailText?: string;
	avatar?: AvatarType;
	kickerText?: string;
	showPulsingDot?: boolean;
	showSlash?: boolean;
	supportingContent?: DCRSupportingContent[];
	snapData?: DCRSnapType;
	containerType?: DCRContainerType;
	showAge?: boolean;
	discussionId?: string;
	isDynamo?: true;
```

What do you think? Maybe we can just have these props set explicitly here and default the rest to whatever is passed in on the `trail` object?

In the future, some of these can be removed, like `showPulsingDot` or `showQuotes`, which can be worked out inside the Card using `format`

oliverlloyd (commented):
I am a big fan of this PR. You make a really good case in the description for why this abstraction is worth it

dotcom-rendering/src/web/components/FrontCard.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * commonly overridden params to make it easier for a developer to know if they actually need to override these values.
+ *
+ * @param {SmallHeadlineSize} [headlineSize="medium"] - Defaults to "medium"
+ * @param {ImagePositionType} [imagePosition="top"] - Defaults to "top"
+ * @param {ImagePositionType} [imagePositionOnMobile="left"] - Defaults to "left"
+ * @param {ImageSizeType} [imageSize="medium"] - Defaults to "medium"
+ * @param {DCRSupportingContent[]} [supportingContent=undefined] - Defaults to undefined, set to trail.supportingContent if you want this card to show sublinks.
+ */
+export const FrontCard = (props: Props) => {
+	const { trail } = props;
```
bryophyta, line 24: If I'm following this correctly, then `props` still has the `trail` key on line 49 when it gets passed to `Card`. But the `Props` type for `Card` doesn't have a `trail` key. So this is a small point but I feel like it could be a bit confusing that we're passing this prop that won't be used?

@mxdvl 's [recent comment](https://github.com/guardian/dotcom-rendering/pull/5383/commits/5bcae7d406cb7fb03f90d1a21e71a7cbd507da17#r920304759) reminded me that we can spread within a destructuring assignment, so we could possibly do something like:

```const { trail, ...rest } = props;```

here, and then just pass `{ ...defaultProps, ...rest }` on line 49? I don't think this is a big deal either way though :)

bryophyta (approved):
This looks great to me! I definitely agree that having an abstraction here is useful, and this seems like a really good way to do that abstraction!

dotcom-rendering/src/web/components/FrontCard.tsx
AshCorr: Fancy! Thanks for the tip, I'll give it a go!

OllysCoding (approved):
+1 This looks really good to me!

dotcom-rendering/src/web/components/FrontCard.tsx
AshCorr: I'm open to the option of being a bit more restrictive with what props we allow overriding but I'm curious to what the benefit of this would be and if its worth the extra complexity of doing this?

AshCorr: Going to merge this down for now and raise a subsequent PR to restrict the props!

PR #5371, opened 2022-07-12
Add check to recipeMarkup to avoid undefined json-ld

OllysCoding:
We are accidentally adding an 'undefined' json-ld script to the page:
image

This is upsetting google:
image
image

This affects all pages that aren't receiving the recipe markup, so we should look to resolve this quickly to avoid any impact on our search rankings!

Before:
2 json LD tags, the second of which is 'undefined'
image

After:
Only 1 json LD tag, which is always populated
image


joecowton1 (approved):

mxdvl:
cc @frederickobrien for visibility

PR #5372, opened 2022-07-12
add logic for adding the pinned post in the live layout in AR

marjisound:
## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

### DIFF
--- a/apps-rendering/src/components/Layout/LiveLayout.tsx
+++ b/apps-rendering/src/components/Layout/LiveLayout.tsx
@@ -23,12 +23,13 @@ import MainMedia from 'components/MainMedia';
 import Metadata from 'components/Metadata';
 import RelatedContent from 'components/RelatedContent';
 import Tags from 'components/Tags';
-import { getFormat } from 'item';
+import { getFormat, Item } from 'item';
 import type { DeadBlog, LiveBlog } from 'item';
 import { toNullable } from 'lib';
 import type { LiveBlock } from 'liveBlock';
 import type { FC } from 'react';
 import { articleWidthStyles, darkModeCss, onwardStyles } from 'styles';
+import { OptionKind } from '@guardian/types';
 
 // ----- Component ----- //
 
@@ -111,6 +112,10 @@ const keyEvents = (blocks: LiveBlock[]): KeyEvent[] =>
 		[],
 	);
 
+const showPinnedPost = (item: LiveBlog | DeadBlog) =>
+	item.pagedBlocks.currentPage.pageNumber === 1 &&
+	item.pinnedBlock.kind === OptionKind.Some;
+
 interface Props {
 	item: LiveBlog | DeadBlog;
 }
@@ -181,6 +186,8 @@ const LiveLayout: FC<Props> = ({ item }) => {
 						`}
 						id="liveblock-container"
 					>
+						{/* This is a placeholder for the pinned post */}
+						{showPinnedPost(item) && <></>}
 						{item.pagedBlocks.currentPage.pageNumber > 1 &&
 							pagination}
 						<LiveBlocks
--- a/apps-rendering/src/item.ts
+++ b/apps-rendering/src/item.ts
@@ -82,6 +82,7 @@ interface LiveBlog extends Fields {
 	blocks: LiveBlock[];
 	totalBodyBlocks: number;
 	pagedBlocks: LiveBlogPagedBlocks;
+	pinnedBlock: Option<LiveBlock>;
 }
 
 interface DeadBlog extends Fields {
@@ -89,6 +90,7 @@ interface DeadBlog extends Fields {
 	blocks: LiveBlock[];
 	totalBodyBlocks: number;
 	pagedBlocks: LiveBlogPagedBlocks;
+	pinnedBlock: Option<LiveBlock>;
 }
 
 interface Review extends Fields {
@@ -387,6 +389,7 @@ const fromCapiLiveBlog =
 			blocks: parsedBlocks,
 			pagedBlocks,
 			totalBodyBlocks: content.blocks?.totalBodyBlocks ?? body.length,
+			pinnedBlock: fromNullable(parsedBlocks.find((b) => b.isPinned)),
 			...itemFields(context, request),
 		};
 	};
--- a/apps-rendering/src/liveBlock.ts
+++ b/apps-rendering/src/liveBlock.ts
@@ -21,6 +21,7 @@ type LiveBlock = {
 	lastModified: Date;
 	body: Body;
 	contributors: Contributor[];
+	isPinned: Boolean;
 };
 
 // ----- Functions ----- //
@@ -66,6 +67,7 @@ const parse =
 				contributorTags(block.contributors, tags),
 				context,
 			),
+			isPinned: block.attributes.pinned ?? false,
 		});
 	};
 


marjisound:
Just realised I have to do some adjustment to this PR. Please wait a bit until it's done :)

abeddow91 (approved):
Nice work!

apps-rendering/src/components/LiveBlocks/index.tsx
```diff
@@ -30,12 +32,27 @@ const contributorToBlockContributor = (
...
 
-const LiveBlocks: FC<LiveBlocksProps> = ({ blocks, format }) => {
+const LiveBlocks: FC<LiveBlocksProps> = ({
+	blocks,
+	format,
+	pinnedPost,
+	pageNumber,
+}) => {
+	const pinnedPostId =
+		pinnedPost.kind === OptionKind.Some ? pinnedPost.value.id : '';
```
georgeblahblah, line 46: there might be a neater way to express this (I'm still very new to the AR codebase!), but this might be a good opportunity to use [`withDefault`](https://github.com/guardian/dotcom-rendering/blob/56997678c39261f489651da147775cfabca2ec77/apps-rendering/src/option.ts#L52) and [`map`](https://github.com/guardian/dotcom-rendering/blob/56997678c39261f489651da147775cfabca2ec77/apps-rendering/src/option.ts#L73):

```
const pinnedPostId =
		compose(withDefault(''), map(pinnedPost => pinnedPost.id))(pinnedPost)
```

apps-rendering/src/components/Layout/LiveLayout.tsx
```diff
@@ -186,6 +186,8 @@ const LiveLayout: FC<Props> = ({ item }) => {
 						<LiveBlocks
 							blocks={item.pagedBlocks.currentPage.blocks}
 							format={item}
+							pinnedPost={item.pinnedBlock}
```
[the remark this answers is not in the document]
marjisound: Good idea @georgeblahblah 

Because we are using the shared component `LiveBlockContainer` and in there the relevant properties are called `isPinnedPost` & `isOriginalPinnedPost` I changed the field in the `item` model to be `pinnedPost` rather than `pinnedBlock` to make the naming consistent. Do you think it's OK?

apps-rendering/src/components/LiveBlocks/index.tsx
```diff
@@ -44,8 +61,12 @@ const LiveBlocks: FC<LiveBlocksProps> = ({ blocks, format }) => {
 					blockTitle={block.title}
 					blockFirstPublished={Number(block.firstPublished)}
 					blockId={block.id}
-					// TODO pass this value in when available
+					// This is false because it's only true when it's
+					// used for the actual pinned post on top of the page
 					isPinnedPost={false}
+					isOriginalPinnedPost={
+						block.isPinned && block.id === pinnedPostId
```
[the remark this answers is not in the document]
marjisound: Good question. From my testing with the CAPI data, I don't think this should ever happen because as soon as a post is pinned, the pinned attribute in previous pinned post would be set to false. 

But I noticed that DCR is doing this check so I thought there's no harm to do so in here too. Just in case if get 2 blocks that have their pinned attribute set to true. 

But I'm interested to hear what you or others think :)

apps-rendering/src/components/LiveBlocks/index.tsx
marjisound: Very interesting, I made the change 👍

apps-rendering/src/components/LiveBlocks/index.tsx
```diff
@@ -44,8 +61,12 @@ const LiveBlocks: FC<LiveBlocksProps> = ({ blocks, format }) => {
 					blockTitle={block.title}
 					blockFirstPublished={Number(block.firstPublished)}
 					blockId={block.id}
-					// TODO pass this value in when available
+					// This is false because it's only true when it's
+					// used for the actual pinned post on top of the page
 					isPinnedPost={false}
```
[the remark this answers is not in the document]
georgeblahblah: Oh I see! That makes perfect sense, thanks!

apps-rendering/src/components/LiveBlocks/index.tsx
```diff
@@ -30,12 +32,29 @@ const contributorToBlockContributor = (
...
+	)(pinnedPost);
+
+	const showPinnedPost =
+		pageNumber === 1 && pinnedPost.kind === OptionKind.Some;
+
 	return (
 		<>
 			{/* Accordion? */}
+			{/* below is placeholder for pinned post component */}
+			{showPinnedPost && <></>}
```
JamieB-gu, line 57: Out of curiosity, what made you choose to render this inside the `LiveBlocks` component instead of the `LiveLayout` component?

apps-rendering/src/components/LiveBlocks/index.tsx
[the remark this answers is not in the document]
JamieB-gu: A possible alternative implementation that removes the need to set `pinnedPostId` to `''`:

```suggestion
						pipe(
							pinnedPost,
							map(pinned => block.isPinned && block.id === pinned.id),
							withDefault(false),
						)
```

Also just wondering: if the current block id is equal to the pinned block id, do you also need the check for `isPinned`? You've already used `isPinned` to get the pinned block id in the first place I think?

apps-rendering/src/item.ts
```diff
@@ -387,6 +389,7 @@ const fromCapiLiveBlog =
 			blocks: parsedBlocks,
 			pagedBlocks,
 			totalBodyBlocks: content.blocks?.totalBodyBlocks ?? body.length,
+			pinnedPost: fromNullable(parsedBlocks.find((b) => b.isPinned)),
```
JamieB-gu, line 392: What's the thinking behind storing `pinnedPost` in `Item`? So far I think it's only used in once place, in `LiveBlock`, so technically it could be derived directly from `blocks` there? Are there other advantages to having it derived earlier here?

JamieB-gu (commented):
Looks good! Just a couple of comments/questions.

apps-rendering/src/components/LiveBlocks/index.tsx
JamieB-gu: Oh nice spot George! This is an interesting discussion. I think for this kind of thing it's reasonable to use either approach, whichever you find easier to reason about. You'll see examples of `map`, `if` and `switch...case` used to work with these elsewhere in the AR codebase, and I think you'll also see `if (isSome)`, `match` and `map` used in Scala codebases.

@marjisound I'm guessing TS has probably pushed you to add a type annotation in your implementation of George's suggestion, because it struggles to infer the types with the argument order for `compose`? If you'd prefer to avoid this `pipe` may serve as an alternative:

```ts
pipe(
  pinnedPost,
  map(pinned => pinned.id),
  withDefault(''),
)
```

**Note:** Whatever option you go with I'd recommend using `pinned` (or some other name) over `pinnedPost` inside the function to avoid variable shadowing. Variable shadowing can sometimes cause problems: https://eslint.org/docs/latest/rules/no-shadow

apps-rendering/src/components/LiveBlocks/index.tsx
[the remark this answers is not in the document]
marjisound: You're absolutely right, we can only rely on the block id here 👍

apps-rendering/src/item.ts
marjisound: Very great point, as you said we only ever need to know about the pinned post within the `LiveBlocks` component so it makes sense to derive it in `LiveBlocks` rather than adding it to the `Item` model. Made the changes for this 👍

apps-rendering/src/components/LiveBlocks/index.tsx
marjisound: It'd be fine to render it in the `LiveLayout` component, I just thought it might be cleaner to move it into the LiveBlocks, because when the `PinnedPost` component is created in future PR, we need to pass `LiveBlockContainer` as children to it, and that could make the `LiveLayout` a bit messy (AR `LiveLayout` is currently very easy to follow and a lot simpler than the DCR one :) )


ALSO, now that I removed the `pinnedPost` field from `Item` and am deriving it within the `LiveBlocks` component, I think it's ever more important that we render the `PinnedPost` component within `LiveBlocks`

apps-rendering/src/components/LiveBlocks/index.tsx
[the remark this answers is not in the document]
marjisound: The pipe option looks great. I just had to explicitly add the type for `withDefault` otherwise the storybook build would have failed with the following error (the error is so strange 🤔 ):
```
      TS2345: Argument of type '(optA: Option<LiveBlock>) => Option<boolean>' is not assignable to parameter of type '(_a: Option<LiveBlock>) => Option<false>'.
  Type 'Option<boolean>' is not assignable to type 'Option<false>'.
    Type 'Some<boolean>' is not assignable to type 'Option<false>'.
      Type 'Some<boolean>' is not assignable to type 'Some<false>'.
        Type 'boolean' is not assignable to type 'false'.
```

But the build was successful with this:
```
pipe(
  pinnedPost,
  map((pinned) => block.id === pinned.id),
  withDefault<boolean>(false),
)
```

apps-rendering/src/components/LiveBlocks/index.tsx
marjisound: At the end because I took the approach of your other [suggestion](https://github.com/guardian/dotcom-rendering/pull/5372#discussion_r920339393) , I ended up not needing to derive the pinnedPostId any more. 

Your other suggestion:
```
isOriginalPinnedPost={pipe(
  pinnedPost,
  map((pinned) => block.id === pinned.id),
  withDefault<boolean>(false),
)}
```

apps-rendering/src/components/LiveBlocks/index.tsx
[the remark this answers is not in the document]
JamieB-gu: I think the type inference has probably narrowed things too far here. It's inferred the type as the `false` when it's actually `boolean`, which is `true | false`. It's then become unhappy because it can't assign the wider type `boolean` to `false`.

I don't really understand why this would only come up in the storybook build though, given that the main build doesn't seem to have this problem...

apps-rendering/src/components/LiveBlocks/index.tsx
JamieB-gu: That all makes sense, thanks for explaining!

JamieB-gu (approved):
🎉

PR #5376, opened 2022-07-13
Replace triangle icon on captions with camera

georgeblahblah:
## What does this change?

- Replaces the triangle icon on captions with the camera icon for AR & Editions

### AR

| Before      | After      |
|-------------|------------|
| Screenshot 2022-07-13 at 09 03 31 | Screenshot 2022-07-13 at 09 03 08 |

### Editions

| Before      | After      |
|-------------|------------|
| Screenshot 2022-07-13 at 09 04 35 | Screenshot 2022-07-13 at 09 04 06 |

### DIFF
--- a/common-rendering/src/components/figCaption.tsx
+++ b/common-rendering/src/components/figCaption.tsx
@@ -5,35 +5,38 @@ import { css } from '@emotion/react';
 import { remSpace } from '@guardian/source-foundations';
 import { brandAltText, neutral, text } from '@guardian/source-foundations';
 import { textSans } from '@guardian/source-foundations';
+import { SvgCamera } from '@guardian/source-react-components';
 import type { ArticleFormat } from '@guardian/libs';
 import { ArticleDesign } from '@guardian/libs';
 import type { Option } from '@guardian/types';
 import { OptionKind } from '@guardian/types';
 import type { FC, ReactNode } from 'react';
-import { fill } from '../editorialPalette';
 import { darkModeCss } from '../lib';
 
 // ----- Sub-Components ----- //
 
-interface TriangleProps {
+interface CameraProps {
 	format: ArticleFormat;
 	supportsDarkMode: boolean;
 }
 
-const triangleStyles = (
-	format: ArticleFormat,
-	supportsDarkMode: boolean,
-): SerializedStyles => css`
-	fill: ${fill.icon(format)};
-	height: 0.8em;
+const cameraStyles = (supportsDarkMode: boolean): SerializedStyles => css`
+	display: inline-block;
+	height: 1.2rem;
+	vertical-align: middle;
 	padding-right: ${remSpace[1]};
 
+	svg {
+		height: 100%;
+		fill: ${text.supporting};
+	}
+
 	${darkModeCss(supportsDarkMode)`
-        fill: ${fill.iconDark(format)};
+        fill: ${brandAltText.supporting};
     `}
 `;
 
-const Triangle: FC<TriangleProps> = ({ format, supportsDarkMode }) => {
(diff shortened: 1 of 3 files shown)


georgeblahblah:
I think we should also (in a separate PR) move `figCaption` out of `common-rendering` for now as it's only being used in AR

georgeblahblah:
@mxdvl I have a feeling that the way I've handled the SVG here isn't ideal, if you wouldn't mind taking a look?

mxdvl (commented):
I would:
- keep the `fill` attribute at the top level
- specify the width rather than the height, or specifiy both!
- force `display: block;` on the SVG
- remove the `vertical-align`

```ts
css`
	display: inline-block;
	width: 1.2rem;
	padding-right: ${remSpace[1]};

	svg {
		display: block;
		width: 100%;
	}
	
	fill: ${text.supporting};
		fill: ${brandAltText.supporting};
	`}
`;
 ```

**Comment by georgeblahblah** [comment]
Thanks so much @mxdvl! I had a go with your suggested approach but ran into some more problems which I _think_ are related to the sizing of the SVG in Source.

I had good results with adopting the handling of the caption icon from [DCR's implementation](https://github.com/guardian/dotcom-rendering/blob/a401bd0fbb839916af4fade22e8d1a3b2974b373/dotcom-rendering/src/web/components/Caption.tsx#L153) - hoping I haven't missed any important context from that!

File: common-rendering/src/components/figCaption.tsx [hunk 1]
```diff
@@ -5,49 +5,57 @@ import { css } from '@emotion/react';
...
+	svg {
+		position: absolute;
+		top: 0;
+		left: 0;
+		width: 100%;
+	}
 
 	${darkModeCss(supportsDarkMode)`
-        fill: ${fill.iconDark(format)};
+        fill: ${brandAltText.supporting};
```
JamieB-gu, line 44: I think `text` and `brandAltText` were deprecated in v4 of `source-foundations`? The closest available I can find in the current version are:

- `text.supporting` -> `neutral[46]`
- `brandAltText.supporting` -> `neutral[60]`

You may want to add these to the `fill` module in `editorialPalette`, similar to how the fill colour is managed in the DCR example you linked above. Perhaps `cameraCaptionIcon` and `cameraCaptionIconDark`? I think, in general, if you're setting a colour (via `color`, `background-color`, `fill`, `border` etc.) it's worth considering whether it could live in `editorialPalette`.

JamieB-gu (commented):
Looks good! One suggestion.

mxdvl:
> Thanks so much @mxdvl! I had a go with your suggested approach but ran into some more problems which I _think_ are related to the sizing of the SVG in Source.
> 
> I had good results with adopting the handling of the caption icon from [DCR's implementation](https://github.com/guardian/dotcom-rendering/blob/a401bd0fbb839916af4fade22e8d1a3b2974b373/dotcom-rendering/src/web/components/Caption.tsx#L153) - hoping I haven't missed any important context from that!

I’m not conviced that our approach is that great either, to be honest. It might be worth getting this merged as-is for consistency, and set some time aside next week to review both?

common-rendering/src/components/figCaption.tsx
georgeblahblah: Oh yes, `text` and `brandAltText` are deprecated! Strangely, VSCode wasn't marking the imports as deprecated - I think because the version of `source-foundations` used hadn't added the JSDoc yet.

That's a great point about adding these colours to `editorialPalette`, thanks!

JamieB-gu (approved):

georgeblahblah:
> It might be worth getting this merged as-is for consistency, and set some time aside next week to review both?

@mxdvl sounds good, I'll set something up! Thanks again for taking a look

PR #5377, opened 2022-07-13
Feat: hide pinned post when filtering

abeddow91:
## What does this change?
Hides the pinned post when the user is filtering a liveblog.

## Why?
Editorial have requested we hide the pinned post when the user is filtering as it doesn't necessarily relate to the filter in use. 
In order to be consistent, filtering here refers to filtering using the key events toggle or an auto filter button. 

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before-button][] | ![after-button][] |

[before-button]: https://user-images.githubusercontent.com/20416599/178691997-1c7a8d3d-5c75-4c5d-9e56-87e393dd8755.png
[after-button]: https://user-images.githubusercontent.com/20416599/178691791-2b6f195a-3284-4709-aa1c-abff55250565.png

| Before      | After      |
|-------------|------------|
| ![before-toggle][] | ![after-toggle][] |

[before-toggle]: https://user-images.githubusercontent.com/20416599/178692115-782e1dfe-e1e5-44fc-9c0d-71884704fca8.png
[after-toggle]: https://user-images.githubusercontent.com/20416599/178692153-cf863273-c142-4caf-8f19-981f7238a76f.png

### DIFF
--- a/dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
+++ b/dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
@@ -59,7 +59,8 @@ export const LiveBlogRenderer = ({
 	availableTopics,
 	selectedTopics,
 }: Props) => {
-	const filtered = !!selectedTopics || filterKeyEvents;
+	const filtered =
+		(selectedTopics && selectedTopics.length > 0) || filterKeyEvents;
 
 	return (
 		<>
--- a/dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
+++ b/dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
@@ -59,7 +59,7 @@ export const LiveBlogRenderer = ({
 	availableTopics,
 	selectedTopics,
 }: Props) => {
-	const filtered = selectedTopics?.length || filterKeyEvents;
+	const filtered = !!selectedTopics || filterKeyEvents;
 
 	return (
 		<>
--- a/dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
+++ b/dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
@@ -59,9 +59,11 @@ export const LiveBlogRenderer = ({
 	availableTopics,
 	selectedTopics,
 }: Props) => {
+	const filtered = selectedTopics?.length || filterKeyEvents;
+
 	return (
 		<>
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
```diff
@@ -59,9 +59,11 @@ export const LiveBlogRenderer = ({
 	availableTopics,
 	selectedTopics,
 }: Props) => {
+	const filtered = selectedTopics?.length || filterKeyEvents;
```
marjisound, line 62: I may be wrong but I think filtered could end up with a value of number or boolean. 
0 || true => true
0 || false => false
1 || false => 1
1 || true => 1

Would it be not better to make sure it's value is a boolean? maybe like this:

`const filtered = selectedTopics?.length > 0 || filterKeyEvents;`

marjisound (commented):
Nice work Anna, just left a question too see what you think :)

dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
abeddow91: Sounds sensible! I'll swap this out

abeddow91, line 62:
```suggestion
	const filtered = selectedTopics?.length > 0 || filterKeyEvents;
```

marjisound (approved):
Looks great Anna, thanks :)

dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
abeddow91: After taking another look at this I've actually opted for a double bang here instead on selectedTopics as `[]` or `undefined` are falsey and so will return false when combined with `!!`. That way we don't need the additional length check.

marjisound: Maybe I'm doing it wrong but just did a test in js console and empty list seems to come as truthy 🤔 
image

oliverlloyd: Ah, the joys of javascript. Great spot!

abeddow91: You're totally right @marjisound, let me revert this!

PR #5378, opened 2022-07-13
Fix author / contributor edge cases

mxdvl:
## What does this change?

Refactor `byline` to stop depending on the author field, which we’ve marked as depracated.

Instead, look at the list of contributors, and check if we have a sole contributor, in which case we may display their twitter handle, image, bio, etc.

## Why?

Resolves https://github.com/guardian/frontend/issues/25220

## Screenshots

(The picture is of Ben-Thomas, not Smith)

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/179212935-28a8211c-a1a3-428a-8875-d12d7d2c7302.png
[after]: https://user-images.githubusercontent.com/76776/179213039-5f53ae0c-bc58-4274-a821-e375bda11e0f.png


dotcom-rendering/src/lib/byline.ts
```diff
@@ -1,11 +1,38 @@
...
+interface ContributorTag extends TagType {
+	type: 'Contributor';
+}
+
+const isContributor = (tag: TagType): tag is ContributorTag =>
+	tag.type === 'Contributor';
+
+/**
+ * Get the sole contributor of an article.
+ * This only works if all of the following are true:
```
bryophyta, line 10: 'This only works' implies to me that it might break if some of the conditions don't hold. But my understanding is that it won't break, and that it's actually the desired behaviour that it will only return a contributor in the case where all the conditions hold.

If this is correct, maybe this comment could be changed to something like `An article has a sole contributor only if all of the following are true:`?

dotcom-rendering/src/lib/byline.ts
```diff
@@ -1,11 +1,38 @@
...
+ * - there is only one `Contributor` tag
+ * - that tag’s `title` is the same as the article byline
+ *
+ * @returns A sole contributor if there is one, undefined otherwise.
+ */
+export const getSoleContributor = (
+	tags: TagType[],
+	byline: string | undefined,
+): ContributorTag | undefined => {
+	const [soleContributor, ...otherContributors] = tags.filter(isContributor);
```
bryophyta, line 20: if I understand the code correctly, then if the right hand side is an empty array, `soleContributor` will be `undefined` and `otherContributors` will be `[]`, right? This seems like a very neat way to write this bit of logic! And I think the naming of the variables makes it easy to get a sense of what's going on. But I wonder whether the underlying mechanisms might be a bit opaque to future readers? (I know that I had to double check what using rest decomposition on an empty array would yield, at least.)

dotcom-rendering/src/lib/byline.ts
bryophyta: And given that a lot of the rendering decisions which invoke `getSoleContributor` are really relying on both of these conditions, I wonder if it's worth spelling out in a bit more detail here what it is that makes these conditions relevant to those rendering decisions? (i.e. why we don't show any images if there is more than one contributor, and why it is that byline and tag title can diverge?)

bryophyta (approved):
I haven't dug into the existing CAPI/Frontend model behaviour, but with that caveat the code changes here look good to me!

Just a couple of small comments about documenting the functionality. I think that the `getSoleContributor` abstraction is a really nice way to simplify a lot of the logic in the rendering stages, but because it carries this weight, my feeling is that we want to make it as clear as possible what work it's doing and why..?

dotcom-rendering/src/lib/byline.ts
mxdvl: I'll add a comment about what destructuring does on an empty array.

If we enabled `strictNullChecks` in our `.tsconfig`, the compiler would know that soleContributor could be `undefined`.

PR #5379, opened 2022-07-13
GHA: Ignore changes which only affect dotcom docs

### LINKED ISSUE #5362: Exclude `docs` directories from (most) GHA workflows?
Currently most of our GHA workflows are configured so that if there is a push that only affects files in `apps-rendering`, then the workflow won't be triggered.[^1] This uses [the `paths-ignore` field](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths):

```yml
name: DCR jest 🤔
on:
    push:
        paths-ignore:
            - "apps-rendering/**"
```

As far as I can tell, most of our Actions are also irrelevant for the `docs` directory within `dotcom-rendering` (exceptions might include `prettier.yml`).

Generally speaking, I don't think that the financial costs of this are significant directly (especially now we've enabled Turbosnap), but there is still an overhead in terms of time when trying to make a quick docs update. In addition, while the environmental impact of running these jobs is not large, it may not be entirely trivial.

If there isn't a significant value-add for running the whole GHA suite when only docs files have changed, I suggest that we add `dotcom-rendering/docs/**` to the list of ignored paths for any unnecessary tests. I'm happy to open a PR to add this to the relevant workflows individually, but it's perhaps also worth noting that @jamesgorrie's [GHA workflows RFC](https://github.com/guardian/dotcom-rendering/pull/5170) might make this sort of change easier to implement and maintain in the future.

One potential drawback with excluding directories from workflows is that subsequent changes to the codebase might change which directories need coverage. e.g. if a directory was excluded from Chromatic testing because it didn't have any rendering-relevant code, we couldn't guarantee that this would remain the case, but this is hard to get visibility on. I think this speaks in favour of maintaining broad coverage in general, but I don't think we should be putting production-relevant code into `docs` as a matter of principle.

If this does introduce a gap in our CI/CD that I'm not aware of, please let me know.

[^1]: "When all the path names match patterns in paths-ignore, the workflow will not run. If any path names do not match patterns in paths-ignore, even if some path names match the patterns, the workflow will run." ([docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths))

bryophyta:
## What does this change?

Add `dotcom-rendering/docs/**` to the 'ignore' list in workflow 'on' conditions for workflows which don't affect docs.

## Why?

To reduce unnecessary workflow runs (closes #5362 )

### DIFF
--- a/.github/workflows/amp-validation.yml
+++ b/.github/workflows/amp-validation.yml
@@ -3,6 +3,7 @@ on:
     push:
         paths-ignore:
             - "apps-rendering/**"
+            - "dotcom-rendering/docs/**"
 
 jobs:
     amp_validation:
--- a/.github/workflows/build-check.yml
+++ b/.github/workflows/build-check.yml
@@ -3,6 +3,7 @@ on:
     push:
         paths-ignore:
             - "apps-rendering/**"
+            - "dotcom-rendering/docs/**"
 
 jobs:
     build_check:
--- a/.github/workflows/bundle-analyser.yml
+++ b/.github/workflows/bundle-analyser.yml
@@ -3,7 +3,7 @@ on:
     push:
         paths-ignore:
             - "apps-rendering/**"
-
+            - "dotcom-rendering/docs/**"
 jobs:
     build_check:
         name: DCR Bundle Analyser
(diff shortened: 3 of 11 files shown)


AshCorr (approved):
👍

PR #5380, opened 2022-07-13
Add `/FrontJSON` to view enhanced json for fronts

OllysCoding:
## What does this change?

Historically in DCR, the `.json?dcr` data for an article was almost exactly the same format as what made its way through our rendering code - this made it especially easy to diagnose issues, and figure out exactly where to look in the code. 

With fronts, we're now doing significantly more transformation in DCR, which means looking at a `.json?dcr` variant of a front does not show you the same data that might be consumed during rendering!

By adding this `/FrontJSON` endpoint, anyone with a DCR local server up and running can quickly view the 'enhanced' (transformed) data that we're passing directly to the rendering layer, making local debugging & testing much easier!
 
## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |
|           | image |

[before]: https://user-images.githubusercontent.com/9575458/178722276-d9cbe30b-11bf-4434-bf99-88e27146d797.png
[after]: https://user-images.githubusercontent.com/9575458/178721966-e4752338-189d-4cd6-8e5b-067751f96ed2.png

### DIFF
--- a/dotcom-rendering/src/server/prod-server.ts
+++ b/dotcom-rendering/src/server/prod-server.ts
@@ -12,6 +12,7 @@ import {
 	renderPerfTest as renderArticlePerfTest,
 	renderBlocks,
 	renderFront,
+	renderFrontJson,
 	renderInteractive,
 	renderKeyEvents,
 } from '../web/server';
@@ -57,6 +58,7 @@ export const prodServer = (): void => {
 	app.post('/Blocks', logRenderTime, renderBlocks);
 	app.post('/KeyEvents', logRenderTime, renderKeyEvents);
 	app.post('/Front', logRenderTime, renderFront);
+	app.post('/FrontJSON', logRenderTime, renderFrontJson);
 
 	// These GET's are for checking any given URL directly from PROD
 	app.get(
@@ -96,6 +98,21 @@ export const prodServer = (): void => {
 			// Eg. http://localhost:9000/Front?url=https://www.theguardian.com/uk/sport
 			try {
 				return renderFront(req, res);
+			} catch (error) {
+				console.error(error);
+			}
+		},
+	);
+
+	app.get(
+		'/FrontJSON',
+		logRenderTime,
+		// TODO: ensure getContentFromURLMiddleware supports fronts
+		getContentFromURLMiddleware,
+		async (req: Request, res: Response) => {
+			// Eg. http://localhost:9000/FrontJSON?url=https://www.theguardian.com/uk/sport
+			try {
+				return renderFrontJson(req, res);
 			} catch (error) {
 				console.error(error);
 			}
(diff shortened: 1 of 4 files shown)


mxdvl (approved):

bryophyta (approved):
This seems like a really useful endpoint to have!

oliverlloyd (approved):
Nice and simple

dotcom-rendering/src/server/prod-server.ts
```diff
@@ -102,6 +104,21 @@ export const prodServer = (): void => {
 		},
 	);
 
+	app.get(
+		'/FrontJSON',
```
jamesgorrie, line 108: How would you get to this on prod?

OllysCoding: This will allow you to get the enhanced endpoint when running things like `make run-ci` which run off of the prod server.

If we find this useful and want to add it to 'prod' then we could add the endpoint in Frontend when we query with something like `www.theguardian.com/uk.json?dcr&enhanced` for example - but I would first want to discuss whether are comfortable exposing that data publicly

jamesgorrie: > then we could add the endpoint in Frontend

I can't think of a reason - I just forgot `prod` is just config rather than where it i hosted so has more applications.

jamesgorrie:
@OllysCoding What workflow instigated this? I was wondering as I wondered if some tests might make it easier to understand what is happening to the JSON, and why. There seems to be pretty good coverage of the enhancers, but wondered if there could be more, or if it could be clearer as to where they are enhancing?

e.g. I make a change, something is unexpected - I could go look at the relevant tests and see what's happening.

This is definitely a great workflow enhancement, but personally I don't like to check my browser for JSON changes, but prefer to have tests telling me what's going on, so wondered if we could support both (I have also not worked much in this part of the code, so forgive if I am asking irrelevant things)?

OllysCoding:
> What workflow instigated this?

I've found this workflow is very common for me, specifically;
1. We get a message with a link saying something like 'Why does article this look the way it does'
2. I'll look at the article, then go straight to the `.json?dcr` to examine the data - this data is in the same format as it is in the rendering layer
3. I can immediately validate basic things like; is the weird behaviour coming from upstream (frontend & capi) or the rendering layer
4. If it's the rendering layer - I can immediately consume the keys / objects of the json data scructure to find it in DCR and start looking at the logic
5. If it's upstream - I can check the CAPI response to determine if it's a Frontend problem or a CAPI problem.

I do it this way because these queries are generally specific to the data coming into DCR, rather than general expected behaviour (which can be identified with things like reading tests). 

In that sense it's less about _what_ is happening to the data, and more about being able to visualise what the data is - and using it to get quick access to the right part of the rendering layer, or identifying upstream problems :)

PR #5382, opened 2022-07-13
Add containers to all inline ad slots and refactor slot css

Jakeii:
## What does this change?
CSS on article and liveblog inline advert containers and slots.

## Why?
We have been having issues with 'expanding' adverts escaping their slots (see below before image(s). We have added containers to all inline slots to align and constrain their widths.

This PR apply's the relevant CSS as well as refactoring the slot CSS to be more concise and readable.

Changes on frontend https://github.com/guardian/frontend/pull/25236

## Screenshots

| Before      | After      |
|-------------|------------|
| image | |
Screenshot 2022-07-13 at 13 14 03 | Screenshot 2022-07-13 at 13 15 05 |


chrislomaxjones (approved):
Looks good, and much simpler 👍

PR #5383, opened 2022-07-13
`isBoosted`

oliverlloyd:
## What does this change?
This adds the `isBoosted` property and uses it in `dynamic/package` to change the layout and size of the headline

## Why?
We want to respect the contracts around dynamic package and make sure it provides the same features as Frontend

| Before      | After      |
|-------------|------------|
| Screenshot 2022-07-13 at 13 23 11 |Screenshot 2022-07-13 at 13 24 36 |


dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -14,9 +14,106 @@ export const DynamicPackage = ({
 	containerPalette,
 	showAge,
 }: Props) => {
-	const primary = trails[0];
-	const remaining = trails.slice(1, 4);
+	const primary = [...trails].slice(0)[0];
```
bryophyta, line 17: I'm probably just missing something here, but do we need both the spread and the slice here?

iiuc `slice` shouldn't mutate the array being sliced, so spreading here would only seem necessary if we were going on to mutate `primary`. But are we mutating `primary`? If so, should we be?

mxdvl:
```ts
const [primary, ...remaining] = cards;
```

oliverlloyd: Sweet

I think I started here but then got stuck on wanting the last card so went back to slice. But now I see that I'm no longer needing the last card so destructuring works!

PR #5305, opened 2022-07-05
Smaller Islands - Re-enable code splitting for webpack chunks

arelra:
## What does this change?

The current Webpack browser script is generating chunks with duplicated dependencies.

Hence each (dynamically imported) island contains duplicated dependencies e.g. `preact`, `@guardalin/libs`, etc.

This means increased download, parse and compilation times for larger chunks and is likely contributing to a lower Lighthouse / web vitals score for network or CPU constrained devices.

After some investigation I found that we had previously disabled Webpack's code splitting aross entry points. The reason for the [initial change](https://github.com/guardian/dotcom-rendering/pull/1492/commits/e03b72c3dd1a952b035408079e986c031a5a0006) is unclear but in our post-loadable, post-islands world there doesn't seem to be any adverse effects of re-enabling - **please verify!**

This PR re-enables code splitting across entry points.

This results in an overall reduction of:

| Chunk | Type | Before | After | Reduction | ~% Reduction |
|--|--|--|--|--|--|
|all |parsed |2.05MB |1.44MB 💾 |**0.60MB**  | 30% |
|all |gzip |627kB |512kB |115kB| 19% |

Some islands have seen very large reductions e.g. `OnwardsUpper-importable`:

| Chunk | Type | Before | After | Reduction | ~% Reduction |
|--|--|--|--|--|--|
|`OnwardsUpper-importable` |parsed |88kB |21kB |67kB | 76% |
|`OnwardsUpper-importable` |gzip |23kB |8kB |15kB| 65% |

## Why?

Reduce the amount of JavaScript we ship to the client.

Browsers spend less time downloading, parsing and compiling duplicated code.

### Further Improvements

Webpack will in some cases, not split chunks if it deems the chunk to be within configured bounds so **there is still dependency duplication**.

I think we could improve this further by configuring cache groups using [SplitChunksPlugin](https://webpack.js.org/plugins/split-chunks-plugin/)

However care needs to be taken when specifying custom cache groups. Splitting chunks too finely or in an unsuitable manner can risk degrading performance by introducing waterfall module loading or saturating network requests.

### Webpack Stats

| Before      | After      |
|-------------|------------|
| ![before-stats][] | ![after-stats][] |
| ![before-onwu][] | ![after-onwu][] |

[before-stats]: https://user-images.githubusercontent.com/7014230/177203108-6e5427b4-8493-4090-82a2-0cfd28f9957d.png
[...]

PR #5384, opened 2022-07-13
Use Braze to power `My Account` notifications

tjmw:
## What does this change?

This wires up Braze to power the notifications UI added in #5241 using Braze [content cards](https://www.braze.com/docs/user_guide/message_building_by_channel/content_cards).

## Why?

There's a strategy to try using more persistent notifications on the web, as opposed to in-app messages (like in the banner and epic slots, which are more single-shot).

## Screenshots

Screenshot 2022-07-12 at 15 12 25

## Things to highlight

* This is behind a switch which is currently set to `false` in prod

* I've added an `id` field to the `DropdownLinkType` which is what we're using to target the relevant link in Braze. Interested in thoughts on this approach!

* The `useBraze` custom hook now returns an object which has this type:

```ts
{
	brazeMessages: BrazeMessagesInterface | undefined;
	brazeCards: BrazeCardsInterface | undefined;
}
```

We decided to keep the existing messages class separate from the new content cards class as their use cases are quite different. Most call sites will use one or the other, though the work to initialise them both is identical).

* `useBraze` is now used in `Links.importable.tx` to get the `brazeCards` instance.

### DIFF
--- a/dotcom-rendering/src/web/components/Links.importable.tsx
+++ b/dotcom-rendering/src/web/components/Links.importable.tsx
@@ -7,11 +7,18 @@ import {
 	from,
 	textSans,
 } from '@guardian/source-foundations';
+import { useEffect, useState } from 'react';
 import { createAuthenticationEventParams } from '../../lib/identity-component-event';
 import ProfileIcon from '../../static/icons/profile.svg';
 import SearchIcon from '../../static/icons/search.svg';
 import { getZIndex } from '../lib/getZIndex';
+import {
+	addNotificationsToDropdownLinks,
+	mapBrazeCardsToNotifications,
+} from '../lib/notification';
+import type { Notification } from '../lib/notification';
 import { useApi } from '../lib/useApi';
+import { useBraze } from '../lib/useBraze';
 import type { DropdownLinkType } from './Dropdown';
 import { Dropdown } from './Dropdown';
 
@@ -20,6 +27,7 @@ type Props = {
 	discussionApiUrl: string;
 	idUrl?: string;
 	mmaUrl?: string;
+	idApiUrl: string;
 };
 
 const linkStyles = css`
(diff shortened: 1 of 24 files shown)


dotcom-rendering/src/web/lib/braze/buildBrazeMessaging.ts
```diff
@@ -89,18 +105,22 @@ export const buildBrazeMessages = async (
...
+			LocalMessageCache,
+			errorHandler,
+		);
+		return { brazeMessages, brazeCards };
 	} catch {
-		return new NullBrazeMessages();
+		return {
+			brazeMessages: new NullBrazeMessages(),
+			brazeCards: new NullBrazeCards(),
+		};
```
mxdvl, line 124: I find early returns easy to reason about, and you can achieve a similar thing with try-catch blocks.

Instead of returning this in three different instances, including the two failure states–no local storage, or the depedencies check are unsucessful–you could throw on failure and find yourself in the catch block. You would have to wrap the whole function body inside the try block.

dotcom-rendering/src/web/lib/useBraze.ts
```diff
@@ -1,25 +1,42 @@
...
 import useSWRImmutable from 'swr/immutable';
-import { buildBrazeMessages } from './braze/buildBrazeMessages';
+import { buildBrazeMessaging } from './braze/buildBrazeMessaging';
 
 /**
- * Returns brazeMessages as BrazeMessagesInterface
+ * Returns brazeMessaging as BrazeMessagesInterface and BrazeCardsInterface
+ *
+ * BrazeMessages is used to show single-impression messages (like ad impressions).
+ * In contrast, BrazeCards can prodivde persistent user notifications.
```
mxdvl, line 16: I love the cSpell extension to catch these early!

```suggestion
 * In contrast, BrazeCards can provide persistent user notifications.
```

mxdvl (approved):
This looks fine by me. I think the changes in the bundle sizes is due to how Webpack chooses to chunk the entrypoints? We’ve merged #5305 recently, which may lead to more variations for the compressed size action.

dotcom-rendering/src/web/components/Links.importable.tsx
```diff
@@ -147,15 +155,19 @@ const SignIn = ({ idUrl }: { idUrl: string }) => (
 	</a>
 );
 
-const MyAccount = ({
-	mmaUrl,
-	idUrl,
-	discussionApiUrl,
-}: {
+interface MyAccountWithNotificationsProps {
```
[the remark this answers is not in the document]
tjmw: This file defines more than one component, each of which have different props, which is why this has such a specific name. Maybe that's an argument for refactoring to split across multiple files. I'll have a look at doing that.

dotcom-rendering/src/web/lib/braze/buildBrazeMessaging.ts
tjmw: Ah that's interesting. I've always slightly shied away from using exceptions for flow control as I think it _can_ make it harder to follow the flow of the code. Is it a pattern used elsewhere in DCR? I do agree that the multiple points where we return the same thing is a bit sub-optimal, I'll see if I can improve that.

mxdvl: I don’t think it’s a pattern I would recommend, but in some cases it may make sense? Again, we’ve got TypeScript making sure that the return types match, so it’s not really worth improving on much further here, as it’s only duplicated twice and not worth creating an abstraction over.

PR #5386, opened 2022-07-13
Use flex for GalleryImage instead of absolute positioning

iainjchambers-guardian:
## What does this change?
Changes the gallery sections to us flex instead of a fixed width on the image and then placing the caption to the right with absolute positioning.
## Why?
The caption being absolutely positioned meant that the height of each row wouldn't grow/shrink with the height of the caption contents. This also meant that adding padding to the right of the caption without causing the height to change and then the text overlapping the element below.
## Screenshots

| Before      | After      |
|-------------|------------|
| image | image |
| image | image |


apps-rendering/src/components/editions/galleryImage/index.tsx
```diff
@@ -50,6 +58,23 @@ const styles = css`
 	}
 `;
 
+const imgStyles = css`
+	${from.tablet} {
+		flex: 0 0 538px;
```
georgeblahblah, line 63: probably nothing to change here, but I'd like to understand a bit more about which cases we use `px` vs `rem` units

apps-rendering/src/components/editions/layout/index.tsx
```diff
@@ -62,32 +62,47 @@ const headerStyles = css`
...
-	padding-top: ${remSpace[3]};
-	padding-bottom: ${remSpace[3]};
-	iframe {
-		width: 100%;
-		border: none;
-	}
-
-	figcaption {
-		background: ${background.primary};
+const bodyStyles = (item: Item): SerializedStyles => {
```
JamieB-gu, line 65: Might be worth passing `Format` here rather than the full `Item`?

JamieB-gu (approved):
One suggestion, otherwise looks good! 👍

PR #5387, opened 2022-07-13
fix: Use npm ci when installing apps-rendering in CI environment

AshCorr:
## What does this change?

 - Adds a condition to the `postinstall` script to use `npm ci` instead of `npm install` when being ran in Github Actions
 - Removes a lot of superfluous yarn & npm install steps(Probably should have been done in a seperate PR, got a bit ahead of myself!)

## Why?

Currently Turbosnap is being skipped on every PR because `npm install` keeps updating the `package-lock.json` file for apps-rendering. Turbosnap is very cautious about when to run tests and if it detects that the package-lock.json has changed it will revert to the standard Chromatic behaviour of running all tests.

### Yarn --ignore-scripts

Yarn has an --ignore-scripts option which can be used to disable the postinstall script, but it seems like it applies this restriction to all dependencies too.

### Turbosnap CLI

Turbosnap has a CLI option to specifically ignore certain files from the "Should I run all stories?" failsafe, but I couldn't see how/if this can be enabled using the Github action options.

### DIFF
--- a/.github/workflows/amp-validation.yml
+++ b/.github/workflows/amp-validation.yml
@@ -18,9 +18,6 @@ jobs:
             # Cache npm dependencies using https://github.com/bahmutov/npm-install
             - uses: bahmutov/npm-install@v1
 
-            - name: Install
-              run: yarn
-
             - name: Validate AMP
               run: make ampValidation
               working-directory: dotcom-rendering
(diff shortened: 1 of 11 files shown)


JamieB-gu (approved):
Good spot!

jamesgorrie:
Part of #5940

PR #5390, opened 2022-07-14
Liveblogs : Main Media Redesign

abeddow91:
## What does this change?
This PR is part of a wider body of work to move the main media above the headline on liveblogs.

In order for D&I to understand the impact this change has had on engagement, they have requested this change be tested in a 50/50 A/B test for one day. Once this is completed, all ab test checks will be removed. 

This PR introduces the following changes behind a/b checks

- Moves the main media below the headline until desktop
- Moves the main media below the scoreboard for minute by minute football 
- Removes gutters
- Removes the left and right bleed on tablets
- Removes the accordion component as this is no longer required

Previously, the video caption was moved onto the video. This design work is still TBC so it has been removed from this PR. Once this is finalised, this work will sit within the YouTube component that we import.

### Mobile - Standard
| Before      | After      |
|-------------|------------|
| ![before-mobile][] | ![after-mobile][] |

[after-mobile]: https://user-images.githubusercontent.com/20416599/188852555-dd20e214-b35b-4670-b8f9-2762ea8529c5.png
[before-mobile]: https://user-images.githubusercontent.com/20416599/188852579-cd654b91-5f48-4231-8ddd-27d23364e692.png

### Tablet - Standard
| Before      | After      |
|-------------|------------|
| ![before-tab][] | ![after-tab][] |

[after-tab]: https://user-images.githubusercontent.com/20416599/188852601-f4984d9b-b497-4fe6-8932-ca757f55abf4.png
[before-tab]: https://user-images.githubusercontent.com/20416599/188852654-5e40e37b-e64a-4b9c-ba0f-16a5c4c633cf.png

### Mobile - Football MBM
| Before      | After      |
|-------------|------------|
| ![before-mobile-ftbl][] | ![after-mobile-ftbl][] |

[before-mobile-ftbl]: https://user-images.githubusercontent.com/20416599/188852896-c8e22ab9-4517-48bc-b90a-4f4faad13baa.png
[after-mobile-ftbl]: https://user-images.githubusercontent.com/20416599/188852918-610c42eb-702d-4e16-bf5f-8dcafa73916c.png

### Tablet - Football MBM

| Before      | After      |
|-------------|------------|
| ![before-tab-ftbl][] | ![after-tab-ftbl][] |

[before-tab-ftbl]: https://user-images.githubusercontent.com/20416599/188852960-35098999-8edc-4a97-9700-edab634c8c03.png
[after-tab-ftbl]: https://user-images.githubusercontent.com/20416599/188852963-d26d4663-d863-493a-9054-00ba37bc94d0.png


dotcom-rendering/src/web/components/CaptionToggle.tsx
```diff
@@ -0,0 +1,59 @@
...
+		left: 0;
+		margin: auto;
+		position: absolute;
+		fill: white;
+	}
+`;
+const Toggle = () => (
+	<>
+		<label htmlFor="the-checkbox" css={toggleStyles}>
+			<svg width="6" height="14" fill="white" viewBox="0 0 6 14">
```
abeddow91, line 29: Note: This will be replaced by source once it has been added to the icons library. This is blocked as Akemi is currently on AL

JamieB-gu:
@abeddow91 if the test is successful will this change be coming to apps too?

HarryFischer:
@abeddow91 Sent you an email about this, would be good to chat before this is merged in

dotcom-rendering/src/web/components/CaptionToggle.tsx
```diff
@@ -0,0 +1,73 @@
...
+		{/* Toggle contains the input with id #the-checkbox */}
+		<>
+			<label
+				htmlFor="the-checkbox"
+				css={[
+					toggleStyles,
+					mediaType === 'Video' && alignTopRightStyles,
+				]}
+			>
+				<svg width="6" height="14" fill="white" viewBox="0 0 6 14">
```
abeddow91, line 64: This will be replaced by source once the icon is added to source

dotcom-rendering/src/web/components/Caption.tsx
```diff
@@ -147,6 +147,15 @@ const hideIconBelowLeftCol = css`
 	}
 `;
 
+const videoCaptionOverlayStyles = css`
```
oliverlloyd, line 150: Is this change related to repositioning the main media?

dotcom-rendering/src/web/components/CaptionToggle.tsx
```diff
@@ -0,0 +1,73 @@
...
+		fill: white;
+	}
+`;
+
+const alignTopRightStyles = css`
+	bottom: auto;
+	top: 5px;
+`;
+
+export const CaptionToggle = ({
```
oliverlloyd, line 47: Is this part of the work to reorder the elements?

dotcom-rendering/src/web/components/MainMedia.tsx
```diff
@@ -48,15 +48,22 @@ const immersiveWrapper = css`
...
 		case ArticleDisplay.Standard: {
 			switch (format.design) {
 				case ArticleDesign.LiveBlog:
 				case ArticleDesign.DeadBlog:
-					return '';
+					if (showMediaAboveHeadline) {
+						return noGutters;
+					} else {
+						return '';
+					}
```
oliverlloyd, line 66: As above, where possible we should avoid nesting booleans into switch statements for format if `format` already provides this data (is a live blog). I think the only decision needed here is the temporary one of if the reader is in the test bucket or not

dotcom-rendering/src/web/components/YoutubeBlockComponent.importable.tsx
```diff
@@ -88,11 +90,29 @@ export const YoutubeBlockComponent = ({
 	duration,
 	origin,
 	stickyVideos,
+	showOverlayCaption,
 }: Props) => {
 	const [consentState, setConsentState] = useState<ConsentState | undefined>(
 		undefined,
 	);
 
+	useEffect(
```
marjisound, line 99: If I remember correctly, we usually try to create functions outside of the useeffect so that it'll be created once and then the reference of the function is used in the useeffect. Maybe something like this: 

```
	const hideOverlayCaptionOnPlay = (showCaption: boolean | undefined) => {
		const hideOverlayCaption = () => {
			if (showCaption) {
				const caption: HTMLElement | null =
					document.querySelector('.overlay-caption');
				if (caption) caption.style.display = 'none';
			}
		};
		document.addEventListener('video:play', hideOverlayCaption);

		return () =>
			document.removeEventListener('video:play', hideOverlayCaption);
	};

	useEffect(
		() => hideOverlayCaptionOnPlay(showOverlayCaption),
		[showOverlayCaption],
	);
```

dotcom-rendering/src/web/components/YoutubeBlockComponent.importable.tsx
```diff
@@ -228,6 +254,24 @@ export const YoutubeBlockComponent = ({
 					isMainMedia={isMainMedia}
 				/>
 			)}
+
+			{showOverlayCaption && (
```
marjisound, line 258: What if `hideCaption` is true in this component? Are we still going to show the caption?

dotcom-rendering/src/web/components/YoutubeBlockComponent.importable.tsx
```diff
@@ -88,11 +90,29 @@ export const YoutubeBlockComponent = ({
...
+					if (caption) caption.style.display = 'none';
+				}
+			};
+			document.addEventListener('video:play', hideOverlayCaption);
+
+			return () =>
+				document.removeEventListener('video:play', hideOverlayCaption);
+		},
+		[showOverlayCaption],
+	);
```
oliverlloyd, line 114: I think this logic needs to be part of the YouTubeBlock component itself. We already have [code there](https://github.com/guardian/atoms-rendering/blob/main/src/YoutubeAtom.tsx#L96) to react to the play event and hide overlays.

marjisound (commented):
It looks great, well done  Just left a few minor questions

oliverlloyd (changes requested):
Apologies for the 'request changes' but I don't think we should be adding caption overlays to videos inside DCR. We already have an overlay as part of the YouTube component that we import. If there's a requirement to enhance that then that change should happen in that component. Very happy to work with you on this.

In general, regarding the re-ordering of components. Is there an opportunity to use the [css order](https://developer.mozilla.org/en-US/docs/Web/CSS/order) property here?

JamieB-gu:
> Is there an opportunity to use the [css order](https://developer.mozilla.org/en-US/docs/Web/CSS/order) property here?

Just a reminder that `order` is visual only, and will not change the order of content for assistive technology (screen readers): https://developer.mozilla.org/en-US/docs/Web/CSS/order#accessibility_concerns

dotcom-rendering/src/web/components/Caption.tsx
abeddow91: Yes, it is. As the video has moved up, we now don't have the caption below the video. Instead, it appears as an overlay on the top of the video. Here's the design in a screengrab
Screenshot 2022-08-10 at 09 33 58

dotcom-rendering/src/web/components/CaptionToggle.tsx
abeddow91: Yes, as mentioned above this is because we now have a caption toggle on the video. As this is used in multiple olaces now, I thought it made sense to refactor it into its own component but can refactor if this isn't DCR convention.

dotcom-rendering/src/web/components/MainMedia.tsx
abeddow91: Good point! I'll refactor this!

dotcom-rendering/src/web/components/YoutubeBlockComponent.importable.tsx
abeddow91: Perfect, great tip! Thanks Marjan. I'll refactor this

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -88,9 +88,10 @@ const HeadlineGrid = ({ children }: { children: React.ReactNode }) => (
 					grid-template-areas: 'title	headline';
 				}
 				${until.desktop} {
-					grid-template-columns: 1fr; /* Main content */
+					grid-template-columns: minmax(0, 1fr); /* Main content */
```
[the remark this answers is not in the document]
abeddow91: Yes, adding in the 0 sets a minimum. By doing so, we prevent grid blowout which is where content that can scale breaks out of the constraints on the grid as there is no minimum width set.

dotcom-rendering/src/web/components/Caption.tsx
```diff
@@ -182,7 +184,7 @@ const videoIconStyle = css`
 
 const captionLink = (palette: Palette) => css`
 	a {
-		color: ${palette.text.captionLink};
+		color: white;
```
abeddow91, line 187:
```suggestion
		color: ${palette.text.captionLink};
```

dotcom-rendering/src/web/components/ArticleHeadline.tsx
```diff
@@ -731,7 +733,9 @@ export const ArticleHeadline = ({
 											color: ${isMatch
 												? palette.text.headlineWhenMatch
 												: palette.text.headline};
-											padding-bottom: ${space[9]}px;
+											padding-bottom: ${isMainMediaTest
```
abeddow91, line 736: This is being checked with Design to see if we need to make this change as 32px is not source standard.

dotcom-rendering/src/web/components/Caption.tsx
```diff
@@ -208,14 +211,20 @@ const CameraIcon = ({ palette, format }: IconProps) => {
...
 				iconStyle(palette),
 				format.display === ArticleDisplay.Immersive &&
 					hideIconBelowLeftCol,
 				videoIconStyle,
+				// Temporary for AB test. Once complete this will be moved to palette
+				format.design === ArticleDesign.LiveBlog &&
+					isMainMediaTest &&
+					css`
+						fill: ${neutral[100]};
+					`,
```
oliverlloyd, line 227: The risk here is that this code lives on. I think we've all universally had similar intentions to go back and fix things later but then not done it. It's really incredible how quickly and easily we move onto the next, more important, thing and leave these todos hanging.

dotcom-rendering/src/web/components/ElementContainer.tsx
```diff
@@ -29,7 +30,25 @@ const topBorderStyles = (colour: string) => emoCss`
...
+	isMainMediaTest?: boolean,
+) => {
+	if (format?.design === ArticleDesign.LiveBlog && isMainMediaTest) {
+		return emoCss`
+		background-color: ${neutral[97]};
+		${from.desktop} {
+			background-color: ${colour};
+		}
+	`;
+	}
```
oliverlloyd, line 45: We should be making colour decisions in `decidePalette` 🙏

dotcom-rendering/src/web/components/ImageComponent.tsx
```diff
@@ -379,7 +381,15 @@ export const ImageComponent = ({
...
-					<Hide when="above" breakpoint="tablet">
+					<Hide
+						when="above"
+						breakpoint={
+							format.design === ArticleDesign.LiveBlog &&
+							isMainMediaTest
+								? 'desktop'
+								: 'tablet'
+						}
+					>
```
oliverlloyd, line 392: This seems to be adding a lot of complexity, is this really needed? It would be great is both the code and the designs were simpler as complexity on both sides has costs which tend to endure for years and these costs should be weighed against the benefits to adding this variation

dotcom-rendering/src/web/components/MainMedia.tsx
```diff
@@ -49,13 +49,26 @@ const immersiveWrapper = css`
...
+						return css`
+							${until.desktop} {
+								margin-left: -20px;
+								margin-right: -20px;
+							}
+
+							${until.mobileLandscape} {
+								margin-left: -10px;
+								margin-right: -10px;
+							}
```
oliverlloyd, line 69: It feels strange that this is not the same as `noGutters`. This variation on breakpoint seems as though it might be a very small design tweak which is exploding complexity

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -409,7 +412,9 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 					<Section
 						showTopBorder={false}
 						backgroundColour={palette.background.matchNav}
+						innerBackgroundColour={palette.background.matchNav}
 						borderColour={palette.border.headline}
+						showSideBorders={isMainMediaTest ? false : true}
```
oliverlloyd, line 417: 👍 thanks for making this explicit, this reads really well

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -452,13 +459,34 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 								}
 							/>
 						</Island>
+						{isMainMediaTest && (
+							<Hide from="desktop">
```
oliverlloyd, line 463: This approach of hiding the main media means this html will exist twice on the page. 

This has impacts in terms of page weight and could be a risk in terms of other code looking for these elements (and not expecting to find more than one).

Could we look to use the [css order property](https://developer.mozilla.org/en-US/docs/Web/CSS/order) here instead?

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -452,13 +459,34 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
...
 				) : (
 					<Section
 						fullWidth={true}
 						showTopBorder={false}
 						backgroundColour={palette.background.header}
+						innerBackgroundColour={palette.background.header}
+						showSideBorders={isMainMediaTest ? false : true}
 						borderColour={palette.border.headline}
+						format={format}
+						isMainMediaTest={isMainMediaTest}
```
oliverlloyd, line 489: I can't see this border vary in the screenshots, can you elaborate more on why the test is affecting this?

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -966,11 +1060,7 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 											</ArticleContainer>
 										</div>
 									) : (
-										<Accordion
-											supportsDarkMode={false}
-											accordionTitle="Live feed"
-											context="liveFeed"
-										>
+										<div css={paddingBody}>
```
oliverlloyd, line 1063: Is the removal of the accordion something we could break out?

abeddow91: yep - i'll revert this

dotcom-rendering/src/web/layouts/LiveLayout.tsx
abeddow91: Sure! The borders are shown by default. Previously we've shown a vertical bleed next to the image on tablet which bleeds into the borders. The new designs remove this bleed but without removing the borders we have a thin pillar line around the image. Here's some pictures to show more clearly...

`showSideBorders={true}` - default
Screenshot 2022-09-20 at 09 21 27

`showSideBorders={false}`
Screenshot 2022-09-20 at 09 20 56

dotcom-rendering/src/web/layouts/LiveLayout.tsx
abeddow91: Thanks for this, I can see this  and definitely happy to chat about this further.

I have taken a look at css order and, from what I understand, I don't think it's suitable here for the following:
Firstly, if i understand the property correctly, the elements need to be within the same contain which wouldn't be the case here; below desktop is nested within the headline grid whilst above desktop is nested within the live grid. 
Secondly, as @JamieB-gu mentioned, order is visual only, and will not change the order of content for assistive technology. This is a [known concern ](https://developer.mozilla.org/en-US/docs/Web/CSS/order#accessibility_concerns) for this property and means we'll have a different experience for users using assistive technology

dotcom-rendering/src/web/components/ArticleHeadline.tsx
abeddow91: design are happy to align with source here.

dotcom-rendering/src/web/components/Caption.tsx
abeddow91: Yep, this is totally reasonable. Would having a removal PR ready before we merge these changes in abate these concerns?

PR #5394, opened 2022-07-14
Use HTTP3 for downloading Assets if part of HTTP3 experiment

AshCorr:
## What does this change?

Paired on with @paulmr 

As part of an experiment that @paulmr is working on we'd like to try and load some of our JS and other assets over a HTTP3 connection. This is done by appending ?http3 to our asset urls.

There will be a corresponding change made over on the Assets VCL to support this query parameter.

## Why?

Fastly has recently added support for HTTP3 connections on their CDN and we've been asked to trial it out. In theory we should be able to improve stats such as TTI as HTTP3 uses the QUIC protocol for transferring assets. QUIC offers, among other things, improved multiplexing (Parallel downloads over a single connection) and a much faster handshake for establishing connections.

We recently added a new experiment which users can opt into and receive HTTP3 on the main www.theguardian.com domain, unfortunately this doesn't work for our other domains such as Assets and Images as we can't access a users experiment cookies in the VCL for those domains. As a workaround we'd like to append ?http3 to our asset URL's if we detect a user is part of the experiment, this will then allow us to check for that query parameter in the Asset VCL and offer HTTP3 if its detected.

HTTP3 is backwards compatible with older browsers, so in theory we can just enable it for everyone, we'd like to try and get some measure of its impact before we do that though.

This article describes HTTP3 far better than I ever could and includes some nice benchmarks https://requestmetrics.com/web-performance/http3-is-fast

## This isn't an ideal solution

Ideally we'd like to be able to test our entire site using HTTP3, (especially images!) but besides just turning on HTTP3 by default we couldn't really think of a great way of doing this.

One alternative might be to modify the DOM at the edge and rewrite any URL's to point to their HTTP3 alternatives but that doesn't really deal with URL's in JS and might skew our experiment by adding extra page load time to mutate the page.

### DIFF
--- a/dotcom-rendering/src/lib/assets.ts
+++ b/dotcom-rendering/src/lib/assets.ts
@@ -49,7 +49,6 @@ export const ASSET_ORIGIN = decideAssetOrigin(process.env.GU_STAGE, isDev);
 
 export const getScriptArrayFromFile = (
 	file: `${string}.js`,
-	offerHttp3: boolean,
 ): { src: string; legacy?: boolean }[] => {
 	if (!file.endsWith('.js'))
 		throw new Error('Invalid filename: extension must be .js');
@@ -60,17 +59,16 @@ export const getScriptArrayFromFile = (
 		: legacyManifest[file];
 
 	const scripts = [];
-	const http3Suffix = offerHttp3 ? '?http3' : '';
 
 	if (filename) {
 		scripts.push({
-			src: `${ASSET_ORIGIN}assets/${filename}${http3Suffix}`,
+			src: `${ASSET_ORIGIN}assets/${filename}`,
 			legacy: false,
 		});
 	}
 	if (legacyFilename) {
 		scripts.push({
-			src: `${ASSET_ORIGIN}assets/${legacyFilename}${http3Suffix}`,
+			src: `${ASSET_ORIGIN}assets/${legacyFilename}`,
 			legacy: true,
 		});
 	}
--- a/dotcom-rendering/src/lib/assets.ts
+++ b/dotcom-rendering/src/lib/assets.ts
@@ -49,6 +49,7 @@ export const ASSET_ORIGIN = decideAssetOrigin(process.env.GU_STAGE, isDev);
 
 export const getScriptArrayFromFile = (
 	file: `${string}.js`,
+	offerHttp3: boolean,
 ): { src: string; legacy?: boolean }[] => {
 	if (!file.endsWith('.js'))
 		throw new Error('Invalid filename: extension must be .js');
@@ -59,16 +60,17 @@ export const getScriptArrayFromFile = (
 		: legacyManifest[file];
 
 	const scripts = [];
+	const http3Suffix = offerHttp3 ? '?http3' : '';
 
 	if (filename) {
 		scripts.push({
-			src: `${ASSET_ORIGIN}assets/${filename}`,
+			src: `${ASSET_ORIGIN}assets/${filename}${http3Suffix}`,
 			legacy: false,
 		});
 	}
 	if (legacyFilename) {
 		scripts.push({
-			src: `${ASSET_ORIGIN}assets/${legacyFilename}`,
+			src: `${ASSET_ORIGIN}assets/${legacyFilename}${http3Suffix}`,
 			legacy: true,
 		});
 	}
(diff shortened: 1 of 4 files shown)


dotcom-rendering/src/web/server/articleToHtml.tsx
```diff
@@ -117,6 +117,10 @@ export const articleToHtml = ({ data }: Props): string => {
 			'model.dotcomrendering.pageElements.TweetBlockElement',
 	);
 
+	// Evaluating the performance of HTTP3 over HTTP2
+	// See: https://github.com/guardian/dotcom-rendering/pull/5394
+	const offerHttp3 = CAPIArticle.config.switches.offerHttp3;
```
mxdvl, line 122:
```suggestion
	const { offerHttp3 } = CAPIArticle.config.switches;
```

dotcom-rendering/src/web/server/frontToHtml.tsx
```diff
@@ -77,21 +77,26 @@ export const frontToHtml = ({ front, NAV }: Props): string => {
 	const polyfillIO =
 		'https://assets.guim.co.uk/polyfill.io/v3/polyfill.min.js?rum=0&features=es6,es7,es2017,es2018,es2019,default-3.6,HTMLPictureElement,IntersectionObserver,IntersectionObserverEntry,URLSearchParams,fetch,NodeList.prototype.forEach,navigator.sendBeacon,performance.now,Promise.allSettled&flags=gated&callback=guardianPolyfilled&unknown=polyfill&cacheClear=1';
 
+	//
+	// See :https://github.com/guardian/dotcom-rendering/pull/5394
+	const offerHttp3 = front.config.switches.offerHttp3;
```
mxdvl, line 82: same as above

mxdvl (commented):
Instead of changing the signature of getScriptArrayFromFile, I would probably simply map over the `src` of each file?

```ts
const priorityScriptTags = generateScriptTags([
    { src: polyfillIO },
    ...getScriptArrayFromFile('bootCmp.js'),
    ...getScriptArrayFromFile('ophan.js'),
    /* … */
    ...getScriptArrayFromFile('islands.js'),
    ...expeditedIslands.flatMap((name) =>
		...expeditedIslands.flatMap((name) =>
        getScriptArrayFromFile(`${name}.js`),
    ),
	].map((script) => offerHttp3 ? ({ ...script, src: `${script.src}?http3` }) : script)
)
```

dotcom-rendering/src/web/server/frontToHtml.tsx
AshCorr, line 82:
```suggestion
	// Evaluating the performance of HTTP3 over HTTP2
	// See: https://github.com/guardian/dotcom-rendering/pull/5394
	const { offerHttp3 } = CAPIArticle.config.switches;
```

AshCorr:
> Instead of changing the signature of getScriptArrayFromFile, I would probably simply map over the `src` of each file?

This is much cleaner, I was a bit worried this might not work since this would apply the ?http param to the commercial and polyfill bundles, but these also both come from assets.guim.co.uk so it should be fine! One slight problem is that we need to account for URL's having other query parameters now (polyfill) so we can't just blindly add more `?`'s into the URL.

dotcom-rendering/src/lib/assets.ts
```diff
@@ -60,17 +59,16 @@ export const getScriptArrayFromFile = (
...
 	if (filename) {
 		scripts.push({
-			src: `${ASSET_ORIGIN}assets/${filename}${http3Suffix}`,
+			src: `${ASSET_ORIGIN}assets/${filename}`,
 			legacy: false,
 		});
 	}
 	if (legacyFilename) {
 		scripts.push({
-			src: `${ASSET_ORIGIN}assets/${legacyFilename}${http3Suffix}`,
```
AshCorr: Realistically legacy browsers won't even support HTTP3, so I don't think we really need to add it to the legacy bundles anyway.

dotcom-rendering/src/web/lib/getHttp3Url.ts
```diff
@@ -0,0 +1,7 @@
+// Evaluating the performance of HTTP3 over HTTP2, this file can be removed once the experiment is concluded.
+// See: https://github.com/guardian/dotcom-rendering/pull/5394
+export const getHttp3Url = (url: string): string => {
+	const urlObject = new URL(url);
+	urlObject.searchParams.append('http', 'true');
```
mxdvl, line 5:
```suggestion
	urlObject.searchParams.append('http3', 'true');
```

mxdvl (approved):

dotcom-rendering/src/web/lib/getHttp3Url.ts
AshCorr: Good catch 😅

PR #5396, opened 2022-07-14
[DevTime]: Extract types out of ambient declaration file

mxdvl:
## What does this change?

Refactor validate to make it simpler and stricter.

Extract types related to front to `src/types/front.ts` and import them explicitely where needed.

## Why?

The `index.d.ts` is extremely long, and contains nearly all kinds of types, making it hard to understand where they are used, and potentially creating conflicts between types declared locally.

We favour being explicit rather than implicit. Most of the frontend types are only imported at the outer edge of the application.


oliverlloyd (approved):
Very nice, much less magical

PR #5397, opened 2022-07-14
Dblatcher/update gen schema after merge 5291

dblatcher:
## What does this change?
Updates the schema and fixtures.

## Why?

All checks passed for https://github.com/guardian/dotcom-rendering/pull/5291, but after merging, there was a schema check warning on main: https://github.com/guardian/dotcom-rendering/actions/runs/2671630751

I'm unsure how this happens.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |


AshCorr:
Might be because we had a bit of an accident on main (#5393), thanks for fixing this nonetheless!

dblatcher:
> It started breaking before your PR was merged it seems, I think it might be because we had a bit of an accident on main earlier(#5393), thanks for fixing this nonetheless!

Thanks - I thought I'd broken something and didn't know how! I'll go through the UI Tests and approve them if it looks right - I think it should just be changes to the content used on the fixtures.

PR #5404, opened 2022-07-15
Add styling to <span> wrapper on byline date

### LINKED ISSUE #5025: Add underline to dateline on Safari
Can't work out why this doesn't get applied in Safari on hover
Has both 
```
-webkit-text-decoration: underline;
text-decoration: underline;
```

Image

bryophyta:
## What does this change?

- Adds a `<span>` to wrap the contents of the `<summary>` element
- Add the hover styling directly to the span, instead of the summary

## Why?

The intention here is to fix a bug whereby the underline wasn't being displayed on Safari. 

It's possible that the original has something to do with the way that Safari renders `<summary>` elements in general, where it uses a webkit pseudo-element for the summary marker.

Closes #5025 

## Potential limitations

Unfortunately I haven't been able to properly work out why the original bug occurred, so this is a bit of a workaround, but I don't think that the fix is likely to have unintended side-effects, so it seems reasonable to use the workaround, rather than sinking time into working out where the bug came from.

## Screenshots

| Before      | After      |
|-------------|------------|
| image| image |

### DIFF
--- a/dotcom-rendering/src/web/components/Dateline.tsx
+++ b/dotcom-rendering/src/web/components/Dateline.tsx
@@ -16,11 +16,14 @@ const datelineStyles = css`
 const primaryStyles = css`
 	list-style: none;
 	cursor: pointer;
-	:hover {
-		text-decoration: underline;
-	}
 	&::-webkit-details-marker {
 		display: none;
+	}
+`;
+
+const hoverUnderline = css`
+	:hover {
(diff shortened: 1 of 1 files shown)


mxdvl (approved):
Nice fix!

Agreed that figuring out the exact reason for the issue migth not be worth investing too much time into, but we could always flag a bug on WebKit.

PR #5406, opened 2022-07-15
chore: Restrics use of props on FrontCard

AshCorr:
## What does this change?

I merged down #5364 so that people could start using the new FrontCard component to make layouts but there was an outstanding point on weather we should restrict what props can be used in the component.

Follow up on https://github.com/guardian/dotcom-rendering/pull/5364#discussion_r920428971

I have mixed feelings about this, I feel like the FrontCard should work identically to a normal Card but accept an additional `trails` property to auto-fill all the props, at the same time I think the intellisense in VSCode works better if we're explicit about what props a Component accepts instead of using a `Partial<CardProps>`

### DIFF
--- a/dotcom-rendering/src/web/components/FrontCard.tsx
+++ b/dotcom-rendering/src/web/components/FrontCard.tsx
@@ -4,7 +4,26 @@ import { Card } from './Card/Card';
 
 type Props = {
 	trail: TrailType;
-} & Partial<CardProps>;
+	headlineSize?: SmallHeadlineSize;
+	showQuotes?: boolean;
+	showByline?: boolean;
+	imageUrl?: string;
+	imagePosition?: ImagePositionType;
+	imagePositionOnMobile?: ImagePositionType;
+	imageSize?: ImageSizeType;
+	trailText?: string;
+	avatar?: AvatarType;
+	kickerText?: string;
+	showPulsingDot?: boolean;
+	showSlash?: boolean;
+	supportingContent?: DCRSupportingContent[];
+	snapData?: DCRSnapType;
+	containerPalette?: DCRContainerPalette;
+	containerType?: DCRContainerType;
+	showAge?: boolean;
+	discussionId?: string;
+	isDynamo?: true;
+};
 
 /**
  * A wrapper around the normal Card component providing sensible defaults for Cards on front containers.


dotcom-rendering/src/web/components/FrontCard.tsx
```diff
@@ -4,7 +4,26 @@ import { Card } from './Card/Card';
 
 type Props = {
 	trail: TrailType;
-} & Partial<CardProps>;
+	headlineSize?: SmallHeadlineSize;
+	showQuotes?: boolean;
+	showByline?: boolean;
```
oliverlloyd, line 9: I think we can remove these. We probably only ever want to respect the values set by the tool. If this changes later we can add them in

dotcom-rendering/src/web/components/FrontCard.tsx
```diff
@@ -4,7 +4,26 @@ import { Card } from './Card/Card';
...
+	imageUrl?: string;
+	imagePosition?: ImagePositionType;
+	imagePositionOnMobile?: ImagePositionType;
+	imageSize?: ImageSizeType;
+	trailText?: string;
+	avatar?: AvatarType;
```
oliverlloyd, line 15: I think this can be removed. We only show this for certain `format`s and only if it is present

dotcom-rendering/src/web/components/FrontCard.tsx
```diff
@@ -4,7 +4,26 @@ import { Card } from './Card/Card';
...
+	kickerText?: string;
```
oliverlloyd, line 16: Can be removed. I _think_ we simply show kicker text when given by the tool?

dotcom-rendering/src/web/components/FrontCard.tsx
```diff
@@ -4,7 +4,26 @@ import { Card } from './Card/Card';
...
+	showPulsingDot?: boolean;
+	showSlash?: boolean;
```
oliverlloyd, line 18: I think the defaults should be rules for these. I don't think we have any use cases where we would want to override these rules.

dotcom-rendering/src/web/components/FrontCard.tsx
```diff
@@ -4,7 +4,26 @@ import { Card } from './Card/Card';
...
+	supportingContent?: DCRSupportingContent[];
+	snapData?: DCRSnapType;
+	containerPalette?: DCRContainerPalette;
```
oliverlloyd, line 21: Should we just always pass this is as a required field?

dotcom-rendering/src/web/components/FrontCard.tsx
```diff
@@ -4,7 +4,26 @@ import { Card } from './Card/Card';
...
+	containerType?: DCRContainerType;
+	showAge?: boolean;
+	discussionId?: string;
```
oliverlloyd, line 24: I think this can just be the default trail value

oliverlloyd (approved):
I think we can reduce this list a bit. My guess is for a lot of these values the default values can simply be the static values and they don't need to be overridable

mxdvl (commented):
I’d be tempted to start with a restricted set of overridable keys, with `Partial<Pick<CardProps, "…">>`, like so:

```ts
type Props = {
	trail: TrailType;
} & Partial<
	Pick<
		CardProps,
		| 'showQuotes'
		| 'showByline'
		| 'imageUrl'
		| 'imagePosition'
		| 'imagePositionOnMobile'
		| 'imageSize'
		| 'trailText'
		| 'avatar'
		| 'kickerText'
		| 'showPulsingDot'
		| 'showSlash'
		| 'supportingContent'
		| 'snapData'
		| 'containerPalette'
		| 'containerType'
		| 'showAge'
		| 'discussionId'
		| 'isDynamo'
	>
>;
```

PR #5411, opened 2022-07-15
Fix fluid ad width on mobile

Jakeii:
## What does this change?
Another follow up to #5382 to fix fluid ad widths.

Uses flex box so that ads of any size can be centred on mobile.

## Screenshots

| Before      | After      |
|-------------|------------|
| Screenshot 2022-07-15 at 16 12 31 | Screenshot 2022-07-15 at 16 12 11 |

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleContainer.tsx
+++ b/dotcom-rendering/src/web/components/ArticleContainer.tsx
@@ -70,28 +70,21 @@ const adStyles = css`
 		margin: ${space[3]}px auto;
 		/* this is centring the ad iframe as they are display: inline; elements by default */
 		text-align: center;
-
-		${from.mobile} {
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/components/ArticleContainer.tsx
```diff
@@ -70,30 +70,23 @@ const adStyles = css`
 		margin: ${space[3]}px auto;
 		/* this is centring the ad iframe as they are display: inline; elements by default */
 		text-align: center;
-
-		${from.mobile} {
-			width: 300px;
-		}
+		display: flex;
```
arelra, line 73: Great to see flexbox finally make an appearance in our CSS!

arelra (approved):
🎉

PR #5413, opened 2022-07-15
Dblatcher/sign up tracking format and styling

dblatcher:
## What does this change?

 - changes the format of the ophan component event sent by `SecureSignupIframe` to include UTC timestamps and event descriptions matching those used on similar frontend components
 - adjust internal padding and margins of `EmailSignup` to fit design
 - change to neutral colour for the links in the privacy text
 - create a separate component for the Privacy text 

## Why?
 - keeps the tracking data provided to the data team consistent
 - aligns to design input
 - reduces duplication between `SecureSignup` and `SecureSignupMock`

## Screenshots

| Before      | After      |
|-------------|------------|
| oldMobile | newMobile |
| desktopOld | desktopNew |

## Design brief
Screenshot 2022-07-15 at 17 11 59


dblatcher:
> It seems like there are multiple unrelated things happening in this PR. Would it be possible to separate the work out into multiple PRs?
> 
> I know this is an annoying comment to get but it's not just about helping reviewers, this also means the git history for this codebase is easier to consume, future developers trying to learn about our motivations will thank us! It's not just that though, if there are any issues with this code it will be easier to debug/revert if the code is more targeted.

Sure, sorry should have done that to begin with.

PR #5420, opened 2022-07-18
Remove `OnwardsLower`

oliverlloyd:
## What does this change?
This PR removes the `OnwardsLower` component and deletes all associates files, tests and stories

## Why?
When we [migrated from onwards upper and lower to using the carousel](https://github.com/guardian/dotcom-rendering/pull/3172), we put both the top and bottom rows for the onwards content into the `OnwardsUpper` component, so we now see two carousels on most articles, a top and a bottom one. But we didn't completely remove the `OnwardsLower` component and so it sometimes (an article must be both part of a series and appear in a story package) pops up causing _three_ containers to show.

This PR completes that migration work and removes the possibility of three onwards containers showing

This simplification will also make any future work in this area a lot easier as we will have fewer use cases to support and understand.

### But what if I want to use different card layouts later, like `MoreThanFive`?
Now that we are building card layout container components as part of the fronts migration we should focus on using that code. If a different card layout were wanted at the bottom of articles we could either repurpose an existing front container or create a new one using the same patterns.

### Labs stories now use `Carousel`
As a consequence of this change onwards content on labs articles, which was excluded from using the Carousel for testing purposes, will now also appear inside a carousel, and not in the older style container that was previously being used

| Before      | After      |
|-------------|------------|
| Screenshot 2022-07-18 at 09 49 22 | Screenshot 2022-07-18 at 10 11 09 |


oliverlloyd:
It's been agreed with the Labs team + Comm dev (@Amouzle ) that we will make this change but that we should keep an eye on the numbers to ensure we don't degrade engagement. We don't expect this to be the case as we already have positive results based on AB testing the carousel.

OllysCoding (approved):
+1 Win!

PR #5422, opened 2022-07-18
Fix inter-scroller ads stealing mouse events from around them

Jakeii:
## What does this change?
Adds `overflow: hidden;` to inline ad slots.

## Why?
Setting `overflow: hidden;` will prevent the contents stealing mouse events outside of it's bounds.


dotcom-rendering/src/web/components/ArticleContainer.tsx
```diff
@@ -75,18 +75,23 @@ const adStyles = css`
...
 		}
 
+		.ad-slot {
+			${from.tablet} {
+				/* from tablet the ad slot will stretch to the full width of the container and the iframe will be centred by the text-align: center; on the container */
+				flex: 1;
+			}
+
+			/* this fixes inter-scrollers stealing mouse events */
+			overflow: hidden;
```
Jakeii, line 92: This should not have any effect on any other kinds of ads AFAIK

ashishpuliyel (approved):
Pretty neat, if this is all it takes to fix it!

arelra (approved):

PR #5425, opened 2022-07-18
fix: Height of flexbox ad slot

chrislomaxjones:
Co-authored-by: Jake Lee Kennedy <jake.kennedy@guardian.co.uk>

## What does this change?

Apply `align-self: flex-start` to ad slots.

## Why?

Applying flex to the container causes the height of the child element (the slot) to fill the full height of the container. Applying this changes causes the slot to take on `auto` height, which allows it to be sticky.

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleContainer.tsx
+++ b/dotcom-rendering/src/web/components/ArticleContainer.tsx
@@ -79,6 +79,8 @@ const adStyles = css`
 			.ad-slot {
 				/* from tablet the ad slot will stretch to the full width of the container and the iframe will be centred by the text-align: center; on the container */
 				flex: 1;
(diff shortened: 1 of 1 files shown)


Jakeii (approved):

dotcom-rendering/src/web/components/ArticleContainer.tsx
```diff
@@ -79,6 +79,7 @@ const adStyles = css`
 			.ad-slot {
 				/* from tablet the ad slot will stretch to the full width of the container and the iframe will be centred by the text-align: center; on the container */
 				flex: 1;
+				align-self: flex-start;
```
arelra, line 82: Can we add a comment please to explain the magic :)

chrislomaxjones: You've got it!

PR #5426, opened 2022-07-18
Fix sport liveblog header text colour

bryophyta:
## What does this change?

Adds a new prop, `isMatch`, to ArticleHeadline.tsx, which allows it to use a different text colour for match liveblogs.

## Why?

Before, the headline was rendered as white text on a yellow background.

The solution of adding `isMatch` and `palette.text.headlineWhenMatch` was chosen because it follows the pattern established by the `seriesTitleWhenMatch` property (h/t @OllysCoding).

## Screenshots

| Before      | After      |
|-------------|------------|
| image| image |

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -83,6 +83,7 @@ type Colour = string;
 type Palette = {
 	text: {
 		headline: Colour;
+		headlineWhenMatch: Colour;
 		seriesTitle: Colour;
 		sectionTitle: Colour;
 		seriesTitleWhenMatch: Colour;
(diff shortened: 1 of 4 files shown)


oliverlloyd (approved):
Nice. +1 for following the established pattern, I think this is the right route forward for this problem at this point in time.

For future Dotcommers looking at perhaps a similar issue, it might be worth considering a deeper approach. Ideally, we would encapsulate this state using `format`. We decide if a piece is a 'match' based on `CAPIArticle.matchType === 'FootballMatchType' && CAPIArticle.matchUrl` which means this decision could be done further up the chain, maybe in an enhancer to start, and then assigned to a new `Design` type.

bryophyta:
> For future Dotcommers looking at perhaps a similar issue, it might be worth considering a deeper approach. Ideally, we would encapsulate this state using `format`. We decide if a piece is a 'match' based on `CAPIArticle.matchType === 'FootballMatchType' && CAPIArticle.matchUrl` which means this decision could be done further up the chain, maybe in an enhancer to start, and then assigned to a new `Design` type.

this sounds like a good idea to me 👍

PR #5427, opened 2022-07-18
`dynamic/fast` refactor - groups, layouts & boosting

OllysCoding:
## What is this PR?

This PR does too much 😡 (I'm sorry).

### Groups

We introduce the concept of _grouping_ to DCR. Dynamic layouts in fronts tool are configured with the cards put in different groups:

<details closed>
<summary>Example of fronts tool 'groups'</summary>
<br>
image
</details>

These groups are passed down from fronts tool into DCR as the `FEFrontCard.card.group`. Unhelpfully these groups are passed as ids (`"0"`, `"1"`, `"2"`, etc) - which represent different groups depending on the layout (`dynamic/slow` & `dynamic/fast` use "huge" (3) , "very big" (2), "big" (1) and "standard" (0), `dynamic/package` uses "snap" (1) and "standard" (0)).

This PR adds [groupCards.ts](https://github.com/guardian/dotcom-rendering/blob/olly/fronts-use-card-container-groups/dotcom-rendering/src/model/groupCards.ts) which is responsible for sorting the curated & backfilled cards into a `grouped` object which has arrays named after each group type:

<details closed>
<summary>DCRGroupedTrails</summary>
<br>
image
</details>

> Note: While I did consider that components could sort the groups themselves; an advantage of doing it further up the stack is that it helps make defining & reading UI stories simple & clear ([example](https://github.com/guardian/dotcom-rendering/blob/olly/fronts-use-card-container-groups/dotcom-rendering/src/web/components/DynamicFast.stories.tsx#L37))

### `dynamic/fast` Layout refactor

One of the most difficult parts of this PR was working out the logic to layout the dynamic/fast component. The existing logic in this component worked for a specific layout set (2 primaries, 1st boosted, 2 bigs, 6 standards) - however it didn't respect the groups and didn't work with the variety of layouts these groups can create.

<details closed>
<summary>Existing DCR implementation</summary>
<br>
This layout is a great starting point; but doesn't respect boosting, the various primary configurations, or the groups cards are given in fronts tool. This is what I set out to solve in this PR.
image
</details>

Putting together the layout in the PR came from a combination of testing how frontend looks with different configurations of dynamic/fast, as well as analysing the existing Frontend logic (located [here](https://github.com/guardian/frontend/blob/main/common/app/layout/slices/DynamicContainer.scala), and [here](https://github.com/guardian/frontend/blob/main/common/app/layout/slices/DynamicFast.scala)).

<details closed>
<summary><b>Primaries</b> (First Slice)</summary>
<br>
Primaries, known as 'optionalFirstSlice' in Frontend is the cards on first row of the container. There was some very clear pre-existing logic in Frontend we were able to use for this:
<ol>
<li>If there are any huges, use the first one for the entire first row</li>
<li>If there is 1 very big, use this for the entire first row</li>
<li>If there are 2 very bigs, take them both and split them 50/50 with horizontal space (boosting these cards will be covered later)</li>
</ol>

Any leftovers from the 'huge' and 'very big' groups are passed down to the 'bigs' group later in the rendering flow.

Primaries / the first slice will therefor like to look like this (not taking into account boosted layouts):

| 1 Card      | 2 Cards      | 
|-------------|------------|
| ![primaries1card][] | ![primaries2cards][] |

[primaries1card]: https://user-images.githubusercontent.com/9575458/179565540-2637d888-2b12-40bc-ae6c-3461aa3cd612.png
[primaries2cards]: https://user-images.githubusercontent.com/9575458/179565422-e1c70dbb-f8f0-46bb-9cb5-c9d9bda02206.png
</details>

<details closed>
<summary><b>Bigs & standards</b> </summary>
<br>

Left over after primaries, are any _huges_ and _very bigs_ not used - and the bigs & standards which we haven't looked at yet.

Some examples of this fall through logic would be: 
- There is 1 huge, 2 very bigs, and 3 bigs. 
    - The first huge is taken as the primary
    - The 2 very bigs and prepended at the start of the 3 bigs - giving us 5 total
- There are 3 very bigs, and 1 big
    - The first 2 very bigs are taken as primary
    - The remaining very big is prepended to the 1 existing big, giving us 2 total

There can be up to 4 bigs - after this point no more cards will be rendered, and it should fall through to 'show more' if enabled.

Each 'big' which is rendered takes up 25% of the horizontal space in the row. This dictates how many standard cards can be included in the front.

We show up to 3 standard cards per 1/4 of available space For example, with 3 bigs we can fit 3 standards:
image

When the max number of standards aren't provided, the container should adapt automatically (max 6, 5 provided):
image

</details>

### `dynamic/fast` Support boosting

'Boosts' are properties that can be applied to cards in Fronts tool, with a button looking something like:

image

This switch is available for all cards on fronts tool, but it's down to the platform to decide where to support a card being 'boosted' or not. (nb. we should look to move this logic to the tools as this will reduce complexity & increase platform consistency)

On `dynamic/fast` we support boosting of the 'first 3' cards. Colloquially these are referred to as 'first boosted', 'second boosted' and 'third boosted' - but these names aren't always factually accurate to the actual configuration. In the code we break it into 'primary boosted' and 'first big boosted'.

'primary boosted' and 'first big boosted' can be combined, as their layout impacts only affect their own row.

<details closed>
<summary><b>'Primary' boosting</b></summary>
<br>
There can either be 1 or 2 primaries, as reviewed earlier. Boosting primaries only has an effect when there are 2 primaries selected.

The default state for 2 primaries is a 50/50 split:
image

Either the first, or the second primary can be boosted, resulting in a 75/25 split or 25/75 split

First boosted:
image
Second boosted:
image

Boosting of primaries only affects the primary row, meaning the layout of all cards (and therefor the 'first big boosted') operates independently.

</details>

<details closed>
<summary><b>'First big' boosting</b></summary>
<br>
'First big' boosting is when the first card considered a 'big' has boost enabled - this is an important distinction as 'bigs' can fall through from the 'huge' or 'very big' categories depending on the configuration in fronts tools.

Boosting is only supported for the first big, if a 2nd or 3rd big is boosted it will be ignored

The unboosted layout with just 1 big would be:
(the left most card being a big, and the rest being standard)
image

A similar layout with just 1 big, but the big is boosted would be
(note: there are less standards chosen in this example)
image

As you can see, the first big being boosted doubles the width of the card to 50%, and moves the image to the top!

**First big + another big**

What happens when the first big is boosted and there are other bigs? counterintuitively instead of just stacking more 25% width bigs after the boosted big, the layout switches so the right 50% of the big row becomes columns - this column layout supports just one more 'big', (in a different layout than before) & all the other bigs are rendered as standards.

This layout looks like:
bigboosted+anotherbig explained

As you can see, the first big remains the same, but the remaining space has been shifted to a column based layout where:
- The next big is shown slightly larger than the standards, with an image
- Any remaining bigs are 'demoted' to standard
- standards & remaining bigs are stacked in 2 columns

The column layout is necessary due to the larger 'big' - if we used a row layout, the adjacent standard would be given as much vertical room.
</details>

## Screenshots

(More Coming soon hopefully)

| Frontend      | DCR Before      | DCR After      |
| ![frontend1][] | ![before1][] | ![after1][] |

[frontend]: https://user-images.githubusercontent.com/9575458/180793488-14eafc3f-1795-4fc0-a6b6-c44d74cc68b4.png
[before]: https://user-images.githubusercontent.com/9575458/180793568-fd9f90e3-ce31-469c-a591-8ddccedfcd4e.png
[after]: https://user-images.githubusercontent.com/9575458/180793621-0e1e23c2-b686-4831-bc8d-cd8e6462d2d0.png

[frontend1]: https://user-images.githubusercontent.com/9575458/180793874-92aee8f2-2c44-44ca-b379-f345d1a02559.png
[before1]: https://user-images.githubusercontent.com/9575458/180793941-b4caacc7-3a51-4d13-a19f-a9012de0dfaf.png
[after1]: https://user-images.githubusercontent.com/9575458/180794004-c8d66bfa-8177-4013-81b8-53c555087513.png

[More examples of DCR in this PRs storybook](https://5dfcbf3012392c0020e7140b-ozsoqdyhrt.chromatic.com/?path=/story/components-dynamicfast--df-2-very-bigs-2-bigs-5-standards)

### Sublinks
Sublinks will only show up where intended on `dynamic/fast` thanks to https://github.com/guardian/dotcom-rendering/pull/5364, however sublinks do not render correctly (in the eyes of Frontend) - as sometimes they will be stacked vertically where they would be horizontal Frontend. 

The solution to this would likely to be to give some optional control (`sublinkStacking?: "horizontal" | "vertical"`) in the `<Card />` API so containers can provide information on how to layout sublinks.

### `dynamic/slow` and `dynamic/package`

This PR introduces `DCRGroupedTrails` on `DCRCollectionType.grouped` which is able to group cards with support for `dynamic/slow` and `dynamic/package` - however the `DCRCollectionType.grouped` is not consumed by their respective components, as each may require a similar amount of work to that done in this PR (especially dynamic/slow). This will be left for future PRs.


dotcom-rendering/src/web/components/Card/components/LI.tsx
```diff
@@ -52,6 +52,10 @@ const decideSize = (percentage?: CardPercentageType, stretch?: boolean) => {
 	if (percentage) {
 		sizeStyle = css`
 			flex-basis: ${percentage};
+			${stretch &&
+			css`
+				flex-grow: 1;
+			`}
```
OllysCoding, line 58: Here I added the ability to enable 'stretch={true}` while also setting a percentage. This was needed to achieve some of the layouts required

dotcom-rendering/src/model/groupCards.ts
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * For 'dynamic' container types in fronts tool, cards can be grouped by sizes:
+ *  - Snap (dynamic/package only)
+ *  - Huge
+ *  - Very big
+ *  - Big
+ *  - Standard
+ *
+ * This can be consumed by the rendering layer to tell us how we should size each card, which
+ * can very by container type.
```
oliverlloyd, line 20:
```suggestion
 * can vary by container type.
```

dotcom-rendering/src/web/components/Card/components/LI.tsx
OllysCoding: (this seems to have had some side effects else where which I'm not looking into)

oliverlloyd (commented):
It looks like the dynamic fast container has a substantial number of possible variations based on different variables. My feeling is that we should consider defining a subset of these possibilities - based on usage - and only build in support for these instead of trying to capture all of them. I think there's appetite from all sides for this.

What do you think? Maybe we can work with Harry and the eds to try and create a contract?

OllysCoding:
> My feeling is that we should consider defining a subset of these possibilities

I've been thinking a lot about this while working on this component, and perhaps I've gotten too far into the logic, but I worry (especially from both the conversations we've had with editorial already, and from seeing usages on the site) that all of the features I've implemented here are used on a near daily basis.

I think if you tried to break down the layout into a set of 'use cases' as suggested, but we can't get rid of any of the features, we'll end up either implementing all the edge cases this PR covers, or leaving broken functionality around the outside of the use cases, which I believe is most likely to come from odd numbers of cards in different groups. 

---

I've still got some tidying to do in this PR to get the code where I want it, but I'm inclined to push for keeping the full functionality of this component for the time being - and allow the editorial/design simplification work to allow us to simplify this component more concretely in the future.

I'm also happy I've been able to catch all the variations in behaviour with the stories for the component - with storybook generally giving us a powerful tool here to understand exactly how the component works.

I'm open to whatever route we want to go down, but my gut reaction is that we should make this component as simple to work on as possible, but try and keep the functionality matching Frontend for now.

OllysCoding:
If there was one thing I'd like to get rid of which would help de-complicate the component though it would be the `ThirdBoostedPlusBig` layout which is just unnecessarily complicated and completely different than all the other layouts :/

Unfortunately I've seen it used multiple times in 'Headlines' on the UK front over the past few days

oliverlloyd:
> that all of the features I've implemented here are used on a near daily basis

I think you're right, I am very sure that there are active examples of strange layouts in use right now and that edge cases will also be used occasionally. And yes, the third boosted layout is _very_ common, I see this daily.

But how much of this is intended layout decisions and how much is just what happens based on how the tool works now? I am honestly unsure, maybe we could do some sort of audit to see how this container is used? I say maybe though because I think it's good to balance usage with code paths and also business desires - these all play a part in defining what we should build.

My hope is that there is only a limited number of variations required/in use and that some of these variations could be represented as their own layout in DCR. A good example of this is the third boosted option. This really should be its own layout. The problem with using 'boosting' to change the layout is it only works on certain cards - which is magical and confusing! But a separate layout file better represents things. If this works, then I think it's reasonable to extend the change up the stack and into the tool, adding the new layout there

> I'm inclined to push for keeping the full functionality of this component for the time being - and allow the editorial/design simplification work to allow us to simplify this component more concretely in the future.

What do you think about introducing some sort of enhancer pattern here?

My main worry is that this code is hard to reason about when coming at the file cold. This isn't a reflection on your work by the way, I just think the business logic here is super complex! But maybe we can abstract that weird logic into an enhancer and then the outcome could be a decision on which layout file to use where there would be additional layout files to capture these variations. E.g.: `DynamicFastThirdBoosted` (horrible name, but you get the idea).

An abstracted enhancer would also be easier to change / migrate away from later if we simplify.

Another advantage is this approach is this would allow is to keep `trails` as a single array. It's true that the raw data does contain this concept of big, huge, small, whatever, but this is a very non human way to read the data. The reality is some layouts do special things with the first or third cards, and this is related to these groups, but only loosely and inconsistently. I feel a single array better represents how my brain thinks about layouts, it's a list of cards laid out left to right, top to bottom where some are bigger based on layout position. That's all very hand wavy but I feel this is something you can communicate in words.

oliverlloyd:
I should also point out that this PR is an amazing piece of work! 💎

I tend to just assume you are going to churn out great PRs @OllysCoding and then dive straight in on the technical discussions but I completely recognise the amount of work that went into this and how impressive it is that you've been able to capture it both in your brain and in the code

OllysCoding:
Having discussed this more with @oliverlloyd and @HarryFischer - for the time being I think this complexity may be necessary, while we'd love this component to be simplified into even just a few different layouts - attempting this during the migration may add further to the overall complexity of the migration itself, even if it might simplify this component.

This component is difficult to approach at best, but we've definitely come away with some wins. This PR, the code & storybook/chromatic testing mean we are the closest to understanding how this component works than we have been in a long time, even if 30 or so minutes of head scratching is still needed.

---

I'd love some more reviews; especially on where the code can make the business logic behind this component more approachable, as this is an area I believe some more work could be done, but I think I'm not too close to the logic to see it so well!

dotcom-rendering/index.d.ts
```diff
@@ -696,7 +696,7 @@ type LineEffectType = 'labs' | 'dotted' | 'straight';
 
 type LeftColSize = 'compact' | 'wide';
 
-type CardPercentageType = '25%' | '34%' | '50%' | '66%' | '75%' | '100%';
+type CardPercentageType = '25%' | '33.333%' | '50%' | '66%' | '75%' | '100%';
```
OllysCoding, line 699: This change was necessary for the layout to look correct, from examining other areas where this % was used, I don't think we'll see a big impact on the site from it

mxdvl (commented):
This is an enormous PR, which makes hard to review, but congratulations on the effort.

Would it be possible to run the grouping on a subsequent step in the enhancer? Otherwise we are enhancing all the cards twice.

dotcom-rendering/src/model/groupCards.ts
```diff
@@ -0,0 +1,90 @@
...
+ *  - Standard
+ *
+ * This can be consumed by the rendering layer to tell us how we should size each card, which
+ * can vary by container type.
+ *
+ * Backfilled cards are always considered 'standard'.
+ *
+ * @param curated
+ * @param backfill
+ * @param containerPalette
```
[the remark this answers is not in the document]
oliverlloyd: Agreed. Typescript already tells us what props there are

dotcom-rendering/src/model/groupCards.ts
```diff
@@ -0,0 +1,90 @@
...
+						.filter((card) => card.card.group === '0')
+						.concat(backfill),
+					containerPalette,
+				),
+			};
+		case 'dynamic/package':
+			return {
+				huge: [],
+				veryBig: [],
+				big: [],
```
oliverlloyd, line 66: Dynamic package doesn't support huge/very big/etc. So do we need this?

dotcom-rendering/src/model/groupCards.ts
```diff
@@ -0,0 +1,90 @@
...
+				// Only 'snap' and 'standard' are supported by dynamic/package
```
oliverlloyd, line 67: 🤔 oh, wait, so dynamic package _does_ supporting some form of grouping? Interesting

dotcom-rendering/src/web/components/Carousel.tsx
```diff
@@ -353,7 +353,6 @@ export const CarouselCard: React.FC<CarouselCardProps> = ({
 	branding,
 }: CarouselCardProps) => (
 	<LI
-		stretch={true}
```
oliverlloyd, line 356: So having this didn't do anything?

It might be worth double checking this on Firefox and friends - I vaguely recall there being some issues around this

dotcom-rendering/src/web/components/DynamicFast.stories.tsx
```diff
@@ -22,17 +22,246 @@ export default {
...
+		centralBorder="partial"
+	>
+		<DynamicFast
+			groupedTrails={{
+				snap: [],
+				huge: [trails[0]],
+				veryBig: [trails[1]],
+				big: [trails[2], trails[3]],
+				standard: [trails[4], trails[5], trails[6]],
+			}}
```
oliverlloyd, line 71: There is something to say for this api. It makes it clearer at this point what we're asking the component to do

dotcom-rendering/src/web/components/DynamicFast.stories.tsx
```diff
@@ -22,17 +22,246 @@ export default {
...
+				snap: [],
+				huge: [],
+				veryBig: [
+					trails[0],
+					{ isBoosted: true, ...trails[1] },
+					trails[2],
+				],
+				big: [trails[3], trails[4], trails[5]],
+				standard: [],
+			}}
```
oliverlloyd, line 168: These stories are very readable, nice work 👍

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -1,58 +1,427 @@
...
+		}
+	`;
+
+	/**
+	 * The row & columns in this component are a little outside of the standard of what the standard
+	 * LI and UL components we have support - it made more sense to manually create the CSS for this use case
+	 * than to expand & muddy the APIs of both components to support this layout type.
+	 */
+	return (
+		<li
```
oliverlloyd, line 150: I suppose the risk here is any changes to LI or UL will not be picked up by this code. This makes me a little uncomfortable in that I feel that we're breaching a contract here. At the moment, every single Card list is put through a UL/LI combo but now we're adding an exception.

I can appreciate though that these abstractions are very heavy and changing them isn't easy

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -1,58 +1,427 @@
...
+								// If there are cards in the second col, we always want padding
+								standardsCol2.length > 0 ||
+								cardIndex < standardsCol1.length - 1
+							}
+						>
+							<FrontCard
+								trail={card}
+								containerPalette={containerPalette}
+								showAge={showAge}
+								imageUrl={undefined}
```
oliverlloyd, line 193:
```suggestion
```

I think removing this is the same?

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -1,58 +1,427 @@
...
+							padBottom={cardIndex < standardsCol2.length - 1}
+							padBottomOnMobile={
+								cardIndex < standardsCol2.length - 1
+							}
+						>
+							<FrontCard
+								trail={card}
+								containerPalette={containerPalette}
+								showAge={showAge}
+								imageUrl={undefined}
```
oliverlloyd, line 216: No need

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -1,58 +1,427 @@
...
+		];
+	} else if (standardsConfig.columns > 0) {
+		// We don't need to select any 'left over' bigs here as if there are
+		// 4 bigs there is no room for standards.
+		standards = groupedTrails.standard.splice(
+			0,
+			// We allow 3 standards per columns
+			standardsConfig.columns * 3,
+		);
+	}
```
oliverlloyd, line 379: If I'm honest, this is where I get stuck. From discussing this it is clear that it simply is the case that there is a lot of complex business logic here but I wonder if there is a way to refactor this to be easier to reason about? Of course, asking questions about wondering is very easy!

Are there perhaps too many comments here? In a way I think I am maxed out by trying to read those.

It also doesn't help that it's 17:41 on a Monday and my brain is mush!

oliverlloyd (commented):
This is intimidating but there's a lot of great work here. I'd be up for trying to see if there's a way to iron out some of the more narly parts if you're happy to spend some time on it

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -1,58 +1,427 @@
...
+}: {
+	standards: TrailType[];
+	big: TrailType;
+	showAge?: boolean;
+	containerPalette?: DCRContainerPalette;
+}) => {
+	// 'col1' has 1 big & up to 2 standards
+	const standardsCol1 = standards.splice(0, 2);
+	// 'col2' has up to 5 standards
+	const standardsCol2 = standards.splice(0, 5);
```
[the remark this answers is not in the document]
mxdvl: it’s `slice`!

mxdvl (approved):
I’ve tried a handful of ways to see if this could be made any simpler, and I could not find any.

I think this is as simple as it can get. 💎

dotcom-rendering/src/web/components/Carousel.tsx
OllysCoding: Prior to this PR, the conditional in the `<LI >` styles meant if you specified a `percentage` (which this element does) then stretch styles would not be applied.

I updated the LI element in this PR so that `percentage` and `stretch` properties can be applied at once (which is needed to make the dynamic/fast layout work) - doing this broke the carousel so I removed the prop!

dotcom-rendering/index.d.ts
OllysCoding: I haven't tested this - are there browsers which don't support flex-basis being set to decimals? From what I remember I saw frontend use decimals so I feel it should be okay?

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -1,58 +1,427 @@
...
+ * When there are 2 very bigs in the primaries row, they each take 50% of the horizontal space.
+ * Either of the 2 very bigs can 'boosted', in which case the boosted card will receive
+ * 75% of the horizontal space.
+ */
+const Primaries = ({
+	primaries,
+	showAge,
+	containerPalette,
+}: {
+	primaries: TrailType[];
```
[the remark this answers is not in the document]
OllysCoding: Going to give this a shot as it would go a long way to making the types clearer!

dotcom-rendering/src/model/groupCards.ts
OllysCoding: It does! I've not seen it used in a while (if at all?)

Here's what it looks like:
image
(with this set in fronts tool)
image

Here's how it renders in DCR right now:
image

This should be a simple fix for `dynamic/package` - if we update it to consume `groupedTrails` - then keep the existing logic pulling from `groupedTrails.standard` and simply add the top slice with `groupedTrails.snap`. (I've not tested what happens when you add a second snap yet - feel free to mess around with my front to test: https://m.code.dev-theguardian.com/olly)

dotcom-rendering/src/web/components/DynamicFast.stories.tsx
OllysCoding: Yes! This is what eventually led me to decide to leave the `groupTrails` functionality in the enhancer rather than in the component itself - it adds a bit more complexity to our collection types but this interface is so readable from a testing perspective

dotcom-rendering/src/web/components/DynamicFast.tsx
OllysCoding: [imageUrl is set in FrontCard](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/FrontCard.tsx#L39) so if you want no image to be shown you need to explicitly set it to be undefined

dotcom-rendering/src/web/components/DynamicFast.tsx
OllysCoding: Yes I can't help but agree - the comments are there to try and explain the business logic but I have a tendancy to overcomment so I will review their usefulness and see if we can get away with less/more concise.

I'll keep staring at this peace of logic and see how it can be tidied up more (if at all) to be more readable.

It is split into essentially 3 sections,
1. 'primaries' (also known as first slice, would this be a better name?)
2. bigs
3. standards

They need to be created in that order as they each have logic which requires the earlier to be known - but perhaps the flow could be more clear!

dotcom-rendering/src/model/groupCards.ts
oliverlloyd: Thanks Olly, makes sense.

I'm guessing that this view is probably not used but if we can get DCR to support everything then in many ways this makes life easier

dotcom-rendering/src/web/components/DynamicFast.tsx
oliverlloyd: Oh, that's new. That must be part of the `FrontCard` work. Fair enough, this api does make sense, I just didn't know about it, thanks for showing it to me!

dotcom-rendering/src/web/components/DynamicFast.tsx
oliverlloyd: Thanks. But as discussed offline lets focus on moving this one forward, there's a risk you could lose a lot of time refactoring here

dotcom-rendering/src/web/components/Carousel.tsx
oliverlloyd: Thanks for explaining 👍

oliverlloyd (approved):
Great work on this

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -1,58 +1,434 @@
...
 	containerPalette?: DCRContainerPalette;
 	showAge?: boolean;
 };
 
-export const DynamicFast = ({ trails, containerPalette, showAge }: Props) => {
-	const primary = trails[0];
-	const secondary = trails[1];
-	const bigCards = trails.slice(2, 4);
-	const smallCards = trails.slice(4, 10);
+/**
```
bryophyta, line 14: great documentation to have here!

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -1,58 +1,434 @@
...
-						imagePositionOnMobile="top"
-						imageSize="large"
-						supportingContent={primary.supportingContent}
-						trailText={primary.trailText}
-					/>
-				</LI>
+		<UL direction="row" padBottom={true}>
+			{primaries.map((card, index) => {
+				// Boosted Layout
+				if (primaries.length > 1 && boostedIndex !== -1) {
```
bryophyta, line 41: fwiw, this is the section that tripped me up when reading it. But, like other reviewers, I agree that the complexity is pretty inherent to the setup here, and I can't think of a better alternative!

It feels to me like there are two main sources of complexity here when trying to understand this section. One is the number of different layouts that can be generated. The other is that the `isBoosted` property that we get through from tools/Frontend doesn't actually mean that the card will be rendered as boosted (hence having to check against `boostedIndex` here, rather than being able to check against `isBoosted`). Does that sound right?

I think the point is implicit in your comment on ln.34, but it might be helpful to make it a bit more explicit? i.e. something like "A maximum of one card can be boosted in this slice, but this requirement is not enforced in the data we receive, so we are using `boostedIndex` here to determine whether a card should actually be boosted, rather than relying on the `isBoosted` property directly."?

bryophyta:
I just wanted to echo some of the other comments above! Great work on grappling with such an awkward piece of business logic, and thank you for taking the time to write such a clear and thorough PR for it, too! I wonder if we should link to this PR in our docs somewhere, or add an inline comment in the `DynamicFast.tsx` which points people towards it? I feel like it would be a really helpful resource for people who are trying to understand how these layouts work

bryophyta (approved):

dotcom-rendering/src/web/components/DynamicFast.tsx
OllysCoding: This was a good suggestion & from reading I think you understand it all well - I will update this comment to be more explicit now!

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -1,58 +1,475 @@
...
-	const smallCards = trails.slice(4, 10);
+/**
+ * 'Primaries' renders the first row of dynamic/fast
+ * This row will consist of either 1 huge, 1 very big, or 2 very bigs.
+ *
+ * When there is 1 huge or 1 very big, these will be treated the same, and take
+ * up 100% of the horizontal space
+ *
+ * When there are 2 very bigs in the primaries row, they each take 50% of the horizontal space.
+ * Either of the 2 very bigs can 'boosted', in which case the boosted card will receive
```
oliverlloyd, line 30:
```suggestion
 * Either of the 2 very bigs can be 'boosted', in which case the boosted card will receive
```

oliverlloyd (approved):
This is a lot clearer, well done!. Thanks for spending the extra time on this

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -5,6 +5,14 @@ import { LI } from './Card/components/LI';
 import { UL } from './Card/components/UL';
 import { FrontCard } from './FrontCard';
 
+/**
+ * Not sure where to start? This PR documents a lot of the key features
```
bryophyta, line 9: 🤩

PR #5430, opened 2022-07-18
Replace hardcoded colour with Source colour in sports match headers

bryophyta:
## What does this change?

Replaces a hard-coded hex value with a reference to the corresponding value in Source's palette.

## Why?

I believe that this is the recommended practice in DCR. (Helps to improve consistency and maintainability.)

## Screenshots

There is no visible difference in the output (verified locally).

### DIFF
--- a/dotcom-rendering/src/web/lib/decidePalette.ts
+++ b/dotcom-rendering/src/web/lib/decidePalette.ts
@@ -998,7 +998,7 @@ const borderCardSupporting = (format: ArticleFormat): string => {
 };
 
 const backgroundMatchNav = (): string => {
-	return '#FFE500';
+	return brandAlt[400];
 };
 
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/lib/decidePalette.ts
```diff
@@ -998,7 +998,7 @@ const borderCardSupporting = (format: ArticleFormat): string => {
 };
 
 const backgroundMatchNav = (): string => {
-	return '#FFE500';
+	return brandAlt[400];
```
bryophyta, line 1001: https://github.com/guardian/source/blob/1c4d34c5f61be7226b274d231277ffb927213038/packages/%40guardian/source-foundations/src/colour/palette.ts#L63

AshCorr (approved):

PR #5431, opened 2022-07-18
feat: amend layout for Newsletter Signup

cemms1:
## What does this change?

Changes the layout for newsletter signup pages but does not include it in the `DecideLayout` component yet since it's a WIP. Follow up PRs will add to this layout until it is ready to use.

## Why?

The Newsletters team are aiming to increase the volume of signups for individual newsletters so the pages have been redesigned with a new look and feel to make it easier to do so.
The aim is for these signup pages to be created using Composer, rather than a bespoke interactive layout each time (trialled for First Edition).

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/43961396/179521530-df38206f-4402-4bca-947a-d1a502937c9e.png
[after]: https://user-images.githubusercontent.com/43961396/179521708-a2c86c70-30f8-4403-9872-e6b89b516141.png
[before2]: https://user-images.githubusercontent.com/43961396/179521838-828d25cf-61f9-4768-a1a1-3dd29166be66.png
[after2]: https://user-images.githubusercontent.com/43961396/179521856-62dea1ad-a9cb-4100-88b8-7fe4ae27e0c8.png

### DIFF
--- a/dotcom-rendering/src/web/lib/decidePalette.ts
+++ b/dotcom-rendering/src/web/lib/decidePalette.ts
@@ -813,7 +813,11 @@ const fillShareIcon = (format: ArticleFormat): string => {
 				return specialReport[300];
 		}
 	}
-	if (format.design === ArticleDesign.NewsletterSignup) return BLACK;
+	if (
+		format.design === ArticleDesign.NewsletterSignup &&
+		format.display === ArticleDisplay.Standard
+	)
+		return BLACK;
 
 	if (format.theme === ArticleSpecial.Labs) return BLACK;
 	if (format.theme === ArticleSpecial.SpecialReport)


mxdvl (dismissed):
Nice touch! 

Is it the intended design to have the Nav Bar splitting the two headers? I find it a little jarring visually, but it could grow on me.

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -242,38 +336,150 @@ export const NewsletterSignupLayout = ({ CAPIArticle, NAV, format }: Props) => {
...
+
+							{/* TODO:
+								- This data will come from the Newsletters API
+								- Only render this part if preview link exists
+								- Add onClick handler or link?
+								- Accessibility?
+							*/}
+							{showNewsletterPreview && (
+								<div css={previewButtonWrapperStyle}>
+									<Button
```
oliverlloyd, line 376: If this is a plain link, we can use a `LinkButton` here. That way we won't need javascript 🎉

dotcom-rendering/src/web/components/ShareIcons.tsx
```diff
@@ -67,6 +67,14 @@ const decideIconColor = (format: ArticleFormat, context: Context) => {
...
+	if (
+		format.design === ArticleDesign.NewsletterSignup &&
+		context === 'ArticleMeta'
+	) {
+		return css`
+			fill: ${neutral[7]};
+		`;
+	}
 	return css`
 		fill: ${palette.fill.shareIcon};
```
mxdvl, line 79: Colour variations by format, we have a concept of `palette`. Have a look at `decidePalette`, which is where we centralise all colour decisions that depend on an article format.

mxdvl (commented):
Colour variations on an article’s format should live in `decidePalette`, but otherwise you’re good to go :)

dotcom-rendering/src/web/components/ShareIcons.tsx
```diff
@@ -67,6 +67,14 @@ const decideIconColor = (format: ArticleFormat, context: Context) => {
 			fill: ${palette.fill.shareIconGrayBackground};
 		`;
 	}
+	if (
+		format.design === ArticleDesign.NewsletterSignup &&
+		context === 'ArticleMeta'
```
oliverlloyd, line 72: If design is equal to NewsletterSignup then I think this context value is always going to be `ArticleMeta`?

If so, I think we can move this decision logic into `decidePalette`. 

### `decidePalette`? What is this?
This is where we make decisions about what colour to use for things in DCR. We built this (big, long) abstraction as a way to centralise where decisions are made about colour based on `format`. It is designed to try and make this kind of logic easier to reason about and more scalable. We often get requests and features to edit this kind of code so it's nice to have a structure to it.

dotcom-rendering/src/web/components/ShareIcons.tsx
```diff
@@ -87,7 +95,18 @@ const decideIconColorOnHover = (format: ArticleFormat, context: Context) => {
...
+	) {
+		return css`
+			:hover {
+				background-color: ${neutral[7]};
+				border-color: ${neutral[7]};
+				fill: white;
+			}
+		`;
+	}
```
oliverlloyd, line 109: Ditto, `decidePalette`

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -101,6 +115,77 @@ const leftColWrapperStyle = css`
...
+		display: none;
+	}
+`;
+
+const mainGraphicWrapperStyle = css`
+	border-radius: ${space[2]}px;
+	overflow: hidden;
+	margin: ${space[4]}px 0;
+
+	& > figcaption {
```
oliverlloyd, line 130: We don't typically use these types of selectors in DCR. Instead, we favour using components and having local styles applied there. One of the goal of DCR is componentised, isolated code. This is also true for styles but by having css selectors that also apply styles to descendants you create a coupling that can cause problems when scaling things.

I think you can simply create a new  css property - captionStyles - and add this there and then target it directly.

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -101,6 +115,77 @@ const leftColWrapperStyle = css`
...
+		display: flex;
+
+		${until.desktop} {
+			display: none;
+		}
+
+		background-color: ${brandAlt[400]};
+		align-items: center;
+		padding: ${space[1]}px;
+		${textSans.medium({ fontWeight: 'bold', lineHeight: 'tight' })}
```
oliverlloyd, line 140: Perfection 👍

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -101,6 +115,77 @@ const leftColWrapperStyle = css`
...
+
+	return (
+		<div css={containerStyle}>
+			<SvgEnvelope size="small" />
+			<span>{text}</span>
+		</div>
+	);
+};
+
 interface Props {
```
oliverlloyd, line 189: Nitpick: There's a convention to keep the props for a component at the top of the file. It can help when opening new files and trying to understand them

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -101,6 +115,77 @@ const leftColWrapperStyle = css`
...
+			flex-shrink: 0;
+		}
+	}
+`;
+
+interface NewsletterDetailProps {
+	text: string;
+	showUntil?: Breakpoint;
+}
+const NewsletterDetail = ({ text, showUntil }: NewsletterDetailProps) => {
```
oliverlloyd, line 153: It's classically hard to know when this is a good idea or not - and I'm not really on top of this code enough to make the call myself - but you could consider if this component need breaking out into it's own file?

Pros: You can create a story for it, it's easier to expand later
Cons: It separates this logic away from where it's being used and hides it.

My instinct would be to break it out. It's nice if the layout file has a little css in it as possible and is only positioning components on the page but I'm happy to leave the call with you.

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -101,6 +115,77 @@ const leftColWrapperStyle = css`
...
+		${textSans.medium({ fontWeight: 'bold', lineHeight: 'tight' })}
+
+		svg {
+			margin-right: ${space[1]}px;
+			flex-shrink: 0;
+		}
+	}
+`;
+
+interface NewsletterDetailProps {
```
oliverlloyd, line 149: For no particular reason, there's a convention to use `type` over `interface` when defining props 🤷 

```suggestion
type NewsletterDetailProps = {
```

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -101,6 +115,77 @@ const leftColWrapperStyle = css`
...
+		padding-bottom: ${space[2]}px;
+
+		svg {
+			background-color: ${brandAlt[400]};
+			border-radius: 50%;
+			padding: 2px;
+			margin-right: ${space[2]}px;
+		}
+
+		span {
```
oliverlloyd, line 176: As above can we target this css please

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -101,6 +115,77 @@ const leftColWrapperStyle = css`
...
+		svg {
+			margin-right: ${space[1]}px;
+			flex-shrink: 0;
+		}
+	}
+`;
+
+interface NewsletterDetailProps {
+	text: string;
+	showUntil?: Breakpoint;
```
oliverlloyd, line 151: In other places in DCR we do similar things as here using the `Hide` component, which has similar props for breakpoints. Given we only use `showUntil` with `leftCol` maybe that works?

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -125,6 +210,13 @@ export const NewsletterSignupLayout = ({ CAPIArticle, NAV, format }: Props) => {
 
 	const palette = decidePalette(format);
 
+	/** Logic to show preview button or image wrapper
+		TODO:
+		- Include logic here for whether preview exists for the newsletter
+		(has preview URL available?)
+	*/
+	const showNewsletterPreview = true;
```
oliverlloyd, line 218: 👍

oliverlloyd (approved):
This is an awesome first DCR PR!

I've made some comments but nothing major and I'm happy for you to manage things and make the changes as you see best

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
oliverlloyd: Before. Most components have the first line after the imports as `type Props = ...`

PR #5433, opened 2022-07-18
`/Onwards`

oliverlloyd:
## What does this change?
This PR introduces the `/Onwards` endpoint. When passed `CAPIOnwardsType` this endpoint will return the html needed to render the trails given in the request.

```ts
type CAPIOnwardsType = {
	heading: string;
	trails: CAPITrailType[];
	description?: string;
	url?: string;
	ophanComponentName: OphanComponentName;
	format: CAPIFormat;
	isCuratedContent?: boolean;
};
```

The request contains properties of the CAPI type. These are transformed to their DCR version before being rendered

DCR Onwards type:
```ts
type OnwardsType = {
	heading: string;
	trails: TrailType[];
	description?: string;
	url?: string;
	ophanComponentName: OphanComponentName;
	format: ArticleFormat;
	isCuratedContent?: boolean;
};
```

This DCR data is then given to the `Carousel` component and the resulting html returned.

## Why?
This PR is part of a wider effort to move the rendering logic for onwards content onto the server. We want to ensure that the `Card` component is not being rendered in the browser (because it is using `IMAGE_SALT` and `crypto` neither of which are available on the client.

Subsequent PRs to follow:

1) Add a query param (say, `html=true`) to the onwards api endpoints such that when given, the Frontend code will pass the trail json data to DCR to be rendered to html (using this PR's code). If not given or `html=false`, then the response will be the same json we get now

2) Once the `html=true` param is available, we can have a version of DCR that expects html in the response and inserts that into the DOM rather than asking React to generate the html for us.

### DIFF
--- a/dotcom-rendering/src/web/components/OnwardsData.tsx
+++ b/dotcom-rendering/src/web/components/OnwardsData.tsx
@@ -3,13 +3,13 @@ import { useEffect } from 'react';
 import { decideTrail } from '../lib/decideTrail';
 import { revealStyles } from '../lib/revealStyles';
 import { useApi } from '../lib/useApi';
+import { Carousel } from './Carousel';
 import { Placeholder } from './Placeholder';
 
 type Props = {
 	url: string;
 	limit: number; // Limit the number of items shown (the api often returns more)
 	ophanComponentName: OphanComponentName;
-	Container: React.FC<OnwardsType>;
 	format: ArticleFormat;
 	isCuratedContent?: boolean;
 };
@@ -29,7 +29,6 @@ export const OnwardsData = ({
 	url,
 	limit,
 	ophanComponentName,
-	Container,
 	format,
 	isCuratedContent,
 }: Props) => {
@@ -75,7 +74,7 @@ export const OnwardsData = ({
 		return (
 			<div css={[minHeight, revealStyles]} className="onwards">
 				<div className="pending">
-					<Container
+					<Carousel
 						heading={data.heading || data.displayname} // Sometimes the api returns heading as 'displayName'
 						trails={buildTrails(data.trails, limit)}
 						description={data.description}
--- a/dotcom-rendering/src/web/server/index.ts
+++ b/dotcom-rendering/src/web/server/index.ts
@@ -12,6 +12,7 @@ import { articleToHtml } from './articleToHtml';
 import { blocksToHtml } from './blocksToHtml';
 import { frontToHtml } from './frontToHtml';
 import { keyEventsToHtml } from './keyEventsToHtml';
+import { onwardsToHtml } from './onwardsToHtml';
 
 function enhancePinnedPost(format: CAPIFormat, block?: Block) {
 	return block ? enhanceBlocks([block], format)[0] : block;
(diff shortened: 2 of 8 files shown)


dotcom-rendering/src/web/components/OnwardsData.tsx
```diff
@@ -75,7 +74,7 @@ export const OnwardsData = ({
 		return (
 			<div css={[minHeight, revealStyles]} className="onwards">
 				<div className="pending">
-					<Container
+					<Carousel
```
oliverlloyd, line 77: We used to support the `Container` being passed in as a prop but now we always render the same component (Carousel) there's no longer any need to do this

dotcom-rendering/src/web/server/onwardsToHtml.tsx
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * onwardsToHtml is used by the /Onwards endpoint to render the Carousel at the bottom of articles
+ * It takes an array of json key-event blocks and returns the resulting html string
+ *
+ * @returns string (the html)
+ */
+export const onwardsToHtml = ({
+	heading,
+	// description,
+	// url,
```
oliverlloyd, line 22: These two properties are not currently supported by `Carousel` but were used by the previous onwards components. I've left them here for now until we see what we decide to pass in from Frontend.

dotcom-rendering/src/web/server/onwardsToHtml.tsx
```diff
@@ -0,0 +1,41 @@
...
+	isCuratedContent,
+}: CAPIOnwardsType): string => {
+	const format = decideFormat(CAPIFormat);
+
+	const html = renderToString(
+		<Carousel
+			heading={heading}
+			ophanComponentName={ophanComponentName}
+			isCuratedContent={isCuratedContent}
+			trails={buildTrails(trails, 8)}
```
oliverlloyd, line 35: Do we need this limit of 8 cards? I guess we could pass more to carousel?

dotcom-rendering/src/web/server/index.ts
```diff
@@ -189,12 +190,43 @@ export const renderKeyEvents = (
...
+	try {
+		const {
+			heading,
+			description,
+			url,
+			ophanComponentName,
+			trails,
+			format,
+			isCuratedContent,
+		} = body;
```
OllysCoding, line 206: As a general note here, we use schemas on articles (incl. liveblogs) & fronts to ensure the data passed is correct, but don't do the same for `/blocks`, `/keyevents`, etc - I wonder if we are happy with this?

(If we want to its pretty easy to add new schemas to [get-schema.js](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/scripts/json-schema/get-schema.js), then expand [validate.ts](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/model/validate.ts))

OllysCoding (approved):
+1 Looks good :)

dotcom-rendering/src/web/server/index.ts
oliverlloyd: I've made a note to look at this. I'm wondering what we want the outcome to be if the schema validation fails. Should that fails the request? I guess it should but it's a bit annoying as ideally, you'd want to know that is going to happen ahead of time but I suppose the separation between repos won't allow this.

PR #5437, opened 2022-07-19
Dependabot should ignore aws-cdk & related dependencies for AR

### LINKED ISSUE #5436: Disable dependabot for aws-cdk
`aws-cdk` is a peer dependency of `@guardian/cdk`. We should disable dependabot for `aws-cdk`.

Instead, we should bump versions of `@guardian/cdk` which will also result in upgrades to `aws-cdk`.

georgeblahblah:
## What does this change?

- Ignores `aws-cdk`, `@aws-cdk/core`, `@aws-cdk/assert` and `@aws-cdk/cloudformation-include` from dependabot's configuration for Apps Rendering

## Why?

These dependencies are peer deps of `@guardian/cdk` and so the version here should match exactly the version specified by @guardian/cdk. It is not helpful to have Dependabot open PRs to upgrade these dependencies alone

Fixes #5436

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,6 +18,13 @@ updates:
       labels:
         - "AR Dependency"
       rebase-strategy: "disabled"
+      ignore:
+        # aws-cdk is a peer dependency of @guardian/cdk
+        # The version of the aws-cdk dependencies should match exactly the version specified by @guardian/cdk
+        - dependency-name: 'aws-cdk'
+        - dependency-name: '@aws-cdk/core'
+        - dependency-name: '@aws-cdk/assert'
+        - dependency-name: '@aws-cdk/cloudformation-include'
     - package-ecosystem: "npm"
       directory: "/dotcom-rendering"
       schedule:


JamieB-gu:
@jacobwinch @akash1810 what do you think of this idea? I think the aim is to stop dependabot upgrading these one at a time, which would make them go out of sync with the GuCDK version. Would this be useful for other projects that use CDK?

akash1810 (approved):
> @jacobwinch @akash1810 what do you think of this idea? I think the aim is to stop dependabot upgrading these one at a time, which would make them go out of sync with the GuCDK version. Would this be useful for other projects that use CDK?

👍🏽 we do this within DevX too ([example](https://github.com/guardian/cdk-playground/blob/main/.github/dependabot.yml)).

Worth noting that if you move to v41.0.0 of GuCDK you'll start to use the mono-packaged AWS-CDK, meaning your list of exclusions collapse to:

```
- dependency-name: "aws-cdk"
- dependency-name: "aws-cdk-lib"
```

PR #5441, opened 2022-07-19
Dblatcher/sign up styling

dblatcher:
## What does this change?
 - adjust internal padding and margins of `EmailSignup` to fit design
 - change to neutral colour for the links in the privacy text
 - create a separate component for the Privacy text 

## Why?
 - aligns to design input
 - reduces duplication between `SecureSignup` and `SecureSignupMock`

## Screenshots

| Before      | After      |
|-------------|------------|
| oldMobile | newMobile |
| desktopOld | desktopNew |

## Design brief
Screenshot 2022-07-15 at 17 11 59


dotcom-rendering/src/web/components/NewsletterPrivacyMessage.tsx
```diff
@@ -0,0 +1,55 @@
...
+		}
+	}
+	strong {
+		color: ${neutral[0]};
+		font-weight: bold;
+	}
+`;
+
+export const NewsletterPrivacyMessage = () => (
+	<>
```
oliverlloyd, line 24:
```suggestion
```

I don't think this fragment is needed

dotcom-rendering/src/web/components/EmailSignup.tsx
```diff
@@ -60,7 +69,9 @@ export const EmailSignup = ({ newsletter }: Props) => {
 				<p css={titleStyles(theme)}>
 					Sign up to <span>{name}</span> today
 				</p>
-				<NewsletterFrequency frequency={frequency} />
+				<div>
```
oliverlloyd, line 72: Is this div intentional?

dotcom-rendering/src/web/components/EmailSignup.tsx
```diff
@@ -18,15 +18,25 @@ type Props = {
...
 	border-radius: 12px;
-	padding: 10px;
 	margin-bottom: ${space[3]}px;
+	padding: ${space[2]}px;
+
+	${from.tablet} {
+		padding: ${space[2]}px ${space[3]}px;
+		& > div {
+			margin-bottom: -4px;
+		}
```
oliverlloyd, line 28: We try to avoid these types of selectors in DCR, they create couplings between components that can make it hard to reason about later. Can we target the css using local css instead.

That said, negative margins are a bit of a flag - is there a better way to achieve the spacing we want?

dotcom-rendering/src/web/components/NewsletterPrivacyMessage.tsx
dblatcher: You're right, i forgot to remove it when I combined the two spans into one

dotcom-rendering/src/web/components/EmailSignup.tsx
dblatcher: There was some odd behaviour  in the `NewsletterFrequency` component that was making it taller than it should be, but I've made a change to that so we don't need the div or the negative margin.

PR #5443, opened 2022-07-19
Standfirsts & Sublinks

oliverlloyd:
## Standfirsts & Sublinks
Also known as `trailText` & `supportingContent` but the alliteration was too hard to ignore.

This PR adds back support for `trailText` and `supportingContent` on DCR fronts

## Why?
These properties are only used on certain cards in certain containers and need to be explicitly specified.

| Before      | After      |
|-------------|------------|
| Screenshot 2022-07-19 at 17 13 16 | Screenshot 2022-07-19 at 17 12 38 |


mxdvl (approved):
seems fine!

PR #5447, opened 2022-07-19
Older richllinks don't always have `imageAsset`

oliverlloyd:
## What does this change?
This removes the absolute requirement for a rich link to have the `imageAsset` property

## Why?
Some [older articles](https://www.theguardian.com/media/2007/mar/28/pressandpublishing.royalsandthemedia) don't provide this when inserted as richlinks and this was causing javascript errors when the `imageAsset.fields` property was accessed.


mxdvl (approved):
Thanks for this!

Issue #4241, opened 2022-03-14
CSS Variables For Dark Mode


JamieB-gu:
A rough proposal for using CSS variables to power dark mode. There's _lots_ of prior art on this topic here: https://github.com/guardian/source/discussions/1060 - many/most of the ideas in here come from @oliverlloyd @SiAdcock @mxdvl and @iainjchambers-guardian.

### Editorial Palette

The editorial palette already provides a framework for selecting semantic light and dark mode colours for a given piece of UI. Here's a simplified example of what the `text` API looks like:

```ts
const text = {
  headline,
  headlineDark,
  standfirst,
  standfirstDark,
  ...
}
```

### CSS Variables

What if we were to leverage this to provide a group of CSS variables for each colour mode?

```ts
const lightPalette = css`
  --text-headline: ${text.headline(format)};
  --text-standfirst: ${text.standfirst(format)};
  --background-headline: ${background.headline(format)};
`

const darkPalette = css`
  --text-headline: ${text.headlineDark(format)};
  --text-standfirst: ${text.standfirstDark(format)};
  --background-headline: ${background.headlineDark(format)};
`
```

**Note:** This is for demo purposes, we may be able to reduce the boilerplate.

### Components

Then in our components:

```tsx
const styles = css`
  color: var(--text-headline);
  background-color: var(--background-headline);
`

const Headline = ({ copy }) =>
  <h1 css={styles}>{copy}</h1>
```

**Note:** Again this is for demo purposes, we may want to import these variable strings, instead of entering them manually in each place, for type safety. Or maybe that's overkill and this is fine.

### Switching Modes

Now we simply activate either `lightPalette` or `darkPalette` at the top level of our app to have everything switch over automatically. We could do this via a media query (most likely option in apps):

```ts
const topLevelStyles = css`
  ${lightPalette}

  @media (prefers-color-scheme: dark) {
    ${darkPalette}
  }
`
```

Or via other mechanisms, like JS or a `<link>` tag (more likely in dotcom):

```html
<link
  rel="stylesheet"
  href="light.css"
  media="(prefers-color-scheme: light)"
/>

<link
  rel="stylesheet"
  href="dark.css"
  media="(prefers-color-scheme: dark)"
/>
```

### Storybook

[...]

PR #5454, opened 2022-07-20
RFC: CSS Variables

oliverlloyd:
## What does this change?
This PR adds the usage of [css variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) to DCR. What do you think?

## Why?
CSS variables have long been wanted on DCR but we did not adopt them because they are not supported on IE11. Now that we no longer offer official support for this browser we have the opportunity to adopt them.

This particular PR is providing the building blocks for future work. In particular, it offers us the ability to have page level colours - such a pillar colour - which removes the need for local props on components. This simplifies the rendering of some content and makes it easier to do certain refactors, such as moving content from the client to the server

## Future enhancements
This PR is very basic but is built with future enhancements in mind

### Script generation
Rather than create this file manually we should look to build a script that generates it based on imports from `Source`. This script can then be rerun to add new properties or reflect changes to the `source` libs.

### Dark mode
How we implement dark mode on the site is [discussed in this document](https://github.com/guardian/source/discussions/1060). One of the key elements of that solution is the use of css variables.

This PR does not go all the way down the dark path but if we wanted to do that later then we'd need to add a second css file, `dark.css`. We would then need to consider how we mark the reader's preference in the document, such as via a class on the `body` tag. @mxdvl made a [very good suggestion](https://github.com/guardian/source/discussions/1060#discussioncomment-1913032) for how this setting might be made, which is likely to be the best method.

The `link` tag could have a media property

```html
<link
	rel="stylesheet"
	href="${CDN}static/frontend/css/dark.css"
	media="(prefers-color-scheme: dark)"
/>
```
We should consider if we want to allow readers the option to switch between modes on the site (a toggle somewhere) or if we are okay using the system setting (e.g.: system prefs).

Consider if we would need a listener against changes to the system preferences which updated the setting (the `.light` class) on the site

### DIFF
--- a/dotcom-rendering/src/static/css/light.css
+++ b/dotcom-rendering/src/static/css/light.css
@@ -1,0 +1,4 @@
+:root {
+	--white: #ffffff;
+	--black: #121212;
+}
--- a/dotcom-rendering/src/web/lib/decidePalette.ts
+++ b/dotcom-rendering/src/web/lib/decidePalette.ts
@@ -27,8 +27,8 @@ import type { DCRContainerPalette } from '../../types/front';
 import { decideContainerOverrides } from './decideContainerOverrides';
 import { transparentColour } from './transparentColour';
 
-const WHITE = neutral[100];
-const BLACK = neutral[7];
+const WHITE = 'var(--white)';
+const BLACK = 'var(--black)';
 
 const blogsGrayBackgroundPalette = (format: ArticleFormat): string => {
 	switch (format.theme) {
--- a/dotcom-rendering/src/web/server/articleTemplate.ts
+++ b/dotcom-rendering/src/web/server/articleTemplate.ts
@@ -286,6 +286,10 @@ https://workforus.theguardian.com/careers/product-engineering/
                 <style class="webfont">${getFontsCss()}</style>
                 <style>${resets.resetCSS}</style>
 				${css}
+				<link
+					rel="stylesheet"
+					href="${ASSET_ORIGIN}static/frontend/css/light.css"
+				/>
 				<link rel="stylesheet" media="print" href="${ASSET_ORIGIN}static/frontend/css/print.css">
 			</head>
 


dotcom-rendering/src/static/css/light.css
```diff
@@ -0,0 +1,4 @@
+:root {
+	--white: #ffffff;
+	--black: #121212;
```
mxdvl, line 3: Could this be generated programatically from source?

mxdvl (dismissed):
Interesting! See also the proposed Spike in #4241

dotcom-rendering/src/static/css/light.css
mxdvl: I meant could we add a comment, sorry, you’ve obviously mentioned this in the PR description

sndrs:
i;'m not clear what the benefit is here, given the CSS is currently all generated in JS which already gives you variables?

if you have different values available in two different domains (JS and CSS), how do you know the value of the thing your looking at in a piece of code?

for some historical context, DCR began life as an attempt to overcome the CSS bloat of frontend by binding the html markup to the styling explicitly in code, then using css-in-js to only serve what we needed inline per page. this meant authors could safely work on components without needing to hold the full context of the global CSS env in their heads as they worked on a specific problem.

so i'm wary of pushing config-type data _back_ into CSS land (or rather, splitting it across domains). 

what is the driver for this? is it mainly light/dark mode stylesheets or are their other benefits the existing set up prevents you taking advantage of?

bryophyta:
> In particular, it offers us the ability to have page level colours - such a pillar colour - which removes the need for local props on components.

I think this could be a real win if we can pull it off, but my sense is that it will take quite a lot of careful thought if we are planning to use css variables in this way, because of the different inheritance models for css vs. components. I think we'd need to design the css API with care to avoid re-introducing fine-grained props lower in the component tree again at a later date to account for edge cases, which would then be messier than what we have now due to the css not being scoped. (Potentially leading to the kind of situation @sndrs mentions above.)

One suggestion: as a test case, would it make sense to try this out a relatively low branch of the component tree, like a card?  This might help to clarify some of the potential advantages and challenges for implementing the pattern more broadly?

OllysCoding:
> so i'm wary of pushing config-type data back into CSS land (or rather, splitting it across domains).

I'd put forward similar concerns as Alex here; it's not clear to me what benefit we're going to get from CSS variables? This PR seems to add another layer to `decidePalette.ts` but I'm not entirely clear what its supposed to do yet! 

To be clear, I'm not against CSS variables, but it'd be nice to see some solid use cases it will benefit us, as well as some clear lines in the sand to ensure it's used properly & ensure we don't end up with CSS mixed between vars & js (preferably in the form of eslint rules, etc).

oliverlloyd:
> i;'m not clear what the benefit is here, given the CSS is currently all generated in JS which already gives you variables?
> 
> what is the driver for this? is it mainly light/dark mode stylesheets or are their other benefits the existing set up prevents you taking advantage of?

Yeah, this PR isn't really putting a good case forward for the why here. sorry - there were some other reasons that triggered this work but really, the core motivation is dark mode. We support this on AR but not DCR. It is true that there are no immediate plans to have dark mode on the site but nonetheless we do want to be able to have shared approaches for how we render content such as interactives, embeds, etc. In order to work out the best approach for this type of content having an established pattern for dark mode is useful.


> if you have different values available in two different domains (JS and CSS), how do you know the value of the thing your looking at in a piece of code?

You don't. But I would argue `decidePalette` is already heavily abstracting the decision on colour so adding an additional level of abstraction (pointing to a css var instead of a source variable) would not be onerous.

oliverlloyd:
> This PR seems to add another layer to `decidePalette.ts` but I'm not entirely clear what its supposed to do yet!

The benefit to css vars lies in how we approach dark mode. If each colour is represented by a css variable then the task of switching and responding to a reader's preference an happen in one place, `<body class="light'>`. This centralised class can be set using a small js snippet and then you alternate between `light.css` and `dark.css` to set the variables.

AR currently uses a different approach which is more config/props based but this has limitations and there have been discussion around having more alignment

bryophyta:
To (hopefully) clarify my initial comment above, I think that using css vars at a root level, like for dark mode, avoids most of the issues that I described. It's replacing logic in components with logic controlling css vars higher up in the tree that seems particularly delicate to me, and would require some significant planning (imo).

sndrs:
> this PR isn't really putting a good case forward for the why

> the core motivation is dark mode

> there are no immediate plans to have dark mode on the site

this makes it quite hard to understand the motivation. i can see the cost, but i'm not sure i see the benefit

> we do want to be able to have shared approaches for how we render content

is this what's at root? are you fundamentally proposing a contract whereby partial styling information will be stored in global CSS scope for non-platform teams to hook into?

> `decidePalette` is already heavily abstracting the decision on colour so adding an additional level of abstraction (pointing to a css var instead of a source variable) would not be onerous

would you be able to expand on this? my immediate reaction is that it would exacerbate an existing source of difficulty

oliverlloyd:
> this makes it quite hard to understand the motivation. i can see the cost, but i'm not sure i see the benefit

Okay, I'm doing a terrible job of explaining this. I created this PR casually based on an assumption that there agreement on the approach of using css vars for dark mode and then thought this PR would be a rubber stamp. Let me fix that.

So the use case for using css vars is mostly laid out in [this discussion](https://github.com/guardian/source/discussions/1060). But there is also [this article](https://epicreact.dev/css-variables/) lobbying for using css vars vs. things like themes in Emotion. In many ways, the impact to DCR (and also AR) is smaller than you might normally experience because we already abstract the logic for deciding colours using `decidePalette`. The only place we would actually implement the `var(...)` is inside this one file.

From discussions with AR, it's my impression that there's agreement that this approach is the right one for both codebases. I don't want to speak for AR too much here though so @JamieB-gu can confirm more. Either way, I think this alignment is important. We share code between the two platforms and there are more components being built (and even libraries being created) that are intended to be used cross platform but right now this shared code does not handle dark mode on apps.

Screenshot 2022-07-28 at 10 09 12

If we can say, yes, this approach is how we want to implement dark mode in both DCR and AR then that allows us to start helping teams that are creating content that is meant for both platforms to add this dark mode support to their code.

On the other hand, if we don't reach consensus then it highlights that more work is needed in this area and it is far preferable to do that work as early as possible to reduce the amount of refactoring needed later to provide a consistent experience for readers.

oliverlloyd:
> are you fundamentally proposing a contract whereby partial styling information will be stored in global CSS scope for non-platform teams to hook into?

I think this is essentially true, yes.

Ultimately, dark mode is global state. This is unavoidable. You need the various parts of the page to be aware of this state and to respond to changes in it. One way to do this is using props but that is limited to hydrated react code of which there isn't much. We could also set a global flag somewhere and then give this contract to each component and ask them to implement the decision logic but this would create fragmentation.

Having a light.css and dark.css doesn't magically solve these problems but if you couple it with `decidePalette` and `format` then all together this potentially forms a solution that, if it works well, could scale. To be honest, I haven't completely thought this through so bear with me.

### How might it look?
Say I'm writing an embed that I want to work cross platform. Let's say it's an email embed. Right now, I enter the colours I use for things directly `background: white;`. This does not work in AR when the reader is in dark mode.

But what if in my embed code I did this

```
import { decidePalette } from @guardian/somehwhere;

myStyles = css`background: ${palette.background.emailEmbed};`
```

This would allow me, the email signup developer, to not think about dark mode but to gain support for it.

A down side to this approach is that I would have to ensure my colours exist in the `decidePalette` code, which means doing the update npm packages dance. In theory we could mitigate this by exposing curated sets of colours that would already have their dark and light versions decided, I guess these would be like themes, but in my experience everyone wants to use brand new colours so I'm not sure how much traction that would get.

And maybe there are other downsides?

Another approach might be to setup the css vars on the site and then offer these up directly to people to use in their code, without the need to import `decidePalette`. Maybe this is a better approach, not sure to be honest. What do you think?

JamieB-gu:
Coming late to this, sorry.

I think there are some advantages and disadvantages to using global CSS variables to implement dark mode. However, based on the previous discussions we've had on this (many of which @oliverlloyd referenced above) I think this is potentially a net positive result, and at least worth experimenting with.

The live apps are, I think, the only place we currently support dark mode. The way it's currently implemented in AR does not rely on CSS variables, but I think we gain some benefits by moving to this approach (not least, support for dark mode in storybook, a major gap in our current regression testing). I also think it's unlikely DCR will be able to implement dark mode using AR's existing solution, due to the expected requirement for an in-page toggle (this is handled natively in apps).

I agree with @oliverlloyd that we should be able to build on the existing `decidePalette`/`editorialPalette` patterns to do this, to reduce the scale of the changes required.

I also agree with @sndrs that nebulous global state in CSS is something to be very wary of. I do think it may be worth it in this case, but I would be against extending this approach to _anything_ beyond the colour palettes required to swap between light and dark mode. I think this is a specific solution to the problem of dark mode, not a general one. On the plus side, I think this would actually be likely to reduce the size of the CSS currently included in each page in AR: there are currently a _lot_ of inline "prefers-color-scheme" media queries that we would drop as a result of this.

I previously wrote up #4241 which expands upon many of these points and includes some possible implementation examples. I hadn't really considered the potential for embeds (email sign-up, interactives etc.); the ideas for those that you describe above sound interesting @oliverlloyd !

oliverlloyd:
> I also agree with @sndrs that nebulous global state in CSS is something to be very wary of. I do think it may be worth it in this case, but I would be against extending this approach to anything beyond the colour palettes required to swap between light and dark mode. I think this is a specific solution to the problem of dark mode, not a general one.

💯 Global css is not something we should ad lightly and should absolutely not be extended just because it exists. It should happen in one place - decidePalette/editorialPalette - and for one thing - colour.


> I previously wrote up #4241 which expands upon many of these points and includes some possible implementation examples. I hadn't really considered the potential for embeds (email sign-up, interactives etc.); the ideas for those that you describe above sound interesting @oliverlloyd !

I'm hearing more and more examples of embeds having issues with the colours like this, not just because of dark mode but also because different types of articles have different colour backgrounds and some embeds need to know what these colours are.

PR #5457, opened 2022-07-20
Iain/gallery layout

iainjchambers-guardian:
## What does this change?
Adds the gallery layout and all necessary components for the header and body


apps-rendering/src/components/FigCaption/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+			);
+
+		default:
+			return null;
+	}
+};
+
+// ----- Exports ----- //
+
+export default FigCaption;
```
JamieB-gu, line 138: Would you be able to talk a little bit about your plan for this component? There's currently an existing `FigCaption` component in `common-rendering`. Are you planning to migrate it into the AR directory and, if so, are we able to delete the common-rendering version? Also, if this is derived from that component is it perhaps based on a slightly older version? @georgeblahblah made some changes in #5376 to remove the triangle in favour of a camera, but I can see this version still has a triangle?

apps-rendering/src/components/Headline/GalleryHeadline.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { grid } from 'grid/grid';
+
+// ----- Component ----- //
+
+const styles = css`
+	${headline.medium({ fontWeight: 'bold' })}
+	background-color: ${neutral[7]};
+	color: ${neutral[100]};
+	padding: 0 ${remSpace[5]} ${remSpace[9]} 0;
+	${grid.between('centre-column-start', 'centre-column-end')}
```
JamieB-gu, line 20: This could be `grid.column.centre` I think? (I've just noticed this is taken from the immersive headline, I'll tweak that too)

```suggestion
	${grid.column.centre}
```

apps-rendering/src/components/Headline/GalleryHeadline.tsx
```diff
@@ -0,0 +1,69 @@
...
+	}
+`;
+
+const backgroundStyles = css`
+	background-color: ${neutral[7]};
+	${grid.between('viewport-start', 'centre-column-end')}
+	grid-row: 3 / 5;
+
+	${from.mobileLandscape} {
+		${grid.between('viewport-start', 'viewport-end')}
```
JamieB-gu, line 39: This could be `grid.column.all` I think?

```suggestion
		${grid.column.all}
```

apps-rendering/src/components/Headline/GalleryHeadline.tsx
```diff
@@ -0,0 +1,69 @@
...
+interface Props {
+	headline: string;
+	format: ArticleFormat;
+}
+
+const GalleryHeadline: React.FC<Props> = ({ headline }) => (
+	<>
+		<div css={backgroundStyles} />
+		<h1 css={styles}>{headline}</h1>
+		<div css={standfirstBackgroundStyles} />
```
JamieB-gu, line 63: Would you be able to talk a little bit about what this is doing? I can see you've also introduced a `GalleryStandfirst` component separately?

apps-rendering/src/components/Layout/Layout.stories.tsx
```diff
@@ -206,6 +207,18 @@ export const DeadBlog = (): ReactElement => (
 );
 DeadBlog.story = { name: 'DeadBlog ' };
 
+export const GalleryItem = (): React.ReactNode => {
+	return (
+		<Gallery item={{ ...article, design: ArticleDesign.Gallery }}>
```
JamieB-gu, line 212: Just FYI, I think galleries only ever include images in the body, whereas I believe this fixture has text, atoms, pullquotes etc. The snapshot may end up looking a bit unusual because we won't have introduced gallery styles for any of those other elements (because we don't need to)?

One solution to this might be to introduce a gallery fixture that only contains images. This could be done in a follow-up PR if necessary?

apps-rendering/src/components/MainMedia/MainMediaImage/index.tsx
```diff
@@ -23,12 +24,16 @@ interface Props {
...
 
-const MainMediaImage: FC<Props> = ({ className, image, format }: Props) => {
+const MainMediaImage: FC<Props> = ({ image, format }: Props) => {
 	if (format.display === ArticleDisplay.Immersive) {
 		return <ImmersiveMainMediaImage image={image} format={format} />;
 	}
 
 	switch (format.design) {
+		case ArticleDesign.LiveBlog:
+			return <ImmersiveMainMediaImage image={image} format={format} />;
```
JamieB-gu, line 34: I don't think we want to use the immersive main media for liveblogs?

```suggestion
```

apps-rendering/src/components/Standfirst/index.tsx
```diff
@@ -42,6 +43,14 @@ const Standfirst: React.FC<Props> = ({ item }) => {
 		case ArticleDesign.DeadBlog:
 			return <DeadBlogStandfirst item={item} />;
 		case ArticleDesign.Gallery:
+			return (
+				<GalleryStandfirst
+					standfirst={item.standfirst}
+					format={format}
+					byline={item.byline}
+					bylineHtml={item.bylineHtml}
```
JamieB-gu, line 51: These props aren't in the `GalleryStandfirst` props type?

apps-rendering/src/components/Layout/GalleryLayout.tsx
```diff
@@ -0,0 +1,54 @@
...
+		<article css={wrapperStyles}>
+			<header css={headerStyles}>
+				<MainMedia
+					mainMedia={item.mainMedia}
+					format={getFormat(item)}
+				/>
+				<Series item={item} />
+				<Headline item={item} />
+				<Standfirst item={item} />
+				<Metadata item={item} />
```
JamieB-gu, line 41: I may have missed it (if so let me know!), but I can't see anywhere that we're rendering a byline? Should that be included in the metadata component?

JamieB-gu (commented):
This is great! Really nice use of CSS grid 👍 

A few questions and suggestions.

georgeblahblah (commented):
This looks really good! Just a few little comments, mostly stylistic

apps-rendering/src/components/FigCaption/index.tsx
```diff
@@ -0,0 +1,138 @@
+// ----- Imports ----- //
```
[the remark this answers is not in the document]
georgeblahblah: oops sorry I see @JamieB-gu also mentioned this in his review! Sorry for the overlap

apps-rendering/src/components/Dateline/index.tsx
```diff
@@ -55,10 +58,12 @@ const blogDatelineStyles = (color: string): SerializedStyles => css`
 
 const getDatelineStyles = (format: ArticleFormat): SerializedStyles => {
 	switch (format.design) {
+		case ArticleDesign.Gallery:
+			return blogDatelineStyles(neutral[100], neutral[100]);
```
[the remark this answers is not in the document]
iainjchambers-guardian: Yeah you're right there, I've made that confusing, needs a more generic name on that function

apps-rendering/src/components/Dateline/index.tsx
```diff
@@ -37,12 +37,15 @@ const commentDatelineStyles = css`
 	${darkStyles}
 `;
 
-const blogDatelineStyles = (color: string): SerializedStyles => css`
+const blogDatelineStyles = (
+	color: string,
+	desktopColor: string,
```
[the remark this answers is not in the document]
iainjchambers-guardian: Yep, well spotted!

apps-rendering/src/components/BodyImage/GalleryBodyImage.tsx
```diff
@@ -0,0 +1,131 @@
...
+			position: absolute;
+			top: 0;
+			bottom: 0;
+			border-left: 1px solid ${neutral[20]};
+		}
+
+		&::before {
+			left: 0;
+			transform: translateX(-10px);
+		}
```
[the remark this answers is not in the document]
iainjchambers-guardian: Well spotted, copied from above and not tidied

apps-rendering/src/components/FigCaption/index.tsx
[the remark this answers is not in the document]
iainjchambers-guardian: Ignore this additional component, this was an attempt to change figCaption and bring it in to AR but I found a better way then forgot I'd added this while playing around with it

apps-rendering/src/components/Headline/GalleryHeadline.tsx
iainjchambers-guardian: Yes, fair question. So, in the immersive headline you added another empty div that sits behind the headline text and gives it a background that stretches to the viewport end. Gallery needed the same behind the standfirst, or so I thought at the time but I can see now that it's closer to finished and I've got the correct colours everywhere I can see that this is overkill

apps-rendering/src/components/Layout/GalleryLayout.tsx
iainjchambers-guardian: No you're absolutely right and now you've pointed that out it does seem like a glaring omission. However, I can't see a byline in the Dotcom galleries either so that's how that happened. Who do I talk to about this?

apps-rendering/src/components/Layout/Layout.stories.tsx
iainjchambers-guardian: Yeah I'm going to remove this for now, at one point I was thinking it might be easier to build and test in storybook and then discovered that the fixture isn't there and that would be a lot of work that I'll most likely need your help on.

apps-rendering/src/components/MainMedia/MainMediaImage/index.tsx
iainjchambers-guardian: Yeah that's a really weird one. Not sure how that happened, I think it might have been a bad merge. Fixed now, whatever happened

apps-rendering/src/components/BodyImage/BodyImage.defaults.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+const DefaultBodyImage: FC<BodyImageProps> = ({
+	image,
+	format,
+	supportsDarkMode,
+	lightbox,
+	caption,
+	leftColumnBreakpoint,
+}) => (
+	<figure css={getDefaultStyles(image.role, leftColumnBreakpoint)}>
```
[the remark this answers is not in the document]
iainjchambers-guardian: Ideally yes, the issue with this component is that there's two (potentially three) items that need to be styled, the `<figure />` wrapper, the `` and the `<FigCaption />`. So it's not as simple as adding the two props you mentioned and passing a `css` prop to the component, it will have to be something more like:
```
figureStyles: SerializedStyles;
imgStyles: SerializedStyles;
captionStyles: SerializedStyles; 
```

apps-rendering/src/components/Layout/GalleryLayout.tsx
JamieB-gu: I suspect not all galleries have one, but many do. Here's an example: https://www.theguardian.com/artanddesign/gallery/2022/jul/22/the-week-around-the-world-in-20-pictures

gallery-byline

iainjchambers-guardian: Brilliant, thanks for digging that out, I'll make sure that gets added in

JamieB-gu (commented):
Looks good, last few small suggestions.

JamieB-gu (approved):
🎉

PR #5467, opened 2022-07-21
Refactor FECollectionConfigType as separate type

bryophyta:
This is to allow easier reference to the config type (e.g. in an upcoming PR where config will be passed from Frontend to a DCR endpoint, but not as part of an FECollectionType.

(It was originally part of #5116, but has been picked out to try to slim the PR down.)

This should be a self-contained edit which doesn't change any behaviour or affect any existing interfaces.

Co-authored-by: Ioanna Kokkini <ioannakok@users.noreply.github.com>

### DIFF
--- a/dotcom-rendering/src/types/front.ts
+++ b/dotcom-rendering/src/types/front.ts
@@ -233,6 +233,24 @@ export type DCRSnapType = {
 	embedCss?: string;
 };
 
+type FECollectionConfigType = {
+	displayName: string;
+	metadata?: { type: FEContainerPalette }[];
+	collectionType: FEContainerType;
+	href?: string;
+	groups?: string[];
+	uneditable: boolean;
+	showTags: boolean;
+	showSections: boolean;
+	hideKickers: boolean;
+	showDateHeader: boolean;
+	showLatestUpdate: boolean;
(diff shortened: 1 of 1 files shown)


mxdvl (approved):

dotcom-rendering/src/types/front.ts
```diff
@@ -233,6 +233,24 @@ export type DCRSnapType = {
 	embedCss?: string;
 };
 
+type FECollectionConfigType = {
+	displayName: string;
+	metadata?: { type: FEContainerPalette }[];
```
jamesgorrie, line 238: Is this optional - or would it return an empty list (optional lists are fiddly to deal with so it's best to try and avoid them unless there's a reason)?

bryophyta: I think that it is optional. This is a collection config from the current UK front:
image

My understanding is that Frontend excludes keys with `None` values during serialisation, although I think you're right that there's an argument for passing an empty array when the `Option` is a list of some sort.

mxdvl: Yes, we should return `Nil` instead of `None`.

PR #5473, opened 2022-07-22
Server render 'More on this Story'

oliverlloyd:
** This PR is dependent on https://github.com/guardian/frontend/pull/25267 **

## What does this change?
This PR moves the work to render the More on this Story onwards container from the client, to the server

## Why?
Because we can. The CAPI response has always included this data, we just never sent it down. Instead we sent a boolean `hasStoryPackage` and if this was true we would make an api call to fetch the data. So now we no longer need to make that api call, which is nice.

### Why is `hasStoryPackage` still there?
We continue to use this boolean to ensure that the decision logic in `OnwardsUpper` does not fallback to another type of request. E.g.: If an article has no story package we often replace it with related content instead. At a later stage we can look to refactor this logic.

### Why are we adding `carouselImages` and `isLiveBlog`?
After running `make gen-fixtures` (to ensure I had the `storyPackage` property) I started to get errors around `CAPITrailType`. I'm not sure if these properties are new or existing but they are recent additions to the data being sent over from Frontend (in fixtures) and need to be typed.


mxdvl (commented):
Shouldn't the schema be updated?

I've done a similar thing with byline, but marked the previous, now unused key as deprecated and unknown. We might want to take the same approach here.

oliverlloyd:
> marked the previous, now unused key as deprecated and unknown

How did you mark it as deprecated? We currently continue to use `hasStoryPackage` (to ensure we display the correct fallback) but I'm interested in any case!

oliverlloyd:
> Shouldn't the schema be updated?

Yes! This is now done.

oliverlloyd:
This code works - the content is server rendered and is then hydrated. This is verified on CODE using https://github.com/guardian/frontend/pull/25267.

But the size of the island you end up with when hydrating is not pretty
Screenshot 2022-07-22 at 15 06 27

mxdvl:
> > marked the previous, now unused key as deprecated and unknown
> 
> How did you mark it as deprecated? We currently continue to use `hasStoryPackage` (to ensure we display the correct fallback) but I'm interested in any case!

Using a `@deprecated` JSDoc comment.

https://github.com/guardian/dotcom-rendering/blob/8cdb1bafb6beafe51046cec1ae6db4a287949e26/dotcom-rendering/index.d.ts#L551-L552

From https://github.com/guardian/dotcom-rendering/pull/5378/commits/8cdb1bafb6beafe51046cec1ae6db4a287949e26

dotcom-rendering/src/web/layouts/CommentLayout.tsx
```diff
@@ -696,6 +698,22 @@ export const CommentLayout = ({ CAPIArticle, NAV, format }: Props) => {
 					/>
 				</ElementContainer>
 
+				{CAPIArticle.storyPackage && (
+					<ElementContainer>
+						<Island deferUntil="visible">
```
OllysCoding, line 703: The 'island' behaviour here is just for hydrating the carousel, right?

oliverlloyd: Yes 😐

dotcom-rendering/src/model/article-schema.json
```diff
@@ -321,6 +321,24 @@
         "hasStoryPackage": {
             "type": "boolean"
         },
+        "storyPackage": {
```
oliverlloyd, line 324: Frontend now sends this data to DCR (thanks to https://github.com/guardian/frontend/pull/25267)

dotcom-rendering/src/web/components/Carousel.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import { ArticleDesign, ArticleDisplay, ArticlePillar } from '@guardian/libs';
 import { breakpoints } from '@guardian/source-foundations';
-import { Carousel } from './Carousel';
+import { Carousel } from './Carousel.importable';
```
oliverlloyd, line 3: Previously (and elsewhere) we hydrate this component as part of `OnwardsUpper`. But here it sits alone and needs hydrating by itself

OllysCoding (approved):
+1 All looks good :)

PR #5475, opened 2022-07-22
DCR: 🛠 Debug bookmarklet 🛠 (10% time)

OllysCoding:
A _lot_ of stuff happens when DCR renders a page - sometimes it's hard to figure out exactly what's going on!

The DCR debug bookmarklet is here to the rescue, with handy annotations for elements*, islands, front containers & layouts!

_*only some elements supported - [bareElements](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/lib/renderElement.tsx#L769) are not currently supported._

## How does it work?

We've created a new entry-point into the DCR client bundle: this script downloads some CSS onto the page which does all of the heavy lifting here.

Importantly, the output of this is not hashed in the webpack bundle, allowing us to access it without needing to know the latest hash (if we didn't do this, nobodies bookmarklets would update :()

To enable the annotation of the layout, we've added `data-layout` to the `<main >` element in each of the layout files!

## How do I use it?!?!?!?!

Until this PR is merged & deployed, you can clone this branch and run the dev server locally which will allow you to use the local bookmarklet:
```
javascript: (() => {
    document.body.appendChild(document.createElement('script')).src =
        'http://localhost:3030/assets/debug.js';
})();
```
⚠️ The local script will only work on localhost:3030 articles, not on prod articles!

Once this PR is merged you can grab the live version:
```
javascript: (() => {
    document.body.appendChild(document.createElement('script')).src =
        'https://assets.guim.co.uk/assets/debug.js';
})();
```

## Sreenshots:

Note: in this early version the styling is far from perfect, the information however is very usable!
image
image
image
image


dotcom-rendering/package.json
```diff
@@ -184,6 +184,7 @@
     "stylelint-config-styled-components": "^0.1.1",
     "stylelint-processor-styled-components": "^1.10.0",
     "swr": "^1.3.0",
+    "to-string-loader": "^1.2.0",
```
OllysCoding, line 187: We referenced 'to-string-loader' in our webpack browser bundle, but it wasn't actually installed! This lets us get the contents of the `debug.css` file and add it to the document!

dotcom-rendering/src/web/layouts/StandardLayout.tsx
```diff
@@ -458,7 +458,7 @@ export const StandardLayout = ({ CAPIArticle, NAV, format }: Props) => {
 				<AdSlot position="survey" display={format.display} />
 			)}
 
-			<main>
+			<main data-layout="StandardLayout">
```
oliverlloyd, line 461: It's annoying that it seems so hard to get the name of a component from inside itself

oliverlloyd (commented):
curb-your-enthusiasm-there-you-go

dotcom-rendering/src/web/layouts/StandardLayout.tsx
OllysCoding: Yeah this is frustrating, but I think for one use per layout there's not even much value in trying to find an abstraction for it

oliverlloyd: Agreed

dotcom-rendering/scripts/webpack/webpack.config.browser.js
```diff
@@ -31,9 +32,15 @@ module.exports = ({ isLegacyJS, sessionId }) => ({
...
 		relativeTime: './src/web/browser/relativeTime/init.ts',
 		initDiscussion: './src/web/browser/initDiscussion/init.ts',
+		debug: './src/web/browser/debug/init.ts',
 	},
 	output: {
-		filename: generateName(isLegacyJS),
+		filename: (data) => {
+			// We don't want to hash the debug script so it can be used in bookmarklets
+			if (data.chunk.name === 'debug')
+				return generateName(isLegacyJS, false);
```
jamesgorrie, line 41: Could we remove the abstraction and just use:
```suggestion
				return `[name].js`
```

As we don't want to use this anywhere else and don't need to support chunkname nor legacy?

dotcom-rendering/scripts/webpack/webpack.config.dev-server.js
```diff
@@ -30,6 +30,7 @@ module.exports = {
 			directory: path.join(__dirname, '..', '..', 'src', 'static'),
 			publicPath: '/static/frontend',
 		},
+		allowedHosts: ['r.thegulocal.com'],
```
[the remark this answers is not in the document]
jamesgorrie: Was this needed for this PR? Happy for it to stay, I just couldn't work out if it as needed for this.

dotcom-rendering/src/web/browser/debug/README.md
```diff
@@ -0,0 +1,28 @@
...
+javascript: (() => {
+    document.body.appendChild(document.createElement('script')).src =
+        'http://localhost:3030/assets/debug.js';
+})();
+```
+
+## After publishing
+
+Assets on assets.guim.co.uk are cached by fastly - to allow users to receive the latest update of this bookmarklet,
+you should log into the fastly dashboard and clear the cache!
```
jamesgorrie, line 28: If we want to keep the readme close to the code, could we also create a link in the main readme for discoverability?

dotcom-rendering/src/web/layouts/StandardLayout.tsx
jamesgorrie: Agree on the abstraction - it feels like it would be fiddly. I wonder if we could write a test to ensure all layouts have a name for people adding a future one to be prompted and to describe why it's there, but I'm not sure if that's any less fiddly.

jamesgorrie (commented):
This is a great tool, and feels like it will help us describe some core principles to future Dotcom-ers in a really clear way.

A couple of nit picks, but looking great.

dotcom-rendering/scripts/webpack/webpack.config.browser.js
mxdvl: Agree with this suggestion!

dotcom-rendering/scripts/webpack/webpack.config.dev-server.js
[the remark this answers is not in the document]
OllysCoding: It isn't -  there was a point where I got lost down a rabbit hole of testing the nginx script in DCR & noticed it wouldn't work due to this variable not being set.

dotcom-rendering/src/web/browser/debug/debug.css
```diff
@@ -0,0 +1,43 @@
+/* Layout: Annotate */
+main[data-layout]::before {
+	content: 'layout: ' attr(data-layout);
```
mxdvl, line 3:
```suggestion
	content: '📰 layout: ' attr(data-layout);
```

dotcom-rendering/scripts/webpack/webpack.config.browser.js
OllysCoding: Wouldn't doing it this way mean the legacy bundle would just overwrite the original bundle when generated, since we're giving it the same name?

mxdvl (approved):
Love it!

What do you think about using a Guardian palette? Are we trying to contrast as much as possible with the existing content?

dotcom-rendering/scripts/webpack/webpack.config.browser.js
mxdvl: Does it matter? We don’t really need two files. But you could always turn it into a ternary operator.

An alternative would be to simply add a copy step and serve the files untransformed, without the Webpack step. You would have to import the css in a JS way, however, so this might be more effort.

dotcom-rendering/src/web/browser/debug/debug.css
```diff
@@ -0,0 +1,43 @@
+/* Layout: Annotate */
+main[data-layout]::before {
```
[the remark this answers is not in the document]
OllysCoding: Disadvantage of this route is that the layout would (without additional styling) be annotated at the bottom of the page rather than the top as it is now

OllysCoding:
> What do you think about using a Guardian palette? Are we trying to contrast as much as possible with the existing content?

I think by using these bold 'dev design' type styles are important in distinguishing that these debug annotations are not meant to be seen (in the production sense). I think its purpose is to be different from the rest of the site, if that makes sense?

jamesgorrie:
> I think its purpose is to be different from the rest of the site

This makes a lot of sense. I also like that we just use `red` or `blue` or whatever, as then the thought is put into the tool and being clear, rather than spending time thinking of the right one (I'd normally use `hotpink`).

jamesgorrie (approved):
Looking good!

dotcom-rendering/src/web/browser/debug/debug.css
jamesgorrie: I wonder about the emojis and accessibility here, would this not read out as "Newspaper layout..." which is quite confusing.

dotcom-rendering/src/web/browser/debug/README.md
```diff
@@ -0,0 +1,30 @@
...
+```js
+javascript: (() => {
+    document.body.appendChild(document.createElement('script')).src =
+        'http://localhost:3030/assets/debug.js';
+})();
+```
+
+## After publishing
+
+Assets on assets.guim.co.uk are cached by fastly - to allow users to receive the latest update of this bookmarklet,
```
bryophyta, line 29:
```suggestion
Assets on assets.guim.co.uk are cached by fastly. To allow users to receive the latest version of this bookmarklet after an update,
```

bryophyta: I tripped up slightly on the original phrasing, as I thought it was saying that it was cached in order to allow users to receive the latest updates. This might just be me though!

dotcom-rendering/src/web/browser/debug/debug.css
bryophyta: I'd be tempted to go for `Page layout` here, or something like that? Just because we also have things like `ContainerLayout`, which are different

bryophyta (approved):
A couple of very small suggestions on phrasing, but nothing blocking. This looks great to me!

dotcom-rendering/src/web/browser/debug/debug.css
OllysCoding: I've gone with page layout - mostly for descriptiveness, though the emojis are kind of cool!

oliverlloyd:
Just a thought, could we update the PR with a hint on how to use this javascript snippet. I _think_ this is a chrome extension thing? But I'm not sure.

```js
javascript: (() => {
    document.body.appendChild(document.createElement('script')).src =
        'https://assets.guim.co.uk/assets/debug.js';
})();
```

oliverlloyd:
> Just a thought, could we update the PR with a hint on how to use this javascript snippet. I _think_ this is a chrome extension thing? But I'm not sure.
> 
> ```js
> javascript: (() => {
>     document.body.appendChild(document.createElement('script')).src =
>         'https://assets.guim.co.uk/assets/debug.js';
> })();
> ```

Ignore me, reading the README I can see there is more detail on this 👍

dotcom-rendering/src/web/components/ContainerLayout.tsx
```diff
@@ -34,6 +34,7 @@ type Props = {
 	format?: ArticleFormat;
 	ophanComponentName?: string;
 	ophanComponentLink?: string;
+	containerName?: string;
```
oliverlloyd, line 37: I'm not super strong on this so feel free to ignore me but is adding this prop increasing the cognitive complexity of things here? This component has a lot of props!

Maybe if it were marked as debug only somehow then future developers would not need to worry about what it is?

oliverlloyd (approved):
This is very cool. I am especially going to enjoy knowing which container I'm looking at!

dotcom-rendering/src/web/components/ContainerLayout.tsx
OllysCoding: Yeah I didn't love this either - the other name I'd considered was `dataContainerName` to try and suggest its just there for a data attribute, but we could also try `debugContainerName` - my fear here was introducing a new pattern for debug properties could increase the cognitive load just as much if you're a developer trying to figure out what that means.

oliverlloyd: Would it work if we used the `ophanComponentName` for debugging?

oliverlloyd: Fair

PR #5481, opened 2022-07-25
feat(eslint): catch falsy JSX fragments

mxdvl:
## What does this change?

Add and enforce the following [strict-logical-expressions](https://github.com/hpersson/eslint-plugin-jsx-expressions/blob/master/docs/rules/strict-logical-expressions.md) ESLint /JSX rule.

## Why?

We have had a number of times where this has bitten us:
- #5344
- #4712
- #3149
- #1441

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/180758412-27c079ad-ab02-44b5-bbd6-cdf4879eb699.png
[after]: https://user-images.githubusercontent.com/76776/180758261-dc7974b1-2327-46eb-b066-bb669e3fe930.png

### DIFF
--- a/dotcom-rendering/src/web/components/Dropdown.tsx
+++ b/dotcom-rendering/src/web/components/Dropdown.tsx
@@ -380,10 +380,9 @@ export const Dropdown = ({
 											)}
 										</a>
 
-										{l.notifications?.length &&
-											l.notifications.length > 0 && (
-												<RedDot />
-											)}
+										{!!l.notifications?.length && (
+											<RedDot />
+										)}
 									</li>
 								))}
 							</ul>
--- a/dotcom-rendering/src/web/components/FilterButton.importable.tsx
+++ b/dotcom-rendering/src/web/components/FilterButton.importable.tsx
@@ -62,7 +62,7 @@ const valueStyles = css`
 const Label = ({ value, count }: LabelProps) => (
 	<>
 		<span css={valueStyles}>{value}</span>{' '}
-		{count && <span css={countStyles}>({count})</span>}
+		{!!count && <span css={countStyles}>({count})</span>}
 	</>
 );
 
--- a/dotcom-rendering/src/web/components/ImageComponent.tsx
+++ b/dotcom-rendering/src/web/components/ImageComponent.tsx
@@ -306,7 +306,7 @@ export const ImageComponent = ({
 					isLazy={!isMainMedia}
 					isMainMedia={isMainMedia}
 				/>
-				{title && (
+				{!!title && (
 					<ImageTitle title={title} role={role} palette={palette} />
 				)}
 			</div>
@@ -337,8 +337,10 @@ export const ImageComponent = ({
 					isLazy={!isMainMedia}
 					isMainMedia={isMainMedia}
 				/>
-				{starRating && <PositionStarRating rating={starRating} />}
-				{title && (
+				{typeof starRating === 'number' && (
+					<PositionStarRating rating={starRating} />
+				)}
+				{!!title && (
 					<ImageTitle title={title} role={role} palette={palette} />
 				)}
 			</div>
@@ -407,8 +409,10 @@ export const ImageComponent = ({
 						</Row>
 					</Hide>
 				)}
-				{starRating && <PositionStarRating rating={starRating} />}
-				{title && (
+				{typeof starRating === 'number' && (
+					<PositionStarRating rating={starRating} />
+				)}
+				{!!title && (
 					<ImageTitle title={title} role={role} palette={palette} />
 				)}
 			</div>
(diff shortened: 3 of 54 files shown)


JamieB-gu:
This looks like a good idea to me, but I'll let the Dotcom team decide/approve given this is DCR 🙂. Didn't realise there was a rule for this, might add it to AR too ⚙️.

OllysCoding:
Is this too strict? 

For example, to my understanding

```jsx
// inputs
{myOptionalString && <element />}
// possible outputs
<element />
// or
"" // - This is just an empty string, this shouldn't cause problems?
// or 
null / undefined // Again I don't think this adds anything to the DOM? 

// inputs
{myOptionalNumber && <element />
// possible outputs
<element />
// or
0 // - This is bad
// or 
null / undefined // I don't think this adds anything to the DOM? 
```

Am I wrong in this? Just trying to figure out if it's only numbers we need to target or if there are other truthy/fasley conditions that pose risk?



 

```

**Review by OllysCoding — APPROVED** [review]
+1 This looks good, 1 question added around how strict this is :)

**Review by oliverlloyd — COMMENTED** [review]
That's a lot of bangs... Is there a way to have a rule that only applies when the variable is numeric? I'm guessing not but I can see this blanket rule being pesky.

Arguably, pesky is preferable to publishing `0` though!

**Comment by mxdvl** [comment]
> Is this too strict? @OllysCoding 

> That's a lot of bangs... Is there a way to have a rule that only applies when the variable is numeric? I'm guessing not but I can see this blanket rule being pesky. @oliverlloyd 

This rule is auto-fixable, meaning that there’s no user input required, it will simply be fixed once you save the file. While it’s true that adding an empty string has the same result, I think we favour being explicit, and would opt for keeping the rule as-is, and not only target numerical values.

[According to the React docs](https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored), empty strings are **not** part of what is ignored by the renderer.

File: dotcom-rendering/src/web/components/Dropdown.tsx [hunk 1]
```diff
@@ -380,10 +380,9 @@ export const Dropdown = ({
 											)}
 										</a>
 
-										{l.notifications?.length &&
-											l.notifications.length > 0 && (
-												<RedDot />
-											)}
+										{!!l.notifications?.length && (
+											<RedDot />
+										)}
```
mxdvl, line 385: Small refactor here, @guardian/tx-engineers. If the length is `0`, it would be falsy and therefore not render the dot.

dotcom-rendering/src/web/components/FilterButton.importable.tsx
```diff
@@ -62,7 +62,7 @@ const valueStyles = css`
 const Label = ({ value, count }: LabelProps) => (
 	<>
 		<span css={valueStyles}>{value}</span>{' '}
-		{count && <span css={countStyles}>({count})</span>}
+		{!!count && <span css={countStyles}>({count})</span>}
```
mxdvl, line 65: Hey @guardian/journalism-dev I’m assuming here that you don’t want to show labels that have a count of `0`, but if the logic should be different, let me know.

dotcom-rendering/src/web/components/ImageComponent.tsx
```diff
@@ -337,8 +337,10 @@ export const ImageComponent = ({
 					isLazy={!isMainMedia}
 					isMainMedia={isMainMedia}
 				/>
-				{starRating && <PositionStarRating rating={starRating} />}
-				{title && (
+				{typeof starRating === 'number' && (
+					<PositionStarRating rating={starRating} />
+				)}
```
mxdvl, line 342: Here we want to handle a rating of `0`, as [it is valid and happens](http://content.guardianapis.com/search?api-key=test&page-size=100&show-fields=starRating&star-rating=0).

dotcom-rendering/src/web/components/RichLink.tsx
```diff
@@ -282,21 +282,21 @@ export const RichLink = ({
 									{mainContributor}
 								</div>
 							)}
-							{starRating && starRating > 0 && (
+							{!!starRating && (
```
mxdvl, line 285: This is a fix, as star ratings of `0` [are actually valid](http://content.guardianapis.com/search?api-key=test&page-size=100&show-fields=starRating&star-rating=0)!

dotcom-rendering/src/web/layouts/StandardLayout.tsx
```diff
@@ -487,42 +487,40 @@ export const StandardLayout = ({ CAPIArticle, NAV, format }: Props) => {
...
-													CAPIArticle.headline
-												}
-												tags={CAPIArticle.tags}
-												webPublicationDateDeprecated={
-													CAPIArticle.webPublicationDateDeprecated
-												}
-											/>
-										</Island>
-									)}
+								{isMatchReport && (
```
mxdvl, line 490: The compiler knows that if `isMatchReport` is `true`, `matchUrl` is defined.

mxdvl (commented):
Some notes about the refactor where it’s not simply the automatic fix:

PR #5482, opened 2022-07-25
Bug/fix 0 when no caresoul

marjisound:
## What does this change?
This PR fixes a bug where the number `0` was showing up when there was no key events.
## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

### DIFF
--- a/dotcom-rendering/src/web/layouts/LiveLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/LiveLayout.tsx
@@ -598,7 +598,7 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 						</GridItem>
 					</StandFirstGrid>
 				</ElementContainer>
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -598,7 +598,7 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 						</GridItem>
 					</StandFirstGrid>
 				</ElementContainer>
-				{showKeyEventsCarousel && CAPIArticle.keyEvents.length && (
+				{showKeyEventsCarousel && CAPIArticle.keyEvents.length > 0 ? (
```
oliverlloyd, line 601: Am I right that _technically_ making this a comparison means it will return a boolean so we don't need, need the ternary? @mxdvl @OllysCoding 

Either way, having it here means we're solid

oliverlloyd (approved):

Issue #5093, opened 2022-05-30
Avatar on comment cards


oliverlloyd:
`Card` already has the avatar included but it is not appearing on DCR Fronts 

<img width="233" alt="Screenshot 2022-05-30 at 16 44 11" src="https://user-images.githubusercontent.com/1336821/171026243-6dbcc7f3-4d41-43a5-aa7f-9b3e6a8dafc1.png">

HarryFischer:
Size and positioning slightly changed - see here:
https://www.figma.com/file/5CfbWOeZPRi15lXBD5u1rW/?node-id=0%3A1

PR #5484, opened 2022-07-25
Card avatars

oliverlloyd:
## What does this change?
This adds the logic to display avatars on `Card`s

### When should an avatar be displayed?
This PR is stating that an avatar should only display if:

1) The container layout supports the use of an image
Cards that do not have the `imageUrl` property set in their layout will also not show an avatar

2) There is a single contributor tag and its title matches the byline

3) There is an `image` property and its `type` is equal to 'Cutout'.

| Before      | After      |
|-------------|------------|
| Screenshot 2022-07-25 at 12 34 23 | Screenshot 2022-07-25 at 12 33 53 |

Fixes https://github.com/guardian/dotcom-rendering/issues/5093

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -685,11 +685,6 @@ type SmallHeadlineSize =
 	| 'huge'
 	| 'ginormous';
 
-type AvatarType = {
-	src: string;
-	alt: string;
-};
-
 type MediaType = 'Video' | 'Audio' | 'Gallery';
 
 type LineEffectType = 'labs' | 'dotted' | 'straight';
--- a/dotcom-rendering/src/model/enhanceCards.ts
+++ b/dotcom-rendering/src/model/enhanceCards.ts
@@ -78,10 +78,9 @@ const enhanceSupportingContent = (
 };
 
 const decideAvatarUrl = (
-	tags?: TagType[],
+	tags: TagType[] = [],
 	byline?: string,
 ): string | undefined => {
-	if (!tags || tags.length === 0) return undefined;
 	const soleContributor = getSoleContributor(tags, byline);
 	return soleContributor?.bylineLargeImageUrl ?? undefined;
 };
--- a/dotcom-rendering/src/model/enhanceCards.ts
+++ b/dotcom-rendering/src/model/enhanceCards.ts
@@ -81,9 +81,8 @@ const decideAvatarUrl = (
 	tags?: TagType[],
 	byline?: string,
 ): string | undefined => {
-	if (!tags || tags.length === 0) return;
+	if (!tags || tags.length === 0) return undefined;
 	const soleContributor = getSoleContributor(tags, byline);
-	// eslint-disable-next-line consistent-return -- because I want to return undefined and you can't stop me
 	return soleContributor?.bylineLargeImageUrl ?? undefined;
 };
 
--- a/dotcom-rendering/src/model/enhanceCards.ts
+++ b/dotcom-rendering/src/model/enhanceCards.ts
@@ -1,10 +1,12 @@
 import { ArticleDesign, ArticleSpecial } from '@guardian/libs';
+import { getSoleContributor } from '../lib/byline';
 import type {
 	DCRContainerPalette,
 	DCRFrontCard,
 	DCRSupportingContent,
 	FEFrontCard,
 	FESupportingContent,
+	FETagType,
 } from '../types/front';
 import { decideFormat } from '../web/lib/decideFormat';
 import { getDataLinkNameCard } from '../web/lib/getDataLinkName';
@@ -75,6 +77,38 @@ const enhanceSupportingContent = (
 	});
 };
 
+const decideAvatarUrl = (
+	tags?: TagType[],
+	byline?: string,
+): string | undefined => {
+	if (!tags || tags.length === 0) return;
+	const soleContributor = getSoleContributor(tags, byline);
+	// eslint-disable-next-line consistent-return -- because I want to return undefined and you can't stop me
+	return soleContributor?.bylineLargeImageUrl ?? undefined;
+};
+
+const enhanceTags = (tags: { properties: FETagType }[]): TagType[] => {
+	return tags.map((tag) => {
+		const {
+			id,
+			tagType,
+			webTitle,
+			twitterHandle,
+			bylineImageUrl,
+			contributorLargeImagePath,
+		} = tag.properties;
+
+		return {
+			id,
+			type: tagType,
+			title: webTitle,
+			twitterHandle,
+			bylineImageUrl,
+			bylineLargeImageUrl: contributorLargeImagePath,
+		};
+	});
+};
+
 export const enhanceCards = (
 	collections: FEFrontCard[],
 	containerPalette?: DCRContainerPalette,
@@ -92,6 +126,11 @@ export const enhanceCards = (
 			faciaCard.display.isBoosted ? '+' : ''
 		}`;
 		const dataLinkName = getDataLinkNameCard(format, group, index + 1);
+
+		const tags = faciaCard.properties.maybeContent?.tags.tags
+			? enhanceTags(faciaCard.properties.maybeContent.tags.tags)
+			: [];
+
 		return {
 			format,
 			dataLinkName,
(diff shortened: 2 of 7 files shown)


dotcom-rendering/index.d.ts
```diff
@@ -685,11 +685,6 @@ type SmallHeadlineSize =
 	| 'huge'
 	| 'ginormous';
 
-type AvatarType = {
-	src: string;
-	alt: string;
-};
-
```
oliverlloyd, line 692: This type did not align with the wider usage of avatarUrl and there was no good solution for `alt` text. We now use the byline for alt text

dotcom-rendering/src/model/enhanceCards.ts
```diff
@@ -75,6 +77,38 @@ const enhanceSupportingContent = (
 	});
 };
 
+const decideAvatarUrl = (
```
oliverlloyd, line 80: This function uses the new `FETagType` type (which is the data we get from FE) and transforms the tags array into `TagType[]`. We need this transformation to be able to pass `tags` to the `getSoleContributor` function to decide if an avatar should show or not

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -277,6 +277,11 @@ export const Card = ({
 		return <Snap snapData={snapData} />;
 	}
 
+	// Decide what type of image to show, main media, avatar or none
+	let imageType: 'mainmedia' | 'avatar' | undefined;
```
oliverlloyd, line 281: This variable is used to determine when an avatar should show. By using a variable like this we ensure we never show _both_ the main media image and the avatar at the same time

dotcom-rendering/src/model/enhanceCards.ts
```diff
@@ -75,6 +77,38 @@ const enhanceSupportingContent = (
...
+	tags?: TagType[],
+	byline?: string,
+): string | undefined => {
+	if (!tags || tags.length === 0) return;
+	const soleContributor = getSoleContributor(tags, byline);
+	// eslint-disable-next-line consistent-return -- because I want to return undefined and you can't stop me
```
mxdvl, line 86: The problem is that you’re returning inconsistently two lines above, which I don’t think you need to do.

```suggestion
	const soleContributor = getSoleContributor(tags, byline);
	// eslint-disable-next-line consistent-return -- because I want to return undefined and you can't stop me
```

dotcom-rendering/src/model/enhanceCards.ts
mxdvl: Can you require that tags are defined?

dotcom-rendering/src/types/front.ts
```diff
@@ -90,6 +90,21 @@ export type DCRContainerPalette =
...
+export type FETagType = {
+	id: string;
+	url: string;
+	tagType: string;
+	sectionId: string;
+	sectionName: string;
+	webTitle: string;
+	webUrl: string;
+	twitterHandle?: string;
+	// bio is html
```
mxdvl, line 102: JSDoc please 🥺

dotcom-rendering/src/web/components/Card/Card.tsx
mxdvl: I think the best approach for stuff like this is to extract it to a helper function, rather that mutating values inline.

```ts
const getImage = ({ imageUrl, avatarUrl }): ['avatar' | 'mainmedia', string] | [undefined, undefined] => {
  if(avatarUrl) return ['avatar', avatarUrl];
  if(imageUrl) return ['mainmedia', imageUrl]:
  return [undefined, undefined];
}

// inside the component:
const [imageType, image] = getImage({ imageUrl, avatarUrl });
```

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -277,6 +277,11 @@ export const Card = ({
...
+	if (imageUrl && avatarUrl) imageType = 'avatar';
```
mxdvl, line 282: Do we only want to show an avatar if there is an image, is that right?

If not, you can use my helper function above as-is.

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -348,12 +353,12 @@ export const Card = ({
 								/>
 							) : undefined}
 						</HeadlineWrapper>
-						{avatar && (
+						{imageType === 'avatar' && avatarUrl && (
```
mxdvl, line 356: Using the helper method, you wount no longer have to check for the truthiness of `avatarUrl`, but rather can assume that `image` is always string if `imageType` is defined.

mxdvl (commented):
Some thoughts below:

dotcom-rendering/src/web/components/Card/Card.tsx
oliverlloyd: I added the extra check here to keep typescript happy. We know from reading the code that if imageType === avatar then avatarUrl can't be undefined but ts doesn't

dotcom-rendering/src/web/components/Card/Card.tsx
mxdvl: I don’t think the curly braces are much different.

The benefit of the abstraction is that you can lean more heavily on the compiler, which will know that if `imageType` is set to `"avatar"`, an `avatarUrl` will be defined. It seems to not be able to infer that from the if logic.

dotcom-rendering/src/web/components/Card/Card.tsx
mxdvl: Hence my proposal to create a helper that can help discriminate.

dotcom-rendering/src/web/components/Card/Card.tsx
mxdvl: If we only ever show one image, that is either the avatar or the image, could we discriminate the type for this?

oliverlloyd: I tried creating an abstraction here but I am still getting typescript errors 😐 

```ts
	const decideGraphic = (imageUrl?: string, avatarUrl?: string) => {
		if (imageUrl && avatarUrl) {
			return 'avatar';
		} else if (imageUrl) {
			return 'image';
		}
		return 'none';
	};
```

I think the only way to avoid these would be to create a _new_ variable for the `imageUrl` and `avatarUrl`, like `theImageSourceToUse`. But adding extra complexity and abstraction to rename two variables to one value just to avoid a guard in the code doesn't seem like a good approach to me.

dotcom-rendering/src/model/enhanceCards.ts
oliverlloyd: It still needs the check for `tags` but it seems ts doesn't recognise `return` as being the same as `return undefined`, so I helped it out by saying that

oliverlloyd:
> Can you require that tags are defined?

I didn't follow this comment (GH isn't assigning it to a line) - can you elaborate 🙏

dotcom-rendering/src/model/enhanceCards.ts
mxdvl: You could forgo checking for tags by setting a default in where you’re calling `getSoleContributor` like so:

```ts
getSoleContributor(tags ?? [], byline);
```

dotcom-rendering/src/model/enhanceCards.ts
mxdvl: (This is where I’ve added the comment)

I was thinking that `tags` should be required for this, or at lease default to an empty array.

oliverlloyd: Yep, now done 👍

dotcom-rendering/src/model/enhanceCards.ts
oliverlloyd: Added a default

dotcom-rendering/src/web/components/Card/Card.tsx
mxdvl: Fair enough.

mxdvl (approved):
Would be nice to squash some of these commits 😉

PR #5495, opened 2022-07-25
Hover styles for cards inside palette containers

### LINKED ISSUE #5268: Hover colours for palette cards
When a `Card` has a `containerPalette` applied it needs special hover colours.

`CardWrapper` currently manages the hover colour using `filter(brightness)` but perhaps we I think we need a more explicit decision. Perhaps we can use `decidePalette` here instead?

oliverlloyd:
## What does this change?
If a `Card` is sitting in a container that has a `containerPalette` applied to it then this PR ensures that this card is given a specific hover style

Closes #5268 

## Why?
Cards inside palette containers have their colours overriden so we need the hover styles to be set based on this.

| Before      | After      |
|-------------|------------|
|  |  |

### DIFF
--- a/dotcom-rendering/src/web/components/Card/components/CardWrapper.tsx
+++ b/dotcom-rendering/src/web/components/Card/components/CardWrapper.tsx
@@ -21,6 +21,7 @@ const cardStyles = (
 	format: ArticleFormat,
 	palette: Palette,
 	isDynamo?: true,
+	containerPalette?: DCRContainerPalette,
 ) => {
 	const baseCardStyles = css`
 		display: flex;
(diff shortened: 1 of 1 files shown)


ioannakok (approved):

dotcom-rendering/src/web/components/Card/components/CardWrapper.tsx
```diff
@@ -47,6 +48,39 @@ const cardStyles = (
...
+				return `84%`;
+			case 'LongRunningAltPalette':
+				return `95%`;
+			case 'SombrePalette':
+				return `90%`;
+			case 'SombreAltPalette':
+				return `85%`;
+			default:
+				return `90%`;
+		}
```
OllysCoding, line 71: Is there a world where this is set in [decideContainerOverrides.ts](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/lib/decideContainerOverrides.ts)? 

I'm conscious that when adding snippets like this, we fragment the changes required when a new palette is created - it's now another place you need to know where to go when editing / creating palettes

oliverlloyd: We can't set these values in `decidePalette` (or friends) because that only returns colours (strings) but the hover css is... well, I guess it wouldn't be a colour but it could be a string.... Hmm. I'll have a look 👍

PR #5498, opened 2022-07-25
WIP: Test different Webpack chunking strategies

mxdvl:
## What does this change?

Serve many new webpack chunks and compare their impact on our performance metrics.

## Why?

Since #5305, we’ve noticed slightly worse performance metrics, despite reverting to Webpack’s defaults.

### DIFF
--- a/dotcom-rendering/scripts/webpack/islands.js
+++ b/dotcom-rendering/scripts/webpack/islands.js
@@ -1,0 +1,13 @@
+// @ts-check
+
+const { readdirSync } = require('fs');
+// eslint-disable-next-line @typescript-eslint/unbound-method -- it’s the node:path way
+const { resolve } = require('path');
+
+const islands = Object.fromEntries(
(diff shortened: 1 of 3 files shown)


arelra:
> Since #5305, we’ve noticed slightly worse performance metrics, despite reverting to Webpack’s defaults.

Interesting!

Do we have more detail/graphs on the worsening performance?

A lot of volatility in metrics is caused by commercial i.e ad rendering so would be good to see a breakdown by `adFree`.

arelra:
Also worth looking at more fine grained control of cache groups e.g.:
https://webpack.js.org/plugins/split-chunks-plugin/#splitchunkscachegroupscachegrouptest
https://webpack.js.org/plugins/split-chunks-plugin/#split-chunks-example-3

arelra:
Could we use the existing server side A/B testing framework to test builds?

More here: https://github.com/guardian/dotcom-rendering/issues/5519

We could then have D&I report on web vitals pre and post #5305.

PR #5509, opened 2022-07-26
Ensure `index` is unique when`UnsafeEmbedBlockcomponent`s exist in pinned posts

oliverlloyd:
## What does this change?
This ensures that the `UnsafeEmbedBlockcomponent` uses a unique index value when there are two of the same embed on the page, one in a pinned post and the other in the original post in the article body

## Why?
[This article](https://www.theguardian.com/politics/live/2022/jul/26/tory-leadership-race-liz-truss-rishi-sunak-tv-debate-uk-politics-live) has two embeds the same on the page but they both have the same index (`2`) and this is preventing the iframe resizing logic from working

| Before      | After      |
|-------------|------------|
| Screenshot 2022-07-26 at 15 21 04 | Screenshot 2022-07-26 at 15 20 39 |

### DIFF
--- a/dotcom-rendering/src/web/components/LiveBlock.tsx
+++ b/dotcom-rendering/src/web/components/LiveBlock.tsx
@@ -80,6 +80,7 @@ export const LiveBlock = ({
 					isAdFreeUser={isAdFreeUser}
 					isSensitive={isSensitive}
 					switches={switches}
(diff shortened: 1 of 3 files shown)


marjisound (approved):
Great work thanks 👍

OllysCoding (approved):
+1 Good catch!

Issue #5510, opened 2022-07-26
Latest articles where `iframe` have no titles


mxdvl:
**WIP** Do not attempt to fix these individually.

Querying http://content.guardianapis.com/search?api-key=test&query-fields=body&q=iframe&show-fields=body&page-size=100

## Results

- [ ] [Stuart Daulman: the 10 funniest things I have ever seen (on the internet)](https://www.theguardian.com/culture/2022/jul/14/stuart-daulman-the-10-funniest-things-i-have-ever-seen-on-the-internet)
- [X] [With inflation rising in Australia, an uncomfortable truth is we waste a lot of what we have | Jessica Mizrahi](https://www.theguardian.com/commentisfree/2022/jul/19/with-inflation-rising-in-australia-an-uncomfortable-truth-is-we-waste-a-lot-of-what-we-have)
- [ ] [‘I’ve denied it for years, but Taylor Swift really can write’: Ben Bailey Smith’s honest playlist](https://www.theguardian.com/culture/2022/jul/18/ben-bailey-smith-honest-playlist)
- [ ] [‘Jailhouse Rock made prison seem like fun’: Billy Gibbons’s honest playlist](https://www.theguardian.com/culture/2022/jul/25/billy-gibbons-zz-top-honest-playlist)
- [ ] [Bridie Connell: the 10 funniest things I have ever seen (on the internet)](https://www.theguardian.com/culture/2022/jul/06/bridie-connell-the-10-funniest-things-i-have-ever-seen-on-the-internet)
- [ ] [Cutting up the dancefloor: the glorious, dubious history of the disco edit](https://www.theguardian.com/music/2022/jul/13/cutting-up-the-dancefloor-the-glorious-dubious-history-of-the-disco-edit)
- [X] [Are wine clubs worth joining? | Fiona Beckett on drinks](https://www.theguardian.com/food/2022/jul/15/are-wine-clubs-worth-joining-fiona-beckett)
- [X] [A recession is not inevitable in Australia – and certainly not necessary | Greg Jericho](https://www.theguardian.com/commentisfree/2022/jul/14/a-recession-is-not-inevitable-in-australia-and-certainly-not-necessary)
- [X] [Crisp white wines to whet the appetite](https://www.theguardian.com/food/2022/jul/22/crisp-white-wines-to-whet-the-appetite-aperitif-fiona-beckett)
[...]

Issue #4698, opened 2022-04-22
Fronts Ophan tag parity (`data-component`)


mxdvl:
## Remaining Front components mismatches

Comparing **`data-component`** on the international Front between [Frontend][] & [DCR][].

[Frontend]: https://www.theguardian.com/international?dcr=false
[DCR]: https://www.theguardian.com/international?dcr=true

### Missing from DCR (1 / 25)

- [X] **`palette-styles-new-do-not-delete`** &rarr; `<SECTION/>`

### Tag mismatch (3 / 25)

- [X] **`nav3`** : `<DIV />` &cross; should be &rarr; `<NAV/>`
- [X] **`nav2`** : `<DIV />` &cross; should be &rarr; `<NAV/>`
- [X] **`footer`** : `<DIV />` &cross; should be &rarr; `<FOOTER/>`

---

### Identical match (21 / 25)

- [X] **`headlines`** &rarr; `<SECTION/>`
- [X] **`ukraine-invasion`** &rarr; `<SECTION/>`
- [X] **`spotlight`** &rarr; `<SECTION/>`
- [X] **`opinion`** &rarr; `<SECTION/>`
- [X] **`sport`** &rarr; `<SECTION/>`
- [X] **`wordiply`** &rarr; `<SECTION/>`
- [X] **`around-the-world`** &rarr; `<SECTION/>`
- [X] **`climate-crisis`** &rarr; `<SECTION/>`
- [X] **`contact-the-guardian`** &rarr; `<SECTION/>`
- [X] **`culture`** &rarr; `<SECTION/>`
- [X] **`cotton-capital`** &rarr; `<SECTION/>`
- [X] **`lifestyle`** &rarr; `<SECTION/>`
- [X] **`documentaries`** &rarr; `<SECTION/>`
- [X] **`explore`** &rarr; `<SECTION/>`
- [X] **`take-part`** &rarr; `<SECTION/>`
- [X] **`today-in-focus`** &rarr; `<SECTION/>`
- [X] **`videos`** &rarr; `<SECTION/>`
- [X] **`video-playlist`** &rarr; `<DIV/>`
- [X] **`in-pictures`** &rarr; `<SECTION/>`
- [X] **`most-viewed`** &rarr; `<SECTION/>`
- [X] **`trending-topics`** &rarr; `<SECTION/>`

mxdvl:
See also #4662

mxdvl:
[...]

PR #5511, opened 2022-07-26
feat(a11y): list iframes without a title

mxdvl:
## What does this change?

Add the script for #5510 

## Why?

It allows us to track accessibility issues.

### DIFF
--- a/scripts/deno/iframe-titles.ts
+++ b/scripts/deno/iframe-titles.ts
@@ -1,0 +1,187 @@
+import { array, object, string } from 'https://cdn.skypack.dev/zod@3.17?dts';
+import {
+	DOMParser,
+	Element,
+	Node,
+} from 'https://deno.land/x/deno_dom@v0.1.22-alpha/deno-dom-wasm.ts';
+import { octokit } from './github.ts';
+
+const searchParams = new URLSearchParams({
+	'api-key': 'test',
+	'query-fields': 'body',
+	q: 'iframe',
+	'show-fields': 'body',
+	'page-size': String(100),
+});
+const url = new URL(
+	`search?${searchParams}`,
+	'http://content.guardianapis.com/',
+);
+
+const schema = object({
+	response: object({
+		status: string(),
+		results: array(
+			object({
+				id: string(),
+				webTitle: string(),
+				fields: object({
+					body: string(),
+				}),
+			}),
+		),
+	}),
+});
+
+const {
+	response: { results },
+} = await fetch(url.toString())
+	.then((r) => r.json())
+	.then(schema.parse);
+
+const parser = new DOMParser();
+
+const isElement = (n: Node): n is Element => n instanceof Element;
+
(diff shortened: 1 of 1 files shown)


bryophyta (approved):
+1

scripts/deno/iframe-titles.ts
```diff
@@ -0,0 +1,187 @@
...
+	Node,
+} from 'https://deno.land/x/deno_dom@v0.1.22-alpha/deno-dom-wasm.ts';
+import { octokit } from './github.ts';
+
+const searchParams = new URLSearchParams({
+	'api-key': 'test',
+	'query-fields': 'body',
+	q: 'iframe',
+	'show-fields': 'body',
+	'page-size': String(100),
```
AshCorr, line 14: Out of curiosity, whats the difference between this and just using a string literal `'100'`?

AshCorr (approved):

scripts/deno/iframe-titles.ts
```diff
@@ -0,0 +1,187 @@
+import { array, object, string } from 'https://cdn.skypack.dev/zod@3.17?dts';
+import {
+	DOMParser,
+	Element,
+	Node,
+} from 'https://deno.land/x/deno_dom@v0.1.22-alpha/deno-dom-wasm.ts';
+import { octokit } from './github.ts';
+
+const searchParams = new URLSearchParams({
+	'api-key': 'test',
```
AshCorr, line 10: Is this key public knowledge already?

mxdvl: Yes, it’s listed as an example in https://open-platform.theguardian.com/explore/

scripts/deno/iframe-titles.ts
mxdvl: I prefer the syntax highlighting that way.

georgeblahblah:
@guardian/newsletters this might be interesting for Newsletters, a similar script could be used to find articles using old-style newsletter embeds in CAPI if you ever want to convert them to the new style.

I was also thinking it could be used in the meantime to find embeds which are improperly formatted (e.g missing the `data-component` attribute)

scripts/deno/iframe-titles.ts
bryophyta: this doesn't apply to the current code, but a potential gotcha that I always find a little surprising is that `new String(100)` returns something different to `String(100)`:

```js
typeof String(100) === 'string'  // true
typeof new String(100) === 'string'  // false

typeof new String(100) === 'object'  // true
```

mxdvl:
> I was also thinking it could be used in the meantime to find embeds which are improperly formatted (e.g missing the `data-component` attribute)

We are doing exactly that for fronts in #4698, the code lives in [`scripts/deno/ophan-components.ts`](https://github.com/guardian/dotcom-rendering/blob/803644712f7a538861d68d4c336a7c83cdd8e4b9/scripts/deno/ophan-components.ts)

PR #5522, opened 2022-07-27
Eslint refactor: no-fallthough

OllysCoding:
Co-Authored-By: Max Duval <hi@mxdvl.com>

Disabling the ESLINT rules 'default-case' & 'no-fallthrough' as these do not respect / understand typescript - throwing errors when they were not expected.

To replace this rule we turned on `noFallthroughCasesInSwitch` in the tsconfig - ensuring that these rules are still enforced while respecting typescript types!


mxdvl (approved):
Yes, we paired on this and it’s a worthfile addition.

Just wanted to mentione the `.tsconfig` diff is a bit messy–but we probably can remove all the comments as VSCode now gives these throught IntelliSense.

oliverlloyd:
Yes !!!

PR #5524, opened 2022-07-27
ESLint: `restrict-template-expressions`

mxdvl:
## What does this change?

Enforce `restrict-template-expressions`.

Fixes all associated warnings, mostly by coercing `any` values to strings via `String(value)`

## Why?

This rule is part of @typescript-eslint’s recommended rules.


dotcom-rendering/src/amp/components/OnwardContainer.tsx
```diff
@@ -118,9 +118,8 @@ export const OnwardContainer = ({
 		<MoustacheTemplate>
 			<MoustacheSection name="showContent">
 				<div
-					css={`
-						${inner} js-has-click-event
-					`}
+					className="js-has-click-event"
+					css={inner}
```
mxdvl, line 122: I have no idea how this code ever worked!

dotcom-rendering/src/web/lib/getCommentContext.ts
```diff
@@ -101,9 +92,9 @@ export const getCommentContext = async (
 ): Promise<CommentContextType> => {
 	const url = joinUrl([ajaxUrl, 'comment', commentId.toString(), 'context']);
 	const filters = initFiltersFromLocalStorage();
-	const params = buildParams(filters);
+	const params = new URLSearchParams(buildParams(filters));
```
mxdvl, line 95: This is a refactor, but I'm confident it does the same thing and we polyfill `URLSearchParams`.

dotcom-rendering/src/web/lib/getZIndex.tsx
```diff
@@ -74,9 +74,9 @@ const indices = [
 // Implementation code - you don't need to change this to get a new index
 export type ZIndex = typeof indices[number];
 
-const decideIndex = (name: ZIndex): number | null => {
+const decideIndex = (name: ZIndex): number => {
 	const index = indices.indexOf(name);
-	if (index === -1) return null; // indexOf returns -1 if there is no match
+	if (index === -1) return -1; // indexOf returns -1 if there is no match
```
mxdvl, line 79: This is in line with the usage, returning a value that will put the element below any other one.

This should never happen anyway, as the input is typed.

dotcom-rendering/src/web/server/articleToHtml.tsx
```diff
@@ -206,24 +206,25 @@ export const articleToHtml = ({ data }: Props): string => {
...
+	fjs.parentNode.insertBefore(js, fjs);
+
+	t._e = [];
+	t.ready = function(f) {
+	t._e.push(f);
+	};
+
+	return t;
+}(document, "script", "twitter-wjs"));
+</script>`;
```
mxdvl, line 227: Should we make this script `async` and `defer`?

mxdvl (commented):
This touches a lot of @guardian/tx-engineers's code. Is there a reason you log errors to the console rather than using Sentry?

tjmw (approved):
Looks good from a TX perspective 👍 I've made a note to review `console.log` usage and see if reporting to sentry would be more appropriate.

PR #5529, opened 2022-07-27
Add fixed/medium/slow-vii container

### LINKED ISSUE #5137: "fixed/medium/slow-VII" (`125`) container

ioannakok:
Co-authored-by: Daniel Clifton <daniel.clifton@guardian.co.uk>

Closes #5137 

## What does this change?
Adds support for `fixed/medium/slow-vii` container.

## Why?
As part of the Fronts migration to DCR (see: https://github.com/guardian/dotcom-rendering/issues/4720)

## Screenshots

| Before (Frontend prod)     | After (DCR local) |
|-------------|------------|
| image | image |
| image | image |
| image | image |

### DIFF
--- a/dotcom-rendering/src/web/components/FixedMediumSlowVII.tsx
+++ b/dotcom-rendering/src/web/components/FixedMediumSlowVII.tsx
@@ -1,0 +1,105 @@
+import type { DCRContainerPalette } from '../../types/front';
+import type { TrailType } from '../../types/trails';
+import { LI } from './Card/components/LI';
+import { UL } from './Card/components/UL';
+import { FrontCard } from './FrontCard';
+
+type Props = {
+	trails: TrailType[];
+	containerPalette?: DCRContainerPalette;
+	showAge?: boolean;
+};
+
+export const FixedMediumSlowVII = ({
+	trails,
+	containerPalette,
+	showAge,
+}: Props) => {
+	const primary = trails[0];
+	const firstSlice = trails.slice(1, 3);
+	const secondSlice = trails.slice(3, 7);
+
+	return (
+		<>
+			<UL direction="row" padBottom={true}>
+				<LI
+					key={primary.url}
+					padSides={true}
+					showDivider={false}
+					padBottomOnMobile={true}
+					percentage="50%"
+				>
+					<FrontCard
+						trail={primary}
+						format={primary.format}
+						containerPalette={containerPalette}
+						showAge={showAge}
+						headlineSize="large"
+						imagePositionOnMobile="top"
+						imageSize="large"
+						supportingContent={primary.supportingContent}
+					/>
+				</LI>
+				{firstSlice.map((trail, index) => {
+					return (
+						<LI
+							key={trail.url}
+							padSides={true}
+							showDivider={true}
+							padBottomOnMobile={index !== 1}
+							percentage="25%"
+						>
+							<FrontCard
+								trail={trail}
+								format={trail.format}
+								containerPalette={containerPalette}
+								showAge={showAge}
+								headlineSize="medium"
+								imagePositionOnMobile="left"
+								imageSize="medium"
+								trailText={
+									trail.supportingContent &&
+									trail.supportingContent.length > 0
+										? undefined
+										: trail.trailText
+								}
+								supportingContent={
+									trail.supportingContent &&
+									trail.supportingContent.length > 2
+										? trail.supportingContent.slice(0, 2)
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/Card/components/LI.tsx
```diff
@@ -92,6 +98,7 @@ type Props = {
 	// Prevent the divider from spanning the LI's bottom padding. To be used when you know that the
 	// LI will have bottom padding, but won't have another card in the same container directly below it.
 	offsetBottomPaddingOnDivider?: boolean;
+	showOnMobile?: boolean;
```
ioannakok, line 101: Added this in case in DCR we want to keep the functionality where [some containers show less cards in breakpoints below tablet](https://www.figma.com/file/sx2vMFHbL7SsUo0LcpsKNe/%E2%AC%A3--Front-container?node-id=0%3A1). Will remove if not. Do we want to keep this? @HarryFischer

ioannakok: If we keep this, it will affect how we implement `Show More` functionality since in certain containers different cards will be shown in different breakpoints. My personal preference would be to not keep it as this will make things simpler.

ioannakok:
The last 3 Chromatic diffs seem false positives to me

dotcom-rendering/src/web/components/Card/components/LI.tsx
ioannakok: The requirement by digital editors is they want all containers, both fixed and dynamic, to keep the same count of cards on desktop vs mobile, so I am removing this.

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -423,7 +423,7 @@ export const Card = ({
 					</div>
 				</ContentWrapper>
 			</CardLayout>
-			{hasSublinks && noOfSublinks > 2 ? (
+			{hasSublinks && noOfSublinks >= 2 ? (
```
ioannakok, line 426: I made the changes in this file to achieve parity with frontend for the last screenshot in the PR description. Is there a requirement that `supportingContent` should look like this on mobile?

image

mxdvl: I don’t know, might be worth getting @HarryFischer’s input.

dotcom-rendering/src/web/components/FixedMediumSlowVII.tsx
```diff
@@ -0,0 +1,104 @@
...
+};
+
+export const FixedMediumSlowVII = ({
+	trails,
+	containerPalette,
+	showAge,
+}: Props) => {
+	const primary = trails[0];
+	const firstSlice = trails.slice(1, 3);
+	const secondSlice = trails.slice(3, 7);
```
mxdvl, line 19: Maybe we could keep things consistent?

```suggestion
	const primary = trails.slice(0,1);
	const firstSlice = trails.slice(1, 3);
	const secondSlice = trails.slice(3, 7);
```

mxdvl (approved):
This looks good.

It would be good to get a designer’s take on the supporting content on mobile.

dotcom-rendering/src/web/components/Card/Card.tsx
ioannakok: I have verified with @HarryFischer and we are keeping the change to make full width.

dotcom-rendering/src/web/components/FixedMediumSlowVII.tsx
ioannakok: Thanks for the suggestion! I kept this as `trails[0]` because I wanted primary to be a `TrailType` so that I can call directly `primary.format` etc. `trails.slice(0, 1)` gives me a `TrailType[]` with one element which I will then have to extract from the array or call a `map` on it below to access its properties. For this reason I would like to keep it as it is but let me know if I'm missing something and there is a simpler way of doing this!

dotcom-rendering/src/web/components/Card/Card.tsx
ioannakok: Just realised this change creates an unwanted diff: https://www.chromatic.com/test?appId=5dfcbf3012392c0020e7140b&id=62fd2da1519f2836f2cffc8a. I will revert this for this PR and will fix it in a following one with another bug regarding `SupportingContent` that I noticed.

PR #5533, opened 2022-07-28
Use `String.prototype.padStart` to pad minutes

georgeblahblah:
## What does this change?

- Uses `String.prototype.padStart` to pad minutes

## Why?

- A bit easier to read!

### DIFF
--- a/apps-rendering/src/date.ts
+++ b/apps-rendering/src/date.ts
@@ -125,24 +125,24 @@ const month = (date: Date): string => months[date.getUTCMonth()];
 
 const fullMonth = (date: Date): string => fullMonths[date.getUTCMonth()];
 
-const padZero = (n: number): string => String(n).padStart(2, '0');
+const padStart = (n: number): string => String(n).padStart(2, '0');
 
 const time = (date: Date, separator: string): string =>
-	`${padZero(date.getUTCHours())}${separator}${padZero(
+	`${padStart(date.getUTCHours())}${separator}${padStart(
 		date.getUTCMinutes(),
 	)}`;
 
 const time12hr = (date: Date, separator: string): string => {
 	const utcHours = date.getUTCHours();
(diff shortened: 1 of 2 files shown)


apps-rendering/src/date.ts
```diff
@@ -125,7 +125,7 @@ const month = (date: Date): string => months[date.getUTCMonth()];
 
 const fullMonth = (date: Date): string => fullMonths[date.getUTCMonth()];
 
-const padZero = (n: number): string => (n < 10 ? `0${n}` : n.toString());
+const padZero = (n: number): string => String(n).padStart(2, '0');
```
OllysCoding, line 128: Maybe a hot take but I prefer `n.toString().padStart(2, '0')` than `String(n).padStart(2, '0')`!

georgeblahblah: oh interesting! I don't think I have a strong preference for either, but your comment did send me down a rabbit hole to investigate the differences, so thank you for that!

What I learned:

- `n.toString()` will throw a `TypeError` if `n` is `undefined` - makes sense, not really a concern in this case. Whereas `String(n)` will always give you a `string` back, even if the input is `undefined` unless `n` is missing `toString` or `valueOf`
- `String(n)` will call `n.toString()` if it exists

I enjoyed reading this article about [Object to Primitive conversions](http://www.adequatelygood.com/Object-to-Primitive-Conversions-in-JavaScript.html)

mxdvl (commented):
What about replacing the use of `padZero` with `padStart`?

georgeblahblah:
> What about replacing the use of `padZero` with `padStart`?

Good suggestion @mxdvl!

mxdvl (approved):
LGTM.

Interestingly, this implementation does not handle 3-digit numbers. Thankfully these are unlikely to happen with hours or minutes.

PR #5536, opened 2022-07-28
Use an article that will expire on 31/12/2022 in paid.content cypress test 

### LINKED ISSUE #5531: Find a proper solution for when paid article content expires and paid cypress tests are failing
We have `paid.content` cypress test that tests we send some information to Google Analytics when a logo is clicked in a labs paid article and we' re testing against real articles. The problem is this content is not stable and gets removed quite regularly (here is the one of the previous examples we used: https://www.theguardian.com/you-could-be-here/2022/may/13/kos-crete-corfu-and-mykonos-a-guide-to-greeces-favourite-islands). Ask Commercial whether they can provide us with a permanent test.

ioannakok:
Closes #5531 

## What does this change?
The URL of the article used for `paid.content` cypress test. Paid content can be removed in articles used for this test and it starts failing. We have asked from Commercial to provide us with a permanent test page. Until then we're using the URL with the longest end date.

Also removing a comment that is no longer relevant. After this PR https://github.com/guardian/dotcom-rendering/pull/4187 we can do `cy.visit(url, { headers : { 'x-gu-geolocation': 'ip.dummytext,country:UK' } })` and the article edition will default to the value in the header. We don't need to do this with this new article we're using because this paid content has the same commercialProperties for all regions.

### DIFF
--- a/dotcom-rendering/cypress/e2e/parallel-6/paid.content.cy.js
+++ b/dotcom-rendering/cypress/e2e/parallel-6/paid.content.cy.js
@@ -4,17 +4,8 @@ import { cmpIframe } from '../../lib/cmpIframe';
 import { privacySettingsIframe } from '../../lib/privacySettingsIframe';
 import { storage } from '@guardian/libs';
 
-// It is important to use this article for this test because its commercialProperties
-// coming from the CAPI object are the same for all editions. This way we are making sure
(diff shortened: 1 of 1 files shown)


bryophyta (approved):
looks good!

PR #5543, opened 2022-07-29
Remove size mapping data attrs from ad slots

Jakeii:
## What does this change?
Remove size mapping data attributes from ad slots

## Why?
They are redundant now, size mappings are now loaded from [commercial-core](https://github.com/guardian/commercial-core/blob/main/src/ad-sizes.ts#L143) as of this frontend PR https://github.com/guardian/frontend/pull/25124


arelra (approved):
A great sight to see - lots of deleted code 🎉

PR #5544, opened 2022-07-29
Revert: Fix contrasts for branding on liveblogs mobile #5526 

bryophyta:
## What does this change?

Reverts the changes from #5526

[Diff between this branch and pre-merge `main`](https://github.com/guardian/dotcom-rendering/compare/4416ad15772a1cf24d758a9a3636ec2cc2d92872...pf/revert-liveblog-header) shows no differences in the relevant files.

## Why?

The CI was run on an older version of a cypress fixture. With the new fixture it didn't pass on TeamCity, thereby preventing deployment of `main`.

There's at least on genuine bug revealed by the new fixture which wasn't accounted for in the original issue/PR, so I'm going to reopen the issue and revert the changes to unblock Prod.

PR #5526, opened 2022-07-27
Fix contrasts for branding on liveblogs mobile

bryophyta:
## What does this change?
The background colour for the meta on liveblogs is dark on mobile, light
on desktop.

This commit uses light colours and the light variant of the logo on
smaller breakpoints. (Largely following the pattern already used in
`Contributor.tsx`.)

## Why?
Closes #5486 

## Screenshots

https://www.theguardian.com/football/live/2022/jul/26/england-v-sweden-womens-euro-2022-semi-final-live?dcr&live=true

| Before      | After      |
|-------------|------------|
| image | image |

[before]: https://example.com/before.png
[after]: https://example.com/after.png

PR #5546, opened 2022-07-29
Fix: contrast logo on branded liveblogs

bryophyta:
## What does this change?

This re-implements #5526, and adds some logic to check whether the article being rendered has a liveblog format, before deciding whether branded logos in the `meta` header should change between smaller and larger screens. (The background of the `meta` section changes between breakpoints on liveblogs, whereas it's constant on other branded articles.)

## Why?

This fixes a bug which led to the previous PR being reverted in #5544. The underlying bug was that the light version of a branded logo was being used for smaller screens on branded articles of all formats, when this should have been limited to liveblog formats.

## Screenshots

(The first set of screenshots should differ, and they do; the second set shouldn't, and they don't.)

https://www.theguardian.com/football/live/2022/jul/26/england-v-sweden-womens-euro-2022-semi-final-live?dcr&live=true

https://www.theguardian.com/a-vision-for-better-food/2022/jul/22/a-kitchen-in-a-quarry-why-charlie-bighams-food-campus-was-named-a-riba-building-of-the-year

| Before      | After      |
|-------------|------------|
| image | image |
|image|image|

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleMeta.tsx
+++ b/dotcom-rendering/src/web/components/ArticleMeta.tsx
@@ -327,7 +327,11 @@ export const ArticleMeta = ({
 			<div css={meta(format)}>
 				{branding && (
 					<Island deferUntil="visible">
-						<Branding branding={branding} palette={palette} />
+						<Branding
+							branding={branding}
+							palette={palette}
+							format={format}
+						/>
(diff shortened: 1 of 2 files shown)


ioannakok (approved):
Thanks for the explanation!

PR #5550, opened 2022-07-31
`onwardsType`

oliverlloyd:
## What does this change?
This PR replaces the property `ophanComponentName` with `onwardsType`.

## Why?
We still use the 'type' of onwards as the string sent to Ophan but this type is also useful in other contexts and makes more sense as a separate property.

For example, we can now remove the `isCuratedContent` prop and replace it with `const isCurated = onwardsType === 'curated-content'`

In addition, we've now renamed the typescript type `OphanComponentName` to `OnwardsType`. We use `ophanComponentName` as a property in other places but in those cases we type it as a string so the name here did not make sense in this new context.

```ts
type OnwardsType =
	| 'series'
	| 'more-on-this-story'
	| 'related-stories'
	| 'related-content'
	| 'more-media-in-section'
	| 'more-galleries'
	| 'curated-content'
	| 'default-onwards'; 
```

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -790,12 +790,12 @@ type CAPIOnwardsType = {
 	trails: CAPITrailType[];
 	description?: string;
 	url?: string;
-	ophanComponentName: OphanComponentName;
+	onwardsType: OnwardsType;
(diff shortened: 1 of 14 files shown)


mxdvl (approved):
Excellent!

PR #5551, opened 2022-07-31
A more flexible `onwards`

oliverlloyd:
Depends on
https://github.com/guardian/dotcom-rendering/pull/5550
&
https://github.com/guardian/dotcom-rendering/issues/5566

Closes https://github.com/guardian/dotcom-rendering/issues/5571

## What does this change?
This PR adds a new property to the article model, `onwards: CAPIOnwardsType[];`

```ts
type CAPIOnwardsType = {
	heading: string;
	trails: CAPITrailType[];
	description?: string;
	url?: string;
	onwardsType: OnwardsType;
	format: CAPIFormat;
};
```

It also adds a new abstraction `Onwards` which accepts this new property and decides how to render the content.

Frontend can now send an array of different styles of onwards data and DCR will dynamically render it based on what's sent. There are two possible outcomes:

### Server render the trails
If `trails` is a non zero length array of trail objects then `Onwards` will server render these items and then hydrate the `Carousel` on the client

### Fetch the onwards data
If the `trails` array contains no data and a `url` property is provided, then `Onwards` will setup the browser to make an api call to fetch the data and render it from there

## Dependencies
For this PR to work, we need Frontend to make the decisions about what onwards data should be shown or fetched. Currently this logic exists in `OnwardsUpper` but there is an intention to move this to the server. The work to move this logic to the server has already started [with this PR](https://github.com/guardian/dotcom-rendering/pull/5473) and also [this one](https://github.com/guardian/frontend/pull/25267); other, similar work is in play.

**This work needs to complete before this PR can be merged.**


oliverlloyd:
Closed in favour of https://github.com/guardian/dotcom-rendering/pull/5578 because that can be merged

PR #5557, opened 2022-08-01
`Treats`

### LINKED ISSUE #5493: `Treats`
Also known as `Nuggets` and only on desktop, these links are provided in the data model

Image

Is the crossword link also part of the treat?

Image

oliverlloyd:
## What does this change?
Closes #5493 by adding support for `Treats`, those little link thingies in the bottom left of some containers

Screenshot 2022-08-01 at 12 05 55

## Screenshots

| Before      | After      |
|-------------|------------|
| Screenshot 2022-08-01 at 12 00 32 | Screenshot 2022-08-01 at 11 56 59 |

### Also
I spotted a small issue where the trail text was showing for the second card  on `FixedMediumSlowVI` when it shouldn't, so I added a drive by for that. Here is a screenshot for how the container looked before this fix

Screenshot 2022-08-01 at 12 01 33

### DIFF
--- a/dotcom-rendering/src/model/enhanceTreats.ts
+++ b/dotcom-rendering/src/model/enhanceTreats.ts
@@ -1,9 +1,7 @@
 import type { FEFrontCard, TreatType } from '../types/front';
 
 export const enhanceTreats = (treats: FEFrontCard[]): TreatType[] =>
-	treats.map((treat) => {
-		return {
-			text: treat.header.headline,
-			linkTo: treat.properties.href ?? treat.header.url,
-		};
-	});
+	treats.map((treat) => ({
+		text: treat.header.headline,
+		linkTo: treat.properties.href ?? treat.header.url,
+	}));
--- a/dotcom-rendering/src/model/enhanceTreats.ts
+++ b/dotcom-rendering/src/model/enhanceTreats.ts
@@ -1,0 +1,9 @@
+import type { FEFrontCard, TreatType } from '../types/front';
+
+export const enhanceTreats = (treats: FEFrontCard[]): TreatType[] =>
+	treats.map((treat) => {
+		return {
+			text: treat.header.headline,
+			linkTo: treat.properties.href ?? treat.header.url,
+		};
+	});
--- a/dotcom-rendering/src/web/components/ContainerLayout.tsx
+++ b/dotcom-rendering/src/web/components/ContainerLayout.tsx
@@ -2,7 +2,7 @@ import { css } from '@emotion/react';
 import type { ArticleFormat } from '@guardian/libs';
 import { ArticleDesign } from '@guardian/libs';
 import { from, space, until } from '@guardian/source-foundations';
-import type { DCRContainerPalette } from '../../types/front';
+import type { DCRContainerPalette, TreatType } from '../../types/front';
 import { decideContainerOverrides } from '../lib/decideContainerOverrides';
 import { hiddenStyles } from '../lib/hiddenStyles';
 import { ContainerTitle } from './ContainerTitle';
@@ -11,6 +11,7 @@ import { Flex } from './Flex';
 import { Hide } from './Hide';
 import { LeftColumn } from './LeftColumn';
 import { ShowHideButton } from './ShowHideButton';
+import { Treats } from './Treats';
 
 type Props = {
 	title?: string;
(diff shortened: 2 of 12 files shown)


HarryFischer:
Looks good - can you send me a screenshot of leftCol breakpoint please @oliverlloyd . Just want to check the line-height when the text wraps

dotcom-rendering/src/model/enhanceTreats.ts
```diff
@@ -0,0 +1,9 @@
+import type { FEFrontCard, TreatType } from '../types/front';
+
+export const enhanceTreats = (treats: FEFrontCard[]): TreatType[] =>
+	treats.map((treat) => {
+		return {
+			text: treat.header.headline,
+			linkTo: treat.properties.href ?? treat.header.url,
+		};
+	});
```
mxdvl, line 9: Might as well return directly…

```suggestion
	treats.map((treat) => ({
		text: treat.header.headline,
		linkTo: treat.properties.href ?? treat.header.url,
	}));
```

mxdvl (approved):
What a treat!

oliverlloyd:
> Looks good - can you send me a screenshot of leftCol breakpoint please @oliverlloyd . Just want to check the line-height when the text wraps

Screenshot 2022-08-01 at 12 58 53

Here you go, as discussed offline, it looks good. We're using Source components here so we get these features off the shelf

dotcom-rendering/src/model/enhanceTreats.ts
```diff
@@ -0,0 +1,7 @@
+import type { FEFrontCard, TreatType } from '../types/front';
+
+export const enhanceTreats = (treats: FEFrontCard[]): TreatType[] =>
```
jamesgorrie, line 3: nit: most other enhance methods return a `DCRType`. Should we stick to that convention?

dotcom-rendering/src/web/components/ContainerLayout.tsx
```diff
@@ -38,6 +39,7 @@ type Props = {
 	innerBackgroundColour?: string;
 	showDateHeader?: boolean;
 	editionId?: EditionId;
+	treats?: TreatType[];
```
jamesgorrie, line 42: Is this ever optional as [the parent type](https://github.com/guardian/dotcom-rendering/pull/5557/files#diff-06ec888dd5f5b04e928175cbb9dbe5377633973349b1fc48b069b5fd49d66dbaR21) is required?

dotcom-rendering/src/web/components/ContainerLayout.tsx
```diff
@@ -164,19 +167,39 @@ export const ContainerLayout = ({
...
+								}
+								description={description}
+								url={url}
+								containerPalette={containerPalette}
+								showDateHeader={showDateHeader}
+								editionId={editionId}
+							/>
+							{leftContent}
+						</div>
+						{treats && (
```
jamesgorrie, line 194: Worth adding a guard against `treat.length > 0` to not add an empty list?
(This might be tied into me not understanding the `undefined` case of `treats` in the model)

jamesgorrie (approved):
Couple of nit picks, 😋

dotcom-rendering/src/model/enhanceTreats.ts
oliverlloyd: There's no frontend version of a treat. It's just a card (FECardType) so I figured there was no need for the DCR prefix. We also use this type outside of 'fronts' so that felt like another reason to keep it simple

dotcom-rendering/src/web/components/ContainerLayout.tsx
oliverlloyd: `ContainerLayout` is used in lots of places outside of the fronts flow and in those cases - at this time - no treats are given

dotcom-rendering/src/web/components/ContainerLayout.tsx
oliverlloyd: I added a check to the Treats component itself

bryophyta (approved):
This looks good to me 👍

dotcom-rendering/src/model/enhanceTreats.ts
jamesgorrie: I like the simplicity.

Do you think it's ever worth using the `DCR` prefix?

oliverlloyd: I think the prefixes make sense when you have two types that refer to the same thing, the prefix lets you know which is the FE version and which is the DCR version.

PR #5559, opened 2022-08-01
Key Events Carousel: underline links on hover

silvacb:
## What does this change?
Adds a text-decoration property to `textStyles`.

## Why?
This is a design decision for Key Events Carousel links to be underlined on hover.

## Screenshots

Example1:
Screenshot 2022-08-01 at 15 23 30

Example2:
Screenshot 2022-08-01 at 15 32 36

### DIFF
--- a/dotcom-rendering/src/web/components/KeyEventCard.tsx
+++ b/dotcom-rendering/src/web/components/KeyEventCard.tsx
@@ -42,12 +42,10 @@ const linkStyles = (palette: Palette) => css`
 		span {
 			border-bottom: 1px solid ${palette.hover.keyEventLink};
 		}
-	}
-
-	&:hover > div {
-		text-decoration: underline;
-		text-underline-offset: 3px;
-		text-decoration-color: ${palette.text.keyEvent};
+		div {
+			text-decoration: underline ${palette.text.keyEvent};
+			text-underline-offset: 3px;
+		}
 	}
 `;
 
--- a/dotcom-rendering/src/web/components/KeyEventCard.tsx
+++ b/dotcom-rendering/src/web/components/KeyEventCard.tsx
@@ -94,6 +94,7 @@ const textStyles = (palette: Palette) => css`
 
 	&:hover {
 		text-decoration: underline;
+		text-underline-offset: 3px;
 		text-decoration-color: ${palette.text.keyEvent};
 	}
 `;
--- a/dotcom-rendering/src/web/components/KeyEventCard.tsx
+++ b/dotcom-rendering/src/web/components/KeyEventCard.tsx
@@ -91,6 +91,11 @@ const listItemStyles = (palette: Palette) => css`
 const textStyles = (palette: Palette) => css`
 	${textSans.small({ fontWeight: 'regular', lineHeight: 'regular' })};
 	color: ${palette.text.keyEvent};
+
+	&:hover {
+		text-decoration: underline;
+		text-decoration-color: ${palette.text.keyEvent};
+	}
 `;
 
 const timeStyles = (palette: Palette) => css`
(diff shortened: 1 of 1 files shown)


HarryFischer:
Looks good 👍

dotcom-rendering/src/web/components/KeyEventCard.tsx
```diff
@@ -91,6 +91,12 @@ const listItemStyles = (palette: Palette) => css`
 const textStyles = (palette: Palette) => css`
 	${textSans.small({ fontWeight: 'regular', lineHeight: 'regular' })};
 	color: ${palette.text.keyEvent};
+
+	&:hover {
```
alinaboghiu, line 95: this is very cool! just to be pedantic, I think we should underline when we hover over any part of the link (e.g. when we hover over the timestamp or the dot)

alinaboghiu: but just to be clear, I don't think we should underline the timestamp, just that the text should become underlined when we hover over the timestamp - I think that's the correct behaviour?

silvacb: @alinaboghiu gotcha, I have changed this now to include hover over the whole card. Please let me know what you think.

dotcom-rendering/src/web/components/KeyEventCard.tsx
```diff
@@ -43,6 +43,12 @@ const linkStyles = (palette: Palette) => css`
 			border-bottom: 1px solid ${palette.hover.keyEventLink};
 		}
 	}
+
+	&:hover > div {
```
abeddow91, line 47: We can move this into the `&:hover` selector above, how the span is being selected, to make the code a little cleaner

dotcom-rendering/src/web/components/KeyEventCard.tsx
```diff
@@ -43,6 +43,12 @@ const linkStyles = (palette: Palette) => css`
...
+		text-decoration: underline;
+		text-underline-offset: 3px;
+		text-decoration-color: ${palette.text.keyEvent};
```
abeddow91, line 50: we can consolidate this line into line 48 by using the `text-decoration: underline ${palette.text.keyEvent}`; shorthand which would allow us to remove this extra property.

abeddow91 (commented):
Looks great, a couple of small refactors to tidy it up but otherwise looks great!

dotcom-rendering/src/web/components/KeyEventCard.tsx
abeddow91: Yeah it sure is, a little extra reading here if you're interested https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration

paperboyo:
This will make them greatly more usable, 👏! Any reason screenshots do not look like using `text-decoration-skip-ink`? (I thought this is now on by default in all browsers…🤔)

HarryFischer:
> This will make them greatly more usable, 👏! Any reason screenshots do not look like using `text-decoration-skip-ink`? (I thought this is now on by default in all browsers…🤔)

I think it's because I asked for `text-underline-offset: 3px;` so it now doesn't touch ink? @paperboyo

paperboyo:
I don’t know CSS for squat, but it does very much touch ink in the screenshot, that’s why I asked…

silvacb:
@paperboyo can you take a screenshot of those screenshots? :) to me it looks like the underline doesn't seem to be touching the ink, but maybe I'm mistaken?

paperboyo:
@silvacb Ha! I just zoomed in and indeed! Must be my eyesight (they are close).

alinaboghiu (approved):

abeddow91 (commented):
Nice work! 🚀

PR #5563, opened 2022-08-01
Add an edition path parameter when serving locally

georgeblahblah:
## What does this change?

- Added an optional `edition` path parameter for the local server GET routes, and include it in the mock `RenderingRequest` model

### DIFF
--- a/apps-rendering/src/server/server.ts
+++ b/apps-rendering/src/server/server.ts
@@ -2,6 +2,7 @@
 
 import 'source-map-support/register'; // activating the source map support
 import path from 'path';
+import { Edition } from '@guardian/apps-rendering-api-models/edition';
 import type { RelatedContent } from '@guardian/apps-rendering-api-models/relatedContent';
 import type { RenderingRequest } from '@guardian/apps-rendering-api-models/renderingRequest';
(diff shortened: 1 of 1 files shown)


apps-rendering/src/server/server.ts
```diff
@@ -311,6 +312,21 @@ async function serveEditionsArticlePost(
 	}
 }
 
+function getEdition(maybeEdition: string): Edition {
```
georgeblahblah, line 315: I'm interested to discover if there's a better way of doing this! It feels verbose (which I like in this case), but maybe it's not the right style.

I also struggled with naming this function & parameter

JamieB-gu: I think this is the correct way of doing this 👍

Possible alternative naming:

```ts
function editionFromString(edition: string): Edition {}
function editionFromString(editionString: string): Edition {}
function editionFromString(str: string): Edition {}
function editionFromString(s: string): Edition {}
```

georgeblahblah: Thanks! I went with `editionFromString(editionString: string)` as well as combining the 'uk' case and `default`

JamieB-gu (approved):

PR #5572, opened 2022-08-02
Override commercial bundle URL with an environment variable

chrislomaxjones:
## What does this change?

Add the ability to override the commercial bundle URL passed in the DCR data model with an environment variable - `COMMERCIAL_BUNDLE_URL`.

Currently this won't be actually used anywhere, but it'll be possible to invoke it via:

```bash
COMMERCIAL_BUNDLE_URL=http://localhost:3031/graun.standalone.commercial.js make dev
```

This takes inspiration from the way [support-dotcom-components are developed locally against DCR](https://github.com/guardian/dotcom-rendering/blob/41ab963f48cbd61507c31f6bb614b2ae38329d60/dotcom-rendering/src/web/lib/contributions.ts#L245-L247).

## Why?

This is a step towards becoming more flexible when running our commercial code against various platforms. By allowing use of this environment variable, we can point DCR to use a commercial bundle being served locally, rather than via a local or prod Frontend.

The benefits include:
- Running our commercial code just against DCR and not Frontend. As more content is served by DCR, this will become increasingly useful.
- Running End-to-end tests in CI. By spinning up a production build of DCR and our commercial code, we could run our end-to-end tests in a Github action.

### DIFF
--- a/dotcom-rendering/src/web/server/articleToHtml.tsx
+++ b/dotcom-rendering/src/web/server/articleToHtml.tsx
@@ -135,7 +135,9 @@ export const articleToHtml = ({ data }: Props): string => {
 			...getScriptArrayFromFile('bootCmp.js'),
 			...getScriptArrayFromFile('ophan.js'),
 			CAPIArticle.config && {
-				src: CAPIArticle.config.commercialBundleUrl,
+				src:
+					process.env.COMMERCIAL_BUNDLE_URL ??
+					CAPIArticle.config.commercialBundleUrl,
 			},
 			...getScriptArrayFromFile('sentryLoader.js'),
 			...getScriptArrayFromFile('dynamicImport.js'),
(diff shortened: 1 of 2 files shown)


dotcom-rendering/src/lib/commercial-bundle.ts
```diff
@@ -0,0 +1,3 @@
+export const decideCommercialBundleUrl = (
+	commercialBundleUrl: string,
+): string => process.env.COMMERCIAL_BUNDLE_URL ?? commercialBundleUrl;
```
oliverlloyd, line 3: Bit of a nitpick but could this code just be inlined?

It's true that this would create duplication but given this is just one line of code maybe the benefit of having the code be more readable (I don't need to open a new file to understand what's happening) is worth the price? I really don't have a problem with this level of duplication.

oliverlloyd (approved):

arelra (approved):
A step closer to having our e2e tests running in a CI pipeline 🎉

Jakeii (approved):

dotcom-rendering/src/lib/commercial-bundle.ts
chrislomaxjones: I think that's a fair point, will make the change. Thanks!

ashishpuliyel:
Amazing. Super excited about this one

Issue #5571, opened 2022-08-02
A more flexible onwards


oliverlloyd:
Accept the `onwards` property from Frontend and render onwards content based on this

```ts
type CAPIOnwardsType = {
	heading: string;
	trails: CAPITrailType[];
	description?: string;
	url?: string;
	onwardsType: OnwardsType;
	format: CAPIFormat;
	isCuratedContent?: boolean;
};

onwards: CAPIOnwardsType[];
```

Issue #5566, opened 2022-08-02
Frontend decides what onwards content to show


oliverlloyd:
The logic for which onwards slices to show is currently contained in [OnwardsUpper](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/OnwardsUpper.importable.tsx). We want to move this to Frontend, simplifying the task of DCR to that of a renderer.

That logic looks complex but it is likely that it can be greatly simplified. The essential flow is

** For the top row (the first item in the array): **
Do I have a story package, if yes, render `more-on-this-story`
Do I have a series tag, if yes, render `series`
Do I have a popular tag, if yes, render `popular-in-tag`
Else, render `related-stories`

** For the bottom row (the second item in the array): **
Render `curated-content` - always

### The general flow
When as article request is made the data is fetched from CAPI and Frontend runs some transformations. At a certain point in this process we need to insert a new property `onwards` with the type:

```ts
type CAPIOnwardsType = {
	heading: string;
	trails: CAPITrailType[];
	description?: string;
	url?: string;
	onwardsType: OnwardsType;
	format: CAPIFormat;
	isCuratedContent?: boolean;
};

onwards: CAPIOnwardsType[];
```

DCR will use [this PR](https://github.com/guardian/dotcom-rendering/pull/5551) to read and render that property.

## Why?
This is part of a larger effort to move the rendering of `Cards` to the server. As part of this work, the various different onwards types are going to be rendered in different places

### Now
`more-on-this-story` - This is already server side rendered with the data provided by CAPI
`popular-in-tag` - This is currently [fetched as json](https://github.com/guardian/dotcom-rendering/issues/5568#issuecomment-1215160054) by the client and rendered on the client
`series` - This is currently [fetched as json](https://github.com/guardian/dotcom-rendering/issues/5569#issuecomment-1215215257) by the client and rendered on the client
`related-stories` - This is currently fetched as json by the client and rendered on the client
`curated-content` - This is currently [fetched as json](https://github.com/guardian/dotcom-rendering/issues/5567#issuecomment-1215260498) by the client and rendered on the client

### Target
`more-on-this-story` - No change
`popular-in-tag` - This will be read from a cache on Frontend, inserted into the DCR model and then server side rendered by DCR
`series` - This will be read from a cache on Frontend, inserted into the DCR model and then server side rendered by DCR
`related-stories` - This will still be fetched from the client but as html, not json
[...]

PR #5578, opened 2022-08-02
A more flexible `onwards`

oliverlloyd:
## What does this change?
Closes https://github.com/guardian/dotcom-rendering/issues/5571

This PR adds a new property to the article model, `onwards: CAPIOnwardsType[];`

```ts
type CAPIOnwardsType = {
	heading: string;
	trails: CAPITrailType[];
	description?: string;
	url?: string;
	onwardsType: OnwardsType;
	format: CAPIFormat;
};
```

It also adds a new abstraction `Onwards` which accepts this new property and decides how to render the content.

Frontend can now send an array of different styles of onwards data and, if it is present, DCR will dynamically render it based on what's sent. If it isn't sent DCR will default to the old (current) method for onwards content.

Where this new array of onwards data is provided, there are two possible outcomes for each item in the array:

### Server render the trails
If `trails` is a non zero length array of trail objects then `Onwards` will server render these items and then hydrate the `Carousel` on the client

### Fetch the onwards data
If the `trails` array contains no data and a `url` property is provided, then `Onwards` will setup the browser to make an api call to fetch the data and render it from there

### When will the old method be removed?
Once [this ticket](https://github.com/guardian/dotcom-rendering/issues/5566) is complete we will be able to remove the old `OnwardsUpper` component and associated logic and all onwards data can be sent via this new property.

### DIFF
--- a/dotcom-rendering/src/model/article-schema.json
+++ b/dotcom-rendering/src/model/article-schema.json
@@ -515,7 +515,6 @@
         "main",
         "mainMediaElements",
         "nav",
-        "onwards",
         "openGraphData",
         "pageFooter",
         "pageId",
--- a/dotcom-rendering/src/model/article-schema.json
+++ b/dotcom-rendering/src/model/article-schema.json
@@ -339,6 +339,61 @@
                 "trails"
             ]
         },
+        "onwards": {
+            "type": "array",
+            "items": {
+                "description": "Onwards",
+                "type": "object",
+                "properties": {
+                    "heading": {
+                        "type": "string"
+                    },
+                    "trails": {
+                        "type": "array",
+                        "items": {
+                            "$ref": "#/definitions/CAPITrailType"
+                        }
+                    },
+                    "description": {
+                        "type": "string"
+                    },
+                    "url": {
+                        "type": "string"
+                    },
+                    "onwardsType": {
+                        "$ref": "#/definitions/OnwardsType"
+                    },
+                    "format": {
+                        "type": "object",
+                        "properties": {
+                            "design": {
+                                "$ref": "#/definitions/CAPIDesign"
+                            },
+                            "theme": {
+                                "$ref": "#/definitions/CAPITheme"
+                            },
+                            "display": {
+                                "$ref": "#/definitions/CAPIDisplay"
+                            }
+                        },
+                        "required": [
+                            "design",
+                            "display",
+                            "theme"
+                        ]
+                    },
+                    "isCuratedContent": {
+                        "type": "boolean"
+                    }
+                },
+                "required": [
+                    "format",
+                    "heading",
+                    "onwardsType",
+                    "trails"
+                ]
+            }
+        },
         "beaconURL": {
             "type": "string"
         },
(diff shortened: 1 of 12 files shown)


dotcom-rendering/src/model/article-schema.json
```diff
@@ -4168,6 +4223,19 @@
                 "sponsorName"
             ]
         },
+        "OnwardsType": {
+            "enum": [
+                "curated-content",
```
jlieb10, line 4228: Are these arbitrary strings or are they matched elsewhere/in CAPI? Curious to know how easy it is to add to this as necessary.

dotcom-rendering/src/web/components/DecideOnwards.tsx
```diff
@@ -0,0 +1,53 @@
...
+				);
+			}
+
+			if (url) {
+				return (
+					<ElementContainer key={onwardsType}>
+						<Island
+							clientOnly={true}
+							deferUntil="visible"
+							placeholderHeight={600}
```
jlieb10, line 37: How did we decide on the placeholder height?

dotcom-rendering/src/web/layouts/CommentLayout.tsx
```diff
@@ -698,48 +699,59 @@ export const CommentLayout = ({ CAPIArticle, NAV, format }: Props) => {
...
-								trails={CAPIArticle.storyPackage.trails.map(
-									decideTrail,
-								)}
-								onwardsType="more-on-this-story"
+				{CAPIArticle.onwards ? (
+					<DecideOnwards
+						onwards={CAPIArticle.onwards}
+						format={format}
+					/>
+				) : (
```
jlieb10, line 707: Would it be cleaner to pass the CAPIArticle to <DecideOnwards> and do all of the possible onwards rendering in that component? It seems like we have onwards rendering logic both places now.

jlieb10: (Applies to all layouts)

dotcom-rendering/src/model/article-schema.json
oliverlloyd: This type was [added in this PR](https://github.com/guardian/dotcom-rendering/pull/5550). It's arbitrary in the sense that you can add anything to it but the type constrains what can be sent by Frontend

dotcom-rendering/src/web/layouts/CommentLayout.tsx
oliverlloyd: There's a rule in DCR to never pass `CAPIArticle` as a prop. We have this because this object is many thousands of lines long and can reach upwards of 30Kbs in size. It's preferable to keep props as focussed and explicit as possible to make the work React has to do easier and to keep the components easy to reason about.

dotcom-rendering/src/web/components/DecideOnwards.tsx
oliverlloyd: This is the existing value (not added here) but it's essentially a best guess based on how cards render. Longer or shorter headlines will change the actual height

PR #5581, opened 2022-08-02
Reduce Percy snapshot breakpoints to stay under monthly limit

arelra:
## What does this change?

We are currently taking snapshots for every breakpoint and browser (Chrome, Firefox, Edge & Safari).

This change reduces snapshot breakpoints to just one breakpoint.

I've also configured the dotcom-rendering project in percy.io to only snapshot for Chrome.

## Why?

Screenshot 2022-08-02 at 15 24 49

We have exceeded our monthly limit of snapshots under our current plan.

This is currently blocking commercial from running visual tests as our plan falls under the same organisation.

Until we are using Percy for real we can limit the snapshots without any impact.

I will follow up with Browserstack about our current plan.


bryophyta (approved):
this looks good to me!

PR #5582, opened 2022-08-02
Refactor `extract-nav`

mxdvl:
## What does this change?

Stop parsing the data in `extract-nav`, and instead introduce type signatures on the expected data.

Consolidate Nav-related types in `extract-nav.ts` and update type imports accordingly.

Decouple `SimpleLinkType` and `CAPILinkType` for data pre- and post-transform.

## Why?

We are already parsing/validating the data upstream, so let’s trust the compiler.

A partial clean-up was done in #663 after we moved to v2 of AJV.

Reduce the surface are for bugs. Remove a few `any` and the useless `ObjectType`.

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -4,6 +4,7 @@
 
 type DCRSnapType = import('./src/types/front').DCRSnapType;
 type DCRSupportingContent = import('./src/types/front').DCRSupportingContent;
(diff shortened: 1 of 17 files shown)


bryophyta (approved):
tbh I haven't fully gone through the original extract-nav function in detail, but assuming what you've said about its role then this looks good to me!

If you're wanting any more informed feedback on this part then I'm happy to spend the time to understand the original code; just lmk?

Issue #5554, opened 2022-08-01
Deeply Read Experiment


mxdvl:
# Deeply Read Experiment

We want to run an experiment where users would be shown some of the most deeply read articles on the Guardian, sorted by edition.

## Where do we want to run this experiment?
- [ ] Network fronts
- [ ] Tag fronts?
- [ ] Articles?

## Creating the Agents
The agents need to handle all editions.

- [x] https://github.com/guardian/dotcom-rendering/issues/5664
- [x] https://github.com/guardian/dotcom-rendering/issues/5662
- [x] https://github.com/guardian/dotcom-rendering/issues/5660
- [ ] https://github.com/guardian/frontend/pull/25358

## Setting the test
- [ ] Create a server-side test for 5% of users
- [ ] Measure impact with D&I

## Drive-by Improvements
- [ ] Server-render “Most Viewed” on DCR
- [ ] Server-render “Across the Guardian” on Frontend + DCR

## What data is powering what?

![onwards-mostpopular](https://user-images.githubusercontent.com/31692/184098480-efa7383b-e9d7-4541-982c-9bdee38fb5ec.png)

---

## Previous 🎨 

- [Deeply Read test doc - June 2022](https://docs.google.com/document/d/1dutAGHapw7Y-jZc6jMv_64OdDLxm6tu6SZVuF989CHc/edit)

- ['Deeply Read' API endpoint in Ophan](https://docs.google.com/document/d/1FSfB47-bZ2CFvbATzDqFst6j-6JNCkgsSPKQKFnQRhU/edit)

Original implementation & test in late 2020 / early 2021

- https://github.com/guardian/frontend/pull/23364
- https://github.com/guardian/frontend/pull/23365
- https://github.com/guardian/frontend/pull/23366
- https://github.com/guardian/frontend/pull/23367
- https://github.com/guardian/frontend/pull/23368
- https://github.com/guardian/frontend/pull/23369
- https://github.com/guardian/frontend/pull/23254
- https://github.com/guardian/frontend/pull/23376
- https://github.com/guardian/frontend/pull/23377
- https://github.com/guardian/frontend/pull/23379

- https://github.com/guardian/frontend/pull/23665

mxdvl:
See test cleanup PR [frontend#23665](https://github.com/guardian/frontend/pull/23665)

VDuczekW:
Web X will not work on this - the ticket link has been passed on to EDX

PR #5585, opened 2022-08-03
feat: add MostPopularContainer and use instead of MostViewed on fronts

jamesgorrie:
## What does this change?

This removes the `MostViewed` component from the fronts, as it currently doesn't work as expected, and replaces it with a `MostPopular` component. The naming is to reflect the container name from fronts (`news/most-popular`).

This new container renders the content from the fronts tool. 

## Expected from DCR

- Has backfill content from a CAPI query put in the config of the container in fronts tool config editor 
- Has editionalised `most-viewed`

fronts-editor-capi-backfill

most-viewed-container

### Current from DCR
- Has editionalised `most-viewed`
Screenshot 2022-08-03 at 10 43 26

A few things...

### We loose "Across the guardian" for now

There is some work in #5554 to make the [MostPopularAgent](https://github.com/guardian/frontend/blob/main/onward/app/feed/MostPopularAgent.scala) available to fronts, which means that we could get this data on the server

### We loose "Most commented" and "Most shared" for now

I'm not there is a solution for this right now... but it forces us to think about it

### What about deeply read

This should set us up to be able to add another data source, especially as static as `DeeplyRead` relatively easily, and make it accessible via the server. This would simplify the implementation of the test significantly as we wouldn't have to work with server+client rendered data.


jamesgorrie:
Closing in favour of #5903

PR #5589, opened 2022-08-03
AR - replace hv by rem to make it work in android

marjisound:
## What does this change?
This PR replaces the `40vh` by `22.4rem` for the pinned post collapsible body `max-height` because the `vh` wasn't working properly in android web view. 

## Why?
Making it work in Android devices

## Screenshots
Android:
| Before      | After      |
|-------------|------------|
| image | image |

IOS (**Unchanged**):
| Before      | After      |
|-------------|------------|
| image | image |

### DIFF
--- a/apps-rendering/src/components/PinnedPost/index.tsx
+++ b/apps-rendering/src/components/PinnedPost/index.tsx
@@ -134,7 +134,7 @@ const fakeButtonStyles = (format: ArticleFormat): SerializedStyles => css`
 `;
 
 const collapsibleBody = css`
-	max-height: 40vh;
+	max-height: 22.4rem;
 	overflow: hidden;
 `;
 


abeddow91 (approved):
Good spot! All approved 🚀

georgeblahblah:
This is a curious problem! I'm probably reading the [compatability tables](https://caniuse.com/viewport-units) wrong, but it looks like to me that `vh` _should_ be supported in Android WebViews. I wonder if this bug is being triggered by some other behaviour, like resizing of the webview which [can cause problems on mobile devices](https://css-tricks.com/the-trick-to-viewport-units-on-mobile/).

Another thing maybe worth noting is that `rem` does not behave the same way as `vh`. `rem` is a relative unit, so the actual space `1rem` takes up is determined by the `font-size` set on the root `html` element.

The app allows the user to scale the font size, which means that `22.4rem` is ≈ `40vh` at the default font size. But if the user changes the app's font size, `22.4rem` might be quite a lot different to `40vh`. In the screenshot below, I increased the root font size to `20px` which increases the space the pinned post container takes up to ~80%.
<details><summary>Screenshot</summary>

</details>

It might be that this isn't a problem in this case, but I wanted to raise the discussion in case the impact wasn't obvious from the `rem`/`vh` units alone.

marjisound:
Thanks @georgeblahblah for taking your time and letting me know about the `vh` support. Yes, the version of the android app I was using was very old (over 1 year old) and was probably using a very old android os and therefore chrome. After getting the latest version, the `vh` is working fine on my emulator. I created a revert PR to undo this `vh` replacement https://github.com/guardian/dotcom-rendering/pull/5597

PR #5597, opened 2022-08-04
Revert "AR - replace hv by rem to make it work in android"

marjisound:
Reverts guardian/dotcom-rendering#5589

In previous PR I had replaced `hv` by `rem`  because it wasn't working on my emulator. But thanks to @georgeblahblah for his comment https://github.com/guardian/dotcom-rendering/pull/5589#issuecomment-1204059856 I realized that I was using a very old version of the app which was running a very old version of the web view. It seems that `vh` is fully supported in `chrome for android` 

After I tested it with the latest version of `android-news-app` , I can now see it's working fine using `hv`
image

### DIFF
--- a/apps-rendering/src/components/PinnedPost/index.tsx
+++ b/apps-rendering/src/components/PinnedPost/index.tsx
@@ -134,7 +134,7 @@ const fakeButtonStyles = (format: ArticleFormat): SerializedStyles => css`
 `;
 
 const collapsibleBody = css`
-	max-height: 22.4rem;
+	max-height: 40vh;
 	overflow: hidden;
 `;
 


marjisound:
@JamieB-gu  @georgeblahblah  We can still think what's better in this situation. `vh` or `rem`? 

If I increase the text size to max in my android emulator, with `vh` it'd look like this:
image

the same large font size, using `rem` for the height, it'd look like this:
image

JamieB-gu:
> We can still think what's better in this situation. `vh` or `rem`?

I think that depends on what you're trying to achieve? If it's the amount of content in that box that's more important and it doesn't matter too much how tall it is, opt for `rem`. If the height of the box is important and you're less worried about exactly how much content is visible opt for `vh`.

This might be a question for UX/Design?

marjisound:
> > We can still think what's better in this situation. `vh` or `rem`?
> 
> I think that depends on what you're trying to achieve? If it's the amount of content in that box that's more important and it doesn't matter too much how tall it is, opt for `rem`. If the height of the box is important and you're less worried about exactly how much content is visible opt for `vh`.
> 
> This might be a question for UX/Design?

Yes that's right @JamieB-gu . Unfortunately Ben Wuersching is away on holiday right now. I can leave this PR open and will confirm with him when he's back

marjisound:
@JamieB-gu & @georgeblahblah After having a conversation with Ben, he suggested that `vh` would be a better option for the apps because we don't want the pinned post to take more and more space of the webview if the user increases the text size to maximum.

If you are happy and approve this PR, I will merge :)

georgeblahblah (approved):

PR #5598, opened 2022-08-04
Varying webpack bundles for tests

mxdvl:
## What does this change?

Allows us to serve a variant version of the bundle to a subset of users as a server side experiment.

Three bundles can now be generated at build time:
- `modern` – served to most users
- `variant` – served to a small subset of users in a server-side test
- `legacy` – served to browser that don’t support ES modules.

For consistency, the “nameless” default is now clearly labelled as modern: `manifest.json` becomes `manifest.modern.json`.

> **Note**
> In order to build the `variant` bundle, make sure to set the value of `BUILD_VARIANT` to `true` in `bundles.js`.
>
> Currently, the `variant` and `modern` are identical.
> When running an actual test, they should probably differ.

## Why?

We want to test the impact of different bundles on performance.

We have a several projects awaiting a way of testing new bundles:
- #5206 
- #5172 
- #5498
- #5899 

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/187447627-0b9b9430-386c-4bba-bee2-b6734e7497b3.png
[after]: https://user-images.githubusercontent.com/76776/187657460-e956ba03-ec65-48c4-b178-57ce39986cea.png

### DIFF
--- a/dotcom-rendering/scripts/test/build-check.js
+++ b/dotcom-rendering/scripts/test/build-check.js
@@ -7,6 +7,7 @@
 
 const find = require('find');
 const loadJsonFile = require('load-json-file');
+const { BUILD_VARIANT } = require('../webpack/bundles');
 
 const errorAndThrow = (error) => {
 	console.error(error);
@@ -28,7 +29,7 @@ const fileExists = async (glob) => {
 	// Check that the manifest files exist
 	await fileExists('manifest.modern.json');
 	await fileExists('manifest.legacy.json');
-	await fileExists('manifest.variant.json');
+	if (BUILD_VARIANT) await fileExists('manifest.variant.json');
 
 	// Check that the manifest files return values for all the chunks
 	const manifest = await loadJsonFile('./dist/manifest.modern.json');
--- a/dotcom-rendering/scripts/test/build-check.js
+++ b/dotcom-rendering/scripts/test/build-check.js
@@ -13,6 +13,7 @@ const errorAndThrow = (error) => {
 	throw new Error(error);
 };
 
+/** @type {(glob: string) => Promise<void>} */
 const fileExists = async (glob) => {
 	await find.file(glob, `./dist/`, (files) => {
 		if (files.length === 1) {
@@ -25,11 +26,12 @@ const fileExists = async (glob) => {
 
 (async () => {
 	// Check that the manifest files exist
-	await fileExists('manifest.json');
+	await fileExists('manifest.modern.json');
 	await fileExists('manifest.legacy.json');
+	await fileExists('manifest.variant.json');
 
 	// Check that the manifest files return values for all the chunks
-	const manifest = await loadJsonFile('./dist/manifest.json');
+	const manifest = await loadJsonFile('./dist/manifest.modern.json');
 	const legacyManifest = await loadJsonFile('./dist/manifest.legacy.json');
 
 	[
@@ -44,7 +46,7 @@ const fileExists = async (glob) => {
 		'newsletterEmbedIframe.js',
 		'relativeTime.js',
 		'initDiscussion.js',
-	].map((name) => {
+	].forEach((name) => {
 		if (manifest[name]) {
 			console.log(`Manifest returned value ${name}`);
 		} else {
--- a/dotcom-rendering/scripts/test/build-check.js
+++ b/dotcom-rendering/scripts/test/build-check.js
@@ -32,8 +32,13 @@ const fileExists = async (glob) => {
 	if (BUILD_VARIANT) await fileExists('manifest.variant.json');
 
 	// Check that the manifest files return values for all the chunks
-	const manifest = await loadJsonFile('./dist/manifest.modern.json');
-	const legacyManifest = await loadJsonFile('./dist/manifest.legacy.json');
+	const manifests = [
+		await loadJsonFile('./dist/manifest.modern.json'),
+		await loadJsonFile('./dist/manifest.legacy.json'),
+	];
+	if (BUILD_VARIANT) {
+		manifests.push(await loadJsonFile('./dist/manifest.variant.json'));
+	}
 
 	[
 		'sentryLoader.js',
@@ -48,18 +53,12 @@ const fileExists = async (glob) => {
 		'relativeTime.js',
 		'initDiscussion.js',
 	].forEach((name) => {
-		if (manifest[name]) {
-			console.log(`Manifest returned value ${name}`);
-		} else {
-			errorAndThrow(`Manifest did not return a value for ${name}`);
-		}
-
-		if (legacyManifest[name]) {
-			console.log(`Legacy manifest returned value ${name}`);
-		} else {
-			errorAndThrow(
-				`Legacy Loadabl manifest did not return a value for ${name}`,
-			);
+		for (const manifest of manifests) {
+			if (manifest[name]) {
+				console.log(`A manifest returned value ${name}`);
+			} else {
+				errorAndThrow(`A manifest did not return a value for ${name}`);
+			}
 		}
 	});
 })();
--- a/dotcom-rendering/scripts/webpack/webpack.config.browser.js
+++ b/dotcom-rendering/scripts/webpack/webpack.config.browser.js
@@ -46,7 +46,7 @@ const getLoaders = (bundle) => {
 					},
 				},
 			];
-		case 'modern':
+		case 'variant':
 			return [
 				{
 					loader: 'babel-loader',
@@ -74,7 +74,7 @@ const getLoaders = (bundle) => {
 					},
 				},
 			];
-		case 'variant':
+		case 'modern':
 			return [
 				{
 					loader: 'babel-loader',
@@ -84,8 +84,10 @@ const getLoaders = (bundle) => {
 							[
 								'@babel/preset-env',
 								{
-									targets:
-										'extends @guardian/browserslist-config',
+									bugfixes: true,
+									targets: {
+										esmodules: true,
+									},
 								},
 							],
 						],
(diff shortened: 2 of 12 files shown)


dotcom-rendering/scripts/webpack/webpack.config.browser.js
```diff
@@ -4,20 +4,112 @@ const GuStatsReportPlugin = require('./plugins/gu-stats-report-plugin');
...
+					},
+				},
+				{
+					loader: 'ts-loader',
+					options: {
+						configFile: 'tsconfig.build.json',
+						transpileOnly: true,
+					},
+				},
+			];
```
mxdvl, line 76: This is the same as `modern` for the time being.

dotcom-rendering/src/lib/assets.ts
```diff
@@ -47,31 +37,63 @@ const isDev = process.env.NODE_ENV === 'development';
...
+		getScripts(manifestPaths, file);
+
+export const generateScriptTags = (scripts: Array<string | false>): string[] =>
+	scripts.filter(isString).map((script) => {
+		if (script.includes('.legacy.')) {
+			return `<script defer nomodule src="${script}"></script>`;
+		}
+		if (script.includes('.modern.')) {
+			return `<script type="module" src="${script}"></script>`;
+		}
```
mxdvl, line 93: This is a lot simpler, but assumes that “external” does not contain these patterns.

dotcom-rendering/src/lib/assets.ts
```diff
@@ -47,31 +37,63 @@ const isDev = process.env.NODE_ENV === 'development';
...
+	return manifestPaths.map((manifestPath) => {
+		const manifest = getManifest(manifestPath);
+		const filename = manifest[file];
+
+		return `${ASSET_ORIGIN}assets/${filename}`;
+	});
 };
+
+export const getScriptsFromManifest =
+	(manifestPaths: Array<`./manifest.${string}.json`>) =>
```
OllysCoding, line 82: Should this be `Array<ManifestPath>`?

dotcom-rendering/scripts/webpack/webpack.config.js
```diff
@@ -114,18 +114,27 @@ module.exports = [
...
+	),
+	merge(
+		commonConfigs({
+			platform: 'browser.variant',
 		}),
 		require(`./webpack.config.browser`)({
-			isLegacyJS: false,
+			bundle: 'variant',
 			sessionId,
 		}),
```
OllysCoding, line 139: Should we _always_ be building the variant webpack, or should this be something that is only enabled when we have a test we want to run? I'm thinking that this will increase build times, etc.

dotcom-rendering/src/lib/assets.ts
```diff
@@ -47,31 +37,63 @@ const isDev = process.env.NODE_ENV === 'development';
...
+			readFileSync(resolve(__dirname, path), { encoding: 'utf-8' }),
+		);
+		if (typeof assetHash != 'object' || assetHash === null)
+			throw new Error('Not a valid AssetHash type');
+
+		/** @TODO validate the object */
+		return assetHash as AssetHash;
+	} catch (e) {
+		console.error('Could not load manifest in: ', path);
+		console.error('Some filename lookups will fail');
```
OllysCoding, line 52: Would this produce console errors in `make dev` as the legacy bundle does not exist?

dotcom-rendering/src/lib/assets.ts
mxdvl:
```suggestion
	(manifestPaths: ManifestPath[]) =>
```

dotcom-rendering/scripts/webpack/webpack.config.js
mxdvl: I don’t think there is another way to do this, as the building is indepedent from the test bucketing.

dotcom-rendering/src/lib/assets.ts
mxdvl: No, because we’re not using the manifests in development mode.

We’re returning early on lines 66-71 inside `getScripts` if `isDev`.

dotcom-rendering/scripts/webpack/webpack.config.js
OllysCoding: If we only 'enable' the variant build in the webpack config when we have a test to run, then in the request logic it should be relatively simple to try and 'read' in the variant manifest, and if it doesn't exist we should fall back to the modern/legacy variants.

mxdvl: I have looked at implementing this in bcabdd463d95de4ab3425281ea1be9dbbf258af8. I guess the key value is to make sure that we want the ability to build variant bundles in the future.

We have to make sure not to muddy the waters of the results, because it will be possible for a reader to be in the variant bucker of the [frontend#25443](https://github.com/guardian/frontend/pull/25443) experiment, but not be served the variant because we have not built it in DCR.

dotcom-rendering/src/lib/assets.ts
```diff
@@ -47,31 +37,63 @@ const isDev = process.env.NODE_ENV === 'development';
 
 export const ASSET_ORIGIN = decideAssetOrigin(process.env.GU_STAGE, isDev);
 
-export const getScriptArrayFromFile = (
+const getManifest = (path: string): AssetHash => {
+	try {
+		const assetHash: unknown = JSON.parse(
+			readFileSync(resolve(__dirname, path), { encoding: 'utf-8' }),
```
OllysCoding, line 43: Does this file get re-read for every request? 

I think previously this file being would be read at app startup, and then would exist in-memory from that point, I'd be interested whether re-reading it would have any impact on rendering latency?

dotcom-rendering/src/lib/assets.ts
```diff
@@ -47,31 +37,63 @@ const isDev = process.env.NODE_ENV === 'development';
...
-	return scripts;
+	return manifestPaths.map((manifestPath) => {
+		const manifest = getManifest(manifestPath);
+		const filename = manifest[file];
+
+		return `${ASSET_ORIGIN}assets/${filename}`;
+	});
 };
+
+export const getScriptsFromManifest =
```
OllysCoding, line 81: I would love if there was a better naming scheme we could use that makes it clearer the function currying?

dotcom-rendering/src/lib/assets.ts
OllysCoding: Could be a little bit safer `.endsWidth('.modern.js')`?

dotcom-rendering/scripts/test/build-check.js
```diff
@@ -25,11 +27,12 @@ const fileExists = async (glob) => {
...
 	// Check that the manifest files exist
-	await fileExists('manifest.json');
+	await fileExists('manifest.modern.json');
 	await fileExists('manifest.legacy.json');
+	if (BUILD_VARIANT) await fileExists('manifest.variant.json');
 
 	// Check that the manifest files return values for all the chunks
-	const manifest = await loadJsonFile('./dist/manifest.json');
+	const manifest = await loadJsonFile('./dist/manifest.modern.json');
 	const legacyManifest = await loadJsonFile('./dist/manifest.legacy.json');
```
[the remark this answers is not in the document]
mxdvl: Fair point.

dotcom-rendering/src/lib/assets.ts
mxdvl: I cannot really think of any name… 😢

dotcom-rendering/src/lib/assets.ts
mxdvl: I guess that is a valid point, I do not have a strong feeling about this, but feels like this could be revisited at a later point if we ever see any actual performance impact. I would hope reading files from disk is not our bottleneck.

The easiest thing to recreate the existing behaviour would be to create an object one with the mappings:

```ts 
const manifests: { [key: string]: AssetHash } = {
  './manifest.modern.json': getManifest('./manifest.modern.json'),
  './manifest.legacy.json': getManifest('./manifest.legacy.json'),
  './manifest.variant.json': getManifest('./manifest.variant.json'),
} ;
```

dotcom-rendering/src/lib/assets.ts
mxdvl: Not quite, because of the fact that we might have a hash. We’ve implemented a regular expression matching in ec3c5e6aef1bb8f78ce33b4cf7bcf3a441f133b9.

dotcom-rendering/src/web/server/articleToHtml.tsx
```diff
@@ -108,6 +89,14 @@ export const articleToHtml = ({ article: CAPIArticle }: Props): string => {
 			'model.dotcomrendering.pageElements.TweetBlockElement',
 	);
 
+	const manifestPaths: ManifestPath[] =
+		BUILD_VARIANT &&
+		CAPIArticle.config.abTests.dcrJsBundleVariant === 'variant'
+			? ['./manifest.variant.json', './manifest.legacy.json']
+			: ['./manifest.modern.json', './manifest.legacy.json'];
+
+	const getScripts = getScriptsFromManifest(manifestPaths);
```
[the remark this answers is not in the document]
mxdvl: We have opted for keeping the original name to reduce the diff and avoid answering that question.

jlieb10 (approved):
Excellent work @mxdvl  - thanks for walking through this PR with me

mxdvl:
Deployed on CODE and saw the assets come through as expected

image

PR #5545, opened 2022-07-29
feat: newsletter signup layout main section 

cemms1:
_Once [this PR](https://github.com/guardian/dotcom-rendering/pull/5599) is merged, base will change back to `main`_

### Improves the implementation of the main section of the individual newsletter signup layout
- Removed Article Body from the newsletter signup layout since the intention is to keep the newsletter signup page minimal
- Intend to use main media caption as the newsletter preview URL (since these will be highly curated signup pages, not normal articles)
- Use new style of email signup iframe rather than relying on old embed methods 
- Allowed option of removing the privacy wording underneath the email signup input, with caveat that override should only happen if we know the text will be displayed elsewhere on the page (override needed for the newsletter signup layout design)

## Why?

Newsletters team work to create a new sign-up layout for individual newsletters:
[As per Figma designs](https://www.figma.com/file/Nwcf5adXfG6hgHbwuD1OoQ/FINAL%3A-Single-signup-page-designs?node-id=321%3A52138)

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/43961396/181764914-7c3a8aa6-0222-4b6e-b2f0-020b10817958.png
[after]: https://user-images.githubusercontent.com/43961396/182652425-84171f65-7e62-4e47-a5fd-28ff6454ee1a.png

PR #5599, opened 2022-08-04
refactor: tweaks to newsletter signup components

cemms1:
## What does this change?

- ~Moved newsletters-related components to their own directory~ (_reverted as recommended_), added stories for some components
- Distinguished between NewsletterDetail and NewsletterFrequency components since the former design depends on context whereas the frequency text will only ever be under the signup form.
- Rearranged some of the styling into emotion css blocks separate from React component for better readability
- Split out SignupForm to allow the storybook mock to be a closer representation to the component

## Why?

Splitting out changes from [original PR](https://github.com/guardian/dotcom-rendering/pull/5545). These are the refactoring-type changes which are mainly separating out CSS styles, moving components to a new directory, adding storybook stories for them to better keep track of changes.

## Screenshots

See Chromatic diff for comparison of components

### DIFF
--- a/dotcom-rendering/src/web/components/EmailSignup.tsx
+++ b/dotcom-rendering/src/web/components/EmailSignup.tsx
@@ -85,7 +85,7 @@ export const EmailSignup = ({
 		<aside css={containerStyles}>
 			<div css={stackBelowTabletStyles}>
 				<p css={titleStyles(theme)}>
-					Sign up to <span>{name}</span> today
+					Sign up to <span>{name}</span>
 				</p>
 				<div css={noHeightFromTabletStyles}>
 					<NewsletterDetail text={frequency} />
--- a/apps-rendering/api-models/scala/version.sbt
+++ b/apps-rendering/api-models/scala/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "0.12.1-SNAPSHOT"
+version in ThisBuild := "1.0.1-SNAPSHOT"
--- a/apps-rendering/api-models/thrift/src/main/thrift/appsRendering.thrift
+++ b/apps-rendering/api-models/thrift/src/main/thrift/appsRendering.thrift
@@ -17,6 +17,13 @@ enum RelatedItemType {
     COMMENT
 }
 
+enum Edition {
+	UK,
+	US,
+	AU,
+	INTERNATIONAL
+}
+
 struct Image {
     1: required string url
     2: required i32 height
@@ -121,4 +128,5 @@ struct RenderingRequest {
     6: optional list<Campaign> campaigns
     7: optional RelatedContent relatedContent
     8: optional FootballContent footballContent
+    9: optional Edition edition
 }
--- a/apps-rendering/package-lock.json
+++ b/apps-rendering/package-lock.json
@@ -7929,54 +7929,60 @@
       "integrity": "sha512-iO9ZQHkZxHn4mSakYV0vFHAVDyEOIJQrV2uZ06HxEPcx+mt8swXoZHIbaaJ2crJYFfErySgktuTZ3BeLz+XmFA=="
     },
     "@typescript-eslint/eslint-plugin": {
-      "version": "4.31.2",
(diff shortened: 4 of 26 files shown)


dotcom-rendering/src/web/components/Newsletters/NewsletterFrequency.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const spanStyles = css`
+	margin-left: ${space[1]}px;
+	${textSans.xsmall()}
+	strong {
+		${textSans.xsmall({ fontWeight: 'bold' })}
+	}
+`;
+
+export const NewsletterFrequency: React.FC<Props> = ({ frequency }) => {
```
oliverlloyd, line 22: Is this component used anywhere? I can only see it being imported into the story

dotcom-rendering/src/web/components/SecureSignup.tsx
```diff
@@ -92,7 +61,11 @@ const generateForm = (
 	return { html, styles };
 };
 
-export const SecureSignup = ({ newsletterId, successDescription }: Props) => {
+export const SecureSignup: React.FC<Props> = ({
```
oliverlloyd, line 64: We don't use the ` React.FC<Props>` style of typing components in DCR. There's a [nice write up for why in this article](https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript) but essentially it's mainly because it adds a `children` prop which then isn't used but also because we want to keep one way for how things are done.

oliverlloyd (commented):
Thanks for breaking these changes out, this really helps 💙 

I've listed a few things below which possibly look like a lot but actually none of them are major, mostly it's around conventions. I realise stuff like this can seem nitpicky but we try to maintain consistent approaches to things in DCR in an attempt to keep this code maintainable for as long as possible

## Sub directories
This change moves some of the newslettery components into a sub directory called 'Newsletters' however DCR, _typically_, uses a flat structure for components. I know there are some exceptions to this so it makes this very unclear!

Reasons why things are flat

1) there's an argument that once you reach a certain size then having things in directories doesn't aid discovery and most people resort to using shortcuts like `CMD+P` to find files

2) Our island solution requires this flat structure. You cannot hydrate a component that is not placed at the root of the /components folder

1 is arguably subjective but 2 is not and using sub directories will prevent hydration of these components in the future

## `React.FC<Props>`
We don't use this style of typing in DCR

## Unused code
It looks like this PR is adding some components that are not used. I'm guessing this is a side effect of the breaking out work?

oliverlloyd:
> it's been tricky to find the ones that are useful! Is there any documentation or best practice stuff I can read up on to help get up to speed faster?

Yeah, it's a classic problem with larger repos; discovering what already exists is never easy.

To help, we've got a [list of stories](https://www.chromatic.com/library?appId=5dfcbf3012392c0020e7140b) that visually display most components. The section naming is a bit off, if you ignore the ones under 'AR' and focus on those under 'Components' that should help

> I had a look for a linting rule this morning but couldn't find one for our needs so it might be worth creating a rule to discourage the use of the React.FC generic since it's used quite a lot in the dotcom-rendering dir at the moment.

Done: https://github.com/guardian/dotcom-rendering/pull/5622

dotcom-rendering/src/web/components/EmailSignup.tsx
```diff
@@ -85,13 +88,14 @@ export const EmailSignup = ({
 					Sign up to <span>{name}</span> today
 				</p>
 				<div css={noHeightFromTabletStyles}>
```
dblatcher, line 90: Really minor point - we should probably update the comment on line 55 that still refers to a `NewsletterFrequency` component.

dblatcher (approved):
Looks good - I was thinking before that it would make sense to combine the 2 components that used the SvgNewsletters into one .

bryophyta:
Very helpful discussion above, thanks @cemms1 and @oliverlloyd!

Does anyone know whether the decisions about sub-directories and avoiding `React.FC` are currently documented anywhere in our repo? I'm trying to pick up a thread of work that I started a while ago to make some of these sorts of conventions a bit more discoverable.

PR #5600, opened 2022-08-04
Remove stale comment to deleted document

georgeblahblah:
## What does this change?

- Removes comment linking to a deleted Google sheet

It's possible the document still exists and I just don't have access, though!


oliverlloyd (approved):
I can't see it either; I think it's gone...

PR #5601, opened 2022-08-04
limit filters for topics with counts of 3 and more

marjisound:
## What does this change?
This PR limits the topic filters to only the topics that have been mentioned in at least 3 blocks. 

## Why?
Requested by the editorials as there's not much point in showing a filter category that only 1 or 2 blocks are mentioning it. 

## Screenshots

| Before      | After      |
|-------------|------------|
| image | image |

### DIFF
--- a/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
+++ b/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
@@ -86,7 +86,7 @@ export const TopicFilterBank = ({
 	const selectedTopic = selectedTopics?.[0];
 	const topFiveTopics = availableTopics
 		.slice(0, 5)
-		.filter((topic) => topic.count && topic.count > 2);
+		.filter((topic) => !!topic.count && topic.count > 2);
 
 	if (selectedTopic) {
 		const selectedIndex = availableTopics.findIndex((availableTopic) =>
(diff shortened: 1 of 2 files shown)


abeddow91 (approved):
Looks great! Nice story coverage too 🤩

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
```diff
@@ -84,7 +84,9 @@ export const TopicFilterBank = ({
 }: Props) => {
 	const palette = decidePalette(format);
 	const selectedTopic = selectedTopics?.[0];
-	const topFiveTopics = availableTopics.slice(0, 5);
+	const topFiveTopics = availableTopics
+		.slice(0, 5)
+		.filter((topic) => !!topic.count && topic.count > 2);
```
mxdvl, line 89: Do you want to `filter` first and then `slice`? Otherwise you might have miss topics with more count that are outside the first 5.

Event if there is an assumption that the topics are sorted by count from the backend, that’s quite a loose contract, so it’s probably worth updating this.

marjisound: yea good idea. I'll make another PR to change the order of filter and slice 👍

PR #5605, opened 2022-08-04
Add star rating to DCR frontCards

### LINKED ISSUE #5532: Support star rating in DCR front cards
Repo: https://github.com/guardian/dotcom-rendering/tree/dc-star-rating-dcr

We need to show star rating in DCR fronts cards when they exist, like in frontend.

Frontend
Image

DCR
Image

The new design is different, with the stars sitting below the headline. See here: https://www.figma.com/file/5CfbWOeZPRi15lXBD5u1rW/Card-layout-system?node-id=0%3A1

We already get this information in the JSON frontend sends to DCR:

Image

So we need to add `starRating` in the `DCRFrontCard` and pass it further down where needed.

DanielCliftonGuardian:
## What does this change? 
This adds star ratings to DCR front cards

Closes #5532 

## Why? 
This change matches the figma designs: https://www.figma.com/file/5CfbWOeZPRi15lXBD5u1rW/Card-layout-system?node-id=0%3A1

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/110032454/182878611-30cd17ee-d45e-4051-8052-6766c15428a9.png
[after]: https://user-images.githubusercontent.com/110032454/182878389-5b14b7ba-e8c4-429d-b04d-a86022786fbd.png

### DIFF
--- a/dotcom-rendering/src/web/components/DynamicFast.tsx
+++ b/dotcom-rendering/src/web/components/DynamicFast.tsx
@@ -68,6 +68,7 @@ const Primaries = ({
 						>
 							<FrontCard
 								trail={card}
+								starRating={card.starRating}
 								containerPalette={containerPalette}
 								showAge={showAge}
 								supportingContent={card.supportingContent}
@@ -112,6 +113,7 @@ const Primaries = ({
 					>
 						<FrontCard
 							trail={card}
+							starRating={card.starRating}
 							containerPalette={containerPalette}
 							showAge={showAge}
 							supportingContent={card.supportingContent}
@@ -186,6 +188,7 @@ const FirstBigBoostedPlusBig = ({
 				>
 					<FrontCard
 						trail={big}
+						starRating={big.starRating}
 						containerPalette={containerPalette}
 						showAge={showAge}
 						supportingContent={big.supportingContent}
@@ -210,6 +213,7 @@ const FirstBigBoostedPlusBig = ({
 						>
 							<FrontCard
 								trail={card}
+								starRating={card.starRating}
 								containerPalette={containerPalette}
 								showAge={showAge}
 								imageUrl={undefined}
@@ -233,6 +237,7 @@ const FirstBigBoostedPlusBig = ({
 						>
 							<FrontCard
 								trail={card}
+								starRating={card.starRating}
 								containerPalette={containerPalette}
 								showAge={showAge}
 								imageUrl={undefined}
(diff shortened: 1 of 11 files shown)


dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -68,6 +68,7 @@ const Primaries = ({
 						>
 							<FrontCard
 								trail={card}
+								starRating={card.starRating}
```
oliverlloyd, line 71: Do we need these changes?

I think `FrontCard` will pass this `starRating` property through by default

oliverlloyd (approved):
I give this PR ⭐ ⭐ ⭐ ⭐ ⭐ s

Just one comment, I suspect we don't need to explicitly pass through the starRating property in all the containers as the `FrontsCard` component picks it up by default (which is the thing that's cool about that abstraction).

If we _didn't_ want to show stars we could have `starRating={undefined}` but I am guess we would never want to do that and if the tool asks use to show starts then we should

mxdvl (commented):
Nice!

Echoing @oliverlloyd’s comment.

oliverlloyd:
> If we _didn't_ want to show stars we could have `starRating={undefined}` but I am guessing we would never want to do that and if the tool asks use to show starts then we should

Thinking about this more, I'm not sure if this is true or not. We previously displayed the stars above the image which meant that if a card didn't have an image then, even if the model provided it, the stars would not show. But now that the stars are decoupled from the image, this is no longer the case. Even small cards with no image will show stars, see screenshot.

Screenshot 2022-08-05 at 08 02 04

Honestly, I kind of like it, but this seems to have an editorial impact, are we making these review articles more prominent? @HarryFischer what do you think?

HarryFischer:
I think it's good to always show stars when something is a review. In my Figma design I'd adjusted the height of the stars as font size of the headline changes. Can't tell if that logic is in this PR but when I tested out the design I don't think we were adding too much prominence.

oliverlloyd:
> I think it's good to always show stars when something is a review. In my Figma design I'd adjusted the height of the stars as font size of the headline changes. Can't tell if that logic is in this PR but when I tested out the design I don't think we were adding too much prominence.

Sounds good. In this case I think this PR is good to go as it is and we can have another ticket to ensure that we reduce star size in line with the headline 👍

PR #5606, opened 2022-08-05
Remove `DCRServerDocumentData`

oliverlloyd:
## What does this change?
This PR refactors the code around `articleToHtml` and also a bit around `frontToHtml`. It makes it so that we now only pass one property to these functions. By doing this I was able to remove the `DCRServerDocumentData` type

## Why?
Previously, we had the logic to break out values from the data sent to DCR in different places as well as some unused props that were being passed down but never used. This wasn't intentional, it just ended up this way.

Having `index` be the place where functions are simply called and `articleToHtml` the place where data is extracted and set helps rationalise things a bit and will hopefully make future maintenance and expansion easier.

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -841,12 +841,6 @@ interface GADataType {
 // General DataTypes //
 // ----------------- //
 
-interface DCRServerDocumentData {
-	CAPIArticle: CAPIArticleType;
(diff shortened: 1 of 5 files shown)


OllysCoding (approved):
+1 This seems like a sensible refactor - removes the cognitive load of needing to understand a whole additional type just to work on these files

PR #5607, opened 2022-08-05
Bug fix: provide fonts css to signup iframe

dblatcher:
## What does this change?
`SecureSignUpIframe` add the "GuardianTextSans" fontFaces to the iframe document.

## Why?
`SecureSignUpIframe` has a `styles` prop for passing the stringified styles for needed for the iframe contents, but this does not include the fonts css included as part of the page template. The `font-family` values were being  set to use the guardianTextSans, but the iframe didn't have the font definitions, so was using the fallback font.

## Screenshots 
(note the lowercase "L" in "Enter your email address")
| Before      | After      |
|-------------|------------|
| Screenshot 2022-08-05 at 12 27 11| Screenshot 2022-08-05 at 12 26 40|


dblatcher:
It should be noted this added extra HTML payload weight to the SecureSignUpIframe component - the stringified HTML and css passed as props to the Island was 5049 characters before. Adding the font adds an extra 2113, bringing it up to 7162.

dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
```diff
@@ -283,6 +283,58 @@ export const SecureSignupIframe = ({
...
+			}
+		};
+
+		const iframeFontFaceSet = iframe?.contentDocument?.fonts;
+		if (!iframeFontFaceSet) {
+			return fallbackMethod();
+		}
+
+		// The ts.dom interface for FontFaceSet does not contain the .add method,
+		// browser support is not universal.
```
sndrs, line 307: i think this is only true for IE, which we do not need to support anymore. maybe an alternative approach is to use `.add` where supported, and fallback to system fonts where not?

dblatcher: sounds reasonable - the function can just exit if the method isn't supported

dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
```diff
@@ -283,6 +283,58 @@ export const SecureSignupIframe = ({
...
 	};
 
+	const addFontToIframe = (requiredFontNames: string[]) => {
+		const { current: iframe } = iframeRef;
+
+		// fallback: clone the whole fonts node into the iframe.
+		// used if the browser does not support FontFaceSet.add
+		const fallbackMethod = () => {
+			const fontNode = document
+				.querySelector('style.webfont')
```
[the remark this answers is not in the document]
dblatcher: agreed. I've removed the fallback function and refactors to make the logic a bit clearer.

dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
```diff
@@ -283,6 +288,40 @@ export const SecureSignupIframe = ({
...
+			// but still ts considers it possibily undefined
+			if (iframeFontFaceSet.add) {
+				iframeFontFaceSet.add(font);
+			}
+		});
+	};
+
+	const onIFrameLoad = (): void => {
+		attachListenersToIframe();
+		addFontToIframe(['GuardianTextSans']);
```
alinaboghiu, line 322:
```suggestion
		addFontsToIframe(['GuardianTextSans']);
```

dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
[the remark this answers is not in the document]
sndrs: :👍

dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
```diff
@@ -283,6 +288,40 @@ export const SecureSignupIframe = ({
 		resetIframeHeight();
 	};
 
+	const addFontToIframe = (requiredFontNames: string[]) => {
```
sndrs, line 291: i think this is looking pretty nice. how would you feel about pulling this up to https://github.com/guardian/libs? this is an interesting problem that wasn't obvious to how to solve at first. would be a good one to make more generally available/public?

dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
dblatcher: 👍

dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
dblatcher: That sounds great - I'll look at doing a PR in /libs. We've got some time pressure on this feature so I'd probably  need to merge in this PR and then do another PR to use the import from /libs when the update is available in DCR.

sndrs: ok great - yeah whenever it works for you

sndrs (approved):

PR #5610, opened 2022-08-05
Better sublink colours

oliverlloyd:
## What does this change?
This fixes the decision logic for how we decide which `format` property to use on sublinks

## Why?
We were not taking into account the sublink's own `format`

## Screenshots

| Before      | After      |
|-------------|------------|
| Screenshot 2022-08-05 at 14 53 37 | Screenshot 2022-08-05 at 14 52 45 |


ioannakok (approved):
Thanks for fixing this! It was annoying indeed!

OllysCoding (approved):

PR #5611, opened 2022-08-05
ESLint: stop warning on danger

mxdvl:
## What does this change?

Stop warning on dangerous React methods, like `dangerouslySetInnerHtml`.

## Why?

We use `dangerouslySetInnerHTML` in several components. The function name is explicit enough as it is, and we have valid use cases for it, [including accessibility](https://github.com/guardian/dotcom-rendering/blob/11c68a72d16d093762615d7b686a9a4a2fe1cd3f/dotcom-rendering/docs/architecture/022-NoJS-navigation-and-accessibility.md).


oliverlloyd (approved):
Yolo

PR #5612, opened 2022-08-05
Varying webpack bundles for tests II

arelra:
## What does this change?

Updates #5598 with

- promote conditional webpack bundling from presets -> loaders (we will need this for #5172 )
- align naming
- partial application of `getScriptsFromManifest` to avoid having to pass `manifestPath` for every call

### DIFF
--- a/dotcom-rendering/scripts/webpack/webpack.config.browser.js
+++ b/dotcom-rendering/scripts/webpack/webpack.config.browser.js
@@ -4,7 +4,7 @@ const GuStatsReportPlugin = require('./plugins/gu-stats-report-plugin');
 const DEV = process.env.NODE_ENV === 'development';
 
 /**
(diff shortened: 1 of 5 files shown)


mxdvl (commented):
I think you're missing the legacy bundle which should always be served, regardless of `variant`/`modern`.

Maybe we should rename `modern` to `control`

arelra:
> I think you're missing the legacy bundle which should always be served, regardless of `variant`/`modern`.
> 
> Maybe we should rename `modern` to `control`

It's still there - I just changed to order to legacy, modern, variant.

Have tested this locally

mxdvl (approved):
Ignore my above comment, the legacy manifest is always captured independently.

Renaming `modern` to `control` still stands.

PR #5613, opened 2022-08-05
Reduce email signup sizing

dblatcher:
## What does this change?
Reduces the size of the text and form components used within the EmailSignup Component

## Why?
We've had some feedback that the design should be smaller.

## Screenshots

| Before(218px high)      | After (190px high)      |
|-------------|------------|
| Screenshot 2022-08-05 at 16 02 30| Screenshot 2022-08-05 at 16 02 18|

### DIFF
--- a/dotcom-rendering/__mocks__/SecureSignupMock.tsx
+++ b/dotcom-rendering/__mocks__/SecureSignupMock.tsx
@@ -1,5 +1,5 @@
 import { css } from '@emotion/react';
-import { neutral, space } from '@guardian/source-foundations';
+import { neutral, space, textSans } from '@guardian/source-foundations';
 import { Button, Label, TextInput } from '@guardian/source-react-components';
 import { NewsletterPrivacyMessage } from '../src/web/components/NewsletterPrivacyMessage';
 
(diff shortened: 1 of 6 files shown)


dotcom-rendering/__mocks__/SecureSignupMock.tsx
```diff
@@ -9,21 +9,34 @@ export const SecureSignup = ({ newsletterId }: Props) => {
 	return (
 		<div title="This is a mock for the SecureSignup component which is not compatible with storybook. Changes to the real component will not be automatically reflected in this mock.">
 			<form id={`secure-signup-${newsletterId}`}>
-				<Label text="Enter your email address" />
+				<Label
+					text="Enter your email address"
+					css={css`
```
oliverlloyd, line 14: Does the `Label` component have a `cssOverrides` prop?

dotcom-rendering/__mocks__/SecureSignupMock.tsx
```diff
@@ -9,21 +9,34 @@ export const SecureSignup = ({ newsletterId }: Props) => {
...
 						css={css`
 							margin-right: ${space[3]}px;
+							margin-bottom: ${space[2]}px;
 							flex-basis: 335px;
 							flex-shrink: 1;
+
+							input {
+								height: 36px;
+								margin-top: 0;
+							}
```
oliverlloyd, line 39: Do you think you'd be able to target this css onto the component itself?

dotcom-rendering/__mocks__/SecureSignupMock.tsx
dblatcher: Yes - I'll use that instead and make the same change on the non-mocked version of the component.

PR #5617, opened 2022-08-07
`DynamicSlowMPU`

### LINKED ISSUE #5147: "dynamic/slow-mpu" (`42`) container

oliverlloyd:
## What does this change?
This adds support for the `DynamicSlowMPU` container. Closes #5147 

This container supports four possible layouts. The decision for which of the four to show is based on the number of 'big' cards and - in the case where there are 2 big cards - if the first big card is boosted or not

### No big cards
If no big cards are provided a single slice is rendered using `Card33_ColumnOfThree33_Ad33`

Screenshot 2022-08-08 at 07 16 59

### One big card
If only one big card is provided the first standard card is promoted such that there are two big cards. See 'Two big cards' 👇 .

### Two bigs cards
If two big cards are provided they are placed side by side with equal spacing using `Card50_Card50` and the remaining, standard, cards are placed in `ColumnOfThree50_Ad50` in the slice underneath

Screenshot 2022-08-07 at 16 38 18

### Two big cards, with the first boosted
If the first of the two big cards is promoted it is given more width using `Card75_Card25`

Screenshot 2022-08-07 at 16 41 40

### Three big cards
When there are three bigs cards the `Card50_Card25_Card25` layout is used

Screenshot 2022-08-07 at 16 43 28

### More than three bigs
Any additional big cards are demoted to standards with no changes to the layout

## What are these `Card50_Card50` slice abstractions?
In order to get a handle on the variations that are possible with this container I created some slice abstractions. These are inspired by [the existing work in Frontend](https://github.com/guardian/frontend/blob/main/common/app/layout/slices/Slice.scala) and also based on ideas and feedback from @bryophyta, @AshCorr and others in the team.

Early abstractions are to be avoided but I think we have now built up a good library of knowledge for what the different possibilities are here. These are local abstractions for now but could serve well in other situations at a later stage by moving them out into their own files.

#### Naming convention
Naming things is hard. I was tempted to use the original [monster naming convention](https://github.com/guardian/frontend/pull/5930) that @robertberry added because that was pretty cool but hopefully the names I ended up picking for these abstractions work well at communicating what they do. `Card50_Card50` is two single cards side by side with equal widths. `Card33_ColumnOfThree33_Ad33` is three equal spaced columns, the first with a single card, the second has a column with three cards in it, and the third is a ad (mpu).

I've added similar ascii art to that [in Frontend](https://github.com/guardian/frontend/pull/5930#discussion-diff-17486301) to provide another visual aid.

Using underscores in the names like this breaks the eslint rule [@typescript-eslint/naming-convention](https://typescript-eslint.io/rules/naming-convention/). I think this is okay for this use case. Having the underscore improves readability of the name and better communicates what the component does.


paperboyo:
Just to note that for years, _Two bigs cards_ had wrongly-sized images (too low res). Hopefully DCR will finally remedy that!

dotcom-rendering/src/web/components/DynamicSlowMPU.stories.tsx
```diff
@@ -0,0 +1,164 @@
...
+				breakpoints.desktop,
+				breakpoints.leftCol,
+				breakpoints.wide,
+			],
+		},
+	},
+};
+
+const bigs = trails.slice(0, 3);
+const standards = trails.slice(3);
```
oliverlloyd, line 26: These are arbitrary

dotcom-rendering/src/web/components/DynamicSlowMPU.tsx
```diff
@@ -0,0 +1,382 @@
...
+				<ColumnOfThree50_Ad50
+					cards={standardCards}
+					containerPalette={containerPalette}
+					showAge={showAge}
+				/>
+			</>
+		);
+	}
+
+	// Else bigCards.length === 3
```
oliverlloyd, line 367: Logically, bigCards has a length of three here but typescript isn't following along so I'm defaulting to this outcome.

dotcom-rendering/src/web/components/DynamicSlowMPU.tsx
```diff
@@ -0,0 +1,382 @@
...
+		<LI percentage="33.333%">
+			<Hide until="tablet">
+				{/* TODO: Replace mostpop with a more appropriate value */}
+				<AdSlot position="mostpop" />
+			</Hide>
+		</LI>
+	</UL>
+);
+
+/* ._________________._________________.
```
bryophyta, line 77: just an aesthetic preference, but what do you think of this style?

image

Happy to make them if you like them, but I don't want to impose if not.

bryophyta: either way, +1 on having the ascii diagrams! I find them really helpful

oliverlloyd: > just an aesthetic preference, but what do you think of this style?
> 
> image
> 
> Happy to make them if you like them, but I don't want to impose if not.

Super cool! How did you do that?

dotcom-rendering/src/web/components/DynamicSlowMPU.tsx
```diff
@@ -0,0 +1,382 @@
...
+	</UL>
+);
+
+/**
+ * DynamicSlowMPU
+ *
+ * This container only allows big and standard cards (from groupedTrails)
+ *
+ * Layout is dynamic deopending on the number of big cards. You're only
+ * allowed to have 2 or 3 big cards. If you pass 1 a standrd card will
```
bryophyta, line 291:
```suggestion
 * allowed to have 2 or 3 big cards. If you pass 1 a standard card will
```

dotcom-rendering/src/web/components/DynamicSlowMPU.tsx
bryophyta: I could definitely be wrong here, but: is typescript complaining because it's possible that `promoted` in ln.322 is just an empty array, in which case at this point `bigCards.length === 1`?

bryophyta: I'm finding the way that length checks are done both before and after the array is mutated seems to be making it a bit more difficult to follow the logic here. Have you considered any alternative ways of implementing this logic? One option that comes to mind is to break into two steps, something like `decidePromotion` and then `decideContainer` (though not sure that they'd need to be actually separate functions as such).

I'm happy to try and sketch what I have in mind in a bit more detail. If you think it sounds promising though then lmk and I'll try to describe it a bit clearer than I've done here

bryophyta:
This looks good to me! I like the level of abstraction that you've gone with, and the naming convention seems very suitable to me. I could maybe be tempted to add a jsdoc comment to the functions explaining the naming scheme though -- i.e. that the number refers to percentage width, rather than e.g. number of cards. But I think when you look at the code itself it's pretty clear.

I've put in a couple of very small comments, and one slightly more substantive one. I don't think it's blocking as such, but seems worth discussing a little if possible, before merging?

dotcom-rendering/src/web/components/DynamicSlowMPU.tsx
bryophyta: Using this set of characters:

```
    ┌─┬┐ top
    │ ││ head
    ├─┼┤ head_row
    │ ││ mid
    ├─┼┤ row
    ├─┼┤ foot_row
    │ ││ foot
    └─┴┘ bottom
```
I think I originally came across them when I was looking into the [rich](https://github.com/Textualize/rich/blob/a6f2d84511cda7f367dca527177d7b586264755b/rich/box.py) package for Python.

They don't render great in all fonts, but they seem to be fine in most monospace, including in [comments rendered in github](https://github.com/guardian/dotcom-rendering/blob/1be769c18136cab1785e93fec513a00e95ccd8ad/dotcom-rendering/src/web/components/ExtraCardsContainer.tsx#L25-L33).

bryophyta: thought I'd open a quick [PR](https://github.com/guardian/dotcom-rendering/pull/5627) while I was thinking about this, though it'd probably need a bit of tweaking. Totally up to you whether you want to use them though :)

dotcom-rendering/src/web/components/DynamicSlowMPU.tsx
oliverlloyd: > I could definitely be wrong here, but: is typescript complaining because it's possible that `promoted` in ln.322 is just an empty array, in which case at this point `bigCards.length === 1`?

The fact that each branch is wrapped in a conditional means typescript will complain that there is not a certain return path, even if we know that it is sure that one of the branches will be followed.

I think this is fine. We're saying that the default is to render `Card50_Card25_Card25`.

oliverlloyd: > I'm finding the way that length checks are done both before and after the array is mutated seems to be making it a bit more difficult to follow the logic here. Have you considered any alternative ways of implementing this logic? One option that comes to mind is to break into two steps, something like `decidePromotion` and then `decideContainer` (though not sure that they'd need to be actually separate functions as such).

You are 100% spot on with this. I implemented your suggestion and the code reads so much better. Thanks!

dotcom-rendering/src/web/components/DynamicSlowMPU.tsx
oliverlloyd: These versions of the diagrams defo look clicker but I struggled to represent the image position. Do you have any tricks for indicating where the image sits? I tried this but it looks a bit off

```
/*
 * ┌───────────┐┌───────────┐
 * ├──###──────┤│    MPU    │
 * ├──###──────┤│           │
 * └──###──────┘└───────────┘
 */
```

bryophyta: ah, good point... maybe something to leave for a potential refactor then? but I'll have a think and let you know if I come up with a promising solution

dotcom-rendering/src/web/components/DynamicSlowMPU.tsx
bryophyta: glad to hear it was a helpful suggestion :) and yeah fwiw I find it a lot easier to follow the logic now 👍

bryophyta (approved):

OllysCoding:
A bit late here but wanted to say I _love_ how this is written! Great job @oliverlloyd, I'm taking notes 🖊️

PR #5621, opened 2022-08-08
Add a new class for blog rich links

DanielCliftonGuardian:
Co-Authored-By: Joshua <joshuathomaslieberman@gmail.com>

## What does this change?
This fixes text size bug.
## Why?
On smaller screens live blog rich links were too small

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]:  https://user-images.githubusercontent.com/110032454/183407276-f06ff566-e859-49fe-b101-6746193ef0b6.png
[after]: https://user-images.githubusercontent.com/110032454/183407168-0e0fc8d3-7855-40f0-bc6d-0202629fc8dd.png

### DIFF
--- a/dotcom-rendering/src/web/components/RichLink.tsx
+++ b/dotcom-rendering/src/web/components/RichLink.tsx
@@ -84,8 +84,9 @@ const richLinkHeader = css`
 	color: ${neutral[0]};
 `;
 
-const blogRichLinkTitle = css`
+const richLinkTitleStyle = (parent: boolean) => css`
 	${headline.xxxsmall()};
+	font-size: ${!parent ? '14px' : undefined};
 	padding-top: 1px;
 	padding-bottom: 1px;
 	font-weight: 400;
@@ -93,11 +94,6 @@ const blogRichLinkTitle = css`
 		${headline.xxsmall()};
 		padding-bottom: 5px;
 	}
-`;
-
-const richLinkTitle = css`
-	${blogRichLinkTitle};
-	font-size: 14px;
 `;
 
 const labsRichLinkTitle = css`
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/components/RichLink.tsx
```diff
@@ -96,6 +95,11 @@ const richLinkTitle = css`
 	}
 `;
 
+const richLinkTitle = css`
+	${blogRichLinkTitle};
+	font-size: 14px;
+`;
```
oliverlloyd, line 101: Semantically it's a bit confusing if `richLinkTitle` extends `blogRichLinkTitle` because in the switch statement I feel like I'm either going to get labs, blog or normal. But actually I end up with a combination of them.

We don't often use this pattern (where one style includes another one). Typically we have functions that return styles based on values passed in, e.g.:

```
const myStyle = (aValue: valueType) => css`
  color: ${decideColour(aValue)};
`
```

Would that work here?

Failing that I guess we could just duplicate things? A little duplication is no bad thing.

oliverlloyd (approved):
Just one comment on how we decide css which I'm happy to leave up to you but otherwise great stuff

dotcom-rendering/src/web/components/RichLink.tsx
```diff
@@ -84,9 +84,9 @@ const richLinkHeader = css`
 	color: ${neutral[0]};
 `;
 
-const richLinkTitle = css`
+const richLinkTitleStyle = (parent: boolean) => css`
 	${headline.xxxsmall()};
-	font-size: 14px;
+	font-size: ${!parent ? '14px' : undefined};
```
oliverlloyd, line 89:
```suggestion
	${parentIsBlog && 'font-size: 14px'};
```
?

jlieb10 (commented):
Hey @DanielCliftonGuardian and @oliverlloyd - great refactor!!

PR #5622, opened 2022-08-08
Ban usage of `React.FC` and similar

oliverlloyd:
## What does this change?
This bans the `React.FC` type

We're warning for now, but might upgrade to `error` soon (if anyone has the urge to refactor away all existing use cases)

## Why?
[This article](https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript) provides a good summary and [this Github issue](https://github.com/facebook/create-react-app/pull/8177) has even more details but ultimately because we don't want to add the implicit `children` prop but also because we want one consistent way of doing things in DCR

### DIFF
--- a/dotcom-rendering/.eslintrc.js
+++ b/dotcom-rendering/.eslintrc.js
@@ -152,6 +152,14 @@ module.exports = {
 						types: {
 							'JSX.Element': 'Prefer type inference',
 							'EmotionJSX.Element': 'Prefer type inference',
+							'React.StatelessComponent':
+								'Please use const MyThing = ({foo, bar}: Props) instead',
+							'React.FunctionalComponent':
+								'Please use const MyThing = ({foo, bar}: Props) instead',
+							'React.SC':
+								'Please use const MyThing = ({foo, bar}: Props) instead',
+							'React.FC':
+								'Please use const MyThing = ({foo, bar}: Props) instead',
 						},
 						extendDefaults: true,
 					},


jamesgorrie (commented):
I am very much in favour of this, and it's been recommended against for quite a while.

If people want to use `children`  - what solution are we recommending they use? Currently is seems to be [explicitly setting the prop as a mix of `JSX.Element` and  `React.ReactNode`](https://github.com/guardian/dotcom-rendering/search?q=children).

An ADR linked to in the code could stop us flip-flipping on this later making the reaosn for this really visible. If it is this PR, we could just add a link to the rules so future Dotcommers know where to look for the reason.

Why would we not just go full error? I am wondering what this intermediary step offers.

AshCorr (approved):
The article explains this very well! This seems like a good change and will make things a lot more consistent.

oliverlloyd:
> I am very much in favour of this, and it's been recommended against for quite a while.
> 
> If people want to use `children` - what solution are we recommending they use? Currently is seems to be [explicitly setting the prop as a mix of `JSX.Element` and `React.ReactNode`](https://github.com/guardian/dotcom-rendering/search?q=children).

Typing `children` in React is a open discussion with lots of input. I think `React.ReactNode` is often good but there's a lot of opinions here and I don't think any answer is universally right.

> 
> An ADR linked to in the code could stop us flip-flipping on this later making the reason for this really visible. If it is this PR, we could just add a link to the rules so future Dotcommers know where to look for the reason.

Hopefully this PR provides enough information on the motivations for why we want to make this change. I doubt this will change in future. React.FC is now something of a legacy way to type react components with the general consensus moving away from it (for the reasons given above)

> 
> Why would we not just go full error? I am wondering what this intermediary step offers.

There is still code in DCR using the old way, once that is refactored away we can go full error

jamesgorrie:
@oliverlloyd 

> Hopefully this PR provides enough information on the motivations for why we want to make this change.

This makes sense - I wonder if a link in the code to the motivation (this PR) would be useful for people who come across this rule with no familiarity with it. Makes it easier than searching gitHub.

jamesgorrie:
> My guess would be that we do actually use React.FC in a lot of places (most of them unecessarily) and would be a fair bit of work for us to change all of them all at once in this one PR.

I just wondered if now we've given ourselves a whole bunch or warnings, I'm not sure someone would notice when introducing a new one. Given this would instigate people to remove them as they go, but just wondered if we could have skipped that step given that we are our own biggest clients and deprecation isn't always necessary as you would expect if a library or similar.

oliverlloyd:
> I wonder if a link in the code to the motivation (this PR)

Isn't the PR already linked to the code? Certainly in my IDE I can click through to the PR for each line in the code - this is why we write our motivations for things in PRs. I guess there's a risk additional commits will cover that up but in those cases I use blame to step back through time. My IDE makes this fairly painless.

oliverlloyd:
> > My guess would be that we do actually use React.FC in a lot of places (most of them unecessarily) and would be a fair bit of work for us to change all of them all at once in this one PR.
> 
> I just wondered if now we've given ourselves a whole bunch or warnings, I'm not sure someone would notice when introducing a new one. Given this would instigate people to remove them as they go, but just wondered if we could have skipped that step given that we are our own biggest clients and deprecation isn't always necessary as you would expect if a library or similar.

The reason for this PR is rooted in a comment by @cemms1 [on another PR](https://github.com/guardian/dotcom-rendering/pull/5599#issuecomment-1207909480). They rightly pointed out that when searching the codebase to try and decide which pattern to use they found both use cases so it wasn't clear which one should be adopted. This PR was designed to make that journey better by adding the warning.

I could have also done the work to refactor away all the existing use cases of `React.FC` but that's a lot more work which comes with risk. There's nothing intrinsically _wrong_ with using this type, so removing it doesn't solve any problems; it's just preferable that we be more consistent and we want to give a signal to people new to the codebase which one to pick.

jamesgorrie:
> Isn't the PR already linked to the code? Certainly in my IDE I can click through to the PR for each line in the code 

Assuming this is GitLens. I've generally just found it useful to say "Here's some code that might need extra context to know why it's here as it's not immediately obviously". 

> The reason for this PR is rooted in a comment by @cemms1 on another 

Makes tonnes of sense - I wondered if a person implemented this, if the warning would get lost in the already existing ones,  - but I was completely forgetting the IDE integration (again) which would flag this up straight away (I don't always use IDEs with eslint integration so rely on the overall test run, but this is an edge case).

PR #5623, opened 2022-08-08
refactor: create signup form component for better storybook mocking

cemms1:
## What does this change?

Puts the newsletter signup form into its own component

## Why?

For better readability and separation, and so that changes can be more easily be reflected in Storybook


dblatcher (approved):
Great idea - thanks for breaking that out!

Issue #5603, opened 2022-08-04
Live Blog cards metadata colour -> neutral.100


bryophyta:
**Live Blog cards** The kicker and all metadata needs to change to neutral.100, other than Labs.

- [x] DCR
- [ ] Frontend

_Part of an issue raised by @Bryony-Szekeres in https://github.com/guardian/dotcom-rendering/issues/4953#issuecomment-1139546312_

bryophyta:
In our storybook we currently have stories for liveblogs on some palettes which don't have dark backgrounds, and also aren't Labs. e.g. `LongRunningAltPalette`: https://www.chromatic.com/component?appId=5dfcbf3012392c0020e7140b&csfId=layouts-palettes--long-running-alt-palette&buildNumber=14363&k=62eba7ec7afd5a3ca5789b26-1200-interactive-true&h=14&b=-2

Should we be converting the metadata colours for the dark palettes to `neutral.100` (and leaving the light background ones as they are)? Or is it the case that some of these palettes will never actually have liveblogs associated with them?

(cc. @Bryony-Szekeres, @oliverlloyd, @HarryFischer)

oliverlloyd:
> In our storybook we currently have stories for liveblogs on some palettes which don't have dark backgrounds, and also aren't Labs. e.g. `LongRunningAltPalette`: https://www.chromatic.com/component?appId=5dfcbf3012392c0020e7140b&csfId=layouts-palettes--long-running-alt-palette&buildNumber=14363&k=62eba7ec7afd5a3ca5789b26-1200-interactive-true&h=14&b=-2
> 
> Should we be converting the metadata colours for the dark palettes to `neutral.100` (and leaving the light background ones as they are)? Or is it the case that some of these palettes will never actually have liveblogs associated with them?
> 
> (cc. @Bryony-Szekeres, @oliverlloyd, @HarryFischer)
> 
> <img alt="image" width="777" src="https://user-images.githubusercontent.com/37048459/182867964-cbb2bbc9-5633-417c-841e-d47f1639fccf.png">

When you put any card inside a container that has a 'Palette' (e.g.: `containerPalette`) applied to it, then that card has its normal colours overridden by the special palette, regardless of the type of card.

bryophyta:
[...]

Issue #4953, opened 2022-05-20
Accessibility colour improvements for DCR cards


HarryFischer:
Culture kicker fails currently

Bryony-Szekeres:
The old and new colours are laid out side-by-side in this Figma file:
https://www.figma.com/proto/5CfbWOeZPRi15lXBD5u1rW/Card-layout-system?page-id=0%3A1&node-id=45%3A43534&viewport=-12253%2C897%2C0.39&scaling=min-zoom

This is a summarised list of the things that need to change:
**All Opinion cards** The opinion.400 colour is being changed within Source to #C74600 which will make the text accessible
**All Culture cards** The culture.350 colour is being changed within Source to #866D50 which will make the text accessible
**Standard Labs card** The labs.400 colour needs to be replaced with the new labs.200 #0C7A73
**Live Blog cards** The kicker and all metadata needs to change to neutral.100, other than Labs
**Media cards** Make the metadata neutral.86 across all variants
**Special Report Comment card** Replace opinion.550 with brandAlt.400 and change the multiline colour to neutral.46
**Special Report cards** Make the metadata neutral.86 across all variants
**Metadata across all cards** should be either neutral.46 for pale cards, neutral.86 for dark cards, neutral.100 for live blog cards, and #606060 for some fronts palettes cards

bryophyta:
I'm happy to pick this up, assuming that no-one's already actively working on it?

bryophyta:
> Standard Labs card The labs.400 colour needs to be replaced with the new labs.200 #0C7A73

This colour is currently being set by the default fall-through of using `pillarPalette[format.theme].main`:

https://github.com/guardian/dotcom-rendering/blob/e6810b910f1cb656d75100af5239e9b108e38e04/dotcom-rendering/src/web/lib/decidePalette.ts#L468-L469

So it could be resolved by updating [the value used for `main` for labs](https://github.com/guardian/dotcom-rendering/blob/e6810b910f1cb656d75100af5239e9b108e38e04/dotcom-rendering/src/lib/pillars.ts#L128), but these palettes are deprecated, so it might be better to refactor and add an explicit case for Labs kickers.

@oliverlloyd it looks like you were involved in deprecating `pillars.ts` -- is there any context you can provide on how we should approach it moving forwards for refactors like this?

oliverlloyd:
> is there any context you can provide on how we should approach it moving forwards for refactors like this?

[...]

PR #5624, opened 2022-08-08
Use `neutral.100` for liveblog cards metadata

bryophyta:
## What does this change?

Uses `neutral.100` as the colour for liveblog cards metadata, unless the card is a Labs card, or part of a container with a `containerPalette` applied.

## Why?

This improves the contrast between text and background, which is important for accessibility. This is the DCR part of closing #5603 and contributes to #4953.

## Screenshots

**Query:** Without the colour difference, there's not so much variation between the kicker and the standfirst for sublinked stories (e.g. at the bottom of the Sport liveblog card, pictured below). Is this okay? Or do we need to think about adding a `/` here, or other visual cue?

| Before      | After      |
|-------------|------------|
| image| image |
|image|image|
|image|image|

### DIFF
--- a/dotcom-rendering/src/web/lib/decidePalette.ts
+++ b/dotcom-rendering/src/web/lib/decidePalette.ts
@@ -473,21 +473,10 @@ const textCardFooter = (format: ArticleFormat): string => {
 			}
 		case ArticleDesign.LiveBlog:
 			switch (format.theme) {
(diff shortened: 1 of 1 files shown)


AshCorr (approved):
Looks good! I'm not sure on the sublink query, to me the bold style makes is distinct enough but perhaps a question for UX.

PR #5619, opened 2022-08-08
Resize key event carousel button

silvacb:
## What does this change?
Decreases the Key Event Carousel button.

## Why?
Design change.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: 
https://user-images.githubusercontent.com/55602675/183386318-5720401b-203d-4d90-8125-da26f7036b76.png

[after]: https://user-images.githubusercontent.com/55602675/183386403-8cfb4165-6cc1-4402-856f-62e75cd3ff59.png

PR #5625, opened 2022-08-08
Fix padding in `KeyEventCard`

silvacb:
## What does this change?
Changes the `padding-bottom` to 12px for desktop.

## Why?
As part of the redesign, the Key Events Carousel buttons have been resized (#5619) and thus the padding between them and the carousel cards had to be reduced.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: 
https://user-images.githubusercontent.com/55602675/183416034-8b8facaf-1bd6-4ddd-9c2f-f725fdc63dec.png

[after]: 
https://user-images.githubusercontent.com/55602675/183452355-7c864499-e321-4545-a3d6-0d3319ab0991.png

### DIFF
--- a/dotcom-rendering/src/web/components/KeyEventCard.tsx
+++ b/dotcom-rendering/src/web/components/KeyEventCard.tsx
@@ -71,7 +71,7 @@ const listItemStyles = (palette: Palette) => css`
 	scroll-snap-align: start;
 
 	${from.desktop} {
-		padding-bottom: ${space[5]}px;
+		padding-bottom: 13px;
 		background-color: ${palette.background.keyEventFromDesktop};
 		width: 200px;
 		padding-right: ${space[5]}px;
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/components/KeyEventCard.tsx
```diff
@@ -71,7 +71,7 @@ const listItemStyles = (palette: Palette) => css`
 	scroll-snap-align: start;
 
 	${from.desktop} {
-		padding-bottom: ${space[5]}px;
+		padding-bottom: 13px;
```
marjisound, line 74: If you use `${space[3]}px;` would it look good enough? Cause I think it's usually recommended to use these spaces if possible. And I know spece[3] would result is 12 but I think maybe 1 px wouldn't be much. @oliverlloyd do you have any advice on this?

silvacb: Sure, I don't think it would be an issue if we used `${space[3]}px;`.

abeddow91: +1 to this comment; definitely use source here if we can!

silvacb: Thanks @marjisound and @abeddow91, this has now been updated.

marjisound (approved):
great work 🥳

PR #5505, opened 2022-07-26
feat(#5136): Add support for fixed/small/slow-I

AshCorr:
## What does this change?

Add support for fixed/small/slow-I

Closes #5136 

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |
| ![before2][] | ![after2][] |

[before]: https://user-images.githubusercontent.com/21217225/183670449-1f0d13fe-2c08-4172-b5d8-5f6f03243bf6.png
[after]: https://user-images.githubusercontent.com/21217225/183670547-623f18a1-193c-41e4-a445-8196d40e6183.png

[before2]: https://user-images.githubusercontent.com/21217225/183667960-e8b9e991-890a-4b6f-86f7-021d3641e0a9.png
[after2]: https://user-images.githubusercontent.com/21217225/183667863-c916c2d5-547e-41ad-b7e3-a7b5bd60d33d.png

PR #5626, opened 2022-08-08
Adds `headlineSizeOnMobile` parameter to Cards

AshCorr:
## What does this change?

Allows manually setting the mobile headline size on Cards.

## Why?

On certain front Containers such as `fixed/small/slow-I` some Cards don't follow the usual headline logic of using the next size lower on Mobile.

Part of #5505

## Screenshot

image

(We don't have a mobile size for sizes under Medium)

### DIFF
--- a/dotcom-rendering/src/web/components/CardHeadline.tsx
+++ b/dotcom-rendering/src/web/components/CardHeadline.tsx
@@ -28,6 +28,7 @@ type Props = {
 	showSlash?: boolean;
 	showQuotes?: boolean; // Even with design !== Comment, a piece can be opinion
 	size?: SmallHeadlineSize;
+	sizeOnMobile?: SmallHeadlineSize;
 	byline?: string;
 	showByline?: boolean;
 	showLine?: boolean; // If true a short line is displayed above, used for sublinks
@@ -48,12 +49,6 @@ const fontStyles = ({
 	switch (size) {
 		case 'ginormous':
 			return css`
-				${until.mobileLandscape} {
-					${headline.medium(options)};
-				}
-				${between.mobileLandscape.and.desktop} {
-					${headline.large(options)};
-				}
 				${from.desktop} {
 					${headline.large(options)};
 					font-size: 50px;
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/CardHeadline.tsx
```diff
@@ -252,6 +282,11 @@ export const CardHeadline = ({
 									? 'bold'
 									: 'regular',
 						  }),
+					format.theme !== ArticleSpecial.Labs &&
+						fontStylesOnMobile({
+							size: sizeOnMobile || size,
+							fontWeight: containerPalette ? 'bold' : 'regular',
+						}),
```
oliverlloyd, line 289: I could be confused here (very possible) but is it that we only want to set this if `sizeOnMobile` exists?

If so, could we guard against undefined on this before running the function so we don't need to fallback to `size`? A bit like:

```
						sizeOnMobile && fontStylesOnMobile({
							size: sizeOnMobile,
							fontWeight: containerPalette ? 'bold' : 'regular',
						}),
```

oliverlloyd (approved):
This makes sense. I think this follows on from `imagePositionOnMobile` where we can't predict these things and want more flexibility

dotcom-rendering/src/web/components/CardHeadline.tsx
AshCorr: I think we'd like to be able to fallback to size, it means that we don't need to change any of existing Components that use Card. In almost all cases we can just use the normal `size` for deciding what style should be applied on mobile, its only in some very few scenarios where we actually need to specify what `sizeOnMobile` should be.

oliverlloyd: Ah, I think I see what's happening here now. We _always_ want to set _both_ `fontStyles` **and** `fontStylesOnMobile`. But sometimes we override the `size` value when setting `fontStylesOnMobile`.

This all makes sense now 👍

oliverlloyd (approved):

PR #5630, opened 2022-08-09
Dblatcher/define apps rendering signup component

dblatcher:
## What does this change?
 - Adds a non-functional `NewsletterSignup` component with stories
 - Adds a new `BodyElement` type with a new value in the `ElementKind` enum to represent the component

## Why?
We want to replace the iframe-based email sign-up embeds with new sign up components. The feature is being trialled on DCR and we want to also implement it in AR. The DCR implementation is described on https://github.com/guardian/dotcom-rendering/pull/5410

This PR is to implement the design as an AR component. Expected follow up PR's:
 - getting newsletter data from MAPI
 - inserting the NewsletterSignup BodyElement into the article
 - Adding the sign-up functionality (May involve native UI - not determined yet)

## Screenshots

Screenshot 2022-08-09 at 09 55 30

### DIFF
--- a/apps-rendering/src/bodyElement.ts
+++ b/apps-rendering/src/bodyElement.ts
@@ -112,14 +112,12 @@ interface AudioAtom {
 
 interface NewsletterSignUp {
 	kind: ElementKind.NewsletterSignUp;
-	newsletter: {
-		id: string;
-		displayName: string;
-		frequency: string;
-		description: string;
-		group: string;
-		theme: string;
-	};
+	id: string;
+	displayName: string;
+	frequency: string;
+	description: string;
+	group: string;
+	theme: string;
 }
 
 type BodyElement =
--- a/apps-rendering/src/bodyElement.ts
+++ b/apps-rendering/src/bodyElement.ts
@@ -110,6 +110,18 @@ interface AudioAtom {
 	title: string;
 }
 
+interface NewsletterSignUp {
+	kind: ElementKind.NewsletterSignUp;
+	newsletter: {
+		id: string;
+		displayName: string;
+		frequency: string;
+		description: string;
+		group: string;
+		theme: string;
+	};
+}
+
 type BodyElement =
 	| Text
 	| Image
@@ -157,7 +169,8 @@ type BodyElement =
 	| ChartAtom
 	| AudioAtom
 	| KnowledgeQuizAtom
-	| PersonalityQuizAtom;
+	| PersonalityQuizAtom
+	| NewsletterSignUp;
 
 type Elements = BlockElement[] | undefined;
 
@@ -402,4 +415,5 @@ export {
 	TimelineAtom,
 	AudioAtom,
 	parseElements,
+	NewsletterSignUp,
 };
--- a/apps-rendering/src/bodyElement.ts
+++ b/apps-rendering/src/bodyElement.ts
@@ -5,6 +5,7 @@ import type { TimelineEvent } from '@guardian/atoms-rendering/dist/types/types';
 import type { Atoms } from '@guardian/content-api-models/v1/atoms';
 import type { BlockElement } from '@guardian/content-api-models/v1/blockElement';
 import { ElementType } from '@guardian/content-api-models/v1/elementType';
+import type { ArticleTheme } from '@guardian/libs';
 import type { Option, Result } from '@guardian/types';
 import {
 	err,
@@ -117,7 +118,7 @@ interface NewsletterSignUp {
 	frequency: string;
 	description: string;
 	group: string;
-	theme: string;
+	theme: ArticleTheme;
 }
 
 type BodyElement =
--- a/apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
+++ b/apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
@@ -44,10 +44,10 @@ const EmailSignupForm: FC<Props> = ({ newsletterId }) => {
 			/>
 			<form css={formStyle} action={undefined}>
 				<TextInput
-					type={'email'}
+					type="email"
 					width={30}
 					hideLabel
-					label={'Enter your email address'}
+					label="Enter your email address"
 					cssOverrides={css`
 						height: 2.25rem;
 						margin-right: ${remSpace[3]};
--- a/apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
+++ b/apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
@@ -58,8 +58,8 @@ const EmailSignupForm: FC<Props> = ({ newsletterId }) => {
 					title="Sign up"
 					cssOverrides={css`
 						background-color: ${neutral[0]};
-						margin-bottom: ${space[2]}px;
-						flex-basis: 118px;
+						margin-bottom: ${remSpace[2]};
+						flex-basis: ${pxToRem(118)}rem;
 						justify-content: center;
 					`}
 				>
--- a/apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
+++ b/apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
@@ -1,7 +1,12 @@
 // ----- Imports ----- //
 
 import { css } from '@emotion/react';
-import { neutral, pxToRem, remSpace, textSans } from '@guardian/source-foundations';
+import {
+	neutral,
+	pxToRem,
+	remSpace,
+	textSans,
+} from '@guardian/source-foundations';
 import { Button, Label, TextInput } from '@guardian/source-react-components';
 import type { FC } from 'react';
 
--- a/apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
+++ b/apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
@@ -1,0 +1,66 @@
+// ----- Imports ----- //
+
+import { css } from '@emotion/react';
+import { space, neutral, textSans } from '@guardian/source-foundations';
+import { Button, Label, TextInput } from '@guardian/source-react-components';
+import { NewsletterSignUp } from 'bodyElement';
+import type { FC } from 'react';
+
+// ----- Component ----- //
+
+interface Props {
+	newsletter: NewsletterSignUp['newsletter'];
+}
+
+const formStyle = css`
+	display: flex;
+	align-items: center;
+	flex-wrap: wrap;
+`;
+
+const EmailSignupForm: FC<Props> = ({ newsletter }) => {
+	const { id } = newsletter;
+	return (
+		<>
+			<Label
+				text="Enter your email address"
+				cssOverrides={css`
+					div {
+						${textSans.xsmall({ fontWeight: 'bold' })};
+					}
+				`}
+			/>
+			<form css={formStyle} action={undefined}>
+				<TextInput
+					type={'email'}
+					width={30}
+					hideLabel
+					label={'Enter your email address'}
+					cssOverrides={css`
+						height: 36px;
+						margin-right: ${space[3]}px;
+						margin-top: 0;
+						margin-bottom: ${space[2]}px;
+						flex-basis: 335px;
+					`}
+				/>
+				<Button
+					size="small"
+					title="Sign up"
+					cssOverrides={css`
+						background-color: ${neutral[0]};
+						margin-bottom: ${space[2]}px;
+						flex-basis: 118px;
+						justify-content: center;
+					`}
+				>
+					Sign up
+				</Button>
+			</form>
+		</>
+	);
+};
+
+// ----- Exports ----- //
+
+export default EmailSignupForm;
--- a/apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
+++ b/apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
@@ -23,7 +23,6 @@ const formStyle = css`
  * when implemented
  */
 const EmailSignupForm: FC<Props> = ({ newsletterId }) => {
-
 	const handleSubmit = (): void => {
 		console.log({ newsletterId });
 	};
--- a/apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
+++ b/apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
@@ -18,6 +18,11 @@ const formStyle = css`
 	flex-wrap: wrap;
 `;
 
+/**
+ * NOTE: this component is non functional and is for demonstration only.
+ * The UI for the NewsletterSignup might not use an HTML form for apps
+ * when implemented
+ */
 const EmailSignupForm: FC<Props> = ({ newsletter }) => {
 	const { id } = newsletter;
 	return (
--- a/apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
+++ b/apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
@@ -1,7 +1,7 @@
 // ----- Imports ----- //
 
 import { css } from '@emotion/react';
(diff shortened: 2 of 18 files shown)


apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+						${textSans.xsmall({ fontWeight: 'bold' })};
+					}
+				`}
+			/>
+			<form css={formStyle} action={undefined}>
+				<TextInput
+					type={'email'}
+					width={30}
+					hideLabel
+					label={'Enter your email address'}
```
JamieB-gu, line 48: Do this work?

```suggestion
					type="email"
					width={30}
					hideLabel
					label="Enter your email address"
```

apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+					cssOverrides={css`
+						height: 36px;
+						margin-right: ${space[3]}px;
+						margin-top: 0;
+						margin-bottom: ${space[2]}px;
+						flex-basis: 335px;
```
JamieB-gu, line 54: We use `rem`s in AR to make sure the UI scales evenly when the user increases their zoom level. I've demonstrated a few different ways to get `rem` values here:

- `remSpace` from Source, this is equivalent to `space`
- Write the `rem` value directly instead of the `px` value
- Use the `pxToRem` helper from Source

```suggestion
						height: 2.25rem;
						margin-right: ${remSpace[3]};
						margin-top: 0;
						margin-bottom: ${remSpace[2]};
						flex-basis: ${pxToRem(335)}rem;
```

apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+					`}
+				/>
+				<Button
+					onClick={handleSubmit}
+					size="small"
+					title="Sign up"
+					cssOverrides={css`
+						background-color: ${neutral[0]};
+						margin-bottom: ${space[2]}px;
+						flex-basis: 118px;
```
JamieB-gu, line 64:
```suggestion
						margin-bottom: ${remSpace[2]};
						flex-basis: ${pxToRem(118)}rem;
```

apps-rendering/src/bodyElement.ts
```diff
@@ -110,6 +110,18 @@ interface AudioAtom {
...
 
+interface NewsletterSignUp {
+	kind: ElementKind.NewsletterSignUp;
+	newsletter: {
+		id: string;
+		displayName: string;
+		frequency: string;
+		description: string;
+		group: string;
+		theme: string;
```
JamieB-gu, line 121: Is `theme` equivalent to `ArticleTheme` in the `Format` type, or does it just happen to have the same name?

apps-rendering/src/components/NewsletterSignup/PrivacyWording.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+import { css } from '@emotion/react';
+import { neutral, text, textSans } from '@guardian/source-foundations';
+import { Link } from '@guardian/source-react-components';
+import type { FC } from 'react';
+
+// ----- Component ----- //
+
+interface Props {
+	notUsingCaptcha?: boolean;
```
JamieB-gu, line 11: Does this prop have to be optional? Can it just be a required `boolean`?

If it does have to be optional we have a convention for using the `Option` type and `maybeRender` in AR. This helps to avoid the complexities that can come with implicit truthiness and the `&&` operator. Happy to talk through how these work if you'd like?

apps-rendering/src/components/NewsletterSignup/PrivacyWording.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+const termsStyle = css`
+	${textSans.xxsmall({ lineHeight: 'tight' })}
+	color: ${text.supporting};
```
JamieB-gu, line 16: I think `text.supporting` was deprecated in v4 of Source? The closest equivalent I can find is `neutral[46]`.

apps-rendering/src/components/NewsletterSignup/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+const containerStyles = (format: ArticleFormat): SerializedStyles => css`
+	clear: both;
+	border: ${neutral[0]} 3px dashed;
+	border-radius: ${space[3]}px;
+	margin-bottom: ${space[3]}px;
+	padding: ${space[2]}px;
+
+	${from.tablet} {
+		padding: ${space[2]}px ${space[3]}px;
+	}
```
JamieB-gu, line 37:
```suggestion
	border: ${neutral[0]} 0.1875rem dashed;
	border-radius: ${remSpace[3]};
	margin-bottom: ${remSpace[3]};
	padding: ${remSpace[2]};

	${from.tablet} {
		padding: ${remSpace[2]} ${remSpace[3]};
	}
```

apps-rendering/src/components/NewsletterSignup/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+`;
+
+const stackBelowTabletStyles = css`
+	display: flex;
+	flex-direction: column;
+	margin-bottom: ${space[2]}px;
+
+	${from.tablet} {
+		flex-direction: row;
+		margin-bottom: ${space[1]}px;
```
JamieB-gu, line 47:
```suggestion
	margin-bottom: ${remSpace[2]};

	${from.tablet} {
		flex-direction: row;
		margin-bottom: ${remSpace[1]};
```

apps-rendering/src/components/NewsletterSignup/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+	${headline.xxsmall({ fontWeight: 'bold' })}
+	flex-grow: 1;
+	span {
+		color: ${theme === 'news' ? sport[500] : 'inherit'};
+	}
+`;
+
+const descriptionStyles = css`
+	${textSans.xsmall({ lineHeight: 'tight' })}
+	margin-bottom: ${space[2]}px;
```
JamieB-gu, line 61:
```suggestion
	margin-bottom: ${remSpace[2]};
```

apps-rendering/src/components/NewsletterSignup/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+`;
+
+const iconHolderStyles = css`
+	display: flex;
+	align-items: center;
+	svg {
+		background-color: ${brandAlt[400]};
+		border-radius: 50%;
+		margin-right: ${space[2]}px;
```
JamieB-gu, line 70:
```suggestion
		margin-right: ${remSpace[2]};
```

apps-rendering/src/components/NewsletterSignup/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+	}
+`;
+
+// When in a row with the title, the Icon in the NewsletterFrequency
+// component should not affect the spaceing between the title text and
+// the description text, which should be 4px (space [1]).
+// When stacked below the title, there should be 8px (space[2]) between
+// the title and the Icon and then 8px between the Icon and the description
+const noHeightFromTabletStyles = css`
+	margin-top: ${space[2]}px;
```
JamieB-gu, line 84:
```suggestion
	margin-top: ${remSpace[2]};
```

apps-rendering/src/components/NewsletterSignup/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+	}
+`;
+
+/**
+ * NOTE: this component is non functional and is for demonstration only.
+ */
+const NewsletterSignup: FC<Props> = ({ format, element }) => {
+	const { displayName, frequency, description, theme } = element.newsletter;
+	return (
+		<aside css={containerStyles(format)}>
```
JamieB-gu, line 99: 👍 for the use of `aside`!

JamieB-gu (commented):
Nice job getting all of this working in AR so quickly! 🚀 

A few comments, mostly on platform conventions and specifically `px` vs `rem`. A couple of questions for my own understanding also 😅.

General question: do we have dark mode designs for this component?

dblatcher:
> General question: do we have dark mode designs for this component?

We don't yet, but I can ask for one to be done if we'll need it. Would that be something to add to this PR or a separate one?

apps-rendering/src/bodyElement.ts
```diff
@@ -110,6 +110,18 @@ interface AudioAtom {
...
+	};
```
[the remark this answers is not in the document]
dblatcher: It can be flattened - it might prove useful down the line to have a `newsletter` type with all the relevant datapoint from the API, but lets keep it flat for now.

apps-rendering/src/components/NewsletterSignup/EmailSignupForm.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { neutral, space, textSans } from '@guardian/source-foundations';
+import { Button, Label, TextInput } from '@guardian/source-react-components';
+import type { NewsletterSignUp } from 'bodyElement';
+import type { FC } from 'react';
+
+// ----- Component ----- //
+
+interface Props {
+	newsletter: NewsletterSignUp['newsletter'];
+}
```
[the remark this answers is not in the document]
dblatcher: will do - it might eventually need some extra data points depending on how the functionality ends up working, but they can also be separate props

apps-rendering/src/bodyElement.ts
dblatcher: I think it was intended to be equivalent to `ArticleTheme`, but this isn't formalised in the codebase for the newsletters API - so the value should be an `ArticleTheme`, but it wouldn't be safe to type it as one rather than as a `string`

apps-rendering/src/components/NewsletterSignup/PrivacyWording.tsx
dblatcher: It doesn't have to be optional - I'll make it required. It probably best as a required prop as it should be a conscious, explicit decision to include the captcha wording or not.

apps-rendering/src/components/NewsletterSignup/PrivacyWording.tsx
dblatcher: it doesn't seem to be marked as deprecated (but there is a warning in the docs about the colour contrast on some backgrounds)- maybe it was restored at some point?

JamieB-gu: > it doesn't seem to be marked as deprecated

It won't appear as deprecated yet in VSCode because I think the deprecation warning was added to a slightly later version of `source-foundations` than the one we're on in DCR/AR. However if you go to the docs for `source-foundations` it's in the list of deprecated colour tokens for v4.

apps-rendering/src/bodyElement.ts
JamieB-gu: Makes sense. In that case I think we can narrow the type when we parse the data coming from MAPI? That way we can give the `theme` field here in this body element a more accurate type. A function something like:

```ts
const themeFromString = (theme: string): ArticleTheme => {
  switch (theme) {
    case 'opinion':
      return ArticlePillar.Opinion;
    ...
    case 'news':
    default:
      return ArticlePillar.News;
  }
}
```

This is very similar to how we derive the main theme field in `Format`; we take the pillar `string` CAPI gives us and turn it into an `ArticleTheme`. In fact we can probably make that function a bit more generic so we don't duplicate the logic.

apps-rendering/src/components/NewsletterSignup/PrivacyWording.tsx
dblatcher: 👍  thanks, I'll change to neutral[46] - https://guardian.github.io/source/?path=/docs/packages-source-foundations-palette--palette

apps-rendering/src/bodyElement.ts
dblatcher: sounds good - using the ArticleTheme enum instead of a string now!

apps-rendering/src/components/NewsletterSignup/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+						<SvgNewsletter size="small" />
+						<b>{frequency}</b>
+					</div>
+				</div>
+			</div>
+
+			<p css={descriptionStyles}>{description}</p>
+
+			<EmailSignupForm newsletterId={id} />
+			<PrivacyWording notUsingCaptcha={false} />
```
georgeblahblah, line 117: similar point as above, I think this might read a bit more easily as `useCaptcha={true}`. Just an opinion though!

apps-rendering/src/components/NewsletterSignup/SvgNewsletter.tsx
```diff
@@ -0,0 +1,61 @@
...
+}) => (
+	<>
+		<Svg size={size} />
+		{isAnnouncedByScreenReader && (
+			<span
+				css={css`
+					${visuallyHidden}
+				`}
+			>
+				undefined
```
georgeblahblah, line 53: what does `undefined` do here?

georgeblahblah (commented):
This looks great! I had a comment around prop naming, as well as around capitalisation of `reCAPTCHA`

apps-rendering/src/components/NewsletterSignup/PrivacyWording.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+// ----- Component ----- //
+
+interface Props {
+	notUsingCaptcha: boolean;
+}
+
+const termsStyle = css`
+	${textSans.xxsmall({ lineHeight: 'tight' })}
+	color: ${text.supporting};
```
JamieB-gu, line 16:
```suggestion
	color: ${neutral[46]};
```

JamieB-gu (approved):
One suggestion, otherwise looks good!

apps-rendering/src/components/NewsletterSignup/PrivacyWording.tsx
```diff
@@ -0,0 +1,76 @@
...
+				<Link
+					data-ignore="global-link-styling"
+					href="https://www.theguardian.com/help/privacy-policy"
+					rel="noopener noreferrer"
+				>
+					Privacy Policy
+				</Link>
+				.
+			</span>{' '}
+			{!notUsingCaptcha && (
```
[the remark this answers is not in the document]
dblatcher: good point - I originally had this as an optional boolean prop and wanted the default behaviour to be to include the extra wording so it wasn't left off by mistake, but now its required,  `useCaptcha` is clearer and just as safe.

apps-rendering/src/components/NewsletterSignup/index.tsx
dblatcher: 👍

apps-rendering/src/components/NewsletterSignup/SvgNewsletter.tsx
dblatcher: It's the text of the visually hidden label that would be rendered if the `isAnnouncedByScreenReader` prop was set to true. 
I adapted this file from [source-components ](https://github.com/guardian/source/blob/main/packages/%40guardian/source-react-components/vendor/icons/SvgNewsletter.tsx) because the icon isn't available in the version of source used on dotcom-rendering yet. I'm guessing the text being "undefined" in source is a mistake - the icons components get code-genned by referencing a figma file, so maybe the screenreader name is missing. I'll raise that with the team looking after source.

apps-rendering/src/components/NewsletterSignup/PrivacyWording.tsx
dblatcher: sorry, I thought I'd changed that already!

PR #5632, opened 2022-08-09
Show `HeaderAdSlot` above front pages

### LINKED ISSUE #5512: `HeaderAdSlot` on fronts
Front pages render an ad slot at the top of the page above the header. This is very similar to the one seen on articles.

Screenshot 2022-07-27 at 09 57 45

oliverlloyd:
## What does this change?
This adds the `HeaderAdSlot` component above front pages

Closes #5512 

## Why?
💰


AshCorr (approved):

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -48,6 +50,21 @@ export const FrontLayout = ({ front, NAV }: Props) => {
...
+					<Stuck>
+						<ElementContainer
+							showTopBorder={false}
+							showSideBorders={false}
+							padded={false}
+							shouldCenter={false}
+						>
+							<HeaderAdSlot
+								isAdFreeUser={front.isAdFreeUser}
+								shouldHideAds={false}
```
AshCorr, line 62: Not entirely sure what this prop does, but it sounds like it shouldn't always be false? 
```
shouldHideAds={front.isAdFreeUser}
```

oliverlloyd: Good question.

`shouldHideAds` is a property that gets set on an article if the content is sensitive. It's false here because there's no equivalent for fronts

oliverlloyd: My next PR will refactor these props to make this read better

PR #5635, opened 2022-08-09
Cards: Only set `discussionId` if `isCommentable`

### LINKED ISSUE #5634: Remove un-populated comment count links from cards
The comment count links on cards for stories which don't have discussion enabled are selectable using the keyboard. They should be removed.

Image

bryophyta:
## What does this change?

Set the value of `discussionId` in the `enhanceCards` function for a
card only if `isCommentable` for that card is true.

This means that the `CommentCount` component will no longer be rendered
on cards which don't have a discussion to display.

There are some potential edge-cases where discussion might have been opened and
then closed, so there might be non-zero comment counts which we don't
display. However, because the discussion section on articles is also
controlled by `isCommentable`, the edge cases shouldn't be visible, and
if we did want to show comments and comment counts for these cases, this
seems like a separate issue.

## Why?

Empty `<a>` elements were being rendered on cards which didn't have discussion enabled. This will close #5634.

### DIFF
--- a/dotcom-rendering/src/model/enhanceCards.ts
+++ b/dotcom-rendering/src/model/enhanceCards.ts
@@ -162,7 +162,9 @@ export const enhanceCards = (
 						containerPalette,
 				  )
 				: undefined,
-			discussionId: faciaCard.discussion.discussionId,
+			discussionId: faciaCard.discussion.isCommentable
+				? faciaCard.discussion.discussionId
+				: undefined,
 			// nb. there is a distinct 'byline' property on FEFrontCard, at
 			// card.properties.byline
 			byline:


AshCorr:
I'm still a little bit confused about when we should show the comment count. On Frontend I see this section on [the music front](https://www.theguardian.com/music)

image

At the time of writing commenting is disabled on this article but it still displays the comment count on Fronts, I think based on the logic in this PR you shouldn't be able to see the comment count if commenting is disabled?

DanielCliftonGuardian (approved):
Nice! Looks good to me.

bryophyta:
> At the time of writing commenting is disabled on this article but it still displays the comment count on Fronts, I think based on the logic in this PR you shouldn't be able to see the comment count if commenting is disabled?

It's a good question, and I'm still not 100% sure I understand the interaction between the different props here, but for the Ferry article the reason it shows the 'closed for comments' message in the article is because `isClosedForComments` is set to `true` in the Front json:

image

Because this PR just looks at `isCommentable`, it does actually show the comment count for this card:

image

Does that answer your question?

AshCorr (approved):
> Does that answer your question?

Oh I see! So even though its closed it will still have isCommentable set to true... That helps, thanks!

bryophyta:
> So even though its closed it will still have isCommentable set to true

Yep, that's my understanding anyway, and `isClosedForComments` seems to be set by seeing whether the discussion expiry date is in the past:

https://github.com/guardian/frontend/blob/df34addb5177c39386c2e6de5421a21eacb3fa1e/common/app/model/trails.scala#L64

That said, there doesn't seem to be anything to enforce the constraint that only `isCommentable` discussions will have a discussion expiry date, because we do (occasionally) see all the possible combinations of `true/true`, `true/false`, `false/true`, `false/false` in these two properties.

PR #5643, opened 2022-08-09
Don't render ad slots at all if we're not serving them

oliverlloyd:
## What does this change?
Previously we passed the properties to decide if an advert should show or not down into the ad components and left it up to them to prevent the ad showing. This PR hoists this decision up such that the component won't even be rendered on the server.

This change only applies to the `top-above-nav`, `merchandising`, `merchandising-high` and `survey` slots.

## Why?
Like this we will send less code to the client


OllysCoding:
Would this impact shadypie and/or potentially have an impact on layouts expecting the structure of the ad slot to be there?

oliverlloyd:
> Would this impact shadypie and/or potentially have an impact on layouts expecting the structure of the ad slot to be there?

I skipped shadypie for this reason. I'm only applying this logic to the header, mechandising and survey slots. I should probably update the description with this information!

dotcom-rendering/src/web/layouts/CommentLayout.tsx
```diff
@@ -297,24 +297,25 @@ export const CommentLayout = ({ CAPIArticle, NAV, format }: Props) => {
 
 	const contributionsServiceUrl = getContributionsServiceUrl(CAPIArticle);
 
+	const renderAds = !CAPIArticle.isAdFreeUser && !CAPIArticle.shouldHideAds;
```
chrislomaxjones, line 300: Like in your PR description, is it worth adding a TSDoc comment to these `renderAds` values to make it clear they don't apply to all slots?

chrislomaxjones (approved):
Nice improvement 👍

dotcom-rendering/src/web/layouts/CommentLayout.tsx
bryophyta: I'd agree with this suggestion 👍 

Just for my own understanding, is it the case that the logic for whether ads should be shown is in fact captured by `renderAds`, but there are reasons for not using this particular variable to control all ads at the moment? (i.e. I don't actually know what shadypie is, but I'm assuming it follows the same logic of being shown only if `!CAPIArticle.isAdFreeUser && !CAPIArticle.shouldHideAds;`?

bryophyta (approved):
I second Chris's suggestion about adding a jsdoc on `renderAds`, but aside from that this looks good to me!

dotcom-rendering/src/web/layouts/CommentLayout.tsx
oliverlloyd: Shadypie is designed to fallback to a static subscription ask in the event that an ad blocker is detected. I think we could revisit these other slots on a case by case basis to think about if we want to render them or not. There's also some questions around how much of the commercial code we might be able to filter out (if we don't do this already). But I'd rather approach that in a separate thread so that we're being incremental here

chrislomaxjones: Nice summary, and thanks for adding the comment @oliverlloyd!

PR #5644, opened 2022-08-09
Refactor: Add focus styling to Fronts Cards

### LINKED ISSUE #5177: Focus styling for CardLinks
In the onwards carousel-type component on Fronts, there is no visible focus styling for the article cards, which makes keyboard navigation difficult. At the same time, using the keyboard one is able to highlight and click through to a `#comments` link for an article, even if comments are not enabled and there's no visible link. (Screenshot below is of the comments link being focused -- small blue square in the bottom right corner of the middle card.)

Image

- [x] Visible focus highlights should be added to the cards (existing hover styling is not very pronounced, so might be better to use something else, e.g. the focus ring from Source?).
- [x] #5634

bryophyta:
## What does this change?

- Add focusHalo to CardLink on focus.
- Remove unneeded styling from CardLink:
  - Styling of CardLink itself was redundant because opacity was set to 0.
  - The styling applied indirectly to `.image-overlay` was redundant because it
  didn't actually select anything (perhaps copied mistakenly from CardWrapper?)

## Why?

There was no visible focus indicator on these cards before. Closes #5177 

## Queries
I couldn't see what the code that I've deleted from `CardLink.tsx` was doing, but that doesn't mean that it wasn't doing anything! If anyone has more information, please let me know!

## Screenshots
Before (no visible styling when focused):
image

After: 
image

### DIFF
--- a/dotcom-rendering/src/web/components/Card/components/CardLink.tsx
+++ b/dotcom-rendering/src/web/components/Card/components/CardLink.tsx
@@ -10,29 +10,7 @@ const fauxLinkStyles = css`
 	right: 0;
 	bottom: 0;
 	left: 0;
-`;
-
-const baseLinkStyles = css`
-	display: flex;
-	/* a tag specific styles */
-	color: inherit;
-	text-decoration: none;
 	background-color: transparent;
-
-	/* The whole card is one link so we card level styles here */
-	width: 100%;
-
-	/* Sometimes a headline contains it's own link so we use the
-       approach described below to deal with nested links
-       See: https://css-tricks.com/nested-links/ */
-	:before {
-		content: '';
-		position: absolute;
-		left: 0;
-		top: 0;
-		right: 0;
-		bottom: 0;
-	}
 
 	:focus {
 		${focusHalo};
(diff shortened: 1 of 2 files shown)


oliverlloyd:
Ha, great spot! Looks like we had the same hover state code set in two places, here and - like you say - in `CardWrapper`. I have no idea how that came about but well done for digging into this and fixing it

dotcom-rendering/src/web/components/Card/components/CardLink.tsx
```diff
@@ -13,104 +10,46 @@ const fauxLinkStyles = css`
...
-		:before {
-			content: '';
-			position: absolute;
-			left: 0;
-			top: 0;
-			right: 0;
-			bottom: 0;
-		}
-
-		:hover .image-overlay {
```
bryophyta, line 42: this rule is also present in `CardWrapper`, and I couldn't see any scenario where there would be an element with `.image-ovoerlay` inside the `CardLink`, though I may have missed something?

bryophyta:
Thanks @oliverlloyd ! I've looked over it again and  removed a bit more css. I can't see any ways in which this affects the final rendering, but I'm not 100% sure -- would you mind casting another eye over the most recent commit at some point, please?

oliverlloyd (approved):
I checked this out locally and can verify that the hover states for both the whole card overlay and also the sublinks work as expected 👍 

Nice pruning work

bryophyta:
> Nice pruning work

Thanks! 🧑‍🌾 (TIL that there's aren't many good gardening emojis 😔 )

I'll merge tomorrow 👍

PR #5670, opened 2022-08-10
textCardFooter for media cards returns neutral.86

### LINKED ISSUE #5604: Card meta colours: Media
**Media cards** Make the metadata neutral.86 across all variants

The old and new colours are laid out side-by-side in this Figma file:
https://www.figma.com/file/5CfbWOeZPRi15lXBD5u1rW/?node-id=160%3A4026

- [ ] DCR
- [ ] Frontend

_Part of [this issue](https://github.com/guardian/dotcom-rendering/issues/4953#issuecomment-1139546312)_

bryophyta:
Co-authored by Daniel Clifton <daniel.clifton@theguardian.com>

## What does this change?

It changes the colour for footer text in media Fronts cards to `neutral.86`.

## Why?

Increase the contrast with the background, for accessibility purposes. Closes #5604 as part of #4953.

## Screenshots

| Before      | After      |
|-------------|------------|
| image|image|

### DIFF
--- a/dotcom-rendering/src/web/lib/decidePalette.ts
+++ b/dotcom-rendering/src/web/lib/decidePalette.ts
@@ -481,20 +481,7 @@ const textCardFooter = (format: ArticleFormat): string => {
 		case ArticleDesign.Gallery:
 		case ArticleDesign.Audio:
 		case ArticleDesign.Video:
-			switch (format.theme) {
-				case ArticleSpecial.SpecialReport:
-					return brandAlt[400];
-				case ArticlePillar.News:
-					return news[600];
-				case ArticlePillar.Sport:
-					return sport[600];
-				case ArticlePillar.Opinion:
-					return opinion[550];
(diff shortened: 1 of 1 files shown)


oliverlloyd (approved):
Looks good.

This is also being applied to the media icon but the designs call for this to keep the pillar colour
Screenshot 2022-08-10 at 17 30 02

I think this is fine in the context of this change but maybe we will need to create a new property `cardMeta` at a later stage to allow for these icons to have their own colour?

bryophyta:
> This is also being applied to the media icon but the designs call for this to keep the pillar colour

Ah good call -- I hadn't spotted that!

> maybe we will need to create a new property cardMeta at a later stage to allow for these icons to have their own colour?

Yep I agree that it probably makes sense to handle this in a different  PR, as it might require a bit of refactoring of other cards, too? I'll create an issue 👍

PR #5682, opened 2022-08-11
fix pinned post bug on twitter posts

marjisound:
## What does this change?
This PR fixes a bug where the pinned post `show more` button would have become hidden on some tweeter posts. 

# changes: 
1- We were using the `MutationObserver` to observe the `pinnedPost` element, but when I was testing locally, I realised that this observer is not really getting triggered on twitter blocks. Updating the config from `{ childList: true }` to `{ childList: true, subtree: true }`, fixed the triggering issue for the observer. 
-  childList: Set to true if additions and removals of the target node's child elements
-  subtree: Set to true if mutations to not just target, but also target's descendants are to be observed

2- This was not enough to fix the issue. Because for some twitter posts, at first, the scroll height of the component was less than the client height, so the `show more` button would get hidden. After twitter enhancement, the `MutationObserver` would get triggered, but in the call back we were only taking action for hiding the `show more` button, not for showing the hidden `show more` button. 

I refactored `hideShowMore` function into `toggleShowMore` so that it can now show the hidden button, depending on the boolean `show` parameter. 

## Screenshots

| Before      | After      |
|-------------|------------|
|  image | image |

### DIFF
--- a/dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
+++ b/dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
@@ -18,17 +18,19 @@ const pinnedPostContent: HTMLElement | null = !isServer
 	: null;
 
 /**
- * hide show more button and overlay on pinned post
+ * toggle show more button and overlay on pinned post
  */
-function hideShowMore() {
+function toggleShowMore(show: boolean) {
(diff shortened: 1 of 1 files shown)


oliverlloyd (approved):
Great to see this bug fixed, nice one!

Thanks for the great description on this PR. That really helped make this understandable.

PR #5688, opened 2022-08-11
Display dates & times in the edition's timezone

georgeblahblah:
## What does this change?

- Adds `edition` to `Item`
- Editionalises the dates & timestamps. This means a date or timestamp will be displayed in the timezone of the chosen edition, with the relevant timezone displayed. If an edition is not available, it defaults to UK
- Removes the client side scripts for updating elements matching `.js-date` or `time[data-date]` to local time

### UK

| Before | After |
| - | - |
| Screenshot 2022-08-18 at 12 09 20 | Screenshot 2022-08-18 at 12 08 58 |

### AU

| Before | After |
| - | - |
| Screenshot 2022-08-18 at 12 10 35 | Screenshot 2022-08-18 at 12 10 11 |


apps-rendering/src/components/Dateline/index.tsx
```diff
@@ -17,6 +18,7 @@ import { darkModeCss as darkMode } from 'styles';
 interface Props {
 	date: Option<Date>;
 	format: ArticleFormat;
+	edition?: Edition;
```
georgeblahblah, line 21: ideally, this would not be optional (as we set a default value in `server/page.tsx`). Setting it as optional here means we don't have to pass `edition` to every instance of this component yet

apps-rendering/src/components/Dateline/index.tsx
```diff
@@ -75,18 +78,14 @@ const getDatelineStyles = (format: ArticleFormat): SerializedStyles => {
...
-			<time
-				css={getDatelineStyles(format)}
-				data-date={d}
-				className="date js-date"
-			>
+			<time css={getDatelineStyles(format)}>
 				{format.design === ArticleDesign.LiveBlog ||
 				format.design === ArticleDesign.DeadBlog
-					? fullyFormatDate(d)
+					? datetimeFormat(edition ?? Edition.UK)(d)
```
georgeblahblah, line 88: again, this ideally would always have a value but it's optional for now so we don't have to pass an edition everywhere yet

apps-rendering/src/components/LastUpdated/index.tsx
```diff
@@ -21,10 +21,7 @@ const LastUpdated: FC<Props> = ({ lastUpdatedDisplay, lastUpdated }: Props) => (
 			`}
 		`}
 	>
-		<time
-			className="js-last-updated-time"
-			dateTime={lastUpdated.toISOString()}
-		>
+		<time dateTime={lastUpdated.toISOString()}>
```
georgeblahblah, line 24: I removed the `js-` class to prevent this value being overwritten by a client side script

apps-rendering/src/components/Dateline/index.tsx
```diff
@@ -75,18 +78,14 @@ const getDatelineStyles = (format: ArticleFormat): SerializedStyles => {
...
 
-const Dateline: FC<Props> = ({ date, format }) =>
+const Dateline: FC<Props> = ({ date, format, edition }) =>
 	pipe(
 		date,
 		map((d) => (
-			<time
-				css={getDatelineStyles(format)}
-				data-date={d}
-				className="date js-date"
```
JamieB-gu, line 85: Will removing this class also prevent the datetime being updated client-side for non-blog articles? Should we do one of the following:

1. Leave this class in place until we've migrated those articles over?
2. Migrate those articles over to the new format now as well?

apps-rendering/src/components/Layout/LiveLayout.tsx
```diff
@@ -114,9 +115,10 @@ const keyEvents = (blocks: LiveBlock[]): KeyEvent[] =>
 
 interface Props {
 	item: LiveBlog | DeadBlog;
+	edition: Edition;
```
JamieB-gu, line 118: Should `edition` live on the `Item` object? `Item` is a description of the article to be rendered, and I think the edition could be considered part of that description?

JamieB-gu (commented):
Looks good 👍, couple of suggestions.

Is the plan for those optional props (`?`) to become required by the end of this migration?

georgeblahblah:
> Is the plan for those optional props (`?`) to become required by the end of this migration?

@JamieB-gu yes! I wondered if it was too much for one PR, so I've opened another (#5766) which completes the migration and turns the props to required

apps-rendering/src/components/Dateline/index.tsx
georgeblahblah: Hmm, that's a good point. I'm leaning to option 2 - there isn't significant work to complete the entire migration. I've made the remaining changes in #5766 which could easily be merged here if it's not too much to review at once

apps-rendering/src/components/Layout/MediaLayout.tsx
```diff
@@ -59,6 +59,7 @@ const MediaLayout: FC<Props> = ({ item, children }) => {
 							publicationDate={item.publishDate}
 							className={articleWidthStyles}
 							item={item}
+							edition={item.edition}
```
JamieB-gu, line 62: If this already takes `item` does it also need to take `edition`?

JamieB-gu (approved):
Nice! One suggestion about `byline` props.

apps-rendering/src/components/Layout/MediaLayout.tsx
georgeblahblah: It doesn't! I'll remove the `edition` prop

PR #5694, opened 2022-08-12
`dynamic/package` - support snap

OllysCoding:
## What does this change?

This PR updates `dynamic/package` to support grouping.

`dynamic/package` supports 2 card groups, 'snap' and 'standard'. Currently it is written to assume all cards are 'standard', so this PR updates the container to also consume the 'snap' which is placed on top of all standard cards.

Only one 'snap' is supported, so additional snap grouped cards will fall through to the top of 'standards'

## Why?

We're on the way to ✨ _parity_ ✨

## Screenshots

| Frontend      | DCR Before      | DCR After   |

[frontend]: https://user-images.githubusercontent.com/9575458/184338959-74d8c294-e2ed-443c-9e1b-ca09b73471eb.png
[before]: https://user-images.githubusercontent.com/9575458/184339158-1b48724c-eb22-450d-91ff-a1e2888f69d7.png
[after]: https://user-images.githubusercontent.com/9575458/184339393-9c99bef6-7192-4fbe-b9e6-43157cdd2103.png

### DIFF
--- a/dotcom-rendering/src/web/components/Card/components/ContentWrapper.tsx
+++ b/dotcom-rendering/src/web/components/Card/components/ContentWrapper.tsx
@@ -31,15 +31,21 @@ const flexBasisStyles = ({
 			`;
 		case 'medium':
 			return css`
-				flex-basis: 50%;
+				${from.tablet} {
+					flex-basis: 50%;
+				}
 			`;
 		case 'large':
 			return css`
-				flex-basis: 34%;
+				${from.tablet} {
+					flex-basis: 34%;
+				}
 			`;
 		case 'jumbo':
 			return css`
-				flex-basis: 25%;
+				${from.tablet} {
+					flex-basis: 25%;
+				}
 			`;
 	}
 };
--- a/dotcom-rendering/src/web/components/DynamicPackage.tsx
+++ b/dotcom-rendering/src/web/components/DynamicPackage.tsx
@@ -19,7 +19,7 @@ const Snap100 = ({
 	showAge?: boolean;
 }) => {
 	return (
-		<UL direction="row">
+		<UL>
 			<LI padSides={true} padBottom={true}>
 				<FrontCard
 					trail={snap}
@@ -62,7 +62,7 @@ export const DynamicPackage = ({
 						showAge={showAge}
 					/>
 				)}
-				<UL direction="row">
+				<UL>
 					<LI padSides={true}>
 						<FrontCard
 							trail={primary}
--- a/dotcom-rendering/src/web/components/DynamicPackage.tsx
+++ b/dotcom-rendering/src/web/components/DynamicPackage.tsx
@@ -44,7 +44,7 @@ export const DynamicPackage = ({
 	showAge,
 }: Props) => {
 	// Take the first 'snap' - all others are treated as standards
-	const snaps = groupedTrails.snap.slice(0, 1);
+	const snap = [...groupedTrails.snap].shift();
 	const [primary, ...remaining] = [
 		...groupedTrails.snap.slice(1),
 		...groupedTrails.standard,
@@ -55,9 +55,9 @@ export const DynamicPackage = ({
 	if (primary.isBoosted) {
 		return (
 			<>
-				{snaps.length > 0 && (
+				{!!snap && (
 					<Snap100
-						snap={snaps[0]}
+						snap={snap}
 						containerPalette={containerPalette}
 						showAge={showAge}
 					/>
@@ -107,9 +107,9 @@ export const DynamicPackage = ({
 	}
 	return (
 		<>
-			{snaps.length > 0 && (
+			{!!snap && (
 				<Snap100
-					snap={snaps[0]}
+					snap={snap}
 					containerPalette={containerPalette}
 					showAge={showAge}
 				/>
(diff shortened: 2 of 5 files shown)


dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -1,24 +1,68 @@
...
+	snap: TrailType;
+	containerPalette?: DCRContainerPalette;
+	showAge?: boolean;
+}) => {
+	return (
+		<UL direction="row">
+			<LI padSides={true} padBottom={true}>
+				<FrontCard
+					trail={snap}
+					starRating={snap.starRating}
```
oliverlloyd, line 26: Is this needed? I have a feeling this falls through by default

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -1,24 +1,68 @@
...
+
 export const DynamicPackage = ({
-	trails,
+	groupedTrails,
 	containerPalette,
 	showAge,
 }: Props) => {
-	const [primary, ...remaining] = trails;
+	// Take the first 'snap' - all others are treated as standards
+	const snap = groupedTrails.snap.shift();
```
oliverlloyd, line 48: This may be fine but my gut tells me to avoid shift as it mutates the original array. Can this be rewritten somehow using `slice(0, 1)`?

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -1,24 +1,68 @@
...
+		...groupedTrails.snap,
+		...groupedTrails.standard,
+	];
 
+	// TODO: Different layouts are required when there are 4+ 'remaining' cards
+	// dynamic/package will display up to 8 'remaining' cards before moving to 'show more'
 	if (primary.isBoosted) {
 		return (
 			<>
+				{!!snap && (
```
oliverlloyd, line 59: I think this is probably fine for now and this is one for a later PR but I think this logic (er, that I wrote) could be improved by separating out the decision logic for what gets shown into a pre render step

dotcom-rendering/src/web/components/FrontCard.tsx
```diff
@@ -19,6 +19,7 @@ type Props = {
  * @param [imagePositionOnMobile] - Defaults to "left"
  * @param [imageSize] - Defaults to "medium"
  * @param [supportingContent] - Defaults to undefined, set to trail.supportingContent if you want this card to show sublinks.
+ * @param [trailText] - Defailts to undefined, set to trail.trailTrext if you want this card to show trail text.
```
oliverlloyd, line 22: 👍 nice

dotcom-rendering/src/web/components/Card/components/ContentWrapper.tsx
```diff
@@ -31,15 +31,21 @@ const flexBasisStyles = ({
...
+				${from.tablet} {
+					flex-basis: 34%;
+				}
 			`;
 		case 'jumbo':
 			return css`
-				flex-basis: 25%;
+				${from.tablet} {
+					flex-basis: 25%;
+				}
```
oliverlloyd, line 48: Thanks!

oliverlloyd (approved):
Yep. Aligning on `groupedTrails` like this really feel like the right way to go. That was a great abstraction you created before!

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -1,24 +1,68 @@
...
+	snap,
+	containerPalette,
+	showAge,
+}: {
+	snap: TrailType;
+	containerPalette?: DCRContainerPalette;
+	showAge?: boolean;
+}) => {
+	return (
+		<UL direction="row">
```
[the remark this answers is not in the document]
OllysCoding: image
Looks like it defaults to `column`?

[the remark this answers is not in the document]
oliverlloyd: Does it change how it renders though? I would have thought if there's only one item then... But if not, ho hum

dotcom-rendering/src/web/components/DynamicPackage.tsx
OllysCoding: Yeah definitely - I decided I wouldn't refactor the rendering logic for this PR, and leave that for a later one (where we cover the edge cases the container currently misses)

dotcom-rendering/src/web/components/DynamicPackage.tsx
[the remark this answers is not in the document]
OllysCoding: Tried it out & you're right - I'd copied it from elsewhere so will update both places :)

dotcom-rendering/src/web/components/DynamicPackage.tsx
OllysCoding: Had a fiddle around with this - I believe our tsconfig is set up so that typescript assumes any array access (e.g myArray[150]) will always assume that it is not undefined - not great for us. 

Because of this I moved to using .slice(0, 1) and keeping `snaps` as an array (even though we know it will only be 0 or 1 items). Not my favourite solution but not a lot of great options

oliverlloyd (approved):
Nice

dotcom-rendering/src/web/components/DynamicPackage.tsx
OllysCoding: Update: Used array deconstructing with .shift to work around this instead

PR #5695, opened 2022-08-12
Better behaviour around Braze switch

tjmw:
## What does this change?

When checking the dependencies for loading the Braze web SDK, check for the Braze API key before the switch.

## Why?

So that if the switch is ever turned off, we still have access to the API key in the caller in order to use the SDK wipe local data.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |


mxdvl (approved):
LGTM.

Just to be clear, turning the switch off should wipe all local user data?

tjmw:
Thanks @mxdvl

> Just to be clear, turning the switch off should wipe all local user data?

Yep that's right, the same behaviour as when a user logs out or removes consent 👍

PR #5699, opened 2022-08-12
Refactor container padding

oliverlloyd:
## What does this change?
This moves control for where we add padding to the bottom of containers from the internal content to the container itself

## Why?
Using the container is more semantic and provides more consistent results. Previously we were sometimes adding different levels of padding and sometimes none at all. Now we always have 36 pixels underneath each container.


oliverlloyd:
Superseded by https://github.com/guardian/dotcom-rendering/pull/5701

PR #5700, opened 2022-08-12
Dynamic Package refactor

OllysCoding:
This PR refactors the `dynamic/package` component to use conventions similar to that of https://github.com/guardian/dotcom-rendering/pull/5617

This builds on the use of **slices** seen in #5617:

A horizontal 'row' of cards can be described a a `slice`. This comes straight from Frontend & is helpful as it allows us to break down cards into more simple chunks to be consumed by rendering logic which is unaware of any upstream logic.

`dynamic/fast` can have _up to_ four slices, the first always being an option `snap`, with the remaining three being available for cards marked as 'standard'.

**'The file is so large!'**

A side effect of adopting the conventions proposed is that the rendering logic leans on the side of over-verbose. There would be room to abstract down some of these rendering-only components into multi-functioning forms that could do many different types of layouts - but the risk if that we create our own business logic that requires additional cognitive load & time to understand, while verbose, working on this component is very easy & approachable, and there are minimal side effects when making changes or fixing bugs! 

## Layouts

`dynamic/package` supports seven different layouts; each is variable as it can also support a single `snap` as demonstrated in: https://github.com/guardian/dotcom-rendering/pull/5694

This PR also introduces new layouts for when there are 1, 2 or 5+ standard cards - previously all of these would have fallen into the `fourOrLessStandards` layout. This brings us up to parity with Frontend, with any more than 9 cards being moved to 'show more'

**oneStandard**

image

**twoStandards**

> ⚠️ Note: I deliberately did not implement 'boosting' on the twoStandards layout. This is because it looks very broken on frontend - I imagine it was never used.

<details closed>
<summary>(How it looked boosted on frontend):</summary>
<br>
image

Because of how bad it looks - boosting simply will not work on this layout. I've added a comment in the code to signal this.
</details>

image

**threeOrFourStandards**

Three standards:
image

Four standards:
image

**threeOrFourStandardsBoosted**

Three standards:
image

Four standards:
image

**fiveStandards**

image

**sixStandards**

image

**sevenStandards**

image

**eightStandards**

image

**nineStandards**
image

each of `[five-nine]Standards` layouts supports the first standard being boosted, adding a larger headline font:
image

## Why this way?

This convention for writing dynamic components helps separate out the business logic & rendering - and breaks down the rendering into reusable chunks which are approachable for development.

With it already used for `dynamic/slow-mpu` - this PR stands almost as a proposal to use it for _all_ dynamic container types - avoiding a segmented world where each dynamic container has its own style of logic & rendering, as well as hopefully providing the most simple & approachable solution we can currently find for working on these components

### DIFF
--- a/dotcom-rendering/src/web/components/DynamicPackage.tsx
+++ b/dotcom-rendering/src/web/components/DynamicPackage.tsx
@@ -70,6 +70,45 @@ const Card100 = ({
 	);
 };
 
+const Card75_Card25 = ({
+	cards,
+	containerPalette,
+	showAge,
+}: {
+	cards: TrailType[];
+	containerPalette?: DCRContainerPalette;
+	showAge?: boolean;
+}) => {
+	if (cards.length < 2) return null;
+
+	return (
+		<UL direction="row">
+			<LI padSides={true} padBottomOnMobile={true} percentage="75%">
+				<FrontCard
+					trail={cards[0]}
+					containerPalette={containerPalette}
+					containerType="dynamic/package"
+					showAge={showAge}
+					headlineSize="large"
+					headlineSizeOnMobile="huge"
+					imagePosition="right"
+					imagePositionOnMobile="bottom"
+					imageSize="medium"
+					trailText={cards[0].trailText}
+				/>
+			</LI>
+			<LI padSides={true} padBottomOnMobile={false} percentage="25%">
+				<FrontCard
+					trail={cards[1]}
+					containerPalette={containerPalette}
+					containerType="dynamic/package"
+					showAge={showAge}
+				/>
+			</LI>
+		</UL>
+	);
+};
+
 const Card25_Card25_Card25_Card25 = ({
 	cards,
 	containerPalette,
@@ -403,8 +442,10 @@ export const DynamicPackage = ({
 	showAge,
 }: Props) => {
 	let layout:
-		| 'primaryBoosted'
-		| 'fourOrLessStandards'
+		| 'oneStandard'
+		| 'twoStandards'
+		| 'threeOrFourStandardsBoosted'
+		| 'threeOrFourStandards'
 		| 'fiveStandards'
 		| 'sixStandards'
 		| 'sevenStandards'
@@ -456,21 +497,64 @@ export const DynamicPackage = ({
 			secondSlice = cards.slice(0, 1);
 			thirdSlice = cards.slice(1);
 			break;
+		case 2:
+			layout = 'twoStandards';
+			firstSlice = snaps;
+			secondSlice = cards.slice(0, 2);
+			break;
+		case 1:
+			layout = 'oneStandard';
+			firstSlice = snaps;
+			secondSlice = cards.slice(0, 1);
+			break;
 		default:
 			if (cards[0].isBoosted) {
-				layout = 'primaryBoosted';
+				layout = 'threeOrFourStandardsBoosted';
 				firstSlice = snaps;
 				secondSlice = cards.slice(0, 1);
 				thirdSlice = cards.slice(1);
 			} else {
-				layout = 'fourOrLessStandards';
+				layout = 'threeOrFourStandards';
 				firstSlice = snaps;
 				secondSlice = cards;
 			}
 	}
 
 	switch (layout) {
-		case 'primaryBoosted':
+		case 'oneStandard':
+			return (
+				<>
+					<Snap100
+						snaps={firstSlice}
+						containerPalette={containerPalette}
+						showAge={showAge}
+					/>
+					<Card100
+						cards={secondSlice}
+						containerPalette={containerPalette}
+						showAge={showAge}
+					/>
+				</>
+			);
+		case 'twoStandards':
+			return (
+				<>
+					<Snap100
+						snaps={firstSlice}
+						containerPalette={containerPalette}
+						showAge={showAge}
+					/>
+					{/* Card75_Card25 does not support the first card being boosted - on Frontend the 75% card would
+						receive a ginourmous headline size - however this broke the layout visually, so we decided not
+						to support boosting for the twoStandards layout */}
+					<Card75_Card25
+						cards={secondSlice}
+						containerPalette={containerPalette}
+						showAge={showAge}
+					/>
+				</>
+			);
+		case 'threeOrFourStandardsBoosted':
 			return (
 				<>
 					<Snap100
@@ -490,7 +574,7 @@ export const DynamicPackage = ({
 					/>
 				</>
 			);
-		case 'fourOrLessStandards':
+		case 'threeOrFourStandards':
 			return (
 				<>
 					<Snap100
--- a/dotcom-rendering/src/web/components/DynamicPackage.tsx
+++ b/dotcom-rendering/src/web/components/DynamicPackage.tsx
@@ -405,12 +405,12 @@ export const DynamicPackage = ({
 }: Props) => {
 	let layout:
 		| 'primaryBoosted'
-		| '4orLessStandards'
-		| '5Standards'
-		| '6Standards'
-		| '7Standards'
-		| '8Standards'
-		| '9Standards';
+		| 'fourOrLessStandards'
+		| 'fiveStandards'
+		| 'sixStandards'
+		| 'sevenStandards'
+		| 'eightStandards'
+		| 'nineStandards';
 
 	const snaps = [...groupedTrails.snap].slice(0, 1);
 
@@ -425,29 +425,29 @@ export const DynamicPackage = ({
 	let thirdSlice: TrailType[] = [];
 	switch (cards.length) {
 		case 9: {
+			layout = 'nineStandards';
 			firstSlice = cards.slice(0, 1);
 			secondSlice = cards.slice(1, 5);
 			thirdSlice = cards.slice(5);
-			layout = '9Standards';
 			break;
 		}
 		case 8:
-			layout = '8Standards';
+			layout = 'eightStandards';
 			firstSlice = cards.slice(0, 1);
 			secondSlice = cards.slice(1);
 			break;
 		case 7:
-			layout = '7Standards';
+			layout = 'sevenStandards';
 			firstSlice = cards.slice(0, 1);
 			secondSlice = cards.slice(1);
 			break;
 		case 6:
-			layout = '6Standards';
+			layout = 'sixStandards';
 			firstSlice = cards.slice(0, 1);
 			secondSlice = cards.slice(1);
 			break;
 		case 5:
-			layout = '5Standards';
+			layout = 'fiveStandards';
 			firstSlice = cards.slice(0, 1);
 			secondSlice = cards.slice(1);
 			break;
@@ -457,8 +457,8 @@ export const DynamicPackage = ({
 				firstSlice = cards.slice(0, 1);
 				secondSlice = cards.slice(1);
 			} else {
+				layout = 'fourOrLessStandards';
 				firstSlice = cards;
-				layout = '4orLessStandards';
 			}
 	}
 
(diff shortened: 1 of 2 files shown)


dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -10,19 +11,20 @@ type Props = {
 };
 
 const Snap100 = ({
-	snap,
+	snaps,
 	containerPalette,
 	showAge,
 }: {
-	snap: TrailType;
+	snaps: TrailType[];
```
oliverlloyd, line 18: 👍 this isn't ideal _semantically_ but it is consistent

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -31,63 +33,132 @@ const Snap100 = ({
...
-					/>
-				)}
-				<UL>
-					<LI padSides={true}>
+	ginormousHeadline,
+}: {
+	cards: TrailType[];
+	containerPalette?: DCRContainerPalette;
+	showAge?: boolean;
+	ginormousHeadline?: boolean;
```
oliverlloyd, line 52: This kind of thing is so tricky. I guess you could have a `GinormousCard100` or maybe a `isBoosted` prop 🤷 

My vote would be to have a new component like `GinormousCard100` (we might want to shorten the biggness verbiage there 🤣 ) but I don't think there's a right or wrong here. It's true that once you add code it tends to be sticky but I think it's okay to try things and change them later.

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -97,79 +168,448 @@ export const DynamicPackage = ({
...
+							trail={card}
+							containerPalette={containerPalette}
+							containerType="dynamic/package"
+							showAge={showAge}
+							supportingContent={card.supportingContent}
+						/>
+					</LI>
+				);
+			})}
+			{columns.map((columnCards) => {
```
oliverlloyd, line 217: 👍 this reads nice

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -97,79 +168,448 @@ export const DynamicPackage = ({
...
+				return (
+					<LI
+						showDivider={true}
+						showTopMarginWhenStacked={true}
+						percentage="25%"
+					>
+						<UL direction="column">
+							{columnCards.map((card, cardIndex) => {
```
oliverlloyd, line 225: Nice

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -97,79 +168,448 @@ export const DynamicPackage = ({
...
+	cards,
+	containerPalette,
+	showAge,
+}: {
+	cards: TrailType[];
+	containerPalette?: DCRContainerPalette;
+	showAge?: boolean;
+}) => {
+	const bigs = cards.slice(0, 1);
+	const columns = [cards.slice(1, 3), cards.slice(3, 5), cards.slice(5, 7)];
```
oliverlloyd, line 268: This is a nice approach. It scales well like this

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -97,79 +168,448 @@ export const DynamicPackage = ({
...
+							})}
+						</UL>
+					</LI>
+				);
+			})}
+		</UL>
+	);
+};
+
+const Card75_ColumnOfX25 = ({
```
oliverlloyd, line 334:
```suggestion
const Card75_ColumnOfCards25 = ({
```

What do you think? 'X' is perhaps a clearer indication of 'any number' but somehow it's maybe nice to keep things wordy

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -97,79 +168,448 @@ export const DynamicPackage = ({
...
+	showAge,
+}: Props) => {
+	let layout:
+		| 'primaryBoosted'
+		| '4orLessStandards'
+		| '5Standards'
+		| '6Standards'
+		| '7Standards'
+		| '8Standards'
+		| '9Standards';
```
oliverlloyd, line 414:
```suggestion
		| 'fourOrLessStandards'
		| 'fiveStandards'
		| 'sixStandards'
		| 'sevenStandards'
		| 'eightStandards'
		| 'nineStandards';
```

I think my brain want this is because in some languages you can't start a variable using a numeric, is that true?

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -97,79 +168,448 @@ export const DynamicPackage = ({
...
+
+	const snaps = [...groupedTrails.snap].slice(0, 1);
```
oliverlloyd, line 416: 🤔 hmm, that pluralisation feels off here. I think I was pro pluralisation before but now I'm less sure

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -97,79 +168,448 @@ export const DynamicPackage = ({
...
+
+	let firstSlice: TrailType[] = [];
+	let secondSlice: TrailType[] = [];
+	let thirdSlice: TrailType[] = [];
+	switch (cards.length) {
+		case 9: {
+			firstSlice = cards.slice(0, 1);
+			secondSlice = cards.slice(1, 5);
+			thirdSlice = cards.slice(5);
+			layout = '9Standards';
```
oliverlloyd, line 432:
```suggestion
			layout = '9Standards';
			firstSlice = cards.slice(0, 1);
			secondSlice = cards.slice(1, 5);
			thirdSlice = cards.slice(5);
```

It's nice if the layout definition is always the first thing set

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -97,79 +168,448 @@ export const DynamicPackage = ({
...
+			secondSlice = cards.slice(1);
+			break;
+		default:
+			if (cards[0].isBoosted) {
+				layout = 'primaryBoosted';
+				firstSlice = cards.slice(0, 1);
+				secondSlice = cards.slice(1);
+			} else {
+				firstSlice = cards;
+				layout = '4orLessStandards';
```
oliverlloyd, line 462:
```suggestion
				layout = '4orLessStandards';
				firstSlice = cards;
```

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -97,79 +168,448 @@ export const DynamicPackage = ({
...
+
+	// We support up to 9 cards
+	const cards = [
+		...groupedTrails.snap.slice(1),
+		...groupedTrails.standard,
+	].slice(0, 9);
+
+	let firstSlice: TrailType[] = [];
+	let secondSlice: TrailType[] = [];
+	let thirdSlice: TrailType[] = [];
```
oliverlloyd, line 426: I think it would make more sense if we had first, second, third and forth slices and the first one is make equal to the `snap`. Visually when you look at what you get it's divided into slices, a snap is still a slice.

Also, and I'm not sure what the typescript impacts might be with this, but could these vars start as `undefined` and only get set if they're being used, that feels more aligned with what we end up doing with them (which is sometimes utterly nothing)

oliverlloyd (commented):
I've added a bunch of comments here but overall I just want to say that this is fantastic work! This container is, by a fair margin, the most complicated of them all but this PR really handles that super well

dotcom-rendering/src/web/components/DynamicPackage.tsx
OllysCoding: I like `GinormousCard100` a lot, but if we went for it, where would the logic for picking this sit? Would we create duplicate layouts for each no. of standards, e.g `'9StandardsBoosted' | '9Standards' | '8StandardsBoosted' | '8Standards' | ...`, or would we put the logic in the rendering section, e.g

```tsx
case '8Standards':
return <>
    ....
    {firstSlice[0].isBoosted ? <GinormousCard100 /> : <Card100>
</>
```

dotcom-rendering/src/web/components/DynamicPackage.tsx
OllysCoding: I just changed this because it didn't work but thankyou 😢

dotcom-rendering/src/web/components/DynamicPackage.tsx
OllysCoding: That is true - in js/ts you cannot start a variable using a numeric value, since this is a string we get away with it - but I'm happy to switch over because it does feel more readable anyway!

dotcom-rendering/src/web/components/DynamicPackage.tsx
OllysCoding: This can actually be just `groupedTrails.snap.slice(0,1)` I changed my mind a few times and apparently got mixed up

OllysCoding:
> This container is, by a fair margin, the most complicated of them all
dynamic/fast is still to come! 😫😫😫

Thank you for the feedback though - implementing a lot of these changes now

dotcom-rendering/src/web/components/DynamicPackage.tsx
oliverlloyd: Oh, does it actually switch here on if it is boosted or not? If so... `BoostedCard100`?

As for where, I think something like `isBoosted ? <BoostedCard100 /> : <Card100>` could work.

🤔 I also think you could have a `isBoosted: boolean` prop on `Card100`. That is maybe a more reasonable balance?

dotcom-rendering/src/web/components/DynamicPackage.tsx
OllysCoding: I agree on the snap as firstSlice thing - doing this now.

Unfortunately when I tried setting these as undefined, typescript wasn't smart enough to figure out that it would _always_ be defined because it was set in the first switch statement, so alas I fear it would make the jsx much more complicated

oliverlloyd: Cool, cool

dotcom-rendering/src/web/components/DynamicPackage.tsx
OllysCoding: Going to try the `isBoosted` props on `Card100`. I realise now that in `Card100` we could also do card.isBoosted check as well, but do we want that to happen implicitly for Card100?

oliverlloyd: > in Card100 we could also do card.isBoosted check as well

That makes sense to me. We check other properties on the card and vary what gets displayed (supportingContent / trailText) so adding this check feels valid

AshCorr (approved):
This is a great way of doing this!

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -456,21 +497,64 @@ export const DynamicPackage = ({
...
-		case 'primaryBoosted':
+		case 'oneStandard':
+			return (
+				<>
+					<Snap100
+						snaps={firstSlice}
+						containerPalette={containerPalette}
+						showAge={showAge}
+					/>
+					<Card100
```
AshCorr, line 532: I was actually waiting for this PR to be merged so that I could do this, but I think you've already done it, thanks!

This is fixes https://github.com/guardian/frontend/issues/25194 on DCR right?

OllysCoding: Yes it does :)

dotcom-rendering/src/web/components/DynamicPackage.tsx
mxdvl: You could possible have a tuple like `[TrailType]`?

PR #5701, opened 2022-08-12
Consistent container padding

oliverlloyd:
## What?
This PR ensures consistent spacing of 36px below each container on fronts

## Why?
Previously, we sometimes had different levels of padding below each container and sometimes none at all

### Before
Screenshot 2022-08-12 at 17 51 27

### After
Screenshot 2022-08-12 at 17 48 01

### DIFF
--- a/dotcom-rendering/src/web/layouts/FrontLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/FrontLayout.tsx
@@ -56,7 +56,7 @@ export const FrontLayout = ({ front, NAV }: Props) => {
 						<ElementContainer
 							showTopBorder={false}
 							showSideBorders={false}
-							padded={false}
+							padSides={false}
 							shouldCenter={false}
 						>
 							<HeaderAdSlot
@@ -70,7 +70,7 @@ export const FrontLayout = ({ front, NAV }: Props) => {
 					<ElementContainer
 						showTopBorder={false}
 						showSideBorders={false}
-						padded={false}
+						padSides={false}
 						backgroundColour={brandBackground.primary}
 						element="header"
 					>
@@ -92,7 +92,7 @@ export const FrontLayout = ({ front, NAV }: Props) => {
 						showSideBorders={true}
 						borderColour={brandLine.primary}
 						showTopBorder={false}
-						padded={false}
+						padSides={false}
 						backgroundColour={brandBackground.primary}
 						element="nav"
 					>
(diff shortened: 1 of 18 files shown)


dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -147,6 +147,14 @@ export const FrontLayout = ({ front, NAV }: Props) => {
 					// There are some containers that have zero trails. We don't want to render these
 					if (trails.length === 0) return null;
 
+					// This is a legacy container used to add palette styling on Frontend. DCR ignores it
+					if (
+						collection.displayName ===
+						'Palette styles new do not delete'
```
bryophyta, line 153: is there a case to be made  for filtering this  container  out  before `map`-ing  on the list? It might  enable  us to simplify some of the index logic below, e.g. https://github.com/guardian/dotcom-rendering/blob/d65f5aac0fc48e383065f931ee517c21f40d5e10/dotcom-rendering/src/web/layouts/FrontLayout.tsx#L183-L192

bryophyta (approved):
Nice! This looks good to me 👍 

Just a couple of non-blocking questions:
1. Would it  make sense  to filter out the 'padding styles' container before mapping the list of containers in FrontLayout? (See inline comment.)
2. Now that there's a `padBottom` prop on `ElementContainer` to use for fixed/thrasher, might it make sense to move the bottom padding on other containers into their ElementContainers too, instead of in the `<UL>` element? If we do want this padding on all fronts containers, it seems like  setting it in this one  place on FrontLayout  might be helpful for expressing that intention?

I don't think these should be taken as blocking considerations, though!

dotcom-rendering/src/web/layouts/FrontLayout.tsx
oliverlloyd: Nice idea! I think this makes a lot of sense. I think this would work well as an enhancer.

I've got another thrasher related PR pending so I will either put this change in there or do it after that one

oliverlloyd:
> * Would it  make sense  to filter out the 'padding styles' container before mapping the list of containers in FrontLayout? (See inline comment.)

I agree, this would be much better. WIll action in another PR

> * Now that there's a `padBottom` prop on `ElementContainer` to use for fixed/thrasher, might it make sense to move the bottom padding on other containers into their ElementContainers too, instead of in the `<UL>` element? If we do want this padding on all fronts containers, it seems like  setting it in this one  place on FrontLayout  might be helpful for expressing that intention?

I just landed a small (😉) refactor PR to this area so I think this will need revisiting but I essentially agree, we want this type of container level padding to be set at this level.

There might be some trickiness here however because we also need the internal border, the one that shows in-between the left col and the content, to reach the top and bottoms of the container. For this to work the padding has to exist inside the container, not on the wrapping element. But I could be misunderstanding things, lets see where we land

PR #5703, opened 2022-08-15
`Section`

oliverlloyd:
## What does this change?
This is a refactoring PR which does two main things, adding a `fullWidth` prop to `ContainerLayout` and then renaming it to `Section`.

In terms of how content appears on the page, this PR does nothing. This is evidenced by the lack of any visual diffs (other than where we have renamed a component so they got marked as new in Chromatic).

### 80 files changed?! 😢 
Yeah, there's a lot of changes here which I appreciate might look scary to review but I have been careful to make each separate change its own commit so I think if you use Github's commit by commit review feature this will hopefully make things easier.
Screenshot 2022-08-15 at 08 51 21

The truth is these components are just used in a lot of places so any changes to them will explode like this. I suspect you will find a lot of repetition here and can hopefully lean on Chromatic to provide assurance.

## Why?
By using `fullWidth`, we are able to rationalise `ContainerLayout` and `ElementContainer` into one abstraction.

Screenshot 2022-08-14 at 20 07 42

These two components have been a source of confusion so by bringing them together we help improve the developer experience for using DCR. `ElementContainer` continues to exist but is now only used in one place, inside `Section` (previously known as `ContainerLayout`).

Screenshot 2022-08-15 at 09 04 17

### Why 'Section'?
My first thought was to use 'Container' here as the general catch all name. But looking through Frontend and reading older PRs I learnt that `fc-container` was originally meant as a Facia Card Container; in addition, the term 'container' has a lot of associations to fronts. 'Section' is more generic and seems a better general term for what we're trying to do here, which is divide pages into horizontal sections.

#### But it isn't always an actual `section` tag
This is true, we default to `section` as the type of html tag to use for this component but we also allow it to be overridden, so using 'Section' as the overall name does not perfectly capture things but I think the word section can also be seen as an overall catch all for dividing content up.


common-rendering/src/components/Pagination.tsx
```diff
@@ -23,7 +23,7 @@ type Props = {
 	format: ArticleFormat;
 };
 
-const Container = ({ children }: { children: React.ReactNode }) => (
+const NavWrapper = ({ children }: { children: React.ReactNode }) => (
```
oliverlloyd, line 26: I wanted to make sure there were no other places where the names 'Container' and 'Section' were being used so there's a bunch of renaming commits included in this PR

dotcom-rendering/src/web/components/DynamicFast.stories.tsx
```diff
@@ -168,21 +138,15 @@ export const DF3VeryBigs3Big0StandardsSecondBoosted = () => (
...
 
 DF3VeryBigs3Big0StandardsSecondBoosted.story = {
 	name: '3 Very Bigs - 3 Big - 0 Standards (Second boosted)',
 };
 
 export const DF2VeryBigs3Big6StandardsThirdBoosted = () => (
-	<ContainerLayout
-		title="DynamicFast"
-		showTopBorder={true}
-		sideBorders={true}
```
oliverlloyd, line 182: These two props default to true so we're not carrying them over

dotcom-rendering/src/web/components/ElementContainer.tsx
```diff
@@ -75,8 +78,8 @@ export const ElementContainer = ({
 					id={sectionId}
 					css={[
 						shouldCenter && center,
-						showSideBorders && sideBorders(borderColour),
-						showTopBorder && topBorder(borderColour),
+						showSideBorders && sideBorderStyles(borderColour),
```
AshCorr, line 89: Cool drive by refactor :)

dotcom-rendering/src/web/components/Section.tsx
```diff
@@ -14,33 +14,89 @@ import { ShowHideButton } from './ShowHideButton';
...
+	/**
+	 * 🛠️ DEBUG ONLY 🛠️
+	 * Used to highlight the name of a container when DCR debug mode is enabled
+	 *
+	 * @see https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/browser/debug/README.md
+	 */
 	containerName?: string;
+	/** Fronts containers can have their styling overidden using a `containerPalette` */
 	containerPalette?: DCRContainerPalette;
+	/** Defautls to `false`. If true a Hide button is show top right allowing this section
```
AshCorr, line 77:
```suggestion
	/** Defaults to `false`. If true a Hide button is show top right allowing this section
```

dotcom-rendering/src/web/layouts/ImmersiveLayout.tsx
```diff
@@ -227,14 +229,15 @@ export const ImmersiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
...
+				<Section
+					fullWidth={true}
 					showTopBorder={false}
 					showSideBorders={false}
 					backgroundColour={decidePalette(format).background.article}
 					element="article"
 				>
 					<ImmersiveGrid>
-						{/* Above leftCol, the Caption is controled by ContainerLayout ^^ */}
+						{/* Above leftCol, the Caption is controled by Section ^^ */}
```
AshCorr, line 240:
```suggestion
						{/* Above leftCol, the Caption is controlled by Section ^^ */}
```

dotcom-rendering/src/web/layouts/InteractiveImmersiveLayout.tsx
```diff
@@ -300,14 +300,15 @@ export const InteractiveImmersiveLayout = ({
...
+				<Section
+					fullWidth={true}
 					showTopBorder={false}
 					showSideBorders={false}
 					backgroundColour={palette.background.article}
 					element="article"
 				>
 					<InteractiveImmersiveGrid>
-						{/* Above leftCol, the Caption is controled by ContainerLayout ^^ */}
+						{/* Above leftCol, the Caption is controled by Section ^^ */}
```
AshCorr, line 311:
```suggestion
						{/* Above leftCol, the Caption is controlled by Section ^^ */}
```

dotcom-rendering/src/web/components/Section.tsx
```diff
@@ -14,33 +14,90 @@ import { ShowHideButton } from './ShowHideButton';
 import { Treats } from './Treats';
```
OllysCoding, line 15: Could you add a comment here something like

```tsx
/**
 * ----- First time here? -----
 * If you're unsure how to use this component, check out `Section.stories.tsx` and
 * try running it in storybook to get an idea of how this component is used!
 */
```

I am conscious this component is very large and unwieldy - and giving future developers a small hint on how to start working with it might be quite helpful!

OllysCoding (approved):
+1 My browser crashed so many times trying to review this I've given up and approved it :p

(In all seriousness - this looks really good & well thought out, good work!)

dotcom-rendering/src/web/components/Section.tsx
oliverlloyd: Sounds like a good idea.

I'm also thinking it might be nice - at a later stage - to break things up into `FrontSection` (or maybe `Container`) and `Section` so that all the front specific logic can live by itself. I haven't really got my head into that though

PR #5704, opened 2022-08-15
Add support for the `culture-treat`

### LINKED ISSUE #5515: Culture thrasher (is it a thrasher?) incorrectly positioned
Screenshot 2022-07-27 at 10 11 29

oliverlloyd:
## What does this change?
This PR closes #5515 by adding support for the culture treat

## How?
We're achieving the goal of supporting this special thrasher by moving the contract that existed between the thrasher code and Frontend into DCR. This contract broadly states:

That each container on a front would use a `section` tag and that the `ul` element wrapping treats would have the class `treats__container`. In addition, video, thrasher and mobile ad slot containers are marked with `fc-container--thrasher `, `fc-container--video` and `fc-container__mpu--mobile` respectively so that they can be [ignored by the treats thrasher code](https://github.com/guardian/thrashers/blame/master/embeds/culture-nugget/_source/main.js#L4) when deciding where to insert the nugget.

### Is coupling code like this a good idea?
Ultimately, no. There have long been issues with this approach - I found [this one](https://github.com/guardian/frontend/pull/7044#issuecomment-62557428) from back in 2014 - and these problems will continue for as long as we continue to add brittle code couplings like this.

The problem is these sorts of dependencies are undocumented. Short of reading every line of thrasher code it's not possible to know what couplings there are or what might break so when making changes to the DCR platform it's very difficult to be sure you won't have external impacts.

### So what could we do instead?
The goal here is to be able to insert an image with some associated text and have that link out to a page. This image and text should sit at the bottom of the left column.

This requirement could be achieved via tools. A change to the fronts to could perhaps allow the image, text and url to be specified directly. Or perhaps it could allow a snap url to be given that returns the nugget. Either way, a tools change would allow this as an enhancement made to a specific container. This would be preferable to running code in a thrasher and expecting that to be able to find and insert content into another container.

### So why don't we implement the better solution?
Because changes to the tools is hard and it's easier for the Dotcom team to move forward with our migration when we're only changing code that we control 😐 

| Before      | After      |
|-------------|------------|
| Screenshot 2022-08-15 at 10 22 29 | Screenshot 2022-08-15 at 10 22 10 |

### DIFF
--- a/dotcom-rendering/src/web/components/ContainerLayout.tsx
+++ b/dotcom-rendering/src/web/components/ContainerLayout.tsx
@@ -41,6 +41,7 @@ type Props = {
 	showDateHeader?: boolean;
 	editionId?: EditionId;
 	treats?: TreatType[];
+	legacyClassName?: string;
 };
 
 const containerStyles = css`
@@ -145,6 +146,7 @@ export const ContainerLayout = ({
 	editionId,
 	containerName,
 	treats,
+	legacyClassName,
 }: Props) => {
 	const overrides =
 		containerPalette && decideContainerOverrides(containerPalette);
@@ -164,6 +166,7 @@ export const ContainerLayout = ({
 			ophanComponentName={ophanComponentName}
 			containerName={containerName}
 			innerBackgroundColour={innerBackgroundColour}
+			className={legacyClassName}
 		>
 			<Flex>
 				<LeftColumn
--- a/dotcom-rendering/src/web/components/Treats.tsx
+++ b/dotcom-rendering/src/web/components/Treats.tsx
@@ -55,7 +55,7 @@ export const Treats = ({
 								a later date after we've migrated, but for now a little bit
 								of evil is a good thing.
 							*/
-							padding-top: ${space[1]}px !important;
+							padding-top: ${space[1]}px !important; /* stylelint-disable-line declaration-no-important */
 							padding-left: ${space[2]}px;
 						`}
 					>
--- a/dotcom-rendering/src/web/components/Treats.tsx
+++ b/dotcom-rendering/src/web/components/Treats.tsx
@@ -10,13 +10,24 @@ export const Treats = ({
 	treats: TreatType[];
 	borderColour?: string;
 }) => {
-	if (treats.length === 0) return null;
 	return (
 		<ul
 			css={css`
 				display: flex;
 				flex-direction: column;
+				/*
+				   The legacy javascript executed by the culture-treat adds
+				   css to show a line below the li element, presumably because
+				   on Frontend this was needed. This css removes this line
+				   because we don't need it on DCR
+				*/
+				.treat-separator:before {
+					display: none;
+				}
 			`}
+			// The legacy javascript executed by the culture-treat depends
+			// on this class existing
+			className="treats__container"
 		>
 			{treats.map((treat) => {
 				return (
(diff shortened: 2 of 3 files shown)


OllysCoding:
Perhaps the most frustrating question to ask - but is there a world where we simply just don't support this in DCR? We're adding a decent amount of code here, enabling old patterns & adding complexity and cognitive load to our code.

We could (if we really wanted to) blacklist certain thrashers or ask editorial/visuals to remove them - and enforce a boundary that better solutions been to be found for this kind of visual. This would likely kickstart cross-platform work to resolve this if it's seen as urgent/important, and if not - should we be cluttering our code with it? 

Given the many, many ways there are to provide a 'clickable' piece of content there are on fronts - is this one we really need to carry forward?

p.s I feel I could expand on this for paragraphs on why this is a bad idea - but I'm sure we already know all this, so for the sake of documentation, if we do proceed with this implementation, it would be great to have written down for the future _why_ it was the best decision for us at the moment to do so!

AshCorr:
I don't think theres many places where we use the culture nugget, I wonder if we could hardcode/configure the model in Frontend and inject them into the nuggets part of our pressed front model. Atleast until we find the time / support we need to build this into the Fronts tool.

The way I kinda saw this working is that the fronts tool would let you add Snaps to the nugget slot of a container, if we can somehow manage to do this in the model it might be a good first step to getting it supported in the Fronts tool.

OllysCoding:
> I wonder if we could hardcode/configure the model in Frontend, similar to what we do for Badges

I would definitely be open to doing this if it allowed us to expand our model so that this doesn't need to be a snap/thrasher & instead we can position it with our own styling.

for example ``container.config.imageTreatUrl` = 'xxx'` - then it can be styled without all of this hassle - and we can push for it being installed upstream in tools later?

oliverlloyd:
> I don't think it ever changes all that often

The url for these netflix articles changes monthly

> > I wonder if we could hardcode/configure the model in Frontend, similar to what we do for Badges

Just so I understand, you're saying we create a model in Frontend where we pass down the image, url and text to use? And then changes to this are done in FE?

Would it also work if we made the change in DCR?

In general, I'm in favour of this solution. Having to create a PR once a month isn't ideal but it's not terrible.

> we can push for it being installed upstream in tools later?

This makes sense. If the workaround is still problematic then we have additional steps we can take to make this easier but I'm prefer taking small steps first and maybe that gets us far enough to be happy?

oliverlloyd:
> add Snaps to the nugget slot of a container,

Or maybe just add a new type of treat? Not sure how the model would end up looking but treats are just links so it makes sense for this image to sit alongside them (you might already have been suggesting this).

This has the nice bonus of also allowing for that Crossword image which also appears in a list of treats somewhere

AshCorr:
> The url for these netflix articles changes monthly

I think this change happens in the Snap/Thrasher itself though right? The snap link `https://interactive.guim.co.uk/thrashers/culture-nugget/source.json` would always stay the same. We wouldn't need to update the link every month as that would be done by Visuals in the HTML of the thrasher/snap itself.

> Just so I understand, you're saying we create a model in Frontend where we pass down the image, url and text to use? And then changes to this are done in FE?

I was suggesting that we just allow a Snap to be inserted into the nugget slot, although this could also work instead but would mean that we'd have to update the URL every month.

oliverlloyd:
@OllysCoding @AshCorr thanks for the comments, I'm going to close this PR (which I am happy to do) and work on a better solution. Based on a conversation with @AshCorr and @HarryFischer we think that we could replace this thrasher with an enhancer in DCR which has a hard coded list of these special image treats which it adds to certain containers based on name and edition.

Any changes to these nuggets would require a PR, similar to Badges, but the trade offs here feel right

PR #5716, opened 2022-08-16
`PLATFORM_TREATS`

oliverlloyd:
## What does this change?
This PR adds the `PLATFORM_TREATS` pattern which offers the ability to control treats from within the DCR platform.

To add a new treat you only need to edit the value of `PLATFORM_TREATS` in the [enhanceTreats.ts](https://github.com/guardian/dotcom-rendering/compare/oliver/platform-treats?expand=1#diff-e5ce3bd6173d3c49874329ef2365cee9f0305f48a9123489569cc746f284163e) file. Treats will be inserted in every container that has the same `displayName` value, you can optionally also filter further using `editionId` and `pageId`

If more than one link is passed in the `links` array then this will be shown underneath and the background colour will alternate between black and piller 300.

### What about that little Don't Miss arrow thing?
This arrow is an additional image which was positioned alongside the main one using the thrasher code. I'm removing this here for now as I don't think the complexity of supporting two images and deciding when to show the arrow image or not is justified. Perhaps we can put the arrow into the main image?

### What about image width, will it always be 130 pixels?
This width seems to work for the use cases I can see at the moment but if we need more flexibility in the future we could add `imageWidth` and `imageHeight` to `TreatType` and allow this to be specified

## Why?
This PR replaces #5704. Rather than trying to support the existing treats by adding brittle contracts this PR creates a platform based solution that does not have any external dependencies.

### Alternatives
Another solution would be to configure these treats using the fronts tool. This is preferable because it gives the greatest control and flexibility and avoids having to use `displayName` to target containers and avoids the need to merge PRs to make changes. The work in this PR should be seen a step towards this; in the event the tools were to offer this ability then so long as we keep the same model then no (or very few) changes would be needed in DCR

| Before      | After      |
|-------------|------------|
| Screenshot 2022-08-16 at 10 09 20 | Screenshot 2022-08-17 at 17 01 53 |

### DIFF
--- a/dotcom-rendering/src/model/enhanceCollections.ts
+++ b/dotcom-rendering/src/model/enhanceCollections.ts
@@ -4,10 +4,19 @@ import { enhanceCards } from './enhanceCards';
 import { enhanceTreats } from './enhanceTreats';
 import { groupCards } from './groupCards';
 
+const supportedContainers = (collection: FECollectionType) => {
+	switch (collection.displayName) {
+		case 'Palette styles new do not delete':
+			return false;
+		default:
+			return true;
+	}
+};
+
 export const enhanceCollections = (
 	collections: FECollectionType[],
 ): DCRCollectionType[] => {
-	return collections.map((collection) => {
+	return collections.filter(supportedContainers).map((collection) => {
 		const { id, displayName, collectionType } = collection;
 		const containerPalette = decideContainerPalette(
 			collection.config.metadata?.map((meta) => meta.type),
--- a/dotcom-rendering/src/model/enhanceCollections.ts
+++ b/dotcom-rendering/src/model/enhanceCollections.ts
@@ -4,23 +4,20 @@ import { enhanceCards } from './enhanceCards';
 import { enhanceTreats } from './enhanceTreats';
 import { groupCards } from './groupCards';
 
-const supportedContainers = (collection: FECollectionType) => {
-	switch (collection.displayName) {
-		case 'Palette styles new do not delete':
-		case 'culture-treat':
-		case 'newsletter treat':
-			return false;
-		default:
-			return true;
-	}
-};
+const FORBIDDEN_CONTAINERS = [
+	'Palette styles new do not delete',
+	'culture-treat',
+	'newsletter treat',
+];
+const isSupported = (collection: FECollectionType): boolean =>
+	!FORBIDDEN_CONTAINERS.includes(collection.displayName);
 
 export const enhanceCollections = (
 	collections: FECollectionType[],
 	editionId: EditionId,
 	pageId: string,
 ): DCRCollectionType[] => {
-	return collections.filter(supportedContainers).map((collection) => {
+	return collections.filter(isSupported).map((collection) => {
 		const { id, displayName, collectionType } = collection;
 		const containerPalette = decideContainerPalette(
 			collection.config.metadata?.map((meta) => meta.type),
--- a/dotcom-rendering/src/model/enhanceCollections.ts
+++ b/dotcom-rendering/src/model/enhanceCollections.ts
@@ -18,6 +18,7 @@ const supportedContainers = (collection: FECollectionType) => {
 export const enhanceCollections = (
 	collections: FECollectionType[],
 	editionId: EditionId,
+	pageId: string,
 ): DCRCollectionType[] => {
 	return collections.filter(supportedContainers).map((collection) => {
 		const { id, displayName, collectionType } = collection;
@@ -37,7 +38,12 @@ export const enhanceCollections = (
 			),
 			curated: enhanceCards(collection.curated, containerPalette),
 			backfill: enhanceCards(collection.backfill, containerPalette),
-			treats: enhanceTreats(collection.treats, editionId, displayName),
+			treats: enhanceTreats(
+				collection.treats,
+				displayName,
+				editionId,
+				pageId,
+			),
 			config: {
 				showDateHeader: collection.config.showDateHeader,
 			},
--- a/dotcom-rendering/src/model/enhanceCollections.ts
+++ b/dotcom-rendering/src/model/enhanceCollections.ts
@@ -7,6 +7,8 @@ import { groupCards } from './groupCards';
 const supportedContainers = (collection: FECollectionType) => {
 	switch (collection.displayName) {
 		case 'Palette styles new do not delete':
+		case 'culture-treat':
+		case 'newsletter treat':
 			return false;
 		default:
 			return true;
@@ -15,6 +17,7 @@ const supportedContainers = (collection: FECollectionType) => {
 
 export const enhanceCollections = (
 	collections: FECollectionType[],
+	editionId: EditionId,
 ): DCRCollectionType[] => {
 	return collections.filter(supportedContainers).map((collection) => {
 		const { id, displayName, collectionType } = collection;
@@ -34,7 +37,7 @@ export const enhanceCollections = (
 			),
 			curated: enhanceCards(collection.curated, containerPalette),
 			backfill: enhanceCards(collection.backfill, containerPalette),
-			treats: enhanceTreats(collection.treats),
+			treats: enhanceTreats(collection.treats, editionId, displayName),
 			config: {
 				showDateHeader: collection.config.showDateHeader,
 			},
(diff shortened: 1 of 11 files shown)


dotcom-rendering/src/model/enhanceTreats.ts
```diff
@@ -1,7 +1,64 @@
...
+
+const getPlatformTreats = (
+	editionId: EditionId,
+	displayName: string,
+): TreatType[] => {
+	return PLATFORM_TREATS.filter((treat) => {
+		// We decide if a treat should be shown for a container based on
+		// if the editionId matches and if the displayName text is the same.
+		return (
+			treat.displayName === displayName && treat.editionId === editionId
```
AshCorr, line 45: I'm not entirely sure this check will work as intended, I think with how its setup at the moment it would add the Culture treat to the first container of this [page](https://www.theguardian.com/uk/culture) which doesn't currently have a Culture treat in Frontend.

Could we use something like the PageId or the webUrl instead of editionId?

oliverlloyd: > I think with how its setup at the moment it would add the Culture treat to the first container of this [page](https://www.theguardian.com/uk/culture) which doesn't currently have a Culture treat in Frontend

This is a good point. I didn't think about this.

But maybe this is a good thing? That container is very similar to the Culture one on the main front and having treats show in all places where 'Culture' shows is arguably correct.

AshCorr: > But maybe this is a good thing? That container is very similar to the Culture one on the main front and having treats show in all places where 'Culture' shows is arguably correct.

Quite possibly! I suppose thats a question for editorial though!

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -143,26 +143,15 @@ export const FrontLayout = ({ front, NAV }: Props) => {
...
-						'Palette styles new do not delete'
-					) {
-						return null;
-					}
-
 					const ophanName = ophanComponentId(collection.displayName);
-					const ophanComponentLink = `container-${
-						index + 1
-					} | ${ophanName}`;
+					const ophanComponentLink = `container-${index} | ${ophanName}`;
```
AshCorr, line 154: Good catch!

AshCorr (approved):

dotcom-rendering/src/model/enhanceTreats.ts
OllysCoding: I'd definitely lean toward pageId/webUrl for the time being - otherwise the way these are controlled may end up being quite flakey

oliverlloyd: So replace editionId = "UK" with pageId = "https://www.theguardian.com/uk"?

This would tighten the targeting

OllysCoding: Does that work across environments, I wondered if there was a relative `/uk` string as part of the data somewhere?

oliverlloyd: After chatting with Harry, I'm proposing a more advanced solution: You can now choose to match on `containerName`, `editionId` OR `pageId`. This allows the culture treat to be shown for all Culture containers, if this is desired, or it can be targeted to one specific page

dotcom-rendering/src/model/enhanceCollections.ts
```diff
@@ -4,10 +4,23 @@ import { enhanceCards } from './enhanceCards';
...
+const supportedContainers = (collection: FECollectionType) => {
+	switch (collection.displayName) {
+		case 'Palette styles new do not delete':
+		case 'culture-treat':
+		case 'newsletter treat':
+			return false;
+		default:
+			return true;
+	}
+};
```
OllysCoding, line 16: Maybe tidier? 
```ts
const FORBIDDEN_CONTAINERS = ['Palette styles new do not delete',  'culture-treat', 'newsletter treat'];
const isSupportedContainer = (collection: FECollectionType): boolean => !FORBIDDEN_CONTAINERS.includes(collection.displayName)

```

oliverlloyd: This is much better - done 👍

dotcom-rendering/src/model/enhanceTreats.ts
```diff
@@ -1,7 +1,99 @@
...
+	});
+};
+
+export const enhanceTreats = (
+	treats: FEFrontCard[],
+	displayName: string,
+	editionId: EditionId,
+	pageId: string,
+): TreatType[] => {
+	const classicTreats = treats.map((treat) => ({
```
OllysCoding, line 78: `existingTreats`?

dotcom-rendering/src/model/enhanceTreats.ts
```diff
@@ -1,7 +1,99 @@
...
+		/**
+		 * We decide if a treat should be shown for a container based on the container
+		 * name, the edition and the page id.
+		 *
+		 * Matching on edition or page id is optional. If either of these are not
+		 * provided then we return true for that check
+		 */
+		const matchesContainer = treat.containerTitle === displayName;
+		const matchesEdition =
+			!treat.editionId || treat.editionId === editionId;
```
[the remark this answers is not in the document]
oliverlloyd: > As I understand this `editionId` check, a treat targeted to edition `US` & container `Spotlight` - then _all_ pages with 'Spotlight' will have this treat when the users edition is `US`.
> 

Yes. This is true. If you want to target the treat to _one specific page_ then you can also pass `pageId: '/us'` and that will ensure the treat only appears in this one place

dotcom-rendering/src/web/components/Treats.tsx
```diff
@@ -19,29 +111,61 @@ export const Treats = ({
...
+								/>
+							))}
+						</>
+					);
+				}
+
+				if (
+					treat.imageUrl &&
+					treat.altText &&
+					(treat.theme || treat.theme === 0) // ArticlePillar.News is zero
```
OllysCoding, line 142: `treat.theme !== undefined`?

dotcom-rendering/src/model/enhanceTreats.ts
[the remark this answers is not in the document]
OllysCoding: Thinking about this - the most likely people to encounter this issue or be confused by it is us (developers), CP & digital eds?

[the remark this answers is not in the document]
OllysCoding: Just to confirm, my main concern is that the we're not filtering by the edition of the page, but instead the edition of the user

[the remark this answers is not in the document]
oliverlloyd: > Just to confirm, my main concern is that the we're not filtering by the edition of the page, but instead the edition of the user

Ah, I see. I guess this same situation is true for all places where we switch by edition. If I *really* want to see content from another edition I can get to it, but any code decisions are done using `editionId`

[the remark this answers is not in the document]
OllysCoding: Could we add a comment to this effect on the platform treat definition? Just confirm it's targeting the edition of the user specifically - We should definitely be cautious of when this is used!

OllysCoding (approved):
+1 This is a nice solution :)

dotcom-rendering/src/model/enhanceTreats.ts
```diff
@@ -1,7 +1,95 @@
...
+		 *
+		 * If editionId is given without pageId it will match against all readers who
+		 * have that edition cookie, regardless of which page they are viewing
+		 */
+		const matchesContainer = treat.containerTitle === displayName;
+		const matchesEdition =
+			!treat.editionId || treat.editionId === editionId;
+		const matchesPage = !treat.pageId || treat.pageId === pageId;
+
+		return matchesContainer && matchesEdition && matchesPage;
```
mxdvl, line 71: Have you considered early returns? I find it to help break down logical statements.

```ts
if(treat.containerTitle !== displayName) return false;
// treat.editionId is optional
if(treat.editionId && treat.editionId !== editionId) return false;
// treat.pageId is optional
if(treat.pageId && treat.pageId !== pageId) return false;

return true;

```

PR #5717, opened 2022-08-16
Remove legacy key events container from live layout

abeddow91:
## What does this change?
Removes the key events container as the key events carousel is now production ready.  It also removes the switch to show the key events toggle in its old position on mobile and refactors the live blog renderer to remove unnecessary prop drilling and use the switches property instead.
## Why?
When we were testing the key events carousel in production, we kept the previous key events container in the live layout behind the key events carousel switch. This is no longer needed as the carousel is production ready.


oliverlloyd (approved):
This is great, thanks for keeping things tidy 🧹

PR #5527, opened 2022-07-27
Move overflow hidden to interscroller specific class

Jakeii:
## What does this change?
moves the `overflow: hidden` to a more specific classname.

## Why?

The fix for interscrollers stealing mouse events around them was too broad https://github.com/guardian/dotcom-rendering/pull/5422, tweak the interscroller creative template to add an interscroller specific class name to the slot.

PR #5719, opened 2022-08-16
Fix viewability tracking on interscroller native ads

Jakeii:
## What does this change?
Adds css to position the iframe absolutely within the slot.
`z-index: -1;` makes sure the ad is still clickable as the link is not in the iframe.

## Why?
Interscrollers do not display their content inside the iframe, elements are added outside the iframe but within the slot to show the actual ad.

The ad iframe is pushed down after actual interscroller ad content, which means that previously the viewability tracker would fire very late, and after recent changes in #5422 & #5527 would not fire at all.

Positioning the iframe absolutely behind the ad content fixes the viewability tracking.

This can be tested by watching for network requests to: `https://pagead2.googlesyndication.com/pcs/activeview...`

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |
| <video src="https://user-images.githubusercontent.com/1731150/184867991-a8115a08-b6c7-4930-8177-116345399274.mov" autoplay /> | <video src="https://user-images.githubusercontent.com/1731150/184868071-adc0b268-13d4-4932-96a4-a7f2f92dff7c.mov" autoplay /> |

[before]:  https://user-images.githubusercontent.com/1731150/184863058-ef64d464-fe1d-4a2a-8fc6-7197d38cabf0.png
[after]: https://user-images.githubusercontent.com/1731150/184863104-2f4af384-1da7-4437-b0d9-1cbee5331c4a.png

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleContainer.tsx
+++ b/dotcom-rendering/src/web/components/ArticleContainer.tsx
@@ -95,6 +95,19 @@ const adStyles = css`
 		.ad-slot--interscroller {
 			/* this fixes inter-scrollers stealing mouse events */
 			overflow: hidden;
+			position: relative;
+
+			/* position the iframe absolutely (relative to the slot) so that it is in the correct position to detect viewability */
+			.ad-slot__content {
+				position: absolute;
+				height: 100%;
+				left: 0;
+				top: 0;
+				right: 0;
+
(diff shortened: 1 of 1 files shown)


ashishpuliyel (approved):
Neat! Looks good.

PR #5728, opened 2022-08-16
AR analysis design tweaks

marjisound:
## What does this change?
This PR makes AR headline bold in Standard Analysis articles. 
It also changes the background colour for the Analysis articles. 

This is part of work for the Editorial Design team detailed here: https://docs.google.com/document/d/1MqSldxfgR86kOJaXTFNaBs5LzPzSa7jC4D4YwqqMFwg/edit# 

Note:
Background colour is always news[800] for light mode no matter which pillar. and is always the 1A1A1A for dark mode which is neutral[10]

TODO: 
The background colour of the tag items at the bottom of article also needs to change. 
## Why?
This is part of the changes requested by the Editorial Design team. 
## Screenshots

| Before      | After      |
|-------------|------------|
| image | image |

### DIFF
--- a/apps-rendering/src/components/Headline/Headline.defaults.tsx
+++ b/apps-rendering/src/components/Headline/Headline.defaults.tsx
@@ -12,9 +12,13 @@ import { articleWidthStyles, darkModeCss } from 'styles';
 
 export const defaultStyles = (format: ArticleFormat): SerializedStyles => {
 	const baseStyles = css`
-		${standardFont}
+		${headline.small()}
 		color: ${text.headline(format)};
 		margin: 0;
+
+		${from.tablet} {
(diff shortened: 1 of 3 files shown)


apps-rendering/src/components/Headline/Headline.defaults.tsx
```diff
@@ -49,6 +57,20 @@ interface DefaultProps {
 	styles: SerializedStyles;
 }
 
+export const standardFont = css`
```
abeddow91, line 60: think we can inline this inside baseStyles

abeddow91 (approved):
One non blocking change but otherwise looks great!

abeddow91 (approved):

PR #5729, opened 2022-08-16
AR: Create Analysis standfirst

abeddow91:
## What does this change?
Adds a new standfirst component for design.analysis with

- A title of `Analysis`
- Specific `analysis` font styling
- relevant story

## Why?
This is to align the AR templates with the new template designs from Editorial Design.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/20416599/184913152-e540cc19-8297-471a-aced-9f9ab3a328ee.png
[after]: https://user-images.githubusercontent.com/20416599/184913195-93415edd-8136-4229-9267-4993ffdf3479.png

### DIFF
--- a/apps-rendering/src/components/Standfirst/AnalysisStandfirst.tsx
+++ b/apps-rendering/src/components/Standfirst/AnalysisStandfirst.tsx
@@ -8,6 +8,12 @@ interface Props {
 	item: Item;
 }
 
+const analysisStandfirstStyles = css`
+	${headline.xxxsmall()}
+	p:first-child {
+		padding-top: 0;
+	}
+`;
 const AnalysisStandfirst: React.FC<Props> = ({ item }) => (
 	<>
 		<p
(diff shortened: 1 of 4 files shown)


apps-rendering/src/components/Standfirst/AnalysisStandfirst.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { headline } from '@guardian/source-foundations';
+import type { Item } from 'item';
+import { getFormat } from 'item';
+import DefaultStandfirst, { defaultStyles } from './Standfirst.defaults';
+
+interface Props {
+	item: Item;
+}
+
+const analysisStandfirstStyles = css`
```
georgeblahblah, line 11: I think we could probably call this `styles` as we're in a clearly named module which indicates Analysis & Standfirst

georgeblahblah (approved):
Looks good! I had one non-blocking comment about naming

georgeblahblah:
it does seem a bit strange to me to have this static text & (in the screenshot of this PR) "Analysis:" in the standfirst itself - but perhaps that's just the design

oliverlloyd:
> it does seem a bit strange to me to have this static text & (in the screenshot of this PR) "Analysis:" in the standfirst itself - but perhaps that's just the design

Yeah, that is pretty strange. And there might be other examples of this but the goal is we will update that standfirst text to remove any of these prefixes and prefer hard coding the text like in this PR

apps-rendering/src/components/Standfirst/AnalysisStandfirst.tsx
abeddow91: Makes sense, I'll update this :)

PR #5730, opened 2022-08-16
Dblatcher/apps darkmode newsletter signup

dblatcher:
## What does this change?
(no publicity visible changes)
Adds darkmode CSS to the (currently unused) `NewsletterSignup` apps component.

### Testing 
*(please let me know if there's a better way to do this!)*
To "force" storybook to showing the darkmode version remove lines 40 and 42 from "./apps-rendering/src/styles.ts"  as below then run storybook as normal:

``` tsx
export const darkModeCss = (
	styles: TemplateStringsArray,
	...placeholders: string[]
): SerializedStyles => {
	const darkStyles = styles
		.map((style, i) => `${style}${placeholders[i] ? placeholders[i] : ''}`)
		.join('');
	return css`

			${darkStyles}

	`;
};
```

## Screenshots

| Standard      | darkmode      |
|-------------|------------|
| Screenshot 2022-08-16 at 15 15 24 | Screenshot 2022-08-16 at 15 16 21 |


JamieB-gu (commented):
Looks good, one suggestion: can we consider moving these colours into `editorialPalette` and referencing them from there? You can see examples of this across the AR codebase and in DCR with `decidePalette`. I'm also happy to pair on this if you'd find that helpful?

> (please let me know if there's a better way to do this!)

There isn't unfortunately; lack of storybook/Chromatic coverage of dark mode is a known problem in AR. A proposal for a possible long-term solution can be found in #4241. Please feel free to comment on that issue if you have any thoughts or suggestions!

dblatcher:
> Looks good, one suggestion: can we consider moving these colours into `editorialPalette` and referencing them from there? You can see examples of this across the AR codebase and in DCR with `decidePalette`. I'm also happy to pair on this if you'd find that helpful?

I've pushed some changes to do this - I think I followed the pattern right, but do let me know if not.

common-rendering/src/editorialPalette/hover.ts
```diff
@@ -33,10 +34,15 @@ const pagination = (format: ArticleFormat): Colour => {
 	}
 };
 
+const signUpFormButton = (_format: ArticleFormat): Colour => neutral[7];
+const signUpFormButtonDark = (_format: ArticleFormat): Colour => '#d1d1d1';
```
JamieB-gu, line 38: Do you know why we're not able to use a Source colour for this? Is that a question for Design?

JamieB-gu (approved):
> I've pushed some changes to do this - I think I followed the pattern right, but do let me know if not.

Excellent job, that's exactly right! 💥

JamieB-gu (approved):

Issue #5645, opened 2022-08-09
Automate Publishing Of API Models To npm


JamieB-gu:
We have several other projects that publish npm packages automatically using different methods:

- atoms-rendering - using changesets
- libs - using semantic release
- cdk - using semantic release
- source - using changesets

DevX can help us understand the current best practice and set it up on this project.

We'll want releases of both packages to happen together and to remain in sync (version numbers etc.). We'll need to take into account that we're publishing more than just a JavaScript/TypeScript package. As a result, we may want to opt for tooling that can understand multiple languages and package managers (semantic release for example).

PR #5735, opened 2022-08-17
feat: add newsletter data to thrift

cemms1:
## What does this change?

Adds newsletter data to the thrift model for MAPI

## Why?

As part of a project to implement the new newsletter signup forms (embeds) in the apps, we need to allow the `RenderingRequest` from MAPI to expect the `promotedNewsletter` field which contains the data required to render the newsletter signup form.

### DIFF
--- a/apps-rendering/api-models/thrift/src/main/thrift/appsRendering.thrift
+++ b/apps-rendering/api-models/thrift/src/main/thrift/appsRendering.thrift
@@ -126,9 +126,7 @@ struct Newsletter {
     3: required string theme
     4: required string description
     5: required string frequency
-    6: required i32 listId
-    7: required string group
-    8: required string successDescription
+    6: required string successDescription
 }
 
 struct RenderingRequest {


apps-rendering/api-models/thrift/src/main/thrift/appsRendering.thrift
```diff
@@ -119,6 +119,19 @@ struct FootballContent {
     8: optional string venue
 }
 
+
+struct Newsletter {
+    1: required string identityName
```
georgeblahblah, line 124: out of interest, how will `identityName` be used in Apps Rendering? Is it the same as `listId`?

apps-rendering/api-models/thrift/src/main/thrift/appsRendering.thrift
```diff
@@ -119,6 +119,19 @@ struct FootballContent {
...
+    2: required string name
+    3: required string theme
+    4: required string description
+    5: required string frequency
+    6: required i32 listId
+    7: required string group
+	8: required string successDescription
```
georgeblahblah, line 131: I think there's something a bit off with the indentation here

apps-rendering/api-models/thrift/src/main/thrift/appsRendering.thrift
```diff
@@ -129,4 +142,5 @@ struct RenderingRequest {
     7: optional RelatedContent relatedContent
     8: optional FootballContent footballContent
     9: optional Edition edition
+	10: optional Newsletter promotedNewsletter
```
georgeblahblah, line 145: another indentation oddity!

georgeblahblah (commented):
Looks good! I had a question around the usage of some of the fields and I noticed some identation looks a bit off, but nothing major!

apps-rendering/api-models/thrift/src/main/thrift/appsRendering.thrift
georgeblahblah: Ok, makes sense! I think with `thrift`, it's a good idea to keep the model as small as possible. We can always add fields later if we need them, but it's not possible to remove fields in this case as they're marked as required.

georgeblahblah (approved):
Looks good! I think there are 2 related things:

1. the release process for this package is manual. Once this is merged to `main`, I will create tickets to release the packages, `npm` (which Apps Rendering will use) and maybe @frankie297 can release the Scala packages to Sonatype (which MAPI will use). We hope to make this package easier to release in the future (#5645, #5646)
2. there is a related type in AR which seems related, [here](https://github.com/guardian/dotcom-rendering/blob/main/apps-rendering/src/bodyElement.ts#L114), which might need updating based on the definition of this type in the thrift model. This can happen separately though as I don't think the two are connected yet!

PR #5737, opened 2022-08-17
a11y: Don't use an empty link to wrap Cards.

AshCorr:
## What does this change?

Adds text to the wrapper links used by Cards.

Tested on the [uk front](https://www.theguardian.com/uk) using https://wave.webaim.org/

## Why?

This link is positioned to cover the entire card so that you can click anywhere on the card and be linked to the article. Because its used like this visually there was no need to have any text in the link, however, any screen reader that tries to read the link will have issues because it can't find any meaningful text to explain where the link goes.

Alternatively we could add `aria-hidden` and `tabindex="-1"` to the Link but this doesn't seem to hide the errors from wave. Interestingly Frontend adds text to the link AND makes it aria-hidden.

https://webaim.org/standards/wcag/checklist#sc2.4.4

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/21217225/185096813-434b2c42-f155-4017-81ee-8e3f5880ae9e.png
[after]: https://user-images.githubusercontent.com/21217225/185096669-5020bf23-4db0-4587-af68-51f296d381d9.png

### DIFF
--- a/dotcom-rendering/src/web/components/Card/components/CardLink.tsx
+++ b/dotcom-rendering/src/web/components/Card/components/CardLink.tsx
@@ -12,6 +12,10 @@ const fauxLinkStyles = css`
 	left: 0;
 	background-color: transparent;
 
+	opacity: 0;
+	text-indent: 200%;
+	overflow: hidden;
+
 	:focus {
 		${focusHalo};
 	}
@@ -19,12 +23,19 @@ const fauxLinkStyles = css`
 
 type Props = {
 	linkTo: string;
+	headlineText: string;
 	containerPalette?: DCRContainerPalette;
 	dataLinkName?: string;
 };
 
-export const CardLink = ({ linkTo, dataLinkName = 'article' }: Props) => {
+export const CardLink = ({
+	linkTo,
+	headlineText,
+	dataLinkName = 'article',
+}: Props) => {
 	return (
-		<a href={linkTo} css={fauxLinkStyles} data-link-name={dataLinkName} />
+		<a href={linkTo} css={fauxLinkStyles} data-link-name={dataLinkName}>
+			{headlineText}
+		</a>
 	);
 };
--- a/dotcom-rendering/src/web/components/Card/components/CardLink.tsx
+++ b/dotcom-rendering/src/web/components/Card/components/CardLink.tsx
@@ -12,10 +12,6 @@ const fauxLinkStyles = css`
 	left: 0;
 	background-color: transparent;
 
-	opacity: 0;
-	text-indent: 200%;
-	overflow: hidden;
-
 	:focus {
 		${focusHalo};
 	}
@@ -34,8 +30,11 @@ export const CardLink = ({
 	dataLinkName = 'article',
 }: Props) => {
 	return (
-		<a href={linkTo} css={fauxLinkStyles} data-link-name={dataLinkName}>
-			{headlineText}
-		</a>
+		<a
+			href={linkTo}
+			css={fauxLinkStyles}
+			data-link-name={dataLinkName}
+			aria-label={headlineText}
+		/>
 	);
 };
--- a/dotcom-rendering/src/web/components/Card/Card.tsx
+++ b/dotcom-rendering/src/web/components/Card/Card.tsx
@@ -294,7 +294,11 @@ export const Card = ({
 			containerType={containerType}
 			isDynamo={isDynamo}
 		>
-			<CardLink linkTo={linkTo} dataLinkName={dataLinkName} />
+			<CardLink
+				linkTo={linkTo}
+				headlineText={headlineText}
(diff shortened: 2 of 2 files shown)


OllysCoding (approved):
+1 Great change!

oliverlloyd (approved):
This is very cool.

I wonder, is there any benefit in this text containing more contextual information? We spend a considerable amount of effort in using styles that convey context, like blue for sport, red for news, etc., should we also add this information to the text that gets reads out here?

AshCorr:
> I wonder, is there any benefit in this text containing more contextual information? We spend a considerable amount of effort in using styles that convey context, like blue for sport, red for news, etc., should we also add this information to the text that gets reads out here?

Thats a great idea! I think with this current solution you wouldn't be able to tell if it was Liveblog or if its in the Culture pillar, etc... I wonder if we could prefix the link with "Culture article:" or "Liveblog:"

oliverlloyd:
> I wonder if we could prefix the link with "Culture article:" or "Liveblog:"

A prefix makes sense. I can imagine it is nice having the context already established prior to hearing the headline. If nothing else, it will make it easier to, say, skip the sport content (if you were not a fan).

bryophyta:
haha great minds think alike I guess! I was just about to open a PR to fix the same issue and then I saw this 🤦‍♂️ ([here](https://github.com/guardian/dotcom-rendering/compare/pf/cardlink-wraps-cardheadline?expand=1), just for reference)

I was a bit uncertain about my approach anyway because it involves some css which is potentially a little fragile, and I think your approach looks much better from this perspective!

bryophyta:
> is there any benefit in this text containing more contextual information? 

Contextual info sounds good to me, although I'm not sure about adding it as a prefix, because e.g. on a sports front you're likely to get dozens of links all starting with 'sport...', which could be a bit frustrating to have to listen through if you already know you're on the sport page.

Heydon Pickering has an interesting discussion of a related, but not identical issue, in [their post on cards](https://inclusive-components.design/cards/). Their preferred solution was this:

> For sighted screen reader users there's potential for a little confusion here, since "read more" will not be announced despite the element appearing to take focus. Fortunately, we can attach "read more" to the link as a description, using aria-describedby. Now users will hear "Card design woes, link, read more". The description is always read last.

Our case isn't precisely the same, because the extra context isn't always the same like it is with 'read more', and also we don't already have the pillar in text on the card (afaik). But thought it might be worth thinking about something similar here.

AshCorr:
> I was a bit uncertain about my approach anyway because it involves some css which is potentially a little fragile, and I think your approach looks much better from this perspective!

Yea I was a bit wary in wrapping everything in the link, I assumed it was done this way for a reason, I guess because like you said it makes the CSS a bit fragile.

dotcom-rendering/src/web/components/Card/components/CardLink.tsx
```diff
@@ -12,19 +12,30 @@ const fauxLinkStyles = css`
 	left: 0;
 	background-color: transparent;
 
+	opacity: 0;
```
bryophyta, line 15: setting `opacity: 0` here hides the focus styling for the card.. do you think it could be applied to an inner `<span>` containing the text instead, or something like that?

AshCorr: > setting `opacity: 0` here hides the focus styling for the card.. do you think it could be applied to an inner `<span>` containing the text instead, or something like that?

Hm I hadn't noticed that, good catch!

AshCorr: Seems like just using `aria-label` is enough to get rid of the error

dotcom-rendering/src/web/components/Card/components/CardLink.tsx
```diff
@@ -34,8 +30,11 @@ export const CardLink = ({
...
 }: Props) => {
 	return (
-		<a href={linkTo} css={fauxLinkStyles} data-link-name={dataLinkName}>
-			{headlineText}
-		</a>
+		<a
+			href={linkTo}
+			css={fauxLinkStyles}
+			data-link-name={dataLinkName}
+			aria-label={headlineText}
```
oliverlloyd, line 37: 👍 this is easier to understand when looking at the code

AshCorr:
Going to merge this down as in and raise a seperate PR for `aria-description` as its going to be a little trickier to add a valuable description!

dotcom-rendering/src/web/components/Card/components/CardLink.tsx
OllysCoding: Nicer for when you inspect the DOM too!

mchv:
Nice one 💎

PR #5738, opened 2022-08-17
fix pinned post overlay visibilitty

marjisound:
## What does this change?
This PR is fixing the bug that's showing an overlay on top of share buttons which is preventing the user from clicking the share buttons. 
image

A while ago a PR was made to fix the visibility of the `Show More` button in pinned post when pinned post contains twitter embed https://github.com/guardian/dotcom-rendering/pull/5682 This PR fixed the visibility of the button, but caused another bug on the overlay section that's the bottom of the collapsed pinned post. 

image

The pinned post enhancer, was adding the `display: none` when the scroll height of the pinned post was less than the client height. Then as part of that PR, `display: none` was replaced by `display: block` when after the twitter enhancement, the scroll hight was changed and became larger than the client height. 

But because  `display: block` was added through js, it would go as an inline element styles, which would then have priority over the actual component styles. 

## Screenshots

| Before      | After      |
|-------------|------------|
| image | image |

### DIFF
--- a/dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
+++ b/dotcom-rendering/src/web/components/EnhancePinnedPost.importable.tsx
@@ -27,10 +27,22 @@ function toggleShowMore(show: boolean) {
 	const pinnedPostOverlay = document.querySelector<HTMLElement>(
 		'#pinned-post-overlay',
 	);
-	if (pinnedPostButton)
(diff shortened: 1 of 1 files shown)


oliverlloyd (approved):
👍

abeddow91 (approved):
Great stuff! Very explicit :)

PR #5750, opened 2022-08-17
update libs to introduce explainer type

marjisound:
## What does this change?
This PR does the following:
- Updates `guardian/libs` version to `7.1.0` in all **DCR**, **AR** & **common rendering**
- fixes the compiler errors in **DCR** due to the breaking change in https://github.com/guardian/libs/releases/tag/v7.0.0 ( Thanks @mxdvl for helping out on resolving these)
- Adds `Explainer` design type in **AR** item.ts
- Ran `make gen-stories` for **DCR** as it was necessary after the new format design `Explainer`

## Why?
The new type Explainer is required as part of the new changes the Editorial Design teams are introducing. More info can be found here: 
https://docs.google.com/document/d/1MqSldxfgR86kOJaXTFNaBs5LzPzSa7jC4D4YwqqMFwg/edit#


dotcom-rendering/src/web/lib/countryCodes.ts
```diff
@@ -0,0 +1,251 @@
+export const countries = [
```
mxdvl, line 1: This could make its way to libs!

mxdvl (approved):
Happy with the custom type guards, on which I paired.

JamieB-gu (approved):
Looks good from the AR side; I'll let Dotcom review the DCR side.

From what I can tell, at the moment if a piece is published with this tag AR will parse it correctly as an explainer. There won't, however, be a layout for it so it'll fall back to the "not implemented" page in the layout component:

https://github.com/guardian/dotcom-rendering/blob/5abc09e96401a0b9e3fcabc686aae40a6886e552/apps-rendering/src/components/Layout/index.tsx#L127

Just checking this is expected behaviour, and that we'll be implementing a layout for this before content of this type gets published? When that happens we'll also want to add a story to `Layout.stories.tsx` to get some snapshot testing.

marjisound:
> From what I can tell, at the moment if a piece is published with this tag AR will parse it correctly as an explainer. There won't, however, be a layout for it so it'll fall back to the "not implemented" page in the layout component:


Thanks @JamieB-gu for the reminder. So I think there's a risk right now that some of the production articles with tag `tone/explainer` may be rendered by AR. So I think before merging this PR, I should either filter out explainer articles from https://github.com/guardian/mobile-apps-api/blob/main/common-play/src/main/scala/data/RenderedItemUtils.scala

or another option is, remove the part that's adding this design in item.ts in this PR and only add it back when we have completed the layout design?

abeddow91:
How are explainer tags currently rendered? Would another option be to add Design.Explainer to the existing layout in switch in AR whilst the new template is being built?

marjisound:
> How are explainer tags currently rendered? Would another option be to add Design.Explainer to the existing layout in switch in AR whilst the new template is being built?

I think currently, they are rendered depending on their other tags. Maybe mostly they would go through the standard layout. I think it by removing the condition `else if (isExplainer(tags))...` from item.ts, we are letting them get rendered by standard layout

marjisound:
@JamieB-gu I added the Explainer design as part of the default Standard layout. Is that OK?

marjisound:
[Create Explainer design type in apps rendering](https://trello.com/c/UmBENXG2/534-create-explainer-design-type-in-apps-rendering)

PR #5616, opened 2022-08-06
Extract CAPIArticleType out of the ambient declaration

mxdvl:
## What does this change?

Extract types out of the ambient type declaration `index.d.ts` into their individual files: `badge`, `branding`, `commercial`, `config`, `edition`, `footer`, `front`, `frontend`, `onwards` and `trails`.

This refactor does not change the [output of the schema](https://github.com/guardian/dotcom-rendering/commit/1b6ae23f89c832903a5cf8fad96155d295466210#diff-13f8fac22130e08f3778bf25e2b99d1f55223b462545def5b164c793a13376bd), which only has two descriptive comments added.

It might be easier to review commits individually:

- fix(types): keep imported types local – ab0bce47f
- refactor(types): extract Badge – a23a4ed95
- refactor(types): import NavType explicitly – 4e8a74331
- refactor(types): extract Config – 676f2c9f9
- refactor(types): extract Footer – debd42e77
- refactor(types): extract Edition – debf09930
- refactor(types): extract TrailType – 8ca8532ef
- refactor(types): extract Branding – 055b40c3b
- refactor(types): extract Onwards – 9e2db1195
- refactor(types): extract Commercial – 09af97422
- fix(types): simplify ServerSideTests – 33655e5a3
- refactor(types): extract CAPIArticleType – d8d5ff668
- chore(fixtures): regenerate – 909086222

## Why?

We favour being explicit rather than implicit.

Knowing which types are are exported and where they are imported is now easily understood. Imports in ambient type declarations do not work well and we have ESLint errors. Refactors will be easiers when the relationships between types are explicit.

```mermaid
graph TD
    branding -->|Branding| commercial
    edition -->|EditionId| commercial

    trails -->|CAPITrailType| onwards

    badge --->|BadgeType| frontend
    commercial -->|CommercialProperties| frontend
    config --->|ConfigType| frontend
    edition --->|EditionId| frontend
    footer --->|FooterType| frontend
    onwards -->|CAPIOnwardsType| frontend
    trails -->|CAPITrailType| frontend
    
    frontend --> type((CAPIArticleType))
```

PR #5753, opened 2022-08-17
Extract `Palette` type

mxdvl:
## What does this change?

Extract the `Palette` type to its own file and out of the ambient declaration file.

## Why?

Follow-up on #5616.


bryophyta (approved):
+1

PR #5754, opened 2022-08-18
Adds `tableOfContents`

oliverlloyd:
## What does this change?
This adds the `enhanceTableOfContents` enhancer

## But we don't have any h3s or h4s
The code here is assuming that the document will contain pure `<h3>` or `<h4>` tags but these do not natively come from Composer. However there is work in play to support enhancing faux h3s (`<p><strong>I'm a faux h3</strong></p>`) so that they become h3 tags and also potential work to add h4 tag support as well

## Why?
We want to extract an object containing a nested stack of all h2/h3s or h3/h4s so that we can later render them in a `<TableOfContents />` component

## Isn't there already code to do this?
There is [another enhancer](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/model/enhance-interactive-contents-elements.ts) which does a similar heading extraction. This code is designed to add support for a legacy table of contents element. It looks for articles that contain this element and replaces it with a [new element](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/InteractiveContentsBlockComponent.tsx).

The goal for our table is to add support for *all* analysis articles dynamically , without a dependency on inserting a special atom element. At a later stage we could look to migrate all articles that contain this legacy element to our new dynamic pattern.

### DIFF
--- a/dotcom-rendering/src/model/enhanceTableOfContents.ts
+++ b/dotcom-rendering/src/model/enhanceTableOfContents.ts
@@ -46,6 +46,10 @@ const hasH3s = (blocks: Block[]) => {
 	return blocks.find((block) => block.elements.find(isH3));
 };
 
+const hasH4s = (blocks: Block[]) => {
+	return blocks.find((block) => block.elements.find(isH4));
+};
+
 const hasInteractiveContentsElement = (blocks: Block[]): boolean => {
 	return !!blocks.find((block) =>
 		block.elements.find(
@@ -87,7 +91,7 @@ export const enhanceTableOfContents = (
 						});
 					} else {
 						// This is an orphan h3 with no parent h2 element to assign it to
-						// so we drop it from the table
+						// so we don't show it in the table
 					}
 				}
 			});
@@ -102,21 +106,21 @@ export const enhanceTableOfContents = (
 						nested: [],
 					});
 				}
+			});
+		});
+	} else if (hasH4s(blocks)) {
+		blocks.forEach((block) => {
+			block.elements.forEach((element) => {
 				if (isH4(element)) {
-					if (tocs.length > 0) {
-						const lastItem = tocs[tocs.length - 1];
-						lastItem.nested.push({
-							id: element.elementId,
-							title: extractText(element),
-						});
-					} else {
-						// This is an orphan h4 with no parent h3 element to assign it to
-						// so we drop it from the table
-					}
+					tocs.push({
+						id: element.elementId,
+						title: extractText(element),
+						nested: [],
+					});
 				}
 			});
 		});
 	}
 
-	return tocs.length > 0 ? tocs : undefined;
+	return tocs.length >= 3 ? tocs : undefined;
 };
--- a/dotcom-rendering/src/model/enhanceTableOfContents.ts
+++ b/dotcom-rendering/src/model/enhanceTableOfContents.ts
@@ -1,0 +1,122 @@
+import { JSDOM } from 'jsdom';
+
+const isHeading = (element: CAPIElement, level: 'H2' | 'H3' | 'H4') => {
+	if (
+		element._type !==
+			'model.dotcomrendering.pageElements.TextBlockElement' &&
+		element._type !==
+			'model.dotcomrendering.pageElements.SubheadingBlockElement'
+	)
+		return false;
+	const frag = JSDOM.fragment(element.html);
+	return frag.firstElementChild?.nodeName === level;
+};
+
+const isH2 = (
+	element: CAPIElement,
+): element is SubheadingBlockElement | TextBlockElement => {
+	return isHeading(element, 'H2');
+};
+
+const isH3 = (
+	element: CAPIElement,
+): element is SubheadingBlockElement | TextBlockElement => {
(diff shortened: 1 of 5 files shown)


dotcom-rendering/src/model/enhanceTableOfContents.ts
```diff
@@ -0,0 +1,122 @@
+import { JSDOM } from 'jsdom';
+
+const isHeading = (element: CAPIElement, level: 'H2' | 'H3' | 'H4') => {
+	if (
```
abeddow91, line 4: Why do we need to check for these types here when we type them in each of the `isH` functions below? Is this just an extra layer of security?

dotcom-rendering/src/model/enhanceTableOfContents.ts
```diff
@@ -0,0 +1,122 @@
...
+	);
+};
+
+export const enhanceTableOfContents = (
+	format: CAPIFormat,
+	blocks: Block[],
+): TOCType[] | undefined => {
+	if (
+		format.design !== 'ArticleDesign' ||
+		hasInteractiveContentsElement(blocks)
```
abeddow91, line 65: Nice! super clear

dotcom-rendering/src/model/enhanceTableOfContents.ts
```diff
@@ -0,0 +1,122 @@
...
+	return isHeading(element, 'H2');
+};
+
+const isH3 = (
+	element: CAPIElement,
+): element is SubheadingBlockElement | TextBlockElement => {
+	return isHeading(element, 'H3');
+};
+
+const isH4 = (
```
abeddow91, line 27: Happy to see H4s accounted for :)

dotcom-rendering/src/web/server/index.ts
```diff
@@ -19,18 +20,23 @@ function enhancePinnedPost(format: CAPIFormat, block?: Block) {
...
-			data.format,
-			data.promotedNewsletter,
-		),
+		blocks: enhancedBlocks,
 		pinnedPost: enhancePinnedPost(data.format, data.pinnedPost),
 		standfirst: enhanceStandfirst(data.standfirst),
 		commercialProperties: enhanceCommercialProperties(
 			data.commercialProperties,
 		),
+		tableOfContents: data.config.switches.tableOfContents
```
abeddow91, line 37: Do we want to consider driving this from format as well here so we can limit it to just Analysis and Explainer or are we happy to open this up to all articles?

abeddow91 (approved):
Looks great! A few non-blocking questions but looks good to go 🚀

marjisound (approved):
Looks great 👍

dotcom-rendering/src/model/enhanceTableOfContents.ts
```diff
@@ -0,0 +1,122 @@
...
+		blocks.forEach((block) => {
+			block.elements.forEach((element) => {
+				if (isH2(element)) {
+					tocs.push({
+						id: element.elementId,
+						title: extractText(element),
+						nested: [],
+					});
+				}
+				if (isH3(element)) {
```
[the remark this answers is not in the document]
oliverlloyd: I just caught up with Harry around this and he proposed simplifying things

If there are h2s and h3s then these are nested. Otherwise there is no nesting. I'll make those changes now

dotcom-rendering/src/model/enhanceTableOfContents.ts
oliverlloyd: I played around with these checks to try and avoid bloating things but this if statement is needed it seems.

Because `element` can be any type of element, it needs the code to narrow things down. If I change the type of `element` to be `SubheadingBlockElement | TextBlockElement` then the calling function has a problem because that is passing in CAPIElement, which could be anything

dotcom-rendering/src/model/enhanceTableOfContents.ts
oliverlloyd: Yep, Harry asked for this 👍

dotcom-rendering/src/web/server/index.ts
oliverlloyd: I did initially start with adding this switch check inside the component itself, along with the check on `format.design` but then I realised I'd have to pass in a new prop to include the switch and then it seemed more sensible to just do the check externally.

The check for `format` though is something we do inside the enhancer functions in several other places so I left that there to keep things consistent.

I'm not strong on this to be honest so if you think it works better to have all the checks externally or all internally I'm open to this, what do you think?

oliverlloyd: > or are we happy to open this up to all articles?

I should have said, we have a check inside the enhancer like this

```ts
	if (
		format.design !== 'ArticleDesign' ||
		hasInteractiveContentsElement(blocks)
	) {
		return undefined;
	}
```

So it's only for Analysis for now

oliverlloyd: Oh, wait, that says Article, not analysis...

Fixing!

PR #5762, opened 2022-08-18
Add H3 Enhancer

abeddow91:
## What does this change?
Hoists the existing h3 enhancer into its own enhancer and separates existing h3 enhancement tests into its own test suite.
It also updates the deprecated `.substr` function.

## Why?
h3 elements do not exist in composer. Instead, Editorial use "faux" h3 tags that are comprised of `<p><strong>` tags and DCR converts these to h3. Previously this was only done whilst enhancing numbered lists. This PR hoists this functionality into its own enhancer so h3 are available across all TextBlockElements. 

numbered list example article https://www.theguardian.com/world/2021/feb/08/we-get-daily-abuse-uk-frontline-workers-on-the-covid-second-wave


dotcom-rendering/src/model/enhance-H3s.ts
```diff
@@ -0,0 +1,103 @@
...
+		if (
+			thisElement._type ===
+				'model.dotcomrendering.pageElements.TextBlockElement' &&
+			isFalseH3(thisElement)
+		) {
+			const h3Text = extractH3(thisElement);
+
+			// To avoid having to depend on the ordering of the enhancer (which could easily be a source of bugs)
+			// We determine if previous items are `ItemLinkBlockElement` through type and `isItemLink` functions
+			const isPreviousItemLink =
```
oliverlloyd, line 72: I think this is a numbered list thing. I'm not certain (it's confusing!) but we might need to narrow what we're extracting out a bit further here

dotcom-rendering/src/model/enhance-H3s.ts
```diff
@@ -0,0 +1,103 @@
...
+				previousItem?._type ===
+					'model.dotcomrendering.pageElements.ItemLinkBlockElement' ||
+				(previousItem && isItemLink(previousItem));
+
+			withH3s.push(
+				{
+					_type: 'model.dotcomrendering.pageElements.DividerBlockElement',
```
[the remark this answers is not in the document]
abeddow91: aha! I see, I've refactored this back into numbered lists and amended the test

dotcom-rendering/src/model/enhance-numbered-lists.ts
```diff
@@ -401,8 +347,8 @@ const enhance = (
 		new Enhancer(elements, format)
 			// Add the data-ignore='global-h2-styling' attribute
 			.removeGlobalH2Styles()
-			// Turn false h3s into real ones
-			.addH3s()
```
oliverlloyd, line 405: 💯 thanks for going the extra mile and renaming this to make sense

dotcom-rendering/src/model/enhance-H3s.test.ts
```diff
@@ -0,0 +1,200 @@
...
+						elementId: 'mockId',
+						html: '<h3>Strong 1 Strong 2</h3>',
+					},
+				],
+			},
+		];
+		expect(enhanceH3s(input)).toEqual(expectedOutput);
+	});
+
+	it('does set a h3 if there is more than one strong tag', () => {
```
oliverlloyd, line 33: This might be left over from the move but this test has a duplicate name to the one above

```suggestion
	it('handles multiple strong tags where one contains an a tag inside it', () => {
```

oliverlloyd (approved):
👍

PR #5765, opened 2022-08-18
Fixes body scrolling when side nav is open and burger button displaying above top ad slots

### LINKED ISSUE #4437: Subnav: make close icon sticky
Until tablet, if the menu is expanded, the menu close icon should stick. Currently we can scroll and the close icon is out of view.

https://user-images.githubusercontent.com/45561419/160574681-028a1c63-a9d2-475d-9eb1-e70203c4e51f.mp4

AshCorr:
## What does this change?

- Prevent the document body from scrolling when the navigation bar is open.
- Display the Navigation burger behind top slot ads on Tablet when the navigation bar is closed.

## Why?

Fixes #4437 
Partially fixes guardian/frontend#25253 (Still broken in Frontend but now fixed in DCR)

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/21217225/185447563-5eeebe93-364f-42f6-b8cf-0a067bd5ed91.png
[after]: https://user-images.githubusercontent.com/21217225/185446694-e8fdcd10-91e5-40fb-8277-37a5e8fa25dd.png
[before2]: https://user-images.githubusercontent.com/21217225/185447630-f7bf2945-dba3-4cad-bd84-825690d0de87.gif
[after2]: https://user-images.githubusercontent.com/21217225/185447106-8968c68f-340a-41f3-822f-986f814a62fe.gif

### DIFF
--- a/dotcom-rendering/src/web/components/Nav/Nav.tsx
+++ b/dotcom-rendering/src/web/components/Nav/Nav.tsx
@@ -1,4 +1,4 @@
-import { css, ThemeProvider } from '@emotion/react';
+import { css, Global, ThemeProvider } from '@emotion/react';
 import { ArticleDisplay, ArticleSpecial } from '@guardian/libs';
 import { space, until, visuallyHidden } from '@guardian/source-foundations';
 import {
@@ -71,6 +71,17 @@ export const Nav = ({ format, nav, subscribeUrl, editionId }: Props) => {
 
 	return (
 		<div css={rowStyles}>
+			<Global
+				styles={css`
+					/* We apply this style when the side navigation is open the prevent the document body from scrolling */
+					/* See Nav.tsx */
+					.nav-is-open {
+						${until.desktop} {
+							overflow: hidden;
+						}
+					}
+				`}
+			/>
 			{/*
                 IMPORTANT NOTE: Supporting NoJS and accessibility is hard.
 
--- a/dotcom-rendering/src/web/components/Nav/Nav.tsx
+++ b/dotcom-rendering/src/web/components/Nav/Nav.tsx
@@ -109,6 +109,8 @@ export const Nav = ({ format, nav, subscribeUrl, editionId }: Props) => {
 						if (!navInputCheckbox) return; // Sticky nav replaces the nav so element no longer exists for users in test.
 
                         navInputCheckbox.addEventListener('click',function(){
+                          document.body.classList.toggle('nav-is-open')
+
                           if(!navInputCheckbox.checked) {
 							firstColLabel.setAttribute('aria-expanded', 'false')
                             showMoreButton.setAttribute('data-link-name','nav2 : veggie-burger: show')
--- a/dotcom-rendering/src/web/components/ArticlePage.tsx
+++ b/dotcom-rendering/src/web/components/ArticlePage.tsx
@@ -1,11 +1,6 @@
 import { css, Global } from '@emotion/react';
 import { ArticleDesign } from '@guardian/libs';
-import {
-	brandAlt,
-	focusHalo,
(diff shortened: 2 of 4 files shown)


dotcom-rendering/src/web/components/Article.tsx
```diff
@@ -44,6 +49,14 @@ export const Article = ({ CAPIArticle, NAV, format }: Props) => {
 						background: ${brandAlt[400]};
 						color: ${neutral[7]};
 					}
+
+					/* We apply this style when the side navigation is open the prevent the document body from scrolling */
+					/* See Nav.tsx */
+					.nav-is-open {
```
AshCorr, line 55: Bit of a shame that we need to use a CSS class for this, I'm not entirely sure if theres another solution for this though. Anyone got any thoughts?

jamesgorrie: This does couple `Article.tsx` to it's child `Nav.tsx`, which is a shame.

I can think of a few options:
- rename `nav-is-open` to a more utility style name like `no-scroll`, `lock-scroll` or even `overflow-hidden`. While this seems like a superficial change, it moves the responsibility of this functionality, locking the scrolling on the body, to the `Article.tsx`, exposing it to be used by other components.
- remove the coupling completely and keep everything in `Nav.tsx`. We could achieve this by adding a `style` tag in `Nav.tsx` that does all the `nav-is-open` business, keeping it in one place. This is also more explicit than the above option as making it more abstract isn't great, as we don't actually have more use cases for it.

jamesgorrie (commented):
A couple of other options I could think of to try keep the functionality in 1 place.

dotcom-rendering/src/web/components/Article.tsx
AshCorr: Actually its coupling `Article.tsx`, `Nav.tsx`, and `VeggieBurgerMenu.tsx` 😰 

> rename nav-is-open to a more utility style name like no-scroll, lock-scroll or even overflow-hidden. While this seems like a superficial change, it moves the responsibility of this functionality, locking the scrolling on the body, to the Article.tsx, exposing it to be used by other components.

This would help quite a bit, but nav-is-open can't actually be all that generic... It needs to only apply `overflow: hidden` when the display is tablet or smaller, otherwise we'd be blocking scroll when the desktop navigation bar is open.

> remove the coupling completely and keep everything in Nav.tsx. We could achieve this by adding a style tag in Nav.tsx that does all the nav-is-open business, keeping it in one place. This is also more explicit than the above option as making it more abstract isn't great, as we don't actually have more use cases for it.

I quite like this idea, but it's still affected by the aforementioned issue of only blocking scrolling on a tablet sized device. If we were to apply the `overflow` style programatically we'd have to use Javascript to check what the screen size is before applying the style which comes with a host of its own issues.
Like the idea of making it more generic, it might make the CSS in VeggieBurgerMenu a bit confusing though, but can probably just resolve that with a comment.

What I could actually do is move the Global styles into `Nav.tsx`, that way atleast the styles are close to what uses them

jamesgorrie (approved):

dotcom-rendering/src/web/components/Nav/Nav.tsx
```diff
@@ -69,6 +69,17 @@ export const Nav = ({ format, nav, subscribeUrl, editionId }: Props) => {
 
 	return (
 		<div css={rowStyles}>
+			<Global
```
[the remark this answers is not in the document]
AshCorr: This is updated so now only the page body is linked with this css class, not `ExpandedMenu`

PR #5771, opened 2022-08-19
Add best paperbacks to nugget

HarryFischer:
This is sort of a test PR so I can see how the process of adding/removing links each month would work. Test because it's non user facing at the moment.


oliverlloyd (approved):
Approving but it seems to me this change doesn't really need approval, maybe we can put this list into a separate file and open up access to it?

PR #5776, opened 2022-08-19
Add styles to <sup /> element

iainjchambers-guardian:
## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

### DIFF
--- a/apps-rendering/src/renderer.ts
+++ b/apps-rendering/src/renderer.ts
@@ -321,7 +321,18 @@ const textElement =
 			case 'SUB':
 				return h('sub', { key }, children);
 			case 'SUP':
-				return h('sup', { key }, children);
+				return styledH(
+					'sup',
+					{
+						key,
+						css: {
+							position: 'relative',
+							top: '-0.5em',
+							fontSize: '75%',
+						},
+					},
+					children,
+				);
 			default:
 				return null;
 		}


georgeblahblah (commented):
This is interesting!

Out of curiosity, I [queried CAPI](http://content.guardianapis.com/search?api-key=test&show-fields=body&query-fields=body&q=sup) to see what the usage of `<sup>` is like. Without checking too closely, there are up to 658 articles using the element.

Some of them are using `sup` in hacky ways to style copy, like [this Comment Is Free article](https://www.theguardian.com/commentisfree/2022/jul/11/would-be-mothers-parent-solo-dating-apps-children). 
Here, the subheadings are wrapping the text in a `sup` element, which renders on this branch like:

Untitled-2022-08-12-1046

Interestingly, Dotcom renders these headings normally (as if it ignores the `sup`):

Screenshot 2022-08-19 at 16 29 13

I'm not sure if there's anything to be done about it in this PR. Some options might be to

- have an allowlist for children of `<h2>` elements (or allow only text) - this should be easier following the changes in #5767 
- scope the styling of `sup` so it does not apply to subheadings
- something else?

paperboyo:
Interesting! And, once again, not at all surprising to see users hack structure to get presentation (they think they) want. I would be against any allow lists platforms would have as this would need to be sorted in the tools. One can imagine (and find in the wild) examples of genuine HTML tags in H2s (links, super- or subscript elements).

Maybe Composer should disallow styling **the whole** of H2 as super- or subscript? But parts of it – should be allowed.

Feel free to initiate a discussion in Editorial Tools channel.

apps-rendering/src/renderer.ts
```diff
@@ -321,7 +321,18 @@ const textElement =
...
 			case 'SUP':
-				return h('sup', { key }, children);
+				return styledH(
+					'sup',
+					{
+						key,
+						css: {
+							position: 'relative',
+							top: '-0.5em',
+							fontSize: '75%',
```
JamieB-gu, line 331: Do you think it would make sense to use the rules the browsers (Safari and Firefox) use for superscript? `vertical-align` in particular has a specific keyword for this [^1].

```suggestion
							fontSize: 'smaller',
							verticalAlign: 'super',
```

[^1]: https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align#super

JamieB-gu (commented):
Good spot! One suggestion about the specific CSS rules used.

> have an allowlist for children of `<h2>` elements (or allow only text)

This is likely to come as part of the work we're doing with Journalism on tables of contents. We're changing the way headings are rendered.

> I would be against any allow lists platforms would have as this would need to be sorted in the tools.

This already happens to some extent. I can't speak for certain on DCR, but AR already renders on an element-by-element basis.

> One can imagine (and find in the wild) examples of genuine HTML tags in H2s (links, super- or subscript elements).

I think this is true. Links and basic formatting like emphasis will already be supported. If we come across an instance of superscript or subscript we can add support for that too.

iainjchambers-guardian:
So do we have a consensus about what we're going to do about the `<sup />`s in H2s? Think all we can do about it for now is figure out how to not apply styles if it's a child of a heading but that seems silly to me, putting in a hack to cover for a hack, can't the journalists just **not** do that, if it doesn't do anything on DCR anyway?

paperboyo:
> can't the journalists just not do that, if it doesn't do anything on DCR anyway?

Haha. If it's available in the Tools, not only someone's gonna do it. Also, platforms should never cover for that by ignoring tools' signals. Because this way, there's gonna be more of it if only as a result of mistakes (which then are invisible in Preview). All Composer styling options should have a result on platforms. 

Partial <sup> and <sub> should obviously be allowed and rendered accordingly as they can be totally legit.

Whether the whole of H2 should be allowed or if it could be prevented is a question for tools (for meetup?).

IMHO and all.

JamieB-gu:
> So do we have a consensus about what we're going to do about the `<sup />`s in H2s?

#5767 changes the way `h2` elements are rendered. I'll make a suggestion there to include support for `sup` and `sub`, as I think @paperboyo is correct that they're valid in headings. I don't think any changes are required in this PR, we may instead want to add support for them to DCR.

JamieB-gu:
> Interestingly, Dotcom renders these headings normally (as if it ignores the `sup`):

@georgeblahblah from what I can tell, DCR does actually render the `sup`. I suspect it may be caused by the same thing as @iainjchambers-guardian uncovered on AR: the reset CSS removing browser defaults for these elements.

iainjchambers-guardian:
> Haha. If it's available in the Tools, not only someone's gonna do it.

Yeah...I knew that in my heart but I'm not immune to optimism.

> I don't think any changes are required in this PR;

Does that mean it can be approved and merged? 🤞

JamieB-gu:
> Does that mean it can be approved and merged? 🤞

I think there's still an outstanding comment above?

apps-rendering/src/renderer.ts
iainjchambers-guardian: Weird, thought I had but no. Just fixed that

JamieB-gu (approved):

PR #5777, opened 2022-08-19
feat(tsconfig): check all indexed access for object and arrays

mxdvl:
## What does this change?

Turn [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) on in our TypeScript config.

> TypeScript has a way to describe objects which have unknown keys but known values on an object, via index signatures.
> Turning on `noUncheckedIndexedAccess` will add `undefined` to any un-declared field in the type.

## Why?

This guards us from a very common class of errors, where we access a member of an object or an item in an array, but don’t check if the value is actually defined.

```ts
type AbTests = { [key: string]: boolean };
const tests = {
  a: true,
  b: false
}
tests.c; // By default, TypeScript infers `boolean`, but in reality this is `undefined`!
```


dotcom-rendering/fixtures/manual/key-events.ts
```diff
@@ -14,7 +14,7 @@ const format = {
 	theme: ArticlePillar.News,
 };
 
-export const events: KeyEventCard[] = [
+export const events: [KeyEventCard, ...KeyEventCard[]] = [
```
OllysCoding, line 17: Does this change convince TS there is always atleast 1 item in this array?

OllysCoding: Is there no world where TS understands a statically declared array will always have at least as many elements as its originally defined with?
Or does the existence of modify-in-place function like Array.splice ruin this for us?

mxdvl: You could use `as const`, but then you run into other kinds of issues where you’re passing on `readonly` types to mutable types.

I think this could be a worthwile endeavour, but I think we should discuss as a team the various options.

dotcom-rendering/src/amp/lib/find-adslots.ts
```diff
@@ -84,8 +84,9 @@ const hasForwardBuffer = (
 
 	const enoughCharsForward = meetsThreshold || noForwardsEmbeds;
 
-	const neighbourSuitable = elements[index]
-		? suitableAdNeighbour(elements[index])
+	const neighbour = elements[index];
+	const neighbourSuitable = neighbour
+		? suitableAdNeighbour(neighbour)
```
mxdvl, line 89: Typescript needs the `neighbour` extracted to know that it is defined in the left side of the ternary.

dotcom-rendering/src/amp/lib/find-adslots.ts
```diff
@@ -144,8 +146,12 @@ export const findAdSlots = (elements: CAPIElement[]): number[] => {
 	const elementsWithLength = getElementsWithLength(elements);
 
 	for (let i = 0; i < elementsWithLength.length; i += 1) {
-		if (adCount < AD_LIMIT && isTextElement(elements[i])) {
-			charsSinceLastAd += elementsWithLength[i].length;
+		const element = elements[i];
+		const elementWithLength = elementsWithLength[i];
+		if (!element || !elementWithLength) throw new Error('Invalid element');
```
mxdvl, line 151: Throwing in the outer scope if the elements are not defined.

dotcom-rendering/src/amp/lib/image-fit.ts
```diff
@@ -16,6 +16,7 @@ export const bestFitImage = (
 		.sort((a, b) => a.width - b.width)
 		.find((img) => img.width >= containerWidth);
 
+	// @ts-expect-error -- we’re checking the array is not empty above
```
mxdvl, line 19: I gave up on rewriting this code as there's already a type-guard above, but TypeScript doesn't understand it.

dotcom-rendering/src/model/enhance-dividers.ts
```diff
@@ -1,6 +1,7 @@
 import { JSDOM } from 'jsdom';
 
-const isDinkus = (element: CAPIElement): boolean => {
+const isDinkus = (element?: CAPIElement): boolean => {
+	if (!element) return false;
```
mxdvl, line 4: It felt cleaner to rework the signature of this method rather than extracting the previous element below.

dotcom-rendering/src/model/enhance-numbered-lists.test.ts
```diff
@@ -4,14 +4,16 @@ import { images } from '../../fixtures/generated/images';
 import { blockMetaData } from '../../fixtures/manual/block-meta-data';
 import { enhanceNumberedLists } from './enhance-numbered-lists';
 
+const [image] = images as [ImageBlockElement];
+
```
mxdvl, line 8: This is a test, so failures will be caught, if the type assertion is not correct.

dotcom-rendering/src/server/lib/get-content-from-url.js
```diff
@@ -82,7 +82,7 @@ const parseURL = (requestURL, requestPath) => {
 	 */
 
 	let url = requestURL.includes('url=')
-		? requestURL.split('url=')[1]
+		? requestURL.split('url=')[1] ?? 'unknown'
```
mxdvl, line 85: This could fail catastrophically, but it only happens on the development server.

dotcom-rendering/src/web/browser/bootCmp/init.ts
```diff
@@ -4,10 +4,9 @@ import {
 	onConsentChange,
 } from '@guardian/consent-management-platform';
 import type { ConsentState } from '@guardian/consent-management-platform/dist/types';
-import { getCookie, log } from '@guardian/libs';
+import { getCookie, loadScript, log } from '@guardian/libs';
```
mxdvl, line 7: Not sure why it wasn't working before, but getting no errors with this import from `@guardian/libs`

dotcom-rendering/src/web/browser/ophan/ophan.ts
```diff
@@ -81,9 +81,9 @@ export const abTestPayload = (tests: {
 	[key: string]: string;
 }): OphanABPayload => {
 	const records: { [key: string]: OphanABEvent } = {};
-	Object.keys(tests).forEach((testName) => {
+	Object.entries(tests).forEach(([testName, variantName]) => {
 		records[`ab${testName}`] = {
-			variantName: tests[testName],
+			variantName,
```
mxdvl, line 86: By using `Object.entries`, we can ensure that we only check existing keys!

dotcom-rendering/src/web/components/BigNumber/BigNumber.tsx
```diff
@@ -27,5 +27,5 @@ export const BigNumber = ({ index }: { index: number }) => {
 		<Number10 />,
 	];
 
-	return numbers[index];
+	return numbers[index] ?? null;
```
mxdvl, line 30: If the number we want is not in the 0-10 range, return nothing.

dotcom-rendering/src/web/components/Callout/CheckboxSelect.tsx
```diff
@@ -15,7 +15,9 @@ export const CheckboxSelect = ({ formField, formData, setFormData }: Props) => (
 				// data related to this field is mapped to `formData` using `formField.id`
 				// We cannot assume that the data exists, so we need to check if `formField.id` key exists in `formData`
 				const selectedCheckboxesArray =
-					formField.id in formData ? formData[formField.id] : [];
+					formField.id in formData
+						? formData[formField.id] ?? []
+						: [];
```
mxdvl, line 20:
```suggestion
				formData[formField.id] ?? [];
						
```

dotcom-rendering/src/web/components/CricketScoreboard.tsx
```diff
@@ -107,21 +107,21 @@ export const CricketInnings = ({
 	// There can be up to 2 innings per team in a cricket match
 	switch (teamInnings.length) {
 		case 1:
-			return (
+			return teamInnings[0] ? (
```
mxdvl, line 110: This is somewhat redundant, as we are already checking the length of the array.

This class of refactors could remove the switch pattern entirely, but would probably need to start with the longest option first.

dotcom-rendering/src/web/components/Doughnut.test.tsx
```diff
@@ -12,19 +12,23 @@ const mockSections = [
 		label: 'Dat2',
 		color: '#ce070c ',
 	},
-];
+] as const;
 
 describe('Doughnut', () => {
 	it('should display the given labels', () => {
-		const { getByText } = render(<Doughnut sections={mockSections} />);
+		const { getByText } = render(<Doughnut sections={[...mockSections]} />);
```
mxdvl, line 19: Ee can't assign a readonly object to the mutable `SectionType`. There may be a cleaner way to do this?

dotcom-rendering/src/web/components/DynamicFast.stories.tsx
```diff
@@ -1,8 +1,25 @@
...
+	TrailType,
+	TrailType,
+	TrailType,
+	TrailType,
+	TrailType,
+	TrailType,
+	TrailType,
+	TrailType,
+];
+
```
mxdvl, line 22: Quite repetitive, we could build convenience `Tuple` types in the future.

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -80,7 +80,7 @@ const Card75_Card25 = ({
 	containerPalette?: DCRContainerPalette;
 	showAge?: boolean;
 }) => {
-	if (cards.length < 2) return null;
+	if (!(cards[0] && cards[1])) return null;
```
mxdvl, line 83: We only care about the first two items being defined.

dotcom-rendering/src/web/components/DynamicSlowMPU.stories.tsx
```diff
@@ -22,7 +23,7 @@ export default {
 	},
 };
 
-const bigs = trails.slice(0, 3);
+const bigs = trails.slice(0, 3) as [TrailType, TrailType, TrailType];
```
mxdvl, line 26: This is a story, if the type assertion is wrong, it will differ at runtime when building storybook.

dotcom-rendering/src/web/components/DynamicSlowMPU.tsx
```diff
@@ -29,57 +29,58 @@ const Card33_ColumnOfThree33_Ad33 = ({
...
-		</LI>
-		<LI percentage="33.333%">
-			<Hide until="tablet">
-				<AdSlot position="inline" index={index} />
-			</Hide>
-		</LI>
-	</UL>
-);
+}) =>
+	cards[0] && cards[1] && cards[2] && cards[3] ? (
```
mxdvl, line 33: We make sure the cards exist.

An alternative would be to type the cards array as a tuple of 4 items, and handle errors one level above.

dotcom-rendering/src/web/components/FixedMediumSlowXIIMPU.tsx
```diff
@@ -288,35 +288,46 @@ export const FixedMediumSlowXIIMPU = ({
 			return null;
 		}
 		case 1: {
+			/** assert type as we know there is only 1 item */
+			const single = trails as [TrailType];
```
mxdvl, line 292: We are not relying on the compiler here, but we are in the switch case where there is only one item.

It would be worthwhile to reconsider this logic.

dotcom-rendering/src/web/components/InteractiveContentsBlockComponent.stories.tsx
```diff
@@ -3,28 +3,31 @@ import { NumberedList } from '../../../fixtures/generated/articles/NumberedList'
...
-			block._type ===
-			'model.dotcomrendering.pageElements.InteractiveContentsBlockElement',
-	);
+const interactiveContentsBlock = enhanceInteractiveContentsElements(
+	NumberedList.blocks,
+)[0]?.elements.find(
+	(block): block is InteractiveContentsBlockElement =>
+		block._type ===
+		'model.dotcomrendering.pageElements.InteractiveContentsBlockElement',
+);
```
mxdvl, line 12: Remove the error expectation, and handle the case in the actual story below.

dotcom-rendering/src/web/components/InteractiveContentsBlockComponent.tsx
```diff
@@ -221,21 +221,19 @@ export const InteractiveContentsBlockComponent = ({
...
 					endElement &&
 					endElement.boundingClientRect.y > 0 &&
-					lastElement.ref &&
+					lastElement?.ref &&
 					lastElement.ref.getBoundingClientRect().y < 0
 				) {
-					return setStickyNavCurrentHeader(
-						enhancedSubheadings[enhancedSubheadings.length - 1],
-					);
+					return setStickyNavCurrentHeader(lastElement);
```
mxdvl, line 236: We've already extracted `lastElement`, so let's use it!

dotcom-rendering/src/web/components/Liveness.importable.tsx
```diff
@@ -229,6 +230,8 @@ export const Liveness = ({
 		if (!topOfBlog) return () => {};
 
 		const observer = new window.IntersectionObserver(([entry]) => {
+			if (!entry) return;
+
```
mxdvl, line 234: Can you have an empty entry? This might have to be improved in the typescript lib itself.

dotcom-rendering/src/web/components/MultiImageBlockComponent.tsx
```diff
@@ -96,7 +96,8 @@ export const MultiImageBlockComponent = ({
 	const imageCount = images.length;
 
 	switch (imageCount) {
-		case 1:
+		case 1: {
+			if (!images[0]) return null;
```
mxdvl, line 100: Somewhat redundant, but makes the compiler happy.

dotcom-rendering/src/web/components/SignInGateSelector.importable.tsx
```diff
@@ -102,7 +102,7 @@ const ShowSignInGate = ({
 	useEffect(() => {
 		submitViewEventTracking({
 			component: withComponentId(
-				signInGateTestIdToComponentId[abTest.id],
+				signInGateTestIdToComponentId[abTest.id] ?? 'unknown',
```
mxdvl, line 105: @guardian/identity you may want to have a peek.

dotcom-rendering/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
```diff
@@ -324,7 +324,7 @@ const RemoteBanner = ({
 		window
 			.guardianPolyfilledImport(module.url)
 			.then((bannerModule: { [key: string]: JSX.Element }) => {
-				setBanner(() => bannerModule[module.name]); // useState requires functions to be wrapped
+				setBanner(() => bannerModule[module.name] ?? null); // useState requires functions to be wrapped
```
mxdvl, line 327: @guardian/tx-engineers you might want to have a peek.

dotcom-rendering/src/web/lib/buildNewsletterSignUpText.tsx
```diff
@@ -5,8 +5,9 @@ const specialCasedValues: Record<string, string> = {
 export const buildDetailText = (input: string) => {
 	const normalisedInput = input.toLowerCase().trim();
 
-	if (normalisedInput in specialCasedValues) {
-		return `Free ${specialCasedValues[normalisedInput]} newsletter`;
+	const specialCase = specialCasedValues[normalisedInput];
+	if (specialCase) {
+		return `Free ${specialCase} newsletter`;
```
mxdvl, line 10: Needed to extract this to make the compiler happy.

@guardian/newsletters you might want to have a peek.

mxdvl (commented):
Some pointers.

dotcom-rendering/src/web/lib/buildNewsletterSignUpText.tsx
dblatcher: looks good - thanks for flagging.

dotcom-rendering/src/web/components/CricketScoreboard.tsx
mxdvl: Alternative:

```ts
	// There can be up to 2 innings per team in a cricket match
	const [one, two] = teamInnings;
	if (one && two) {
		return (
			<p>
				{cricketScore({ innings: one, short: true })}
				{' & '}
				{cricketScore({ innings: two })} ({two.overs} overs)
			</p>
		);
	}

	if (one) {
		return (
			<p>
				{cricketScore({ innings: one })} ({one.overs} overs)
			</p>
		);
	}

	return <p>Yet to bat</p>;
```

dotcom-rendering/src/web/components/Doughnut.test.tsx
mxdvl: I have opted for destructuring `mockSections` into its consitutuents, as we’re using these in the tests explicitely.

dotcom-rendering/src/server/lib/get-content-from-url.js
mxdvl: I have simplified this logic. If the url does not contain `url=`, then accessing the `[1]` will return undefined, and we default to the `requestUrl`. Simples.

dotcom-rendering/src/web/components/StickyBottomBanner/ReaderRevenueBanner.tsx
tjmw: I think this looks good! 👍

Strictly speaking this wasn't owned by TX, we looked after the similar `BrazeBanner.tsx` which sits along side this. Though now we've merged with Marketing Tools maybe this is us :-) /cc @tomrf1 in case you want to take a look.

dotcom-rendering/src/model/enhance-dividers.ts
mxdvl: I opted for defining a previous element, in the end.

dotcom-rendering/src/web/components/CricketScoreboard.tsx
mxdvl: Opted for a `isTuple` helper.

dotcom-rendering/src/web/components/DynamicFast.stories.tsx
mxdvl: `isTuple` helps is now used.

dotcom-rendering/src/web/components/SignInGateSelector.importable.tsx
ob6160: just curious about the choice of `unknown` as a default here — is this a standard thing?

mxdvl: Not particularly, hence the nudge.

I'm not sure what would make the most sense here, maybe you want to throw an error?

dotcom-rendering/fixtures/manual/key-events.ts
mxdvl: Here’s a propsal: https://github.com/guardian/dotcom-rendering/pull/5817/files

PR #5781, opened 2022-08-22
add analysis tag to articles with tone `Analysis`

alinaboghiu:
## What does this change?

Adding an 'Analysis' tag on articles with this tone in Apps Rendering

## Why?

This is part of the editorial design team's work

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/1229808/185879898-45c0afd5-51be-4f4a-bdaa-4810d5f9b579.png
[after]: https://user-images.githubusercontent.com/1229808/185879870-8bbaa6d6-3b7c-4408-8e52-6530888051e1.png


marjisound (approved):
LGTM 👍

apps-rendering/src/components/HeadlineTag/index.tsx
```diff
@@ -24,7 +29,9 @@ type Props = {
 };
 
 const HeadlineTag: FC<Props> = ({ tagText, format }) => (
-	<span css={headlineTagStyles(format)}>{tagText}</span>
+	<div css={headlineTagWrapper}>
```
marjisound, line 32: I'm just wondering, what happens if this component is used in other places? Do we need `articleWidthStyles` for all diferent use cases?

georgeblahblah (approved):

apps-rendering/src/components/HeadlineTag/index.tsx
alinaboghiu: I can't imagine when we wouldn't - the only other place HeadlineTag is used atm is Interviews and InterviewHeadline sets the width explicitly (perhaps this should also use articleWidthStyles instead - but didn't want to change too much)

alinaboghiu: I think we always want the HeadlineTag to align with the rest of the article and I think that's what using articleWidthStyles here achieves? But I could be wrong - would any issues be highlighted in chromatic tests?

marjisound: Yea you're right. It may be better if the interview also uses the articleWidthStyles. But I think right now there are 2 layers of padding applied. One in `InterviewHeadline` and one in `HeadlineTag`

abeddow91 (approved):

apps-rendering/src/components/HeadlineTag/index.tsx
marjisound: Maybe in the `HeadlineTag` we cvan just apply this spacing for when it's `Analysis`?

alinaboghiu:
The storybook changes need to be fixed on the main feature branch as they are related to DesignTag rather than HeadlineTag so will merge this into feature-new-analysis-design

JamieB-gu:
FYI, if the article is part of a series it'll look like this. Just checking that's expected?

https://www.theguardian.com/environment/2022/aug/19/weather-tracker-flash-floods-europe-heatwave-ends-thunderstorms

series-tag

PR #5783, opened 2022-08-22
Create and use humanised unique slug as h2 id

abeddow91:
## What does this change?
Enhances h2 tags by using a slugified, unique human-readable id from the text of the h2. This PR also adds a test for this feature. 
## Why?
We were previously using the element id as the id for h2s so that we can navigate. However, this produces a less than perfect url for our users. By creating a slugified version of the h2 text, we create a much friendlier nav link. Ensuring these are unique will also make sure we navigate to the correct header.

## Screenshots

Screenshot 2022-08-22 at 10 21 59

`<h2 id='fd332fba-32cb-4b05-a1bd-110049bcfb60'>What about a national treaty?</h2>`
becomes
`<h2 id='what-about-a-national-treaty'>What about a national treaty?</h2>`


dotcom-rendering/src/model/enhance-H2s.test.ts
```diff
@@ -1,5 +1,5 @@
 import { blockMetaData } from '../../fixtures/manual/block-meta-data';
-import { enhanceH2s } from './enhance-H2s';
```
oliverlloyd, line 2: I think it would be easier to understand / change this code in the future if we leave the `getUnique` as an implementation detail inside the enhancer and then add a test here to verify what happens if I pass multiple h2s with the same title.

It's nicer if the tests execute at as higher level as possible, focussing on results and leaving it up to the things being tested to decide how they do things

dotcom-rendering/src/model/enhance-H2s.ts
```diff
@@ -1,15 +1,54 @@
...
+	if (array.includes(slug)) {
+		const occurenceCount = array.filter(
+			(currentItem) => currentItem == slug,
+		).length;
+		return `${slug}-${occurenceCount}`;
+	}
+	return slug;
+};
+
+const slugify = (text: string) => {
```
oliverlloyd, line 23: I always like to leave a link to the source location that code came in from - it's super useful to see what the thinking behind a chunk of code was

dotcom-rendering/src/model/enhance-H2s.ts
```diff
@@ -1,15 +1,54 @@
...
 	const enhanced: CAPIElement[] = [];
 	elements.forEach((element) => {
-		// If this element is a subheading, insert the element ID as its ID
+		// If this element is a subheading, insert a humanised slug of the text as its ID
 		if (
 			element._type ===
 			'model.dotcomrendering.pageElements.SubheadingBlockElement'
 		) {
+			const text = extractText(element);
+			const slug = text ? slugify(text) : element.elementId;
```
oliverlloyd, line 43: Can `slugify` ever return a falsey value?

dotcom-rendering/src/model/enhance-H2s.ts
```diff
@@ -1,15 +1,54 @@
...
+			const uniqueSlug = getUnique(slug, slugifiedIds);
```
oliverlloyd, line 44: I'm not really on top of this code so feel free to take this or leave it but how about something more like

```
const slug = slugify( ... )
if (slugifiedIds.includes(slug)) {
  const positionInArray = findPosition( ... )
  slug += `-${positionInArray}`
}
```

Or something like this. I think it's more about having the check for if the slug is unique inline, that helps me understand the code better I find.

oliverlloyd (approved):
Looks good. This is a very cool change to make, it's really nice to see this level of care given to an id!

I've made some comment but I'm happy to leave them with you, feel free to action as you see fit and merge from there

JamieB-gu (commented):
Looks reasonable to me 👍

I'll let Dotcom approve since this code is in DCR, but I'm assuming you plan to use the same algorithm in AR? If so, it may be worth pulling it into `libs` at some point? It might be useful for other teams, and if both DCR and AR are using the same function we can make sure they remain in sync. It's useful to keep link formats the same between dotcom and apps, because it allows links shared from either to work across both.

Is it worth adding some tests for the `slugify` function itself to test the regexes? For example, testing against URLs that have non-words, non-normalised characters, multiple `-`s and so on.

dotcom-rendering/src/model/enhance-H2s.ts
JamieB-gu: `''` is falsy I believe? I think it's possible for it to return an empty string if the processing filters out all characters. For example, if the input is a space: `' '`.

abeddow91: Yep, I think you're right here. I'll put a check in place to make sure we always get a viable id here

dotcom-rendering/src/model/enhance-H2s.test.ts
abeddow91: Absolutely, I'll refactor this.

dotcom-rendering/src/model/enhance-H2s.ts
abeddow91: Good point! i'll add this in

PR #5785, opened 2022-08-22
update explainer headline

marjisound:
## What does this change?
Thsi PR updates the DCR headline for Explainer articles. 
- headline font weight is bold now
- The Explainer tag will be on top of headline
## Why?
Requested by editorial design team https://docs.google.com/document/d/1MqSldxfgR86kOJaXTFNaBs5LzPzSa7jC4D4YwqqMFwg/edit#
## Screenshots

| Before      | After      |
|-------------|------------|
| image | image |


dotcom-rendering/src/web/components/ArticleHeadline.tsx
```diff
@@ -670,7 +670,7 @@ export const ArticleHeadline = ({
 									css={[
 										format.theme === ArticleSpecial.Labs
 											? labsFont
-											: standardFont,
+											: boldFont,
```
oliverlloyd: This is a change to Analysis here, was this intentional?

dotcom-rendering/src/web/components/ArticleHeadline.tsx
```diff
@@ -689,6 +689,39 @@ export const ArticleHeadline = ({
 							)}
 						</div>
 					);
+				case ArticleDesign.Explainer:
```
oliverlloyd, line 692: Does this one need a new branch? Is the return code for feature/review etc. the same as this?

E.g. Could we do this:

```
switch (format.design) {
				case ArticleDesign.Review:
				case ArticleDesign.Recipe:
				case ArticleDesign.Explainer:
				case ArticleDesign.Feature:
					return (
						<div
```

This is just a guess, I could be wrong!

dotcom-rendering/src/web/lib/decideDesign.ts
```diff
@@ -56,6 +56,8 @@ export const decideDesign = ({
 			return ArticleDesign.FullPageInteractive;
 		case 'NewsletterSignupDesign':
 			return ArticleDesign.NewsletterSignup;
+		case 'ExplainerDesign':
```
oliverlloyd, line 59: Nice 👍

oliverlloyd (commented):
Now that we're adding a new design, should we also add new stories? Maybe in ArticleHeadline and in Layout?

🤔 For the layout stuff we'd need a new fixture added [in the gen-fixtures](https://github.com/guardian/dotcom-rendering/blob/feature-new-analysis-design/dotcom-rendering/scripts/test-data/gen-fixtures.js#L29) file. And then we can run `make gen-fixtures` and add it [to the Layouts.stories file here](https://github.com/guardian/dotcom-rendering/blob/feature-new-analysis-design/dotcom-rendering/src/web/layouts/Layout.stories.tsx#L32).

This will generate a lot of diffs but I think this is the right place to add this story

dotcom-rendering/src/web/components/ArticleHeadline.tsx
oliverlloyd: Ah, it's already gone 👍

dotcom-rendering/src/web/components/ArticleHeadline.tsx
marjisound: Great point, I wasn't sure which of the existing switch branches was appropriate for Explainer. This seems to be good. Just fixed it 👍 

The only point worth mentioning is that, for Analysis we are now having a different font style when theme is labs.  but now for Explainer, we don't apply labs font styles any more

marjisound:
@oliverlloyd thanks so much for also explaining how to do the new layout story :) 

I added the new `Explainer` layout story following the steps you mentioned. Also added a new headline story for explainer.

oliverlloyd (approved):
Yep 👍

PR #5788, opened 2022-08-22
Clickable `DesignTag`

oliverlloyd:
## What does this change?
Makes `DesignTag` clickable

## Why?
To keep the page content semantically marked up


mxdvl (approved):
Nice one!

Is it worth using the `LinkButton` from source here if we’re resetting so much of its css?

oliverlloyd:
> Is it worth using the `LinkButton` from source here if we’re resetting so much of its css?

Yeah, it's a lot of overrides!

My feeling though is if I use the Source component I preserve the onion skin philosophy and don't need to worry about what may or may not exist in Source. Maybe someone adds some tracking code there or a special accessibility feature, like this that code will flow through

marjisound (approved):
Amazing :)

abeddow91 (approved):

PR #5791, opened 2022-08-22
Add useParticipations display rule for Sign In Gate

vlbee:
This PR adds a new display rule, which will run to determine whether or not to show the sign in gate. 

**`useParticipations`** - this display rule will either set an AB test participation in local storage or check local storage for participation when determining whether to display the gate. 

**Why do we need `useParticipations`?** 

We would like to be able to run long-running tests over several weeks, but only accumulate browsers in the test bucket for part of that time. This will allow us to AB test behaviour over time.  We cannot use the `mvt_id` cookie alone, as this is set by Fastly and we crucially do not know when it was set on a given browser. The solution implemented repurposes the existing `gu.ab.participations` local storage key to flag browsers bucketed into a test within a pre-defined "cutoff" window. 

_Caveats_ - this solution does not require a new cookie with timestamp. This solution also means all users in the `mvt_id` test bucket range encountering the test after the cutoff window will _never_ see the sign in gate until the test is is turned off.  

TODO
- Investigate if there a way to default back to the standard sign in gate if a test "canShow" function returns false?

### DIFF
--- a/dotcom-rendering/src/web/components/SignInGate/displayRule.ts
+++ b/dotcom-rendering/src/web/components/SignInGate/displayRule.ts
@@ -1,6 +1,7 @@
 // use the dailyArticleCount from the local storage to see how many articles the user has viewed in a day
 import { onConsent } from '@guardian/consent-management-platform';
 import type { ConsentState } from '@guardian/consent-management-platform/dist/types';
+import { getLocale } from '@guardian/libs';
 import {
 	getParticipationsFromLocalStorage,
 	setParticipationsInLocalStorage,
--- a/dotcom-rendering/src/web/components/SignInGate/displayRule.ts
+++ b/dotcom-rendering/src/web/components/SignInGate/displayRule.ts
@@ -91,7 +91,7 @@ export const useParticipations = (
 	currentTest: CurrentSignInGateABTest,
 ): Promise<boolean> => {
 	const now = Date.now(); // UTC
-	const { id, name, variant } = currentTest;
+	const { id, variant } = currentTest;
 
 	// BEFORE cutoff date all users will be bucketed and potentially shown the gate if they satisfy all subsequent conditions
 	if (now < cutOff) {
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/SignInGate/displayRule.ts
```diff
@@ -102,6 +120,41 @@ export const hasRequiredConsents = (): Promise<boolean> => {
...
+		});
+		return Promise.resolve(true); //  we only want to bucket until the cutoff
+	}
+
+	// AFTER cutoff date, browsers without local storage test participation will not be shown the gate
+	const participations = getParticipationsFromLocalStorage();
+	// if the test participation exists, show the test gate
+	return !!participations[id] && participations[id]?.variant === variant
+		? Promise.resolve(true)
+		: Promise.resolve(false);
```
mxdvl, line 155: The conditional chaining helps you avoid the check for the existence of `participations[id]`.

```suggestion
	return participations[id]?.variant === variant
		? Promise.resolve(true)
		: Promise.resolve(false);
```

dotcom-rendering/src/web/components/SignInGate/displayRule.ts
```diff
@@ -30,6 +34,20 @@ export const isIOS9 = (): boolean => {
...
+		return navigator.maxTouchPoints > 0;
+	} else {
+		const ua = navigator.userAgent;
+		return (
+			/\b(BlackBerry|webOS|iPhone|IEMobile)\b/i.test(ua) ||
+			/\b(Android|Windows Phone|iPad|iPod)\b/i.test(ua)
+		);
+	}
+};
+
```
mxdvl, line 50: Can this be moved to a separate PR?

What’s the reason for introducing this `isMobile` check? I don’t think we tend to vary the behaviour of the website based on the reader’s device type.

vlbee: Its a requirement for the new long AB test we need to run - more info[ here ](https://docs.google.com/document/d/10sa-pzxABeOQ6_gXzqTpdh5AKoiK_t69BU2xsLTeIc8/edit#heading=h.2pyb60zfmnx6)

I'll also be adding a display rule to check region, specifically EU for this test. Will put that and `isMobile` into separate PRs

Issue #5038, opened 2022-05-27
Issue ID: DAC_Bypass_Blocks2


VDuczekW:
Report Link: https://drive.google.com/file/d/1EiqM0vMziEvdG_N8ne0GtIg-MsDydvqy/view?usp=sharing

Full issue list: https://docs.google.com/spreadsheets/d/1vWQtZfD49uq7uNCNELuYN_HcIj-BKiE4t4Hwda3tJaQ/edit#gid=0

bryophyta:
>**Note**
>This is an automatically-generated comment. Feel free to edit the content, but please don't edit this 'note' block.
 
# Intro to report

Copying the intro here to aid discoverability of the report's content:
```
Issue ID: DAC_Bypass_Blocks2
URL: https://www.theguardian.com/artanddesign/ng-interactive/2018/mar/02/how-tomake-a-pearl-the-san-francisco-man-who-lives-in-darkness
Page title: How to Make a Pearl - the San Francisco man who lives in darkness | Art and
design | The Guardian
Journey: 1.6
Screenshot:

The first link that users encounter is the ‘Skip to main content’. The destination of the link is
the ‘maincontent’ ID but this is not present on the page. Due to the broken reference of the
link users will think that the bypass method is broken on the site, and they are unable to
skip to the main content.
```

PR #5792, opened 2022-08-22
a11y : Add id for skip to main content for full page interactive layout

DanielCliftonGuardian:
## What does this change?

This adds the ID to skip to main content for full page interactive layouts.

This resolves https://github.com/guardian/dotcom-rendering/issues/5038

## Why?

This was highlighted by the DAC review: https://drive.google.com/file/d/1EiqM0vMziEvdG_N8ne0GtIg-MsDydvqy/view

### DIFF
--- a/dotcom-rendering/src/web/layouts/FullPageInteractiveLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/FullPageInteractiveLayout.tsx
@@ -328,23 +328,25 @@ export const FullPageInteractiveLayout = ({
 				backgroundColour={palette.background.article}
 				element="main"
 			>
-				<article>
-					<Renderer
-						format={format}
-						elements={
-							CAPIArticle.blocks[0]
-								? CAPIArticle.blocks[0].elements
-								: []
-						}
-						host={host}
-						pageId={CAPIArticle.pageId}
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/layouts/FullPageInteractiveLayout.tsx
```diff
@@ -328,23 +328,25 @@ export const FullPageInteractiveLayout = ({
...
-						pageId={CAPIArticle.pageId}
-						webTitle={CAPIArticle.webTitle}
-						ajaxUrl={CAPIArticle.config.ajaxUrl}
-						switches={CAPIArticle.config.switches}
-						isAdFreeUser={CAPIArticle.isAdFreeUser}
-						isSensitive={CAPIArticle.config.isSensitive}
-					/>
-				</article>
+				<div id="maincontent" tabIndex={-1}>
+					<article>
```
oliverlloyd, line 332: Does this change break any contracts we might have with interactive content? These articles are each written slightly differently and sometimes there is javascript which reads the dom expecting a certain structure.

Unfortunately, we have no visibility on what these contracts are but it's possible adding in this div could create issues. To mitigate this, would it work if we move the id and tabIndex onto the `article` element instead?

Issue #5794, opened 2022-08-22
Rationalise Onwards / MostPopular types


jamesgorrie:
## Use the concept of `MostPopular` and `Onwards`
We are proposing we have `MostPopular` and `Onwards`, which are both quite similar, but have different data sources, and are rendered on differently in different parts of the page.

Would leave us with `DotcomRenderingDataModel` of:
```scala
trait OnwardSource
object OnwardSource {
  case object Series extends OnwardSource
  case object PopularInTag extends OnwardSource
  case object MostOnThisStory extends OnwardSource
  case object RelatedStories extends OnwardSource
  case object CuratedContent extends OnwardSource
}
case class OnwardsCollection(heading: String, trails: Seq[Trail], onwardsSource: OnwardSource)

trait MostPopularSource
object MostPopularSource {
  case object MostViewed extends MostPopularSource
  case object MostViewInSection extends MostPopularSource
  case object DeeplyRead extends MostPopularSource
}
case MostPopularCollection(heading: String, trails: Seq[Trail], mostPopularSource: MostPopularSource)

case class DotcomRenderingDataModel(
  // ...
  onwards: Option[Seq[OnwardsCollection]],
  mostPopular: Option[Seq[MostPopularCollection]]
)
```

![onwards-most-popular](https://user-images.githubusercontent.com/31692/185965706-256b555f-56cd-4ad5-9618-0463e76ab7a4.png)

## Rename [`OnwardItem`](https://github.com/guardian/frontend/blob/main/common/app/model/dotcomrendering/MostPopular.scala#L14-L34) => `Trail`
As `OnwardsItem` is now used in `MostPopular` and `Onwards`, we can also standardise this type to be more generic and useful across multiple types of containers.

https://github.com/guardian/frontend/pull/25420

## Rename `CAPIOnwards` type in DCR to `FEOnwardsContainer`

This would bring both frontend and DCR inline with eachother.

jamesgorrie:
This has been closed in favour of https://github.com/guardian/dotcom-rendering/issues/6036

PR #5795, opened 2022-08-22
Rename `OnwardsSource` type

mxdvl:
## What does this change?

Rename types to be more explicit:

- `OnwardsType` &rarr; `OnwardsSource`
- `CAPIOnwardsType` &rarr; `CAPIOnwards` 
- 'default-onwards' &rarr; 'unknown-source'

## Why?

Together with @jamesgorrie, we think it helps to think about these data types better. See also #5794.

### DIFF
--- a/dotcom-rendering/src/web/server/index.ts
+++ b/dotcom-rendering/src/web/server/index.ts
@@ -7,7 +7,7 @@ import { enhanceStandfirst } from '../../model/enhanceStandfirst';
 import { validateAsCAPIType, validateAsFrontType } from '../../model/validate';
 import type { DCRFrontType, FEFrontType } from '../../types/front';
 import type { CAPIArticleType } from '../../types/frontend';
-import type { CAPIOnwardsType } from '../../types/onwards';
+import type { CAPIOnwards } from '../../types/onwards';
 import { articleToHtml } from './articleToHtml';
 import { blocksToHtml } from './blocksToHtml';
 import { frontToHtml } from './frontToHtml';
@@ -179,17 +179,18 @@ export const renderKeyEvents = (
 };
 
 export const renderOnwards = (
-	{ body }: { body: CAPIOnwardsType },
+	{ body }: { body: CAPIOnwards },
 	res: express.Response,
 ): void => {
 	try {
-		const { heading, description, url, onwardsType, trails, format } = body;
+		const { heading, description, url, onwardsSource, trails, format } =
+			body;
 
 		const html = onwardsToHtml({
 			heading,
 			description,
 			url,
-			onwardsType,
+			onwardsSource,
 			trails,
 			format,
 		});
--- a/dotcom-rendering/src/model/article-schema.json
+++ b/dotcom-rendering/src/model/article-schema.json
@@ -361,8 +361,8 @@
                     "url": {
                         "type": "string"
                     },
-                    "onwardsType": {
-                        "$ref": "#/definitions/OnwardsType"
+                    "onwardsSource": {
+                        "$ref": "#/definitions/OnwardsSource"
                     },
                     "format": {
                         "type": "object",
@@ -390,7 +390,7 @@
                 "required": [
                     "format",
                     "heading",
-                    "onwardsType",
+                    "onwardsSource",
                     "trails"
                 ]
             }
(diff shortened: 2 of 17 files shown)


oliverlloyd (commented):
I think the move from `onwardsType` to `onwardsSource` makes a lot of sense!

The move away from `CAPIOnwardsType` to `FEOnwardsContainer` however seems to be merging front and article terminology, which I personally find more confusing.

At present, the CAPI prefix is used for article data from Frontend and the FE prefix is used for data from Facia. The Type suffix is used to denote a type where there is potential for a naming clash. In addition, in my brain, onwards is an article concept and containers a fronts thing so having a `FEOnwardsContainer` makes it hard for me to know what this means 🙃

jamesgorrie:
@oliverlloyd `CAPI` vs `FE` - interesting, I didn't realise this was to differentiate between `frontend.facia` and `frontend.article`. I thought we were leaning into using `FE` as `CAPI` is confusing as it's not from `CAPI`. Maybe for this PR we can continue to use `CAPI`, but potentially revisit it.

Would `CAPIOnwardsContainer` make sense, as that's what we're calling it when it comes from `frontend` i.e. `OnwardsContainer`, and the components that renders this type is `OnwardContainer`.

oliverlloyd:
> Would `CAPIOnwardsContainer` make sense, as that's what we're calling it when it comes from `frontend` i.e. `OnwardsContainer`, and the components that renders this type is `OnwardContainer`?

We don't use Frontend names anywhere else in DCR so I'm not sure we should start to use this pattern. DCR and Frontend don't share types at all at the moment and my worry is that by using the same name would give the illusion of a shared type without it actually being the same thing.

Fundamentally, DCR and Frontend are very different things with different data and goals. I don't think we should be creating any couplings like this between the two codebases unless they are enforced by actual code.

mxdvl:
> The move away from `CAPIOnwardsType` to `FEOnwardsContainer` however seems to be merging front and article terminology, which I personally find more confusing. (@oliverlloyd)

I was under the same impression as @jamesgorrie, here:

> @oliverlloyd `CAPI` vs `FE` - interesting, I didn't realise this was to differentiate between `frontend.facia` and `frontend.article`. I thought we were leaning into using `FE` as `CAPI` is confusing as it's not from `CAPI`. Maybe for this PR we can continue to use `CAPI`, but potentially revisit it.

Happy to keep the `CAPI` -> `FE` prefix renaming out of this, but I was under the impression that we were planning to rename all of these at some point, and the reason they are present for the fronts is only a matter of having been created after we agreed that using the `CAPI` prefix was both confusing and not entirely true.

jamesgorrie:
@oliverlloyd re: Container - this makes a lot of sense, it really ties it to component-style thinking. 

In frontend we are going to stick to the `Collection` naming [as suggested here](https://github.com/orgs/guardian/projects/12) - maybe that's useful?

> when done together, as one PR?

I think this would be the case, but we were unaware that these decisions were so far reaching. Maybe we can open a mini RFC about prefixes, get consensus, then put the decision into code.

For this PR could we
- rename source
- follow up with `OnwardsType` => `OnwardsCollection`

mxdvl:
Thanks both, that makes a lot of sense.

I just pushed a commit going with `CAPIOnwards` here, as it’s neither sharing a name with Frontend nor introducing the `FE` prefix. Should I stick with the existing `CAPIOnwardsType` for now?

dotcom-rendering/src/web/server/index.ts
```diff
@@ -179,17 +179,18 @@ export const renderKeyEvents = (
 };
 
 export const renderOnwards = (
-	{ body }: { body: CAPIOnwardsType },
+	{ body }: { body: FEOnwardsContainer },
```
oliverlloyd, line 182: I presume this is a glitch?

oliverlloyd (approved):
So we're saying:

Rename `default-onwards` > `unknown-source`,
rename `onwardsType` > `onwardsSource`
and rename `CAPIOnwardsType` to `CAPIOnwards`

I think using `onwardsSource` and `unknown-source` are both excellent changes.

I think dropping the Type suffix is not really needed but I don't feel strongly about it. I recognise that using a suffix for type names is a long standing area of debate and there are pros and cons to both camps.

dotcom-rendering/src/web/server/index.ts
mxdvl: Yes, VSCode wasn’t greedy enough!

jamesgorrie (approved):

PR #5800, opened 2022-08-23
Add global h3 styles

abeddow91:
## What does this change?
Provides global h3 styles to all h3 elements, not just those in a numbered list. 

## Why?
Composer doesn't provide h3 tags, it provides "faux" h3 tags (`<p><strong>`). DCR previously converted these faux tags to h3 tags for numbered lists only. We have recently hoisted this logic to enhance all h3s. As such, we need to add additional styles. 

Whilst undertaking this work, we also noticed that these h3 tags were being wrapped in a span due to React injecting them via dangerouslySetHtml. We have updated the unwrapping logic to stop h3 tags from being wrapped in a unnecessary span. 

Paired with @oliverlloyd


marjisound (approved):
Looks great :)

PR #5802, opened 2022-08-23
Add h3 enhancer

abeddow91:
## What does this change?
Hoists the existing h3 enhancer into its own enhancer and separates existing h3 enhancement tests into its own test suite.
It also updates the deprecated `.substr` function.

## Why?
h3 elements do not exist in composer. Instead, Editorial use "faux" h3 tags that are comprised of `<p><strong>` tags and DCR converts these to h3. Previously this was only done whilst enhancing numbered lists. This PR hoists this functionality into its own enhancer so h3 are available across all articles that contain chaptered UI. These are Explainer, Analysis and Numbered Lists. 

numbered list example article https://www.theguardian.com/world/2021/feb/08/we-get-daily-abuse-uk-frontline-workers-on-the-covid-second-wave


oliverlloyd (approved):

marjisound (approved):
Very neat 👌

dotcom-rendering/src/model/enhance-H3s.test.ts
```diff
@@ -0,0 +1,245 @@
...
+					},
+				],
+			},
+		];
+		expect(enhanceH3s(input, AnalysisArticle.format)).toEqual(
+			expectedOutput,
+		);
+	});
+
+	it('does set an h3 and extract all text even if it is between stong tages', () => {
```
mxdvl, line 96:
```suggestion
	it('does set an h3 and extract all text even if it is between strong tags', () => {
```

dotcom-rendering/src/model/enhance-H3s.ts
```diff
@@ -0,0 +1,92 @@
...
+		containsStrongtags &&
+		doesNotContainLinks &&
+		startStrong &&
+		endsStrong
+	);
+};
+
+const extractH3 = (element: CAPIElement): string => {
+	// Extract the text based on the convention: <p><strong><H3 text</strong></p>
+	const textElement = element as TextBlockElement;
```
mxdvl, line 34: 1. Could the comment be moved to a JSDoc style?

2. Why does the function not take a TextBlockElement, but rather does a type assertion? I can help with creating a type-guard and we have other examples in the enhancers doing exactly this

3. You're doing the `isFalseH3` check twice. It might be better to create a new type-guard especially for this and I'm happy to pair on it.

dotcom-rendering/src/model/enhance-H3s.ts
```diff
@@ -0,0 +1,92 @@
...
+
+const enhance = (elements: CAPIElement[]): CAPIElement[] => {
+	/**
+	 * Why not just add H3s in Composer?
+	 * Truth is, you can't. So to get around this there's a convention that says if
+	 * you insert <p><strong>Faux H3!</strong>,</p> then we replace it with a h3 tag
+	 * instead.
+	 */
+	const withH3s: CAPIElement[] = [];
+	elements.forEach((thisElement) => {
```
mxdvl, line 56: I know this is being used elsewhere, but I think it's an anti-pattern.

A `map` is much more explicit and consistent. It should be trivial to refactor.

mxdvl (commented):
Excellent work on the tests. Some comments below.

dotcom-rendering/src/model/enhance-H3s.ts
oliverlloyd: I don't think it's entirely fair to call `forEach` an anti-pattern 🤣 . It's a valid feature of the javascript language

mxdvl: `forEach` has its uses. It's a function that invokes side-effects for every item in an array.

In this case, however, these side effects have the property of mutating an array which ends up being the exact same length as the input array. There's a more explicit method that does exactly that: `map`. We generally favour being explicit and try to follow the principle of least surprise.

So, for this reason, using `forEach` to perform imperatively the exact same thing as `map` is “a bit of an anti-pattern”. If you wanted an array of a different length than the original array, or if you wanted to register listeners, or log things to the console, I would probably still recommend a `for…of` loop over `forEach` as it becomes more explicit that you will have side effects.

mxdvl: For comparison, the resulting code is nearly identical, but you also get the added benefit of the TypeScript compiler catching the “passthrough“ case if you had forgotten to push it.

```ts
const withH3s: CAPIElement[] = elements.map((thisElement) => {
	if (isTextBlockElement(thisElement) && isFalseH3(thisElement)) {
		const h3Text = extractH3(thisElement);

		return {
			...thisElement,
			html: `<h3>${h3Text}</h3>`,
		};
	} else {
		// Pass through
		return thisElement;
	}
});
```

dotcom-rendering/src/model/enhance-H3s.ts
```diff
@@ -0,0 +1,92 @@
...
+	// Extract the text based on the convention: <p><strong><H3 text</strong></p>
+	const textElement = element as TextBlockElement;
+	const frag = JSDOM.fragment(textElement.html);
+	if (isFalseH3(element)) {
+		return (
+			frag.firstElementChild?.innerHTML
+				.split('<strong>')
+				.join('')
+				.split('</strong>')
+				.join('') ?? ''
```
mxdvl, line 42: It might be more explicit to use `replaceAll` here, if what we’re doing is stripping away the `strong` tags.

```suggestion
			frag.firstElementChild?.innerHTML
				.replaceAll('<strong>', '')
				.replaceAll('</strong>', '') ?? ''
```

dotcom-rendering/src/model/enhance-H3s.ts
oliverlloyd: I think map is a perfectly fine alternative here for the reasons you've given, and thanks for the example, that's useful!

I also think forEach is a valid solution; it's in use in lots of places and has not caused any problems. The term 'anti-pattern' has negative connotations which I don't think are fair given the differences here are ones of paradigm (functional programming vs. imperative) which are subjects of debate rather than concrete facts.

My personal preference these days is to use `map` and I generally think the community is moving this way but I am happy to work on code that uses `forEach` and tend to leave these sorts of decisions up to developers.

dotcom-rendering/src/model/enhance-H3s.ts
```diff
@@ -0,0 +1,92 @@
...
+	const startStrong = html.startsWith('<p><strong>');
+	const endsStrong = html.endsWith('</strong></p>');
+
+	return (
+		hasPwrapper &&
+		containsStrongtags &&
+		doesNotContainLinks &&
+		startStrong &&
+		endsStrong
+	);
```
mxdvl, line 29: If you want to avoid unncessary work, you can follow the pattern of early returns:

```ts
const frag = JSDOM.fragment(element.html);
if (!frag.firstElementChild) return false;

// The following things must be true for an element to be a faux H3
if (frag.firstElementChild.nodeName !== 'P') return false;
if (!frag.firstElementChild.outerHTML.includes('<strong>')) return false;
if (frag.firstElementChild.outerHTML.includes('<a>')) return false;

// note: we don’t need to extract html until now.
const html = frag.firstElementChild.outerHTML;
if (!html.startsWith('<p><strong>')) return false;
if (!html.endsWith('</strong></p>')) return false;

return true;
```

dotcom-rendering/src/model/enhance-H3s.ts
mxdvl: I was unaware of the connotation of anti-pattern. I was trying to indicate that in this case there is an alternative that is more explicit, and therefore [better aligned to the guardian’s “coding with empathy” recommendations](https://github.com/guardian/recommendations/blob/main/coding-with-empathy.md#recommendations).

dotcom-rendering/src/model/enhance-H3s.ts
alinaboghiu: I think the first version is easier to read. 

        hasPwrapper &&
		containsStrongtags &&
		doesNotContainLinks &&
		startStrong &&
		endsStrong

Makes it very clear what needs to be true for the element to be considered an h3.

I believe a list of ORed negations is much more difficult to get one's head around, hence the need for the comment there. 

Since the performance gain in this case it negligible I would prefer keeping the original version if you don't mind?

In general I do agree returning early is preferable.

dotcom-rendering/src/model/enhance-H3s.ts
alinaboghiu: Happy to pair on this today if you have a little time? Anna is not working today.

dotcom-rendering/src/model/enhance-H3s.ts
alinaboghiu: Thank you both! Will change this to a map as it makes sense in this case.

dotcom-rendering/src/model/enhance-H3s.ts
mxdvl: Agree, readability trumps micro-performance optimisations.

Keep it as-is.

mxdvl (approved):

PR #5804, opened 2022-08-23
fix: coerce switches to boolean

mxdvl:
## What does this change?

Cooerce switches to booleans. Undefined switches will be explicitely `false`.

## Why?

Relying on the falsiness of these switches is not as explicit.

These will throw errors when we have `noUncheckedIndexedAccess`.

#5777


ioannakok (approved):
+1!

OllysCoding (approved):
Simple change :)

PR #5805, opened 2022-08-23
Dblatcher/signup placement tweak

dblatcher:
## What does this change?

Updates the Signup block placement logic:
 - The maximum distance from the center is now 3 places - was 4
 - When sorting the possible positions, previous floating elements more than 4 paragraph above the possible insert position are not taken into account
 - when sorting the possible places, prefer places that are further down (but still within 3 places of the center) - before, they were sorted to prefer places closer to the center

## Why?
Editorial noted that on some article the signup block was appearing sooner than they would want it. This was caused by the sorting logic preferring places that were not after a floating element. If a slot before/above the floating element was in range, it would always take that that slot, leading to a tendency to 'go high' on articles with a floating element around the middle of the articles.

Previous floating elements do still need to be taken into account for the reasons outlined [here](https://github.com/guardian/dotcom-rendering/pull/5410#discussion_r930899305), but if the floating element is more than 4 paragraph above, there *should* be no extra spacing (or very little if the floating element is very tall and/or all four paragraphs are very short)

## Screenshots

| Before      | After      |
|-------------|------------|
| BBC- current | bbc-second-change |
| minions-current| minions-second-change |

### DIFF
--- a/dotcom-rendering/src/model/insertPromotedNewsletter.ts
+++ b/dotcom-rendering/src/model/insertPromotedNewsletter.ts
@@ -51,29 +51,30 @@ const MINIMUM_DISTANCE_AFTER_FLOATING_ELEMENT = 3;
 //    | TextTextTextTextTextText  | +------+
 //    | TextTextTextText          |
 
-const MAXIMUM_DISTANCE_AFTER_FLOATING_ELEMENT_TO_TAKE_ACCOUNT_OF_WHEN_SORTING = 4;
-// The maximum distance is set so that the sort function does not always place
-// the SignUp as high as possible to be above a (fairly distant) floating element
-// resulting in the block being higher than is ideal.
-
-//	 0 TEXT
-//	 1 TEXT
-//	 2 TEXT
-//    [too high] <--  Without this rule the SignUp
-//	 3 RICHLINK     | tends to go too high so it's
-//	 4 TEXT         | above the floating element
-//	 5 TEXT         |
-//	 6 TEXT         |
-//	 7 TEXT         |
-//     [target] ----+ Staying here would probably be
-//	 8 TEXT           fine as there are four paragraphs
-//	 9 TEXT           between this spot and the richlink
-//	10 TEXT
-//	11 TEXT
-//	12 TEXT
-//	13 TEXT
-//	14 TEXT
-
+/**
+ * The maximum distance is set so that the sort function does not always place
+ * the SignUp as high as possible to be above a (fairly distant) floating element
+ * resulting in the block being higher than is ideal.
+ * @example
+ *	 0 TEXT
+ *	 1 TEXT
+ *	 2 TEXT
+ *    [too high] <--  Without this rule the SignUp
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/model/insertPromotedNewsletter.ts
```diff
@@ -52,6 +51,29 @@ const MINIMUM_DISTANCE_AFTER_FLOATING_ELEMENT = 4;
...
+//	 7 TEXT         |
+//     [target] ----+ Staying here would probably be
+//	 8 TEXT           fine as there are four paragraphs
+//	 9 TEXT           between this spot and the richlink
+//	10 TEXT
+//	11 TEXT
+//	12 TEXT
+//	13 TEXT
+//	14 TEXT
+
```
mxdvl, line 76: By using JSDoc comments, you get intellisense whenever you use these values!

```suggestion
/**
 * The maximum distance is set so that the sort function does not always place
 * the SignUp as high as possible to be above a (fairly distant) floating element
 * resulting in the block being higher than is ideal.
 * @example
 *	 0 TEXT
 *	 1 TEXT
 *	 2 TEXT
 *    [too high] <--  Without this rule the SignUp
 *	 3 RICHLINK     | tends to go too high so it's
 *	 4 TEXT         | above the floating element
 *	 5 TEXT         |
 *	 6 TEXT         |
 *	 7 TEXT         |
 *     [target] ----+ Staying here would probably be
 *	 8 TEXT           fine as there are four paragraphs
 *	 9 TEXT           between this spot and the richlink
 *	10 TEXT
 *	11 TEXT
 *	12 TEXT
 *	13 TEXT
 *	14 TEXT
 */
const MAXIMUM_DISTANCE_AFTER_FLOATING_ELEMENT = 4;
```

image

dotcom-rendering/src/model/insertPromotedNewsletter.ts
```diff
@@ -72,14 +94,26 @@ const getDistanceAfterFloating = (
 				['supporting', 'thumbnail', 'richLink'].includes(element.role),
 		);
 
+	// it would be more logical to return Infinity here,
+	// but having an outsized finite value simplifies the sort function
+	// (Infinity - Infinity == NaN)
+	const outsizedValue = elements.length * 10;
```
mxdvl, line 100: Same as above regarding JSDoc.

Would `elements.length + 1` be enough, as it’s simply one more than the amount of elements? Otherwise, maybe you want [`Number.MAX_VALUE`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_VALUE)

mxdvl (approved):

dotcom-rendering/src/model/insertPromotedNewsletter.ts
dblatcher: `elements.length +1` would be enough; it just needs to be a number higher than the function could otherwise return. I'll use that instead

PR #5806, opened 2022-08-23
feat: add type-guards for filtering arrays

mxdvl:
## What does this change?

Use type-guards when filtering arrays to get narrower types on the filtered items.

## Why?

These will throw errors when we have `noUncheckedIndexedAccess`.

#5777

### DIFF
--- a/dotcom-rendering/src/web/components/InteractiveContentsBlockComponent.stories.tsx
+++ b/dotcom-rendering/src/web/components/InteractiveContentsBlockComponent.stories.tsx
@@ -3,28 +3,31 @@ import { NumberedList } from '../../../fixtures/generated/articles/NumberedList'
 import { enhanceInteractiveContentsElements } from '../../model/enhance-interactive-contents-elements';
 import { InteractiveContentsBlockComponent } from './InteractiveContentsBlockComponent';
 
-// @ts-expect-error: we know that NumberedList fixture has an interactive content block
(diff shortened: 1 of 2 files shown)


dotcom-rendering/src/web/components/InteractiveContentsBlockComponent.stories.tsx
```diff
@@ -3,28 +3,31 @@ import { NumberedList } from '../../../fixtures/generated/articles/NumberedList'
...
 
-// @ts-expect-error: we know that NumberedList fixture has an interactive content block
-const interactiveContentsBlock: InteractiveContentsBlockElement =
-	enhanceInteractiveContentsElements(NumberedList.blocks)[0].elements.find(
-		(block) =>
-			block._type ===
-			'model.dotcomrendering.pageElements.InteractiveContentsBlockElement',
-	);
+const interactiveContentsBlock = enhanceInteractiveContentsElements(
+	NumberedList.blocks,
```
ioannakok, line 7: I can see now we're getting an ESLint error: `Unsafe member access .blocks on an `any` value`. Is that okay?

ioannakok: Just realised this was already happening!

ioannakok (approved):
+1

PR #5769, opened 2022-08-19
New designs for `Analysis` and `Explainer` articles

alinaboghiu:
## What does this change?

This PR introduces the new designs for Analysis and Explainer articles. This is the main feature branch. Most of these changes were developed and reviewed on separate branches but this could also benefit from a final review.

## Why?

The Editorial Design team have created these new designs for Analysis and Explainer articles. the Journalism team have taken a week to implement these some of these changes in DCR and AR. 

https://docs.google.com/document/d/1MqSldxfgR86kOJaXTFNaBs5LzPzSa7jC4D4YwqqMFwg/edit#

## Screenshots

| Before      | After      |
|-------------|------------|
| ![beforeAnalysis][] | ![afterAnalysis][] |
| ![beforeExplainer][] | ![afterExplainer][] |

[beforeAnalysis]: https://user-images.githubusercontent.com/1229808/185647519-857357f5-2a4e-479e-8ad2-0808bd72b605.png
[afterAnalysis]: https://user-images.githubusercontent.com/1229808/185647549-a3df309a-93c0-4025-9d6f-b42b7513e391.png
[afterExplainer]: https://user-images.githubusercontent.com/1229808/186193753-2d5cca26-b8a0-479a-879c-10f1e5c2eef5.png
[beforeExplainer]: https://user-images.githubusercontent.com/1229808/186193762-1b80fb9c-d1ce-47d5-970d-f03a7c0963d4.png

PR #5808, opened 2022-08-23
Dblatcher/apps rendering data model with newsletter

dblatcher:
## What does this change?
(no user facing changes)
- Upgrades the apps-rendering-api-models package to version 1.1.0 (adds 'promotedNewsletter' field with new type of `Newsletter` to `renderingRequest`)
- Uses the new type in the definition of the `NewsletterSignup` body element
- updates the prop names in the (work in progress)  `NewsletterSignup` component to match

## Why?
When [MAPI is updated](https://github.com/guardian/mobile-apps-api/pull/2061), the 'promotedNewsletter' field will be included on rendering requests when a newsletter promotion tag is on the article. This PR is to bring the preview work on the apps-rendering `NewsletterSignup` in line with the data model.


apps-rendering/src/bodyElement.ts
```diff
@@ -111,13 +112,8 @@ interface AudioAtom {
 	title: string;
 }
 
-interface NewsletterSignUp {
+interface NewsletterSignUp extends Omit<Newsletter, 'theme'> {
```
georgeblahblah, line 115: We have `theme` on both the `Newsletter` type and [`Item`](https://github.com/guardian/libs/blob/main/src/format/ArticleFormat.ts#L6). Are we anticipating that they will both have different values? Do you think there's an opportunity to always use the Article's theme?

apps-rendering/src/components/NewsletterSignup/NewsletterSignup.stories.tsx
```diff
@@ -31,13 +32,13 @@ const NewsTheme: FC = () => (
 	<NewsletterSignup
 		element={{
 			kind: ElementKind.NewsletterSignUp,
-			id: 'monring-briefing',
+			identityName: 'monring-briefing',
```
georgeblahblah, line 35:
```suggestion
			identityName: 'morning-briefing',
```

apps-rendering/src/components/NewsletterSignup/NewsletterSignup.stories.tsx
```diff
@@ -31,13 +32,13 @@ const NewsTheme: FC = () => (
...
 			description:
 				'Archie Bland and Nimo Omer take you through the top stories and what they mean, free every weekday morning',
-			displayName: 'First Edition',
+			name: 'First Edition',
 			frequency: 'Every weekday',
 			theme: ArticlePillar.News,
-			group: 'News',
+			successDescription: "We'll send you first edition every weekday",
```
georgeblahblah, line 41:
```suggestion
			successDescription: "We'll send you First Edition every weekday",
```

georgeblahblah (commented):
Looks good! I had one question about whether the `Newsletter` `theme` property could be inferred from the `Item` theme, or whether they need to be separate.

Additionally, I noticed that the changes to `package-lock.json` include quite a few diffs which don't seem related. Would you mind taking a closer look? It's worth checking the `npm` and `node` versions in use for `apps-rendering`. Mine are:

```
✔︎ dotcom-rendering/apps-rendering % node --version
v14.18.3
✔︎ dotcom-rendering/apps-rendering % npm --version
6.14.15
```

apps-rendering/src/bodyElement.ts
dblatcher: The theme of the Article and the theme of the Newsletter promoted in the article won't necessarily be the same - eg we might promote a 'NEWS' newsletter in an 'OPINION' article . It's the Newsletter's theme that is used to adjust the styling of the sign up component, so we'd need a separate 'theme' field on the `Newsletter`

dblatcher: looking back at this bit, the type definition is a bit hard to follow - the idea is that the `NewsletterSignUp` bodyElement should have all same fields as the `Newsletter` API model, except that the `theme` should be an ArticleTheme enum value rather than a string - it would probably be better to keep it as a string to match the API model and convert it to the enum value as needed.

dblatcher:
> Additionally, I noticed that the changes to `package-lock.json` include quite a few diffs which don't seem related. Would you mind taking a closer look? It's worth checking the `npm` and `node` versions in use for `apps-rendering`. Mine are:
> 
> ```
> ✔︎ dotcom-rendering/apps-rendering % node --version
> v14.18.3
> ✔︎ dotcom-rendering/apps-rendering % npm --version
> 6.14.15
> ```

I've re-synced with main, double checked the versions are the same and re-installed the dependencies. I'm still getting the same diff on `package-lock.json`.

The unexpected changes are all on the integrity hashes, and switch from sha512 to shorter sha1 hashes. Afraid I'm not familiar with how these are generated or why there would be a change. 

If we need to look into this further, do you have any suggestions where to look or who to ask?

georgeblahblah:
> If we need to look into this further, do you have any suggestions where to look or who to ask?

thanks for checking your versions! Generally I wouldn't expect to see hashes going from `sha-512` to `sha-1`. There's a bit more info in [this StackOverflow post](https://stackoverflow.com/questions/47638381/why-did-package-lock-json-change-the-integrity-hash-from-sha1-to-sha512) (ignore the part about `npm i -g npm`!), but what happens if you:

1. reset any changes to `package-lock.json` by taking the version from main
  - from your branch, inside the `apps-rendering` directory, run the following command to copy the `apps-rendering` `package-lock.json` from `main`:
  - `git restore --source main -- package-lock.json`
1. From inside the `apps-rendering` dir, `rm -rf node_modules`
1. again, from inside `apps-rendering`, run `npm install`

When I tried this locally, my `package-lock.json` diff with `main` looked much more sensible with the only change closer to what I would expect for this package upgrade:

```
"@guardian/apps-rendering-api-models": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/@guardian/apps-rendering-api-models/-/apps-rendering-api-models-1.1.0.tgz",
-      "integrity": "sha512-C3JETz5d76oAktdfYhbcf5NbFI9HN7d0ydqhhTs+JhRhxxtedQksPSVg/OPTKlS7Kq4AGiGXpDB4yXIDlwjwaA==",
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/@guardian/apps-rendering-api-models/-/apps-rendering-api-models-1.0.0.tgz",
+      "integrity": "sha512-6PAb0X+Vm8GuTFDiNS/GK5C/TvVXD6+N0CUlD5Tasd1454616YWz18oLfjHO3QlAYDHM4SeuZ9dHkLh6Lnevow==",
```

Happy to look at this together in a pairing! Feel free to ping me or put something in the calendar :)

dblatcher:
> Just to hop in here, these files changed really recently... https://github.com/guardian/dotcom-rendering/pull/5769/files#diff-9b6128b747f27002e5b17b7a3e1e481de39dc7455e40c6e2aa8f54c8efdcea47 so I _think_ it's possible that David is actually changing them _back_ to what they should be in this PR and not the other way around

Thanks both - I followed the George's steps and it has brought the file back in line with main, so I think its sorted. I don't really understand the underlying mechanism but I gather `sha512` is what we want, so I think we're good :)

georgeblahblah:
@cemms1 that's interesting! I hadn't seen the changes in that PR! We've been experiencing this flip-flopping for some time but from what I understand (quite a shallow understanding!) `sha-512` is what we want. There's a longer discussion [here](https://github.com/npm/npm/issues/16938) about why this might happen. We are hopeful that the problem might become less frequent as we upgrade node versions.

@dblatcher that's great! I'm not 100% on the underlying mechanism either, but this diff looks sensible for the changes in `package.json`

georgeblahblah (approved):

apps-rendering/src/item.ts
```diff
@@ -65,6 +66,7 @@ interface Fields extends ArticleFormat {
 	relatedContent: Option<ResizedRelatedContent>;
 	logo: Option<Logo>;
 	webUrl: string;
+	promotedNewsletter?: Newsletter;
```
JamieB-gu, line 69: Can I suggest using an `Option` for this (see the other fields above for examples)? We try to avoid use of optional properties (`A | undefined`) in AR.

```suggestion
	promotedNewsletter: Option<Newsletter>;
```

apps-rendering/src/item.ts
```diff
@@ -307,6 +309,7 @@ const itemFields = (
 		),
 		logo: paidContentLogo(content.tags),
 		webUrl: content.webUrl,
+		promotedNewsletter: request.promotedNewsletter,
```
JamieB-gu, line 312: This will automatically convert to an `Option`.

```suggestion
		promotedNewsletter: fromNullable(request.promotedNewsletter),
```

JamieB-gu (commented):
Looks good! One suggestion.

PR #5812, opened 2022-08-23
Thrift And Scrooge Updates

JamieB-gu:
## Why?

Updates a number of Thrift-related dependencies. This brings all the versions into line with the ones in https://github.com/guardian/content-api-models/pull/210 because we rely heavily on the CAPI libraries in these models.

## Changes

- Upgraded thrift and scrooge plugins and dependencies
- Added `scroogeDisableStrict` to suppress errors about reserved keywords


georgeblahblah (approved):
@frankie297 and I tested this on MAPI CODE yesterday, looks like it's working well! We will merge to `main` and release non-snapshot versions from there

PR #5813, opened 2022-08-23
Create explainer type

abeddow91:
## What does this change?
Adds an explainer type and fixture to apps rendering

## Why?
We have a new design type called Explainer.


marjisound (approved):
Nice 👍

alinaboghiu (approved):

PR #5817, opened 2022-08-23
feat: generate tuples for fixtures

mxdvl:
## What does this change?

Instead of an array of any length, `ImageBlockElement[]`, make it explicit that the array contains a fixed number of elements: e.g. `[ImageBlockElement, ImageBlockElement, ImageBlockElement]` for 3 images.

## Why?

This way we can safely access these elements in tests using them. Preparation for #5777


OllysCoding (approved):
In my head this change was going to affect more than one fixture but it seems it's only applicable to images!

mxdvl:
I think there's also a manual trail. Let me add it.

PR #5757, opened 2022-08-18
Add idle-loading of interactives to `CommercialEndOfQuarterMegaTest`

simonbyford:
## What does this change?

Adds logic to selectively enable the idle-loading of interactives for users in the [end of quarter mega test](https://github.com/guardian/frontend/pull/25396).

More info about this functionality: https://github.com/guardian/dotcom-rendering/pull/4942

## Why?

To test the revenue impact of this change in conjunction with all the other changes from Q1

PR #5815, opened 2022-08-23
A11y add aria expanded sub nav

DanielCliftonGuardian:
## What does this change?

This adds the aria-expanded attribute to the veggie burger "button" with dynamic true / false toggling.

This has been tested with VoiceOver.

This resolves https://github.com/guardian/dotcom-rendering/issues/5048 

## Why?

This was raised by the DAC report  https://drive.google.com/file/d/1hhrWd99v9tB7SNIDRWHy0svS3JyaKxeE/view

## Screenshots

image

PR #5819, opened 2022-08-23
fix(types): import ServerSideTests from config

mxdvl:
## What does this change?

Imports `ServerSideTests` explicitely

## Why?

#5757 was created before the #5616 was created


mxdvl:
Fixed in #5815

simonbyford (approved):

PR #5825, opened 2022-08-24
Moved location of maincontent id to article component

DanielCliftonGuardian:
## What does this change?
Moved location of maincontent id to article component
## Why?
Avoid potential complication from adding new div.


oliverlloyd (approved):
🙌

PR #5830, opened 2022-08-24
refactor: use Object.entries()

mxdvl:
## What does this change?

When we want both keys and values, make this explicit.

## Why?

This also prevents possibly `undefined` values if/when `noUncheckedIndexedAccess` is turned on.


ioannakok (approved):
Nice!

bryophyta (approved):
great stuff!

georgeblahblah (approved):
Looks good! I will check if there's opportunities for similar refactors in AR

PR #5831, opened 2022-08-24
refactor(loadScript): use @guardian/libs version

mxdvl:
## What does this change?

Uses loadScript from `@guardian/libs`

## Why?

The type checks now pass.


ioannakok (approved):
+1

PR #5832, opened 2022-08-24
Add 4px margin to AR headline tag

alinaboghiu:
## What does this change?

Adds a 4px margin below the headline tag in Apps Rendering. 

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[after]: https://user-images.githubusercontent.com/1229808/186411664-ef6dde83-ded2-4e1e-a0e4-f1cb10e17f09.png
[before]: https://user-images.githubusercontent.com/1229808/186411676-3572499d-7f71-4c01-95e8-353ae0fedd1e.png

### DIFF
--- a/apps-rendering/src/components/HeadlineTag/index.tsx
+++ b/apps-rendering/src/components/HeadlineTag/index.tsx
@@ -19,7 +19,6 @@ const headlineTagStyles = (format: ArticleFormat): SerializedStyles => css`
 	display: inline-block;
 	box-decoration-break: clone;
 	padding: 0 0.375rem 0.125rem;
-	margin-bottom: 0.25rem;
 	${from.tablet} {
 		${headline.xxsmall({ fontWeight: 'bold', lineHeight: 'loose' })}
 	}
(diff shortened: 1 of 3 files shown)


apps-rendering/src/components/HeadlineTag/index.tsx
```diff
@@ -19,6 +19,7 @@ const headlineTagStyles = (format: ArticleFormat): SerializedStyles => css`
 	display: inline-block;
 	box-decoration-break: clone;
 	padding: 0 0.375rem 0.125rem;
+	margin-bottom: 0.25rem;
```
georgeblahblah, line 22: I think you could use the `remSpace` helper from Source here!

```suggestion
	margin-bottom: ${remSpace[1]};
```

georgeblahblah (approved):
Looks good!

I had a small suggestion about using the `remSpace` helper from Source. I did a code suggestion, but it will also need the corresponding import in the file so it's not a standalone suggestion, sorry!

apps-rendering/src/components/HeadlineTag/index.tsx
alinaboghiu: ah good catch thank you - will have to move this to the layout files anyway because it doesn't work for interviews so have put back to draft but will go with remSpace 👍

marjisound (approved):

georgeblahblah (approved):

PR #5834, opened 2022-08-24
Add gallery byline and dark mode styles

iainjchambers-guardian:
## What does this change?
Adds the byline to Galleries and dark mode styles

## Why?
Both were missing

## Screenshots

| Before      | After      |
|-------------|------------|
| image | image |


apps-rendering/package-lock.json
```diff
@@ -18632,7 +18632,7 @@
     "which-module": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/which-module/-/which-module-2.0.0.tgz",
-      "integrity": "sha512-B+enWhmw6cjfVC7kS8Pj9pCrKSc5txArRyaYGe088shv/FGWH+0Rjx/xPgtsWfsUtS27FkP697E4DDhgrgoc0Q=="
+      "integrity": "sha1-2e8H3Od7mQK4o6j6SzHD4/fm6Ho="
     },
     "which-typed-array": {
       "version": "1.1.8",
```
JamieB-gu, line 18638: You may want to drop the changes to `package-lock.json`? I'm not sure they're related to this change?

apps-rendering/src/components/BodyImage/GalleryBodyImage.tsx
```diff
@@ -103,6 +104,10 @@ const captionStyles = (format: ArticleFormat): SerializedStyles => css`
 			left: -10px;
 		}
 	}
+
+	${darkModeCss`
+		color: ${neutral[86]};
```
JamieB-gu, line 109: One for the palette? And perhaps the `color` specified a bit higher up?

apps-rendering/src/components/Footer/GalleryFooter.tsx
```diff
@@ -0,0 +1,93 @@
...
+		<DefaultFooter
+			css={css(defaultStyles(format), footerStyles)}
+			isCcpa={isCcpa}
+		/>
+	</div>
+);
+
+// ----- Exports ----- //
+
+export default ImmersiveFooter;
```
JamieB-gu, line 93:
```suggestion
const GalleryFooter: FC<Props> = ({ format, isCcpa }) => (
	<div css={styles(format)}>
		<div css={spacerStyles} />
		<DefaultFooter
			css={css(defaultStyles(format), footerStyles)}
			isCcpa={isCcpa}
		/>
	</div>
);

// ----- Exports ----- //

export default GalleryFooter;
```

apps-rendering/src/components/Footer/GalleryFooter.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+		${darkModeCss`
+			&::before {
+				content: '';
+				position: absolute;
+				top: 0;
+				bottom: 0;
+				border-left: 1px solid ${neutral[20]};
+				left: -10px;
+			}
```
JamieB-gu, line 51: This `before` only exists in dark mode?

apps-rendering/src/components/Footer/GalleryFooter.tsx
```diff
@@ -0,0 +1,93 @@
...
+	padding-right: 0;
+	grid-row: 1;
+
+	${from.desktop} {
+		${grid.between('centre-column-start', 'right-column-end')}
+	}
+
+	${darkModeCss`
+		background-color: ${neutral[10]};
+		color: ${neutral[86]};
```
JamieB-gu, line 72: Two for the palette?

```suggestion
		background-color: ${background.footerDark(format)};
		color: ${text.footerDark(format)};
```

apps-rendering/src/components/Headline/GalleryHeadline.tsx
```diff
@@ -28,6 +29,11 @@ const styles = css`
 		${headline.xlarge({ fontWeight: 'bold' })}
 		${grid.span('centre-column-start', 8)}
 	}
+
+	${darkModeCss`
+		background-color: ${neutral[10]};
+		color: ${neutral[86]};
```
JamieB-gu, line 35: A couple for the palette? And perhaps the `color` above?

```suggestion
		background-color: ${background.headlineDark(format)};
		color: ${text.headlineDark(format)};
```

apps-rendering/src/components/Headline/GalleryHeadline.tsx
```diff
@@ -43,6 +49,10 @@ const backgroundStyles = css`
 		${grid.between('centre-column-start', 'viewport-end')}
 		margin-left: calc(${grid.columnGap} * -1/2);
 	}
+
+	${darkModeCss`
+		background-color: ${neutral[10]};
```
JamieB-gu, line 54: One for the palette?

```suggestion
		background-color: ${background.headlineDark(format)};
```

JamieB-gu (commented):
Looks good!

A number of small comments, mainly about using `editorialPalette`.

apps-rendering/src/components/Standfirst/GalleryStandfirst.tsx
```diff
@@ -24,16 +27,60 @@ const styles = css`
...
+				{rendered}
+				<address css={bylineStyles}>
+					<p>by {renderStandfirstText(bylineHtml.value, format)}</p>
+				</address>
+			</>
+		);
+	}
+
+	return rendered;
+};
```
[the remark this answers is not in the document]
paperboyo: They shouldn’t (I know of no examples). And the example given should be sent to Central Prod as I can’t see a reason why it’s been misedited such! (there were some genuine reasons where the byline was used in standfirst as a part of a sentence, but not here).

Galleries have their own silly quirks:
1. Sometimes extensive time-sensitive info is written [into first](https://www.theguardian.com/artanddesign/gallery/2020/sep/16/the-other-side-of-rene-burri-swiss-photographer-in-pictures) (or last) caption, because Editorial feels it would be too big as part of standfirst.
2. Sometimes, All photographs by, instead of being written into Byline (like [here](https://www.theguardian.com/stage/gallery/2020/sep/16/from-top-girls-to-the-skriker-the-best-of-caryl-churchill-in-pictures-tristram-kenton)), is written [into first](https://www.theguardian.com/artanddesign/gallery/2020/sep/16/the-other-side-of-rene-burri-swiss-photographer-in-pictures), [or last](https://www.theguardian.com/artanddesign/gallery/2020/oct/10/mesmerising-urban-photography-in-pictures), image caption (no idea what’s the reasoning, tried to fight it, lost).
3. Editorial expressed a need for a footer for when a legal change had to be communicated (just once).
4. Dotcom is doing weird stuff enhancing first emboldened para into an H2 (because Composer Gallery editor doesn’t allow H2, ma do in the future).

Some of the above are Editorial (bad) habits. But platforms and tools (pt. 4) can maybe help alleviate just some of these…

apps-rendering/package-lock.json
iainjchambers-guardian: Grr this sha issues...

apps-rendering/src/components/Standfirst/GalleryStandfirst.tsx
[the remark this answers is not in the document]
iainjchambers-guardian: Ah ok I think I see what's happening here. So I've made it more complex that it needs to be by using the immersive work as a base. I'm thinking I can just separate this out and add a `<GalleryByline />` component. Sound good?

apps-rendering/package-lock.json
iainjchambers-guardian: Have removed the changes to package-lock by checking out the version on main

apps-rendering/src/components/Byline/GalleryByline.tsx
```diff
@@ -0,0 +1,29 @@
...
+import type { Option } from '@guardian/types';
+import { grid } from 'grid/grid';
+import { defaultAnchorStyles, DefaultByline } from './Byline.defaults';
+
+const bylineStyles = css`
+	padding-top: ${remSpace[2]};
+	padding-bottom: ${remSpace[2]};
+	${grid.span('centre-column-start', 3)}
+	grid-row: 6/7;
+`;
```
JamieB-gu, line 13: Out of curiosity, how do the gallery byline text colour and font get set?

apps-rendering/src/components/Standfirst/GalleryStandfirst.tsx
```diff
@@ -24,16 +27,24 @@ const styles = css`
...
+
+	${darkModeCss`
+		color: ${text.standfirstDark(format)};
+	`}
 `;
 
 type Props = {
 	standfirst: Option<DocumentFragment>;
+	byline: string;
+	bylineHtml: Option<DocumentFragment>;
```
JamieB-gu, line 39: Are these still needed?

JamieB-gu (commented):
Thanks for making all those palette changes!

Couple of final small questions.

apps-rendering/src/components/Standfirst/GalleryStandfirst.tsx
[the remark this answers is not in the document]
JamieB-gu: > I'm thinking I can just separate this out and add a `<GalleryByline />` component. Sound good?

Sounds good, it seems we don't need to worry about this problem for galleries.

@paperboyo couple of questions on (3) and (4). On 3), what sort of footer? Do you have an example piece we could look at to make sure this works? On (4) are we talking about paragraphs within captions? AR has logic to do this too, although I wonder if a `<strong>` (which is what CAPI gives us) might be more appropriate?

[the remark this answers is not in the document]
paperboyo: Ad. 3. Sadly, can’t find an example. Will fwd you the only request that was made about it. It didn’t go anywhere, but the need may be there. Something to ponder only, really.
Ad. 4. Yes, emboldened (first) paras within captions. One might think, Composer should add H2s to image captions. But practically, I’m not sure: it would need to be restricted to one, opening H2 (sic!) and would require massive habit change for Editorial. Maybe we should just live with the silly cleaner. Although it consistently barfs for when something else is emboldened further down or copy/paste fails, haha, see [here](https://www.theguardian.com/news/gallery/2022/aug/24/ukraine-independence-day-and-reading-festival-wednesdays-best-photos#img-11)

image

CAPI:
image

apps-rendering/src/components/Standfirst/GalleryStandfirst.tsx
iainjchambers-guardian: Nope, well spotted

apps-rendering/src/components/Byline/GalleryByline.tsx
iainjchambers-guardian: It was being set by the anchor styles, I noticed that the DCR galleries didn't have the word "by" before the anchor. That said, I've set some styles explicitly on the parent just in case

JamieB-gu (approved):

PR #5837, opened 2022-08-24
Refactor Object.keys => Object.entries

georgeblahblah:
## What does this change?

- Uses `Object.entries` instead of `Object.keys`, prompted by #5830 
- This function uses both the keys & values

Tested on CODE


JamieB-gu (approved):
Might be worth checking this functionality still works as expected in a live example?

PR #5841, opened 2022-08-24
Ar table of content component

marjisound:
## What does this change?
This Pr creates `TableOfContent` component and adds this component on `Explainer` articles. 

The table of content is opened by default unless if the number of headings are more than 5 items. 

## Why?
The table of content in Explainer articles has been requested by the Editorial Design team. More details can be found here https://docs.google.com/document/d/1MqSldxfgR86kOJaXTFNaBs5LzPzSa7jC4D4YwqqMFwg/edit#

### Desktop
| Before      | After (Initially opened)     |
|-------------|------------|
| image | image |

| Before      | After (Initially closed)    | Opened      |
| image | image | image |

### Mobile
| Before      | After      |
|-------------|------------|
| image | image |

### Dark mode
| opened      | closed     |
|-------------|------------|
| image | image |

### DIFF
--- a/apps-rendering/src/components/Layout/Layout.stories.tsx
+++ b/apps-rendering/src/components/Layout/Layout.stories.tsx
@@ -13,6 +13,7 @@ import {
 	article,
 	comment,
 	editorial,
+	explainer,
 	feature,
 	interview,
 	letter,
@@ -197,6 +198,18 @@ export const Analysis = (): React.ReactNode => {
 	);
 };
 Analysis.story = { name: 'Analysis' };
+
+export const Explainer = (): React.ReactNode => {
+	return (
+		<Standard item={explainer}>
+			{renderAll(
+				formatFromItem(analysis, some(ArticleDisplay.Standard)),
+				partition(analysis.body).oks,
+			)}
+		</Standard>
+	);
+};
+Explainer.story = { name: 'Explainer' };
 
 export const LiveBlog = (): ReactElement => (
 	<Live
--- a/apps-rendering/src/components/Layout/Layout.stories.tsx
+++ b/apps-rendering/src/components/Layout/Layout.stories.tsx
@@ -203,7 +203,7 @@ export const Explainer = (): React.ReactNode => {
 	return (
 		<Standard item={explainer}>
 			{renderAll(
-				formatFromItem(analysis, some(ArticleDisplay.Standard)),
+				formatFromItem(explainer, some(ArticleDisplay.Standard)),
 				partition(explainer.body).oks,
 			)}
 		</Standard>
--- a/apps-rendering/src/components/Layout/Layout.stories.tsx
+++ b/apps-rendering/src/components/Layout/Layout.stories.tsx
@@ -204,7 +204,7 @@ export const Explainer = (): React.ReactNode => {
 		<Standard item={explainer}>
 			{renderAll(
 				formatFromItem(analysis, some(ArticleDisplay.Standard)),
-				partition(analysis.body).oks,
+				partition(explainer.body).oks,
 			)}
 		</Standard>
 	);
--- a/apps-rendering/src/components/Layout/StandardLayout.tsx
+++ b/apps-rendering/src/components/Layout/StandardLayout.tsx
@@ -135,11 +135,14 @@ const StandardLayout: FC<Props> = ({ item, children }) => {
 						<Metadata item={item} />
 						<Logo item={item} />
 					</section>
+
 					{item.design === ArticleDesign.Explainer && (
-						<TableOfContent
-							format={getFormat(item)}
-							outline={item.outline}
-						/>
+						<section css={articleWidthStyles}>
+							<TableOfContent
+								format={getFormat(item)}
+								outline={item.outline}
+							/>
+						</section>
 					)}
 				</header>
 				<Body className={[articleWidthStyles]} format={item}>
--- a/apps-rendering/src/components/Layout/StandardLayout.tsx
+++ b/apps-rendering/src/components/Layout/StandardLayout.tsx
@@ -41,6 +41,7 @@ import {
 	onwardStyles,
 } from 'styles';
 import { themeToPillarString } from 'themeStyles';
+import TableOfContent from 'components/TableOfContent';
 
 // ----- Styles ----- //
 const backgroundStyles = (format: ArticleFormat): SerializedStyles => css`
@@ -134,6 +135,12 @@ const StandardLayout: FC<Props> = ({ item, children }) => {
 						<Metadata item={item} />
 						<Logo item={item} />
 					</section>
+					{item.design === ArticleDesign.Explainer && (
+						<TableOfContent
+							format={getFormat(item)}
+							outline={item.outline}
+						/>
+					)}
 				</header>
 				<Body className={[articleWidthStyles]} format={item}>
 					{children}
--- a/apps-rendering/src/components/Layout/StandardLayout.tsx
+++ b/apps-rendering/src/components/Layout/StandardLayout.tsx
@@ -23,6 +23,7 @@ import Metadata from 'components/Metadata';
 import RelatedContent from 'components/RelatedContent';
 import Series from 'components/Series';
 import Standfirst from 'components/Standfirst';
+import TableOfContent from 'components/TableOfContent';
 import Tags from 'components/Tags';
 import { getFormat } from 'item';
 import type {
@@ -41,7 +42,6 @@ import {
 	onwardStyles,
 } from 'styles';
 import { themeToPillarString } from 'themeStyles';
-import TableOfContent from 'components/TableOfContent';
 
 // ----- Styles ----- //
 const backgroundStyles = (format: ArticleFormat): SerializedStyles => css`
--- a/apps-rendering/src/components/Layout/StandardLayout.tsx
+++ b/apps-rendering/src/components/Layout/StandardLayout.tsx
@@ -136,14 +136,15 @@ const StandardLayout: FC<Props> = ({ item, children }) => {
 						<Logo item={item} />
 					</section>
 
-					{item.design === ArticleDesign.Explainer && (
-						<section css={articleWidthStyles}>
-							<TableOfContent
-								format={getFormat(item)}
-								outline={item.outline}
-							/>
-						</section>
-					)}
+					{item.design === ArticleDesign.Explainer &&
+						item.outline.length > 0 && (
+							<section css={articleWidthStyles}>
+								<TableOfContent
+									format={getFormat(item)}
+									outline={item.outline}
+								/>
+							</section>
+						)}
 				</header>
 				<Body className={[articleWidthStyles]} format={item}>
 					{children}
--- a/apps-rendering/src/components/Layout/StandardLayout.tsx
+++ b/apps-rendering/src/components/Layout/StandardLayout.tsx
@@ -23,7 +23,7 @@ import Metadata from 'components/Metadata';
 import RelatedContent from 'components/RelatedContent';
 import Series from 'components/Series';
 import Standfirst from 'components/Standfirst';
-import TableOfContent from 'components/TableOfContent';
+import TableOfContents from 'components/TableOfContents';
 import Tags from 'components/Tags';
 import { getFormat } from 'item';
 import type {
@@ -139,7 +139,7 @@ const StandardLayout: FC<Props> = ({ item, children }) => {
 					{item.design === ArticleDesign.Explainer &&
 						item.outline.length > 0 && (
 							<section css={articleWidthStyles}>
-								<TableOfContent
+								<TableOfContents
 									format={getFormat(item)}
 									outline={item.outline}
 								/>
--- a/apps-rendering/src/components/TableOfContent/TableOfContent.stories.tsx
+++ b/apps-rendering/src/components/TableOfContent/TableOfContent.stories.tsx
@@ -1,4 +1,3 @@
-import { css } from '@emotion/react';
 import {
 	ArticleDesign,
 	ArticleDisplay,
@@ -15,12 +14,8 @@ const format: ArticleFormat = {
 	theme: ArticlePillar.News,
 };
 
-const wrapperStyles = css``;
-
 const Default = (): ReactElement => (
-	<div css={wrapperStyles}>
-		<TableOfContent outline={explainer.outline} format={format} />
-	</div>
+	<TableOfContent outline={explainer.outline} format={format} />
 );
 
 export default {
--- a/apps-rendering/src/components/TableOfContent/TableOfContent.stories.tsx
+++ b/apps-rendering/src/components/TableOfContent/TableOfContent.stories.tsx
@@ -1,12 +1,8 @@
-import {
-	ArticleDesign,
-	ArticleDisplay,
-	ArticleFormat,
-	ArticlePillar,
-} from '@guardian/libs';
+import type { ArticleFormat } from '@guardian/libs';
+import { ArticleDesign, ArticleDisplay, ArticlePillar } from '@guardian/libs';
 import { explainer } from 'fixtures/item';
-import { ReactElement } from 'react';
-import TableOfContent from '.';
+import type { ReactElement } from 'react';
+import TableOfContent from './';
 
 const format: ArticleFormat = {
 	design: ArticleDesign.Standard,
--- a/apps-rendering/src/components/TableOfContent/TableOfContent.stories.tsx
+++ b/apps-rendering/src/components/TableOfContent/TableOfContent.stories.tsx
@@ -1,0 +1,31 @@
+import { css } from '@emotion/react';
+import {
+	ArticleDesign,
+	ArticleDisplay,
+	ArticleFormat,
+	ArticlePillar,
+} from '@guardian/libs';
+import { explainer } from 'fixtures/item';
+import { ReactElement } from 'react';
+import TableOfContent from '.';
+
+const format: ArticleFormat = {
+	design: ArticleDesign.Standard,
+	display: ArticleDisplay.Standard,
+	theme: ArticlePillar.News,
+};
+
+const wrapperStyles = css``;
+
+const Default = (): ReactElement => (
+	<div css={wrapperStyles}>
+		<TableOfContent outline={explainer.outline} format={format} />
+	</div>
+);
+
+export default {
+	component: TableOfContent,
+	title: 'AR/TableOfContent',
+};
+
+export { Default };
--- a/apps-rendering/src/components/TableOfContent/index.tsx
+++ b/apps-rendering/src/components/TableOfContent/index.tsx
@@ -1,0 +1,57 @@
+import { ArticleFormat } from '@guardian/libs';
+import Anchor from 'components/Anchor';
+import ListItem from 'components/ListItem';
+import OrderedList from 'components/OrderedList';
+import { Outline } from 'outline';
+import { FC } from 'react';
+
+interface Props {
+	format: ArticleFormat;
+	outline: Outline;
+}
+
+interface SubContentsProps {
+	format: ArticleFormat;
+	subheadings: {
+		id: string;
+		doc: Node;
+	}[];
+}
+
+const SubContents: FC<SubContentsProps> = ({ format, subheadings }) => {
+	return (
+		<OrderedList>
+			{subheadings.map((heading) => (
+				<ListItem format={format}>
+					<Anchor format={format} href={'#' + heading.id}>
+						{heading.doc.textContent}
+					</Anchor>
+				</ListItem>
+			))}
+		</OrderedList>
+	);
+};
+
+const TableOfContent: FC<Props> = ({ format, outline }) => {
+	return (
+		<div>
+			<OrderedList>
+				{outline.map((ol) => (
+					<ListItem format={format}>
+						<Anchor format={format} href={'#' + ol.id}>
+							{ol.doc.textContent}
+						</Anchor>
+						{ol.subheadings && (
+							<SubContents
+								format={format}
+								subheadings={ol.subheadings}
+							/>
+						)}
+					</ListItem>
+				))}
+			</OrderedList>
+		</div>
+	);
+};
+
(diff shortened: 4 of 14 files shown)


marjisound:
[Ar table of content component](https://trello.com/c/gcOfvLfj/554-ar-table-of-content-component)

apps-rendering/src/components/Layout/StandardLayout.tsx
```diff
@@ -134,6 +135,16 @@ const StandardLayout: FC<Props> = ({ item, children }) => {
 						<Metadata item={item} />
 						<Logo item={item} />
 					</section>
+
+					{item.design === ArticleDesign.Explainer &&
+						item.outline.length > 0 && (
```
abeddow91, line 140: Will `outline` always exist as a property of item? If not I think we need to check outline exists first before checking the length. I think `item.outline?.length > 0` should work

apps-rendering/src/components/TableOfContent/TableOfContent.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { ReactElement } from 'react';
+import TableOfContent from '.';
+
+const format: ArticleFormat = {
+	design: ArticleDesign.Standard,
+	display: ArticleDisplay.Standard,
+	theme: ArticlePillar.News,
+};
+
+const wrapperStyles = css``;
```
abeddow91, line 18: Can we remove this if its empty styles?

apps-rendering/src/components/TableOfContent/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+				<span className="is-on" css={arrowPosition}>
+					<SvgChevronUpSingle size="xsmall" />
+				</span>
+			</summary>
+			<OrderedList className={listStyles}>
+				{outline.map((ol) => (
+					<ListItem className={listItemStyles}>
+						<Anchor
+							format={format}
+							href={'#' + ol.id}
```
abeddow91, line 113: Could we interpolate this instead? `#${ol.id}`

apps-rendering/src/components/TableOfContent/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+const arrowPosition: SerializedStyles = css`
+	position: absolute;
+	right: ${remSpace[1]};
+	top: 0;
+`;
+
+const TableOfContent: FC<Props> = ({ format, outline }) => {
+	return (
+		<details open={outline.length < 5} css={detailsStyles}>
```
abeddow91, line 98: nice!

apps-rendering/src/components/TableOfContent/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+							{ol.doc.textContent}
+						</Anchor>
+					</ListItem>
+				))}
+			</OrderedList>
+		</details>
+	);
+};
+
+export default TableOfContent;
```
abeddow91, line 125: This looks great! Do we need to hydrate this to use on the client?

apps-rendering/src/fixtures/item.ts
```diff
@@ -85,6 +94,18 @@ const docFixture = (): Node => {
 	return doc.firstChild!;
 };
 
+const elementFixture = (element: string, innerText: string): Node => {
```
abeddow91, line 97: This is a great edition! thanks!

abeddow91 (commented):
A few comments / questions but this looks really great!

apps-rendering/src/components/TableOfContent/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+	svg {
+		height: 2rem;
+	}
+`;
+
+const titleStyle: SerializedStyles = css`
+	${textSans.xsmall({ lineHeight: 'regular' })}
+	color: ${neutral[46]};
+	${darkModeCss`
+		color: ${neutral[86]};
```
JamieB-gu, line 86: Could these live in the `text` palette?

apps-rendering/src/components/TableOfContent/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+			</summary>
+			<OrderedList className={listStyles}>
+				{outline.map((ol) => (
+					<ListItem className={listItemStyles}>
+						<Anchor
+							format={format}
+							href={'#' + ol.id}
+							className={anchorStyles(format)}
+						>
+							{ol.doc.textContent}
```
JamieB-gu, line 116: I think it's possible we want to run this through a rendering step that's similar to what we use for the `h2`s in the body? If we just use the `textContent` we'll lose any emphasis, superscript, subscript etc.?

It may not necessarily be an identical rendering step, however. For example, in the body we allow anchors in the `h2`, whereas if we allow those here they'll conflict with the enclosing anchor.

JamieB-gu (commented):
This looks great! A few small comments.

apps-rendering/src/components/TableOfContent/index.tsx
JamieB-gu: Are there client-side features planned for this component?

apps-rendering/src/components/Layout/StandardLayout.tsx
marjisound: A great question. @JamieB-gu please correct me if I'm not explaining correctly. 
Item can be one of the following types:
```
type Item =
	| LiveBlog
	| DeadBlog
	| Review
	| Comment
	| Standard
	| Interactive
	| MatchReport
	| Letter
	| Obituary
	| Editorial
	| Correction
	| Interview
	| Analysis
	| Explainer;
```
And Explainer interface looks like this:
```
interface Explainer extends Fields {
	design: ArticleDesign.Explainer;
	body: Body;
	outline: Outline;
}
```

So in this case when we check `item.design === ArticleDesign.Explainer` we are sure that we're dealing with only the `Explainer` Item type and `outline` is a required field in this type.

apps-rendering/src/components/TableOfContent/index.tsx
marjisound: the # needs to be a string I think. I could also make it work this way:
```
href={`#${ol.id}`}
```

Or I could also move it to a function like this: 
```
const getId = (heading: {id: string}) => `#${heading.id}`;

href={getId(ol)}
```

Which do you think would be clearer? Or if you know a better way?

abeddow91: Aha yes I meant the first way, i think github formatted by text because of the backticks . I meant 
href={`#${ol.id}`}

apps-rendering/src/components/TableOfContent/index.tsx
abeddow91: Perfect, no I don't think so

apps-rendering/src/components/TableOfContent/index.tsx
marjisound: Thanks @JamieB-gu for your suggestion, it's a valid point. I did add a new `TocTextElement` within the table of content which is now rendering the node considering the nested sub elements. Could you check if it looks correct or I couldv'e done it in a better way? 

These changes can be found in this commit [render toc contents with their existing nested html elemenst](https://github.com/guardian/dotcom-rendering/pull/5841/commits/0e8de40ff2788bb943a60e9db54fabb4ef91dc10)

apps-rendering/src/components/TableOfContent/index.tsx
marjisound: I added `tableOfContentsTitle` & `tableOfContentsTitleDark` in the `text` pallet

apps-rendering/src/components/TableOfContent/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+`;
+
+const listItemStyles: SerializedStyles = css`
+	${textSans.xsmall({ fontWeight: 'bold', lineHeight: 'regular' })}
+	border-bottom: ${line.primary} 1px solid;
+	padding-top: ${remSpace[2]};
+`;
+
+const detailsStyles: SerializedStyles = css`
+	margin-bottom: 1.25rem;
```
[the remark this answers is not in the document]
marjisound: Done

apps-rendering/src/components/Layout/Layout.stories.tsx
```diff
@@ -197,6 +198,18 @@ export const Analysis = (): React.ReactNode => {
 };
 Analysis.story = { name: 'Analysis' };
 
+export const Explainer = (): React.ReactNode => {
+	return (
+		<Standard item={explainer}>
+			{renderAll(
+				formatFromItem(analysis, some(ArticleDisplay.Standard)),
+				partition(analysis.body).oks,
```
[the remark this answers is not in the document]
marjisound: oops, good spot :)

apps-rendering/src/components/TableOfContent/index.tsx
marjisound: Great, done :)

apps-rendering/src/components/TableOfContent/TableOfContent.stories.tsx
marjisound: Good spot, done

JamieB-gu (commented):
Thanks for making those changes, just a few more comments (sorry!).

apps-rendering/src/components/TableOfContent/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+	const children = Array.from(node.childNodes).map((item, i) => {
+		console.log(`I'm children node ${item.nodeName}`);
+		return <TocTextElement node={item} key={i} />;
+	});
+
+	console.log(`I'm node ${node.nodeName}`);
+
+	switch (node.nodeName) {
+		case 'H2':
+			return <span>{children}</span>;
```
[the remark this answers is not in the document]
marjisound: None of the above worked, for the first option it's erroring with this message:

`Type 'Element[]' is missing the following properties from type 'ReactElement<any, string | JSXElementConstructor<any>>': type, props, key`

For the second option the error message is:
`Type '{ children: EmotionJSX.Element[]; }' is not assignable to type 'ReactElement<any, string | JSXElementConstructor<any>>'.
  Object literal may only specify known properties, and 'children' does not exist in type 'ReactElement<any, string | JSXElementConstructor<any>>'.`

But it works if I do this: `return <>children</>;`

Shall I go with the empty fragment option?

apps-rendering/src/components/TableOfContent/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+					<ListItem className={listItemStyles}>
+						<Anchor
+							format={format}
+							href={`#${ol.id}`}
+							className={anchorStyles(format)}
+						>
+							<TocTextElement
+								node={ol.doc}
+								key={0}
+							></TocTextElement>
```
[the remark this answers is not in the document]
marjisound: Thanks, by the way, do you think using 0 for key is OK? I'm so confused about this key thing in react 🤷

apps-rendering/src/components/Layout/Layout.stories.tsx
```diff
@@ -197,6 +198,18 @@ export const Analysis = (): React.ReactNode => {
 };
 Analysis.story = { name: 'Analysis' };
 
+export const Explainer = (): React.ReactNode => {
+	return (
+		<Standard item={explainer}>
+			{renderAll(
+				formatFromItem(analysis, some(ArticleDisplay.Standard)),
```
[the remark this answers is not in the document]
marjisound: Sorry I had missed this, done now

apps-rendering/src/components/TableOfContent/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+	node,
+	key,
+}): ReactElement => {
+	const text = node.textContent ?? '';
+	const children = Array.from(node.childNodes).map((item, i) => {
+		console.log(`I'm children node ${item.nodeName}`);
+		return <TocTextElement node={item} key={i} />;
+	});
+
+	console.log(`I'm node ${node.nodeName}`);
```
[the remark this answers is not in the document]
marjisound: Done

apps-rendering/src/fixtures/item.ts
```diff
@@ -86,6 +95,39 @@ const docFixture = (): Node => {
...
+const h2ElementWithSub = () => {
+	{
+		const doc = new DocumentFragment();
+
+		const el = document.createElement('H2');
+		const sub = document.createElement('SUB');
+		sub.innerText = 'student';
+
+		const el2 = document.createElement('text');
+		el2.innerText = 'loan forgiveness? ';
```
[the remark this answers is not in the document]
marjisound: Nice one Jamie, `createTextNode` is what I needed 👍

apps-rendering/src/components/Layout/StandardLayout.tsx
```diff
@@ -134,6 +135,16 @@ const StandardLayout: FC<Props> = ({ item, children }) => {
...
+							<section css={articleWidthStyles}>
+								<TableOfContents
```
georgeblahblah, line 142: should this also be added to the Analysis layout?

apps-rendering/src/components/TableOfContents/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+const summaryStyles: SerializedStyles = css`
+	cursor: pointer;
+	position: relative;
+	list-style: none;
+	align-items: center;
+	padding-left: ${remSpace[3]};
+	padding-top: 0.44rem;
+	padding-bottom: 0.375rem;
+	border-bottom: ${line.primary} 1px solid;
+	border-top: ${line.primary} 1px solid;
```
georgeblahblah, line 78: It's possible that it doesn't matter, but I think the CSS spec indicates the following order for these properties: `width style color`, which would give:

```suggestion
	border-bottom: 1px solid ${line.primary};
	border-top: 1px solid ${line.primary};
```

There's a bit more from the spec on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/border#formal_syntax).

apps-rendering/src/components/TableOfContents/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+	margin: 0;
+
+	> li {
+		padding-left: ${remSpace[3]};
+	}
+`;
+
+const listItemStyles: SerializedStyles = css`
+	${textSans.xsmall({ fontWeight: 'bold', lineHeight: 'regular' })}
+	border-bottom: ${line.primary} 1px solid;
```
georgeblahblah, line 54: similar comment as below for ordering the properties as `width style color`

apps-rendering/src/components/TableOfContents/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+	summary::-webkit-details-marker {
+		display: none;
+	}
+`;
+
+const summaryStyles: SerializedStyles = css`
+	cursor: pointer;
+	position: relative;
+	list-style: none;
+	align-items: center;
```
georgeblahblah, line 73: I think this property is not needed, as this element isn't a flex or grid container

```suggestion
```

apps-rendering/src/components/TableOfContents/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+	node: Node;
+	key: number;
+}
+
+const anchorStyles = (format: ArticleFormat): SerializedStyles => css`
+	color: ${text.paragraph(format)};
+	border-bottom: none;
+	:hover {
+		border-bottom: 0.0625rem solid ${neutral[86]};
+	}
```
georgeblahblah, line 37: Perhaps I'm missing a use-case here, but when do we expect to see the `:hover` pseudoclass activated from mobile devices?

georgeblahblah (commented):
This is looking great! I had a a couple of non-blocking comments, and some suggestions

apps-rendering/src/components/Layout/StandardLayout.tsx
georgeblahblah: ignore this, I remembered the PR description specifying Explainer articles only

apps-rendering/src/components/TableOfContent/index.tsx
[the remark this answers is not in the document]
JamieB-gu: Isn't that the second option ^, or have I missed something? 😅

apps-rendering/src/components/TableOfContents/index.tsx
JamieB-gu: Mobile devices can have pointers too! 🙂

https://developer.apple.com/design/human-interface-guidelines/inputs/pointing-devices/
https://support.apple.com/en-us/HT210546

apps-rendering/src/components/TableOfContent/index.tsx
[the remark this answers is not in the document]
JamieB-gu: This is a technical look at why keys are important in React: https://reactjs.org/docs/reconciliation.html#recursing-on-children

This is an overview of how to use keys with lists: https://reactjs.org/docs/lists-and-keys.html

In summary, my understanding is that they're a way of making re-renders more efficient, by only updating parts of a list that have changed, rather than re-rendering the whole list. I think this is unlikely to matter on the server-side, where we only render once, but you get the warnings anyway.

Strictly speaking, you're best not using an array index as the key. As mentioned in both those doc pages, I think this is because it can confuse React if the list gets re-ordered. However, if this list is only rendered server-side and never gets re-ordered, then this probably fine.

Alternatively, if there's a way to turn off these warnings on the server-side, that might be preferable; but I've never come across one.

Just cc'ing @arelra @oliverlloyd @mxdvl to see if they have any further thoughts on this as knowledgeable React people ⚛️.

apps-rendering/src/components/TableOfContent/index.tsx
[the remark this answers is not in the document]
marjisound: OMG my eyes had seen your second option as just `{children}` 😅 so done then 👍

apps-rendering/src/components/TableOfContent/index.tsx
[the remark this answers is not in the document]
marjisound: Thanks @JamieB-gu , very well explained. I made some changes and am now using the h2 id for the key. Though, because the `TocTextElement` is recursive, the nested elements within the h2, such as sub or sup would also needed the key as well. I had to use the index of the child nodes for that: `<TocTextElement node={item} key={i.toString()} />`

apps-rendering/src/components/TableOfContents/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+				<h2 css={titleStyle(format)}>Jump to...</h2>
+				<span className="is-off" css={arrowPosition}>
+					<SvgChevronDownSingle size="xsmall" />
+				</span>
+				<span className="is-on" css={arrowPosition}>
+					<SvgChevronUpSingle size="xsmall" />
+				</span>
+			</summary>
+			<OrderedList className={listStyles}>
+				{outline.map((ol, index) => (
```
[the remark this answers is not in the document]
marjisound: Nice, done 👍

apps-rendering/src/components/Layout/StandardLayout.tsx
marjisound: This is a good question, I was also not sure and because I couldn't find any Analysis article with h2 tags, I assumed they may not need the TOC. I'll talk to Harry tomorrow to confirm, thanks for pointing out 👍

apps-rendering/src/components/TableOfContents/index.tsx
marjisound: Done

apps-rendering/src/components/TableOfContents/index.tsx
marjisound: Done

apps-rendering/src/components/TableOfContents/index.tsx
marjisound: Thanks @georgeblahblah Done

apps-rendering/src/components/Layout/StandardLayout.tsx
marjisound: Harry in on holiday, and we aren't sure if he wants the table of content on analysis articles at this stage or not. So to be on the safe side, we can just release the explainer version until he's back

apps-rendering/src/components/TableOfContents/index.tsx
georgeblahblah: D'oh, of course! I think I wrote this in response to Firefox's responsive viewer supressing the `:hover` states, but it makes sense to keep this selector!

georgeblahblah (approved):
Looks great to me! I can't wait to see this live 🎊

apps-rendering/src/components/TableOfContents/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+	summary::-webkit-details-marker {
+		display: none;
+	}
+`;
+
+const summaryStyles: SerializedStyles = css`
+	cursor: pointer;
+	position: relative;
+	list-style: none;
+	//align-items: center;
```
JamieB-gu, line 73:
```suggestion
```

apps-rendering/src/components/TableOfContents/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+				<span className="is-off" css={arrowPosition}>
+					<SvgChevronDownSingle size="xsmall" />
+				</span>
+				<span className="is-on" css={arrowPosition}>
+					<SvgChevronUpSingle size="xsmall" />
+				</span>
+			</summary>
+			<OrderedList className={listStyles}>
+				{outline.map((outlineItem, index) => (
+					<ListItem className={listItemStyles} key={index}>
```
JamieB-gu, line 173:
```suggestion
					<ListItem className={listItemStyles} key={outlineItem.id}>
```

JamieB-gu (approved):
Couple of small suggestions, otherwise looks good!

PR #5849, opened 2022-08-25
fix: safe `performance.getEntriesByType`

mxdvl:
## What does this change?

Some browsers do not support `window.performance.getEntriesByType`, so we check for its existence before measuring the time taken.

## Why?

This is the top reported `TypeError` in Sentry.

Fixes CLIENT-SIDE-PROD-CSXR
Fixes CLIENT-SIDE-PROD-CTRH

This method was Introduced in #5298.
Related to the fix in #5299

### DIFF
--- a/dotcom-rendering/src/web/browser/islands/doHydration.ts
+++ b/dotcom-rendering/src/web/browser/islands/doHydration.ts
@@ -48,6 +48,8 @@ export const doHydration = (
 			return { clientOnly, timeTaken };
 		})
 		.then(({ clientOnly, timeTaken }) => {
+			if (!('getEntriesByType' in window.performance)) return;
+
 			// Log performance info
 			const { duration: download = -1 } =
 				window.performance


bryophyta (approved):
great catch!

bryophyta:
I guess there are a few non-equivalent ways of checking whether an object has a property/method in js, right?  I'm never clear on exactly which to use when. `in` looks to me like it should work here, but I'd be interested to know if you've got any guidelines on when to use what test?

mxdvl:
> I guess there are a few non-equivalent ways of checking whether an object has a property/method in js, right? I'm never clear on exactly which to use when. `in` looks to me like it should work here, but I'd be interested to know if you've got any guidelines on when to use what test?

Thanks for the question. The reason I am using the `in` keyword here is because we have a linting rule against checking optional chaining on defined members of an object. As this method is defined in [`lib.dom.d.ts`](https://github.com/microsoft/TypeScript/blob/c4eb37c8a0d5915d64e3992a91a34724aab19fe6/lib/lib.dom.d.ts#L11076), we need to circumvent this rule.

```ts
window.performance.getEntriesByType?.('measure');
// Via https://typescript-eslint.io/rules/no-unnecessary-condition/ ESLint auto-fixes to:
window.performance.getEntriesByType('measure');
```

An alternative would be to disable that rule locally, which might be more explicit? What do you think, @bryophyta:

```ts
// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- not all browsers
window.performance.getEntriesByType?.('measure');
```

PR #5854, opened 2022-08-25
Fix the hover colour for standfirst links

oliverlloyd:
## What does this change?
We now always use the text colour to style the underline when hovering links in the standfirst

## Why?
To simplify this code and fix an issue where we were not highlighting the hover properly

| Before      | After      |
|-------------|------------|
|  |  |


marjisound (approved):
LGTM

HarryFischer:
👍

PR #5855, opened 2022-08-25
Don't always stack the byline on Analysis pieces

oliverlloyd:
## What does this change?
This PR fixes a bug where the byline looked wrong when an Analysis piece had multiple contributors stacked on top of each other

| Before      | After      |
|-------------|------------|
| Screenshot 2022-08-25 at 11 59 57 | Screenshot 2022-08-25 at 11 56 14 |


HarryFischer:
👍

alinaboghiu (approved):

paperboyo:
+1 if only because it renders the comma from the content now. I could imagine all sorts of disasters when hiding commas. Like, total societal breakdown. For, example.

Issue #5856, opened 2022-08-25
Network front Thrasher sizes


mxdvl:
# Largest thrashers on network fronts

## [UK Front](https://www.theguardian.com/uk) 

| Name | Total size | Embed size | Resources |
| ---- | ---------- | ---------- | --------- |
| Documentaries | 901 kB | 25.4 kB | `jpg`: 196 kB, `mp4`: 288 kB, `woff2`: 391 kB, `js`: 1.54 kB |
| First Edition newsletter | 381 kB | 30.8 kB | `png`: 40 kB, `woff2`: 300 kB, `js`: 9.65 kB |
| Contact the Guardian | 325 kB | 22.7 kB | `woff2`: 300 kB, `js`: 1.92 kB |
| Palette styles new do not delete | 112 kB | 111 kB | `js`: 962 B |
| culture-treat | 103 kB | 9.79 kB | `png`: 89.8 kB, `gif`: 1.1 kB, `js`: 2.82 kB |

JamieB-gu:
@ashishpuliyel can we discuss ways to manage this? For example, one suggestion was to automate notifications to people who build these when they cross a certain threshold. Perhaps we could discuss in a WebX workshop session?

PR #5857, opened 2022-08-25
Thrasher tracker

mxdvl:
## What does this change?

Adds a script to measure the size of thrashers over time. From a mobbing session on 10% day last week.

## Why?

We want to know if they grow, reduce, or which ones to tackle first. Updates #5856 

## Example

> 
> ## [UK Front](https://www.theguardian.com/uk) 
> 
> | Name | Total size | Embed size | Resources |
> | ---- | ---------- | ---------- | --------- |
> | Documentaries | 901 kB | 25.4 kB | `woff2`: 391 kB, `mp4`: 288 kB, `jpg`: 196 kB, `js`: 1.54 kB |
> | First Edition newsletter | 381 kB | 30.8 kB | `woff2`: 300 kB, `png`: 40 kB, `js`: 9.65 kB |
> | Contact the Guardian | 325 kB | 22.7 kB | `woff2`: 300 kB, `js`: 1.92 kB |
> | Palette styles new do not delete | 112 kB | 111 kB | `js`: 962 B |
> | culture-treat | 103 kB | 9.79 kB | `png`: 89.8 kB, `js`: 2.82 kB, `gif`: 1.1 kB |

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -1,4 +1,4 @@
-name: DCR Fronts Ophan Components
+name: Scheduled daily jobs
 on:
   schedule:
     # Every work day of the week at 08:08
@@ -8,7 +8,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  ophan-components:
+  scheduled:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout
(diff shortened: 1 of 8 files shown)


paperboyo:
Interesting. Looks like reusing platform fonts would give us biggest gain?

mxdvl:
> Interesting. Looks like reusing platform fonts would give us biggest gain?

It's not that simple, as 1 kB of JavaScript has to be parsed, evaluated and can load additional resources, whereas 100 kB of images or fonts can be cached, quickly processed and not load additional resources.

But in general, yes, sharing fonts would be a big win for some thrashers!

OllysCoding (approved):
I mobbed on this but it looks really good!

OllysCoding:
> It's not that simple, as 1 kB of JavaScript has to be parsed, evaluated and can load additional resources ...

Also interesting, is that 'Embed size' is the size of the text getting added to the html page (prior to gzipping ofc!) - this then increases the size of the initial page request, slowing down nearly _all_ other parts of page loading. Getting rid of 'Palette styles do not delete' as part of the DCR migration will be a huge win!

mxdvl:
> I mobbed on this but it looks really good!

Not only did you mob, you were the instigator 💎

PR #5858, opened 2022-08-25
Add isMobile Sign In Gate display rule

vlbee:
## What does this change?

This PR adds a new display rule to check device type, which determines whether or not to show the sign in gate.

'isMobile' - checks if a browsing device is a mobile device or not, based on the [latest Mozilla guidance](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent) to use maxTouchpoints

## Why?

This is required to implement an upcoming Sign In Gate AB test

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

### DIFF
--- a/dotcom-rendering/src/web/components/SignInGate/displayRule.test.ts
+++ b/dotcom-rendering/src/web/components/SignInGate/displayRule.test.ts
@@ -171,17 +171,19 @@ describe('SignInGate - displayRule methods', () => {
 			'get',
 		);
 
-		// FIXME sometimes this navigator property does not exist in testing env
-		!window.navigator.maxTouchPoints &&
+		// Sometimes this navigator property does not exist in testing env
+		if (!window.navigator.maxTouchPoints) {
 			Object.defineProperty(window.navigator, 'maxTouchPoints', {
 				configurable: true,
 				get: () => '',
 			});
+		}
 		const maxTouchPointGetter = jest.spyOn(
 			window.navigator,
 			'maxTouchPoints',
 			'get',
 		);
+
 		test('is mobile device', () => {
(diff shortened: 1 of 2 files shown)


oliverlloyd:
Hi @vlbee

Thanks for breaking this code out 🙏 

Would you be able to provide more context and information for the requirements around this change? This is a new precedent for DCR, we've never used javascript to try and detect the device a reader is using, we've only ever used breakpoints

oliverlloyd:
> more context and information for the requirements around this change?

To be clear, I ask because I'm not sure it's certain if theguardian.com should offer a varied experience based on device. We are moving much of the site away from using client side javascript to decide what readers see and experience to a world where content is server rendered. But on the server we are not aware what device a reader is using.

We are doing this work because we want to improve the performance of the page. Better performance leads to better Core Web Vital scores, which leads to an improved reader experience, better Google ranking, improve ad revenue and, ultimately, more money.

Nothing is decided in this area but adding this function does feel as though it could be somewhat of a watershed, which, if we can, it might be good to avoid.

mxdvl (commented):
Thanks for isolating this requirement.

If this is indeed a requirement for an upcoming test, I would be very explicit about it, and extract this method into its own file, with a comment stating that it shall be removed afterwards. As @oliverlloyd mentioned, we would not want this pattern to make its way into the libs for anyone to use.

You can also mark the method as deprecated with JSDoc to prevent others from using it.

```ts
/**
 * DO NOT USE - special method for Sep 2022 test
 * @deprecated
 */
const isMobile = () => false
```

PR #5859, opened 2022-08-25
use Option<Newsletter> instead of optional Newsletter

dblatcher:
## What does this change?
use `Option<Newsletter>` instead of `Newsletter | undefined` at the type for `fields.promotedNewsletter`

## Why?
Consistency with the rest of the AR structure. 
https://github.com/guardian/dotcom-rendering/pull/5808#discussion_r954108482


JamieB-gu (approved):
🎉

PR #5865, opened 2022-08-26
refactor(types): extract Switches to config

mxdvl:
## What does this change?

Extract the `Switches` type to `types/config.ts` and import it explicitely.
Add the possibility that values are `undefined`, which is more explicit.

## Why?

Follow-up on #5804, which was pre-emptively coercing these switches to `boolean`. The TS compiler would now raise these as errors.


OllysCoding (approved):
+1 This is a major improvement !

PR #5866, opened 2022-08-26
Iain/remove rich link image

iainjchambers-guardian:
## What does this change?
Fixes the original PR on this subject so that it only affects blogs

## Why?
Misunderstanding of the requirements on my part


georgeblahblah (approved):
Looks good! I wondered if we wanted to conver the `RichLink` component to JSX as we're touching it here, but not necessary!

PR #5871, opened 2022-08-26
Varying webpack bundles for tests - Sentry tag

arelra:
## What does this change?

Add a Sentry tag based on the bundle variant

## Why?

So we can trace errors back to the bundle type

### DIFF
--- a/dotcom-rendering/src/web/browser/sentryLoader/sentry.ts
+++ b/dotcom-rendering/src/web/browser/sentryLoader/sentry.ts
@@ -51,6 +51,11 @@ Sentry.init({
 	},
 });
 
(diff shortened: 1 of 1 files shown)


mxdvl:
Is this enough? Do we want to make it explicitly `control` / `variant`?

Should we only set this tag for the variant?

arelra:
> Is this enough? Do we want to make it explicitly `control` / `variant`?

The value of the test is `'variant' | 'control'` already - seemed unnecessary to redefine.

Happy to make it explicit if you prefer though.

> Should we only set this tag for the variant?

I thought having a tag for control and variant would make it easier to compare in Sentry using tags?

Issue #5353, opened 2022-07-11
Numbered list ignores weighting


mxdvl:

bryophyta:
Update: another email came through related to this issue yesterday. It sounds like there's a relatively easy fix in the editorial tools, so perhaps if we aren't able to fix this in the near future we could try to publicise the tools-side fix?

paperboyo:
> we could try to publicise the tools-side fix

Not really a fix as much as users having to remember to change tools switches that are ignored by DCR. They will never remember to do it, even more so because, apart from an image going blurry, the preview looks exactly the same.

What happens is, Design decided that Numbered list template is rendered with always effectively a Showcase Main media weighting no matter what is the state of weighting switch in Composer. So a platform is making a decision of ignoring a switch. Always dodgy idea (think what happens in ten years with those*), but if upheld – the platform in question should make sure it’s requesting a correctly dimensioned image (effectively a Showcase-sized one).

*platforms ignoring switches inevitably leads to inconsistencies in legacy content which will become visible once the design changes. Some of the Numbered list articles will have this “fixed” (Showcase), some not (“Inline”). One day, when platform(s) will go back to honouring content switches, they will look… random.

/end rant, super-minor, but wanted to make the point

HarryFischer:
Is a solution to this to stop forcing showcase main media? Allow inline and showcase?

oliverlloyd:
> Is a solution to this to stop forcing showcase main media? Allow inline and showcase?

For me, this makes sense. In composer, `numberedList` is a display hint which mainly affects how ul/li tags are rendered. But it also has the effect of forcing the main media to showcase, but this feels wrong. Ideally, we'd respect the weighting choice made in Composer and allow the main media to be inline as well as showcase.

If we did this, then there's a case to be made that `Display.NumberedList` should be migrated to `Design.Numbered` list. This would be a lot more semantic - it never felt right being able to combine the `NumberedList` display type with other designs. Whereas having a showcase numbered list or an immersive numbered list both feel like good ways to extend the model.

paperboyo:
> For me, this makes sense

There was this thinking of moving Main media weighting to live where Display hints live. This would prevent it being possible. Just saying. cc @HarryFischer

paperboyo:
[...]

PR #5875, opened 2022-08-29
Prioritise `Display` vs. `role` when chosing image sizes

### LINKED ISSUE #5878: Poor image quality on immersive images
If a main media image is given the role of `showcase` (you can choose between default and showcase) and then the article is given the display hint of immersive then this results in the image being shown as immersive but given the sources based on the showcase role. This means the quality is reduced

oliverlloyd:
## What does this change?
This PR fixes #5878 by ensuring we prefer the use of `format.display` when deciding which image sizes rather than `image.role`

## Why?
Because there was an issue before where an immersive article could still have its main media given a role of `showcase`.

In Composer, the main media can have two possible weightings, `default` and `showcase`. If you choose `showcase` then this has the effect of making the main media larger and repositioning some elements on the page. DCR defines this as `Display.Showcase`. However, if you continue to edit the article in Composer to add the `displayHint` of `immersive` then this has the effect of 'overriding' the showcase role and the article becomes `Display.Immersive`. This strange behaviour by the tool was not anticipated by DCR and so we were still choosing the image sizes assuming showcase but the main media was being rendered full width immersive.

Article.Display is always correct so we now favour this over `role`

### DIFF
--- a/dotcom-rendering/src/web/components/Picture.tsx
+++ b/dotcom-rendering/src/web/components/Picture.tsx
@@ -45,8 +45,22 @@ const decideImageWidths = ({
 	format: ArticleFormat;
 }): ImageWidthType[] => {
 	if (isMainMedia) {
-		switch (role) {
-			case 'showcase':
+		switch (format.display) {
+			case ArticleDisplay.Immersive: {
+				// If display is Immersive then main media should *always*
+				// use these larger image sources
+				return [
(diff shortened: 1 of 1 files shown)


mxdvl (approved):
Looks good. Does this also fix #5353 ?

oliverlloyd:
> Does this also fix #5353 ?

No, but I think we could use a very similar approach as we're using here to solve that problem. I'll add that ticket to my list

PR #5876, opened 2022-08-29
Use `makeWindowGuardian` everywhere

oliverlloyd:
## What?
This PR merges `makeFrontWindowGuardian` with `makeWindowGuardian`

## Why?
We want to be able to serve more than two different types of page with DCR but without loosing too much common functionality in terms of how we construct the page. We previously merged `articleTemplate` and `frontTemplate` for the same reasons.

### DIFF
--- a/dotcom-rendering/src/model/window-guardian.ts
+++ b/dotcom-rendering/src/model/window-guardian.ts
@@ -1,4 +1,4 @@
-import type { ServerSideTests, Switches } from '../types/config';
+import type { ConfigType, ServerSideTests, Switches } from '../types/config';
 import type { EditionId } from '../types/edition';
 import type { DCRFrontType } from '../types/front';
 import type { CAPIArticleType } from '../types/frontend';
@@ -72,9 +72,20 @@ interface WindowGuardianFrontConfig {
 export const makeWindowGuardian = ({
 	CAPIArticle,
 	GAData,
+	unknownConfig = {},
 }: {
 	CAPIArticle: CAPIArticleType;
 	GAData: GADataType;
+	/**
+	 * In the case of articles we don't know the exact values that need to exist
+	 * on the window.guardian.config.page property so rather than filter them we
+	 * allow the entire object to be passed through here and we then extend it
+	 * using Object.assigns
+	 *
+	 * This is obviously rubbish but has to be weighed against the risk of the
+	 * commercial code failing because it depended on a property we removed
+	 */
+	unknownConfig?: ConfigType | object;
 }): {
 	// The 'config' attribute is derived from CAPIArticle and contains
 	// all the data that, for legacy reasons, for instance compatibility
@@ -98,7 +109,7 @@ export const makeWindowGuardian = ({
 			isDev: process.env.NODE_ENV !== 'production',
 			stage: config.stage,
 			frontendAssetsFullURL: config.frontendAssetsFullURL,
-			page: Object.assign(config, {
+			page: Object.assign(unknownConfig, {
 				dcrCouldRender: true,
 				contentType: CAPIArticle.contentType,
 				edition: CAPIArticle.editionId,
--- a/dotcom-rendering/src/model/window-guardian.ts
+++ b/dotcom-rendering/src/model/window-guardian.ts
@@ -98,15 +98,15 @@ export const makeWindowGuardian = ({
 	dfpAccountId: string;
 	adUnit: string;
 	ajaxUrl: string;
-	shouldHideReaderRevenue: boolean;
-	isPaidContent: boolean;
 	googletagUrl: string;
 	switches: Switches;
 	abTests: ServerSideTests;
 	editionId: EditionId;
+	shouldHideReaderRevenue?: boolean;
+	isPaidContent?: boolean;
 	contentType?: string;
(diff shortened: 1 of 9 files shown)


dotcom-rendering/src/model/window-guardian.ts
```diff
@@ -119,80 +102,43 @@ export const makeWindowGuardian = (
...
+				brazeApiKey,
+			}),
+			libs: {
+				googletag: googletagUrl,
+			},
+			switches: switches,
+			tests: abTests,
+			ophan: {
+				pageViewId: '',
+				browserId: '',
```
mxdvl, line 139: I know this was already set up that way, but what is the benefit of this? Should we not set it to `null` instead? Is this because we have typed these as `string` but they can only be determined client-side?

Not quite one for this PR to fix, but worth thinking about.

dotcom-rendering/src/model/window-guardian.ts
```diff
@@ -119,80 +102,43 @@ export const makeWindowGuardian = (
...
-		},
-		modules: {
-			sentry: {
-				reportError: () => null,
+		config: {
+			// This indicates to the client side code that we are running a dotcom-rendering rendered page.
+			isDotcomRendering: true,
+			isDev: process.env.NODE_ENV !== 'production',
+			stage: stage,
+			frontendAssetsFullURL: frontendAssetsFullURL,
```
mxdvl, line 113: How do we feel about shorthand properties? I find them easier to parse, especially with longer properties such as this one, and it helps keep the same variable names accross.

We are using it extensively in the `page` object below, with `sentryPublicApiKey`, `dfpAccountId`, `ajaxUrl`, etc.

```suggestion
			frontendAssetsFullURL,
```

mxdvl: There’s an [ESLint rule for this](https://eslint.org/docs/latest/rules/object-shorthand.html), so will raise a PR and we can take that discussion there.

dotcom-rendering/src/web/server/articleToHtml.tsx
```diff
@@ -172,7 +173,38 @@ export const articleToHtml = ({ article: CAPIArticle }: Props): string => {
...
+					sectionName: CAPIArticle.sectionName,
+					contentType: CAPIArticle.contentType,
+					tags: CAPIArticle.tags,
+					pageId: CAPIArticle.pageId,
+					editionId: CAPIArticle.editionId,
+					beaconURL: CAPIArticle.beaconURL,
+				}),
+				unknownConfig: CAPIArticle.config,
+			}),
+		),
```
mxdvl, line 207: Nice and explicit, I love it!

dotcom-rendering/src/model/window-guardian.ts
```diff
@@ -36,77 +36,60 @@ export interface WindowGuardianConfig {
...
+	/**
+	 * In the case of articles we don't know the exact values that need to exist
+	 * on the window.guardian.config.page property so rather than filter them we
+	 * allow the entire object to be passed through here and we then extend it
+	 * using Object.assigns
+	 *
+	 * This is obviously rubbish but has to be weighed against the risk of the
+	 * commercial code failing because it depended on a property we removed
+	 */
+	unknownConfig?: ConfigType | object;
```
mxdvl, line 91: With the current [commercial bundle TypeScript conversion sitting above of 80%](https://guardian.github.io/commercial-bundle-progress/), this risk is much reduced than when this decision was originally made.

We should probably revisit this and see if we can move forward with an implementation. One approach might be for the @guardian/commercial-dev team to define the relevant configuration in [@guardian/commercial-core](https://github.com/guardian/commercial-core) so types can be shared accross frontend and DCR.

mxdvl (approved):
’twas the right abstraction!

oliverlloyd:
This relates to https://github.com/guardian/dotcom-rendering/issues/5881

dotcom-rendering/src/model/window-guardian.ts
oliverlloyd: Defo. I missed this and would have appreciated a nudge from the tooling!

dotcom-rendering/src/model/window-guardian.ts
oliverlloyd: Are we perhaps at the point where we could take the leap and agree a model for `config`? This would be a _huge_ win and I think it's worth some time investigating

dotcom-rendering/src/model/window-guardian.ts
oliverlloyd: I think you're right. I'd be much happier if these were `null` or `undefined` as that better represents their true state while they wait to get set by the script in the head of the page.

My guess is the reason we've not done this is we probably have contracts with external codebases reading this property which perhaps expect a string

oliverlloyd: > Is this because we have typed these as `string` but they can only be determined client-side?

Yes, there's a piece of javascript that gets executed super early in the head of the page which sets `pageViewId` and `browserId` is set by Fastly and there's an init script which runs and updates this window property with it.

Arguably, we don't really need to use the window object to set `browserId`, you could just read the cookie directly 🤔

PR #5877, opened 2022-08-29
Avatar position refactor

### LINKED ISSUE #5678: Avatar position in Card

oliverlloyd:
## What does this change?
This closes #5678 by refactoring how we position avatars

The essence of this PR is that it now solves the problem of where to position the avatar at different breakpoints using a different approach. It removes the previous pattern of having two avatars in the card, showing and hiding them based on breakpoint, replacing it with an approach that treats the avatar similar to the main card image. In this PR, both the avatar and then main card image are repositioned using `flex-direction`, with `imageType` being passed down to adjust things as needed.

## Why?
Because we need to support more complex use cases now that DCR is rendering fronts

Screenshot 2022-08-29 at 16 04 54


dotcom-rendering/src/web/components/ArticleMeta.tsx
```diff
@@ -244,7 +244,7 @@ const shouldShowContributor = (format: ArticleFormat) => {
 	}
 };
 
-const AvatarContainer = ({ children }: { children: React.ReactNode }) => (
+const MetaAvatarContainer = ({ children }: { children: React.ReactNode }) => (
```
oliverlloyd, line 247: I renamed this to avoid confusion with the other component also called `AvatarContainer`

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -307,80 +306,79 @@ export const Card = ({
 					imageUrl !== undefined ? imagePositionOnMobile : 'top'
 				}
 				minWidthInPixels={minWidthInPixels}
+				imageType={imageType}
 			>
-				{imageType === 'mainmedia' && (
```
oliverlloyd, line 311: A key part of this PR is that avatars and mainmedia images now share the same location (and associated wrappers)

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -307,80 +306,79 @@ export const Card = ({
...
-					</ImageWrapper>
-				)}
+					)}
+				</ImageWrapper>
 				<ContentWrapper
+					imageType={imageType}
 					imageSize={imageSize}
 					imagePosition={imagePosition}
 				>
-					<Flex>
```
oliverlloyd, line 330: One less div 🎉

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -307,80 +306,79 @@ export const Card = ({
...
-							<Hide when="above" breakpoint="tablet">
-								<AvatarContainer>
-									<Avatar
-										imageSrc={avatarUrl}
-										imageAlt={byline ?? ''}
-										containerPalette={containerPalette}
-										format={format}
-									/>
-								</AvatarContainer>
-							</Hide>
```
oliverlloyd, line 381: The previous logic solved the problem for where to position the avatar on mobile by using two avatars, show/hiding them at different breakpoints. We no longer do this, instead we share the `flex-direction` logic used by the main image

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -395,18 +393,6 @@ export const Card = ({
...
-						)}
```
oliverlloyd, line 409: As above, we no longer use this show/hide approach

dotcom-rendering/index.d.ts
```diff
@@ -315,6 +315,8 @@ type ImagePositionType = 'left' | 'top' | 'right' | 'bottom' | 'none';
 
 type ImageSizeType = 'small' | 'medium' | 'large' | 'jumbo';
 
+type CardImageType = 'mainmedia' | 'avatar';
```
OllysCoding, line 318: `mainMedia`?

dotcom-rendering/src/web/components/Card/Card.tsx
OllysCoding: This is a really nice solution!

dotcom-rendering/src/web/components/Card/components/ImageWrapper.tsx
```diff
@@ -58,9 +60,15 @@ export const ImageWrapper = ({
 		<div
 			css={[
 				isHorizontal &&
+					imageType === 'mainmedia' &&
 					flexBasisStyles({
 						imageSize,
 					}),
```
OllysCoding, line 66: Very much styleistic but could the order be 
```js
imageType === 'mainmedia' &&
isHorizontal &&
flexBasisStyles({
	imageSize,
}),
```

dotcom-rendering/index.d.ts
oliverlloyd: Makes sense. Done.

dotcom-rendering/src/web/components/Card/components/ImageWrapper.tsx
oliverlloyd: Yeah, this looks a bit nicer. Done 👍

OllysCoding (approved):
+1 I really like this change :)

PR #5880, opened 2022-08-30
ESLint: object shorthand

mxdvl:
## What does this change?

Enable the [`object-shorthand`](https://eslint.org/docs/latest/rules/object-shorthand.html) ESLint rule

## Why?

Make this [_de facto_](https://en.wikipedia.org/wiki/De_facto) rule explicit.


oliverlloyd (approved):
👍 : 👍

mxdvl:
> 👍 : 👍

I see what you did there!

PR #5882, opened 2022-08-30
fix: allow thrasher-tracker URLs

mxdvl:
## What does this change?

Fix scripts/deno/thrasher-tracker.ts URLs

## Why?

The action is failing on `main` :( It’s an unfortunate quirk of Github that if you create a new action, you cannot run it until it’s merged in, so we could not test this as part of #5857 .

- Failure: https://github.com/guardian/dotcom-rendering/actions/runs/2954611157
- Success: https://github.com/guardian/dotcom-rendering/actions/runs/2955299417

### DIFF
--- a/scripts/deno/thrasher-tracker.ts
+++ b/scripts/deno/thrasher-tracker.ts
@@ -1,9 +1,11 @@
 import { array, object, string } from 'https://deno.land/x/zod@v3.17.3/mod.ts';
 import { fetchJSON } from './json.ts';
-import prettyBytes from 'https://esm.sh/pretty-bytes';
+import prettyBytes from 'https://esm.sh/pretty-bytes@6.0.0';
 import { octokit } from './github.ts';
 
 // -- Constants -- //
+
+const gu = 'https://www.theguardian.com/';
 
 const fronts = ['uk', 'us', 'international', 'au'] as const;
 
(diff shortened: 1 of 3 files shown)


scripts/deno/thrasher-tracker.ts
```diff
@@ -1,10 +1,12 @@
 import { array, object, string } from 'https://deno.land/x/zod@v3.17.3/mod.ts';
 import { fetchJSON } from './json.ts';
-import prettyBytes from 'https://esm.sh/pretty-bytes';
+import prettyBytes from 'https://esm.sh/pretty-bytes@6.0.0';
 import { octokit } from './github.ts';
 
 // -- Constants -- //
 
+const gu = new URL('https://www.theguardian.com/');
```
oliverlloyd, line 8: Does this need to be a URL object? I didn't know you could pass in `new URL()` to itself but it seems you can. But do we need to if we're hard coding the string like this?

oliverlloyd (approved):
Looks good. Just a question about URL usage but nothing blocking

scripts/deno/thrasher-tracker.ts
mxdvl: Your scientists were so preoccupied with whether or not they could, they didn’t stop to think if they should.

PR #5889, opened 2022-08-30
Add support for `fixed/video` container

### LINKED ISSUE #5149: "fixed/video" (`24`) container

Please update [FaciaPicker in Frontend](https://github.com/guardian/frontend/blob/main/facia/app/services/dotcomrendering/FaciaPicker.scala) once this container is supported by DCR

After proposing a basic solution (see [this comment](https://github.com/guardian/dotcom-rendering/issues/5149#issuecomment-1231682026)) it's now been agreed that this container should be fully migrated with the same features as Frontend (larger cards, play video in place, scoll when clicking offscreen cards).

This work may involve a refactor of Card and Carousel or it might make more sense to create new ones, maybe `MediaCard` and `MediaCarousel`?

oliverlloyd:
## What does this change?
This adds support for the `fixed/video` container using the `Carousel` component.

This closes #5149 

This is a stopgap. The existing video container is very different to `Carousel`, it allows videos to be played insitue and clicking partially off-screen cards triggers a scroll, none of the 'cards' are links. In this PR, we simply add classic Cards for each video article.

## Why?
Because taking this approach took me exactly 19 minutes to implement. If I were to create a new carousel to match the one that currently shows on front pages I would have to invest considerably more time. We may well still implement this different behaviour but having a stopgap is perhaps useful in terms of making a DCR AB test happen.

Screenshot 2022-08-30 at 14 51 40

### DIFF
--- a/dotcom-rendering/src/web/components/Carousel.importable.tsx
+++ b/dotcom-rendering/src/web/components/Carousel.importable.tsx
@@ -30,7 +30,7 @@ type Props = {
 	description?: string;
 	url?: string;
 	onwardsSource: OnwardsSource;
-	format: ArticleFormat;
+	format?: ArticleFormat;
 };
 
 // Carousel icons - need replicating from source for centring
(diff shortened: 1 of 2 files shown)


OllysCoding:
Just to confirm, is this the container we're implementing?
image

oliverlloyd:
> is this the container we're implementing

Yes

mxdvl (commented):
I’m not sure I’m a big fan of making format optional in this case, as it introduces another level of complexity.

If this is explicitely a stopgap, I would be tempted to duplicate the Carousel component and create a version that does not take a format, so we’re not polluting our abstractions.

oliverlloyd:
I'm closing this after talking to @HarryFischer

PR #5891, opened 2022-08-30
Adds test run for new AB test long-running trial implementation using localStorage participations

vlbee:
## What does this change?

- Adds new AB testing utility called `setOrUseParticipations` that can be used as a test definition's `canRun` function. This function will set a localStorage ab testing key, or use the presence of that key to determine if a test can run, based on a switch flag.
   - Caveats: to simplify implementation and avoid scope creep, this function assumes only one variant per test. 
   
- Add AB tests definitions for a test run of the Sign In Gate long running mandatory test that will use this localstorage key. 
   - Test run includes 4 regions, 1% each. 
   - Dependant PR - https://github.com/guardian/frontend/pull/25437
   - Notes - during the bucketing stage, a single browser can be bucketed into two test participations if they move regions during that time (eg. travel from the US to UK). The test will only run for their current region. This can be avoided by staggering test offsets but not really neccessary for a test run. 

## Why?

- We need to run a test on browser behaviour over time, but need a way to limit new browsers being added to the test bucket during that time. This PR adds an implementation of for a two-stage  (bucketing and testing) AB test. 
- We need to have a test run of the two-stage bucketing implemented here. 

### Local

**`setParticipationsFlag = true`**
| mvt in range | geo location matches test | participation is set correctly | ophan fires with correct test registration | tested? |
| ------------ | ------------------------- | -------------------- | ------------------------------------------ | ---- |
| y            | y                         | y                    | y                                          | ✔ |
| y            | n                         | n                    | n                                          |  ✔|
| n            | y                         | n                    | n                                          |  ✔ |
| n            | n                         | n                    | n                                          | ✔ |

**`setParticipationsFlag = false`**
| mvt in range | geo location matches test | correct participation present | ophan fires with correct test registration |        |
| ------------ | ------------------------- | ----------------------------- | ------------------------------------------ | ------ |
| y            | y                         | y                             | y                                          | ✔️ |
| y            | y                         | y - multiple                  | y                                          | ✔️     |
| y            | n                         | y - multiple                  | n                                          | ✔️     |
| n            | y                         | y - multiple                  | n                                          | ✔️     |
| n            | n                         | y - multiple                  | n                                          | ✔️     |
| y            | y                         | n                             | n                                          | ✔️     |
| y            | n                         | n                             | n                                          | ✔️     |
| n            | y                         | n                             | n                                          | ✔️     |
| n            | n                         | n                             | n                                          | ✔️     |

### DIFF
--- a/dotcom-rendering/cypress/e2e/parallel-1/sign-in-gate.cy.js
+++ b/dotcom-rendering/cypress/e2e/parallel-1/sign-in-gate.cy.js
@@ -59,7 +59,7 @@ describe('Sign In Gate Tests', function () {
 		beforeEach(function () {
 			disableCMP();
 			// sign in gate main runs from 0-900000 MVT IDs, so 500 forces user into test
-			setMvtCookie('500');
+			setMvtCookie('500000');
 
 			// set article count to be min number to view gate
 			setArticleCount(3);
--- a/dotcom-rendering/src/web/experiments/lib/ab-exclusions.ts
+++ b/dotcom-rendering/src/web/experiments/lib/ab-exclusions.ts
@@ -12,7 +12,8 @@ import {
  * After the cut off date users without the local storage participation key won't see the gate at all.
  *
  * Caveats - There will be natural churn of users to account for. Users can manually delete localStorage at any time.
- * Some browsers also automatically delete local storage specifically on Webkit browsers (Safari/iOS) since iOS and iPadOS 13.4, Safari 13.1 on macOS
+ * Some browsers also automatically delete local storage after some time.
+ * Notably: ITP in Safari since iOS and iPadOS 13.4, Safari 13.1 on macOS @see https://webkit.org/tracking-prevention/#intelligent-tracking-prevention-itp
  *
  */
 export const setParticipations = (
--- a/dotcom-rendering/src/web/experiments/lib/ab-exclusions.ts
+++ b/dotcom-rendering/src/web/experiments/lib/ab-exclusions.ts
@@ -32,9 +32,7 @@ export const setOrUseParticipations = (
 
 	if (setParticipationsFlag) {
 		// check if participation already exists
-		if (participations[abTestId]?.variant === variantId) {
-			return true;
-		}
+		if (participations[abTestId]?.variant === variantId) return true;
 
 		// else set add participation to local storage
 		setParticipations({
--- a/dotcom-rendering/src/web/experiments/lib/ab-exclusions.ts
+++ b/dotcom-rendering/src/web/experiments/lib/ab-exclusions.ts
@@ -1,0 +1,51 @@
+import { Participations } from '@guardian/ab-core';
+import {
+	getParticipationsFromLocalStorage,
+	setParticipationsInLocalStorage,
+} from './ab-localstorage';
+
+/**
+ * Occasionally we may want to track long-running behavior of browsers in a certain test,
+ * without adding new browsers to the test bucket. We cannot rely on the GU_mvt_id bucketing alone to do this.
+ * You can use this method to show the gate based on local storage participation key ("gu.ab.participations") instead.
+ *
+ * After the cut off date users without the local storage participation key won't see the gate at all.
+ *
+ * Caveats - There will be natural churn of users to account for. Users can manually delete localStorage at any time.
+ * Some browsers also automatically delete local storage specifically on Webkit browsers (Safari/iOS) since iOS and iPadOS 13.4, Safari 13.1 on macOS
+ *
+ */
+export const setParticipations = (
+	currentTestParticipation: Participations,
+): void => {
+	// sets participation under local storage key: 'gu.ab.participations'
+	setParticipationsInLocalStorage(currentTestParticipation);
+};
+
+// Put this LAST in any row of canRun boolean functions so we don't set the flag unless all other criteria are true
+export const setOrUseParticipations = (
+	setParticipationsFlag: boolean, // If setParticipationsFlag is true, will set flag and run test, else will run based on flag
+	abTestId: string,
+	variantId: string, // IMPORTANT: Can only be used for single variant AB Tests!
+): boolean => {
+	const participations = getParticipationsFromLocalStorage();
+
+	if (setParticipationsFlag) {
+		// check if participation already exists
+		if (participations[abTestId]?.variant === variantId) {
+			return true;
+		}
+
+		// else set add participation to local storage
+		setParticipations({
+			...participations,
+			[abTestId]: {
+				variant: variantId,
+			},
+		});
+		return true; // test will be run
+	} else {
+		// if the test participation exists in localstorage, run the test
+		return participations[abTestId]?.variant === variantId;
+	}
+};
--- a/dotcom-rendering/src/web/experiments/lib/ab-localstorage.ts
+++ b/dotcom-rendering/src/web/experiments/lib/ab-localstorage.ts
@@ -1,0 +1,30 @@
+import type { Participations } from '@guardian/ab-core';
+import { storage } from '@guardian/libs';
+
+/**
+ * These utils have been lifted from the equivalent Frontend file
+ * https://github.com/guardian/frontend/blob/18f5d8aa4a4191f4efb58b37dc39685860369e04/static/src/javascripts/projects/common/modules/experiments/ab-local-storage.ts
+ */
+
+const participationsKey = 'gu.ab.participations';
+
+// -------
+// Reading
+// -------
+export const getParticipationsFromLocalStorage = (): Participations =>
+	(storage.local.get(participationsKey) as Participations | undefined) ?? {};
+
+// -------
+// Writing
+// -------
+
+// Wipes all localStorage participations
+export const clearParticipations = (): void => {
+	storage.local.remove(participationsKey);
+};
+
+export const setParticipationsInLocalStorage = (
+	participations: Participations,
+): void => {
+	storage.local.set(participationsKey, participations);
+};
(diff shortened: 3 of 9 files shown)


dotcom-rendering/src/web/experiments/lib/ab-exclusions.ts
```diff
@@ -0,0 +1,51 @@
...
+): void => {
+	// sets participation under local storage key: 'gu.ab.participations'
+	setParticipationsInLocalStorage(currentTestParticipation);
+};
+
+// Put this LAST in any row of canRun boolean functions so we don't set the flag unless all other criteria are true
+export const setOrUseParticipations = (
+	setParticipationsFlag: boolean, // If setParticipationsFlag is true, will set flag and run test, else will run based on flag
+	abTestId: string,
+	variantId: string, // IMPORTANT: Can only be used for single variant AB Tests!
```
vlbee, line 29: I've hardcoded this to assume one variant. I am not sure how difficult it would be or if it's worth trying to replicate the variant bucketing here that is implemented in the testing library.

dotcom-rendering/src/web/experiments/lib/ab-localstorage.ts
```diff
@@ -0,0 +1,30 @@
...
+ * https://github.com/guardian/frontend/blob/18f5d8aa4a4191f4efb58b37dc39685860369e04/static/src/javascripts/projects/common/modules/experiments/ab-local-storage.ts
+ */
+
+const participationsKey = 'gu.ab.participations';
+
+// -------
+// Reading
+// -------
+export const getParticipationsFromLocalStorage = (): Participations =>
+	(storage.local.get(participationsKey) as Participations | undefined) ?? {};
```
mxdvl, line 15: Could we do a bit of parsing or validation there? What happens if the object is malformed instead of simply missing?

```ts
const isParticipations = (participations: unknown): participations is Participations => {
  if(!isObject(participations)) return false;
  if(Object.values(maybeParticipations).some(participation => !isString(participation.variant)) return false;
  return true;
}
```

vlbee: I've added in 👍

dotcom-rendering/src/web/experiments/lib/ab-exclusions.ts
```diff
@@ -0,0 +1,49 @@
...
+): boolean => {
+	const participations = getParticipationsFromLocalStorage();
+
+	if (setParticipationsFlag) {
+		// check if participation already exists
+		if (participations[abTestId]?.variant === variantId) return true;
+
+		// else set add participation to local storage
+		setParticipations({
```
vlbee, line 38: Ophan records any initialised test when a page loads. This can be used to track number of unique browserids in the test bucket in the data lake so an additional ophan call is not needed here. 

https://github.com/guardian/dotcom-rendering/blob/025b8e2be443a7813cd9e58dc67e65aa56c550cd/dotcom-rendering/src/web/browser/ophan/ophan.ts#L91

dotcom-rendering/src/web/experiments/tests/sign-in-gate-mandatory-long-testrun.ts
```diff
@@ -14,25 +14,26 @@ export const signInGateMandatoryLongTestRunUk: ABTest = {
...
 		{
 			id: 'mandatory-long-testrun-uk',
 			test: (): void => {},
+			canRun: () =>
+				['GB'].includes(getCountryCodeSync() ?? '') &&
+				setOrUseParticipations(
+					setParticipationsFlag,
+					'SignInGateMandatoryLongTestRunUk', // test id
+					'mandatory-long-testrun-uk', // variant id - can only be used for single variant test
+				),
```
vlbee, line 36: Had to move this to the Variant's optional `canRun` function because of the [order in which the ab testing framework calcutes what it can run](https://github.com/guardian/ab-testing/blob/9cfc419c49f92d08bc41c9185a1642602d0e3164/packages/ab-core/src/core.ts#L110-L115) - which is:

1/  Forced test exceptions
2/ ABTest definition `canRun` function
3/ Test Variant based on `mvt_id` cookie
4/ Variant definition `canRun` function if present

If we used the test's canRun function we risk setting the participation for a test that falls outside of the present `mvt_id` cookie bucket range but is in the correct reason.

This is only the case on non overlapping tests.

dotcom-rendering/cypress/e2e/parallel-1/sign-in-gate.cy.js
```diff
@@ -59,7 +59,7 @@ describe('Sign In Gate Tests', function () {
 		beforeEach(function () {
 			disableCMP();
 			// sign in gate main runs from 0-900000 MVT IDs, so 500 forces user into test
-			setMvtCookie('500');
+			setMvtCookie('500000');
```
mxdvl, line 62: For readability, this is identical in modern JS:

```suggestion
			setMvtCookie('500_000');
```

dotcom-rendering/src/web/experiments/lib/ab-exclusions.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Occasionally we may want to track long-running behavior of browsers in a certain test,
+ * without adding new browsers to the test bucket. We cannot rely on the GU_mvt_id bucketing alone to do this.
+ * You can use this method to show the gate based on local storage participation key ("gu.ab.participations") instead.
+ *
+ * After the cut off date users without the local storage participation key won't see the gate at all.
+ *
+ * Caveats - There will be natural churn of users to account for. Users can manually delete localStorage at any time.
+ * Some browsers also automatically delete local storage specifically on Webkit browsers (Safari/iOS) since iOS and iPadOS 13.4, Safari 13.1 on macOS
```
mxdvl, line 15:
```suggestion
 * Some browsers also automatically delete local storage after some time.
 * Notably: ITP in Safari since iOS and iPadOS 13.4, Safari 13.1 on macOS @see https://webkit.org/tracking-prevention/#intelligent-tracking-prevention-itp
```

dotcom-rendering/src/web/experiments/tests/sign-in-gate-mandatory-long-testrun.ts
```diff
@@ -0,0 +1,189 @@
...
+
+export const signInGateMandatoryLongTestRunUk: ABTest = {
+	id: 'SignInGateMandatoryLongTestRunUk',
+	start: '2022-09-20',
+	expiry: '2022-10-01',
+	author: 'vlbee',
+	description:
+		'Test run for long mandatory test - Show sign in gate to global users on 3rd article view of simple article templates, with higher priority over banners and epic.',
+	audience: 0.0025,
+	audienceOffset: 0.89, // 890001 - 892500
```
mxdvl, line 17: Not a blocker, but do you want have a test that ensures none of these tests ever overlap?

dotcom-rendering/src/web/experiments/tests/sign-in-gate-mandatory-long-testrun.ts
mxdvl: I think the canRun method of a test is optional, so you could remove it, if unused.

dotcom-rendering/src/web/experiments/tests/sign-in-gate-mandatory-long-testrun.ts
```diff
@@ -0,0 +1,189 @@
...
+// or use localstorage participation key to decide canRun result (false)
+const setParticipationsFlag = true;
+
+export const signInGateMandatoryLongTestRunUk: ABTest = {
+	id: 'SignInGateMandatoryLongTestRunUk',
+	start: '2022-09-20',
+	expiry: '2022-10-01',
+	author: 'vlbee',
+	description:
+		'Test run for long mandatory test - Show sign in gate to global users on 3rd article view of simple article templates, with higher priority over banners and epic.',
```
mxdvl, line 15: For data that is identical over many tests, you could create a base that you spread in actual tests:

```
const sharedTestData = {
	start: '2022-09-20',
	expiry: '2022-10-01',
	author: 'vlbee',
	description:
		'Test run for long mandatory test - Show sign in gate to global users on 3rd article view of simple article templates, with higher priority over banners and epic.',
} as const;

// later…

export const signInGateMandatoryLongTestRunUk: ABTest = {
	...sharedTestData,
}
```

dotcom-rendering/src/web/experiments/tests/sign-in-gate-mandatory-long-testrun.ts
```diff
@@ -0,0 +1,189 @@
...
+				setOrUseParticipations(
+					setParticipationsFlag,
+					'SignInGateMandatoryLongTestRunAunz', // test id
+					'mandatory-long-testrun-aunz', // variant id - can only be used for single variant test
+				),
+		},
+	],
+};
+
+const EuropeList = [
```
mxdvl, line 105:
```suggestion
const EuropeList: CountryCode[] = [
```

mxdvl (approved):
This is very thorough and diligent work! 🚀 

Few non-blocking comments below.

dotcom-rendering/src/web/experiments/tests/sign-in-gate-mandatory-long-testrun.ts
vlbee: Its only optional on the Variant type, required on the ABTest type so this needs to default to true

dotcom-rendering/cypress/e2e/parallel-1/sign-in-gate.cy.js
vlbee: 🤯

dotcom-rendering/src/web/experiments/tests/sign-in-gate-mandatory-long-testrun.ts
vlbee: Yes, as browsers can move between regions during the bucketing and therefore same browser might be in different tests. the data team thought best for the test run to keep them non overlapping

PR #5899, opened 2022-08-31
feat(bundle): use @guardian/browserslist-config

mxdvl:
## What does this change?

Run an experiment for the variant bundle, using [the guardian’s browserlist config](https://github.com/guardian/csnx/tree/main/libs/%40guardian/browserslist-config) as a target.

Note that we are [currently rolling out out the `DCRJSBundleVariant` to 1%](https://github.com/guardian/frontend/blob/6709656637a612120b06892f264273ca9ea9473a/common/app/experiments/Experiments.scala#L19-L26).

## Why?

We want to check the impact on performance of using this target in DCR.

We will compare:
- Core Web Vitals for readers in the control and variant. We expect the variant to perform as good or better than the control. 
- Attention time is not significantly worse.
- Any difference in errors reported by Sentry for the `dcr.bundle` tag.

### DIFF
--- a/dotcom-rendering/package.json
+++ b/dotcom-rendering/package.json
@@ -63,6 +63,7 @@
     "@guardian/ab-react": "^2.0.1",
     "@guardian/atoms-rendering": "^23.6.0",
     "@guardian/braze-components": "^7.4.0",
+    "@guardian/browserslist-config": "^2.0.3",
     "@guardian/commercial-core": "^4.3.0",
     "@guardian/consent-management-platform": "10.11.1",
     "@guardian/discussion-rendering": "^10.2.1",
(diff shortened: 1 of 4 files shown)


bryophyta (approved):
This looks good to me!

mxdvl:
Realised that this will not work because the tests are not named properly. The following two PRs should fix it

- …
- https://github.com/guardian/frontend/pull/25578

arelra:
Thanks for pushing this through @mxdvl looking forward to seeing the results!

mxdvl:
We can roll this out to everyone, as we got this from Guy, our data analyst:

> Just checked and no notable or significant difference between the control and variant for CWV or attention time

Issue #5902, opened 2022-08-31
I can identify which trail represents `mostCommented` and `mostShared` in the `news/most-popular` container


oliverlloyd:

bryophyta:
The description's pretty minimal, so does anyone know anything about this issue? Do we think this has been closed by our MostViewed PRs? @jamesgorrie @OllysCoding

VDuczekW:
no comment and not enough detail therefore closing

PR #5903, opened 2022-08-31
Support `news/most-popular` container

### LINKED ISSUE #5142: "news/most-popular" (`82`) container

oliverlloyd:
## What does this change?
This PR closes #5142 by adding support for the `news/most-popular` container on fronts. To achieve this I've refactored the most viewed components such that there is a separate `MostViewedFooter` component that takes the right level of structured data that works for both fronts and articles

## Why?
We were previously using the same client side approach as articles to insert this data on fronts but given the `news/most-popular` container has the trail data that we need, doing this work on the server makes more sense

## What about?
There are three issues that this PR does not address and that have been broken out into their own pieces of work

### Where are the most commented and most shared items?
They're not included in this PR because I was not able to find them in the data model sent by Frontend. But the `mostShared` & `mostCommented` properties do exist on the component so can be passed through once available. I've created a [ticket to add this information to the model](https://github.com/guardian/dotcom-rendering/issues/5902).

### Why does it not use `containerTitle` and instead just says 'Most viewed'?
Because this is what Frontend does. There's [a ticket to question this](https://github.com/guardian/dotcom-rendering/issues/5910) however. I considered putting that work here, in this PR, but after I saw it also touched on articles I decided to break it out.

### Why don't we show tab titles?
Right now we are hardcoding one tab with 10 trails because FE is not sending any other extra data to DCR. There's [a ticket](https://github.com/guardian/frontend/issues/25448) to investigate this.

| Before      | After      |
|-------------|------------|
| Screenshot 2022-09-01 at 08 41 17 | Screenshot 2022-09-01 at 08 40 26 |

### DIFF
--- a/dotcom-rendering/src/web/layouts/LiveLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/LiveLayout.tsx
@@ -1274,6 +1274,7 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 							data-print-layout="hide"
 							data-link-name="most-popular"
 							data-component="most-popular"
+							leftColSize="wide"
 						>
 							<MostViewedFooterLayout>
 								<Island clientOnly={true} deferUntil="visible">


dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -174,6 +185,60 @@ export const FrontLayout = ({ front, NAV }: Props) => {
...
+											<AdSlot
+												position="mostpop"
+												display={format.display}
+											/>
+										</div>
+									)}
+								</div>
+							</Section>
+						);
+					}
```
OllysCoding, line 240: I'm not a big fan of the pattern we have building here of lots of 'large' (no. of lines of code wise) of 'exceptions' building up here, in my opinion it makes the code hard to read, as you can easily have your entire IDE filled with JSX inside the map, and get lost as to where you are. 

In my opinion maps should be relatively concise so you can tell exactly what's going on in them at a glance.

Would this component fit into <DecideContainer>? If not perhaps we could look into some other kind of abstraction in a future PR to make this map more readable?

dotcom-rendering/src/web/components/AdSlot.tsx
```diff
@@ -96,6 +96,19 @@ const merchandisingAdStyles = css`
 	min-height: 250px;
 `;
 
+const mostpopAdStyles = css`
```
OllysCoding, line 99: `mostPopAdStyles`?

dotcom-rendering/src/web/components/MostViewedFooterData.importable.tsx
```diff
@@ -1,44 +1,22 @@
...
 import { useAB } from '../lib/useAB';
 import { useApi } from '../lib/useApi';
-import { MostViewedFooterGrid } from './MostViewedFooterGrid';
-import { MostViewedFooterSecondTierItem } from './MostViewedFooterSecondTierItem';
+import { MostViewedFooter } from './MostViewedFooter';
 
 interface Props {
 	sectionName?: string;
-	palette: Palette;
+	format: ArticleFormat;
```
OllysCoding, line 16: +1

dotcom-rendering/src/web/components/AdSlot.tsx
oliverlloyd: I started off with `mostPopAdStyles` but went with `mostpopAdStyles` in the end to align with the value of `position` (`mostpop`).

But I'm not strong here at all and I agree that consistent camel case is better so I'll make the change. Arguably we should also change `position` to match...

oliverlloyd: Done. I briefly looked at updating `position` but the string `mostpop` is also used as part of a classname and we really don't want to be messing with css classname conventions that bleed into FE and commercial code 😬

dotcom-rendering/src/web/layouts/FrontLayout.tsx
oliverlloyd: Thanks for raising this. I had a think about this and then saw that there's a way to simplify everything. In [this commit](https://github.com/guardian/dotcom-rendering/pull/5903/commits/85182142bd7b59755381fa139f4b4f6664b71b5b) I've moved this layout logic into `MostViewedLayout` and made that component accept children so that it will work for both articles (where we fetch the data using an api call) and now also fronts (where we render the data directly).

This also has the effect of making the stories for this component simpler

oliverlloyd: > In my opinion maps should be relatively concise so you can tell exactly what's going on in them at a glance

💯 agree on this. Thanks for calling it out!

arelra (approved):
ad style updates lgtm 👍

OllysCoding (approved):
+1 I think this looks good :) Well thought through change & nice to see the palette => format refactor!

bryophyta (approved):
I haven't been following the context around this very closely, but no particular issues are jumping out to me!

PR #5904, opened 2022-08-31
refactor(gap): use css flex gap property

mxdvl:
## What does this change?

Use flex’s [gap](https://developer.mozilla.org/en-US/docs/Web/CSS/gap) property instead of padding top and bottoms with margins.

Some bottom margins are kept for the `<UL>` component.

## Why?

It’s more semantic, it removes code and it looks nearly identical. There’s a 2px difference for vertical spacing, as we are now using Source’s `space[3]`. This achieved better design consistency, at 12px, up from 10px.

We can get rid of much of the complex decision-making around `padBottom` or `padBottomOnMobile` in `<LI>`s.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/188116319-46b3399e-215a-4629-8baa-c35c0135fb2d.png
[after]: https://user-images.githubusercontent.com/76776/188657913-2c2f494e-13d9-4efc-b2f5-ece194a95726.png

## Known issues

Some older browsers may not support this property. Tested with Chrome 83, this does not look _too_ bad, as we still have horizontal which visually indicate a break between cards:

image

### DIFF
--- a/dotcom-rendering/src/web/components/Card/components/LI.tsx
+++ b/dotcom-rendering/src/web/components/Card/components/LI.tsx
@@ -1,50 +1,38 @@
 import { css } from '@emotion/react';
-import { from, until } from '@guardian/source-foundations';
+import { from, space, until } from '@guardian/source-foundations';
 import { verticalDivider } from '../../../lib/verticalDivider';
 import { verticalDividerWithBottomOffset } from '../../../lib/verticalDividerWithBottomOffset';
+
+/**
+ * This value needs to match the one set
+ * in the `verticalDividerWithBottomOffset`
+ */
+const GAP_SIZE = `${space[3]}px`;
 
 const liStyles = css`
 	/* This position relative is needed to contain the vertical divider */
 	position: relative;
 
 	display: flex;
+	row-gap: ${GAP_SIZE};
 `;
 
-const sidePaddingStyles = (
-	padSidesOnMobile: boolean,
-	paddingSize: string,
-) => css`
+const sidePaddingStyles = (padSidesOnMobile: boolean) => css`
 	/* Set spacing on the li element */
 	${padSidesOnMobile && until.tablet} {
-		padding-left: ${paddingSize};
-		padding-right: ${paddingSize};
+		padding-left: 10px;
+		padding-right: 10px;
 	}
 
 	${from.tablet} {
-		padding-left: ${paddingSize};
-		padding-right: ${paddingSize};
+		padding-left: 10px;
+		padding-right: 10px;
 	}
 `;
 
 const snapAlignStartStyles = css`
 	/* Snap start of card */
 	scroll-snap-align: start;
-`;
-
-const paddingBottomStyles = (paddingSize: string) => css`
-	padding-bottom: ${paddingSize};
-`;
-
-const mobilePaddingBottomStyles = (paddingSize: string) => css`
-	${until.tablet} {
-		padding-bottom: ${paddingSize};
-	}
-`;
-
-const marginTopStyles = css`
-	${until.tablet} {
-		margin-top: 12px;
-	}
 `;
 
 const decideSize = (percentage?: CardPercentageType, stretch?: boolean) => {
@@ -69,28 +57,30 @@ const decideSize = (percentage?: CardPercentageType, stretch?: boolean) => {
 	return sizeStyle;
 };
 
-function decideDivider(
+const decideDivider = (
 	offsetBottomPaddingOnDivider: boolean,
 	paddingSize: string,
-) {
-	return offsetBottomPaddingOnDivider
+) =>
+	offsetBottomPaddingOnDivider
 		? verticalDividerWithBottomOffset(paddingSize)
 		: verticalDivider;
-}
 
 type Props = {
 	children: React.ReactNode;
-	percentage?: CardPercentageType; // Used to give a particular LI more or less weight / space
-	stretch?: boolean; // When true, the card stretches based on content
-	showDivider?: boolean; // If this LI wraps a card in a row this should be true
-	padSides?: boolean; // If this LI directly wraps a card this should be true
-	padSidesOnMobile?: boolean; // Should be true if spacing between cards is desired on mobile devices
-	padBottom?: boolean;
-	padBottomOnMobile?: boolean; // Should be true if spacing below is desired on mobile devices
-	showTopMarginWhenStacked?: boolean;
-	snapAlignStart?: boolean; // True when snapping card when scrolling e.g. in carousel
-	// Prevent the divider from spanning the LI's bottom padding. To be used when you know that the
-	// LI will have bottom padding, but won't have another card in the same container directly below it.
+	/** Used to give a particular LI more or less weight / space */
+	percentage?: CardPercentageType;
+	/** When true, the card stretches based on content */
+	stretch?: boolean;
+	/** If this LI wraps a card in a row this should be true */
+	showDivider?: boolean;
+	/** If this LI directly wraps a card this should be true */
+	padSides?: boolean;
+	/** Should be true if spacing between cards is desired on mobile devices */
+	padSidesOnMobile?: boolean;
+	/** True when snapping card when scrolling e.g. in carousel */
+	snapAlignStart?: boolean;
+	/** Prevent the divider from spanning the LI's bottom padding. To be used when you know that the
+	LI will have bottom padding, but won't have another card in the same container directly below it. */
 	offsetBottomPaddingOnDivider?: boolean;
 };
 
@@ -101,18 +91,11 @@ export const LI = ({
 	showDivider,
 	padSides = false,
 	padSidesOnMobile = false,
-	padBottom,
-	padBottomOnMobile,
-	showTopMarginWhenStacked,
 	snapAlignStart = false,
 	offsetBottomPaddingOnDivider = false,
 }: Props) => {
 	// Decide sizing
 	const sizeStyles = decideSize(percentage, stretch);
-	// paddingSize is set here because the offset value used for the
-	// verticalDividerWithBottomOffset needs to match the value used for
-	// paddingBottom
-	const paddingSize = '10px';
 
 	return (
 		<li
@@ -120,11 +103,8 @@ export const LI = ({
 				liStyles,
 				sizeStyles,
 				showDivider &&
-					decideDivider(offsetBottomPaddingOnDivider, paddingSize),
-				padSides && sidePaddingStyles(padSidesOnMobile, paddingSize),
-				padBottom && paddingBottomStyles(paddingSize),
-				padBottomOnMobile && mobilePaddingBottomStyles(paddingSize),
-				showTopMarginWhenStacked && marginTopStyles,
+					decideDivider(offsetBottomPaddingOnDivider, GAP_SIZE),
+				padSides && sidePaddingStyles(padSidesOnMobile),
 				snapAlignStart && snapAlignStartStyles,
 			]}
 		>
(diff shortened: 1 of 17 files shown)


OllysCoding:
:O I didn't know this was a thing but it would simplify working with LI/UL's so much! 😍

oliverlloyd:
Using `gap` is an inspired idea. Modern css to the rescue!

There's some visual diffs here highlighting issues with missing vertical spacing below UL slices which might need attention but hopefully that's simple.

Regarding the move from 12px to 10px. There's a long standing convention to use 12px vertical spacing for elements at the Guardian, would it be possible to preserve this? I think if we were going to make any change I suspect there's more appetite to adjust the 10px row spacing to 12 but this would require a lot of refactoring in different areas.

dotcom-rendering/src/web/components/Card/components/LI.tsx
```diff
@@ -69,28 +54,30 @@ const decideSize = (percentage?: CardPercentageType, stretch?: boolean) => {
...
+	showDivider?: boolean;
+	/** If this LI directly wraps a card this should be true */
+	padSides?: boolean;
+	/** Should be true if spacing between cards is desired on mobile devices */
+	padSidesOnMobile?: boolean;
+	/** True when snapping card when scrolling e.g. in carousel */
+	snapAlignStart?: boolean;
+	/** Prevent the divider from spanning the LI's bottom padding. To be used when you know that the
+	LI will have bottom padding, but won't have another card in the same container directly below it. */
 	offsetBottomPaddingOnDivider?: boolean;
```
mxdvl, line 81: Using JSDoc here for Intellisense information when using the `LI` component.

mxdvl, line 81: Using JSDoc here so we get Intellisense when using the `LI` component.

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -179,15 +179,11 @@ const FirstBigBoostedPlusBig = ({
...
 		>
-			<ul css={manualUlStyles}>
-				<LI
-					showDivider={true}
-					padSides={true}
-					padBottom={standardsCol1.length > 0}
-					padBottomOnMobile={standardsCol1.length > 0}
-				>
+			<ul css={[manualUlStyles, verticalDivider]}>
+				<LI showDivider={true} padSides={true}>
```
mxdvl, line 186: We want the divider on the `UL` more than the `LI`.

dotcom-rendering/src/web/components/DynamicPackage.tsx
```diff
@@ -524,7 +456,7 @@ export const DynamicPackage = ({
 	switch (layout) {
 		case 'oneStandard':
 			return (
-				<>
+				<Stack space={3}>
```
mxdvl, line 459: This needs some spacing!

dotcom-rendering/src/web/components/FixedMediumSlowXIIMPU.tsx
```diff
@@ -361,9 +313,12 @@ export const FixedMediumSlowXIIMPU = ({
 		default: {
 			const topThree = trails.slice(0, 3);
 			const remainingCards = trails.slice(3, 9);
-			const lengthIsEven = remainingCards.length % 2 === 0;
+			const evenLength =
+				remainingCards.length % 2 === 0
+					? remainingCards.length
+					: remainingCards.length - 1;
```
mxdvl, line 319: There’s a more concise way to do this, but I thought it might make it harder to follow.

```ts
const evenLength = remainingCards.length - (remainingCards.length % 2)
```

dotcom-rendering/src/web/components/FixedMediumSlowXIIMPU.tsx
```diff
@@ -387,33 +342,12 @@ export const FixedMediumSlowXIIMPU = ({
...
-												  )
-										}
 										offsetBottomPaddingOnDivider={
-											lengthIsEven
-												? false
-												: shouldOffsetWhenOdd(
-														trailIndex,
-														remainingCards.length,
-												  )
+											trailIndex <= evenLength - 2
```
mxdvl, line 346: The logic is much simplified here.

bryophyta:
Nice! I thought that gap didn't have enough support still, but looks like it's past 95% now!

Not relevant to this PR but I wanted to just note that there are a few versions of `gap`-properties, and I think that some of them have less support still than `row-gap` for flex. (The [caniuse results](https://caniuse.com/?search=gap) on this are a bit hard to parse though tbh!)

dotcom-rendering/src/web/lib/verticalDividerWithBottomOffset.ts
```diff
@@ -16,7 +17,7 @@ export function verticalDividerWithBottomOffset(
 				width: 1px;
 				/* 100% is a reasonable fallback for browsers which don't support calc() */
 				height: 100%;
-				height: calc(100% - ${bottomPaddingSize});
+				height: calc(100% + ${bottomPaddingSize});
```
mxdvl, line 20: This has flipped because the offset now has to grow into the `gap` rather than shrink away from the padding.

mxdvl (commented):
Some pointers below 👇

mxdvl:
> Nice! I thought that gap didn't have enough support still, but looks like it's past 95% now!

It is also worth noting that the failure in browsers like Chrome 83 and below is not catastrophic, as we still have horizontal which visually indicate a break between cards:

image

mxdvl:
> Using `gap` is an inspired idea. Modern css to the rescue!
> 
> There's some visual diffs here highlighting issues with missing vertical spacing below UL slices which might need attention but hopefully that's simple.
>

dotcom-rendering/src/web/components/Card/components/LI.tsx
```diff
@@ -1,28 +1,32 @@
 import { css } from '@emotion/react';
-import { from, until } from '@guardian/source-foundations';
+import { from, space, until } from '@guardian/source-foundations';
 import { verticalDivider } from '../../../lib/verticalDivider';
 import { verticalDividerWithBottomOffset } from '../../../lib/verticalDividerWithBottomOffset';
 
+/**
+ * This value needs to match the one set
+ * in the `verticalDividerWithBottomOffset`
+ */
```
ioannakok, line 9: Which value of `verticalDividerWithBottomOffset` this needs to match and how is this accomplished? Sorry, this might be a silly question but I'm not very familiar with this and not sure I'm following here.

ioannakok:
This is a great refactoring! 👏 I didn't know about `gap`, that's very cool! I only have one non-blocking question for my understanding.

dotcom-rendering/src/web/components/Card/components/LI.tsx
mxdvl: It’s the one in [`lib/verticalDividerWithBottomOffset`](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/lib/verticalDividerWithBottomOffset.ts) imported just above.

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -434,7 +412,7 @@ export const DynamicFast = ({
 		: groupedTrails.standard.splice(0, standardsDisplayConfig.columns * 3);
 
 	return (
-		<>
+		<Stack space={3}>
```
oliverlloyd, line 415: Stack is a neat abstraction from Source (which I didn't know about!) but if we use it here are we then adding an extra div? If we have the `ul` tag add the spacing below itself it feels like we will end up with more semantic html structure

oliverlloyd (commented):
This PR is really great, it's cool how it is [already fixing bugs](https://www.chromatic.com/test?appId=5dfcbf3012392c0020e7140b&id=6312356bed7cb3b4f9da3d5f)!

I had a question about `Stack`. Is this a trade off between one extra div giving us simpler rendering logic vs. making the `UL` abstraction more complex but keeping the html a bit more semantic? Happy to be persuaded one way or the other!

dotcom-rendering/src/web/components/DynamicFast.tsx
mxdvl: I think we should strive for the flattest and most semantic DOM. In this case, we have already deeply nested [`ul`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul) which do not really adequately represent a front card.

I’m happy to separate the introduction of `Stack` into its own PR, though.

mxdvl: I’ve removed `<Stack>` from this PR for the time being.

dotcom-rendering/src/web/components/DynamicPackage.tsx
mxdvl: No longer using `<Stack>`

dotcom-rendering/src/web/components/DynamicFast.tsx
```diff
@@ -517,22 +491,24 @@ export const DynamicFast = ({
...
 										<LI
 											key={card.url}
 											percentage={
 												standardsDisplayConfig.cardWidth
 											}
 											stretch={true}
-											showDivider={true}
+											showDivider={
+												cardIndex % columns !== 0
+											}
```
oliverlloyd, line 507: This is a drive by fix to a divider bug I presume? Just checking there's not a connection to the gap I'm missing

dotcom-rendering/src/web/components/FixedLargeSlowXIV.tsx
```diff
@@ -66,20 +65,15 @@ export const FixedLargeSlowXIV = ({
...
 							percentage="25%"
-							showDivider={
-								cardIndex !== 0 &&
-								cardIndex !== 4 &&
-								cardIndex !== 8
-							}
-							padBottom={cardIndex < 4}
-							padBottomOnMobile={
-								cardIndex !== thirdSlice.length - 1
+							showDivider={cardIndex % 4 !== 0}
```
oliverlloyd, line 74: Is this is a drive by refactor unrelated to the gap change in the PR?

dotcom-rendering/src/web/components/FixedMediumSlowXIIMPU.tsx
```diff
@@ -245,34 +227,6 @@ const Card33_Card33_Ad33 = ({
...
- *
- * When there is an odd number of cards the last card spans
- * both columns so we don't want to push the divider for the
- * previous two cards above it down. If we did it would
- * touch
- */
-function shouldOffsetWhenOdd(i: number, noOfCards: number) {
-	return i === noOfCards - 2 || i === noOfCards - 3;
-}
-
```
oliverlloyd, line 275: Bye 👋

dotcom-rendering/src/web/components/FixedMediumSlowXIIMPU.tsx
oliverlloyd: Is this affected to the gap css or just an unrelated diff?

oliverlloyd (commented):
I _think_ these changes are good to go. I _think_ they're brilliant. But to be honest I got a little lost trying to work out what code diffs are caused by the move to `row-gap and what are not which makes adding a thumbs up a lot harder to do.

I think in general the odd drive by fix or improvement is fine but in this case the changes seem to be very close to the code actually being modified by gap so it's hard to tease things apart

mxdvl:
> I think in general the odd drive by fix or improvement is fine but in this case the changes seem to be very close to the code actually being modified by gap so it's hard to tease things apart

I have to agree. I got a little carried away fixing some of the visual issues and simplifying the abstractions, but I'll refactor and keep them in a separate PR for simplicity.

dotcom-rendering/src/web/components/FixedMediumSlowXIIMPU.tsx
mxdvl: Drive-by, moved to #6025

dotcom-rendering/src/web/components/FixedMediumSlowXIIMPU.tsx
mxdvl: Moved to #6025

oliverlloyd (approved):
🎉

PR #5905, opened 2022-08-31
Sensitive shady pie

emma-imber:
## What does this change?
The shady pie component (promotes a digital subscription when an ad blocker is being used) was still showing on some articles marked as sensitive. This update only returns a shady pie component if the content is suitable for showing ads.

## Why?
On certain types of articles we probably shouldn't be showing any kind of promotional material.

## Screenshots

| Before      | After      |
|-------------|------------|
| Screenshot 2022-08-31 at 15 54 41 | Screenshot 2022-08-31 at 15 54 50 |

### DIFF
--- a/dotcom-rendering/src/web/components/AdSlot.tsx
+++ b/dotcom-rendering/src/web/components/AdSlot.tsx
@@ -20,6 +20,7 @@ type InlineProps = {
 	index: number;
 	shouldHideReaderRevenue?: boolean;
 	isPaidContent?: boolean;
+	shouldHideAds?: boolean;
 };
 
 type NonInlineProps = {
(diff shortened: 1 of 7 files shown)


oliverlloyd (commented):
This is awesome, thank you for making this change 🙏 

Do you think it would be possible to hoist the place where we check that boolean up a bit such that the component simply never renders? There's an example of this approach [in this PR](https://github.com/guardian/dotcom-rendering/pull/5643)

emma-imber:
> This is awesome, thank you for making this change 🙏
> 
> Do you think it would be possible to hoist the place where we check that boolean up a bit such that the component simply never renders? There's an example of this approach [in this PR](https://github.com/guardian/dotcom-rendering/pull/5643)

Thanks for this suggestion! I've made the changes, it's definitely a neater way of doing it

oliverlloyd (approved):
Nice. Thanks again for making these changes, it's important that we respect these flags

mxdvl (approved):

PR #5911, opened 2022-09-01
Add errorRetryCount to FetchOnwardsData

bryophyta:
We don't expect live updates for this data, so if the API call has returned an error 3 times, it doesn't seem like there's much reason to keep trying.

Not very urgent, but I thought it merited a pass-by refactor. Is this a pattern we're happy with?

### DIFF
--- a/dotcom-rendering/src/web/components/FetchOnwardsData.importable.tsx
+++ b/dotcom-rendering/src/web/components/FetchOnwardsData.importable.tsx
@@ -32,7 +32,9 @@ export const FetchOnwardsData = ({
 	onwardsSource,
 	format,
 }: Props) => {
-	const { data, loading, error } = useApi<OnwardsResponse>(url);
+	const { data, loading, error } = useApi<OnwardsResponse>(url, {
+		errorRetryCount: 3,
+	});
 
 	const buildTrails = (
 		trails: CAPITrailType[],


oliverlloyd (commented):
I think the existing suite of features that [SWR provides](https://swr.vercel.app/docs/error-handling) around request management is probably okay for our use case. By default we get:

Errors are retried on a five second interval
Retry requests are backed off exponentially
Requests are not made when a tab or page is backgrounded (and automatically restarted when focus returns)
The retry interval is extended if a slow network is detected

dotcom-rendering/src/web/components/FetchOnwardsData.importable.tsx
```diff
@@ -32,7 +32,9 @@ export const FetchOnwardsData = ({
 	onwardsSource,
 	format,
 }: Props) => {
-	const { data, loading, error } = useApi<OnwardsResponse>(url);
+	const { data, loading, error } = useApi<OnwardsResponse>(url, {
+		errorRetryCount: 3,
```
jamesgorrie, line 36: Is there a reason 3? [SWR uses an exponential backoff algorithm](https://swr.vercel.app/docs/error-handling#error-retry) which I would be wary of changing unless we had a specific reason.

Something to bare in mind is that it is not only server errors, but timeouts due to connections, network errors etc which can cause these, which is why the above algorithm is useful.

bryophyta:
sorry for the delay @oliverlloyd & @jamesgorrie -- didn't see these comments until now!

I didn't realise that SWR had those defaults; I agree that they largely negate the need for any extra controls on retries (although there's an argument for not retrying `404`s and maybe some others, too often?)

Having said that, from looking at the requests showing up in my console, it's not totally clear that the default rules are actually being followed?

image

I don't think this is a big deal, and the defaults sound very sensible to me, so I'm happy to close this PR, but just thought I'd flag the thing about the actual behaviour in case it signals an issue. (I had a quick look through our code and SWR's implementation and I can't see any obvious way in which we're inadvertently overriding the defaults, for instance.)

mxdvl:
@bryophyta is you screenshot a real example of something that is tried over and over again on a specific page? This might be worth investigating.

bryophyta:
it is a live example from prod, in this case caused by this issue: https://github.com/guardian/frontend/issues/25453

Issue #5913, opened 2022-09-01
Intermittent Instance Health Check Failures


JamieB-gu:
Occasionally AR instances are failing the ELB health check and getting cycled out of use.

JamieB-gu:
Investigated with @jacobwinch @marjisound @georgeblahblah .

## Why?

Based on metrics from the ELB, it looks like these are failures to respond to health check requests within the acceptable timeout window. The lack of obvious errors in the AR logs seems to back up this assumption: the requests aren't failing and therefore generating error logs, they're just succeeding too late.

Looking at some of the instance metrics, our theory is that the application may be running up against the resource limits (CPU, memory or both) of our current instance size, `t4g.micro`:

https://github.com/guardian/dotcom-rendering/blob/e6ca0805fc0786ea109c5d4844680127848631c0/apps-rendering/cdk/lib/mobile-apps-rendering.ts#L85-L88

Given that we've been slowly increasing the percentage of pageviews that can be served by AR over the last few months, a corresponding increase in traffic to the application would be expected. An increase in traffic (more requests served) would tend to correspond to heavier resource usage.

## What next?

To test our theory, we're going to try increasing the instance size to `t4g.small` for a few days. We can then monitor whether these failures continue to happen, and what some of the metrics like CPU utilisation look like on the new instances.

If this doesn't solve the problem we will reconvene and keep investigating.

## A Comparison With DCR

Interestingly, DCR, which is a very similar application, appears to be using `t4g.small` already:

https://github.com/guardian/dotcom-rendering/blob/e6ca0805fc0786ea109c5d4844680127848631c0/dotcom-rendering/scripts/deploy/riff-raff.yaml#L13-L17

JamieB-gu:
We haven't had any new health check failures since migrating to `t4g.small`. We're going to stick with this new instance size for now.

PR #5915, opened 2022-09-01
Change AR To `t4g.small`

JamieB-gu:
## Why?

We'd like to try running AR on EC2's `t4g.small` instances for a few days, which is a step up from our current `t4g.micro` instances.

A more detailed explanation why can be found in #5913.

Information on the different EC2 instance types can be found here: https://aws.amazon.com/ec2/instance-types/

## Changes

- Change EC2 instance type to `t4g.small`

### DIFF
--- a/apps-rendering/cdk/lib/mobile-apps-rendering.ts
+++ b/apps-rendering/cdk/lib/mobile-apps-rendering.ts
@@ -84,7 +84,7 @@ export class MobileAppsRendering extends GuStack {
 			},
 			instanceType: InstanceType.of(
 				InstanceClass.T4G,
-				InstanceSize.MICRO,
+				InstanceSize.SMALL,
 			),
 			certificateProps: {
 				CODE: {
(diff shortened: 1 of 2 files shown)


apps-rendering/cdk/lib/mobile-apps-rendering.ts
```diff
@@ -84,7 +84,7 @@ export class MobileAppsRendering extends GuStack {
 			},
 			instanceType: InstanceType.of(
 				InstanceClass.T4G,
-				InstanceSize.MICRO,
+				InstanceSize.SMALL,
```
georgeblahblah, line 87: this changes the instance size for both the live AR and the one which reads from preview - would we want to keep the instance size for the preview deployment as micro?

I'd imagine because the preview deployment receives less traffic it is less likely to be experiencing the same problem, though I can see some advantages to keeping them the same.

georgeblahblah (approved):
Looks good! I had a question about whether we should keep the Preview deployment as micro. Given that this experiment is intended for a few days, it's not an urgent question.

apps-rendering/cdk/lib/mobile-apps-rendering.ts
JamieB-gu: Good question. If it's simple to provision preview separately in CDK maybe we should keep it on the smaller size?

What do you think @jacobwinch ?

JamieB-gu: #5966

PR #5919, opened 2022-09-02
`WitnessBlockComponent` may not have a caption

mxdvl:
## What does this change?

Mark the caption field as optional.

## Why?

Prevent some older articles with _guardian**witness**_ elements from failing validation.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/188111380-adac6083-4f76-4e27-97d3-132fcc0e859c.png
[after]: https://user-images.githubusercontent.com/76776/188111275-c3a0d8e1-23dc-47f9-a8f5-eb77c2423e15.png


OllysCoding (approved):
Nice catch :)

PR #5921, opened 2022-09-02
Add explainer to supported article designs

iainjchambers-guardian:
## What does this change?
Adds `Explainer` to the article designs supported by Editions

## Why?
Explainer is a new article type that wasn't previously considered

### Note
This is a very simple _Occam's Razor_ fix, there may well be differences with the Explainer type that I haven't considered and will cause issues so discussion is welcome. Just put this up on the off chance that this is all that needs to be done to save the editions publishing team the work of recreating all these articles to get them to display.

### DIFF
--- a/apps-rendering/src/components/editions/layout/index.tsx
+++ b/apps-rendering/src/components/editions/layout/index.tsx
@@ -170,6 +170,7 @@ const getSectionStyles = (item: ArticleFormat): SerializedStyles[] => {
 
 const Layout: FC<Props> = ({ item }) => {
 	if (
+		item.design === ArticleDesign.Explainer ||
 		item.design === ArticleDesign.Analysis ||
 		item.design === ArticleDesign.Standard ||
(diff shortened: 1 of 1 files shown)


JamieB-gu (approved):
Good catch @iainjchambers-guardian !

I think this will have been a consequence of separating explainers out into a new `Design` [^1][^2]. Previously I think they would most likely have been parsed as `Standard`, and would have been captured by this check. When making this change we realised that we'd have to add a new condition to handle this for the live apps [^3], but I think probably missed that we'd want to make the same change for Editions 😬.

FYI @alinaboghiu @abeddow91 @marjisound 

[^1]: https://github.com/guardian/libs/pull/386
[^2]: https://github.com/guardian/dotcom-rendering/pull/5750
[^3]: https://github.com/guardian/dotcom-rendering/pull/5750#pullrequestreview-1076171989

Issue #5931, opened 2022-09-05
Redirect unsupported AMP pages


mxdvl:
> **Note**
> Follow-up on #5930

If we don’t support an article type (Immersive, Quiz), we should redirect `amp.theguardian.com` URLs to `www.theguardian.com` ones.

bryophyta:
This issue is in the 'In Progress' column in the Rota board, but no-one's assigned to it. Do we know if it's currently in progress? (cc. @mxdvl )

mxdvl:
I think it may have been fixed by #6287 ?

PR #5933, opened 2022-09-05
Stop adding AMP link if it is unsupported

mxdvl:
## What does this change?

Remove the `ampLink` from the page template if an article is Immersive.

## Why?

It’s unsupported in frontend. We are sending the wrong information about what we support.

- https://github.com/guardian/frontend/pull/11397 (the PR is light on details…)
- https://github.com/guardian/frontend/pull/25451
- #5931

### DIFF
--- a/dotcom-rendering/src/web/server/articleToHtml.tsx
+++ b/dotcom-rendering/src/web/server/articleToHtml.tsx
@@ -207,16 +207,23 @@ export const articleToHtml = ({ article: CAPIArticle }: Props): string => {
 		),
 	);
 
-	const hasAmpInteractiveTag = CAPIArticle.tags.some(
-		(tag) => tag.id === 'tracking/platformfunctional/ampinteractive',
-	);
+	const getAmpLink = (tags: TagType[]) => {
+		// We don’t support AMP for Immersive articles in frontend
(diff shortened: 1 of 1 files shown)


dotcom-rendering/src/web/server/articleToHtml.tsx
```diff
@@ -207,16 +207,23 @@ export const articleToHtml = ({ article: CAPIArticle }: Props): string => {
 		),
 	);
 
-	const hasAmpInteractiveTag = CAPIArticle.tags.some(
-		(tag) => tag.id === 'tracking/platformfunctional/ampinteractive',
-	);
+	const getAmpLink = (tags: TagType[]) => {
+		// We don’t support AMP for Immersive articles in frontend
+		// https://github.com/guardian/frontend/blob/149c8d3be273edf784465a780ee332bd58e2a9b3/common/app/model/content.scala#L95
+		if (CAPIArticle.format.display === 'ImmersiveDisplay') return undefined;
```
bryophyta, line 213: I feel as though ideally if the decision is being made in Frontend then we'd probably want to add a property before it gets to DCR saying `doNotRenderAmpLink` or similar.

But given that there was already a check for `InteractiveDesign` here, it would be a bigger refactor to move the logic to Frontend, so this looks good to me.

And I think moving the logic into the `getAmpLink` function really helps readability here 👍

bryophyta (approved):
this looks good to me!

Issue #5932, opened 2022-09-05
Galleries Visual Tweaks


JamieB-gu:
A few design tweaks, some of which bring AR galleries into line with dotcom (frontend) galleries.

1. Main media caption font size. The credit and the caption are currently using different font sizes, they should both be `textSans.xsmall`.

![main-media-caption](https://user-images.githubusercontent.com/53781962/188431628-4705c086-79f6-4c39-afec-a69f528746f9.jpg)

2. Standfirst should be using `headline.xxxsmall` and bold; it's currently using Text Egyptian.

| AR | Dotcom |
| - | - |
| ![standfirst-ar] | ![standfirst-dotcom] |

3. Standfirst links should be text colour, using border underline.

| AR | Dotcom |
| - | - |
| ![standfirst-ar] | ![standfirst-dotcom] |

4. Most copy should be `neutral.86` rather than `neutral.100`; everything except the headline, which remains `neutral.100`.

| AR | Dotcom |
| - | - |
| ![text-ar] | ![text-dotcom] |

5. Caption `h2`s should be `headline.xxxsmall`.

| AR | Dotcom |
| - | - |
| ![caption-h2-ar] | ![caption-h2-dotcom] |

6. The series should be white text on a theme-coloured background. Is it possible it's currently theme-coloured text on a white background?

| AR | Dotcom |
| - | - |
| ![series-ar] | ![series-dotcom] |

[standfirst-dotcom]: https://user-images.githubusercontent.com/53781962/188432046-0f83eeb4-ab2b-441f-944d-768dc4e38978.jpg
[standfirst-ar]: https://user-images.githubusercontent.com/53781962/188432050-c1c6b660-44ea-4f64-93d4-5c856ac85ac5.jpg

[text-ar]: https://user-images.githubusercontent.com/53781962/188433374-31a5294b-6484-4460-a804-f4ae94e58c9c.jpg
[text-dotcom]: https://user-images.githubusercontent.com/53781962/188433381-bcec4a7b-bdd4-4fa5-83f0-7b7b342f3255.jpg

[caption-h2-dotcom]: https://user-images.githubusercontent.com/53781962/188434218-c892cd38-4683-4380-b3f6-4b1cf059be5d.jpg
[caption-h2-ar]: https://user-images.githubusercontent.com/53781962/188434219-4183af03-6a1e-4382-930a-2e36b7f657c3.jpg

[...]

PR #5934, opened 2022-09-05
Add tweaks from review

iainjchambers-guardian:
## What does this change?
Implements tweaks suggested in [#5932](https://github.com/guardian/dotcom-rendering/issues/5932)

### DIFF
--- a/apps-rendering/src/components/Series/GallerySeries.tsx
+++ b/apps-rendering/src/components/Series/GallerySeries.tsx
@@ -28,8 +28,8 @@ const styles = css`
 `;
 
 const linkStyles = (format: ArticleFormat): SerializedStyles => css`
-	background-color: ${text.seriesTitle(format)};
-	color: ${background.series(format)};
+	color: ${text.seriesTitle(format)};
+	background-color: ${background.series(format)};
 	${headline.xxxsmall({ fontWeight: 'bold' })}
 	line-height: 2;
 	text-decoration: none;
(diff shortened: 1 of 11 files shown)


apps-rendering/src/components/Series/GallerySeries.tsx
```diff
@@ -28,8 +28,8 @@ const styles = css`
 `;
 
 const linkStyles = (format: ArticleFormat): SerializedStyles => css`
-	color: ${text.seriesTitle(format)};
-	background-color: ${background.series(format)};
+	background-color: ${text.seriesTitle(format)};
+	color: ${background.series(format)};
```
JamieB-gu, line 32: Should these be the other way round in the palette? That way we don't end up using `background` for `color` and `text` for `background-color`?

common-rendering/src/editorialPalette/text.ts
```diff
@@ -669,6 +668,7 @@ const standfirstLink = (format: ArticleFormat): Colour => {
 					return specialReport[300];
 			}
 		case ArticleDesign.Gallery:
+			return neutral[100];
```
JamieB-gu, line 671:
```suggestion
			return neutral[86];
```

JamieB-gu (commented):
Nice, thanks for getting this updated so quickly!

Couple of final suggestions:

- The standfirst links should also be `neutral.86`, with a border underline of `neutral.46` (closest I could find in the Source palette).
- I think the datetime should also be `neutral.86`.

JamieB-gu (approved):

PR #5936, opened 2022-09-05
Always show tab titles

### LINKED ISSUE #5910: Rationalise container titles for most viewed
Review and rationalise the text that displays for the 'Most viewed' container title. Remove 'popular' and add more context where needed.

**The word 'popular'**
On fronts, we hardcode this container title to 'Most viewed'. On articles we hardcode it to 'Most popular'.

There's [previous work](https://github.com/guardian/frontend/pull/14970) which suggests using the term 'popular' was not desirable as it was in poor taste to suggest some (heavily viewed) articles were 'popular'. We should avoid this terminology.

Image

**Lack of context**
On fronts, we only ever show the text 'most viewed' but the actual container title contains more context. It might say 'Most viewed in Sport' or '...in the UK'. We omit this context, I believe, because we expect tab headings to show which will also display this context. But these headings are often not appearing and in this situation simply saying 'Most viewed' is confusing.

UK front with no context

Image

Sport front _with_ context

Image

The same issue exists for articles. In some cases, there are no tab headings and even when the list of articles is clearly limited to a particular section of the site, the labelling suggest it isn't. E.g.: All the articles listed below are from Culture but the title suggests otherwise

Image

oliverlloyd:
## What does this change?
The PR fixes #5910 by removing the restriction that prevented the tab header from being displayed if there was only one tab. It also changes the css such that the mouse pointer is no longer shown when hovering over the active tab

## Why?
Because if no tab header shows then the reader is not given any context for what the list of trails being shown relate to. We hard code the container title to be `Most viewed` but sometimes this is most viewed globally but it might just be 'in sport' or 'in culture'. By showing the tab header we give that context.

| Before      | After      |
|-------------|------------|
| Screenshot 2022-09-05 at 17 20 19 | Screenshot 2022-09-05 at 17 18 41 |

### DIFF
--- a/dotcom-rendering/src/web/components/MostViewedFooterGrid.tsx
+++ b/dotcom-rendering/src/web/components/MostViewedFooterGrid.tsx
@@ -57,7 +57,7 @@ const unselectedStyles = css`
 	}
 `;
 
-const tabButton = css`
+const buttonStyles = (isSelected: boolean) => css`
 	${headline.xxxsmall()};
 	color: ${neutral[7]};
 	margin: 0;
@@ -73,7 +73,7 @@ const tabButton = css`
 	width: 100%;
 
 	&:hover {
-		cursor: pointer;
+		cursor: ${isSelected ? 'default' : 'pointer'};
 	}
 `;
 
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/MostViewedFooterGrid.tsx
```diff
@@ -130,7 +130,7 @@ export const MostViewedFooterGrid = ({
 	const [selectedTabIndex, setSelectedTabIndex] = useState<number>(0);
 	return (
 		<>
-			{Array.isArray(data) && data.length > 1 && (
+			{Array.isArray(data) && (
```
bryophyta, line 133: would `data` ever not be an array? If not, there seems something slightly odd about checking that it is, but also not (afaict) having any defined behaviour for cases where it's not an array? Not a big deal though I don't think

bryophyta (approved):
nice! just added one small comment but shouldn't be blocking I don't think

PR #5937, opened 2022-09-05
Stop Rendering Dinkus Paragraphs

JamieB-gu:
## Why?

Don't render paragraphs when they contain a dinkus. It results in `<p><hr></p>`, which is invalid and the browser converts to `<p></p><hr><p></p>`. Both variants break the dropcap CSS.

## Changes

- Don't render paragraph tags when they contain a dinkus

### DIFF
--- a/apps-rendering/src/renderer.ts
+++ b/apps-rendering/src/renderer.ts
@@ -242,7 +242,12 @@ const textElement =
 		);
 		switch (node.nodeName) {
 			case 'P': {
+				if (text === '* * *') {
+					return children;
+				}
+
 				const showDropCap = shouldShowDropCap(text, format, isEditions);
+
 				return h(
 					Paragraph,
 					{ key, format, showDropCap, isEditions },


apps-rendering/src/renderer.ts
```diff
@@ -242,7 +242,12 @@ const textElement =
 		);
 		switch (node.nodeName) {
 			case 'P': {
+				if (text === '* * *') {
```
georgeblahblah, line 245: I wonder if we should extract this to a helper? it's also appearing in:

* [`components/HeadingTwo`](https://github.com/guardian/dotcom-rendering/blob/main/apps-rendering/src/components/HeadingTwo/index.tsx#L131)
* [`renderer`](https://github.com/guardian/dotcom-rendering/blob/main/apps-rendering/src/renderer.ts#L130)

georgeblahblah (approved):
Looks good! I had one non-blocking comment about possibly extracting this to a helper

apps-rendering/src/renderer.ts
JamieB-gu: Yeah I think you're right, it's a bit scattered.

In the heading code we're actually already doing this correctly, it's not rendering the `h2` if there's a dinkus.

JamieB-gu: I've merged this fix for now, but we should consider the linked issue as a better long-term solution.

mxdvl: I think it’s also possible to get `***` or `•••`, according to this check in DCR:

https://github.com/guardian/dotcom-rendering/blob/93597a1d00613f641d4415e19946418ae2ed513c/dotcom-rendering/src/model/enhance-dividers.ts#L15-L20

PR #5956, opened 2022-09-06
Set colour to gallery's neutral[86] for follow link

iainjchambers-guardian:
## What does this change?
Changes the colour of the follow link in the gallery metadata component
## Why?
Using theme colours on the dark background does not meet a11y standards
## Screenshots

| Before      | After      |
|-------------|------------|
| image | image |

### DIFF
--- a/common-rendering/src/editorialPalette/text.ts
+++ b/common-rendering/src/editorialPalette/text.ts
@@ -419,6 +419,10 @@ const headlineTag = (_format: ArticleFormat): Colour => neutral[100];
 const headlineTagDark = (_format: ArticleFormat): Colour => neutral[10];
 
 const follow = (format: ArticleFormat): Colour => {
+	if (format.design === ArticleDesign.Gallery) {
+		return neutral[86];
+	}
+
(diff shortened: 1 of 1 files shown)


common-rendering/src/editorialPalette/text.ts
```diff
@@ -419,6 +419,10 @@ const headlineTag = (_format: ArticleFormat): Colour => neutral[100];
 const headlineTagDark = (_format: ArticleFormat): Colour => neutral[10];
 
 const follow = (format: ArticleFormat): Colour => {
+	if (format.design === ArticleDesign.Gallery) {
```
bryophyta, line 422: given that we switch on `format.design` below on line 428, it would be nice if there was a way to bring this check for `ArticleDesign.Gallery` into the main switch-case block. Having said that, I can't think of a clean way to do so! Just thought I'd flag in case you can think of a good way of doing it :)

bryophyta (approved):
Good catch!

JamieB-gu (approved):

Issue #5952, opened 2022-09-06
The table of contents on this article is no longer working


oliverlloyd:
The TOC on this article is expecting h2s to have ids set using their element id but they have text instead.

https://www.theguardian.com/help/privacy-policy

AshCorr:
This seems like its broken again!

PR #5961, opened 2022-09-07
Use `elementId` for h2 ids if there's not an existing TOC on the page

oliverlloyd:
## What does this change?
This PR fixes https://github.com/guardian/dotcom-rendering/issues/5952 by updating the `enhanceH2s` function to factor in if a page already has a legacy table of contents inserted, adjusting the ids that get set on h2s in this case to use the element id instead of a human readable version

## Why?
Because existing TOCs expect to be using elementIds.

### Was there a different approach we could have taken here?
In a kinder world where all is well and the birds sing with joy I would have refactored the `InteractiveContentsBlockComponent` to use these human readable ids. And I did start down this happy joyful path. But then I saw that in some cases these legacy TOC elements also work with different types of element, not just h2s. For example, [this piece](https://www.theguardian.com/technology/2019/dec/17/best-smartphone-2019-iphone-oneplus-samsung-and-huawei-compared-and-ranked) uses ids on `figure` tags, but a figure has no text so can't have a human readable id. At this point I choose to back out and make the change inside the h2 enhancer itself

### DIFF
--- a/dotcom-rendering/src/model/enhance-H2s.ts
+++ b/dotcom-rendering/src/model/enhance-H2s.ts
@@ -1,5 +1,15 @@
 import { JSDOM } from 'jsdom';
 
+const hasInteractiveContentsElement = (blocks: Block[]): boolean => {
+	return !!blocks.find((block) =>
+		block.elements.find(
+			(element) =>
+				element._type ===
+				'model.dotcomrendering.pageElements.InteractiveContentsBlockElement',
+		),
+	);
+};
+
 const extractText = (element: SubheadingBlockElement): string => {
 	const frag = JSDOM.fragment(element.html);
 	if (!frag.firstElementChild) return '';
@@ -78,10 +88,16 @@ const enhance = (elements: CAPIElement[]): CAPIElement[] => {
 	return enhanced;
 };
 
-export const enhanceH2s = (blocks: Block[]): Block[] =>
-	blocks.map((block: Block) => {
+export const enhanceH2s = (blocks: Block[]): Block[] => {
+	if (hasInteractiveContentsElement(blocks)) {
+		// Don't enhance h2s if there's an existing table of contents element on the page
+		return blocks;
+	}
+
+	return blocks.map((block: Block) => {
 		return {
 			...block,
 			elements: enhance(block.elements),
 		};
 	});
+};
--- a/dotcom-rendering/src/model/enhance-H2s.ts
+++ b/dotcom-rendering/src/model/enhance-H2s.ts
@@ -45,11 +45,14 @@ const slugify = (text: string) => {
 };
 
 /**
- * This function creates a sluggified string if we have test available. If not, then it retuns the element id so that we always have an id.
+ * This function attempts to create a slugified string to use as the id. It fails over to elementId.
  */
-const generateId = (text: string, elementId: string) => {
-	if (!text) return elementId;
-	return slugify(text) || elementId;
+const generateId = (element: SubheadingBlockElement, existingIds: string[]) => {
+	const text = extractText(element);
+	if (!text) return element.elementId;
+	const slug = slugify(text);
+	if (!slug) return element.elementId;
+	return getUnique(slug, existingIds);
 };
 
 /**
@@ -65,17 +68,15 @@ const enhance = (elements: CAPIElement[]): CAPIElement[] => {
 			element._type ===
 			'model.dotcomrendering.pageElements.SubheadingBlockElement'
 		) {
-			const text = extractText(element);
 			const id = shouldUseElementId
 				? element.elementId
-				: generateId(text, element.elementId);
-			const uniqueId = getUnique(id, slugifiedIds);
+				: generateId(element, slugifiedIds);
 
 			slugifiedIds.push(id);
 			const withId = element.html.replace(
 				'<h2>',
 				// add ID to H2 element
-				`<h2 id='${uniqueId}'>`,
+				`<h2 id='${id}'>`,
 			);
 			enhanced.push({
 				...element,
--- a/dotcom-rendering/src/model/enhance-H2s.ts
+++ b/dotcom-rendering/src/model/enhance-H2s.ts
@@ -22,7 +22,7 @@ const getUnique = (slug: string, array: string[]): string => {
 	const occurenceCount = array.filter(
 		(currentItem) => currentItem === slug,
 	).length;
-	return occurenceCount > 1 ? `${slug}-${occurenceCount - 1}` : slug;
+	return occurenceCount > 0 ? `${slug}-${occurenceCount}` : slug;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/model/enhance-H2s.ts
```diff
@@ -1,5 +1,13 @@
 import { JSDOM } from 'jsdom';
 
+const hasInteractiveContentsElement = (elements: CAPIElement[]): boolean => {
+	return !!elements.find(
+		(element) =>
+			element._type ===
+			'model.dotcomrendering.pageElements.InteractiveContentsBlockElement',
+	);
```
mxdvl, line 8: Might be more semantic to use `some`?

```suggestion
	return elements.some(
		(element) =>
			element._type ===
			'model.dotcomrendering.pageElements.InteractiveContentsBlockElement',
	);
```

dotcom-rendering/src/model/enhance-H2s.ts
```diff
@@ -78,10 +87,11 @@ const enhance = (elements: CAPIElement[]): CAPIElement[] => {
 	return enhanced;
 };
 
-export const enhanceH2s = (blocks: Block[]): Block[] =>
-	blocks.map((block: Block) => {
+export const enhanceH2s = (blocks: Block[]): Block[] => {
+	return blocks.map((block: Block) => {
 		return {
```
mxdvl, line 92: Why wrapt this in a block?

dotcom-rendering/src/model/enhance-H2s.ts
```diff
@@ -11,13 +19,10 @@ const extractText = (element: SubheadingBlockElement): string => {
...
-		const occurenceCount = array.filter(
-			(currentItem) => currentItem === slug,
-		).length;
-		return `${slug}-${occurenceCount}`;
-	}
-	return slug;
+	const occurenceCount = array.filter(
+		(currentItem) => currentItem === slug,
+	).length;
+	return occurenceCount > 1 ? `${slug}-${occurenceCount - 1}` : slug;
```
mxdvl, line 25: I think if it’s already in the array, you want to pad the slug.

```suggestion
	return occurenceCount > 0 ? `${slug}-${occurenceCount - 1}` : slug;
```

mxdvl (commented):
I think there’s a small bug around `occurenceCount`

abeddow91 (approved):

dotcom-rendering/src/model/enhance-H2s.ts
oliverlloyd: Why not wrap it in a block? 🙃

dotcom-rendering/src/model/enhance-H2s.ts
```diff
@@ -78,10 +87,11 @@ const enhance = (elements: CAPIElement[]): CAPIElement[] => {
 	return enhanced;
 };
 
-export const enhanceH2s = (blocks: Block[]): Block[] =>
-	blocks.map((block: Block) => {
+export const enhanceH2s = (blocks: Block[]): Block[] => {
+	return blocks.map((block: Block) => {
```
abeddow91, line 91: Do we need this return?

abeddow91 (approved):

dotcom-rendering/src/model/enhance-H2s.ts
abeddow91: +1

dotcom-rendering/src/model/enhance-H2s.ts
oliverlloyd: No. But there are _lots_ of examples of this pattern in the codebase and I don't think DCR has (or should have) a rule banning needless `return`s.

The argument for them is that they provide a more explicit statement for what is happening. The argument against them is that the code will be shorter and more concise. The language allows developers to choose which approach they take and I think we should keep this choice open rather than be too prescriptive for things like this.

dotcom-rendering/src/model/enhance-H2s.ts
oliverlloyd: Nice!

dotcom-rendering/src/model/enhance-H2s.ts
oliverlloyd: This code does work but I agree it's confusing 😅 . The confusion is that the array of existingIds is already updated with the new id at this point (it happens in the calling function) so when it's counting occurrences it's also counting itself.

I've made a change to mutate that array _after_ the `getUnique` function is called

dotcom-rendering/src/model/enhance-H2s.ts
mxdvl: Because it wasn’t wrapped in a block before!

Do you think we could take a stab at implementing https://eslint.org/docs/latest/rules/arrow-body-style for consistency?

dotcom-rendering/src/model/enhance-H2s.ts
mxdvl: Arg… I find this really hard to reason about.

Have we actually run some numbers on how often this double-id thing happens? It seems like we might be introducing a lot of complexity here for minimal gains.

dotcom-rendering/src/model/enhance-H2s.ts
mxdvl: I actually think we _should be prescriptive_. Consistency is helpful as it ensures exposure to patterns.

If a developer is unfamiliar with arrow functions, I think they might be wondering what the difference is between the two approaches. I know I did when I first encoutered them. There’s an ESLint rule called [`arrow-body-style`](https://eslint.org/docs/latest/rules/arrow-body-style).

mxdvl (approved):

dotcom-rendering/src/model/enhance-H2s.ts
oliverlloyd: I do think that lint rules and prettier are great things and generally support their use as a way to avoid bike-shedding around formatting but I think we need to be careful as seeing them as a solution for all formatting disagreements. I think that there are sometimes reasons why a developer might choose one approach over the other and we need to be careful to preserve those developer intentions.

In general, the two styles are very similar:

```js
// Implicit return
const add = (num1, num2) => num1 + num2
```

vs

```js
// Explicit return
const add = (num1, num2) => {
    return num1 + num2
}
```

In principle, forcing the code to use one approach over the other makes little difference but I find one of the disadvantages of some lint rules is they will override intentional formatting and can sometimes make, longer, more complex sequences of code, messy. I think we've all experienced that disappointment when you've written a lovely block of beautiful code and then you hit save and Prettier mangles it 😢 . My fear here is that there are places in DCR where this might happen with this particular lint rule

oliverlloyd: Also, yes, happy to have this as a separate discussion to unblock this PR

Issue #4784, opened 2022-04-29
Football lineups on articles could use columns as per Liveblogs at mobile view


rcrphillips:
👋 👋 

Matchstats area on mobile articles(match reports) stacks the lineups instead of putting them into columns side by side as per liveblogs for the same matches.
https://www.theguardian.com/football/live/2022/apr/27/liverpool-v-villarreal-champions-league-semi-final-first-leg-live-score-updates
https://www.theguardian.com/football/2022/apr/27/liverpool-villarreal-champions-league-semi-final-first-leg-match-report
![Screenshot_20220429-162309](https://user-images.githubusercontent.com/2051501/165977488-00f7202c-1293-40a7-a7da-765680b91d3a.png)
![Screenshot_20220429-162249](https://user-images.githubusercontent.com/2051501/165977499-943a0d75-baba-41c2-9192-72eedbfa3285.png)

PR #5965, opened 2022-09-07
fix: Use two columns on mobile, the same as we do for blogs

oliverlloyd:
## What does this change?
This PR fixes https://github.com/guardian/dotcom-rendering/issues/4784 by changing the grid layout we use for non blog articles on mobile from a single column (100%) to a two column view (50% 50%).

## Why?
Because @rcrphillips asked us to 😸  (Sorry this took so long Rob!)

| Before      | After      |
|-------------|------------|
| Screenshot 2022-09-07 at 15 23 26 | Screenshot 2022-09-07 at 15 22 11 |

### DIFF
--- a/dotcom-rendering/src/web/components/MatchStats.tsx
+++ b/dotcom-rendering/src/web/components/MatchStats.tsx
@@ -121,16 +121,15 @@ const StatsGrid = ({
 							}
 
 							${until.phablet} {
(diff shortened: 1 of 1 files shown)


mxdvl (approved):
👏

rcrphillips (approved):
Muchas gracias!

PR #5971, opened 2022-09-08
Scale up servers from 9 > 18

oliverlloyd:
## What?
Scales things up

## Why?
Just to be safe

### DIFF
--- a/dotcom-rendering/cloudformation.yml
+++ b/dotcom-rendering/cloudformation.yml
@@ -79,7 +79,7 @@ Mappings:
       Value: rendering
   StageMap:
     PROD:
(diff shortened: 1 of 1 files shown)


dotcom-rendering/cloudformation.yml
```diff
@@ -79,7 +79,7 @@ Mappings:
       Value: rendering
   StageMap:
     PROD:
-      MinCapacity: 9
+      MinCapacity: 18
       MaxCapacity: 36
```
ioannakok, line 83: We need to change Max also to 72

AshCorr (approved):

PR #5972, opened 2022-09-08
Bump Node to v14.20.0

mxdvl:
## What does this change?

Bump Node to the latest LTS version Fermium: [v14.20.0](https://nodejs.org/es/blog/release/v14.20.0/)

## Why?

Keep things up-to-date.


JamieB-gu:
Will we be bumping the AR and DCR AMIgo recipes too?

AshCorr:
Hm, interestingly Turbosnap doesn't consider a node bump worthy of running all tests again.

AshCorr:
Hm, It seems like Turbosnap doesn't recognize changes in .nvmrc and has skipped all snapshots

mxdvl:
> Will we be bumping the AR and DCR AMIgo recipes too?

I have already updated the DCR one.

mxdvl:
> Following the discussion in last week's Client Side Infra, should we consider a bump to 16 instead?
> – @georgeblahblah 

I have attempted to do this in #5577, but it’s a bigger task due to peer deps and the way npm now words.

georgeblahblah (approved):

PR #5974, opened 2022-09-09
fix: add class .js-native-video to MainMediaVideo wrapper for client script

georgeblahblah:
## What does this change?

- Adds `.js-native-video` class to the `MainMediaVideo`
- Renames `.native-video` on media atom renderer to `.js-native-video`

## Why?

The client side script [`nativeCommunication.ts`](https://github.com/guardian/dotcom-rendering/blob/main/apps-rendering/src/client/nativeCommunication.ts) looks for elements with a specific class name to pass video information via Bridget to the apps. For `MainMediaVideo`, this class was missing and so the elements were not found.

### DIFF
--- a/apps-rendering/src/client/nativeCommunication.ts
+++ b/apps-rendering/src/client/nativeCommunication.ts
@@ -188,7 +188,7 @@ function slideshow(): void {
 }
 
 function getVideoSlots(): VideoSlot[] {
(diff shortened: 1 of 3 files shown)


tkgnm (approved):
Looks great

JamieB-gu (approved):

PR #5986, opened 2022-09-12
fix: Labs header overlapping right advert

chrislomaxjones:
## What does this change?

This is a fix to prevent the Labs header from overlapping the right advert, once a user starts scrolling and the header becomes sticky. This is achieved by applying an offset of `55px` (the height of the Labs header) plus a padding of `6px` (to match the padding of the advert when it's not scrolling) to the `top` of the advert.

## Why?

Prevent the Labs header from overlapping the top of an advert.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/8000415/189696495-887405ac-9c8d-4565-8940-9cddc4ff918f.png
[after]: https://user-images.githubusercontent.com/8000415/189696290-9e9a2135-7c34-4302-937c-37a0c0f49724.png

### DIFF
--- a/dotcom-rendering/src/web/components/TopRightAdSlot.importable.tsx
+++ b/dotcom-rendering/src/web/components/TopRightAdSlot.importable.tsx
@@ -1,12 +1,12 @@
 import type { SerializedStyles } from '@emotion/react';
 import { css } from '@emotion/react';
 import { getCookie } from '@guardian/libs';
+import { LABS_HEADER_HEIGHT } from '../lib/labs-constants';
 import { useAdBlockInUse } from '../lib/useAdBlockInUse';
 import { ShadyPie } from './ShadyPie';
 
 const isServer = typeof window === 'undefined';
 const MOSTVIEWED_STICKY_HEIGHT = 1059;
-const LABS_HEADER_HEIGHT = 55;
 
 /**
  * TopRightAdSlot decides if we should render the ShadyPie or not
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/TopRightAdSlot.importable.tsx
```diff
@@ -6,6 +6,7 @@ import { ShadyPie } from './ShadyPie';
 
 const isServer = typeof window === 'undefined';
 const MOSTVIEWED_STICKY_HEIGHT = 1059;
+const LABS_HEADER_HEIGHT = 55;
```
chrislomaxjones, line 9: I originally pulled this value out of the [labs header file](https://github.com/guardian/dotcom-rendering/blob/17a61d25aa2701e3ce69da4d940d89bbab888ac5/dotcom-rendering/src/web/components/LabsHeader.importable.tsx#L21) itself. I've since gone back on this as it causes this island's bundle to pull in the entire labs header script (bundle analysis below).

Screenshot 2022-09-12 at 16 33 40

 Redefining the value here seems like the simplest way to avoid this.

mxdvl: Not a blocker, but I would rather we had an explicit coupling if there is one.

Could you create a module that contains these constant and import these pixel values from there?

```ts
// commercialConstants.ts
const LABS_HEADER_HEIGHT = 55;

// TopRightAdSlot.importable.tsx
import { LABS_HEADER_HEIGHT } from "./commercialConstants.ts";
css`
  top: ${LABS_HEADER_HEIGHT}px;
`

// LabsHeader.importable.tsx
import { LABS_HEADER_HEIGHT } from "./commercialConstants.ts";
css`
  height: ${LABS_HEADER_HEIGHT}px;
`
```

mxdvl (approved):

dotcom-rendering/src/web/components/TopRightAdSlot.importable.tsx
chrislomaxjones: Good point. I'll add this here whilst this is still open.

SiAdcock (approved):
Nice 👍

PR #5991, opened 2022-09-13
Remove tabindex from skip to maincontent links

DanielCliftonGuardian:
## What does this change?
This removes tabindex from maincontent links.
## Why?
The tabindex was not providing any benefit.

It is not generally used and not mentioned on webaim:
https://webaim.org/techniques/skipnav/

A tab index of -1 was suggested by DAC however this may have been due to tabindexes already being present on some links.

### DIFF
--- a/dotcom-rendering/src/web/components/ArticleBody.tsx
+++ b/dotcom-rendering/src/web/components/ArticleBody.tsx
@@ -187,7 +187,6 @@ export const ArticleBody = ({
 	}
 	return (
 		<div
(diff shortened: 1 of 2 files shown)


mxdvl (approved):
Looks good to me. 

Might be worth adding back the a11y ESLint rule that prevents `tabindex` on non-interactive elements!

mxdvl:
The idea to implement a rule to catch all a11y issues is implemented in #5997

Issue #4434, opened 2022-03-29
Subnav: until desktop ensure all expected links are displayed


DavidLawes:
At mobile & phablet we're missing the search bar:

| Frontend      | DCR      |
|-------------|------------|
| ![frontend][] | ![dcr][] |

[frontend]: https://user-images.githubusercontent.com/45561419/160570963-a406b72d-cd19-46d8-8d0d-0cd4f6e9bf00.png
[dcr]: https://user-images.githubusercontent.com/45561419/160570885-c390d9b0-45bc-48e6-8a9d-d2563afb2904.png

We are also missing some styling on on the 'make a contribution' and 'subscribe' links:

| Frontend      | DCR      |
|-------------|------------|
| ![frontend2][] | ![dcr2][] |

[frontend2]: https://user-images.githubusercontent.com/45561419/160571634-6eb1768d-7251-46e1-836c-e12886d5408a.png
[dcr2]: https://user-images.githubusercontent.com/45561419/160571735-8e7cf5b0-9a12-47b3-b2a8-b06f6ff139e3.png

^^ Note also the missing edition switch

We're also missing facebook and twitter links at the bottom of the menu:

| Frontend      | DCR      |
|-------------|------------|
| ![frontend3][] | ![dcr3][] |

[frontend3]: https://user-images.githubusercontent.com/45561419/160572357-5ad3b690-e137-44ae-94d1-745cffe3f650.png
[dcr3]: https://user-images.githubusercontent.com/45561419/160572419-6dc32619-d9d0-4752-84d4-b92d57bcae2b.png

PR #5996, opened 2022-09-13
Added selectableMenuItem class to allow proper tabbing

DanielCliftonGuardian:
## What does this change?

This adds the search bar to the sub nav menu until desktop in line with Frontend. This is implemented as a button redirecting to google advanced search inline with other DCR search buttons.

This relates to issue https://github.com/guardian/dotcom-rendering/issues/4434

The editions switch will be added in a subsequent PR.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/110032454/185429099-3b35bf24-d526-4bc4-8a86-a791296f1ad0.png
[after]: https://user-images.githubusercontent.com/110032454/190172265-b3c59623-3fc8-49c9-8fd4-c3278c7b40ea.png

### DIFF
--- a/dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
+++ b/dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
@@ -15,7 +15,7 @@ import {
 	SvgMagnifyingGlass,
 } from '@guardian/source-react-components';
 import type { NavType } from '../../../../model/extract-nav';
-import { Column } from './Column';
+import { Column, lineStyle } from './Column';
 import { MoreColumn } from './MoreColumn';
 import { ReaderRevenueLinks } from './ReaderRevenueLinks';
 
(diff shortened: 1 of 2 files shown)


dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -116,6 +124,97 @@ const brandExtensionLink = css`
...
+	max-width: 23.75rem;
+	position: relative;
+	margin-bottom: 24px;
+	margin-right: 41px;
+	padding-bottom: 15px;
+`;
+
+const searchInput = css`
+	${textSans.large()}
+	background-color: rgba(255,255,255, .1);
```
mxdvl, line 154: There’s a `transparentColour` helper you can import, and use like so: `transparentColour(neutral[100], 0.1)`

mxdvl (commented):
I am not sure this should be a text input until this is mirrored in the desktop version.

Currently, on DCR, we simply redirect to the Google Advance Search page.

mxdvl (approved):
Nice one, that looks good!

We discussed this offline, but it would be good to rethink the search experience on DCR, as the current method of redirecting to Google Advanced Search is not very user-friendly. Maybe you could raise an issue so we can discuss as a team and prioritise?

PR #5997, opened 2022-09-13
feat: enforce recommended a11y jsx

mxdvl:
## What does this change?

Enforce recommended Accessibility JSX ESLint rules.

Make an exception for the `aria-role` rule on custom components, because the attribute has a specific editorial meaning: how should an image display, aka _weighting_. We do not want to see errors when using a prop with this name outside of actual html tags.

```ts
// aka weighting. RoleType affects how an image is placed. It is called weighting
// in Composer but role in CAPI. We respect CAPI so we maintain this nomenclature
// in DCR
type RoleType =
	| 'immersive'
	| 'supporting'
	| 'showcase'
	| 'inline'
	| 'thumbnail'
	| 'halfWidth';
```

## Why?

The plugin was introduced in #114, but disabled at some point since.

### DIFF
--- a/dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
+++ b/dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
@@ -329,6 +329,8 @@ export const SecureSignupIframe = ({
 	return (
 		<>
 			<iframe
+				// @TODO check with @dblatcher if there is a better title
+				title={`Sign up to ${newsletterId}`}
 				ref={iframeRef}
 				css={css`
 					width: 100%;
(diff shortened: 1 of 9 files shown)


ioannakok (approved):
We mobbed on this!

dotcom-rendering/src/web/components/SecureSignupIframe.importable.tsx
```diff
@@ -329,6 +329,8 @@ export const SecureSignupIframe = ({
 	return (
 		<>
 			<iframe
+				// @TODO check with @dblatcher if there is a better title
+				title={`Sign up to ${newsletterId}`}
```
mxdvl, line 333: Hey @guardian/newsletters & @dblatcher, is there a better title we could pass to this `iframe`?

PR #6002, opened 2022-09-14
Adds isInRegion ABTest canRun function util for Mandatory Sign in Gate test

vlbee:
## What does this change?

- Adds a small utility called `isInRegion' to the abtest-exclusions file plus tests.
- Adds the list of countries comprising 'Europe' as includes too many countries

## Why?

This function will be used along with `setOrUseParticipations` in the `canRun` function of the AB Test defininitions of the mandatory sign in gate test. If the canRun function returns false, the test is not run and ophan tracking when the test framework initialises sends empty query parameters (eg. `&abTestRegister={}`)

Notes: AB test framework is syncronous but the `getCountryCode` utility is async as it falls back on to an API request, so I have used `getCountryCodeSync` which only checks for the cookie value.  I don't see this having significant impact on numbers in the test bucket but we can check in the test run. 

### Example usage

```
canRun: () =>
		isInRegion(['US', 'CA']) &&
		setOrUseParticipations(
			setParticipationsFlag,
			'SignInGateMandatoryLongTestRun', // test id
			'mandatory-long-testrun', // variant id - can only be used for single variant test
		),
```

ROW example:
```
canRun: () =>
		!isInRegion(['GB', 'US', 'CA', 'AU', 'NZ', ...EuropeList ]) &&
		setOrUseParticipations(
			setParticipationsFlag,
			'SignInGateMandatoryLongTestRun', // test id
			'mandatory-long-testrun', // variant id - can only be used for single variant test
		),
```

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

### DIFF
--- a/dotcom-rendering/src/web/experiments/lib/ab-exclusions.ts
+++ b/dotcom-rendering/src/web/experiments/lib/ab-exclusions.ts
@@ -1,4 +1,5 @@
 import { Participations } from '@guardian/ab-core';
+import { getCountryCodeSync } from '../../lib/getCountryCode';
 import {
 	getParticipationsFromLocalStorage,
 	setParticipationsInLocalStorage,
@@ -8,8 +9,6 @@ import {
  * Occasionally we may want to track long-running behavior of browsers in a certain test,
  * without adding new browsers to the test bucket. We cannot rely on the GU_mvt_id bucketing alone to do this.
  * You can use this method to show the gate based on local storage participation key ("gu.ab.participations") instead.
- *
- * After the cut off date users without the local storage participation key won't see the gate at all.
  *
  * Caveats - There will be natural churn of users to account for. Users can manually delete localStorage at any time.
  * Some browsers also automatically delete local storage specifically on Webkit browsers (Safari/iOS) since iOS and iPadOS 13.4, Safari 13.1 on macOS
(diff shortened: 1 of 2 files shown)


dotcom-rendering/src/web/experiments/lib/ab-exclusions.ts
```diff
@@ -47,3 +46,64 @@ export const setOrUseParticipations = (
...
 	}
 };
+
+/**
+ * Regional Exclusions
+ */
+export const isInRegion = (region: string[]): boolean => {
+	const countryCode = getCountryCodeSync();
+	return !!countryCode && region.includes(countryCode);
+};
```
mxdvl, line 56: Does this need to be synchronous?

What’s the benefit of this abstraction over an `includes`? 

```ts
const regions: CountryCode = ["US", "CA"];
regions.includes(getCountryCodeSync() ?? "");
```

mxdvl (commented):
I am not entirely sure what the benefit of this abstraction is, especially as it forces us to mock values in a test.

```ts
{
	canRun: () =>
		!['GB', 'US', 'CA', 'AU', 'NZ', ...EuropeList ].includes(getCountryCodeSync()) &&
		setOrUseParticipations(
			setParticipationsFlag,
			'SignInGateMandatoryLongTestRun', // test id
			'mandatory-long-testrun', // variant id - can only be used for single variant test
		),
}
```

dotcom-rendering/src/web/experiments/lib/ab-exclusions.ts
vlbee: Fair point. I'm happy to close this PR and just use getCountryCodesSync in the test def

PR #6005, opened 2022-09-14
Add the labs traffic driver shady pie

emma-imber:
## What does this change?
Adds a new version of the shady pie to be shown on lifestyle articles to promote a labs article. Different labs articles are shown for the US and UK editions, and for other editions the standard shady pie (digital subscription) should be shown.

## Why?
We want to help improve traffic to the labs articles, and this is one way that we can make use of the empty ad space when an ad blocker is in use.

## Screenshots

| Standard      | UK Lifestyle      | US Lifestyle
|Screenshot 2022-09-12 at 10 51 50|Screenshot 2022-09-12 at 10 51 50|Screenshot 2022-09-12 at 10 51 50


dotcom-rendering/src/web/components/ShadyPie.tsx
```diff
@@ -10,17 +13,45 @@ params.set(
...
-			<img
-				src="https://uploads.guim.co.uk/2020/10/02/Digisubs_MPU_c1_my_opt.png"
-				width="300"
-				alt=""
+export const ShadyPie = ({
+	sectionName,
+	editionId,
+}: {
+	sectionName: string;
+	editionId: EditionId;
```
chrislomaxjones, line 21: You could make these props optional and avoid having to pass in default values. I don't think you'd have to change your if statements further down, since they'll cover the case when either property is undefined (you'll get the original shady pie)

```suggestion
	sectionName?: string;
	editionId?: EditionId;
```

chrislomaxjones (commented):
Nice work! Just had a few suggestions

chrislomaxjones:
Just an idea, but as a follow-up it might be cool to create a Storybook story for this new Labs component. It would potentially make it easier to develop against (just run `yarn storybook`) without having to reach for an ad blocker and so on.

dotcom-rendering/src/web/layouts/StandardLayout.tsx
```diff
@@ -761,6 +761,8 @@ export const StandardLayout = ({ CAPIArticle, NAV, format }: Props) => {
 												CAPIArticle.pageType
 													.isPaidContent
 											}
+											sectionName={CAPIArticle.pillar}
```
oliverlloyd, line 764: Could we replace `CAPIArticle.pillar` with `format`, please 🙏 `CAPIArticle.pillar` is a legacy property which although not quite deprecated it could soon be. We now use `format` when making decisions based on the pillar of an article.

dotcom-rendering/src/web/components/ShadyPie.tsx
```diff
@@ -10,17 +13,45 @@ params.set(
...
+}) => {
+	let pieFilling;
+	if (sectionName == 'lifestyle' && editionId == 'UK') {
+		pieFilling = (
```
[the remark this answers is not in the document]
oliverlloyd: I agree, the naming is fantastic, but it is better here to just return early

dotcom-rendering/src/web/components/LabsShadyPie.tsx
```diff
@@ -0,0 +1,105 @@
...
+const Logo = () => (
+	<Link href="https://www.theguardian.com/guardian-labs">
+		<LabsLogo />
+	</Link>
+);
+
+export const LabsShadyPie = ({
+	articleTitle,
+	articleImageLink,
+	articleLink,
```
oliverlloyd, line 68: It might be better removing 'article' from the naming here, it's not unreasonable to think we might end up using this component in other, non article, contexts in future so it might be best to keep it simple

dotcom-rendering/src/web/components/LabsShadyPie.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+const parentContainer = css`
+	display: flex;
+	align-items: flex-end;
+`;
+
+const labsLogoContainer = css`
+	width: 180px;
+	display: inline-block;
+	padding: 0.5rem;
```
oliverlloyd, line 51: We use pixels in dcr. It's annoying, how do you know when to use one or the other?! It's a whole thing. I think we might align better in the future cross team / codebases but for now DCR uses pixels everywhere.

dotcom-rendering/src/web/components/LabsShadyPie.tsx
```diff
@@ -0,0 +1,105 @@
...
+	sponsorLogoLink: string;
+}) => {
+	return (
+		<div css={shadyBackground}>
+			<div css={shadyBanner}>
+				<p>Paid content</p>
+			</div>
+			<a
+				css={shadyStyle}
+				href={articleLink + '?utm_source=shady-pie'}
```
oliverlloyd, line 83: I guess we could just add this query param to the url in the previous component? That is probably simper and removes any chance of problems later with the url having it's own query params when passed in and this string concatenation adding a second '?'

dotcom-rendering/src/web/components/ShadyPie.tsx
```diff
@@ -10,17 +13,45 @@ params.set(
...
+			<LabsShadyPie
+				articleTitle="It’s all about the berries: meet one of the farmers growing blackcurrants for Ribena"
+				articleImageLink="https://i.guim.co.uk/img/media/f98f93c22a563b3107394d64e8410a062172de7d/0_37_2122_1273/master/2122.jpg?width=620&quality=45&fit=max&dpr=2&s=9e01623babd043c1a2fbaec9db29fca2"
```
oliverlloyd, line 28: Can we reduce this width value from `width=620`? That's pretty big and it looks like we only ever need a smaller size so we can keep the overall page size smaller by adjusting this url

oliverlloyd (commented):
This is great work. The code is simple, easy to understand and there is some fabulous variable naming going on!

Just some small comments and one request to replace `sectionName` with `format`

oliverlloyd:
Small aside: Do you think it would be possible to capture `ShadyPie` in a story. It might be nice to expose this component now that we're starting to add variation to it?

dotcom-rendering/src/web/components/ShadyPie.tsx
emma-imber: Thanks for the feedback Chris! Have implemented these suggestions 🙂

chrislomaxjones: Thanks!

dotcom-rendering/src/web/components/LabsShadyPie.tsx
```diff
@@ -0,0 +1,105 @@
...
+	articleLink,
+	sponsorLogoLink,
+}: {
+	articleTitle: string;
+	articleImageLink: string;
+	articleLink: string;
+	sponsorLogoLink: string;
+}) => {
+	return (
+		<div css={shadyBackground}>
```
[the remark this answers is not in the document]
emma-imber: Thanks for all the feedback Oliver! This was really useful for me - good to know about sizing in pixels for future reference 🙂

dotcom-rendering/src/web/components/AdSlot.tsx
```diff
@@ -28,6 +31,8 @@ type NonInlineProps = {
 	index?: never;
 	shouldHideReaderRevenue?: boolean;
 	isPaidContent?: boolean;
+	pillar?: ArticleTheme;
```
oliverlloyd, line 34: Can we make this `format` and then access it via `format.theme`. We take the approach of passing `format` by convention even if the code only accesses one property from the object. Sorry, I wasn't clear about that before!

oliverlloyd (commented):
This is looking great. Just a couple of additional comments, one is just a tiny refactor to meet the convention of passing `format` vs. `pillar`.

The other is that I think we also need to update the other places where we insert this `AdSlot`, this PR is updating `StandardLayout` but there are other layouts that render the same slot. You can find them all by searching for the string 'position="right"'

emma-imber:
> Small aside: Do you think it would be possible to capture `ShadyPie` in a story. It might be nice to expose this component now that we're starting to add variation to it?

Have just added a story for it - thank you both for the suggestion!

dotcom-rendering/src/web/components/TopRightAdSlot.importable.tsx
```diff
@@ -42,7 +47,7 @@ export const TopRightAdSlot = ({
 		!isServer
 	) {
 		// Show a fixed image asking people to subscribe
```
chrislomaxjones, line 49: Tiny one - can we remove this comment? It feels out of date now we also show labs articles
```suggestion
```

chrislomaxjones (approved):
Awesome! Thanks for looking at all those suggestions.

Now we have some stories for this component, I think you can go ahead and approve them in the [Chromatic UI](https://www.chromatic.com/build?appId=5dfcbf3012392c0020e7140b&number=15596) (if you're happy with them, of course). 

There may also be some unrelated false positives, which you can safely approve if they seem to be completely unrelated to the presentational changes in this PR.

oliverlloyd (approved):
Very nice work!

dotcom-rendering/src/web/components/ShadyPie.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+export const Default = () => {
+	return (
+		<div style={{ width: '300px' }}>
+			<ShadyPie format={defaultFormat} editionId={'UK'} />
+		</div>
+	);
+};
+Default.story = {
+	name: 'Default',
+};
```
oliverlloyd, line 31: Just a thought, what happens if you don't supply either `format` or `editionId`? Given these are optional is that the true default?

emma-imber: It is the true default - turns out we don't need to supply format or editionId, so I've simplified the code for this story now

PR #6012, opened 2022-09-16
Dc subnav add editions switch

DanielCliftonGuardian:
## What does this change?
This adds the editions switch to DCR subnav
## Why?
Parity with frontend
## Screenshots

| Before      | After      |
|------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/110032454/191212435-6d3f1ab1-8480-48c0-a2bf-1308f477a82a.png

[after]: https://user-images.githubusercontent.com/110032454/190626387-8a815acc-662a-4100-8a48-721bf1e79789.png

### DIFF
--- a/dotcom-rendering/src/web/components/Nav/ExpandedMenu/Column.tsx
+++ b/dotcom-rendering/src/web/components/Nav/ExpandedMenu/Column.tsx
@@ -208,7 +208,7 @@ export const Column = ({
 	//Replace whitespace with hyphen https://stackoverflow.com/questions/3794919/replace-all-spaces-in-a-string-with/3795147#3795147
 	const columnInputId = `${column.title}-checkbox-input`.split(' ').join('-');
 	const collapseColumnInputId = `${column.title}-button`.split(' ').join('-');
-	const ariaControls = `${column.title.toLowerCase()}-Links`
+	const ariaControls = `${column.title.toLowerCase()}Links`
 		.split(' ')
 		.join('-');
 


dotcom-rendering/src/model/extract-nav.ts
```diff
@@ -14,6 +15,10 @@ interface LinkType extends BaseLinkType {
 	more?: boolean;
 }
 
+interface EditionLinkType extends LinkType {
+	id: EditionId;
```
oliverlloyd, line 19: Maybe `editionId: EditionId;` is clearer here?

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Column.tsx
```diff
@@ -34,7 +37,7 @@ const pillarDivider = css`
 	}
 `;
 
-const columnLinkTitle = css`
+export const columnLinkTitle = css`
```
oliverlloyd, line 40: Do we need these exported? I guessing this is left over from some earlier work

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -116,46 +122,140 @@ const brandExtensionLink = css`
...
+				<Column
+					column={column}
+					key={column.title.toLowerCase()}
+					index={i}
+					isLastColumn={i !== nav.pillars.length - 1}
+				/>
+			))}
+
+			<li>
+				<ThemeProvider theme={{ ...buttonThemeBrand }}>
```
oliverlloyd, line 196: I know that there is an existing use case for using Emotion theming but I'm nervous adding more as Emotion using React context to provide this value and I suspect in this case we don't need such an advanced solution and can style this button using a more basic approach. Happy to pair on looking at this in more detail.

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -116,46 +122,140 @@ const brandExtensionLink = css`
...
+					<div css={searchBar}>
+						<LinkButton
+							href="https://www.google.co.uk/advanced_search?q=site:www.theguardian.com"
```
oliverlloyd, line 199: Is this code from a different branch or are we also adding in the search feature here? Or am I reading this all wrong, also possible!

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -116,46 +122,140 @@ const brandExtensionLink = css`
...
 			<Column
-				column={column}
-				key={column.title.toLowerCase()}
-				index={i}
+				column={{
+					...activeEdition,
+					children: remainingEditions,
+				}}
+				index={10}
+				isLastColumn={true}
```
oliverlloyd, line 229: I think this is where we're inserting the edition dropdown?

Just a subjective thought, it might be nice to add some comment labels to this file so we can quickly see what's what. I'm not strong on this though

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Column.tsx
```diff
@@ -260,6 +266,7 @@ export const Column = ({
 					columnLinks,
 					index === 0 && firstColumnLinks,
 					!!column.pillar && pillarColumnLinks,
+					index === 10 && pillarColumnLinks,
```
oliverlloyd, line 269: Both these lines return the same css, can we merge them into one?

oliverlloyd: Also, index === 10 seems oddly specific, is this intentional?

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -116,46 +122,140 @@ const brandExtensionLink = css`
...
+			css={columnsStyle(format.display)}
+			role="menubar"
+			data-cy="nav-menu-columns"
+		>
+			{nav.pillars.map((column, i) => (
+				<Column
+					column={column}
+					key={column.title.toLowerCase()}
+					index={i}
+					isLastColumn={i !== nav.pillars.length - 1}
```
oliverlloyd, line 191: `isLastColumn` sounds like it will be true if the pillar is the last one but in fact it is being used in the opposite context

Maybe we can be more imperative here and use `showLineBelow`? And then refactor around this

oliverlloyd (commented):
This is looking pretty great. I've got some thoughts and comments on some small areas but nothing major.

Also, @HarryFischer are there any design requirements for how this edition dropdown should look on mobile?

Oh, and one more very small thing: would it be possible to update the before and after images to be _DCR_ before and after. I think that before image is of Frontend. We could also include a screenshot of frontend and have three columns?

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
oliverlloyd: Ah, cool. I think in this case if you merge `main` into this branch then that should clear things up automatically

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
mxdvl: This PR only moves things around, and the Theming was already in place, which is the recommended approach from Source. I don’t think this should be a blocker for this PR. I don’t believe any of this makes its way to the client-side in any case.

mxdvl (approved):
I would be happy for this to go through, but it seems like our accessibility metrics have seriously gone down in our Lighthouse test.

Any idea why that is? Would need sorted before merging.

PR #6018, opened 2022-09-20
Revert "Revert "Revert "Scale up servers from 9 > 18"""

ioannakok:
No need to be scaled up anymore.
Reverts guardian/dotcom-rendering#6013


AshCorr (approved):
Revert Revert Revert 😂 I think we should keep reverting these PR's every time we need to scale up and see how long we can make the PR title!

LGTM!

PR #6029, opened 2022-09-21
Cypress runs headless Chrome locally and mutes audio

arelra:
## What does this change?

- Updates all local Cypress tasks (make and Yarn) to force headless Chrome rather than the default Electron browser.
- Mute audio when running headless 🙉 

## Why?

- aligns with CI script which already forces Chrome
- Chrome allows much more control of its behaviour via command line arguments (controlling audio for example :)
- anecdotally it's approx 30% faster than Electron when running tests locally

### DIFF
--- a/dotcom-rendering/cypress.config.js
+++ b/dotcom-rendering/cypress.config.js
@@ -27,6 +27,13 @@ module.exports = defineConfig({
   },
   e2e: {
     setupNodeEvents(on, config) {
+	  on('before:browser:launch', (browser = {}, launchOptions) => {
+	    // mute audio in Chrome when running headless
+	    if (browser.family === 'chromium' && browser.isHeadless) {
+	      launchOptions.args.push('--mute-audio');
+	    }
+		return launchOptions;
+      });
       return plugins(on, config)
     },
     baseUrl: 'http://localhost:9000/',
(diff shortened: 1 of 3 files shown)


jamesgorrie (commented):
You can [set this in the config](https://docs.cypress.io/guides/guides/launching-browsers#Customize-available-browsers) avoiding needing to have it in all the run commands. Would that be cleaner?

OllysCoding (approved):
LGTM!

arelra:
> You can [set this in the config](https://docs.cypress.io/guides/guides/launching-browsers#Customize-available-browsers) avoiding needing to have it in all the run commands. Would that be cleaner?

I _think_ that's only for `cypress open` rather the headless mode:

> config object and return the list of browsers you want available for selection during cypress open.

Looks like there's an open PR requesting configuration of the default headless browser via `cypress.config`:

https://github.com/cypress-io/cypress/issues/6646

jamesgorrie (approved):

dotcom-rendering/cypress.config.js
```diff
@@ -4,31 +4,38 @@ import plugins from './cypress/plugins';
...
+	retries: {
+		runMode: 2,
+		openMode: 0,
+	},
+	e2e: {
+		setupNodeEvents(on, config) {
+			on('before:browser:launch', (browser = {}, launchOptions) => {
+				// mute audio in Chrome when running headless
+				if (browser.family === 'chromium' && browser.isHeadless) {
+					launchOptions.args.push('--mute-audio');
```
mxdvl, line 33: This did not work for me locally… Did you try this?

PR #6039, opened 2022-09-22
Cypress runs headless Chrome locally and mutes audio

arelra:
Reattempt of https://github.com/guardian/dotcom-rendering/pull/6029

---

## What does this change?

- Updates all local Cypress tasks (make and Yarn) to force headless Chrome rather than the default Electron browser.
- Mute audio when running headless 🙉 
- TeamCity build remains on Electron as Chrome is not available for ARM64 environments

## Why?

- aligns with CI script which already forces Chrome
- Chrome allows much more control of its behaviour via command line arguments (controlling audio for example :)
- anecdotally it's approx 30% faster than Electron when running tests locally


dotcom-rendering/makefile
```diff
@@ -95,6 +95,11 @@ dev-legacy: clear clean-dist install gen-dotenv
 
 cypress: export NODE_ENV=production
 cypress: clear clean-dist install build
+	$(call log, "starting frontend PROD server for Cypress")
+	@DISABLE_LOGGING_AND_METRICS=true start-server-and-test 'node dist/frontend.server.js' 9000 'cypress run --browser chrome --spec "cypress/e2e/**/*"'
+
+cypress-arm64: export NODE_ENV=production
+cypress-arm64: clear clean-dist install build
```
arelra, line 102: New makefile Cypress task specifically for TeamCity / ARM64 that runs with the default Electron browser

oliverlloyd (approved):

dotcom-rendering/makefile
mxdvl: Is there no way around that? The number of environments we have seems to keep growing all the time…

mxdvl (approved):
Peace and quiet…

dotcom-rendering/makefile
jamesgorrie: Could be worth noting in the file - maybe noting that we should deprecate this. I've added it to this issue:

PR #6040, opened 2022-09-22
Add `canonical` link to `pageTemplate`

jamesgorrie:
## What does this change?
Adds the `<link rel="canonical" href="{url}" />` to article pages.

## Why?

While digging into [Google search console](https://search.google.com/search-console) there are some reasons why pages aren't being indexed. The second most frequent reason, and the only one that seems to be growing steadily over time was "[Duplicate page without canonical tag](https://support.google.com/webmasters/answer/7440203#duplicate_page_without_canonical_tag)". This lead me to [this solution](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls?visit_id=637994434325810631-1629966483&rd=1) - which is what is implemented here.

Here's a sample of the URLs it's suggesting aren't being indexed:
1. https://www.theguardian.com/business/2021/feb/14/uk-airlines-warn-of-job-losses-as-they-lose-business-to-brexit?CMP=Share_iOSApp_Other
2. https://www.theguardian.com/technology/2016/aug/01/amazon-noise-cancelling-headphones-know-your-name?utm_content=buffer42bd9&utm_medium=social&utm_source=twitter.com&utm_campaign=buffer
3. https://www.theguardian.com/books/live/2017/feb/10/neil-gaiman-webchat-norse-mythology?page=with:block-58a2f6a8e4b09b65848eb7a1

Here's the growing graph (numbers and links omitted for privacy):
Screenshot 2022-09-22 at 12 36 49

This makes sense that we should be setting a canonical, as the canonical for `1.` should be `https://www.theguardian.com/business/2021/feb/14/uk-airlines-warn-of-job-losses-as-they-lose-business-to-brexit`.

We should probably make this non-optional and add it to fronts, but thought I would start with this.

## Screenshots

URL: http://localhost:3030/Article?url=https://www.theguardian.com/business/2021/feb/14/uk-airlines-warn-of-job-losses-as-they-lose-business-to-brexit?CMP=Share_iOSApp_Other

### before
Screenshot 2022-09-22 at 14 08 32

### after
Screenshot 2022-09-22 at 14 08 17

### DIFF
--- a/dotcom-rendering/src/web/server/articleToHtml.tsx
+++ b/dotcom-rendering/src/web/server/articleToHtml.tsx
@@ -275,5 +275,6 @@ window.twttr = (function(d, s, id) {
 				: undefined,
 		recipeMarkup,
 		offerHttp3,
+		canonicalURL: url,
 	});
 };
--- a/dotcom-rendering/src/web/server/pageTemplate.ts
+++ b/dotcom-rendering/src/web/server/pageTemplate.ts
@@ -21,6 +21,7 @@ export const pageTemplate = ({
 	twitterData,
 	initTwitter,
 	recipeMarkup,
+	canonicalURL,
 }: {
 	css: string;
 	html: string;
@@ -38,6 +39,7 @@ export const pageTemplate = ({
 	twitterData?: { [key: string]: string };
 	initTwitter?: string;
 	recipeMarkup?: string;
+	canonicalURL?: string;
 }): string => {
 	const favicon =
 		process.env.NODE_ENV === 'production'
@@ -184,6 +186,11 @@ https://workforus.theguardian.com/careers/product-engineering/
 			    ${weAreHiringMessage}
                 <title>${title}</title>
                 <meta name="description" content="${he.encode(description)}" />
+				${
+					canonicalURL !== undefined
+						? '<link rel="canonical" href="${canonicalURL}" />'
+						: '<!-- no canonical URL -->'
+				}
                 <meta charset="utf-8">
 
                 <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">


oliverlloyd:
Is there a way that we can test/verify these sorts of changes? It'd be great to understand their impact

jamesgorrie:
@oliverlloyd 

> Is there a way that we can test/verify these sorts of changes? It'd be great to understand their impact

I would expect that graph in Google search console to drop, and become stable.

OllysCoding (approved):
+1 It'll be important to monitor whether we end up with duplicate canonical URLs after this change. 

From previous experience, picking the canonical url means things like ensuring they are case sensitive, www. vs not www., etc. It's not immediately clear to me from this whether that will be a problem or not, but from my understanding it likely won't be. The only way to know for sure will be to monitor search console of course!

oliverlloyd:
> > Is there a way that we can test/verify these sorts of changes? It'd be great to understand their impact
> 
> I would expect that graph in Google search console to drop, and become stable, indicating that Google is indexing pages that for this reason.

But the motivation for this change is, I presume, in response to indexing concerns and if we're hoping to move that dial it'd be great to have a way to attribute any impacts. Maybe we could  look at some form of AB test?

It might be that we can't do this, but if that is the case we should, at a minimum, be careful to only make one change at a time so that we don't see results improve or degrade and not have the ability to understand why.

mxdvl (approved):
Should this be added to AMP, as well?

jamesgorrie:
>  in response to indexing concerns

It's more a response to graph I saw in the search console, and the recommendation that was suggested through Google's documentation.

The issue with the A/B test is that the data we're using for validation is from Google, which doesn't (and shouldn't) know about tests i.e. I can't ask Google to search for `variantA` within articles that aren't being indexed for not having a canonical rel.

I have tried to see if we can search for maybe section URLs or similar in the console, and no dice, so ruling out applying this across a section or similar.

> be careful to only make one change at a time

Assuming we mean [this PR](https://github.com/guardian/frontend/pull/25517) - I wouldn't have expected that to affect this graph at all, but take that it might affect other SEO data - so agree it might be prudent to hold off on that (which @OllysCoding has brought to attention that I'm not sure how to see what effect that would have). I've marked that as draft for now.

jamesgorrie:
> Should this be added to AMP, as well?

It is: https://github.com/guardian/dotcom-rendering/blob/3ac24426c459fb8456c81b5148da6c883872189e/dotcom-rendering/src/amp/server/document.tsx#L59

oliverlloyd:
> The issue with the A/B test is that the data we're using for validation is from Google, which doesn't (and shouldn't) know about tests i.e. I can't ask Google to search for `variantA` within articles that aren't being indexed for not having a canonical rel.
> 
> I have tried to see if we can search for maybe section URLs or similar in the console, and no dice, so ruling out applying this across a section or similar.

You can hash the url and use a mod of that to bucket particular articles into test groups. Like this a particular page will only ever be served with the control / variant, never anything else. This approach is mostly implemented in [this branch](https://github.com/guardian/frontend/tree/oliver/google-indexing-test).

jamesgorrie:
@oliverlloyd 
> > The issue with the A/B test is that the data we're using for validation is from Google, which doesn't (and shouldn't) know about tests i.e. I can't ask Google to search for `variantA` within articles that aren't being indexed for not having a canonical rel.
> > I have tried to see if we can search for maybe section URLs or similar in the console, and no dice, so ruling out applying this across a section or similar.
> 
> You can hash the url and use a mod of that to bucket particular articles into test groups. Like this a particular page will only ever be served with the control / variant, never anything else. This approach is mostly implemented in [this branch](https://github.com/guardian/frontend/tree/oliver/google-indexing-test).

But the data that I am using to confirm that we no longer get this error is from Google, as they are where this error is reported - so I can't analyse the data irrespective of how we bucket internally.

oliverlloyd:
> But the data that I am using to confirm that we no longer get this error is from Google, as they are where this error is reported - so I can't analyse the data irrespective of how we bucket internally.

I was referring to testing for any impact to indexing, not error graphs. If you're strong on not using an AB test for this I'm happy to concede. I just thought it would be good to have a controlled approach to any changes as often in these situations it is very tempting to make lots of changes because they look like they could help but it's hard to ever know if they did

jamesgorrie:
> I was referring to testing for any impact to indexing, not error graphs. 

Ah, OK - I was getting a bit confused as the graph I am using is a graph for why things aren't being indexed, whereas the graph we're creating internally is to help ascertain why things are being indexed slowly.

I am very tempted to see these as two separate things, but if you think this would make it harder to analyse the speed graphs, happy to sit and try come up with a plan.

OllysCoding:
I would say that from my experience canonicals are treated as no-brainers when it comes to SEO, I don't think we need to go to massive extents to test this - but rather a good practice we should implement while we have heightened focus on this area!

mchv:
I would recommend to merge that as soon as possible

rtyley:
Let's keep a note of when deployment has completed for this PR...!

Right now, if I execute this request, I don't get a canonical back:

```
curl --silent "https://www.theguardian.com/world/2022/sep/22/covid-hospitalisations-rise-by-nearly-20-in-a-week-in-england" | grep canonical
```

rtyley:
Looks like theres a bug? `${canonicalURL}` is not being evaluated...

```
% curl --silent "https://www.theguardian.com/environment/2022/sep/23/rees-mogg-constituency-fracked-tory-shale-gas-drilling" | grep canonical
				<link rel="canonical" href="${canonicalURL}" />
```

mxdvl:
> Looks like theres a bug? `${canonicalURL}` is not being evaluated...

@rtyley  should be fixed by #6048

PR #6042, opened 2022-09-22
Snapshot using fewer breakpoints

oliverlloyd:
## What does this change?
This reduces the number of breakpoints we use to take Chromatic snapshots.

## Why?
This will make it easier to review diffs and speed up our test suite.

## Why not?
The argument against this change is that we reduce the level of test coverage. It's true that by not running regression tests at all breakpoints we risk a problem slipping through, but in my experience we very rarely see visual bugs are breakpoints other than desktop or mobile so I think this is a fair tradeoff.

| Before      | After      |
|-------------|------------|
| Screenshot 2022-09-22 at 16 56 07  | Screenshot 2022-09-22 at 16 55 50 |

### DIFF
--- a/apps-rendering/src/components/Layout/Layout.stories.tsx
+++ b/apps-rendering/src/components/Layout/Layout.stories.tsx
@@ -240,7 +240,11 @@ export default {
 		layout: 'fullscreen',
 		chromatic: {
 			diffThreshold: 0.4,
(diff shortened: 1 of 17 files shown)


mxdvl (approved):
I think that is fair according to my experience reviewing Chromatic! Why look at `wide` over `desktop`?

We can always revisit in the future if we do stop a regression.

Issue #5647, opened 2022-08-09
Move AR API Models To New Repository


JamieB-gu:
*This may or may not be necessary.*

Some automated publishing methods trigger a release every time a PR is merged. This repository contains far more than just the API models, so the vast majority of PRs should not trigger a release. We could figure out some way to filter the automated releases, or we could move the models to a dedicated repository and have all PRs trigger releases.

JamieB-gu:
Let's have a meeting to confirm we want to do this.

PR #6043, opened 2022-09-22
chore: remove apps-rendering-api-models

### LINKED ISSUE #6041: Remove `apps-rendering-api-models` from `apps-rendering`
As `apps-rendering-api-models` are moving to their own repository (#5647), we should delete the files from the `dotcom-rendering` repository

georgeblahblah:
## What does this change?

- Deletes `apps-rendering-api-models` files, as they have been moved to [`guardian/apps-rendering-api-models`](https://github.com/guardian/apps-rendering-api-models), as they have been moved to their own repository in #5647
- Documents the new home for the models, and the purpose of the models

## Why?

Maintaining `apps-rendering-api-models` in this repository seems reasonable at first glance. These models _are_ related to `apps-rendering` after all, and so it makes sense to co-locate them with the code. However, this co-location has caused some issues for us:

- `apps-rendering-api-models` are a Scala project with a Scala build chain. `apps-rendering` is a node/Typescript project. Having two sets of build tools can be confusing
- The models are also used by `mobile-apps-api`. Why should they be kept here, instead of living `mobile-apps-api`?
- Up until now, the release process has been entirely manual. We would like to automate as much of this process as possible (see #5645, #5646). However, introducing new workflows and tooling here is complicated by the fact we share this repository with `dotcom-rendering`
- Many other similar models (https://github.com/guardian/content-atom, https://github.com/guardian/content-api-models) have their own repositories, so this is an established pattern
- We would like to take advantage of [GitHub Releases](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository) so people can easily see changelogs between model versions. If the models remain here and we create Releases for them, it is confusing to have those Releases associated with `dotcom-rendering` - the two are only tangentially related.

This fixes #6041


JamieB-gu (approved):
👍 

Are there any docs we should be adding/updating to explain what the API models are and where to find them?

georgeblahblah:
> 👍
> 
> Are there any docs we should be adding/updating to explain what the API models are and where to find them?

Great idea. I've added some docs in our README to explain the purpose of the models, along with where to find them.

PR #6047, opened 2022-09-23
Run native element positioning on 'load'

ashishpuliyel:
## Why?
We have a recurring issue where the advertising (which is rendered natively) appears higher than the ad-slot and obscures page content.

Both videos and advertising are rendered natively and layered over the webview at the expected position in the view to overlay the placehodler that the web-layer draws.

It's proven difficult to replicate this issue reliably. One theory is that the web layer tells the native layer where the slots should be, and then something additional loads (an image or a figure?) which pushes down the article content, but it doesn't update the native layer with the new position for the slot (which is now further down), and the native ad floats over article content.

Looking at the [source](https://github.com/guardian/dotcom-rendering/blob/main/apps-rendering/src/client/nativeCommunication.ts) I see it essentially does three things: 

1. `265: void document.fonts.ready.then(() => {` when the fonts are loaded
2. `251:  window.addEventListener( 'resize',` when the window is resized (I think the most common case for this will be orientation change)
3. `261: const observer = new MutationObserver(callback);` on DOM tree mutation. Note: there are changes, (loading images or text changes for example) that will cause reflow, but won't cause the mutationObserver to fire.

For each of these events it checks to see if the position has changed (line 239), and then if it has, it updates the advert position.

This PR adds a single additional invocation of the check for positioning against the DOM 'load' event, which should occur after all images and subviews are loaded.

The other change this PR makes it it moves the initial `reportNativeElementPositionChanges();` to be the last in a chain of events in rendering an article, so now it runs after as much of 'work' is done as possible. This is a belt-and-braces approach and I could be persuaded to drop that part for now.

## Screenshots of the bug
Screenshot_20220731-124547


apps-rendering/src/client/article.ts
```diff
@@ -475,3 +474,4 @@ initAudioAtoms();
 hydrateAtoms();
 richLinks();
 hydrateClickToView();
+reportNativeElementPositionChanges();
```
georgeblahblah, line 477: What do you think about adding a comment here? I think it might be useful to explain we want `reportNativeElementPositionChanges()` to come last and possibly link back to this PR

ashishpuliyel: A good suggestion which I've taken

georgeblahblah (approved):
I think this is a sensible approach to try and solve the problem. I've put the branch on CODE Apps Rendering. In my unscientific testing, I haven't yet managed to trigger the bug whereas it was easier for me to do so before deploying this branch.

ashishpuliyel:
> In my unscientific testing, I haven't yet managed to trigger the bug whereas it was easier for me to do so before deploying this branch.
That's good news. Thanks for your help. Lets merge early next week.

@mohammad-haque FYI

mohammad-haque:
> > In my unscientific testing, I haven't yet managed to trigger the bug whereas it was easier for me to do so before deploying this branch.
> 
> That's good news. Thanks for your help. Lets merge early next week.
> 
> @mohammad-haque FYI

Thanks for this PR, hopefully this will fix this bug, we will only verify it once it goes out to users.

PR #6050, opened 2022-09-23
Use `canonicalUrl`

oliverlloyd:
## What?
This replaces the use of `webUrl` as the canonical url, which was derived locally, with `canonicalUrl` which is provided by Frontend

## Why?
There [is code in Frontend](https://github.com/guardian/frontend/blob/main/common/app/common/LinkTo.scala#L93) which makes a more informed decision about what the canonical url should be based on which query params are being used. This is of particular importance for live blogs


mxdvl:
📇

mxdvl (approved):
📇

Issue #4238, opened 2022-03-14
Ads On Apps Deadblogs


JamieB-gu:
Ads on the apps are rendered natively. In AR we provide placeholders and then report their locations to the native layers via Bridget.

In normal articles these placeholders typically appear every few paragraphs. In live and dead blogs we'll instead want them to appear as placeholder _blocks_ in the list of content blocks. The ads themselves will then still be inserted via Bridget as normal.

Let's talk to the Commercial team to figure out the best algorithm for deciding how many blocks should be placeholders and where they should appear.

SiAdcock:
Hi @JamieB-gu, thanks for raising this, and I look forward to seeing AR-rendered liveblogs and deadblogs soon!

> Let's talk to the Commercial team to figure out the best algorithm for deciding how many blocks should be placeholders and where they should appear.

Ought the plan be to render ads in LiveBlogs and DeadBlogs in the same positions as they are currently rendered by mobile-apps-article-templates? If I'm not mistaken, MPUs are currently inserted [before the 2nd and 7th block](https://github.com/guardian/mobile-apps-article-templates/blob/70b8eb9456e90ebd94801d1b35e7ac3ee9a51cf6/ArticleTemplates/assets/js/modules/ads.js#L72-L83).

Alternatively, you might be considering the migration as an opportunity to revisit this approach. If so, I'd be interested to hear more about the thinking behind this 😊

Note that I'm going on holiday for 2 weeks after 29/7/22, so it would be helpful to keep @guardian/commercial-dev in the loop.

JamieB-gu:
> Ought the plan be to render ads in DeadBlogs in the same positions as they are currently rendered by mobile-apps-article-templates?

Yup we can do that if you think that's best 👍.

> MPUs are currently inserted after the 2nd and 7th block.

Should we do this on each page of the blog?

ashishpuliyel:
Can we come back to you early next week with more canonical advice, please? I think we've been wanting to revise the 2/7 rule for some time (because it's quite different from the web behaviour). But I'll need to do some digging before I can recommend a new rule.

ashishpuliyel:
@JamieB-gu I know this is a lot later than 'early next week'. Sorry. For now lets go with a modified version of 2nd and 7th:
2nd and every 5 blocks after that. So 2nd & 7th ... & 12th & 17th etc.

[...]

PR #6051, opened 2022-09-23
Add AdSlot component to deadblogs

iainjchambers-guardian:
## What does this change?
Adds ads to deadblogs at certain intervals between blocks

## Screenshots
 image image

### DIFF
--- a/apps-rendering/src/components/LiveBlocks/index.tsx
+++ b/apps-rendering/src/components/LiveBlocks/index.tsx
@@ -1,7 +1,8 @@
 // ----- Imports ----- //
 
 import type { Edition } from '@guardian/apps-rendering-api-models/edition';
-import { ArticleDesign, ArticleFormat } from '@guardian/libs';
+import type { ArticleFormat } from '@guardian/libs';
+import { ArticleDesign } from '@guardian/libs';
 import type { Option } from '@guardian/types';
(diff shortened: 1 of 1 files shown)


apps-rendering/src/components/LiveBlocks/index.tsx
```diff
@@ -29,6 +31,12 @@ const LiveBlocks: FC<LiveBlocksProps> = ({
 	const showPinnedPost =
 		pageNumber === 1 && pinnedPost.kind === OptionKind.Some;
 
+	const showAd = (index: number): boolean =>
+		// This can be removed when LiveBlogs are deployed
+		format.design === ArticleDesign.DeadBlog &&
+		// Add an AdSlot at 2nd and every other 5 blocks
+		(index === 1 || (index - 1) % 5 === 0);
```
georgeblahblah, line 38: coming back to this again today, I found it hard to figure out what this was doing (even with the comment!). I wonder if it would help to give a name to the value `1` here, so the calculations are easier to read. I thought of something like `firstAdOffset` but maybe there's a better option! What do you think?

```suggestion
		(index === firstAdOffset || (index - firstAdOffset) % 5 === 0);
```

georgeblahblah (approved):

JamieB-gu:
Fixes #4238

PR #6057, opened 2022-09-26
Actually serve smallest image on fallback

mxdvl:
## What does this change?

Actually pick the smallest image for fallback.
Set the `width` and `height` explicitely for the `img` tag based on that image’s dimensions by calculating its ratio.

## Why?

We are sorting the sources by width in a descending order. This means browsers without sources support get the largest source.

## Screenshots

| Before      | After      |
|-------------|------------|
| 21 kB | 6 kB |
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/192315759-4b93df40-f7b4-415a-b153-96e1e6e59e83.png
[after]: https://user-images.githubusercontent.com/76776/192315597-b112fc38-b51c-420a-9c52-5b5341abafd8.png

### DIFF
--- a/dotcom-rendering/src/web/components/Picture.tsx
+++ b/dotcom-rendering/src/web/components/Picture.tsx
@@ -283,7 +283,7 @@ export const Picture = ({
 						calculate width. The value of 167vh relates to an assumed image ratio of 5:3
 						which is equal to 167 (viewport height) : 100 (viewport width)
 
-						If either of these media queires match then the browser will choose an image from the
+						If either of these media queries match then the browser will choose an image from the
 						list of sources in srcset based on the viewport list. If the media query doesn't match
 						it continues checking using the standard sources underneath
 					*/}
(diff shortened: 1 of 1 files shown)


jlieb10 (approved):
Great catch 💯

dotcom-rendering/src/web/components/Picture.tsx
```diff
@@ -260,6 +260,14 @@ export const Picture = ({
 			};
 		});
 
+	const ratio = parseInt(height, 10) / parseInt(width, 10);
+	/**
+	 * The assumption here is readers on devices that do not support srcset
+	 * are likely to be on poor network connections so we're going
+	 * to fallback to the smallest image.
+	 */
+	const [fallbackSource] = sources.slice(-1);
```
jamesgorrie, line 269: Could we mention the order of the sources in the comment?

jamesgorrie: Completely sideline - [but could we use](https://caniuse.com/mdn-javascript_builtins_array_at) [Array.prototype.at](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at)? It is available at [polyfill.io](https://polyfill.io/) - but not sure what our rules are to adding to that is.

mxdvl: Yes, this could be a valuable refactor.

It’s only used in three instances in the codebase, so I don’t think it’s major. One blocker is that I could not get ESLint to understand that the result was indeed a `string | undefined` and not just an `any`…

image

Issue #5562, opened 2022-08-01
Prevent hiding Headlines


mxdvl:
We should not let users decide to hide the Headlines container, if we are to match frontend.

This is also a potential cause of CLS, as hiding the container is saved as a preference in the browser’s storage. Returning users who have hidden this container will get a short flash of it, before the island is hydrated.

Issue #5146, opened 2022-06-07
"nav/list" (`56`) container


bryophyta:
Part of https://github.com/guardian/dotcom-rendering/issues/4720

[Example](https://www.theguardian.com/info/complaints-and-corrections) (Useful links)
<img width="1335" alt="Screenshot 2022-10-06 at 16 55 42" src="https://user-images.githubusercontent.com/31692/194360993-63b3786f-a316-42d4-95be-02c713cb641d.png">

Please update [FaciaPicker in Frontend](https://github.com/guardian/frontend/blob/main/facia/app/services/dotcomrendering/FaciaPicker.scala) once this container is supported by DCR

PR #6064, opened 2022-09-27
Prevent hiding headlines in network fronts and 1st container in non-network fronts

ioannakok:
Closes https://github.com/guardian/dotcom-rendering/issues/5562

## What does this change?
* Brings `isNetworkFront` one level up in JSON schema and makes it required field
* Prevents hiding "Headlines" in network fronts
* Prevents hiding the 1st container of a non-network pressed front
* Prevents hiding `nav/list` and nav/media-list`

## Why?
To achieve parity with frontend 

## Screenshots

| Before      | After      |
|-------------|------------|
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |

### DIFF
--- a/dotcom-rendering/src/model/front-schema.json
+++ b/dotcom-rendering/src/model/front-schema.json
@@ -372,6 +372,9 @@
         },
         "isAdFreeUser": {
             "type": "boolean"
+        },
+        "isNetworkFront": {
+            "type": "boolean"
         }
     },
     "required": [
@@ -381,6 +384,7 @@
         "editionLongForm",
         "guardianBaseURL",
         "isAdFreeUser",
+        "isNetworkFront",
         "nav",
         "pageFooter",
         "pageId",
@@ -2580,9 +2584,6 @@
                             "uneditable"
                         ]
                     }
-                },
-                "isNetworkFront": {
-                    "type": "boolean"
                 }
             },
             "required": [
--- a/dotcom-rendering/src/web/layouts/FrontLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/FrontLayout.tsx
@@ -8,7 +8,7 @@ import {
 } from '@guardian/source-foundations';
 import { StraightLines } from '@guardian/source-react-components-development-kitchen';
 import type { NavType } from '../../model/extract-nav';
-import type { DCRFrontType } from '../../types/front';
+import type { DCRCollectionType, DCRFrontType } from '../../types/front';
 import { AdSlot } from '../components/AdSlot';
 import { Footer } from '../components/Footer';
 import { Header } from '../components/Header';
@@ -54,6 +54,14 @@ const shouldInsertMerchHigh = (
 	}
 
 	return index === desiredPosition;
+};
+
+const isToggleable = (index: number, collection: DCRCollectionType) => {
+	return (
+		index != 0 &&
+		collection.collectionType != 'nav/list' &&
+		collection.collectionType != 'nav/media-list'
+	);
 };
 
 export const FrontLayout = ({ front, NAV }: Props) => {
@@ -251,7 +259,7 @@ export const FrontLayout = ({ front, NAV }: Props) => {
 								ophanComponentName={ophanName}
 								containerName={collection.collectionType}
 								containerPalette={collection.containerPalette}
-								toggleable={true}
+								toggleable={isToggleable(index, collection)}
 								sectionId={collection.id}
 								showDateHeader={
 									collection.config.showDateHeader
--- a/dotcom-rendering/src/web/layouts/FrontLayout.tsx
+++ b/dotcom-rendering/src/web/layouts/FrontLayout.tsx
@@ -56,12 +56,24 @@ const shouldInsertMerchHigh = (
 	return index === desiredPosition;
 };
 
-const isToggleable = (index: number, collection: DCRCollectionType) => {
+const isNavList = (collection: DCRCollectionType) => {
 	return (
-		index != 0 &&
-		collection.collectionType != 'nav/list' &&
-		collection.collectionType != 'nav/media-list'
+		collection.collectionType == 'nav/list' ||
+		collection.collectionType == 'nav/media-list'
 	);
+};
+
+const isToggleable = (
+	index: number,
+	collection: DCRCollectionType,
+	isNetworkFront: boolean,
+) => {
+	if (isNetworkFront) {
+		return (
+			collection.displayName.toLowerCase() != 'headlines' &&
+			!isNavList(collection)
+		);
+	} else return index != 0 && !isNavList(collection);
 };
 
 export const FrontLayout = ({ front, NAV }: Props) => {
@@ -259,7 +271,11 @@ export const FrontLayout = ({ front, NAV }: Props) => {
 								ophanComponentName={ophanName}
 								containerName={collection.collectionType}
 								containerPalette={collection.containerPalette}
-								toggleable={isToggleable(index, collection)}
+								toggleable={isToggleable(
+									index,
+									collection,
+									front.isNetworkFront,
+								)}
 								sectionId={collection.id}
 								showDateHeader={
 									collection.config.showDateHeader
@@ -282,7 +298,7 @@ export const FrontLayout = ({ front, NAV }: Props) => {
 							</Section>
 							{shouldInsertMerchHigh(
 								index,
-								front.pressedPage.isNetworkFront,
+								front.isNetworkFront,
 								front.pressedPage.collections.length,
 								front.pressedPage.frontProperties.isPaidContent,
 							) && (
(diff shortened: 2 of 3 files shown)


mxdvl (commented):
Simple, but I wonder if this check should happen one level up and as part of `toggleable`.

That would mean updating [`FrontLayout`](https://github.com/guardian/dotcom-rendering/blob/78554590f4c06751d9cf5c4a86866d2dd85d1753/dotcom-rendering/src/web/layouts/FrontLayout.tsx#L254) to set toggleable to `title != 'Headlines'`.

davidfurey:
It is sad that this is based on the container title, but since it is - it would be nice for dotcom and apps to use exactly the same logic - https://github.com/guardian/mobile-apps-api/blob/c782e8011a69e3f41a0e531ef3b81371b84308a9/mobile-fronts/app/data/facia/MobileFacia.scala#L473

ioannakok:
> 

Ah I missed that! Yes, that makes perfect sense, thanks!

ioannakok:
> It is sad that this is based on the container title, but since it is - it would be nice for dotcom and apps to use exactly the same logic - https://github.com/guardian/mobile-apps-api/blob/c782e8011a69e3f41a0e531ef3b81371b84308a9/mobile-fronts/app/data/facia/MobileFacia.scala#L473

Good point, thanks for the suggestion. I will do this.

jamesgorrie:
This feels like a silly question - but could this just be inactive for first of the containers - and then it would work across all fronts, like it currently does?

[There's a few other rules here](https://github.com/guardian/frontend/blob/1618e08cd9e7c3ef194157461dfc17f7a37b6be1/common/app/views/support/GetClasses.scala#L150-L151).

e.g. https://www.theguardian.com/education
Screenshot 2022-09-27 at 14 59 08

ioannakok:
Thanks all for your suggestions! I investigated how [the logic](https://github.com/guardian/frontend/blob/1618e08cd9e7c3ef194157461dfc17f7a37b6be1/common/app/views/support/GetClasses.scala#L150-L151) in frontend works and explaining here the reasons for the approach this PR is taking. 

TL;DR is we shouldn't aim to replicate the frontend logic as it is trying to deal with multiple types of fronts and has different effects on each type. Also, there are a couple of bugs / unexpected findings. Please see PR description for a more detailed summary. We cannot replicate the MAPI logic either because in frontend the toggle is not showing for all the 1st containers of pressed fronts (not just "Headlines").

## Frontend rules and what do they mean

```
val showToggle = !disableHide && !container.exists(!slices.Container.showToggle(_)) && !isFirst && hasTitle && !isHeadlines
```

* `disableHide`: Set to true only for [tag containers (rendered on index fronts)](https://github.com/guardian/frontend/blob/main/common/app/views/support/GetClasses.scala#L119-L133). Index fronts are currently out of scope so we're not moving this logic to DCR at this point.
* `container.exists(!slices.Container.showToggle(_))`: [false only for `nav/list` and `nav/media/list`](https://github.com/guardian/frontend/blob/main/common/app/layout/slices/Container.scala#L84-L88). I have never seen a `nav/list` or `nav/media/list`, especially in the first position of a front, but have moved this logic into DCR. Open to take this out though.
* `isFirst` is different in the context of frontend and DCR (see below)
* `hasTitle` : `displayName.isDefined`
* `isHeadlines`: `displayName.contains("headlines")` --> This is always false, even if the container's display name is "Headlines". The check should have been `displayName.contains("Headlines")` or `displayName.exists(_.toLowerCase().contains("headlines"))`.

## What is currently happening in frontend?

### Pressed but non-network fronts (e.g. https://www.theguardian.com/education)
The most straightforward case. The 1st collection in JSON is the same that is rendered both in frontend and DCR. `showToggle` is false when `isFirst`.

| JSON | What is rendered |
|------|------------------|
| image | image |


### Network fronts

#### frontend
1st collection in JSON is always the palette. Headlines is the first container rendered but the 2nd collection in JSON. Also the check `isHeadlines` returns always `false` (see above).

| JSON | What is rendered |
|------|------------------|
| image | image |

As a result, the toggle's presence is not defined by this logic. The toggle is present in the DOM and [gets hidden with CSS](https://github.com/guardian/frontend/blob/main/static/src/stylesheets/module/facia-garnett/_container.scss#L478-L481).

Screenshot 2022-10-04 at 16 51 02
Screenshot 2022-10-04 at 16 51 52

#### DCR
JSON is coming from frontend with the palette initially as the 1st collection in `FECollectionType` but [gets cleaned up in the enhancement stage](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/model/enhanceCollections.ts#L8-L21). 1st collection is "Headlines" in `DCRCollectionType`.

### Index fronts (e.g. https://www.theguardian.com/sport/cricket)
They don't have the toggle but they're out of scope.

## What is this PR doing?
Not showing the toggle for the first container or if the container's type is `nav/list` / `nav/media/list`.

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -56,6 +56,14 @@ const shouldInsertMerchHigh = (
 	return index === desiredPosition;
 };
 
+const isToggleable = (index: number, collection: DCRCollectionType) => {
+	return (
+		index != 0 &&
+		collection.collectionType != 'nav/list' &&
```
ioannakok, line 62: I have never seen a `nav/list` or `nav/media/list`, especially in the first position of a front, but have moved this logic into DCR. Open to take this out.

jamesgorrie:
@ioannakok re: `nav/list` - coincidently found an example yesterday:
https://github.com/guardian/dotcom-rendering/issues/5146

Saying that I would really like to get a UX perspective on this as I think consistency would be better that odd use cases that aren't transparent readers.

> 1st collection in JSON is always the palette.

This isn't always guaranteed, it's added in fronts tool. Maybe we can filter this out as we have another way of doing this in DCR? Or, I think we already do this [here](https://github.com/guardian/dotcom-rendering/blob/169bc0bbc63b9708ec167b9c71489c7625c127e1/dotcom-rendering/src/model/enhanceCollections.ts#L8-L12) so can believe the first container is a renderable one? Not sure if this feels more fitting in `frontend` as our data layer.

dotcom-rendering/src/web/layouts/FrontLayout.tsx
jamesgorrie: As per my comment, maybe consistency is better than having an arbitrary rule that readers need to some how reason about?

jamesgorrie (approved):

OllysCoding:
Great explanation, thank you @ioannakok 

Once question I have is whether it's worth keeping the 'headlines' title check for when there is a breaking story which takes the top slot in the network container. 

It's an edge case, but it might be that a user would hide the headlines container when its not in the top spot - then when it returns to the top spot they would not be able to show it again, making it hidden permanently.

ioannakok:
@jamesgorrie 

> This isn't always guaranteed

Sorry, if I wasn't clear. I mean _1st collection in JSON is always the palette in a network front on frontend_. On DCR it is not for the reason you're mentioning.

ioannakok:
@OllysCoding @jamesgorrie Thanks for your comments / suggestions! I am adding @isabellapurchas in the conversation to get some UX input! Happy to explain further or have a chat about this Bella to give you some context!

ioannakok:
> @OllysCoding @jamesgorrie Thanks for your comments / suggestions! I am adding @isabellapurchas in the conversation to get some UX input! Happy to explain further or have a chat about this Bella to give you some context!

Discussed today with @isabellapurchas and this is the behaviour this PR is introducing (and achieves parity with frontend):
In pressed fronts we prevent hiding:
* "Headlines" in network fronts. This rule covers the edge case @OllysCoding mentions. 
* The first container of a non-network front. These essentially act as the "headline" of the front, e.g. the 1st container of https://www.theguardian.com/education has displayName "Education"
* `nav-list` and `nav/media/list`. We move this functionality for parity with frontend in fronts like https://www.theguardian.com/info/complaints-and-corrections. However, UX / Design will revisit these fronts in the future.

dotcom-rendering/src/model/front-schema.json
```diff
@@ -372,6 +372,9 @@
         },
         "isAdFreeUser": {
             "type": "boolean"
+        },
+        "isNetworkFront": {
```
ioannakok, line 376: This needs to be one level up: https://github.com/guardian/frontend/pull/25519

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -56,6 +56,26 @@ const shouldInsertMerchHigh = (
...
+	index: number,
+	collection: DCRCollectionType,
+	isNetworkFront: boolean,
+) => {
+	if (isNetworkFront) {
+		return (
+			collection.displayName.toLowerCase() != 'headlines' &&
+			!isNavList(collection)
+		);
+	} else return index != 0 && !isNavList(collection);
```
mxdvl, line 76: Might be a bit odd to reason about the `isNavList` twice? Maybe we could have an early return if `isNavList`?

```
const isToggleable = (
	index: number,
	collection: DCRCollectionType,
	isNetworkFront: boolean,
) => {
	if (isNavList(collection)) return false;
	if (isNetworkFront) {
		return collection.displayName.toLowerCase() != 'headlines';
	}
	return index != 0;
};
```

PR #6066, opened 2022-09-27
refactor(Card): extract image

mxdvl:
## What does this change?

Settle on a single image value to avoid parallel branches that are harder to reason about.
Allow optional parameters not to be set.

## Why?

I was looking at this code, and this seemed like a worthwile refactor.

This reduces the amount HTML on Cards without an image, including the `img` tag without a `src` attribute.

```html
 <div><img alt="" role="presentation"></div>
```

## Screenshots

No change, as can be glanced by the absence of Chromatic diffs.

### DIFF
--- a/dotcom-rendering/src/web/components/Card/Card.tsx
+++ b/dotcom-rendering/src/web/components/Card/Card.tsx
@@ -310,8 +310,8 @@ export const Card = ({
 				dataLinkName={dataLinkName}
 			/>
 			<CardLayout
-				imagePosition={image ? imagePosition : 'top'}
-				imagePositionOnMobile={image ? imagePositionOnMobile : 'top'}
+				imagePosition={imagePosition}
+				imagePositionOnMobile={imagePositionOnMobile}
 				minWidthInPixels={minWidthInPixels}
 				imageType={image?.type}
 			>
--- a/dotcom-rendering/src/web/components/Card/Card.tsx
+++ b/dotcom-rendering/src/web/components/Card/Card.tsx
@@ -166,6 +166,18 @@ const CommentFooter = ({
 	);
 };
 
+const getImage = ({
+	imageUrl,
+	avatarUrl,
+}: {
+	imageUrl?: string;
+	avatarUrl?: string;
+}): { type: CardImageType; src: string } | undefined => {
+	if (avatarUrl) return { type: 'avatar', src: avatarUrl };
+	if (imageUrl) return { type: 'mainMedia', src: imageUrl };
+	return undefined;
+};
+
 export const Card = ({
 	linkTo,
 	format,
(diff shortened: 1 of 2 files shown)


dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -166,6 +166,18 @@ const CommentFooter = ({
...
 };
 
+const getImage = ({
+	imageUrl,
+	avatarUrl,
+}: {
+	imageUrl?: string;
+	avatarUrl?: string;
+}): { type: CardImageType; src: string } | undefined => {
+	if (avatarUrl) return { type: 'avatar', src: avatarUrl };
```
OllysCoding, line 176: This logic looks to be wrong? The previous lines:
```ts
if (imageUrl && avatarUrl) {
	imageType = 'avatar';
}
```

So this should be 
```ts
if (avatarUrl && imageUrl) return { type: 'avatar', src: avatarUrl };
```

That said I don't think I fully understand _why_ this logic is the way it is

dotcom-rendering/src/web/components/Card/components/CardLayout.tsx
```diff
@@ -37,8 +37,8 @@ const decideWidth = (minWidthInPixels?: number) => {
 };
 
 const decidePosition = (
-	imagePosition: ImagePositionType,
-	imagePositionOnMobile: ImagePositionType,
+	imagePosition?: ImagePositionType,
+	imagePositionOnMobile?: ImagePositionType,
```
ioannakok, line 41: Not sure I understand why these two are changing to optional. In the `Props` types they remain non-optional: https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/Card/components/CardLayout.tsx#L7-L8

And in `Card.tsx` above I see this:
image

dotcom-rendering/src/web/components/Card/Card.tsx
```diff
@@ -301,41 +310,37 @@ export const Card = ({
 				dataLinkName={dataLinkName}
 			/>
 			<CardLayout
-				imagePosition={imageUrl !== undefined ? imagePosition : 'top'}
-				imagePositionOnMobile={
-					imageUrl !== undefined ? imagePositionOnMobile : 'top'
-				}
+				imagePosition={image ? imagePosition : 'top'}
+				imagePositionOnMobile={image ? imagePositionOnMobile : 'top'}
```
mxdvl, line 314:
```suggestion
				imagePosition={imagePosition}
				imagePositionOnMobile={imagePositionOnMobile}
```

dotcom-rendering/src/web/components/Card/Card.tsx
mxdvl: I could not figure it out either, and this causes no visual difference, so tempted to simplify it like so.

If we want to capture a missing `imageUrl` and a set `avatarUrl`, we should have a story against it.

dotcom-rendering/src/web/components/Card/components/CardLayout.tsx
mxdvl: You are right, I have now fixed this in [b27392f](https://github.com/guardian/dotcom-rendering/pull/6066/commits/b27392f7ff6593bfc479bea0b1c7dfdece1d1d4d)

ioannakok (approved):
Looks good to me!

dotcom-rendering/src/web/components/Card/Card.tsx
ioannakok: I agree the previous logic is a bit weird and I prefer the new version. I'll leave this up to both of you. On my side, I'm happy with this PR so I have approved.

PR #6067, opened 2022-09-27
Add shady pie clickthrough test

emma-imber:
## What does this change?
Adds the shady pie clickthrough rate measurement test.

## Why?
We want to measure the clickthrough rate of the new labs shady pie component.

### DIFF
--- a/dotcom-rendering/src/web/components/TopRightAdSlot.importable.tsx
+++ b/dotcom-rendering/src/web/components/TopRightAdSlot.importable.tsx
@@ -1,8 +1,9 @@
 import type { SerializedStyles } from '@emotion/react';
 import { css } from '@emotion/react';
-import { getCookie } from '@guardian/libs';
+import { ArticlePillar, getCookie } from '@guardian/libs';
 import type { EditionId } from '../../types/edition';
 import { LABS_HEADER_HEIGHT } from '../lib/labs-constants';
+import { useAB } from '../lib/useAB';
 import { useAdBlockInUse } from '../lib/useAdBlockInUse';
 import { ShadyPie } from './ShadyPie';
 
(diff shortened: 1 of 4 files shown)


dotcom-rendering/src/web/components/TopRightAdSlot.importable.tsx
```diff
@@ -37,50 +39,74 @@ export const TopRightAdSlot = ({
...
+	useEffect(() => {
+		const isSignedIn =
+			!isServer && !!getCookie({ name: 'GU_U', shouldMemoize: true });
+		setIsShady(
+			(adBlockerDetected &&
+				!isSignedIn &&
+				!shouldHideReaderRevenue &&
+				!isPaidContent &&
+				!isServer) ||
+				!!userInShadyPieTestVariant,
```
chrislomaxjones, line 59: I think you can entirely delete all references to `isServer` - it's only ever used inside the body of the `useEffect` and `useEffect` doesn't run on the server, so it'd always be `false`.

chrislomaxjones, line 59: Wondering if we might want to adjust these rules slightly so that even if you're in the variant you won't see shady pie if you're signed in, it's paid content or reader-revenue should be hidden?

```suggestion
			(adBlockerDetected ||
				!!userInShadyPieTestVariant) &&
				!isSignedIn &&
				!shouldHideReaderRevenue &&
				!isPaidContent,
```

Maybe I've misunderstood the setup of this test though so ignore me if this is the intended behaviour!

dotcom-rendering/src/web/experiments/tests/shady-pie-click-through.ts
```diff
@@ -0,0 +1,22 @@
...
+	audienceOffset: 0,
+	audienceCriteria: 'All pageviews',
+	successMeasure:
+		'Getting a statistically significant measure of the click through rate',
+	canRun: () => true,
+	variants: [
+		{
+			id: 'variant',
+			test: (): void => {},
+		},
```
chrislomaxjones, line 20: Historically we've included an additional control group as one of these variants. This is to ensure we collect measurements for a group without the `variant` logic enabled. Do we need one for this test here?

```
		{
			id: 'variant',
			test: (): void => {},
		},
		{
			id: 'control,
			test: (): void => {},
		},
```

emma-imber: We don't need one for this test as our control is effectively the users with an ad blocker

dotcom-rendering/src/web/components/TopRightAdSlot.importable.tsx
emma-imber: Good catch! That sounds sensible

dotcom-rendering/src/web/experiments/tests/shady-pie-click-through.ts
chrislomaxjones: Thanks for clarifying!

chrislomaxjones (approved):
Look forward to the results.

PR #6068, opened 2022-09-27
Edit mostpop height when ad block is in use

emma-imber:
## What does this change?
Quick fix to stop the overlapping of the most popular articles when using an ad blocker

## Why?
The new labs shady pie is taller than the standard size shady pie, so this can cause the most popular articles to overlap when on a lifestyle page and using an ad blocker

## Screenshots

Before:

Screenshot 2022-09-27 at 15 01 33

After:

Screenshot 2022-09-27 at 15 02 28


OllysCoding (approved):
lgtm

Issue #5520, opened 2022-07-27
Full width ads on fronts


oliverlloyd:
The sport front for uk has a full width ad showing in the 3rd position
<img width="1440" alt="Screenshot 2022-07-27 at 15 00 20" src="https://user-images.githubusercontent.com/1336821/181266774-b4d06f43-297b-4c97-a5a5-b91eae53ea4f.png">

OllysCoding:
TODO: Mobile Ads

PR #6071, opened 2022-09-28
Mobile ads on fronts

DanielCliftonGuardian:
## What does this change?
This adds mobile ads to fronts. The positioning logic is taken from HtmlCleaner.scala 

Also moved the render logic of merchandising high slot as this was not being checked on some indexes.

## Why?
Parity with Frontend

Linked: https://github.com/guardian/dotcom-rendering/issues/5520
## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]:https://user-images.githubusercontent.com/110032454/194596649-e293289e-6d67-4f6b-bb1d-6fda24e6ce60.png

[after]: https://user-images.githubusercontent.com/110032454/194596532-262622db-1bb1-4322-9237-8ef5ffd56ee4.png


dotcom-rendering/src/web/components/AdSlot.tsx
```diff
@@ -459,6 +459,33 @@ export const AdSlot = ({
...
+			return (
+				<div
+					id={`dfp-ad--mobile-front`}
+					className={[
+						'js-ad-slot',
+						'ad-slot',
+						`ad-slot--mobile-front`,
+						'ad-slot--container-inline',
+						'ad-slot--rendered',
+					].join(' ')}
```
OllysCoding, line 472: Are these the same classes applied to ads on Frontend mobile ads?

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -34,26 +35,64 @@ const spaces = / /g;
...
+	return index === getMerchHighPosition(collectionCount, isNetworkFront);
+};
+// On mobile, we remove the first container if it is a thrasher
+// and remove a container if it, or the next sibling, is a commercial container
+// we also exclude any containers that are directly before a thrasher
+// then we take every other container, up to a maximum of 10, for targeting MPU insertion
+
+const mobileAddPosition: number[] = [];
+
+const getMobileAddPosition = (
```
OllysCoding, line 71:
```suggestion
const getMobileAdPosition = (
```

OllysCoding, line 71:
```suggestion
const mobileAdPosition: number[] = [];

const getMobileAdPosition = (
```

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -76,6 +115,13 @@ export const FrontLayout = ({ front, NAV }: Props) => {
 	 */
 	const renderAds = !front.isAdFreeUser;
 
+	if (front.isNetworkFront) {
+		getMobileAddPosition(
+			front.isNetworkFront,
+			front.pressedPage.collections,
+		);
+	}
```
OllysCoding, line 123: Right now, `mobileAddPosition` is defined in the root of the file, this is probably not best practice & potentially could cause issues with server side rendering (ad slots may persist between renders depending on how things work under the hood). 

Perhaps a refactor so that `getMobileAddPosition` returns an array would work better?

```js
// This can be declared at the root of the file
const getMobileAdPositions = (
	isNetworkFront: boolean | undefined,
	collections: DCRCollectionType[],
) => {
    let positions = number[]
	const merchHighPosition = getMerchHighPosition(
		collections.length,
		isNetworkFront,
	);

	collections.forEach((el, ind, arr) => {
		if (
			(ind === 0 && el.collectionType === 'fixed/thrasher') ||
			ind === merchHighPosition ||
			ind + 1 === merchHighPosition ||
			arr[ind + 1]?.collectionType === 'fixed/thrasher'
		) {
			return false;
		} else if (ind % 2 === 0) {
			positions.push(ind);
		}
		return false;
	});
    // separate suggestion for this incoming
	if (positions.length > 10) {
		positions.length = 10;
	}
};

// ....

export const FrontLayout = ({ front, NAV }: Props) => {
     // ...
     const mobileAdPositions = front.isNetworkFront ? getMobileAdPositions(...) : []
    // ...
```

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -34,26 +35,64 @@ const spaces = / /g;
...
+		) {
+			return false;
+		} else if (ind % 2 === 0) {
+			mobileAddPosition.push(ind);
+		}
+		return false;
+	});
+	if (mobileAddPosition.length > 10) {
+		mobileAddPosition.length = 10;
+	}
```
OllysCoding, line 95: I haven't tested this, but perhaps a better way to limit the length would be to slice the array. This would have to work in conjunction with my changes above, 


```ts
// Limit the number of ad positions to 10
mobileAdPosition = mobileAdPosition.slice(0, 10); 

```

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -34,26 +35,64 @@ const spaces = / /g;
...
 
-	return index === desiredPosition;
+	return index === getMerchHighPosition(collectionCount, isNetworkFront);
+};
+// On mobile, we remove the first container if it is a thrasher
+// and remove a container if it, or the next sibling, is a commercial container
+// we also exclude any containers that are directly before a thrasher
+// then we take every other container, up to a maximum of 10, for targeting MPU insertion
+
+const mobileAddPosition: number[] = [];
```
OllysCoding, line 69: We probably don't want to be defining this array at the root level of the document, see my other comments

dotcom-rendering/src/web/components/AdSlot.tsx
```diff
@@ -459,6 +459,33 @@ export const AdSlot = ({
...
+							position: relative;
+							min-height: 274px;
+							min-width: 300px;
+							width: 300px;
+							margin: 12px auto;
+						`,
+						adStyles,
+					]}
+					data-link-name={`ad slot mobile-front`}
+					data-name={`mobile-front`}
```
OllysCoding, line 484: I was having a look at mobile ads on Frontend on the /uk front, and it seems like all of the ads there (though I could be looking at the wrong ones) come through with the data-link-name as `ad slot inlineX` (where X is the index on the page) and data name `inlineX`. 

Are we purposefully changing the name here (if so have we discussed this with @guardian/commercial-dev) & should we be including the index in the naming scheme to ensure ad slots don't get mixed up?

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -182,18 +228,42 @@ export const FrontLayout = ({ front, NAV }: Props) => {
...
+									front.pressedPage.collections.length,
+									front.pressedPage.frontProperties
+										.isPaidContent,
+								) && (
+									<AdSlot
+										data-print-layout="hide"
+										position="merchandising-high"
+										display={format.display}
+									/>
+								)}
```
OllysCoding, line 256: Was the absence of potentially inserting the merchandising high slot here a bug we're fixing alongside adding mobile ads? If so it could be worth calling this our separately in the PR description!

OllysCoding (commented):
Overally things are looking really good - have left a few comments about refactors & general questions about the layout of the code, would be happy to go through these on a call if more information is needed!

dotcom-rendering/src/types/front.ts
```diff
@@ -254,6 +254,7 @@ export type DCRFrontCard = {
 	byline?: string;
 	showByline?: boolean;
 	avatarUrl?: string;
+	editionBrandings?: { edition: { id: EditionId } }[];
```
ioannakok, line 257: I don't think we need this anymore? If I remember correctly, it was an idea we had at some point to check whether a container is labs container but it turns out there is no need for this check.

oliverlloyd (commented):
This is great work. I know there was a lot of work working out this logic and bringing everything together so well done!

At this point my comments are all refactors and often subjective but I think this is a core area of the codebase so I think it's worth some extra thought.

One thing I was wondering is if it would be possible to refactor things such that instead of having a pre-defined list of mobile positions (`mobileAdPositions`) we had a sort of `shouldInsertMobileAd` function call? This would be similar to `shouldInsertMerchHigh` and would perhaps makes things easier to reason about?

Another advantage to having a should... function is we'd be able to pass in the result of the merchandising decision, rather than having to decide it again, which I think would better guard against potential future bugs if that code were ever modified. 

So you could have something like

```js
const insertMerchandisingSlot = shouldInsertMerchHigh( ...blah );
const insertMobileAd = shouldInsertMobileAd({
  position: index,
  hasMerchSlot: insertMerchandisingSlot,
  etc...
})
```

Very happy to chat this through more

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -34,26 +35,64 @@ const spaces = / /g;
...
+		isNetworkFront,
+	);
+
+	collections.forEach((el, ind, arr) => {
+		if (
+			(ind === 0 && el.collectionType === 'fixed/thrasher') ||
+			ind === merchHighPosition ||
+			ind + 1 === merchHighPosition ||
+			arr[ind + 1]?.collectionType === 'fixed/thrasher'
+		) {
```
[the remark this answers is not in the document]
mxdvl: I’m very much in favour of this approach.

I would also suggest this enables a cleaner, purely functional approach:

```ts
const positions = collections.reduce((positions, collection, index) => {
  if(shouldExcludePosition(index, collection, collections)) return positions.concat(index);
  return positions;
}, []) )
  // some comment about why we only pick the first ten…
  .slice(0,10)
return positions;
```

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -34,26 +35,91 @@ const spaces = / /g;
...
+		if (isNearMerchandising) return false;
+		if (isNearThrasher) return false;
+		else if (
+			collectionIndex % 2 === 0 &&
+			collectionIndex < collections.length - 1
+		) {
+			positions.push(collectionIndex);
+		}
+		return false;
+	});
```
mxdvl, line 85: What followes is a code style suggestion you may want to ignore!

If you are keen to use a purely functional style without side-effects, you could use `reduce`, or a series of `map` & `filter`. This is not a major issue, as the side effects do not escape the scope of this function, but happy to offer support if you would like to refactor now or at some later point.

For example:

```ts
const positions = collections
  .map((collection, collectionIndex) => {
    const isThrasher = collection.collectionType === "fixed/thrasher";
    const isFirst = collectionIndex === 0;
    const isNearMerchandising =
      collectionIndex === merchHighPosition ||
      collectionIndex + 1 === merchHighPosition;
    const isNearThrasher =
      collections[collectionIndex + 1]?.collectionType === "fixed/thrasher";
    if (isFirst && isThrasher) return false;
    if (isNearMerchandising) return false;
    if (isNearThrasher) return false;
    else if (
      collectionIndex % 2 === 0 &&
      collectionIndex < collections.length - 1
    ) {
      // This is where we avoid the mutation
      return true;
    }
    return false;
  })
  .map((shouldDisplayAd, collectionIndex) =>
    shouldDisplayAd ? collectionIndex : undefined
  )
  // we filter out `undefined` values
  .filter((index) => typeof index === "number")
  .slice(0, 10);
```

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -34,26 +35,91 @@ const spaces = / /g;
...
+		return 4;
 	}
+};
+
+// On mobile, we remove the first container if it is a thrasher
+// and remove a container if it, or the next sibling, is a commercial container
+// we also exclude any containers that are directly before a thrasher
+// then we take every other container, up to a maximum of 10, for targeting MPU insertion
+
+const getMobileAdPositions = (
```
mxdvl, line 56: If you turn this into JSDoc, it can help with documentation.

```suggestion
/**
  * On mobile, we remove the first container if it is a thrasher
  * and remove a container if it, or the next sibling, is a commercial container
  * we also exclude any containers that are directly before a thrasher
  * then we take every other container, up to a maximum of 10, for targeting MPU insertion
  */
const getMobileAdPositions = (
```

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -34,26 +35,91 @@ const spaces = / /g;
...
+				<AdSlot
+					index={index}
+					data-print-layout="hide"
+					position="mobile-front"
+					display={format}
+				/>
+			</Hide>
+		);
+	}
+	return false;
```
mxdvl, line 122: Generally we favour returning `null` or an empty fragment when we don’t want to display anything.

mxdvl (approved):
This looks sound!

dotcom-rendering/src/web/layouts/FrontLayout.tsx
oliverlloyd: +1 to using JSDoc here, the integration with VSCode and other IDEs is great

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -34,26 +35,91 @@ const spaces = / /g;
...
+		if (isNearThrasher) return false;
+		else if (
+			collectionIndex % 2 === 0 &&
+			collectionIndex < collections.length - 1
+		) {
+			positions.push(collectionIndex);
+		}
+		return false;
+	});
+	//Should insert no more than 10 ads
```
oliverlloyd, line 86:
```suggestion
	// Should insert no more than 10 ads
```

Complete nitpick but I think my linter might fall over on this so it is probably better to have a space here

dotcom-rendering/src/web/layouts/FrontLayout.tsx
oliverlloyd: +1 The [React docs say](https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering) you should return `null`

oliverlloyd (approved):
Pending the small comments and nitpicks, LGTM!

dotcom-rendering/src/web/layouts/FrontLayout.tsx
```diff
@@ -76,6 +55,100 @@ const isToggleable = (
...
+				collectionIndex < collections.length - 1
+			) {
+				return true;
+			}
+			return false;
+		})
+		.map((shouldDisplayAd, collectionIndex) =>
+			shouldDisplayAd ? collectionIndex : undefined,
+		)
+		.filter((index) => typeof index === 'number')
```
mxdvl, line 111: If you turn this into a type predicate, you won't have to assign the type type of positions and it will be `number[]`

```suggestion
		.filter((index): index is number => typeof index === 'number')
```

mxdvl (approved):
🚢

PR #6078, opened 2022-09-28
feat: helper to merge Dependabot branches

mxdvl:
## What does this change?

Add a helper to [get dependabot branches](https://github.com/guardian/dotcom-rendering/pulls/app%2Fdependabot) and easily merge them in.

See #6076 for example usage.

## Why?

Updating PRs one by one is mandraulic.

## Screenshots

image

### DIFF
--- a/scripts/deno/dependabot.ts
+++ b/scripts/deno/dependabot.ts
@@ -1,0 +1,41 @@
+/**
+ * A regex to filter matching branches
+ */
+const REGEX = Deno.args[0];
+
+const process = Deno.run({
+	cmd: [
+		'git',
+		'branch',
+		'-r',
+		'--list',
+		'origin/dependabot/*',
+		'--no-merged',
+	],
+	stdout: 'piped',
+});
+
+const branches = new TextDecoder()
+	.decode(await process.output())
+	.split('\n')
+	.map((line) => line.trim())
+	.filter((branch) => branch.match(REGEX));
+
+console.warn('About to merge the following branches:', branches);
+
+const shouldProceed = confirm('Do you want to merge these branches?');
+if (!shouldProceed) {
+	console.info('Not merging anything');
+	Deno.exit();
+}
+
+for (const branch of branches) {
+	const process = Deno.run({
+		cmd: ['git', 'merge', branch],
+	});
+	const { code } = await process.status();
+	if (code !== 0) Deno.exit(code);
+	console.info('merged', branch.split('/').at(-1));
+}
+
+Deno.exit();


JamieB-gu:
Would it be worth adding some docs to explain how this should be used? For example I'm assuming, based on your example PR, that:

1. You'd usually want to select specific branches using the command line argument, rather than merging _all_ dependabot branches
2. You'd want to run this command from a base branch that isn't `main`, and open a PR from that branch

?

mxdvl:
> Interesting idea! 💡 Would it be worth adding some docs to explain how this should be used?
> - @JamieB-gu 

Yes, I could have kept this a draft PR. The main thing was that I was trying to remember [what the magic command was](https://github.com/mxdvl/mxdvl/blob/d448dcc4b4433e14e0f50b909a8482fd5d8d3632/env/git.zsh#L7-L10) and thought it would be best to write a little helper. This is still very much half-baked.

jfsoul:
Hey! 👋 

As it happens DevX are trialling a new workflow to solve the same problem (I think!) More on that below, but first a disclaimer that the gold standard would probably be to build out an advanced CI/CD/monitoring pipeline that you trust enough to automerge (at least patch/minor) dependency updates to `main` if they pass tests. Even without that, I would recommend taking a look at renovate for batching updates together in the JS ecosystem - maybe talk to CSTI about their experience with it?

OTOH renovate doesn't fully support SBT/Scala and we've found that the manual effort to review/test/resolve/merge all the individual PRs on all the projects to be too much, so we've built a basic set of GitHub action workflows to batch the updates together. You target Dependabot/Scala Steward to a non-default branch, auto-merge PRs, and then every week/month a PR containing all the updates is generated for you. Check out the docs [here](https://github.com/guardian/.github/blob/main/.github/workflows/docs/pr-batching.md)

bryophyta (approved):
This looks good to me  👍 

If we are including this in the repo, would it make sense to add some kind of documentation? I don't think  it would need to be anything extensive, but maybe just some comments at the top of the file explaining how to run it and e.g. what's expected for the positional argument used to populate `REGEX`?

mxdvl:
> I would recommend taking a look at renovate for batching updates together in the JS ecosystem - maybe talk to CSTI about their experience with it?

I’m a big fan of Mend Renovate and use it in anger on all my personal repos. Last I checked, it had not been approved company-wide, hence my reluctance to use it there. We would love to be able to use it instead of dependabot as it:
- bumps related packages at once
- supports pnpm
- has a great dashboard view
- favours pinned dependencies

PR #6088, opened 2022-09-29
Liveblog epic fixes

tomrf1:
The liveblog epic is implemented differently from the standard epic and banner. The main difference is that it uses hooks instead of async calls for its dependencies.

There are 3 issues with the current implementation:
1. The `force-liveblog-epic` querystring parameter is not being handled correctly. The SDC library already has correct handling for this. I've created `useSDC.tsx`, which exports hooks for making requests to the SDC API. These use [SWR](https://swr.vercel.app/) (a standard way of doing api calls in DCR) and wrap the functions provided by the SDC library, which correctly handle the querystring param.
2. It gets the article count directly from localstorage, which is wrong. I've added a hook, `useArticleCounts`, which uses the existing `getArticleCounts` function.
3. The existing `useConsent` hook was preventing _any_ epic from showing if the user has not consented to article count. This is not the correct behaviour. I've replaced this with a `useConsentForArticleCount` hook.


dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
```diff
@@ -37,7 +37,7 @@ type Props = {
 	idApiUrl: string;
 	stage: string;
 	pageId: string;
-	keywordsId: string;
+	keywordIds: string;
```
tomrf1, line 40: While adding `keywordIds` for the liveblog epic I noticed it had a typo in various other places so fixed it

dotcom-rendering/src/web/lib/useSDC.tsx
```diff
@@ -0,0 +1,40 @@
...
+import useSWR from 'swr';
+
+const options: SWRConfiguration = {
+	revalidateIfStale: false,
+	revalidateOnFocus: false,
+	revalidateOnReconnect: false,
+};
+
+/**
+ * Hooks for making requests to the support-dotcom-components API
```
tomrf1, line 22: Currently only `useSDCLiveblogEpic` is used, as the other components use the SDC library directly. But in future we should migrate them to use these hooks so that we're consistent in using SWR

dotcom-rendering/src/web/components/LiveBlogEpic.importable.tsx
```diff
@@ -1,43 +1,31 @@
...
 
 type Props = {
 	section: string;
 	shouldHideReaderRevenue: boolean;
 	isPaidContent: boolean;
 	tags: TagType[];
 	contributionsServiceUrl: string;
-};
-
-const useConsent = () => {
```
tomrf1, line 29: `useConsent` and `consentGivenForArticleCounts` below are replaced with `useConsentForArticleCount`

dotcom-rendering/src/web/components/LiveBlogEpic.importable.tsx
```diff
@@ -122,18 +87,25 @@ const usePayload = ({
...
 	shouldHideReaderRevenue: boolean;
 	section: string;
 	isPaidContent: boolean;
 	tags: TagType[];
-}) => {
-	const consent = useConsent();
+	pageId: string;
+	keywordIds: string;
+}): EpicPayload | undefined => {
+	const articleCounts = useArticleCounts(pageId, keywordIds);
```
tomrf1, line 100: `useArticleCounts` is used instead of getting the data straight from localstorage

dotcom-rendering/src/web/components/SlotBodyEnd.importable.tsx
tomrf1, line 40: While adding `keywordIds` for the liveblog epic I noticed it had a typo in various other places so fixed it

dotcom-rendering/src/web/lib/contributions.ts
```diff
@@ -183,6 +184,21 @@ export const hasOptedOutOfArticleCount = async (): Promise<boolean> => {
 	return !hasCmpConsent || hasArticleCountOptOutCookie();
 };
 
+// A hook to find out if a user has opted out of article counting
+export const useHasOptedOutOfArticleCount = (): boolean | 'Pending' => {
+	const [hasOptedOut, setHasOptedOut] = useState<boolean | 'Pending'>(
+		'Pending',
+	);
+
+	useEffect(() => {
```
tjmw, line 193: DCR has a `useOnce` hook, which I think you could use in place of `useEffect` with an empty dependency array. I like it as the name makes it clear what the behaviour is.

dotcom-rendering/src/web/lib/useSDC.tsx
```diff
@@ -0,0 +1,40 @@
...
+	getPuzzlesBanner,
+} from '@guardian/support-dotcom-components';
+import type {
+	BannerPayload,
+	EpicPayload,
+} from '@guardian/support-dotcom-components/dist/dotcom/src/types';
+import type { SWRConfiguration } from 'swr';
+import useSWR from 'swr';
+
+const options: SWRConfiguration = {
```
tjmw, line 15: I think instead of specifying these options you could use `useSWRImmutable` which will give the same behaviour:  https://swr.vercel.app/docs/revalidation#disable-automatic-revalidations

dotcom-rendering/src/web/lib/useSDC.tsx
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Hooks for making requests to the support-dotcom-components API
+ */
+type UseSDC<T> = (
+	baseUrl: string,
+	payload: T,
+) => ModuleDataResponse | undefined;
+
+export const useSDCEpic: UseSDC<EpicPayload> = (baseUrl, payload) =>
+	useSWR('epic', () => getEpic(baseUrl, payload), options).data;
```
tjmw, line 30: `useSWR` also returns an `.error` property. I wonder if the underlying request fails would we want to log that somehow? Or perhaps that's already handled upstream (in `useEpic` for example)?

tjmw (approved):
I think this looks good 👍  I left a few suggestions, but nothing that's a blocker (i.e. feel free to ignore!)

dotcom-rendering/src/web/lib/useSDC.tsx
tomrf1: nice 👍

dotcom-rendering/src/web/lib/useSDC.tsx
tomrf1: the functions from SDC just throw an error - https://github.com/guardian/support-dotcom-components/blob/main/packages/dotcom/src/requests.ts#L49
I'll have a go at logging errors here

dotcom-rendering/src/web/lib/useSDC.ts
```diff
@@ -0,0 +1,43 @@
...
+	getBanner,
+	getEpic,
+	getLiveblogEpic,
+	getPuzzlesBanner,
+} from '@guardian/support-dotcom-components';
+import type {
+	BannerPayload,
+	EpicPayload,
+} from '@guardian/support-dotcom-components/dist/dotcom/src/types';
+import useSWRImmutable from 'swr';
```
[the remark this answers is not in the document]
tomrf1: is this causing an issue?

PR #6092, opened 2022-09-29
Use h2 tag for Advertisement header

### LINKED ISSUE #6091: Do not use `h1` tag for Advertisement header
The heading "Advertisement" above adverts should not be an `h1`, as we should have only one `h1` tag per page.

We should instead use non-heading markup, like `p` and/or `strong` tags. Although this isn't semantically correct, it is the best thing we can do for our readers using assistive technologies.

georgeblahblah:
## What does this change?

- Uses `h2` for the Advertisement header, instead of `h1`. This improves accessibility

## Why?

- Fixes #6091 

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

### DIFF
--- a/apps-rendering/src/adSlot.tsx
+++ b/apps-rendering/src/adSlot.tsx
@@ -13,7 +13,7 @@ import {
 	until,
 } from '@guardian/source-foundations';
 import { Button, buttonThemeBrandAlt } from '@guardian/source-react-components';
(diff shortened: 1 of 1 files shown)


georgeblahblah:
Closing because we don't think `h2` is the right solution for this.

PR #6097, opened 2022-09-29
Subnav add editions switch

DanielCliftonGuardian:
## What does this change?
This adds the editions switch to DCR subnav
## Why?
Parity with frontend
## Screenshots

| Frontend      | Before      | After      |
|------------|------------|------------|

[frontend]: https://user-images.githubusercontent.com/110032454/196415270-7e4d7b35-311d-4084-a857-1512ea57d0b6.png

[before]: https://user-images.githubusercontent.com/110032454/196415143-13d26a70-f878-4b1c-a144-d6c61d0b448a.png

[after]: https://user-images.githubusercontent.com/110032454/196415020-85444254-8dd3-45eb-badc-75797cd69e17.png

This PR's initial submission featured a visual bug on the desktop menu:

image

This was fixed by hiding the editions switch component from desktop.

Also fixed desktop nav bug by hiding lineStyle from desktop: 
image


dotcom-rendering/src/web/components/Nav/ExpandedMenu/Column.tsx
```diff
@@ -195,15 +198,19 @@ const columnStyle = css`
...
 }) => {
 	// As the elements are dynamic we need to specify the IDs here
-	const columnInputId = `${column.title}-checkbox-input`;
-	const collapseColumnInputId = `${column.title}-button`;
+	//Replace whitespace with hyphen https://stackoverflow.com/questions/3794919/replace-all-spaces-in-a-string-with/3795147#3795147
+	const columnInputId = `${column.title}-checkbox-input`.split(' ').join('-');
+	const collapseColumnInputId = `${column.title}-button`.split(' ').join('-');
+	const ariaControls = `${column.title.toLowerCase()}Links`
+		.split(' ')
+		.join('-');
```
OllysCoding, line 213:
```suggestion
	const columnId = column.title.toLowerCase().split(' ').join('-')
	const columnInputId = `${columnId}-checkbox-input`;
	const collapseColumnInputId = `${columnId}-button`;
	const ariaControls = `${columnId}Links`;
```

To me this reads a little bit better? It does mean they're all lower case, but I wasn't sure if that would be a problem?

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -131,74 +133,133 @@ const searchBar = css`
...
+		title: 'AU edition',
+	},
+];
+
+const getEdition = (editionId: EditionId): EditionLinkType => {
+	return (
+		editionList.find((edition) => edition.editionId === editionId) ??
+		editionList[1]
+	);
+};
```
OllysCoding, line 168: I think a lot of this is actually defined in [editions.ts](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/lib/edition.ts) - could that code be reused & upgraded to achieve this?

OllysCoding (commented):
Overall looks good! Could we include some screenshots on the desktop nav & a short description of what was changed from the previous PR to fix the bug with the desktop nav?

bryophyta:
A (hopefully quick!) question:

The styling of the dropdown here is different from how it is in Frontend. The edition dropdown has been promoted to the same level as e.g. Lifestyle in this PR. Was a decision already taken to make this change, or does it need to be discussed with design?
Frontend:
image

This PR (running locally):
image

(My concern is that the new styling makes it look slightly more like a section heading, rather than a switch, e.g. if you ignore the chevron on the side, it kind-of looks like `Search Jobs` etc. is part of `UK Edition`. I don't have super strong opinions on this but wanted to check whether it was a conscious decision or not.

image

oliverlloyd:
> The styling of the dropdown here is different from how it is in Frontend. The edition dropdown has been promoted to the same level as e.g. Lifestyle in this PR. Was a decision already taken to make this change, or does it need to be discussed with design?

I think you're right @bryophyta , This seems like an unintentional change and for the reasons you list it should be updated to better match how it looks in Frontend.

Great spot!

oliverlloyd:
> Thanks for taking a look. The search button styles were merged in PR #5996. I have spoken with Harry now and changed the styles here #6215

I think @bryophyta was referencing the Edition dropdown, not the search button. To be fair, when I first read the comment and looked at the screenshots I did initially think it was talking about the Search action as well!

bryophyta:
> > Thanks for taking a look. The search button styles were merged in PR #5996. I have spoken with Harry now and changed the styles here #6215
> 
> I think @bryophyta was referencing the Edition dropdown, not the search button. To be fair, when I first read the comment and looked at the screenshots I did initially think it was talking about the Search action as well!

yep that's right! -- Daniel and I discussed on another channel earlier but I forgot to update here 👍

PR #6101, opened 2022-09-30
Add `mask-image property` to make gate transparent

silvacb:
## What does this change?
Adds the `mask-image` property to `shared.tsx` file to make the gate overlay transparent.

## Why?
This is a design-related decision.

### Desktop

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/55602675/193283312-f325a8dc-e6d0-4f15-ad72-bc9d0024311c.png

[after]: https://user-images.githubusercontent.com/55602675/193283425-6c68adfa-fa57-44f1-b90c-8a5e76aafc95.png

### Mobile

| Before      | After      |
|-------------|------------|

[before2]: https://user-images.githubusercontent.com/55602675/193283983-79384716-dda6-485b-a90c-cb674f6140c8.png

[after2]: https://user-images.githubusercontent.com/55602675/193284034-4aa7cb81-b089-42b7-87b5-982272dfb9fe.png

### DIFF
--- a/dotcom-rendering/src/web/components/SignInGate/gateDesigns/shared.tsx
+++ b/dotcom-rendering/src/web/components/SignInGate/gateDesigns/shared.tsx
@@ -136,6 +136,7 @@ export const firstParagraphOverlay = (isComment: boolean) => css`
 		70%,
 		rgba(255, 255, 255, 0)
 	);
+	mask-image: linear-gradient(rgba(0, 0, 0, 0), transparent);
 `;
 
 // This css does 3 things


ioannakok (approved):
I paired on this! Approving but it's good to also double check with someone from Journalism that this is the intended styling. You might have already done this, just noting down this thought I had!

Also, on a second PR it would be good for us to document how to load the sign in gate locally. Happy to pair, do it myself or let you do it. Totally up to you 🙂

marjisound (approved):
Great work 👍

mxdvl:
From the screenshot, I cannot see how the mask-image is doing anything at all.  I think the design was meant to fade out to the background colour, so the `mask-image` would have to be applied to the underlying text, or pick up the very background colour.

I think the rule should have removed the `background-image` and have a linear gradient from `black` to `transparent`.

```css
margin-top: -250px;
width: 100%;
height: 250px;
position: absolute;
mask-image: linear-gradient(black, transparent);
```

PR #6102, opened 2022-10-03
Sign-in gate fixes

mxdvl:
## What does this change?

Applies the CSS introduced in #6101 to the article paragraphs themselves.

## Why?

The gradient mask was lost.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/195044068-0258b0af-e977-402f-ba9b-775c14a9b47c.png
[after]: https://user-images.githubusercontent.com/76776/195044147-4d80c7ad-a829-4a3f-bbc1-62cf226b6dd0.png

### DIFF
--- a/dotcom-rendering/src/web/components/SignInGateSelector.importable.tsx
+++ b/dotcom-rendering/src/web/components/SignInGateSelector.importable.tsx
@@ -1,4 +1,4 @@
-import { ArticleDesign, getCookie } from '@guardian/libs';
+import { getCookie } from '@guardian/libs';
 import { useEffect, useState } from 'react';
 import { constructQuery } from '../../lib/querystring';
 import { useOnce } from '../lib/useOnce';
@@ -91,7 +91,6 @@ const generateSignInUrl = ({
 // fires a VIEW ophan component event
 // and show the gate component if it exists
 const ShowSignInGate = ({
-	format,
 	abTest,
 	setShowGate,
 	signInUrl,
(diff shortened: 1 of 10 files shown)


ioannakok (approved):
LGTM!

dotcom-rendering/src/web/components/SignInGateSelector.importable.tsx
```diff
@@ -121,9 +120,6 @@ const ShowSignInGate = ({
 			},
 			abTest,
 			ophanComponentId: signInGateTestIdToComponentId[abTest.id],
-			isComment:
-				format.design === ArticleDesign.Comment ||
-				format.design === ArticleDesign.Editorial,
```
oliverlloyd, line 126: Elsewhere we take the approach of just prop drilling `format` itself. This tends to be more scalable because once you have format in scope it's a lot easier to use it for branching decisions (which is desirable). Is that an option here?

mxdvl: Not sure I understand… we are removing the need to rely on `format` at all, so why pass it?

oliverlloyd: Oh wow, I completely read that red as green - I thought you were adding the `isComment` prop, not removing it. Sorry, ignore me, nothing to see here

PR #6104, opened 2022-10-03
Lockfile Changes For `typescript-json-schema` `v0.54.0`

JamieB-gu:
## Why?

These seem to have been missed by dependabot in #6081. Currently running `yarn` on `main` introduces these changes, so we should commit them.

## Changes

- Lockfile changes for `typescript-json-schema` `v0.54.0`, generated by running `yarn` on `main`


mxdvl (approved):
You had one job, @dependabot… why? WHY!?

PR #6128, opened 2022-10-05
Add CONTRIBUTING as to make our contributing docs more discoverable

jamesgorrie:
## What does this change?

Creates the CONTRIBUTING document which links through to our existing contribution documents to help surface these in the various places it gets surfaced that [you can see here](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors).

This is more a discovery PR, there is a lot more we could do, this is to just get a start.

## Why?

As we have more and more contributors, it would be good to have a place for people to get an understanding of how to contribute more effectively, and hopefully break down barriers when contributing.

Also to help the platform team evolve these documents s the platform matures.

### DIFF
--- a/dotcom-rendering/docs/contributing/README.md
+++ b/dotcom-rendering/docs/contributing/README.md
@@ -1,0 +1,5 @@
+# Contributing
+
+- [Detailed setup guide](./detailed-setup-guide.md)
+- [Code style](./code-style.md)
+- [Where should my code live?](./where-should-my-code-live.md)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,11 @@
+# Welcome to the the Guardian's Apps & Web rendering platforms
+
+Read our [Code of conduct](./CODE_OF_CONDUCT.md) to help keep things approachable and respectful.
+
+Interested in our web rendering platform? [Start here ➡️](./dotcom-rendering/docs/contributing/README.md)
+
+Interested in our apps rendering platform? [Start here ➡️](./apps-rendering/README.md)
+
+
+
+


bryophyta (approved):
this is a great idea!

OllysCoding (approved):
Looks good to me, but would be great to expand this with more information soon!

bryophyta:
we might want to add `CONTRIBUTING.md` to the `paths-ignore` rules for our GHA workflows, like we do for docs?

https://github.com/guardian/dotcom-rendering/blob/883f3422771854a57361d55ac11a7bfd45493732/.github/workflows/amp-validation.yml#L4-L6

Happy to raise this as a separate PR after this one's been merged though, if that's preferable.

dotcom-rendering/docs/contributing/README.md
```diff
@@ -0,0 +1,5 @@
+# Contributing
+
+- [Detailed setup guide](./detailed-setup-guide.mddetail)
```
JamieB-gu, line 3:
```suggestion
- [Detailed setup guide](./detailed-setup-guide.md)
```

jamesgorrie:
@JamieB-gu I've updated the readme to rather have starting points dependant on your information.

We could have a set of contributions documentation, and then technical jumping off points into the relevant projects?

jamesgorrie:
> we might want to add `CONTRIBUTING.md` to the `paths-ignore` rules for our GHA workflows, like we do for docs?

@bryophyta I wonder if we make that a `**/*.md` instead (as part of another PR)?

apps-rendering/package-lock.json
```diff
@@ -3376,7 +3376,7 @@
         "ms": {
```
JamieB-gu, line 3376: You may want to drop the changes to this file?

JamieB-gu (commented):
> I've updated the readme to rather have starting points dependant on your information.

Looks great, thanks!

bryophyta:
> @bryophyta I wonder if we make that a **/*.md instead (as part of another PR)?

yes, that sounds like a good idea!

georgeblahblah:
I think that this is a great idea! I think there's a good opportunity to write some contributing docs for AR

apps-rendering/package-lock.json
mxdvl: Agreed!

PR #6141, opened 2022-10-06
fix: Remove paths_ignore from DCR chromatic builds

AshCorr:
## What does this change?

Always run Chromatic, even if the only files that have changed are the docs files. Turbosnap should kick in and prevent Chromatic from taking any snapshots if the only files changed are doc files.

## Why?

Currently we require the "UI Checks" status to be reported on a PR before we allow it to be merged, this has issues when the only file the PR touches are doc files as Chromatic will never run and report the "UI Checks" status.

Alternatively we could remove the required UI Checks status from our Github settings.

image

I've seen Github fail to kick-off actions one or two times recently and I'm a little hesitant to remove this requirement.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -3,11 +3,7 @@ name: Chromatic 👓
 on:
   push:
     branches: [main]
-    paths-ignore:
-      - "dotcom-rendering/docs/**"
   pull_request:
-    paths-ignore:
-        - "dotcom-rendering/docs/**"
 
 jobs:
   chromatic:


OllysCoding (approved):
+1 This makes sense to me

jamesgorrie:
I am still confused as to when Chromatic ever returns red - is it when someone rejects a change?

I ask as the workflow we follow seems to be that we check our own work, so they report orange (still merge-able) and rarely, if ever, return red? If this is the case, why do we have them as required?

And if there isn't a reason, this might help inform how we do this.

AshCorr:
Going to merge this down to unblock some of the Docs PR's, I think we still have some benefit to keeping "UI Checks" as required!

PR #6065, opened 2022-09-27
Add `CardPicture` and use in in `Card`

mxdvl:
## What does this change?

Create a new implementation of `Picture` specific for fronts called `FrontPicture`. The decisions for the sizing of a image on fronts and articles is very different, so we choose to start afresh and be open to merging the logic in the future if there is enough overlap.

This is not a perfect implementation as not all possible sizes on all possible breakpoints are accounted for. However, the aim of this change is to provide a lot of value for minimum effort and keep changes small. With the current implementation of Card, there is not quite enough information to know the precise size of an image at every breakpoint. Covering this exhaustively is going to be a significant undertaking that can be decoupled from introducing the picture tag, so we favour this approach for now.

Currently, the image decision logic lives in `decideImageWidths` and only takes in the four existing `imageSizeType`s: `small`, `medium`, `large` and `jumbo`.

We added a new image type, `carousel`, to ensure that Carousel images are alaways 220 pixels at all breakpoints.

## Why?

Fixes #6062 

Alternative to #4611 

## Screenshots

(images are only partly loaded in “before” as they are so large)

| Before      | After      |
|-------------|------------|
| ![front before][] | ![front after][] |
| ![carousel before][] | ![carousel after][] |

[front before]: https://user-images.githubusercontent.com/76776/192526771-932d4303-d9cc-4e2b-be89-c2988496b379.png
[front after]: https://user-images.githubusercontent.com/76776/192526529-ff84fcd3-f7ef-41bc-8db5-506012c6390e.png

[carousel before]: https://user-images.githubusercontent.com/76776/196407918-d2529b4d-d1a3-46af-8fd4-6c336614a580.png
[carousel after]: https://user-images.githubusercontent.com/76776/196407534-643725b1-a1a7-4b98-9bcd-77f1ea893ba8.png

Issue #4460, opened 2022-03-30
Issue ID: DAC_Magnified_Images1 - Feature: Gallery


VDuczekW:
Report link: https://drive.google.com/file/d/1myaQdkLNCNhmrIoe_BPRwOxvpPxvT7w0/view
Full issue list: https://docs.google.com/spreadsheets/d/1vWQtZfD49uq7uNCNELuYN_HcIj-BKiE4t4Hwda3tJaQ/edit#gid=0

AshCorr:
@OllysCoding we had a brief look at this in our team.

It looks like the new <dialog> element might be the way to go, this article summarises it pretty nicely: https://www.scottohara.me/blog/2019/03/05/open-dialog.html

Dialogs make it a lot easier to have accessible modals by trapping focus in the modal and moving focus to the modal when opened.

jamesgorrie:
We're looking into the activation of the dialogue with Voice Over and Voice Control.

jamesgorrie:
Note on ordering of the DOM with styling:

Using styling to “smartly” order DOM on different breakpoints can cause confusion with how a person would navigate using tabindex e.g. a screenreader.

An example is [here](https://www.theguardian.com/science/gallery/2018/mar/14/the-life-of-stephen-hawking-in-pictures) where the `figcaption` is before the "Expand image" button in the DOM.

This makes sense when the design is left to right:
<img width="1421" alt="Screenshot 2022-06-14 at 11 51 42" src="https://user-images.githubusercontent.com/31692/173561628-24157a8d-e21b-4af1-8f12-9443ec9707f8.png">


But not top to bottom>:
<img width="790" alt="Screenshot 2022-06-14 at 11 52 30" src="https://user-images.githubusercontent.com/31692/173561574-a3ddc2c8-4584-4502-8a10-e410ba6c8899.png">

PR #6154, opened 2022-10-08
(Tech Demo / Proposal) Lightbox on DCR 

OllysCoding:
> In a way I like to think this PR is also functioning a tech demo of the maturity of DCR as a platform, showing just how far we've come in the past few years in terms of platform features, devX & code heath, and just how easy it can be to implement  the fundamentals of a new, complex feature.

A WIP project to build lightbox in DCR, taking advantage of a few key platform improvements:

- Islands: This PR enhances islands with `deferUntil="interaction"`, meaning lightbox JS will not load until someone tries to open an image in full screen.
- Unsigned Images: Thanks to our recent changes to fastly-image-service, we're able to create images on the client, simplifying the architecture significantly.

This project is still in early phases, and thought hasn't been put into UI, Accessibility, or improvements to the user flow in general. 

TODO:

- [x] Refactor into two components (Lightbox wrapper, lightbox)
- [x] Use `<dialog>` element for more accessible modals 
- [x] Look at loading states between images
- [ ] Improve UI
- [ ] Look at images with different aspect ratios ([example](https://www.theguardian.com/food/2022/oct/09/chips-in-an-air-fryer-they-are-dull-dry-and-very-sad-as-am-i))
- [ ] Support #img-n in the url?
- [ ] Loading state / animation behind images?
- [ ] & More

Questions:
- On Frontend, when lightbox is opened, all the other images are loaded - do we want this on DCR, or do we want a loading state to conserve bandwidth for us and our users? (Update: DCR does this for the time being, to avoid weirding loading problems between images) 
- Frontend does not use dpr=2, we get this for free in Frontend, is this omitted from Frontend Lightbox deliberately, and if so should we remove it from DCR? 

Problems:
- A potential problem is that if you open lightbox on one image, close it, then open it on another image, a second modal is created on the DOM. This isn't really a massive issue, the code is still only downloaded once, images are cached on the disk etc, but there is a risk it could cause some confusion from a development perspective in the future.

Demo:

https://user-images.githubusercontent.com/9575458/194728031-119d6bea-72e5-479f-92ad-cb7187840ead.mov

(Note: ~~in this demo, all the images were already cached on my disk - without this, the loading between images on is still quite buggy!~~ This was fixed by including all the images in the dom at once, but we might wanna investigate this further)


.vscode/settings.json
```diff
@@ -1,28 +1,25 @@
...
+    "titleBar.activeForeground": "#000000"
+  },
+  "jest.jestCommandLine": "yarn workspaces run test",
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "gitlens.advanced.blame.customArguments": [
+    "--ignore-revs-file",
+    ".git-blame-ignore-revs"
+  ],
+  "git.branchProtection": ["main"],
+  "cSpell.words": ["Lightbox"]
```
OllysCoding, line 24: We recommend a spell checker, so I had to add this as it's not recognised as a word

dotcom-rendering/src/web/browser/islands/doHydration.ts
```diff
@@ -17,7 +16,7 @@ import { initPerf } from '../initPerf';
  */
 export const doHydration = (
 	name: string,
-	data: Attributes | null,
```
OllysCoding, line 20: I'm not sure how we get away with `Attributes` here - but it's not actually accurate to the types

dotcom-rendering/src/web/browser/islands/getIslandsByName.ts
```diff
@@ -0,0 +1,8 @@
+export const getIslandsByName = (name: string): HTMLElement[] => {
```
OllysCoding, line 1: I added this under the `web/browser/islands` code because it adds a link in our codebase to gu-islands exist, and putting it here makes things easier if we migrate away from islands one day.

dotcom-rendering/src/web/components/Lightbox.importable.tsx
```diff
@@ -0,0 +1,125 @@
...
+	const canNext = () => index < elements.length - 1;
+
+	const { master, alt, caption } =
+		index !== -1 ? elements[index] : initialElementProps;
+
+	return (
+		<>
+			<button type="button" onClick={() => setOpen(true)}>
+				Open Lightbox
+			</button>
```
OllysCoding, line 41: I'm interested if there's a way to make `deferUnti=interaction` more intuitive, but I haven't had the time to look into it yet

dotcom-rendering/src/web/components/LightboxPicture.tsx
```diff
@@ -0,0 +1,122 @@
...
+		{
+			breakpoint: breakpoints.mobileLandscape,
+			width: breakpoints.mobileLandscape,
+		},
+		{
+			breakpoint: breakpoints.mobileMedium,
+			width: breakpoints.mobileMedium,
+		},
+		{ breakpoint: breakpoints.mobile, width: breakpoints.mobile },
+	];
```
OllysCoding, line 31: We should update this to multiple the width by 0.85 for image width

dotcom-rendering/src/web/components/LightboxPicture.tsx
```diff
@@ -0,0 +1,122 @@
...
+	padding-top: ${(3 / 5) * 100}%;
+	position: relative;
+	& > * {
+		position: absolute;
+		top: 0;
+		left: 0;
+		width: 100%;
+		height: 100%;
+	}
+`;
```
OllysCoding, line 55: TODO: This was copied from https://github.com/guardian/dotcom-rendering/pull/6065 and I should review if its what we want here.

dotcom-rendering/src/web/components/Lightbox.importable.tsx
```diff
@@ -0,0 +1,125 @@
...
+			(element) => element.master === initialElementProps.master,
+		),
+	);
+	const [isOpen, setOpen] = useState(true);
+
+	const canPrev = () => index > 0;
+	const canNext = () => index < elements.length - 1;
+
+	const { master, alt, caption } =
+		index !== -1 ? elements[index] : initialElementProps;
```
OllysCoding, line 35: Don't love this but it works for now - we shouldn't need this data for the server render, which is when index is negative 1

OllysCoding (commented):
My thoughts so far on my work:

paperboyo:
> couldn't figure out how to change it

image

Yay! Finally. The whole of Picture Desk eagerly awaits this missing functionality! Some logic is captured here:
https://trello.com/c/08teLVKQ/311-image-lightbox

We should also look at fixing the [regression in user experience](https://github.com/guardian/frontend/pull/25114) introduced ostensibly because of [accessibility concerns](https://github.com/guardian/dotcom-rendering/issues/4460). While we want everything accessible to all, it should not come at the expense of reduced ease of use: the whole of the image should be clickable. Ideally, icon should appear only on hover too, signifying available action, but not covering image area. Not sure how to square it with [the report](https://drive.google.com/file/d/1myaQdkLNCNhmrIoe_BPRwOxvpPxvT7w0/view), but recommendations there mean less useful and worse-looking Lightbox, used by fewer readers (this is a hypothesis as I’m not aware of any metrics captured to asses the introduction of that change: a problem in itself). I hope fully accessible solution can be found that doesn’t reduce the functionality, discoverability and ease of use of the feature.

dotcom-rendering/src/web/components/Lightbox.tsx
```diff
@@ -0,0 +1,135 @@
...
+						padding: ${space[2]}px;
+					`}
+				>
+					<div
+						css={css`
+							margin-bottom: ${space[2]}px;
+							color: white;
+						`}
+					>
+						<span>
```
oliverlloyd, line 102: Is this span needed?

dotcom-rendering/src/web/components/Lightbox.tsx
```diff
@@ -0,0 +1,135 @@
...
+							type="button"
+							disabled={!canPrev()}
+							onClick={() => canPrev() && setIndex(index - 1)}
+						>
+							Prev
+						</button>
+						<button
+							type="button"
+							disabled={!canNext()}
+							onClick={() => canNext() && setIndex(index + 1)}
```
oliverlloyd, line 115: Is this `canNext` check needed if we're disabling the button when it is false?

dotcom-rendering/src/web/components/Lightbox.tsx
```diff
@@ -0,0 +1,135 @@
...
+					<div
+						css={css`
+							margin-bottom: ${space[2]}px;
+							color: white;
+						`}
+					>
+						<span>
+							{index + 1}/{elements.length}
+						</span>
+						<button
```
oliverlloyd, line 105: Just a thought, is there benefit to using a source component here? I tend to lean towards using them even when it means a lot of style overrides but I'm open to questioning that approach. I know you mentioned there was still UI work needed here though so maybe this was always intended

dotcom-rendering/src/web/components/Lightbox.tsx
```diff
@@ -0,0 +1,135 @@
...
+				})}
+				<div
+					css={css`
+						flex-basis: 15%;
+						align-items: flex-end;
+						display: flex;
+						flex-direction: column;
+						padding: ${space[2]}px;
+					`}
+				>
```
oliverlloyd, line 95: I presume you can't merge these two divs because of the `figcaption`?

dotcom-rendering/src/web/components/LightboxPicture.tsx
```diff
@@ -0,0 +1,122 @@
+import { css } from '@emotion/react';
+import { breakpoints } from '@guardian/source-foundations';
+import React from 'react';
+import type { ImageWidthType } from './Picture';
```
oliverlloyd, line 4: Maybe we could have a `picture` file for these sorts of shared types? That seems to be the pattern

dotcom-rendering/src/web/components/LightboxPicture.tsx
```diff
@@ -0,0 +1,122 @@
...
+			breakpoint: breakpoints.mobileMedium,
+			width: breakpoints.mobileMedium,
+		},
+		{ breakpoint: breakpoints.mobile, width: breakpoints.mobile },
+	];
+};
+
+/**
+ * Used on `picture` and `img` to prevent having a line-height,
+ * as these elements are which are `inline` by default.
```
oliverlloyd, line 36:
```suggestion
 * Used on `picture` and `img` to prevent having a line-height,
 * as these elements are `inline` by default.
```

dotcom-rendering/src/web/components/LightboxPicture.tsx
```diff
@@ -0,0 +1,122 @@
...
+				// height={fallbackSource.hei / ratio}
+				css={block}
+			/>
+		</picture>
+	);
+};
+
+// We use disableLazyLoading to decide if we want to turn off lazy loading of images site wide. We use this
+// to prevent false negatives on Chromatic snapshots (see /.storybook/config)
+LightboxPicture.disableLazyLoading = false;
```
oliverlloyd, line 122: If we're not using the `lazy` prop on the `img` then we don't need this

dotcom-rendering/src/web/lib/getZIndex.tsx
```diff
@@ -23,6 +23,7 @@
  */
 const indices = [
 	// Modals will go here at the top
+	'lightbox',
```
oliverlloyd, line 26: 👍 👍

dotcom-rendering/src/web/components/LightboxWrapper.importable.tsx
```diff
@@ -0,0 +1,41 @@
...
+	const elements = getIslandsByName('LightboxWrapper');
+	return elements.map((element) => getProps(element) as Props);
+};
+
+export const LightboxWrapper = (initialElementProps: Props) => {
+	const [isOpen, setOpen] = useState(true);
+	const [elements, setElements] = useState<Props[]>([]);
+	useEffect(() => {
+		setElements(getLightboxElements());
+	}, []);
```
oliverlloyd, line 23: This is an interesting mix of plain js and react. Are there any edge cases to this approach that might cause issues?

oliverlloyd (commented):
It's about time DCR had a lightbox

dotcom-rendering/src/web/components/ImageComponent.tsx
```diff
@@ -301,6 +303,13 @@ export const ImageComponent = ({
 					}
 				`}
 			>
+				<Island deferUntil="interaction">
+					<LightboxWrapper
```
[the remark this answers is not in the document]
OllysCoding: Fair question, perhaps it's a `container`? Interestingly we use the 'wrapper' name on atoms we 'wrap' in .importable components for making them into islands

dotcom-rendering/src/web/components/Island.tsx
```diff
@@ -20,13 +20,21 @@ interface ClientOnlyProps {
 	expediteLoading?: boolean;
 }
 
+interface InteractionProps {
+	deferUntil: 'interaction';
+	clientOnly?: false;
+	placeholderHeight?: never;
+	children: JSX.Element;
+	expediteLoading?: false;
+}
```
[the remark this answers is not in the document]
OllysCoding: No - We always need a server-rendered version because (as it works now) the client must expose some kind of UI for the user to interact with, which would trigger hydration!

dotcom-rendering/src/web/components/LightboxPicture.tsx
OllysCoding: I think I want to look into whether the the lazy prop would be useful for us - the mobile experience isn't an area I've looked at yet but it might change things up a bit!

oliverlloyd: `loading=lazy` on an image defers the download until the browser [thinks the image is needed](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading) (based on if it is in the viewport or not)

> Tells the user agent to hold off on loading the image until the browser estimates that it will be needed imminently. For instance, if the user is scrolling through the document, a value of lazy will cause the image to only be loaded shortly before it will appear in the window's [visual viewport](https://developer.mozilla.org/en-US/docs/Glossary/Visual_Viewport).

This isn't 100% clear but I think this means that if an `img` is off page, as, a `dialog` is not visible, then the image will not be loaded. 

There's also the question of these images affecting the page load event

> The [load](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event) event is fired when the document has been fully processed. When images are loaded eagerly (which is the default), every image in the document must be fetched before the load event can fire.
> 
> By specifying the value lazy for loading, you prevent the image from delaying the load attribute by the amount of time it takes to request, fetch, and process the image.

None of this conclusively makes the case for using `loading-lazy` but my feeling is we probably want to at least test and investigate using it

PR #6164, opened 2022-10-11
fix: Dependabot PR's should run with full chromatic tests

AshCorr:
## What does this change?

Add an extra condition to force Dependabot PR's to run a full chromatic build if the PR is opened by dependabot.

## Why?

It seems like the dependency labels are being added to the PR after dependabot has already opened it and the Chromatic github action has already started running.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Chromatic - Non Dependency PR / main
         uses: chromaui/action@v1
-        if: ${{ github.event_name == 'push' || !contains(github.event.pull_request.labels.*.name, 'dependencies') }}
+        if: ${{ github.event_name == 'push' || !contains(github.event.pull_request.labels.*.name, 'dependencies') || github.event.pull_request.user.login == 'dependabot'}}
         with:
           projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


AshCorr:
> I may be missing something, but are "Dependency PRs" meant to be the ones opened by `dependabot`? If so, should these conditions be the other way around? Do you instead want the user to be `dependabot` for "Dependency PR"s and not for "Non Dependency PR"s?

You are absolutely right! It should be the other way around!

.github/workflows/chromatic.yml
```diff
@@ -43,7 +43,7 @@ jobs:
 
       - name: Chromatic - Dependency PR
         uses: chromaui/action@v1
-        if: ${{ github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'dependencies') }}
+        if: ${{ github.event_name == 'pull_request' && (contains(github.event.pull_request.labels.*.name, 'dependencies') || github.event.pull_request.user.login == 'dependabot') }}
```
oliverlloyd, line 46: Is it possible that a PR could be opened by a different user login than 'dependabot' and be given the label `dependencies`?

If not then can we simplify this to:

```suggestion
        if: ${{ github.event_name == 'pull_request' && github.event.pull_request.user.login == 'dependabot' }}
```

?

Same for above

AshCorr: I was thinking that if you kept both conditions then people could still run a full chromatic test on manual dependency updates by adding the Dependency label

bryophyta (approved):
I remember that there was some discussion about what the intended/desired behaviour is for chromatic and turbosnap last week, in connection the Source update. I'm not sure whether that affects the strategy implemented in this PR or not. But the code itself in this  branch looks good to me 👍

PR #6166, opened 2022-10-12
If no keyEvents or availableTopics, don't show the filters or the filter heading in FilterTopicBank

sophie-macmillan:
## What does this change?
This change updates the check on availableTopics, to ensure that the array has some items in it. If not, we don't show the FilterTopic bank.

## Why?
UX Improvement - We've seen a case when availableTopics exists, but it empty, which meant the Filter headings were visible with no Filter buttons.

## Screenshots

| Before      | After      |
|-------------|------------|
| image| image|

### DIFF
--- a/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
+++ b/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
@@ -83,6 +83,9 @@ export const TopicFilterBank = ({
 	filterKeyEvents = false,
 	id,
 }: Props) => {
+	const showFilters = keyEvents?.length || availableTopics.length;
+	if (!showFilters) return null;
+
 	const palette = decidePalette(format);
 	const selectedTopic = selectedTopics?.[0];
 	const topFiveTopics = availableTopics
--- a/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
+++ b/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
@@ -83,9 +83,6 @@ export const TopicFilterBank = ({
 	filterKeyEvents = false,
 	id,
 }: Props) => {
-	const showFilters = keyEvents?.length || availableTopics.length;
-	if (!showFilters) return null;
-
 	const palette = decidePalette(format);
 	const selectedTopic = selectedTopics?.[0];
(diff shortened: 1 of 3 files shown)


dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
```diff
@@ -83,6 +83,9 @@ export const TopicFilterBank = ({
 	filterKeyEvents = false,
 	id,
 }: Props) => {
+	const showFilters = keyEvents?.length && availableTopics.length;
```
oliverlloyd, line 86: My brain melts in these situations so there's a non trivial chance I'm talking rubbish but should this be OR here or?

```suggestion
	const showFilters = keyEvents?.length || availableTopics.length;
```

sophie-macmillan: Oops yeah, you're absolutely right!

mxdvl (approved):

oliverlloyd (approved):
👍

sophie-macmillan:
I originally avoided this as the `TopicFilterBank` is used in both `LiveLayout.tsx` and `LiveBlogRenderer.tsx`, to reduce code duplication. If reducing javascript load is the priority, I could just duplicate this logic in both files?

oliverlloyd:
Reducing javascript load

i-need-that-so-bad-marissa-rachel

OllysCoding (approved):
+1 Looks good!!

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -750,7 +751,8 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 								</Hide>
 
 								{showTopicFilterBank &&
-									CAPIArticle.availableTopics && (
+									CAPIArticle.availableTopics &&
```
bryophyta, line 754: if I'm understanding correctly then is looks like this set of 3 conditions is equivalent to `!showToggle`. If that's the intention, there might be scope to rename `showToggle` and then use it here as well as where it's used above.

Just a suggestion if I've understood correctly, and if it seems like it'll help to make the logic easier to follow. Not a blocking suggestion for this PR  though!

PR #6167, opened 2022-10-12
Add `AssetHash` type guard

mxdvl:
## What does this change?

Add a AssetHash type-guard.

## Why?

Left as a todo in #5598


bryophyta (approved):
lgtm!

PR #6170, opened 2022-10-12
Remove floating promises linting rule

mxdvl:
## What does this change?

Use the `void` keyword to indicate promises are not awaited.

## Why?

It was not allowed to use this `void` keyword previously so we had to use disable the ESLint rule in #4513.

### DIFF
--- a/dotcom-rendering/src/web/components/CoreVitals.importable.tsx
+++ b/dotcom-rendering/src/web/components/CoreVitals.importable.tsx
@@ -41,9 +41,7 @@ export const CoreVitals = () => {
 			Object.keys(window.guardian.config.tests).includes(test),
 		);
 
-	/* eslint-disable @typescript-eslint/no-floating-promises -- they’re async methods */
-
-	initCoreWebVitals({
+	void initCoreWebVitals({
 		browserId,
 		pageViewId,
 		isDev,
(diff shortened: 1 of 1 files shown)


oliverlloyd:
This seems like a perfectly good change to make but just for clarity I wanted to check something. We added the lint exceptions in the [older PR](https://github.com/guardian/dotcom-rendering/pull/4513) you linked to and here we're removing them, but what happened to mean that they can now be removed?

mxdvl:
@oliverlloyd the [`.eslintrc` has changed enormously](https://github.com/guardian/dotcom-rendering/compare/fe800dd761e263ac46fe86cba3607181ae57353b..7476195dde3a75b5526d5bb257e6cf7d3ba988df#diff-81b4e412f2d2c3b3d9c0c950617624cc658ad191981fad94a4587560eec9af45) between these times.

I’m not entirely sure why it was prevented then, but there was a clash between the `no-floating-promises` and `no-void` rules, which is [a known issue](https://typescript-eslint.io/rules/no-floating-promises#ignorevoid). Since we switched to the [@guardian/eslint-config-typescript](https://github.com/guardian/csnx/blob/main/libs/%40guardian/eslint-config-typescript/index.js) the clash seem to have been resolved.

oliverlloyd:
> Since we switched to the [@guardian/eslint-config-typescript](https://github.com/guardian/csnx/blob/main/libs/%40guardian/eslint-config-typescript/index.js) the clash seem to have been resolved.

Ah, it was this - gotcha 👍

oliverlloyd (approved):

PR #6173, opened 2022-10-12
Illustrate private vpcs

nicl:
**Note, the first commit is a formatting fix. So see the second for the actual changes here.**

Move DCR instances to private subnets following their creation (See https://github.com/guardian/platform/pull/1497.) This is a security improvement as it ensures instances aren't directly addressable over the internet. It is also a prerequisite for migration to our EC2 CDK patterns.

A similar change can and should be made to other Dotcom-owned apps - for example, Frontend services.

Deployment-wise, existing instances are preserved when the `VPCZoneIdentifier` is updated, so we can deploy this safely.

### DIFF
--- a/dotcom-rendering/cloudformation.yml
+++ b/dotcom-rendering/cloudformation.yml
@@ -1,9 +1,14 @@
 AWSTemplateFormatVersion: '2010-09-09'
 Description: Frontend rendering service
 Parameters:
-  Subnets:
+  PublicSubnets:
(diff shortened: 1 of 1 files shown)


mxdvl (approved):
Looks good to me, looking specifically at 7f1c3e5206dcfc9653a42e7421c6de21301b111e

ioannakok (approved):
+1!

mxdvl:
We could merge this tomorrow!

PR #6179, opened 2022-10-12
Fix Sentry tag

mxdvl:
## What does this change?

Fix the Sentry tag used for the dcr-javascript-bundle test

## Why?

Forgot to update it in #6168


bryophyta (approved):
This looks good to me 👍

PR #6180, opened 2022-10-12
Shady pie test updates

emma-imber:
## Why?

- Extending the duration of the shady pie clickthrough rate test
- Updating the shady pie component logic so that the labs shady pie only renders when a reader is in the AB test
- Updates the shady pie story to temporarily comment out the test variants to prevent errors


mxdvl (approved):
Not a blocker, but I think I would favour a clean commit that removes the story that can be reverted in the future over commenting out parts of the code.

PR #6184, opened 2022-10-13
Update node version manager recommendations

bryophyta:
## What does this change?
- Update recommendations from `nvm` to `fnm`.
- Add section on installing Yarn to the detailed setup guide.
- Re-run `createtoc`. (nb. this seems to have picked up some prior changes which hadn't been reflected in the TOC until now.)
- Update the checks and hints in `check-node.js`.

## Why?
My understanding from recent conversations in the department is that `nvm` is no longer the recommended way of managing Node versions. I believe the official department-wide recommendation is currently to use `asdf`. But people have also reported difficulties with using it, and if a new contributor is in the position of needing to install a Node version manager then it seems like the extra difficulty of `asdf` isn't worth it, given that its benefits seem to be seen mainly when managing other kinds of environments.

I'm very happy for the `fnm` recommendation to be challenged, but thought that this PR could be seen as a 'soft RFC' for the recommendation itself. (Suggested changes to the wording are also welcome, of course!)

I added the Yarn docs because it wasn't listed before  but it seems like a necessary step for running the project, unless I've missed something?

### DIFF
--- a/dotcom-rendering/README.md
+++ b/dotcom-rendering/README.md
@@ -6,25 +6,24 @@ Frontend rendering framework for theguardian.com. It uses [React](https://reactj
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 <!-- Automatically created with yarn run createtoc and on push hook -->
 
-- [Dotcom Rendering](#dotcom-rendering)
-	- [Quick start](#quick-start)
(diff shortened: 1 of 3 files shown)


bryophyta:
@OllysCoding On your suggestion I checked for other references to nvm in the scripts. I've updated the `check-env` script and haven't found any others, but if you know of any then lmk?

mxdvl (approved):
Nice!

PR #6185, opened 2022-10-13
Should we ask people to raise an RFC prior to making changes to the platform?

oliverlloyd:
## What?
This PR proposes asking teams to raise an RFC issue before making changes to Dotcom Rendering.

## Why?
To make the developer experience when making changes to DCR better.

We've noticed that teams have had a more positive experience with making changes when a member of Dotcom is involved early on and that, conversely, there have been examples where late involvement from the team has led to long PRs with lots of comments and requests for changes.

## Isn't this gate-keeping 😡 🚫 ?
No. We absolutely do not want this RFC process to be seen as gate-keeping. We want to be very explicit about the fact we see it as our responsibility to keep these RFCs moving through to successful completion, not the teams raising them.

We want this process to be seen as a positive, helpful experience, enabling teams to reach their goals sooner, with less friction.

## What about Apps Rendering?
This process is, initially, only for DCR and the naming convention signifies this. It could be that we extend it later to encompass both platforms.

### DIFF
--- a/.github/ISSUE_TEMPLATE/request-for-change.md
+++ b/.github/ISSUE_TEMPLATE/request-for-change.md
@@ -1,5 +1,5 @@
 ---
-name: DCR Request for change
+name: DCR Request for comments
 about: Tell us about a feature that you're planning to implement in dotcom rendering
 title: "[RFC]"
 labels: RFC, dotcom-rendering
--- a/.github/ISSUE_TEMPLATE/request-for-change.md
+++ b/.github/ISSUE_TEMPLATE/request-for-change.md
@@ -10,4 +10,4 @@ assignees: ''
 The Dotcom team would like to help you with your proposed change. Please provide us with a brief and concise explanation for what you're looking to do we will try to make the process easier with early feedback.
 
 **Who is responsible for completing this RFC process?**
-Once the RFC is raised, it is the responsibility of the _Dotcom team_ to complete this process
+Once the RFC is raised, it is the responsibility of the _Dotcom team_ to provide feedback and help you get your change to the finish line.
--- a/.github/ISSUE_TEMPLATE/request-for-change.md
+++ b/.github/ISSUE_TEMPLATE/request-for-change.md
@@ -15,4 +15,4 @@ This is an optional process. If you own the code you're working on or feel confi
 However, if your change affects core platform code or is introducing a new concept or pattern, or if you're just looking for more eyes and input then this process is for you.
 
 **Who is responsible for completing this RFC process?**
-Once the RFC is raised, it is the responsibility of the _Dotcom team_ to provide feedback and help you get your change to the finish line.
+Once the RFC is raised, it is the responsibility of the _Dotcom team_ (@guardian/dotcom-platform) to provide feedback and help you get your change to the finish line.
--- a/.github/ISSUE_TEMPLATE/request-for-change.md
+++ b/.github/ISSUE_TEMPLATE/request-for-change.md
@@ -12,7 +12,7 @@ The Dotcom team would like to help you with your proposed change. Please provide
 **When should this RFC be used?**
 This is an optional process. If you own the code you're working on or feel confident about the change you're making then you should simply raise a PR as normal. We want developers and teams to feel empowered to make the choice about when to reach out.
 
-But if your change affects core platform code or is introducing a new concept or pattern, or if you're just looking for more eyes and input then this process is for you.
+However, if your change affects core platform code or is introducing a new concept or pattern, or if you're just looking for more eyes and input then this process is for you.
 
 **Who is responsible for completing this RFC process?**
 Once the RFC is raised, it is the responsibility of the _Dotcom team_ to provide feedback and help you get your change to the finish line.
--- a/.github/ISSUE_TEMPLATE/request-for-change.md
+++ b/.github/ISSUE_TEMPLATE/request-for-change.md
@@ -1,8 +1,8 @@
 ---
-name: Request for change
+name: DCR Request for change
 about: Tell us about a feature that you're working on
 title: "[RFC]"
-labels: RFC
+labels: RFC, dotcom-rendering
 assignees: ''
 
 ---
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/request-for-change.md
```diff
@@ -0,0 +1,13 @@
...
+title: "[RFC]"
+labels: RFC
+assignees: ''
+
+---
+
+The Dotcom team would like to help you with your proposed change. Please provide us with a brief and concise explanation for what you're looking to do we will try to make the process easier with early feedback.
+
+**Who is responsible for completing this RFC process?**
+Once the RFC is raised, it is the responsibility of the _Dotcom team_ to complete this process
```
AshCorr, line 13:
```suggestion
Once the RFC is raised, it is the responsibility of the _Dotcom team_ to provide feedback and help you get your change to the finish line.
```

Maybe? I feel like it currently implies that we're going to be the ones executing the RFC!

georgeblahblah:
what do you think about scoping the template to specify DCR in the title, label and name? I think Apps Rendering will look to implement something similar as part of our focus on AR as a platform, but having separate templates might make filtering issues simpler whilst reducing confusion.

.github/ISSUE_TEMPLATE/request-for-change.md
```diff
@@ -0,0 +1,13 @@
+---
+name: DCR Request for change
+about: Tell us about a feature that you're working on
+title: "[RFC]"
+labels: RFC, dotcom-rendering
+assignees: ''
+
+---
+
+The Dotcom team would like to help you with your proposed change. Please provide us with a brief and concise explanation for what you're looking to do we will try to make the process easier with early feedback.
```
kenoir, line 10: Is it worth indicating the type of change an RFC is necessary for? In principal this seems like a great idea, but it would be good to lay down the scale of change this applies for. 

There are also other teams that are part responsible for some code in dotcom-rendering (thinking commercial / T&C etc.) Would BAU changes to code those teams are concerned with require an RFC, i'd guess not but worth laying it out?

.github/ISSUE_TEMPLATE/request-for-change.md
```diff
@@ -0,0 +1,13 @@
...
+
+**Who is responsible for completing this RFC process?**
```
kenoir, line 12: Will this change be enforced by repository rules / codeowners configuration, or will we rely on Dotcom team members commenting on PRs and nudging them towards the RFC process?

.github/ISSUE_TEMPLATE/request-for-change.md
oliverlloyd: This is a good point. I don't think anyone wants to be asked to jump through hoops when making changes to their 'own' code so we need to be clear about when this might or might now apply.

I think in reality what we will end up on is asking people to use their own best judgement. We're a clever bunch here at the Guardian and, more than that, we're very good at working autonomously and making the right choices and so maybe am optional approach would work. But equally, there are benefits to automation taking over and reducing the amount of thinking engineers need to do. What do you think?

kenoir: > I think in reality what we will end up on is asking people to use their own best judgement.

This makes sense, but for new folk it's probably worth being clear (perhaps with examples) in the template about when an RFC would be appreciated.

oliverlloyd: 👍  I've added more information to the description to make it clearer about when this RFC should apply

.github/ISSUE_TEMPLATE/request-for-change.md
oliverlloyd: Initially, we should keep the flow optional and only start to enforce things as we learn more about usage

shtukas (approved):

.github/ISSUE_TEMPLATE/request-for-change.md
```diff
@@ -0,0 +1,18 @@
...
+assignees: ''
+
+---
+
+The Dotcom team would like to help you with your proposed change. Please provide us with a brief and concise explanation for what you're looking to do we will try to make the process easier with early feedback or advice.
+
+**When should this RFC be used?**
+This is an optional process. If you own the code you're working on or feel confident about the change you're making then you should simply raise a PR as normal. We want developers and teams to feel empowered to make the choice about when to reach out.
+
+But if your change affects core platform code or is introducing a new concept or pattern, or if you're just looking for more eyes and input then this process is for you.
```
mxdvl, line 15: I love a “but”, but wonder if this reads nicer?


```suggestion
However, if your change affects core platform code or is introducing a new concept or pattern, or if you're just looking for more eyes and input then this process is for you.
```

.github/ISSUE_TEMPLATE/request-for-change.md
```diff
@@ -0,0 +1,18 @@
...
+
+**Who is responsible for completing this RFC process?**
+Once the RFC is raised, it is the responsibility of the _Dotcom team_ to provide feedback and help you get your change to the finish line.
```
mxdvl, line 18:
```suggestion
Once the RFC is raised, it is the responsibility of the _Dotcom team_ (@guardian/dotcom-platform) to provide feedback and help you get your change to the finish line.
```

mxdvl (approved):
This seems like a good improvement for long-term cross-team collaborations :)

.github/ISSUE_TEMPLATE/request-for-change.md
oliverlloyd: Yep

.github/ISSUE_TEMPLATE/request-for-change.md
oliverlloyd: Better

jamesgorrie (approved):
Looks great.

I wonder if we could give some recomendations on an RFCs, similar to how we do in  [ADRs](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/docs/architecture/000-architecture-decision-records.md), or [pull requests](https://github.com/guardian/recommendations/blob/main/pull-requests.md) - but maybe this will come from running this process for a bit.

PR #6186, opened 2022-10-13
Move isDev and enableSentryReporting checks

bryophyta:
A 'beforeSend' hook was being used inside the Sentry options to check whether `isDev` or `enableSentryReporting` were true.

These values are available prior to Sentry being loaded at all, so this moves the check earlier in the process. (There doesn't seem much point in loading the code if errors are never going to be sent.)

**Disclaimer**: This was a pass-by refactor. I don't fully understand the Sentry setup so I might be missing something about how it works. However, I thought that raising the PR would provide an easier way to discuss this. @oliverlloyd I think you worked on the lazy loading code previously -- do you know if I'm missing something?

### DIFF
--- a/dotcom-rendering/src/web/browser/sentryLoader/init.ts
+++ b/dotcom-rendering/src/web/browser/sentryLoader/init.ts
@@ -3,6 +3,12 @@ import { isAdBlockInUse } from '@guardian/commercial-core';
 import { startup } from '../startup';
 
 const init = async (): Promise<void> => {
+	// We don't send errors on the dev server, or if the enableSentryReporting switch is off.
+	const {
+		switches: { enableSentryReporting },
+		isDev,
+	} = window.guardian.config;
+	const dontSend = isDev || !enableSentryReporting;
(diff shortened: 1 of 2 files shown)


dotcom-rendering/src/web/browser/sentryLoader/init.ts
```diff
@@ -3,14 +3,20 @@ import { isAdBlockInUse } from '@guardian/commercial-core';
 import { startup } from '../startup';
 
 const init = async (): Promise<void> => {
+	// We don't send errors on the dev server, or if the enableSentryReporting switch is off.
+	const {
+		switches: { enableSentryReporting },
+		isDev,
+	} = window.guardian.config;
+	const dontSend = isDev || !enableSentryReporting;
```
oliverlloyd, line 11:
```suggestion
	const sentryDisabled = isDev || !enableSentryReporting;
```

What do you think of this?

oliverlloyd (approved):
Thanks for spotting this and making this change, great stuff!

PR #6024, opened 2022-09-20
Adds the `FixedMediumFastXI` container

oliverlloyd:
## What does this change?
This PR closes #5140 by adding support for the `FixedMediumFastXI` container

As part of introducing this container I extended the pattern of having slices abstracted in their own file. This abstraction felt right here and seems to be scaling well.

Screenshot 2022-09-20 at 14 36 04

PR #5651, opened 2022-08-10
`FixedMediumSlowXIIMPU`

oliverlloyd:
## What does this change?
This closes #5141 by adding support for the `FixedMediumSlowXIIMPU` container.

This is a fixed container so in theory we would only expect a fixed number of trails passed in but looking at how Frontend and the fronts tool currently works there is support for rendering this container with fewer cards. So this PR aligns with that, supporting all possible counts from 1-9. After nine, the additional cards overflow to backfill.

PR #6187, opened 2022-10-13
Adds `FixedMediumFastXI` container

### LINKED ISSUE #5140: "fixed/medium/fast-XI" (`94`) container

Please update [FaciaPicker in Frontend](https://github.com/guardian/frontend/blob/main/facia/app/services/dotcomrendering/FaciaPicker.scala) once this container is supported by DCR

oliverlloyd:
## What does this change?
This PR closes #5140 by adding support for the `FixedMediumFastXI` container. This PR replaces #6024 which got stymied with merge conflicts after I foolishly left it open and went on holiday 😄 

As part of introducing this container I extended the pattern of having slices abstracted in their own file. This abstraction felt right here and seems to be scaling well.

Although this container is fixed and designed to accept 11 trails, it is possible that it's used with fewer than this number and so each of these possibilities is handled gracefully here.

Screenshot 2022-10-13 at 15 00 30


dotcom-rendering/src/web/components/FixedMediumFastXI.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+import { trails } from '../../../fixtures/manual/trails';
+import { FixedMediumFastXI } from './FixedMediumFastXI';
+import { Section } from './Section';
+
+export default {
+	component: FixedMediumFastXI,
+	title: 'Components/FixedMediumFastXI',
+	parameters: {
+		chromatic: {
+			viewports: [
```
AshCorr, line 11: Is this many breakpoints intentional?

oliverlloyd: No, not intentional at all! This slipped in from a me copying old code, thanks for spotting

mxdvl (commented):
I find the pattern of handling trails very confusing. Is this container meant to have eleven (XI) elements? Why handle a different count? I believe CP has said that they would occasionally be short, but this forces us to support 10 cases that should never happen.

I would much rather we either:
- failed to render this code
  - catastrophically by throwing an error
  - softly by rendering nothing
- padded the array of trails with bogus articles if there are less than 11

This is a discussion that spans a bit wider than this individual component, but I think we should settle on a consistent approach for trails of unexpected length.

dotcom-rendering/src/web/components/FixedMediumFastXI.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { UL } from './Card/components/UL';
+import { FrontCard } from './FrontCard';
+
+type Props = {
+	trails: TrailType[];
+	containerPalette?: DCRContainerPalette;
+	showAge?: boolean;
+};
+
+const decideOffset = ({
```
ioannakok, line 18: Non-blocking naming suggestion: Since this returns a boolean would it make more sense to name this something along the lines of `hasOffset` or `shouldHaveOffset`?

ioannakok (approved):
This looks good to me! I only have a naming non-blocking suggestion

oliverlloyd:
> This is a discussion that spans a bit wider than this individual component, but I think we should settle on a consistent approach for trails of unexpected length.

This was discussed with CP & DigiEds (as part of [this PR](https://github.com/guardian/dotcom-rendering/pull/5651))and the agreement made was that DCR would, broadly, do the same as Frontend, which is to gracefully support fewer than the expected number of trails, but that DCR would not match what Frontend does exactly. This means we _should_ render fewer than 11 trails in a way that looks correct and isn't broken but that we do not need to make the additional effort it would take to have exact parity with Frontend.

If we wanted to change this, and remove support for fewer than the fixed target number, then I think that change has to first happen in the Fronts Tool. But until then, so long as the tool allows editors to create containers with fewer than n trails, the rendering platforms have to support this.

PR #6190, opened 2022-10-13
update footer urls for UK/INTL and AUS

dblatcher:
## What does this change?
Updates the hard coded relative URLs used on the 'Sign-up for our email' buttons  (for UK/INTL and Aus)in the `Footer` component. The old pages are currently redirected to the new ones.

## Why?
We've updated the sign-up pages for the newsletters and they have new URL. Updating the buttons so they work without redirecting the user.

equivalent change for frontend: https://github.com/guardian/frontend/pull/25588


oliverlloyd (approved):
I tested each url, looks good 👍

PR #6205, opened 2022-10-17
Break up metadata

iainjchambers-guardian:
## What does this change?
This is a refactor of the metadata component to bring it inline with the per-design component with a single index file based controller component. 

## Notes
This is a **WIP** PR. I wanted to have a discussion about how far to go with this one, the reason being that currently the Metadata is  broken up into three components, a _simple_ version, an _extended_ version that includes a byline and a much more complex version that was only for live/dead blogs. I have combined the first two into one using some simple conditional rendering logic and separated out the blog version. Apart from the blog version, every other design uses the default component.

The question here is; do we want a separate file for all designs when most will be identical except for a boolean prop to show/hide the byline. Without doing that this component currently feels very slim and like our pattern is only half implemented but at the same time, I say it _feels_  that way because I don't think any of our components that are structured this way have a component for absolutely every design we support, they've always used as few as possible to remove code reuse, but they've usually contained more variations than this.


apps-rendering/src/components/Metadata/LiveBlogMetadata.tsx
```diff
@@ -0,0 +1,185 @@
...
+	${from.desktop} {
+		padding: 0;
+		width: 100%;
+	}
+`;
+
+// This styling function is only temprarily used and will be removed
+// after the liveblog header is completed and background colours
+// are added
+const tempraryBackgroundStyle = (format: ArticleFormat): SerializedStyles => {
```
JamieB-gu, line 50: It sounds like these may not be temporary any more, so they could probably be merged into the `blogStyles`?

apps-rendering/src/components/Metadata/Metadata.defaults.tsx
```diff
@@ -0,0 +1,42 @@
...
+	padding-top: ${withByline ? remSpace[1] : 0};
+`;
+
+type Props = {
+	item: Item;
+	withByline: boolean;
+};
+
+export const DefaultMetadata: FC<Props> = ({ item, withByline }: Props) => (
+	<div css={defaultStyles}>
```
JamieB-gu, line 29: I think the pattern you've used in other places is to have styles come in as a `className` prop, and then consumers pass in the `defaultStyles` to that prop? The `Styleable` type can be used to automate creating the Emotion props for this:

https://github.com/guardian/dotcom-rendering/blob/94b90145584680365ebd327570205c174ce9d846/apps-rendering/src/lib.ts#L19-L22

Here's an example of it in use:

https://github.com/guardian/dotcom-rendering/blob/94b90145584680365ebd327570205c174ce9d846/apps-rendering/src/components/HeadingTwo/HeadingTwo.defaults.tsx#L17-L23

JamieB-gu (commented):
Interesting question, I see your point 🤔. I guess your options are something like:

1. This solution: use a `withByline` prop
2. Create two sub-components, one with a byline and one without. Then you have to decide which one is the "default".
3. Create different components for every `Design` that we have to add a byline to.

I don't know which is preferable. As you say, this solution (1) looks a bit different to our other versions of this. It seems reasonable though, so perhaps we should stick with it unless or until we run into problems? To me it looks more manageable than the single large file we had before 👍

apps-rendering/src/components/Metadata/LiveBlogMetadata.tsx
iainjchambers-guardian: Yeah actually that whole thing is complete overkill anyway, the only difference between the two cases was the background colour on mobile.

apps-rendering/src/components/Metadata/Metadata.defaults.tsx
iainjchambers-guardian: Yep, totally true, can do that for this. The annoying thing is that there's two sets of styles, on the main component and on the text. So I can pass the `css` prop to the component and it'll be magically transmogrified to `className` but the text styles will stay of type `SerializedStyles`. Which is annoying, and the reason I was leaving this until after the discussion I raised in the description.

iainjchambers-guardian:
> I don't know which is preferable. As you say, this solution (1) looks a bit different to our other versions of this. It seems reasonable though, so perhaps we should stick with it unless or until we run into problems? To me it looks more manageable than the single large file we had before 👍

Personally, I think this is fine. There's just an itch in the back of my brain telling me I've not applied the pattern that we worked out so it's only a half measure. At the same time, I think sticking too rigidly to a pattern would in this case lead to overall less readability and would be kind of arbitrary

apps-rendering/src/components/Metadata/Metadata.defaults.tsx
```diff
@@ -0,0 +1,41 @@
...
+	flex-grow: 1;
+	padding-top: ${remSpace[1]};
+`;
+
+type Props = {
+	item: Item;
+	withByline: boolean;
+	css?: SerializedStyles;
+	className?: string;
+};
```
JamieB-gu, line 23: Are we able to use the `Styleable` helper for consistency?

```suggestion
type Props = Styleable<{
	item: Item;
	withByline: boolean;
}>;
```

JamieB-gu (commented):
🚀 nice! Couple of small comments, otherwise looks great.

apps-rendering/src/components/Metadata/Metadata.defaults.tsx
```diff
@@ -0,0 +1,42 @@
...
+export const defaultStyles = css`
+	flex-grow: 1;
+	padding-top: ${remSpace[1]};
+`;
+
+type Props = Styleable<{
+	item: Item;
+	withByline: boolean;
+	css?: SerializedStyles;
+	className?: string;
```
JamieB-gu, line 23:
```suggestion
```

JamieB-gu (approved):

PR #6215, opened 2022-10-17
Convert DCR mobile subnav search button to input

DanielCliftonGuardian:
## What does this change?
Brings search input styling in line with frontend.
## Why?
This is the desired styling
## Screenshots

| Frontend      | Before      | After      |
|-------------|-------------|------------|

[frontend]: https://user-images.githubusercontent.com/110032454/196665770-02bbec15-3013-4a87-869c-383cc490e077.png
[before]: https://user-images.githubusercontent.com/110032454/196233855-0896c35b-944f-4a15-be9d-b4f580155a01.png
[after]:https://user-images.githubusercontent.com/110032454/196665579-e85a970e-2e00-4aec-8734-12416bad7db6.png

### DIFF
--- a/dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
+++ b/dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
@@ -255,6 +255,7 @@ export const Columns: React.FC<{
 						placeholder="Search"
 						data-link-name="nav2 : search"
 						className="selectableMenuItem"
+						tabIndex={-1}
 					/>
 
 					<Label hideLabel={true} text="google-search">
@@ -276,7 +277,7 @@ export const Columns: React.FC<{
 						cssOverrides={searchSubmit}
 						data-link-name="nav2 : search : submit"
 						type="submit"
-						className="selectableMenuItem"
+						tabIndex={-1}
 					></Button>
 					<input
 						type="hidden"


oliverlloyd:
Do we want the same styling as frontend here? On Frontend this element is an actual input but in DCR it is a button so I think we should keep it looking like a button. Arguably we could be more buttony

Chloe-Guardian:
Hi all, thanks for @ me on this. I shared this with Akemi, who looks after our design system. Her response is:

"we have a DAC slot to ask some questions around form elements accessibility on 2nd Nov.
We will ask DAC about this search input field as well. (eg. colour contrast, the behaviour: the input field appears after clicking the text ‘search’ is it accessible?, etc)

Before that, I will explore how the search field would look if we were to use the current input style. (The rounded corner input is an old style) We will keep you informed!"

@HarryFischer would you have time to sync up with Akemi about any changes, particularly in light of accessibility? I know you are stacked, so feel free to push to me if you're too busy. 

Side note... it would be good to know what the process is for design to reach this stage. e.g I assume Alex B will have approved any change that comes through here but is there a process/review I can join (or create)? Any thoughts so welcome.... also please accept my advanced apologies for my essay and possibly miss using github .

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -162,30 +235,45 @@ export const Columns: React.FC<{
...
-				</div>
-			</ThemeProvider>
+			<form css={searchBar} action="https://www.google.co.uk/search">
+				<TextInput
+					hideLabel={true}
+					label="Search input"
+					cssOverrides={searchInput}
+					name="q"
+					placeholder="Search"
+					data-link-name="nav2 : search"
```
jamesgorrie, line 245: Does this work on inputs?

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -162,30 +235,45 @@ export const Columns: React.FC<{
...
+						<SvgMagnifyingGlass
+							isAnnouncedByScreenReader={true}
+							size="medium"
+						/>
+					</div>
+				</Label>
+				<Button
+					icon={
+						<SvgArrowRightStraight
+							isAnnouncedByScreenReader={true}
```
jamesgorrie, line 259: I'd be interested as to how [this text](https://github.com/guardian/source/blob/529f25a8321ec14a51bb0045604c81a48102e1a0/packages/%40guardian/source-react-components/vendor/icons/SvgArrowRightStraight.tsx#L37) being read out is useful?

HarryFischer:
Looks good - I think this is better then the button
Cheers @DanielCliftonGuardian

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -127,12 +130,82 @@ const searchBar = css`
...
 	margin-bottom: 24px;
 	margin-right: 41px;
 	padding-bottom: 15px;
 `;
 
+const searchInput = css`
+	${textSans.large()}
+	background-color: rgba(255,255,255, .1);
+	border: 0;
+	border-radius: 1000px;
```
oliverlloyd, line 147: Maybe we can use a percentage here?

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -127,12 +130,82 @@ const searchBar = css`
 	${from.desktop} {
 		display: none;
 	}
-	margin-left: 45px;
+	box-sizing: border-box;
+	display: block;
+	margin-left: 13px;
+	max-width: 23.75rem;
```
oliverlloyd, line 136: Rems are not used in dcr, we favour pixels (1rem = 16pixels so this would be 380px)

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -127,12 +130,82 @@ const searchBar = css`
...
+	color: ${neutral[100]};
+	height: 36px;
+	padding-left: 38px;
+	vertical-align: middle;
+	width: 100%;
+	&::placeholder {
+		color: ${neutral[100]};
+	}
+	&:focus {
+		outline: none;
```
oliverlloyd, line 158: I think we probably want to keep the default [focus styles](https://theguardian.design/2a1e5182b/p/08dc26-interaction-states/t/314e46) that Source gives us

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -127,12 +130,82 @@ const searchBar = css`
...
+	border-radius: 1000px;
+	box-sizing: border-box;
+	color: ${neutral[100]};
+	height: 36px;
+	padding-left: 38px;
+	vertical-align: middle;
+	width: 100%;
+	&::placeholder {
+		color: ${neutral[100]};
+	}
```
oliverlloyd, line 156: Does it work if we preserve the existing styles from Source here? I know this ticket is about aligning with Frontend but I think within that goal we can also allow some deviation to preserve the design system standards for text inputs

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -174,28 +247,44 @@ export const Columns: React.FC<{
...
+						name="q"
+						placeholder="Search"
+						data-link-name="nav2 : search"
+						tabIndex={-1}
+					/>
 
+					<Label hideLabel={true} text="google-search">
+						<div css={searchGlass}>
+							<SvgMagnifyingGlass
+								isAnnouncedByScreenReader={true}
```
oliverlloyd, line 264: What will the screen reader announce here? I'm not 100% sure how this works but my guess is it will look for the label text but it seems as though there's none here

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -174,28 +247,44 @@ export const Columns: React.FC<{
...
-							Search
-						</LinkButton>
-					</div>
-				</ThemeProvider>
+				<form css={searchBar} action="https://www.google.co.uk/search">
+					<TextInput
+						hideLabel={true}
+						label="Search input"
+						cssOverrides={searchInput}
+						name="q"
```
oliverlloyd, line 255: Does this name represent the `q` query param sent to google in which case it _needs_ to be 'q'?

If so, can we add a comment explaining it just in case anyone lands here in the future thinking it needs renaming (which was my first instinct!)

If not, can we use a longer, more descriptive name?

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -174,28 +247,44 @@ export const Columns: React.FC<{
...
+						placeholder="Search"
+						data-link-name="nav2 : search"
+						tabIndex={-1}
```
oliverlloyd, line 258: Is this the reason I can't tab to this control? I appreciate there's a few things going on here (a button and another hidden input) so it's not that simple but when testing locally focus jumps over this control which I _presume_ is unintentional?

dotcom-rendering/src/web/components/Nav/ExpandedMenu/Columns.tsx
```diff
@@ -174,28 +247,44 @@ export const Columns: React.FC<{
...
+					</Label>
+					<Button
+						icon={
+							<SvgArrowRightStraight
+								isAnnouncedByScreenReader={false}
+								size="medium"
+							/>
+						}
+						aria-label="Search with google"
+						cssOverrides={searchSubmit}
```
oliverlloyd, line 277: Not sure where the css is that needs to change for this but the focus styles for this button seem a bit off (see image)

Screenshot 2022-10-25 at 10 21 58

oliverlloyd (approved):
This is very cool, I tried it locally and it works so well, great work!

One thing that would be nice here is if we can build out the description a little to explain what is happening here, why we're doing it and what the history of this approach is. This feels like a PR that future developers will look at to try to understand things better and it's a wonderful thing when developers from the past speak to you about what motivated them and why they did things. It's also a great thing to do in terms of helping us learn about what we're doing, the more you explain something to others, the better we understand it ourselves.

The other area I wondered about is how far down the parity path we needed to go. It could be there are some styling decisions taken by Source that we want to preserve and that some deviation from Frontend is okay?

Random question: Do we need to support [this feature](https://github.com/guardian/frontend/pull/17458)? My guess is not but it might be good to confirm this with Ophan / Rhianna

Happy to  give this one a 👍 and leave it up to you how to work through the comments and decide when to merge

PR #6225, opened 2022-10-18
Adds support for Fixed/Small/FastVIII container

Georges-GNM:
## What does this change?

Adds support for the fixed/small/fastVIII container, closing issue #5145 

There _is_ a discrepancy with the image shown in the Figma file (which displays 7 cards instead of 8) but this PR is on par with the Frontend patter and can revise it if needed. 

The screenshots comparison also shows different colours on some of the cards, and the crossword card is missing its picture in the DCR container (see issue #6236), however these issues don't seem related to the implementation of the container itself (and so should be looked at separately). 

## Why?
Furthers the fronts migration.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![Frontend before][] | ![DCR before][] |
| ![Frontend before][] | ![DCR after][] |

[Frontend before]: https://user-images.githubusercontent.com/102960844/196447142-ee7502a0-f341-46d7-a6d0-7789726533e5.png

[DCR before]: 
https://user-images.githubusercontent.com/102960844/196443597-9d45a807-495f-4c10-97cc-ec6be65d1358.png
 
[DCR after]: https://user-images.githubusercontent.com/102960844/196464453-5443c0d3-237c-4df9-b688-28492ed57149.png

Co-authored by: @bryophyta


mxdvl:
Is the colour issue due to the inferred formats of the cards? Do we not support lifestyle cards?

PR #6229, opened 2022-10-18
No legacy JS (experiment)

mxdvl:
## What does this change?

For users in the `dcrJavascriptBundleVariant`, stop serving legacy JavaScript, as per the current [Three Javascript Experiences RFC](https://docs.google.com/document/d/1Fhu_g4Q9-a18jWftnGt1n6Scn8x88X-7jX2Ygr8LJrk/edit).

- feat: create `Bundle` type
- feat: drop legacy bundle (test)
- chore(deps): bump yarn.lock

## Why?

We believe older browsers should benefit from being sent less JS.

Reimplementation of #5206, now that we have a more robust bundle testing framework introduced in #5598

## Tested

- [ ] On browserstack
- [ ] In a legacy browser 
- [ ] Flagged this to all teams

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/196492078-e78880d2-57fd-43a3-9e09-7c77da2700b1.png
[after]: https://user-images.githubusercontent.com/76776/196491989-363428c3-5dc4-4d84-b93e-0c94fff494e5.png

### DIFF
--- a/dotcom-rendering/src/web/server/pageTemplate.ts
+++ b/dotcom-rendering/src/web/server/pageTemplate.ts
@@ -1,5 +1,6 @@
 import { brandBackground, resets } from '@guardian/source-foundations';
 import he from 'he';
+import { ophanNomoduleScript } from '../../../scripts/webpack/bundles';
 import { ASSET_ORIGIN } from '../../lib/assets';
 import { getFontsCss } from '../../lib/fonts-css';
 import { getHttp3Url } from '../lib/getHttp3Url';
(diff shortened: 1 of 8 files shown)


dotcom-rendering/src/web/server/pageTemplate.ts
```diff
@@ -326,6 +327,7 @@ https://workforus.theguardian.com/careers/product-engineering/
 					).toString()}" />
                 </noscript>
                 ${priorityScriptTags.join('\n')}
+				${ophanNomoduleScript}
```
arelra, line 330: Minor - indentation

arelra (commented):
Great to see the new testing strategy moving forwards! 🚀 

I am a little confused about what the sample of this test would represent.

iiuc the intention behind this test is to verify if we can switch users who currently get a legacy bundle because their browser doesn't support modules to a no-JS experience.

However is it right to say in this test we will be serving a no-JS experience to a sample of _all_ users vs a sample of users who currently get the legacy bundle?

So it could be the users sampled by the test don't strictly overlap with the users the test result would affect?

That might be fine and acceptable but I just wanted to check.

Issue #6181, opened 2022-10-12
RFC: Migrating 'show more'


bryophyta:
# RFC: Migrating Show More

This is to discuss approaches to closing #4604.

## Requirements (unknowns)

* Do we need to preserve the current convention for `data-link-name`?
   * Probably not, as per discussion with Guy in Data + DCR.
* Is adFree in scope?
   * Assuming yes, if it’s straightforward.
* Do we need format information to be passed to DCR?
   * Yes (well; containerPalette)
* How much deduping do we need to do?
   * Current situation: Cards are deduped on the server side and the client side. The latter catches some edge cases which can’t be caught on the server.
   * There’s been some discussion as to how much we should care about those edge cases, but no clear decision was reached.

## Current Solution

1. User presses Show More on a Fronts container.
2. API Call is made to e.g. `/uk/show-more/headlines.json` (`/*path/show-more/*containerId.json`)
3. Frontend fetches full PressedPage from S3 and extracts the collection with the requested ID.
4. Frontend uses ContainerLayout to filter out all cards that are already visible on the client.
5. Frontend renders the filtered cards using twist/twirl/thingy
6. Rendered HTML is returned by the API and inserted client side
7. Duplicated cards are removed from “Show More” section using DOM manipulation

## Schematic description of the main DCR options

We can think of the current solution as one way of achieving the following tasks:

- I. Locate and retrieve the pressed collection data for the relevant container.
- II. Remove cards that are already visible on the Front.
- III. Remove paid cards if the request isAdFree.
- IV. Determine the correct palette for the container.
- V. Render the collection JSON as HTML.
- VI. Remove cards which have been duplicated due to the collection changing between the original page load and the `/show-more/` endpoint call.

(I) and (III) are naturally done in Frontend.
(IV) is naturally done in DCR.
(VI) can only really be done on the client side.
(V) is easy, but the choice of where to do the rendering (DCR server, or client) is consequential.
(II) is probably the most complex problem, given the current implementation,[^1] and the choice of whether to render the HTML on the client or on the server doesn’t really affect this problem. However, it’s worth noting that (VI) and (II) could essentially be treated as a single problem (that of ‘deduplicating’ the new cards against the existing cards in the collection) if they are both performed on the client side. Again, this could be done whether the initial markup is rendered on the server or the client (although it’ll be simpler to implement with client-side rendering).
  
[...]

PR #6230, opened 2022-10-18
🏝️ Islands: deferUntil="interaction"

OllysCoding:
Working with @oliverlloyd, this PR adds support for deferring the loading of an island until its content is 'interacted' with.

## Why do we want this pattern?

The ability to defer loading javascript for a feature until the user interacts with it will enable us to fundamentally change how we think about adding javascript to the page.

Until now, when writing javascript features on DCR, we had to consider that the javascript would be downloaded by every user on the page it would appear on. We improved things with the deferUntil options `idle` (waiting till the CPU thread isn't busy) and `visible` (great if the content is much lower on the page), however fundamentally, building complex javascript features would mean adding weight to the page for everyone, even if they weren't used by the majority of users.

However with `deferUntil="interaction"`, we can put off loading _any_ additional code on the client until a user actually interacts with the islands content, theoretically allowing us to develop feature-rich experiences that don't impact users who aren't interested in them.

_However,_ we should still consider
- We need to keep considering how writing javascript can impact accessibility & our no JS experience
- Maintaining complex javascript features can requires additional dev work & time
- The HTML of the gu-island & its props will add to the initial page weight, even if the JS isn't downloaded ([example](https://user-images.githubusercontent.com/9575458/196423239-489b0b79-9e5f-40ad-b9dd-dcd383f2aa20.png))

### Some examples of where this might be useful are:
- Lightbox on articles (https://github.com/guardian/dotcom-rendering/pull/6154)
- 'Show more' on fronts (https://github.com/guardian/dotcom-rendering/issues/6181)
- Existing islands like 'SubNav'?

## How does this work? 

Upgraded from the original trial implementation in #6154, this relatively simple implementation works by:
- Islands marked as `deferUntil="interactive"` will receive a 'click' event listener when the islands/init script runs on page load
- If the user interacts with (clicks) in the island area, the island will be downloaded and hydrated
- Once the island is hydrated, we 'replay' the click event, enabling the island to now receive it, and act as it would have if it was downloaded initially (e.g incrementing a counter)
- After this the event listener is removed - the island is now hydrated, and can work as a normal react component

The 'event replay' approach is really quite powerful, as it enables us to write the our `.importable` components just as we'd write any other island, with no special considerations needed

## How do I implement it?

Since the initial interaction (e.g button press) is repeated once the island is hydrated, you can essentially write your island as you would write any regular 'hydratable' island:

```tsx
// MyIsland.importable.tsx
import { useState } from 'react';

export const TestInteraction = ({ title }: { title: string }) => {
	const [clicks, setClicks] = useState(0);

	return (
		<>
			<h2>Hello World: {clicks}</h2>
			<button type="button" onClick={() => setClicks(clicks + 1)}>
				Click me
			</button>
		</>
	);
};
```
Then, add your island to the page
```tsx
<Island deferUntil="interaction">
	<MyIsland title="Hello World" />
</Island>
```

Because your component _must_ be rendered on the server as well as the client (otherwise there would be no content for the user to interact with), you cannot pass `clientOnly` or `placeholderHeight`, you also cannot pass `expediteLoading`, as this would defeat the purpose.

### Demo

Initial load click on interactive element (button)

https://user-images.githubusercontent.com/9575458/196464122-119c9706-a2ab-40a9-9dbf-09c4c8f94948.mov

Initial load click on non-interactive element (empty space / text in the island)

https://user-images.githubusercontent.com/9575458/196464303-9ee41ba5-65ff-4436-8f9d-5d045beee351.mov

Initial load click on interactive element with keyboard controls (button - pressed enter)

https://user-images.githubusercontent.com/9575458/196464424-22d79959-9a6f-44cf-8ebe-29040492ad81.mov

### DIFF
--- a/dotcom-rendering/index.d.ts
+++ b/dotcom-rendering/index.d.ts
@@ -574,7 +574,7 @@ declare namespace JSX {
 	interface IntrinsicElements {
 		'gu-island': {
 			name: string;
-			deferUntil?: 'idle' | 'visible';
+			deferUntil?: 'idle' | 'visible' | 'interaction';
 			clientOnly?: boolean;
 			expediteLoading?: boolean;
 			props: any;
--- a/dotcom-rendering/src/web/browser/initDiscussion/init.ts
+++ b/dotcom-rendering/src/web/browser/initDiscussion/init.ts
@@ -20,7 +20,7 @@ function forceHydration() {
 		props.expanded = true;
 
 		// Force hydration
-		doHydration(name, props, guElement);
+		void doHydration(name, props, guElement);
 	} catch (err) {
 		// Do nothing
 	}
--- a/dotcom-rendering/src/web/browser/islands/doHydration.ts
+++ b/dotcom-rendering/src/web/browser/islands/doHydration.ts
@@ -1,6 +1,5 @@
-/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-member-access -- necessary for calling our async loaded modules */
 import { log } from '@guardian/libs';
-import type { Attributes } from 'preact';
 import { h, hydrate, render } from 'preact';
 import { initPerf } from '../initPerf';
 
@@ -15,11 +14,11 @@ import { initPerf } from '../initPerf';
  * @param data The deserialised props we want to use for hydration
  * @param element The location on the DOM where the component to hydrate exists
  */
-export const doHydration = (
+export const doHydration = async (
 	name: string,
-	data: Attributes | null,
+	data: { [key: string]: unknown } | null,
 	element: HTMLElement,
-): void => {
+): Promise<void> => {
 	// If this function has already been run for an element then don't try to
 	// run it a second time
 	const alreadyHydrated = element.dataset.guReady;
@@ -27,7 +26,7 @@ export const doHydration = (
 
 	const { start, end } = initPerf(`hydrate-${name}`);
 	start();
-	import(
+	return import(
 		/* webpackInclude: /\.importable\.tsx$/ */
 		/* webpackChunkName: "[request]" */
 		`../../components/${name}.importable`
--- a/dotcom-rendering/src/web/browser/islands/doHydration.ts
+++ b/dotcom-rendering/src/web/browser/islands/doHydration.ts
@@ -26,7 +26,7 @@ export const doHydration = async (
 
 	const { start, end } = initPerf(`hydrate-${name}`);
 	start();
-	return import(
+	await import(
 		/* webpackInclude: /\.importable\.tsx$/ */
 		/* webpackChunkName: "[request]" */
 		`../../components/${name}.importable`
--- a/dotcom-rendering/src/web/browser/islands/initHydration.ts
+++ b/dotcom-rendering/src/web/browser/islands/initHydration.ts
@@ -1,6 +1,7 @@
 import { doHydration } from './doHydration';
 import { getName } from './getName';
 import { getProps } from './getProps';
+import { onInteraction } from './onInteraction';
 import { whenIdle } from './whenIdle';
 import { whenVisible } from './whenVisible';
 
@@ -27,18 +28,28 @@ export const initHydration = (elements: NodeListOf<Element>): void => {
 			switch (deferUntil) {
 				case 'idle': {
 					whenIdle(() => {
-						doHydration(name, props, element);
+						void doHydration(name, props, element);
 					});
 					break;
 				}
 				case 'visible': {
 					whenVisible(element, () => {
-						doHydration(name, props, element);
+						void doHydration(name, props, element);
+					});
+					break;
+				}
+				case 'interaction': {
+					onInteraction(element, (targetElement) => {
+						void doHydration(name, props, element).then(() => {
+							targetElement.dispatchEvent(
+								new MouseEvent('click'),
+							);
+						});
 					});
 					break;
 				}
 				default: {
-					doHydration(name, props, element);
+					void doHydration(name, props, element);
 				}
 			}
 		}
--- a/dotcom-rendering/src/web/browser/islands/onInteraction.ts
+++ b/dotcom-rendering/src/web/browser/islands/onInteraction.ts
@@ -1,0 +1,21 @@
+/**
+ * Use this function to delay execution of something until an element is interacted
+ * with
+ *
+ * @param element : The html element that we want to wait for an interaction on;
+ * @param callback : This is fired when the element is clicked on
+ */
+export const onInteraction = (
+	element: HTMLElement,
+	callback: (e: HTMLElement) => void,
+): void => {
+	element.addEventListener(
+		'click',
+		(e) => {
+			if (e.target instanceof HTMLElement) {
+				callback(e.target);
+			}
+		},
+		{ once: true },
+	);
+};
--- a/dotcom-rendering/src/web/components/Island.tsx
+++ b/dotcom-rendering/src/web/components/Island.tsx
@@ -2,7 +2,7 @@
 
 import { Placeholder } from './Placeholder';
 
-type When = 'idle' | 'visible' | 'interaction';
+type When = 'idle' | 'visible';
 
 interface HydrateProps {
 	deferUntil?: When;
@@ -22,10 +22,10 @@ interface ClientOnlyProps {
 
 interface InteractionProps {
 	deferUntil: 'interaction';
-	clientOnly?: false;
+	clientOnly?: never;
 	placeholderHeight?: never;
 	children: JSX.Element;
-	expediteLoading?: false;
+	expediteLoading?: never;
 }
 
 /**
@@ -60,7 +60,7 @@ const decideChildren = (
  * namimg convention
  *
  * @param {HydrateProps | ClientOnlyProps} props - JSX Props
- * @param {When} props.deferUntil - Delay when client code should execute
+ * @param {When | 'interaction'} props.deferUntil - Delay when client code should execute
  * 		- idle - Execute when browser idle
  * 		- visible - Execute when component appears in viewport
  *      - interaction - Execute when component is clicked on in the viewport
--- a/dotcom-rendering/src/web/components/Island.tsx
+++ b/dotcom-rendering/src/web/components/Island.tsx
@@ -2,7 +2,7 @@
 
 import { Placeholder } from './Placeholder';
 
-type When = 'idle' | 'visible';
+type When = 'idle' | 'visible' | 'interaction';
 
 interface HydrateProps {
 	deferUntil?: When;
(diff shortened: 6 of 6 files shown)


dotcom-rendering/index.d.ts
```diff
@@ -574,7 +574,7 @@ declare namespace JSX {
 	interface IntrinsicElements {
 		'gu-island': {
 			name: string;
-			deferUntil?: 'idle' | 'visible';
+			deferUntil?: 'idle' | 'visible' | 'interaction';
```
oliverlloyd, line 577: `clicked`?

I think if you pressed me I would admit to preferring `interaction` but I'm putting this out there because maybe this works better 🤷 ?

dotcom-rendering/src/web/browser/initDiscussion/init.ts
```diff
@@ -20,7 +20,7 @@ function forceHydration() {
 		props.expanded = true;
 
 		// Force hydration
-		doHydration(name, props, guElement);
+		void doHydration(name, props, guElement);
```
oliverlloyd, line 23: This relates to [this change](https://github.com/guardian/dotcom-rendering/pull/6170) I believe

dotcom-rendering/src/web/browser/islands/initHydration.ts
```diff
@@ -27,18 +28,28 @@ export const initHydration = (elements: NodeListOf<Element>): void => {
...
+					break;
+				}
+				case 'interaction': {
+					onInteraction(element, (targetElement) => {
+						void doHydration(name, props, element).then(() => {
+							targetElement.dispatchEvent(
+								new MouseEvent('click'),
+							);
+						});
 					});
```
oliverlloyd, line 48: 😻 this is very clear

dotcom-rendering/src/web/browser/islands/onInteraction.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @param element : The html element that we want to wait for an interaction on;
+ * @param callback : This is fired when the element is clicked on
+ */
+export const onInteraction = (
+	element: HTMLElement,
+	callback: (e: HTMLElement) => void,
+): void => {
+	element.addEventListener(
+		'click',
```
oliverlloyd, line 13: Does the click event also get fired if I press the enter key when an element has focus?

dotcom-rendering/src/web/components/Island.tsx
```diff
@@ -2,7 +2,7 @@
 
 import { Placeholder } from './Placeholder';
 
-type When = 'idle' | 'visible';
+type When = 'idle' | 'visible' | 'interaction';
```
oliverlloyd, line 5: Should we add this here? Perhaps, now that we have different use cases for different 'whens', we could remove the entire `When` type and just inline `'idle' | 'visible'`?

dotcom-rendering/src/web/components/Island.tsx
```diff
@@ -20,13 +20,21 @@ interface ClientOnlyProps {
 	expediteLoading?: boolean;
 }
 
+interface InteractionProps {
+	deferUntil: 'interaction';
+	clientOnly?: false;
+	placeholderHeight?: never;
+	children: JSX.Element;
+	expediteLoading?: false;
```
oliverlloyd, line 28:
```suggestion
	expediteLoading?: never;
```

Does this work with never? If so I guess that's a better dev ex.

dotcom-rendering/src/web/components/Island.tsx
```diff
@@ -20,13 +20,21 @@ interface ClientOnlyProps {
 	expediteLoading?: boolean;
 }
 
+interface InteractionProps {
+	deferUntil: 'interaction';
+	clientOnly?: false;
```
oliverlloyd, line 25:
```suggestion
	clientOnly?: never;
```

?

oliverlloyd (approved):
This is a very exciting improvement to islands!

I've approved as I think the implementation is sound however I've made a few comments, happy for you to manage those. I'd also love to understand how this works for keyboard users?

dotcom-rendering/index.d.ts
bryophyta: I'd second `clicked` because there are multiple kinds of interaction and click events are the kind that are actually being invoked here, e.g. tabbing to focus an `<input>` won't trigger a click event (afaik)

dotcom-rendering/src/web/browser/initDiscussion/init.ts
mxdvl: Yes, we are now able to use the `void` keyword, and it is recommended for promises that are not awaited.

dotcom-rendering/src/web/browser/islands/onInteraction.ts
mxdvl: There’s a demo video suggesting that it does!

dotcom-rendering/src/web/browser/islands/doHydration.ts
```diff
@@ -15,19 +14,19 @@ import { initPerf } from '../initPerf';
...
+): Promise<void> => {
 	// If this function has already been run for an element then don't try to
 	// run it a second time
 	const alreadyHydrated = element.dataset.guReady;
 	if (alreadyHydrated) return;
 
 	const { start, end } = initPerf(`hydrate-${name}`);
 	start();
-	import(
+	return import(
```
mxdvl, line 29: I think this is refactor that is orthogonal to this change. Could it be split in its own PR?

Making this function async is not really giving us anything as we are not awaiting it anywhere.

mxdvl (commented):
This is great. Would second the naming of `clicked` as it’s a specific kind of interaction.

I would prefer the changes related to promises to be extracted into their own PR.

dotcom-rendering/src/web/browser/islands/initHydration.ts
```diff
@@ -27,18 +28,28 @@ export const initHydration = (elements: NodeListOf<Element>): void => {
...
-						doHydration(name, props, element);
+						void doHydration(name, props, element);
+					});
+					break;
+				}
+				case 'interaction': {
+					onInteraction(element, (targetElement) => {
+						void doHydration(name, props, element).then(() => {
+							targetElement.dispatchEvent(
+								new MouseEvent('click'),
```
bryophyta, line 45: this is a great idea!

I'm wondering how the replay functionality will work in the case where the  placeholder is clicked more than once before the hydration is complete?

1. Will it attempt to hydrate again? I guess SWR will be caching the fetch call, but afaict `doHydration` might end up running again because the `guReady` flag is only set after the hydration is complete.
  - We wouldn't want to prevent retries entirely, in the case where it fails, though, so perhaps in practice a simple 'throttling' function would do the trick?
2. If the user clicks more than once before the hydration is complete, will all of the clicks be stored to be replayed after hydration? (E.g. in the counter case we'd plausibly want to replay multiple clicks, though that's debatable.)

oliverlloyd: Great point!

To my understanding any clicks made in the dead zone (aka the uncanny valley) after the initial one that triggers hydration and before the component has been fully hydrated will be lost. Lost clicks are an accepted part of the hydration problem, even if you hydrate the whole page or give something the highest priority for hydration, there's always a gap where interaction gets lost.

In our case I think the trade off we're making is worth it. Deferring until click does increase the risk there will be interaction during the dead zone but I think the fact we capture and replay the first event means we probably offer a better experience for most readers.

bryophyta (approved):
This looks really great! I added a question about whether the island will ever try to hydrate itself more than once, but that seems fairly minor to me.

My feeling is that we should test this new pattern a bit more before using it in production, and also add some documentation on when/how it should be used. But imo these are separate issues and I'm v happy to approve this PR now 👍

dotcom-rendering/index.d.ts
OllysCoding: I'd actually change the question to which forms of interaction _should_ we support, and do we want to support multiple, e.g some kind of interface

```tsx
<island
    deferUntil="interaction"
    events={["click", "keypress"]}
>
    ....
</island>
```

dotcom-rendering/src/web/browser/initDiscussion/init.ts
OllysCoding: I quite like it!

dotcom-rendering/src/web/browser/islands/doHydration.ts
oliverlloyd: `await import ...` is probably a nicer read if that's an option

dotcom-rendering/src/web/browser/islands/initHydration.ts
OllysCoding: The current functionality is quite primitive, the event listener will only run once, which triggers the hydration, and then the event replays. As Oliver says, any other clicks will be lost - but we don't risk trying to hydrate it multiple times.

Right now the 'decoupled' nature of the the 'trigger' and the actual hydration process means that a more advanced system would require more work, but perhaps that's what we need?

mxdvl: I would focus on what we need right now, not what we might need in the future. Islands load rather quickly, especially if they are cached!

dotcom-rendering/src/web/browser/islands/doHydration.ts
OllysCoding: Would definitely support this refactor :)

dotcom-rendering/index.d.ts
bryophyta: this sounds good to me! I was thinking along similar lines to your `events` idea but I was wary of suggesting a premature abstraction; if you think we're likely to have use-cases for other event types any time soon then I agree that this is the way to go!

AshCorr (approved):
Looks good!

One thought though, for islands such as the front Hide/Show button we have a pattern of having a kind of "controller" island that controls all instances of the button on the page from a single island, this was to avoid downloading and executing the island on every "Show/Hide" button.

If you had multiple of the same "interaction" island would would the user need to re-download the Island bundle every time they click on different instance of the island? If not then great, maybe we can refactor Show/Hide slightly too!

OllysCoding:
> If you had multiple of the same "interaction" island would would the user need to re-download the Island bundle every time they click on different instance of the island?

No, the bundle will only every be downloaded once! Of course though, each button click still requires a react instance to be spun up & hydrated - which is an overhead (though relatively low)

I thought about whether this could be used for show/hide - but wasn't sure if you'd still need an island loaded early to actually 'hide' the containers on page load, but perhaps we keep that as its own thing, and still use "interaction" islands for the actual show/hide buttons?

dotcom-rendering/index.d.ts
OllysCoding: To avoid premature abstraction, I'm going to keep it as on="interaction", as things like hitting enter on a button will cause it to hydrate (technically it's a click event, which is somewhat counter-intuitive). Definitely like the idea of the `events` api for the future though

OllysCoding:
Thanks everyone for their feedback! 

I think there's going to be a few improvements to made down the line, but I want to merge this PR as-is, and get it in the hands of people actually building features with it, then we can improve as needed!

PR #6232, opened 2022-10-19
docs(adr): specify Three Javascript Experiences

mxdvl:
## What does this change?

Introduce the Three Javascript Experiences ADR.

## Why?

The proposal has been put to client-side teams for a long time and no further comments have emerged.


mxdvl:
This will be revisited by the Open Journalism team.

PR #6235, opened 2022-10-19
Hides the Topic Filter Bank if there are no availableTopics (with count > 2)

sophie-macmillan:
## What does this change?
This fixes a bug where if there were availableTopics, but none of them had a count greater than two, the filter topic bank would render with a heading but no filter buttons.
  - Checks that there are relevant topics before rendering the TopicFilterBank in both LiveLayout.tsx and LiveBlogRenderer.tsx
  - Adds a check to make sure that there are keyEvents or relevantTopics in the TopicFilterBank itself
  - Adds tests for the TopicFilterBank component

## Why?
In edge cases the TopicFilterBank component was rendering with a heading but no buttons (see image below)

## Screenshots

| Before      | After      |
|-------------|------------|
| image | image|

### DIFF
--- a/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
+++ b/dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
@@ -5,7 +5,7 @@ import { decidePalette } from '../lib/decidePalette';
 import { FilterButton } from './FilterButton.importable';
 
 type Props = {
-	availableTopics: Topic[];
+	availableTopics?: Topic[];
 	selectedTopics?: Topic[];
 	format: ArticleFormat;
 	filterKeyEvents: boolean;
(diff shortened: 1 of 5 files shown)


marjisound (approved):
It looks amazing, well done 👏

dotcom-rendering/src/web/layouts/LiveLayout.tsx
```diff
@@ -285,10 +288,12 @@ export const LiveLayout = ({ CAPIArticle, NAV, format }: Props) => {
 	const cricketMatchUrl =
 		CAPIArticle.matchType === 'CricketMatchType' && CAPIArticle.matchUrl;
 
-	const showTopicFilterBank = !!CAPIArticle.config.switches.automaticFilters;
-	const hasAvailableTopics = !!CAPIArticle.availableTopics?.length;
+	const showTopicFilterBank =
```
abeddow91, line 291: much cleaner!

dotcom-rendering/src/web/lib/LiveBlogRenderer.tsx
```diff
@@ -111,7 +114,7 @@ export const LiveBlogRenderer = ({
 				<></>
 			)}
 
-			{switches.automaticFilters && !!availableTopics?.length && (
+			{switches.automaticFilters && hasRelevantTopics(availableTopics) && (
```
abeddow91, line 117: much more readable 👏

abeddow91 (approved):
Looks great, and so good to see a boost in test coverage - thanks!

oliverlloyd (commented):
Great stuff, just one question and I'm not completely familiar with this area so I could well be wrong on this but is the business logic here that if there are some key events but zero available topics, we still want to render?

dotcom-rendering/src/web/components/TopicFilterBank.importable.tsx
```diff
@@ -75,19 +75,27 @@ const isEqual = (selectedTopic: Topic, availableTopic: Topic) =>
 	availableTopic.type === selectedTopic.type &&
 	availableTopic.value === selectedTopic.value;
 
+const getTopFiveTopics = (availableTopics: Topic[]) =>
+	availableTopics
+		.slice(0, 5)
+		.filter((topic) => !!topic.count && topic.count > 2);
```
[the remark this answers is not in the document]
sophie-macmillan: Slicing first probably speeds this up (depending on the length of the array), but filtering first means we're more likely to get a subset of topics with high enough count. From the data i've seen they look like they're ordered by descending count (but I couldn't find anything in code to prove that), so I think we're ok to leave it as is

oliverlloyd:
That [other visual diff](https://www.chromatic.com/test?appId=5dfcbf3012392c0020e7140b&id=635171f201451d79e366dbd2) that was showing up is a false negative we sometimes see. I've approved it through.

sophie-macmillan:
> Great stuff, just one question and I'm not completely familiar with this area so I could well be wrong on this but is the business logic here that if there are some key events but zero available topics, we still want to render?

If the feature is enabled and
  - There are availableTopics and keyEvents -> show the filter bank
  - keyEvents only -> hide the filter bank (and show the keyEvents toggle)

PR #6172, opened 2022-10-12
Add support for `nav/list` container type

AshCorr:
## What does this change?

Adds support for the `nav/list` container type.

## Why?

One step closer to DCR fronts!

Closes #5146 
Merges https://github.com/guardian/frontend/pull/25581

## Screenshots

| Frontend      | DCR      |
|-------------|------------|
| ![before][] | ![after][] |
| ![before2][] | ![after2][] |
| ![before3][] | ![after3][] |

[before]: https://user-images.githubusercontent.com/21217225/195337481-417bf364-83d3-4a0d-bdfc-ab2675748bce.png
[after]:https://user-images.githubusercontent.com/21217225/195337513-9c93f44b-e209-4c91-8d50-fa01e63c8b7f.png
[before2]: https://user-images.githubusercontent.com/21217225/195349392-f51953b8-c608-40ae-be9e-0488810c237b.png
[after2]:https://user-images.githubusercontent.com/21217225/195349429-8112e38e-3035-41a8-936d-c2e8d70aa609.png
[before3]: https://user-images.githubusercontent.com/21217225/195351108-d399e6a9-7fa4-4e8b-a240-4d14131e0952.png
[after3]:https://user-images.githubusercontent.com/21217225/195351093-bce87883-ef60-432f-832d-801a3ea3776d.png

PR #6248, opened 2022-10-20
feat: Allow container dividers to be controllable through the palette

AshCorr:
## What does this change?

- Use containerPalette to decide vertical divider colour
- Stop using containerPalette to decide container margin border colour
- Update container border colours as per https://www.figma.com/file/agVeiMZULSWlf4JxNgkng4/Fronts-Palettes

## Why?

Previously the container border colour was being used to decide what colour the margin of a container should be, this thing here: 
image

Acording to the figma it looks like this border should not actually change colour and instead the container border property should be used to control the dividers between cards, these things here:

image

A comment was made on https://github.com/guardian/dotcom-rendering/pull/6172#discussion_r993651916 about why we weren't changing divider colour when the containerPalette was set, this PR should address that comment.

## Screenshot

image


dotcom-rendering/src/web/lib/decideContainerOverrides.ts
```diff
@@ -325,5 +346,6 @@ export const decideContainerOverrides = (
 		topBar: {
 			card: topBarCard(containerPalette),
 		},
+		divider: divider(containerPalette),
```
oliverlloyd, line 349: Would it work if we use `border.container` instead of creating a new property? I presume we'd always want these colours to match?

AshCorr: Good question that I'm not sure the answer is to yet!

AshCorr: Seems like we should be using border.container!

dotcom-rendering/src/web/components/NavList.tsx
```diff
@@ -28,7 +28,10 @@ const ulStyles = css`
 	}
 	column-count: 4;
 
-	column-rule: 1px solid ${border.secondary};
+	column-rule: 1px solid
+		${containerOverrides
+			? containerOverrides.border.container
+			: border.secondary};
```
mxdvl, line 34: What follows is information about Emotion and does not have to be implemented as part of this PR.

In order to keep the dynamic and static parts of the CSS distinct, which reduces the runtime work Emotion has to do, you could extract the [`column-rule-color`](https://developer.mozilla.org/en-US/docs/Web/CSS/column-rule-color) to its own style declaration, like so:

```ts
columnRuleColor = (containerOverrides?: ContainerOverrides) => css`
	column-rule-color: ${containerOverrides
		? containerOverrides.border.container
		: border.secondary};
`;

/* later… */
<ul css={[ulStyles, columnRuleColor(containerOverrides)}>
```

dotcom-rendering/src/web/lib/decideContainerOverrides.ts
```diff
@@ -245,21 +245,21 @@ const textContainerToggle = (containerPalette: DCRContainerPalette): string => {
 const borderContainer = (containerPalette: DCRContainerPalette): string => {
 	switch (containerPalette) {
 		case 'LongRunningPalette':
-			return 'rgba(0,0,0, 0.2)';
+			return transparentColour(neutral[86], 0.4);
```
mxdvl, line 248: Neat! 😉

mxdvl (approved):
Lovely attention to detail! 🔍 

Left a small comment about Emotion for the sole purpose of knowledge sharing.

PR #6256, opened 2022-10-21
fix(source): revert package bump

### LINKED ISSUE #6255: Underlines everywhere
Since #6178 we have underline on all links, and it’s too much!

Remediation:
- ~revert PR~ we will attempt to fix-forward
- capture visual regression in discussions
- bump packages in discussion rendering

mxdvl:
## What does this change?

Revert Source bump

## Why?

The new packages introduce underlines everywhere. See screenshots.

Fixes #6255 

## Screenshots

image

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -1,11 +1,11 @@
 <script>
-	document.cookie = 'gu-cmp-disabled=true';
+    document.cookie = 'gu-cmp-disabled=true';
 </script>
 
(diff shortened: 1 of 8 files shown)


AshCorr (approved):
Looks good, would be good to see if we can fix without reverting though.

mxdvl:
We’ve identified this as P2 instead of P1 and will opt for a fix forward instead.

PR #6258, opened 2022-10-21
Fix `subdued` on buttons adding underlines

AshCorr:
## What does this change?

A source version bump removed the subdued prop on Buttons which controlled underlines.

https://github.com/guardian/source/pull/1437 They were removed for a good reason (a11y) so even though we're re-removing the underlines in this PR we probably want to revisit these designs at some point @HarryFischer

### DIFF
--- a/dotcom-rendering/src/web/components/Callout/Form.tsx
+++ b/dotcom-rendering/src/web/components/Callout/Form.tsx
@@ -175,10 +175,12 @@ export const Form = ({ onSubmit, formFields, error }: FormProps) => {
 					`}
 				>
 					<Link
-						subdued={true}
 						priority="secondary"
 						target="_blank"
 						href="https://www.theguardian.com/help/terms-of-service"
+						cssOverrides={css`
+							text-decoration: none;
+						`}
 					>
 						Terms and conditions
 					</Link>
--- a/dotcom-rendering/src/web/components/ShowHideButton.tsx
+++ b/dotcom-rendering/src/web/components/ShowHideButton.tsx
@@ -19,6 +19,8 @@ const showHideButtonCss = (
 	position: relative;
 	align-items: bottom;
 
+	text-decoration: none;
+
 	${from.wide} {
 		position: absolute;
 		top: 0;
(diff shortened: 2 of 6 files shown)


mxdvl (approved):
A good, albeit temporary, fix!

AshCorr:
I do think that there is definitely room to add underlines to some places, this diff in particular is quite striking https://www.chromatic.com/test?appId=5dfcbf3012392c0020e7140b&id=63527be765c9bf9e37b90f2d as it seems like the "Terms and Conditions" button really isn't highlighted enough.

But something for design to review perhaps

dotcom-rendering/src/web/components/Callout/Form.tsx
```diff
@@ -175,10 +175,12 @@ export const Form = ({ onSubmit, formFields, error }: FormProps) => {
 					`}
 				>
 					<Link
-						subdued={true}
 						priority="secondary"
 						target="_blank"
 						href="https://www.theguardian.com/help/terms-of-service"
+						cssOverrides={css`
+							text-decoration: none;
+						`}
```
mxdvl, line 183: I think we can remove this one

dotcom-rendering/src/web/components/ShowHideButton.tsx
```diff
@@ -19,6 +19,8 @@ const showHideButtonCss = (
 	position: relative;
 	align-items: bottom;
 
+	text-decoration: none;
+
```
mxdvl, line 23: I think we can remove this one

dotcom-rendering/src/web/components/Treats.tsx
```diff
@@ -35,6 +35,7 @@ const TextTreat = ({
 			subdued={true}
 			cssOverrides={css`
 				${textSans.xsmall()}
+				text-decoration: none;
```
mxdvl, line 38: We should discuss this one

dotcom-rendering/src/web/layouts/NewsletterSignupLayout.tsx
```diff
@@ -124,6 +124,7 @@ const previewCaptionStyle = css`
 	background-color: ${brandAlt[400]};
 	padding: ${space[1]}px ${space[3]}px;
 	${textSans.medium({ fontWeight: 'bold' })};
+	text-decoration: none;
```
mxdvl, line 127: I think we can remove this one

mxdvl (commented):
All but the Treats I think we can remove, actually!

PR #6262, opened 2022-10-21
Show more island

### LINKED ISSUE #4604: `Show More`

### LINKED ISSUE #4737: Standardise the spinning animation on the '+ More Opinion' button to align with other loading buttons styles in Source

Georges-GNM:
Co-authored by: @bryophyta
Co-authored by: @AshCorr  

## What does this change?

Creates the "show more" island, i.e. when there are lots of cards in a container, this is the button that allows a user to click "show more" and display more cards.

### Error message

The island gets hydrated when the user interacts with the 'Show More' button; the cards get displayed as they currently do in frontend. If the show more logic errors and the page is unable to display more cards, the page will then show an error using [Source's inline error component](https://guardian.github.io/source/?path=/docs/packages-source-react-components-inlineerror--inline-error-default-theme), which leads to a couple of points of disparity between the frontend and DCR error messages:
- There is now the "alert" triangle just before the error message;
- In frontend, the error message will fade and disappear after a few seconds (but remains in the DOM and doesn't retry automatically), whereas the DCR implementation keeps the message on the page keeps displaying the error until the user clicks the button again or the issue is resolved (which will load the expected cards).

### Deduplication

We have also fixed some bugs in the client-side deduplication logic. For more details, see #6181 and #4604.

### Button styling

The button styling has also been changed, which should close #4737.

### Storybook testing

We've added `@storybook/testing-library` in order to be able to simulate a button click in the story for the ShowMore component. This is (as far as we know) a new pattern in this codebase.

| Frontend | DCR |
| ---------|------|
| image| image |
|image|image|
| | Card colour styling differences are due to a separate issue. |

## Why?

Furthers the [fronts migration work](https://docs.google.com/spreadsheets/d/1JzLvT7a_C_Ap3rKxiORyZwmpgz8bkHP0tZkKJ0QjeC8/edit#gid=0), closes the [associated issue](https://github.com/guardian/dotcom-rendering/issues/4604) and gets us closer to running a test!

Closes #4604.

## Demo

Showing first the current functionality in frontend, then in DCR running locally. 

[DCR working](https://user-images.githubusercontent.com/102960844/199550706-dd9bc30a-b841-470a-8268-f6259786723a.mov)

[DCR error](https://user-images.githubusercontent.com/102960844/199550696-afbe9740-4938-4e99-80ef-87f3f70de53c.mov)

## Accessibility

We've made some changes to the Frontend markup which we hope will improve accessibility, especially for screen reader users. However, there is at least one known issue remaining. Because this issue already existed in the Frontend version, and because the solution requires input from UX and potentially also our accessibility partners, we have created a separate ticket for this issue: #6343

### DIFF
--- a/dotcom-rendering/src/model/enhanceCollections.ts
+++ b/dotcom-rendering/src/model/enhanceCollections.ts
@@ -45,7 +45,7 @@ export const enhanceCollections = (
 			config: {
 				showDateHeader: collection.config.showDateHeader,
 			},
-			hasMore,
+			hasMore: hasMore && !collection.config.hideShowMore,
 		};
 	});
 };
--- a/dotcom-rendering/src/model/enhanceCollections.ts
+++ b/dotcom-rendering/src/model/enhanceCollections.ts
@@ -45,7 +45,7 @@ export const enhanceCollections = (
 			config: {
 				showDateHeader: collection.config.showDateHeader,
 			},
-			hasMore: hasMore && !collection.config.hideShowMore,
+			canShowMore: hasMore && !collection.config.hideShowMore,
 		};
 	});
 };
--- a/dotcom-rendering/src/model/enhanceCollections.ts
+++ b/dotcom-rendering/src/model/enhanceCollections.ts
@@ -19,7 +19,7 @@ export const enhanceCollections = (
 	pageId: string,
 ): DCRCollectionType[] => {
 	return collections.filter(isSupported).map((collection) => {
-		const { id, displayName, collectionType } = collection;
+		const { id, displayName, collectionType, hasMore } = collection;
 		const containerPalette = decideContainerPalette(
 			collection.config.metadata?.map((meta) => meta.type),
 		);
@@ -45,6 +45,7 @@ export const enhanceCollections = (
 			config: {
 				showDateHeader: collection.config.showDateHeader,
 			},
+			hasMore,
 		};
 	});
 };
--- a/dotcom-rendering/src/web/components/ShowMore.importable.tsx
+++ b/dotcom-rendering/src/web/components/ShowMore.importable.tsx
@@ -1,4 +1,5 @@
 import { css } from '@emotion/react';
+import { isUndefined } from '@guardian/libs';
 import { from, space } from '@guardian/source-foundations';
 import { Button, SvgMinus, SvgPlus } from '@guardian/source-react-components';
 import { useState } from 'react';
@@ -43,9 +44,13 @@ export const ShowMore = ({ containerTitle, path, containerId }: Props) => {
 		);
 	}
 
+	const [cardLinks, setCardLinks] = useState<string[]>([]);
+
 	useOnce(() => {
-		const cardLinks = findCardLinks();
-		console.log(cardLinks);
+		const linksArray: string[] = findCardLinks().filter(
+			(item): item is string => !!item,
+		);
+		setCardLinks(linksArray);
 	}, [containerId]);
 
 	const [isOpen, setIsOpen] = useState(false);
@@ -63,7 +68,9 @@ export const ShowMore = ({ containerTitle, path, containerId }: Props) => {
 		: undefined;
 	const { data, loading } = useApi<FEFrontCard[]>(url);
 
-	//data = data?.filter(cardLinks
+	const filteredData =
+		data &&
+		enhanceCards(data).filter((card) => !cardLinks.includes(card.url));
 
 	/**
 		@todo: Semantics for the button and the new container (sub-container?)
@@ -71,15 +78,14 @@ export const ShowMore = ({ containerTitle, path, containerId }: Props) => {
 		@todo: Appropriate semantics for indicating state (aria-live, aria-expanded etc.)?
 		@todo: can we avoid hard-coding the API URL?
 		@todo: handle errors
-		@todo: deduping
-			@todo: get visibleCardIds from DOM instead of as a prop
 		@todo: rename props to match Front type?
 		@todo: make sure that the new content doesn't shift the top of the viewport when it loads
+		@todo: make sure the new cards that get loaded are layed out properly (not just as a list)
 	*/
 
 	return (
 		<>
-			{data && (
+			{filteredData && (
 				<>
 					<div
 						css={css`
@@ -87,7 +93,7 @@ export const ShowMore = ({ containerTitle, path, containerId }: Props) => {
 						`}
 					/>
 					<UL>
-						{enhanceCards(data).map((trail) => (
+						{filteredData.map((trail) => (
 							<LI key={trail.url} padSides={true}>
 								<FrontCard trail={trail} imageUrl={undefined} />
 							</LI>
--- a/dotcom-rendering/src/web/components/ShowMore.importable.tsx
+++ b/dotcom-rendering/src/web/components/ShowMore.importable.tsx
@@ -40,6 +40,7 @@ type Props = {
 	path: string;
 	containerId: string;
 	showAge: boolean;
+	baseUrl: string;
 	containerPalette?: DCRContainerPalette;
 };
 
@@ -48,13 +49,11 @@ export const ShowMore = ({
 	path,
 	containerId,
 	showAge,
+	baseUrl,
 	containerPalette,
 }: Props) => {
 	const [existingCardLinks, setExistingCardLinks] = useState<string[]>([]);
 	const [isOpen, setIsOpen] = useState(false);
-	const [baseUrl, setBaseUrl] = useState(
-		'https://api.nextgen.guardianapps.co.uk',
-	);
 
 	/**
 		@todo: Fix focus behaviour on expand/collapse: @see https://github.com/guardian/dotcom-rendering/issues/6343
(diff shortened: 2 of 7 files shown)


dotcom-rendering/src/web/components/ShowMore.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+const containerTitle = 'Opinion';
+const path = 'uk/lifestyle';
+const containerId = '5011-3940-8793-33a9';
+
+export default {
+	component: ShowMore,
+	title: 'Components/ShowMore',
+};
+
+export const ShowMoreSuccess = () => {
```
bryophyta, line 14: Currently these stories will only show the success/error states if a user clicks on them. So they're useful for documentation purposes but not so useful for Chromatic.

I don't know if we have a good way of showing the component in an 'open' state in Storybook, or whether we should wait until we've decided how we want to use Percy?

dotcom-rendering/src/web/components/ShowMore.importable.tsx
```diff
@@ -0,0 +1,200 @@
...
+		return containerLinksArray.map(
+			(Element) => Element.attributes.getNamedItem('href')?.value,
+		);
+	}
+
+	const [cardLinks, setCardLinks] = useState<string[]>([]);
+
+	useOnce(() => {
+		const linksArray: string[] = findCardLinks().filter(
+			(item): item is string => !!item,
```
bryophyta, line 67: are we happy with inlined type guards like this, from a readability perspective? the purpose here is just to tell tsc something that it should already know but which it doesn't because of some quirks around filtering.

dotcom-rendering/src/web/components/ShowMore.importable.tsx
```diff
@@ -0,0 +1,200 @@
...
+		? `https://api.nextgen.guardianapps.co.uk/${path}/show-more/${containerId}.json?dcr=true`
+		: undefined;
+	const { data, error, loading } = useApi<FEFrontCard[]>(url);
+
+	const filteredData =
+		data &&
+		enhanceCards(data).filter((card) => !cardLinks.includes(card.url));
+
+	/**
+		@todo: Desired focus behaviour on expand/collapse?
```
bryophyta, line 92: there's an issue open for this todo. The rest seem like they can be tackled optionally/at a later date, too.

The content shifting behaviour is something that Frontend seems to have fixed, but which we aren't sure how to address..

dotcom-rendering/src/web/components/ShowMore.stories.tsx
AshCorr: Alternatively you could pass the initial isOpen state as a prop and default it to `false`

dotcom-rendering/src/web/components/ShowMore.importable.tsx
```diff
@@ -0,0 +1,200 @@
...
+	function toggleOpen() {
+		setIsOpen(!isOpen);
+	}
+
+	/** We only pass an actual URL to SWR when 'showMore' is true.
+	 * Toggling 'showMore' will trigger a re-render
+	 *   @see https://swr.vercel.app/docs/conditional-fetching#conditional
+	 */
+	const url = isOpen
+		? `https://api.nextgen.guardianapps.co.uk/${path}/show-more/${containerId}.json?dcr=true`
```
bryophyta, line 73: I don't think should be a blocker, but if anyone knows of an easy way to avoid hard-coding this URL then that could be good!

dotcom-rendering/src/web/components/ShowMore.importable.tsx
```diff
@@ -0,0 +1,188 @@
...
+	path,
+	containerId,
+	showAge,
+	containerPalette,
+}: Props) => {
+	const [existingCardLinks, setExistingCardLinks] = useState<string[]>([]);
+	const [isOpen, setIsOpen] = useState(false);
+
+	/**
+		@todo: Fix focus behaviour on expand/collapse: @see https://github.com/guardian/dotcom-rendering/issues/6343
```
bryophyta, line 57: there's an issue for this; also including it as an inline todo to aid discovery

dotcom-rendering/src/web/components/ShowMore.stories.tsx
```diff
@@ -0,0 +1,66 @@
+import { userEvent, within } from '@storybook/testing-library';
+import fetchMock from 'fetch-mock';
+import { trails } from '../../../fixtures/manual/show-more-trails';
+import { ShowMore } from './ShowMore.importable';
+
+/**
+ * Clicks the 'show more' button so that Chromatic can capture it the component
+ * in its 'open' state.
+ */
+const play = ({ canvasElement }: { canvasElement: HTMLElement }) => {
```
bryophyta, line 10: shout out to @AshCorr for suggesting this!

Given that it required installing a new package (`@storybook/testing-library`) I'm guessing we don't use this pattern already. It seems to work fine with Chromatic, but if people have reservations about introducing  it here then I'm happy to split it off into a  separate PR, as it's not part of the core of this particular issue and does have  broader  implications.

mxdvl: I think this is a very nice feature, but in the spirit of keeping PRs as atomic as possible it would be better to extract this change into its own.

dotcom-rendering/src/web/components/ShowMore.importable.tsx
AshCorr: I don't think its too problematic to have the URL here, but ideally we'd check if we were on a CODE front and switch to the CODE API domain instead of it always being prod

mxdvl: I think there's an apiUrl in the config?

bryophyta: that could be useful! do you know where it is @mxdvl? I haven't been able to find one. I was thinking of pulling the stage from `window.guardian.config.stage` and switching on that, but if the actual URL is available somewhere then that would be great 👍

dotcom-rendering/src/web/components/ShowMore.stories.tsx
bryophyta: I've removed the commit from this branch, and put it on a new one; I'll open a PR once this is merged into `main`:
https://github.com/guardian/dotcom-rendering/compare/ap/show-more-island...pf/story-interactions-for-show-more?expand=1

dotcom-rendering/src/web/components/ShowMore.importable.tsx
bryophyta: our current type definitions seem to assume that `window.guardian.config.page.ajaxUrl` will always return a string, in which case we wouldn't need to provide a default hard-coded value (here I'm storing it in state because afaik `window` access within islands needs to happen within `useOnce`). And we also don't need to check whether it does return a string at runtime before we use it.

I don't know too much about the `window.guardian` setup though -- do those seem like reasonable assumptions or should we be building in more safeguards when using it? cc. @AshCorr & @mxdvl 

this is the sort of thing I've got at the moment, and the type checker seems happy with it:

```js
	const [baseUrl, setBaseUrl] = useState(
		'https://api.nextgen.guardianapps.co.uk',
	);

	useOnce(() => {
                 // snip
		setBaseUrl(window.guardian.config.page.ajaxUrl);
	}, []);

	// snip
	const url = isOpen
		? `${baseUrl}/${path}/show-more/${containerId}.json?dcr=true`
		: undefined;
```

mxdvl: This is coming through as a prop on the object, so you also use it as a prop!

bryophyta: oh great, that's a much better solution! thanks!

bryophyta (approved):
We paired on this and it looks good to me 👍

AshCorr (approved):
Also paired on this, looks good!

dotcom-rendering/src/model/enhanceCollections.ts
```diff
@@ -45,6 +45,7 @@ export const enhanceCollections = (
 			config: {
 				showDateHeader: collection.config.showDateHeader,
 			},
+			hasMore,
```
AshCorr, line 48: I think additionally we'll want to check if the container config has hideShowMore enabled, and hide show more if it is!

image

bryophyta: ooh, good call! For some reason I thought that Frontend was looking at `hideShowMore` when it was deciding whether to set `hasMore`, but I guess not!

AshCorr: So did I! I only noticed whilst reviewing #6392 !

PR #6265, opened 2022-10-24
fix: Don't enhance H2s if theres a Table Of Contents

AshCorr:
## What does this change?

Fixes Table Of Content atoms by only enhancing H2's if there is no table of contents on the page.

## Why?

Editorial like having functional table of contents 😄


oliverlloyd (approved):
I paired on this but the changes look sound to me on review and they tested well so I'm happy to see this one go in

👍

PR #6136, opened 2022-10-05
Add AB test to test two-stage test bucketing

vlbee:
## What does this change?

Introduces an AB test to test two-stage bucketing in isolation. 

Frontend AB Test Switch Code: https://github.com/guardian/frontend/pull/25553

We require an implementation of two-stage bucketing for the long-running mandatory sign in gate trial.

#### How it should work
- During a bucketing window, new browsers are added to the test bucket based on their mvt_id cookie range but we also set a local storage key for the ab test under `"gu.ab.participations"` . 
- After the bucketing window, we stop setting the local storage key and check for its presence instead.  The ab test variant _should only run_ if the correct ab test is set in `"gu.ab.participations"`. This means that new browsers falling within the mvt_id bucket range should _**not**_ appear in the test, becuase the local storage key will not be set for them. 

#### How it is implemented
- This is implemented by a function called 'setOrUseParticipations'[ here](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/experiments/lib/ab-exclusions.ts)
- This function can be used within an AB Test Definition's Variant's `canRun` function 
- The bucketing is controlled by a boolean flag passed into  'setOrUseParticipations'

#### Data expectations

We expect to see new, unique browsers (represented by BrowserIds in the pageview table) entering the test and seeing the gate during the bucketing phase.

After the bucketing phase (running phase), we expect:
- Continued ab test component events firing for the unqiue browser ids from the bucketing phase. 
- No new browser ids bucketed in the test after the bucketing phase. 
- No ab test component events fired for NEW browser ids 

## Why?

We did not observe the expected data in the[ previous test run](https://github.com/guardian/dotcom-rendering/pull/5891), so are testing the local storage participation bucketing in isolation over a 48hr period. 

24hrs with flag set to true (bucketing)
24hrs with flag set to false (no bucketing)

## Local Testing Observations

**1st Set up - Ensure the setParticipationsFlag is true**
- Run branch - `make watch`
- Frontend PR - set the test's AB switch to "ON".
- Run Frontend PR client - `make dev` 
- Run Frontend PR server side `./sbt  project article run` 

1. Open new, clean browser window. 
2. Navigate to local Article page. Accept CMP cookies. 
3. Set 'GU_mvt_id' cookie in range eg. 897502
4. Refresh page. 
 
|   |   Expectation      | 
|----------|:-------------:|
| ab.participation |  newly set | 
| GU_mvt_id |   Set   | 
[...]

PR #6280, opened 2022-10-25
Fix typo

simonbyford:
## What does this change?

Changes the AB test participations localstorage key in a spec from: 

`'gu.ab_participations'` 

to  

`'gu.ab._participations'`

## Why?

It's `'gu.ab._participations'` everywhere else so I believe this was a typo.

### DIFF
--- a/dotcom-rendering/src/web/experiments/lib/ab-exclusions.test.ts
+++ b/dotcom-rendering/src/web/experiments/lib/ab-exclusions.test.ts
@@ -26,7 +26,7 @@ describe('canRun using participations', () => {
 	});
 
 	test('canRun using participations returns true if setParticipationsFlag is true and participation already set', () => {
-		storage.local.set('gu.ab_participations', currentTestParticipation);
+		storage.local.set('gu.ab._participations', currentTestParticipation);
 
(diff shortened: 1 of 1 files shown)


OllysCoding:
Should this have caused the test to fail?

oliverlloyd (commented):
Is there a use case where existing state stored in local storage needs to be used and if we change the expected name for that property do we risk losing access to this previous information?

mxdvl (changes requested):
I think @vlbee has the most context for introducing this in #6136.
Please double-check that this will not break ongoing tests before merging.

simonbyford:
> Is there a use case where existing state stored in local storage needs to be used and if we change the expected name for that property do we risk losing access to this previous information?

This change only affects a test file, so we should be ok

mxdvl (approved):

simonbyford:
> Should this have caused the test to fail?

Since the wrong key was being used, the test essentially became a clone of a [separate test](https://github.com/guardian/dotcom-rendering/blob/30b576ed40c191c885c01f1d8c7ae572e7a388e7/dotcom-rendering/src/web/experiments/lib/ab-exclusions.test.ts#L21) where participations are _not_ set. So it "accidentally" passed.

Now it still passes with the correct setup.

PR #6281, opened 2022-10-25
Better page titles

### LINKED ISSUE #5748: Render correct `<title>` for network fronts
DCR currently passes the top-level `.webTitle` property as the `<title>` for Fronts. This works for many cases, but for some (e.g. the `/uk` homepage) `.webTitle` is 'Network Front', which is not the right value to use.

Identified discrepancies

|url|Frontend title|DCR title|
|-|-|-|
|https://www.theguardian.com/uk|`News, sport and opinion from the Guardian's UK edition \ The Guardian`|`Network Front`|
|https://www.theguardian.com/us?INTCMP=CE_US|`News, sport and opinion from the Guardian's US edition \ The Guardian`|`Network Front`|
|others?|||

I haven't been able to find prior mention of this issue.

@oliverlloyd I feel like you might have some insight here -- is this something that should be resolved in the fronts tool, or do  we need to e.g. hard-code a mapping from url to title in DCR for these  special cases?

oliverlloyd:
## What?
This PR improves the [page titles](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) we use for DCR fronts. Fixes #5748 

There are two properties that can provide the title text, `pressedPage.seoData.title` and `pressedPage.seoData.webTitle`. 
`title` was not part of the DCR model so it has now been added and, when present, we now favour this value as it provides a longer, more descriptive indication of the page contents. If `title` is not available, we fallback to `webTitle`

In addition, we now append '| The Guardian' to the string we use.

## Why?
Having a descriptive title improves our SEO and offers readers better looking results when using search engines.

### Why append 'The Guardian'?
The text provided for the page title lacks the context of the fact it is a _Guardian_ page so by adding '| The Guardian' we surface this information in search results better.

### DIFF
--- a/dotcom-rendering/src/model/front-schema.json
+++ b/dotcom-rendering/src/model/front-schema.json
@@ -411,6 +411,9 @@
                         "webTitle": {
                             "type": "string"
                         },
+                        "title": {
+                            "type": "string"
+                        },
                         "description": {
                             "type": "string"
                         }
--- a/dotcom-rendering/src/types/front.ts
+++ b/dotcom-rendering/src/types/front.ts
@@ -415,6 +415,7 @@ type FESeoDataType = {
 	id: string;
 	navSection: string;
 	webTitle: string;
+	title?: string;
 	description: string;
 };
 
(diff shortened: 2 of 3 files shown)


OllysCoding (approved):
+1 This has been on my mind for a while, looks good!

PR #6282, opened 2022-10-25
Remove shady pie test and labs shady pie

emma-imber:
- Removes the shady pie test
- Removes the labs version of the shady pie, due to the decision not to use it having looked at the test results


oliverlloyd (approved):
LGTM!

PR #6285, opened 2022-10-26
Use @guardian-browserlist-config

mxdvl:
## What does this change?

Use `@guardian-browserlist-config` for our modern build.

## Why?

We ran a test in #5899 & #6168  on 1% of our users and our data analyst has confirmed that it has had no negative impact on our core web vitals or engagement.


jamesgorrie (approved):
We were talking about co-location earlier, I wonder if there's a way to put this all in one place?

mxdvl:
> We were talking about co-location earlier, I wonder if there's a way to put **this** all in one place?

What do you mean by “this”, here, @jamesgorrie?

Issue #4060, opened 2022-02-23
Don't tell amp we can serve an amp page when we can't


jamesgorrie:
from: Dotcom team
via: standup
on: 23 Feb 2022

More info coming soon

OllysCoding:
@OllysCoding to add more info

VDuczekW:
refactor into better ticket

PR #6287, opened 2022-10-26
Only include amp tag if we support amp

Georges-GNM:
Co-Authored-By: @AshCorr
Co-Authored-By: @OllysCoding 

## What does this change?

This specifies which elements are supported by AMP in DCR and should when AMP isn't supported by DCR (e.g. interactives, quizzes, liveblogs and if there is a formstack form on the page. This should mean that a page only has the amp tag when it is actually supported. 

## Why?

Related to the following issues: #4060, #5931 & [#25566](https://github.com/guardian/frontend/issues/25566). There was a suggestion of adding a redirect to the website if an unsupported amp page is trying to be loaded, but this seems like a sufficient solution for the moment.

## Tracking
https://metrics.gutools.co.uk/d/QuBd4eH4k/amp-metrics


dotcom-rendering/src/amp/components/Elements.tsx
```diff
@@ -21,6 +22,79 @@ import { VideoYoutubeBlockComponent } from './elements/VideoYoutubeBlockComponen
...
+	main: string,
+): boolean => {
+	if (format.design === 'InteractiveDesign') {
+		const hasAmpInteractiveTag = tags.some(
+			(tag) => tag.id === 'tracking/platformfunctional/ampinteractive',
+		);
+		if (!hasAmpInteractiveTag) return false;
+	}
+
+	if (format.design === 'ArticleDesign') {
```
AshCorr, line 75: Hmm, sorry I've been bouncing back and forth on this, but I think we do actually want to do something like
`tags.includes('type/article')` instead of `format.design === 'ArticleDesign'` to deal with all the different  types of articles such as Analysis, Explainers, etc

OllysCoding:
I wonder if we should also be using the `isAmpSupported` at the entrypoint to rendering an AMP article? Rather than relying on code elsewhere failing - this way it will force us / future engineers to keep this function up-to-date

AshCorr (approved):
I paired on this, but it looks good!

https://github.com/guardian/frontend/pull/25616 this PR needs to be merged down first.

dotcom-rendering/src/amp/components/Elements.tsx
Georges-GNM: Now updated!

dotcom-rendering/src/web/server/articleToHtml.tsx
```diff
@@ -205,12 +206,16 @@ export const articleToHtml = ({ article: CAPIArticle }: Props): string => {
...
 
 	const getAmpLink = (tags: TagType[]) => {
-		if (CAPIArticle.format.design === 'InteractiveDesign') {
-			const hasAmpInteractiveTag = tags.some(
-				(tag) =>
-					tag.id === 'tracking/platformfunctional/ampinteractive',
-			);
-			if (!hasAmpInteractiveTag) return undefined;
+		if (
+			!isAmpSupported({
```
mxdvl, line 210: One method to handle them all.

mxdvl (approved):

PR #6291, opened 2022-10-26
Capture CWV & Commercial metrics on fronts

mxdvl:
## What does this change?

Always capture Core Web Vitals and Commercial metrics when a user is in the DCR fronts test.

## Why?

If a user is in the test variant, we want to capture these metrics all the time, as there is only a small number of users that are currently in the test–it’s opt-in only.

We want to be able to get a throrough reporting, jointly with @guardian/commercial-dev.


jamesgorrie (approved):
Looks great!

PR #6267, opened 2022-10-24
Handle static guim domain images

mxdvl:
## What does this change?

Handle static domain for images

## Why?

Some older articles are not displaying the proper carousel images

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/197539098-f92c8622-78da-4a36-a4e9-091ecb78ba52.png
[after]: https://user-images.githubusercontent.com/76776/197539032-a2197157-1018-4935-9c4a-33e69576e09f.png

PR #6307, opened 2022-10-28
Use 'masterImage' for onwards trails

OllysCoding:
## What does this change?

In a recent PR (https://github.com/guardian/frontend/pull/25610) we added a 'masterImage' property to the API response for onwards trails.

> 🚨🚨🚨 Some trails where we would previously would have received a 'sized' image from Frontend no longer show an image due to them not having a 'master image' - this might be something we want to look at, but if we can we should try and fix this in Frontend, before merging this PR

## Why?

This enables us to remove the hack added in https://github.com/guardian/dotcom-rendering/pull/6267 and https://github.com/guardian/dotcom-rendering/pull/6065, as we can use the master image directly

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |
| ![before1][] | ![after1][] |

[before]: https://user-images.githubusercontent.com/9575458/198642379-4d2450db-ff42-4187-ab47-2c80cc6e1221.png
[after]: https://user-images.githubusercontent.com/9575458/198642572-c43dd2a9-8580-4f59-b957-1ef410f4d749.png
[before1]: https://user-images.githubusercontent.com/9575458/198643081-36901d56-4967-4a82-aeb8-f4ddd86d7fb3.png
[after1]: https://user-images.githubusercontent.com/9575458/198643186-0c51ff58-f4b8-4559-a980-b13e687c6591.png


mxdvl (commented):
I think we need to flag this quite broadly before merging, as it’s quite a big change.

The code is so much cleaner/clearer, though!

PR #6321, opened 2022-10-31
SDC 1.0.6

tomrf1:
## What does this change?
Updates the support-dotcom-components dependency to 1.0.6

## Why?
This version no longer requires the `edition` field for header links targeting. (See https://github.com/guardian/support-dotcom-components/pull/779)
This is good because we're adding a new EUR edition.

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |


OllysCoding (approved):
+1

PR #6333, opened 2022-11-01
Add element type Callout to AR model

abeddow91:
## What does this change?
Adds the new Callout element type to AR model and `bumps apps-rendering-api-models` and `content-api-models`. It also prevents embed callouts from being parsed as these are not currently supported by the platform and allows us to update the existing Callout type, rather than adding a new variant.


apps-rendering/package.json
```diff
@@ -36,11 +36,11 @@
     "@babel/preset-env": "^7.15.6",
     "@creditkarma/thrift-server-core": "^1.0.4",
     "@emotion/jest": "^11.10.0",
-    "@guardian/apps-rendering-api-models": "^1.1.1",
+    "@guardian/apps-rendering-api-models": "^0.0.0-2022-11-01-SNAPSHOT",
```
georgeblahblah, line 39: just adding a note here that we need to remember to release non-snapshot versions of these libraries before merging this! I've accidentally merged snapshot versions before 🙊

georgeblahblah (commented):
Looks great to me! I'm happy to take another look when it comes out of Draft

apps-rendering/src/bodyElement.ts
```diff
@@ -335,41 +335,52 @@ const parse =
...
-					if (!campaign) {
-						return Result.err('No matching campaign');
-					}
-
-					const description = context.docParser(
-						campaign.fields.description ?? '',
+				if (isCallout)
+					return Result.err(
+						'Embed callouts are deprecated in Apps Rendering. Please use element type Callout',
 					);
```
JamieB-gu, line 383:
```suggestion
				if (isCallout) {
					return Result.err(
						'Embed callouts are deprecated in Apps Rendering. Please use element type Callout',
					);
				}
```

JamieB-gu (commented):
Nice! This approach seems sensible to me 👍

apps-rendering/package.json
JamieB-gu: Just a reminder that dependencies are managed differently since #6339. When you come to update these you may need to merge/rebase `main` and then use `yarn` to lock the changes.

apps-rendering/src/bodyElement.ts
```diff
@@ -335,41 +335,52 @@ const parse =
...
+			case ElementType.CALLOUT: {
+				const {
+					campaignId: campaignId,
+					isNonCollapsible: isNonCollapsible,
+				} = element.calloutTypeData ?? {};
+				if (!campaignId || !isNonCollapsible) {
+					return Result.err(
+						'No campaignId or isNonCollapsible field on calloutTypeData',
+					);
+				}
```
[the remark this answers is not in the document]
abeddow91: Yep you're right - thanks for the suggestion, I'll commit this

apps-rendering/package.json
abeddow91: Perfect, have done!

georgeblahblah (commented):
Looks good! We are waiting on a round of testing on CODE before merging

apps-rendering/src/campaign.ts
```diff
@@ -11,23 +11,17 @@ const getCallout = (
...
-		campaign.fields.kind === 'callout' &&
-		campaign.fields.callout.tagName === id
-	) {
+	const campaign = campaigns.find((campaign) => campaign.id === id);
+
+	if (campaign?.fields.kind === 'callout') {
 		return Optional.some(campaign.fields.callout);
 	}
 
-	return getCallout(id, tail);
```
abeddow91, line 30: As we are using the campaign id now, we can search the campaigns for the specific campaign and no longer need this to be recursive

sophie-macmillan (approved):

georgeblahblah (approved):
Looks great! Thanks for this!

PR #6335, opened 2022-11-01
Add SpecialReportAltPalette

ioannakok:
## What does this change?
Adds `SpecialReportAltPalette`. DCR should be able to consume the [change in the fronts API](https://github.com/guardian/facia-scala-client/pull/281). 

## Why?
To match [designs](https://www.figma.com/file/agVeiMZULSWlf4JxNgkng4/Fronts-Palettes?node-id=1%3A2)

## Screenshots
image

Please ignore the wrong sublinks kicker colours. There is a [ticket](https://github.com/guardian/dotcom-rendering/issues/6336) to fix this more general issue.

### DIFF
--- a/dotcom-rendering/src/web/lib/decideContainerOverrides.ts
+++ b/dotcom-rendering/src/web/lib/decideContainerOverrides.ts
@@ -1,5 +1,7 @@
 import type { DCRContainerPalette } from '../../types/front';
 import type { ContainerOverrides } from '../../types/palette';
+import { transparentColour } from './transparentColour';
+import { neutral } from '@guardian/source-foundations';
 
 const textCardHeadline = (containerPalette: DCRContainerPalette): string => {
 	switch (containerPalette) {
@@ -19,6 +21,8 @@ const textCardHeadline = (containerPalette: DCRContainerPalette): string => {
 			return '#041F4A';
 		case 'EventAltPalette':
 			return '#041F4A';
+		case 'SpecialReportAltPalette':
+			return '#2b2b2a';
 	}
 };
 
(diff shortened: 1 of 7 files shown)


dotcom-rendering/src/web/lib/decideContainerOverrides.ts
```diff
@@ -19,6 +21,8 @@ const textCardHeadline = (containerPalette: DCRContainerPalette): string => {
 			return '#041F4A';
 		case 'EventAltPalette':
 			return '#041F4A';
+		case 'SpecialReportAltPalette':
+			return '#2b2b2a';
```
ioannakok, line 25: I went with hex here for consistency but wondered why we're not using the source palette references?

ioannakok:
I'm getting some false positive Chromatic diffs for the `VimeoComponent` which I think it's okay to accept but also some `KeyEvents` diffs unrelated to my changes

HarryFischer:
@ioannakok Please can we remove the bold headlines and just have them as normal (Bold kicker, medium headline) ?

dotcom-rendering/src/web/lib/decideContainerOverrides.ts
mxdvl: I believe it’s because there is no source value for that specific colour.

mxdvl (approved):

HarryFischer:
> > @ioannakok Please can we remove the bold headlines and just have them as normal (Bold kicker, medium headline) ?
> 
> Done - looks a lot better indeed! Sorry, it was in the designs and missed that.

Looks good now - thanks

HarryFischer (approved):

dotcom-rendering/src/web/lib/decideContainerOverrides.ts
ioannakok: I am keen to attempt replacing these hex values with palette references at some point but indeed I don't know if they all exist in source off the top of my head. Will give it a go at some point and we'll see how that goes!

